PR #2090, opened 2016-09-01
fixed Thing comparison to consider labels and locations

### LINKED ISSUE #2089: Thing label changes are not considered
When changing a label of a Thing in the DSL, this change is not reflected while the framework is running.

sjka:
So far, Things where those fields changed have been considered to
be equal and therefore were not updated in the registry.

fixes #2089

### DIFF
--- a/bundles/core/org.eclipse.smarthome.core.thing/src/main/java/org/eclipse/smarthome/core/thing/util/ThingHelper.java
+++ b/bundles/core/org.eclipse.smarthome.core.thing/src/main/java/org/eclipse/smarthome/core/thing/util/ThingHelper.java
@@ -26,6 +26,7 @@ import org.eclipse.smarthome.core.thing.internal.BridgeImpl;
 import org.eclipse.smarthome.core.thing.internal.ThingImpl;
 
 import com.google.common.base.Joiner;
+import com.google.common.base.Objects;
 
 /**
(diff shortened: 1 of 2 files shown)


bundles/core/org.eclipse.smarthome.core.thing/src/main/java/org/eclipse/smarthome/core/thing/util/ThingHelper.java
```diff
@@ -66,6 +66,20 @@ public static boolean equals(Thing a, Thing b) {
...
+        if (a.getLabel() != null && !a.getLabel().equals(b.getLabel())) {
+            return false;
+        }
+        // location
+        if (a.getLocation() == null && b.getLocation() != null) {
+            return false;
+        }
+        if (a.getLocation() != null && !a.getLocation().equals(b.getLocation())) {
+            return false;
+        }
```
maggu2810: Using Objects.equals will result in less code and is IMHO better readable.
Missed I the reason you don't use it?
https://docs.oracle.com/javase/7/docs/api/java/util/Objects.html#equals(java.lang.Object,%20java.lang.Object)

sjka:
very valid point, thanks!
I've changed the existing ones too. Which while git tries its best to guess line matches makes the diff a bit confusing to read...

maggu2810:
All fine, thanks

Issue #2049, opened 2016-08-26
Paper UI: rule handling


maggu2810:
If a new rule is added using the Paper UI a post method is done to path /rest/rules and the payload contain the rule.
The response header contains the key "location" to the "path" of the newly created rule.

* The Paper UI does not use this URL to fetch the rule (perhaps a modified one IIRC we have a autoconnection feature) and to add it to its rule storage. Instead it loads the whole list of all rules again.
IMHO this could be improved.

* If a new rule has been added an event is submitted (SSE / eventsource), so the PaperUI is aware of this change. But also in this case the PaperUI ignores the event and does not load this rule. If more then one user control the system using the PaperUI I would expect that a new rule is added to my list without forcing a refresh as in the old days without ajax and all the async stuff.

sjka:
@aounhaider1 could you please have a look?

PR #2093, opened 2016-09-01
Support receiving rules events

aounhaider1:
1) Receive rule events and update local repo accordingly.
2) Rely more on rule repository (local) than rule service (remote).
3) Added Status info of rule.
fixes https://github.com/eclipse/smarthome/issues/2049

### DIFF
--- a/extensions/ui/org.eclipse.smarthome.ui.paper/web-src/js/services.repositories.js
+++ b/extensions/ui/org.eclipse.smarthome.ui.paper/web-src/js/services.repositories.js
@@ -161,8 +161,48 @@ angular.module('PaperUI.services.repositories', []).factory('bindingRepository',
     var repository = new Repository($q, $rootScope, itemService, 'items')
     $rootScope.data.items = [];
     return repository;
-}).factory('ruleRepository', function($q, $rootScope, ruleService) {
-    var repository = new Repository($q, $rootScope, ruleService, 'rules')
+}).factory('ruleRepository', function($q, $rootScope, ruleService, eventService) {
+    var repository = new Repository($q, $rootScope, ruleService, 'rules', true)
     $rootScope.data.rules = [];
+
+    eventService.onEvent('smarthome/rules/*/updated', function(topic, ruleUpdate) {
+
+        var existing = repository.find(function(rule) {
+            return rule.uid === ruleUpdate[0].uid;
+        });
+        if (existing) {
+            existing.name = ruleUpdate[0].name;
+            existing.description = ruleUpdate[0].description;
+            existing.triggers = ruleUpdate[0].triggers;
+            existing.actions = ruleUpdate[0].actions;
+            existing.conditions = ruleUpdate[0].conditions;
+        }
+    });
+
+    eventService.onEvent('smarthome/rules/*/added', function(topic, rule) {
+        repository.add(rule);
+    });
+
+    eventService.onEvent('smarthome/rules/*/removed', function(topic, removedRule) {
+        var existing = repository.find(function(rule) {
+            return rule.uid === removedRule.uid;
+        });
+        repository.remove(existing);
+    });
+
+    eventService.onEvent('smarthome/rules/*/state', function(topic, rule) {
+        var existing = repository.find(function(rule) {
+            return rule.uid === topic.split('/')[2];
+        });
+        existing.status = existing.status ? existing.status : {};
+        existing.status.status = rule.status;
+        existing.status.statusDetails = rule.status;
+        if (rule.status.toUpperCase() === "DISABLED") {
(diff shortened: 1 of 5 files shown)


kaikreuzer:
@maggu2810 I leave this to you to test and merge.

maggu2810:
* I merged this PR locally
* I opened the Rules list in two different browsers
* I added a new rule in browser A
* if I press an the + button in browser B (so the contect menu "new rule" "rule from template" is shown), the list is repainted and I see the newly created rule


So it seems the event is correctly handled and the new rule is added internally.
I would expect that the new rule is also added to the currently shown list (e.g. at the end to not change the order). Why do we need an user interaction to show the new rule?

extensions/ui/org.eclipse.smarthome.ui.paper/web-src/js/services.repositories.js
```diff
@@ -161,8 +161,48 @@ angular.module('PaperUI.services.repositories', []).factory('bindingRepository',
...
+
+    eventService.onEvent('smarthome/rules/*/added', function(topic, rule) {
+        repository.add(rule);
+    });
+
+    eventService.onEvent('smarthome/rules/*/removed', function(topic, removedRule) {
+        var existing = repository.find(function(rule) {
+            return rule.uid === removedRule.uid;
+        });
+        repository.remove(existing);
```
maggu2810: Shouldn't we use `if (existing)` similar to the update event here, too?

extensions/ui/org.eclipse.smarthome.ui.paper/web-src/js/services.repositories.js
```diff
@@ -161,8 +161,48 @@ angular.module('PaperUI.services.repositories', []).factory('bindingRepository',
...
+    });
+
+    eventService.onEvent('smarthome/rules/*/state', function(topic, rule) {
+        var existing = repository.find(function(rule) {
+            return rule.uid === topic.split('/')[2];
+        });
+        existing.status = existing.status ? existing.status : {};
```
maggu2810: `if (existing)`?

extensions/ui/org.eclipse.smarthome.ui.paper/web-src/js/services.repositories.js
```diff
@@ -161,8 +161,48 @@ angular.module('PaperUI.services.repositories', []).factory('bindingRepository',
...
+        existing.status.status = rule.status;
+        existing.status.statusDetails = rule.status;
```
maggu2810: Why set we status.status to rule.status and stats.statusDetails also to rule.status? Shouldn't the details point to the details?

extensions/ui/org.eclipse.smarthome.ui.paper/web-src/js/services.repositories.js
aounhaider1: Not necessarily, I would expect remove function in repository to take care of that.

aounhaider1:
The new rule list is sorted so every time a new rule is added angular insert it on the right position. I think it makes sense to maintain the order however, we can have the same highlight feature as we have in things to make it easier to identify the newly created rules. What do you say?

```Why do we need an user interaction to show the new rule?```
I did not understand what you mean here?

maggu2810:
> (1) I merged this PR locally
> (2) I opened the Rules list in two different browsers
> (3) I added a new rule in browser A
> (4) if I press an the + button in browser B (so the contect menu "new rule" "rule from template" is shown), the list is repainted and I see the newly created rule

See number (4), if I don't press e.g. on the + button the new rule is notshown in browser B. I need to interact with the UI so it gets refreshed.

aounhaider1:
The scope should refresh now when new rule events are received.

extensions/ui/org.eclipse.smarthome.ui.paper/web-src/js/services.repositories.js
aounhaider1: You are right it should be rule.statusDetails.

aounhaider1: Changed it!

extensions/ui/org.eclipse.smarthome.ui.paper/web-src/js/services.repositories.js
aounhaider1: Added if found check on remove function.

maggu2810:
Looks good now, thanks

maggu2810:
Have you finished your changes?

aounhaider1:
Yes.

maggu2810:
Thanks

Issue #658, opened 2015-12-04
Strip down core.scheduler to only contain Quartz classes that are really required


sjka:
migrated from Bugzilla [#483175](https://bugs.eclipse.org/bugs/show_bug.cgi?id=483175)
status NEW severity *normal* in component *Automation* for *---*
Reported in version *unspecified* on platform *PC*
Assigned to: Project Inbox

On 2015-11-27 07:50:31 -0500, Kai Kreuzer wrote:

> In order to shrink the minimal runtime size, the bundle core.scheduler should be revisited to only contain Quartz classes that are required for ESH purposes.
> There are many parts about JDBC job persistence etc. that are definitely not required.

On 2015-11-27 08:41:57 -0500, Simon Kaufmann wrote:

> do you by any chance already know if there is anything to consider regarding IP/license in that case?
> I'm not sure if "stripping down" a jar technically creates some derived work, or if it's considered to be "unmodified" in the sense that the actual code won't be changed...

On 2015-11-27 08:45:51 -0500, Kai Kreuzer wrote:

> I am pretty certain that if you have the CQ approved for some code, you are free to take any subset of it as well. The class files stay "unmodified", it is only a different jar, but this is what is done with many things in Orbit as well by re-packaging the 3rd party content.

sjka:
there is some work in progress to replace quartz with an own implementation, see #869

kaikreuzer:
The new scheduler service has been merged with https://github.com/eclipse/smarthome/pull/1239, so my suggestion is to close this issue and have the timer module type adapted to use the new scheduler instead.

kgoderis:
Yes, indeed

PR #2094, opened 2016-09-01
allow non-model based persistence configuration

maggu2810:
Currently a persistence service could be only configured using the DSL / models files (.persist).
This PR moves the persistence manager logic (the non-model based one) to persistence core and allows a programmatically configuration of persistence services, too. So we could add a REST interface some time.

It is far from complete, but it should at least work.
I used some "stupid" names, starting with "Simple". This has been done to prevent collisions between the names used in the model at development but will be changed later.

Can you have a look at?
@cdjackson As you seems to be a professional developer of the ESH persistence interfaces, could you have a look at, too?
@kaikreuzer The most files are touched has been written by you. Your comments are very welcome.

There is a service for `org.eclipse.smarthome.core.persistence.PersistenceManager` now.
This service provides functions to:
- add persistence service configuration (will also trigger 'start event handling' for this persistence service)
- remove persistence service  configuration (will also trigger 'stop event handling' for this persistence service)
- start event handling for persistence service
- stop event handling for persistence service


bundles/core/org.eclipse.smarthome.core.persistence/META-INF/MANIFEST.MF
```diff
@@ -7,13 +7,20 @@ Bundle-ManifestVersion: 2
 Bundle-License: http://www.eclipse.org/legal/epl-v10.html
 Bundle-SymbolicName: org.eclipse.smarthome.core.persistence
 Import-Package: org.eclipse.smarthome.config.core,
+ org.eclipse.smarthome.core.common.registry,
  org.eclipse.smarthome.core.items,
  org.eclipse.smarthome.core.library.types,
  org.eclipse.smarthome.core.persistence,
  org.eclipse.smarthome.core.types,
+ org.quartz,
```
kaikreuzer: This is a problem. You might remember that we tried to get rid off Quartz in the core bundles (see https://github.com/eclipse/smarthome/issues/658), since it is not compatible with compact2 profile and also is pretty fat. So when overhauling the persistence APIs, I would definitely refrain from using this library.
I would hope that you can use @kgoderis' new scheduler implementation instead.

kaikreuzer:
Thanks @maggu2810, making the persistence stuff independent from Xtext+EMF is definitely an important step (which was far too low on my personal todo list ;-)).

So apart from the Quartz-refactoring, I would like to start the discussion what other refactorings we might want to do in order to adapt the persistence API / configuration possibilities to the needs.
The current setup with the strategies etc. is something I came up with a very long time ago and it hasn't evolved since then. Now is a chance to do changes and enhancements to it, so use the chance!

bundles/core/org.eclipse.smarthome.core.persistence/META-INF/MANIFEST.MF
maggu2810: I know :wink: 
I used quartz just temporary to not change to much code.
I will use the new scheduler implementation (the automation should move its timer trigger, too) after the concept of the split is okay.

maggu2810:
> So apart from the Quartz-refactoring

Done, no Quarts anymore :wink:

kaikreuzer:
> Done, no Quartz for persistence anymore 😉

Cool. Did you also test that it is working? I never tried the new scheduler with cron stuff myself yet...

So what remains is my question in https://github.com/eclipse/smarthome/pull/2094#issuecomment-244128221.
I'd also like to invite other to think about any requirements and wishes here - @cdjackson, @kgoderis, @sjka and others - please join in!

maggu2810:
> Cool. Did you also test that it is working?

I tried this strategy `everySecond : "* * * * * ?"` and this one is working using the ESH scheduler (ExpressionThreadPoolExecutor).

kgoderis:
@maggu2810 @kaikreuzer Damn... 8hours to late ;-) I have this piece of code in the pipe : https://github.com/kgoderis/smarthome/commit/826fc78270c34e1e99578a0ad2704939c0027798

I had already the de-quartzifying on my to do list - the url here above could serve as an example.

kgoderis:
@maggu2810 btw, forget about cron, use RFC 5545 ;-)

maggu2810:
> Damn... 8hours to late ;-) I have this piece of code in the pipe : kgoderis@826fc78

@kgoderis I removed Quartz from the persistence service only.
So, all the remaining usages could be changed by you :wink:

kgoderis:
> So, all the remaining usages could be changed by you 😉

Yeah Yeah - the todo list is growing every day ....

kaikreuzer:
I'd be fine to go ahead with this PR. @maggu2810 Do you still consider it WIP? Or shall we merge it as a working replacement for the current code and discuss further evolution in future?

maggu2810:
I think I need to revert the change to use the ESH scheduler instead of Quartz because cron does not work with the current ESH scheduler implementation as expected (see #2105)

kaikreuzer:
I didn't follow #2105 in detail, but is this something that cannot easily be fixed by @kgoderis and we simply wait for that?

maggu2810:
> but is this something that cannot easily be fixed by @kgoderis

I don't know if it is such easily. But I assume @kgoderis knows.

> and we simply wait for that?

That is the reason why I kept the [WIP] in front of the title.

> Or shall we merge it as a working replacement for the current code and discuss further evolution in future?

I assume it is simple to merge the split of model.persistence now (without the Quartz scheduler removal) and merge the Quartz scheduler removal (the commits could be simple split out of this PR) after the ESH scheduler has been fixed.

kaikreuzer:
> I assume it is simple to merge the split of model.persistence now (without the Quartz scheduler removal) 

Not really, because this step moves the usage of Quartz from model to core. As the model bundles are optional, not everyone has this dependency, but if we move it to core, it will force people to add (the fat) Quartz to their stack. I'd therefore rather would want to wait for a solution of #2105.

maggu2810:
> this step moves the usage of Quartz from model to core

Okay :wink:, missed that point

kgoderis:
> I didn't follow #2105 in detail, but is this something that cannot easily be fixed by @kgoderis and we simply wait for that?

Yoda says: Easy it is not

I have not revisted https://github.com/eclipse/smarthome/issues/2105 since the discussion we had on the subject, as we need to decide on a strategy (see that thread) before moving forward. @kaikreuzer maybe you can voice your thoughts on the matter  if you happen to find some time?

kaikreuzer:
> if you happen to find some time

I won't for the next days, but WHEN I find time, I will try to thing about it and share my thoughts...

kgoderis:
@maggu2810 Pls see PR #2270

maggu2810:
#2270 has been merged, so #2105 should be fixed

bundles/core/org.eclipse.smarthome.core.persistence/src/main/java/org/eclipse/smarthome/core/persistence/internal/PersistenceManagerImpl.java
```diff
@@ -0,0 +1,420 @@
...
+     * Delete all {@link Job}s of the group <code>persistModelName</code>
+     *
+     * @throws SchedulerException if there is an internal Scheduler error.
+     */
+    private void removeTimers(String persistModelName) {
+        if (!persistenceJobs.containsKey(persistModelName)) {
+            return;
+        }
+        for (final Runnable job : persistenceJobs.get(persistModelName)) {
+            boolean success = scheduler.remove(job);
```
kaikreuzer: I just started testing this PR, but in this line success is always false and I get:
```
2016-11-02 15:25:55.180 [WARN ] [o.e.s.c.p.PersistenceManager  :317  ] - Failed to delete cron job for dbId 'rrd4j```
in my log (when updating the file rrd4j.persist).
This is probably an issue in the ESH scheduler implementation, but I'd like to know if this has worked for you? In my tests, the scheduler keeps creating new jobs and never removes them again...

File: bundles/core/org.eclipse.smarthome.core.persistence/src/main/java/org/eclipse/smarthome/core/persistence/internal/PersistenceManagerImpl.java [hunk 3]
```diff
@@ -0,0 +1,420 @@
...
+    private final ExpressionThreadPoolExecutor scheduler;
+
+    private ItemRegistry itemRegistry;
+
+    final Map<String, PersistenceService> persistenceServices = new HashMap<>();
+    final Map<String, PersistenceServiceConfiguration> persistenceServiceConfigs = new HashMap<>();
+    private final Map<String, Set<Runnable>> persistenceJobs = new HashMap<>();
+
+    public PersistenceManagerImpl() {
+        scheduler = ExpressionThreadPoolManager.getExpressionScheduledPool("persist");
```
kaikreuzer: wouldn't it be better to do this in activate() and de-reference and dispose it again in deactivate()?

bundles/core/org.eclipse.smarthome.core.persistence/src/main/java/org/eclipse/smarthome/core/persistence/internal/PersistenceManagerImpl.java
kaikreuzer: I have created https://github.com/eclipse/smarthome/pull/2391/files#diff-6ab8869df39e4b2f961a094a0437b037R34 for further investigation and fixing.

maggu2810: I assume this should be fixed by #2435, now. Correct?

bundles/core/org.eclipse.smarthome.core.persistence/src/main/java/org/eclipse/smarthome/core/persistence/internal/PersistenceManagerImpl.java
maggu2810: Yes, I think so.

Will change it.

maggu2810:
Hm, what are we missing here?

kaikreuzer:
Trust in the new CRON scheduler implementation...
Sorry, I didn't yet do any further tests, but my feeling is that it would be good to test it on a real system running for at least 24 hours.

maggu2810:
The PR do allow non-madel based persistence configuration also use the ESH scheduler instead of the Quartz one.
For sure we should not merge it without some tests.
Have you some ideas how such a test case should look like?
I could keep a system running for some days that use different CRON expression to store a value.
As only the CRON scheduler needs to be tested, there is no "real" hardware or state changes necessary. Correct?
We need only to ensure that the CRON expression are still triggered after some days. Correct?

kaikreuzer:
Correct. What should be checked is whether there is any change in average system load and in the memory footprint. My plan was to test it with the openHAB demo setup, which has a couple of cron persistence configurations. Didn't yet find the time, tough.

kaikreuzer:
Ok, I finally did some testing. I couldn't make out any unusual CPU or memory consumption, even under high (persistence) load, so it lgtm.

The problem is that the initial job does not seem to be correctly scheduled. After a startup, no items are persisted. Once I touch the rrd4j.perist file, I see a
```
13:56:39.800 [WARN ] [.core.persistence.PersistenceManager] - Failed to delete cron job for dbId 'rrd4j'
```
in my log and afterwards it starts working.

Not sure whether this is an issue of this PR or of the new scheduler service. Will need to investigate, before this can be merged.

kaikreuzer:
Ok, I tracked it down to a bug in the scheduler: https://github.com/eclipse/smarthome/issues/2746
As long as this isn't fixed, we cannot merge this PR here.

maggu2810:
Thanks for tracking that down @kaikreuzer

kaikreuzer:
#2746 is fixed, so we are good to merge!

PR #2096, opened 2016-09-01
Fixed bug, which resulted in duplicate items in the item registry.

### LINKED ISSUE #2095: Warning during startup of rules
In the ESH demo setup, I find the following in the logs:

```
17:02:52.965 DEBUG o.e.s.m.r.r.i.e.RuleEngineImpl[:282] - Executing startup rule 'Initialize demo items'
17:03:07.808 WARN  o.e.s.m.r.j.RulesJvmModelInferrer[:201] - Duplicate field: 'DemoDimmer'. Ignoring 'org.eclipse.smarthome.core.library.items.DimmerItem'.
```
Analyzing it, I noticed that this is not a problem of Xtext+co, but that the cause is the fact that there are two DemoDimmer items available in the ItemRegistry, which is clearly a bug.

kaikreuzer:
Additionally, implemented a delay before the provider actively creates its items to avoid too many short living item creations.

Fixes #2095

### DIFF
--- a/bundles/core/org.eclipse.smarthome.core.thing/src/main/java/org/eclipse/smarthome/core/thing/internal/ChannelItemProvider.java
+++ b/bundles/core/org.eclipse.smarthome.core.thing/src/main/java/org/eclipse/smarthome/core/thing/internal/ChannelItemProvider.java
@@ -9,13 +9,13 @@ package org.eclipse.smarthome.core.thing.internal;
 
 import java.util.Collection;
 import java.util.Collections;
-import java.util.Date;
 import java.util.HashMap;
 import java.util.HashSet;
 import java.util.Locale;
 import java.util.Map;
 import java.util.Set;
 import java.util.concurrent.Executors;
+import java.util.concurrent.TimeUnit;
 
 import org.eclipse.smarthome.core.common.registry.ProviderChangeListener;
 import org.eclipse.smarthome.core.common.registry.RegistryChangeListener;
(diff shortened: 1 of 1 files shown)


bundles/core/org.eclipse.smarthome.core.thing/src/main/java/org/eclipse/smarthome/core/thing/internal/ChannelItemProvider.java
```diff
@@ -55,10 +57,12 @@
     private Map<String, Item> items = null;
 
     private boolean enabled = true;
+    private boolean initialized = false;
+    private long lastUpdate = new Date().getTime();
```
maggu2810: I would prefer to use `System.nanoTime()` because this one should be really a monotonic time base.

kaikreuzer: Done.

bundles/core/org.eclipse.smarthome.core.thing/src/main/java/org/eclipse/smarthome/core/thing/internal/ChannelItemProvider.java
```diff
@@ -141,13 +145,24 @@ protected synchronized void modified(Map<String, Object> properties) {
...
-            logger.debug("Enabling channel item provider.");
-            for (ProviderChangeListener<Item> listener : listeners) {
-                for (Item item : getAll()) {
-                    listener.added(this, item);
+            Executors.newSingleThreadExecutor().submit(new Runnable() {
+                @Override
+                public void run() {
+                    // we wait until no further new links or items are announced in order to avoid creation of
+                    // items which then must be removed again immediately.
+                    while (lastUpdate > new Date().getTime() - 2000L) {
```
maggu2810: I assume you would like to use nanoTime here, too (and perhaps TimeUnit.SECONDS.toNanos(2))

kaikreuzer: Indeed - thanks a lot, fixed now.

Issue #1757, opened 2016-06-28
Test failure(s): automation.module.timer.internal.RuntimeRuleTest


maggu2810:
commit id: d279efc780f28abfeb7f906239c39c9612ac3013

```text
Tests run: 2, Failures: 1, Errors: 0, Skipped: 0, Time elapsed: 10.818 sec <<< FAILURE! - in org.eclipse.smarthome.automation.module.timer.internal.RuntimeRuleTest
assert that timerTrigger works(org.eclipse.smarthome.automation.module.timer.internal.RuntimeRuleTest)  Time elapsed: 10.462 sec  <<< FAILURE!
java.lang.AssertionError: 
Expected: is <ON>
     but: was <OFF>
	at org.hamcrest.MatcherAssert.assertThat(MatcherAssert.java:20)
	at org.junit.Assert.assertThat(Assert.java:865)
	at org.junit.Assert.assertThat(Assert.java:832)
	at org.junit.Assert$assertThat.callStatic(Unknown Source)
	at org.eclipse.smarthome.automation.module.timer.internal.RuntimeRuleTest$_assert_that_timerTrigger_works_closure10.doCall(RuntimeRuleTest.groovy:120)
	at org.eclipse.smarthome.automation.module.timer.internal.RuntimeRuleTest$_assert_that_timerTrigger_works_closure10.doCall(RuntimeRuleTest.groovy)
	at sun.reflect.GeneratedMethodAccessor6.invoke(Unknown Source)
	at sun.reflect.DelegatingMethodAccessorImpl.invoke(DelegatingMethodAccessorImpl.java:43)
	at java.lang.reflect.Method.invoke(Method.java:498)
	at org.codehaus.groovy.reflection.CachedMethod.invoke(CachedMethod.java:90)
	at groovy.lang.MetaMethod.doMethodInvoke(MetaMethod.java:324)
	at org.codehaus.groovy.runtime.metaclass.ClosureMetaClass.invokeMethod(ClosureMetaClass.java:278)
	at groovy.lang.MetaClassImpl.invokeMethod(MetaClassImpl.java:1016)
	at org.codehaus.groovy.runtime.callsite.PogoMetaClassSite.call(PogoMetaClassSite.java:39)
	at org.codehaus.groovy.runtime.callsite.CallSiteArray.defaultCall(CallSiteArray.java:45)
	at org.codehaus.groovy.runtime.callsite.AbstractCallSite.call(AbstractCallSite.java:108)
[...]

Issue #2057, opened 2016-08-28
Test failure(s): automation.module.timer.internal.RuntimeRuleTest


maggu2810:
Related to: #1757, #2054

The last changes shows the error in more details.
The rule itself is using wildcards for the cron job timer.
The rule should be disabled but is not disabled after 10 seconds.
I don't know if the rule is triggered such frequently that the disabling does not work anymore.
@danchom Could you give me some informations why the disable does not work?

```text
-------------------------------------------------------
 T E S T S
-------------------------------------------------------
Running org.eclipse.smarthome.automation.module.timer.internal.RuntimeRuleTest
16:51:26.895 [main] DEBUG o.e.s.c.c.registry.AbstractRegistry - Provider 'com.sun.proxy.$Proxy7' has been added.
16:51:26.900 [main] INFO  o.e.s.a.c.internal.RuleRegistryImpl - Rule Managed Provider: org.eclipse.smarthome.automation.core.internal.ManagedRuleProvider@aafcffa is added.
16:51:26.900 [main] INFO  o.e.s.a.c.internal.RuleRegistryImpl - Rule provider: org.eclipse.smarthome.automation.core.internal.ManagedRuleProvider@aafcffa is added.
16:51:26.903 [main] DEBUG o.e.s.c.c.registry.AbstractRegistry - Provider 'org.eclipse.smarthome.automation.core.internal.ManagedRuleProvider' has been added.
16:51:26.933 [main] DEBUG o.e.s.a.i.c.p.AutomationResourceBundlesEventQueue - Process bundle event 2, for automation bundle 'org.eclipse.smarthome.automation.module.core' 
16:51:26.935 [main] DEBUG o.e.s.a.i.c.p.AutomationResourceBundlesEventQueue - Process bundle event 2, for automation bundle 'org.eclipse.smarthome.automation.module.timer' 
16:51:26.960 [Timer-0] DEBUG org.quartz.utils.UpdateChecker - Checking for available updated version of Quartz...
16:51:26.964 [Automation Provider Processing Queue] DEBUG o.e.s.a.i.c.p.RuleResourceBundleImporter - Parse rules from bundle 'org.eclipse.smarthome.automation.module.core' 
16:51:26.965 [main] DEBUG o.e.s.c.c.registry.AbstractRegistry - Provider 'org.eclipse.smarthome.core.items.ManagedItemProvider' has been added.
[...]

PR #2102, opened 2016-09-02
demonstrate that disable of rule fails some times

maggu2810:
After a rule has been executed, it state is set to IDLE.
This is done regardless if the rule has been deactivated before (at the time it has been executed).

@danchom, @marinmitev Can you have a look at the fix?

Fixes: #1757, #2057

### DIFF
--- a/bundles/automation/org.eclipse.smarthome.automation.core/src/main/java/org/eclipse/smarthome/automation/core/internal/RuleEngine.java
+++ b/bundles/automation/org.eclipse.smarthome.automation.core/src/main/java/org/eclipse/smarthome/automation/core/internal/RuleEngine.java
@@ -901,28 +901,33 @@ public class RuleEngine
      * @param td {@link TriggerData} object containing new values for {@link Trigger}'s {@link Output}s
      */
     protected void runRule(RuntimeRule rule, RuleEngineCallbackImpl.TriggerData td) {
-        RuleStatus ruleStatus = getRuleStatus(rule.getUID());
+        final String uid = rule.getUID();
+        RuleStatus ruleStatus = getRuleStatus(uid);
         if (ruleStatus == RuleStatus.IDLE) {
             try {
 
                 // change state to RUNNING
-                setRuleStatusInfo(rule.getUID(), new RuleStatusInfo(RuleStatus.RUNNING));
+                setRuleStatusInfo(uid, new RuleStatusInfo(RuleStatus.RUNNING));
                 clearContext(rule);
 
-                setTriggerOutputs(rule.getUID(), td);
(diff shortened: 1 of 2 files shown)


maggu2810:
@chriskn Can you test this one on your machine(s)?

maggu2810:
Updated PR: I think I have found the real cause of the failing module timer tests.

bundles/automation/org.eclipse.smarthome.automation.core/src/main/java/org/eclipse/smarthome/automation/core/internal/RuleEngine.java
```diff
@@ -901,28 +901,33 @@ private void removeMissingModuleTypes(Collection<String> moduleTypes) {
...
+                    logger.debug("The rule '{}' is NOT executed, since it has unsatisfied conditions.", uid);
                 }
             } catch (Throwable t) {
                 logger.error("Fail to execute rule '{}': {}", new Object[] { rule.getUID(), t.getMessage() }, t);
             }
 
-            // change state to IDLE
-            setRuleStatusInfo(rule.getUID(), new RuleStatusInfo(RuleStatus.IDLE));
+            // change state to IDLE only if the rule has not been DISABLED.
+            synchronized (this) {
```
marinmitev: Please note that this synchronization affects the execution of all rules which in theory should run in parallel and should not depend on each other. It seems better to synchronize with the rule object itself so other threads executing rules should not be blocked

marinmitev: The rest of the code in the commit looks good to me. Thanks for the fix!

maggu2810: The object `this` is used for synchronization in `setRuleStatusInfo`. That is the reason why it has been choosen.
Shouldn't the `get` and `set` methods for an information use the same sync. object?

marinmitev:
ok, I see that 'this' object is used also to synchronize the access to 'statusMap' which holds all rule statuses
I would suggest to @danchom to consider synchronizing with 'statusMap' instead 'this' in such places.
But since he will be back on Wednesday I will open a new issue about the synchronizations

maggu2810:
> I will open a new issue about the synchronizations

@marinmitev Perhaps you have seen that I fixed another usage of the rule status handling (same method, some lines above).
Have you created the issue for @danchom already (before it gets lost)? I haven't realize it.

PR #2074, opened 2016-08-30
Issue #2069 Implemented tests for FolderObserver

m-memova:
There are 8 test cases testing the functionality of
org.eclipse.smarthome.model.core.internal.folder.FolderObserver.
They check if the correct ModelRepository's methods are invoked
when certain events in the watched directory are triggered.

Signed-off-by: Mihaela Memova <mihaela.memova@musala.com>

PR #1597, opened 2016-06-02
WatchService Provides File Name Instead of Relative Path

dstivanov:
The file path, passed to the AbstractWatchQueueReader.processWatchEvent is now full (till now it was containing only the name of the target file, but not its relative to the watched directory).
Added tests for the AbstractWatchService to verify the behavior above.

**Why do we need the full path**:
When implementing the AbstractWatchService, we can provide an option to track the sub-directories by implementing the AbstractWatchService.watchSubDirectories() method. However, when the sub-directories are tracked, the AbstractWatchService.registerDirectory(Path subDir) will be called for each sub-directory. 
So we have two options - either to implement separate watch service for each of the sub-directories or to have the relative path as an argument when a watch event is detected. With this PR I assume that the second option is more convenient.

Signed-off-by: Dimitar Ivanov dstivanov@gmail.com

PR #2108, opened 2016-09-02
Implemented tests for ConfigDispatcher

petarnv:
There are 24 tests, 5 of which are ignored, because I think the behavior of the ConfigDispatcher in them is unexpected:
    1) **Incorrect Overriding** - I think that the configuration should be updated with the last modified value for a specific property. In ConfigDispatcher files are processed as listFiles() method ordered them and this order is never guaranteed(https://docs.oracle.com/javase/7/docs/api/java/io/File.html#listFiles()), i.e. the order can be alphabetical or some other order, which may cause the latest value for a property to be overridden by another value of the same property.
    2) **No Update On Delete** - On delete the file should be processed and the configurations should be updated, but in ConfigDispatcher, when file is deleted, it is only processed, but not updated. An update happens only when new content is added, but not when it is deleted.

As we have seen from https://github.com/eclipse/smarthome/pull/1597 and https://github.com/eclipse/smarthome/pull/2074, the API for files watching is platform dependent. These proposed tests are behaving exactly the same way on Windows and MAC.

### DIFF
--- a/bundles/config/org.eclipse.smarthome.config.dispatch/src/main/java/org/eclipse/smarthome/config/dispatch/internal/ConfigDispatcher.java
+++ b/bundles/config/org.eclipse.smarthome.config.dispatch/src/main/java/org/eclipse/smarthome/config/dispatch/internal/ConfigDispatcher.java
@@ -1,5 +1,5 @@
 /**
- * Copyright (c) 2014-2015 openHAB UG (haftungsbeschraenkt) and others.
+ * Copyright (c) 2014-2016 by the respective copyright holders.
  * All rights reserved. This program and the accompanying materials
  * are made available under the terms of the Eclipse Public License v1.0
  * which accompanies this distribution, and is available at
@@ -19,6 +19,8 @@ import java.nio.file.Path;
 import java.nio.file.WatchEvent;
 import java.nio.file.WatchEvent.Kind;
 import java.nio.file.WatchService;
+import java.util.Arrays;
+import java.util.Comparator;
 import java.util.Dictionary;
 import java.util.HashMap;
 import java.util.List;
@@ -67,6 +69,7 @@ import org.slf4j.LoggerFactory;
  * "pid: com.acme.smarthome.security".
  *
  * @author Kai Kreuzer - Initial contribution and API
+ * @author Petar Valchev - Added sort by modification time, when configuration files are read
  */
 public class ConfigDispatcher extends AbstractWatchService {
 
@@ -192,6 +195,13 @@ public class ConfigDispatcher extends AbstractWatchService {
         File dir = new File(getSourcePath());
         if (dir.exists()) {
             File[] files = dir.listFiles();
+            // Sort the files by modification time,
+            // so that the last modified file is processed last.
+            Arrays.sort(files, new Comparator<File>() {
+                public int compare(File left, File right) {
+                    return Long.valueOf(left.lastModified()).compareTo(right.lastModified());
+                }
+            });
             for (File file : files) {
                 try {
                     processConfigFile(file);
--- a/bundles/config/org.eclipse.smarthome.config.dispatch.test/.classpath
+++ b/bundles/config/org.eclipse.smarthome.config.dispatch.test/.classpath
@@ -1,0 +1,8 @@
+<?xml version="1.0" encoding="UTF-8"?>
+<classpath>
+	<classpathentry kind="con" path="org.eclipse.jdt.launching.JRE_CONTAINER/org.eclipse.jdt.internal.debug.ui.launcher.StandardVMType/JavaSE-1.7"/>
+	<classpathentry kind="con" path="org.eclipse.pde.core.requiredPlugins"/>
+	<classpathentry kind="src" path="src/test/groovy"/>
+	<classpathentry exported="true" kind="con" path="GROOVY_DSL_SUPPORT"/>
+	<classpathentry kind="output" path="target/classes"/>
+</classpath>
--- a/bundles/config/org.eclipse.smarthome.config.dispatch.test/.project
+++ b/bundles/config/org.eclipse.smarthome.config.dispatch.test/.project
@@ -1,0 +1,40 @@
+<?xml version="1.0" encoding="UTF-8"?>
+<projectDescription>
+	<name>org.eclipse.smarthome.config.dispatch.test</name>
+	<comment></comment>
+	<projects>
+	</projects>
+	<buildSpec>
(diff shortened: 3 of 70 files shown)


petarnv:
I saw that the ip-validation failed. I will fix my CLA.

maggu2810:
> 1) Incorrect Overriding - I think that the configuration should be updated with the last modified value for a specific property. In ConfigDispatcher files are processed as listFiles() method ordered them and this order is never guaranteed(https://docs.oracle.com/javase/7/docs/api/java/io/File.html#listFiles()), i.e. the order can be alphabetical or some other order, which may cause the latest value for a property to be overridden by another value of the same property.

I assume we are able to sort the list returned by listFiles() to a specific order.
We can order it by the file name, the last modification time, etc.
It need at least be defined.
The config dispatcher extends the AbstractWatchService so at runtime a modification will result on a reload of the last modified file.
Doesn't make a sort by modification time sense here?

> 2) No Update On Delete - On delete the file should be processed and the configurations should be updated, but in ConfigDispatcher, when file is deleted, it is only processed, but not updated. An update happens only when new content is added, but not when it is deleted.

Looking at the sources it seems that only ENTRY_CREATE and ENTRY_MODIFY are handled and delete is ignored at all.

petarnv:
> I assume we are able to sort the list returned by listFiles() to a specific order.
We can order it by the file name, the last modification time, etc.
It need at least be defined.
The config dispatcher extends the AbstractWatchService so at runtime a modification will result on a reload of the last modified file.
Doesn't make a sort by modification time sense here?

Yes - a modification will result in a reload of the last modified file and in this case the configuration will be updated properly and no sort is even needed. But that is not the the thing I test in the test methods from https://github.com/eclipse/smarthome/issues/2109. After activated, the ConfigDispatcher reads all the configuration files and processes them. And in the ignored tests I want to test precisely the processing after activation of the ConfigDispatcher, not the processing on modification of a file (I have another test that tests the processing on modification of a file). And the problem is that when the ConfigDispatcher is started, it processes files as listFiles() ordered them (not guaranteed order). So, if any change happened in the content of these files before the activation of the ConfigDispatcher, а modification event will not be handled and all the files (modified or not) will be processed as listFiles() ordered them and the configuration might not be updated properly.

> Looking at the sources it seems that only ENTRY_CREATE and ENTRY_MODIFY are handled and delete is ignored at all.

Now I see that my explanation can be misleading. What I meant was that in the  ConfigDispatcher, when the content of a file(not the file itself) is deleted, it is only processed, but not updated. It is processed because when we delete the content of the file this is considered ENTRY_MODIFY, not ENTRY_DELETE. And as I have said - an update happens only when new content is added, but not when it is deleted.

maggu2810:
>> I assume we are able to sort the list returned by listFiles() to a specific order.
We can order it by the file name, the last modification time, etc.

> And the problem is that when the ConfigDispatcher is started, it processes files as listFiles() ordered them (not guaranteed order). 

As written above, we can sort and process the files that are returned by listFiles() to grant that also in this case the one with the last modification timestamp wins. Doesn't it solve your non deterministic behaviour if we state "also on startup the files are read in order of the modification timestamp"?

petarnv:
I have added a sort by modification time of the files that are read at the activation of the ConfigDispatcher, so that the last modified file is processed last. I also removed the `@Ignore` tag in the tests, related to the incorrect overriding, because they are not failing after the change.

maggu2810:
@petarnv Sorry about the delay. Can you give me a short summary what is missing? Are there still ignored tests?

petarnv:
@maggu2810, there is still one ignored test left, because when a content of a file is deleted, the configuration is not updated properly (it should be updated without the deleted from the file values), but I am working on fixing this problem.

maggu2810:
Thanks. If there is no urgent reason to merge this PR (with ignoring one test) now, I will wait for the fix, so we do not need to ignore a test. Okay?

petarnv:
Okay

petarnv:
I have updated the PR. I have fixed the problem with the configuration not being updated, when content of a file is deleted and I have added a couple more tests, that show the fix is really working. The new tests, as well as the old ones, are behaving exactly the same way on Windows, Linux and MAC. There are no more ignored tests.

kaikreuzer:
Sorry to come late to the party.

> Looking at the sources it seems that only ENTRY_CREATE and ENTRY_MODIFY are handled and delete is ignored at all.

This is actually done this way on purpose. With the current implementation, the behavior is identical whether your runtime is up or down at the moment when you remove a file.
With your implementation, the parameters are removed from ConfigAdmin if you delete the file while the runtime is up. But if you delete it while the runtime is down and you start it up, the settings in ConfigAdmin are kept (because you have no clue that this file ever existed).

So while it is not ideal, it at least is consistent. The only way to get rid off settings is to set parameters to null (empty string?) - but configurations themselves can only be deleted through other mechanisms (config admin through console, starting the OGSi framework with clean option, etc.).

bundles/config/org.eclipse.smarthome.config.dispatch/src/main/java/org/eclipse/smarthome/config/dispatch/internal/ConfigDispatcher.java
```diff
@@ -38,39 +42,47 @@
 import org.slf4j.LoggerFactory;
 
 /**
- * This class provides a mean to read any kind of configuration data from
```
petarnv: Sorry about this. I will revert to the original formatting, when I update the PR.

petarnv:
@kaikreuzer, I agree with you, that the change, related to the configuration delete would not behave the same way, if the runtime is down. I can remove it, when I update the PR. But what should I do with the tests, that would fail, if that change is removed - should I ignore them, or should I remove them.

And also, there is a second change in the ConfigDispatcher, that I have made - the sort of the files by modification time, when they are read. Do you approve this change?

kaikreuzer:
> But what should I do with the tests, that would fail, if that change is removed - should I ignore them, or should I remove them.

Probably ok to remove them, if they were meant for a functionality that actually isn't there.

> And also, there is a second change in the ConfigDispatcher, that I have made - the sort of the files by modification time, when they are read. Do you approve this change?

Yes, I saw this. I would consider it an edge case, because in theory, no two files should ever deal with the same configuration parameters - but IF this situation occurs, your fix indeed makes sense.

petarnv:
@kaikreuzer, I have updated my PR with the changes, that we discussed.

kaikreuzer:
Great, many thanks!

Issue #2121, opened 2016-09-06
Item initialization / item channel link handling


maggu2810:
This issue should continue the non-topic discussion started here: #2087 

@maggu2810 (https://github.com/eclipse/smarthome/pull/2087#issuecomment-244746937)

> Hi @sbussweiler,
I see a lot of
```text
Received update for non-existing item: Item '...' could not be found in the item registry
```
now.
After that messages, I see
```text
Enabling channel item provider.
```
and all the
```text
Item '...' has been added.
```

> That is a little bit confusing.

> I need to inspect if the updates are generated by my handlers or the framework itself.

> If them are generated by my handlers:
My thing handlers call `updateState(channelIdWithoutGroup, state);` so if the framework tries to update items, it already knows the item names that are linked to the channels and tries to delegate that updates but the items does not yet exist.
If the item is created I would assume that a refresh is called, so the thing handler is able to send the status update of the channel linked to that item again (we added this to the framework some time ago, see #1677). But if I look at the item state it is NULL.

@kaikreuzer (https://github.com/eclipse/smarthome/pull/2087#issuecomment-244875272)

> @maggu2810 Upon some analysis, I would assume that these log messages are due to https://github.com/eclipse/smarthome/pull/2096 (and not due to this PR). In #2096, I introduced a small delay before the ChannelItemProvider creates its items. The ThingManager assumes that if there is a link, the referenced item MUST exist as well and thus it sends events for it.
I am not sure how to solve this best; we currently cannot technically guarantee that items are always loaded/created before any link is loaded/created. The solution might also be related to the discussion in https://github.com/eclipse/smarthome/issues/1896, i.e. when we can be sure that all stuff is fully loaded after startup.

maggu2810:
@kaikreuzer Thank you for the response.
I assume also if the startup is done completely we cannot assume that all items has been created.
IIRC the ChannelItemProvider (correct me if I am wrong) service can be deactivated and is so an optional feature.
If this is correct, a channel item link exist but the item is not available as long as it is created by some other logic / by the user / ...
Can't we send a refresh command to an item (perhaps only ones a link exist for) if the item has been created?

sjka:
on a first look, I think the root cause lies within the ´AbstractLinkRegistry.getLinkedItems(UID)´ method. Although it claims that it returns items (or better: the names of the linked items), it actually only returns the pointers to the items without checking if they actually exist.
[...]

PR #2122, opened 2016-09-06
respect link and item existence for channel-linked notification

maggu2810:
Fixes: https://github.com/eclipse/smarthome/issues/2121

Does this make sense?
At least it is IMHO no real overhead and does not break anything.

### DIFF
--- a/bundles/core/org.eclipse.smarthome.core.thing/src/main/java/org/eclipse/smarthome/core/thing/link/ItemChannelLinkRegistry.java
+++ b/bundles/core/org.eclipse.smarthome.core.thing/src/main/java/org/eclipse/smarthome/core/thing/link/ItemChannelLinkRegistry.java
@@ -10,12 +10,8 @@ package org.eclipse.smarthome.core.thing.link;
 import java.util.Collection;
 import java.util.HashSet;
 import java.util.LinkedHashSet;
-import java.util.LinkedList;
 import java.util.Set;
 
-import org.eclipse.smarthome.core.common.registry.Provider;
-import org.eclipse.smarthome.core.common.registry.RegistryChangeListener;
-import org.eclipse.smarthome.core.items.Item;
 import org.eclipse.smarthome.core.items.ItemRegistry;
 import org.eclipse.smarthome.core.thing.ChannelUID;
 import org.eclipse.smarthome.core.thing.Thing;
@@ -23,77 +19,19 @@ import org.eclipse.smarthome.core.thing.ThingRegistry;
 import org.eclipse.smarthome.core.thing.ThingUID;
 import org.eclipse.smarthome.core.thing.UID;
 import org.eclipse.smarthome.core.thing.link.events.LinkEventFactory;
-import org.osgi.service.component.ComponentContext;
 
 /**
  * {@link ItemChannelLinkRegistry} tracks all {@link ItemChannelLinkProvider}s
  * and aggregates all {@link ItemChannelLink}s.
  *
  * @author Dennis Nobel - Initial contribution
- * @author Markus Rathgeb - Respect item existence
+ * @author Markus Rathgeb - Linked items returns only existing items
  *
  */
 public class ItemChannelLinkRegistry extends AbstractLinkRegistry<ItemChannelLink> {
 
-    /*
-     * Methods of the AbstractRegistry does not respect the component life cycle.
-     * The method AbstractRegistry.addProvider executes logic if the component has not been activated, too.
-     * This could result in some wired runtime behavior because member variables that are injected by the service
-     * management could be present or not.
-     */
-    boolean active = false;
-    private final LinkedList<Provider<ItemChannelLink>> waitingProviders = new LinkedList<>();
-
     private ThingRegistry thingRegistry;
     private ItemRegistry itemRegistry;
-
-    protected void activate(ComponentContext context) {
-        itemRegistry.addRegistryChangeListener(itemRegistryChangeListener);
-        synchronized (this) {
-            active = true;
-            while (!waitingProviders.isEmpty()) {
-                super.addProvider(waitingProviders.removeFirst());
-            }
-        }
-    }
-
-    protected void deactivate() {
-        synchronized (this) {
-            active = false;
-        }
-        itemRegistry.removeRegistryChangeListener(itemRegistryChangeListener);
-    }
-
-    private final RegistryChangeListener<Item> itemRegistryChangeListener = new RegistryChangeListener<Item>() {
-        @Override
-        public void added(Item element) {
-            final String itemName = element.getName();
-            for (final ItemChannelLink itemChannelLink : getAll()) {
-                if (itemChannelLink.getItemName().equals(itemName)) {
-                    notifyListenersAboutAddedElement(itemChannelLink, true);
-                }
-            }
-        }
-
-        @Override
-        public void removed(Item element) {
-            final String itemName = element.getName();
-            for (final ItemChannelLink itemChannelLink : getAll()) {
-                if (itemChannelLink.getItemName().equals(itemName)) {
-                    notifyListenersAboutRemovedElement(itemChannelLink, true);
-                }
-            }
-        }
-
-        @Override
-        public void updated(Item oldElement, Item element) {
-            if (!oldElement.equals(element)) {
-                this.removed(oldElement);
-                this.added(element);
-            }
-        }
-
-    };
 
     /**
      * Returns a set of bound channels for the given item name.
@@ -128,7 +66,7 @@ public class ItemChannelLinkRegistry extends AbstractLinkRegistry<ItemChannelLin
         final Set<String> linkedItems = new LinkedHashSet<>();
         for (final AbstractLink link : getAll()) {
             final String itemName = link.getItemName();
-            if (link.getUID().equals(uid) && itemRegistry.containsItem(itemName)) {
+            if (link.getUID().equals(uid) && itemRegistry.get(itemName) != null) {
                 linkedItems.add(itemName);
             }
         }
@@ -176,18 +114,6 @@ public class ItemChannelLinkRegistry extends AbstractLinkRegistry<ItemChannelLin
         this.itemRegistry = null;
     }
 
-    @Override
-    protected void addProvider(Provider<ItemChannelLink> provider) {
-        // Workaround for addProvider: see description of member 'active' above
-        synchronized (this) {
-            if (active) {
-                super.addProvider(provider);
-            } else {
-                waitingProviders.add(provider);
-            }
-        }
-    }
-
     public void removeLinksForThing(ThingUID thingUID) {
         if (this.managedProvider != null) {
             ((ManagedItemChannelLinkProvider) this.managedProvider).removeLinksForThing(thingUID);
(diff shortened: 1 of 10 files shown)


kaikreuzer:
It only addressed the second part of https://github.com/eclipse/smarthome/issues/2121#issuecomment-244891078 (which is not the cause of the problem that you describe).

What is imho missing is that ThingHandlerCallback.stateUpdated and ThingHandlerCallback.postCommand need to be updated. And for this it is probably best to adapt ItemChannelLinkRegistry.getLinkedItems to only return item names that really exist (i.e. ItemChannelLinkRegistry will have to depend on ItemRegistry as well. Does that make sense?

maggu2810:
Hm, I assume it makes sense to hide links that item is not present.
I added another commit that tries to implement this.

I detected (has this done by intention?) a wired (perhaps I have done something wrong) runtime behavior of the AbstractRegistry as it does not respect the component life-cycle.

maggu2810:
I will have a look at the test failures after we agree about a solution at all.

bundles/core/org.eclipse.smarthome.core.thing/src/main/java/org/eclipse/smarthome/core/thing/link/ItemChannelLinkRegistry.java
```diff
@@ -9,24 +9,91 @@
...
+                if (itemChannelLink.getItemName().equals(itemName)) {
+                    notifyListenersAboutRemovedElement(itemChannelLink, true);
+                }
+            }
+        }
+
+        @Override
+        public void updated(Item oldElement, Item element) {
+            if (!oldElement.equals(element)) {
+                this.removed(oldElement);
```
kaikreuzer: I'd say that this method should do nothing as the item still exists, so the link is valid - that's all that should be important here.

bundles/core/org.eclipse.smarthome.core.thing/src/main/java/org/eclipse/smarthome/core/thing/link/ItemChannelLinkRegistry.java
```diff
@@ -99,19 +198,40 @@ public void removeLinksForThing(ThingUID thingUID) {
...
 
     @Override
     protected void notifyListenersAboutUpdatedElement(ItemChannelLink oldElement, ItemChannelLink element) {
         super.notifyListenersAboutUpdatedElement(oldElement, element);
         // it is not needed to send an event, because links can not be updated
     }
+
+    protected void notifyListenersAboutAddedElement(final ItemChannelLink element, boolean itemExists) {
+        if (itemExists) {
+            super.notifyListenersAboutAddedElement(element);
```
kaikreuzer: You now completely ignore the fact that links are created by even suppressing their events. I don't think we should do that. The link is there when it is created, it should merely not be assumed that the item exists as well.

I would expect that the ChannelItemProvider stops working with your current code, since it actually listens for these events and only creates its items when a new link is created.

bundles/core/org.eclipse.smarthome.core/src/main/java/org/eclipse/smarthome/core/items/ItemRegistry.java
```diff
@@ -25,6 +25,14 @@
 public interface ItemRegistry extends Registry<Item, String> {
 
     /**
+     * This methods checks if an item is part of the registry.
+     *
+     * @param name the item name
+     * @return true if there is an item with given name, otherwise false
+     */
+    public boolean containsItem(String name);
```
kaikreuzer: Not sure we should touch the ItemRegistry interface here - a 'get(name)!=null' would do as well. If we feel we need an "exists" method, we should imho rather add it as 'Registry.exists(K)'.

maggu2810: The registry will hold (for the ItemRegistry) elements of type Item, so we cannot rely on "name" (an item attribute) in the Regitry class.
We could use getItem(name) and catch ItemNotFoundExcpetion. I am not a fan of an exception based control flow, if you like that. Okay

bundles/core/org.eclipse.smarthome.core.thing/src/main/java/org/eclipse/smarthome/core/thing/link/ItemChannelLinkRegistry.java
maggu2810: For the link itself you are right.

But what I want to solve is that all my item states are wrong now and so I tried to use the event to trigger a refresh command and a correct state update of the items.
If the items differ it could be because the new item has another state (UnDefType.NULL).

If we do not want to use the channel link events here, then trigger a link notification on "item added after link" is wrong, too.
Isn't it?

bundles/core/org.eclipse.smarthome.core.thing/src/main/java/org/eclipse/smarthome/core/thing/link/ItemChannelLinkRegistry.java
maggu2810: Okay, sound reasonable.

Should we inform the listeners on link creation and item creation (same question as above: should we use 'added'. 'removed' of the items here)?
Perhaps I misunderstand your comments I assumed you want to use that events in the channel link to refresh the channels, too.
Can we send the event twice on link creation and on item creation?

bundles/core/org.eclipse.smarthome.core/src/main/java/org/eclipse/smarthome/core/items/ItemRegistry.java
kaikreuzer: ItemRegistry extends Registry and thus has `public E get(K key);`, where K is String (the item name) and which returns null, if the element does not exist - you do not have to deal with ItemNotFoundExcpetion here.

bundles/core/org.eclipse.smarthome.core.thing/src/main/java/org/eclipse/smarthome/core/thing/link/ItemChannelLinkRegistry.java
kaikreuzer: "updated" here is called if the item meta data changes - its category, label or whatever. It is NOT called for state changes, so this should be completely independent here.

bundles/core/org.eclipse.smarthome.core.thing/src/main/java/org/eclipse/smarthome/core/thing/link/ItemChannelLinkRegistry.java
kaikreuzer: > Should we inform the listeners on link creation and item creation

I would say only on link creation, because this is what they register for -> keep the behavior on this part as it was.

> Perhaps I misunderstand your comments I assumed you want to use that events in the channel link to refresh the channels, too.

Right, which you correctly implemented in https://github.com/eclipse/smarthome/pull/2122/commits/cbe1eb15fde14ab92eef250b92d6ced8c15b4ec5 - so I would suggest you keep this code in place as well.

> Can we send the event twice on link creation and on item creation?

No, but the ThinkLinkManager can register on both evens (links and items) and call the "channelLinked()" method when it is sure that both link+item exist.

bundles/core/org.eclipse.smarthome.core/src/main/java/org/eclipse/smarthome/core/items/ItemRegistry.java
maggu2810: The interface is okay, this implementation is another topic...
Don't need to discuss it here, not related to this changes.
But if you have a look at ItemRegistryImpl you see why I am are not comfortable with the current implementation (`get(itemName)` is using `getItem(itemName)` and catches the exception. `getItem` calls `getItems` that uses `getAll` that creates an copy of...).

maggu2810:
Sorry, I will give it another try... Wait a moment... :wink:

maggu2810:
Next try :wink:

* ItemChannelLinkRegistry.getLinkedItems returns no longer the names of non-existing items
* ThingLinkManager informs the handlers about linked channels if the link AND the item exists (no more if only the link exists)
* ThingLinkManager informs the handlers about linked channels if the link or the item no longer exists (previously it has not been called if the item has been removed but the link still exists)

kaikreuzer:
Thanks sounds and looks good to me. Can you confirm yourself that your issue #2121 is solved with it?

So last thing to do is to fix this test:
```
Failed tests: 
  ThingLinkManagerOSGiTest.assert that channelLinked and channelUnlinked at ThingHandler is called:136->OSGiTest.waitForAssert:-1->OSGiTest.waitForAssert:181 
Expected: is <hue:lamp:lamp1:1>
     but: was null
```

@sjka Do you also want to have a look before it is merged?

maggu2810:
> Thanks sounds and looks good to me.

Okay, will have a look at the tests in the next free minutes and ping you later.

kaikreuzer:
Travis is still failing :-(

bundles/core/org.eclipse.smarthome.core.thing/src/main/java/org/eclipse/smarthome/core/thing/link/ThingLinkManager.java
```diff
@@ -115,6 +171,11 @@ public void added(ItemChannelLink itemChannelLink) {
 
         @Override
         public void removed(ItemChannelLink itemChannelLink) {
+            if (itemRegistry.get(itemChannelLink.getItemName()) == null) {
+                // Don't inform about the unlink if the item does not exist.
+                // The handler is informed if the item has been removed.
+                return;
```
maggu2810: @kaikreuzer I assume this here is problematic.

If a link is removed and the item does not exist, this lines assume that the unlink is already sent by the item removal listener.

See: https://github.com/eclipse/smarthome/blob/6a8cb49/bundles/core/org.eclipse.smarthome.core.thing/src/main/java/org/eclipse/smarthome/core/thing/internal/ChannelItemProvider.java#L307
If a link is removed the ChannelItemProvider removes the item (if it is owned by him), too.

If the item and link is removed before the listener functions are called, no unlink of the thing manager is called.
Here we have "zero or one" handler unlink notifiaction(s).

Do you think the "removal" could be send for both, if the item is removed and the link is removed?
Then we have "one or two" handler unlink notifiaction(s).

I assume all that event handling is async. and the only way to really "one" notification is if we cache all item names internally. But I don't think this is memory efficient.

WDYT?

maggu2810:
> Travis is still failing :-(

I have written a comment the same time:
https://github.com/eclipse/smarthome/pull/2122#discussion_r77658573

bundles/core/org.eclipse.smarthome.core.thing/src/main/java/org/eclipse/smarthome/core/thing/link/ThingLinkManager.java
kaikreuzer: I would be fine to actually only call channelUnlinked() when a link is removed and don't care about items.
The only thing that could happen is that an item disappears unnoticed and that the handler tries to send a command/update nonetheless. But as this can anyhow happen anytime (since hardly any handler first checks via "isLinked()", I don't think that there is any issue with that.

bundles/core/org.eclipse.smarthome.core.thing/src/main/java/org/eclipse/smarthome/core/thing/link/ThingLinkManager.java
```diff
@@ -99,10 +113,52 @@ public boolean isAutoLinksEnabled() {
...
+        }
+
+    };
+
     private final RegistryChangeListener<ItemChannelLink> itemChannelLinkRegistryChangeListener = new RegistryChangeListener<ItemChannelLink>() {
 
         @Override
         public void added(ItemChannelLink itemChannelLink) {
+            if (itemRegistry.get(itemChannelLink.getItemName()) == null) {
+                // Don't inform about the unlink if the item does not exist.
```
sjka: "link", not "unlink"

bundles/core/org.eclipse.smarthome.core.thing/src/main/java/org/eclipse/smarthome/core/thing/link/ThingLinkManager.java
maggu2810: > I would be fine to actually only call channelUnlinked() when a link is removed and don't care about items.

Okay

> The only thing that could happen is that an item disappears unnoticed and that the handler tries to send a command/update nonetheless. But as this can anyhow happen anytime (since hardly any handler first checks via "isLinked()", I don't think that there is any issue with that.

The test that recently fails needs at least one call to the unlink method and does not succeed if there is no such call.
So, I will ignore the existence of the item if a link is removed.

bundles/core/org.eclipse.smarthome.core.thing/src/main/java/org/eclipse/smarthome/core/thing/link/ThingLinkManager.java
maggu2810: Thanks

bundles/core/org.eclipse.smarthome.core.thing/src/main/java/org/eclipse/smarthome/core/thing/link/ThingLinkManager.java
sjka: > I would be fine to actually only call channelUnlinked() when a link is removed and don't care about items.

makes sense. we should be aware though that bindings might get two "linked" events without a "unlinked" in between, if just the item dis- and reappears for whatever reason. Would be okay as far as I can see, but should be documented.

sjka:
> @sjka Do you also want to have a look before it is merged?

yes, thanks, lgtm.

maggu2810:
> Can you confirm yourself that your issue #2121 is solved with it?

Yes

> Travis is still failing :-(

It succeeded - at least once :wink:

kaikreuzer:
Congrats - that's enough, let's merge 👍

Issue #2091, opened 2016-09-01
Thing updates are not reflected in the PaperUI


sjka:
having fixed #2089, I noticed that thing updates (e.g. label changes) are not reflected in the PaperUI. I had to refresh the page in order to see them.
I saw that the PaperUI is not listening to `smarthome/things/*/updated` events yet (unlike for items, where it does it already). Wouldn't it make sense to do so for Things?

sjka:
@aounhaider1 wdyt?

aounhaider1:
Actually this should not happen. Currently, View thing is being loaded with a force update from the rest endpoint instead of serving from local storage. 
However, listening to thing updates can potentially improve the number of requests sent to the server.

PR #2123, opened 2016-09-06
Thing update event

aounhaider1:
fixes https://github.com/eclipse/smarthome/issues/2091

### DIFF
--- a/extensions/ui/org.eclipse.smarthome.ui.paper/web-src/js/controllers.configuration.js
+++ b/extensions/ui/org.eclipse.smarthome.ui.paper/web-src/js/controllers.configuration.js
@@ -470,6 +470,7 @@ angular.module('PaperUI.controllers.configuration', [ 'PaperUI.constants' ]).con
     }
 
     $scope.configChannel = function(channel, thing, event) {
+
         var channelType = this.getChannelFromChannelTypes(channel.channelTypeUID);
 
         $mdDialog.show({
@@ -479,7 +480,7 @@ angular.module('PaperUI.controllers.configuration', [ 'PaperUI.constants' ]).con
             hasBackdrop : true,
             locals : {
                 channelType : channelType,
-                channel : channel,
+                channelUID : channel.uid,
                 thing : thing
             }
         });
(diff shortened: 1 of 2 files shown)


maggu2810:
@sjka As you created #2091, can you check that changes, too?

sjka:
The `Thing` itself now gets properly updated, e.g. label changes are reflected instantly. Nice!

However, not its channels. Let's assume you have the following:

```
yahooweather:weather:losangeles "Los Angeles 1" [ location=2442047, refresh=60 ] {
    Channels: 
        String : foo
}
```

Navigate to Configuration -> Things -> Los Angeles 1. You should see the three original channels plus the "foo" one. Now change the file to the following:

```
yahooweather:weather:losangeles "Los Angeles 1" [ location=2442047, refresh=60 ] {
    Channels: 
        String : foo
        String : bar
}
```

The "bar" channel will only show up after a reload of the page.

aounhaider1:
Updated accordingly! However, if you are editing a channel's configurations in two browsers and save one of them, the change will not be directly visible unless you close and reopen the dialog. Implementing it should be possible but I am not sure of it is worth the effort.

aounhaider1:
Needs updating.

aounhaider1:
It should now support adding/removing channels.

extensions/ui/org.eclipse.smarthome.ui.paper/web-src/js/controllers.configuration.js
```diff
@@ -662,7 +666,7 @@ angular.module('PaperUI.controllers.configuration', [ 'PaperUI.constants' ]).con
         thingService.update({
             thingUID : thing.UID
         }, $scope.thing, function() {
-            thingRepository.update($scope.thing);
+            // thingRepository.update($scope.thing);
```
sjka: if that's not needed here anymore, could you please delete it?

sjka:
thanks. I only found a left-over commented-out code line. Could you please remove it? 
Otherwise lgtm

aounhaider1:
Updated.

Issue #2117, opened 2016-09-04
bug Rule : Script interpreter couldn't be obtain


igorrobertifoc:
I have the same problem posted on
https://github.com/eclipse/smarthome/issues/1393.
and discussed on
https://community.openhab.org/t/intermittant-rule-issue/11663/8

it seem that rules, when are updated ( or refreshed by the system ), the "old" version stay on memory and generate the error :Script interpreter couldn't be obtain

my test.items
```
Number Scena  "Scena" <scena>
String VoiceCommand
```
my test.rules
```
rule testVoice
when
Item VoiceCommand received command
then
var String command = VoiceCommand.state.toString.toLowerCase

if (command.contains("luce relax") {
    Scena.sendCommand(1)
}
if (command.contains("luce lettura") {
    Scena.sendCommand(2)
}
if (command.contains("accendi luce") {
    Scena.sendCommand(0)
}
if (command.contains("spegni luce") {
    Scena.sendCommand(4)
}   
end
```
With openhab android is necessary just speak to generate the error

igorrobertifoc:
Just an information can maybe help, I installed the new rules manager V2 (maybe some conflict)

kaikreuzer:
After a looong, loooong debugging session, I hope I have finally tackled the problem!

igorrobertifoc:
Goooood! Bravo!
where it was the problem just to know?

Thank you

maggu2810:
You can have a look at #2125

PR #2125, opened 2016-09-06
Fixes startup and reload behaviour of DSL rules

kaikreuzer:
- Make sure that only loaded rule resources are used.
- When items are changed, notify the rule engine that all rules are reloaded.


kaikreuzer:
Fixes #2117, which was introduced by the Xtext upgrade a while ago.

maggu2810:
Without learning Xtext I assume I cannot understand the real impact of using e.g. eResource method instead of the eIsProxy method.
As this PR does not break anything but the DSL rule behaviour itself if it is not working and I assume you have tested the changes, should we wait for further comments or merge ASAP?

kaikreuzer:
You cannot even understand it after learning Xtext/EMF, trust me ;-)
I would tend to say that it at least makes stuff better and not worse and therefore it should be merged.

igorrobertifoc:
Hello,
I updated with the nightly build on
https://openhab.ci.cloudbees.com/job/openHAB-Distribution/ws/distributions/openhab-offline/target/apt-repo/

but it give me the same error. Is the patch updated?

Thank you

kaikreuzer:
Better try the zipped distro to make sure there is nothing left from an earlier install.

PR #2126, opened 2016-09-06
fix another rule status change

maggu2810:
We must not check if the status is ... (IDLE in this case) and set it to another status (RUNNING in this case) without using synchronization. At least this does not work.

```java
        if (ruleStatus == RuleStatus.IDLE) {
            try {
                // change state to RUNNING
                setRuleStatusInfo(uid, new RuleStatusInfo(RuleStatus.RUNNING));
```

If between the above lines the rule status has been set to disabled, this change has been lost and the rule is enabled again.


maggu2810:
@kaikreuzer I assume it is okay to merge as it is rather similar to #2102. Perhaps we can get some green Travis CI build some time.

PR #2132, opened 2016-09-07
add / remove provider logic requires active component

maggu2810:
After some modifications of the ItemChannelLinkRegistry yesterday (which
have been never going upstream) I realized that the AbstractRegistry
implementation relies on an active component if provider are added (and
removed).
Using the multiple (0..n) cardinality and policy static the providers
could be added / removed if the component is active but also if the
component is not active.
Regardless if the error is currently thrown this is an error in the
design and should be fixed.

This PR introduce a service tracker to track the providers as long as
the component is active.

### DIFF
--- a/bundles/core/org.eclipse.smarthome.core.thing/src/main/java/org/eclipse/smarthome/core/thing/link/AbstractLinkRegistry.java
+++ b/bundles/core/org.eclipse.smarthome.core.thing/src/main/java/org/eclipse/smarthome/core/thing/link/AbstractLinkRegistry.java
@@ -12,6 +12,7 @@ import java.util.LinkedHashSet;
 import java.util.Set;
 
 import org.eclipse.smarthome.core.common.registry.AbstractRegistry;
+import org.eclipse.smarthome.core.common.registry.Provider;
 import org.eclipse.smarthome.core.thing.UID;
 
 /**
@@ -23,7 +24,12 @@ import org.eclipse.smarthome.core.thing.UID;
  * @param <L>
  *            Concrete type of the abstract link
  */
-public abstract class AbstractLinkRegistry<L extends AbstractLink> extends AbstractRegistry<L, String> {
+public abstract class AbstractLinkRegistry<L extends AbstractLink, P extends Provider<L>>
+        extends AbstractRegistry<L, String, P> {
+
+    protected AbstractLinkRegistry(final Class<P> providerClazz) {
+        super(providerClazz);
+    }
 
     /**
      * Returns if an item for a given item name is linked to a channel or thing for a
--- a/bundles/core/org.eclipse.smarthome.core/src/main/java/org/eclipse/smarthome/core/common/registry/AbstractRegistry.java
+++ b/bundles/core/org.eclipse.smarthome.core/src/main/java/org/eclipse/smarthome/core/common/registry/AbstractRegistry.java
@@ -14,6 +14,9 @@ import java.util.concurrent.CopyOnWriteArraySet;
 
 import org.eclipse.smarthome.core.events.Event;
 import org.eclipse.smarthome.core.events.EventPublisher;
+import org.osgi.framework.BundleContext;
+import org.osgi.framework.ServiceReference;
+import org.osgi.util.tracker.ServiceTracker;
 import org.slf4j.Logger;
 import org.slf4j.LoggerFactory;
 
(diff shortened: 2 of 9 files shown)


kaikreuzer:
This is now a different solution than you chose in https://github.com/eclipse/smarthome/pull/2122/commits/80ff507d5c85b714815c36c66f6dcfc2b923aa80. Can you elaborate on this?

And since ItemChannelLinkRegistry overrides the activate method (without calling super), this looks like a problem to me, right?

maggu2810:
> This is now a different solution than you chose in 80ff507. 

Don't see a realtionship between this two PRs.

> And since ItemChannelLinkRegistry overrides the activate method (without calling super)

Really? Didn't find them https://github.com/eclipse/smarthome/blob/master/bundles/core/org.eclipse.smarthome.core.thing/src/main/java/org/eclipse/smarthome/core/thing/link/ItemChannelLinkRegistry.java

maggu2810:
At I worked on 80ff507 I realized that the AbstractRegistry is used in Components only but does not respect the Component life cycle. The error does not occur anymore as I finished 80ff507 because the problem that should be solved has been solved in a way that the life-cycle violation it hidden as before.

But only because the violation is not visible it does not mean that we should not handle it correctly.

kaikreuzer:
> Really? Didn't find them

Ok, I had looked to much at your commits yesterday, so I thought that code still exists - I didn't notice that you reverted it in [next try](https://github.com/eclipse/smarthome/pull/2122/commits/61799032eddc250ccd50090cc25d74919439a499).

> Don't see a realtionship between this two PRs.

I linked to your commit, not to the PR. In the commit, you introduced the field "waitingProviders" and kept the use of DS (and not used ServiceTrackers). I assumed that was meant to solve the same lifecycle issue that you refer to in this PR?

maggu2810:
> I linked to your commit, not to the PR. In the commit, you introduced the field "waitingProviders" and kept the use of DS (and not used ServiceTrackers). I assumed that was meant to solve the same lifecycle issue that you refer to in this PR?

Okay, I was a little bit confused about the usage of the activator and get lost in the different changes (I still have in my brain), too.
It worked around the problem but not in a good fashion. ;)

kaikreuzer:
Tested it, seems to work well, thanks.

bundles/core/org.eclipse.smarthome.core.thing/src/main/java/org/eclipse/smarthome/core/thing/link/AbstractLinkRegistry.java
```diff
@@ -23,7 +24,12 @@
  * @param <L>
  *            Concrete type of the abstract link
  */
```
afuechsel: Although this PR has been merged a while ago, I just now stumbled across the extension and found that the second parameter has not been added to the javadoc. I personally feel uncomfortable with so many generics so it would be good for understanding to document all generic type parameters. Especially, since I still have some implementation of the removed ItemThingLinkRegistry, that is based on this class and must now be changed too.

bundles/core/org.eclipse.smarthome.core/src/main/java/org/eclipse/smarthome/core/common/registry/AbstractRegistry.java
```diff
@@ -30,7 +33,8 @@
  * @param <E>
  *            type of the element
  */
-public abstract class AbstractRegistry<E, K> implements ProviderChangeListener<E>, Registry<E, K> {
+public abstract class AbstractRegistry<E, K, P extends Provider<E>>
```
afuechsel: Also here the type parameter are not completely documented.

maggu2810:
> Only two comments about missing javadoc. WOuld be good to add it in a small PR.

@afuechsel Nothing to comment. You are right, I will add it. Will try to find some time this week end or next week.

tomhoefer:
I have one more question regarding this PR... why do we need to inject the managed provider explicitly by DS? Is there any ManagedProvider that does not implement the interface of the given ProviderClazz?

maggu2810:
> I have one more question regarding this PR... why do we need to inject the managed provider explicitly by DS?

@tomhoefer This PR does not change the managed provider handling at all. It has been injected before by DS and this mechanism has not been changed.
There could be multiple provders and one managed provider. This has also not been changed.

PR #2138, opened 2016-09-07
use rule DTO for rule events

maggu2810:
Fixes: https://github.com/eclipse/smarthome/issues/2137

This PR has not been tested by myself but I assume @aounhaider1 you can cherry-pick the PR and report it.

PR #2145, opened 2016-09-08
fix failing tests (caused by usage of rule DTO in events)

maggu2810:
Related to: #2138

I have missed to run the tests. Mea culpa...

The formatter changed most of the stuff only the four changes of UID to uid (because different member names of RuleDTO and Rule) has been done by myself.

A build (with all tests) is currently running on my machine. I will report if it succeeds.


maggu2810:
Succeeds on my machine and on Travis CI.

kaikreuzer:
Thanks!

PR #2156, opened 2016-09-12
WIP: Add test to ensure that group functions persist

mkhl:
With 1cf093e we can create group items with group functions via the REST API, but the group functions are not currently persisted: ManagedItemProvider.PersistedItem is missing affordances to store group functions and ManagedItemProvider is missing code making use of them.

This change adds a test to ensure that group functions are saved and restored correctly. The test is currently failing but might benefit the future implementer.

Signed-off-by: Martin Kühl martin.kuehl@gmail.com


kaikreuzer:
https://github.com/eclipse/smarthome/pull/2309 implemented this feature - would you want to rebase your test on the latest master and see whether it succeeds?

mkhl:
Done.

The tests succeed, and probably don’t add much value to the ones added in #2309. I’ll happily close this PR if you think they don’t.

I needed to adapt the tests slightly, passing `null` as the second argument to the `GroupItem` constructor lets them fail in non-obvious ways (the groups are restored with the default function), but I’ not sure that that’s a situation one would encounter “in the wild”.

kaikreuzer:
Thanks, full ok to merge!

maggu2810:
I removed the "WIP" in front of the topic :wink:

Issue #2150, opened 2016-09-09
Paper UI: state for r/o items with options broken


maggu2810:
HEAD: 9a7ea3237656cd4380c7ac99a8284a8c30a53e4d

I have a thing that is using a channel of this type
```xml
	<channel-type id="opt">
		<item-type>String</item-type>
		<label>Options</label>
		<state readOnly="true">
			<options>
				<option value="A">alpha</option>
				<option value="B">bravo</option>
				<option value="C">charlie</option>
			</options>
		</state>
	</channel-type>
```

I added a thing and have an item (automatically created) connected to the channel.

Previously the Paper UI shows me 
* "alpha" if the item state is A
* "beta" if the item state is B
* "charlie" if the item state is C

Now I see nothing value in the Paper UI.

```text
osgi> smarthome items list test_test_thing_d53e92d5_opt
test_test_thing_d53e92d5_opt (Type=StringItem, State=NULL, Label=Options, Category=null)
```
```text
osgi> smarthome update test_test_thing_d53e92d5_opt A
Update has been sent successfully.
osgi> 16:50:57.363 INFO  smarthome.event.ItemStateEvent[:43] - test_test_thing_d53e92d5_opt updated to A
16:50:57.364 DEBUG o.e.s.c.t.i.ThingManager[:455] - Delegating update 'A' for item 'test_test_thing_d53e92d5_opt' to handler for channel 'test:test-thing:d53e92d5:opt'
16:50:57.368 INFO  s.event.ItemStateChangedEvent[:43] - test_test_thing_d53e92d5_opt changed from NULL to A
```
```text
osgi> smarthome items list test_test_thing_d53e92d5_opt
test_test_thing_d53e92d5_opt (Type=StringItem, State=A, Label=Options, Category=null)
```

The Paper UI still shows only the "Options" line, but without a value.

maggu2810:
I need to add, that it is broken for read only channels but working for the read / writable ones.

maggu2810:
I assume it has been broken by 81a0fdf0e47c0e1d4ebca22227a36a94bb56f963

PR #2158, opened 2016-09-12
Show state of readonly string item

aounhaider1:
fixes https://github.com/eclipse/smarthome/issues/2150

### DIFF
--- a/extensions/ui/org.eclipse.smarthome.ui.paper/web-src/partials/control.html
+++ b/extensions/ui/org.eclipse.smarthome.ui.paper/web-src/partials/control.html
@@ -114,7 +114,7 @@
 														<div ng-show="isReadOnly(item)">
 															<h2 class="state" ng-if="isReadOnly(item)">{{getStateText(item)}}</h2>
 														</div>
-														<div layout="row" layout-align="end center" ng-if="!isReadOnly(item) && !isOptionList(item)">
+														<div layout="row" layout-align="end center" ng-if="!isReadOnly(item) && !isOptionList(item) && !shouldRenderSlider(item)">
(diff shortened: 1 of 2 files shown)


maggu2810:
I don't think that solves the option list state correctly, because it shows the state and not the respective option.

I tried a modified get state text method, but I don't know how to solve it the best way (see: https://github.com/maggu2810/smarthome/commit/862b1fbc1c9533e6471af99ca8f3c6c573cfaa9a).

WDYT @aounhaider1

aounhaider1:
So you mean if a read only item has options then its state should be used as a key to get the option from the option list? Right?

maggu2810:
I think so.
Isn't that the same what is done for read / writable items with options list?

aounhaider1:
Your logic for getting the state text is correct. The html conditions from this commit and js logic from yours should solve the issue.

maggu2810:
@aounhaider1 You can use that code in your PR. I don't create a PR for that JS and HTML stuff because I have no knowledge about it.

extensions/ui/org.eclipse.smarthome.ui.paper/web-src/partials/control.html
```diff
@@ -227,8 +227,8 @@ <h2 class="state edit" ng-if="!editMode" ng-click="editState()">{{formattedState
 												</div>
 
 												<!-- default text output -->
-												<div layout="row" layout-align="start center" ng-show="!isOptionList(item)">
-													<div ng-show="isReadOnly(item)" layout="row" layout-align="start center" flex>
+												<div layout="row" layout-align="start center" ng-show="isReadOnly(item) || (!isReadOnyl(item) && !isOptionList(item))">
```
maggu2810: Does this work for you? I assume it should not work because of the typo in the function name: `isReadOnyl`

Issue #2162, opened 2016-09-13
javasound : NPE in AudioPlayer:convertAudioFormat()


kgoderis:
At various places in the codebase we create AudioFormats with a lot of null values for things like frequency (for example JavaSoundAudioSink line #45). Feeding those AudioFormat objects to the method mentioned above, in order to convert them to an javax.sound.sampled.audioformat will throw an NPE in lines 113 and beyond. In my use case, feeding a .wav file (from the filesystem) with an AudioFormat with a bunch of null's for frequency,... , to the javasound sink throws the NPE as it goes through the convertAudioFormat() method

kaikreuzer:
Having null values was only meant for format matching, when a service declares what it supports. A concrete AudioFormat (of a stream) was supposed to have all meta-data correctly filled in.

But I see your issue: When reading WAV or mp3 from the file system, you stand no chance to figure out their bitrate etc. So yes, the code should be adapted to also be able to cope with incomplete format information.

PR #2163, opened 2016-09-13
Sonos audiosink, voice improvements, jlzoom mp3 support

kgoderis:
Not formally signed off as this needs to be played around. It contains AudioSink support for Sonos through a dedicated audio servlet, playback of mp3 files through the jlzoom player, as well as console command extensions to list sources and sinks, and to play a file to a given sink. Playback of .wav files to javasound still has a bug (see https://github.com/eclipse/smarthome/issues/2162), for the rest it is more or less behaving as it should. Still to be done : add support for playSound() and similar commands to Rules

Just give it a try with "smarthome voice say javasound hello world" on the console, or "smarthome sound play piano.wav RINCON_000E58D6F1A601400" (note: I should switch around the order of the arguments for the play command)


kgoderis:
Oh, and @kaikreuzer , the bundles will need shuffling based on what you want to have where. For now I have put everything under .io.sound to make things easy

kaikreuzer:
Thanks! It will also need some shuffling towards OH2, since the jl library cannot be in ESH...
I'll have a look as soon as I find time, but it will probably take a moment, so please allow me some time.

kgoderis:
//TODO : Move to thread pool instead of threads, make sound playback as sync as possible between Sinks
//TODO : Extend AudioSink API with mechanism to set/get volume
//TODO : Introduce wildcards in <sink>
//TODO : play().... in Rules
//TODO : Improve lifetime management of temporary files

kgoderis:
@kaikreuzer question :would modifications in Audio.java in the org.openhab.io.multimedia.actions package in the org.openhab.core.compat1x bundle, that is part of the openhab-core repo, break anything with respect to OH1? e.g. anything that is in compat1x can be changed solely in function of ESH and OH2 APIs?

kaikreuzer:
No, it wouldn't break anything in OH1. But I would actually hope that we can completely remove the Audio action from the compat bundle and rather have that part of ESH...

kgoderis:
Ok - We are aligned then;  I presume it only means re-implementing that AudioActionService.java stuff as a org.eclipse.smarthome.model.script.engine.action.ActionService

kaikreuzer:
Yep, I guess that would be the right place.

bundles/core/org.eclipse.smarthome.core.voice/src/main/java/org/eclipse/smarthome/core/voice/internal/VoiceConsoleCommandExtension.java
```diff
@@ -44,7 +44,8 @@ public VoiceConsoleCommandExtension() {
 
     @Override
     public List<String> getUsages() {
-        return Arrays.asList(new String[] { buildCommandUsage(SUBCMD_SAY + " <text>", "speaks a text"),
+        return Arrays.asList(new String[] {
+                buildCommandUsage(SUBCMD_SAY + " <sink> <text>", "speaks a text through the given audio sink"),
```
kaikreuzer: This makes the command line usage pretty awkward - I intentionally did it this way to keep it simple. The user can anyhow configure his defaultSink, which is used by the console commands.

kgoderis: mmhm.. I tend not follow that as I have a scenario in mind whereby I want to redirect say output to different sinks (read: sonos zone players in different rooms). Maybe it is through for the console, but definitely not for ActionService.... Let's see.

kaikreuzer: No, that's clear, for the rule action it is definitively required to chose a sink!

kgoderis: Would keep it for now, it is useful for testing the code

kaikreuzer:
Good news, I got that working nicely: https://twitter.com/kaikreuzer/status/776879796357853185

I had to do some adaptions and other fixes though.
Still needs quite some refactoring to get the parts at the right place.
Shall I deal with that? Or shall we discuss what needs to be done and you do the changes?
Note that I am at JavaOne the next week, so I won't have much availability...

kgoderis:
No - pls go ahead as you see fit. I hope you did catch the latest code as I did some bug fixing yesterday/today. Maybe you can put things into the right place, I am happy to incorporate your fixes and in line with that I can do a further review of the code whilst you are at JavaOne?

kaikreuzer:
I think I won't touch my local code for the next week because it could easily destroy my demo setup, which is highly fragile - so don't expect me to provide any changes the next days. Feel free to further add fixes to the code; once I start with the refactoring, I will then take your latest version and ask you for a code freeze :-)

kgoderis:
Ok - Anything critical you bumped into that did not work at your side? I only did some simple testing on the console and in some Rules, and they worked, apart from some random Timeout I sometime get  in the ThreadFactory

kaikreuzer:
No, nothing serious, really.

kgoderis:
//TODO : the Sonos zoneplayer can not play .wav generated by the MacTTS service because of  a mismatch in sample frequency

kaikreuzer:
> the Sonos zoneplayer can not play .wav generated by the MacTTS service because of a mismatch in sample frequency

Right, I fixed that by changing the frequency to 44100, which works. But the negociation of the right AudioFormat should indeed be improved here.

kaikreuzer:
FTR, I have pulled this branch now and will start with the refactoring. I am closing this PR and will come up with some new ones that include different changes.

kaikreuzer:
FTR, I just created two PRs with my first refactorings: https://github.com/eclipse/smarthome/pull/2225 and https://github.com/eclipse/smarthome/pull/2224

Still to do:
- add rule actions
- create a javazoom sound bundle for openHAB2 (cannot be in ESH due to its license)
- add audioservlet to core.audio
- adapt Sonos binding

kgoderis:
Ok - I understand you want to generalise the servlet bit. As I could not immediately find a way to debug the filesystem part in the IDE, I just used a folder in USERDATA, you might want to change that.  I leave it in your good hands ;-)

kaikreuzer:
Right, the servlet will need some generalization, so that it can be used by different bindings in the end.

Btw, for the "AudioSink" support within ThingHandlers, I am thinking about formalizing this as a new Channel kind (similar to the Trigger channel kind). We would then have "AudioLinks" with it, which would act like an alias for the sinks (and at the same time leave it to the user, which sinks of the Thing should be potentially used and which not). So within rules, you could then refer to your sink as "sonos1" instead of "RINCON3442354:abc" (or however this looked like today). Would that make sense?

kgoderis:
Yes. Obviously, not all Things have audio capabilities, so it can be a bit awkward at first. As long as it can properly be defined through the DSL I am fine....

kaikreuzer:
@maggu2810 I would actually like to use a chromecast binding for an initial PoC of the concept described in https://github.com/eclipse/smarthome/pull/2163#issuecomment-250352838. As you already worked on one: Is this in a state that you would create a PR for it? Or could I at least build upon it and make it round?

maggu2810:
As it was not possible for me to work with the upstream project, I forked the library and have done the necessary changed there:
* https://github.com/maggu2810/shk/tree/master/bundles/shk-library-chromecast-api

The binding is pretty simple and could be found here:
* https://github.com/maggu2810/shk/tree/master/bundles/shk-binding-chromecast

A Karaf Feature could be found here:
* https://github.com/maggu2810/shk/blob/master/features/shk-base-wo-esh/src/main/feature/feature.xml#L26

The development has been stalled caused by the Chromecast Library. I don't think I have time to update the library myself on new Chromcast SDK / API changes and the changes necessary for a reliable binding implementation has not been merged upstream...

The current state implements this channels:
https://github.com/maggu2810/shk/blob/master/bundles/shk-binding-chromecast/src/main/resources/ESH-INF/thing/channel-type-all.xml

I assume "play uri", "set volume" and "play / pause" should be enough for an PoC of the audio stuff.

kaikreuzer:
Thanks @maggu2810, I'll have a look and see where I can go from there.

PR #2175, opened 2016-09-15
added more commands to German voice interpretation and made interpret…

kaikreuzer:
…er more robust

### DIFF
--- a/bundles/core/org.eclipse.smarthome.core.voice/src/main/java/org/eclipse/smarthome/core/voice/text/AbstractRuleBasedInterpreter.java
+++ b/bundles/core/org.eclipse.smarthome.core.voice/src/main/java/org/eclipse/smarthome/core/voice/text/AbstractRuleBasedInterpreter.java
@@ -26,6 +26,8 @@ import org.eclipse.smarthome.core.library.types.DecimalType;
 import org.eclipse.smarthome.core.library.types.StringType;
 import org.eclipse.smarthome.core.types.Command;
 import org.eclipse.smarthome.core.types.State;
+import org.slf4j.Logger;
+import org.slf4j.LoggerFactory;
 
 /**
  * A human language command interpretation service.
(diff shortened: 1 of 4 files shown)


bundles/core/org.eclipse.smarthome.core.voice/src/main/java/org/eclipse/smarthome/core/voice/text/AbstractRuleBasedInterpreter.java
```diff
@@ -480,10 +485,17 @@ protected String executeSingle(ResourceBundle language, String[] labelFragments,
                     if (oldState.equals(newState)) {
                         String template = language.getString("state_already_singular");
                         String cmdName = "state_" + command.toString().toLowerCase();
-                        String stateText = language.getString(cmdName);
+                        String stateText = null;
+                        try {
+                            stateText = language.getString(cmdName);
+                        } catch (Exception e) {
+                            stateText = language.getString("state_current");
```
maggu2810: For error a new const has been created above, for state_current the magic variable is used here.

maggu2810: Only one minor comment...

kaikreuzer: I did it the same way as the code looked like for the English version...

kaikreuzer: There are (already were) quite some other strings for which there is no constant defined in here - I did it similar to the state_already_* ones. But yes, all of these strings could be defined as constants throughout this class.

kaikreuzer: updated!

kaikreuzer:
You are also allowed to merge then ;-)

maggu2810:
The merge request builder (Travis CI) has not been finished till now.

kaikreuzer:
Ah, ok!

PR #2178, opened 2016-09-15
execute xtend-maven-plugin only for necessary artifacts

maggu2810:
The plugin is currently executed for every artifact.
This consumes build time and wastes the log.

### DIFF
--- a/bundles/designer/org.eclipse.smarthome.designer.core/pom.xml
+++ b/bundles/designer/org.eclipse.smarthome.designer.core/pom.xml
@@ -1,7 +1,7 @@
 <?xml version="1.0" encoding="UTF-8"?>
 <project xmlns:xsi="http://www.w3.org/2001/XMLSchema-instance"
-  xmlns="http://maven.apache.org/POM/4.0.0"
-  xsi:schemaLocation="http://maven.apache.org/POM/4.0.0 http://maven.apache.org/maven-v4_0_0.xsd">
+         xmlns="http://maven.apache.org/POM/4.0.0"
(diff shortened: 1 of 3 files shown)


kaikreuzer:
Weird, I would have assumed that it is just the other way round... I always thought that a "pluginManagement" entry does not have any effect and only provides a default configuration for a plugin...

kaikreuzer:
It is not a proof, but some indication that this change does not seem speed up the build:

screen shot 2016-09-15 at 17 16 45

maggu2810:
> I always thought that a "pluginManagement" entry does not have any effect and only provides a default configuration for a plugin...

That is correct.
But it has not been placed in pluginManagement only.
It has been also placed in the plugins section of build (without management).
So the plugin has been enabled for the root pom and all children.

maggu2810:
You are right, using Travis CI builds for comparison is no proof

* 21m 36s https://travis-ci.org/eclipse/smarthome/builds/160114784
* 42m 53s https://travis-ci.org/eclipse/smarthome/builds/159808377

I have written it wastes time, not that the build is significant faster.

I don't see any reason to execute the xtend plugin for non-xtend projects.

I have not measured if we win at least one second but if we can drop the plugin execution for a lot of artifacts, why not?

kaikreuzer:
Sorry, I didn't see [this change](https://github.com/eclipse/smarthome/pull/2178/files#diff-600376dffeb79835ede4a0b285078036L136) in your PR before, so I wondered how this PR could actually achieve its intension to NOT execute the plugin. Now it is clear (and yes, I agree that it is not about saving time, but about not executing stuff that is not relevant or could even cause issues).

PR #2187, opened 2016-09-19
avoid NPE in LIFX handler if no url info is set yet

kaikreuzer:


kgoderis:
I guess you meant Sonos Zone Player instead of Lifx ;-)

kaikreuzer:
Haha, I probably meant "some binding done by Karel" ;-)

PR #2188, opened 2016-09-19
Improve logging of ConfigStatus events

kaikreuzer:
Adapted implementation to be similar to ThingStatus events, which enables a proper toString() implementation (and thus improves event log files).


kaikreuzer:
@tomhoefer Could you please check, if this change is ok from your perspective?

tomhoefer:
ok for me

Issue #2180, opened 2016-09-16
Modify org.eclipse.smarthome.automation.core to use DS


vilchev:
By now automation.core module was activated through BundleActivator. Dependencies were tracked by implementing ServiceTrackers.

Extract ModuleTypeRegistry, RuleEventFactory, RuleRegistry, TemplateRegistry as declared components

vilchev:
RuleRegistryImpl is tracking **TemplateProviders** via ServiceTracker
  
```
@Override
public void modifiedService(ServiceReference reference, Object service) {
  TemplateProvider provider = (TemplateProvider) bc.getService(reference);
  templateUpdated(provider.getTemplates(null));
}
```


Here implementation uses **modifiedService** to be notified when the TemplateProvider service has been modified (templates from that TemplateProvider have been edited).

Along with **bind**, **unbind** methods from **Event strategy** ( one of the strategies for accessing services ) there is **updated** which can be used in this particular case:

> 
> • bind – The bind method is called to bind a new service to the component that matches the
> selection criteria. If the policy is dynamic then the bind method of a replacement service can be
> called before its corresponding unbind method.
> • updated – The updated method is called when the service properties of a bound services are mod-
> ified and the resulting properties do not cause the service to become unbound because it is no
> longer selected by the target filter.
> • unbind – The unbind method is called when the SCR needs to unbind the service.



I can find that **updated** is supported in OSGi R4.3, but cannot find info for it in OSGi R4.2.

danchom:
As I remember, we have changed to use modified service method because the providers (especially script provider) to have opportunity to update supported module types (same with templates). I'm not sure if this functionality is still needed.  For me bind and unbind is enough to introduce module types and templates from providers. But if the  supper of OSGi R4.2 is not requirement, the updated can be used too.
Kai what do you think?

vilchev:
In OSGi R4.2

**Bound Service Replacement**

> If an active component configuration has a dynamic reference with unary
> cardinality and the bound service is modified or unregistered and ceases to
> be a target service, SCR must attempt to replace the bound service with a
> new target service. SCR must first bind a replacement target service and
> then unbind the outgoing service. If the dynamic reference has a mandatory
> cardinality and no replacement target service is available, the component
> configuration must be deactivated because the cardinality constraints will
> be violated.
> 
> If a component configuration has a static reference and a bound service is
> modified or unregistered and ceases to be a target service, SCR must deacti-
[...]

PR #2194, opened 2016-09-20
Refactored automation.core & adapted to use Declarative Services

vilchev:
Changes Made:

0. RuleManager , TemplateManager, ModuleTypeManger wrapper classes (over RuleRegistry, TemplateRegistry, ModuleTypeRegistry) are now removed. Directly refer to registries where needed.

1. RuleRegistry, RuleEventFactory, ModuleTypeRegistry, TemplateRegistry  extracted as DS componentes

2. ManagedRuleProvider is now component with dependency 1..1 with storage service

3. RuleRegistry now is not dependent on StorageService and ManagedProvider (rule registry becomes active and stays active regardless of StorageService and ManagedProvider availability)
When storage is not available -> setEnabled throws IllegalStateException (cannot persist disabled state)
If rules are added before storage is available -> rules are added with disabled state.
When storage becomes available -> all rules that are not in the disabled rule storage are set back to enabled.

4. When ManagedRuleProvider is removed (not available) rules from it are now removed from engine.

5. Fixed bug in integration tests

Signed-off-by: Vasil Ilchev <v.ilchev@prosyst.bg>


bundles/automation/org.eclipse.smarthome.automation.core.test/src/test/java/org/eclipse/smarthome/automation/core/internal/RuleEngineTest.java
```diff
@@ -41,18 +41,19 @@
 
     Logger log = LoggerFactory.getLogger(RuleEngineTest.class);
 
-    private RuleEngine createRuleEngine() {
+    private RuleEngine createRuleEngine() throws Exception {
```
maggu2810: I comment this only here on the first occurrence (as it is used multiple times):
As it is only a test I don't care very much, but I don't think it is good practice to throw the type "Exception".
There are so many different exceptions that normally are not handled all the same way.

But if it is only for testing...

bundles/automation/org.eclipse.smarthome.automation.core/META-INF/MANIFEST.MF
```diff
@@ -17,7 +17,8 @@ Import-Package:
  org.eclipse.smarthome.core.storage,
  org.osgi.framework,
  org.osgi.service.cm,
+ org.osgi.service.component;version="1.2.2",
```
maggu2810: https://www.eclipse.org/smarthome/documentation/development/guidelines.html
"The manifest must not have any version constraint on package imports, unless this is thoughtfully added. Note that Eclipse automatically adds these constraints based on the version in the target platform, which might be too high in many cases."

So, if you know that this bundle does not work using a previous version (e.g. "1.2.1") of the package, I am fine with the version constraint.
If the constraint has been added because the target version uses 1.2.2, you should remove that / replace that with the correct minimum version.

bundles/automation/org.eclipse.smarthome.automation.core/build.properties
```diff
@@ -1,5 +1,6 @@
 output.. = target/classes/
 bin.includes = META-INF/,\
                .,\
+			   OSGI-INF/,\
```
maggu2810: Are there mixed spaces and tabs in that file?

maggu2810:
I stop the review as it seems you still change the commits (e.g. force push) of this PR.

vilchev:
Just finished the PR (fixed current remarks by @maggu2810 ).

vilchev:
Fixes: #2180

vilchev:
Travis build failed:

> [ERROR] Failed to resolve target definition /home/travis/build/eclipse/smarthome/targetplatform/smarthome.target: Failed to load p2 metadata repository from location http://download.eclipse.org/eclipse/updates/4.4: Unable to read repository at http://download.eclipse.org/eclipse/updates/4.4. Communication with repository at http://download.eclipse.org/eclipse/updates/4.4/bug445122 failed. Read timed out -> [Help 1]

maggu2810:
I have seen this two (Hudson failed, Travis failed, all fails).
It is pretty annoying that the eclipse repos are down / unreachable from time to time.

vilchev:
It seems that there is some synchronization issue.. Travis builds are always failing:

> [INFO] Eclipse SmartHome Automation Integration Tests .... FAILURE [51.096s]

Have tested heavily on 2 of our machines - never reproduced that test failing..

Keep researching what the problem could be.

maggu2810:
It's not always reproducible on every machine setup (see e.g. https://github.com/eclipse/smarthome/issues/2057#issuecomment-243779763) but normally there is an error.

vilchev:
Build(with integration tests) is stable on Travis now too. This PR is now ready for review.

bundles/automation/org.eclipse.smarthome.automation.core/src/main/java/org/eclipse/smarthome/automation/core/internal/RuleRegistryImpl.java
```diff
@@ -91,87 +91,166 @@
...
 
     @Override
-    protected void setManagedProvider(ManagedProvider<Rule, String> provider) {
-        super.setManagedProvider(provider);
-        logger.info("Rule Managed Provider: {} is added.", provider);
+    protected void removeProvider(Provider<Rule> provider) {
+        LOGGER.info("Rule provider: {} is removed.", provider);
+        if (provider instanceof ManagedRuleProvider) {
+            if (ruleRegistryReg != null) {
+                ruleRegistryReg.unregister();
```
vilchev: Old implementation was dependent on ManagedRuleProvider (Storage service is available)
I saw that AbstractRegistry handles situtation on **add/updated/remove** operations when ManagedProvider has **null** value (e.g. throws runtime exception).


Do we need this dependency here?

bundles/automation/org.eclipse.smarthome.automation.core/src/main/java/org/eclipse/smarthome/automation/core/internal/RuleRegistryImpl.java
```diff
@@ -91,87 +91,166 @@
...
+        super.unsetEventPublisher(eventPublisher);
+    }
+
     @Override
     protected void addProvider(Provider<Rule> provider) {
-        logger.info("Rule provider: {} is added.", provider);
+        LOGGER.info("Rule provider: {} is added.", provider);
         super.addProvider(provider);
+        if (provider instanceof ManagedRuleProvider) {
+            ruleRegistryReg = bundleContext.registerService(RuleRegistry.class.getName(), this, null);
```
vilchev: Old impl was registering the **RuleRegistry** only when **ManagedRuleProvider**  Do we really need dependency on ManagedRuleProvider as AbstractRegistry handles situations when managed provider has **null** value?

bundles/automation/org.eclipse.smarthome.automation.core/src/main/java/org/eclipse/smarthome/automation/core/internal/RuleRegistryImpl.java
danchom: The RuleRegistry must not be registered if the managed provider is not available. Otherwise a rule from other provider with same uid as the ones from managed provider can be loaded into rule engine and all changes stored in the managed provider will be not visible (the rule from managed provider will not be available at all). For this reason the RuleEngine first has to be  initialized with rules from managed provider and after then it is registered and give change to other providers to make make add their rules in the RuleEngine.
May be the RuleRegistry must not be unregistered when managed provider is disappeared, because it can return available rules, only update command will throw exception

vilchev: Actually old implementation didn't unregister RuleRegistry when **ManagedRuleProvider** is removed. I thought that if **registration** of  **RuleRegistry** is dependent on **ManagedRuleProvider** - **unregister** should be dependent on **ManagedRuleProvider** availability too.

Now I rather think that this was a feature than a bug.

danchom: Yes I think so.

danchom:
LGTM

bundles/automation/org.eclipse.smarthome.automation.core/src/main/java/org/eclipse/smarthome/automation/core/internal/RuleRegistryImpl.java
```diff
@@ -91,87 +91,162 @@
...
     }
 
     @Override
-    protected void removeProvider(Provider<Rule> provider) {
-        logger.info("Rule provider: {} is removed.", provider);
-        super.removeProvider(provider);
+    protected void setManagedProvider(ManagedProvider<Rule, String> managedProvider) {
+        LOGGER.info("Rule Managed Provider: {} is added.", managedProvider);
+        super.setManagedProvider(managedProvider);
+        this.managedRuleProviderReg = bundleContext.registerService(RuleProvider.class.getName(), this.managedProvider,
```
kaikreuzer: What is this good for? If the goal of this PR is a refactoring to DS, `bundleContext.registerService` should imho nowhere be used anymore.

bundles/automation/org.eclipse.smarthome.automation.core/src/main/java/org/eclipse/smarthome/automation/core/internal/RuleRegistryImpl.java
```diff
@@ -91,87 +91,162 @@
 public class RuleRegistryImpl extends AbstractRegistry<Rule, String, RuleProvider>
         implements RuleRegistry, StatusInfoCallback {
 
+    private static final String SOURCE = RuleRegistryImpl.class.getSimpleName();
+    private static final Logger LOGGER = LoggerFactory.getLogger(RuleRegistryImpl.class.getName());
```
vilchev: Ok, agree

bundles/automation/org.eclipse.smarthome.automation.core/src/main/java/org/eclipse/smarthome/automation/core/internal/RuleRegistryImpl.java
```diff
@@ -91,87 +91,162 @@
...
+     * Unbind the {@link TemplateRegistry} service - called from DS.
+     *
+     * @param templateRegistry templateRegistry service.
+     */
+    protected void unsetTemplateRegistry(TemplateRegistry templateRegistry) {
+        this.templateRegistry = null;
+    }
+
+    protected void setStorageService(StorageService storageService) {
+        // this.storageService = storageService;
```
danchom: Kai, we have following dependencies:
1) ManagedRuleProvider  has 1:1 dependency StorageService, but not 1:1 to RuleRegistry,
2) RuleRegistry waits ManagedRuleProvider  to register itself (because the rules from the ManagedProvider have priority then other providers)
3) RuleRegistry depends only to ModuleType registry and Template registry. It starts RuleEngine. 
The RuleEngine has to work without RuleRegistry.   These means when the ManagedService is disappeared the RuleEngine has to continue to work. I also think the RuleRegistry must not disappeared when managed provider goes off (you can list rules from other providers)

kaikreuzer: 1) Ok, RuleRegistry can have a 0:1 dependency on ManagedRuleProvider.
2) It it is a static dependency, the RuleRegistry can be sure to have the ManagedRuleProvider injected before activate() is called - so it can deal with the priority.
3) "It starts RuleEngine." -> imho, it shouldn't. The RuleEngine could be a service on its own, having a dependency on the RuleRegistry.
4) Then it definitely should be its on DS component and not be started by the RuleRegistry.

danchom: 4) I'm agree that the RuleRegistry has to be separate component, but I think it should't have dependency to RuleRegistry. For me the RuleRegistry is just an interface through which you can add, update, remove or list rules available in the RuleEngine. The RuleEngine must continue to work (execute available rules) without this interface. For example the alarm system must not be switch off just because the RuleRegistry is disappeared.

kaikreuzer: 3+4) Why is there the necessity to have RuleEngine be independent of RuleRegistry? For me, the RuleRegistry is merely an aggregation service for rules - and I would expect ALL rules to be registered there (i.e. nothing to be passed directly into the RuleEngine). The RuleEngine can have a 1:1 dependency on the RuleRegistry. Disappearing of the RuleRegistry should simply be the same as if the RuleEngine itself disappears, in which case rules aren't executed anymore either.

danchom: The RuleRegistry is using RuleEngine in its implementation. How the RuleRegistry will get the RuleEngine component without public interface? For me the RuleEngine must not be component at all when having 1:1 relation between RuleRegistry and RuleEngine.
RuleRegistry must have dynamic policy to ManagedProvider. (RuleRegistry must not be deactivated when the storage is disappeared)
Our proposal is that:
1) RuleRegistry to be a component
2) RuleRegistry has dynamic relation to ManageProvider 
3) RuleRegistry creates a RuleEngine (internally, it will not be a component - it is part of the RuleRegistry and they will have 1:1 dependency)
There is one problem that the RuleRegistry can be available without ManagedProvider and all RuleRegistry.add method will throws exception. The applications have to listen for ManagedRuleProvider instead of RuleRegistry to add (update, remove) rules.

danchom: Kay, do you have some pattern for relation between other registries and their managed providers?
Are they 1:1? As I see in the ThinkRegistry it has 0:1 dynamic relation to ManagedThingProvider.
This means if ManagedThingProvider is disappeared the ThinkRegistry continues to work, but you can't add,remove and update thinks. Is this desired behavior in the rules? As I remember there was a reported problem that if you want to add rule you have to wait for two services RuleRegistry and Managed provider (which was too complex) and then start the add operation.
Isn't be easier to have 1:1 dependency to managed provider because the managed provider depends only to storage service and it is a system service which shouldn't disappeared frequently?

kaikreuzer: > Kay, do you have some pattern for relation between other registries and their managed providers?

In general, I regard the registry to be a simple "aggregation" of providers, which might come and go (and the registry should not be impacted by this). Imho, it is perfectly fine to have only "read-only" providers being present (and not the managed provider), if your system is not meant to be "interactive". In this case, the add/update/remove functions will simply throw an IllegalStateException ([as written in the JavaDoc](https://github.com/eclipse/smarthome/blob/6dd1df77136a70ec55da6746be05bf65b78ba15b/bundles/core/org.eclipse.smarthome.core/src/main/java/org/eclipse/smarthome/core/common/registry/Registry.java#L62)). I think this behavior makes sense for Things and Rules likewise.

But yes, it indeed brings up the question how clients can know, when (and if) they can actually add something to the registry. I wonder if we should allow them to directly refer to the ManagedProvider in such a case?

danchom: Ok we will implement relation between RuleRegistry and ManagedRuleProvider to be 0:1 dynamic relation

vilchev:
Mistery of failing tests only on remote is now Solved. Build is now stable.

The problem was mixture of:

1. refactoring of core module has fixed existing bug


 ```
public void removedService(ServiceReference reference, Object service) {
 ....
 else if (service instanceof RuleProvider) {
                    if (ruleRegistry != null) {
                        RuleProvider rp = (RuleProvider) service;
                        if (rp instanceof ManagedRuleProvider) { -- rules from managedRuleProvider were not removed from rule engine
                            ruleRegistry.removeManagedProvider((ManagedProvider<Rule, String>) rp);
                        } else {
                            ruleRegistry.removeProvider(rp);
                        }
                    }
```

Now this is handled by AbstractRegistry -> rules from ManagedRuleProvider are removed when ManagedProvider is removed

2. integration test includes 3 test classes and packs json resources with rules(test was awlays failing because of missing rules uploaded from json resources)

AutomationIntegrationJsonTest
AutomationIntegrationTest
HostFragmentSupportTest

Two of them (AutomationIntegrationJsonTest & AutomationIntegrationTest) are registering a static reference of storage

    public static def VolatileStorageService VOLATILE_STORAGE_SERVICE = new VolatileStorageService()//keep storage with rules imported from json files

The third onw (HostFragmentSupportTest) - creates new instance and register it as storage service

3. Order of tests classes run are reversed - local vs remote 

local order
1. AutomationIntegrationJsonTest (uses static storage) 
2. AutomationIntegrationTest (uses static storage)
3. HostFragmentSupportTest (uses new instance)

remote order 
1. HostFragmentSupportTest (uses new instance)
2. AutomationIntegrationTest (uses static storage)
3. AutomationIntegrationJsonTest (uses static storage) 


3.1 Parser of json rules parses rules and adds them through ManagedProvider.. This is done only once (when bundle was resolved) - remotely rules were added in storage of HostFragmentSupportTest (which was not shared among the other 2 test classes) because it was the first testclass run that registers storage service from that bundle  - after removing once ManagedProvider now rules are removed from rule engine too - next time(AutomationIntegerationTest has started) managedprovider is up - it faces empty storage  - no rules from json


---
HostFragmentSupportTest now uses(registers) the shared storage too.

 public static def VolatileStorageService VOLATILE_STORAGE_SERVICE = new VolatileStorageService()//keep storage with rules imported from json files

**Comment by vilchev** [comment]
Updated my initial commit comment with made changes in this PR

**Comment by danchom** [comment]
> RuleManager , TemplateManager, ModuleTypeManger are now removed. 

They are not needed anymore.

> RuleRegistry now is not dependent on StorageService and ManagedProvider 

RuleRegistry should have any dependency. It always has to list available rules, event they come from readOnly providers, or they are disabled because of missing of storage.

>When storage is not available -> setEnabled throws IllegalStateException (cannot persist disabled state)

Yes, this is right behavior, also the  IllegalStateException is thrown when the user tries to update readOnly rule (impl by AbstractRegistry). The rules from ManagedProvider can't be edit also in this case because the managed provider has 1:1 static dependence to storage and these rules will be removed when the storage is removed.  

> If rules are added before storage is available -> rules are added with disabled state.

This is the right behavior because the RuleEngine must not start disable rule when the system is restarted and the storage is not available.

> When storage becomes available -> all rules that are not in the disabled rule storage are set back to enabled.

At this moment it is safety for the RuleEngine to start the enabled rules.

> ManagedRuleProvider is now component with dependency 1..1 with storage service
The managed provider can't work without storage

> When ManagedRuleProvider is removed (not available) rules from it are now removed from engine.
It has to work as any other rule provider

LGTM

**Comment by danchom** [comment]
@kaikreuzer could you please check this PR because it is blocker for us.

**Comment by kaikreuzer** [comment]
Sorry, was on holiday last week and am at EclipseCon this week - will try to review as soon as I find time this week.

File: bundles/automation/org.eclipse.smarthome.automation.core/src/main/java/org/eclipse/smarthome/automation/core/internal/RuleRegistryImpl.java [hunk 9]
```diff
@@ -91,75 +90,148 @@
...
-                    }
+    /**
+     * Activates this component. Called from DS.
+     *
+     * @param componentContext this component context.
+     * @throws Exception
+     */
+    protected void activate(ComponentContext componentContext) throws Exception {
+        this.bundleContext = componentContext.getBundleContext();
+        this.ruleEngine = new RuleEngine(bundleContext, moduleTypeRegistry);
```
kaikreuzer: What is the reason that the RuleEngine is not realized as a DS component? I thought we had discussed that it should have a 1..1 relationship to the RuleRegistry? 
With the current code, you still have OSGi dependencies in RuleEngine and RuleRegistryImpl - I do not really see a necessity for this.

vilchev: RuleRegistryImpl uses RuleEngine in its implementation. If RuleEngine is Component - How could RuleRegistry retreive RuleEngine(without having some new Interface for the RuleEngine)?

kaikreuzer: If you consider RuleEngine as an internal implementation detail, you should not register it as a ManagedService, but rather have the RuleRegistry take a service configuration - this way, you would be able to remove the dependency on the bundle context and thus remove the OSGi references from the code.

vilchev: > you should not register it as a ManagedService, but rather have the RuleRegistry take a service configuration - this way, you would be able to remove the dependency on the bundle context and thus remove the OSGi references from the code.

Agree, if it is internal implementation we can remove its OSGi references and use the RuleRegistry as a proxy. If you are Ok with that I would update the PR

danchom: Kai, thanks for quick review. I want mention that the RuleEngine can't be separate component, because it works with RuntimeRules, which are not public rules. Even if we move conversion from rules to run-time once into the RuleEngine we have to introduce a new public interface which has to be used by the RuleRegistry to communicate with RuleEngine. Unfortunately this interface can be use to violate all the logic coming from the RuleRegistry (persist rule changes, not permit add, update and remove operation for rules coming from not managed providers and etc). The purpose of the RuleEngine is just to manage run-time rules (initialize/deinitalize, enable/disable and execute rules).

bundles/automation/org.eclipse.smarthome.automation.core/src/main/java/org/eclipse/smarthome/automation/core/internal/RuleRegistryImpl.java
```diff
@@ -89,77 +94,177 @@
...
+     */
+    protected void unsetTemplateRegistry(TemplateRegistry templateRegistry) {
+        this.templateRegistry = null;
+    }
 
-                });
-        templateProviderTracker.open();
+    protected void setStorageService(StorageService storageService) {
+        this.storageService = storageService;
+        setDisabledRuleStorage(
```
vilchev: Possible NPE for rule engine - Storage Service is dynamic dependency now - there is no guarantee setStorageService bind method will be called after activate method is called (in activate we instantiate ruleEngine). @kaikreuzer please do not merge yet

vilchev: With latest update of the PR possible NPE was fixed. PR is ready for a review

danchom:
LGTM

bundles/automation/org.eclipse.smarthome.automation.core/src/main/java/org/eclipse/smarthome/automation/core/internal/RuleRegistryImpl.java
```diff
@@ -89,77 +94,177 @@
...
+                storageService.<Boolean>getStorage("automation_rules_disabled", this.getClass().getClassLoader()));
```
kaikreuzer: could we make a constant out of this string?

bundles/automation/org.eclipse.smarthome.automation.core/src/main/java/org/eclipse/smarthome/automation/core/internal/RuleRegistryImpl.java
```diff
@@ -89,77 +94,177 @@
  * @author Benedikt Niehues - added events for rules
  */
 public class RuleRegistryImpl extends AbstractRegistry<Rule, String, RuleProvider>
-        implements RuleRegistry, StatusInfoCallback {
+        implements RuleRegistry, StatusInfoCallback, ManagedService {
```
kaikreuzer: as suggested before, I would very much prefer to not implement ManagedService, but use the DS configuration mechanisms (passing them through activate() or modified() methods). Would that make sense?

vilchev: Ok. Now I got your initial idea and yes it makes sense to make use of the component configuration(of RuleRegistry) rather than impl ManagedService.

bundles/automation/org.eclipse.smarthome.automation.core/src/main/java/org/eclipse/smarthome/automation/core/internal/RuleRegistryImpl.java
vilchev: Yes, of course.

Issue #1569, opened 2016-05-25
Implement a file system TemplateProvider and ModuleTypeProvider


kaikreuzer:
For templates and composite module types it would be nice to be able to directly put the json files to a folder in the file system instead of having to deploy them in an OSGi bundle.

danchom:
Implemented by #2070

PR #2070, opened 2016-08-30
Implement a file system TemplateProvider and ModuleTypeProvider

adimova:
implements issue #1569
replace PR https://github.com/eclipse/smarthome/pull/1648
Signed-off-by: Ana Dimova a.dimova@prosyst.bg

PR #2196, opened 2016-09-20
implements issue #1569

adimova:


adimova:
this PR fixes the comments in #2070

maggu2810:
What's the reason to create new PRs?
I think this makes it more complicated to see if all review points has been done etc.

kaikreuzer:
Btw, @adimova, I would prefer if you could use a "speaking" tile for the PRs to make clear what it does. You can then add a comment like "fixes #xyz" or "resolves #xyz" to it, which will make the [auto close feature](https://help.github.com/articles/closing-issues-via-commit-messages/) kick in.

adimova:
PR #2070 is recovered. So this PR can be closed

maggu2810:
Thanks @adimova

PR #2199, opened 2016-09-20
Make thingTypeDTO's lighter

### LINKED ISSUE #1729: Introduce an index for thing type definitions
Analyzing performance issues on the Paper UI lately, I noticed that the REST API response on /thing-types is about 1.5MB with Z-Wave binding installed and 62KB without.

This clearly indicates that there are some bindings with a huge number of thing types and that we need to optimize the way to access this information.

A while ago, @cdjackson already predicted this and suggested a kind of index of thing types within the bindings. This index should only have the most relevant information (like id, label, description, what else?) and we should have a REST url to access only this information. If the full glory is required, this should be specifically requested for a (or a list of) given thing type UID(s).

Ideally, we would even generate this index during the build process, but besides optimizing the UI access, it will also reduce the startup time as the full XMLs should only be parsed upon request and not upfront anymore.

Mixajlo:
Make thingTypeDTO's lighter by not setting properties, configParameters and channels
Removed unused method
Fixes #1729 part 1

Signed-off-by: Miki Jankov <miki.jankov87@gmail.com>

### DIFF
--- a/bundles/io/org.eclipse.smarthome.io.rest.core/src/main/java/org/eclipse/smarthome/io/rest/core/thing/ThingTypeResource.java
+++ b/bundles/io/org.eclipse.smarthome.io.rest.core/src/main/java/org/eclipse/smarthome/io/rest/core/thing/ThingTypeResource.java
@@ -31,6 +31,8 @@ import org.eclipse.smarthome.config.core.dto.ConfigDescriptionParameterGroupDTO;
 import org.eclipse.smarthome.core.thing.ThingTypeUID;
 import org.eclipse.smarthome.core.thing.dto.ChannelDefinitionDTO;
 import org.eclipse.smarthome.core.thing.dto.ChannelGroupDefinitionDTO;
+import org.eclipse.smarthome.core.thing.dto.StrippedThingTypeDTO;
+import org.eclipse.smarthome.core.thing.dto.StrippedThingTypeDTOMapper;
 import org.eclipse.smarthome.core.thing.dto.ThingTypeDTO;
 import org.eclipse.smarthome.core.thing.type.BridgeType;
 import org.eclipse.smarthome.core.thing.type.ChannelDefinition;
@@ -60,6 +62,7 @@ import io.swagger.annotations.ApiResponses;
  * @author Chris Jackson - Added parameter groups, advanced, multipleLimit,
  *         limitToOptions
  * @author Yordan Zhelev - Added Swagger annotations
+ * @author Miki Jankov - Introducing StrippedThingTypeDTO
  */
 @Path(ThingTypeResource.PATH_THINGS_TYPES)
 @Api(value = ThingTypeResource.PATH_THINGS_TYPES)
@@ -91,13 +94,14 @@ public class ThingTypeResource implements RESTResource {
 
     @GET
     @Produces(MediaType.APPLICATION_JSON)
-    @ApiOperation(value = "Gets all available things types.", response = ThingTypeDTO.class, responseContainer = "Set")
+    @ApiOperation(value = "Gets all available thing types without config description, channels and properties.", response = StrippedThingTypeDTO.class, responseContainer = "Set")
     @ApiResponses(value = @ApiResponse(code = 200, message = "OK") )
     public Response getAll(
             @HeaderParam(HttpHeaders.ACCEPT_LANGUAGE) @ApiParam(value = HttpHeaders.ACCEPT_LANGUAGE) String language) {
         Locale locale = LocaleUtil.getLocale(language);
-        Set<ThingTypeDTO> thingTypeDTOs = convertToThingTypeDTOs(thingTypeRegistry.getThingTypes(locale), locale);
-        return Response.ok(thingTypeDTOs).build();
+        Set<StrippedThingTypeDTO> strippedThingTypeDTOs = convertToStrippedThingTypeDTOs(
+                thingTypeRegistry.getThingTypes(locale), locale);
+        return Response.ok(strippedThingTypeDTOs).build();
     }
 
     @GET
@@ -115,13 +119,6 @@ public class ThingTypeResource implements RESTResource {
         } else {
             return Response.noContent().build();
         }
-    }
-
-    public Set<ThingTypeDTO> getThingTypeDTOs(String bindingId, Locale locale) {
-
-        List<ThingType> thingTypes = thingTypeRegistry.getThingTypes(bindingId);
-        Set<ThingTypeDTO> thingTypeDTOs = convertToThingTypeDTOs(thingTypes, locale);
-        return thingTypeDTOs;
     }
 
     private ThingTypeDTO convertToThingTypeDTO(ThingType thingType, Locale locale) {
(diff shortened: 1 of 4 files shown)


bundles/io/org.eclipse.smarthome.io.rest.core/src/main/java/org/eclipse/smarthome/io/rest/core/thing/ThingTypeResource.java
```diff
@@ -91,7 +92,7 @@ protected void unsetConfigDescriptionRegistry(ConfigDescriptionRegistry configDe
 
     @GET
     @Produces(MediaType.APPLICATION_JSON)
-    @ApiOperation(value = "Gets all available things types.", response = ThingTypeDTO.class, responseContainer = "Set")
+    @ApiOperation(value = "Gets all available things types without config parameters, channels and properites.", response = ThingTypeDTO.class, responseContainer = "Set")
```
IOOOTAlan: There's a typo (`properites`)

Mixajlo: :(

Mixajlo: I will fix this, as soon as someone does a review, so I can do it in one commit

kaikreuzer: The Swagger documentation still claims that you are returning ThingTypeDTOs, which might be a bit misleading (technically, it is of course right, but the user might expect that the values are provided and not null).
I wonder if we should introduce something like "StrippedDTOs" in addition to our "EnrichedDTOs" that make it more explicit, which information is returned. wdyt?

maggu2810: Sounds reasonable.

Mixajlo: Agreed

kaikreuzer:
Thanks.
Just a general question for discussion: Are we ok that the stripped version is REST specific and thus resides in io.rest and not in core.thing.dto? The consequences are that they can not be used e.g. in events or other integrations (MQTT, whatever). Additionally, the "Full" DTO cannot extend the strippedDTO and thus one needs to add new potential properties to both classes. Is this ok? I am still a bit undecided on this...

bundles/io/org.eclipse.smarthome.io.rest.core/src/main/java/org/eclipse/smarthome/io/rest/core/thing/ThingTypeResource.java
```diff
@@ -211,19 +206,26 @@ private ThingTypeDTO convertToThingTypeDTO(ThingType thingType, Locale locale) {
...
         }
 
-        return thingTypeDTOs;
+        return strippedThingTypeDTOs;
     }
 
+    /*
+     * Convert thingType to StrippedDTO, which excludes configDescription, channels and properties
+     */
+    private StrippedThingTypeDTO convertToStrippedThingTypeDTO(ThingType thingType, Locale locale) {
```
kaikreuzer: We usally have "Mapper" classes for such purposes. I would prefer it over a method in this class here.

Mixajlo: > Are we ok that the stripped version is REST specific and thus resides in io.rest and not in core.thing.dto?

This means that all "EnrichedDTO" and "DTOMappers" should be moved to core.thing.dto together with "official" DTO's?
> Additionally, the "Full" DTO cannot extend the strippedDTO

This crossed my mind. I was first thinking for the ThingTypeDTO to extend the StrippedDTO and the Enriched to extend the "Full" DTO... But not the best design.

Mixajlo:
> Are we ok that the stripped version is REST specific and thus resides in io.rest and not in core.thing.dto?

This means that all "EnrichedDTO" and "DTOMappers" should be moved to core.thing.dto together with "official" DTO's?
> Additionally, the "Full" DTO cannot extend the strippedDTO

This crossed my mind. I was first thinking for the ThingTypeDTO to extend the StrippedDTO and the Enriched to extend the "Full" DTO... But not the best design.

kaikreuzer:
> This means that all "EnrichedDTO" and "DTOMappers" should be moved to core.thing.dto together with "official" DTO's?

No, the EnrichedDTOs are intentionally in the REST bundle, since they are only meant to enhance the DTO for the REST usage. So they will stay. My question was whether we claim the same for the stripped version: Are they really only relevant for REST or not?
I am e.g. a bit afraid that sooner or later we have too much traffic on the event bus (also SSE), if we always post the "full" entities on a change/update. It imho would make more sense to also use a stripped version there and if a client requires the full information, he would have to do a specific call to get it.

tomhoefer:
First of all in my opinion this change is not consistent. I would expect that a single element of the retrieved thing type collection matches the corresponding element if it is retrieved by ```getByUID```. 

I prefer to send the actual entities via the event bus and not the stripped ones. If the entity is too heavy to be sent via the event bus oftentimes then we should think about to refactor the entity, e.g. then it could contain a list of UIDs as parameters instead of the complete list of domain objects (in case of the thing type it could contain only a list of channel definition ids instead of the list of all corresponding channel definitions)

kaikreuzer:
> I would expect that a single element of the retrieved thing type collection matches the corresponding element if it is retrieved by getByUID.

Hm, I am not sure I understand your comment. This whole PR is about making a difference here, i.e. to have reduced information in the lists.

> I prefer to send the actual entities via the event bus and not the stripped ones.

Why, if the information is not needed and only blows up the required bandwidth? I would rather go the opposite way and say that in events it is often good enough to simply say "thing with id xyz has changed" without providing ANY further details - whoever is then interested, can request the current version. In 99% no one will be interested, the data is just trash.

> it could contain a list of UIDs as parameters instead of the complete list of domain objects

This does not work as you would force the UI to not only do a single REST request but potentially 100 in order to get all the required information. We are just discussing here how to best reduce the requests and traffic between the server and the UI...

maggu2810:
* Do we need Enriched...DTO in REST only (future use cases)
* Do we need Stripped...DTO in REST only (@kaikreuzer already pointed out e.g. MQTT)
* Do you think that there is / should be only one Stripped...DTO in the framework or is it possible we need different stripped down version at different places? 

If Enrichted...DTO should extent ...DTO and Stripped...DTO, the ...DTO needs to extend Stripped...DTO. So it cannot be placed in the rest package anymore.
* Do you see any problems to move Stripped...DTO and Enriched...DTO to the package ...DTO is placed?

kaikreuzer:
I actually thought to remember that the EnrichedDTOs had dependencies to stuff that we wouldn't want to have in core.dto packages; but doing a quick scan, I cannot find any of those and thus it should imho be an option to move them as well. I agree that this would probably the nicest solution to have all "levels" of DTOs in a common package.

Does anyone see a reason that speaks against this? As this is a bit bigger refactoring, we can do this as a separate PR not to block this change here too long (as it addresses a real issue). But if we go this way, I would suggest to change this PR in a way that the StrippedDTOs already reside in the core.dto package and that the regular DTO extends them.

tomhoefer:
Ok. I understand your ideas better now... @kaikreuzer I dont see any reason that speaks against your idea

Mixajlo:
I agree with @kaikreuzer as this PR was created to reduce the time that takes to send thing types when requested via REST.
>Are they really only relevant for REST or not?

Kai, I see you point here. And I agree that the stripped versions can be used to be sent via event bus if its necessary to reduce traffic..

Mixajlo:
Not all DTO's have mappers, maybe we should consider to implement mappers(maybe not for all) and the mappers could be moved to different package (for ex. core.thing.dto.mapper)

kaikreuzer:
Why would you want to put the mappers into a separate package?

Mixajlo:
Too many DTO's and Mappers in one package :)

kaikreuzer:
That's no valid argument 😛 API-wise the mappers belong to the DTO, hence they should be in the same package.

Mixajlo:
ok, you are the expert :)

Issue #2193, opened 2016-09-19
[LIFX] Bulb offline issues


wborn:
The LIFX binding is still pretty unstable for me when using it with my 34 bulbs.

Bulbs go offline a lot. I've been trying to get it stable. Bulbs go offline less when I use `MAXIMUM_POLLING_RETRIES = 4;` in the `LifxLightHandler`. So maybe the value can be increased or made configurable e.g. via Paper UI?

Also when bulbs go offline often the following error is logged:

```
22:09:51.993 [INFO ] [me.event.ThingStatusInfoChangedEvent] - 'lifx:colorlight:XXXXXXXXXXXX' changed from OFFLINE (COMMUNICATION_ERROR) to ONLINE
22:09:39.663 [ERROR] [inding.lifx.handler.LifxLightHandler] - An exception orccurred while communicating with the bulb : 'null'
```

The `orccurred` typo is actually very helpful because it helps to pin point which of the two similar log statement needs to be adjusted to also log a the stack trace.

This is the exception that gets logged with the modified log statement.

```
22:45:46.013 [ERROR] [inding.lifx.handler.LifxLightHandler] - An exception orccurred while communicating with the bulb
java.util.ConcurrentModificationException
	at java.util.HashMap$HashIterator.nextNode(HashMap.java:1437)[:1.8.0_101]
	at java.util.HashMap$KeyIterator.next(HashMap.java:1461)[:1.8.0_101]
	at org.eclipse.smarthome.binding.lifx.handler.LifxLightHandler$1.run(LifxLightHandler.java:430)[214:org.eclipse.smarthome.binding.lifx:0.9.0.201609192042]
	at java.util.concurrent.Executors$RunnableAdapter.call(Executors.java:511)[:1.8.0_101]
	at java.util.concurrent.FutureTask.runAndReset(FutureTask.java:308)[:1.8.0_101]
	at java.util.concurrent.ScheduledThreadPoolExecutor$ScheduledFutureTask.access$301(ScheduledThreadPoolExecutor.java:180)[:1.8.0_101]
	at java.util.concurrent.ScheduledThreadPoolExecutor$ScheduledFutureTask.run(ScheduledThreadPoolExecutor.java:294)[:1.8.0_101]
	at java.util.concurrent.ThreadPoolExecutor.runWorker(ThreadPoolExecutor.java:1142)[:1.8.0_101]
[...]

PR #2201, opened 2016-09-20
LIFX : fix #2191, #2193 and other improvements

### LINKED ISSUE #2191: [LIFX] Deadlock may occur when sending broadcast packet
When I added my LIFX bulbs via `smarthome:inbox approve lifx:colorlight:....` the binding ran into deadlock when it was trying get locks for sending a broadcast packet. 

The two deadlocked threads had the following stacktraces:

```
Name: safeCall-5
State: WAITING on java.util.concurrent.locks.ReentrantLock$NonfairSync@5e28855 owned by: ESH-thingHandler-2
Total blocked: 2  Total waited: 183

Stack trace: 
sun.misc.Unsafe.park(Native Method)
java.util.concurrent.locks.LockSupport.park(LockSupport.java:175)
java.util.concurrent.locks.AbstractQueuedSynchronizer.parkAndCheckInterrupt(AbstractQueuedSynchronizer.java:836)
java.util.concurrent.locks.AbstractQueuedSynchronizer.acquireQueued(AbstractQueuedSynchronizer.java:870)
java.util.concurrent.locks.AbstractQueuedSynchronizer.acquire(AbstractQueuedSynchronizer.java:1199)
java.util.concurrent.locks.ReentrantLock$NonfairSync.lock(ReentrantLock.java:209)
java.util.concurrent.locks.ReentrantLock.lock(ReentrantLock.java:285)
org.eclipse.smarthome.binding.lifx.internal.LifxNetworkThrottler.lock(LifxNetworkThrottler.java:64)
org.eclipse.smarthome.binding.lifx.handler.LifxLightHandler.broadcastPacket(LifxLightHandler.java:597)
org.eclipse.smarthome.binding.lifx.handler.LifxLightHandler.initialize(LifxLightHandler.java:223)
org.eclipse.smarthome.core.thing.internal.ThingManager$7$1.call(ThingManager.java:759)
org.eclipse.smarthome.core.thing.internal.ThingManager$7$1.call(ThingManager.java:1)
org.eclipse.smarthome.core.common.SafeMethodCaller$CallableWrapper.call(SafeMethodCaller.java:177)
java.util.concurrent.FutureTask.run(FutureTask.java:266)
java.util.concurrent.ThreadPoolExecutor.runWorker(ThreadPoolExecutor.java:1142)
java.util.concurrent.ThreadPoolExecutor$Worker.run(ThreadPoolExecutor.java:617)
java.lang.Thread.run(Thread.java:745)



Name: ESH-thingHandler-2
State: WAITING on java.util.concurrent.locks.ReentrantLock$NonfairSync@1a82ed98 owned by: safeCall-5
Total blocked: 0  Total waited: 288

Stack trace: 
sun.misc.Unsafe.park(Native Method)
java.util.concurrent.locks.LockSupport.park(LockSupport.java:175)
java.util.concurrent.locks.AbstractQueuedSynchronizer.parkAndCheckInterrupt(AbstractQueuedSynchronizer.java:836)
java.util.concurrent.locks.AbstractQueuedSynchronizer.acquireQueued(AbstractQueuedSynchronizer.java:870)
java.util.concurrent.locks.AbstractQueuedSynchronizer.acquire(AbstractQueuedSynchronizer.java:1199)
java.util.concurrent.locks.ReentrantLock$NonfairSync.lock(ReentrantLock.java:209)
java.util.concurrent.locks.ReentrantLock.lock(ReentrantLock.java:285)
org.eclipse.smarthome.binding.lifx.internal.LifxNetworkThrottler.lock(LifxNetworkThrottler.java:64)
org.eclipse.smarthome.binding.lifx.handler.LifxLightHandler.broadcastPacket(LifxLightHandler.java:597)
org.eclipse.smarthome.binding.lifx.handler.LifxLightHandler.access$14(LifxLightHandler.java:583)
org.eclipse.smarthome.binding.lifx.handler.LifxLightHandler$1.run(LifxLightHandler.java:537)
java.util.concurrent.Executors$RunnableAdapter.call(Executors.java:511)
java.util.concurrent.FutureTask.runAndReset(FutureTask.java:308)
java.util.concurrent.ScheduledThreadPoolExecutor$ScheduledFutureTask.access$301(ScheduledThreadPoolExecutor.java:180)
java.util.concurrent.ScheduledThreadPoolExecutor$ScheduledFutureTask.run(ScheduledThreadPoolExecutor.java:294)
java.util.concurrent.ThreadPoolExecutor.runWorker(ThreadPoolExecutor.java:1142)
java.util.concurrent.ThreadPoolExecutor$Worker.run(ThreadPoolExecutor.java:617)
java.lang.Thread.run(Thread.java:745)
```

After looking at the code, I think an explanation for this deadlock could be that values were being added to the `locks` ConcurrentHashMap in the `LifxNetworkThrottler`. When the size of a `HashMap` changes `values()` could start returning the locks in a different order. Then the two threads could end up waiting on eachothers locks.

kgoderis:
This should fix some concurrency issues. 

@wborn If would be helpful if you could test this in your LIFX "swarm". I have put the number of retries for polling to 4. I prefer not to put that as a configurable parameter as it would clutter the simple and clean configuration interface of the bulbs, but instead I would propose to put it to the value that works for you, since you have the biggest collection of LIFX bulbs I know of ;-)

### DIFF
--- a/extensions/binding/org.eclipse.smarthome.binding.lifx/src/main/java/org/eclipse/smarthome/binding/lifx/internal/LifxNetworkThrottler.java
+++ b/extensions/binding/org.eclipse.smarthome.binding.lifx/src/main/java/org/eclipse/smarthome/binding/lifx/internal/LifxNetworkThrottler.java
@@ -7,17 +7,21 @@
  */
 package org.eclipse.smarthome.binding.lifx.internal;
 
+import java.util.List;
+import java.util.Map;
 import java.util.concurrent.ConcurrentHashMap;
+import java.util.concurrent.CopyOnWriteArrayList;
 import java.util.concurrent.locks.ReentrantLock;
 
 import org.slf4j.Logger;
 import org.slf4j.LoggerFactory;
 
 /**
- * The {@link LifxNetworkThrottler} is helper class that regulates the frequency at which messages/packets are sent to
- * LIFX bulbs. the LIFX LAN Protocol Specification states that bulbs can process up to 20 messages per second, not more
+ * The {@link LifxNetworkThrottler} is a helper class that regulates the frequency at which messages/packets are sent to
+ * LIFX bulbs. The LIFX LAN Protocol Specification states that bulbs can process up to 20 messages per second, not more.
  *
  * @author Karel Goderis - Initial Contribution
+ * @author Wouter Born - Deadlock fix
  */
 public class LifxNetworkThrottler {
 
@@ -25,54 +29,69 @@ public class LifxNetworkThrottler {
 
     public final static long PACKET_INTERVAL = 50;
 
-    private static ConcurrentHashMap<String, ReentrantLock> locks = new ConcurrentHashMap<String, ReentrantLock>();
-    private static ConcurrentHashMap<String, Long> timestamps = new ConcurrentHashMap<String, Long>();
+    /**
+     * Tracks when the last packet was sent to a LIFX bulb. The packet is sent after obtaining the lock and before
+     * releasing the lock.
+     */
+    private static class LifxLightCommunicationTracker {
 
-    public static void lock(String key) {
-        if (!locks.containsKey(key)) {
-            locks.put(key, new ReentrantLock());
+        private long timestamp;
+
+        private ReentrantLock lock = new ReentrantLock();
+
+        public void lock() {
+            lock.lock();
         }
 
-        locks.get(key).lock();
+        public void unlock() {
+            // When iterating over all trackers another thread may have inserted this object so this thread may not
+            // have a lock on it. When the thread does not have the lock, it also did not send a packet.
+            if (lock.isHeldByCurrentThread()) {
+                timestamp = System.currentTimeMillis();
+                lock.unlock();
+            }
+        }
 
-        if (timestamps.get(key) != null) {
+        public long getTimestamp() {
+            return timestamp;
+        }
+    }
 
-            long lastStamp = timestamps.get(key);
-            long timeToWait = Math.max(PACKET_INTERVAL - (System.currentTimeMillis() - lastStamp), 0);
-            if (timeToWait > 0) {
-                try {
-                    Thread.sleep(timeToWait);
-                } catch (InterruptedException e) {
-                    logger.error("An exception occurred while putting the thread to sleep : '{}'", e.getMessage());
+    /**
+     * A separate list of trackers is maintained when locking all lights in case of a broadcast. Iterators of
+     * {@link ConcurrentHashMap}s may behave non-linear when inserts take place to obtain more concurrency. When the
+     * iterator of {@code values()} of {@link #macTrackerMapping} is used for locking all lights, it could sometimes
+     * cause deadlock.
+     */
+    private static List<LifxLightCommunicationTracker> trackers = new CopyOnWriteArrayList<>();
+
+    private static Map<String, LifxLightCommunicationTracker> macTrackerMapping = new ConcurrentHashMap<String, LifxLightCommunicationTracker>();
+
+    public static void lock(String mac) {
+        LifxLightCommunicationTracker tracker = getOrCreateTracker(mac);
+        tracker.lock();
+        waitForNextPacketInterval(tracker.getTimestamp());
+    }
+
+    private static LifxLightCommunicationTracker getOrCreateTracker(String mac) {
+        LifxLightCommunicationTracker tracker = macTrackerMapping.get(mac);
+        if (tracker == null) {
+            // for better performance only synchronize when necessary
+            synchronized (trackers) {
+                // another thread may just have added a tracker in this synchronized block, so reevaluate
+                tracker = macTrackerMapping.get(mac);
+                if (tracker == null) {
+                    tracker = new LifxLightCommunicationTracker();
+                    trackers.add(tracker);
+                    macTrackerMapping.put(mac, tracker);
                 }
             }
         }
-
+        return tracker;
     }
 
-    public static void unlock(String key) {
-
-        if (locks.containsKey(key)) {
-            timestamps.put(key, System.currentTimeMillis());
-            locks.get(key).unlock();
-        }
-    }
-
-    public static void lock() {
-
-        for (ReentrantLock aLock : locks.values()) {
-            aLock.lock();
-        }
-
-        long lastStamp = 0;
-
-        for (Long aStamp : timestamps.values()) {
-            if (aStamp > lastStamp) {
-                lastStamp = aStamp;
-            }
-        }
-
-        long timeToWait = Math.max(PACKET_INTERVAL - (System.currentTimeMillis() - lastStamp), 0);
+    private static void waitForNextPacketInterval(long timestamp) {
+        long timeToWait = Math.max(PACKET_INTERVAL - (System.currentTimeMillis() - timestamp), 0);
         if (timeToWait > 0) {
             try {
                 Thread.sleep(timeToWait);
@@ -82,15 +101,25 @@ public class LifxNetworkThrottler {
         }
     }
 
-    public static void unlock() {
-
-        for (String key : locks.keySet()) {
-            if (locks.get(key).isHeldByCurrentThread()) {
-                timestamps.put(key, System.currentTimeMillis());
-                locks.get(key).unlock();
-            }
+    public static void unlock(String mac) {
+        if (macTrackerMapping.containsKey(mac)) {
+            macTrackerMapping.get(mac).unlock();
         }
+    }
 
+    public static void lock() {
+        long lastStamp = 0;
+        for (LifxLightCommunicationTracker tracker : trackers) {
+            tracker.lock();
+            lastStamp = Math.max(lastStamp, tracker.getTimestamp());
+        }
+        waitForNextPacketInterval(lastStamp);
+    }
+
+    public static void unlock() {
+        for (LifxLightCommunicationTracker tracker : trackers) {
+            tracker.unlock();
+        }
     }
 
 }
--- a/extensions/binding/org.eclipse.smarthome.binding.lifx/src/main/java/org/eclipse/smarthome/binding/lifx/handler/LifxLightHandler.java
+++ b/extensions/binding/org.eclipse.smarthome.binding.lifx/src/main/java/org/eclipse/smarthome/binding/lifx/handler/LifxLightHandler.java
@@ -24,6 +24,7 @@ import java.nio.channels.SelectionKey;
 import java.nio.channels.Selector;
 import java.nio.channels.SocketChannel;
 import java.util.ArrayList;
+import java.util.ConcurrentModificationException;
 import java.util.Enumeration;
 import java.util.Iterator;
 import java.util.List;
@@ -85,9 +86,9 @@ public class LifxLightHandler extends BaseThingHandler {
     private static long NETWORK_INTERVAL = 50;
     private static int ECHO_POLLING_INTERVAL = 15;
     private static int STATE_POLLING_INTERVAL = 3;
-    private static int MAXIMUM_POLLING_RETRIES = 2;
+    private static int MAXIMUM_POLLING_RETRIES = 4;
     private static int lightCounter = 1;
-    private static ReentrantLock lighCounterLock = new ReentrantLock();
+    private static ReentrantLock lightCounterLock = new ReentrantLock();
 
     private long source;
     private int service;
@@ -121,38 +122,58 @@ public class LifxLightHandler extends BaseThingHandler {
 
     @Override
     public void dispose() {
-        if (networkJob != null && !networkJob.isCancelled()) {
-            networkJob.cancel(true);
-            networkJob = null;
-        }
-
-        currentColorState = null;
-        currentPowerState = null;
-
         try {
-            if (selector != null) {
-                selector.close();
-            }
-        } catch (IOException e) {
-            logger.warn("An exception occurred while closing the selector : '{}'", e.getMessage());
-        }
+            lock.lock();
 
-        if (broadcastKey != null) {
-            try {
-                broadcastKey.channel().close();
-            } catch (IOException e) {
-                // TODO Auto-generated catch block
-                e.printStackTrace();
+            if (networkJob != null && !networkJob.isCancelled()) {
+                networkJob.cancel(true);
+                networkJob = null;
             }
-        }
 
-        if (unicastKey != null) {
+            currentColorState = null;
+            currentPowerState = null;
+
             try {
-                unicastKey.channel().close();
+                if (selector != null) {
+
+                    selector.wakeup();
+
+                    boolean isContinue = true;
+                    while (isContinue) {
+                        try {
+                            for (SelectionKey selectionKey : selector.keys()) {
+                                selectionKey.channel().close();
+                                selectionKey.cancel();
+                            }
+                            isContinue = false; // continue till all keys are cancelled
+                        } catch (ConcurrentModificationException e) {
+                            logger.warn("An exception occurred while closing a selector key : '{}'", e.getMessage());
+                        }
+                    }
+
+                    selector.close();
+                }
             } catch (IOException e) {
-                // TODO Auto-generated catch block
-                e.printStackTrace();
+                logger.warn("An exception occurred while closing the selector : '{}'", e.getMessage());
             }
+
+            if (broadcastKey != null) {
+                try {
+                    broadcastKey.channel().close();
+                } catch (IOException e) {
+                    logger.warn("An exception occurred while closing the broadcast channel : '{}'", e.getMessage());
+                }
+            }
+
+            if (unicastKey != null) {
+                try {
+                    unicastKey.channel().close();
+                } catch (IOException e) {
+                    logger.warn("An exception occurred while closing the unicast channel : '{}'", e.getMessage());
+                }
+            }
+        } finally {
+            lock.unlock();
         }
 
     }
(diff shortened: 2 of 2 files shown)


wborn:
@kgoderis thanks for the quick response! I'll put the swarm to work with your updated code and report back on how well it all works with it. With using 4 retries I already know that only very occasionally a bulb goes offline.

#2191 does does not occur frequently. I'll try to reproduce it by adding/removing the bulbs a lot of times. 

I think when your code works with my setup it should work with everyone's. I'm still running everything on my fast development PC instead of a less capable Raspberry Pi or so. So this setup is a concurrency nightmare for your code. ;-)

wborn:
@kgoderis I still ran into deadlock, I'll add some review comments on where I see room for improvement.

This is the thread dump I now got of the deadlock

```
Name: ESH-discovery-1
State: WAITING on java.util.concurrent.locks.ReentrantLock$NonfairSync@68f44d4b owned by: ESH-thingHandler-3
Total blocked: 0  Total waited: 14

Stack trace: 
sun.misc.Unsafe.park(Native Method)
java.util.concurrent.locks.LockSupport.park(LockSupport.java:175)
java.util.concurrent.locks.AbstractQueuedSynchronizer.parkAndCheckInterrupt(AbstractQueuedSynchronizer.java:836)
java.util.concurrent.locks.AbstractQueuedSynchronizer.acquireQueued(AbstractQueuedSynchronizer.java:870)
java.util.concurrent.locks.AbstractQueuedSynchronizer.acquire(AbstractQueuedSynchronizer.java:1199)
java.util.concurrent.locks.ReentrantLock$NonfairSync.lock(ReentrantLock.java:209)
java.util.concurrent.locks.ReentrantLock.lock(ReentrantLock.java:285)
org.eclipse.smarthome.binding.lifx.internal.LifxNetworkThrottler.lock(LifxNetworkThrottler.java:66)
org.eclipse.smarthome.binding.lifx.internal.LifxLightDiscovery.broadcastPacket(LifxLightDiscovery.java:229)
org.eclipse.smarthome.binding.lifx.internal.LifxLightDiscovery.doScan(LifxLightDiscovery.java:210)
org.eclipse.smarthome.binding.lifx.internal.LifxLightDiscovery$2.run(LifxLightDiscovery.java:144)
java.util.concurrent.Executors$RunnableAdapter.call(Executors.java:511)
java.util.concurrent.FutureTask.runAndReset(FutureTask.java:308)
java.util.concurrent.ScheduledThreadPoolExecutor$ScheduledFutureTask.access$301(ScheduledThreadPoolExecutor.java:180)
java.util.concurrent.ScheduledThreadPoolExecutor$ScheduledFutureTask.run(ScheduledThreadPoolExecutor.java:294)
java.util.concurrent.ThreadPoolExecutor.runWorker(ThreadPoolExecutor.java:1142)
java.util.concurrent.ThreadPoolExecutor$Worker.run(ThreadPoolExecutor.java:617)
java.lang.Thread.run(Thread.java:745)


Name: ESH-thingHandler-3
State: WAITING on java.util.concurrent.locks.ReentrantLock$NonfairSync@55ef972f owned by: ESH-discovery-1
Total blocked: 0  Total waited: 9,858

Stack trace: 
sun.misc.Unsafe.park(Native Method)
java.util.concurrent.locks.LockSupport.park(LockSupport.java:175)
java.util.concurrent.locks.AbstractQueuedSynchronizer.parkAndCheckInterrupt(AbstractQueuedSynchronizer.java:836)
java.util.concurrent.locks.AbstractQueuedSynchronizer.acquireQueued(AbstractQueuedSynchronizer.java:870)
java.util.concurrent.locks.AbstractQueuedSynchronizer.acquire(AbstractQueuedSynchronizer.java:1199)
java.util.concurrent.locks.ReentrantLock$NonfairSync.lock(ReentrantLock.java:209)
java.util.concurrent.locks.ReentrantLock.lock(ReentrantLock.java:285)
org.eclipse.smarthome.binding.lifx.internal.LifxNetworkThrottler.lock(LifxNetworkThrottler.java:38)
org.eclipse.smarthome.binding.lifx.handler.LifxLightHandler.sendPacket(LifxLightHandler.java:601)
org.eclipse.smarthome.binding.lifx.handler.LifxLightHandler.access$12(LifxLightHandler.java:589)
org.eclipse.smarthome.binding.lifx.handler.LifxLightHandler$1.run(LifxLightHandler.java:556)
java.util.concurrent.Executors$RunnableAdapter.call(Executors.java:511)
java.util.concurrent.FutureTask.runAndReset(FutureTask.java:308)
java.util.concurrent.ScheduledThreadPoolExecutor$ScheduledFutureTask.access$301(ScheduledThreadPoolExecutor.java:180)
java.util.concurrent.ScheduledThreadPoolExecutor$ScheduledFutureTask.run(ScheduledThreadPoolExecutor.java:294)
java.util.concurrent.ThreadPoolExecutor.runWorker(ThreadPoolExecutor.java:1142)
java.util.concurrent.ThreadPoolExecutor$Worker.run(ThreadPoolExecutor.java:617)
java.lang.Thread.run(Thread.java:745)

```

extensions/binding/org.eclipse.smarthome.binding.lifx/src/main/java/org/eclipse/smarthome/binding/lifx/internal/LifxNetworkThrottler.java
```diff
@@ -25,8 +27,8 @@
 
     public final static long PACKET_INTERVAL = 50;
 
-    private static ConcurrentHashMap<String, ReentrantLock> locks = new ConcurrentHashMap<String, ReentrantLock>();
-    private static ConcurrentHashMap<String, Long> timestamps = new ConcurrentHashMap<String, Long>();
+    private static Map<String, ReentrantLock> locks = Collections.synchronizedMap(new HashMap<String, ReentrantLock>());
```
wborn: This data structure still does not solve the issue that when as the `Map` grows it starts returning the locks in a different order when `values()` is called. All threads should really lock in the same order to get rid of the deadlock.

You can clearly see it happen when you make a simple test printing `String`s as map values. It kicks even in earlier than I thought. Especially when there are a lot of bulbs, the order starts changing a lot each time a new lock is added:

```
    Map<String, String> keyValueMap = Collections.synchronizedMap(new HashMap<String, String>());

    @Test
    public void valuesTest()
    {
        String values = "";

        for (int i = 0; i < 10; i++)
        {
            keyValueMap.put("key" + i, "value" + i);
            values = StringUtils.join(keyValueMap.values(), ",");

            System.out.println(values);

        }
    }


The output of system is then:

value0
value1,value0
value1,value2,value0
value1,value2,value0,value3
value1,value2,value0,value3,value4
value1,value2,value0,value5,value3,value4
value1,value2,value0,value5,value6,value3,value4
value1,value2,value0,value5,value6,value3,value4,value7
value1,value2,value0,value5,value6,value3,value4,value7,value8
value1,value2,value0,value5,value6,value3,value4,value9,value7,value8
```

extensions/binding/org.eclipse.smarthome.binding.lifx/src/main/java/org/eclipse/smarthome/binding/lifx/internal/LifxNetworkThrottler.java
```diff
@@ -25,8 +27,8 @@
...
+    private static Map<String, Long> timestamps = Collections.synchronizedMap(new HashMap<String, Long>());
 
     public static void lock(String key) {
         if (!locks.containsKey(key)) {
```
wborn: I don't think it is thread safe to first inspect the map and then add the key in the next statement. Most of the time it will work though. ;-) But in theory two threads may inspect the map. Conclude that the key is not there, and then one thread would overwrite the others value.

kgoderis:
> The Javadoc of synchronizedMap(..) suggest it might be a good idea to synchronize on the Map itself when iterating over it. I might give that a try tomorrow and see what happens.

mmh... my understanding was that synchronizedMap() does lock on both read and write operations, in contrast to the ConcurrentHashMap. 

I fail to see as to why there is a deadlock in the first place. Does the order of the .values() really matter?  So, the deadlock occurs when a tread calls .lock(somestring), and then another thread does a .lock() on all locks (e.g. broadcast). 

Imagine that order:
1. lock(string) locks a single lock
2. lock() iterates over all locks, and is stopped when getting to the  individual lock that is already locked in 1. 
3. unlock(string) gets called by the first thread
4. the lock() continues to lock the remaining locks

what are we missing here?

putting a synchronised() the map itself is a solution, but I am wondering what performance impact that could have on a large number of bulbs

wborn:
@kgoderis I refactored the `LifxNetworkThrottler` a bit myself now to see if I can make the deadlock go away. I've reinstated the `ConcurrentHashMap` because I really noticed some downgraded performance due to the `synchronizedMap`.

Also I've introduced an additional `CopyOnWriteArrayList` for storing/getting the locks that would normally be returned by `.values()`. That should take care of any influence `put(..)` operations have on `.values()` might have. So far no deadlocks... I'll keep testing this for a couple of days.

The bulbs do stay online much better again with the 4 retries and there are only occasionally messages about bulbs that are missing in action. :+1:

kgoderis:
> Also I've introduced an additional CopyOnWriteArrayList for storing/getting the locks that would normally be returned by .values(). That should take care of any influence put(..) operations have on .values() might have. So far no deadlocks... I'll keep testing this for a couple of days.

I too found some documentation on ConcurrentHashMap vs syncrhonizedMap performance, it should a factor 4 difference. 

Ok - just do a PR on my repo so that I can integrate it all in a clean PR on the main repo

wborn:
I think the root cause of the deadlock is that the `ConcurrentHashMap` resorts to tricks to keep things highly concurrent: http://www.ibm.com/developerworks/java/library/j-jtp07233/index.html

> ConcurrentHashMap achieves higher concurrency by slightly relaxing the promises it makes to callers. A retrieval operation will return the value inserted by the most recent completed insert operation, and may also return a value added by an insertion operation that is concurrently in progress (but in no case will it return a nonsense result). Iterators returned by ConcurrentHashMap.iterator() will return each element once at most and will not ever throw ConcurrentModificationException, but may or may not reflect insertions or removals that occurred since the iterator was constructed. No table-wide locking is needed (or even possible) to provide thread-safety when iterating the collection. ConcurrentHashMap may be used as a replacement for synchronizedMap or Hashtable in any application that does not rely on the ability to lock the entire table to prevent updates.

wborn:
I also think your `org.openhab.binding.oceanic.internal.SerialPortThrottler` has the same issues ;-)

kgoderis:
> I also think your org.openhab.binding.oceanic.internal.SerialPortThrottler has the same issues ;-)

Yeah, I know, but that is on a single serial port, and normally there is no more than 1 thread accessing the throttler.

kgoderis:
@wborn Any further observations ?

wborn:
The deadlock no longer occurred after restarting OH dozens of times with with my refactored version of the `LifxNetworkThrottler` for the last few days. So I'll put in a PR with your repo.


There is still a `ConcurrentModificationException` when bulbs come back `ONLINE` when they have been `OFFLINE`.

```
21:13:31.788 [ERROR] [inding.lifx.handler.LifxLightHandler] - An exception orccurred while communicating with the bulb
java.util.ConcurrentModificationException
	at java.util.HashMap$HashIterator.nextNode(HashMap.java:1437)[:1.8.0_101]
	at java.util.HashMap$KeyIterator.next(HashMap.java:1461)[:1.8.0_101]
	at org.eclipse.smarthome.binding.lifx.handler.LifxLightHandler$1.run(LifxLightHandler.java:459)[214:org.eclipse.smarthome.binding.lifx:0.9.0.201609221856]
	at java.util.concurrent.Executors$RunnableAdapter.call(Executors.java:511)[:1.8.0_101]
	at java.util.concurrent.FutureTask.runAndReset(FutureTask.java:308)[:1.8.0_101]
	at java.util.concurrent.ScheduledThreadPoolExecutor$ScheduledFutureTask.access$301(ScheduledThreadPoolExecutor.java:180)[:1.8.0_101]
	at java.util.concurrent.ScheduledThreadPoolExecutor$ScheduledFutureTask.run(ScheduledThreadPoolExecutor.java:294)[:1.8.0_101]
	at java.util.concurrent.ThreadPoolExecutor.runWorker(ThreadPoolExecutor.java:1142)[:1.8.0_101]
	at java.util.concurrent.ThreadPoolExecutor$Worker.run(ThreadPoolExecutor.java:617)[:1.8.0_101]
	at java.lang.Thread.run(Thread.java:745)[:1.8.0_101]
```

It is easy to reproduce. Just disconnect power from a bulb. Wait for it to go offline. Then restore power to the bulb. When it goes online the exception occurs.


Also I wonder why in `LifxLightHandler.sendPacket` `macAddress.getAsLabel()` is used and not `macAddress.getHex()`? The last one would be a less expensive operation.


I still see some `A messages with sequence number '54' has already been sent to the bulb. Is it missing in action?` messages, even though my bulbs remain online. Do you also have those? They could ofcourse be valid when there is no response. But I'd rather see them gone when they are false positives due to some software bug. :-)

kgoderis:
> Also I wonder why in LifxLightHandler.sendPacket macAddress.getAsLabel() is used and not macAddress.getHex()? The last one would be a less expensive operation.

No particular reason but you are right ;-)

kgoderis:
> There is still a ConcurrentModificationException when bulbs come back ONLINE when they have been OFFLINE.

Wouter, I was unable to reproduce this. Apart from that, given that a lock is acquired in the run(), I wondering how that CME is possible in the first place. Did you investigate it further?

kgoderis:
> I still see some A messages with sequence number '54' has already been sent to the bulb. Is it missing in action? messages, even though my bulbs remain online. Do you also have those?

I have less of these since we fixed the error you found. Now, the sequence number is between 1 and 255, and when a lot of messages are sent to a bulb, the sequence number runs over 255 and starts again at 1. If there are previously sent messages still stuck in the sent queue (because there was no answer to it due to whatever problem), then the warning is produced. To overcome this annoying message/situation, we could opt for a mechanism whereby we use the timestamp (a stamp is created when the class instance is instantiated) of the sent packets in the sent queue, and retire these packets after a given amount of time so that the queue is cleared over time. For example, do not keep sent packets older than 60 seconds?

wborn:
> Wouter, I was unable to reproduce this. Apart from that, given that a lock is acquired in the run(), I wondering how that CME is possible in the first place. Did you investigate it further?

I have not yet investigated this issue but I can give it a try.

I did  have a look at:
> I still see some A messages with sequence number '54' has already been sent to the bulb. Is it missing in action? messages, even though my bulbs remain online. Do you also have those? They could ofcourse be valid when there is no response. But I'd rather see them gone when they are false positives due to some software bug. :-)

After tweaking the logging and filtering requests/responses per bulb, my conclusion is that these are not false positives but indeed packages that never got a response.

> To overcome this annoying message/situation, we could opt for a mechanism whereby we use the timestamp (a stamp is created when the class instance is instantiated) of the sent packets in the sent queue, and retire these packets after a given amount of time so that the queue is cleared over time. For example, do not keep sent packets older than 60 seconds?

I think we should address the issue of packets that get lost in another (new) issue. For developers working on the binding it is useful information to know that there were indeed packets that got lost. If we were to do something about it in this issue, I think changing the log level to debug would be the most simple short-term solution for now. End users will most of the time just be nagged with the warning.

Also it really depends how bad it really is that a packet got lost. Most of the packets that are sent to bulbs are just continuously querying its current state (GetRequest/GetServiceRequest/GetLightPowerRequest). When there is no response there will be another request, several seconds later, that will most likely get a response. If not, then the bulb will most likely be offline and the end user will see the bulb go from `ONLINE` to `OFFLINE` anyways some time later.

When a packet gets lost that is send via `handleCommand(..)` it is an issue. I sometimes see this happen in my own network when a bulb does not go on/off. But then I would rather have that the binding recovers from it by resending the command instead of logging warnings. ;-) 

Even with `handleCommand(..)` packets it is sometimes undesirable to recover from lost packets. When a creative end user uses the binding for executing cool lighting effects, I think the binding should only make sure that the last command is successfully executed.

So I think we are done with issues #2191 and #2193 when the `ConcurrentModificationException` is resolved.

wborn:
>> Wouter, I was unable to reproduce this. Apart from that, given that a lock is acquired in the run(), I wondering how that CME is possible in the first place. Did you investigate it further?

> I have not yet investigated this issue but I can give it a try.

OK, I also had again some troubles to always reproduce it. So I reprogrammed the binding to automatically throw a bulb each time offline. This increased the chance the `ConcurrentModificationException` occurs. Apparently it occurs due to registration of the `unicastChannel` in the `selector` in `handlePacket(..)`.

```
    private void handlePacket(Packet packet, InetSocketAddress address) {
...
            if (packet instanceof StateServiceResponse) {
...
                                unicastKey = unicastChannel.register(selector,
                                        SelectionKey.OP_READ | SelectionKey.OP_WRITE);
...
```

This is actually a method called from the iteration over the `selector.selectedKeys()` in `run()` according to the call hierarchy. So it makes sense for a `ConcurrentModificationException` to occur. 

Now I just need to come up with a solution.

wborn:
Karel I put in a new PR with your that should fix the CME. Also I made some minor logging improvement and used the `getHex` when sending packets. That should resolve all remaining issues of #2191 and #2193.

If you want, you can do something about the nagging warnings in case of lost packets. I'm OK with putting it on debug or not doing anything about it now. I would rather see effort go into making a structural solution. Something that could look like what I outlined in one of my comments above.

kgoderis:
@kaikreuzer This is LGTM and contains necessary improvements for the LIFX binding. Thanks to @wborn for providing insights and fixes.

kaikreuzer:
Thank you both!

Issue #2182, opened 2016-09-16
Letter inside of Circle on Paper UI Thing Screen Still Relies on ThingType being part of the ThingUID


arstorey:
In the file:  smarthome/extensions/ui/org.eclipse.smarthome.ui.paper/web-src/partials/configuration.html

The code that renders a circle with a letter beside it in the Thing view of the paper UI is getting its text from the 2nd field of the ThingUID string at index 1.  Since we now have the ability to change the TypeType of a Thing after its creation, this field is no longer relevant and changes have been made to ThingUID.java to indicate that the field is deprecated.  I changed my binding to stop including the ThingType as part of my ThingUID so now I just have a blank circle next to my Things in the UI.

Changes will be necessary to configuration.html to account for this and to obtain the first character of the Thing's current ThingType another way.

kaikreuzer:
@aounhaider1 Could you change this?

PR #2203, opened 2016-09-21
Use thingTypeUID field for thing's icon character

aounhaider1:
fixes https://github.com/eclipse/smarthome/issues/2182


aounhaider1:
No. But is it possible for thing to not have a thingtype uid at all? In that case we can show the first letter of thing's label itself.

kaikreuzer:
Yes, this is possible. Also ok to leave it empty in such a case, I just want to make sure that there isn't an error thrown.

aounhaider1:
Updated!

kaikreuzer:
Thanks!

PR #2204, opened 2016-09-21
Bug fix: show error message on failed rest request

aounhaider1:

### DIFF
--- a/extensions/ui/org.eclipse.smarthome.ui.paper/web-src/js/services.js
+++ b/extensions/ui/org.eclipse.smarthome.ui.paper/web-src/js/services.js
@@ -6,8 +6,8 @@ angular.module('PaperUI.services', [ 'PaperUI.constants' ]).config(function($htt
     $httpProvider.interceptors.push(function($q, $injector) {
         return {
             'responseError' : function(rejection) {
-                var showError = !!rejection.showError;
-                if (showError) {
+                var showError = rejection.showError;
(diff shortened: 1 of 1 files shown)


extensions/ui/org.eclipse.smarthome.ui.paper/web-src/js/services.js
```diff
@@ -6,8 +6,8 @@ angular.module('PaperUI.services', [ 'PaperUI.constants' ]).config(function($htt
     $httpProvider.interceptors.push(function($q, $injector) {
         return {
             'responseError' : function(rejection) {
-                var showError = !!rejection.showError;
-                if (showError) {
+                var showError = rejection.showError;
+                if (showError !== false) {
```
maggu2810: As my knowledge about JS is about zero, could you explain me, what is the difference now?
Why has the "not not" been wrong?

aounhaider1: Error should be shown when showError is not set i.e. when it is undefined. When rejection.showError is undef first negation returns true and second one converts it to false and the error is not shown. The error should only be shown when showError is EXACTLY equal to false. i.e. when it is set.

maggu2810: Thanks @aounhaider1, perhaps I will understand that JS some time better :wink:

Issue #2202, opened 2016-09-20
OH2 new-gen rule engine only shows two conditions


Sheriff-Donnerknall:
I am using OH2 beta 4 with the new-gen rules. When I enter more than two conditions for my rule OH2 seems to save the rule fine. But when I open the rule again it only displays two conditions. I am using a temperature and solar azimuth and solar elevation. But the last one of the solar conditions I enter survives and the first that I have entered is no longer shown. No matter which one I enter first or second. The temperature condition always stays in the rule.
When I click on the rule name to see the JSON representation of the rule it actually includes both solar conditions as shown below:

```
{
  "enabled": false,
  "status": {
    "status": "DISABLED",
    "statusDetail": "NONE"
  },
  "triggers": [
    {
      "id": "trigger_1",
      "label": "Temperaturänderung",
      "description": "Temperatur Studio",
      "configuration": {
        "itemName": "homematic_HG_HM_TC_IT_WM_W_EU_HMCFGUSB2_NEQ0362419_2_ACTUAL_TEMPERATURE"
      },
      "type": "ItemStateUpdateTrigger"
    },
    {
      "id": "trigger_5",
      "label": "Timer Trigger",
      "description": "This triggers a rule based on a cron expression",
      "configuration": {
        "cronExpression": "0 */2 * * * ?"
      },
      "type": "TimerTrigger"
    }
  ],
  "conditions": [
    {
      "inputs": {},
      "id": "condition_6",
      "label": "SonneUeberHorizont",
      "description": "Sonne ueber Horizont",
      "configuration": {
        "itemName": "astro_sun_local_position_elevation",
        "state": "0.0",
        "operator": ">"
      },
      "type": "ItemStateCondition"
    },
    {
      "inputs": {},
      "id": "condition_3",
      "label": "Temperatur über 23.3",
      "description": "Thermostat Studio",
      "configuration": {
        "itemName": "homematic_HG_HM_TC_IT_WM_W_EU_HMCFGUSB2_NEQ0362419_1_TEMPERATURE",
        "state": "23.3",
        "operator": ">"
      },
      "type": "ItemStateCondition"
    },
    {
      "inputs": {},
      "id": "condition_6",
      "label": "Sonnenposition",
      "description": "Sonne weiter als 180 Grad sued",
      "configuration": {
        "itemName": "astro_sun_local_position_azimuth",
        "state": "180",
        "operator": ">"
      },
      "type": "ItemStateCondition"
    }
  ],
[...]

PR #2208, opened 2016-09-22
Bug fix: module ids not unique

aounhaider1:
fixes https://github.com/eclipse/smarthome/issues/2202

### DIFF
--- a/extensions/ui/org.eclipse.smarthome.ui.paper/web-src/js/shared.properties.js
+++ b/extensions/ui/org.eclipse.smarthome.ui.paper/web-src/js/shared.properties.js
@@ -27,9 +27,9 @@ angular.module('PaperUI.controllers').service('sharedProperties', function() {
         },
         updateModule : function(type, value) {
             var modArr = this.getModuleArray(type);
-
+            var maxId = this.getMaxModuleId(modArr);
             if (!value.id) {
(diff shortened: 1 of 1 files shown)


extensions/ui/org.eclipse.smarthome.ui.paper/web-src/js/shared.properties.js
```diff
@@ -31,16 +29,14 @@ angular.module('PaperUI.controllers').service('sharedProperties', function() {
             var modArr = this.getModuleArray(type);
 
             if (!value.id) {
-                value.id = type + "_" + tId;
+                value.id = type.substring(0, 1) + "_" + Date.now();
```
kaikreuzer: Will this now add a timestamp as part of the id? I would still prefer "simple" ids, e.g. using the first positive integer that isn't yet used within the rule. Would something like that be possible?

aounhaider1: Updated accordingly.

aounhaider1: But it uses the max number in the respective module and not in the rule overall. Is that fine?

Issue #2164, opened 2016-09-13
Sonos binding: new channel to know what source is being played


lolodomo:
It can be useful when working with automation rules to have a way to know what a Sonos device is currently playing. For example, I would like to know if a particular TuneIn station is being played.
So we need to add a new channel that will match CurrentAVTransportURI.

I will implement this enhancement.

lolodomo:
@kgoderis & @R-Edelhoff : I have a question for you because I discovered something a little strange I don't really understand. It concerns the call to getMediaInfo() every minute (by default). What is used in return to this call is only CurrentURI and CurrentURIMetaData. But these information are already provided in pseudo real time through the event subscription: AVTransportURI and AVTransportURIMetaData.
Do I miss something ? Is there a reason to not rely on event subscription ?

lolodomo:
In fact, the track data are already updated when receiving CurrentTrackMetaData from the event subscription. Perect. I am just surprised that sometimes in the code "CurrentURIMetaData" is used, something that is set only by a call to updateMediaInfo() every minute.
And my feeling is that we should consider AVTransportURI being notified in real time rather than CurrentURI being set only every minute.

lolodomo:
Independently of the previous remarks, I have now proposed a PR to add 2 new channels (URI).

PR #2213, opened 2016-09-24
Sonos binding: 2 new channels currenttrackuri and currenttransporturi

lolodomo:
This PR implements #2164.

### DIFF
--- a/extensions/binding/org.eclipse.smarthome.binding.sonos/src/main/java/org/eclipse/smarthome/binding/sonos/handler/ZonePlayerHandler.java
+++ b/extensions/binding/org.eclipse.smarthome.binding.sonos/src/main/java/org/eclipse/smarthome/binding/sonos/handler/ZonePlayerHandler.java
@@ -560,6 +560,16 @@ public class ZonePlayerHandler extends BaseThingHandler implements UpnpIOPartici
                     updateCurrentURIFormatted(value);
                     break;
                 }
+                case "AVTransportURI": {
+                    updateState(CURRENTTRANSPORTURI, (stateMap.get("AVTransportURI") != null)
+                            ? new StringType(stateMap.get("AVTransportURI")) : UnDefType.UNDEF);
+                    break;
+                }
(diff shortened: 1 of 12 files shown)


extensions/binding/org.eclipse.smarthome.binding.sonos/src/main/java/org/eclipse/smarthome/binding/sonos/handler/ZonePlayerHandler.java
```diff
@@ -560,6 +560,16 @@ public void onValueReceived(String variable, String value, String service) {
                     updateCurrentURIFormatted(value);
                     break;
                 }
+                case "AVTransportURI": {
+                    updateState(CURRENTTRANSPORTURI, (stateMap.get("AVTransportURI") != null)
```
maggu2810: There is no need to use braces around the stateMap.get call.

extensions/binding/org.eclipse.smarthome.binding.sonos/src/main/java/org/eclipse/smarthome/binding/sonos/handler/ZonePlayerHandler.java
```diff
@@ -560,6 +560,16 @@ public void onValueReceived(String variable, String value, String service) {
...
+                            ? new StringType(stateMap.get("AVTransportURI")) : UnDefType.UNDEF);
+                    break;
+                }
+                case "CurrentTrackURI": {
+                    updateState(CURRENTTRACKURI, (stateMap.get("CurrentTrackURI") != null)
```
maggu2810: There is no need to use braces around the stateMap.get call.

lolodomo:
@maggu2810 :  I put braces because all other cases in the switch have braces. But I agree they are not necessary.

maggu2810:
You are right. I just have had a look at the changes not on the existing code. So it is okay but perhaps we could refactor the code some time.

PR #2225, opened 2016-09-28
Split general audio from voice support

kaikreuzer:
- moved core.audio into a separate bundle
- introduced AudioManager and made VoiceManager depend on it
- added audio console commands
- introduced URLAudioStream
- introduced FileAudioStream
- added volume control to sinks

Also-by: Karel Goderis <karel.goderis@me.com>


bundles/core/org.eclipse.smarthome.core.audio/src/main/java/org/eclipse/smarthome/core/audio/AudioManager.java
```diff
@@ -0,0 +1,460 @@
...
+
+    // service maps
+    private Map<String, AudioSource> audioSources = new HashMap<>();
+    private Map<String, AudioSink> audioSinks = new HashMap<>();
+
+    /**
+     * default settings filled through the service configuration
+     */
+    private String defaultSource = null;
+    private String defaultSink = null;
```
maggu2810: It is okay to do this explicitly, but no need, as it is null implicitly.

bundles/core/org.eclipse.smarthome.core.audio/src/main/java/org/eclipse/smarthome/core/audio/AudioManager.java
```diff
@@ -0,0 +1,460 @@
...
+     * @param audioStream The audio stream to play
+     */
+    public void play(AudioStream audioStream) {
+        play(audioStream, null);
+    }
+
+    /**
+     * Plays the passed audio stream on the given sink.
+     *
+     * @param audioStream The filename to play
```
maggu2810: filename => audio stream

bundles/core/org.eclipse.smarthome.core.audio/src/main/java/org/eclipse/smarthome/core/audio/AudioManager.java
```diff
@@ -0,0 +1,460 @@
...
+        playFile(fileName, null);
+    }
+
+    /**
+     * Plays an audio file from the "sounds" folder using the given audio sink.
+     *
+     * @throws AudioException in case the file does not exist or cannot be opened
+     */
+    public void playFile(String fileName, String sink) throws AudioException {
+        File file = new File(System.getProperty(ConfigConstants.CONFIG_DIR_PROG_ARGUMENT) + File.separator + SOUND_DIR
```
maggu2810: I would prefer to use `ConfigConstants.getConfigFolder()`.

bundles/core/org.eclipse.smarthome.core.audio/src/main/java/org/eclipse/smarthome/core/audio/AudioManager.java
```diff
@@ -0,0 +1,460 @@
...
+
+    /**
+     * Stream audio from the passed url to the given sink
+     *
+     * @param url The url to stream from
+     * @param sinkId The id of the audio sink to use or null
+     * @throws AudioException in case the url stream cannot be opened
+     */
+    public void stream(String url, String sinkId) throws AudioException {
+
```
maggu2810: As you already need to touch this file, I think this empty line could be removed.

bundles/core/org.eclipse.smarthome.core.audio/src/main/java/org/eclipse/smarthome/core/audio/AudioManager.java
```diff
@@ -0,0 +1,460 @@
...
+            try {
+                return sink.getVolume();
+            } catch (IOException e) {
+                logger.error("An exception occured while getting the volume of sink {} : '{}'", sink.getId(),
+                        e.getMessage());
+            }
+        }
+
+        return 0;
+
```
maggu2810: As you already need to touch this file, I think this empty line could be removed.

bundles/core/org.eclipse.smarthome.core.audio/src/main/java/org/eclipse/smarthome/core/audio/AudioManager.java
```diff
@@ -0,0 +1,460 @@
...
+     * @param sinkId the sink to set the volume
+     * @param volume the volume to set as a value between 0 and 1
+     */
+    public void setVolume(String sinkId, float volume) {
+        AudioSink sink = null;
+        if (sinkId == null) {
+            sink = getSink();
+        } else {
+            sink = audioSinks.get(sinkId);
+        }
```
maggu2810: This logic has been used in different methods of this file.
Perhaps we could create a private method, that handles that logic.

bundles/core/org.eclipse.smarthome.core.audio/src/main/java/org/eclipse/smarthome/core/audio/AudioManager.java
```diff
@@ -0,0 +1,460 @@
...
+                logger.error("An exception occured while setting the volume of sink {} : '{}'", sink.getId(),
+                        e.getMessage());
+            }
+        }
+    }
+
+    /**
+     * Retrieves an AudioSource.
+     * If a default name is configured and the service available, this is returned. Otherwise, the first available
+     * service is returned.
```
maggu2810: This does not fit to the description of the return value and the implementation.
If a default name is configured, but the service is not available, null is returned and not the first available service.

bundles/core/org.eclipse.smarthome.core.audio/src/main/java/org/eclipse/smarthome/core/audio/AudioManager.java
```diff
@@ -0,0 +1,460 @@
...
+        } else {
+            logger.debug("No AudioSource service available!");
+        }
+        return source;
+    }
+
+    /**
+     * Retrieves an AudioSink.
+     * If a default name is configured and the service available, this is returned. Otherwise, the first available
+     * service is returned.
```
maggu2810: See above

bundles/core/org.eclipse.smarthome.core.audio/src/main/java/org/eclipse/smarthome/core/audio/AudioManager.java
```diff
@@ -0,0 +1,460 @@
...
+        return sink;
+    }
+
+    /**
+     * Retrieves the ids of all sources
+     *
+     * @return ids of all sources
+     */
+    public Set<String> getSourceIds() {
+        return audioSources.keySet();
```
maggu2810: The keySet is a view... So, the caller could run into a CME using the returned Set it the Map changes.
Wouldn't it be better to return a new set created by the content of the view set?

bundles/core/org.eclipse.smarthome.core.audio/src/main/java/org/eclipse/smarthome/core/audio/AudioManager.java
```diff
@@ -0,0 +1,460 @@
...
+    }
+
+    /**
+     * Retrieves the ids of all sinks
+     *
+     * @return ids of all sources
+     */
+    public Set<String> getSinkIds() {
+        return audioSinks.keySet();
```
maggu2810: See above

bundles/core/org.eclipse.smarthome.core.audio/src/main/java/org/eclipse/smarthome/core/audio/AudioManager.java
```diff
@@ -0,0 +1,460 @@
...
+            // Return preferred AudioFormat
+            return format;
+        }
+
+        // Return null indicating failure
+        return null;
+    }
+
+    protected void addAudioSource(AudioSource audioSource) {
+        this.audioSources.put(audioSource.getId(), audioSource);
```
maggu2810: This method is used by DS -- so it is called async.
The Map is used by other functions in this class, too.
There is no sync. used.

You can use a ConcurrentHashMap or add sync statements to prevent errors.

bundles/core/org.eclipse.smarthome.core.audio/src/main/java/org/eclipse/smarthome/core/audio/AudioManager.java
```diff
@@ -0,0 +1,460 @@
...
+    }
+
+    protected void removeAudioSource(AudioSource audioSource) {
+        this.audioSources.remove(audioSource.getId());
```
maggu2810: See above

bundles/core/org.eclipse.smarthome.core.audio/src/main/java/org/eclipse/smarthome/core/audio/AudioManager.java
```diff
@@ -0,0 +1,460 @@
...
+    }
+
+    protected void addAudioSink(AudioSink audioSink) {
+        this.audioSinks.put(audioSink.getId(), audioSink);
```
maggu2810: See above

bundles/core/org.eclipse.smarthome.core.audio/src/main/java/org/eclipse/smarthome/core/audio/AudioManager.java
```diff
@@ -0,0 +1,460 @@
...
+    }
+
+    protected void removeAudioSink(AudioSink audioSink) {
+        this.audioSinks.remove(audioSink.getId());
```
maggu2810: See above

bundles/core/org.eclipse.smarthome.core.audio/src/main/java/org/eclipse/smarthome/core/audio/FileAudioStream.java
```diff
@@ -0,0 +1,65 @@
...
+
+    private AudioFormat audioFormat;
+    private InputStream inputStream;
+
+    public FileAudioStream(File file) throws AudioException {
+        this.inputStream = getInputStream(file);
+        this.audioFormat = getAudioFormat(file);
+    }
+
+    private AudioFormat getAudioFormat(File file) throws AudioException {
```
maggu2810: Minor comment: could be static to signal, that we don't access object members (so really save to use in constructor).

bundles/core/org.eclipse.smarthome.core.audio/src/main/java/org/eclipse/smarthome/core/audio/FileAudioStream.java
```diff
@@ -0,0 +1,65 @@
...
+        if (file.getName().toLowerCase().endsWith(".wav")) {
+            return new AudioFormat(AudioFormat.CONTAINER_WAVE, AudioFormat.CODEC_PCM_SIGNED, false, 16, 705600, 44100L);
+        } else if (file.getName().toLowerCase().endsWith(".mp3")) {
+            return new AudioFormat(AudioFormat.CODEC_MP3, AudioFormat.CODEC_MP3, null, null, null, null);
+        } else {
+            throw new AudioException("Unsupported file extension!");
+        }
+    }
+
+    private InputStream getInputStream(File file) throws AudioException {
```
maggu2810: Minor comment: could be static to signal, that we don't access object members (so really save to use in constructor).

features/karaf/src/main/feature/feature.xml
```diff
@@ -41,6 +41,7 @@
 
     <bundle start-level="75">mvn:org.eclipse.smarthome.core/org.eclipse.smarthome.core.thing.xml/${project.version}</bundle>
     <bundle>mvn:org.eclipse.smarthome.core/org.eclipse.smarthome.core.transform/${project.version}</bundle>
+    <bundle>mvn:org.eclipse.smarthome.core/org.eclipse.smarthome.core.audio/${project.version}</bundle>
```
maggu2810: It does need to be target in this PR, but is voice and audio really a base part that needs to be available?
Isn't that some optional feature that could be added...

maggu2810:
I added some comments. Please have a look :wink:

features/karaf/src/main/feature/feature.xml
kaikreuzer: well, the audio package was already part of smarthome.core bundle, so this is actually not a change, just a refactoring while keeping the status quo. If you have a requirement to split this off the core, feel free to suggest a separate PR.

kaikreuzer:
Thanks for the review, I have updated the PR accordingly.

PR #2227, opened 2016-09-29
there is no resources directory in core.audio

maggu2810:
Without this changes the IDE complains about missing resource directory.
As an empty directory gets lost using Git, we should remove it from the build path


kaikreuzer:
Thanks, I just came across the same :-)

Issue #1798, opened 2016-07-03
RawType base64-encoded value is written to log file every time state changes


dominicdesu:
Hi,

whenever an item changes, the ItemStateChangedEvent is logged (on info level), containing the old and the new state of an item:
`itemName + " changed from " + oldItemState.toString() + " to " + itemState.toString();`

If a RawType item has changed, the toString method of the RawType outputs its Base64-encoded value:
`PortableBase64.getEncoder().encode(bytes);`

This is not really nice as I don't need the binary file in my log file and this can increase the size of the log file quite fast.

kaikreuzer:
I agree, this should be avoided.
It would be nicest to implement RawType.toString() so that it comes up with something short like "binary content", but as the toString method is also used to have a full string representation from which types can be reconstructed again, this won't work.
So the better fix is probably to change the toString method of the item events in a way that they behave differently for RawTypes.

thhart:
> but as the toString method is also used to have a full string representation from which types can be reconstructed again, this won't work

Can you explain this further please. Sorry for being this direct but IMHO it is not very good design to have a full Base64 encoding nor a full data return in a toString() method.

maggu2810:
IIRC:
Currenty every type should have a static method valueOf(String) that returns an instance of that type.
The instance should be equal to the one the toString method is called to receive the string.
So `Type.valueOf(type.toString()).equals(type)` should be true all the time.

maggu2810:
I think there are two options:
We could create a static function that returns a simplified string of the state for debugging purpose etc.
It can then skip the value encoding for RawType
```java
    protected String getSimpleStateString(final State state) {
        if (state instanceof RawType) {
            return "<skip-raw-type-encoding>";
        } else {
            return state.toString();
        }
    }
```
This method could be used in [Group]ItemStateChangedEvent or whenever a message is logged.

Okay, but that does not look good as a static method does need to decide how a state should be logged and not the state class itself.

> Currenty every type should have a static method valueOf(String) that returns an instance of that type.
The instance should be equal to the one the toString method is called to receive the string.
So Type.valueOf(type.toString()).equals(type) should be true all the time.

For the first step we could (for the transition phase)
* change the Type interface to define two methods `toStringFull()` and `toStringSimple()`
* `toStringSimple()` should return a simplified string that does not need to contain the whole state (e.g. for RawType the first bytes + "...")
* `toStringFull()` behaves similar as the current `toString()` method (full string representation of the internal type)
[...]

PR #2230, opened 2016-09-29
Type: differ between a full and simple string representation

maggu2810:
* The full string representation must be compatible to the static `valueOf(String)` method of the respective Type implementation.
* The simple string representation does not need to be compatible to the `valueOf(String)` method of the respective Type implementation.

So, whenever you need a representation that can be consumed by `valueOf(String)` later, you should consider to use the full string representation.

Until now `toString()` returns the full string representation.
To keep backward compatibility this mechanism is unchanged.
The toString() of a type should be implemented this way:
`return TypeToStringTransitionUtils.handleStringTransition(this);`

If you don't set trace log level for `org.eclipse.smarthome.core.internal.types.TypeToStringTransitionUtils` there should not be any performance impact.

If you set trace log level for that class, every usage (except ignored ones) of toString() is logged.
This has been done, so you are able to easily change your code to explicit use the 'full' method.

Related to: https://github.com/eclipse/smarthome/issues/1798

### DIFF
--- a/bundles/core/org.eclipse.smarthome.core/src/main/java/org/eclipse/smarthome/core/items/events/GroupItemStateChangedEvent.java
+++ b/bundles/core/org.eclipse.smarthome.core/src/main/java/org/eclipse/smarthome/core/items/events/GroupItemStateChangedEvent.java
@@ -46,8 +46,6 @@ public class GroupItemStateChangedEvent extends ItemStateChangedEvent {
 
     @Override
     public String toString() {
-        return itemName + " changed from " + oldItemState.toString() + " to " + itemState.toString() + " through "
-                + memberName;
-
+        return String.format("%s through %s", super.toString(), memberName);
     }
 }
--- a/bundles/core/org.eclipse.smarthome.core/src/main/java/org/eclipse/smarthome/core/items/events/ItemEventFactory.java
+++ b/bundles/core/org.eclipse.smarthome.core/src/main/java/org/eclipse/smarthome/core/items/events/ItemEventFactory.java
@@ -219,7 +219,7 @@ public class ItemEventFactory extends AbstractEventFactory {
         assertValidArguments(itemName, state, "state");
         String topic = buildTopic(ITEM_STATE_EVENT_TOPIC, itemName);
         ItemEventPayloadBean bean = new ItemEventPayloadBean(state.getClass().getSimpleName(),
-                state.toFullTypeString());
+                state.toFullString());
         String payload = serializePayload(bean);
         return new ItemStateEvent(topic, payload, itemName, state, source);
     }
@@ -253,8 +253,8 @@ public class ItemEventFactory extends AbstractEventFactory {
         assertValidArguments(itemName, newState, "state");
         String topic = buildTopic(ITEM_STATE_CHANGED_EVENT_TOPIC, itemName);
         ItemStateChangedEventPayloadBean bean = new ItemStateChangedEventPayloadBean(
-                newState.getClass().getSimpleName(), newState.toFullTypeString(), oldState.getClass().getSimpleName(),
-                oldState.toFullTypeString());
+                newState.getClass().getSimpleName(), newState.toFullString(), oldState.getClass().getSimpleName(),
+                oldState.toFullString());
         String payload = serializePayload(bean);
         return new ItemStateChangedEvent(topic, payload, itemName, newState, oldState);
     }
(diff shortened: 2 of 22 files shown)


kaikreuzer:
Hm, this is an awful lot of code change for a very minor logging issue... I am honestly not too happy about that. It even introduces a dependency of core API classes to a new (internal) package (which can be an issue if we would ever want to split API from Impl bundles).

Why not simply keep toString() as is and only additionally introduce a toShortString()/toSimpleString() additionally?

maggu2810:
I don't think it is good practice to keep the requirement that toString() needs to export the full state without loose of any information.
We could keep that approach as long as possible, but I still prefer to have a separate method to grant it forever.

> awful lot of code change for a very minor logging issue

It is not about the logging issue but about the requirement that the ESH framework doesn't differ between toString() and get a string representation of the type without loose any information.

> It even introduces a dependency of core API classes to a new (internal) package (which can be an issue if we would ever want to split API from Impl bundles).

Interesting argument. I don't know that this has been a point on any previous contribution (e.g. PortableBase64 etc.).

kaikreuzer:
> I don't think it is good practice to keep the requirement that toString() needs to export the full state without loose of any information.

Well, toString() is supposed to return a "string representation", so this was perfect fit for the Types. It then was a functional requirement that the string representation is parseable again by the class. I still think, this makes perfect sense and I don't think that it is not allowed to add such an additional requirement for your own classes. Note that the Type interface is not meant to be implemented by anyone, but the core framework itself. So there isn't even the need to make sure that people know about this.

So my point is that it is a lot of changed code down in the core framework (with 350 new lines of code more than before). And I do not think that it fixes something as the matter is imho rather a question of perspective than a matter of right or wrong.

> Interesting argument. I don't know that this has been a point on any previous contribution

Right, I only learnt about this recently from @pkriens when diving into bnd, which correctly gives you a warning if you do such things.

maggu2810:
> Note that the Type interface is not meant to be implemented by anyone, but the core framework itself. So there isn't even the need to make sure that people know about this.

But I assume the Types are intended to be consumed by the ones that are using the framework.
The Java API suggests "The result should be a concise but informative representation that is easy for a person to read." (taken from Object's toString method). I assume this does not fit well in "don't loose any information" at least for the raw type.

> with 350 new lines of code more than before

You are right, there are a lot of new lines. Not really 350 if we drop the optional tracking support (that intention was only to find usages) and don't count the formatter cleanups (so, count only the relevant ones -- 3fe220f3a013e852f8b2903df5c1355c1de34870). Still about 300 lines of changes, but IMHO all straight forward changes, as we only introduce separate functions that could handle a separate functionality. Just to note, that the number of lines could be smaller if we would break the compatibility and remove the old behaviour. But I am fine with it, to keep the compatibility as long as necessary.
I would prefer adding more lines of code instead of reusing a method that exist for every object and add some special meaning to it (in the special case for toString it is a function that is used by other functions implicit).

> It even introduces a dependency of core API classes to a new (internal) package (which can be an issue if we would ever want to split API from Impl bundles).

That is really a valid point and it would be perhaps also the best for the whole change of this PR if it would not contain the optional tracking support (I already split it as a separate commit).
Do we consider to create API packages? IIRC you told me some time (can't find the mail) that you don't consider to create separate API packages for ESH.

> I am honestly not too happy about that.

So, how to proceed? As you already stated you don't want a separate function that signals that its intention is to generate something that keeps the whole information and must be consumable by valueOf, I assume we can stop here and close the PR.

kaikreuzer:
Sorry for coming back so late on this, I had to contemplate a bit over it.
I see all your arguments and (as usual), there is hardly anything to say against them - so I do not want to overrule here without any good reason and close the PR.

I therefore digged through the code (on openHAB side) in order to better assess the risk of breaking compatibility; it actually seems that the risk is low as I didn't find any code that relies on the "full" content apart from the core itself.

As introducing a dependency to the internal package might actually cause a runtime incompatibility (and might require a re-compilation), I now rather tend to say that we should not have any transition phase, but rather have things breaking early, if at all. It is imho better now (before a final release of OH2) than changing it afterwards.

So I would vote for only introducing a single new function "toFullString" and change the semantic of toString (while this will actually only have an impact on RawType, since for the rest, the output will be identical, right?).

> Do we consider to create API packages? IIRC you told me some time (can't find the mail) that you don't consider to create separate API packages for ESH.

This is up to us. I never did it as I wanted to keep the number of bundles low and with Tycho-based builds there was not really any benefit. With bnd and "pure Maven" builds, this could change, i.e. it would be nice to have a simple API artifact that can be used as a dependency to compile against (without the need to have the ESH binaries in place). But this is imho something for later, I would not want to go into deeper discussions at this moment.

maggu2810:
> As introducing a dependency to the internal package might actually cause a runtime incompatibility (and might require a re-compilation), I now rather tend to say that we should not have any transition phase, but rather have things breaking early, if at all.

Tracking code removed.

If there are some persistence service that already persist the RawType, I assume that ones need to be changed.

> So I would vote for only introducing a single new function "toFullString" and change the semantic of toString 

Me, too. :wink: 

Code has been changed.

kaikreuzer:
Thanks, lgtm. Just a last remark: 

> So I would vote for only introducing a single new function "toFullString"

The method in the PR is instead `toFullTypeString` - I think we can omit the "Type".

maggu2810:
Done

Issue #1416, opened 2016-04-26
"osgi.os" cannot be used to identify the OS


kaikreuzer:
The "say" console command currently uses the system property "osgi.os" to determine the OS the stack is running on. This property is usually set on Eclipse-based products, but e.g. not within a Karaf instance.

It therefore cannot be used to reliably determine the OS.

My suggestion is to completely remove the platform specific information from the services and rather make them all optional extensions (i.e. extract the Mac TTS service from the io.voice bundle).

kaikreuzer:
@kdavis-mozilla & @tilmankamp I just looked at extracting the MacTTSService implementation and wondered, if you already have further evolved the concept of how the say command should actually be implemented. At the moment, the TTSService still creates the audio stream and outputs it by itself.
As discussed, the TTSService should usually only create the stream which we would then hand over to an AudioSink. Do you already have an adapted say command in place for which you could create a PR?
How would we integrate "old-school" TTS services, such as MacTTS or also the 1.x services like MaryTTS into such a new setup?

kdavis-mozilla:
@kaikreuzer Yes we have an implementation [TTSServiceMacOS.java](https://github.com/mozilla/smarthome/blob/vaani/bundles/io/org.eclipse.smarthome.io.voice/src/main/java/org/eclipse/smarthome/io/voice/internal/TTSServiceMacOS.java) that creates a stream and hands it over to an AudioSink. We also have an adapted say command [SayConsoleCommandExtension.java](https://github.com/mozilla/smarthome/blob/vaani/bundles/io/org.eclipse.smarthome.io.voice/src/main/java/org/eclipse/smarthome/io/voice/internal/extensions/SayConsoleCommandExtension.java)  We've also integrated MaryTTS too [TTSServiceMaryTTS.java](https://github.com/mozilla/smarthome/blob/vaani/bundles/io/org.eclipse.smarthome.io.voice/src/main/java/org/eclipse/smarthome/io/voice/internal/TTSServiceMaryTTS.java).

I think there are one or two issues that have to be addressed before we can do a pull request. But we should be able to do one next week.

kaikreuzer:
Cool, sounds great - I will wait for your contribution then!

kaikreuzer:
@kdavis-mozilla I think I will do the extraction of the MacTTS service into its own bundle anyhow right now - it seems unrelated to your work (and you would have to tear it out of io.voice as well).

[...]

PR #2232, opened 2016-09-29
audio and voice script/rule actions

kaikreuzer:


kgoderis:
@kaikreuzer Maybe we should - likewise the console extension - introduce the usage of wildcards in the sinkId in order to play files/streams directly to a selected number of sinks? That would be handy for example to play a file directly through a bunch of Sonos Players without having to loop through them in the Rule or Script.

kaikreuzer:
@kgoderis I would not want to do this (at least for a start). It would lead to the impression that we support a kind of multi-room setup. Playing the same on multiple Sonos speakers should e.g. be done by grouping the speakers in a zone and then use this as your sink.

But if you think this would be neat for testing and playing around, feel free to create a follow-up PR.

bundles/model/org.eclipse.smarthome.model.script/src/org/eclipse/smarthome/model/script/actions/Audio.java
```diff
@@ -0,0 +1,128 @@
...
+ * This allows a script to use audio features.
+ *
+ * @author Kai Kreuzer
+ */
+public class Audio {
+
+    private static final Logger logger = LoggerFactory.getLogger(Audio.class);
+
+    @ActionDoc(text = "plays a sound from the sounds folder to the default sink")
+    static public void playSound(@ParamDoc(name = "filename", text = "the filename with extension") String filename) {
```
maggu2810: Didn't we use the JLS suggested modifier order all the time? I would prefer to keep a consistent order in the framework.

bundles/model/org.eclipse.smarthome.model.script/src/org/eclipse/smarthome/model/script/actions/Voice.java
```diff
@@ -0,0 +1,83 @@
...
+            VoiceActionService.voiceManager.say(text, voice, sink);
+        }
+    }
+
+    /**
+     * Interprets the given text.
+     *
+     * <p>
+     * This method uses the default Human Language Interpreter and passes the text to it.
+     * </p>
```
maggu2810: There is no closing paragraph tag in the JavaDoc (at least this is not documented by Oracle - or I never found the correct document)

extensions/io/org.eclipse.smarthome.io.javasound/src/main/java/org/eclipse/smarthome/io/javasound/internal/JavaSoundAudioSink.java
```diff
@@ -118,4 +136,7 @@ private void runVolumeCommand(Closure closure) {
         }
     }
 
+    private boolean isMacOS() {
+        return System.getProperty("osgi.os").equals("macosx");
```
maggu2810: is this the last resort?
You previously posted that osgi.os is not reliable to detect the OS (https://github.com/eclipse/smarthome/issues/1416#issue-151054512).

bundles/model/org.eclipse.smarthome.model.script/src/org/eclipse/smarthome/model/script/actions/Audio.java
kaikreuzer: Note that this was old code that was only moved around.
But I have fixed the order now.

bundles/model/org.eclipse.smarthome.model.script/src/org/eclipse/smarthome/model/script/actions/Voice.java
kaikreuzer: Ok, removed.

extensions/io/org.eclipse.smarthome.io.javasound/src/main/java/org/eclipse/smarthome/io/javasound/internal/JavaSoundAudioSink.java
kaikreuzer: Valid point. Again, this is the old code that was in place for long in openHAB 1/2.
I am not sure how to best identify a macos; in my tests, the non-mac implementation actually also worked on a mac, so I am not completely sure whether we need to do this separation at all. @kgoderis do you have any insights here?

maggu2810: It would be great if it is not necessary at all.
But if we need to use it, we should do it using this one:
* https://wiki.searchtechnologies.com/javadoc/osgi/osgi-javadoc-v41/org/osgi/framework/Constants.html#FRAMEWORK_OS_NAME
* https://www.osgi.org/developer/specifications/reference/#ReferenceOS

kaikreuzer: Thanks for the links, I have adapted the code accordingly and it seems to work well.

bundles/model/org.eclipse.smarthome.model.script/src/org/eclipse/smarthome/model/script/actions/Audio.java
maggu2810: You missed exactly that line that has been comment :wink:

bundles/model/org.eclipse.smarthome.model.script/src/org/eclipse/smarthome/model/script/actions/Voice.java
```diff
@@ -0,0 +1,81 @@
...
+ */
+public class Voice {
+
+    /**
+     * Says the given text.
+     *
+     * @param text the text to speak
+     */
+    @ActionDoc(text = "says a given text with the default voice")
+    static public void say(@ParamDoc(name = "text") Object text) {
```
maggu2810: Can you change the order of that file, too?

kaikreuzer:
@maggu2810 Ok, now I should finally have caught all of them!

Issue #2242, opened 2016-10-04
Incorrect update of dateTime channel of the NTP Binding


petarnv:
In the NTP Binding the state of an item, linked to a NTP thing is not updated with a time from the time zone, that is set as a property in the thing configuration. Even though in the NtpHandler the state of the item is updated with a DateTimeType, containing a calendar with the proper time zone (the one set in the configuration of the thing) - https://github.com/eclipse/smarthome/blob/master/extensions/binding/org.eclipse.smarthome.binding.ntp/src/main/java/org/eclipse/smarthome/binding/ntp/handler/NtpHandler.java#L199, this state is serialized in the ThingManager, when an event is posted (https://github.com/eclipse/smarthome/blob/master/bundles/core/org.eclipse.smarthome.core.thing/src/main/java/org/eclipse/smarthome/core/thing/internal/ThingManager.java#L169) by the method createStateEvent() - https://github.com/eclipse/smarthome/blob/master/bundles/core/org.eclipse.smarthome.core/src/main/java/org/eclipse/smarthome/core/items/events/ItemEventFactory.java#L218, where the toString() method of the DateTimeType is used (https://github.com/eclipse/smarthome/blob/master/bundles/core/org.eclipse.smarthome.core/src/main/java/org/eclipse/smarthome/core/library/types/DateTimeType.java#L86), which will return the time from the current time zone(http://stackoverflow.com/questions/11337557/what-is-the-default-timezone-in-java-util-date). Because of this the state of the items is always updated with the current time zone, even if a property for the time zone is set in the thing configuration.

I have made a PR with tests for this: https://github.com/eclipse/smarthome/pull/2243

This behavior can be seen in the following ignored tests:
1) the dateTime channel is updated with the right time zone
2) the calendar of the dateTime channel is updated with the right time zone

marcelrv:
This issue is indeed known. Unfortunately not directly fixable by the binding see

https://github.com/eclipse/smarthome/issues/992

PR #2243, opened 2016-10-04
Implemented tests for NTP Binding

petarnv:
I have implemented 16 tests for the NTP Binding. 2 of the tests are ignored, because I think that the behavior of the NTP Binding in them is unexpected. Here I have explained what is the problem: https://github.com/eclipse/smarthome/issues/2242

Also, I have made small changes in the code of the NtpHandler and NtpBindingConstants:

1) Changed the constant CHANNEL_STING in the NtpBindingConstants class to CHANNEL_STRING (https://github.com/eclipse/smarthome/blob/master/extensions/binding/org.eclipse.smarthome.binding.ntp/src/main/java/org/eclipse/smarthome/binding/ntp/NtpBindingConstants.java#L31)

2) According to the logic in NtpHandler, if the property for formatting in the configuration of the string channel is an empty string, a new SimpleDateFormat object is created with that empty string as an argument (https://github.com/eclipse/smarthome/blob/master/extensions/binding/org.eclipse.smarthome.binding.ntp/src/main/java/org/eclipse/smarthome/binding/ntp/handler/NtpHandler.java#L137). As a result the dates are formatted with a format empty string, which leads to item registry being updated with an empty string as a date. But judging by the comments in the logger the date should be updated with the default format. I think that if we just add a logical NOT operator in the condition of the if : `if (!(dateTimeFormatString == null || dateTimeFormatString.isEmpty()))`, the logic will be correct.

These proposed tests are behaving exactly the same way on Windows and MAC.


extensions/binding/org.eclipse.smarthome.binding.ntp.test/pom.xml
```diff
@@ -0,0 +1,104 @@
+<project xmlns="http://maven.apache.org/POM/4.0.0" xmlns:xsi="http://www.w3.org/2001/XMLSchema-instance"
+  xsi:schemaLocation="http://maven.apache.org/POM/4.0.0 http://maven.apache.org/xsd/maven-4.0.0.xsd">
+
+  <modelVersion>4.0.0</modelVersion>
+  <groupId>org.eclipse.smarthome.binding</groupId>
+  <artifactId>org.eclipse.smarthome.binding.ntp.tests</artifactId>
```
marcelrv: org.eclipse.smarthome.binding.ntp.test  instead of org.eclipse.smarthome.binding.ntp.test**s**

marcelrv: I'll pull it once more. In the first attempt I did not manage to get these tests to run successfully

petarnv: What was the problem? Did any of the tests fail? Did you manage to run the tests successfully after all?

marcelrv:
LGTM

marcelrv:
@petarnv thanks very much.
Good to have these included

PR #2251, opened 2016-10-04
expose HumanLanguageInterpreters on the REST API

kaikreuzer:

### DIFF
--- a/bundles/core/org.eclipse.smarthome.core.voice/src/main/java/org/eclipse/smarthome/core/voice/VoiceManager.java
+++ b/bundles/core/org.eclipse.smarthome.core.voice/src/main/java/org/eclipse/smarthome/core/voice/VoiceManager.java
@@ -473,7 +473,6 @@ public class VoiceManager {
      * Retrieves a TTS service with the given id.
      *
      * @param id the id of the TTS service
-     *
      * @return a TTS service or null, if no service with this id exists
      */
     public TTSService getTTS(String id) {
@@ -483,7 +482,7 @@ public class VoiceManager {
     /**
      * Retrieves all TTS services.
      *
-     * @return a set of TTS services
+     * @return a collection of TTS services
      */
     public Collection<TTSService> getTTSs() {
         return Collections.unmodifiableCollection(ttsServices.values());
@@ -516,7 +515,6 @@ public class VoiceManager {
      * Retrieves a STT service with the given id.
      *
      * @param id the id of the STT service
-     *
      * @return a STT service or null, if no service with this id exists
      */
     public STTService getSTT(String id) {
@@ -526,7 +524,7 @@ public class VoiceManager {
     /**
      * Retrieves all STT services.
      *
-     * @return a set of STT services
+     * @return a collection of STT services
      */
     public Collection<STTService> getSTTs() {
         return Collections.unmodifiableCollection(sttServices.values());
@@ -559,7 +557,6 @@ public class VoiceManager {
      * Retrieves a KS service with the given id.
      *
      * @param id the id of the KS service
-     *
      * @return a KS service or null, if no service with this id exists
      */
     public KSService getKS(String id) {
@@ -569,7 +566,7 @@ public class VoiceManager {
     /**
      * Retrieves all KS services.
      *
-     * @return a set of KS services
+     * @return a collection of KS services
      */
     public Collection<KSService> getKSs() {
         return Collections.unmodifiableCollection(ksServices.values());
@@ -602,7 +599,6 @@ public class VoiceManager {
      * Retrieves a HumanLanguageInterpreter with the given id.
      *
      * @param id the id of the HumanLanguageInterpreter
-     *
      * @return a HumanLanguageInterpreter or null, if no interpreter with this id exists
      */
     public HumanLanguageInterpreter getHLI(String id) {
@@ -612,7 +608,7 @@ public class VoiceManager {
     /**
      * Retrieves all HumanLanguageInterpreters.
      *
-     * @return a set of HumanLanguageInterpreters
+     * @return a collection of HumanLanguageInterpreters
      */
     public Collection<HumanLanguageInterpreter> getHLIs() {
         return Collections.unmodifiableCollection(humanLanguageInterpreters.values());
--- a/bundles/core/org.eclipse.smarthome.core.voice/src/main/java/org/eclipse/smarthome/core/voice/VoiceManager.java
+++ b/bundles/core/org.eclipse.smarthome.core.voice/src/main/java/org/eclipse/smarthome/core/voice/VoiceManager.java
@@ -9,7 +9,6 @@ package org.eclipse.smarthome.core.voice;
 
 import java.util.ArrayList;
 import java.util.Collection;
-import java.util.Collections;
 import java.util.HashMap;
 import java.util.HashSet;
 import java.util.Locale;
@@ -485,7 +484,7 @@ public class VoiceManager {
      * @return a collection of TTS services
      */
     public Collection<TTSService> getTTSs() {
-        return Collections.unmodifiableCollection(ttsServices.values());
+        return new HashSet<>(ttsServices.values());
     }
 
     /**
@@ -527,7 +526,7 @@ public class VoiceManager {
      * @return a collection of STT services
      */
     public Collection<STTService> getSTTs() {
-        return Collections.unmodifiableCollection(sttServices.values());
+        return new HashSet<>(sttServices.values());
     }
 
     /**
@@ -569,7 +568,7 @@ public class VoiceManager {
      * @return a collection of KS services
      */
     public Collection<KSService> getKSs() {
-        return Collections.unmodifiableCollection(ksServices.values());
+        return new HashSet<>(ksServices.values());
     }
 
     /**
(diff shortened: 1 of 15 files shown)


kaikreuzer:
Note that the VoiceResource is meant to be further extended in future to provide the other services (TTS etc.) as well. Nontheless, the current version should be self-contained and valid to be merged as-is.

@maggu2810, question: I decided to put this under a "/voice" url prefix, i.e. interpreters are at "/voice/interpreters". So far we have most resources directly under the root, but I felt that this might mess up the (Swagger) REST API too much (once TTS and the rest is there as well) and that it is nicer to have such things combined under a common prefix. We didn't do it this way for the rule engine (although I think it might have been a good idea there as well), so things are not completely aligned.
Is this ok from your perspective or shall I remove the "voice" prefix?

maggu2810:
@kaikreuzer I fully agree to use the voice prefix. If you consider to introduce this for the automation I will advocate this, too.

bundles/core/org.eclipse.smarthome.core.voice/src/main/java/org/eclipse/smarthome/core/voice/VoiceManager.java
```diff
@@ -469,6 +470,26 @@ public TTSService getTTS() {
     }
 
     /**
+     * Retrieves a TTS service with the given id.
+     *
+     * @param id the id of the TTS service
+     *
```
maggu2810: I don't know about a strong rule here, but I don't think we need to add empty lines between @param and @return here.
IIRC it is normally not used (e.g. http://www.oracle.com/technetwork/java/javase/documentation/index-137868.html#format) and the documentation is something that is consumed by a doc generator, so not need to add empty lines between tags at all.

bundles/core/org.eclipse.smarthome.core.voice/src/main/java/org/eclipse/smarthome/core/voice/VoiceManager.java
```diff
@@ -492,6 +513,26 @@ public STTService getSTT() {
     }
 
     /**
+     * Retrieves a STT service with the given id.
+     *
+     * @param id the id of the STT service
+     *
```
maggu2810: see above, I will not add any additional comments about the empty line between @param and @return (and other tags).

bundles/core/org.eclipse.smarthome.core.voice/src/main/java/org/eclipse/smarthome/core/voice/VoiceManager.java
```diff
@@ -469,6 +470,26 @@ public TTSService getTTS() {
...
+     * @return a TTS service or null, if no service with this id exists
+     */
+    public TTSService getTTS(String id) {
+        return ttsServices.get(id);
+    }
+
+    /**
+     * Retrieves all TTS services.
+     *
+     * @return a set of TTS services
```
maggu2810: * it is not necessary a "set" but a collection
* should we add the information, that the returned collection will throw an exception if someone tried to modify it (so add it is an unmodifiable collection)

bundles/core/org.eclipse.smarthome.core.voice/src/main/java/org/eclipse/smarthome/core/voice/VoiceManager.java
```diff
@@ -492,6 +513,26 @@ public STTService getSTT() {
...
+     * @return a STT service or null, if no service with this id exists
+     */
+    public STTService getSTT(String id) {
+        return sttServices.get(id);
+    }
+
+    /**
+     * Retrieves all STT services.
+     *
+     * @return a set of STT services
```
maggu2810: * it is not necessary a "set" but a collection
* should we add the information, that the returned collection will throw an exception if someone tried to modify it (so add it is an unmodifiable collection)

bundles/core/org.eclipse.smarthome.core.voice/src/main/java/org/eclipse/smarthome/core/voice/VoiceManager.java
```diff
@@ -515,6 +556,26 @@ public KSService getKS() {
...
+     * @return a KS service or null, if no service with this id exists
+     */
+    public KSService getKS(String id) {
+        return ksServices.get(id);
+    }
+
+    /**
+     * Retrieves all KS services.
+     *
+     * @return a set of KS services
```
maggu2810: see above (unmodifiable collection)

bundles/core/org.eclipse.smarthome.core.voice/src/main/java/org/eclipse/smarthome/core/voice/VoiceManager.java
```diff
@@ -538,6 +599,26 @@ public HumanLanguageInterpreter getHLI() {
...
+     * @return a HumanLanguageInterpreter or null, if no interpreter with this id exists
+     */
+    public HumanLanguageInterpreter getHLI(String id) {
+        return humanLanguageInterpreters.get(id);
+    }
+
+    /**
+     * Retrieves all HumanLanguageInterpreters.
+     *
+     * @return a set of HumanLanguageInterpreters
```
maggu2810: see above (unmodifiable collection)

bundles/io/org.eclipse.smarthome.io.rest.voice/META-INF/MANIFEST.MF
```diff
@@ -0,0 +1,16 @@
...
+Bundle-Vendor: Eclipse.org
+Bundle-RequiredExecutionEnvironment: JavaSE-1.7
+Import-Package: io.swagger.annotations;resolution:=optional,
+ javax.ws.rs,
+ javax.ws.rs.core,
+ org.eclipse.smarthome.core.voice,
+ org.eclipse.smarthome.core.voice.text,
+ org.eclipse.smarthome.io.rest,
+ org.slf4j
+Service-Component: OSGI-INF/voicerest.xml
```
maggu2810: I am fine with that, but you prefer `OSGI-INF/*.xml`

bundles/io/org.eclipse.smarthome.io.rest.voice/src/main/java/org/eclipse/smarthome/io/rest/voice/internal/VoiceResource.java
```diff
@@ -0,0 +1,147 @@
...
+/**
+ * This class acts as a REST resource for voice features.
+ *
+ * @author Kai Kreuzer - Initial contribution and API
+ */
+@Path(VoiceResource.PATH_SITEMAPS)
+@Api(value = VoiceResource.PATH_SITEMAPS)
+public class VoiceResource implements RESTResource {
+
+    public static final String PATH_SITEMAPS = "voice";
```
maggu2810: All fine, but "no modifier" would suffer the visibility, too (https://docs.oracle.com/javase/tutorial/java/javaOO/accesscontrol.html).

bundles/core/org.eclipse.smarthome.core.voice/src/main/java/org/eclipse/smarthome/core/voice/VoiceManager.java
kaikreuzer: Afaik, we nowhere mention unmodfiable collections although we return them at many places. Imho collections returned by getter methods should never be expected to by modifiable by the caller.

bundles/io/org.eclipse.smarthome.io.rest.voice/META-INF/MANIFEST.MF
kaikreuzer: Me too :-)

bundles/io/org.eclipse.smarthome.io.rest.voice/src/main/java/org/eclipse/smarthome/io/rest/voice/internal/VoiceResource.java
kaikreuzer: Right, we are not exposing this Resource (I copied it from other resources, which we often have exported).

kaikreuzer:
Thanks, @maggu2810, I have updated the PR.

bundles/core/org.eclipse.smarthome.core.voice/src/main/java/org/eclipse/smarthome/core/voice/VoiceManager.java
```diff
@@ -469,6 +470,25 @@ public TTSService getTTS() {
...
+        return ttsServices.get(id);
+    }
+
+    /**
+     * Retrieves all TTS services.
+     *
+     * @return a collection of TTS services
+     */
+    public Collection<TTSService> getTTSs() {
+        return Collections.unmodifiableCollection(ttsServices.values());
```
maggu2810: HashMap.values() returns a view of the values.
Collections.unmodifiableCollection(...) returns an unmodifiable view.

So, if you use that construct every change to e.g. ttsServices (an add or remove of a service) will be mirrored to the caller that is using the returned view (only the view is unmodifiable not the collection the view resides on).
If the caller iterates over the returned collection and ttsServices is changes, this will lead to a CME.

Same for the other functions, too.

The normal use case is to create a temp. collection to create the unmodifiable view (so the only one that holds the reference is the unmodifiable view at all).
As you need to copy the values to avoid a CME, there is perhaps no need to use unmodifiable at all and leave the ownership of the collection to the caller.

kaikreuzer: You are so right. Shame on me that I do such beginner mistakes...
Should be fixed now.

maggu2810:
I will merge after Travis CI finishes.
Thanks and sorry that I missed the CME on my first review.

PR #2257, opened 2016-10-05
splitting up DSL Thing configurations into config & properties

### LINKED ISSUE #2215: DSL specified properties are not handled correctly
When we add properties using the ```DiscoveryResultBuilder.withProperties``` method, we can add properties and configuration and when the thing is instantiated from the inbox, (if I remember correctly!) the framework uses the configDescription to decide if the entry is a property or a parameter and moves them into the appropriate place.

This is not the case if things are defined in the DSL - in this case it seems that all properties are defined as configuration. It therefore seems to be impossible to add properties (or correct me if I'm wrong ;) ).

sjka:
...and therefore aligning it with the behavior with was already
in place for Things created by the PersistentInbox

In order to reuse the same code, I had to move it to a place where it can be re-used from both, therefore the ThingHelper seemed the most appropriate place IMHO. I only had to pass the ConfigDescriptionRegistry, as it is not an OSGi service itself.

The majority of the changes are related to the fact that the tests so far didn't have config descriptions, hence the parameters ended up as properties, not configuration anymore.

The GenericThingProvider service had to get a dependency to the ConfigDescriptionRegistry - not an issue dependency-wise as it was like that before. I hope I didn't miss anything why such a service dependency could get us in trouble.

fixes #2215


kaikreuzer:
Hm, I am hesitant to say that the issue description might have been a bit misleading (and thus you took this approach in the PR)...
I think the essential message was "It therefore seems to be impossible to add properties [in DSL-defined Things]."
Looking at the [documentation](https://www.eclipse.org/smarthome/documentation/features/dsl.html), it indeed currently does NOT ALLOW to specify properties for a Thing. Note that "Inside the squared brackets configuration parameters of the thing are defined.".
This PR breaks this, and if there is no config description available, it even ends up creating a thing without any configuration.

Imho the correct solution to the problem would be to extend the DSL grammar to allow specifying properties. Note that DSL-things should be considered as one kind of serialization/persistence of a Thing, just like the ManagedProvider. And within a Thing, configurations and properties are always clearly separated, it is only the DiscoveryResult (i.e. when no Thing has been created at all), which does the automatic separation. But this means that the GenericThingProvider should be aligned with the ManagedThingProvider (instead of the DiscoveryResult) implementation.

sjka:
valid point, thanks. I will create another one then

PR #2260, opened 2016-10-05
Switch Oomph setup model from Java 7 to Java 8

vorburger:


kaikreuzer:
Eclipse SmartHome bundles still have to build with Java7, see https://www.eclipse.org/smarthome/documentation/development/guidelines.html#c-language-levels-and-libraries. So I am afraid, we cannot change the compiler level to Java8 yet.

PR #2263, opened 2016-10-06
Fix NPE in RulesItemRefresher

vorburger:


vorburger:
hang on, don't merge - mixed up two local branches..

vorburger:
OK as intended now... This fixes an NPE that I've ran into during container shutdown.

kaikreuzer:
Thanks!

PR #2264, opened 2016-10-06
.gitignore for distribution/smarthome/userdata

vorburger:
This directory is written into during local development, and not having
this on a .gitignore causes issues when switching git branches etc.


kaikreuzer:
We already have https://github.com/eclipse/smarthome/blob/master/.gitignore#L14. Is this outdated or do you maybe only have these files because you are using an OH IDE and not an ESH IDE?

vorburger:
OK I hadn't seen that (parent) .gitignore .. there is a subtle different, for some reason I don't understand - in my workspace, these mapdb files were in distribution/smarthome/userdata/mapdb, whereas https://github.com/eclipse/smarthome/blob/master/.gitignore#L14 is for distribution/smarthome/runtime/etc/mapdb/ - note the additional /etc/. 

I'm not 100% sure what created them in the non-/etc/ dir. in the first place - I think it got created once when I launched openhab-distro/launch/openHAB_Runtime.launch ... but on retrying this now I can't seem to repro. this anymore?!

Which one is right, with or without that /etc/ sub-dir? Do you want to just close this PR until I hit it again, and can explain what causes this, or shall I change this PR to just add the ignore to the non-etc location as well, as that seems to happen some times?

kaikreuzer:
Feel free to add it to the root .gitignore file, if the files are indeed written in your case.

PR #2266, opened 2016-10-06
check for uniqueness of channels

### LINKED ISSUE #2210: Removal of dynamically created channels
At present, if we need to remove the channels which were created dynamically, we need to get the whole channel list using the getChannels() method, convert the immutable list to mutable, remove the channels from the created mutable list and then use the build() method to create the thing again. It would be good if a removeChannel() method is provided.

### LINKED ISSUE #2256: Duplicate channels on Thing
A thing's channels are stored as a List<Channel> , see 

https://github.com/eclipse/smarthome/blob/e0985e90d389350580e59bfbcd7e0adaaccf7519/bundles/core/org.eclipse.smarthome.core.thing/src/main/java/org/eclipse/smarthome/core/thing/internal/ThingImpl.java#L49

Because this is a list, its possible for a Thing to have duplicate channels with the same ID. This also means that calling "getChannel(String id)" will only return the first instance of the duplicate channel. My assumption would be that a thing can only have a single instance of a channel.

sjka:
...in ThingHelper and ThingBuilder so no duplicate channels can be created.

The ChannelUID is used as the only parameter as it defines the identity
of a Channel.

And while touching it anyway, I also added a "removeChannel" method to the ThingBuilder.

fixes #2256
fixes #2210


bundles/core/org.eclipse.smarthome.core.thing/src/main/java/org/eclipse/smarthome/core/thing/util/ThingHelper.java
```diff
@@ -90,10 +95,28 @@ private static String toString(List<Channel> channels) {
...
         ((ThingImpl) thing).getChannelsMutable().addAll(channels);
     }
 
     /**
+     * Ensures that there are no duplicate channels in the list (i.e. not using the same ChannelUID)
+     *
+     * @param channels the {@link List} of channels to check
+     * @throws IllegalArgumentException in case there are duplicate channels found
+     */
+    public static void ensureUnique(List<Channel> channels) {
```
maggu2810: You do you use a List as argument and create a temporary copy of a collection in all the methods above?
Wouldn't it be much simpler to use a "Collection" as argument and loop through all members of the collection?

sjka: You have to distinguish: In ThingBuilder this was done before, to ensure the Thing gets a copy so it cannot be modified. I conveniently use it and only update the Thing with it if it's correct.
In the `ThingHelper. addChannelsToThing()` above it's true, I created that temporary copy. I did that in order to do the check _before_ the actual collection gets modified. There is no value in throwing an IllegalArgumentException if we accept the new Channel anyway...

bundles/core/org.eclipse.smarthome.core.thing/src/main/java/org/eclipse/smarthome/core/thing/binding/builder/ThingBuilder.java
```diff
@@ -56,18 +60,34 @@ public ThingBuilder withLabel(String label) {
...
 
     public ThingBuilder withChannels(List<Channel> channels) {
-        this.thing.setChannels(Lists.newArrayList(channels));
+        ArrayList<Channel> channelList = Lists.newArrayList(channels);
+        ThingHelper.ensureUnique(channelList);
+        this.thing.setChannels(channelList);
+        return this;
+    }
+
+    public ThingBuilder removeChannel(ChannelUID channelUID) {
```
sjka: Yes'n'yes

bundles/core/org.eclipse.smarthome.core.thing/src/main/java/org/eclipse/smarthome/core/thing/util/ThingHelper.java
sjka: Thanks @maggu2810, I took the freedom to include your suggestions here (and adding an also-by line to the commit). Hope that's okay?

maggu2810: Sure, if you are fine with it.

PR #2270, opened 2016-10-06
Scheduler : fix #2105

### LINKED ISSUE #2105: ESH Scheduler: Cron expression has no future
I am using the ESH scheduler (thanks for your work @kgoderis) and add a job using a cron expression.
To execute the job every second I used this expression: `* * * * * ?'`

It runs for about two minutes and stops then
```text
14:37:39.259 DEBUG o.e.s.c.p.PersistenceManager[:298] - Scheduled strategy everySecond with cron expression * * * * * ?
```

```text
14:39:17.561 INFO  o.e.s.c.s.ExpressionThreadPoolManager[:178] - Expression '* * * * * ?' has no future executions anymore
```

@kgoderis Can you explain me, why this crone expression has no future executions anymore?

kgoderis:

### DIFF
--- a/bundles/core/org.eclipse.smarthome.core.test/src/test/java/org/eclipse/smarthome/core/scheduler/CronExpressionTest.java
+++ b/bundles/core/org.eclipse.smarthome.core.test/src/test/java/org/eclipse/smarthome/core/scheduler/CronExpressionTest.java
@@ -13,7 +13,6 @@ import java.text.ParseException;
 import java.util.Calendar;
 import java.util.Date;
 
-import org.eclipse.smarthome.core.scheduler.CronExpression;
 import org.junit.Test;
 
 public class CronExpressionTest {
@@ -63,4 +62,22 @@ public class CronExpressionTest {
 
         assertEquals(checkDate, nextDate);
     }
+
+    @Test
+    public void runForever() throws ParseException, InterruptedException {
+
+        final CronExpression expression;
+        expression = new CronExpression("* * * * * ?");
+
+        Date nextDate = expression.getTimeAfter(Calendar.getInstance().getTime());
+        int counter = 1;
+
+        while (nextDate != null && counter <= 150) {
+            System.out.println("value " + counter + " is " + nextDate);
+            nextDate = expression.getTimeAfter(nextDate);
+            counter++;
+        }
+
+    }
+
 }
--- a/bundles/core/org.eclipse.smarthome.core/src/main/java/org/eclipse/smarthome/core/scheduler/AbstractExpression.java
+++ b/bundles/core/org.eclipse.smarthome.core/src/main/java/org/eclipse/smarthome/core/scheduler/AbstractExpression.java
@@ -29,7 +29,7 @@ public abstract class AbstractExpression<E extends AbstractExpressionPart> imple
 
     private final Logger logger = LoggerFactory.getLogger(this.getClass().getName());
 
-    private int minimumCandidates = 10;
+    private int minimumCandidates = 1;
     private int maximumCandidates = 100;
 
     private String expression;
@@ -207,22 +207,49 @@ public abstract class AbstractExpression<E extends AbstractExpressionPart> imple
 
     @Override
     public Date getTimeAfter(Date afterTime) {
-        if (getCandidates().isEmpty()) {
+
+        Date currentStartDate = getStartDate();
+
+        if (hasFloatingStartDate()) {
+            try {
+                clearCandidates();
+                setStartDate(afterTime);
+            } catch (IllegalArgumentException | ParseException e) {
+                logger.error("An exception occurred while setting the start date : '{}'", e.getMessage());
+            }
+        } else if (getCandidates().isEmpty()) {
             try {
                 setStartDate(afterTime);
-                parseExpression(expression);
             } catch (ParseException e) {
-                logger.error("An exception occurred while parsing the expression : '{}'", e.getMessage());
+                logger.error("An exception occurred while setting the start date : '{}'", e.getMessage());
             }
         }
 
         if (!getCandidates().isEmpty()) {
+            if (getCandidates().size() == 1) {
+                return getCandidates().get(0);
+            } else {
+                while (getCandidates().size() > 1) {
 
-            Collections.sort(getCandidates());
+                    Collections.sort(getCandidates());
 
-            for (Date candidate : getCandidates()) {
-                if (candidate.after(afterTime)) {
-                    return candidate;
+                    Date newStartDate = null;
+
+                    try {
+
+                        for (Date candidate : getCandidates()) {
+                            newStartDate = candidate;
+                            if (candidate.after(afterTime)) {
+                                setStartDate(currentStartDate);
+                                return candidate;
+                            }
+                        }
+
+                        clearCandidates();
+                        setStartDate(newStartDate);
+                    } catch (IllegalArgumentException | ParseException e) {
+                        logger.error("An exception occurred while parsing the expression : '{}'", e.getMessage());
+                    }
                 }
             }
         }
@@ -232,6 +259,9 @@ public abstract class AbstractExpression<E extends AbstractExpressionPart> imple
 
     @Override
     public Date getFinalFireTime() {
+
+        Date currentStartDate = getStartDate();
+
         if (getCandidates().isEmpty()) {
             try {
                 parseExpression(getExpression());
@@ -240,11 +270,36 @@ public abstract class AbstractExpression<E extends AbstractExpressionPart> imple
             }
         }
 
-        if (getCandidates().isEmpty()) {
-            return null;
+        Date lastCandidate = null;
+
+        if (!getCandidates().isEmpty()) {
+            if (getCandidates().size() == 1) {
+                lastCandidate = getCandidates().get(0);
+            } else {
+
+                while (getCandidates().size() == maximumCandidates) {
+                    Collections.sort(getCandidates());
+                    lastCandidate = getCandidates().get(getCandidates().size() - 1);
+                    try {
+                        clearCandidates();
+                        setStartDate(lastCandidate);
+                    } catch (IllegalArgumentException | ParseException e) {
+                        logger.error("An exception occurred while parsing the expression : '{}'", e.getMessage());
+                    }
+
+                }
+
+                lastCandidate = getCandidates().get(getCandidates().size() - 1);
+
+                try {
+                    setStartDate(currentStartDate);
+                } catch (IllegalArgumentException | ParseException e) {
+                    logger.error("An exception occurred while setting the start date : '{}'", e.getMessage());
+                }
+            }
         }
 
-        return getCandidates().get(getCandidates().size() - 1);
+        return lastCandidate;
     }
 
     /**
@@ -273,6 +328,10 @@ public abstract class AbstractExpression<E extends AbstractExpressionPart> imple
 
     protected void setCandidates(ArrayList<Date> candidates) {
         this.candidates = candidates;
+    }
+
+    protected void clearCandidates() {
+        this.candidates = null;
     }
 
     public ArrayList<E> getExpressionParts() {
--- a/bundles/core/org.eclipse.smarthome.core/src/main/java/org/eclipse/smarthome/core/scheduler/CronExpression.java
+++ b/bundles/core/org.eclipse.smarthome.core/src/main/java/org/eclipse/smarthome/core/scheduler/CronExpression.java
@@ -169,7 +169,7 @@ public final class CronExpression extends AbstractExpression<CronExpressionPart>
      * @throws ParseException if the string expression cannot be parsed into a valid <code>CronExpression</code>.
      */
     public CronExpression(final String expression, final Date startTime, final TimeZone zone) throws ParseException {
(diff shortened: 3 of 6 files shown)


kgoderis:
@maggu2810 Can you try out this fix for https://github.com/eclipse/smarthome/pull/2094

bundles/core/org.eclipse.smarthome.core/src/main/java/org/eclipse/smarthome/core/scheduler/AbstractExpression.java
```diff
@@ -207,22 +207,51 @@ protected void prune() {
...
+        if (hasFloatingStartDate()) {
+            try {
+                clearCandidates();
+                setStartDate(afterTime);
+            } catch (IllegalArgumentException | ParseException e) {
+                logger.error("An exception occurred while setting the start date : '{}'", e.getMessage());
+            }
+        }
+
         if (getCandidates().isEmpty()) {
```
maggu2810: We can use a `else if` here. If it is a foating start date, set start date has been already called. If this results in an empty candidates list, is there any reason to call it again?

bundles/core/org.eclipse.smarthome.core/src/main/java/org/eclipse/smarthome/core/scheduler/AbstractExpression.java
```diff
@@ -207,22 +207,51 @@ protected void prune() {
...
-
-            for (Date candidate : getCandidates()) {
-                if (candidate.after(afterTime)) {
-                    return candidate;
+            if (getCandidates().size() == 1) {
+                return getCandidates().get(0);
+            } else {
+                while (getCandidates().size() > 1) {
+
+                    Collections.sort(getCandidates());
```
maggu2810: Wouldn't it make sense to move the sort method to the place the candidates are generated instead of every getTimeAfter call?

bundles/core/org.eclipse.smarthome.core/src/main/java/org/eclipse/smarthome/core/scheduler/AbstractExpression.java
```diff
@@ -207,22 +207,51 @@ protected void prune() {
...
-
-            Collections.sort(getCandidates());
-
-            for (Date candidate : getCandidates()) {
-                if (candidate.after(afterTime)) {
-                    return candidate;
+            if (getCandidates().size() == 1) {
+                return getCandidates().get(0);
+            } else {
+                while (getCandidates().size() > 1) {
```
maggu2810: Why do you use > 1 here?
If this loop is iterated multiple times, isn't it valid, that the candidates consist of exactly one element?
At the end of the iteration you call `setStartDate(newStartDate);` I assume this could result in a candidates list of the minimum size (1, now).
Shouldn't it be `!getCandidates().isEmpty()`?

bundles/core/org.eclipse.smarthome.core/src/main/java/org/eclipse/smarthome/core/scheduler/AbstractExpression.java
```diff
@@ -240,11 +272,36 @@ public Date getFinalFireTime() {
...
-            return null;
+        Date lastCandidate = null;
+
+        if (!getCandidates().isEmpty()) {
+            if (getCandidates().size() == 1) {
+                lastCandidate = getCandidates().get(0);
+            } else {
+
+                while (getCandidates().size() == maximumCandidates) {
+                    Collections.sort(getCandidates());
```
maggu2810: See comment above

bundles/core/org.eclipse.smarthome.core/src/main/java/org/eclipse/smarthome/core/scheduler/AbstractExpression.java
```diff
@@ -240,11 +272,36 @@ public Date getFinalFireTime() {
...
-        if (getCandidates().isEmpty()) {
-            return null;
+        Date lastCandidate = null;
+
+        if (!getCandidates().isEmpty()) {
+            if (getCandidates().size() == 1) {
+                lastCandidate = getCandidates().get(0);
+            } else {
+
+                while (getCandidates().size() == maximumCandidates) {
```
maggu2810: If we used a floating start date, couldn't the size be equal to maximum candidates, too?
You set the minimum number in the constructor to 0 for CronExpression, but is this a grant, that there aren't the maximum number be used?

bundles/core/org.eclipse.smarthome.core.test/src/test/java/org/eclipse/smarthome/core/scheduler/CronExpressionTest.java
```diff
@@ -63,4 +62,22 @@ public void getFinalTimeCheck() throws ParseException {
 
         assertEquals(checkDate, nextDate);
     }
+
+    @Test
+    public void runForever() throws ParseException, InterruptedException {
```
maggu2810: You shouldn't name a test "run forever", this could be a little bit misleading :wink:

maggu2810:
@kgoderis "* * * * * ?" runs for a while now. Thanks.
Can you have a look at the comments?

bundles/core/org.eclipse.smarthome.core/src/main/java/org/eclipse/smarthome/core/scheduler/AbstractExpression.java
kgoderis: size()==1 is to cover DateExpression or alike, which always have getCandidates().size()=1.

In the end we could indeed arrive at a candidate list that contains a single candidate, being the startDate itself. So, if I am not wrong, getCandidates() should never be empty

bundles/core/org.eclipse.smarthome.core/src/main/java/org/eclipse/smarthome/core/scheduler/AbstractExpression.java
kgoderis: Possibly, to be checked that nowhere else we modify the order of that Collection.

bundles/core/org.eclipse.smarthome.core/src/main/java/org/eclipse/smarthome/core/scheduler/AbstractExpression.java
kgoderis: I am not sure I fully understand your comment here. For cron the overall number of candidates can be limited (as per Quartz definition), but for practical reasons, in case there are unlimited number of candidates (like in your infamous example), the YEARS parameters is limited to 2100 anyways. So, infinite series are finite from a practical point of view.

kgoderis: I try to understand why you try to differentiate between floating start date and non-floating start date wrt getfinalfiretime()....

bundles/core/org.eclipse.smarthome.core/src/main/java/org/eclipse/smarthome/core/scheduler/AbstractExpression.java
maggu2810: I comment the else branch, so let's have a look at the code in that branch only (`getCandidates().size() == 1` is checked outside of the else branch).

```java
while (getCandidates().size() > 1) {

    Collections.sort(getCandidates());

    Date newStartDate = null;

    try {

        for (Date candidate : getCandidates()) {
            newStartDate = candidate;
            if (candidate.after(afterTime)) {
                setStartDate(currentStartDate);
                return candidate;
            }
        }

        clearCandidates();
        setStartDate(newStartDate);
    } catch (IllegalArgumentException | ParseException e) {
        logger.error("An exception occurred while parsing the expression : '{}'", e.getMessage());
    }
}
```

If the last call to `setStartDate(newStartDate);` results in a candidate list of size 1, the condition if false and the while loop body is not executed again. Shouldn't also the one remaining candidate be checked if it is after the after time and returned if this is true.

bundles/core/org.eclipse.smarthome.core/src/main/java/org/eclipse/smarthome/core/scheduler/AbstractExpression.java
maggu2810: > the YEARS parameters is limited to 2100 anyways

Didn't know that cron ends working in the year 2100 :wink:

bundles/core/org.eclipse.smarthome.core/src/main/java/org/eclipse/smarthome/core/scheduler/AbstractExpression.java
maggu2810: > It should be called again

Does this mean if `setStartDate(afterTime);` results in an empty candidate list another call to `setStartDate(afterTime)` (using the same aster time) can result in an non empty candidate list? Or why should it be called again?

bundles/core/org.eclipse.smarthome.core/src/main/java/org/eclipse/smarthome/core/scheduler/AbstractExpression.java
maggu2810: > I try to understand why you try to differentiate between floating start date and non-floating start date wrt getfinalfiretime()....

By a mistake :wink:

bundles/core/org.eclipse.smarthome.core/src/main/java/org/eclipse/smarthome/core/scheduler/AbstractExpression.java
kgoderis: Damn. I meant  "should not be called again". my mistake

bundles/core/org.eclipse.smarthome.core/src/main/java/org/eclipse/smarthome/core/scheduler/AbstractExpression.java
kgoderis: Yes, that single final date is checked, but in the previous iteration of the loop, in the "for (Date candidate : getCandidates()) " loop

bundles/core/org.eclipse.smarthome.core/src/main/java/org/eclipse/smarthome/core/scheduler/AbstractExpression.java
kgoderis: > Didn't know that cron ends working in the year 2100

It is the to limit the algorithm. Maybe 2100 is even too long, but one has to be optimist and believe that your ESH/OH2 instance can run unattended and without rebooting for the next 84 years. Maybe for performance reasons we can limit that something like "current time" + 1 year, no?

bundles/core/org.eclipse.smarthome.core/src/main/java/org/eclipse/smarthome/core/scheduler/AbstractExpression.java
maggu2810: Hm, I do not get it.

Isn't at the end in the while loop by using `clearCandidates()` and `setStartDate(newStartDate)` the candidate set new calculated?

Then the next iteration starts with the check for &gt 1...

```java
// First loop iteration
// Let's assume: getCandidates().size() == 5 (at least something > 1)

// The condition "getCandidates().size() > 1" is checked
// All fine


    Collections.sort(getCandidates());
    Date newStartDate = null;
    try {
        for (Date candidate : getCandidates()) {
            newStartDate = candidate;
            if (candidate.after(afterTime)) {
                setStartDate(currentStartDate);
                return candidate;
            }
        }

// No of the current 5 candidates is after(afterTime)
// So we are here

// Clear the candidates and get new candidates by setting the start date
        clearCandidates();
        setStartDate(newStartDate);
// Let's assume the number of caluclated candidates 1: getCandidates().size() == 1

    } catch (IllegalArgumentException | ParseException e) {
        logger.error("An exception occurred while parsing the expression : '{}'", e.getMessage());
    }

// Second iteration loop
// The condition "getCandidates().size() > 1" is checked
// The size is 1 so not greater then 1 and the loop is not entered.
```

bundles/core/org.eclipse.smarthome.core/src/main/java/org/eclipse/smarthome/core/scheduler/AbstractExpression.java
maggu2810: @kgoderis Do you consider to change the `if` to `else if`? I think the remaining stuff is fine and we could merge this PR.

bundles/core/org.eclipse.smarthome.core/src/main/java/org/eclipse/smarthome/core/scheduler/AbstractExpression.java
maggu2810: So, should we keep the sort ATM as it is? And have a look at performance improvements later? Okay for me.

kgoderis:
> So, should we keep the sort ATM as it is? And have a look at performance improvements later? Okay for me.

Yeah - let's do that. Will push a mod for the "else if" in a second and you are good to go. Sorry for the delay, I am currently swamped

PR #2286, opened 2016-10-10
changed volume of AudioSink from float to PercentType

kaikreuzer:
This is imho more in line with our other interfaces where we use our own type system.


kgoderis:
@kaikreuzer in that case the Sonos binding needs to be updated as well, no?

kaikreuzer:
No, as it does not yet implement the AudioSink (I am still working on implementing the infrastructure based on your PoC - this PR is one outcome of it).

PR #2287, opened 2016-10-10
Introduction of an AudioServlet to provide audio streams through HTTP

kaikreuzer:

### DIFF
--- a/bundles/core/org.eclipse.smarthome.core.audio/src/main/java/org/eclipse/smarthome/core/audio/internal/AudioServlet.java
+++ b/bundles/core/org.eclipse.smarthome.core.audio/src/main/java/org/eclipse/smarthome/core/audio/internal/AudioServlet.java
@@ -1,0 +1,229 @@
+/**
+ * Copyright (c) 2014-2016 by the respective copyright holders.
+ * All rights reserved. This program and the accompanying materials
+ * are made available under the terms of the Eclipse Public License v1.0
+ * which accompanies this distribution, and is available at
+ * http://www.eclipse.org/legal/epl-v10.html
+ */
+package org.eclipse.smarthome.core.audio.internal;
+
+import java.io.IOException;
+import java.net.InetAddress;
+import java.net.MalformedURLException;
+import java.net.URL;
+import java.net.UnknownHostException;
+import java.util.Hashtable;
+import java.util.Map;
+import java.util.UUID;
+import java.util.concurrent.ConcurrentHashMap;
+
+import javax.servlet.ServletException;
+import javax.servlet.ServletOutputStream;
+import javax.servlet.http.HttpServlet;
+import javax.servlet.http.HttpServletRequest;
+import javax.servlet.http.HttpServletResponse;
+
+import org.apache.commons.io.IOUtils;
+import org.apache.commons.lang.StringUtils;
+import org.eclipse.smarthome.core.audio.AudioFormat;
+import org.eclipse.smarthome.core.audio.AudioHTTPServer;
+import org.eclipse.smarthome.core.audio.AudioStream;
+import org.osgi.framework.BundleContext;
+import org.osgi.service.http.HttpContext;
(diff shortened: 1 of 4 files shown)


bundles/core/org.eclipse.smarthome.core.audio/src/main/java/org/eclipse/smarthome/core/audio/internal/AudioServlet.java
```diff
@@ -0,0 +1,229 @@
...
+    @Override
+    protected void doGet(HttpServletRequest req, HttpServletResponse resp) throws ServletException, IOException {
+        String streamId = StringUtils.substringAfterLast(req.getRequestURI(), "/");
+        if (!streams.containsKey(streamId)) {
+            logger.debug("Received request for invalid stream id at  {}", req.getRequestURI());
+            resp.sendError(HttpServletResponse.SC_NOT_FOUND);
+        } else {
+            logger.debug("Stream to serve is {}", streamId);
+
+            AudioStream stream = streams.get(streamId);
```
maggu2810: Shouldn't we use `streams.remove(streamId)` here?

kaikreuzer: Definitely yes! Good catch, I only changed that while debugging and missed to replace it again.

kaikreuzer: fixed.

kgoderis:
@kaikreuzer Kai, what is the rationale behind the fact that a stream will only be served a single time?

kaikreuzer:
Security, Resource-cleanup, etc.

Note that I am aware that this does not work for Sonos as they always do two requests.
I plan to introduce further methods like `serve(AudioStream stream, int times, int maxSeconds)` or something like that to also cover cases like Sonos. Makes sense?

kgoderis:
So far, with Sonos, in the PoC, I have taken the route that each player encodes a wav file to be served. If we play or stream or TTS to a bunch of Sonos at the same time, we effectively have a bunch of identical files (apart from their temp. filenames) (well, Audiostreams of course) to be served 2 times. It could be an optimisation, when/if we allow wildcards to be used in play(..) in Rules - and in the context of audioservlet - to generate/copy over the wav only once but stream it many times. Or, foresee a mechanism that an Audiostream can be served to a player that is not necessarily linked to the ThingHandler that made the original serve() call. Maybe it was your intention to *defacto* serve any AudioStream, and somehow embed an URL in the object, so that URL-driven players can simply pick up the URL on the servlet without further processing or copying to be done? e.g. All Audiostreams get  served whether we want it or not, and when passing on the AudioStream to the Sink, let the Sink decide wether to take the AudioStream or directly move on with the URL that is provided

 I am only afraid that when you have a large number of Sinks that need to play the same stream, that because of encoding or other delays, the streams will not be played in sync. Ok, Sonos has the ZoneGroup stuff, but that is not necessarily true for other brands of players. 

Just thinking out loud.

kaikreuzer:
The AudioSink has to deal with any kind of AudioStream that is passed to it, so we cannot assume that it is a file-based stream. Writing any stream to disk just to be potentially able to serve it multiple times imho does not make much sense. Keeping a copy of the whole stream in memory doesn't make sense either. My current approach would be to use "resettable" streams, so that the same stream can be served multiple times (and it is up to the implementation of the stream whether it holds the data in memory or on the file system or anywhere else). I noticed that resettable might cause concurrency issues though, so it might be better to introduce something like stream cloning without duplicating the data (does anything like this exist in Java already...?).

You can get an idea of how Sonos AudioSink support may look like in my branch at https://github.com/kaikreuzer/smarthome/commit/f75b78192917b567848729824aa6c37adcee9588. Feel free to comment.

PR #2300, opened 2016-10-11
Added X-Accel-Buffering=no header to SSE response in order to disable…

kolkoo:
… response buffering when using nginx as a proxy server.

This allows you to use nginx proxy buffering and still have working SSE.

Signed-off-by: IVAN GEORGIEV ILIEV <ivan.iliev@musala.com>

### DIFF
--- a/bundles/io/org.eclipse.smarthome.io.rest.sse/src/main/java/org/eclipse/smarthome/io/rest/sse/SseResource.java
+++ b/bundles/io/org.eclipse.smarthome.io.rest.sse/src/main/java/org/eclipse/smarthome/io/rest/sse/SseResource.java
@@ -51,6 +51,8 @@ public class SseResource {
 
     public final static String PATH_EVENTS = "events";
 
+    private static final String X_ACCEL_BUFFERING_HEADER = "X-Accel-Buffering";
+
     private final SseBroadcaster broadcaster;
 
     private final ExecutorService executorService;
@@ -95,6 +97,10 @@ public class SseResource {
         // the given filter
         final EventOutput eventOutput = new SseEventOutput(eventFilter);
         broadcaster.add(eventOutput);
+
+        // Disables proxy buffering when using an nginx http server proxy for this response.
+        // This allows you to not disable proxy buffering in nginx and still have working sse
+        response.addHeader(X_ACCEL_BUFFERING_HEADER, "no");
 
         if (!SseUtil.SERVLET3_SUPPORT) {
             // if we don't have sevlet 3.0 async support, we want to make sure


kolkoo:
More info about this here http://nginx.org/en/docs/http/ngx_http_proxy_module.html#proxy_buffering

sjka:
lgtm

maggu2810:
@kaikreuzer Do you wait for tests?

maggu2810:
@kaikreuzer Do you wait for tests or Travis?

kaikreuzer:
Was waiting for Travis to finish.

BClark09:
This doesn't seem to be coming through for basicUI. Have tried a NGINX setup with and without buffering on the latest apt openhab. When buffering is off, eventsource items don't come through, and nothing is updated.

image

kaikreuzer:
@kolkoo Could you check and comment?

kolkoo:
@kaikreuzer will do

kolkoo:
@kaikreuzer @BClark09  Correct me if I'm wrong but the basic UI seems to only be subscribing to rest/sitemaps/events and this PR is about /rest/events . 

I've tested PaperUI which uses /rest/events with and without nginx proxy and with or without buffering enabled and it works in all scenarios.

I'm not sure what the problem with BasicUI is but it's definitely not related to this PR.

kolkoo:
@kaikreuzer Actually because this fix was only for /rest/events, naturally it won't work for sitemaps event as they are in a different resource. I also see that servlet 2.4 support will not work for them either, so perhaps it makes sense to decide if both fixes need to be transferred.

BClark09:
Hi @kolkoo, thanks for explaining the issue and my misunderstanding. You are correct, this PR adresses only the events in Paper UI which are indeed now working correctly with buffering off. I had assumed that events from basic UI were obtained in the same way and therefore this PR would resolve this issue.

It's clear that a new issue needs to be created but what is the issue here? Is it that the sitemap events need to be updated to match the rest events or simply just that the same X-Accel-Buffering header needs to be set for sitemap events?

For reference the nginx proxy configuration to see this issue for sitemaps is:

```nginx
server {
        listen                          80;
        server_name                     ***.***.***.***;

        location / {
                proxy_pass                              http://localhost:8080/;
#               proxy_buffering                         off;    #uncomment to get live events.
                proxy_set_header Host                   $http_host;
                proxy_set_header X-Real-IP              $remote_addr;
                proxy_set_header X-Forwarded-For        $proxy_add_x_forwarded_$
                proxy_set_header X-Forwarded-Proto      $scheme;
        }
}
```

kolkoo:
Hey @BClark09 if with proxy_buffering off; everything works fine for you, and it only breaks with buffering on then we just need to add the header to the sitemap events resource method for SSE, just as we've done here. Then it should work in both situations.

BClark09:
That makes sense, I see that it should work with the same two lines in the relative place for [SitemapResource](https://github.com/eclipse/smarthome/blob/master/bundles/io/org.eclipse.smarthome.io.rest.sitemap/src/main/java/org/eclipse/smarthome/io/rest/sitemap/internal/SitemapResource.java) but don't know for certain. Would you be able to make a new PR for this? I'd be more than grateful.

kaikreuzer:
> I also see that servlet 2.4 support will not work for them either, so perhaps it makes sense to decide if both fixes need to be transferred.

I do not really think that servlet 2.4 support is required in this UI, so if you could provide a fix for the buffering on the SitemapResource, it would be fine!

Issue #2282, opened 2016-10-09
Test failure(s): esh.core.thing: FirmwareUpdateServiceOSGiTest


maggu2810:
* Commit: 2538c1d
* Travis job: https://travis-ci.org/eclipse/smarthome/builds/166162273
* raw console log: https://s3.amazonaws.com/archive.travis-ci.org/jobs/166162274/log.txt

Excerpt(s):
```text
Failed tests: 
  FirmwareUpdateServiceOSGiTest.setup:237->OSGiTest.waitForAssert:-1->OSGiTest.waitForAssert:181 
Expected: is <3>
     but: was <2>
```
```text
java.lang.AssertionError: 
Expected: is <3>
     but: was <2>
	at org.hamcrest.MatcherAssert.assertThat(MatcherAssert.java:20)
	at org.junit.Assert.assertThat(Assert.java:865)
	at org.junit.Assert.assertThat(Assert.java:832)
	at org.junit.Assert$assertThat.callStatic(Unknown Source)
	at org.eclipse.smarthome.core.thing.firmware.FirmwareUpdateServiceOSGiTest$_setup_closure4.doCall(FirmwareUpdateServiceOSGiTest.groovy:242)
	at org.eclipse.smarthome.core.thing.firmware.FirmwareUpdateServiceOSGiTest$_setup_closure4.doCall(FirmwareUpdateServiceOSGiTest.groovy)
	at sun.reflect.GeneratedMethodAccessor45.invoke(Unknown Source)
	at sun.reflect.DelegatingMethodAccessorImpl.invoke(DelegatingMethodAccessorImpl.java:43)
	at java.lang.reflect.Method.invoke(Method.java:497)
	at org.codehaus.groovy.reflection.CachedMethod.invoke(CachedMethod.java:90)
	at groovy.lang.MetaMethod.doMethodInvoke(MetaMethod.java:324)
	at org.codehaus.groovy.runtime.metaclass.ClosureMetaClass.invokeMethod(ClosureMetaClass.java:292)
	at groovy.lang.MetaClassImpl.invokeMethod(MetaClassImpl.java:1015)
	at org.codehaus.groovy.runtime.callsite.PogoMetaClassSite.call(PogoMetaClassSite.java:39)
	at org.codehaus.groovy.runtime.callsite.CallSiteArray.defaultCall(CallSiteArray.java:45)
	at org.codehaus.groovy.runtime.callsite.AbstractCallSite.call(AbstractCallSite.java:108)
[...]

PR #2301, opened 2016-10-11
Fix FirmwareUpdateServiceOSGiTest - separate waitForAssert to check t…

tomhoefer:
…hing status and initial firmware status propagation

Signed-off-by: Thomas Höfer <t.hoefer@telekom.de>


tomhoefer:
Should fix #2282

PR #2302, opened 2016-10-11
Some improvements on audio streams

kaikreuzer:
- Introduced FixedLengthAudioStream
- Provide Content-Length in AudioServlet
- Ignore file extensions in urls in AudioServlet
- added ability to reset a FileAudioStream

### DIFF
--- a/bundles/core/org.eclipse.smarthome.core.audio/src/main/java/org/eclipse/smarthome/core/audio/FileAudioStream.java
+++ b/bundles/core/org.eclipse.smarthome.core.audio/src/main/java/org/eclipse/smarthome/core/audio/FileAudioStream.java
@@ -13,27 +13,33 @@ import java.io.FileNotFoundException;
 import java.io.IOException;
 import java.io.InputStream;
 
+import org.apache.commons.io.IOUtils;
+
 /**
  * This is an AudioStream from an audio file
  *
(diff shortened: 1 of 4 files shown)


bundles/core/org.eclipse.smarthome.core.audio/src/main/java/org/eclipse/smarthome/core/audio/FileAudioStream.java
```diff
@@ -13,27 +13,33 @@
...
         this.audioFormat = getAudioFormat(file);
+        this.length = file.length();
     }
 
     private static AudioFormat getAudioFormat(File file) throws AudioException {
         if (file.getName().toLowerCase().endsWith(".wav")) {
             return new AudioFormat(AudioFormat.CONTAINER_WAVE, AudioFormat.CODEC_PCM_SIGNED, false, 16, 705600, 44100L);
         } else if (file.getName().toLowerCase().endsWith(".mp3")) {
-            return new AudioFormat(AudioFormat.CODEC_MP3, AudioFormat.CODEC_MP3, null, null, null, null);
+            return new AudioFormat(AudioFormat.CONTAINER_NONE, AudioFormat.CODEC_MP3, null, null, null, null);
```
maggu2810: We could use AudioFormat.MP3 here, but I don't think that matters.

PR #2305, opened 2016-10-12
added support to serve a single AudioStream multiple times concurrently through HTTP

kaikreuzer:

### DIFF
--- a/bundles/core/org.eclipse.smarthome.core.audio/src/main/java/org/eclipse/smarthome/core/audio/FixedLengthAudioStream.java
+++ b/bundles/core/org.eclipse.smarthome.core.audio/src/main/java/org/eclipse/smarthome/core/audio/FixedLengthAudioStream.java
@@ -10,7 +10,8 @@ package org.eclipse.smarthome.core.audio;
 import java.io.InputStream;
 
 /**
- * This is an {@link AudioStream}, which can provide information about its absolute length and is able to
+ * This is an {@link AudioStream}, which can provide information about its absolute length and is able to provide
+ * cloned streams.
  *
  * @author Kai Kreuzer - Initial contribution and API
  *
--- a/bundles/core/org.eclipse.smarthome.core.audio/src/main/java/org/eclipse/smarthome/core/audio/FixedLengthAudioStream.java
+++ b/bundles/core/org.eclipse.smarthome.core.audio/src/main/java/org/eclipse/smarthome/core/audio/FixedLengthAudioStream.java
@@ -7,8 +7,10 @@
  */
 package org.eclipse.smarthome.core.audio;
 
+import java.io.InputStream;
+
 /**
- * This is an {@link AudioStream}, which can provide information about its absolute length.
+ * This is an {@link AudioStream}, which can provide information about its absolute length and is able to
  *
  * @author Kai Kreuzer - Initial contribution and API
  *
@@ -17,9 +19,17 @@ abstract public class FixedLengthAudioStream extends AudioStream {
 
     /**
      * Provides the length of the stream in bytes.
-     * 
+     *
      * @return absolute length in bytes
      */
     public abstract long length();
 
+    /**
+     * Returns a new, fully independent stream instance, which can be read and closed without impacting the original
+     * instance.
+     *
+     * @return a new input stream that can be consumed by the caller
+     * @throws AudioException if stream cannot be created
+     */
+    public abstract InputStream getClonedStream() throws AudioException;
 }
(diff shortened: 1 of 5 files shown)


bundles/core/org.eclipse.smarthome.core.audio/src/main/java/org/eclipse/smarthome/core/audio/FixedLengthAudioStream.java
```diff
@@ -7,8 +7,10 @@
  */
 package org.eclipse.smarthome.core.audio;
 
+import java.io.InputStream;
+
 /**
- * This is an {@link AudioStream}, which can provide information about its absolute length.
+ * This is an {@link AudioStream}, which can provide information about its absolute length and is able to
```
maggu2810: "... be cloned" or something similar is missing.

bundles/core/org.eclipse.smarthome.core.audio/src/main/java/org/eclipse/smarthome/core/audio/internal/AudioServlet.java
```diff
@@ -37,7 +39,7 @@
 import org.slf4j.LoggerFactory;
 
 /**
- * A servlet that serves audio streams via HTTP.
+ * A servlet that serves audio oneTimeStreams via HTTP.
```
maggu2810: We should keep "streams" here. I assume this has been done by the refactor procedure.

bundles/core/org.eclipse.smarthome.core.audio/src/main/java/org/eclipse/smarthome/core/audio/internal/AudioServlet.java
```diff
@@ -127,32 +140,50 @@ protected void doGet(HttpServletRequest req, HttpServletResponse resp) throws Se
...
-        } else {
-            // TODO: We should also support streams without mark support, but this will mean that we have to read it
-            // first to memory or file system and create a new AudioStream from there.
-            logger.warn("Stream cannot be reset, so it is served without size through HTTP");
-            return serve(stream);
-        }
+    public URL serve(FixedLengthAudioStream stream, int seconds) {
+        String streamId = UUID.randomUUID().toString();
+        multiTimeStreams.put(streamId, stream);
+        streamTimeouts.put(streamId, System.currentTimeMillis() + (seconds * 1000L));
```
maggu2810: I would prefer to use `System.nanoTime()` for such time, as it is independent of the system time and cannot be corrupted by time changes.

bundles/core/org.eclipse.smarthome.core.audio/src/main/java/org/eclipse/smarthome/core/audio/internal/AudioServlet.java
```diff
@@ -127,32 +140,50 @@ protected void doGet(HttpServletRequest req, HttpServletResponse resp) throws Se
...
+            }
+            IOUtils.copy(is, os);
             resp.flushBuffer();
             IOUtils.closeQuietly(stream);
         }
     }
 
+    private synchronized void removeTimedOutStreams() {
+        for (String streamId : multiTimeStreams.keySet()) {
+            if (streamTimeouts.get(streamId) < System.currentTimeMillis()) {
```
maggu2810: See comment below about `System.nanoTime()`

bundles/core/org.eclipse.smarthome.core.audio/src/main/java/org/eclipse/smarthome/core/audio/internal/AudioServlet.java
```diff
@@ -127,32 +140,50 @@ protected void doGet(HttpServletRequest req, HttpServletResponse resp) throws Se
...
+                // the stream has expired, we need to remove it!
+                FixedLengthAudioStream stream = multiTimeStreams.remove(streamId);
+                streamTimeouts.remove(streamId);
+                IOUtils.closeQuietly(stream);
+                stream = null;
```
maggu2810: There is no need to set the local scope reference to null, but if you prefer...

bundles/core/org.eclipse.smarthome.core.audio/src/main/java/org/eclipse/smarthome/core/audio/internal/AudioServlet.java
```diff
@@ -127,32 +140,50 @@ protected void doGet(HttpServletRequest req, HttpServletResponse resp) throws Se
...
+                    is = ((FixedLengthAudioStream) stream).getClonedStream();
+                } catch (AudioException e) {
+                    resp.sendError(HttpServletResponse.SC_INTERNAL_SERVER_ERROR, e.getMessage());
+                }
+            } else {
+                is = stream;
+            }
+            IOUtils.copy(is, os);
             resp.flushBuffer();
             IOUtils.closeQuietly(stream);
```
maggu2810: We should close `is` now to ensure the cloned stream if `multiAccess` is true (if it is false `is` is equal to `stream`).

kaikreuzer: good catch!

kaikreuzer:
Thanks for the review - PR is updated.

PR #2309, opened 2016-10-12
persisting group functions in the ManagedItemProvider

### LINKED ISSUE #2269: group function is not correctly (at all...) persisted in the mapdb
After setting up and restarting ESH the groups created via the new group item function of Paper UI the values like sum, avg and so on get lost.

sjka:
fixes #2269

### DIFF
--- a/bundles/core/org.eclipse.smarthome.core/src/main/java/org/eclipse/smarthome/core/items/ManagedItemProvider.java
+++ b/bundles/core/org.eclipse.smarthome.core/src/main/java/org/eclipse/smarthome/core/items/ManagedItemProvider.java
@@ -8,6 +8,7 @@
 package org.eclipse.smarthome.core.items;
 
 import java.util.ArrayList;
+import java.util.Arrays;
 import java.util.Collection;
 import java.util.HashSet;
 import java.util.Iterator;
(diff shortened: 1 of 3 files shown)


bundles/core/org.eclipse.smarthome.core/src/main/java/org/eclipse/smarthome/core/items/ManagedItemProvider.java
```diff
@@ -36,7 +39,7 @@
  * @author Kai Kreuzer - improved return values
  * @author Alex Tugarev - added tags
  */
-public class ManagedItemProvider extends AbstractManagedProvider<Item, String, PersistedItem> implements ItemProvider {
+public class ManagedItemProvider extends AbstractManagedProvider<Item, String, PersistedItem>implements ItemProvider {
```
kaikreuzer: Are you using Eclipse Mars instead of Neo or why does your auto formatter removing the space?

sjka:
IDE update done - space came back magically. Thanks for the hint

kaikreuzer:
Many thanks for the fix 👍

hfanieng:
Also many thanks from me!

PR #2310, opened 2016-10-12
Audio servlet improvements

maggu2810:
The three commits does not need to be squashed.
The first one is completely independent.
The third one requires the second one, but as long as the commits are left in order, I assume the interface change could be committed separately.


maggu2810:
I already considered that is has been done by intention but I don't see any reason.
The AudioStream already implements the InputStream interface itself and to give a new AudioStream doesn't limit any usage.
In the example you given the rawAudio is a immutable (at least if I understand its usage) private byte array so it could be easily shared between instances.

maggu2810:
I reverted the interface changes and reordered so code to be able to use try-with-resources.
So we should have no resources leaks also with runtime exceptions.

kaikreuzer:
FTR: I used squashed merge because there were revert commits in this PR now.

maggu2810:
:+1:

PR #2311, opened 2016-10-13
adapted MacTTS to use FixedLengthAudioStream

kaikreuzer:

### DIFF
--- a/extensions/voice/org.eclipse.smarthome.voice.mactts/src/main/java/org/eclipse/smarthome/voice/mactts/internal/MacTTSAudioStream.java
+++ b/extensions/voice/org.eclipse.smarthome.voice.mactts/src/main/java/org/eclipse/smarthome/voice/mactts/internal/MacTTSAudioStream.java
@@ -9,12 +9,14 @@ package org.eclipse.smarthome.voice.mactts.internal;
 
 import java.io.File;
 import java.io.FileInputStream;
+import java.io.FileNotFoundException;
 import java.io.IOException;
 import java.io.InputStream;
 
 import org.eclipse.smarthome.core.audio.AudioException;
 import org.eclipse.smarthome.core.audio.AudioFormat;
 import org.eclipse.smarthome.core.audio.AudioStream;
+import org.eclipse.smarthome.core.audio.FixedLengthAudioStream;
 import org.eclipse.smarthome.core.voice.Voice;
 
 /**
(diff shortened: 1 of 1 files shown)


extensions/voice/org.eclipse.smarthome.voice.mactts/src/main/java/org/eclipse/smarthome/voice/mactts/internal/MacTTSAudioStream.java
```diff
@@ -74,19 +79,31 @@ private InputStream createInputStream() throws AudioException {
...
             throw new AudioException("The '" + command + "' has been interrupted", e);
         }
     }
 
+    private InputStream getFileInputStream(File file) throws AudioException {
+        if (file == null) {
+            throw new IllegalArgumentException("file must not be null");
+        }
+        if (file.exists()) {
+            this.length = file.length();
```
maggu2810: The initialization of length needs to be done only once and not all the time `getClonedStream()` is called.

If we would like to handle file length changes it does not work (IIRC) as the file size could change also after the file input stream has been created.

So, I think this line could be moved above after `file = new File(outputFile);`
Or is it because you don't want to handle FileNotFoundException twice?
But reset the size all the time... Hm

WDYT?

extensions/voice/org.eclipse.smarthome.voice.mactts/src/main/java/org/eclipse/smarthome/voice/mactts/internal/MacTTSAudioStream.java
```diff
@@ -74,19 +79,33 @@ private InputStream createInputStream() throws AudioException {
         try {
             Process process = Runtime.getRuntime().exec(command);
             process.waitFor();
-            File file = new File(outputFile);
-            if (file.exists()) {
-                return new FileInputStream(file);
-            } else {
-                throw new AudioException("Temporary file '" + outputFile + "' not found!");
+            file = new File(outputFile);
+            if (file != null) {
```
maggu2810: `new File(...)` should never return null but always a reference (if no exception is thrown by the construction).

If the file does not exist 0L is returned, so I think we do not need any check at all.

kaikreuzer: ok, even better. Updated.

PR #2313, opened 2016-10-13
initial contribution of a web audio sink

kaikreuzer:
This is a new audio sink implementation which simply creates an event with a url for the audio stream to be played from. This event can be retrieved by web clients through SSE.

As a first user of this feature, I would like to see the Paper UI as it already subscribes to all SSE events.

### DIFF
--- a/extensions/io/org.eclipse.smarthome.io.webaudio/src/main/java/org/eclipse/smarthome/io/webaudio/internal/SSEAudioSink.java
+++ b/extensions/io/org.eclipse.smarthome.io.webaudio/src/main/java/org/eclipse/smarthome/io/webaudio/internal/SSEAudioSink.java
@@ -12,6 +12,7 @@ import java.util.HashSet;
 import java.util.Locale;
 import java.util.Set;
 
+import org.apache.commons.io.IOUtils;
 import org.eclipse.smarthome.core.audio.AudioFormat;
 import org.eclipse.smarthome.core.audio.AudioHTTPServer;
 import org.eclipse.smarthome.core.audio.AudioSink;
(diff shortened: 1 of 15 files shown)


extensions/io/org.eclipse.smarthome.io.webaudio/src/main/java/org/eclipse/smarthome/io/webaudio/internal/SSEAudioSink.java
```diff
@@ -0,0 +1,118 @@
...
+        supportedFormats.add(AudioFormat.WAV);
+        supportedFormats.add(AudioFormat.MP3);
+    }
+
+    private AudioHTTPServer audioHTTPServer;
+
+    private EventPublisher eventPublisher;
+
+    @Override
+    public void process(AudioStream audioStream) throws UnsupportedAudioFormatException {
```
maggu2810: Can you give me some details how the audio stream handling is defined for that function?
If someone passes a audio stream to a process method is the audio stream 
* closed by the method in all cases
* closed by the method if no exception is thrown
* should be closed by the caller

I don't see a consistent behaviour here.
* if it is an instance of URLAudioStream it is closed
* if it is an instance of FixedLengthAudioStream it is closed after max. 10 secnonds
* in all other cases it is not closed

As a AudioStream interface extend the InputStream interface it should be closed independent of the specific implementation.
Or miss I something here?

maggu2810: I think I missed a little bit an overview which function that takes an `AudioStream` take the ownership of the stream and close the stream and which functions leave it to the caller to call close.

kaikreuzer: No, you are absolutely right, I missed closing the stream in the "all other" case.
In general, it is the same as with InputStreams: A method that receives a stream must make sure that it is closed; either by itself or by a function to which the stream is passed on.
Does the code look better now?

maggu2810: > A method that receives a stream must make sure that it is closed; either by itself or by a function to which the stream is passed on.

Okay, this is a clear definition.
Thanks.

PR #2314, opened 2016-10-13
Support audio events

aounhaider1:


aounhaider1:
It does not currently work on Safari.

kaikreuzer:
Tested on Chrome and Firefox.
Feel free to come up with follow up PRs if there are ways to also make this work in Safari, IE and Edge.

Issue #2219, opened 2016-09-27
build error: npm


maggu2810:
Hi, it seems that Travis CI and Hudson can build the project. So it could be related to my local machine, but I have no clue (ATM) what causes the error. Perhaps someone can help.

I cannot build the Paper UI anymore.

I already checked, that I have a clean working copy of the sources:

``` sh
rm -rf extensions/ui/org.eclipse.smarthome.ui.paper
git checkout extensions/ui/org.eclipse.smarthome.ui.paper
```

But this does not solve the problem.
I always run into the same error.

``` text
rathgeb@pc05 ~/bin/pkgs/eclipse/esh/git/smarthome/extensions/ui/org.eclipse.smarthome.ui.paper $ mvn clean install
Java HotSpot(TM) 64-Bit Server VM warning: ignoring option PermSize=512m; support was removed in 8.0
Java HotSpot(TM) 64-Bit Server VM warning: ignoring option MaxPermSize=1g; support was removed in 8.0
[INFO] Scanning for projects...
[INFO] Computing target platform for MavenProject: org.eclipse.smarthome.extension.ui:org.eclipse.smarthome.ui.paper:0.9.0-SNAPSHOT @ /home/rathgeb/bin/pkgs/eclipse/esh/git/smarthome/extensions/ui/org.eclipse.smarthome.ui.paper/pom.xml
[INFO] Fetching p2.index from http://download.eclipse.org/tools/orbit/downloads/drops/R20140525021250/repository/ (0B at 0B/s)
[INFO] Adding repository http://download.eclipse.org/tools/orbit/downloads/drops/R20140525021250/repository
[INFO] Fetching p2.index from http://download.eclipse.org/eclipse/updates/4.4/ (0B at 0B/s)
[INFO] Adding repository http://download.eclipse.org/eclipse/updates/4.4
[INFO] Fetching p2.index from http://download.eclipse.org/modeling/tmf/xtext/updates/releases/2.9.2/ (0B at 0B/s)
[INFO] Adding repository http://download.eclipse.org/modeling/tmf/xtext/updates/releases/2.9.2
[INFO] Adding repository http://download.eclipse.org/modeling/emft/mwe/updates/releases/2.8.3
[INFO] Adding repository http://download.eclipse.org/modeling/emf/emf/updates/2.11/core/R201506010402
[INFO] Adding repository http://dist.springsource.org/release/GRECLIPSE/e4.4
[...]

PR #2317, opened 2016-10-13
fix Paper UI build

maggu2810:
I cannot grant that this is a grant fix, but I ran into #2219 now multiple times.
I used a shell while loop to restart `mvn clean install` as long as it does not succeed.

The while loop never finished since hours.

After this fix I can do a clean build again.

I don't know which code of Maven or some of the plugins clean up my target while building but I think this commit makes the situation not more wore and for me it helps (at least today).

Fixes: #2219

### DIFF
--- a/extensions/ui/org.eclipse.smarthome.ui.paper/.gitignore
+++ b/extensions/ui/org.eclipse.smarthome.ui.paper/.gitignore
@@ -1,3 +1,5 @@
+npm_cache/
+npm_tmp/
 node/
 node_modules/
 web-src/bower_components/
--- a/extensions/ui/org.eclipse.smarthome.ui.paper/.project
+++ b/extensions/ui/org.eclipse.smarthome.ui.paper/.project
@@ -38,7 +38,7 @@
 	</natures>
 	<filteredResources>
 		<filter>
-			<id>1465846929365</id>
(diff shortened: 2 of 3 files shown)


kaikreuzer:
I never had issue #2219 and I have no clue what this PR is actually doing. So I'll leave it to @aounhaider1 to comment and am fine with any outcome :-)

maggu2810:
npm downloads the stuff it uses to a temporary directoy.
Normally it is the temporary directory.
Caused by [this issue](https://github.com/npm/npm/issues/6855) we changed that to the build directory.

IMHO: The import change for us has been to use a fixed directory all the time, so not multiple ones are created that are never cleaned up.

There is no need to download the stuff on every build again (the build directory is cleaned up on the `clean` phase) and as it seems there are corner cases the build directory is cleaned up while the frontend-maven-plugin is running...

I am happy to wait for some feedback of @aounhaider1.

kaikreuzer:
@aounhaider1 Still waiting for your approval of this change.

aounhaider1:
Yes, changing tmp directory path should not affect anything. I don't exactly understand why maven build would fail in case of no npm tmp directory but if this fixes the issue then its fine.

PR #2318, opened 2016-10-13
[LIFX] FindBugs issues fix

wborn:
* Unlock lightCounterLock when exceptions occur
* Potential NPE in handlePowerStatus

Signed-off-by: Wouter Born <eclipse@maindrain.net>

### DIFF
--- a/extensions/binding/org.eclipse.smarthome.binding.lifx/src/main/java/org/eclipse/smarthome/binding/lifx/handler/LifxLightHandler.java
+++ b/extensions/binding/org.eclipse.smarthome.binding.lifx/src/main/java/org/eclipse/smarthome/binding/lifx/handler/LifxLightHandler.java
@@ -233,11 +233,16 @@ public class LifxLightHandler extends BaseThingHandler {
                     .setOption(StandardSocketOptions.SO_REUSEADDR, true)
                     .setOption(StandardSocketOptions.SO_BROADCAST, true);
             broadcastChannel.configureBlocking(false);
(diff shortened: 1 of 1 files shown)


extensions/binding/org.eclipse.smarthome.binding.lifx/src/main/java/org/eclipse/smarthome/binding/lifx/handler/LifxLightHandler.java
```diff
@@ -864,8 +869,7 @@ public void handlePowerStatus(StatePowerResponse packet) {
             updateState(CHANNEL_COLOR, currentColorState);
             updateState(CHANNEL_BRIGHTNESS, currentColorState.getBrightness());
         } else {
-            updateState(CHANNEL_COLOR,
-                    new HSBType(currentColorState.getHue(), currentColorState.getSaturation(), PercentType.HUNDRED));
+            updateState(CHANNEL_COLOR, HSBType.WHITE);
```
kaikreuzer: You are changing the functionality here. Before, what was happening is that the brightness was set to 100%, keeping hue and saturation values as before. With your code, your are always setting the color to white instead, which is imho not correct.

kaikreuzer: Ok, you have a point there, thanks. @kgoderis I assume you are ok with this PR?

PR #2320, opened 2016-10-14
Added SatisfiableRESTResource and SatisfiableResourceFilter and refac…

kolkoo:
…tored all ESH RESTResource that have mandatory static dependencies to optional and dynamic and made them implement the new interface.

This is done in order to avoid deactivation/reactivation of RESTResources whenever some of their dependencies go missing as this in turn causes the entire Jersey model to reload which can be very very costly on weak CPU devices.
Instead when a dependency goes missing the SatisfiableResourceFilter will call the newly introduced SatisfiableRESTResource#isSatisfied before the matched resource method is executed and will return 503 Unavailable if the service is not satisfied.
Currently when a RESTResource is deactivated and Jersey is reloading the JAX RS OSGi Connector implementation will return 503 so the behaviour is consistent.

Signed-off-by: IVAN GEORGIEV ILIEV <ivan.iliev@musala.com>


kolkoo:
I've also fixed a *sometimes* failing test in ThingManagerOSGiTest as part of the PR hope that's ok.

tomhoefer:
lgtm

kaikreuzer:
Hey @kolkoo, thanks for this PR.
Just for my understanding: You introduced the new interface in order to prevent the rest services to be called, if they are lacking a dependency, right? Without it, you would have had to change all the services to check themselves whether they have all services that they need for a certain call and return an HTTP error if that's not the case (which probably would be quite some more code changes).
Is there any code left using still the "plain" RESTResource interface? If not, wouldn't it have been an option to add the isSatisfied method simply on that interface without creating a new one?

kolkoo:
Hey Kai, 
Yes, I introduced the interface + filter in order to achieve this in a more global way without having to have a check in each resource method in each resource class.

I've left it as an interface extension because other people using ESH and implementing RESTResource will not be forced to re-implement their service and the code remains backward compatible.

Also there are a few RESTREsources in ESH that do not have any mandatory static dependencies so I've left them as RESTResources only. This way their execution will be slightly faster as they will fail in the filter in the instanceof check and we won't need to also cast them to the SatisfiableRESTResource interface.

kaikreuzer:
Hm, I'd wish we were already on Java8, then you could add isSatisfied to the RESTResource interface with a default implementation that returns true - this would keep backward compatibility in place and all would be fine...

kolkoo:
Well if you decide that backwards compatibility is not that important we can always just add it to RESTResource.

kaikreuzer:
No, I said it is important and I wonder if we should wait for Java 8 ;-)

tomhoefer:
In my opinion it is worth to merge this PR. We can easily change the implementation once our project is running on Java 8

kaikreuzer:
Ok, so let's merge this, but remove the new interface again once we switch to Java8. I have added a link to here on https://github.com/eclipse/smarthome/wiki/Java-8-CP2-cleanup.

PR #2323, opened 2016-10-14
improved logic to retrieve local ip address

kaikreuzer:

### DIFF
--- a/bundles/core/org.eclipse.smarthome.core.audio/src/main/java/org/eclipse/smarthome/core/audio/internal/AudioServlet.java
+++ b/bundles/core/org.eclipse.smarthome.core.audio/src/main/java/org/eclipse/smarthome/core/audio/internal/AudioServlet.java
@@ -9,10 +9,13 @@ package org.eclipse.smarthome.core.audio.internal;
 
 import java.io.IOException;
 import java.io.InputStream;
+import java.net.Inet6Address;
 import java.net.InetAddress;
 import java.net.MalformedURLException;
+import java.net.NetworkInterface;
+import java.net.SocketException;
 import java.net.URL;
-import java.net.UnknownHostException;
+import java.util.Enumeration;
 import java.util.Hashtable;
 import java.util.Map;
 import java.util.UUID;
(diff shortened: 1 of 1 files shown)


bundles/core/org.eclipse.smarthome.core.audio/src/main/java/org/eclipse/smarthome/core/audio/internal/AudioServlet.java
```diff
@@ -195,16 +198,50 @@ public URL serve(FixedLengthAudioStream stream, int seconds) {
...
+        } catch (MalformedURLException e) {
             logger.error("Failed to construct audio stream URL: {}", e.getMessage(), e);
             return null;
         }
     }
 
+    /**
+     * Finds the (non loopback, non localhost) local network interface.
+     */
+    public String getLocalNetworkInterface() {
```
maggu2810: IMHO we mix "interface" and "IPv4" address in this context.

bundles/core/org.eclipse.smarthome.core.audio/src/main/java/org/eclipse/smarthome/core/audio/internal/AudioServlet.java
```diff
@@ -195,16 +198,50 @@ public URL serve(FixedLengthAudioStream stream, int seconds) {
...
+        try {
+            String localInterface = null;
+            Enumeration<NetworkInterface> interfaces = NetworkInterface.getNetworkInterfaces();
+            while (interfaces.hasMoreElements()) {
+                NetworkInterface current = interfaces.nextElement();
+                if (!current.isUp() || current.isLoopback() || current.isVirtual()) {
```
sjka: r u sure it's always a good idea to ignore virtual interfaces?
also, as I mentioned in #2331, IMHO it would be good to have a configuration option to override this automatism, as with multiple interfaces such algorithms tend to always pick the wrong one

kaikreuzer:
Closing as it is superseded by #2331.

PR #2331, opened 2016-10-16
improve "get port" mechanism

maggu2810:
It is not mandatory that the system property is used for setting the
service port. So we should also respect the service property.

The configuration admin is e.g. used by Pax Web.

See also: http://felix.apache.org/documentation/subprojects/apache-felix-http-service.html
The service can both be configured using OSGi environment properties and
using Configuration Admin. ... If you use both methods, Configuration Admin
takes precedence.


maggu2810:
Should we move such a utility function to the core bundle so it could be used from other places, too?

maggu2810:
Same here: https://github.com/eclipse/smarthome/blob/447fd63/bundles/io/org.eclipse.smarthome.io.rest.mdns/src/main/java/org/eclipse/smarthome/io/rest/mdns/internal/MDNSAnnouncer.java#L50 (and needs to be done for the secure port, too)

If you give me a place to put such util methods, I will implement that changes.
E.g. class org.eclipse.smarthome.core.net.NetUtils

maggu2810:
* created utility classes
* add logic from #2323 
* fix mDNS REST announcer to consider configuration admin config for service, too

bundles/core/org.eclipse.smarthome.core/src/main/java/org/eclipse/smarthome/core/net/HttpServiceUtil.java
```diff
@@ -0,0 +1,93 @@
...
+     */
+    public static int getHttpServicePort(final BundleContext bc) {
+        return getHttpServicePortProperty(bc, "org.osgi.service.http.port");
+    }
+
+    public static int getHttpServicePortSecure(final BundleContext bc) {
+        return getHttpServicePortProperty(bc, "org.osgi.service.http.port.secure");
+    }
+
+    // Utility class that could be used for non-secure and secure port.
```
sjka: I don't get that comment...

bundles/core/org.eclipse.smarthome.core/src/main/java/org/eclipse/smarthome/core/net/HttpServiceUtil.java
```diff
@@ -0,0 +1,93 @@
...
+                    serviceRanking = 0;
+                } else {
+                    serviceRanking = (Integer) value;
+                }
+                if (serviceRanking >= candidate) {
+                    candidate = serviceRanking;
+                    port = servicePort;
+                }
+            }
+        } catch (final InvalidSyntaxException ex) {
```
sjka: I know, this in theory cannot happen. However, if it does for whatever crazy circumstances, it's hard do find. So IMHO it can't hurt to log it anyway....

bundles/core/org.eclipse.smarthome.core/src/main/java/org/eclipse/smarthome/core/net/HttpServiceUtil.java
```diff
@@ -0,0 +1,93 @@
...
+        }
+
+        // If the service does not provide the port, try to use the system property.
+        value = bc.getProperty(propertyName);
+        if (value != null) {
+            if (value instanceof String) {
+                try {
+                    return Integer.parseInt(value.toString());
+                } catch (final NumberFormatException ex) {
+                }
```
sjka: Same here: it's a property set by the user - if this property is set with bogus, somebody will be pretty disappointed if it gets ignored silently

bundles/core/org.eclipse.smarthome.core/src/main/java/org/eclipse/smarthome/core/net/NetUtil.java
```diff
@@ -0,0 +1,63 @@
...
+import org.slf4j.LoggerFactory;
+
+/**
+ * Some utility functions related to network interfaces etc.
+ *
+ * @author Markus Rathgeb - Initial contribution and API
+ */
+public class NetUtil {
+
+    private static final Logger LOGGER = LoggerFactory.getLogger(NetUtil.class);
```
sjka: Isn't the convention in ESH to use non-static loggers?

bundles/core/org.eclipse.smarthome.core/src/main/java/org/eclipse/smarthome/core/net/NetUtil.java
```diff
@@ -0,0 +1,63 @@
...
+
+    private NetUtil() {
+    }
+
+    /**
+     * Get the first candidate for a local IPv4 host address (non loopback, non localhost).
+     */
+    public static String getLocalIpv4HostAddress() {
```
sjka: somehow this looks weird to me, IMHO getLocalIPv4HostAddress is more brain-friendly, but that could just be personal preference...

bundles/core/org.eclipse.smarthome.core/src/main/java/org/eclipse/smarthome/core/net/NetUtil.java
```diff
@@ -0,0 +1,63 @@
...
+ */
+public class NetUtil {
+
+    private static final Logger LOGGER = LoggerFactory.getLogger(NetUtil.class);
+
+    private NetUtil() {
+    }
+
+    /**
+     * Get the first candidate for a local IPv4 host address (non loopback, non localhost).
```
sjka: Running the stuff on a server with several network interfaces, such automatisms tend to always pick the wrong one. Shouldn't we also provide some configuration options (i.e. ConfigAdmin + system property) to override this?

bundles/core/org.eclipse.smarthome.core/src/main/java/org/eclipse/smarthome/core/net/NetUtil.java
```diff
@@ -0,0 +1,63 @@
...
+     */
+    public static String getLocalIpv4HostAddress() {
+        try {
+            String hostAddress = null;
+            final Enumeration<NetworkInterface> interfaces = NetworkInterface.getNetworkInterfaces();
+            while (interfaces.hasMoreElements()) {
+                final NetworkInterface current = interfaces.nextElement();
+                if (!current.isUp() || current.isLoopback() || current.isVirtual()) {
```
sjka: are you sure that ignoring virtual interfaces is correct?

sjka: Oh, just realized I'm asking the wrong one... [#2323](https://github.com/eclipse/smarthome/pull/2323/files#diff-1e31442968454e1584828a3f2443a48bR224) ;-)

bundles/core/org.eclipse.smarthome.core/src/main/java/org/eclipse/smarthome/core/net/HttpServiceUtil.java
maggu2810: The logic for "org.osgi.service.http.port" and "org.osgi.service.http.port.secure" is the same only the property name differs. So a private method `getHttpServicePortProperty` has been used that takes the property as argument.

bundles/core/org.eclipse.smarthome.core/src/main/java/org/eclipse/smarthome/core/net/HttpServiceUtil.java
maggu2810: You are right, a log is normally okay.
But using a log in the static method involves the creation of a static logger -- but okay, this could be done, too.

If you have a look at the API documentation the `InvalidSyntaxException` is thrown "If the specified filter contains an invalid filter expression that cannot be parsed."
filter is null and so not used :wink: 
Perhaps a line comment in the code would be nice to explain why the exception is ignored silently.

bundles/core/org.eclipse.smarthome.core/src/main/java/org/eclipse/smarthome/core/net/HttpServiceUtil.java
maggu2810: We would like to parse the port number used by the HTTP servlet.
IMHO If the port is configured in the wrong way, the HTTP servlet has to care about and should report a warning.
As we only consume what the servlet is using -- and if the servlet cannot consume the property, I think it is okay, to ignore it, too.

bundles/core/org.eclipse.smarthome.core/src/main/java/org/eclipse/smarthome/core/net/NetUtil.java
maggu2810: Yes, definitely.
This method has been introduced by Kai.
I moved the method an utility class and make this utility function static.
So, how to use a non-static logger in a static context?

We agreed that in static context we can use a static logger.

bundles/core/org.eclipse.smarthome.core/src/main/java/org/eclipse/smarthome/core/net/NetUtil.java
maggu2810: As written above and comment in the PR itself, this method (and its whole logic) has been migrated from #2323

bundles/core/org.eclipse.smarthome.core/src/main/java/org/eclipse/smarthome/core/net/NetUtil.java
maggu2810: I don't care much. There are different meanings about correct camel case usage.
I can change this.

bundles/core/org.eclipse.smarthome.core/src/main/java/org/eclipse/smarthome/core/net/NetUtil.java
maggu2810: > Oh, just realized I'm asking the wrong one... #2323 ;-)

:wink:

bundles/core/org.eclipse.smarthome.core/src/main/java/org/eclipse/smarthome/core/net/HttpServiceUtil.java
sjka: Ah, okay, got it. Then please change _class_ to  _method_.

bundles/core/org.eclipse.smarthome.core/src/main/java/org/eclipse/smarthome/core/net/HttpServiceUtil.java
sjka: > If the port is configured in the wrong way, the HTTP servlet has to care about and should report a warning.
good point

bundles/core/org.eclipse.smarthome.core/src/main/java/org/eclipse/smarthome/core/net/HttpServiceUtil.java
sjka: I see your point. Nevertheless, having spent way to much time analyzing weird bugs searching in the dark, I would prefer the ad-hoc logger creation in that case. As it's a will-never-happen-if-we-can-rely-on-laws-of-physics-Exception this will not hurt here.

bundles/core/org.eclipse.smarthome.core/src/main/java/org/eclipse/smarthome/core/net/NetUtil.java
sjka: okay

bundles/core/org.eclipse.smarthome.core/src/main/java/org/eclipse/smarthome/core/net/HttpServiceUtil.java
maggu2810: Ah, didn't see that type. Thanks.

bundles/core/org.eclipse.smarthome.core/src/main/java/org/eclipse/smarthome/core/net/HttpServiceUtil.java
maggu2810: Done

maggu2810:
@kaikreuzer Anything that blocks this PR to be merged?

kaikreuzer:
only my holidays - please allow me some time to also have a look at it and review it. It shouldn't be anything urgent, or is it?

Issue #583, opened 2015-12-03
Add support for properties on ItemChannelLinks


kaikreuzer:
migrated from Bugzilla [#456213](https://bugs.eclipse.org/bugs/show_bug.cgi?id=456213)
status NEW severity _enhancement_ in component _Core_ for _---_
Reported in version _unspecified_ on platform _All_
Assigned to: Project Inbox

On 2014-12-27 03:59:13 -0500, Gerhard Riegler wrote:

> A item should have item properties. Let's say a DateTime channel is bound to two items, but for the second item the user wants a item specific offset from the channel DateTime value. This implies, that there should be a possibility to update the state item specific too or, maybe the better way, a callback method in the ThingHandler for each item which is updated with the item properties to have a chance to manipulate the state before it is published to the bus.

On 2014-12-27 15:49:53 -0500, Kai Kreuzer wrote:

> Not the item should have these properties, but the ItemChannelLink (since the configuration can be specific for each channel an item is linked to).

watou:
Many OH1 bindings retain a copy of the bound item's acceptedDataTypes list, so when data arrives, the binding can create an acceptable state with which to update the item.  In some cases the binding has no knowledge of the intended type other than the list of accepted types for the item.  In ESH currently, such a binding would have to define duplicate channels for every possible type.  If the ItemChannelLink knew the linked item's accepted data types and imposed no other constraint, bindings like these could be written for the ESH model without creating duplicate channels for every possible type.

cdjackson:
I would argue that for the stated example above at least, that the conversion should be provided by a middle layer rather than in the binding. I'm not sure where in the system translations sit these days, but I would argue that the binding should provide data in some format (or formats) and it shouldn't be modified each time a user wants a new conversion. I would suggest that such conversions are performed outside of the binding, so that they are (possibly) more user definable, and in any case are available to all bindings/items.

Maybe the above example doesn't represent all possible use cases for this, and maybe there's still a case for passing properties into the binding, but maybe it's not needed?

If we start passing properties (configuration?) from an item, then I assume we also need a configProvider for items?

watou:
[...]

PR #2334, opened 2016-10-17
Converting state into accepted types

### LINKED ISSUE #2253: ThingManager should adapt state updates to linked item type
I have a hue light defined like this in hue.things:

```
Bridge hue:bridge:xxx "Philips Hue Bridge" [ ipAddress="192.168.1.3", userName="xxx" ] {
    0210 1 "Hue White and Color - Living Room Ceiling" @ "Living Room" [ lightId="1" ]
```

I want to have an on/off switch on my sitemap, so I defined an item for it as

```
Switch living_light_ceiling_switch "Living Room Ceiling Switch" <light> (gLights) { channel="hue:0210:xxx:1:color" }
```

I read in the documentation that I need to use the :color channel instead of :switch to make it work. So now I have a sitemap definition as

```
Switch living_light_ceiling_switch "Living Room Ceiling Switch" <light> (gLights) { channel="hue:0210:xxx:1:color" }
```

It works nicely as far as turning the light on and off from the UI but if I try to turn the light on/off from the real Hue app the state doesn't get updated in the classic UI or HABdroid. I see in the basic UI that it gets updated with its HSB value (68,88,0) instead of an ON/OFF state.

This is problematic because in openHAB 1.x I had quite a few rules depending on the ON/OFF state of the item, for example if conditions like (living_light_command is another switch item):

```
if (living_light_ceiling_switch.state != living_light_command.state) { ... }
```

 and now those are not working at all. Can this be fixed in the binding? Either create/enable the :switch channel for the thing, or make the :color channel converted to ON/OFF states when the items is a simple switch.

Issue #866 might be related, but I thought I better open a new issue for this, as that one is not really about this switch state problem. I understand that rewriting my rules to somehow check for the brightness value might work, but it's far from ideal and probably I'm not the only one who will be hit by this during transition to openHAB2.

Oh, and I'm using beta4.

sjka:
This PR is intended to deal with the fact that users might be so creative and link e.g. a SwitchItem to a color channel which maintains a HSBType state. In order to get these items updated successfully whenever values change, the `State` object needs to be converted into something accepted (if possible at all...).

When implementing this, 
- I needed access to the linked items (and not only their names). Therefore I changed the `AbstractLinkRegistry.getLinkedItems()` method to `AbstractLinkRegistry.getLinkedItemNames()` because that is what it actually does and introduced a `ItemChannelLinkRegistry.getLinkedItems()` method with returns what it promises. The alternative would have been to re-add the `ThingManager` -> `ItemRegistry` dependency which I assume we all are happy was dropped.
- I decided to do the conversion in all three places: the `ThingManager`, `ItemUpdater` and `AutoUpdater`. Sending the state update event "as is" and doing the conversion in the `ItemUpdater` would lead to UIs still getting "wrong" events so they won't benefit from it. And doing it only in the `ThingManager` is not enough as there might be other origins for events having the "wrong" type. The `ItemUtil` seems like a good place for it.
- I had to change the order of the `acceptedDataTypes` in some items so they have the most meaningful type at the top of the list. I'm very well aware of the `TypeParser` and therefore I'm wondering even more why they were ordered like that before... I'm just not 100% sure if anywhere else something might rely in this order being the over way round of if if worked until now "by chance" because there was no overlap and the result will always be the same.
- I introduced the `Convertible` interface so that not all types have to implement that method. Alternatively an AbstractState could provide a default implemnentation, if you like that better. I whish we had Java 8...
- I decided to put the conversion logic into the types themselves. The conversion logic doesn't seem to be context specific, so it does not need to be within the items. Alternatively we could have something like a `TypeConverter` which takes a `State` and a target type and does the conversion. I thought it's easier to maintain like this, but if you prefer a clearer SoC, I will happily factory it out.

An example to test it with:

sample.things:

```
Bridge hue:bridge:01234567890 "Philips Hue Bridge" [ ipAddress="192.168.0.42", userName="********" ] {
    0210 1 "Hue Bulb" [ lightId="1" ]
}
```

sample.items:

```
Switch hue_switch "Hue Bulb Switch" <light> { channel="hue:0210:01234567890:1:color" }
Dimmer hue_dimmer "Hue Bulb Dimmer" <light> { channel="hue:0210:01234567890:1:color" }
Color hue_color "Hue Bulb Color" <light> { channel="hue:0210:01234567890:1:color" }
```

sample.sitemap:

```
sitemap demo label="Sample" {
    Frame label="Sample" {
        Switch item=hue_switch
        Slider item=hue_dimmer
        Colorpicker item=hue_color
    }

}

```

fixes #2253
Signed-off-by: Simon Kaufmann simon.kfm@googlemail.com

### DIFF
--- a/bundles/core/org.eclipse.smarthome.core.test/src/test/java/org/eclipse/smarthome/core/library/types/DecimalTypeTest.java
+++ b/bundles/core/org.eclipse.smarthome.core.test/src/test/java/org/eclipse/smarthome/core/library/types/DecimalTypeTest.java
@@ -7,11 +7,11 @@
  */
 package org.eclipse.smarthome.core.library.types;
 
-import static org.junit.Assert.assertEquals;
-import static org.junit.Assert.fail;
+import static org.junit.Assert.*;
 
 import java.text.DecimalFormatSymbols;
 
+import org.eclipse.smarthome.core.types.UnDefType;
 import org.junit.Test;
 
 /**
@@ -83,4 +83,26 @@ public class DecimalTypeTest {
         dt = new DecimalType("11.0");
         assertEquals("11" + sep + "0", dt.format("%.1f")); // "11.0"
     }
+
+    @Test
+    public void testConversionToOnOffType() {
+        assertEquals(OnOffType.ON, new DecimalType("100.0").as(OnOffType.class));
+        assertEquals(OnOffType.ON, new DecimalType("1.0").as(OnOffType.class));
+        assertEquals(OnOffType.OFF, new DecimalType("0.0").as(OnOffType.class));
+    }
+
+    @Test
+    public void testConversionToOpenCloseType() {
+        assertEquals(OpenClosedType.OPEN, new DecimalType("1.0").as(OpenClosedType.class));
+        assertEquals(OpenClosedType.CLOSED, new DecimalType("0.0").as(OpenClosedType.class));
+        assertEquals(UnDefType.UNDEF, new DecimalType("0.5").as(OpenClosedType.class));
+    }
+
+    @Test
+    public void testConversionToUpDownType() {
+        assertEquals(UpDownType.UP, new DecimalType("0.0").as(UpDownType.class));
+        assertEquals(UpDownType.DOWN, new DecimalType("1.0").as(UpDownType.class));
+        assertEquals(UnDefType.UNDEF, new DecimalType("0.5").as(OpenClosedType.class));
+    }
+
 }
--- a/bundles/core/org.eclipse.smarthome.core.test/src/test/java/org/eclipse/smarthome/core/library/types/HSBTypeTest.java
+++ b/bundles/core/org.eclipse.smarthome.core.test/src/test/java/org/eclipse/smarthome/core/library/types/HSBTypeTest.java
@@ -89,4 +89,26 @@ public class HSBTypeTest {
         assertEquals(hsb.getSaturation(), hsbRgb.getSaturation());
         assertEquals(hsb.getBrightness(), hsbRgb.getBrightness());
     }
+
+    @Test
+    public void testConversionToOnOffType() {
+        assertEquals(OnOffType.ON, new HSBType("100,100,100").as(OnOffType.class));
+        assertEquals(OnOffType.ON, new HSBType("100,100,1").as(OnOffType.class));
+        assertEquals(OnOffType.OFF, new HSBType("100,100,0").as(OnOffType.class));
+    }
+
+    @Test
+    public void testConversionToDecimalType() {
+        assertEquals(new DecimalType("1.0"), new HSBType("100,100,100").as(DecimalType.class));
+        assertEquals(new DecimalType("0.01"), new HSBType("100,100,1").as(DecimalType.class));
+        assertEquals(DecimalType.ZERO, new HSBType("100,100,0").as(DecimalType.class));
+    }
+
+    @Test
+    public void testConversionToPercentType() {
+        assertEquals(PercentType.HUNDRED, new HSBType("100,100,100").as(PercentType.class));
+        assertEquals(new PercentType("1"), new HSBType("100,100,1").as(PercentType.class));
+        assertEquals(PercentType.ZERO, new HSBType("100,100,0").as(PercentType.class));
+    }
+
 }
--- a/bundles/core/org.eclipse.smarthome.core.test/src/test/java/org/eclipse/smarthome/core/library/types/OnOffTypeTest.java
+++ b/bundles/core/org.eclipse.smarthome.core.test/src/test/java/org/eclipse/smarthome/core/library/types/OnOffTypeTest.java
@@ -1,0 +1,36 @@
+/**
+ * Copyright (c) 2014-2016 by the respective copyright holders.
+ * All rights reserved. This program and the accompanying materials
+ * are made available under the terms of the Eclipse Public License v1.0
+ * which accompanies this distribution, and is available at
+ * http://www.eclipse.org/legal/epl-v10.html
+ */
+package org.eclipse.smarthome.core.library.types;
+
+import static org.junit.Assert.assertEquals;
+
+import org.junit.Test;
+
+/**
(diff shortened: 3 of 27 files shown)


watou:
Does this PR address any of the conversations at the ESH forum on this subject?  The mechanism in openHAB 1 that allows bindings to determine and adapt to the item's accepted data types at state update time is not present in the ESH APIs, and allowing bindings like HTTP, MQTT, TCP, etc., to be implemented against the ESH APIs would be a nice addition.  (Currently, the only way to get this capability is via OH1 bindings and the compat1x layer).

sjka:
> The mechanism in openHAB 1 that allows bindings to determine and adapt to the item's accepted data types at state update time is not present in the ESH APIs

Wouldn't channels be the right layer for bindings to adapt to? I mean, since opposed to OH1 we have the abstraction between the "real world" (Things) and the presentation layer (Items) in ESH, it sounds odd to me if bindings should know anything about items at all. Channels have an accepted item type which bindings can access. I think I would need to further understand the use-cases you are referring to, but it sounds to me like this discussion is a out-of-scope within this PR. 

This PR addresses the presentation layer, mainly allowing "more simplistic" items to be bound to channels which are defined with more "sophisticated" datatypes (e.g. SwitchItem against a PercentType). Maybe this implicitly also solves some use-cases you have in mind...?

watou:
> it sounds odd to me if bindings should know anything about items at all

The issue is usability.  For example, only the user knows that an MQTT topic publishes decimal numbers, and wants a Number item to be updated when some client publishes on that topic.  There is no feature of the ESH APIs that make this simple configuration possible, while the MQTT binding in openHAB 1.x adapts to the accepted data types of the bound item because that model ties bindings directly to items without channels adding another layer to contend with.

> This PR addresses the presentation layer

If your PR is on a completely different subject that's fine with me, but my interest is seeing how the simple use cases of OH1 could be implemented with ESH APIs and to not miss an opportunity to improve the ESH APIs to address those use cases.  Generally, embedding the concept in the architecture instead of at the presentation layer could potentially yield more benefits.

kaikreuzer:
I think the discussion @watou is referring to is https://github.com/eclipse/smarthome/issues/583.
So in general, there might be a need to have the possibility for a binding to find out what types are linked to it and how these links are configured (note that we do not yet have configurations on links). Imho we should NOT allow it to find out about the individual items that are linked to it, though. #583 probably goes beyond the scope of this PR, but it is fair of @watou to mention it here as it might have an impact on the chosen implementation (which I didn't yet review, so I don't have any opinion on this right now).

@sjka Without having had a look at your code yet, only having read your description: It strikes me odd that you also touch ItemUpdater and AutoUpdateBinding - they only deal with Item events, so on this level, everything should be alright already. There is no need to convert a Percent command to an ON state, because if the item is a switch, a Percent command is anyhow invalid and we do not have to care for such cases (actually we should not). Am I missing something?

sjka:
> There is no need to convert a Percent command to an ON state, because if the item is a switch, a Percent command is anyhow invalid and we do not have to care for such cases (actually we should not).

Okay, I just thought it would be cool as we easily can allow such stuff using the same code, but actually wasn't aware that it should fail by intention. I'll remove it then.

kaikreuzer:
> wasn't aware that it should fail by intention

Well, this was my suggestion - if others think it is worthwhile to also accept invalid commands, I am open to discuss.

kaikreuzer:
>  I'll remove it then.

@sjka Just ftr, I am still waiting for this before merging, right?

sjka:
Hmm. I think I was not precise enough with my cites...

I removed the conversions in the AutoUpdateBinding and from the ItemUpdater.

However, IMHO we should keep the conversion from PercentType to OnOffType. I actually didn't invent it, but refactored it from [DimmerItem.java](https://github.com/eclipse/smarthome/pull/2334/files#diff-543ecd419447346e6a177db8638e3ad0L96). So it's not about sending (wrong) commands at all. Think of a dimmable light, where you have linked a switch item against. In order to make that switch getting updated correctly if somebody changes the brightness, we need this conversion.

kaikreuzer:
Ok, sorry, I actually didn't notice that you had removed the changes in ItemUpdater and AutoUpdateBinding already - my comment referred to these, so all fine!

> I had to change the order of the acceptedDataTypes in some items

I think these changes should be ok - afair, the order was mainly relevant for StringType, since parsing as a String always succeeds (so UNDEF etc. has to come first).

> Alternatively an AbstractState could provide a default implemnentation, if you like that better. I whish we had Java 8...

I agree, a default impl would be much nicer - something to add to https://github.com/eclipse/smarthome/wiki/Java-8-CP2-cleanup, I think.
May I suggest a change in the method name? How about calling it simply `as(Type type)` instead of `convertTo`?

> Alternatively we could have something like a TypeConverter which takes a State

No, I am fine with having the logic in the types. But as you mention "TypeConverter": I wonder if we should make the `as()` method available on `Type` and not only on `State` (once we add it as a Java8 default impl. For now, this would mainly impact the JavaDoc of the Convertible interface)? I don't have a use case at hand right now, but it feels ok to also leave the option to have it implemented on Commands.

sjka:
> May I suggest a change in the method name? How about calling it simply as(Type type) instead of convertTo?

Yes, sure. I changed it.

> I wonder if we should make the as() method available on Type and not only on State (once we add it as a Java8 default impl. For now, this would mainly impact the JavaDoc of the Convertible interface)?

I agree, it feels okay. The main reason why I haven't done that so far is the lovely type erasure in Java: The items actually want a State, not a Type, therefore we'd need to cast it at every usage. This would be solvable if we could change the signature to something like this:

```java
public <T extends Type> T as(Class<T> target);
```

However, Java does not accept any sub-type of Type as T for a return value (i.e. OnOffType.ON is marked as an error). We would clutter the code completely with casts at this side then.

Again, with Java 8 we would get the chance to implement some magic in the interface that does the casting only in a single place, but so far I don't see a straight-forward way of doing this while keeping both sides clean to read. Maybe I missed something...?

kaikreuzer:
> The items actually want a State, not a Type, therefore we'd need to cast it at every usage.

Ok, that's a very valid argument.

PR #2337, opened 2016-10-18
Added setPercentageProgress(int progress, boolean silent) to Progress…

chriskn:
…Callback

Added progress to FirmwareUpdateProgressInfo

setPercentageProgress allows users of the ProgressCallback to set the percentage progress
eigther silent (no event posted) or not silent (event is posted if progress changed).
If silent or not, the updated progress will be part of the next FirmwareUpdateProgressInfo Event (for example if the ProressStep changes).
This allows update processes driven by ProgressStep changes or percentage update progress changes.

Signed-off-by: Christoph Knauf  christoph.knauf@itemis.de

### DIFF
--- a/bundles/core/org.eclipse.smarthome.core.thing.test/src/test/groovy/org/eclipse/smarthome/core/thing/firmware/ProgressCallbackTest.groovy
+++ b/bundles/core/org.eclipse.smarthome.core.thing.test/src/test/groovy/org/eclipse/smarthome/core/thing/firmware/ProgressCallbackTest.groovy
@@ -9,24 +9,24 @@ package org.eclipse.smarthome.core.thing.firmware
 import static org.hamcrest.CoreMatchers.*
 import static org.junit.Assert.*
 
-import java.util.Locale;
+import java.util.Locale
 
 import org.eclipse.smarthome.core.events.Event
 import org.eclipse.smarthome.core.events.EventPublisher
 import org.eclipse.smarthome.core.i18n.I18nProvider
-import org.eclipse.smarthome.core.thing.Thing;
+import org.eclipse.smarthome.core.thing.Thing
 import org.eclipse.smarthome.core.thing.ThingTypeUID
 import org.eclipse.smarthome.core.thing.ThingUID
-import org.eclipse.smarthome.core.thing.binding.firmware.Firmware;
+import org.eclipse.smarthome.core.thing.binding.firmware.Firmware
 import org.eclipse.smarthome.core.thing.binding.firmware.FirmwareUID
-import org.eclipse.smarthome.core.thing.binding.firmware.FirmwareUpdateHandler;
-import org.eclipse.smarthome.core.thing.binding.firmware.ProgressCallback;
-import org.eclipse.smarthome.core.thing.binding.firmware.ProgressStep;
+import org.eclipse.smarthome.core.thing.binding.firmware.FirmwareUpdateHandler
+import org.eclipse.smarthome.core.thing.binding.firmware.ProgressCallback
+import org.eclipse.smarthome.core.thing.binding.firmware.ProgressStep
 import org.junit.After
 import org.junit.Before
 import org.junit.Rule
 import org.junit.Test
-import org.osgi.framework.Bundle;
+import org.osgi.framework.Bundle
 
 /**
  * Testing the {@link ProgressCallback}.
@@ -39,6 +39,7 @@ public final class ProgressCallbackTest {
     List<Event> postedEvents
     ThingUID expectedThingUID
     FirmwareUID expectedFirmwareUID
+    String cancelMessageKey = "update-canceled"
     def usedMessagedKey
 
     @Before
@@ -46,7 +47,7 @@ public final class ProgressCallbackTest {
         def thingType = new ThingTypeUID("thing:type")
         expectedThingUID = new ThingUID(thingType, "thingid")
         expectedFirmwareUID = new FirmwareUID(thingType, "1")
-        postedEvents = new LinkedList<>();
+        postedEvents = new LinkedList<>()
         def publisher = [
             post : { event -> postedEvents.add(event) }
         ] as EventPublisher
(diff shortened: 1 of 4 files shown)


bundles/core/org.eclipse.smarthome.core.thing/src/main/java/org/eclipse/smarthome/core/thing/binding/firmware/ProgressCallback.java
```diff
@@ -74,4 +75,14 @@
...
+    
+    /**
+     * Callback operation to indicate that the percentage progress of the firmware update has changed. 
+     * 
+     * @param progress the progress between 0 and 100
+     * @param silent if true EventSubscriber want get informed 
+     * 
+     * @throws IllegalArgumentException if given progress is < 0 or > 100
+     */
+    void setPercentageProgress(int progress, boolean silent);
```
tomhoefer: I would prefer if the silent parameter is not part of the API... there might be several requirements how the percentage update is to be used by the solutions based on ESH... so if we require such a feature then we should introduce a more generic mechanism to prevent sending of progress update events

bundles/core/org.eclipse.smarthome.core.thing/src/main/java/org/eclipse/smarthome/core/thing/binding/firmware/ProgressCallback.java
```diff
@@ -74,4 +75,14 @@
      * Callback operation to indicate that the firmware update was canceled.
      */
     void canceled();
+    
+    /**
+     * Callback operation to indicate that the percentage progress of the firmware update has changed.
```
tomhoefer: As a binding developer it might be not clear enough how to use this interface after this operation has been introduced... So there are several mechanism now to inform about an update progress.... either using next operation or setPercentageProgress... furthermore it is not clear how this operation corresponds to the defineSequence operation

bundles/core/org.eclipse.smarthome.core.thing/src/main/java/org/eclipse/smarthome/core/thing/binding/firmware/ProgressCallback.java
```diff
@@ -74,4 +75,14 @@
...
      */
     void canceled();
+    
+    /**
+     * Callback operation to indicate that the percentage progress of the firmware update has changed. 
+     * 
+     * @param progress the progress between 0 and 100
+     * @param silent if true EventSubscriber want get informed 
+     * 
+     * @throws IllegalArgumentException if given progress is < 0 or > 100
```
tomhoefer: We should think about another name of this operation... the other operations of this interface does not start with set... maybe next(int progress) or update(int progress)?

bundles/core/org.eclipse.smarthome.core.thing/src/main/java/org/eclipse/smarthome/core/thing/firmware/FirmwareUpdateProgressInfo.java
```diff
@@ -29,6 +30,8 @@
 
     private boolean pending;
 
+    private int progress;
```
tomhoefer: We should use Integer object so that this value can be null if a FirmwareUpdateHandler cannot provide detailed information about the progress

bundles/core/org.eclipse.smarthome.core.thing/src/main/java/org/eclipse/smarthome/core/thing/firmware/FirmwareUpdateProgressInfo.java
```diff
@@ -87,14 +93,33 @@ public ProgressStep getProgressStep() {
...
+     * @return true if pending, false otherwise
+     */
+    public boolean isPending() {
+        return pending;
+    }
+
+    /**
+     * Returns the percentage progress of the firmware update.
+     * 
+     * @return the progress between 0 and 100 or null if no progress was set
```
tomhoefer: "null if no progress was set" is not correct according to the impl

bundles/core/org.eclipse.smarthome.core.thing/src/main/java/org/eclipse/smarthome/core/thing/firmware/ProgressCallbackImpl.java
```diff
@@ -48,8 +49,16 @@
     private Collection<ProgressStep> sequence;
     private Iterator<ProgressStep> progressIterator;
     private ProgressStep current;
-    private boolean finished;
-    private boolean pending;
+    private int progress;
+
+    private enum InternalState {
```
tomhoefer: do we really need this enum? it makes the callback impl much more complex...

bundles/core/org.eclipse.smarthome.core.thing/src/main/java/org/eclipse/smarthome/core/thing/binding/firmware/ProgressCallback.java
chriskn: > So there are several mechanism now to inform about an update progress.... either using next operation or setPercentageProgress...

There is no coherence to the defindeSequence operation. Percentage progress is not bound to ProgressSteps. Maybe there is a coherence but this is implementation specific.

bundles/core/org.eclipse.smarthome.core.thing/src/main/java/org/eclipse/smarthome/core/thing/binding/firmware/ProgressCallback.java
chriskn: Whats about updateProgress(int progress) or updatePercentageProgress(int progress)?

bundles/core/org.eclipse.smarthome.core.thing/src/main/java/org/eclipse/smarthome/core/thing/binding/firmware/ProgressCallback.java
chriskn: Do you have any concrete suggestions for such a mechanism?

bundles/core/org.eclipse.smarthome.core.thing/src/main/java/org/eclipse/smarthome/core/thing/firmware/FirmwareUpdateProgressInfo.java
chriskn: Ok, I agree

bundles/core/org.eclipse.smarthome.core.thing/src/main/java/org/eclipse/smarthome/core/thing/firmware/ProgressCallbackImpl.java
chriskn: Otherwise I would need boolean flags for pending and finished. 

I think this enum makes the possible states of the ProgressCallback more visible and better to understand.  Furthermore it ensures that the callback is always is in a defined state and therefore is more failsafe.

bundles/core/org.eclipse.smarthome.core.thing/src/main/java/org/eclipse/smarthome/core/thing/binding/firmware/ProgressCallback.java
tomhoefer: So how can we make the usage of the API foolproof?

bundles/core/org.eclipse.smarthome.core.thing/src/main/java/org/eclipse/smarthome/core/thing/binding/firmware/ProgressCallback.java
tomhoefer: after rethinking about this it makes no sense to have such a parameter... why should you invoke the callback interface with silent=true if you dont want to inform any clients / subscribers

chriskn: In order to have the updated progress as part of the event when you call next() the next time.

bundles/core/org.eclipse.smarthome.core.thing/src/main/java/org/eclipse/smarthome/core/thing/binding/firmware/ProgressCallback.java
chriskn: ok...

bundles/core/org.eclipse.smarthome.core.thing.test/src/test/groovy/org/eclipse/smarthome/core/thing/firmware/ProgressCallbackTest.groovy
```diff
@@ -54,77 +55,112 @@ public final class ProgressCallbackTest {
...
+    void 'assert that update results not in FirmwareUpdateProgressInfo event if progress not changed'(){
+        sut.defineSequence(ProgressStep.UPDATING)
+        sut.update(10)
+        sut.update(10)
+        assertThatProgressInfoEventIsValid(postedEvents.get(0), ProgressStep.UPDATING, false, 10)
+        assertThat postedEvents.size(), is(1)
+    }
+
+    @Test
+    void 'assert that update results in FirmwareUpdateProgressInfo event if progress is decreased'(){
```
tomhoefer: does this make sense? shoulnt we throw a IllegalArgumentException or IllegalStateException if the value is decreased?

tomhoefer:
I miss an unit test for the FirmwareUpdateService that tests the DFU with a handler that has no sequence defined...

bundles/core/org.eclipse.smarthome.core.thing/src/main/java/org/eclipse/smarthome/core/thing/binding/firmware/ProgressCallback.java
```diff
@@ -74,4 +75,15 @@
      * Callback operation to indicate that the firmware update was canceled.
      */
     void canceled();
+
+    /**
+     * Callback operation to update the percentage progress of the firmware update.
+     * This method can be used to provide detailed progress information additional to the sequence or even without a
```
tomhoefer: I would like to have a description how to use the callback interface on the class / interface level... so that binding developers see how to use it (defining a sequence vs. progress based updateds)

bundles/core/org.eclipse.smarthome.core.thing/src/main/java/org/eclipse/smarthome/core/thing/firmware/FirmwareUpdateProgressInfo.java
```diff
@@ -44,20 +47,45 @@ protected FirmwareUpdateProgressInfo() {
...
+     *            (must not be null)
+     * @param pending the flag indicating if the update is pending
      *
      * @throws NullPointerException if firmware UID or current progress step is null
-     * @throws IllegalArgumentException if sequence is null or empty
      */
     FirmwareUpdateProgressInfo(FirmwareUID firmwareUID, ProgressStep progressStep, Collection<ProgressStep> sequence,
             boolean pending) {
         Preconditions.checkNotNull(firmwareUID, "Firmware UID must not be null.");
+        Preconditions.checkState(sequence != null && !sequence.isEmpty(), "Sequence must not be null or empty.");
```
tomhoefer: I would keep the illegal argument exception... pls add it to the Javadoc again

bundles/core/org.eclipse.smarthome.core.thing/src/main/java/org/eclipse/smarthome/core/thing/binding/firmware/ProgressCallback.java
```diff
@@ -74,4 +87,16 @@
...
+     * previous defined sequence.
+     * 
+     * @param progress the progress between 0 and 100
+     * 
+     * @throws IllegalArgumentException if given progress is < 0 or > 100
+     * @throws IllegalArgumentException if given progress is smaller than old progress
```
tomhoefer: pls check also the method signature for the other operations... most of them throw an IllegalStateException (and an IllegalArgumentException) that is not described

tomhoefer:
I still miss an unit test for the FirmwareUpdateService that tests the DFU with a handler that has no sequence defined...

chriskn:
If you have a look in the existing FirmwareUpdateServiceOSGiTest you will notice that this class does not contain any tests regarding concrete update procedures. In my opinion this class is not suitable for such a test. Furthermore there are no changes in this PR regarding the start of a firmware update in the FirmwareUpdateService. Thats why I introduced the test in the ProgressCallbackTest instead.  

**The changes intoduced in this PR have a test coverage of 100%.**

tomhoefer:
@kaikreuzer lgtm, feel free to review

PR #2343, opened 2016-10-20
allow referencing a channel type in the DSL

sjka:
in the Thing DSL it is possible to define channels manually.
However, it was not possible so far to reference a binding's
channel type which is defined in the XMLs.

### DIFF
--- a/bundles/model/org.eclipse.smarthome.model.thing/src/org/eclipse/smarthome/model/thing/internal/GenericThingProvider.xtend
+++ b/bundles/model/org.eclipse.smarthome.model.thing/src/org/eclipse/smarthome/model/thing/internal/GenericThingProvider.xtend
@@ -44,6 +44,7 @@ import org.eclipse.smarthome.core.thing.type.TypeResolver
 import java.util.Locale
 import org.eclipse.smarthome.core.i18n.LocaleProvider
 import org.eclipse.smarthome.core.thing.type.ChannelKind
+import org.eclipse.smarthome.core.thing.type.ChannelTypeUID
 
 /**
  * {@link ThingProvider} implementation which computes *.things files.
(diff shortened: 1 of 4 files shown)


bundles/model/org.eclipse.smarthome.model.thing/src/org/eclipse/smarthome/model/thing/internal/GenericThingProvider.xtend
```diff
@@ -275,11 +276,30 @@ class GenericThingProvider extends AbstractProvider<Thing> implements ThingProvi
...
+                    itemType = it.type
+                    
+                    val kind = if (it.channelKind == null) "State" else it.channelKind                 
+                    parsedKind = ChannelKind.parse(kind)
+                }
+                
+                var channel = ChannelBuilder.create(new ChannelUID(thingUID, id), itemType)
                     .withKind(parsedKind)
                     .withConfiguration(createConfiguration)
+                    .withType(channelTypeUID)
```
maggu2810: I don't know anything about Xtext, but what about channelTypeUID if `it.channelType == null` (the else branch). It seems to be undefined / null. Is the call to `withType` okay for a non-defined type?

sjka: And more important, channels don't need to have a ChannelTypeUID at all:

```java
    /**
     * Returns the channel type UID
     *
     * @return channel type UID or null if no channel type is specified
     */
    public ChannelTypeUID getChannelTypeUID() {
        return channelTypeUID;
    }
```

docs/documentation/features/dsl.md
```diff
@@ -100,3 +100,34 @@ Thing yahooweather:weather:losangeles [ location=2442047, unit="us", refresh=120
...
+    [...]
+</thing:thing-descriptions>
+``` 
+
+They can be referenced within a thing's channel definition, so that they need to be defined only once and can be reused for many channels. You may do so in the DSL as well:
+
+```
+Thing yahooweather:weather:losangeles [ location=2442047, unit="us", refresh=120 ] {
+    Channels:
+        Type temperature : my_yesterday_temperature
```
sjka: Very good point. If you have a UI where you bind items to channels, you might want to see a nice name (aka label) instead of just the technical id. Therefore I think we should.
I would propose to create a separate PR for it, if that's okay?

kaikreuzer:
Yes, absolutely!

kaikreuzer:
Many thanks!

PR #2344, opened 2016-10-20
Supports playing audio events on Safari and IE

aounhaider1:
Signed-off-by: Aoun Bukhari bukhari@itemis.de


extensions/ui/org.eclipse.smarthome.ui.paper/web-src/js/controllers.js
```diff
@@ -35,15 +35,37 @@ angular.module('PaperUI.controllers', [ 'PaperUI.constants' ]).controller('BodyC
...
     });
     eventService.onEvent('smarthome/webaudio/playurl', function(topic, audioUrl) {
-        if (audioUrl) {
-            var audio = new Audio(audioUrl);
-            audio.load();
-            audio.play();
+        if (prevAudioUrl !== audioUrl) {
+            if (context) {
+                var audioBuffer = null;
+                var arr = audioUrl.split('/');
```
kaikreuzer: Please note that with https://github.com/eclipse/smarthome/pull/2374, the audioUrl is now only a relative url starting with a "/", e.g. "/audio/12345". So I assume you can slightly optimize this code here.

aounhaider1:
Updated.

kaikreuzer:
Thanks! Could you give a hint (some background on what this change does) to @ghys whether https://github.com/openhab/org.openhab.ui.habpanel/pull/53/files should be updated in a similar way?

PR #2346, opened 2016-10-20
Prevent scriptarea from overlapping with outputs

aounhaider1:
Fix: Script field overlaps with output field when focussed.
Before:
screen shot 2016-10-20 at 12 08 26
After:
screen shot 2016-10-20 at 12 09 42

Signed-off-by: Aoun Bukhari bukhari@itemis.de


maggu2810:
Is it bout the space between the text field and the output field?

kaikreuzer:
@aounhaider1 Did you see @maggu2810's question?

aounhaider1:
Yes @maggu2810, it is about space between the text field and the output field

PR #2348, opened 2016-10-21
Changes in the FSInternetRadio binding.

m-memova:
Some changes are made in org.eclipse.smarthome.bindig.fsinternetradio.
They are mainly related to cleaning up some unneccessary checks and
code lines that can never be reached.

### DIFF
--- a/extensions/binding/org.eclipse.smarthome.binding.fsinternetradio/src/main/java/org/eclipse/smarthome/binding/fsinternetradio/internal/FSInternetRadioDiscoveryParticipant.java
+++ b/extensions/binding/org.eclipse.smarthome.binding.fsinternetradio/src/main/java/org/eclipse/smarthome/binding/fsinternetradio/internal/FSInternetRadioDiscoveryParticipant.java
@@ -33,6 +33,7 @@ import org.jupnp.model.meta.RemoteDeviceIdentity;
  * add further supported devices!
  *
  * @author Patrick Koenemann - Initial contribution
+ * @author Mihaela Memova - removed the getLabel(RemoteDevice device) method due to its unreachable code lines
  */
 public class FSInternetRadioDiscoveryParticipant implements UpnpDiscoveryParticipant {
 
@@ -100,25 +101,10 @@ public class FSInternetRadioDiscoveryParticipant implements UpnpDiscoveryPartici
                 }
 
                 final DiscoveryResult result = DiscoveryResultBuilder.create(uid).withProperties(properties)
-                        .withLabel(getLabel(device)).build();
+                        .withLabel(device.getDisplayString()).build();
                 return result;
             }
         }
-        return null;
-    }
-
-    private String getLabel(RemoteDevice device) {
-        if (device.getDisplayString() != null)
-            return device.getDisplayString();
-        final DeviceDetails details = device.getDetails();
-        if (details != null) {
-            if (details.getModelDetails() != null && details.getModelDetails().getModelDescription() != null) {
-                return details.getModelDetails().getModelDescription();
-            }
-            if (details.getFriendlyName() != null)
-                return details.getFriendlyName();
-        }
-        // we did not find anything meaningful...
         return null;
     }
 
--- a/extensions/binding/org.eclipse.smarthome.binding.fsinternetradio/src/main/java/org/eclipse/smarthome/binding/fsinternetradio/internal/radio/FrontierSiliconRadio.java
+++ b/extensions/binding/org.eclipse.smarthome.binding.fsinternetradio/src/main/java/org/eclipse/smarthome/binding/fsinternetradio/internal/radio/FrontierSiliconRadio.java
@@ -15,6 +15,7 @@ import java.io.IOException;
  *
  * @author Rainer Ostendorf
  * @author Patrick Koenemann
+ * @author Mihaela Memova - removed duplicated check for the percent value range
  */
 public class FrontierSiliconRadio {
 
@@ -132,8 +133,7 @@ public class FrontierSiliconRadio {
      * @throws IOException if communication with the radio failed, e.g. because the device is not reachable.
      */
     public void setVolumePercent(int volume) throws IOException {
-        final int newVolumePercent = volume < 0 ? 0 : volume > 100 ? 100 : volume;
-        final int newVolumeAbsolute = (newVolumePercent * 32) / 100;
+        final int newVolumeAbsolute = (volume * 32) / 100;
         final String params = "value=" + newVolumeAbsolute;
         conn.doRequest(REQUEST_SET_VOLUME, params);
         currentVolume = volume;
--- a/extensions/binding/org.eclipse.smarthome.binding.fsinternetradio/src/main/java/org/eclipse/smarthome/binding/fsinternetradio/internal/radio/FrontierSiliconRadioConnection.java
+++ b/extensions/binding/org.eclipse.smarthome.binding.fsinternetradio/src/main/java/org/eclipse/smarthome/binding/fsinternetradio/internal/radio/FrontierSiliconRadioConnection.java
@@ -24,6 +24,7 @@ import org.slf4j.LoggerFactory;
  * @author Rainer Ostendorf
  * @author Patrick Koenemann
  * @author Svilen Valkanov - replaced Apache HttpClient with Jetty
+ * @author Mihaela Memova - changed the calling of the stopHttpClient() method, fixed the hardcoded URL path, fixed the for loop condition part
  */
 public class FrontierSiliconRadioConnection {
 
@@ -37,6 +38,9 @@ public class FrontierSiliconRadioConnection {
 
     /** Port number, usually 80. */
     private final int port;
+
+    /** URL path, must begin with a slash (/) */
+    private static final String path = "/fsapi";
 
     /** Access pin, passed upon login as GET parameter. */
     private final String pin;
@@ -56,6 +60,10 @@ public class FrontierSiliconRadioConnection {
         this.pin = pin;
     }
 
+    protected void deactivate() {
+        stopHttpClient(httpClient);
+    }
+
     /**
      * Perform login/establish a new session. Uses the PIN number and when successful saves the assigned sessionID for
      * future requests.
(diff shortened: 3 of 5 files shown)


extensions/binding/org.eclipse.smarthome.binding.fsinternetradio/src/main/java/org/eclipse/smarthome/binding/fsinternetradio/internal/FSInternetRadioHandlerFactory.java
```diff
@@ -35,13 +35,6 @@ public boolean supportsThingType(ThingTypeUID thingTypeUID) {
...
     protected ThingHandler createHandler(Thing thing) {
-
-        ThingTypeUID thingTypeUID = thing.getThingTypeUID();
-
-        if (thingTypeUID.equals(THING_TYPE_RADIO)) {
-            return new FSInternetRadioHandler(thing);
-        }
-
-        return null;
+        return new FSInternetRadioHandler(thing);
```
maggu2810: I agree that the framework should call this function only if the thing type is part of the supported ones.
@kaikreuzer If there is only one supported thing type, I think we could rely on the fact that the framework calls this function only if the thing consists of that respective type.
We also trust that the OSGi framwork calls our functions (activate, bind, etc.) with the correct arguments.

extensions/binding/org.eclipse.smarthome.binding.fsinternetradio/src/main/java/org/eclipse/smarthome/binding/fsinternetradio/internal/radio/FrontierSiliconRadioConnection.java
```diff
@@ -144,7 +144,7 @@ public FrontierSiliconRadioApiResult doRequest(String requestString) throws IOEx
     public FrontierSiliconRadioApiResult doRequest(String requestString, String params) throws IOException {
 
         // 3 retries upon failure
-        for (int i = 0; i < 2; i++) {
+        for (int i = 0; i < 3; i++) {
```
maggu2810: Now it fits to the comment :wink:

extensions/binding/org.eclipse.smarthome.binding.fsinternetradio/src/main/java/org/eclipse/smarthome/binding/fsinternetradio/internal/radio/FrontierSiliconRadio.java
```diff
@@ -132,8 +132,7 @@ public void setVolumeAbsolute(int volume) throws IOException {
      * @throws IOException if communication with the radio failed, e.g. because the device is not reachable.
      */
     public void setVolumePercent(int volume) throws IOException {
-        final int newVolumePercent = volume < 0 ? 0 : volume > 100 ? 100 : volume;
-        final int newVolumeAbsolute = (newVolumePercent * 32) / 100;
+        final int newVolumeAbsolute = (volume * 32) / 100;
```
maggu2810: It is a public function but a bundle internal one.
So I am fine if you know that every caller already ensures (has this been checked?) that the volume is in range [0;100]
I would prefer to add this to the JavaDoc, too (e.g. @param volume the volume that must be in range [0;100]).

extensions/binding/org.eclipse.smarthome.binding.fsinternetradio/src/main/java/org/eclipse/smarthome/binding/fsinternetradio/internal/FSInternetRadioDiscoveryParticipant.java
```diff
@@ -100,28 +100,13 @@ public DiscoveryResult createResult(RemoteDevice device) {
                 }
 
                 final DiscoveryResult result = DiscoveryResultBuilder.create(uid).withProperties(properties)
-                        .withLabel(getLabel(device)).build();
+                        .withLabel(device.getDisplayString()).build();
```
maggu2810: Why do you remove the fallback to `details.getModelDetails().getModelDescription();` and `details.getFriendlyName();`?

extensions/binding/org.eclipse.smarthome.binding.fsinternetradio/src/main/java/org/eclipse/smarthome/binding/fsinternetradio/internal/FSInternetRadioHandlerFactory.java
m-memova: @maggu2810 , you are right. But when I followed the code logic it turned out that  there is a method called `findThingHandlerFactory(ThingTypeUID thingTypeUID)` in the ThingManager class. And if the factory does not support the given ThingTypeUID, the createHandler method is never called. So in my opinion it can never return null, because the given ThingTypeUID is already checked.

extensions/binding/org.eclipse.smarthome.binding.fsinternetradio/src/main/java/org/eclipse/smarthome/binding/fsinternetradio/internal/radio/FrontierSiliconRadio.java
m-memova: I removed this check from here because the PercentType itself does it. In the `handleCommand `method we check if the given command is a PercentType, so we are sure that it is a value in the range [0-100] before getting in the `setVolumePercent `method

extensions/binding/org.eclipse.smarthome.binding.fsinternetradio/src/main/java/org/eclipse/smarthome/binding/fsinternetradio/internal/FSInternetRadioDiscoveryParticipant.java
m-memova: Well, it's been a long code journey here :) I hope I can explain it.
1. `createResult(RemoteDevice device)`  returns something different from null only if `getThingUID(RemoteDevice device) != null`
2.  `getThingUID(RemoteDevice device)`  can return something different from null only if `getDetails!=null`, `getModelDetails!=null` and `getModelNumber!=null`. 
3. `getLabel(RemoteDevice device)`  is calling the getDisplayString() method which can never return null. It returns either empty or non-empty String. It returns an empty String if `getDetails==null` (but here we are already sure that it's not true) or if `getModelDetails==null && getManifacturerDetails==null)`  (but here we are already sure that it's not true, either)
4. So, basically the code in the [113-122] code line range can never be reached.  I tried to implement tests in order to cover it but I could not do it. getLabel(RemoteDevice device) can never return null, so in my opinion we can use just the `getDisplayString()` method.

maggu2810: Ok, thanks for the explanation.

extensions/binding/org.eclipse.smarthome.binding.fsinternetradio/src/main/java/org/eclipse/smarthome/binding/fsinternetradio/internal/FSInternetRadioHandlerFactory.java
m-memova: I removed that check from here because  when the ThingManager class is trying to register a Handler it calls the `findThingHandlerFactory(ThingTypeUID thingTypeUID)` method and the registerHandler method is never called if the factory does not support the given thingTypeUID. I tried to test it and that `return null;` line can never be covered.

maggu2810:
From my side the minor JavaDoc comment is the only one.

Let's wait for @kaikreuzer if he agree with our both oppinion about createThing (that the framework already checks if the type fits).

m-memova:
Thank you, @maggu2810 !

extensions/binding/org.eclipse.smarthome.binding.fsinternetradio/src/main/java/org/eclipse/smarthome/binding/fsinternetradio/internal/FSInternetRadioHandlerFactory.java
sjka: I'm not Kai, but to my knowledge the check should be and is done in the framework consistently. Therefore I don't see an issue removing the check here unless `supportsThingType()` would return more than one entry.

kaikreuzer: Well, the fact that the framework currently only calls it for thing types that were returned on supportedThingTypes might not be a solid fact - at least I didn't find any such assertions in the JavaDocs or in the code of the BaseThingHandlerFactory (where this method is defined and used).
Looking from a pure Java perspective, this is a method of an interface and as the implementor, I should make the implementation robust within the range that the interface and its JavaDoc allows. I therefore would not approve this change. Besides this, it actually also makes it more difficult for future contributors to figure out what needs to be changed if a new supported thing type is added to the binding - with the old code, this is very obvious.

m-memova: Thank you, @kaikreuzer ! What do you think about the other changes? Are you OK with them? Should I revert this particular line the way it was before and update my PR? Or just leave everything in the binding the way it is?

m-memova:
There's another thing I would like to discuss before updating my PR and I think it would be appropriate to do it here. Currently, the HTTP client is started before every single request and it's stopped after it. Starting and stopping the client are asynchronous by their nature. So the following situation is possible :  one thread is trying to stop the client after a request and another thread is trying to start it again for another request. Currently some of my implemented tests are failing once in 30-40-50 times with   `[ESH-thingHandler-1] DEBUG o.eclipse.jetty.client.HttpReceiver - Response failure HttpResponse[null 0 null]@78e6f075 java.nio.channels.AsynchronousCloseException`. In my opinion the reason for that error is exactly the HTTP client's current behaviour. Has anyone ever have a similar problem?

kaikreuzer:
> the HTTP client is started before every single request and it's stopped after it

Good catch. The HTTP requests were last refactored by @svilenvul and the startHttpClient actually checks whether it is already started and does not try to start again, if it is already started.
What I missed when reviewing it is that after a request, it is indeed always stopped again - I would consider this a bug that indeed needs to be fixed.

m-memova:
@kaikreuzer , I think I found a way to fix it. I moved the `stopHtttpClient()` method into the bundles's `deactivate()` method and everything seems to be working fine on my machine.

Issue #2345, opened 2016-10-20
Test failure(s): NtpOSGiTest


maggu2810:
The build on the HIPP failed:
https://hudson.eclipse.org/smarthome/job/SmartHomeDistribution-Nightly/1387/

```text
-------------------------------------------------------
 T E S T S
-------------------------------------------------------
Running org.eclipse.smarthome.binding.ntp.test.NtpOSGiTest
Tests run: 16, Failures: 1, Errors: 0, Skipped: 2, Time elapsed: 122.594 sec <<< FAILURE! - in org.eclipse.smarthome.binding.ntp.test.NtpOSGiTest
the dateTime channel is updated with the right time zone(org.eclipse.smarthome.binding.ntp.test.NtpOSGiTest) skipped
time is refreshed when string channel is linked(org.eclipse.smarthome.binding.ntp.test.NtpOSGiTest)  Time elapsed: 10.622 sec
if no time zone is set in the configuration, the dateTime channel is updated with the default one(org.eclipse.smarthome.binding.ntp.test.NtpOSGiTest)  Time elapsed: 10.115 sec
time is refreshed when dateTime channel is linked(org.eclipse.smarthome.binding.ntp.test.NtpOSGiTest)  Time elapsed: 10.124 sec
the status of a thing with string channel is updated with communication error, when unknown host is set in the configuraion(org.eclipse.smarthome.binding.ntp.test.NtpOSGiTest)  Time elapsed: 0.121 sec
the string channel is updated with the right formatting(org.eclipse.smarthome.binding.ntp.test.NtpOSGiTest)  Time elapsed: 10.247 sec  <<< FAILURE!
java.lang.AssertionError: The item was not of type StringType
Expected: is an instance of org.eclipse.smarthome.core.library.types.StringType
     but: <NULL> is a org.eclipse.smarthome.core.types.UnDefType
	at org.hamcrest.MatcherAssert.assertThat(MatcherAssert.java:20)
	at org.junit.Assert.assertThat(Assert.java:865)
	at org.junit.Assert$assertThat$0.callStatic(Unknown Source)
	at org.eclipse.smarthome.binding.ntp.test.NtpOSGiTest$_getItemState_closure4.doCall(NtpOSGiTest.groovy:421)
	at org.eclipse.smarthome.binding.ntp.test.NtpOSGiTest$_getItemState_closure4.doCall(NtpOSGiTest.groovy)
[...]

PR #2354, opened 2016-10-23
NTP test fix

maggu2810:
The commits could be left separated.

Fixes: #2345


kaikreuzer:
Hm, Travis failed, but not due to this PR. So let's be optimistic and merge nonetheless...

PR #2359, opened 2016-10-25
First draft of working authentication api for ESH.

splatch:
This commit brings also basic implementation authentication provider based on JAAS realms which allows to plug in any standard java module to work. There are some limitations - first of all authentication is done via REST service layer with SPI provided there. No session tracking is yet available and all works just like basic authentication. Login page is not implemented and login resource for now is just an stub.

There are few weak spots in current implementation:
- At this moment just one authentication provider may be deployed
- Authentication and authorization is controlled by jaxrs service layer (jersey/eclipsesource customization)
-  `GroupPrincipals` and other nested principals set on `Subject` are ignored - just principal names are used.
- There is no form login.
- Authorization for now is enforced just on UUIDResource.
- There is no notion of resource ownership - it's plain role based authorization.

Making long story short - this code works and it is possible to plug in `sql`, `ldap` or basic `property` file as authentication source.


kaikreuzer:
Hi @splatch, I did some first tests today.
In general, the mechanism seems to be working, but I wonder what would be the next steps to make it really usable. The browser must be in some way prompted that authentication is required, either like in OH1 with a simple basic auth prompt or better with a redirect to a login screen. How will the browser know that it has to keep sending an auth header?

splatch:
I get your point @kaikreuzer, user interfeace is in most of cases set of static resources which do not require authorization. We could plug in security at ui path, so browser will prompt for password while accessing user interface.
Browsers automatically propagate basic authentication to sub-urls, however smart home rest endpoints are served at different root ( `/rest`) that user interface (`/ui`) which would require some client side coding to forward this details from one place into another.
Earlier on I prepared basic login resource which could be poked from javascript but in such case we will need some kind of session identifier returned by backend which will get attached to every next request. This is safer option because it does not require credentials all the time.

In both cases some user interface changes will be necessary because with these changes rest endpoints are fully aware of authentication.

kaikreuzer:
Ok, so if I get you right, the cleanest way would be to put these features in the UIs themselves, so that they always have an auth header on every request they do.

Is there no way to do (at least a simple initial) solution on the server side? It would be good for OH2 not having to touch all different UIs, but to have the browser deal with it...
If we require basic-auth on / itself, there wouldn't be any problem with the different paths to the uis and rest or am I missing something? Isn't there some way to configure Jetty/Pax-Web to require authentication on / ?

splatch:
Earlier this year I spent fair amount of time trying to do this and it's not an easy task. Wiring security at this level is hard due to following facts:
A) 👎  Jetty project reformed security handling and it requires quite long sequence of XML calls to configure it fully without webapp.
B) 👎  It's not possible to configure connector level security directly from jetty or pax web level.
C) 👎  Jetty security is WAR-a-like centric while smarthome modules are JARs and deploy single servlets.
D) 👎  It is not possible to centrally configure a basic security at REST level because http context creation is taken over by eclipse source jaxrs integration.
E) 👎  Pax Web uses its own handlers for dispatching requests thus configuration in XML is not applied to things registered in runtime.

It still might be doable with few hours (or days) long debugger session to find matching configuration or workaround. This may work for currently used versions (jetty+pax web) but may get broken with any version upgrade. Security handled in the way OH1 did it is not possible due to points C and D. Even if it would be possible I strongly advise to plug in security at application level so user interfaces will be aware of it as well. Anything else than that will be rather an work around than solution.

kaikreuzer:
> Even if it would be possible I strongly advise to plug in security at application level so user interfaces will be aware of it as well. Anything else than that will be rather an work around than solution.

Ok, thanks a lot for your details insights, let's go the way you propose! I'll come back with some review comments shortly.

bundles/core/org.eclipse.smarthome.core.id/src/main/java/org/eclipse/smarthome/core/id/internal/UUIDResource.java
```diff
@@ -28,6 +29,7 @@
  */
 @Path(UUIDResource.PATH_UUID)
 @Api(value = UUIDResource.PATH_UUID)
+@RolesAllowed({ "administrator" })
```
kaikreuzer: Shall we define a default set of roles as constants in org.eclipse.smarthome.core.auth.Roles that should be used throughout the ESH code?

bundles/core/org.eclipse.smarthome.core/src/main/java/org/eclipse/smarthome/core/auth/Authentication.java
```diff
@@ -0,0 +1,21 @@
...
+ * which accompanies this distribution, and is available at
+ * http://www.eclipse.org/legal/epl-v10.html
+ */
+package org.eclipse.smarthome.core.auth;
+
+/**
+ * Definition of authentication given to username after verification of credentials by authentication provider.
+ *
+ * Each authentication must at least point to some identity (username) and roles.
+ */
```
kaikreuzer: can you add an author tag?

bundles/core/org.eclipse.smarthome.core/src/main/java/org/eclipse/smarthome/core/auth/Authentication.java
```diff
@@ -0,0 +1,21 @@
...
+public interface Authentication {
+
+    String getUsername();
```
kaikreuzer: please add some javadoc

bundles/core/org.eclipse.smarthome.core/src/main/java/org/eclipse/smarthome/core/auth/Authentication.java
```diff
@@ -0,0 +1,21 @@
...
+
+    String[] getRoles();
```
kaikreuzer: We hardly ever work with arrays, rather use `Set<String>` instead.

kaikreuzer: Please add some javadoc

bundles/core/org.eclipse.smarthome.core/src/main/java/org/eclipse/smarthome/core/auth/AuthenticationException.java
```diff
@@ -0,0 +1,27 @@
...
+ * All rights reserved. This program and the accompanying materials
+ * are made available under the terms of the Eclipse Public License v1.0
+ * which accompanies this distribution, and is available at
+ * http://www.eclipse.org/legal/epl-v10.html
+ */
+package org.eclipse.smarthome.core.auth;
+
+/**
+ * Base type for exceptions thrown by authentication layer.
+ */
```
kaikreuzer: please add author tag

bundles/core/org.eclipse.smarthome.core/src/main/java/org/eclipse/smarthome/core/auth/AuthenticationException.java
```diff
@@ -0,0 +1,27 @@
...
+public class AuthenticationException extends RuntimeException {
+
+    public AuthenticationException(String message) {
```
kaikreuzer: please add javadoc to all three methods

bundles/core/org.eclipse.smarthome.core/src/main/java/org/eclipse/smarthome/core/auth/AuthenticationProvider.java
```diff
@@ -0,0 +1,25 @@
...
+ * are made available under the terms of the Eclipse Public License v1.0
+ * which accompanies this distribution, and is available at
+ * http://www.eclipse.org/legal/epl-v10.html
+ */
+package org.eclipse.smarthome.core.auth;
+
+/**
+ * Realizations of this type are responsible for checking validity of various credentials and giving back authentication
+ * which defines access scope for authenticated user or system.
+ */
```
kaikreuzer: please add author tag

bundles/core/org.eclipse.smarthome.core/src/main/java/org/eclipse/smarthome/core/auth/Credentials.java
```diff
@@ -0,0 +1,15 @@
...
+ * All rights reserved. This program and the accompanying materials
+ * are made available under the terms of the Eclipse Public License v1.0
+ * which accompanies this distribution, and is available at
+ * http://www.eclipse.org/legal/epl-v10.html
+ */
+package org.eclipse.smarthome.core.auth;
+
+/**
+ * Super type for credentials which can be handled by authentication providers.
+ */
```
kaikreuzer: please add author tag

bundles/core/org.eclipse.smarthome.core/src/main/java/org/eclipse/smarthome/core/auth/dto/AuthenticationDTO.java
```diff
@@ -0,0 +1,21 @@
+/**
+ * Copyright (c) 2014-2016 by the respective copyright holders.
+ * All rights reserved. This program and the accompanying materials
+ * are made available under the terms of the Eclipse Public License v1.0
+ * which accompanies this distribution, and is available at
+ * http://www.eclipse.org/legal/epl-v10.html
+ */
+package org.eclipse.smarthome.core.auth.dto;
```
kaikreuzer: Do we really need DTOs at all? They are normally only for sending such objects through the REST API or as events, but I doubt we want to do that with any of the auth stuff. So unless required, I would opt for removing them.

bundles/core/org.eclipse.smarthome.core/src/main/java/org/eclipse/smarthome/core/library/auth/DefaultAuthentication.java
```diff
@@ -0,0 +1,35 @@
+/**
+ * Copyright (c) 2014-2016 by the respective copyright holders.
+ * All rights reserved. This program and the accompanying materials
+ * are made available under the terms of the Eclipse Public License v1.0
+ * which accompanies this distribution, and is available at
+ * http://www.eclipse.org/legal/epl-v10.html
+ */
+package org.eclipse.smarthome.core.library.auth;
```
kaikreuzer: I'd suggest to move this class to org.eclipse.smarthome.core.auth as well and remove the library package.

bundles/core/org.eclipse.smarthome.core/src/main/java/org/eclipse/smarthome/core/library/auth/DefaultAuthentication.java
```diff
@@ -0,0 +1,35 @@
...
+
+import org.eclipse.smarthome.core.auth.Authentication;
+
+/**
+ * Basic implementation of authentication type.
+ */
```
kaikreuzer: please add author tag

bundles/core/org.eclipse.smarthome.core/src/main/java/org/eclipse/smarthome/core/library/auth/DefaultAuthentication.java
```diff
@@ -0,0 +1,35 @@
...
+public class DefaultAuthentication implements Authentication {
+
+    private final String username;
+    private final String[] roles;
+
+    public DefaultAuthentication(String username, String... roles) {
```
kaikreuzer: please add javadoc on all public methods

bundles/core/org.eclipse.smarthome.core/src/main/java/org/eclipse/smarthome/core/library/auth/UsernamePasswordCredentials.java
```diff
@@ -0,0 +1,34 @@
...
+ * which accompanies this distribution, and is available at
+ * http://www.eclipse.org/legal/epl-v10.html
+ */
+package org.eclipse.smarthome.core.library.auth;
+
+import org.eclipse.smarthome.core.auth.Credentials;
+
+/**
+ * Credentials which represents user name and password.
+ */
```
kaikreuzer: please add author tag

bundles/core/org.eclipse.smarthome.core/src/main/java/org/eclipse/smarthome/core/library/auth/UsernamePasswordCredentials.java
```diff
@@ -0,0 +1,34 @@
+/**
+ * Copyright (c) 2014-2016 by the respective copyright holders.
+ * All rights reserved. This program and the accompanying materials
+ * are made available under the terms of the Eclipse Public License v1.0
+ * which accompanies this distribution, and is available at
+ * http://www.eclipse.org/legal/epl-v10.html
+ */
+package org.eclipse.smarthome.core.library.auth;
```
kaikreuzer: same here - move to core.auth

bundles/core/org.eclipse.smarthome.core/src/main/java/org/eclipse/smarthome/core/library/auth/UsernamePasswordCredentials.java
```diff
@@ -0,0 +1,34 @@
...
+
+/**
+ * Credentials which represents user name and password.
+ */
+public class UsernamePasswordCredentials implements Credentials {
+
+    private final String username;
+    private final String password;
+
+    public UsernamePasswordCredentials(String username, String password) {
```
kaikreuzer: please add javadoc on all public methods

bundles/io/org.eclipse.smarthome.io.rest.core/META-INF/MANIFEST.MF
```diff
@@ -5,18 +5,23 @@ Bundle-SymbolicName: org.eclipse.smarthome.io.rest.core
 Bundle-Version: 0.9.0.qualifier
 Bundle-Vendor: Eclipse.org
 Bundle-RequiredExecutionEnvironment: JavaSE-1.7
-Import-Package: com.google.common.base,
+Import-Package: com.eclipsesource.jaxrs.provider.security,
```
kaikreuzer: I'd suggest to not make everbody dependent on the jax-rs connector security bundle.
We should move the `SmartHomePrincipal` and the `AbstractSecurityHandler` to a new bundle o.e.s.io.rest.auth and the `BasicSecurityHandler` implementation to a new bundle o.e.s.io.rest.auth.basic.

bundles/io/org.eclipse.smarthome.io.rest.core/OSGI-INF/loginresource.xml
```diff
@@ -0,0 +1,18 @@
...
+<!--
+
+    Copyright (c) 2014-2016 by the respective copyright holders.
+    All rights reserved. This program and the accompanying materials
+    are made available under the terms of the Eclipse Public License v1.0
+    which accompanies this distribution, and is available at
+    http://www.eclipse.org/legal/epl-v10.html
+
+-->
+<scr:component xmlns:scr="http://www.osgi.org/xmlns/scr/v1.1.0" name="org.eclipse.smarthome.io.rest.core.auth.LoginResource">
```
kaikreuzer: According to our discussion, auth must be built into the client apps, so I guess we should remove the LoginResource from this PR as it does not really provide any value.

bundles/io/org.eclipse.smarthome.io.rest.core/src/main/java/org/eclipse/smarthome/io/rest/core/security/AbstractSecurityHandler.java
```diff
@@ -0,0 +1,66 @@
...
+import org.eclipse.smarthome.core.auth.Authentication;
+import org.eclipse.smarthome.core.auth.AuthenticationProvider;
+import org.eclipse.smarthome.core.auth.Credentials;
+
+import com.eclipsesource.jaxrs.provider.security.AuthenticationHandler;
+import com.eclipsesource.jaxrs.provider.security.AuthorizationHandler;
+
+/**
+ * Base type for handling authentication and authorization delegated from rest/jaxrs service layer.
+ */
```
kaikreuzer: please add author tag

bundles/io/org.eclipse.smarthome.io.rest.core/src/main/java/org/eclipse/smarthome/io/rest/core/security/AbstractSecurityHandler.java
```diff
@@ -0,0 +1,66 @@
...
+
+        Authentication authentication = authenticationProvider.authenticate(credentials);
+        if (authentication != null) {
+            return new SmartHomePrincipal(authentication);
+        }
+
+        return null;
+    }
+
+    protected abstract Credentials createCredentials(ContainerRequestContext requestContext);
```
kaikreuzer: please add javadoc

bundles/io/org.eclipse.smarthome.io.rest.core/src/main/java/org/eclipse/smarthome/io/rest/core/security/SmartHomePrincipal.java
```diff
@@ -0,0 +1,37 @@
...
+import java.security.Principal;
+import java.util.Arrays;
+import java.util.HashSet;
+import java.util.Set;
+
+import org.eclipse.smarthome.core.auth.Authentication;
+
+/**
+ * Principal implementation which is set for all authenticated requests.
+ */
```
kaikreuzer: please add author tag

splatch:
Hey @kaikreuzer ,
Thank you very much for review, will fix issues as soon as I can. I will move jaas to separate module so it will be possible to replace it.

Best regards,
Lukasz

kaikreuzer:
Hey @splatch, any progress on addressing the review comments?

kaikreuzer:
I have done the suggested changes and created a new PR with it: https://github.com/eclipse/smarthome/pull/2587

satish-nikam:
loginresource.xml is still there in the OSGI-INF folder, but LoginResource.java file is removed.

kaikreuzer:
@satish-nikam Thanks for noticing! Removed by #2754

PR #2360, opened 2016-10-26
Allow injection of timestamp into AbstractDiscoveryService

afuechsel:
When starting several discovery services for the same thing types at once, each service is started some milliseconds after the other and storing its own timestamp. When calling removeOlderResults it could happen, that discovery results discovered the service that has been started first will be removed by one of the other services started later. Therefore DiscoveryService#startScan now allows to specify the starting timestamp giving the DiscoveryServiceRegistry a chance, to inject the same timestamp into all started discovery services. 

### DIFF
--- a/bundles/config/org.eclipse.smarthome.config.discovery.test/src/test/groovy/org/eclipse/smarthome/config/discovery/inbox/events/InboxEventFactoryTest.groovy
+++ b/bundles/config/org.eclipse.smarthome.config.discovery.test/src/test/groovy/org/eclipse/smarthome/config/discovery/inbox/events/InboxEventFactoryTest.groovy
@@ -29,7 +29,7 @@ class InboxEventFactoryTest {
 
     def THING_UID = new ThingUID("binding:type:id")
 
-    def DISCOVERY_RESULT = new DiscoveryResultImpl(THING_UID, null, null, null, null, 60)
+    def DISCOVERY_RESULT = new DiscoveryResultImpl(THING_UID, null, null, null, null, 60, null)
 
     def INBOX_ADDED_EVENT_TYPE = InboxAddedEvent.TYPE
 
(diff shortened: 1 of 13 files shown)


bundles/config/org.eclipse.smarthome.config.discovery/src/main/java/org/eclipse/smarthome/config/discovery/DiscoveryService.java
```diff
@@ -80,6 +80,25 @@
...
     /**
+     * Triggers this service to start an active scan for new devices.<br>
+     * This method must not block any calls such as {@link #abortScan()} and
+     * must return fast.
+     * <p>
+     * If started, any registered {@link DiscoveryListener} must be notified about {@link DiscoveryResult}s.
+     * <p>
+     * If there is already a scan running, it is aborted and a new scan is triggered.
+     * <p>
+     * In contrast to {@link #startScan(ScanListener)} this method can be given a starting time stamp
```
kaikreuzer: This method does not look meaningful on the public facing API (besides, it might even break compatibility to existing services that might not extend the abstract implementation).
It very much looks like a workaround to some problem and it does not even ask the implementation to mandatorily make use of this timestamp in case removeOlderResults() is called.

Isn't there any way to deal with the situation in the DiscoveryServiceRegistry itself? And is it at all legal that a DiscoveryService removes ALL results, even it they were not created by itself...?

afuechsel: I agree with you that it's not a good idea, to have it on the public interface. But I need a way to call the startScan method from the DiscoveryServiceRegistry with the given timestamp. I see no other way, how the registry could handle this itself. Can we assume, that each DiscoveryService is at least an AbstractDiscoveryService? Then this method doesn't need to be on the interface, only on the abstract class and the registry could use it. But I think, we cannot, as the user might be free to implement the interface completely by itself. 

The other question, whether it's legal or not, that a discovery service removes results that might not have been created by itself, is a good one. However, this would require to mark each result from which service it has been created (or updated lately). Even when implementing it this way, how could we be sure, that no one else has overridden the removeOlderResult method in another DiscoveryService to handle things differently? We should have made this method final...

afuechsel:
PR changed according the discussion using my latest suggestion: 

* Extend the DiscoveryResult with a field to store the store a reference to the source DiscoveryService which is set to null by default. 
* set this field in AbstractDiscoverService
* changed PersistentInbox to remove only results from the calling discovery service

afuechsel:
sorry, this PR is crap. Will close it, rethink, and later create a new one.

PR #2365, opened 2016-10-27
Extending hue binding to use N-UPNP discovery for searching hue bridges

engawel:
Signed-off-by: Awelkiyar Wehabrebi awehabrebi@itemis.de


extensions/binding/org.eclipse.smarthome.binding.hue/src/main/java/org/eclipse/smarthome/binding/hue/internal/discovery/HueBridgeNupnpDiscovery.java
```diff
@@ -0,0 +1,141 @@
...
+    private void discoverHueBridges() {
+        // get bridge list
+        List<BridgeJsonParameters> bridgeList = getBridgeList();
+        if (bridgeList != null) {
+            for (BridgeJsonParameters bridge : bridgeList) {
+                if (bridge.getId() != null && bridge.getInternalipaddress() != null) {
+                    // check if bridge is reachable, and if 'id' contains 'fffe' at the middle
+                    if (isHueBridge(bridge.getInternalipaddress()) && bridge.getId().substring(6, 10).equals("fffe")) {
+                        // 'fffe' should be removed from 'id', to match with 'id' in description.xml file
+                        ThingUID uid = new ThingUID(THING_TYPE_BRIDGE,
```
afuechsel: Why not using bridge.getId().replace("fffe", "") like below? Get it once, store it in a serial number and reuse it below.

extensions/binding/org.eclipse.smarthome.binding.hue/src/main/java/org/eclipse/smarthome/binding/hue/internal/discovery/HueBridgeNupnpDiscovery.java
```diff
@@ -0,0 +1,141 @@
...
+     * Discover available Hue Bridges and then add them in the discovery inbox
+     */
+    private void discoverHueBridges() {
+        // get bridge list
+        List<BridgeJsonParameters> bridgeList = getBridgeList();
+        if (bridgeList != null) {
+            for (BridgeJsonParameters bridge : bridgeList) {
+                if (bridge.getId() != null && bridge.getInternalipaddress() != null) {
+                    // check if bridge is reachable, and if 'id' contains 'fffe' at the middle
+                    if (isHueBridge(bridge.getInternalipaddress()) && bridge.getId().substring(6, 10).equals("fffe")) {
```
afuechsel: All hue bridges have "fffe" in between? Old and new ones? Please use a static constant for "fffe".

extensions/binding/org.eclipse.smarthome.binding.hue/src/main/java/org/eclipse/smarthome/binding/hue/internal/discovery/HueBridgeNupnpDiscovery.java
sjka: > Why not using bridge.getId().replace("fffe", "") 

Disagree - "fffe" could be part of the legitimate mac address in another place as well.

> like below?

For this very reason, the below may be painfully wrong.

afuechsel: Yes, you are right. So the "below" need to be changed. But as we are using the same construct at two places, please make a small method for it. 

Can we be sure, that there is (for whatever reason) always 4 letters that need to be cut out in the reply?

sjka: > Can we be sure, that there is (for whatever reason) always 4 letters that need to be cut out in the reply?

I asked a contact at Philips for a statement regarding this, but didn't get any reply so far

extensions/binding/org.eclipse.smarthome.binding.hue/src/main/java/org/eclipse/smarthome/binding/hue/HueBindingConstants.java
```diff
@@ -51,4 +56,7 @@
     // Light config properties
     public static final String LIGHT_ID = "lightId";
 
+    // Added for the N-UPNP discovery
+    public static final Set<ThingTypeUID> SUPPORTED_THING_TYPES_UID = Collections.singleton(THING_TYPE_BRIDGE);
```
kaikreuzer: This constant should rather be defined within `HueBridgeNupnpDiscovery`. Note that we e.g. also have `HueLightHandler.SUPPORTED_THING_TYPES`.

extensions/binding/org.eclipse.smarthome.binding.hue/src/main/java/org/eclipse/smarthome/binding/hue/internal/discovery/BridgeJsonParameters.java
```diff
@@ -0,0 +1,49 @@
...
+ * http://www.eclipse.org/legal/epl-v10.html
+ */
+package org.eclipse.smarthome.binding.hue.internal.discovery;
+
+/**
+ * The {@link BridgeJsonParameters} class defines JSON object, which
+ * contains bridge attributes like IP address. It is used for bridge
+ * N-UPNP Discovery.
+ *
+ * @author Awelkiyar Wehabrebi
```
kaikreuzer: add: "Initial contribution and API"

extensions/binding/org.eclipse.smarthome.binding.hue/src/main/java/org/eclipse/smarthome/binding/hue/internal/discovery/BridgeJsonParameters.java
```diff
@@ -0,0 +1,49 @@
...
+ *
+ */
+public class BridgeJsonParameters {
+
+    // hue bridge parameters
+    private final String id;
+    private final String internalipaddress;
```
kaikreuzer: I'd prefer this to be camel case.

extensions/binding/org.eclipse.smarthome.binding.hue/src/main/java/org/eclipse/smarthome/binding/hue/internal/discovery/BridgeJsonParameters.java
```diff
@@ -0,0 +1,49 @@
...
+    private final String macaddress;
```
kaikreuzer: I'd prefer this to be camel case.

extensions/binding/org.eclipse.smarthome.binding.hue/src/main/java/org/eclipse/smarthome/binding/hue/internal/discovery/HueBridgeNupnpDiscovery.java
```diff
@@ -0,0 +1,141 @@
...
+ * hue bridges. It uses the 'NUPnP service provided by Philips'.
+ *
+ * @author Awelkiyar Wehabrebi - Initial contribution
+ *
+ */
+public class HueBridgeNupnpDiscovery extends AbstractDiscoveryService {
+
+    private final Logger logger = LoggerFactory.getLogger(HueBridgeNupnpDiscovery.class);
+
+    private static final String url = "https://www.meethue.com/api/nupnp";
```
kaikreuzer: constants must be upper case

extensions/binding/org.eclipse.smarthome.binding.hue/src/main/java/org/eclipse/smarthome/binding/hue/internal/discovery/HueBridgeNupnpDiscovery.java
```diff
@@ -0,0 +1,141 @@
...
+                    if (isHueBridge(bridge.getInternalipaddress()) && bridge.getId().substring(6, 10).equals("fffe")) {
+                        // 'fffe' should be removed from 'id', to match with 'id' in description.xml file
+                        ThingUID uid = new ThingUID(THING_TYPE_BRIDGE,
+                                bridge.getId().substring(0, 6) + bridge.getId().substring(10));
+                        if (uid != null) {
+                            Map<String, Object> properties = new HashMap<>(2);
+                            properties.put(HOST, bridge.getInternalipaddress());
+                            properties.put(SERIAL_NUMBER, bridge.getId().replace("fffe", ""));
+                            DiscoveryResult result = DiscoveryResultBuilder.create(uid).withProperties(properties)
+                                    .withLabel("Philips hue (" + bridge.getInternalipaddress() + ")")
```
kaikreuzer: The HueBridgeDiscoveryParticipant uses `withLabel(device.getDetails().getFriendlyName())` - will this lead to the very same label here?

extensions/binding/org.eclipse.smarthome.binding.hue/src/main/java/org/eclipse/smarthome/binding/hue/internal/discovery/HueBridgeNupnpDiscovery.java
```diff
@@ -0,0 +1,141 @@
...
+     */
+    private boolean isHueBridge(String host) {
+        try {
+            String url = "http://" + host + "/description.xml";
+            String description = HttpUtil.executeUrl("GET", url, timeout);
+            return description.contains("<modelName>Philips hue");
+        } catch (IOException e) {
+            logger.debug("Failure accessing description.xml file of Hue Bridge with IP address - {}.", host);
+        }
+        return false;
```
kaikreuzer: This might not be true. What if the bridge is simply offline?

extensions/binding/org.eclipse.smarthome.binding.hue/src/main/java/org/eclipse/smarthome/binding/hue/internal/discovery/HueBridgeNupnpDiscovery.java
```diff
@@ -0,0 +1,141 @@
...
+     * @return a list of available Hue Bridges
+     */
+    private List<BridgeJsonParameters> getBridgeList() {
+        try {
+            Gson gson = new Gson();
+            String json = HttpUtil.executeUrl("GET", url, timeout);
+            return gson.fromJson(json, new TypeToken<List<BridgeJsonParameters>>() {
+            }.getType());
+        } catch (IOException e) {
+            logger.error("Error accessing the Philips Hue NUPnP service.", e);
```
kaikreuzer: Which is perfectly valid, if there is no Internet access available. Hence this must not result in any errors being logged.

extensions/binding/org.eclipse.smarthome.binding.hue/src/main/java/org/eclipse/smarthome/binding/hue/internal/discovery/BridgeJsonParameters.java
engawel: same here, it should match to the predefined member name. Otherwise, didn't work.

extensions/binding/org.eclipse.smarthome.binding.hue/src/main/java/org/eclipse/smarthome/binding/hue/internal/discovery/BridgeJsonParameters.java
afuechsel: Makes sense. You could use 

````java 
@SerializedName("internalipaddress") private final internalIpAddress
```` 
since the json returned does not using camel case. Similar on all other places.

File: extensions/binding/org.eclipse.smarthome.binding.hue/src/main/java/org/eclipse/smarthome/binding/hue/internal/discovery/HueBridgeNupnpDiscovery.java (hunk 10 above)
Comment (engawel) [inline extensions/binding/org.eclipse.smarthome.binding.hue/src/main/java/org/eclipse/smarthome/binding/hue/internal/discovery/HueBridgeNupnpDiscovery.java]: Do you mean that log should be completely removed?

File: extensions/binding/org.eclipse.smarthome.binding.hue/src/main/java/org/eclipse/smarthome/binding/hue/internal/discovery/HueBridgeNupnpDiscovery.java (hunk 8 above)
Comment (engawel) [inline extensions/binding/org.eclipse.smarthome.binding.hue/src/main/java/org/eclipse/smarthome/binding/hue/internal/discovery/HueBridgeNupnpDiscovery.java]: Yes, the label will be same. It is possible also to distinguish the label by just adding something like, NUPnP - 
withLabel("Philips hue (" + bridge.getInternalipaddress() + ") - NUPnP")

Comment (kaikreuzer) [inline extensions/binding/org.eclipse.smarthome.binding.hue/src/main/java/org/eclipse/smarthome/binding/hue/internal/discovery/HueBridgeNupnpDiscovery.java]: My point is that we must make sure that it is exactly the same label - since the Inbox will "flicker" otherwise in case the one or the other discovery service puts a result in the Inbox for the same UID.

File: extensions/binding/org.eclipse.smarthome.binding.hue/src/main/java/org/eclipse/smarthome/binding/hue/internal/discovery/HueBridgeNupnpDiscovery.java (hunk 9 above)
Comment (kaikreuzer) [inline extensions/binding/org.eclipse.smarthome.binding.hue/src/main/java/org/eclipse/smarthome/binding/hue/internal/discovery/HueBridgeNupnpDiscovery.java]: Implementation yes, but not documentation. It should then rather say "To check if Bridge is reachable" only - and the method name should imho be adapted as well.

File: extensions/binding/org.eclipse.smarthome.binding.hue/src/main/java/org/eclipse/smarthome/binding/hue/internal/discovery/HueBridgeNupnpDiscovery.java (hunk 10 above)
Comment (kaikreuzer) [inline extensions/binding/org.eclipse.smarthome.binding.hue/src/main/java/org/eclipse/smarthome/binding/hue/internal/discovery/HueBridgeNupnpDiscovery.java]: If you want to keep it for debugging, changing the log level to debug is fine for me as well.

File: extensions/binding/org.eclipse.smarthome.binding.hue/src/main/java/org/eclipse/smarthome/binding/hue/internal/discovery/HueBridgeNupnpDiscovery.java (hunk 9 above)
Comment (engawel) [inline extensions/binding/org.eclipse.smarthome.binding.hue/src/main/java/org/eclipse/smarthome/binding/hue/internal/discovery/HueBridgeNupnpDiscovery.java]: Main purpose of the method is to check if Bridge is hue or not, to do that it should be first reachable (reachability check too).  May be I change log message to - logger.debug("Unable to distinguish Bridge type cause Bridge with IP address - {} is not reachable.", host) ?

File: extensions/binding/org.eclipse.smarthome.binding.hue/src/main/java/org/eclipse/smarthome/binding/hue/internal/discovery/HueBridgeNupnpDiscovery.java (hunk 2 above)
Comment (engawel) [inline extensions/binding/org.eclipse.smarthome.binding.hue/src/main/java/org/eclipse/smarthome/binding/hue/internal/discovery/HueBridgeNupnpDiscovery.java]: @afuechsel  I can't find a similar implementation of test, could you please give me some hints for mocking the service.

File: extensions/binding/org.eclipse.smarthome.binding.hue/src/main/java/org/eclipse/smarthome/binding/hue/internal/discovery/HueBridgeNupnpDiscovery.java (hunk 1 above)
Comment (chriskn) [inline extensions/binding/org.eclipse.smarthome.binding.hue/src/main/java/org/eclipse/smarthome/binding/hue/internal/discovery/HueBridgeNupnpDiscovery.java]: Any news here?

**Comment by chriskn** [comment]
@sjka Any news regarding the Bridge Ids?

**Comment by sjka** [comment]
no, not yet - only pretty consistent silence on all channels :-/

**Comment by kaikreuzer** [comment]
So shall we go with the implementation as is for the moment then?
@chriskn The third commit on this PR is a merge commit - could you try to clean this up?

File: extensions/binding/org.eclipse.smarthome.binding.hue.test/src/test/groovy/org/eclipse/smarthome/binding/hue/test/HueBridgeNupnpDiscoveryOSGITest.groovy [hunk 11]
```diff
@@ -0,0 +1,280 @@
...
+            checkDiscoveryResult(results.get(BRIDGE_THING_UID_1), ip1, sn1)
+            assertThat results.get(BRIDGE_THING_UID_2), is(notNullValue())
+            checkDiscoveryResult(results.get(BRIDGE_THING_UID_2), ip2, sn2)
+
+            assertThat inbox.get(new InboxFilterCriteria(BRIDGE_THING_TYPE_UID,null)).size(), is(2)
+        }
+    }
+
+    @Test
+    public void 'assert that invalid birdges are not discovered'(){
```
afuechsel: Typo: birdges instead of bridges

chriskn:
@kaikreuzer Yes, I will clean it up together with @engawel as soon as there are no further comments

chriskn:
> So shall we go with the implementation as is for the moment then?

@kaikreuzer: Yes, imo we should go with this implementation. We can change it if we get an answer

Issue #2221, opened 2016-09-27
ThingManager should notify handlers about commands/updates fully asynchronously


kaikreuzer:
Currently, the ThingManager notifies handlers sequentially using the safemethodcaller. If a call blocks, it times out after 5 seconds and the next handler is informed only then.

As there is no order specified and all calls to the handlers are independent from each other, they should be called fully asynchronously, so that no handler can delay the further processing.

sjka:
let's wait for #2087 to be merged before addressing this.

Issue #2082, opened 2016-08-30
LIFX binding feature request: implement postCommand and UDP broadcast


cbcoursera1:
LIFX bulbs currently change (relatively) slowly and serially when making changes to a group. My understanding is that this can be sped up or made simultaneous in two ways:

**Implementing postCommand**
This would allow multiple commands to be sent in parallel via the event bus. sendCommand, currently used, only allows for a single thread in the event bus.

**Creating a UDP broadcast Thing**
This would allow for simultaneous changing of all LIFX bulbs on the network. I've seen this used in other implementations of the LIFX LAN Protocol.

I'm relatively new to this and don't have the knowledge to add these to the binding myself, but if anyone is able I know that it would be appreciated by LIFX users.

kgoderis:
Pls note that sendCommand and postCommand are only relevant for events from the bulb towards the ESH runtime, not from the runtime to the bulb

On the UDP broadcast it should be noted that even as the command is send to a Group, the ThingHandler get notified individually, eg it does not differentiate between commands send to a group or item directly. So, it would be difficult to group bulbs at the UDP level within the binding, it would also bring complexity as communications have to be shared between ThingHandlers which is something we do not want. 

I have to revisit the code to see if we can render the communications bit

kgoderis:
... Concurrent. 

To be investigated but I understand your concern

Karel

cbcoursera1:
I'm not following your statement about commands from the bulb to the ESH runtime, maybe I haven't explained myself well.

I'm using OpenHAB and trying to write a rule that will simultaneously change several bulbs, and I'm using sendCommand() in my rule script to make the changes to the bulbs. So my thinking is that this is a command from the ESH runtime to the bulb, but maybe I'm not completely understanding.

My thought on the UDP broadcast Thing was this:

In other LIFX LAN Protocol implementations, I understand that simultaneous bulb changing is sometimes done via UDP broadcast. It's simultaneous because you're utilizing the network infrastructure, and it's fast because it's asynchronous. I was thinking that a Thing could be identified on the network at the UDP broadcast IP, and sending commands to that Thing would propagate them across all LIFX bulbs. But, I do understand that this goes counter to the definition of a Thing as being physical.

kgoderis:
> I'm not following your statement about commands from the bulb to the ESH runtime, maybe I haven't explained myself well.
> 
[...]

PR #2366, opened 2016-10-27
Fix #2321: LIFX resend on packet loss, configurable power on brightness and code refactoring

### LINKED ISSUE #2321: LIFX commands sometimes not executed due to UDP packet loss
In the current LIFX implementation commands may not be succesfully executed due to UDP packet loss. As a result bulbs do not turn on/off or change to the correct color. In my setup it occurs quite often since I have a lot of bulbs that are motion activated. Sometimes I end up in a completely unilluminated room which is quite annoying.

I think the binding should:
- keep track of the most recently sent state change that was made with a `SetLightPowerRequest` or `SetColorRequest`
- resend the most recent state change when no `StatePowerResponse` / `StateResponse` for a state change request is received within a certain interval

Futhermore:
- the number of such retries should be limited
- the retries should disregard the online/offline state of bulbs
- it should still be possible to change the state of bulbs using other applications such as the LIFX app

Do you have any thoughts on this @kgoderis? I might also put in my own PR for this.

wborn:
This PR addresses the issue/functionality discussed in #2321 by:
- resending packets that change light power/color when they are not acknowledged after 250ms (max 3 retries)
- adding a channel configuration option to customize the power on brightness level

The binding code has been refactored by decomposing the `LifxLightHandler` into several objects:
- communication related logic has been extracted into the `LifxLightCommunicationHandler`
- a `LifxLightState` object has been introduced for keeping track of the current and pending light state
- the reoccuring task to update the current light state has been extracted into `LifxLightCurrentStateUpdater`
- the reoccuring task to determine the online/offline state has been extracted into `LifxLightOnlineStateUpdater`
- the logic for sending/resending packets to change the light state has been extracted into `LifxLightStateChanger`

The refactoring also includes:
- addition of the `LifxResponsePacketListener` interface, which allows implementors to listen to response packets received by the `LifxLightCommunicationHandler`
- addition of the `LifxLightStateListener` interface, which allows implementors to listen to changes of the current and pending light state
- addition of a thread pool

Other notable improvements include:
- fix of another deadlock scenario by always following the same locking strategy, this is the reason why the locks required for sending a packet have all been moved to `LifxLightCommunicationHandler.sendPacket(Packet packet, InetSocketAddress address, SelectionKey selectedKey)`
- packets are now send/resend using the thread pool, this makes it already possible to concurrently send commands to lights without the implemenation of #2221 (i.e. #2082 can also probably be closed)
- warnings are no longer logged in case of every packet loss (only when the 3 retries fail)
- lights are also powered on when changes are made on the temperature channel
- when there is a pending light state change, updating channels is only done when packets are received that match the pending light state
- due to the introduction of the pending light state, sending multiple commands in quick succession (HSB followed by temperature change) will now correctly update the light state

@kgoderis do you want to review my changes? @kaikreuzer could perhaps have a look at the new `powerOnBrightness` channel configuration? I've also made some remarks in a [#2321 comment](https://github.com/eclipse/smarthome/issues/2321#issuecomment-256198907) about this new option.


kgoderis:
@wborn Wouter, this looks very GTM. That a lot! I tested the code on 3 bulbs only and it works pretty smooth from a functional point of view. ESH and the LIFX app are nicely in sync, and indeed the switch-on delay between bulbs is gone.

I have not yet looked at the code in detail, the only thing that is maybe of concern is that it requires Java 1.8 whereas we come from 1.7. AFAIK, ESH should stick to 1.7. Your only dependency on 1.8 lies in java.util.function.IntUnaryOperator, maybe you have an alternative for that class?

wborn:
@kgoderis thanks for already having a quick look! I'll find a fix for that. :-)

kaikreuzer:
Err, this isn't just some refactoring, this rather looks like a fully new binding :-)
The `powerOnBrightness` looks good to me, thanks.
As it is more than 1000 lines, I actually have to create a CQ for it and have it approved by the IP team.
@wborn let me know once you are ok to freeze this PR, since after creating the CQ there must not be any changes here anymore.

wborn:
Yes I've completed working on this PR @kaikreuzer so you can freeze it. The first comment and further commits should explain what has changed and why. Extracting existing logic away from the `LifxLightHandler` will certainly have added up a lot to those line counts. The `LifxLightCommunicationHandler` logic already makes up for 530 lines alone.

kgoderis:
@kaikreuzer The real changes are indeed rather limited, and it LGTM. Hopefully we can get this PR approved rapidly.

kaikreuzer:
Ok, thanks - I have created CQ 12202 and waiting for its approval now.

kaikreuzer:
Good news, we have check-in approval for this PR!

PR #2367, opened 2016-10-28
fixed UpnpDiscoveryService to call thingDiscovered...

afuechsel:
... for each RemoteDevice already available in the upnp registry


sjka:
makes sense imho

kaikreuzer:
Hm, indeed, this looks like a nasty bug that could explain many upnp discovery issues... Thanks for finding and fixing it!

PR #2369, opened 2016-10-28
Changed PersistentInbox to remove only results created by the same discovery service

afuechsel:
Also extended PersistentInbox#removeOlderResults to remove only the results from the source DiscoveryService

Signed-off-by: Andre Fuechsel andre.fuechsel@telekom.de

### DIFF
--- a/bundles/config/org.eclipse.smarthome.config.discovery/src/main/java/org/eclipse/smarthome/config/discovery/internal/PersistentInbox.java
+++ b/bundles/config/org.eclipse.smarthome.config.discovery/src/main/java/org/eclipse/smarthome/config/discovery/internal/PersistentInbox.java
@@ -307,7 +307,7 @@ public final class PersistentInbox implements Inbox, DiscoveryListener, ThingReg
         for (DiscoveryResult discoveryResult : getAll()) {
             String discovererName = resultDiscovererMap.get(discoveryResult); 
             if (thingTypeUIDs.contains(discoveryResult.getThingTypeUID()) && discoveryResult.getTimestamp() < timestamp
-                    && sourceName.equals(discovererName)) {
+                    && (discovererName == null || sourceName.equals(discovererName))) {
                 ThingUID thingUID = discoveryResult.getThingUID();
                 removedThings.add(thingUID);
                 remove(thingUID);
--- a/bundles/config/org.eclipse.smarthome.config.discovery/src/main/java/org/eclipse/smarthome/config/discovery/internal/PersistentInbox.java
+++ b/bundles/config/org.eclipse.smarthome.config.discovery/src/main/java/org/eclipse/smarthome/config/discovery/internal/PersistentInbox.java
@@ -126,6 +126,8 @@ public final class PersistentInbox implements Inbox, DiscoveryListener, ThingReg
 
     private Storage<DiscoveryResult> discoveryResultStorage;
 
+    private Map<DiscoveryResult, String> resultDiscovererMap = new HashMap<>();
+
     private ScheduledFuture<?> timeToLiveChecker;
 
     private EventPublisher eventPublisher;
@@ -269,6 +271,7 @@ public final class PersistentInbox implements Inbox, DiscoveryListener, ThingReg
                 if (!isInRegistry(thingUID)) {
                     removeResultsForBridge(thingUID);
                 }
+                resultDiscovererMap.remove(discoveryResult);
                 this.discoveryResultStorage.remove(thingUID.toString());
                 notifyListeners(discoveryResult, EventType.removed);
                 return true;
@@ -286,7 +289,9 @@ public final class PersistentInbox implements Inbox, DiscoveryListener, ThingReg
 
     @Override
     public void thingDiscovered(DiscoveryService source, DiscoveryResult result) {
-        add(result);
+        if (add(result)) {
+            resultDiscovererMap.put(result, source.getClass().getName());
+        }
     }
 
     @Override
@@ -298,9 +303,11 @@ public final class PersistentInbox implements Inbox, DiscoveryListener, ThingReg
     public Collection<ThingUID> removeOlderResults(DiscoveryService source, long timestamp,
             Collection<ThingTypeUID> thingTypeUIDs) {
         HashSet<ThingUID> removedThings = new HashSet<>();
+        String sourceName = source.getClass().getName(); 
         for (DiscoveryResult discoveryResult : getAll()) {
-            if (thingTypeUIDs.contains(discoveryResult.getThingTypeUID())
-                    && discoveryResult.getTimestamp() < timestamp) {
+            String discovererName = resultDiscovererMap.get(discoveryResult); 
+            if (thingTypeUIDs.contains(discoveryResult.getThingTypeUID()) && discoveryResult.getTimestamp() < timestamp
+                    && sourceName.equals(discovererName)) {
                 ThingUID thingUID = discoveryResult.getThingUID();
                 removedThings.add(thingUID);
                 remove(thingUID);
--- a/bundles/config/org.eclipse.smarthome.config.discovery/src/main/java/org/eclipse/smarthome/config/discovery/internal/PersistentInbox.java
+++ b/bundles/config/org.eclipse.smarthome.config.discovery/src/main/java/org/eclipse/smarthome/config/discovery/internal/PersistentInbox.java
@@ -18,6 +18,7 @@ import java.util.List;
 import java.util.Map;
 import java.util.Objects;
 import java.util.Set;
+import java.util.concurrent.ConcurrentHashMap;
 import java.util.concurrent.CopyOnWriteArrayList;
 import java.util.concurrent.CopyOnWriteArraySet;
 import java.util.concurrent.ScheduledFuture;
@@ -126,7 +127,7 @@ public final class PersistentInbox implements Inbox, DiscoveryListener, ThingReg
 
     private Storage<DiscoveryResult> discoveryResultStorage;
 
-    private Map<DiscoveryResult, String> resultDiscovererMap = new HashMap<>();
+    private Map<DiscoveryResult, Class<?>> resultDiscovererMap = new ConcurrentHashMap<>();
 
     private ScheduledFuture<?> timeToLiveChecker;
 
@@ -290,7 +291,7 @@ public final class PersistentInbox implements Inbox, DiscoveryListener, ThingReg
     @Override
     public void thingDiscovered(DiscoveryService source, DiscoveryResult result) {
         if (add(result)) {
-            resultDiscovererMap.put(result, source.getClass().getName());
+            resultDiscovererMap.put(result, source.getClass());
         }
     }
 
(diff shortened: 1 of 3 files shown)


bundles/config/org.eclipse.smarthome.config.discovery/src/main/java/org/eclipse/smarthome/config/discovery/DiscoveryResultBuilder.java
```diff
@@ -128,13 +129,24 @@ public DiscoveryResultBuilder withTTL(long ttl) {
     }
 
     /**
+     * Sets the class name of the {@link DiscoveryService} that has discovered this {@link DiscoveryResult}.
+     *
+     * @param service class name of the {@link DiscoveryService}
+     * @return the updated builder
+     */
+    public DiscoveryResultBuilder withDiscoveryService(String service) {
```
sjka: I really don't like this idea. Have you checked how many bindings are using the DiscoveryResultBuilder directly...? Can't this be handles in the `AbstractDiscoveryService.thingDiscovered(...)` transparently, without the need to adapt bazillions of bindings?

bundles/config/org.eclipse.smarthome.config.discovery/src/main/java/org/eclipse/smarthome/config/discovery/internal/UpnpDiscoveryService.java
```diff
@@ -67,7 +67,10 @@ protected void addUpnpDiscoveryParticipant(UpnpDiscoveryParticipant participant)
         if (upnpService != null) {
             Collection<RemoteDevice> devices = upnpService.getRegistry().getRemoteDevices();
             for (RemoteDevice device : devices) {
-                participant.createResult(device);
+                DiscoveryResult result = participant.createResult(device);
```
sjka: hello, #2367

bundles/config/org.eclipse.smarthome.config.discovery/src/main/java/org/eclipse/smarthome/config/discovery/DiscoveryResultBuilder.java
afuechsel: Oh you don't need to adapt gazillions of bindings. It is perfectly valid to just not set the discoveredBy... And: it is already handled in the ABstractDiscoveryService, no worries. I can even remove it from the builder completely, not needed.

afuechsel: Removed

bundles/config/org.eclipse.smarthome.config.discovery/src/main/java/org/eclipse/smarthome/config/discovery/internal/UpnpDiscoveryService.java
afuechsel: removed

bundles/config/org.eclipse.smarthome.config.discovery/src/main/java/org/eclipse/smarthome/config/discovery/internal/DiscoveryResultImpl.java
```diff
@@ -226,4 +239,14 @@ public long getTimestamp() {
...
         return timeToLive;
     }
+
+    @Override
+    public void discoveredBy(String discoveryService) {
+        this.discoveredBy = discoveryService;
+    }
+
+    @Override
+    public String isDiscoveredBy() {
```
kaikreuzer: I think the proper English would be "hasBeenDiscoveredBy". But to make it explicit which of these two methods does what: Wouldn't it be simpler to treat it like a normal bean property and call the methods `setDiscoverer` and `getDiscoverer`?

bundles/config/org.eclipse.smarthome.config.discovery/src/main/java/org/eclipse/smarthome/config/discovery/internal/DiscoveryResultImpl.java
```diff
@@ -49,6 +51,7 @@
      * @param label the human readable label to set (could be null or empty)
      * @param bridgeUID the unique bridge ID to be set
      * @param timeToLive time to live in seconds
+     * @param discoveredBy class name of the {@link DiscoveryService}, that created this result
```
kaikreuzer: I assume you have to make the parameter a string (and not keep it a Class for having strong types) due to the JSON (de)serialization, right? If so, you might want to add "fqcn", so that it is clear that it is not the simple class name that is excepted here.

bundles/config/org.eclipse.smarthome.config.discovery/src/main/java/org/eclipse/smarthome/config/discovery/AbstractDiscoveryService.java
```diff
@@ -248,6 +248,7 @@ protected synchronized void stopScan() {
      *            Holds the information needed to identify the discovered device.
      */
     protected void thingDiscovered(DiscoveryResult discoveryResult) {
+        discoveryResult.discoveredBy(this.getClass().getName());
```
kaikreuzer: Is this really the right place to set this property? After all, the DiscoveryResult interface defines it and thus potential implementors probably feel obliged to fill it (before passing it here). I think at this point, the passed instance should be treated as an immutable (otherwise you shouldn't have added the property to the interface at all, but only to our internal implementation that the builder creates).
There should at least be some safe guard to check whether it is already set (and then do nothing) and if it is not set, try to set it, but catch exceptions.

bundles/config/org.eclipse.smarthome.config.discovery/src/main/java/org/eclipse/smarthome/config/discovery/DiscoveryResultBuilder.java
```diff
@@ -134,7 +134,7 @@ public DiscoveryResultBuilder withTTL(long ttl) {
      */
     public DiscoveryResult build() {
         return new DiscoveryResultImpl(thingTypeUID, thingUID, bridgeUID, properties, representationProperty, label,
-                ttl);
+                ttl, null);
```
kaikreuzer: Is null really the right thing to do?
I would have expected that the builder provides an immutable with this field already set - see my comment in AbstractDiscoveryService.

bundles/config/org.eclipse.smarthome.config.discovery/src/main/java/org/eclipse/smarthome/config/discovery/internal/DiscoveryResultImpl.java
```diff
@@ -226,4 +239,14 @@ public long getTimestamp() {
     public long getTimeToLive() {
         return timeToLive;
     }
+
+    @Override
+    public void discoveredBy(String discoveryService) {
```
kaikreuzer: `discoveryService` is probably a bad name here - maybe call it discoverer, fqcn or classname instead?

bundles/config/org.eclipse.smarthome.config.discovery/src/main/java/org/eclipse/smarthome/config/discovery/DiscoveryResultBuilder.java
kaikreuzer: I agree with @sjka that we must not require users of the Builder to add a new method call everywhere, but maybe it can be derived automatically (or stay null, if we say that it is not a mandatory property)?

afuechsel:
@kaikreuzer is completely right! The source discovery source is passed to each listener, I just didn't realize it, as the PersistentInbox have chosen to ignore it: 

```` java 
    @Override
    public void thingDiscovered(DiscoveryService source, DiscoveryResult result) {
        add(result);
    }
````  

I will completely rework (again) the PR. This makes totally sense!

**Comment by afuechsel** [comment]
Completely reworked according to suggestions of @kaikreuzer. Much less code has been changed now.

File: bundles/config/org.eclipse.smarthome.config.discovery/src/main/java/org/eclipse/smarthome/config/discovery/internal/PersistentInbox.java [hunk 8]
```diff
@@ -298,9 +303,11 @@ public void thingRemoved(DiscoveryService source, ThingUID thingUID) {
     public Collection<ThingUID> removeOlderResults(DiscoveryService source, long timestamp,
             Collection<ThingTypeUID> thingTypeUIDs) {
         HashSet<ThingUID> removedThings = new HashSet<>();
+        String sourceName = source.getClass().getName(); 
         for (DiscoveryResult discoveryResult : getAll()) {
-            if (thingTypeUIDs.contains(discoveryResult.getThingTypeUID())
-                    && discoveryResult.getTimestamp() < timestamp) {
+            String discovererName = resultDiscovererMap.get(discoveryResult);
```
kaikreuzer: We are within the "PersistentInbox" here. The discovererName will be null if the discoveryResult has been loaded from the storage and thus such discoveryResults will never ever be removed again. I would assume that the sourceName will need to be persisted as well in order to correctly implement this.

afuechsel:
Updated PR according the latest discussion with @kaikreuzer

kaikreuzer:
FTR: This discussion meant that we accept the fact that persisted entries are NOT checked against the source, because we do not have this information at hand. Adding it to the storage as well would have meant quite some efforts incl. a required migration for backward compatibility and there wouldn't really have been a practical benefit (as the main reason for this PR is [this problem](https://github.com/eclipse/smarthome/pull/2360#issue-185318827). Persisted results can in any case be regarded as potentially outdated, so I think this is a fair compromise.

bundles/config/org.eclipse.smarthome.config.discovery/src/main/java/org/eclipse/smarthome/config/discovery/internal/PersistentInbox.java
```diff
@@ -126,6 +126,8 @@ private boolean isResultExpired(DiscoveryResult result, long now) {
 
     private Storage<DiscoveryResult> discoveryResultStorage;
 
+    private Map<DiscoveryResult, String> resultDiscovererMap = new HashMap<>();
```
maggu2810: I don't care much, but wouldn't it be simpler to put the Class itself `Class<?>` to the list instead of using the FQN of the class?

kaikreuzer: Would make the code simpler indeed.
Btw, as thingDiscovered() is not synchronized - shouldn't a ConcurrentHashMap be used here?

afuechsel: updated the PR

kaikreuzer:
Thanks!

janssk1:
This change breaks the removeOlderResults functionality in case the bundle gets updated (which happens a lot during development)
Since the new version class != old version class, results that are discovered with the previous version are not removed...

afuechsel:
Looks like you are right. The problem is that there is no id or somethign similar at the DiscoveryService interface to identify the discovery service. Introducing one would break the current interface. Please file a bug, needs to be reworked somehow.

kaikreuzer:
Would comparing the class name instead of the class itself be an option?

afuechsel:
@kaikreuzer didn't we do that in the first place? Cannot remember exactly but there was some issue with comparing the classname only. For instance, if you have the UpnpDiscoveryService, used for different thing types, you don't want to remove the results from the other binding.

afuechsel:
Checked the conversation and found a comment from @maggu2810 on Nov 3 recommending to use the class itself instead of the FQN. But since this makes some troubles as @janssk1 pointed out correctly I would suggest to return to the FQN.

maggu2810:
Can you point me to the old discussion?

afuechsel:
@maggu2810 Just see above. https://github.com/eclipse/smarthome/pull/2369#pullrequestreview-7000001 and open the outdated

maggu2810:
Okay, get the point now. Thanks...

IMHO there is no solution that could suffer all needs.
Using the class, the FQCN or the simple class name.
If I rename my discovery service class this is broken again...

Perhaps every discovery service should contain an ID that is prefixed by `bindingId:`.
The ID could be the same regardless of the class name, package names, etc. and is able to survive also a big refactoring of the bindings itself.

kaikreuzer:
I don't think we need to cover the case where a class is renamed. Note that @janssk1's concern was the issue during development time, where you regularly re-deploy new instances. I guess if a service is renamed, it is ok that you have to delete the results manually.

Issue #2376, opened 2016-10-30
NTP Tests bug, when time is changed


petarnv:
In the NTP Binding Tests the "Europe/Helsinki" time zone is used. But recently there has been a time change in this time zone and that broke two of the tests:

1) 'if no time zone is set in the configuration, the string channel is updated with the default one'

The message is:
The string channel was not updated with the right timezone
Expected: is "EEST"
     but: was "EET"

Aparently the abbreviation EEST(Eastern European Summer Time) is used only in summer and now that the time is changed the abbreviation, that is returned is EET(Eastern European Time) - https://www.timeanddate.com/time/zones/eest

2) 'if no time zone is set in the configuration, the dateTime channel is updated with the default one'

The message is:
The dateTime channel was not updated with the right timezone
Expected: is "+0300"
     but: was "+0200"

Now that the time is changed, the current offset is +0200, not +0300

I have made a PR with a fix for this: https://github.com/eclipse/smarthome/pull/2377

J-N-K:
There is an additional problem (at least on my system) which is probably related and NOT fixed. Build on my Win 10 machine fails because OEZ is returned instead of EET. OEZ is the correct name in germany, so there might be a localization issue involved.

petarnv:
@J-N-K, thank you for finding this problem. It is indeed localization issue. I have made a PR with a fix: https://github.com/eclipse/smarthome/pull/2407. Can you run it on your machine, please?

J-N-K:
Works for me. Thanks.

maggu2810:
Also with a merged #2407 I run into this error:

```text
-------------------------------------------------------
 T E S T S
-------------------------------------------------------
Running org.eclipse.smarthome.binding.ntp.test.NtpOSGiTest
Tests run: 16, Failures: 1, Errors: 0, Skipped: 2, Time elapsed: 21.705 sec <<< FAILURE! - in org.eclipse.smarthome.binding.ntp.test.NtpOSGiTest
the dateTime channel is updated with the right time zone(org.eclipse.smarthome.binding.ntp.test.NtpOSGiTest) skipped
time is refreshed when string channel is linked(org.eclipse.smarthome.binding.ntp.test.NtpOSGiTest)  Time elapsed: 0.22 sec
if no time zone is set in the configuration, the dateTime channel is updated with the default one(org.eclipse.smarthome.binding.ntp.test.NtpOSGiTest)  Time elapsed: 0.138 sec
[...]

PR #2377, opened 2016-10-30
NTP Binding Tests Fix

petarnv:
Fixed the bug in the NTP Binding, related to the time change: https://github.com/eclipse/smarthome/issues/2376


kaikreuzer:
Travis failure is not due to this PR.

kaikreuzer:
Thanks for the fix!

PR #2380, opened 2016-10-31
added a warning in case the sitemap model name does not match the filename

kaikreuzer:


sjka:
+1 as I happened to debug some of this "unexpected behavior" last weekend until I realized my fatal mistake. You were just quicker with this PR ;-)

PR #2387, opened 2016-11-01
Channel translations

### LINKED ISSUE #2356: i18n: label of channel is not resolved
In my binding I have this channel definition:

```
<channels>
    <channel id="press" typeId="system.trigger">
        <label>@text/dashButtonPressChannelLabel</label>
        <description>@text/dashButtonPressChannelLabel</description>
    </channel>
</channels>
```

On the Paper UI the label is not resolved (_@text/dashButtonPressChannelLabel_ is displayed) although the entry is contained in the i18n properties file.

sjka:
added t9n support for channel definitions

The collection even was named `localizedChannelDefinitions`, but actually did not. It contained the plain strings as they came from the XML.

fixes #2356


bundles/core/org.eclipse.smarthome.core.thing.xml.test/src/test/resources/test-bundle-pool/ChannelTypesI18nTest.bundle/ESH-INF/thing/thing-types.xml
```diff
@@ -0,0 +1,36 @@
...
+	<thing-type id="something">
+
+		<label>Something</label>
+		
+		<channels>
+            <channel id="channelPlain" typeId="string">
+                <label>Channel Plain Label</label>
+                <description>Channel Plain Description</description>
+            </channel>
+			<channel id="channelInplace" typeId="channel-with-i18n">
```
kaikreuzer: here seems to be some mix of tabs/spaces

sjka:
will do so - the mix indeed was not intended!

just for the record, the oomph profile is set to tabs for xml:
image
so I reckon I'd rather remove the spaces, right? 
(just bringing this up NOT to raise a discussion about spaces vs. tabs, but just because these settings surprised me, and obviously you too...)

kaikreuzer:
Yeah, indeed, tabs are right here (ftr: this decision was taken since XMLs are being packages in the bundles and using tabs instead of spaces might save a few bytes here and there).

sjka:
> tabs instead of spaces might save a few bytes here and there

oic, thanks for the explanation. I somehow must have missed that and couldn't find it anywhere.

PR has been fixed accordingly.

PR #2391, opened 2016-11-02
Fixed removal of job from the runnable list and added a (failing) test case

kaikreuzer:

### DIFF
--- a/bundles/core/org.eclipse.smarthome.core.test/src/test/java/org/eclipse/smarthome/core/scheduler/ExpressionThreadPoolExecutorTest.java
+++ b/bundles/core/org.eclipse.smarthome.core.test/src/test/java/org/eclipse/smarthome/core/scheduler/ExpressionThreadPoolExecutorTest.java
@@ -1,0 +1,50 @@
+package org.eclipse.smarthome.core.scheduler;
+
+import static org.junit.Assert.*;
+
+import java.text.ParseException;
+
+import org.eclipse.smarthome.core.scheduler.ExpressionThreadPoolManager.ExpressionThreadPoolExecutor;
+import org.junit.Ignore;
+import org.junit.Test;
+
+public class ExpressionThreadPoolExecutorTest {
+
+    private ExpressionThreadPoolExecutor scheduler = ExpressionThreadPoolManager.getExpressionScheduledPool("test");
+    protected boolean success;
+
+    @Test
+    @Ignore
+    public void testScheduleAndRemove() throws ParseException, InterruptedException {
+
+        Runnable runnable = new Runnable() {
+            @Override
+            public void run() {
+                success = true;
+            }
+        };
+
+        success = false;
+        CronExpression expression = new CronExpression("0/1 * * * * ?");
+        scheduler.schedule(runnable, expression);
+        Thread.sleep(1500);
+        assertTrue(success);
+
+        // FIXME: We have to use the method with the expression as a parameter. If we do
+        // scheduler.remove(runnable) instead, the task is never removed (as it is kept in the "scheduled" set).
+        boolean removed = scheduler.remove(expression);
+
+        // FIXME: If TRACE is enabled and thus the execution of the task takes more than 2 seconds, removed is false
+        assertTrue(removed);
+
+        success = false;
+        Thread.sleep(1500);
+        assertFalse(success);
+
+        assertEquals(0, scheduler.getQueue().size());
+
+        // FIXME: What is this task that remains in the queue here? Shouldn't the queue be empty?
+        // assertEquals(0, scheduler.getQueue().size());
+    }
+
+}
--- a/bundles/core/org.eclipse.smarthome.core/src/main/java/org/eclipse/smarthome/core/scheduler/ExpressionThreadPoolManager.java
+++ b/bundles/core/org.eclipse.smarthome.core/src/main/java/org/eclipse/smarthome/core/scheduler/ExpressionThreadPoolManager.java
@@ -225,11 +225,12 @@ public class ExpressionThreadPoolManager extends ThreadPoolManager {
 
         @Override
         public boolean remove(Runnable task) {
-            if (futures.get(task) != null && futures.get(task).cancel(false)) {
-                running.remove(task);
+            Future<?> future = futures.get(task);
+            if (future != null && future.cancel(false)) {
+                running.remove(future);
             }
             futures.remove(task);
-            return super.remove(task);
+            return super.remove((Runnable) future);
         }
 
         public boolean remove(Expression expression) {


bundles/core/org.eclipse.smarthome.core/src/main/java/org/eclipse/smarthome/core/scheduler/ExpressionThreadPoolManager.java
```diff
@@ -225,8 +225,9 @@ public void schedule(final Runnable task, final Expression expression) {
 
         @Override
         public boolean remove(Runnable task) {
-            if (futures.get(task) != null && futures.get(task).cancel(false)) {
-                running.remove(task);
+            Future<?> future = futures.get(task);
+            if (future != null && future.cancel(false)) {
+                running.remove(future);
```
kaikreuzer: This is the fix - @kgoderis from my observations, we need to remove the future from `running`, not the task, right?

kaikreuzer:
This PR imho can be merged as is - the unit test that comes with it is marked as "ignore", so I'd hope that @kgoderis can come up with a separate PR that fixes the bug and enables the test.

bundles/core/org.eclipse.smarthome.core.test/src/test/java/org/eclipse/smarthome/core/scheduler/ExpressionThreadPoolExecutorTest.java
```diff
@@ -0,0 +1,45 @@
...
+        };
+
+        success = false;
+        scheduler.schedule(runnable, new CronExpression("0/1 * * * * ?"));
+        Thread.sleep(1500);
+        assertTrue(success);
+
+        boolean removed = scheduler.remove(runnable);
+
+        // TODO: This currently fails - why?
```
kgoderis: Yep

bundles/core/org.eclipse.smarthome.core/src/main/java/org/eclipse/smarthome/core/scheduler/ExpressionThreadPoolManager.java
kgoderis: I think it should be:
        
@Override
        public boolean remove(Runnable task) {
            Future<?> future = futures.get(task);
            if (future != null && future.cancel(false)) {
                futures.remove(future);
            }
            running.remove(task);
            return super.remove(task);
        }

running holds Runnables.

maggu2810:
As @kgoderis replied to this PR now, I assume it is okay to wait until you both agree.

bundles/core/org.eclipse.smarthome.core.test/src/test/java/org/eclipse/smarthome/core/scheduler/ExpressionThreadPoolExecutorTest.java
```diff
@@ -0,0 +1,45 @@
...
+        assertTrue(removed);
+
+        success = false;
+        Thread.sleep(1500);
+        assertFalse(success);
+
+        // TODO: What is this task that remains in the queue here? Shouldn't the queue be empty?
```
kgoderis: I do not know what it is not empty, but the scheduled queue holds the tasks that need to be scheduled, the scheduler itself holds the scheduled tasks to be execute d a given "delta" time interval, e.g the next occurrence of the task according to the expression.

bundles/core/org.eclipse.smarthome.core/src/main/java/org/eclipse/smarthome/core/scheduler/ExpressionThreadPoolManager.java
kaikreuzer: The runnable that is reached in `beforeExecute`is:
```
java.util.concurrent.ScheduledThreadPoolExecutor$ScheduledFutureTask@358f17ce
```
The one that is reached into `schedule` is
```
org.eclipse.smarthome.core.scheduler.ExpressionThreadPoolExecutorTest$1@6f55137
```

kgoderis: ScheduledFutureTask is implementing both the Runnable and Future interfaces. Could be the source of your confusion? Also, beware, schedule() is overridden. You refer to https://github.com/kaikreuzer/smarthome/blob/7d7ef907d709afe015a0c035b60bd2602307895f/bundles/core/org.eclipse.smarthome.core/src/main/java/org/eclipse/smarthome/core/scheduler/ExpressionThreadPoolManager.java#L212 , but then the other implementation is called at https://github.com/kaikreuzer/smarthome/blob/7d7ef907d709afe015a0c035b60bd2602307895f/bundles/core/org.eclipse.smarthome.core/src/main/java/org/eclipse/smarthome/core/scheduler/ExpressionThreadPoolManager.java#L162

kaikreuzer: It is not me who is confused, it is the JVM. The current implementation simply does not work, as `running` simply does not contain the instances that you try to remove. This PR fixes this.

bundles/core/org.eclipse.smarthome.core.test/src/test/java/org/eclipse/smarthome/core/scheduler/ExpressionThreadPoolExecutorTest.java
kaikreuzer: The test succeeds if I change the return line to
```
            return super.remove((Runnable) future);
```
So here again, the super implementation uses the future in the workQueue.

bundles/core/org.eclipse.smarthome.core.test/src/test/java/org/eclipse/smarthome/core/scheduler/ExpressionThreadPoolExecutorTest.java
kgoderis: Ouch. @maggu2810 and I changed the CronExpression in https://github.com/eclipse/smarthome/pull/2270. More efficient ways possibly, but it would break a bit the common algorithm the various Expression classes use to calculate the next execution time. I need to verify it in the debugger when I have a bit more time at hand.

kaikreuzer:
I have updated my (imho obvious) fixes, which I still think should be merged.
For all further discussions and fixes, I have opened https://github.com/eclipse/smarthome/issues/2394.

kgoderis:
@kaikreuzer I will push a PR shortly that solves the issue. It will partly incorporate the PR here above.

The problem was a bit deeper that we thought. First of all the Scheduler implementations transform Runnables into ScheduledTasks before pushing them through the queues and worker threads, which confused the hell out of us. Secondly, as you pointed out, we can have a multiple instances of a Runnable scheduled if the time to process/run the Runnable is larger than the scheduling interval. e.g. Scheduling a 3-second task every second is a use case we have to deal with. So, in short, we have much more tracking and Future management to do.

maggu2810:
As you have restored your branch would you like to reopen this PR or do we want to wait for another one?

kaikreuzer:
I deleted the branch by accident, but as @kgoderis mentioned that he will come up with a PR, I hope it will also include my work from this PR and thus it can stay closed.

PR #2402, opened 2016-11-03
Align string item value on same row in ControlUI

aounhaider1:
Before:
screen shot 2016-11-03 at 18 02 29

After:
screen shot 2016-11-03 at 18 02 08


aounhaider1:
Ok. But should it break to next line when there is not enough space in the second case?

kaikreuzer:
Yes, that's why I called it a "text area".

aounhaider1:
Updated.

kaikreuzer:
Could you provide new screenshots of how it now looks like?

aounhaider1:
Short text:
screen shot 2016-11-17 at 16 12 45

Long text:
screen shot 2016-11-17 at 16 12 26

kaikreuzer:
Thanks!

Issue #2361, opened 2016-10-26
TemplateProvider & ModuleTypeProvider interfaces should extend Provider interface, ModuleTypeRegistry & TemplateRegistry interfaces should extend Registry interface


adimova:
There is a problem that appears when replacing the service trackers with Declarative Services XMLs in the automation component "org.eclipse.smarthome.automation.providers". Declarative Services do not provide the ServiceRegistration object of a service. This requires to change the current mechanism for notifying the Rule Engine for new module types or rule templates appearing to the system. Currently, this is done via ServiceEvent-s: by invoking ServiceRegistration.setProperties(properties) method of the Provider, which leads to invoking ServiceTrackerCustomizer.modifiedService(ServiceReference reference, Object service) method of the Rule Engine. Relying on Declarative Services will brake this mechanism. This requires a change in the current interfaces ModuleTypeProvider, TemplateProvider, ModuleTypeRegistry and TemplateRegistry. ModuleTypeProvider and TemplateProvider should extend Provider interface, ModuleTypeRegistry and TemplateRegistry should extend Registry interface and their implementations should extend AbstractRegistry. This change will give us possibility to use a Provider-Listener mechanism for informing the Rule Engine for newly added module types or rule templates.
For such a change we need your approval, @kaikreuzer could you please give us your opinion

adimova:
Refers #1568

danchom:
I agree with that,  in that way we will have real registry services, simple implementation and we can introduce ModuleType/Template added, remove and updated events.

kaikreuzer:
Sounds good to me!

kaikreuzer:
This can be closed, right?

PR #2403, opened 2016-11-03
ModuleTypeRegistry & TemplateRegistry interfaces extends Registry int…

adimova:
…erface


adimova:
part of implementation of #2361

PR #2406, opened 2016-11-04
fix ColorItem type conversion

### LINKED ISSUE #2399: Color controls for HSB seem to be independant.
It seems that the 3 controls for color in PaperUI are independant controls?  If I slide the Sat, then I get a saturation, but H and B are 0. Likewise for others - if I change Hue, then S and B are 0.

For example, below I moved the brightness, and I get the following -:
```
2016-11-01 17:43:43.490 [DEBUG] [.c.thing.internal.ThingManager:458  ] - Delegating update '0,0,29' for item 'zwave_device_15785c713e3_node15_color_color' to handler for channel 'zwave:device:15785c713e3:node15:color_color'
```

The controls look as follows -:

image

I would have expected saturation and hue to be values as indicated in the image.

sjka:
fixes #2399

### DIFF
--- a/bundles/core/org.eclipse.smarthome.core.test/src/test/java/org/eclipse/smarthome/core/library/items/ColorItemTest.java
+++ b/bundles/core/org.eclipse.smarthome.core.test/src/test/java/org/eclipse/smarthome/core/library/items/ColorItemTest.java
@@ -1,0 +1,47 @@
+/**
+ * Copyright (c) 2014-2016 by the respective copyright holders.
+ * All rights reserved. This program and the accompanying materials
+ * are made available under the terms of the Eclipse Public License v1.0
+ * which accompanies this distribution, and is available at
+ * http://www.eclipse.org/legal/epl-v10.html
+ */
+package org.eclipse.smarthome.core.library.items;
+
+import static org.junit.Assert.assertEquals;
+
+import org.eclipse.smarthome.core.library.types.DecimalType;
+import org.eclipse.smarthome.core.library.types.HSBType;
+import org.eclipse.smarthome.core.library.types.PercentType;
+import org.junit.Test;
+
+/**
+ *
+ * @author Simon Kaufmann - initial contribution and API
+ *
+ */
+public class ColorItemTest {
+
+    @Test
+    public void testSetStateWithHSBType() {
+        ColorItem item = new ColorItem("test");
+        item.setState(new HSBType(new DecimalType(75), new PercentType(75), new PercentType(75)));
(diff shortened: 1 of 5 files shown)


bundles/core/org.eclipse.smarthome.core.test/src/test/java/org/eclipse/smarthome/core/library/items/ColorItemTest.java
```diff
@@ -0,0 +1,63 @@
...
+        ColorItem item = new ColorItem("test");
+        item.setState(new PercentType(50));
+        assertEquals(new HSBType(new DecimalType(0), new PercentType(0), new PercentType(50)), item.getState());
+    }
+
+    @Test
+    public void testSetStateWithDecimalType() {
+        ColorItem item = new ColorItem("test");
+        item.setState(new DecimalType(0.5));
+        assertEquals(new HSBType(new DecimalType(0), new PercentType(0), new PercentType(50)), item.getState());
```
kaikreuzer: DecimalType is NOT an accepted state - I would therefore expect that it is ignored or an exception is thrown.

Do we currently have situations in the code where a ColorItem has to deal with DecimalTypes?

bundles/core/org.eclipse.smarthome.core.test/src/test/java/org/eclipse/smarthome/core/library/items/ColorItemTest.java
```diff
@@ -0,0 +1,63 @@
...
+        item.setState(new PercentType(50));
+        assertEquals(new HSBType(new DecimalType(75), new PercentType(75), new PercentType(50)), item.getState());
+    }
+
+    @Test
+    public void testUpdateStateWithDecimalType() {
+        ColorItem item = new ColorItem("test");
+        item.setState(new HSBType(new DecimalType(75), new PercentType(75), new PercentType(75)));
+
+        item.setState(new DecimalType(0.5));
```
kaikreuzer: see above

bundles/core/org.eclipse.smarthome.core.test/src/test/java/org/eclipse/smarthome/core/library/items/ColorItemTest.java
sjka: > DecimalType is NOT an accepted state
okay, that's correct, I'll remove it.

> I would therefore expect that it is ignored or an exception is thrown
However, the check is not (and never was) done in the item itself. So far, if bogus was sent, the item was updated with it. I can enhance the ThingManager to really ignore a channel change and not update the item in such a case.

kaikreuzer:
Many thanks!

PR #2412, opened 2016-11-05
[ui] Give meaning to item= attribute on Image and Video sitemap widgets

watou:
`Image` and `Video` sitemap widgets have always accepted the `item=` attribute, but no client has given them meaning.  This PR changes the proxy servlet to first see if the widget has an associated item, if its current state is a `StringType`, and if the string state is a URL per RFC 2396.  If so, it proxies to that URL.  Otherwise, it considers the `url=` attribute as before.  The `url=` attribute is still required, acting as a default value in cases where `item=` is not specified or suitable.  The Classic UI required no changes to support this (since it reloads the sitemap page on updates), but the Basic UI was updated to support dynamic image URLs.  There is a PR in openhab-distro that updates the demo config to demonstrate this and works for both Classic and Basic UIs.

Still to consider:

* In addition to `StringType` containing a URL, it might be possible to add support for `ImageItem/RawType` to deliver images via the proxy servlet.
* The Basic UI does not support `Video item=` .
* The `url=` attribute is still required and acts as a default URL, but it might make sense to change the DSL to make `url=` optional if `item=` is given.
* Other clients should be updated to support dynamic `Image` and `Video` URLs.

Signed-off-by: John Cocula <john@cocula.com>


extensions/ui/org.eclipse.smarthome.ui.basic/web-src/smarthome.js
```diff
@@ -368,12 +368,16 @@
 		_t.image = parentNode.querySelector("img");
 		_t.updateInterval = parseInt(parentNode.getAttribute("data-update-interval"), 10);
 
-		if (_t.updateInterval === 0) {
-			return;
-		}
-
 		_t.url = _t.image.getAttribute("src").replace(/\d+$/, "");
 
+        _t.setValuePrivate = function() {
```
resetnow: I probably should have added this as eslint warning, but in the rest of the file tabs are used for indentation.

wep4you:
Image Resize would be great, if the target couldn't handle another size, the item should have the ability to resize it.

Video Streams for m3u8 seems to didn't work as i have tried for the netatmo welcome camera implementation.

watou:
Thanks Kai!

PR #2415, opened 2016-11-07
Implement tests for AbstractActiveService

m-memova:
I have implemented some tests for org.eclipse.smarthome.core.service.AbstractActiveService. 
If you have any issue with them, please let me know. 

Best Regards,
Mihaela


kaikreuzer:
Thanks @m-memova, although I am afraid, this was more or less in vain...

This class is a remainder from the old OH1 binding APIs that are no longer supported by ESH - actually, this class should have been removed, but that went unnoticed so far.

m-memova:
Well, it's my mistake that I didn't make an issue first :) Thank you, @kaikreuzer !

kaikreuzer:
Feel free to create a PR that deletes this class, though :-)

PR #2418, opened 2016-11-08
Sinope Binding

chaton78:
The sinope binding use the local API of the Sinope gateway. It doesn't require access to the sinope cloud (neviweb).

It supports for now, only the thermostats for baseboard.

For more information on the product:
http://www.sinopetech.com/en/


maggu2810:
Replaced by #2420

PR #2420, opened 2016-11-09
Sinope Binding

chaton78:
The sinope binding use the local API of the Sinope gateway. It doesn't require access to the sinope cloud (neviweb).

It supports for now, only the thermostats for baseboard.

For more information on the product:
http://www.sinopetech.com/en/

Signed-off-by: chaton78 <plarin@gmail.com>


chaton78:
Hum... 

Unable to read repository at http://download.eclipse.org/eclipse/updates/4.4. Unable to read repository at http://download.eclipse.org/eclipse/updates/4.4/R-4.4.2-201502041700/content.xml. Read timed out

Timed out? Do I use an invalid dependency? Or should I retrigger the build? how?

sjka:
Thanks a lot for this very clean contribution! And my apologies that it took so very long until somebody had a look at it.

I left a few comments inside the code. Thanks to the very well organized code they all should be easy to address. 

In addition, I would appreciate if you could rebase your code onto the latest master. I see that you have several merge commits inside your branch. This makes it quite hard to keep the overview.

Nevertheless, there is one bigger issue that we need to sort out. This is the (missing) jar library that you are using for the communication (which is of course a good idea to separated it). I assume it is the one you wrote [here](https://github.com/chaton78/sinope-core), right? You did not put a license to the project, but from looking at the comments inside the code it looks like you intend to put it under the EPL too. If we want to use this library in an Eclipse project, we would need this library to be licensed properly and also have it checked for its license, provenance and potential IP issues by the Eclipse IP process. This is of course not a big issue, we will happily do this if you provide us with a release version which you can after all these checks simply include here in your PR as a binary. But how about the protocol you implemented there? Is it published by sinopé somewhere? Under what conditions? Or did you reverse-engineer it?

These license/IP issues have to be clarified first before we can accept this contribution here. Otherwise we will find another good place for this binding, of course. Maybe https://github.com/openhab/openhab2-addons would be a good choice then...?

chaton78:
I am happy that you took the time to review it. It is appreciated! I'll come back to you shortly.

extensions/binding/org.eclipse.smarthome.binding.sinope/build.properties
```diff
@@ -0,0 +1,7 @@
+source.. = src/main/java/
+output.. = target/classes
+bin.includes = META-INF/,\
+               .,\
+               OSGI-INF/,\
+               lib/sinope-core.jar,\
```
chaton78: This is my own library.

maggu2810:
> These license/IP issues have to be clarified first before we can accept this contribution here. Otherwise we will find another good place for this binding, of course. Maybe https://github.com/openhab/openhab2-addons would be a good choice then...?

Has this been moved?

kaikreuzer:
Not that I am aware of. 
@chaton78 Why did you close this PR?

chaton78:
I closed by mistake.... Since then I redesigned the whole thing with your recommendations. I will move this to openhab2-addons.

PR #2422, opened 2016-11-09
moved event documentation from concepts to features

kaikreuzer:


docs/_includes/documentation-menu.html
```diff
@@ -8,7 +8,6 @@
 			<li><a href="{{docu}}/concepts/items.html">Items</a></li>
 			<li><a href="{{docu}}/concepts/things.html">Things</a></li>
 			<li><a href="{{docu}}/concepts/discovery.html">Inbox & Discovery</a></li>
-			<li><a href="{{docu}}/concepts/events.html">Events</a></li>
```
maggu2810: As the entry is removed from the `Concepts` shouldn't we add one to the `Features`?

kaikreuzer: No, I decided against it, because the content is mainly relevant for the REST API (and users that want to extend it) and from there it is referenced. I would not want to "promote" eventing to a feature on its own.

PR #2426, opened 2016-11-09
extracted interfaces for AudioManager and VoiceManager and have them …

kaikreuzer:
…implement ConfigOptionProvider

### DIFF
--- a/bundles/core/org.eclipse.smarthome.core.audio/src/main/java/org/eclipse/smarthome/core/audio/AudioFormat.java
+++ b/bundles/core/org.eclipse.smarthome.core.audio/src/main/java/org/eclipse/smarthome/core/audio/AudioFormat.java
@@ -7,6 +7,8 @@
  */
 package org.eclipse.smarthome.core.audio;
 
+import java.util.Set;
+
 /**
  * An audio format definition
  *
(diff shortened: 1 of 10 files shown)


kaikreuzer:
FTR: Just pushed a second commit to this branch, both can be squashed when merged. Will keep not push any further stuff, promised ;-)

kaikreuzer:
@maggu2810 I finally resolved the build problems on this branch, so everything ready from my side!

maggu2810:
Okay, it is already part of my task list.

bundles/core/org.eclipse.smarthome.core.audio/src/main/java/org/eclipse/smarthome/core/audio/AudioFormat.java
```diff
@@ -238,6 +240,132 @@ public boolean isCompatible(AudioFormat audioFormat) {
...
 
+    /**
+     * Determines the best match between a list of audio formats supported by a source and a sink.
+     *
+     * @param inputs the supported audio formats of an audio source
+     * @param outputs the supported audio formats of an audio sink
+     * @return the best matching format or null, if source and sink are incompatible
+     */
+    public static AudioFormat getBestMatch(Set<AudioFormat> inputs, Set<AudioFormat> outputs) {
+        AudioFormat preferredFormat = getPreferredFormat(inputs);
```
maggu2810: getPrefferedFormat can return null. Make it sense to look all the outputs (what is the intention of `output.isCompatible(null)`)?

bundles/core/org.eclipse.smarthome.core.audio/src/main/java/org/eclipse/smarthome/core/audio/AudioFormat.java
```diff
@@ -238,6 +240,132 @@ public boolean isCompatible(AudioFormat audioFormat) {
...
+        return null;
+    }
+
+    /**
+     * Gets the first concrete AudioFormat in the passed set or a preferred one
+     * based on 16bit, 16KHz, big endian default
+     *
+     * @param audioFormats The AudioFormats from which to choose
+     * @return The preferred AudioFormat. A passed concrete format is preferred adding
+     *         default values to an abstract AudioFormat in the passed Set.
```
maggu2810: The code states that a null could be returned. This should be documented.

maggu2810: `in the passed Set` - okay, the data type is `Set`, but in the text, I assume we should use a lower case (but I am not an expert) :wink:

bundles/core/org.eclipse.smarthome.core.audio/src/main/java/org/eclipse/smarthome/core/audio/AudioFormat.java
```diff
@@ -238,6 +240,132 @@ public boolean isCompatible(AudioFormat audioFormat) {
...
+
+                // These values must be interdependent (bitRate = bitDepth * frequency)
+                if (null == bitRate) {
+                    if (null == bitDepth) {
+                        bitDepth = new Integer(defaultBitDepth);
+                    }
+                    if (null == frequency) {
+                        frequency = new Long(defaultFrequency);
+                    }
+                    bitRate = new Integer(bitDepth.intValue() * frequency.intValue());
```
maggu2810: Just to be sure: If frequency is a "long", does it make sense to use only the int range for calculation? I assume that frequency is not higher as the max. int, but as long has been choosen, I want to be sure.

bundles/core/org.eclipse.smarthome.core.audio/src/main/java/org/eclipse/smarthome/core/audio/internal/AudioManagerImpl.java
```diff
@@ -0,0 +1,283 @@
...
+
+    @Override
+    public void playFile(String fileName) throws AudioException {
+        playFile(fileName, null);
+    }
+
+    @Override
+    public void playFile(String fileName, String sink) throws AudioException {
+        File file = new File(
+                ConfigConstants.getConfigFolder() + File.separator + SOUND_DIR + File.separator + fileName);
```
maggu2810: I know it is only moved code, but FYI: Theres a method for such path building: https://docs.oracle.com/javase/7/docs/api/java/nio/file/Paths.html#get(java.lang.String,%20java.lang.String...)

kaikreuzer:
Fixed some comments. Please note that ALL of that code was only moved, so it rather prefer not to do any deeper changes in the scope of this PR (like changing longs and ints).

PR #2429, opened 2016-11-10
Fixes #2428 , "mixed content" warning on Paper UI mapviews

### LINKED ISSUE #2428: Mixed content on Paper UI when showing Location items
When displaying Location items, a map segment from http://www.openstreetmap.org is included, which leads to a "mixed content" warning if the Ui is accessed via https (for example when behind a reverse proxy which adds SSL capability)

hakan42:
Signed-off-by: Hakan Tandogan hakan@tandogan.com

### DIFF
--- a/extensions/ui/org.eclipse.smarthome.ui.paper/web-src/js/controllers.control.js
+++ b/extensions/ui/org.eclipse.smarthome.ui.paper/web-src/js/controllers.control.js
@@ -579,7 +579,7 @@ angular.module('PaperUI.controllers.control', []).controller('ControlPageControl
             var bbox = (longitude - 0.01) + ',' + (latitude - 0.01) + ',' + (longitude + 0.01) + ',' + (latitude + 0.01);
             var marker = latitude + ',' + longitude;
             $scope.formattedState = latitude + '°N ' + longitude + '°E';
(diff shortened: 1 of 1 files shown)


kaikreuzer:
I manually checked the CLA, all fine.

maggu2810:
Don't we want to use the common scheme (as it is used by Git itself) anymore (enclose the mail in angle brackets)?

kaikreuzer:
We do, @hakan42 didn't. But I didn't take this as a reason to reject his PR as I am sure he will use correct signed-off-by lines in the future ;-)

hakan42:
Which I will do in the future 😄 

I just had an example commit message laying around in my development environment. Also, honestly, I created my last pull request many many months ago, so I am not up-to-date on the current development requirements.

maggu2810:
@hakan42 If you are using the Git command line client, you can use the "-s" argument on the commit command. This will add a correctly formatted signed-off-line by itself.
Also the EGit plugin for the Eclipse IDE can create a well formatted signed-offline.
The only thing you need to do is to setup the name and mail address in your Git settings once.

@kaikreuzer If we are using the squash + rebase mechanism of Github the person you merge the commit could fix the commit message and so a non-well formatted signed-off-by line :wink:

kaikreuzer:
@maggu2810 You're right, I didn't think of that. Will do in future!

PR #2435, opened 2016-11-10
Cron Expression : Fix #2394

### LINKED ISSUE #2394: Fix bugs in ESH scheduler and add more unit tests
When reviewing https://github.com/eclipse/smarthome/pull/2094, I noticed a couple of issues (see [here](https://github.com/eclipse/smarthome/pull/2094#discussion_r86155377), [here](https://github.com/eclipse/smarthome/pull/2391/files#r86192254) and [here](https://github.com/eclipse/smarthome/pull/2391/files#r86191773)) that definitely would result in regressions wrt cron scheduled rules.

I think we need to do more extensive testing so that we can be sure that we do not break the functionality for existing (mainly OH2) users. There is no test coverage for the scheduler functionality itself yet, only for the expression parsing part - this needs to be addressed as well.

kgoderis:
It does not provide additional tests, but should now correctly deal with cron expressions executing tasks that run longer than the scheduling interval of the expression itself.

Signed-Off-By: Karel Goderis <karel.goderis@me.com>


kaikreuzer:
Are any of the FIXME comments in the unit test addressed? If so, please remove those comments and add an according test (e.g. one that calls `scheduler.remove(runnable)`).

kgoderis:
Did you want a separate test? the one you provided is now testing positive, and also contains .remove().

Squashed.

kaikreuzer:
I was referring to [this FIXME](https://github.com/eclipse/smarthome/pull/2435/commits/88218cd7b41e34a9b86e9b9a82b2a932e2d6dae8#diff-6ab8869df39e4b2f961a094a0437b037R33). I cannot see this being covered by the test?

kgoderis:
Ah ok. You want the method overloading to be fool proof.... Put this PR on hold, I also have to fix an issue whereby cron jobs do not support millisecond definitions

kgoderis:
@kaikreuzer I think we are ok now. I have added a second commit that speeds up the cron expression (see also https://github.com/eclipse/smarthome/issues/2394#issuecomment-258881466) many many times.

kgoderis:
@kaikreuzer I have made some further adjustments wrt usage of milliseconds in Dates. As you know, none of the Expressions today support expressions that deeper than the level of a second, and thus, milliseconds in Dates are not relevant. But since we do a lot Date comparisons in our code, I cleaned up things by setting the Calendar.MILLISECONDS fields of Dates to 0. 

Consider

`Date someDate = someExpression.getTimeAfter(startDate)  
`
with startDate.MILLISECONDS being 0

If startDate is a valid date for the given expression , can someDate = startDate,  or should this rather be the next instance, e.g. startDate + 1 second. Under normal circumstances, when repeatedly calling getTimeAfter, like in a scheduler loop, the computing time elapsed in that loop will be > 1 millisecond, and thus, a next call to getTimeAfter with an updated startDate will yield a Date that advances with the minimal 1 second. Are we ok with this assumption?

kgoderis:
@kaikreuzer not sure why ip-validation if failing for this PR.

kgoderis:
@kaikreuzer Travis is exiting with

Error: JAVA_HOME is not defined correctly.
  We cannot execute /usr/lib/jvm/java-8-oracle-amd64/bin/java
The command "mvn clean install -B -V" exited with 1.

kaikreuzer:
I assume you refer to [this code](https://github.com/eclipse/smarthome/pull/2435/files#diff-7765b0fd1d5b50810208805bd0897fefR181)?
Yes, it sounds reasonable to me to consider it to be the next full second.

What is the general status of this PR? It does not yet include https://github.com/eclipse/smarthome/pull/2391, nor does it fix https://github.com/eclipse/smarthome/issues/2394, so it does not seem to be ready to be merged.

I also repeated my test with the non-model persistence PR and still have the same issue there:

```
2016-11-22 13:46:40.345 [INFO ] [.e.s.m.c.i.ModelRepositoryImpl:114  ] - Refreshing model 'rrd4j.persist'
2016-11-22 13:47:47.496 [WARN ] [o.e.s.c.p.PersistenceManager  :317  ] - Failed to delete cron job for dbId 'rrd4j'
2016-11-22 13:47:47.805 [DEBUG] [o.e.s.c.p.PersistenceManager  :298  ] - Scheduled strategy everyMinute with cron expression 0/1 * * * * ?
```

kgoderis:
> What is the general status of this PR? It does not yet include #2391, nor does it fix #2394, so it does not seem to be ready to be merged.

They should both be solved, for #2391 the code has changed quite a bit so the lines we have been discussing are not there anymore. #2394 I have copied over the test you provided, and it did not fail, so I am a bit curious as what is happening in the persistence bit you are testing

kgoderis:
> I assume you refer to this code?

yes indeed

kgoderis:
> I also repeated my test with the non-model persistence PR and still have the same issue there:

can you point me to which branch/repo you are testing against?

kgoderis:
@kaikreuzer I can not reproduce your problem. FYI I have taken maggu2810/persistence-model, then cherry-picked this PR, made some changes to Expression.java as this was behind in magi's branch, and fired it up. Then modified rrd4j.persist and this is what I get:

```
2016-11-22 22:23:13.476 [INFO ] [.e.s.m.c.i.ModelRepositoryImpl:118  ] - Refreshing model 'rrd4j.persist'
2016-11-22 22:23:13.486 [DEBUG] [o.e.s.m.p.i.PersistenceManager:460  ] - Removed scheduled cron job for strategy 'rrd4j.everyMinute'
2016-11-22 22:23:13.498 [DEBUG] [o.e.s.m.p.i.PersistenceManager:437  ] - Scheduled strategy rrd4j.everyMinute with cron expression 0/1 * * * * ?
```

I am coming in via the DSL, which is not the same as your test I guess, but I was not sure what to do in order to re-enact your scenario

kaikreuzer:
Yes, I did exactly the same in my tests - weird that it behaves differently for you.
Anyhow, this PR seems to make stuff better, so let's merge it. I will continue testing and come up with more detailed bug reports, if I manage to reproduce something. Thanks!

Issue #2437, opened 2016-11-11
Static Code Analysis Checks


svilenvul:
I would like to continue a discussion that started here : [Add Analysis Tool for New Bindings ](https://github.com/openhab/openhab2-addons/issues/1289)

I will give some highlights :) : 

> svilenvul commented on 4 Oct • edited
> I would like to start a discussion about the integration of an analysis tool for PRs in the openhab2-addons repo. This should speed up the review process for new bindings.

> maggu2810 commented on 4 Oct 
>..
>For my projects I use maven plugins for Checkstyle, FindBugs, PMD (you stated them above) etc.
The Jenkins instance checks the Quality Gates (execute all the plugins and continues on error - generate the report only) and Jenkins will fail if the report breaks the limits.

>I also use a pre commit Maven profile that fails if an error occurred and use it before doing a commit (myself).

> MHerbst commented on 4 Oct 
> First of all I like the idea of automatically checking as much as possible. I also would prefer to use only Checkstyle, Findbugs and PMD. We are using them in our company very similar to @maggu2810. And especially the Findbugs checks are really helpful.
>...

> kaikreuzer commented on 5 Oct 
> I agree with @maggu2810 and @MHerbst. It would be ideal if all tools could be used by the developers locally before even creating a PR. A Maven profile should enable those checks and new PRs would be build on Jenkins with this profile.

> What should be tried is to make such a mechanism easily extensible and configurable; we e.g. might e.g. have a different set of checks for bindings than for UIs. And the guidelines might evolve over time. Also note that almost everything is applicable to ESH as well, so openHAB2 might not be the right place for such a feature (at least not for most of it).

>There was a discussion for ESH a long time ago about such automatic checks and I have seen Maven plugins that used Groovy to do more advanced checks - this might be a good solution for all non-Java files.

>Last but not least, it would be nice if a kind of summary/report could be created; I expect that we might have warnings (besides errors), which would NOT stop the build, but which we should nonetheless get to the attention of the developer - if those warnings only appear somewhere in the Maven build log, I doubt that anyone will ever see them.

svilenvul:
I have built a prototype following some of the things discussed above, a PR will be opened shortly.

kaikreuzer:
Great, looking forward to it!

svilenvul:
I have created a simple online poll, where you can vote for the automated code checks that you want to be added to the Static Code Analysis Tool. Please share your opinion, you have only to enter a name to vote.

[...]

PR #2440, opened 2016-11-11
[WIP] Static code analysis tool

svilenvul:
See #2437 

- Plugins for PMD, Checkstyle and FindBugs are integrated with the Maven
build (in a separate profile).
- Created a project for a custom defined rules and configuration data.
- Added a project that creates a summary report for each binding.

Please read "build-tools/README.md" for more information.


kaikreuzer:
Thanks @svilenvul, this looks like some fantastic piece of work!

I only briefly started looking into it, but failed to run the build right from the start. I get the message:

```
[ERROR] Failed to execute goal org.apache.maven.plugins:maven-assembly-plugin:2.6:single (make-assembly) on project static-code-analysis-report: Unable to parse configuration of mojo org.apache.maven.plugins:maven-assembly-plugin:2.6:single for parameter descriptorRefs: Cannot assign configuration entry 'descriptorRefs' with value 'jar-with-dependencies' of type java.lang.String to property of type java.lang.String[] -> [Help 1]
```
Any idea what might be the cause of this?

Another question from a first glance: Wouldn't it make sense to put it into tools instead of creating a new folder build-tools?

Without having a deeper understanding of how it all works yet, I was wondering, how it could be easily transferred to other builds, like e.g. https://github.com/openhab/openhab2-addons (where you first wanted to contribute it) or also single binding repos like https://github.com/openhab/org.openhab.binding.zwave.

My hope was to have a single plugin (that is built here in the ESH build and published to repo.eclipse.org) which could then be added by those other builds into their pom.xml to make use of it. Do you see a chance to wrap the currently single plugins for PMD, Checkstyle, FindBugs and ESH into a single plugin?

svilenvul:
Thanks @kaikreuzer 

>I only briefly started looking into it, but failed to run the build right from the start. I get the message:
...
>Any idea what might be the cause of this?

Yes, I think with the updated version should be ok. 

>Another question from a first glance: Wouldn't it make sense to put it into tools instead of creating a new folder build-tools?

Yes, it will be better.

> Without having a deeper understanding of how it all works yet, I was wondering, how it could be easily transferred to other builds, like e.g. https://github.com/openhab/openhab2-addons (where you first wanted to contribute it) or also single binding repos like https://github.com/openhab/org.openhab.binding.zwave.
> My hope was to have a single plugin (that is built here in the ESH build and published to repo.eclipse.org) which could then be added by those other builds into their pom.xml to make use of it. 

Yes it will be easily added, only a reference in the .pom (export the profile) will be enough.

>Do you see a chance to wrap the currently single plugins for PMD, Checkstyle, FindBugs and ESH into a single plugin?

The first module ('static-code-analysis-config') actually does this. I would say the second one ('static-code-analysis-report') is optional, it is used only to aggregate the reports into single .html page. We should decide if we want to use it or replace it with the [https://maven.apache.org/components/plugins/maven-site-plugin/](Maven Site plugin). If we keep it I can combine both projects into a single one.

extensions/binding/org.eclipse.smarthome.binding.hue.test/src/test/groovy/org/eclipse/smarthome/binding/hue/test/HueBridgeHandlerOSGiTest.groovy
```diff
@@ -228,7 +228,7 @@ class HueBridgeHandlerOSGiTest extends OSGiTest {
     }
 	
     @Test
-    void 'assert that a status configuration message for missing bridge IP is properly returned (IP is null)'() {
+    void 'assert that a status configuration message for missing bridge IP is properly returned - IP is null'() {
```
svilenvul: I had some problems with FindBugs, this method name caused FindBugs to crash.
The Groovy method is compiled in a separate .class and when FindBugs checks the .class file it does not expect the class name to contain braces.

pom.xml
```diff
@@ -440,6 +446,233 @@
...
+                  <version>${project.version}</version>
+                  <type>jar</type>
+                </dependency>
+              </dependencies>
+            </plugin>
+          </plugins>
+        </pluginManagement>
+      </build>
+    </profile>
+    <profile>
```
svilenvul: This may seem a little confusing ...
I use a separate profile to activate the plugins for FindBugs, PMD and Checkstyle and to apply specific configuration data for a set of plug-ins (in this case bundles).

kaikreuzer:
Thanks @svilenvul, I was not able to run the build.
As we seem to have high prio issues, we have to set `fail.on.error` to false, otherwise the ESH build fails, right?

The reports show MANY comments - did you already think about customizing the set of rules, especially for checkstyle? I think we should start with a small set of rules, so that the reports do not flood the reader with unimportant details (there is e.g. a huge number of "is a magic number" hits, which I doubt that we should address at all).

> Yes it will be easily added, only a reference in the .pom (export the profile) will be enough.

Do you have an example of what e.g. would have to be added to https://github.com/openhab/openhab2-addons/blob/master/pom.xml in order to activate checks for bindings (only)?

kaikreuzer:
Another general question about this PR: The idea of this tool was to check if contributors respected [our guidelines](https://www.eclipse.org/smarthome/documentation/development/guidelines.html). Tools like Checkstyle can probably cover most of (A), but do you already have any concept for checking rules from (B), which are more related to OSGi?

svilenvul:
@kaikreuzer 
>As we seem to have high prio issues, we have to set fail.on.error to false, otherwise the ESH build fails, right?

For now I think it is better to set fail.on.error to false. After we set the right priority to each issue we can think to set it back to true to prevent the occurrence of new high prio issues.

> The reports show MANY comments - did you already think about customizing the set of rules, especially for checkstyle? I think we should start with a small set of rules, so that the reports do not flood the reader ..

Yes, I will remove the rules that are flooding the report.

>Do you have an example of what e.g. would have to be added to https://github.com/openhab/openhab2-addons/blob/master/pom.xml in order to activate checks for bindings (only)?

At the moment this could be achieved by adding [these lines](https://github.com/MusalaSoft/openhab2-addons/commit/a4125365598b52358b2e072d96c14705503b643c).

 >Another general question about this PR: The idea of this tool was to check if contributors respected our guidelines. Tools like Checkstyle can probably cover most of (A), but do you already have any concept for checking rules from (B), which are more related to OSGi?

We can create custom rules for PMD e.g. This may be a little bit tricky as at the moment PMD [does not support multiple languages with a single execution](https://issues.apache.org/jira/browse/MPMD-183) and does not offer a support for .MF files. Support for new language can be added according to the documentation - [How to add a new language to PMD](http://pmd.sourceforge.net/pmd-5.4.1/customizing/new-language.html), but at the first glance it looks for me too complicated for our needs.

svilenvul:
@kaikreuzer 
> Tools like Checkstyle can probably cover most of (A), but do you already have any concept for checking rules from (B), which are more related to OSGi?

A [simple Checkstyle check](https://github.com/eclipse/smarthome/pull/2440/commits/942dd4f2c5222fb48cafe45cec9c3693850626c9#diff-f9e9260ce62314e67b07184ad1bc2e97) that parses a MANIFEST.MF file and checks if any internal package is exported is added with a [test case](https://github.com/eclipse/smarthome/pull/2440/commits/942dd4f2c5222fb48cafe45cec9c3693850626c9#diff-7b73d752559f931d569be70bd82ca8ee). So I guess we can continue with checks that cover the different scenarios in (B).

svilenvul:
>My hope was to have a single plugin (that is built here in the ESH build and published to repo.eclipse.org) 

I am working now on merging the both projects into a single one and moving most of the configuration out of the parent .pom. I will update the PR as soon as I have a satisfiable result.

svilenvul:
@kaikreuzer 
I have made most of the changes that we have talked about in this conversation.

Now the tool performs the following checks:
- FindBugs runs around 60 detectors that are listed [here ](https://github.com/eclipse/smarthome/pull/2440/files#diff-d4efb7f8df9cc7d7aa62e109982b3fe6R9). They report only high priority issues and fail the build, if some issue is found. I have found a few violations and fixed them to keep the build passing (when fail.on.error is set to true);
- PMD runs the [following checks](https://github.com/eclipse/smarthome/pull/2440/files#diff-81be318148448f3cf969decac8e4b4c6) that have found as well some broken null checks like [this one ](https://github.com/eclipse/smarthome/pull/2440/commits/8f86281c15541307a995c1fc169c85e38571844f#diff-7bd45e2de9d21e0580d3a0de7c5ba731);
- Checkstyle is configured to check for [this](https://github.com/eclipse/smarthome/pull/2440/files#diff-0099ca9db31f6b777ddc6c2b6ef1179d).

There is still of course a lot to be done - mostly refining the violations that are reported (we have some false positives mostly from Checkstyle), maybe extending some existing checks and adding new custom checks (following the ESH guidelines first).

My question is would you like to merge the tool as it is now (and with the required adjustments of course) or keep it as a long running feature branch ?

And only a side note. In the target folder of the code analysis project the tool generates a summary report that contains links to the individual reports, so you can easily navigate between them.

kaikreuzer:
Thanks @svilenvul!

> My question is would you like to merge the tool as it is now (and with the required adjustments of course) or keep it as a long running feature branch?

As it will require some efforts to get the CQs approved and have it all cleaned and sorted, I'd prefer to only go to merge it as soon as the main functionality is covered - and this is for me still the check of the guidelines, that's really what I need to reduce the work load on reviews.

> In the target folder of the code analysis project the tool generates a summary report that contains links to the individual reports

It seems to still create a folder "build-tools" in the project root, is this intended? Shouldn't the summary report end up in a ./target folder in the project root?

>> Do you have an example of what e.g. would have to be added to https://github.com/openhab/openhab2-addons/blob/master/pom.xml in order to activate checks for bindings (only)?
> At the moment this could be achieved by adding these lines.

Hm, this is still a LOT of code and everybody who uses it (e.g. in individual repos for single bindings) will have to keep this in sync wrt its settings, versions etc. What I had in mind was a single Maven plugin that one could add and which would wrap all of this (and which could be centrally maintained at ESH). Is this just a wild dream or something that you would reckon to be possible?

svilenvul:
Thanks @kaikreuzer !
I will address your comments and update the PR.

svilenvul:
@kaikreuzer 

>Hm, this is still a LOT of code and everybody who uses it (e.g. in individual repos for single bindings) will have to keep this in sync wrt its settings, versions etc. What I had in mind was a single Maven plugin that one could add and which would wrap all of this (and which could be centrally maintained at ESH). Is this just a wild dream or something that you would reckon to be possible?

I have created a single Maven plugin that executes the plugins for PMD, FindBugs and Checkstyle and merges the individual reports. 

When you would like to test it, just keep it mind that you might have to delete the old artifact (static-code-analysis) from the local .m2 repository, otherwise Maven might produce an error that the Maven plugin does not contain descriptor and will not build the new plugin at all.

> It seems to still create a folder "build-tools" in the project root, is this intended? Shouldn't the summary report end up in a ./target folder in the project root?

"build-tools" is deleted and the summary report is moved to target folder in the project root.

tools/static-code-analysis/src/main/java/org/eclipse/smarthome/tools/analysis/tools/CheckstyleChecker.java
```diff
@@ -0,0 +1,99 @@
...
+
+        String key = CHECKSTYLE_RULESET_USER_PROPERTY;
+        URL rulesetLocation = cl.getResource(ruleset);
+        String absoluteLocation = rulesetLocation.toString();
+        log.debug("Config file found at " + absoluteLocation);
+        userProps.setProperty(key, absoluteLocation);
+
+        // Add the static-code-analysis plugin as dependency to maven-checkstyle-plugin, because this plugin contains
+        // custom checks
+        Dependency dep = dependency(SMARTHOME_TOOLS_GROUP_ID, SMARTHOME_TOOLS_ARTIFACT_ID, mavenProject.getVersion());
```
kaikreuzer: This is a subtle bug: This will use the maven version of the project which is using this plugin - I e.g. used it on openHAB2-addons, version 2.1.0-SNAPSHOT, which made the plugin fail with an error (since it tried to find static-code-analysis in version 2.1.0-SNAPSHOT instead of 0.9.0-SNAPSHOT).

Same bug for the other mojos.

svilenvul:
Thanks @kaikreuzer 
I have resolved the conflicts with the master branch.

maggu2810:
@svilenvul WDYT about https://github.com/KengoTODA/findbugs-slf4j
Also a good one to add?

svilenvul:
Thanks @maggu2810 I have added the plugin.

pom.xml
```diff
@@ -447,6 +447,85 @@
...
+      <build>
+        <plugins>
+          <plugin>
+            <groupId>org.eclipse.smarthome.tools</groupId>
+            <artifactId>static-code-analysis</artifactId>
+            <version>${project.version}</version>
+            <configuration>
+              <ruleset>bundle</ruleset>
+              <findbugsPlugins>
+                <findbugsPlugin>
```
maggu2810: Shouldn't this plugin added internal in your static-code-analysis artifact.
I don't think it should be an optional plugin that could be added optional on `static-code-analysis` plugin usage.
It should be a mandatory findbugs plugin which is added by the plugin itself internal.

kaikreuzer:
@svilenvul Many thanks again for your work here. After doing some initial assessment of the license situation, I have the concern that it will be very hard to get that through the Eclipse IP process - especially as I have also found LGPL parts in it.

As I don't think it is necessary to have this tooling within the ESH project itself, I would suggest to maintain it separately and only USE this tool within ESH during build time - this will be much easier regarding IP approval. I discussed this with @dstivanov as well already.

I have therefore taken the liberty to take your code and push it here: https://github.com/openhab/static-code-analysis/commit/4f6b4b3fd3aa8821c23c01774c5690a4c2d4ea26
I'll start [applying it on openHAB2-addons](https://github.com/openhab/openhab2-addons/pull/1987) and we can iterate through further improvements there, before integrating it in the ESH build itself. 

Let's do all further issue tracking and discussions at https://github.com/openhab/static-code-analysis.

PR #2449, opened 2016-11-12
Classic UI: set default values for config settings

lolodomo:


kaikreuzer:
Thanks!

PR #2457, opened 2016-11-13
Sonos binding: clean up ZonePlayerConfiguration (fix #2445)

### LINKED ISSUE #2445: Sonos binding: class ZonePlayerConfiguration
his class is unused except for constants UDN and IDENTICATION.
It should be fixed to match the real configuration of a Sonos thing, that is with only two fields: udn and refresh. Then this class could be used in ZonePlayerHandler. If not used, the class should be simply removed.
IDENTIFICATION constant should probably be moved to SonosBindingConstants.

lolodomo:


kaikreuzer:
@kgoderis Could you please review and comment?

extensions/binding/org.eclipse.smarthome.binding.sonos/src/main/java/org/eclipse/smarthome/binding/sonos/handler/ZonePlayerHandler.java
```diff
@@ -1002,7 +1000,11 @@ public boolean isGroupCoordinator() {
 
     @Override
     public String getUDN() {
```
kgoderis: Is there a specific reason for defining an udn member variable (Line 163), assigning this a value, but never use that again afterwards? why not calling getConfigAs() etc. each time, and be sure to have the latest value each time (even as it does not change ofter or at all)?

lolodomo: This was to avoid checking the thing  config each time. But you are right, i can suppress the udn member and check the thing config each time. Will fix that.

lolodomo:
@kgoderis : PR updated, my last change concerns only the suppress of member udn and the update of the method getUDN().

kgoderis:
@kaikreuzer LGTM. Cosmetic changes for better readability of the code.

PR #2459, opened 2016-11-13
Initial contribution of audio and voice modules for automation engine

kaikreuzer:

### DIFF
--- a/bundles/automation/org.eclipse.smarthome.automation.module.media/src/main/java/org/eclipse/smarthome/automation/module/media/handler/PlayActionHandler.java
+++ b/bundles/automation/org.eclipse.smarthome.automation.module.media/src/main/java/org/eclipse/smarthome/automation/module/media/handler/PlayActionHandler.java
@@ -1,0 +1,51 @@
+/**
+ * Copyright (c) 2016 Kai Kreuzer and others.
+ * All rights reserved. This program and the accompanying materials
+ * are made available under the terms of the Eclipse Public License v1.0
+ * which accompanies this distribution, and is available at
+ * http://www.eclipse.org/legal/epl-v10.html
+ */
+package org.eclipse.smarthome.automation.module.media.handler;
+
+import java.util.Map;
+
+import org.eclipse.smarthome.automation.Action;
+import org.eclipse.smarthome.automation.handler.ActionHandler;
+import org.eclipse.smarthome.automation.handler.BaseModuleHandler;
+import org.eclipse.smarthome.core.audio.AudioException;
+import org.eclipse.smarthome.core.audio.AudioManager;
+import org.slf4j.Logger;
+import org.slf4j.LoggerFactory;
+
+/**
+ * This is an ModuleHandler implementation for Actions that play a sound file from the file system.
+ *
+ * @author Kai Kreuzer - Initial contribution and API
+ *
+ */
+public class PlayActionHandler extends BaseModuleHandler<Action> implements ActionHandler {
+
+    public static final String TYPE_ID = "PlayAction";
+    public static final String PARAM_SOUND = "sound";
+
+    private final Logger logger = LoggerFactory.getLogger(PlayActionHandler.class);
+
+    private AudioManager audioManager;
+
+    public PlayActionHandler(Action module, AudioManager audioManager) {
+        super(module);
+        this.audioManager = audioManager;
+    }
+
+    @Override
+    public Map<String, Object> execute(Map<String, ?> context) {
+        String sound = module.getConfiguration().get(PARAM_SOUND).toString();
+        try {
+            audioManager.playFile(sound);
+        } catch (AudioException e) {
+            logger.error("Error playing sound '{}': {}", sound, e.getMessage());
+        }
+        return null;
+    }
+
+}
--- a/bundles/automation/org.eclipse.smarthome.automation.module.media/src/main/java/org/eclipse/smarthome/automation/module/media/handler/PlayActionHandler.java
+++ b/bundles/automation/org.eclipse.smarthome.automation.module.media/src/main/java/org/eclipse/smarthome/automation/module/media/handler/PlayActionHandler.java
@@ -30,7 +30,7 @@ public class PlayActionHandler extends BaseModuleHandler<Action> implements Acti
 
     private final Logger logger = LoggerFactory.getLogger(PlayActionHandler.class);
 
-    private AudioManager audioManager;
+    private final AudioManager audioManager;
 
     public PlayActionHandler(Action module, AudioManager audioManager) {
         super(module);
--- a/bundles/automation/org.eclipse.smarthome.automation.module.media/src/main/java/org/eclipse/smarthome/automation/module/media/handler/SayActionHandler.java
+++ b/bundles/automation/org.eclipse.smarthome.automation.module.media/src/main/java/org/eclipse/smarthome/automation/module/media/handler/SayActionHandler.java
@@ -1,0 +1,42 @@
+/**
+ * Copyright (c) 2016 Kai Kreuzer and others.
+ * All rights reserved. This program and the accompanying materials
+ * are made available under the terms of the Eclipse Public License v1.0
+ * which accompanies this distribution, and is available at
+ * http://www.eclipse.org/legal/epl-v10.html
+ */
+package org.eclipse.smarthome.automation.module.media.handler;
+
+import java.util.Map;
+
+import org.eclipse.smarthome.automation.Action;
+import org.eclipse.smarthome.automation.handler.ActionHandler;
+import org.eclipse.smarthome.automation.handler.BaseModuleHandler;
+import org.eclipse.smarthome.core.voice.VoiceManager;
+
+/**
+ * This is an ModuleHandler implementation for Actions that trigger a TTS output through "say".
+ *
+ * @author Kai Kreuzer - Initial contribution and API
+ *
+ */
+public class SayActionHandler extends BaseModuleHandler<Action> implements ActionHandler {
+
+    public static final String TYPE_ID = "SayAction";
+    private static final String PARAM_TEXT = "text";
+
+    private VoiceManager voiceManager;
+
+    public SayActionHandler(Action module, VoiceManager voiceManager) {
+        super(module);
+        this.voiceManager = voiceManager;
+    }
+
+    @Override
+    public Map<String, Object> execute(Map<String, ?> context) {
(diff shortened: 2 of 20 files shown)


bundles/automation/org.eclipse.smarthome.automation.module.media/src/main/java/org/eclipse/smarthome/automation/module/media/handler/PlayActionHandler.java
```diff
@@ -0,0 +1,51 @@
...
+ *
+ */
+public class PlayActionHandler extends BaseModuleHandler<Action> implements ActionHandler {
+
+    public static final String TYPE_ID = "PlayAction";
+    public static final String PARAM_SOUND = "sound";
+
+    private final Logger logger = LoggerFactory.getLogger(PlayActionHandler.class);
+
+    private AudioManager audioManager;
```
maggu2810: If its only set in the constructor you could make it final, so it could never be changed again.

bundles/automation/org.eclipse.smarthome.automation.module.media/src/main/java/org/eclipse/smarthome/automation/module/media/handler/PlayActionHandler.java
```diff
@@ -0,0 +1,51 @@
...
+
+    public PlayActionHandler(Action module, AudioManager audioManager) {
+        super(module);
+        this.audioManager = audioManager;
+    }
+
+    @Override
+    public Map<String, Object> execute(Map<String, ?> context) {
+        String sound = module.getConfiguration().get(PARAM_SOUND).toString();
```
maggu2810: Is it guaranteed that ever parameter is available in the configuration? Or is this a potential NPE?

bundles/automation/org.eclipse.smarthome.automation.module.media/src/main/java/org/eclipse/smarthome/automation/module/media/handler/SayActionHandler.java
```diff
@@ -0,0 +1,42 @@
...
+ *
+ * @author Kai Kreuzer - Initial contribution and API
+ *
+ */
+public class SayActionHandler extends BaseModuleHandler<Action> implements ActionHandler {
+
+    public static final String TYPE_ID = "SayAction";
+    private static final String PARAM_TEXT = "text";
+
+    private VoiceManager voiceManager;
```
maggu2810: see above

bundles/automation/org.eclipse.smarthome.automation.module.media/src/main/java/org/eclipse/smarthome/automation/module/media/handler/SayActionHandler.java
```diff
@@ -0,0 +1,42 @@
...
+
+    public SayActionHandler(Action module, VoiceManager voiceManager) {
+        super(module);
+        this.voiceManager = voiceManager;
+    }
+
+    @Override
+    public Map<String, Object> execute(Map<String, ?> context) {
+        String text = module.getConfiguration().get(PARAM_TEXT).toString();
```
maggu2810: see above

bundles/automation/org.eclipse.smarthome.automation.module.media/src/main/java/org/eclipse/smarthome/automation/module/media/internal/MediaActionTypeProvider.java
```diff
@@ -0,0 +1,89 @@
...
+        }
+    }
+
+    @Override
+    public Collection<ModuleType> getModuleTypes(Locale locale) {
+        return Collections.singleton(getPlayActionType(locale));
+    }
+
+    private ModuleType getPlayActionType(Locale locale) {
+        return new ActionType("PlayAction", getConfigDesc(), "Play Sound", "Plays a sound file", null,
```
maggu2810: Could be created in the constructor once and stored in a member variable as it seems like a constant for this class. But also okay to leave it as it is. Just comment and leave the decision up to you.

bundles/automation/org.eclipse.smarthome.automation.module.media/src/main/java/org/eclipse/smarthome/automation/module/media/internal/MediaActionTypeProvider.java
```diff
@@ -0,0 +1,89 @@
...
+     *
+     * @return a list of parameter options representing the sound files
+     */
+    private List<ParameterOption> getOptions() {
+        List<ParameterOption> options = new ArrayList<>();
+        File soundsDir = Paths.get(ConfigConstants.getConfigFolder(), AudioManager.SOUND_DIR).toFile();
+        if (soundsDir.isDirectory()) {
+            for (String fileName : soundsDir.list()) {
+                if (fileName.contains(".")) {
+                    String soundName = StringUtils.capitalize(fileName.substring(0, fileName.indexOf(".")));
```
maggu2810: Should we use "last index of" or really the first one?

Here I come up with a question:
If I have multiple audio sinks, the one supports mp3 streams, the other one wav streams.
IIRC at the moment we have to supple a compatible format.
If I have a doorbell mp3 and wav in my sounds directory, who should choose the correct one?
As the current implementation forces the user to use the correct one, I don't know if we should cut off the file name and drop the information (as a List is used, we have two doorbell entries).

bundles/automation/org.eclipse.smarthome.automation.module.media/src/main/java/org/eclipse/smarthome/automation/module/media/handler/PlayActionHandler.java
kaikreuzer: It is defined as required, so the automation framework should make sure that it is always set.

bundles/automation/org.eclipse.smarthome.automation.module.media/src/main/java/org/eclipse/smarthome/automation/module/media/internal/MediaActionTypeProvider.java
kaikreuzer: The whole point about this class is to construct the type (specifically its config parameter options) dynamically on every new request.

danchom:
LGTM, I think the idea for MediaActionTypeProvider is really cool. I don't know if the VoiceManager can be registered without AudioManager service and vice versa? If it is true may be the static dependency to both services is not needed.

bundles/automation/org.eclipse.smarthome.automation.module.media/src/main/java/org/eclipse/smarthome/automation/module/media/internal/MediaActionTypeProvider.java
kaikreuzer: I think most solutions will decide which formats they (and their sinks) support, so from a (dumb) user perspective I think the format should not be displayed in this list.
What I would like to add in future is another parameter for the sink to use (if not set, the default is used). This could then be taken into account when creating the option list, e.g. filtering out mp3 if the sink does not support it.

kaikreuzer:
> may be the static dependency to both services is not needed.

You mean the AudioAction should be made available if no voice feature is available? In theory yes, but then the cleaner approach would be to create two separate bundles. I decided against this for simplification reasons.

> I think the idea for MediaActionTypeProvider is really cool. 

Actually no. As mentioned [here](https://github.com/eclipse/smarthome/pull/2459/files#diff-4129c37841ca9bc559fc940481b839b8R35), this was only necessary for the dynamic options. Usually we use [ConfigOptionProvider](https://github.com/eclipse/smarthome/blob/master/bundles/config/org.eclipse.smarthome.config.core/src/main/java/org/eclipse/smarthome/config/core/ConfigOptionProvider.java)s in do just this - the type definition could still be done in JSON, only certain options can by dynamically provided.
I would love to be able to do the same for module types and I think this should be possible; what we would need to do is to take the config descriptions and register them in the ConfigDescriptionRegistry; within our Java objects, we would then simply store the uri (e.g. "module-type:PlayAction") while the DTOs would again assemble the content so that they stay as they are. WDYT?

maggu2810:
> WDYT?

Do you want to wait and perhaps change this for this PR before it is merged?
IMHO we could use it as it is and change it later. WDYT :wink:

kaikreuzer:
Yes, I would also say we can merge, but I wanted to show with this code that I think we can optimize something in the automation framework to make things easier for module-type developers.

danchom:
I'm not sure if I understand you in complete. You mean to remove ConfigDescriptionParameter object from Module  and replace it with URI, and use this URI to get  ConfigDescriptionParameter from ConfigDescriptionRegistry? This is very big change and it will break our JSON API.

kaikreuzer:
> This is big change and it will break our JSON API.

@danchom: No, as I said, the JSON should look exactly the same as before right now.
Please have a look at the Thing framework: The "thing-types" (compare to "module-types") are declared in XML (compare JSON) and [contain the config descriptions in-line](https://github.com/eclipse/smarthome/blob/master/extensions/binding/org.eclipse.smarthome.binding.hue/ESH-INF/thing/ColorLight.xml#L22). In the [internal Java representation](https://github.com/eclipse/smarthome/blob/6dd1df77136a70ec55da6746be05bf65b78ba15b/bundles/core/org.eclipse.smarthome.core.thing/src/main/java/org/eclipse/smarthome/core/thing/type/ThingType.java#L40) (compare to ModuleType), it then only holds an uri to the CD (i.e. the XML parser adds the CD to the ConfigDescriptionRegistry and only puts a reference to it in the ThingType instance). The [ThingTypeDTO](https://github.com/eclipse/smarthome/blob/09050cef9d6f857bd2683e90836f4681f0ad975c/bundles/core/org.eclipse.smarthome.core.thing/src/main/java/org/eclipse/smarthome/core/thing/dto/ThingTypeDTO.java#L29) (used for all external exchange, e.h. through REST, compare to ModuleTypeDTO) again holds the full ConfigDescription.

I checked who is using `ModuleType.getConfigurationDescriptions()` and it seems to me that the refactoring should be easily possible without breaking anything externally. The big benefit would be that it would be fully in sync with the Thing framework and features like the `ConfigOptionProvider` could be used as mentioned above (this was introduced to the Thing framework after feedback from developers - creating ThingTypes through Java code simply would have been ugly in many cases, just as I observed it here in this PR for new module types).

PR #2464, opened 2016-11-14
Fix the voice label produced by the voice manager

lolodomo:

With this fix, the label displayed by Paper UI for the default voice config setting will be like "VoiceRSS français (France)" for the voice attached to locale fr-FR.


kaikreuzer:
Thanks! I manually merged it and added a "dash" between the name and the locale as well: https://github.com/eclipse/smarthome/commit/70ff83d03071508839b164aef1f1011de747dd92

Issue #2454, opened 2016-11-12
Classic UI no more able to open certain frames


lolodomo:
With snapshot 583, I discovered that few of my frames cannot be opened in Classic UI. I got this error:
```
22:57:49.814 [WARN ] [eclipse.jetty.servlet.ServletHandler] - /classicui/app
java.lang.NullPointerException
        at org.eclipse.smarthome.ui.classic.internal.render.AbstractWidgetRenderer.getStateAsNumber(AbstractWidgetRenderer.java:192)[184:org.eclipse.smarthome.ui.classic:0.9.0.201611102032]
        at org.eclipse.smarthome.ui.classic.internal.render.SliderRenderer.renderWidget(SliderRenderer.java:59)[184:org.eclipse.smarthome.ui.classic:0.9.0.201611102032]
        at org.eclipse.smarthome.ui.classic.internal.render.PageRenderer.renderWidget(PageRenderer.java:166)[184:org.eclipse.smarthome.ui.classic:0.9.0.201611102032]
        at org.eclipse.smarthome.ui.classic.internal.render.PageRenderer.processChildren(PageRenderer.java:123)[184:org.eclipse.smarthome.ui.classic:0.9.0.201611102032]
        at org.eclipse.smarthome.ui.classic.internal.render.PageRenderer.processPage(PageRenderer.java:86)[184:org.eclipse.smarthome.ui.classic:0.9.0.201611102032]
        at org.eclipse.smarthome.ui.classic.internal.servlet.WebAppServlet.service(WebAppServlet.java:160)[184:org.eclipse.smarthome.ui.classic:0.9.0.201611102032]
        at org.eclipse.jetty.servlet.ServletHolder.handle(ServletHolder.java:812)[79:org.eclipse.jetty.servlet:9.2.14.v20151106]
        at org.eclipse.jetty.servlet.ServletHandler.doHandle(ServletHandler.java:587)[79:org.eclipse.jetty.servlet:9.2.14.v20151106]
        at org.ops4j.pax.web.service.jetty.internal.HttpServiceServletHandler.doHandle(HttpServiceServletHandler.java:70)[170:org.ops4j.pax.web.pax-web-jetty:4.2.4]
        at org.eclipse.jetty.server.handler.ScopedHandler.handle(ScopedHandler.java:143)[78:org.eclipse.jetty.server:9.2.14.v20151106]
[...]

PR #2473, opened 2016-11-16
avoid npe if state is not a number

kaikreuzer:
Fixes https://github.com/eclipse/smarthome/issues/2454

### DIFF
--- a/extensions/ui/org.eclipse.smarthome.ui.classic/src/main/java/org/eclipse/smarthome/ui/classic/internal/render/AbstractWidgetRenderer.java
+++ b/extensions/ui/org.eclipse.smarthome.ui.classic/src/main/java/org/eclipse/smarthome/ui/classic/internal/render/AbstractWidgetRenderer.java
@@ -188,10 +188,16 @@ abstract public class AbstractWidgetRenderer implements WidgetRenderer {
         if (itemName != null) {
             try {
                 Item item = itemUIRegistry.getItem(itemName);
+                State state = item.getState();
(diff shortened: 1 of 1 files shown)


extensions/ui/org.eclipse.smarthome.ui.classic/src/main/java/org/eclipse/smarthome/ui/classic/internal/render/AbstractWidgetRenderer.java
```diff
@@ -188,10 +188,16 @@ protected String getStateAsNumber(Widget w) {
...
-                    return escapeURLPath(item.getStateAs(PercentType.class).toString());
+                    state = item.getStateAs(PercentType.class);
                 } else {
-                    return escapeURLPath(item.getStateAs(DecimalType.class).toString());
+                    state = item.getStateAs(DecimalType.class);
+                }
+                if (state != null) {
+                    return escapeURLPath(state.toString());
+                } else {
+                    logger.debug("State '{}' of item '{}' is not a number!", item.getState(), itemName);
```
maggu2810: Really? You assign `item.getState()` to the variable `state`, so I assume `state == null` does not mean that it is not a number but that there is no state at all.

kaikreuzer: No, item.getState() is never null, it might only be UnDefType.NULL. If state==null here, this is only because the getStateAs() returned null, since it wasn't possible to convert the state.

Issue #2452, opened 2016-11-12
smarthome:audio play with Sonos: stop after few seconds


lolodomo:
I tried (with snapshot 583) the console command "smarthome:audio play" with one of my Sonos as sink.
The playback starts as expected but then stops suddenly after few seconds.

lolodomo:
As I can see in the current code of the Sonos binding, there is a 10 seconds hardcoded for the audio servlet call. i imagine the difficulty is to determine the duration of a MP3 file from the file ?

kaikreuzer:
Can you link to the code which you mean? No, in general the length of the mp3 should not matter - Sonos should play it back until it is finished and then switch back to the original playlist.

lolodomo:
Line 87 in SonosAudioSink
https://github.com/eclipse/smarthome/blob/master/extensions/binding/org.eclipse.smarthome.binding.sonos/src/main/java/org/eclipse/smarthome/binding/sonos/internal/SonosAudioSink.java#L87

kaikreuzer:
Ah, ok. This parameter is actually meant to define how long a stream can be requested from the http servlet (time of the request). Once it started serving it, it should keep running; if this is not the case, it is indeed a bug (but rather in the AudioServlet than in the Sonos binding).

lolodomo:
Ok, so the problem is elsewhere as the playback starts.
The console is "locked", is it normal ? It should start the job in background and then return ? Maybe there is a timeout in the console ?

kaikreuzer:
Yes, blocking is normal; see also https://github.com/eclipse/smarthome/issues/2333 (1).
For the console it might make sense to launch these commands in a separate thread, though, this isn't currently done.

lolodomo:
Please try with a music MP3 file of 4 minutes for example and you will reproduce the issue.
I just forgot to check the logs.

lolodomo:
I just tried again. I can confirm that there is no error logged. What seems to happen is that the Sonos handler decided the end of the playback while not finished. I have the feeling now that the problem could be in the notification sound mechanism and how it detects end of playback.
I will have to add traces to understand what happens exactly.

lolodomo:
Ok found.
waitForFinishedNotification in the Sonos binding handler is using a timeout of 10 seconds.

lolodomo:
[...]

PR #2476, opened 2016-11-17
Added documentation for audio sink support in Sonos

kaikreuzer:
Fixes https://github.com/eclipse/smarthome/issues/2452

### DIFF
--- a/extensions/binding/org.eclipse.smarthome.binding.sonos/README.md
+++ b/extensions/binding/org.eclipse.smarthome.binding.sonos/README.md
@@ -90,7 +90,7 @@ All supported Sonos devices are registered as an audio sink in the framework.
 Audio streams are treated as notifications, i.e. they are fed into the `notificationsound` channel and changing the volume of the audio sink will change the `notificationvolume`, not the `volume`.
 Note that the Sonos binding has a limit of 20 seconds for notification sounds. Any sound that is longer than that will be cut off.
 
-An exception is done if an URL audio stream is passed (e.g. an Internet radio stream). This will be handled by the `playuri` channel.
+URL audio streams (e.g. an Internet radio stream) are an exception and do not get sent to the `notificationsound` channel. Instead, these will be sent to the `playuri` channel.
 
 ## Full Example
 
--- a/extensions/binding/org.eclipse.smarthome.binding.sonos/README.md
+++ b/extensions/binding/org.eclipse.smarthome.binding.sonos/README.md
@@ -10,7 +10,7 @@ This binding integrates the [Sonos Multi-Room Audio system](http://www.sonos.com
 
 ## Supported Things
 
-All available Sonos (playback) devices are supported by this binding. This includes the Play:1, Play:3, Play:5, Connect, Connect:Amp, Playbar, and Sub. The Bridge and Boost are not supported, but these devices do only have an auxiliary role in the Sonos network and do not have any playback capability.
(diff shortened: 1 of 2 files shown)


extensions/binding/org.eclipse.smarthome.binding.sonos/README.md
```diff
@@ -79,6 +84,14 @@ The devices support the following channels:
 | zonegroupid | String | Id of the Zone Group the Zone Player belongs to | all |
 | zonename | String | Name of the Zone Group the Zone Player belongs to | all |
 
+## Audio Support
+
+All supported Sonos devices are registered as an audio sink in the framework.
+Audio streams are treated as notifications, i.e. they are fed into the `notificationsound` channel and changing the volume of the audio sink will change the `notificationvolume`, not the `volume`.
+Note that the Sonos binding has a limit of 20 seconds for notification sounds. Any sound that is longer than that will be cut off.
+
+An exception is done if an URL audio stream is passed (e.g. an Internet radio stream). This will be handled by the `playuri` channel.
```
tavalin: This paragraph does not read well.  I would replace with:

> An exception is raised if a URL audio stream is passed (e.g. an Internet radio stream).  URL audio streams should be sent to the 'playuri' channel.

extensions/binding/org.eclipse.smarthome.binding.sonos/README.md
```diff
@@ -22,7 +22,12 @@ The Sonos devices are discovered through UPnP in the local network and all devic
 
 ## Binding Configuration
 
-The binding does not require any special configuration
+The binding has the following configuration options, which can be set for "binding:sonos":
+
+| Parameter | Name    | Description  | Required |
+|-----------------|------------------------|--------------|------------ |
+| opmlUrl | OPML Service URL | URL for the OPML/tunein.com service | no |
+| callbackUrl | Callback URL | URL to use for playing notification sounds, e.g. http://192.168.0.2:8080 | no |
```
tavalin: I've been having a think about the callbackUrl for audiosinks.  What if I have two similar devices/Things and I want to specify on-the-fly which device plays the audio? What are you thoughts on having the callbackUrl as a Thing config rather than a binding config?

kaikreuzer: This callbackURL specifies the url where the runtime can be reached, not the speaker.
It is mainly required if you have multiple interfaces in your host and thus your external IP cannot be easily determined (what is tried if you leave this value empty). In 99% of all setups, I would claim that all Sonos speakers should be fine with the same url for reaching the runtime.

extensions/binding/org.eclipse.smarthome.binding.sonos/README.md
tavalin: Oh, I see. In that case I still feel it needs rewording as that's ambigious and open to people (like me!) misinterpreting it!

tavalin: Is the situation simply that file system audio streams get sent as a Sonos notification whereas url audio streams get sent down the playuri channel?

kaikreuzer: Yes, right.

tavalin: Maybe something like:

> URL audio streams (e.g. an Internet radio stream) are an exception and do not get sent to the `notificationsound` channel. Instead, these will be sent to the `playuri` channel.

kaikreuzer: Ok, thanks - I have updated the PR accordingly!

kaikreuzer:
Any further comments or ok to merge?

maggu2810:
> Any further comments 

Not from my side

> ok to merge?

Ok for me

PR #2479, opened 2016-11-17
fixed NPE in StringType.equals()

sjka:


maggu2810:
****... I reviewed it already yesterday, but didn't press on finish, so this comment was still in "pending" phase.

sjka:
thanks for the feedback anyway! I created #2507 accordingly.

PR #2480, opened 2016-11-17
Improvements in startup behavior for cron rules

kaikreuzer:
* delay cron job execution until after startup rules have been processed
* only register the rule engine as a listener on other services in activate(), which avoids concurrent execution of addRule()
* if a job already exists for whatever reason, simply replace it instead of failing with an error
* cleaned some javadoc and logging statements

See issues reported at https://community.openhab.org/t/startup-error-for-all-cron-based-rules/14405


bundles/model/org.eclipse.smarthome.model.rule.runtime/src/org/eclipse/smarthome/model/rule/runtime/internal/engine/RuleTriggerManager.java
```diff
@@ -538,15 +546,17 @@ private void removeTimerRule(Rule rule) {
      * @throws SchedulerException if there is an internal Scheduler error.
      */
     private void createTimer(Rule rule, TimerTrigger trigger) throws SchedulerException {
+        synchronized (rule) {
+
```
maggu2810: What is the reason of this empty synchronized block?

kaikreuzer: No reason, that was an accident - will remove it!

kaikreuzer: removed.

Issue #2483, opened 2016-11-17
Connected Lighting Binding using the CoAP protocol - Microchip Technology


gauranggunde:
The CoAP Connected Lighting binding provided by Microchip Technology allows for connectivity with the Lighting setup based on PIC micro-controllers using the Constrained Application (CoAP) protocol. A user would be able to interact with the lighting setup using OpenHAB as the UI through the provided binding.

triller-telekom:
We have californium in our target platform ready to be used. I have also already mentioned it in [your PR](https://github.com/eclipse/smarthome/pull/2936).

@sjka @kaikreuzer This issue can be closed.

kaikreuzer:
@triller-telekom How is that related to #2936? Why should the issue be closed without the PR not being merged (yet)?

triller-telekom:
I read this issue as a request that the coap protocol should be made available in the framework. And it is available now...

gauranggunde:
@triller-telekom This binding uses the Californium library and does special things to communicate with the PIC microcontrollers. It is not "just" enabling the CoAP protocol.

PR #2484, opened 2016-11-17
Connected Lighting Binding using CoAP - Microchip Technology

gauranggunde:
Tagging Issue: #2483

The binding enables the user to interact with the PIC micro-controller based lights and sensors using the Constrained Application (CoAP) protocol.


maggu2810:
Replaced by #2936

PR #2487, opened 2016-11-18
Enforce that handler of a bridge implements BridgeHandler interface.

sbussweiler:
Signed-off-by: Stefan Bussweiler <s.bussweiler@external.telekom.de>


bundles/core/org.eclipse.smarthome.core.thing/src/main/java/org/eclipse/smarthome/core/thing/internal/BridgeImpl.java
```diff
@@ -66,7 +69,13 @@ public void removeThing(Thing thing) {
...
     @Override
     public BridgeHandler getHandler() {
-        return super.getHandler() instanceof BridgeHandler ? (BridgeHandler) super.getHandler() : null;
+        BridgeHandler bridgeHandler = null;
+        if (super.getHandler() instanceof BridgeHandler) {
+            bridgeHandler = (BridgeHandler) super.getHandler();
+        } else if (super.getHandler() != null) {
+            logger.warn("Handler of bridge '{}' must implement BridgeHandler interface.", getUID());
+        }
+        return bridgeHandler;
```
maggu2810: Could you change your code so that super.getHandler() is called only once?

PR #2492, opened 2016-11-18
Sonos binding: fix notification sound when line-in is selected (fix #2474)

### LINKED ISSUE #2474: Sonos notification sound: not working when line-in is playing
Playing with the new TTS feature, I tested the Sonos notification mechanism with a lot of different sources being played and with the queue being filled or not.
It is working very well except in one case, when the current source is the line-in. The behavior is even different, depending on whether the queue is empty or not. In one case, the TTS audio is rendered but the line-in is then not restored. In the other case, the TTS audio is just not rendered at all.

lolodomo:
Define constants for few orefix URI


lolodomo:
Please wait and don't merge.
The title channel is no more updated for tune-in radio, I need first to check if this is relative to this fix or totally independent on it.

lolodomo:
The problem I noticed with title not retrieved for tune-in seems to be relative to the recent suppress of setting opmlurl from the distribution. So noting link to this fix.

Please review (and merge) this fix.

lolodomo:
@kgoderis @R-Edelhoff : plase review this little fix.

R-Edelhoff:
The refactoring is fine. 
If the additional functional change for "handleLineIn" is working like expected (what i assume)
this extension wihtih the PR looks good to me.

PR #2497, opened 2016-11-19
Interpret console command: fix #2493 #2494

### LINKED ISSUE #2493: Interpret console command with system interpreter displays null
My default HLI is set to "Rule-based interpreter".
When I run the command "interpret" in the console, I get just a "null" as result.
```
openhab> smarthome:voice interpret here is a test
null
```
Of course I have an item named VoiceCommand.
I tried to set my locale in Paper UI but no change.

### LINKED ISSUE #2494: Used locale by interpret console command
The "interpret" command considers only the default locale of your machine. Is it expected ?
I would have expected that it used the locale settings of openHAB if defined.
In my case for example, no chance to try with english as my machine is setup to a French locale.

An alternative would be to have the locale as a first optional argument of the command.
Any suggestion ?

lolodomo:
Update the interpret method prototype in the voice manager

### DIFF
--- a/bundles/core/org.eclipse.smarthome.core.voice/src/main/java/org/eclipse/smarthome/core/voice/VoiceManager.java
+++ b/bundles/core/org.eclipse.smarthome.core.voice/src/main/java/org/eclipse/smarthome/core/voice/VoiceManager.java
@@ -52,12 +52,23 @@ public interface VoiceManager {
     void say(String text, String voiceId, String sinkId);
 
     /**
-     * Interprets the passed string using the default services for HLI locale.
+     * Interprets the passed string using the default services for HLI and locale.
      *
      * @param text The text to interpret
      * @throws InterpretationException
+     * @return a human language response
      */
-    void interpret(String text) throws InterpretationException;
+    String interpret(String text) throws InterpretationException;
+
+    /**
+     * Interprets the passed string using a particular HLI service and the default locale.
+     *
+     * @param text The text to interpret
+     * @param hliId The id of the HLI service to use
+     * @throws InterpretationException
+     * @return a human language response
+     */
+    String interpret(String text, String hliId) throws InterpretationException;
 
     /**
      * Determines the preferred voice for the currently set locale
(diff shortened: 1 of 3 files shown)


kaikreuzer:
With this change, we do not need line 87 & 88 anymore, right?

lolodomo:
Yes you're right but if we suppress the null check here, it will be safer to add it in voicemanager.interpret.
I will update this PR to suppress the null check and include the null check in voicemanager.interpret in a next PR adding other little enhancement to voicemanager.interpret (having the hli id as argument).

kaikreuzer:
You are right, the null check is missing in voicemanager.interpret - I would prefer if you could simply update this PR accordingly (feel free to add further improvements).

lolodomo:
Ok

lolodomo:
Ready for review.

lolodomo:
Please wait again, I will fix the ussue #2493 in the same PR.

lolodomo:
That(s time, it is really ok and ready for review and merge.

lolodomo:
@kaikreuzer : tested today the snapshot 607 thinking this PR will be included. I already knew that Eclipse smarthome stuff is not automatically updated every bight. Is there a way for us to know what is the last PR considered from this repo in any nightly build of openHAB ?
Why not considering in the nightly build of openHAB the up-to-date version from Eclipse smarthome ?

kaikreuzer:
> Is there a way for us to know what is the last PR considered from this repo in any nightly build of openHAB ?

Yes, here: https://hudson.eclipse.org/smarthome/job/SmartHomeDistribution-Stable/changes

> Why not considering in the nightly build of openHAB the up-to-date version from Eclipse smarthome ?

Because that would continuously break the openHAB distribution.

maggu2810:
> Is there a way for us to know what is the last PR considered from this repo in any nightly build of openHAB ?

https://github.com/eclipse/smarthome/wiki/ESH-used-in-openHAB

lolodomo:
Thanks guys

kaikreuzer:
Thanks, @maggu2810, good idea 👍

PR #2498, opened 2016-11-19
don't add non-transient logger field

maggu2810:
Fixes: https://github.com/eclipse/smarthome/issues/2495

### DIFF
--- a/bundles/core/org.eclipse.smarthome.core.thing/src/main/java/org/eclipse/smarthome/core/thing/internal/BridgeImpl.java
+++ b/bundles/core/org.eclipse.smarthome.core.thing/src/main/java/org/eclipse/smarthome/core/thing/internal/BridgeImpl.java
@@ -24,7 +24,7 @@ import com.google.common.collect.ImmutableList;
 public class BridgeImpl extends ThingImpl implements Bridge {
 
     private transient List<Thing> things = new CopyOnWriteArrayList<>();
-    private final Logger logger = LoggerFactory.getLogger(BridgeImpl.class);
+    private transient Logger logger = LoggerFactory.getLogger(BridgeImpl.class);
 
     /**
      * Package protected default constructor to allow reflective instantiation.


bundles/core/org.eclipse.smarthome.core.thing/src/main/java/org/eclipse/smarthome/core/thing/internal/BridgeImpl.java
```diff
@@ -24,7 +24,10 @@
 public class BridgeImpl extends ThingImpl implements Bridge {
 
     private transient List<Thing> things = new CopyOnWriteArrayList<>();
-    private final Logger logger = LoggerFactory.getLogger(BridgeImpl.class);
+
+    private static Logger logger() {
```
kaikreuzer: Just for my understanding: What would be the problem to simply add the transient keyword to the existing logger declaration?

maggu2810: IIRC Gson is using sun.misc.Unsafe to create object instances on de-serialization. I don't know if if there is any specification how this function -- that intention is to be used only by the JVM itself and not by external stuff -- initialize the fields using fields initializers. Or at least how the Unsafe object creation works at all.
That's the reason I don't use the transient keyword, because I know this method will work all the time. On the other one, I can just try if it is working currently on my machine, but I don't know about the spec.

kaikreuzer: It should be safe to use transient, see http://stackoverflow.com/a/5889590.
Note that the very same thing is done for the "things" field in the line above as well.

maggu2810:
FYI:
We already added no-args constructors for some classes that are (de)serialized using Gson.
I tested it now in combination with the transient flag.
If the class do not own a no-arg constructor that could be called on de-serialization (I assume caused by the usage of Unsafe if no no-arg constructor is found) the transient fields are not initialized.
So, we should never miss the no-arg constructor for that classes at all!

maggu2810:
FYI:
We already added no-args constructors for some classes that are (de)serialized using Gson.
I tested it now in combination with the transient flag.
If the class do not own a no-arg constructor that could be called on de-serialization (I assume caused by the usage of Unsafe if no no-arg constructor is found) the transient fields are not initialized.
So, we should never miss the no-arg constructor for that classes at all!

If you are interested in an example that shows the difference, you could have a look at:
https://github.com/maggu2810/gson-noarg-constructor

kaikreuzer:
Many thanks for your analysis and providing the fix!

PR #2500, opened 2016-11-21
Deletion of the AbstractActiveService class

m-memova:
The org.eclipse.smarthome.core.service.AbstractActiveService class is deleted since it is a remainder from the old OH1 binding APIs.


kaikreuzer:
Thank you!

PR #2504, opened 2016-11-21
fix state conversion between HSBType and DecimalType

### LINKED ISSUE #2471: Hue group issue since latest snapshot
Hi,

I have upgraded to latest snapshot build from today. Now the Hue groups are not working anymore. The old release from 11.11.16 was working fine. Has something changed since than?

Following error in the log when I turn off and on an light of that group:

```
[pse.smarthome.core.items.GenericItem] - failed notifying listener '[gHueLights_Wohnzimmer_Color (Type=GroupItem, BaseType=ColorItem, Members=8, State=null, Label=Wohnzimmer, Category=colorwheel), Licht_Farbe_WZ_TV_Rack_Rechts, null]' about state update of item java.lang.NullPointerException: {}
```

Also the Paper UI control tab Show internal error 500. I am also not able anymore to switch an Hue Color Group on and off. As suggested by Kai I tried to remove the (:Color:Max) from my goup but still the same behaviour. If I remove the groups from the item definition than all is working correctly also the paper ui control tab is working and showing all things with values again.

item config:
```
Group:Color:MAX gHueLights_Alle_Color "Alle Lampen" 

Color Licht_Farbe_WZ_TV_Rack_Oben       "TV-Rack Oben"       <colorwheel> (gHueLights_Alle_Color)     [ "Lighting" ] { channel="hue:0200:00178816e03f:1:color" }
Color Licht_Farbe_WZ_TV_Rack_Links      "TV-Rack Links"      <colorwheel> (gHueLights_Alle_Color)     [ "Lighting" ] { channel="hue:0200:00178816e03f:2:color" }
```

sitemap:
```
Colorpicker item=gHueLights_Alle_Color label="Alle" 
```
Regards René

sjka:
It looks like #2471 was caused by the missing conversion from DecimalType to HSBType. It got lost in #2334 because it was very well hidden in the super classes (i.e. PercentType/DimmerItem).

fixes #2471

### DIFF
--- a/bundles/core/org.eclipse.smarthome.core/src/main/java/org/eclipse/smarthome/core/library/types/DecimalType.java
+++ b/bundles/core/org.eclipse.smarthome.core/src/main/java/org/eclipse/smarthome/core/library/types/DecimalType.java
@@ -165,6 +165,9 @@ public class DecimalType extends Number implements PrimitiveType, State, Command
             } else {
                 return UnDefType.UNDEF;
             }
+        } else if (target == HSBType.class) {
+            return new HSBType(DecimalType.ZERO, PercentType.ZERO,
+                    new PercentType(this.toBigDecimal().multiply(new BigDecimal(100))));
         } else {
             return StateConverterUtil.defaultConversion(this, target);
         }
--- a/bundles/core/org.eclipse.smarthome.core/src/main/java/org/eclipse/smarthome/core/library/types/DecimalType.java
+++ b/bundles/core/org.eclipse.smarthome.core/src/main/java/org/eclipse/smarthome/core/library/types/DecimalType.java
@@ -41,7 +41,7 @@ public class DecimalType extends Number implements PrimitiveType, State, Command
     }
 
     public DecimalType(long value) {
-        this.value = new BigDecimal(value);
+        this.value = BigDecimal.valueOf(value);
     }
 
     public DecimalType(double value) {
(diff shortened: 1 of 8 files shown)


bundles/core/org.eclipse.smarthome.core/src/main/java/org/eclipse/smarthome/core/library/types/DecimalType.java
```diff
@@ -165,6 +165,9 @@ public State as(Class<? extends State> target) {
             } else {
                 return UnDefType.UNDEF;
             }
+        } else if (target == HSBType.class) {
+            return new HSBType(DecimalType.ZERO, PercentType.ZERO,
+                    new PercentType(this.toBigDecimal().multiply(new BigDecimal(100))));
```
watou: `BigDecimal.valueOf(100)` is preferred to `new BigDecimal(100)`:  "This 'static factory method' is provided in preference to a (long) constructor because it allows for reuse of frequently used BigDecimal values." ([link](https://docs.oracle.com/javase/7/docs/api/java/math/BigDecimal.html#valueOf(long)))

maggu2810: @watou Thanks for that hint, didn't know that. But this valueOf talks about the constructor using a long. Does this argument also fit to ints (I need to look at the code why the static method is better then the constructor.

watou: [This code](http://hg.openjdk.java.net/jdk8/jdk8/jdk/file/687fd7c7986d/src/share/classes/java/math/BigDecimal.java#l1207) shows that any ints between 0 and 10, inclusive, are better to use the `valueOf(long)` static factory method to avoid unnecessary construction of new BigDecimals.  But remaining ignorant of the current implementation of `valueOf` and just going by the JavaDoc's advice seems sensible, should later/other implementations ever cache other values outside that range based on frequent usage (however unlikely we think that might be).  I think there is no `valueOf(int)` static factory method because the `int` is just widened to a `long` without side effects.

watou: An alternative approach to minimizing the number of duplicate immutable objects in the heap, but with more certainty of outcome, would be to define and use a static member HUNDRED like
```
private static final BigDecimal HUNDRED = BigDecimal.valueOf(100);
```

kaikreuzer:
Many thanks @sjka!

rene54321:
Thanks a lot for fixing the failure. If I update now to the latest snaphot is it than implemented allready?

kaikreuzer:
No, you have to wait for this fix to be part of the next ESH stable build, see https://hudson.eclipse.org/smarthome/job/SmartHomeDistribution-Stable/changes.
All OH distros that are built AFTER that will contain the fix.

Issue #2501, opened 2016-11-21
New rule action "interpret" with a given HLI as parameter


lolodomo:
I will add a new action "interpret" that will take a HLI as second parameter.

@kaikreuzer : when I look at the class org.eclipse.smarthome.model.script.actions.Voice, I see that sometimes the parameter "text" (say and interpret actions) is of type Object, sometumes of type String.
Is there a valid reason to not have always the same type ? Should we use type Object or String ?

lolodomo:
Implemented and merged

PR #2505, opened 2016-11-21
New rule action: interpret with HLI parameter (enh #2501)

lolodomo:

I recommend a serious review because the change was not tested (don't know exactly how to test it).
At least, everything is compiling and the change is small.

### DIFF
--- a/bundles/model/org.eclipse.smarthome.model.script/src/org/eclipse/smarthome/model/script/actions/Voice.java
+++ b/bundles/model/org.eclipse.smarthome.model.script/src/org/eclipse/smarthome/model/script/actions/Voice.java
@@ -24,6 +24,8 @@ public class Voice {
     /**
      * Says the given text.
      *
+     * This method uses the default voice and the default audio sink to play the audio.
+     *
      * @param text the text to speak
      */
     @ActionDoc(text = "says a given text with the default voice")
@@ -33,6 +35,8 @@ public class Voice {
 
     /**
      * Says the given text with a given voice.
+     *
+     * This method uses the default audio sink to play the audio.
      *
      * @param text the text to speak
      * @param voice the name of the voice to use or null, if the default voice should be used. If the voiceId is fully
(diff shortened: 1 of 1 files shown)


lolodomo:
@kaikreuzer : without considering this patch, I tried "interpret" in my rule file:
```
rule "Controle par la voix"
when
    Item VoiceCommand received command
then
		interpret(VoiceCommand.state.toString)
end
```
but it leads to this error:
```
23:21:21.366 [ERROR] [.script.engine.ScriptExecutionThread] - Rule 'Controle par la voix': org.eclipse.smarthome.core.voice.VoiceManager.interpret(Ljava/lang/String;)V
```
Could it be due to the fact that the interpret method is delcared with Object as parameter and not String ?
Shall we replace Object by String ?

watou:
> interpret(VoiceCommand.state.toString)

Wouldn't you rather

> interpret(receivedCommand.toString)

lolodomo:
@watou: no change when I use "receivedCommand".
The error is not "talking" but my feeling is that it simply does not find the interpret method.

watou:
I see; I just jumped at commenting on a construct I saw giving someone problems on the forum.  

Is it possible to cast the String argument `as Object` in the scripting language?

lolodomo:
I tried this but it failed too:
```
		var String msg = VoiceCommand.state.toString
		interpret(msg)
```
@kaikreuzer : please comment.

kaikreuzer:
> Could it be due to the fact that the interpret method is delcared with Object as parameter and not String ?

No, it seems to correctly resolve and call the Voice.interpret method, because the error shows that it is something within VoiceManager.interpret that goes wrong. Did you debug to check if some exception is thrown some where within it?

lolodomo:
No, that was with my production environment on RPI.

kaikreuzer:
Then this sounds to me as if you might not have a matching core.voice bundle installed.

lolodomo:
I have of course a core voice bundle loaded (through a new jar file corresponding to the current enhancement), and only one is loaded.
The voice manager is working as the interpret command in console is working as expected and the say action in rule is working well too.

lolodomo:
What jar file has to be deployed to have a rule engine considering my code change in voice action?

lolodomo:
@watou : I tried again with the snapshot 627 and the interpret command in my rule is now working well .
My problem was probably due to the fact that I was loading the voice manager through an additional jar file in my directory addons.
@kaikreuzer : could you review my PR please ?

lolodomo:
I could even enhance this enhancement with what I propose in #2588. Let me know.

kaikreuzer:
#2588 can be treated separately.

PR #2507, opened 2016-11-22
simplified value comparison in StringType

sjka:
...according to the late feedback in #2479
(thanks btw!)


maggu2810:
😉

Issue #2485, opened 2016-11-18
Sonos notification sound: restore a wrong volume level


lolodomo:
When requesting the playback of a sound to a group member that is not the coordinator of the group, the restored volume after notification is the original volume of the group coordinator for all group members (rather than the individual volume for each Sonos device).

lolodomo:
In fact, after more experiments, it looks like the volume is set only for the adressed sonos device and after notification its volume is restored not to its original volume but to the volume of the group coordinator.
Should be easy to fix, I will take a look.

lolodomo:
@R-Edelhoff : what was exactly your intention regarding volume when requesting the playback of a notification ?
Was your intention to set a volume value for all the group members (of the current group) or only the adressed device ?
Finally, of course I imagine your intention was to restore the original volume values for all impacted Sonos devices.

R-Edelhoff:
The behaviour is dependent of the type of group. In case of an Ad-hoc group the notification volume should be applied to the selected sonos speaker. After the notification the previous volume of the  selected speaker should be applied.

In case of a bonded group (stereo pair, 5.1) the volume of the coordinator should be applied after notification.

As i can see it in the code the coordinator volume is restored in every case which is not correct.
In case of an AdHoc group the originalVolume of the selected player should be restored. Taking the volume of the "original" selected ThingHander (originalVolume = getVolume()) should be the correct way. This should also be correct for bonded groups as they have anyway the same volume.

lolodomo:
As you seem to have a clear idea of what has to be changed, can you propose a fix ?

R-Edelhoff:
i could arrange a fix but this would require some time in consideration with testing etc.

kaikreuzer:
Fixed through #2516

PR #2516, opened 2016-11-22
#2485 Sonos notification sound: restore the same volume …

ivivanov-bg:
… for all group members

When saving the current volume - retrieve the data from current player handler and not from the coordinator


lolodomo:
Make sense...
I will test it and give a feedback soon.

PS: what is an adhoc group ? Is it a standard group of several Sonos devices ?

lolodomo:
LGTM

Tested and working well.
I cannot test with Sonos pairs but it is working well with standalone Sonos and group of Sonos.

ivivanov-bg:
'adhoc' group is a normal group of sonos devices, which is not a Stereo Pair or 5.1 (the latter ones are 'bonded' group)

kaikreuzer:
Thanks!

Issue #2127, opened 2016-09-06
Thing Status enhancements


maggu2810:
Does it make sense to allow the control of things that status is not online?

kaikreuzer:
Definitely YES! Because OFFLINE might be an outdated status and it is possible to reach them.
Why did you enter this issue? Is is currently NOT possible to control them?

maggu2810:
> Because OFFLINE might be an outdated status and it is possible to reach them.

What do you mean with outdated status? The Paper UI is using SSE to keep track of the events and already adds / removes stuff in certain views if some information changes. Or do you mean a thing handler does not set the status correctly?

> Why did you enter this issue?

Hm, because I don't see any reasons to be able to control actuators if the handler states that he cannot talk to the devices.

kaikreuzer:
> Or do you mean a thing handler does not set the status correctly?

I mean you never know whether your device is reachable or not if you use radio. In many cases, the ONLINE/OFFLINE status is only set by a handler, if it (un)successfully sent a command to the device over radio. successful->ONLINE, otherwise->OFFLINE. But if you send the next command 5 minutes later, you have no clue whether you can reach it or not. So you have to try it out.
And no, especially on 868MHz with its duty cycle, you do not do a "ping" over air every 10 seconds...

maggu2810:
If a thing handler don't know if a device is online or offline this is a pain but is in responsible of the handler and there needs to be found a way to handle that. But I don't think that the user should see a offline device that could be controlled.

maggu2810:
Wouldn't make another state sense, "unknown (offline or online)" and keep an offline to mark devices offline?

kaikreuzer:
Think of the label "OFFLINE" as "failed communicating with it 10 minutes ago". The status can NEVER be up-to-date, all it tells you is something about the past. Believe me, I had many such discussions in the past already, all with the same result...

maggu2810:
I still need to think about, but at the moment I would still prefer if the thing handler could then use two different offline states (offline_weak, offline_strict, or something similar).
The weak variant could be used if no real online / offline could be detected.
The weak strict could be used where the handler really knows that the device is not reachable.
For example all devices that are accesses directly by an USB connection could be set to offline_strict if the USB connection is not available. We already knows that also a command will not make the device online.
If the handler access the network information of your OS with events the handler is aware that there is currently no connection available and the devices could not be reached at all.
[...]

PR #2022, opened 2016-08-20
add the isInitialized function to the thing itself

maggu2810:
The BaseThingHandler, the ThingManager and the ThingLinkManager contains a method to check if a thing has been initialized.
I moved that logic to a function of the thing itself, so there is only one implementation.

I didn't remove the old (protected) method of the BaseThingHandler but marked it deprecated. So bindings using this method have some time to change their implementation.

PR #2524, opened 2016-11-23
Sonos audio sink: check thing status first

lolodomo:


maggu2810:
Hm, IIRC @kaikreuzer stated that a thing status is always outdated and a thing could be marked offline but gets online if some interaction is done. So with that background I don't think we should skip the play mechanism.
If I found the resp. comment, I will linkt it.

maggu2810:
https://github.com/eclipse/smarthome/issues/2127

lolodomo:
@maggu2810 @kaikreuzer  : The current code is checking the thing status for MP3 audio format and not for WAV audio format.
My fix consists in applying this same check whatever the audio format.
Checking the thing status or not is another question; we could decide to suppress this check.

maggu2810:
@lolodomo You are right. It is not consistent. Missed it.
The question is, if we should introduce the check for WAV or remove it for MP3.

lolodomo:
I let you two decide and then I will adjust my PR if the decision is to remove the check.

kaikreuzer:
Afair, I added the check to make sure that the handler is initialized (because it otherwise might not expect calls to `playNotificationSoundURI()`.
So the correct check would actually be `handler.getThing().getStatus() == ThingStatus.ONLINE || handler.getThing().getStatus() == ThingStatus.OFFLINE`.

lolodomo:
Test updated.

maggu2810:
@kaikreuzer Perhaps I should update #2022 so it could be used here...

kaikreuzer:
@maggu2810 Yes, this would be helpful!

kaikreuzer:
@lolodomo https://github.com/eclipse/smarthome/pull/2022 is merged, so you could change your code to use the isInitialized() method now to make it slightly simpler.

lolodomo:
I can use ThingHandlerHelper.isHandlerInitialized(handler) but I cannot use handler.isInitialized() because isInitialized() is protected.
Is it ok if I use ThingHandlerHelper.isHandlerInitialized(handler)  ?

kaikreuzer:
Right, I forgot that we are not within a handler - yes, your solution is perfect this way, thanks!

Issue #2490, opened 2016-11-18
System interpreter: adding french syntax


lolodomo:
I am interested in adding french syntax to the system rule interpreter.
Will work on it and propose a PR when done.

kaikreuzer:
Nice :-)

lolodomo:
In french, the equivalent of "the" can be "le", "la" or "l'" like for example "le machin", "la chose" or "l'exemple".
Considering that my item has "exemple" for label, will the current code be able to recognize "l'exemple" as being "l'" + my item label ?

kaikreuzer:
No idea - have a try yourself. I fear that it won't work as expected - you might need to adapt the tokenizer to also treat ' as a symbol where words need to be separated.

lolodomo:
I took a look to the method tokenize and I discovered that all ' are first suppressed. Hopefully, it was well thought ;) and we have locale as parameter for this method and I can update the method in case of franch.
I think I should replace any ' by a space so that a text like "faire l'action" will become "faire l action".

lolodomo:
@kaikreuzer : we have the console command "smarthome:voice interpret" but is there an available call to the interpreter we can use in rules ?

kaikreuzer:
Yes, right here: https://github.com/eclipse/smarthome/blob/master/bundles/model/org.eclipse.smarthome.model.script/src/org/eclipse/smarthome/model/script/actions/Voice.java#L73

lolodomo:
Now implemented and merged.

PR #2525, opened 2016-11-23
Add French syntax for system interpreter (enh #2490)

### LINKED ISSUE #2529: System HLI: not accepted command type
The system HLI returns a wrong message when you request a command to an item and the command type is not accepted by the matched item type. The error tells you that no match was found with this label while it should indicate no match with this label and command type.
I would suggest to update the message of the property "no_objects" for all languages to include the command type.

lolodomo:
New returned error message when command type is not accepted (fix #2529)

Suppress OPEN and CLOSE commands from syntax (all languages)


lolodomo:
I will update my PR to suppress OPEN and CLOSE commands from the syntax in all languages.

lolodomo:
... and I will add the implementation of enhancement #2529.
So this PR is finally not yet ready for review/merge.

lolodomo:
Now ready for review.

lolodomo:
@kaikreuzer  :  no feedback  ?

bundles/core/org.eclipse.smarthome.core.voice/src/main/java/org/eclipse/smarthome/core/voice/text/AbstractRuleBasedInterpreter.java
```diff
@@ -479,7 +480,12 @@ protected String executeSingle(ResourceBundle language, String[] labelFragments,
             throws InterpretationException {
         ArrayList<Item> items = getMatchingItems(language, labelFragments, command.getClass());
         if (items.size() < 1) {
-            throw new InterpretationException(language.getString(NO_OBJECTS));
+            if ((command != null) && (getMatchingItems(language, labelFragments, null).size() >= 1)) {
```
maggu2810: Above `command.getClass()` is called, so if command is null, you don't reach this line at all.

lolodomo: In practice, I think command is never null. But I will add a test before command.getClass(( in case it would be.

lolodomo:
@maggu2810 : thank you for your review, I have now taken into account your comment and adjusted my code accordingly.

PR #2526, opened 2016-11-24
no need for polling nupnp in the background

BenediktNiehues:

### DIFF
--- a/extensions/binding/org.eclipse.smarthome.binding.hue/src/main/java/org/eclipse/smarthome/binding/hue/internal/discovery/HueBridgeNupnpDiscovery.java
+++ b/extensions/binding/org.eclipse.smarthome.binding.hue/src/main/java/org/eclipse/smarthome/binding/hue/internal/discovery/HueBridgeNupnpDiscovery.java
@@ -58,17 +58,7 @@ public class HueBridgeNupnpDiscovery extends AbstractDiscoveryService {
     private final Logger logger = LoggerFactory.getLogger(HueBridgeNupnpDiscovery.class);
 
     public HueBridgeNupnpDiscovery() {
-        super(SUPPORTED_THING_TYPES, DISCOVERY_TIMEOUT);
-    }
-
-    @Override
-    protected void startBackgroundDiscovery() {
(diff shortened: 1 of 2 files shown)


maggu2810:
Any reason to override the `startBackgroundDiscovery` and `stopBackgroundDiscovery` at all? The implementation of the super class already "does nothing".

BenediktNiehues:
@maggu2810 just wanted to leave the comments there
and as you can see, travis passed for this branch...

extensions/binding/org.eclipse.smarthome.binding.hue/src/main/java/org/eclipse/smarthome/binding/hue/internal/discovery/HueBridgeNupnpDiscovery.java
```diff
@@ -57,34 +55,20 @@
...
     private static final Set<ThingTypeUID> SUPPORTED_THING_TYPES = Collections.singleton(THING_TYPE_BRIDGE);
 
-    private static final int BRIDGE_SCAN_INTERVAL = 60;
-
     private final Logger logger = LoggerFactory.getLogger(HueBridgeNupnpDiscovery.class);
 
-    private ScheduledFuture<?> bridgeScanningJob;
-
     public HueBridgeNupnpDiscovery() {
         super(SUPPORTED_THING_TYPES, DISCOVERY_TIMEOUT);
```
kaikreuzer: you should actually call super with `false` as a third parameter.

kaikreuzer:
I agree with @maggu2810 that the methods should not be there at all, if the service does not support background discovery.

extensions/binding/org.eclipse.smarthome.binding.hue/src/main/java/org/eclipse/smarthome/binding/hue/internal/discovery/HueBridgeNupnpDiscovery.java
BenediktNiehues: good hint, thanks!

PR #2532, opened 2016-11-24
Fixes of some bugs found by the static analysis tool.

m-memova:
I've tried the static code analysis tool (https://github.com/eclipse/smarthome/pull/2440). 
Since my local build failed several times due to some bugs found by the tool, I decided to fix them.
The fixes are mainly related to closing opened streams and comparing Strings with the equals() method instead of the '==' operator.
In addition, I think that the UNUSED_PUBLIC_OR_PROTECTED_FIELD bug should have a lower  priority (it is defined in the rulesets\findbugs directory)

### DIFF
--- a/extensions/voice/org.eclipse.smarthome.voice.mactts/src/main/java/org/eclipse/smarthome/voice/mactts/internal/MacTTSService.java
+++ b/extensions/voice/org.eclipse.smarthome.voice.mactts/src/main/java/org/eclipse/smarthome/voice/mactts/internal/MacTTSService.java
@@ -15,6 +15,7 @@ import java.util.HashSet;
 import java.util.Locale;
 import java.util.Set;
 
+import org.apache.commons.io.IOUtils;
 import org.eclipse.smarthome.core.audio.AudioException;
 import org.eclipse.smarthome.core.audio.AudioFormat;
 import org.eclipse.smarthome.core.audio.AudioStream;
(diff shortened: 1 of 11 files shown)


extensions/voice/org.eclipse.smarthome.voice.mactts/src/main/java/org/eclipse/smarthome/voice/mactts/internal/MacTTSService.java
```diff
@@ -89,17 +90,21 @@ public AudioStream synthesize(String text, Voice voice, AudioFormat requestedFor
      */
     private final Set<Voice> initVoices() {
         Set<Voice> voices = new HashSet<Voice>();
+        InputStreamReader inputStreamReader = null;
+        BufferedReader bufferedReader = null;
         try {
             Process process = Runtime.getRuntime().exec("say -v ?");
-            InputStreamReader inputStreamReader = new InputStreamReader(process.getInputStream());
-            BufferedReader bufferedReader = new BufferedReader(inputStreamReader);
+            inputStreamReader = new InputStreamReader(process.getInputStream());
```
maggu2810: Why not using try-with-resources here, too?

m-memova: Because in this case I found it easier to do it with a finally block. The ` Process process = Runtime.getRuntime().exec("say -v ?");` line should be surrounded with  try-catch but it is not a closeable object, so it cannot be put in the try-with-recources. So if I was not using the finally block, I would have to put one try-catch block in addition to the try-with-recources block.

maggu2810: > I would have to put one try-catch block in addition to the try-with-recources block.

That's correct. And I don't see anything against the encapsulation of try blocks.
But if you prefer using the finally block here and you are sure that there is no corner case we need to close the inputStreamReader, too (don't know if the BufferedReader constructor could throw a runtime excpection)...

PR #1106, opened 2016-02-29
ClassicUI: full image and chart refresh

watou:
Support the `refresh=` parameter to Image and Chart widgets in all cases in the classic web UI.  Testing feedback should go in [this topic](https://community.openhab.org/t/please-help-test-fix-for-image-and-chart-refresh-in-classic-web-ui/8106).  If wider testing deems this change fit, I will add a comment.

Same change as in https://github.com/openhab/openhab/pull/4117.

PR #2554, opened 2016-11-28
Fix #2544 Wrong chart URLs in Basic UI when using non-default persistence service

### LINKED ISSUE #2544: Wrong chart URLs in Basic UI when using non-default persistence service
I have an outside temperature chart with several periods similar to that of the demo. The only difference is that it uses JDBC as persistence service. I.e. this is how the sitemap looks:

```
Frame {
	Switch item=Outside_Temp_Chart_Period label="Chart period" mappings=[0="Hour", 1="Day", 2="Week", 3="Month", 4="Year"]
	Chart item=Outside_Temp_Chart service="jdbc" period=h refresh=600 visibility=[Outside_Temp_Chart_Period==0]
	Chart item=Outside_Temp_Chart service="jdbc" period=D refresh=3600 visibility=[Outside_Temp_Chart_Period==1]
	Chart item=Outside_Temp_Chart service="jdbc" period=W refresh=3600 visibility=[Outside_Temp_Chart_Period==2]
	Chart item=Outside_Temp_Chart service="jdbc" period=M refresh=3600 visibility=[Outside_Temp_Chart_Period==3]
	Chart item=Outside_Temp_Chart service="jdbc" period=Y refresh=3600 visibility=[Outside_Temp_Chart_Period==4]
}
```

When I initially open the Basic UI page with these charts it shows the correct graph image. However when changing the chart period several times it starts to use the wrong URL to the chart image.

A correct chart URL is:
```
http://machine/chart?groups=Outside_Temp_Chart&period=W&random=1&service=jdbc
```

After several changes a wrong chart URL that is being used is for instance:
```
http://machine/chart?groups=Outside_Temp_Chart&period=h&random=1&service=jdbc9153323982135213
```

When you use a non-default persistence service `org.eclipse.smarthome.ui.basic.internal.render.ChartRenderer` appends that service at the end of the chart URL. The Basic UI however expects the `random` parameter to be at the end of the URL. `function ControlImage(parentNode, callSuper)` in `smarthome.js` always seems to remove the last digit and append a new random to the end of the URL.

wborn:
The following changes have been applied:
* `ChartRenderer` now uses unix timestamp (t) similar to `ImageRenderer`
* `ChartRenderer` puts unix timestamp at the end of the URL so 404 Not Found is fixed when using non-default persistence service
* `ChartRenderer` properly initializes unix timestamp to bust cached versions
* `ControlImage` in `smarthome.js` uses unix timestamp instead of a random to bust caches


extensions/ui/org.eclipse.smarthome.ui.basic/web-src/smarthome.js
```diff
@@ -388,7 +388,7 @@
 					clearInterval(interval);
 					return;
 				}
-				_t.image.setAttribute("src", _t.url + Math.random().toString().slice(2));
+				_t.image.setAttribute("src", _t.url + Date.now());
```
watou: On this line and on line 378, you are still appending a number to the URL which is a bit blind to how the URL is formatted.  The [fix for ClassicUI](https://github.com/eclipse/smarthome/pull/1106/files#diff-ee0239b05dcc713c6860244dfd4d8688R76) instead replaces just the `t` parameter in the URL, which seems safer and more reliable.  Something like

```javascript
_t.image.setAttribute("src", _t.url.replace( /&t=\d+/g , '&t=' + new Date().getTime()));
```

kaikreuzer:
What I am just wondering: Wouldn't the much cleaner solution be to send a correct caching header in the HTTP response of the ChartServlet? We then would not need any of these random numbers...
Does anybody know a reason that speaks against this approach?

watou:
> Does anybody know a reason that speaks against this approach?

I was going to suggest this, as long as all steps in the chain to the client pass along and respect the `Cache-Control: no-store` header.  There could be valid circumstances where the headers are dropped or edited (such as through reverse-proxying CDNs), in which case dropping the cache-busting parameter would have to be reverted.  But if [this tutorial](https://developers.google.com/web/fundamentals/performance/optimizing-content-efficiency/http-caching), for example, is the final word on the subject, then dropping the `t` or `random` parameter would be fine.

kaikreuzer:
Would definitely make sense to me.
Nonetheless, I am fine with merging this PR as well for the time being (after @wborn has considered your review comments as well).

wborn:
Yes I agree that using headers would be better but this requires more thorough testing. So I prefer that this is done in a new issue.

The cache busting parameter is also used for Image widgets that are served through the `ProxyServlet`. This servlet also copies headers so most of the time it should work. I can imagine that some (older) IP cameras do not sent such headers which could cause issues if the correct headers are not returned by the `ProxyServlet`.

watou:
LGTM.  I agree that switching to using headers should be a different PR as needed.

PR #2432, opened 2016-11-10
2341-watch-service-relative-path

svilenvul:
See #2341 

-------
 Fixes bug introduced by #1597 reverted with #1754.

 The implementation of FolderObserver#processWatchEvent() does not get the file name correctly.
 This led to duplicated directory names (see the logs in #1597).

 AbstractWatchServiceTest is modified to run on different OS.

 Signed-off-by: Svilen Valkanov  <svilen.valkanov@musala.com>
 svilenvul )

PR #2557, opened 2016-11-29
cleanup watch service rel. path

maggu2810:
It seems @kaikreuzer and me reviewed the PR #2432 parallel, as it has been merged as I added some comments.
This PR changes the commented stuff.

@svilenvul As my formatter does some additional cleanup (removing trailing whitespaces), I assume that your IDE is not set upped correctly.

### DIFF
--- a/bundles/config/org.eclipse.smarthome.config.dispatch/src/main/java/org/eclipse/smarthome/config/dispatch/internal/ConfigDispatcher.java
+++ b/bundles/config/org.eclipse.smarthome.config.dispatch/src/main/java/org/eclipse/smarthome/config/dispatch/internal/ConfigDispatcher.java
@@ -7,9 +7,7 @@
  */
 package org.eclipse.smarthome.config.dispatch.internal;
 
-import static java.nio.file.StandardWatchEventKinds.ENTRY_CREATE;
-import static java.nio.file.StandardWatchEventKinds.ENTRY_DELETE;
-import static java.nio.file.StandardWatchEventKinds.ENTRY_MODIFY;
+import static java.nio.file.StandardWatchEventKinds.*;
 
 import java.io.File;
 import java.io.FileInputStream;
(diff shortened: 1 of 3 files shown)


kaikreuzer:
And it seems your review was much better ;-) Thanks.

Still, Travis fails since there are apparently code parts that need to be changed.
Is this an API breaking change? I didn't check that...

svilenvul:
[AutomationWatchService](https://github.com/eclipse/smarthome/blob/34e844b5e6aa2817542849fe10c43d3051d54e02/bundles/automation/org.eclipse.smarthome.automation.provider.file/src/main/java/org/eclipse/smarthome/automation/internal/provider/file/AutomationWatchService.java) was merged in the meantime ...

maggu2810:
@svilenvul Can you provide a fix for?

svilenvul:
Yes.

maggu2810:
The fix #2561 has been merged.
I rebased this PR, so we can wait for Travis CI.

kaikreuzer:
Failing unit test: FolderObserverTest

maggu2810:
> Failing unit test: FolderObserverTest

@svilenvul, @kaikreuzer  Do you agree that the changes of this PR are only cosmetic ones?

I don't think that changes of this PR are the source of a failed test and that this PR does not anything that will break anything (that is not already broken).

kaikreuzer:
Which means that https://github.com/eclipse/smarthome/pull/2432 is likely to have broken it? Yes, I would agree. @svilenvul could you check?

m-memova:
I ran the tests 100 times in Eclipse and 10 times in Maven and I cannot make it fail again :)

kaikreuzer:
Thanks for testing, @m-memova!
This time, Travis didn't fail on these tests and Hudson seems happy with the current code as well.
So let's merge :-)

PR #2564, opened 2016-11-29
Improve FirmwareUpdateProgressInfoEvent#toString

chriskn:
A small change to include also the percentage progress in the FirmwareUpdateProgressInfoEvent#toString method

### DIFF
--- a/bundles/core/org.eclipse.smarthome.core.thing/src/main/java/org/eclipse/smarthome/core/thing/firmware/FirmwareUpdateProgressInfoEvent.java
+++ b/bundles/core/org.eclipse.smarthome.core.thing/src/main/java/org/eclipse/smarthome/core/thing/firmware/FirmwareUpdateProgressInfoEvent.java
@@ -103,8 +103,9 @@ public final class FirmwareUpdateProgressInfoEvent extends AbstractEvent {
 
     @Override
     public String toString() {
-        return String.format("The firmware update progress step for thing %s changed to %s.", thingUID,
-                progressInfo.getProgressStep().name());
+        String stepName = progressInfo.getProgressStep() == null ? null : progressInfo.getProgressStep().name();
+        return String.format("The firmware update progress for thing %s changed. Step: %s Progress: %d.", thingUID,
+                stepName, progressInfo.getProgress());
     }
 
 }


bundles/core/org.eclipse.smarthome.core.thing/src/main/java/org/eclipse/smarthome/core/thing/firmware/FirmwareUpdateProgressInfoEvent.java
```diff
@@ -103,8 +103,8 @@ public boolean equals(Object obj) {
 
     @Override
     public String toString() {
-        return String.format("The firmware update progress step for thing %s changed to %s.", thingUID,
-                progressInfo.getProgressStep().name());
+        return String.format("The firmware update progress for thing %s changed. State: %s Progress: %d.", thingUID,
```
tomhoefer: I would prefer step instead of state

bundles/core/org.eclipse.smarthome.core.thing/src/main/java/org/eclipse/smarthome/core/thing/firmware/FirmwareUpdateProgressInfoEvent.java
```diff
@@ -103,8 +103,8 @@ public boolean equals(Object obj) {
...
+                progressInfo.getProgressStep().name(), progressInfo.getProgress());
```
tomhoefer: Due to the introduction of the percentage progress this could result in a NPE if no progress sequence has been defined. So we should check if progress step is not null

chriskn: @tomhoefer: Should we use a String like "UNDEFINED" or something like that if the step is null?

tomhoefer: No, I prefer the actual step which can be null

tomhoefer:
lgtm
however I wonder why the javadoc format has been changed

PR #2566, opened 2016-11-29
WeMo: Fix for 2446 in WemoHandler and WemoLightHandler

hmerk:
Also-by: Jayson Kubilis <jekubili@ktechs.net>


kaikreuzer:
Hm, seems as if there were other failing tests nonetheless, see https://hudson.eclipse.org/smarthome/job/SmartHomeDistribution-Nightly/1480/

@hmerk Could you please check and fix our build again asap? :-)

DigitalBites:
@hmerk @kaikreuzer The failing test cases were fixed in my original branch. They are caused because the device was not set to 'online' until it was responding vs blindly setting it online after initaizliation. The old behavior for that was left in tact as part of the last commit in my old branch. If you would like me to re-make this change in a new pull request i'm happy to do so - let me know.

hmerk:
@DigitalBites
No, thats not rhe reason tests are failing. If you have a closer look, it is claiming the missing InitialDelay for polling. Solution would be either to reimplement or removal from test.

PR #2568, opened 2016-11-30
fix handler initialization if child registration takes too long

### LINKED ISSUE #2559: ThingHandlers stay UNINITIALIZED of registration runs into timeout
The method `ThingManager.registerChildHandlers()` has the following code:

```
                        registerHandler(child);
                        initializeHandler(child);
```
Within `registerHandler` the handler registration is done in a "Safe" call, i.e. it might return after 5 seconds without having successfully registered the handler by that time.
As a result, `initializeHandler` will simply do nothing, since it checks whether the handler is correctly registered already.

The outcome is that the Thing stays in UNINITIALIZED state indefinitely.

[This seems to be like a report](https://community.openhab.org/t/binding-handler-registration-issues/17422) of exactly that situation.

sjka:
fixes #2559

### DIFF
--- a/bundles/core/org.eclipse.smarthome.core.thing/src/main/java/org/eclipse/smarthome/core/thing/internal/ThingManager.java
+++ b/bundles/core/org.eclipse.smarthome.core.thing/src/main/java/org/eclipse/smarthome/core/thing/internal/ThingManager.java
@@ -360,7 +360,7 @@ public class ThingManager extends AbstractItemEventSubscriber
                                 logger.warn("Handler for thing '{}' takes more than {}ms for processing event",
                                         handler.getThing().getUID(), SafeMethodCaller.DEFAULT_TIMEOUT);
                             } catch (Exception ex) {
-                                logger.error("Exception occured while calling handler: " + ex.getMessage(), ex);
+                                logger.error("Exception occured while calling handler: {}", ex.getMessage(), ex);
                             }
                         } else {
                             logger.info(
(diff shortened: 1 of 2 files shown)


sjka:
@sbussweiler could you please also have a look?

bundles/core/org.eclipse.smarthome.core.thing/src/main/java/org/eclipse/smarthome/core/thing/internal/ThingManager.java
```diff
@@ -1020,15 +1008,45 @@ protected void addThingHandlerFactory(ThingHandlerFactory thingHandlerFactory) {
...
+                SafeMethodCaller.call(new SafeMethodCaller.ActionWithException<Void>() {
+                    @Override
+                    public Void call() throws Exception {
+                        registerHandler(thing, thingHandlerFactory);
+                        initializeHandler(thing);
+                        return null;
+                    }
+                });
+            } catch (TimeoutException e) {
+                // TODO Auto-generated catch block
```
kaikreuzer: na, na, na ;-)

bundles/core/org.eclipse.smarthome.core.thing/src/main/java/org/eclipse/smarthome/core/thing/internal/ThingManager.java
```diff
@@ -1020,15 +1008,45 @@ protected void addThingHandlerFactory(ThingHandlerFactory thingHandlerFactory) {
...
+                e.printStackTrace();
+            } catch (ExecutionException e) {
+                // TODO Auto-generated catch block
```
kaikreuzer: dito

sbussweiler:
@sjka lgtm, thanks for the fix.

sjka:
I fixed the auto-generated blocks. 

However, I noticed that registerAndInitializeHandler() -> ... -> doInitializeHandler() leads to nested SafeMethodCaller invocations. 
I will have a look tomorrow in order to sort that out, just in case this would block when the pool runs out of threads.

PR #2570, opened 2016-11-30
Removed rule json view and settings button

aounhaider1:


maggu2810:
Is this PR related to some issue / dicussion or why has the JSON view (I didn't test it but that's the topic) removed?

kaikreuzer:
> why has the JSON view removed?

When it was built in it was only meant as a temporary measure to allow easy debugging while developing the rule engine - now that end users start using the feature, clicking the rule should do what is expected: Show the rule detail view. Developers can still access the JSON details through the REST API.

PR #2574, opened 2016-11-30
Revert "WeMo: Fix for 2446 in WemoHandler and WemoLightHandler"

kaikreuzer:
Reverts eclipse/smarthome#2566 as it has failing unit tests.

@hmerk I reverted in order to take out the time pressure for a fix. Please provide a new PR as soon as you are ready to do so (not urgent), thanks!


hmerk:
Hm, we should ask @svilenvul about the failing tests. Hudson says that testing the command handling is failing, but this PR is only about subscription handling.

@svilenvul could you check why tests are failing again.

hmerk:
@kaikreuzer @svilenvul foget about my last comment.
We removed the initial delay from starting the polling job, thats causing Hudson to fail.
So this should just be 2 lines of code to be changed, but I can not do this before end of next week.

kaikreuzer:
@maggu2810 OK to merge and make the build green again?

Issue #2550, opened 2016-11-28
CompositeTriggerHandler wrongly "dispose" child handlers


vilchev:
Currently CompositeTriggerHandler does: 

```
    @Override
    public void dispose() {
        List<Trigger> children = moduleType.getChildren();
        for (Trigger child : children) {
            TriggerHandler handler = moduleHandlerMap.get(child);
            handler.setRuleEngineCallback(null);
        }
        setRuleEngineCallback(null);
        super.dispose();
    }
```
The idea behind 
`            handler.setRuleEngineCallback(null);`
by now is to stop child handlers from notifying (trigger) the composite handler.

Most of Handler implementations, however, wouldn't expect `ruleCallback=null` (they can even have isNull check and never bother to set the new null value). The right way to notify given handler to stop working/ notifying should be calling the `dispose()`.

Proposal is 
where             

`handler.setRuleEngineCallback(null); `

to be replaced with

`handler.dispose();`

This wouldn't mean that all handlers would dispose all used resources (including ruleCallback=null) as expected (Someone could just forget..). 
Thats why additional check in the CompositeTriggerHandler.triggered(where child handlers notify/trigger) should be performed whether CompositeTriggerHandler is not already disposed.

danchom:
Agree. The RuleEngineCallback has to be set by the setRuleEngineCallback and removed by the handler's dispose method. 
We have to check implementation of current module handles if the remove the reference to the rulecallback.

PR #2577, opened 2016-12-01
CompositeXXXXXHandler now disposes its child handlers.

vilchev:
CompositeXXXXXHandler ,when disposed, disposes its childHandlers Related to: #2550 

Signed-off-by: Vasil Ilchev <v.ilchev@prosyst.bg>

### DIFF
--- a/bundles/automation/org.eclipse.smarthome.automation.core/src/main/java/org/eclipse/smarthome/automation/core/internal/composite/AbstractCompositeModuleHandler.java
+++ b/bundles/automation/org.eclipse.smarthome.automation.core/src/main/java/org/eclipse/smarthome/automation/core/internal/composite/AbstractCompositeModuleHandler.java
@@ -9,6 +9,7 @@ package org.eclipse.smarthome.automation.core.internal.composite;
 
 import java.util.HashMap;
 import java.util.LinkedHashMap;
+import java.util.List;
 import java.util.Map;
 
 import org.eclipse.smarthome.automation.Action;
@@ -41,7 +42,7 @@ import org.eclipse.smarthome.automation.type.TriggerType;
  * @param <H> type of module handler. It can be {@link TriggerHandler}, {@link ConditionHandler} or
  *            {@link ActionHandler}
  */
-public class AbstractCompositeModuleHandler<M extends Module, MT extends ModuleType, H extends ModuleHandler>
+public abstract class AbstractCompositeModuleHandler<M extends Module, MT extends ModuleType, H extends ModuleHandler>
         implements ModuleHandler {
 
     protected LinkedHashMap<M, H> moduleHandlerMap;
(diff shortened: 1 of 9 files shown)


danchom:
LGTM

bundles/automation/org.eclipse.smarthome.automation.core/src/main/java/org/eclipse/smarthome/automation/core/internal/composite/AbstractCompositeModuleHandler.java
```diff
@@ -88,10 +89,14 @@ public AbstractCompositeModuleHandler(M module, MT moduleType, LinkedHashMap<M,
...
     @Override
     public void dispose() {
-        if (moduleHandlerMap != null) {
-            moduleHandlerMap.clear();
-            moduleHandlerMap = null;
+        List<M> children = getChildren();
+        for (M child : children) {
+            ModuleHandler childHandler = moduleHandlerMap.remove(child);
+            childHandler.dispose();
         }
```
kaikreuzer: Wouldn't it be safer to additionally still call 
```
handler.setRuleEngineCallback(null);
```
afterwards, in case a handler "forgets" to dereference the callback in a dispose?
If we translate "dispose->deactivate" and "setRuleEngineCallback(null)->unsetRuleEngineCallback()" in a DS component, it would happen in a similar way: The reference that is injected from externally would also be revoked from externally.
I e.g. see that GenericEventTriggerHandler is NOT releasing the callback.

vilchev: Yes, you have a point, there is a bit of similarity in both mechanisms, but :

1. In DS we have the bind/unbind methods - where  implemenetator for sure knows what to do in the unbind 
In our case we have just the setRuleEnginceCallback, and imho, calling one and the same method , once with real reference, once with null, will only harden the implementation - having always logic isNull then do that, so that there are no unexpected NPE's in some clients handlers where:
1.1 would want to trigger the rule with some initial values, right after the ruleCallback is atached
1.2 would expect setRuleCallback to be called only once with the real reference and in that method to initalize for example an eventHandler - which would again cause a NPE if first dispose is called, where you have released some of the needed references).
 

If we do agree that dispose is not enough I think we have 2 options:
- Adding unsetRuleEngineCallback method in the TriggerHandler interface - matching the mechanism of bind/unbind.
- Propose BaseTriggerHandler class that in its dispose method would just releases the reference of ruleCallback 


But for sure I do not think its a good idea to call setRuleEngineCallback => null, when having another method(dispose) that actually its purpose is to release all used references, because this would work only if all TriggerHandlers: 
1. expect null value and somewhere check if ruleCallBack isNotNull
2. using the setRuleEngineCallback only for initalizing the ruleCallback itself (which is similar to 1.)

@danchom What do you think?

kaikreuzer: I'd be ok with your option 2 (BaseTriggerHandler sets callback to null in dispose) as this at least shows developers that this is needed. But it still means that your PR needs to fix GenericEventTriggerHandler as it does not even call super...

Issue #2214, opened 2016-09-24
Sonos binding: why calling updateMediaInfo() every minute ?


lolodomo:
@kgoderis & @R-Edelhoff : I have a question for you because I discovered something a little strange I don't really understand. It concerns the call to updateMediaInfo() every minute (by default). What is used in return to the call to getMediaInfo is only CurrentURI and CurrentURIMetaData. But these information are already provided in (pseudo) real time through the event subscription (but not used): AVTransportURI and AVTransportURIMetaData.
Do I miss something ? Is there a reason to not rely on event subscription ?
In fact, the track data are already updated when receiving CurrentTrackMetaData from the event subscription. Perect.
My feeling is that we should consider AVTransportURI and AVTransportURIMetaData being notified in real time rather than CurrentURI and CurrentURIMetaData being set only every minute. And then calling updateMediaInfo() every minute is useless.

kgoderis:
@lolodomo I think this is an artefact from the reverse engineering effort I did a looooong time ago. ;-) Yes, if we can get the same info from 2 new event subscriptions, that we should go that way instead. Can you propose a PR ?

lolodomo:
Yes I will. My current feeling after a short analysis of the code yesterday is that the change to do is very simple but then it will require a lot of testing.

lolodomo:
I studied more in depth the current code and the change to do is more important as I first imagined.
I will probably not be able to propose a change soon. But I keep it in my TODO list.

lolodomo:
Fix now merged.

PR #2590, opened 2016-12-04
Sonos binding: simplified update of media info (enh #2214)

lolodomo:
It concerns the update of the channels currenttitle, currentartist, currentalbum and currenttrack.
No need to request updated information every minute, the update is now based only on GENA notifications.

### DIFF
--- a/extensions/binding/org.eclipse.smarthome.binding.sonos/src/main/java/org/eclipse/smarthome/binding/sonos/handler/ZonePlayerHandler.java
+++ b/extensions/binding/org.eclipse.smarthome.binding.sonos/src/main/java/org/eclipse/smarthome/binding/sonos/handler/ZonePlayerHandler.java
@@ -2018,6 +2018,7 @@ public class ZonePlayerHandler extends BaseThingHandler implements UpnpIOPartici
 
                 // stop whatever is currently playing
                 coordinator.stop();
+                coordinator.waitForNotTransportState("PLAYING");
 
                 // clear any tracks which are pending in the queue
                 coordinator.removeAllTracksFromQueue();
@@ -2185,13 +2186,14 @@ public class ZonePlayerHandler extends BaseThingHandler implements UpnpIOPartici
     private void handleNotificationSound(Command notificationURL, ZonePlayerHandler coordinator) {
         String originalVolume = (isAdHocGroup() || isStandalonePlayer()) ? getVolume() : coordinator.getVolume();
         coordinator.stop();
+        coordinator.waitForNotTransportState("PLAYING");
         applyNotificationSoundVolume();
         int notificationPosition = coordinator.getQueue().size() + 1;
-        coordinator.setCurrentURI(QUEUE_URI + coordinator.getUDN() + "#0", "");
         coordinator.addURIToQueue(notificationURL.toString(), "", notificationPosition, false);
+        coordinator.setCurrentURI(QUEUE_URI + coordinator.getUDN() + "#0", "");
         coordinator.setPositionTrack(notificationPosition);
         coordinator.play();
-        waitForFinishedNotification();
+        coordinator.waitForFinishedNotification();
         if (originalVolume != null) {
             setVolumeForGroup(DecimalType.valueOf(originalVolume));
         }
@@ -2202,7 +2204,7 @@ public class ZonePlayerHandler extends BaseThingHandler implements UpnpIOPartici
         switch (nextAction) {
             case "PLAYING":
                 coordinator.play();
-                waitForTransportState("PLAYING");
+                coordinator.waitForTransportState("PLAYING");
                 break;
             case "PAUSED_PLAYBACK":
                 coordinator.pause();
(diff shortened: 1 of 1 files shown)


lolodomo:
I have done tests with all kind of sources I have (tune-in, local library, amazon music, spotify, analog line-in), with and without opmlurl for tune-in radio, with and without group.
The only case I cannot test is with pairs of Sonos and HC configs.

kaikreuzer:
@kgoderis Could you please review?

extensions/binding/org.eclipse.smarthome.binding.sonos/src/main/java/org/eclipse/smarthome/binding/sonos/handler/ZonePlayerHandler.java
```diff
@@ -802,191 +808,120 @@ public boolean isCoordinator() {
...
 
-            if (opmlUrl != null && currentURI.contains(STREAM_URI)) {
+        else if (currentURI.contains(STREAM_URI)) {
+            // Radio stream (tune-in)
+            boolean opmlUrlSucceeded = false;
+            if (opmlUrl != null) {
                 String stationID = StringUtils.substringBetween(currentURI, ":s", "?sid");
-                String previousStationID = stateMap.get("StationID");
-
-                Calendar now = Calendar.getInstance();
```
kgoderis: The throttling of the calls to the OPML should be kept as it is, as they monitor this. On top, all instances of OH uses the same partnerkey, so we have to be prudent with this.

extensions/binding/org.eclipse.smarthome.binding.sonos/src/main/java/org/eclipse/smarthome/binding/sonos/handler/ZonePlayerHandler.java
```diff
@@ -2621,11 +2556,6 @@ public String getCurrentZoneName() {
         return stateMap.get("CurrentZoneName");
     }
 
-    public String getCurrentURIFormatted() {
```
kgoderis: Should this method not be kept in place (without the call to updateCurrentURIFormatted)?

lolodomo: This is one of the few methods of the handler which are not used. If you want, I can restore it with a call to the new update method.

kgoderis: I was just not sure if it is called in handleCommand() linked to one of the Channels of the Thing

extensions/binding/org.eclipse.smarthome.binding.sonos/src/main/java/org/eclipse/smarthome/binding/sonos/handler/ZonePlayerHandler.java
kgoderis: Ok that is fair enough I would say

lolodomo: When you select a new radio station, you will have two notifications in a short delay, one with the state TRANSITIONING and one with the state PLAYING. With my current code, this will trigger 2 OPML requests in a short delay for the same station. To optimize and reduce a little more the OPML requests, we could decide to not update media information when state is TRANSITIONING ?

lolodomo: For a radio station, when state is TRANSITIONING, I could simply use the data I got from the notification (station name, additional informations). When the state moves to PLAYING, I will get the full information from an OPML request.
What will result is a slightly different station name during the TRANSITIONING phase and the PLAYING phase. But it will mean only one OPML request when you change station.

kgoderis: I would keep the 2 calls to OPML, so that there will be consistency in any of the UI's presented to the user

lolodomo:
Since my upgrade to firmware 7.0, the notification mechanism is no more reliable with some of them immediately interrupted. I have to check if it could be due to my change.

extensions/binding/org.eclipse.smarthome.binding.sonos/src/main/java/org/eclipse/smarthome/binding/sonos/handler/ZonePlayerHandler.java
lolodomo: Ok

lolodomo:
I identified the problem in the notification mechanism and certainly found the solution. I just need time to test it correctly and then I will provide an update to my PR.

lolodomo:
Ok, the problem with sound notification is now fixed. I have not squashed my changes to let you see what I have done to fix the last issue. But I can squash everything if required.
@kgoderis : did you test the binding with my first change ?

kgoderis:
@lolodomo No, I did not yet do that, but from the source code itself I do not see major issues here.

lolodomo:
So do you give your agreement for a merge?

kgoderis:
@kaikreuzer LGTM

Issue #2588, opened 2016-12-02
Rule action interpret triggering a TTS in case of interpretation error


lolodomo:
Currently, the interpret action in rule is triggering a command saying the error when the interpreation produces an error. It is working, it is really fun and cool but I think it should be optional.
Additionaly, the text is played by the default audio sink and you have no choice of audio sink.
What do you think about the idea to add a parameter to the interpret action that would be the audio sink to use in case of error. If no audio sink is provided, this will fail silently.

kaikreuzer:
Sounds good. How would you suggest to express that the response should be played on the default audio sink (as it is today)?

lolodomo:
The 2 current "interpret" actions will use TTS with the default audio sink. Unchanged.
I will add a third interpret action with a new parameter that will be the audio sink to use. If this parameter is null, TTS will not be used. If not null, it will be the audio sink to use with TTS.
To disable TTS or to choose another audio sink than the default one, the user will use the new "interpret" action with 3 parameters: text, HLI and audio sink.
Is it ok ?

kaikreuzer:
Sounds good!

lolodomo:
Enhancement merged.

PR #2593, opened 2016-12-04
New rule action "interpret" with audio sink parameter (enh #2588)

### LINKED ISSUE #2610: Rule actions say and interpet: inconsistent parameter type
We have 3 versions of the say rule action. They all provide as first parameter the text to be said.
One expects an Object class while the 2 others expects a String class.
This makes thing confusing for any user.
I would suggest to have a String parameter in all cases.
Same thing for interpret.

lolodomo:
say/interpret actions: fix inconsistent parameter type (fix #2610)


lolodomo:
PR updated to include the fix for issue #2610.

PR #2594, opened 2016-12-05
Full feature item creation

aounhaider1:
1) Show complete item creation options while creating item during channel linking.
2) Allow linking Dimmer and Switch items for channel.itemType=Color and Switch items for channel.itemType=Dimmer. (In addition to original itemType)
3) Show category icon.
4) Suggest an item name based on channel uid.


kaikreuzer:
Lgtm, thanks!

I have two suggestions to optimize the user workflow:
1. If no link is there on a channel, instead of expanding the empty list, it could directly open the "link" dialog.
2. A VERY regular use case will be the creation of an item for an unlinked channel. I think it would be cool if the "create new item" dialog would directly appear, if the round green button is pressed long. This might not be an obvious feature, but it is anyhow only meant as a shortcut for experts.

aounhaider1:
Updated accordingly.

aounhaider1:
It now uses: thing label +(group id) + channel label for item name suggestion. Secondly validates item name on initial form load.

Issue #2582, opened 2016-12-01
Thing stays in INITIALIZING state if initialize() failed once


kaikreuzer:
Scenario:
I create a Thing with a wrong configuration that makes the handler's initialize() method throw an exception (in my case an IllegalArgumentException()).
It is questionable whether this is allowed for the implementation or not - I would say this should not happen, instead the thing status should be set accordingly.
But in any case, this situation (due to a bug) is irrecoverable, because if I now correct the configuration and thus a "Thing update" occurs, the ThingManager logs:
```
14:34:31.193 [WARN ] [ome.core.thing.internal.ThingManager] - Attempt to initialize a handler twice for thing 'lifx:colorlight:30531ecf' at the same time will be ignored.
```
which means that it will never ever again try to initialize my handler - this looks like a bug to me.

sbussweiler:
Can you re-test your mentioned behavior with this [branch](https://github.com/qivicon/smarthome/tree/fixed-thing-update)?

lolodomo:
What would happen if the thing handler initialization exits in timeout ?

sjka:
> What would happen if the thing handler initialization exits in timeout ?

Nothing really. The TimeoutException does not EXIT the initialization, but only makes the rest of the framework continue doing its job. The initialization runs in a separate thread and will continue running anyway. Maybe we should be more clear in the warning message that we print to the logs.

It's the ThingHandler's duty to set the Thing to ONLINE or OFFLINE at the end of the initialization. So since the thread keeps running, this will eventually happen - unless there is a bug in the ThingHandler.initialize(), of course.

lolodomo:
You mean the thing handler init is not interrupted after 5 seconds ?

sbussweiler:
> You mean the thing handler init is not interrupted after 5 seconds ?

Exactly.

lolodomo:
I have since ages an issue with the astro binding that is sometimes remaining in INITIALIZING state. I hope that it could be due to the same reason.
When I run in the console a bundle:stop and then a bundle:start, I got this error:
```
18:30:48.963 [WARN ] [ome.core.thing.internal.ThingManager] - Thing handler for thing 'astro:sun:local' already registered
```
and the thing is remaining in INITIALIZING state.
I even tried to uninstall the astro feature and install it again (through the console) and same result and same warning.

sjka:
@sbussweiler nope. 

[...]

PR #2597, opened 2016-12-06
[LIFX] give the framework a chance to detect failed initialization

sjka:
...as otherwise the Thing will stay in `INITIALIZING` forever as the framework waits for the handler to set it to either `ONLINE` or `OFFLINE`.

### DIFF
--- a/extensions/binding/org.eclipse.smarthome.binding.lifx/src/main/java/org/eclipse/smarthome/binding/lifx/handler/LifxLightHandler.java
+++ b/extensions/binding/org.eclipse.smarthome.binding.lifx/src/main/java/org/eclipse/smarthome/binding/lifx/handler/LifxLightHandler.java
@@ -165,8 +165,9 @@ public class LifxLightHandler extends BaseThingHandler {
             currentStateUpdater.start();
             onlineStateUpdater.start();
             lightStateChanger.start();
-        } catch (Exception ex) {
-            logger.error("Error occured while initializing LIFX handler: " + ex.getMessage(), ex);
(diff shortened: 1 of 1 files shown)


kaikreuzer:
Hm, in https://github.com/eclipse/smarthome/issues/2582#issue-192850072, I wrote:

"It is questionable whether [throwing an exception in initialize()] is allowed for the implementation or not - I would say this should not happen, instead the thing status should be set accordingly."

With this PR, you seem to do exactly this. Maybe we first need to clarify the contract of initialize()?

sjka:
> Maybe we first need to clarify the contract of initialize()?

Maybe. But maybe we first want to make it work again. 

So you are implying that every binding should look like this?

```java
    @Override
    public void initialize() {
        try {
            // ...
        } catch (Exception e) {
            logger.error("Something went seriously wrong: {}", e.getMessage(), e);
            updateStatus(ThingStatus.OFFLINE, ThingStatusDetail.CONFIGURATION_ERROR);
        }
    }
```

Okay, no problem. It's just some additional glue code that clutters the bindings, but technically this is valid. 

In this case there is a subtle difference even: The binding sets it to OFFLINE/CONFIGURATION_ERROR instead of UNINITIALIZED/HANDLER_INITIALIZATION_ERROR, because that's actually the case: the configuration was wrong in a way. The latter is used by the framework and we still need it for unforeseen stuff going seriously wrong. The consequence is bigger though: in the first case the Thing is considered to be initialized, in the second case it's not.

I will amend the PR accordingly.

PR #2598, opened 2016-12-06
run Maven in quiet mode on travis

kaikreuzer:
All Travis builds are failing now since the log exceeds 4MB.
This should be the quickest solution to the problem. If anyone has a good idea how to only partially reduce the logging, feel free to come up with something better!

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -8,6 +8,4 @@ before_install: echo "MAVEN_OPTS='-Xms1g -Xmx2g -XX:PermSize=512m -XX:MaxPermSiz
 #$ mvn install -DskipTests=true -Dmaven.javadoc.skip=true -B -V
 #$ mvn test -B
 
-install: true
-script:
-  - mvn clean install -B -V
+install: travis_wait 60 mvn clean install -B -V -q


maggu2810:
I set install and script previously.
Any reason you remove the script definition?

I assume if you remove it, the default will be used again:
https://docs.travis-ci.com/user/languages/java/#Projects-Using-Maven

kaikreuzer:
> Any reason you remove the script definition?

Yes, because I couldn't figure out what it was good for and how to correctly apply the travis_wait to it. As the check was just successful, it seems to be ok like it is.

Could you elaborate, why the script part is important? The provided link did not clarify this for me, sorry.

maggu2810:
The link above also contains a link to the documentation of the build lifecycle (https://docs.travis-ci.com/user/customizing-the-build/#The-Build-Lifecycle).
My reading of the provided Travis CI documentation has been that:
* `install` is executed before `script`
* `install` could be used to e.g. fetch the necessary dependencies
* On Maven a limited build (phase `install` without tests etc.) is executed 
* `script` provides the build command
* On Maven the phase `test` is called

I don't think the default Maven goals are optimal.
I also did not find a good way to split our build in different steps.

So, I think it is okay to do nothing for `install` and add the correct build command to `script`.

kaikreuzer:
I was looking at https://docs.travis-ci.com/user/common-build-problems/#Build-times-out-because-no-output-was-received, which seemed to suggest that it is a very normal way to call mvn on the `install`.

Looking at the logs, it does not seem that "default Maven goals" are used, but it executes exactly what is specified:
```
$ echo "MAVEN_OPTS='-Xms1g -Xmx2g -XX:PermSize=512m -XX:MaxPermSize=1g'" > ~/.mavenrc
$ travis_wait 60 mvn clean install -B -V -q
Still running (36 of 60): mvn clean install -B -V -q
The command mvn clean install -B -V -q exited with 0.
```
I'd think this is just fine, but if you insist, I can also change it back to install/script split.

maggu2810:
> Looking at the logs, it does not seem that "default Maven goals" are used, but it executes exactly what is specified:

It seems your one is different to mine.

```text
$ grep 'The command' /tmp/132770300.txt 
The command mvn clean install -B -V -q exited with 0.
The command "mvn test -B" exited with 0.
```

maggu2810:
IMHO the call to `mvn test -B` after `mvn clean install -B -V -q` already succeeded, will only waste build resources (e.g. time).

kaikreuzer:
-q still logs errors, so you usually see what causes a failure. Also, all output of the test executions were still logged, which is now missing in your version...

maggu2810:
> all output of the test executions were still logged

Because they are executed twice, first in the -q mode (install) and then again (script).

> which is now missing in your version...

I don't think so. All the output is written to a file. At the end of the build process the last x lines are written to stdout and so visible in the log. It is just a matter how much lines we want to print on failure and on success.
Currently the last 200 lines are shown on success and the last 2000 lines on failure. What do you miss? Do you want to see more lines?

kaikreuzer:
> Do you want to see more lines?

Let's see how the builds go. If we feel it isn't enough, we can extend it as required.
The main issue that the Travis build failed for a few days is solved, that's most important to me.

PR #2599, opened 2016-12-06
Handle handler initialization correctly.

sbussweiler:
Signed-off-by: Stefan Bussweiler <s.bussweiler@external.telekom.de>

### DIFF
--- a/bundles/core/org.eclipse.smarthome.core.thing/src/main/java/org/eclipse/smarthome/core/thing/internal/ThingManager.java
+++ b/bundles/core/org.eclipse.smarthome.core.thing/src/main/java/org/eclipse/smarthome/core/thing/internal/ThingManager.java
@@ -511,8 +511,10 @@ public class ThingManager extends AbstractItemEventSubscriber
                             ex.getMessage(), ex);
                 }
             } else {
-                logger.debug("Cannot notify handler about updated thing {}, because thing is not initialized "
-                        + "(must be in status ONLINE or OFFLINE).", thing.getThingTypeUID());
+                logger.debug(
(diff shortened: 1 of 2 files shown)


sjka:
This indeed fixes the issue we discovered in #2582.

LGTM

bundles/core/org.eclipse.smarthome.core.thing/src/main/java/org/eclipse/smarthome/core/thing/internal/ThingManager.java
```diff
@@ -511,8 +511,10 @@ public Void call() throws Exception {
                             ex.getMessage(), ex);
                 }
             } else {
-                logger.debug("Cannot notify handler about updated thing {}, because thing is not initialized "
-                        + "(must be in status ONLINE or OFFLINE).", thing.getThingTypeUID());
+                logger.debug(
+                        "Cannot notify handler about updated thing '{}', because thing is not initialized (must be in status ONLINE or OFFLINE). Start handler initialization instead.",
```
maggu2810: I have been told some time ago that it is not the thing that is (or is not) initialized, it is the handler only.
It is currently derived by the thing status, but that should be already abstracted by the isHandlerInitialized function above.

sbussweiler: Yes, you are right. I will change the message "... because the handler is not initialized .."

kaikreuzer: @sbussweiler waiting for your update :-)

PR #2608, opened 2016-12-06
work around the travis build output limit

maggu2810:
That is similar to #2598
I created that PR to get a Travis CI build.


maggu2810:
@kaikreuzer WDYT about this one?

PR #2612, opened 2016-12-07
introduce an UNKNOWN thing status

### LINKED ISSUE #2127: Thing Status enhancements

sjka:
...for situations when the thing handler initialization is complete
but due to the nature of the device and/or communication protocol
the thing handler is not yet able to determine if the thing is
ONLINE or OFFLINE.

Like ONLINE and OFFLINE, the UNKNOWN status belongs to the group
of ThingHandler-is-fully-initialized ones, describing rather the status
of the Thing than the status of the ThingHandler.

fixes #2127


sjka:
just found some places which I missed, will amend this PR asap

sjka:
done. Hope I now got'em all...!

bundles/core/org.eclipse.smarthome.core.thing/src/main/java/org/eclipse/smarthome/core/thing/internal/ThingManager.java
```diff
@@ -365,7 +377,7 @@ public Void call() throws Exception {
                         } else {
                             logger.info(
                                     "Not delegating command '{}' for item '{}' to handler for channel '{}', "
-                                            + "because thing is not initialized (must be in status ONLINE or OFFLINE).",
+                                            + "because thing is not initialized (must be in status UNKNOWN, ONLINE or OFFLINE).",
```
kaikreuzer: This should probably be changed to "handler is not initialized", see https://github.com/eclipse/smarthome/pull/2599#discussion_r91108335.

bundles/core/org.eclipse.smarthome.core.thing/src/main/java/org/eclipse/smarthome/core/thing/internal/ThingManager.java
```diff
@@ -415,7 +427,7 @@ public Void call() throws Exception {
                         } else {
                             logger.info(
                                     "Not delegating update '{}' for item '{}' to handler for channel '{}', "
-                                            + "because thing is not initialized (must be in status ONLINE or OFFLINE).",
+                                            + "because thing is not initialized (must be in status UNKNOWN, ONLINE or OFFLINE).",
```
kaikreuzer: This should probably be changed to "handler is not initialized", see https://github.com/eclipse/smarthome/pull/2599#discussion_r91108335.

bundles/core/org.eclipse.smarthome.core.thing/src/main/java/org/eclipse/smarthome/core/thing/internal/ThingManager.java
```diff
@@ -512,7 +525,7 @@ public Void call() throws Exception {
                 }
             } else {
                 logger.debug("Cannot notify handler about updated thing {}, because thing is not initialized "
```
kaikreuzer: This should probably be changed to "handler is not initialized", see https://github.com/eclipse/smarthome/pull/2599#discussion_r91108335.

sjka:
thanks for the review. I corrected the logging statements and rebased the PR.

kaikreuzer:
Thanks. Just came across another question: Shall the `BaseThingHandler.initialize()` method be changed to set the status to UNKNOWN instead of ONLINE now?

sjka:
> Shall the BaseThingHandler.initialize() method be changed to set the status to UNKNOWN instead of ONLINE now?

I didn't dare doing this, just in case some bindings rely on it. So for backward compatibility I think it safer to leave it as it is.

kaikreuzer:
ok, that's an argument, thanks.

kaikreuzer:
@maggu2810 Any further comments from your side before merging as it concerns our discussion at #2127?

maggu2810:
No

maggu2810:
Yes, moment

bundles/core/org.eclipse.smarthome.core.thing/src/main/java/org/eclipse/smarthome/core/thing/util/ThingHandlerHelper.java
```diff
@@ -15,20 +15,30 @@
...
 
     private ThingHandlerHelper() {
     }
 
     /**
+     * Checks if the given state indicates that a thing handler has been initialized.
+     *
+     * @return true if the thing handler has been initialized, otherwise false.
+     */
+    public static boolean isHandlerInitialized(final ThingStatus thingStatus) {
```
kaikreuzer: It seems to be required [here](https://github.com/eclipse/smarthome/pull/2612/files#diff-6d6897e04b52e8efb8749b37d50acbdeR299), a place where before the check against ONLINE/OFFLINE was done. And I do not see how we could solve it differently.

kaikreuzer: > Or do you think it needs to much load?

Yes, this is really just filtering events and I wouldn't want this place to require a lookup in the registry.

markusjw:
The state diagram implies that a change from ONLINE ot OFFLINE to UNKNOWN is not allowed. Why not give full choice to the binding programmer here?
An example: When a bridge goes ONLINE again, a thing triggered by bridgeStatusChanged() might want to switch to UNKNOWN because this is a kind of initial status.

sjka:
I don't have a strong opinion against it. I just though there is no use-case for it: once the handler established a connection to the device, it either has it or not. If a connection is lost, it should switch back to OFFLINE. There is no "I don't know yet, maybe it works" situation anymore.
Do you have any concrete use-cases in mind there switching back to UNKOWN would make sense?

maggu2810:
> There is no "I don't know yet, maybe it works" situation anymore.

Hm, @kaikreuzer told me that every status is outdated e.g. "I don't know if it is still offline, let's try the communication regardless".
https://github.com/eclipse/smarthome/issues/2127

After #2127 I assume every status could be seen as the last one we know about and not the one that is currently the correct one.

PR #2615, opened 2016-12-07
Delay DSL thing creation until all XMLs from the bindings have been l…

### LINKED ISSUE #2521: ClassCastException in HandlerFactories
There are reports from a number of bindings about ClassCastExceptions, when trying to instantiate a handler for a bridge - the Thing framework seems to hand in a Thing (and not a Bridge) instance in that case. Before the Thing refactoring, this code was working, so it seems that this must be some bug in the framework, not in the bindings. See a log here:

```
Caused by: java.lang.ClassCastException: org.eclipse.smarthome.core.thing.internal.ThingImpl cannot be cast to org.eclipse.smarthome.core.thing.Bridge
at org.openhab.binding.homematic.handler.HomematicThingHandlerFactory.createHandler(HomematicThingHandlerFactory.java:50)[204:org.openhab.binding.homematic:2.0.0.201611212345]
at org.eclipse.smarthome.core.thing.binding.BaseThingHandlerFactory.registerHandler(BaseThingHandlerFactory.java:98)[104:org.eclipse.smarthome.core.thing:0.9.0.201611191809]
at org.eclipse.smarthome.core.thing.internal.ThingManager$7.call(ThingManager.java:586)[104:org.eclipse.smarthome.core.thing:0.9.0.201611191809]
at org.eclipse.smarthome.core.thing.internal.ThingManager$7.call(ThingManager.java:1)[104:org.eclipse.smarthome.core.thing:0.9.0.201611191809]
at org.eclipse.smarthome.core.common.SafeMethodCaller$CallableWrapper.call(SafeMethodCaller.java:177)[97:org.eclipse.smarthome.core:0.9.0.201611191809]
... 4 more
```

More infos also at https://github.com/openhab/openhab2-addons/issues/1462

sjka:
…oaded

Reusing the mechanism known from the ThingManager for Handler initialization,
therefore factoring out the common code.

fixes #2521

### DIFF
--- a/bundles/config/org.eclipse.smarthome.config.core/src/main/java/org/eclipse/smarthome/config/core/BundleProcessorVetoManager.java
+++ b/bundles/config/org.eclipse.smarthome.config.core/src/main/java/org/eclipse/smarthome/config/core/BundleProcessorVetoManager.java
@@ -1,0 +1,145 @@
+/**
+ * Copyright (c) 2014-2016 by the respective copyright holders.
+ * All rights reserved. This program and the accompanying materials
+ * are made available under the terms of the Eclipse Public License v1.0
+ * which accompanies this distribution, and is available at
+ * http://www.eclipse.org/legal/epl-v10.html
+ */
+package org.eclipse.smarthome.config.core;
+
+import java.security.AccessController;
+import java.security.PrivilegedAction;
+import java.util.HashSet;
+import java.util.Set;
+
+import org.eclipse.smarthome.config.core.BundleProcessor.BundleProcessorListener;
+import org.osgi.framework.Bundle;
+import org.osgi.framework.BundleReference;
+import org.slf4j.Logger;
+import org.slf4j.LoggerFactory;
+
+import com.google.common.collect.LinkedListMultimap;
+import com.google.common.collect.Multimap;
+import com.google.common.collect.Multimaps;
+
+/**
+ * Utility class in order to delay an action until a binding's XML loading completed.
+ *
+ * It takes an action and whenever it is instructed to execute this action, it is going to check first if there are any
+ * vetoes due to the fact that the XML meta-data of the relevant binding might not be fully loaded. If this is the case,
+ * then the action will be delayed until the loading was finished.
+ *
+ * As this class is not an OSGi service itself, it needs some help from the outside by injecting/removing
+ * BundleProcessors via {@link #addBundleProcessor(BundleProcessor)} and
+ * {@link #removeBundleProcessor(BundleProcessor)}.
+ *
+ * @author Simon Kaufmann - Initial contribution and API
+ */
+public class BundleProcessorVetoManager<T> implements BundleProcessorListener {
+
+    public interface Action<T> {
+        void apply(T object);
+    }
+
+    private final Logger logger = LoggerFactory.getLogger(BundleProcessorVetoManager.class);
+
+    private final Set<BundleProcessor> bundleProcessors = new HashSet<BundleProcessor>();
+    private final Multimap<Long, BundleProcessor> vetoes = Multimaps
+            .synchronizedListMultimap(LinkedListMultimap.<Long, BundleProcessor> create());
+    private final Multimap<Long, T> queue = Multimaps.synchronizedListMultimap(LinkedListMultimap.<Long, T> create());
+    private final Action<T> action;
+
+    /**
+     * Construct a BundleProcessorVetoManager for the given action
+     *
+     * @param action the action to run (potentially delayed)
+     */
+    public BundleProcessorVetoManager(Action<T> action) {
+        this.action = action;
+    }
+
+    @Override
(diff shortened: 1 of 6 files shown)


bundles/config/org.eclipse.smarthome.config.core/src/main/java/org/eclipse/smarthome/config/core/BundleProcessorVetoManager.java
```diff
@@ -0,0 +1,146 @@
...
+ *
+ * As this class is not an OSGi service itself, it needs some help from the outside by injecting/removing
+ * BundleProcessors via {@link #addBundleProcessor(BundleProcessor)} and
+ * {@link #removeBundleProcessor(BundleProcessor)}.
+ *
+ * @author Simon Kaufmann - Initial contribution and API
+ */
+public class BundleProcessorVetoManager<T> implements BundleProcessorListener {
+
+    public static interface Action<T> {
```
maggu2810: If we want to "avoid modifiers which are implied by the context", you can drop "static" here.

bundles/config/org.eclipse.smarthome.config.core/src/main/java/org/eclipse/smarthome/config/core/BundleProcessorVetoManager.java
```diff
@@ -0,0 +1,146 @@
...
+        public void apply(T object);
```
maggu2810: If we want to "avoid modifiers which are implied by the context", you can drop "public" here.

bundles/config/org.eclipse.smarthome.config.core/src/main/java/org/eclipse/smarthome/config/core/BundleProcessorVetoManager.java
```diff
@@ -0,0 +1,146 @@
...
+    private final Logger logger = LoggerFactory.getLogger(BundleProcessorVetoManager.class);
+
+    private final Set<BundleProcessor> bundleProcessors = new HashSet<BundleProcessor>();
+    private final Multimap<Long, BundleProcessor> vetoes = Multimaps
+            .synchronizedListMultimap(LinkedListMultimap.<Long, BundleProcessor> create());
+    private final Multimap<Long, T> queue = Multimaps.synchronizedListMultimap(LinkedListMultimap.<Long, T> create());
+    private final Action<T> action;
+
+    /**
+     *
```
maggu2810: Shouldn't you add at least some very short description here like "Constructor".
Is the short description optional (I don't know)?

bundles/config/org.eclipse.smarthome.config.core/src/main/java/org/eclipse/smarthome/config/core/BundleProcessorVetoManager.java
```diff
@@ -0,0 +1,146 @@
...
+    public BundleProcessorVetoManager(Action<T> action) {
+        this.action = action;
+    }
+
+    @Override
+    public void bundleFinished(final BundleProcessor context, final Bundle bundle) {
+        vetoes.remove(bundle.getBundleId(), context);
+        if (vetoes.get(bundle.getBundleId()).isEmpty()) {
+            logger.debug("Finished loading meta-data of bundle '{}'.", bundle.getSymbolicName());
+            synchronized (queue) {
```
maggu2810: Here you are using synchronized for the queue handling, but you don't do it below if you add something to the queue.

bundles/config/org.eclipse.smarthome.config.core/src/main/java/org/eclipse/smarthome/config/core/BundleProcessorVetoManager.java
```diff
@@ -0,0 +1,146 @@
...
+                if (!vetoes.containsEntry(bundleId, proc)) {
+                    logger.trace("Marking '{}' as vetoed by '{}'", bundle.getSymbolicName(), proc);
+                    vetoes.put(bundleId, proc);
+                }
+            } else {
+                logger.trace("'{}' already finished processing '{}'", proc, bundle.getSymbolicName());
+            }
+        }
+        if (veto) {
+            if (!queue.containsEntry(bundle, object)) {
```
maggu2810: But the whole if stuff in synchronized(queue) -- see above. Or isn't it necessary

sjka: To be honest, I think we can drop the synchronization here entirely. The collection is synchronized itself and the iteration above if over a collection of removed elements, which I don't expect it can change.

sjka:
Thanks for the quick feedback! I have updated the commit accordingly.

kaikreuzer:
@maggu2810 Ok to merge?

maggu2810:
> @maggu2810 Ok to merge?

Sorry, missed to add my approved review comment.

PR #2621, opened 2016-12-08
remove rest.auth from base feature

maggu2810:
There is not need to add the rest.auth bundle to the base feature as we can also run fine without it.

### DIFF
--- a/features/karaf/src/main/feature/feature.xml
+++ b/features/karaf/src/main/feature/feature.xml
@@ -55,7 +55,6 @@
     <bundle>mvn:org.eclipse.smarthome.io/org.eclipse.smarthome.io.monitor/${project.version}</bundle>
     <bundle>mvn:org.eclipse.smarthome.io/org.eclipse.smarthome.io.net/${project.version}</bundle>
     <bundle>mvn:org.eclipse.smarthome.io/org.eclipse.smarthome.io.rest/${project.version}</bundle>
-    <bundle>mvn:org.eclipse.smarthome.io/org.eclipse.smarthome.io.rest.auth/${project.version}</bundle>
     <bundle>mvn:org.eclipse.smarthome.io/org.eclipse.smarthome.io.rest.core/${project.version}</bundle>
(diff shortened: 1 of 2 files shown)


maggu2810:
@kaikreuzer Previously the features has been ordered by their name.
Your PR placed the esh-rest-auth-basic below the esh-auth one. Any reason to group not by name anymore? Perhaps it is for the moment easier to group authentication stuff together but in the long run, I assume alphabetical order is much more easier.

features/karaf/src/main/feature/feature.xml
```diff
@@ -55,7 +55,6 @@
     <bundle>mvn:org.eclipse.smarthome.io/org.eclipse.smarthome.io.monitor/${project.version}</bundle>
     <bundle>mvn:org.eclipse.smarthome.io/org.eclipse.smarthome.io.net/${project.version}</bundle>
     <bundle>mvn:org.eclipse.smarthome.io/org.eclipse.smarthome.io.rest/${project.version}</bundle>
-    <bundle>mvn:org.eclipse.smarthome.io/org.eclipse.smarthome.io.rest.auth/${project.version}</bundle>
```
kaikreuzer: I wouldn't remove this bundle here - it is a part of the ESH base feature and it does not harm. With the same logic, you could otherwise also separate core.id, core.voice, io.console, io.rest.sse etc.

maggu2810: You are wrong! It does harm, because it needs the security provider.

maggu2810: I am not a fan of that optional imports.
It makes the stuff much more difficult than it should be.
Some time someone will change the code and use an optional import and we will run into NCDF exceptions at runtime etc.
I would like to keep the separate bundle and add it to the resp. features as it is done in this PR.

kaikreuzer: Ok then. Although I am not a fan of bundles with just 1.5 classes inside. But yeah, it was me who created it in the first place ;-)

PR #2626, opened 2016-12-08
Applied two little fixes to documentation.

AlexanderKostadinov28:
Fixed wrong URL for ChannelTypeProvider (not sure why the diff shows the whole paragraph changed, it is just the URL, which currently is https://github.com/eclipse/smarthome/blob/master/bundles/core/org.eclipse.smarthome.core.thing/src/main/java/org/eclipse/smarthome/core/thing/type/ChannelTypeProvider.java) and improper indentation of xml snippet element.


kaikreuzer:
Thanks!

Issue #2560, opened 2016-11-29
Rollershutter with slider


aounhaider1:
As the rollershutter item accepts percent value (http://www.eclipse.org/smarthome/documentation/concepts/items.html). Would it make sense to also have a slider for it on control page?

sjka:
maybe it's personal preference, but I don't miss the slider control. 
It's good to have the percent values though in rules.

kaikreuzer:
I like the fact that the Paper UI offers UP/DOWN/STOP buttons for Rollershutter - my suggestion would be to additionally make the percent value editable, so that users can manually post a certain value. This would cover all cases for testing and also provides the most often used controls for users.

bbolte:
I am currently developing a binding with a Rollershutter and like the Paper UI for quick testing of the functionality. The up/down/stop buttons are good, but I miss the possibility to see/set the current position of the Rollershutter in percent for testing the full functionality of the binding and the device. As a workaround, I am currently defining an additional Dimmer channel, which forwards the PercentType commands to the Rollershutter channel.

In my opinion, it makes sense to integrate a position widget for the Rollershutter to avoid such workarounds. I would personally prefer a visual feedback of the Rollershutter position over just a numerical value, e.g. a slider as a device independent representation and/or a window with blinds as a device dependent representation.

Some devices may not support getting/setting the Rollershutter position in percent. I can imagine that a position widget could be confusing for users in this case. Maybe the visibility/appearance of the position widget could be configurable via the channel-type definition. (Options could be to use the category, tags, and/or state settings.)

PR #2628, opened 2016-12-08
Rollershutter state input

aounhaider1:
fixes https://github.com/eclipse/smarthome/issues/2560

### DIFF
--- a/extensions/ui/org.eclipse.smarthome.ui.paper/web-src/js/controllers.control.js
+++ b/extensions/ui/org.eclipse.smarthome.ui.paper/web-src/js/controllers.control.js
@@ -132,7 +132,7 @@ angular.module('PaperUI.controllers.control', []).controller('ControlPageControl
         for (var int = 0; int < $scope.data.items.length; int++) {
             var item = $scope.data.items[int];
             if (item.name === itemName) {
-                if (item.type && (item.type == "Number" || item.groupType == "Number")) {
+                if (item.type && (item.type == "Number" || item.groupType == "Number" || item.type == "Rollershutter")) {
                     var parsedValue = Number(item.state);
                     if (isNaN(parsedValue)) {
                         item.state = null;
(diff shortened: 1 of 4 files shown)


extensions/ui/org.eclipse.smarthome.ui.paper/web-src/js/controllers.control.js
```diff
@@ -527,6 +527,14 @@ angular.module('PaperUI.controllers.control', []).controller('ControlPageControl
     if ($scope.item.state === 'UNDEF' || $scope.item.state === 'NULL') {
         $scope.item.state = '-';
     }
+    $scope.sendStateCommand = function(command) {
+        $scope.sendCommand(command);
+        if (command == "DOWN") {
+            $scope.item.state = 100;
```
kaikreuzer: Hm, are you manually setting the state here? You should not - the runtime should send according events, which should be used to update the displayed status.

extensions/ui/org.eclipse.smarthome.ui.paper/web-src/partials/control.html
```diff
@@ -129,24 +129,31 @@ <h2 class="state edit" ng-if="!editMode" ng-click="editState()">{{getStateText(i
...
 											<div ng-switch-when="Rollershutter" ng-controller="RollershutterItemController">
-												<div layout="row" layout-align="start center">
+												<div layout="row" layout-align="start center" flex="100">
 													<div flex="40" layout="row" layout-align="start center">
 														<p>
 															<i class="material-icons">{{getIcon(item.category, 'format_align_justify')}}</i>{{getLabel(item.category, item.label, 'Position')}}
 														</p>
 													</div>
+													<div flex="60" layout="row" layout-align="end center">
+                                                        <div ng-show="isReadOnly(item)">
```
kaikreuzer: something's messed up with the formatting here.

extensions/ui/org.eclipse.smarthome.ui.paper/web-src/js/controllers.control.js
aounhaider1: This is a click handler and I am just setting the state before sending the command to the runtime. Its being done the same way in all other control widgets.

aounhaider1:
Formatted control.html

extensions/ui/org.eclipse.smarthome.ui.paper/web-src/js/controllers.control.js
kaikreuzer: Correct. Depending on the real hardware, you might get multiple updates with incremental changes.

aounhaider1:
Updated.

Issue #858, opened 2016-01-15
Basic-UI:Selection item


s-simma:
After closing the selection box the mapped text, which belongs to the selected item is not displayed.

kaikreuzer:
@resetnow Could you have a look?

PR #2629, opened 2016-12-08
Allow changing of unmanaged Things

### LINKED ISSUE #1977: Do not persist changes to a Thing when done from the Thing handler itself
As discussed [here](https://www.eclipse.org/forums/index.php?t=msg&th=1079343&goto=1739523&#msg_1739523), we suggest to allow Thing handlers to do dynamic changes to the Thing structure during runtime, indifferently whether where the Thing definition came from (DSL, database, etc.).
These changes should NOT be persisted, since it is assumed that the handler does them upon every start anyhow.

sjka:
So far, changes were allowed only to Things coming from
the ManagedThingProvider. With this change it is now allowed
to change all Things structurally, no matter where they are
from. They will simply not be persisted if its provider is not
managed.

fixes #1977

### DIFF
--- a/bundles/core/org.eclipse.smarthome.core.thing.test/src/test/groovy/org/eclipse/smarthome/core/thing/internal/ThingManagerOSGiTest.groovy
+++ b/bundles/core/org.eclipse.smarthome.core.thing.test/src/test/groovy/org/eclipse/smarthome/core/thing/internal/ThingManagerOSGiTest.groovy
@@ -35,6 +35,7 @@ import org.eclipse.smarthome.core.thing.Channel
 import org.eclipse.smarthome.core.thing.ChannelUID
 import org.eclipse.smarthome.core.thing.ManagedThingProvider
 import org.eclipse.smarthome.core.thing.Thing
+import org.eclipse.smarthome.core.thing.ThingProvider
 import org.eclipse.smarthome.core.thing.ThingRegistry
 import org.eclipse.smarthome.core.thing.ThingStatus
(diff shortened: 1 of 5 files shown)


lolodomo:
@sjka : that means you can now update the configuration of a thing that is defined in a config file ? The only point is that the changes will be lost if you restart OH ?
If true, that is a very interestin gchange.

sjka:
@lolodomo  sorry for the delay, I didn't spot your question... Yes, this is exactly what it means.

sipvoip:
I just tried with #858, the last good build on cloudbees from March 31st.

2017-04-09 13:09:24.344 [INFO ] [ome.io.rest.core.thing.ThingResource] - Received HTTP PUT request for update configuration at 'things/zwave:serial_zstick:controller/config' for an unmanaged thing 'zwave:serial_zstick:controller'.

https://community.openhab.org/t/z-wave-exclue-include-in-openhab-2-1-0-snapshot-with-habmin/26369/17

sjka:
Apparently this did not become clear enough from this PR. If you read the issue that it fixes (#1977) you will see that this change was merely about allowing bindings (i.e. ThingHandlers) to change the configuration. 

The problem I see with removing this check from the REST API is that it is primarily used for UIs. And that means it is used by users, intentionally changing some configuration. And while it would be really convenient in your case to try things out, it is, well... let's say a little surprising that the new configuration is accepted and working well at first, but suddenly gets lost after the next restart.

Issue #2630, opened 2016-12-08
[BUG] digitalSTROM Binding : dssBridge OFFLINE - CONFIGURATION ERROR since change to JsonDB storage


hmerk:
Since thing definitions are stored in JsonDB, the dssBridge is not going ONLINE after openHAB is started.
The error message in PaperUI says that the fields "Sensor update interval", "Total power update interval", "Wait time sensor reading" and "Delays to be slaked trash bin devices" have to be of type number, but aren't.
Going into editing mode and just confirming without any changes brings the dssBridge online.

@MichaelOchel kann you check this please.

hmerk:
I found the solution myself and provides aPR for it.

hmerk:
Fixed with #2685

PR #2633, opened 2016-12-08
[digitalSTROM] fix for #2630 - Configuration error introduced with js…

hmerk:
…onDB storage

### DIFF
--- a/extensions/binding/org.eclipse.smarthome.binding.digitalstrom/src/main/java/org/eclipse/smarthome/binding/digitalstrom/handler/BridgeHandler.java
+++ b/extensions/binding/org.eclipse.smarthome.binding.digitalstrom/src/main/java/org/eclipse/smarthome/binding/digitalstrom/handler/BridgeHandler.java
@@ -9,6 +9,7 @@ package org.eclipse.smarthome.binding.digitalstrom.handler;
 
 import static org.eclipse.smarthome.binding.digitalstrom.DigitalSTROMBindingConstants.*;
 
+import java.math.BigDecimal;
 import java.util.ArrayList;
 import java.util.Collections;
 import java.util.LinkedList;
@@ -199,8 +200,12 @@ public class BridgeHandler extends BaseBridgeHandler
         if (StringUtils
                 .isNotBlank(thingConfig.get(DigitalSTROMBindingConstants.SENSOR_DATA_UPDATE_INTERVAL).toString())) {
             try {
-                config.setSensordataRefreshInterval(Integer.parseInt(
-                        thingConfig.get(DigitalSTROMBindingConstants.SENSOR_DATA_UPDATE_INTERVAL).toString() + "000"));
+                Object param;
+                param = thingConfig.get(DigitalSTROMBindingConstants.SENSOR_DATA_UPDATE_INTERVAL);
+                logger.debug("param = {}", param);
+                if (param instanceof BigDecimal && param != null) {
+                    config.setSensordataRefreshInterval(((BigDecimal) param).intValue());
+                }
             } catch (NumberFormatException e) {
                 if (numberExc == null) {
                     numberExc = new ArrayList<String>();
@@ -211,8 +216,12 @@ public class BridgeHandler extends BaseBridgeHandler
         if (StringUtils
                 .isNotBlank(thingConfig.get(DigitalSTROMBindingConstants.TOTAL_POWER_UPDATE_INTERVAL).toString())) {
             try {
-                config.setTotalPowerUpdateInterval(Integer.parseInt(
-                        thingConfig.get(DigitalSTROMBindingConstants.TOTAL_POWER_UPDATE_INTERVAL).toString() + "000"));
+                Object param;
+                param = thingConfig.get(DigitalSTROMBindingConstants.TOTAL_POWER_UPDATE_INTERVAL);
+                logger.debug("param = {}", param);
+                if (param instanceof BigDecimal && param != null) {
+                    config.setTotalPowerUpdateInterval(((BigDecimal) param).intValue());
+                }
             } catch (NumberFormatException e) {
                 if (numberExc == null) {
                     numberExc = new ArrayList<String>();
@@ -222,8 +231,12 @@ public class BridgeHandler extends BaseBridgeHandler
         }
         if (StringUtils.isNotBlank(thingConfig.get(DigitalSTROMBindingConstants.SENSOR_WAIT_TIME).toString())) {
             try {
-                config.setSensorReadingWaitTime(Integer
-                        .parseInt(thingConfig.get(DigitalSTROMBindingConstants.SENSOR_WAIT_TIME).toString() + "000"));
+                Object param;
+                param = thingConfig.get(DigitalSTROMBindingConstants.SENSOR_WAIT_TIME);
+                logger.debug("param = {}", param);
+                if (param instanceof BigDecimal && param != null) {
+                    config.setSensorReadingWaitTime(((BigDecimal) param).intValue());
+                }
             } catch (NumberFormatException e) {
                 if (numberExc == null) {
                     numberExc = new ArrayList<String>();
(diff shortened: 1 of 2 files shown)


extensions/binding/org.eclipse.smarthome.binding.digitalstrom/src/main/java/org/eclipse/smarthome/binding/digitalstrom/handler/BridgeHandler.java
```diff
@@ -199,8 +200,12 @@ private Config loadAndCheckConfig() {
         if (StringUtils
                 .isNotBlank(thingConfig.get(DigitalSTROMBindingConstants.SENSOR_DATA_UPDATE_INTERVAL).toString())) {
             try {
-                config.setSensordataRefreshInterval(Integer.parseInt(
-                        thingConfig.get(DigitalSTROMBindingConstants.SENSOR_DATA_UPDATE_INTERVAL).toString() + "000"));
+                Object param;
+                param = thingConfig.get(DigitalSTROMBindingConstants.SENSOR_DATA_UPDATE_INTERVAL);
+                logger.debug("param = {}", param);
+                if (param instanceof BigDecimal && param != null) {
```
maggu2810: If you already know that it is an instance of BigDecimal, why do you the additional != null check?

maggu2810:
Why is this necessary using the JSON Db storage but not necassary using the MapDB storage?

extensions/binding/org.eclipse.smarthome.binding.digitalstrom/src/main/java/org/eclipse/smarthome/binding/digitalstrom/handler/BridgeHandler.java
```diff
@@ -199,8 +200,12 @@ private Config loadAndCheckConfig() {
...
+                    config.setSensordataRefreshInterval(((BigDecimal) param).intValue());
```
MichaelOchel: The convertion from seconds to milliseconds ( * 1000) is missing. Also at the next two config parameters.

extensions/binding/org.eclipse.smarthome.binding.digitalstrom/src/main/java/org/eclipse/smarthome/binding/digitalstrom/handler/BridgeHandler.java
```diff
@@ -199,8 +200,12 @@ private Config loadAndCheckConfig() {
...
+                }
             } catch (NumberFormatException e) {
```
MichaelOchel: The catch block is not reachable now.  Use an else-block at the `if( ... instanceof BigDecimal){ ...` for the error message e.g.

if (thingConfig.get(DigitalSTROMBindingConstants.SENSOR_DATA_UPDATE_INTERVAL) instanceof BigDecimal) {
        config.setSensordataRefreshInterval(((BigDecimal)   thingConfig.get(DigitalSTROMBindingConstants.SENSOR_DATA_UPDATE_INTERVAL)).intValue()  * 1000);
        } else {
            if (numberExc == null) {
                numberExc = new ArrayList<String>();
            }
            numberExc.add("\"Sensor update interval\" ( "
                    + thingConfig.get(DigitalSTROMBindingConstants.SENSOR_DATA_UPDATE_INTERVAL) + ")");
        }

Also at the next try-catch-blocks.

MichaelOchel:
But this will only solve the problem at the dssBridge. The same problem will occur by the other number formated configurations, like the scene id, group id and zone id at the scene handler.
I think the easiest way to fix it, it is to add a method that fixes the number String before the number will be parse, like:


private static String fixNumber(String numberString) {
        return numberString.replace(".0", "");
}


E.g. at the scene id it would look like that:
sceneID = Short.parseShort(fixNumber(configSceneID));


A BigDecimal cast is not possible here, because the names (e.g. zone name = kitchen) are also allowed.


Either you will fix this issue, or we address it to the next update. Currently we only need to adjust the readme, some little other things and a full test.


But independently of this, why the number is formatted as floating point number (e.g. 120.0) after starting the server and later on as non floating point number (e.g. 120)?

extensions/binding/org.eclipse.smarthome.binding.digitalstrom/src/main/java/org/eclipse/smarthome/binding/digitalstrom/handler/BridgeHandler.java
hmerk: Fair point, copy and paste "error". will fix it.

extensions/binding/org.eclipse.smarthome.binding.digitalstrom/src/main/java/org/eclipse/smarthome/binding/digitalstrom/handler/BridgeHandler.java
```diff
@@ -199,8 +200,12 @@ private Config loadAndCheckConfig() {
         if (StringUtils
                 .isNotBlank(thingConfig.get(DigitalSTROMBindingConstants.SENSOR_DATA_UPDATE_INTERVAL).toString())) {
             try {
-                config.setSensordataRefreshInterval(Integer.parseInt(
-                        thingConfig.get(DigitalSTROMBindingConstants.SENSOR_DATA_UPDATE_INTERVAL).toString() + "000"));
+                Object param;
+                param = thingConfig.get(DigitalSTROMBindingConstants.SENSOR_DATA_UPDATE_INTERVAL);
+                logger.debug("param = {}", param);
```
hmerk: This is a leftover from my debugging and will be removed.

extensions/binding/org.eclipse.smarthome.binding.digitalstrom/src/main/java/org/eclipse/smarthome/binding/digitalstrom/handler/BridgeHandler.java
hmerk: ok, have to fix this

extensions/binding/org.eclipse.smarthome.binding.digitalstrom/src/main/java/org/eclipse/smarthome/binding/digitalstrom/handler/BridgeHandler.java
hmerk: ok

hmerk:
We should address the question "why this happens" to @cdjackson, who wrote the JSON Db storage.
Might be that this causes more problems as we see at the moment.
@cdjackson any comments?

@MichaelOchel i will test what happens to the scenes. As I am not using them produktive, I did not think that this issue might occur there as well.

hmerk:
@MichaelOchel 
I had a quick look at zoneId and groupId, they are defined as text and therefore be stored as text. I guess this will cause no issue, but will test later.

hmerk:
@maggu2810 @MichaelOchel 
I implemented your comments in the second commit and have successfully tested thing coming online after restart.

maggu2810:
> We should address the question "why this happens" to @cdjackson, who wrote the JSON Db storage.
Might be that this causes more problems as we see at the moment.
@cdjackson any comments?

@hmerk As you found and fixed it: What is the difference between `thingConfig.get(DigitalSTROMBindingConstants.SENSOR_DATA_UPDATE_INTERVAL);` on mabdb and jsondb? So, I assume you got different stuff returned and your changes does now handle it on both cases correctly. What have you received before and what now?

wborn:
I also have issues with the new JSON DB storage. Sometimes integer values that are stored in `org.eclipse.smarthome.core.thing.Thing.json` turn into decimals. I reported them [here](https://community.openhab.org/t/replacement-of-mapdb-by-jsondb/17498/16?u=wborn) and thought they were gone. But they are not. It's just not always reproducible.

Today I reproduced the issue with OH2 build *#*641. I can reproduce it by adding some LIFX lights in the Paper UI (via the Inbox) and then change the integer `fadetime` or `powerOnBrighteness` (on the color channel) configuration properties (using the spinners). The issue does not always occur. Sometimes I need to restart OH2 and modify the properties before it occurs.

So I also think this PR is a workaround for a bigger problem introduced by the JSON DB storage.

wborn:
Today I had a look at the code. I think it is not necessarily an issue that properties are stored as `BigDecimal`s with JSON DB. 

`ConfigUtil.normalizeTypes(configuration, configDescription)` should be able to resolve such issues. But it is easy to create bugs by not using it at the right place. There are various methods on `Configuration` that allow to create/modify a binding/channel `Configuration` so that it does not have normalized types. 

Also I expanded the jsondb/mapdb unit tests and both returned a `BigDecimal` when storing an integer. But I do not know how well my tests simulated the actual behavior.

So it can be a bit misleading that we can now actually see how Things are persisted in the .json files.

kaikreuzer:
@hmerk Do I see it right that this PR is still waiting on you? Please let me know as soon as you think it is ready to be merged!

hmerk:
@kaikreuzer there were no more comments after the second commit so I could finaly squash the two.
I did not find time to test my changes against mapdb storage to answer @maggu2810 's question.
But is this really relevant, as there is no option to choose the storage.

wborn:
I think this PR is no longer necessary now that #2685 has been merged?

hmerk:
Yes, I can confirm that #2665 seems to have fixed the issue for digitalSTROM Binding as well. I can see my devices coming online without any problem.

PR #2641, opened 2016-12-11
Correct ERROR log messages in HTTP action

watou:
It seems that the intent was to print a string message of the exception, but since the last argument is the exception its stack trace is being dumped instead.

Noticed on the forum [here](https://community.openhab.org/t/http-binding-can-the-logging-be-tamed/2764/5?u=watou).


watou:
If this doesn't pass the "minimal change exception" for PRs, I can sign off on a new commit.

kaikreuzer:
Yeah, please sign it off, would be the cleanest - thanks!

watou:
passed ip-validation now

kaikreuzer:
perfect.

PR #2643, opened 2016-12-11
"ü" is not correctly displayed in the browser

### LINKED ISSUE #2639: Use of UTF-8 encoding for i18n properties files
While testing the Kodi binding with the german translation properties file I realized that all german umlauts where not displayed correctly. 
The same problem appears in the Yahoo binding's description:
image
The reason is simple: the properties file where the texts are stored uses encoding ISO 8859-1. 
It seems that all properties files (in Smarthome and openHAB) are using this encoding while all Java files are using UTF-8.  
The german properties file of the Hue binding uses an interesting solution: it contains a Unicode constante for the "ü":
```
Es wurde keine IP Adresse f\u00FCr die Hue Bridge angegeben.
```
Is there any special reason why all properties files are stored with ISO 8859-1? In my opinition it would be better to use generally UTF-8. Otherwise all bindings would have to use unicode constants to get correctly displayed umlauts and other international characters.
For the Kodi binding I have changed the encoding to UTF-8, corrected all umlautes and everything was displayed correctly.

MHerbst:
Therefore it was replaced by the equivalent UTF-8 constant.

Fixes #2639


kaikreuzer:
Closing due to [this discussion](https://github.com/eclipse/smarthome/issues/2639#issuecomment-266389552).

Issue #2640, opened 2016-12-10
Adding an Item in PaperUI fails with no error if Name field contains a space. 


individualist:
To Reproduce: 
Under Configuration->Items click (+) to add an item. 
Fill in all required fields, but in the "Name" field supply a name containing a space, such as "Office Lights" 
Click (+) to add the item.   

Results: You are returned to the Items list without an error and without the item being added.   

Expected results:  
If spaces are not allowed in item names,  a message to that effect should be given, and/or the user not allowed to continue.

kaikreuzer:
@aounhaider1 Is the item name validation broken?

PR #2644, opened 2016-12-12
Itemname allow only alphanumeric chars and underscore

aounhaider1:
fixes https://github.com/eclipse/smarthome/issues/2640


kaikreuzer:
Thanks.
Question: When creating a name suggestion from the Thing/Channel label, do you make sure to follow this rule, i.e. do you remove any other character?

aounhaider1:
Yes, the only allowed chars in that case are a-z,A-Z,0-9 and _. Everything else is removed from the name suggestion.

kaikreuzer:
perfect, thanks.

PR #2646, opened 2016-12-12
updated is called async, sync. access to broker connections

maggu2810:
Signed-off-by: Markus Rathgeb <maggu2810@gmail.com>


maggu2810:
The updated method did not use the `getConnection` method.
If a consumer of the services calls one of the register function at the same time the updated method is called by the OSGi framework two different MqttBrokerConnection could be created and only one is used.
If the updated method wins, the registration is lost and no callback of the listening interface is called.

maggu2810:
I also moved the logic that access the member `brokerConnections` to separate functions, so they could be commented why at some places a synchronization is not necessary.

PR #2657, opened 2016-12-14
[LIFX] Power state bug fixes and code cleanup

wborn:
### Power state bug fixes
* "Power on brightness" was not properly applied when switching ON/OFF White lights. There actually is no difference between switching ON/OFF Color and White lights.
* When switching ON/OFF a light, the brightness should not be changed with `handleHSBCommand` because it always switches a light ON when it is already OFF. I.e. a light that is OFF, and switched OFF again, could go ON and immediately OFF again.
* Power state polling of lights did not work properly. A `GetLightPowerRequest` was sent which results in a `StateLightPowerResponse` and not the `StatePowerResponse` that was used for updating the `PowerState`. Because the power state is also part of the `StateResponse` it is unnecessary to poll for it with a second packet.
* Because power state polling now works properly, issues in the `CurrentLightState` showed up, hence the reworked logic.
* `MAX_STATE_CHANGE_DURATION` has been increased because sometimes power state change packets are not received within this interval. As a result lights were not properly switched ON/OFF because the `CurrentLightState` still contained an old value.
* Because of the above, `Switch`-es and `Dimmer`s bound to the `color`/`brightness` channels now update properly. Also when changing the state of a light via other means such as the LIFX App or a second ESH instance.

### Code cleanup
* Parameters/variables have been renamed for more clarity/consistency.
* Commented code in protocol objects has been cleaned up.
* Missing `expectedResponses` have been added in protocol objects.

### DIFF
--- a/extensions/binding/org.eclipse.smarthome.binding.lifx/ESH-INF/binding/binding.xml
+++ b/extensions/binding/org.eclipse.smarthome.binding.lifx/ESH-INF/binding/binding.xml
@@ -5,7 +5,7 @@
         xsi:schemaLocation="http://eclipse.org/smarthome/schemas/binding/v1.0.0 http://eclipse.org/smarthome/schemas/binding-1.0.0.xsd">
 
     <name>LIFX Binding</name>
-    <description>This is the binding for LIFX bulbs.</description>
+    <description>This is the binding for LIFX lights.</description>
     <author>Karel Goderis</author>
 
 </binding:binding>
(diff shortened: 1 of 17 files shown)


kaikreuzer:
Wow, you are really eager on this binding - thanks!

wborn:
Thanks for reviewing/merging! The binding is almost feature complete and bug free for my uses now. :-)

PR #2658, opened 2016-12-16
reduce severity in io.upnp

sjka:
in order to avoid flooding the logs with "Invoking Action [...]" messages.

They are pretty much spamming the logs with messages like this, e.g. when using sonos:

```
Invoking Action 'GetRunningAlarmProperties' of service 'AVTransport' for participant 'RINCON_[...]'
```

Unless anybody has a good reason to keep them on DEBUG level, I would suggest to reduce them to TRACE, so if we need them we can specifically enable them.


kaikreuzer:
fine for me.

PR #2661, opened 2016-12-16
correctly handle arrays and collections in ConfigurationServices

kaikreuzer:
fixes https://github.com/openhab/openhab2-addons/issues/1575


kaikreuzer:
@maggu2810 This has some priority for me as it came up after rolling out a new service that requires that feature, i.e. it is a kind of "production bug". Would be nice if it could be handled with priority - thanks :-)

kaikreuzer:
Thanks!

PR #2664, opened 2016-12-16
Consistent bundle name for bindings

lolodomo:

### DIFF
--- a/extensions/binding/org.eclipse.smarthome.binding.digitalstrom/META-INF/MANIFEST.MF
+++ b/extensions/binding/org.eclipse.smarthome.binding.digitalstrom/META-INF/MANIFEST.MF
@@ -1,6 +1,6 @@
 Manifest-Version: 1.0
 Bundle-ManifestVersion: 2
-Bundle-Name: DigitalSTROM Binding
+Bundle-Name: Eclipse SmartHome DigitalSTROM Binding
 Bundle-SymbolicName: org.eclipse.smarthome.binding.digitalstrom;singleton:=true
 Bundle-Vendor: Eclipse.org/SmartHome
 Bundle-Version: 0.9.0.qualifier
--- a/extensions/binding/org.eclipse.smarthome.binding.fsinternetradio/META-INF/MANIFEST.MF
+++ b/extensions/binding/org.eclipse.smarthome.binding.fsinternetradio/META-INF/MANIFEST.MF
@@ -1,6 +1,6 @@
 Manifest-Version: 1.0
 Bundle-ManifestVersion: 2
-Bundle-Name: FSInternetRadio Binding
+Bundle-Name: Eclipse SmartHome FSInternetRadio Binding
 Bundle-SymbolicName: org.eclipse.smarthome.binding.fsinternetradio;singleton:=true
 Bundle-Vendor: Eclipse.org/SmartHome
 Bundle-Version: 0.9.0.qualifier
--- a/extensions/binding/org.eclipse.smarthome.binding.ntp/META-INF/MANIFEST.MF
+++ b/extensions/binding/org.eclipse.smarthome.binding.ntp/META-INF/MANIFEST.MF
@@ -1,6 +1,6 @@
 Manifest-Version: 1.0
 Bundle-ManifestVersion: 2
-Bundle-Name: ntp Binding
+Bundle-Name: Eclipse SmartHome NTP Binding
 Bundle-SymbolicName: org.eclipse.smarthome.binding.ntp;singleton:=true
 Bundle-Vendor: Eclipse.org/SmartHome
 Bundle-Version: 0.9.0.qualifier
--- a/extensions/binding/org.eclipse.smarthome.binding.sonos/META-INF/MANIFEST.MF
+++ b/extensions/binding/org.eclipse.smarthome.binding.sonos/META-INF/MANIFEST.MF
@@ -1,6 +1,6 @@
 Manifest-Version: 1.0
 Bundle-ManifestVersion: 2
-Bundle-Name: Sonos Binding
+Bundle-Name: Eclipse SmartHome Sonos Binding
 Bundle-SymbolicName: org.eclipse.smarthome.binding.sonos
 Bundle-Vendor: Eclipse.org/SmartHome
 Bundle-Version: 0.9.0.qualifier
--- a/extensions/binding/org.eclipse.smarthome.binding.yahooweather/META-INF/MANIFEST.MF
+++ b/extensions/binding/org.eclipse.smarthome.binding.yahooweather/META-INF/MANIFEST.MF
@@ -1,6 +1,6 @@
 Manifest-Version: 1.0
 Bundle-ManifestVersion: 2
-Bundle-Name: YahooWeather Binding
+Bundle-Name: Eclipse SmartHome YahooWeather Binding
 Bundle-SymbolicName: org.eclipse.smarthome.binding.yahooweather;singleton:=true
 Bundle-Vendor: Eclipse.org/SmartHome
 Bundle-Version: 0.9.0.qualifier


lolodomo:
With this PR, the bundle name for any binding is starting with "Eclipse SmartHome".

kaikreuzer:
This was done deliberately, because within every solution you will have a mix of ESH bindings and solution-specific bundles. Having different names would make it very ugly in your bundle list.

lolodomo:
@kaikreuzer : please don't close so fast.
My intention was to have consistent names for all Eclipse SmartHome bindings.
Why we have a bundle named "Eclipse SmartHome hue Binding" and another one named "Sonos Binding" ? Where is the consistency ?
```

openhab> bundle:list | grep Binding
 98 | Active   |  80 | 0.9.0.201612091054    | Eclipse SmartHome AutoUpdate Binding
 99 | Active   |  80 | 0.9.0.201612091054    | Eclipse SmartHome Core Binding XML
174 | Active   |  80 | 0.9.0.201612091054    | ntp Binding
175 | Active   |  80 | 2.0.0.201612091712    | Astro Binding
180 | Active   |  80 | 1.9.0.201612100210    | openHAB MiOS Binding
184 | Active   |  80 | 1.9.0.201612100210    | openHAB Weather Binding
198 | Active   |  80 | 0.9.0.201612091054    | Eclipse SmartHome hue Binding
200 | Active   |  80 | 2.0.0.201612091712    | Rfxcom Binding
202 | Active   |  80 | 2.0.0.201612091712    | Freebox Binding
208 | Active   |  80 | 2.0.0.201612092335    | Netatmo Binding
209 | Active   |  80 | 2.0.0.201609101044    | Powermax Binding
210 | Active   |  80 | 0.9.0.201612112336    | Sonos Binding
```

kaikreuzer:
The consistent one would be for now without the prefix, see https://github.com/openhab/openhab2-addons/issues/765#issuecomment-267753746

maggu2810:
> The consistent one would be for now without the prefix, see openhab/openhab2-addons#765 (comment)

I would prefer to keep "Eclipse SmartHome" in front of Eclipse SmartHome stuff!

lolodomo:
Yes, everything ellse seems to be prefixed by "Eclipse SmartHome" except 5 bindings.
I am changing only the name of 5 bindings...

maggu2810:
-- updated my comment above --

lolodomo:
And of course I could update my PR to correct the archetype too.

kaikreuzer:
Sorry, I didn't mean to be rude.
What I tried to explain in https://github.com/openhab/openhab2-addons/issues/765#issuecomment-267753746 is that the current situation is not a bug that should be fixed, but it was a deliberate decision to do it that way when the archetype was created.
Note that the archetype is also meant to be used for any not-Eclipse-hosted binding, so prefixing it with Eclipse SmartHome would be wrong. Ok, we could introduce a new parameter that needs to be set for the archetype, which would provide such a prefix.
A reason for leaving a prefix out of the picture was also to make it easy to move bindings to ESH. Actually, openhab2-addons repo only exists to make it easier for people to get started to contribute and have bindings merged. I could have also asked everyone to directly contribute them to ESH, but this would have meant a huge overhead, which the ESH committers couldn't have dealt with. Nonetheless, the idea was to say at any time that a binding should be moved over (whenever it is relevant for a commercial solution) and the only change should be the package name - a UI facing name (let's consider the console with the BSN a UI) should stay identical to before. Such a move was e.g. done with the Sonos binding, which first lived in OH2 and now in ESH. Note that the same naming scheme (i.e. no ESH prefix) is also used on [some other add-ons](https://github.com/eclipse/smarthome/blob/master/extensions/voice/org.eclipse.smarthome.voice.mactts/META-INF/MANIFEST.MF#L3).

I am not saying that it has to stay like this, I only wanted to provide some background, why things are as they are right now.

Thinking about it, I think we might really make a differerent between core bundles and everything with in the "extensions" folder - in the end, bundles in extensions are a pool of bundles that will be mixed in any ESH-solution with "custom" bundles of the solution, while from a user perspective, they should appear the same way, so I would argue that prefixing is counter-productive.
With an ESH 1.0 release slowly coming closer, I actually think if we have to go a much bigger step and separate the extensions folder out and remove it from the main repo - the versioning lifecycle of the extensions might be pretty different to the one of the core bundles, once 1.0 has to be stable and not change anymore...

lolodomo:
If you don't want a prefix, you have to suppress it from few bindings (like the hue binding for example). Then at least it will be consistent.
PS: note that all 1.x bindings have a prefix (openHAB)

kaikreuzer:
> If you don't want a prefix, you have to suppress it from few bindings (like the hue binding for example). Then at least it will be consistent.

Correct, the hue binding currently is indeed buggy in this sense.

> PS: note that all 1.x bindings have a prefix (openHAB)

Right, because OH1 had it much simpler as it was only ONE project and not split into ESH+OH.

> If you don't want a prefix

If you guys want to convince me that one should be there (and yes, I see arguments for it), I only wanted to highlight that the archetype is the source of everything, so we would have to discuss a solution for it. But I am not too keen on it as I feel that there are more pressing topics at the moment ☹️

lolodomo:
The archetype is not a real problem, as you already said, we could add a new parameter.
But your other arguments make a certain sense.
@maggu2810 : are you ok ?

maggu2810:
> Sorry, I didn't mean to be rude.

Me, too. So sorry about that if this is your feeling.

> I am not saying that it has to stay like this, I only wanted to provide some background, why things are as they are right now.

Okay, then this was really a misunderstanding.
I was not aware that this is the way how it is currently handled.

> a UI facing name (let's consider the console with the BSN a UI) should stay identical to before
> while from a user perspective, they should appear the same way, so I would argue that prefixing is counter-productive.

I would like to see (in such a low level user interface as the console) a BSN where I can differ between e.g. the binding for X provided by the Eclipse SmartHome project and the (e.g. proprietary) binding for X provided by some other guys.
I would also like to read the BSN and be able to differ between SmartHome bundles and other ones (Jetty, Apache Felix, ...).
If I have hundreds of bundles in the OSGi container, so also non ESH related ones, IMHO e.g. "Hue binding" is not enough.
If Jetty and Apache decide to call its BSN of their http client implementation only "HTTP Client" it does not help. So please let's keep some prefix.
For end-user interfaces (I assume the end-user of a product itself is not interested in the correct BSN) it is pretty easy to replace all prefixes by some other ones (or remove it).
I don't think the BSN should be chosen to hide the information if it is an Eclipse SmartHome binding or an openHAB one. If this should be not visible to the user, we could remove "Eclipse SmartHome " and "openHAB " in the BSN before presenting it (how you handle it in openHAB -- if there is something to remove, is up to the openHAB project).

> With an ESH 1.0 release slowly coming closer, I actually think if we have to go a much bigger step and separate the extensions folder out and remove it from the main repo - the versioning lifecycle of the extensions might be pretty different to the one of the core bundles, once 1.0 has to be stable and not change anymore...

I agree.
I assume we also need to think about if some stuff that is currently part of bundles should be moved to extensions.

> If you guys want to convince me that one should be there (and yes, I see arguments for it), I only wanted to highlight that the archetype is the source of everything, so we would have to discuss a solution for it. 

I was not happy about that this PR has been closed without any discussion.
But as written above, I wasn't aware that "we don't use Eclipse SmartHome" as prefix for extensions is the current situation.
My understanding has been we already used that prefix for all bundles as we do it for the artifact names, too.

> But I am not too keen on it as I feel that there are more pressing topics at the moment.

I think we have all enough to do and different persons priories different stuff differently.
IMHO it is important to add stuff like this to the guidelines (package name, BSN, Bundle Vendor, etc.) as it is much easier to do it right from the beginning then fix it later.

lolodomo:
Waiting for further discussion, we could suppress the prefix for the LIFX binding, the Hue binding and the Wemo Binding ?
Note that most of the other extensions (io, ui, transform and voice) have a prefix "Eclipse SmartHome". The prefix will be missing only for the bindings.

PR #2665, opened 2016-12-17
Normalize configuration values when initializing thing handlers

wborn:
On startup stored integer configuration parameters are not normalized. As a result objects with decimals are returned when retrieving these `Configuration` values. This causes exceptions. 

E.g. with the LIFX binding the following warning is logged on startup:

```
18:52:28.940 [WARN ] [inding.lifx.handler.LifxLightHandler] - Invalid value '300.0' for transition time, using default instead.
```

The definition of this parameter is:

```
    <parameter name="fadetime" type="integer" required="false">
        <label>Fade time</label>
        <description>The time to fade to the new color value (in ms).</description>
        <default>300</default>
    </parameter>
```

In this case the code is expecting a Long and not a BigDecimal:

```
    try {
        return Long.parseLong(fadeCfg.toString());
    } catch (NumberFormatException e) {
        logger.warn("Invalid value '{}' for transition time, using default instead.", fadeCfg.toString());
        return FADE_TIME_DEFAULT;
    }
```

Other people have reported similar issues on the [openHAB community forums](https://community.openhab.org/t/replacement-of-mapdb-by-jsondb/17498/30). In PR #2633 a workaround is created for the digitalSTROM binding.

I also did a bit of [research](https://github.com/eclipse/smarthome/pull/2633#issuecomment-266820822) on this issue. The root cause I think is that `Gson` deserializes all integers in a Thing Configuration as a `BigDecimal`. Because this is done using reflection the configuration parameters are never validated/normalized.

This PR addresses the issue by normalizing the Configuration values at the same location where the Configuration defaults are also provided. I've tested it and for me it solves the issue with the LIFX binding. I think it should also fix the issues in other bindings.

### DIFF
--- a/bundles/core/org.eclipse.smarthome.core.thing/src/main/java/org/eclipse/smarthome/core/thing/internal/ThingManager.java
+++ b/bundles/core/org.eclipse.smarthome.core.thing/src/main/java/org/eclipse/smarthome/core/thing/internal/ThingManager.java
@@ -30,6 +30,7 @@ import org.eclipse.smarthome.config.core.BundleProcessorVetoManager;
 import org.eclipse.smarthome.config.core.ConfigDescription;
 import org.eclipse.smarthome.config.core.ConfigDescriptionParameter;
 import org.eclipse.smarthome.config.core.ConfigDescriptionRegistry;
+import org.eclipse.smarthome.config.core.ConfigUtil;
 import org.eclipse.smarthome.config.core.Configuration;
 import org.eclipse.smarthome.core.common.SafeMethodCaller;
 import org.eclipse.smarthome.core.common.ThreadPoolManager;
@@ -638,6 +639,7 @@ public class ThingManager extends AbstractItemEventSubscriber implements ThingTr
                     synchronized (thing) {
                         if (!isInitializing(thing)) {
                             ThingType thingType = getThingType(thing);
+                            normalizeConfiguration(thing, thingType);
                             applyDefaultConfiguration(thing, thingType);
                             if (isInitializable(thing, thingType)) {
                                 setThingStatus(thing,
@@ -657,6 +659,17 @@ public class ThingManager extends AbstractItemEventSubscriber implements ThingTr
                     }
                 }
             });
+
+    private void normalizeConfiguration(Thing thing, ThingType thingType) {
+        Configuration configuration = thing.getConfiguration();
+        if (configuration != null && thingType != null) {
+            ConfigDescription configDesc = configDescriptionRegistry
+                    .getConfigDescription(thingType.getConfigDescriptionURI());
+            if (configDesc != null) {
+                configuration.setProperties(ConfigUtil.normalizeTypes(configuration.getProperties(), configDesc));
+            }
+        }
+    }
 
     private void applyDefaultConfiguration(Thing thing, ThingType thingType) {
         if (thingType != null) {


wborn:
@maggu2810 Yes those are the same concerns I had. It would be nice when it can be done earlier in the call hierarchy when a Thing is deserialized. To me it seemed objects earlier in the call hierarchy are to generic and do not necessarily have all the information.

That integers are being serialized/deserialized with decimals seems to be a common issue when using Gson. See also some of the following Gson issues: 
* https://github.com/google/gson/issues/968
* https://github.com/google/gson/issues/559
* https://github.com/google/gson/issues/525

kaikreuzer:
> In this case the code is expecting a Long and not a BigDecimal

You mean it is expecting an integer, not a floating point number. The normalizer actually will return you a BigDecimal instance - but it will set its scale to 0, so that it is clear that there are no fractional digits.

I'd actually consider this a bug in the JSONStorage. If a BigDecimal is reached in with scale 0, it should be written to the JSON as an int (i.e. "300") and not as a float ("300.0"). From this serialization, it should be possible to also reconstruct a BigDecimal with the correct scale - unfortunately, https://github.com/google/gson/issues/968 indeed seems to always return 300.0, so the storage has no chance to see what was REALLY within the JSON. Maybe one could do a check whether there are fractional digits and set the scale to 0, if not. @cdjackson Any opinion?

cdjackson:
It might be possible. I did a quick search on this a week or so back and I think there is a way to override the (de)serialisation methods used in gson so it might be possible to do something here. I’ve not followed why it seems to be random (i.e. sometimes it writes out with no decimals and other times it writes with decimals (this is what’s happening, right?).

Let me take a look to see if I can do something this afternoon.

kaikreuzer:
> (i.e. sometimes it writes out with no decimals and other times it writes with decimals (this is what’s happening, right?)

My understanding is that the de-serialization is the main problem as it is always done as a "double".

wborn:
Yes I've also seen it deserialize 300 in the .json file  to 300.0 in the Configuration .

cdjackson:
Not in my json files - I have no decimal places - at least none that I can see from a quick scroll through the files…

kaikreuzer:
300.0 in Configuration object, not in JSON. We are talking about DE-serialization.

cdjackson:
Ok - sorry - I thought you mentioned this in the json…

So, some code I copied from elsewhere might be screwing this - it is forcing BigDecimal in the deserialisation.

        if (JsonToken.NUMBER.equals(in.peek())) {
            value = new BigDecimal(in.nextString());
        } else {
            value = valueAdapter.read(in);
        }

This needs to be changed.

I propose on serialisation, to force one decimal place for BigDecimal, and then on deserialisation we can tell the difference.

kaikreuzer:
> I propose on serialisation, to force one decimal place for BigDecimal

I propose to serialize it with the scale that is given in the BigDecimal. If we have a configuration that clearly holds integers, the scale should always be 0. Otherwise you do not stand a chance to recognize those again on deserialization (because you could also have a 300.0000 as a very precise temperature measurement and this should also be de-serialized with this precision).

cdjackson:
Sorry - I wrote that wrong. I meant for BigDecimal it should have a minimum of 1 decimal place. I didn’t mean to reduce precision.

However, that was based on an assumption that config values can be Integer and that might be wrong (it’s been a while since I looked at that). What Java types are supported for numbers? Is everything BigDecimal?

kaikreuzer:
Yes, everything is BigDecimal. If a config description says "INTEGER", it will be a BigDecimal with scale=0, if it says "NUMBER", it will be a BigDecimal with decimal digits.

cdjackson:
Ok - I’ll take a look at this today….

cdjackson:
This could be "solved" with the following code -:

```java
            if (inString.endsWith(".0")) {
                value = new BigDecimal(inString).setScale(0);
            } else {
                value = new BigDecimal(inString);
            }
```

The problem is that we might still change the value such that the original may not be what the source expected if it already wanted a number with x.0.

kaikreuzer:
Is "inString" the raw value from the JSON? As said above, the serialization is correctly doing a 300 (without ".0"), see [here](https://community.openhab.org/t/replacement-of-mapdb-by-jsondb/17498/30?u=kai), so this fix won't work. But if you have that raw value, you could simply check whether it is an integer and if so, set the scale to 0.

cdjackson:
> Is "inString" the raw value from the JSON?

No - as above, there doesn't seem to be any way to actually access the raw string - it already has the .0 on the end when it is presented to the converter !! (which is really stupid - the converter should have access to raw data!).

> see here, so this fix won't work.

Not sure I understand your point. If I read back 8080.0, and remove the .0, I end up with 8080 - isn't that what you need?

> But if you have that raw value, you could simply check whether it is an integer and if so, set the scale to 0.

So I don't have the raw value, but other than that, it's what I'm doing I think?

kaikreuzer:
>  If I read back 8080.0, and remove the .0, I end up with 8080 - isn't that what you need?

Yes. My assumption was that you were talking about the raw data and this might not have had the .0.
But as you clarified this, I agree that this should be ok (although the approach is a bit too eager, as it will also convert a serialized (raw) 80.0 to scale=0, while it should be scale=1).

cdjackson:
> although the approach is a bit too eager, as it will also convert a serialized (raw) 80.0 to scale=0, while it should be scale=1

Yes - this is what I meant above if the source is expecting the x.0. I'm not sure there's any way around that though given there's no access to source data. Unless something changes in gson, we gotta choose I think.

kaikreuzer:
Imho, this is ok, since we do not really use the scale information anywhere (apart from identifying integers).

kaikreuzer:
@wborn Assuming that @cdjackson's fix in the JSONStorage will solve the problem, I suggest to close this PR.

maggu2810:
Hm, I am a little bit confused. Isn't it really not possible to keep the scale using Gson?
I tried a little example myself:

```java
package tmp;

import java.math.BigDecimal;

import com.google.gson.Gson;

public class Test {
    private static class Simple {
        final private BigDecimal bd;

        public Simple(final BigDecimal bd) {
            this.bd = bd;
        }

        @Override
        public String toString() {
            return "Simple [bd=" + bd + "]";
        }

    }

    public static void main(String[] args) {
        final Gson gson = new Gson();

        for (int i = 0; i < 3; ++i) {
            final Simple in = new Simple(new BigDecimal("8").setScale(i));
            final String json = gson.toJson(in, Simple.class);
            final Simple out = gson.fromJson(json, Simple.class);
            System.out.println(String.format("in: %s; out: %s; json: %s", in, out, json));
        }
    }
}
```

This gives me this output:

```text
in: Simple [bd=8]; out: Simple [bd=8]; json: {"bd":8}
in: Simple [bd=8.0]; out: Simple [bd=8.0]; json: {"bd":8.0}
in: Simple [bd=8.00]; out: Simple [bd=8.00]; json: {"bd":8.00}
```

maggu2810:
Is it about that we don't know the class that is deserialized and so need to convert it to the correct class ourself?

cdjackson:
Yes - it works fine if we’re converting to a BigDec etc, but in this case it’s converting to an Object so Gson doesn’t know what to do.

maggu2810:
Can you be patient and explain me what is different between the following and the logic that is done at runtime:

```java
    public static void main(String[] args) {
        final JsonStorage<Object> storage = new JsonStorage<>(new File("/tmp/storage/test.json"), null, 0, 0, 0);

        for (int i = 0; i < 3; ++i) {
            final Simple in = new Simple(new BigDecimal("8").setScale(i));
            storage.put("s" + i, in);
        }

        for (int i = 0; i < 3; ++i) {
            final Object out = storage.get("s" + i);
            System.out.println("out: " + out);
        }
    }
```

maggu2810:
Ah okay, I assume because, is it still known that the member of Simple is a BigDecimal.

cdjackson:
I’ve not checked the source of Gson - I started to look through it, but decided to stop. I assume it’s using reflection to look at the classes, so it knows that the inner type is a BigDecimal and responds appropriately. If the inner class is Object, and you store a BigDecimal in, and then try and read back, I would think it won’t work.

maggu2810:
You are right, thanks for your explanation.

watou:
I think the root problem in Gson is that when [ObjectTypeAdapter encounters a NUMBER](https://github.com/google/gson/blob/master/gson/src/main/java/com/google/gson/internal/bind/ObjectTypeAdapter.java#L79) it calls [`JsonReader.nextDouble()`](https://github.com/google/gson/blob/master/gson/src/main/java/com/google/gson/stream/JsonReader.java#L876), so now all you have is a `Double`, and its normal string representation of whole numbers [always ends in `.0`](https://docs.oracle.com/javase/7/docs/api/java/lang/Double.html#toString(double)).

If this is correct, Gson would need to replace uses of `JsonReader.nextDouble()` with a method like `JsonReader.nextBigDecimal()` and many other related changes.

cdjackson:
No - thats not the problem. The problem is somewhere earlier in the system. We already override the TypeAdapter to change the way a NUMBER is processed, but the problem is that even in this processing, the ‘raw’ text (well, the string that is available at that point) has already been changed to have the x.0 added to the end.

maggu2810:
@cdjackson As I never really used Gson before, I have some interest in this topic.

I modified your JSON storage and added a type adapter for the (de)serialization of the Configuration object.
The type adapter serialize the class of the value (type object) of the map, too. This information is used on deserialization.

```java
public class ConfigurationTypeAdapter implements JsonSerializer<Configuration>, JsonDeserializer<Configuration> {

    private static final String KEY = "key";
    private static final String VALUE_CLASS = "valueClass";
    private static final String VALUE = "value";

    @Override
    public Configuration deserialize(JsonElement json, Type typeOfT, JsonDeserializationContext context)
            throws JsonParseException {
        final Configuration configuration = new Configuration();
        final JsonArray jsonArray = json.getAsJsonArray();
        for (final JsonElement jsonEle : jsonArray) {
            final JsonObject jsonEntry = jsonEle.getAsJsonObject();
            final String key = jsonEntry.get(KEY).getAsString();
            Class valueClass;
            try {
                final String tmp = jsonEntry.get(VALUE_CLASS).getAsJsonPrimitive().getAsString();
                valueClass = Class.forName(tmp);
                final Object value = context.deserialize(jsonEntry.get(VALUE), valueClass);
                configuration.put(key, value);
            } catch (ClassNotFoundException e) {
                e.printStackTrace();
            }
        }
        return configuration;
    }

    @Override
    public JsonElement serialize(Configuration src, Type typeOfSrc, JsonSerializationContext context) {
        final JsonArray jsonArray = new JsonArray();
        for (final Map.Entry<String, Object> entry : src.getProperties().entrySet()) {
            final JsonObject jsonEntry = new JsonObject();
            jsonEntry.addProperty(KEY, entry.getKey());
            jsonEntry.addProperty(VALUE_CLASS, entry.getValue().getClass().getName());
            jsonEntry.add(VALUE, context.serialize(entry.getValue()));
            jsonArray.add(jsonEntry);
        }
        return jsonArray;
    }
}
```

Register type adapter:
```java
        final GsonBuilder builder = new GsonBuilder().registerTypeAdapterFactory(new PropertiesTypeAdapterFactory())
                .setPrettyPrinting();
        builder.registerTypeAdapter(Configuration.class, new ConfigurationTypeAdapter());
        this.mapper = builder.create();
```

Simple test:
```java
    public static void main(String[] args) {
        final JsonStorage<Object> storage = new JsonStorage<>(new File("/tmp/storage/test.json"), null, 0, 0, 0);
        final Configuration in = new Configuration();

        for (int i = 0; i < 3; ++i) {
            final BigDecimal value = new BigDecimal("8").setScale(i);
            in.put("s" + i, value);
        }

        storage.put("conf", in);
        final Object out = storage.get("conf");
        final Configuration confOut = (Configuration) out;
        for (final Map.Entry<String, Object> entry : confOut.getProperties().entrySet()) {
            System.out.println(String.format("key: %s, value: %s", entry.getKey(), entry.getValue()));
        }
        System.out.println("out: " + out);
        System.exit(0);
    }
```
```

Result:
```text
key: s0, value: 8
key: s1, value: 8.0
key: s2, value: 8.00
```

Using such a type adapter should keep the scale (and the real value object) as it has been before.
The current type adapter could handle the new deserialized type only but it should be easy to change the code, so we could deserialize the old and the new one.
WDYT?

cdjackson:
This could be a better solution. I was trying to avoid using JsonSerialiser as the docs state -:
New applications should prefer TypeAdapter <https://google.github.io/gson/apidocs/com/google/gson/TypeAdapter.html>, whose streaming API is more efficient than this interface's tree API.

Maybe it’s not a big issue though (?? - I’m no Gson expert!) and it might be better to use JsonSerialiser. I don’t know what the performance/memory penalties are with the two options - maybe if you have a working example (which you seem to) you could run a quick comparison with a large(ish) json file?

maggu2810:
I could test the efficiency, at least its about to store the real value class and use that if we deserialize it.
So, if all information is present in the deserialized JSON string it should not matter if we use the one or the other interface (hopefully).

I will try to implement a deserialize function that could handle the old and the new layout first (using JsonSerialiser) and later inspect the TypeAdapter interface.

cdjackson:
Sounds good. The problem with the TypeAdapter interface is it doesn’t have access to the raw data - if you inspect the string that is available, it’s not the actual string from the file - eg. the file will have 60, but what you will see is 60.0 already.

watou:
[Gson docs say](https://google.github.io/gson/apidocs/com/google/gson/Gson.html#fromJson-java.io.Reader-java.lang.reflect.Type-) that you should deserialize with the help of a TypeToken and not a Class in order to work around type erasure of generic types, but [your code](https://github.com/eclipse/smarthome/blob/master/bundles/storage/org.eclipse.smarthome.storage.json/src/main/java/org/eclipse/smarthome/storage/json/JsonStorage.java#L213) is using a Class.  Would it make any difference to use a `fromJson(Reader, Token)` instead of `fromJson(Reader, Class)`?

maggu2810:
Can a configuration hold a generic type? I assume the types a configuration can hold is very limited (Boolean, String, BigDecimal). Missed I something?

maggu2810:
But if it is possible to keep more information it should be done. Will have a look at. Thanks

maggu2810:
@cdjackson Can you give it a try? https://github.com/maggu2810/smarthome/tree/json-conf

cdjackson:
Do we really need to store the class here? @maggu2810 did you find a way to actually read the raw string data in the end - I’m thinking not, but I can be hopeful :)

I just feel it’s a shame that we have to add the class “just" to work around the limitations of Gson. We have worked without this in the system so far in other interfaces where json is used (i.e. on the REST interface) so it’s a shame to have to introduce it here. The other issue I have with this is if we’re not careful, we might end up having to add adaptors for other classes - i.e. the system is now not a generic store which is a shame. IMHO the storage should be able to store ‘anything’ and it would be a shame if we had to process specific classes with specific adaptors - it just feels a bit wrong and I would suggest we should avoid it if it’s possible…

Maybe I’m straying off topic, but why are we using Gson (I used it in json storage as it was already used in ESH, but maybe it’s not the best option)? What is used for serialising the REST data (Jackson?)? Maybe we should use this if possible as it doesn’t seem to have the same problem (at least I don’t think so?).

maggu2810:
You are using the word `shame` very often. 😉 
I agree that we should find an option to drop the usage of class information by custom type adapters manually.

Perhaps I am wrong -- need to inspect this again, but:
The time I started with the Configuration type adapter I created a JsonStorage using the Configurattion class and then the JsonElement that has been received by the serialize method (using the debugger view) contains "60", "60.0", ... as it has been present in the JSON file.
If I use a Thing that contains the Configuration, the method receive a JsonElement that already contains "60.0" also if "60" has been stored.
So it could be related to the reading of the Thing...

But need to check that first...

maggu2810:
@cdjackson Our configuration class contains a map that key is a string and that value is declared to be an object. If Gson (or whatever) should be able to differ on derserialization if the JSON value is e.g. a double, long, integer, big decimal, big integer or whatever, will this ever work without storing the type in the serialized representation -- as long as the class that is deserialized (Configuation in this case) does not bring the information about the type?

cdjackson:
You are right - but my point is to see if we can make general assumptions, rather than having to had adaptors for each class (e.g. configuration).

We have the same issue in the REST interface where json is also passed - the system assumes that 60 is an INTEGER (BigDecimal with scale(0)), that 60.0 is a NUMBER, and “60” is a string - my preference is to try to stick to this sort of generalisation rather than adding separate converters. Now, maybe this isn’t possible everywhere - 60 could also be an Integer in some instances, but this is only a problem when we’ve got an Object rather than specific classes.

My hope was that the store would be generic - i.e. it can store anything without having to add a new converter for each type of data we want to store. Maybe the configuration object is a specific case that we need to account for, but (IMHO) it would nice to use the same rules as above, that are used in the REST interface, than to have to add the class.

Given that the json storage is meant to be easy (ish! ??) to edit manually, I personally think that it would be desirable to avoid adding more information into it (i.e. the class) if we have a robust solution that can avoid this.

watou:
In a test here, I skipped registering the PropertiesTypeAdapterFactory and instead just registered a PropertiesTypeAdapter, and my `inString` values are coming back without ".0" added to them.

PR #2675, opened 2016-12-19
fixed some openHAB occurrences

maggu2810:
I tried to only remove wrong occurrences.
References e.g. in the classic UI project and some documentation stuff are kept.


bundles/model/org.eclipse.smarthome.model.item.ui/src/org/eclipse/smarthome/model/ui/contentassist/ItemsProposalProvider.xtend
```diff
@@ -1,3 +1,10 @@
+/**
+ * Copyright (c) 2014-2016 by the respective copyright holders.
+ * All rights reserved. This program and the accompanying materials
+ * are made available under the terms of the Eclipse Public License v1.0
+ * which accompanies this distribution, and is available at
+ * http://www.eclipse.org/legal/epl-v10.html
+ */
 /*
  * generated by Xtext
```
kaikreuzer: As these files are generated as the Javadoc says, they are "inferred" files, i.e. imho not really be treated as source code (because nobody wrote it). As they also get overwritten on a potential re-generation, I would refrain from adding a license header to them.

maggu2810: I agree that we don't need a license header for generated files that are generated by the build process. I don't know if we should drop a license header for files that are checked in (regardless if generated or not).

But the reason the license header has been added now is:
There are also (IIRC two) xtend files with an outdated license header. If we want to be able to update that license headers we need to add the xtend file handling to the license plugin. I added "all" xtend files, as I don't see a change to exclude some by a regex (there is no generated directory used or something similar). We can exclude special xtend files or include some files, but using wildcards, we need a rule that could be used in the plugin configuration.
This has been the reason I added the license header to this files now, as it does not matter and it is much easier for the plugin configuration.
What do you prefer?

kaikreuzer: ok, fine for me, thanks.

PR #2676, opened 2016-12-19
made extension-label-customization backwards compatible

aounhaider1:


kaikreuzer:
@aounhaider1 It seems that this has broken the customization possibility.
I just tested the latest OH2 distro and it now again shows "Extensions" instead of "Add-ons" :-(

aounhaider1:
The label will need to be updated in the OH2's patch. It is not using the label from the previous place anymore. But that in turn makes sure that PaperUI will not completely fail if the customized labels are not present in a patch. I'll create a pull request for OH2 shortly.

kaikreuzer:
thanks @aounhaider1!

PR #2682, opened 2016-12-20
Fixes #2604. Stop execution of rule actions on the first action fail.

### LINKED ISSUE #2604: Change Rule execution: the a actions must stop on first failure
1) When a rule action throws an exception, the message has to be put in status detail and the RuleEngine has to stop execution of rest actions.
2) When a rule condition throws exception the exception must be put in the rule status details.

danchom:


maggu2810:
Triggered a rebuild, it succeeded.

PR #2685, opened 2016-12-20
Json storage BigDecimal scaling

cdjackson:
As discussed in #2665 Gson doesn't provide access to the raw data - in the TypeAdapter class, the string value that is available has already been altered from what is in the file - so a value in the file of 50 is already 50.0. Therefore there is no way to reliably resolve a BigDecimal with scale 1, and a BigDecimal with scale 0.

This currently appears to be the best compromise.


wborn:
Thanks for this fix! I just tested it and my issue seems to be resolved with it. :-)

cdjackson:
Great - thanks for the feedback :)

Issue #2667, opened 2016-12-18
Sonos binding: power off / power on


lolodomo:
I am testing the correct handling of the binding when powering off and powering on a Sonos device.

@kgoderis : I would like to understand what is the reason to call onUpdate() in thingDiscovered ? It is already called when the thing is initialized; this call will in fact do nothing as the job is already started.
https://github.com/eclipse/smarthome/blob/master/extensions/binding/org.eclipse.smarthome.binding.sonos/src/main/java/org/eclipse/smarthome/binding/sonos/handler/ZonePlayerHandler.java#L221
Was the idea to accelerate the process and to not wait for the next job cycle ?

I will propose a little update to be sure that when a thing changes from status OFFLINE to status ONLINE new GENA subscriptions will be requested.
What I noticed is that onServiceSubscribed is called with "failed" only after arround 5 minutes for all subscriptions when the Sonos device is OFF.

When powering off a Sonos, I was expected that thingRemoved to be called but it does never happen. isUpnpDeviceRegistered() still returns true. Normal ? Then when powering on the Sonos, thingRemoved is called and then thingDiscovered. So I am not sure we can rely on these methods thingDiscovered and thingRemoved. By the way, they are only doing something if you are using the default thing UID (not my case).

Finally, I discovered that after powering off a Sonos, if you request a new discovery, your Sonos is till discovered (due to cached data by jupnp ?) but there is an error detected by getSonosRoomName because it tries to read in the description file (which is not more accessible).
So discovery is not working very well in this case.

kgoderis:
@lolodomo First of all it has been a *long* time ago that I have written the Sonos code at the time, so my answers might not be 100% ok here.

> Was the idea to accelerate the process and to not wait for the next job cycle ?

Yes, but if it can be done in a better way, then we should do that

> When powering off a Sonos, I was expected that thingRemoved to be called but it does never happen. isUpnpDeviceRegistered() still returns true. Normal ?

As far as a I remember there is a time-out mechanism at the UPNP level with very long time-outs. I *think* this is the reason we have these side-effects, and why devices are reported registered,... even when they are off-line in reality. We should complement the UPNP mechanism with our own mechanism. That was the reason why we have this separate polling job in the handler to check whether a device is online or not. If we can improve this, or leverage error code from other places (e.g. getSonosRoomName) then maybe we should do that.

lolodomo:
[...]

Issue #2669, opened 2016-12-18
Sonos binding: no value set immediately when linking an item to a channel


lolodomo:
The Sonos binding is not yet handling REFRESH command.
When the user links an item to a channel, the item value is not set until there is an event from the Sonos device providing a new value.

lolodomo:
I think I know how to fix that and that is not very difficult. I will propose a fix.

maggu2810:
👍

lolodomo:
Closed by merged PR #2686.

PR #2686, opened 2016-12-20
Sonos binding: fix issues #2667 #2669

lolodomo:


extensions/binding/org.eclipse.smarthome.binding.sonos/src/main/java/org/eclipse/smarthome/binding/sonos/handler/ZonePlayerHandler.java
```diff
@@ -628,10 +504,143 @@ public void onValueReceived(String variable, String value, String service) {
...
 
-    @Override
-    public void channelLinked(ChannelUID channelUID) {
-        if (channelUID.getId().equalsIgnoreCase(COORDINATOR)) {
-            updateGroupCoordinator();
+    protected void updateChannel(String channeldD) {
+        State newState = UnDefType.UNDEF;
+        switch (channeldD) {
+            case STATE:
+                updateState(STATE, (stateMap.get("TransportState") != null)
```
maggu2810: You are using different logic in the case statements. Wouldn't it be more consistent to use one for all?

newState is already set to undef.

Here for `STATE` you don't use newState, but use the trinary operator in the update state method.

Below for `CONTROL` you first look if the state info is not null and then you don't override newState but set the state in the resp. sections. So no undef is set at all.

Below for `SHUFFLE` you set the newState it the state info is not null and then (in both cases) you set the newState (so undef or the known one).

lolodomo:
@maggu2810 : you were right, I have now updated the code to make it consistent for all channels.
PS: I take advantage of this PR update to update one or two additional things.

maggu2810:
👍

PR #2687, opened 2016-12-21
Implemented tests for org.eclipse.smarthome.core.audio

petarnv:
Signed-off-by: Petar Valchev <petar.valchev@musala.com>


bundles/core/org.eclipse.smarthome.core.audio.test/.classpath
```diff
@@ -0,0 +1,7 @@
+<?xml version="1.0" encoding="UTF-8"?>
+<classpath>
+	<classpathentry kind="con" path="org.eclipse.jdt.launching.JRE_CONTAINER/org.eclipse.jdt.internal.debug.ui.launcher.StandardVMType/JavaSE-1.7"/>
```
maggu2810: Should be changed to Java 1.8

bundles/core/org.eclipse.smarthome.core.audio.test/AudioOSGiTest.launch
```diff
@@ -0,0 +1,47 @@
...
...2.9.v20150224@default:default,org.eclipse.jetty.io*9.2.9.v20150224@default:default,org.eclipse.jetty.osgi.boot@default:default,org.eclipse.jetty.osgi.httpservice@default:default,org.eclipse.jetty.security*9.2.9.v20150224@default:default,org.eclipse.jetty.server*9.2.9.v20150224@default:default,org.eclipse.jetty.servlet*9.2.9.v20150224@default:default,org.eclipse.jetty.util*9.2.9.v20150224@default:default,org.eclipse.jetty.webapp@default:default,org.eclipse.jetty.xml@default:default,org.eclipse.jface.databinding@default:default,org.eclipse.jface@default:default,org.eclipse.osgi.compatibility.state@default:false,org.eclipse.osgi.services@default:default,org.eclipse.osgi@-1:true,org.eclipse.swt.win32.win32.x86_64@default:false,org.eclipse.swt@default:default,org.eclipse.team.core@default:default,org.eclipse.ui.trace@default:default,org.eclipse.ui.workbench@default:default,org.eclipse.ui@default:default,org.hamcrest.core@default:default,org.junit@default:default,org.objectweb.asm@default:default,org.slf4j.api@default:default,org.slf4j.jcl@default:default,org.w3c.css.sac@default:default,org.w3c.dom.events@default:default,org.w3c.dom.smil@default:default,org.w3c.dom.svg@default:default"/>
```
maggu2810: Do we really need all that UI bundles?

bundles/core/org.eclipse.smarthome.core.audio.test/META-INF/MANIFEST.MF
```diff
@@ -0,0 +1,21 @@
+Manifest-Version: 1.0
+Bundle-ManifestVersion: 2
+Bundle-Name: Eclipse Smarthome Audio Test
+Bundle-SymbolicName: org.eclipse.smarthome.core.audio.test
+Bundle-Version: 0.9.0.qualifier
+Fragment-Host: org.eclipse.smarthome.core.audio
+Bundle-RequiredExecutionEnvironment: JavaSE-1.7
```
maggu2810: Should be changed to Java 1.8

bundles/core/org.eclipse.smarthome.core.audio.test/pom.xml
```diff
@@ -0,0 +1,171 @@
...
+              <level>3</level>
+              <autoStart>true</autoStart>
+            </bundle>
+            <bundle>
+              <id>org.eclipse.jetty.osgi.httpservice</id>
+              <level>3</level>
+              <autoStart>true</autoStart>
+            </bundle>
+          </bundleStartLevel>
+          <argLine>-Dorg.osgi.service.http.port=8080 -Djetty.home.bundle=org.eclipse.jetty.osgi.boot</argLine>
```
maggu2810: Should we use another port that does not collide with one that is commonly used?
I assume we don't want to run into build / test failures, only because another application (e.g. ESH in the Eclipse IDE) is currently running.

bundles/core/org.eclipse.smarthome.core.audio/src/main/java/org/eclipse/smarthome/core/audio/internal/AudioManagerImpl.java
```diff
@@ -206,7 +206,7 @@ public AudioSink getSink() {
         String regex = pattern.replace("?", ".?").replace("*", ".*?");
         Set<String> matchedSources = new HashSet<String>();
 
-        for (String aSource : audioSinks.keySet()) {
+        for (String aSource : audioSources.keySet()) {
```
maggu2810: Does this change fix a bug in the current `AudioManagerImpl` implementation?

bundles/core/org.eclipse.smarthome.core.audio/src/main/java/org/eclipse/smarthome/core/audio/URLAudioStream.java
```diff
@@ -90,7 +90,10 @@ private InputStream createInputStream() throws AudioException {
                 os.write(req.getBytes());
                 is = shoutCastSocket.getInputStream();
             } else {
-                is = streamUrl.openStream();
+                // getInputStream() method is more error-proof than openStream(),
+                // because openStream() does openConnection().getInputStream(), 
+                // which opens a new connection and does not reuse the old one.
+                is = connection.getInputStream();
```
maggu2810: Does this change fix a bug in the current `URLAudioStream` implementation?
Shouldn't there be a separate PR for that bugfix?

bundles/core/org.eclipse.smarthome.core.audio.test/.classpath
petarnv: Done.

bundles/core/org.eclipse.smarthome.core.audio.test/AudioOSGiTest.launch
petarnv: I think I removed the unnecessary ones.

bundles/core/org.eclipse.smarthome.core.audio.test/META-INF/MANIFEST.MF
petarnv: Done.

bundles/core/org.eclipse.smarthome.core.audio.test/pom.xml
petarnv: Changed the port to 9090.

bundles/core/org.eclipse.smarthome.core.audio/src/main/java/org/eclipse/smarthome/core/audio/internal/AudioManagerImpl.java
petarnv: The method is `getSourceIds()`, so we should be iterating through sources, not sinks.

bundles/core/org.eclipse.smarthome.core.audio/src/main/java/org/eclipse/smarthome/core/audio/URLAudioStream.java
petarnv: Yes, I think we can call this a bug. `openStream()` method calls `openConnection().getInputStream()`, which opens an entirely new connection and does not reuse the old one. Also, `openStream()` initiates HTTP GET request. For example, this may cause a problem when we use `oneTimeStreams` in the `AudioServlet`, because `oneTimeStreams` can be requested a single time. 

This problem can be solved, if we use `getInputStream()` method on the currently opened connection.

I put it in this PR, because it is a small change, but if you want I can make a separate PR with this change.

petarnv:
I rebased the code to the current master.

bundles/core/org.eclipse.smarthome.core.audio/src/main/java/org/eclipse/smarthome/core/audio/URLAudioStream.java
maggu2810: > I put it in this PR, because it is a small change, but if you want I can make a separate PR with this change.

You should do it, not because I want it, but because that fix is not only related to the test.

Let's assume there is something completely wrong with this test (I don't think so, but ...), if we revert the commit that introduced that test, also the bugfix will be removed.
Let's assume someone is interested why the change has been done (without following every discussion on Github) and uses `git blame` and the Git history as a source of information (as it is intended by that VCS). Then the change is marked to be done because of testing. That is slightly not correct.

maggu2810:
> I rebased the code to the current master.

Thank you for adding the commented changes.
Will have a look at if there is some spare time.

bundles/core/org.eclipse.smarthome.core.audio/src/main/java/org/eclipse/smarthome/core/audio/internal/AudioManagerImpl.java
petarnv: Made a PR with this change: https://github.com/eclipse/smarthome/pull/3119. Ignored the tests, related to removing the change from this branch.

bundles/core/org.eclipse.smarthome.core.audio/src/main/java/org/eclipse/smarthome/core/audio/URLAudioStream.java
petarnv: Made a PR with this change: https://github.com/eclipse/smarthome/pull/3119. Ignored the tests, related to removing the change from this branch.

maggu2810: #3119 has been merged, so you can un-ignore the tests again.

petarnv: I un-ignored the ignored tests.

maggu2810:
Thank you for the tests.

PR #2689, opened 2016-12-21
[LIFX] Discover light labels, improve locking, optimize packet handling

wborn:
This PR adds light label discovery functionality based on the label used in the LIFX App. 
This improves the discovery user experience when working with a lot of lights.

Also the discovery process was further improved. Discovery now:
* only sends requests for missing light data
* waits 200ms before resending requests because it takes time for a response to arrive
* locks on one light only (and not all lights) when sending a packet to a specific light

With the above improvements I've seen the number of sent packets per `networkRunnable.run()` drop significantly. In my setup the number of sent packets per run has dropped by ~99% (from ~12538 to just ~177 packets, including the new label request packets).

Also the code has been reworked:
* it should be easier to add code for discovering additional light properties
* a `MACAddress` instead of a `String` is used to make sure all code is using the same light identifier when getting a lock


kgoderis:
@kaikreuzer lgtm as well (equally without testing)

Issue #689, opened 2015-12-09
New rule engine: CheckNow rule (same as IFTTT CheckNow)


danchom:
smhgit commented on Oct 30
I have posted this issue in the forum (https://www.eclipse.org/forums/index.php/t/860533/) but I think it is more accurate to discuss this option here.

(If you still wants to keep the discussion in the forum I can close this issue).

danchom:
kaikreuzer commented on Nov 1
The feature makes sense for me, but I leave the decision of whether or not to follow up on it here to @marinmitev as this is his branch :-)

danchom:
This is good feature for testing purposes. It can be done as special trigger type which sets context (trigger outputs) base on its configuration.

kaikreuzer:
Ok, picking this up again: I think this is a very useful feature that should be implemented.

The RuleRegistry could have a new method "runNow(ruleUID)", which skips triggers&conditions and directly executes the actions of the rule. This should always be possible unless an action has a mandatory input that is linked to a trigger; in that case, we should not execute the actions, but set the rule status to IDLE with an according message.

maggu2810:
Is it common (the title refers to IFTTT) to skip the condition on a manual trigger?

I would assume, that I trigger the rule manually and so the given triggers are not evaluated but I have assumed that the condition are checked before the action is executed.
Don't know what is common practice.

Perhaps there should be a distinction between trigger action and trigger rule execution.

kaikreuzer:
IFTTT does not have conditions, so this is not a good reference for answering that question.
I as a user would think something is broken, if I click "run now" and nothing happens. So I would assume that skipping the coditions is a logical way to do it.

kgoderis:
> Ok, picking this up again: I think this is a very useful feature that should be implemented.

FYI I had implemented an "empty" trigger to simulate this behaviour, but if we change the API in that way of course that is a better solution.

maggu2810:
> I as a user would think something is broken, if I click "run now" and nothing happens. So I would assume that skipping the coditions is a logical way to do it.

I think this depends if I trigger a rule execution or I trigger the rule's action.

Example
* trigger: motion sensor changed to ON
* condition: illumination < 500
* action: set light to ON

I assume
* if I trigger the rule manually and if the illumination is lower then 500, the light is switched on. If the illumination is greater or equal 500, the light state is unchanged
* if I trigger the action manually the light will be switched on

kaikreuzer:
> I assume [...]

[...]

PR #2691, opened 2016-12-21
Implements RuleRegistry.runNow(ruleId) #689

danchom:
Also fix bug in rule normalization. Default values defined in ConfigDescritptionParameter was not normalized.


kaikreuzer:
Thanks! How about adding this feature to the REST API as well, so that we can include a button in the Paper UI?

danchom:
Yes I was forgot about it. I'll try to extend REST API tomorrow and update the PR.

kaikreuzer:
@danchom Question: Wouldn't it actually make sense to allow the runNow to be executed, even if the rule is "disabled"? When developing/testing rules, I usually have them disabled, so that they are not triggering unexpectedly and I test them by doing a runNow - which is not possible in that case. Wdyt?

maggu2810:
> Question: Wouldn't it actually make sense to allow the runNow to be executed, even if the rule is "disabled"? When developing/testing rules, I usually have them disabled, so that they are not triggering unexpectedly and I test them by doing a runNow

I think this should be allowed as it is IMHO a common use case.

Issue #2589, opened 2016-12-03
PaperUI: translations of "description" is ignored


MHerbst:
I have set the language and locale to "German" resp. "de_DE". The PaperUi shows the translated names of the things but the "...description" entries in the properties file are ignored. 

This is an example from the Yahoo Weather binding:
![image](https://cloud.githubusercontent.com/assets/616792/20861883/7f82ce60-b99b-11e6-98d6-d1da5d67c33f.png)

kaikreuzer:
Did you set the locale of your runtime or the locale of the Paper UI (in Preferences menu)?

MHerbst:
I set everything in the Paper UI (under Preferences I selected German and under System/Locale I set the locale to de_DE).
I tried it again this morning and have some additional information that may help. For the Homematic things the display texts are correct under Things/Configuration:
![image](https://cloud.githubusercontent.com/assets/616792/20865302/88ca711c-ba0e-11e6-9a3c-af413f6cd271.png)

BTW: as you can see in the screenshot also the name of Yahoo binding and its description is not shown in the correct language.

The Homematic binding does not use the i18n properties files. So it seems that the problem happens only with i18n properties files.

MHerbst:
I have some additional information from some tests with the REST API. IF I execute this request with "Accept-Language" set to "de"
```
curl -X GET --header "Accept: application/json" --header "Accept-Language: de" "http://127.0.0.1:8080/rest/bindings"
```
I get the correct german description for the binding. If I set "Accept-Language" to "de_DE" or if I leave it empty.

MHerbst:
It seems so, but I not absolutely sure yet. Maybe it is combination of Paper/Basic UI and server side. In my test I create an own REST request. 
I have planned to perform more tests and log the REST request that were executed by Basic and Paper UI with different combinations of language and locale.

MHerbst:
@kaikreuzer I have executed some more tests and to me it looks like there is a either problem with the language string passed to methods implementing the REST API or the method that tries to parse the language is implemented too simple. In my test environment I used a fresh openHAB installation (Build 644), no explicit locale was set and language in preferences was set to "German" (same as the browser language). 

I have debugged method `org.eclipse.smarthome.io.rest.core.binding.BindingResource#getAll` and these are the results:

**Test 1:** REST request with explicit "Accept-Language: de" ()
```
[...]

PR #2699, opened 2016-12-23
#2589 Support of different Accept-Language HTTP header variants

MHerbst:
The content of the Accept-Language HTTP header variants may contain
comma-separated values. In this case the created Locale was not valid.
Now only the first parameter passed in the header is used.


MHerbst:
Sorry pressed the enter key too fast.

MHerbst:
Issue number corrected in the subject.

PR #2720, opened 2016-12-29
Document label and description tags for a channel (fix #2716)

### LINKED ISSUE #2716: Documentation: label and description for a channel
I discovered looking at the XSD that it is possible to define a label and a description for a channel. It is something not yet documented and I thought it was simply a missing feature.
Documentation only mentions the label and the description XML tags on the channel type, not on the channel.

lolodomo:

### DIFF
--- a/docs/documentation/development/bindings/xml-reference.md
+++ b/docs/documentation/development/bindings/xml-reference.md
@@ -279,6 +279,10 @@ Bridge and *Thing* descriptions must be placed as XML file(s) (with the ending `
 
     <channels>
       <channel id="channelID" typeId="channelTypeID" />
+      <channel id="channelID" typeId="channelTypeID">
+        <label>String</label>
+        <description>String</description>
+      </channel>
       ...
     </channels>
     OR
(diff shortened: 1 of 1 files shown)


docs/documentation/development/bindings/xml-reference.md
```diff
@@ -279,6 +279,10 @@ Bridge and *Thing* descriptions must be placed as XML file(s) (with the ending `
 
     <channels>
       <channel id="channelID" typeId="channelTypeID" />
+      <channel id="channelID" typeId="channelTypeID">
```
kaikreuzer: I'd suggest adding an "OR" above here as this is how it is done in the other places here as well.

docs/documentation/development/bindings/xml-reference.md
```diff
@@ -305,6 +309,10 @@ Bridge and *Thing* descriptions must be placed as XML file(s) (with the ending `
...
```
kaikreuzer: dito

maggu2810:
Using GitHub as committer will break the automatic CLA verification.

PR #2722, opened 2016-12-30
updated unit-api on target platform

kaikreuzer:
Replaced the unit-api 0.8 and unit-ri 0.8 bundles by unit-api 1.0 and uom-se 1.0.2 by https://github.com/eclipse/smarthome/commit/9f23b63a16fbd10ac86ab87acf39859617a09882.

This PR updates the smarthome.target to use the unit-api 1.0 now.
Unfortunately, we cannot yet add uom-se to the tp, since it has a dependency on javax.annotation 1.3.0, which is not available on Orbit. I guess we will need another CQ for it.

### DIFF
--- a/targetplatform/smarthome.target
+++ b/targetplatform/smarthome.target
@@ -1,5 +1,5 @@
 <?xml version="1.0" encoding="UTF-8" standalone="no"?>
-<?pde version="3.8"?><target name="SmartHome" sequenceNumber="104">
+<?pde version="3.8"?><target name="SmartHome" sequenceNumber="105">
 <locations>
 <location includeAllPlatforms="false" includeConfigurePhase="false" includeMode="planner" includeSource="true" type="InstallableUnit">
 <unit id="com.google.inject" version="3.0.0.v201312141243"/>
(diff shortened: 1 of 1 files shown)


clinique:
Please, note that uom-se v1.0.2 will need inclusion of  uom-lib-common-1.0.1.jar as well in the target platform

kaikreuzer:
Oh, indeed, I completely missed this dependency in the pom of uom-se.
@keilw, does this mean we will need yet another CQ? In which repo does uom-lib-common live?

clinique:
Seems to be here : https://github.com/unitsofmeasurement/uom-lib

keilw:
Yes, thanks to @duckAsteroid these are all OSGi bundles, but it seems in Eclipse it might require its own CQ.

keilw:
Only uom-lib-common btw, the GitHub repo contains other libraries like AssertJ etc. you do not need for implementations.

kaikreuzer:
ok, will try to create a CQ for it this week.
Do you see any potential obstacles or is it all clean BSD3 licensed code without any risks?

keilw:
Everything BSD3 code we wrote from scratch. It was not found to match the core API of the JSR at this point but helps standardize implementations and external libraries e.g. with Lambdas, that is why this is a separate module and not part of every implementation.

keilw:
And of course there are also other reusable classes like `BinaryPrefix`. We hoped to get more of them implementation-neutral. Maybe in the future we can, but at least some like `MetricPrefix `use specific versions of `UnitConverter `internally, so we could not factor it out at this point.

PR #2723, opened 2016-12-30
use endtime for last datapoint in boundary query

### LINKED ISSUE #2538: Persistence REST API w/ boundary returns wrong (?) timestamp on last point
The REST API returns values from a chosen persistence service. It queries the persistence service with starttime&endtime and returns these values. If the boundary option is set to true it additionally queries the persistence service for the values at the boundaryies and adds these before and after the other dataset.

In PersistenceResource.java l. 263 it adds the "start" value, using the starttime as the timestamp.
In PersistenceResource.java l. 307 it adds the "end" value, using the timestamp from the persistence service as the timestamp.

This is at least inconstistent. For RRD4J persistence this leads to two values with the same timestamp at the end. 

I suggest changing l. 307 to use the endtime as timestamp for the last datapoint. I'll make the needed changes after it is agreed that this is the proper way.

J-N-K:
Fixes #2538 


kaikreuzer:
@cdjackson As you have implemented the class, can I ask you to review this change?

cdjackson:
LGTM

PR #2732, opened 2017-01-03
readded unit-api and unit-ri 0.8

### LINKED ISSUE #2730: Unable to build ref-0.9.0.b2
I cloned master and checkout to specific tag ref-0.9.0.b2 and then i tried mvn clean install 

.
.
.
[INFO] Computing target platform for MavenProject: org.eclipse.smarthome.test:org.eclipse.smarthome.test:0.9.0.b2 @ /home/shankar/Documents/esh_clone/smarthome/bundles/test/org.eclipse.smarthome.test/pom.xml
[INFO] Fetching p2.index from http://download.eclipse.org/tools/orbit/downloads/drops/R20140525021250/repository/ (0B at 0B/s)
[INFO] Adding repository http://download.eclipse.org/tools/orbit/downloads/drops/R20140525021250/repository
[INFO] Fetching p2.index from http://download.eclipse.org/eclipse/updates/4.4/ (0B at 0B/s)
[INFO] Adding repository http://download.eclipse.org/eclipse/updates/4.4
[INFO] Fetching p2.index from http://download.eclipse.org/modeling/tmf/xtext/updates/releases/2.9.2/ (0B at 0B/s)
[INFO] Adding repository http://download.eclipse.org/modeling/tmf/xtext/updates/releases/2.9.2
[INFO] Adding repository http://download.eclipse.org/modeling/emft/mwe/updates/releases/2.8.3
[INFO] Adding repository http://download.eclipse.org/modeling/emf/emf/updates/2.11/core/R201506010402
[INFO] Adding repository http://dist.springsource.org/release/GRECLIPSE/e4.4
[INFO] Fetching p2.index from http://download.eclipse.org/releases/luna/201406250900/ (0B at 0B/s)
[INFO] Adding repository http://download.eclipse.org/releases/luna/201406250900
[INFO] Adding repository http://www.jupnp.org/update-site/2.2.0
[INFO] Adding repository http://hstaudacher.github.io/osgi-jax-rs-connector/5.3.0
[INFO] Adding repository http://download.eclipse.org/paho
[INFO] Adding repository http://eclipse.github.io/smarthome/third-party/target/repository
[ERROR] Failed to resolve target definition /home/shankar/Documents/esh_clone/smarthome/targetplatform/smarthome.target: Could not find "javax.measure.unit-api/0.8.0" in the repositories of the current location -> [Help 1]
[ERROR] 
[ERROR] To see the full stack trace of the errors, re-run Maven with the -e switch.
[ERROR] Re-run Maven using the -X switch to enable full debug logging.
[ERROR] 
[ERROR] For more information about the errors and possible solutions, please read the following articles:
[ERROR] [Help 1] http://cwiki.apache.org/confluence/display/MAVEN/MavenExecutionException
 

Please can you suggest any fix for this. 
maven version 3.3.9
openjdk 8
ubuntu 16.04 64 bit

kaikreuzer:
fixes #2730


kaikreuzer:
@maggu2810 It seems we have to keep older versions at least until there is next official release.

But imho we will have to live with the fact that some older versions/betas won't be buildable anymore after some time, if they rely on an artifact for which the CQ has been rejected.

maggu2810:
We only add stuff if a CQ has been approved. Didn't we?
So, if a CQ has been rejected after a previous approval it is a ... but then we have to remove the reference in the repo again. Correct?
If we want to ensure that every older tag could be build, we need to keep old artifacts in the third party repo or introduce a versioned repo.
Correct?

maggu2810:
Can we keep stuff in the eclipse repo if the CQ has been declined?
I am on travelling, so if there is no legal problem, I approve this PR. And you could feel free to merge.

kaikreuzer:
> We only add stuff if a CQ has been approved. Didn't we?

It is more complicated. We are an incubation project, so there is a two-step approval process: 1. check-in approval after a quick triage, 2. full approval after an in-depth triage.

unit-api got approval 1, but failed on 2. 2 is required to do an official release, though.
So before we release 0.9.0, we should also remove it again from the repo.

maggu2810:
> It is more complicated. We are an incubation project, so there is a two-step approval process: 1. check-in approval after a quick triage, 2. full approval after an in-depth triage.

We could use two third party repos.
One for "1" and one for "2".
Releases (and perhaps also our ref-tags) must use (and be able to build) using "2" only.
For the development process (snapshot builds), we could use 1 and 2.

kaikreuzer:
That won't help. The ref tags (and thus the beta like in this case) are specifically the "pseudo" releases that do not qualify as a "real" release, so they will use "1", which by definition can potentially break, if we remove stuff. We would have to maintain such a repo outside of Eclipse, then this might work (but we would require different CQs again, etc.)
So I'd suggest to leave it as it is and live with the fact that any ref-tags from a version older than the last official release might break.

maggu2810:
If we differ between "1" and "2", is there any need to remove stuff from 1 anymore?
Is there a requirement that stuff of "2" needs to be to included in an official Eclipse repo (e.g. Orbit)?

maggu2810:
Shouldn't we at least add an comment to the bundles in the pom e.g.
```text
# CQ not fully approved, only kept for old ref tag
```
or something similar?

Wouldn't this be much easier then to remember every entry that needs to be removed some time?

kaikreuzer:
I have created https://github.com/eclipse/smarthome/pull/2755 for it.

PR #2737, opened 2017-01-03
[2639] Handle properties files with encoding ISO-8859-1 correctly

### LINKED ISSUE #2639: Use of UTF-8 encoding for i18n properties files

MHerbst:
Fixes #2639 


eclipsewebmaster:
Issue tracker reference:
https://bugs.eclipse.org/bugs/show_bug.cgi?id=2639

bundles/core/org.eclipse.smarthome.core/src/main/java/org/eclipse/smarthome/core/common/osgi/ResourceBundleClassLoader.java
```diff
@@ -36,10 +36,15 @@
...
  *
  * @author Michael Grammling - Initial Contribution
+ * @author Martin Herbst - ISO-8859-1 support added
  */
 public class ResourceBundleClassLoader extends ClassLoader {
 
-    private static final String[] SUPPORTED_CHARSETS = { "UTF-8" };
+    /**
+     * ISO-8859-1 must be the last array element because all files could be decoded to ISO-8859-1 and
+     * therefore always character set ISO-8859-1 would be assumed.
```
kaikreuzer: Hm, I do not understand this. Does this mean that decoder.decode() will never throw a CharacterCodingException, since every byte is a valid character in ISO-8859-1? And UTF-8 decoding will fail on any non-ASCII character, because these bytes are not valid in an UTF-8 byte array?

Would it be possible to add some unit test in org.eclipse.smarthome.core.test that actually tests this charset detection mechanism? I would feel more comfortable with that :-)

bundles/core/org.eclipse.smarthome.core/src/main/java/org/eclipse/smarthome/core/common/osgi/ResourceBundleClassLoader.java
```diff
@@ -172,15 +187,26 @@ private Charset getResourceCharset(URL url) {
...
 
+    /**
+     * The method tests whether the file content can be decoded with the given character set.
+     *
+     * @param f
+     *            name of the file that will be tested
+     * @param charset
+     *            character set to which the file is tested
+     * @return
+     *         passed character set if the file could be decoded with this character set or null if the decode fails
```
kaikreuzer: Might it make sense to refactor this method to simply do the check and return a boolean? Would make much more sense to me...

bundles/core/org.eclipse.smarthome.core/src/main/java/org/eclipse/smarthome/core/common/osgi/ResourceBundleClassLoader.java
```diff
@@ -172,15 +187,26 @@ private Charset getResourceCharset(URL url) {
...
+     */
     private Charset detectCharset(String f, Charset charset) {
         try {
             InputStream in = super.getResourceAsStream(f);
             BufferedInputStream input = new BufferedInputStream(in);
             CharsetDecoder decoder = charset.newDecoder();
             decoder.reset();
             byte[] buffer = new byte[512];
-            boolean identified = false;
-            while ((input.read(buffer) != -1) && (!identified)) {
```
kaikreuzer: Did this ever work correctly? Didn't the loop always break after the first 512 characters, if they were successfully decoded...?

bundles/core/org.eclipse.smarthome.core/src/main/java/org/eclipse/smarthome/core/common/osgi/ResourceBundleClassLoader.java
MHerbst: > Does this mean that decoder.decode() will never throw a CharacterCodingException, since every byte is a valid character in ISO-8859-1? 

Exactly. I am not sure what happens with values below x'20'. But all values that are greater or equal are accepted as valid characters.

> And UTF-8 decoding will fail on any non-ASCII character, because these bytes are not valid in an UTF-8 byte array?

It simple just checks for valid byte combinations. A byte value that is greater then x7F is only allowed if the preceding byte has a special value.

This means that the routine can only detect if a text is *not* UTF-8.

> Would it be possible to add some unit test in org.eclipse.smarthome.core.test that actually tests this charset detection mechanism? I would feel more comfortable with that :-)

I understand :-). I think it is not necessary to test the whole class (would be complicated because of the bundle mechanism) but only the character detection method detectCharset. It is private but with the help of some reflection this will work.

bundles/core/org.eclipse.smarthome.core/src/main/java/org/eclipse/smarthome/core/common/osgi/ResourceBundleClassLoader.java
MHerbst: Yes, you are right. The method will be easier to understand if it returns a boolean. Will change it.

bundles/core/org.eclipse.smarthome.core/src/main/java/org/eclipse/smarthome/core/common/osgi/ResourceBundleClassLoader.java
MHerbst: No, it never worked correctly. It breaked if any chunk of 512 chars was decoded successfully. In my first test the first 512 were not value UTF-8 and the second 512 bytes were UTF-8. So the routine assumed it is value UTF-8 ...

MHerbst:
I have tested a file with BOM. It was correctly identified as UTF-8.

bundles/core/org.eclipse.smarthome.core/src/main/java/org/eclipse/smarthome/core/common/osgi/ResourceBundleClassLoader.java
kaikreuzer: > but only the character detection method detectCharset.

Yes, that was my idea.
It could be made visible within the package (i.e. default visibility), then the tests could access it without any reflection tricks.

bundles/core/org.eclipse.smarthome.core/src/main/java/org/eclipse/smarthome/core/common/osgi/ResourceBundleClassLoader.java
NorbertHD: detectCharset uses a byte array of 512 Bytes. What happens when the property file is smaller than 512 bytes or if the file size is not a multiple of 512? 
Also if a buffer ends with a part of an utf-8 sequence and the rest is in the next buffer, does this work also?

NorbertHD: detectCharset uses a byte array of 512 Bytes. What happens when the property file is smaller than 512 bytes or if the file size is not a multiple of 512? 
Also if a buffer ends with a part of an utf-8 sequence and the rest is in the next buffer, does this work also?

MHerbst: @NorbertHD I have tested the first two cases (and also added tests to the new unit test class). About the last case you mentioned I am not absolutely sure. I will test it.

MHerbst: @NorbertHD Good point. The test really fails if the UTF-8 character is split into two buffer parts. This means that the method needs some more changes :-(.
@kaikreuzer Please wait with a merge until this is also fixed.

MHerbst: @NorbertHD @kaikreuzer  I have changed the method how the file content is read and checked completely. Looks now much simpler.

kaikreuzer: The downside of this implementation is that it potentially consumes more memory as it does not process the stream in chunks, but loads it completely into memory before starting to process it. As we are talking about text files, I would assume that this is acceptable, though.

One very last question: Are we sure that the stream returned by getResourceAsStream(f) is always correctly closed? I am not sure it IOUtils.toByteArray closes the input stream - looking at the code, it actually does not seem to be the case.

MHerbst: > The downside of this implementation is that it potentially consumes more memory as it does not process the stream in chunks, but loads it completely into memory before starting to process it. As we are talking about text files, I would assume that this is acceptable, though.

I had the same thoughts. But I don't see any other reliable method. But I can optimize the code and create the the byte array only once per file by moving the code out the loop.

>One very last question: Are we sure that the stream returned by getResourceAsStream(f) is always correctly closed?

The documentation is not clear but I found some hints that it would be better to close it. I will add a close method and test it.

kaikreuzer: >  I will add a close method and test it.

I'd suggest to directly go for a try-with-resources.

MHerbst: > I'd suggest to directly go for a try-with-resources.

This is exactly what I did :-). Plus some further little optimizations: the byte array is now created only once per resource. Should be a little bit faster and less memory consuming if a file needs two tries to detect the encoding.

NorbertHD:
Thank you very much!

sjka:
I just had a look at these changes - and although I like the beauty if its simplicity, I'm not really sure if the strategy to always try UTF-8 first is correct. 

I can imagine there are cases where ISO-8859-1 encoded files contain characters which can be mis-interpreted to be valid UTF-8, but in fact should not (remember: ISO-8859-1 is the default for property files in Java...).

So take this ISO-8859-1 example:

```java
assertFalse(testObject.isCharsetValid("Bad encoding Ã¼zÃ¼m baÄlarÄ±".getBytes(Charset.forName("ISO-8859-1")), Charset.forName("UTF-8")));
```

I know, it looks strange, but I want it to do so and I want the users to see exactly this. However, as you can see, the test-case fails (it is mistakenly assumed to be valid UTF-8 also) and I end up getting "Bad encoding üzüm bağları". Which is definitely not what I wanted in this case.

So my point is: As long as we only have characters below 127, it does not matter if we by mistake assume it's UTF-8 because in this range they both are fully compatible. But if not, there is a fair chance that we accidentally pick the wrong one and in this case display wrong characters (again).

IMHO we have no other choice than declaring on character set to be used, and in this case I'd suggest to stick to the standard.

kaikreuzer:
@MHerbst I am afraid, @sjka is right :-/
So let's go for ISO only. The good thing is that the code will become even simpler :-)

MHerbst:
@maggu2810 entered a similar comment here https://github.com/eclipse/smarthome/issues/2639#issuecomment-270401111

I can accept their concerns about the solution. No problem, removing things is easy :-). 
The good thing is that we can remove the charset check completely. So it will need less memory and time. I can also remove the Unit test.

kaikreuzer:
At least we have learned a lot through this exercise 😉

PR #2742, opened 2017-01-04
DiscoveryResults not updated on DiscoveryService change

ivivanov-bg:
When particular DiscoveryService is changed (eg. there are new properties added to the discovery results) the already discovered things are not been updated.

This is due to the cache used in DiscoveryServiceRegistryImpl:

`private HashMultimap<DiscoveryService, DiscoveryResult> cachedResults = HashMultimap.create();`

Javadoc of `HashMultimap` says: 
`Adding a new key-value pair equal to an existing key-value pair has no effect.`

Possible solution is - each time an entry is put to the map - already existing entry to be first removed from the map

### DIFF
--- a/bundles/config/org.eclipse.smarthome.config.discovery.test/src/test/groovy/org/eclipse/smarthome/config/setup/test/discovery/DiscoveryServiceRegistryOSGITest.groovy
+++ b/bundles/config/org.eclipse.smarthome.config.discovery.test/src/test/groovy/org/eclipse/smarthome/config/setup/test/discovery/DiscoveryServiceRegistryOSGITest.groovy
@@ -440,5 +440,60 @@ class DiscoveryServiceRegistryOSGITest extends OSGiTest {
         assertNotNull extendedDiscoveryServiceMock.discoveryServiceCallback.getExistingDiscoveryResult(thingUID)
     }
 
+    @Test
+    void 'assert that updated discovery result is stored properly in cache'() {
+        ThingUID thingUID = new ThingUID(EXTENDED_BINDING_ID, EXTENDED_THING_TYPE, "foo")
+        ThingTypeUID thingTypeUID = new ThingTypeUID(EXTENDED_BINDING_ID, EXTENDED_THING_TYPE);
+
+        ScanListener scanListener =  new ScanListener() {
+                    public boolean discoveryFinished = false;
+                    @Override
+                    void onFinished() {
+                        discoveryFinished = true;
+                    }
+
+                    @Override
+                    void onErrorOccurred(Exception exception) {}
+                }
+
+        AsyncResultWrapper<DiscoveryResult> discoveryListenerResult = new AsyncResultWrapper<DiscoveryResult>()
+        DiscoveryListener discoveryListenerMock = [
+            thingDiscovered: { DiscoveryService source, DiscoveryResult result ->
+                discoveryListenerResult.set result
+            },
+            removeOlderThings: { DiscoveryService source, long timestamp, Collection<ThingTypeUID> thingTypeUIDs ->
+                []
+            }
+        ] as DiscoveryListener;
+
+        inbox.remove(thingUID);
+        boolean discoveryFinished = false;
+        discoveryServiceRegistry.startScan(thingTypeUID, scanListener);
+        waitForAssert ({ assertTrue(scanListener.discoveryFinished) });
+
+        discoveryServiceRegistry.addDiscoveryListener(discoveryListenerMock);
+        waitForAssert ({ assertNotNull(discoveryListenerResult.getWrappedObject()) });
+
+        assertEquals [:], discoveryListenerResult.getWrappedObject().getProperties();
+
+        discoveryServiceRegistry.removeDiscoveryListener(discoveryListenerMock);
+
+        discoveryListenerResult.set null;
+
+        Map<String, Object> props = ["ip":"0.0.0.0"];
+
+        extendedDiscoveryServiceMock.setDiscoveryProperties(props);
+
+        inbox.remove(thingUID);
+        scanListener.discoveryFinished = false;
+        discoveryServiceRegistry.startScan(thingTypeUID, scanListener);
+        waitForAssert ({ assertTrue(scanListener.discoveryFinished) });
+
+        discoveryServiceRegistry.addDiscoveryListener(discoveryListenerMock);
+        waitForAssert ({ assertNotNull(discoveryListenerResult.getWrappedObject()) });
+
+        assertEquals  props, discoveryListenerResult.getWrappedObject().getProperties();
+
+    }
 
 }
--- a/bundles/config/org.eclipse.smarthome.config.discovery.test/src/test/groovy/org/eclipse/smarthome/config/setup/test/discovery/ExtendedDiscoveryServiceMock.groovy
+++ b/bundles/config/org.eclipse.smarthome.config.discovery.test/src/test/groovy/org/eclipse/smarthome/config/setup/test/discovery/ExtendedDiscoveryServiceMock.groovy
@@ -23,9 +23,11 @@ import org.eclipse.smarthome.core.thing.ThingUID
 class ExtendedDiscoveryServiceMock extends DiscoveryServiceMock implements ExtendedDiscoveryService {
 
     public DiscoveryServiceCallback discoveryServiceCallback
+    public Map<String, Object> discoveryProperties;
 
     public ExtendedDiscoveryServiceMock(thingType, timeout, faulty = false) {
         super(thingType, timeout)
+        this.discoveryProperties = null;
     }
 
     @Override
@@ -35,6 +37,11 @@ class ExtendedDiscoveryServiceMock extends DiscoveryServiceMock implements Exten
 
     @Override
     public void startScan() {
-        thingDiscovered(new DiscoveryResultImpl(new ThingUID(thingType, "foo"), null, null, null, null, DEFAULT_TTL))
+        thingDiscovered(new DiscoveryResultImpl(new ThingUID(thingType, "foo"), null, discoveryProperties, null, null, DEFAULT_TTL))
     }
+    
+    public void setDiscoveryProperties(Map<String, Object> discoveryProperties) {
+        this.discoveryProperties = discoveryProperties;
+    }
+    
 }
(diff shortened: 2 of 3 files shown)


sjka:
This affects only `DiscoveryListener`s which are added later in the game, right? 
I'm missing at least a test case in `DiscoveryServiceRegistryOSGITest` for this as part of your PR...

kaikreuzer:
> Adding a new key-value pair equal to an existing key-value pair has no effect.

Can you explain why the new key-value pair is equal to the old one in that case?

ivivanov-bg:
@kaikreuzer - the new value is the same as the old one because the `equals` method of the `DiscoveryResultImpl` only checks the `thingUID` - and in my current scenario - the `thingUID` is the same but there is change in the properties map of the discovery result.

However - changing the `equals` method is not desired because we will end up having two discovery results for the same `thingUID` with different properties 

@sjka - you are correct - this only happens for `DiscoveryListener`s which are added after the 'DiscoveryResult' is been created. And the test is on it's way.

kaikreuzer:
@ivivanov-bg ok, thanks.

bundles/config/org.eclipse.smarthome.config.discovery.test/src/test/groovy/org/eclipse/smarthome/config/setup/test/discovery/DiscoveryServiceRegistryOSGITest.groovy
```diff
@@ -440,5 +440,55 @@ class DiscoveryServiceRegistryOSGITest extends OSGiTest {
         assertNotNull extendedDiscoveryServiceMock.discoveryServiceCallback.getExistingDiscoveryResult(thingUID)
     }
 
+    @Test
+    void 'assert that updated discovery result is stored properly in cache'() {
+        ThingUID thingUID = new ThingUID(EXTENDED_BINDING_ID, EXTENDED_THING_TYPE, "foo")
+        
+        
+        AsyncResultWrapper<DiscoveryResult> discoveryListenerResult = new AsyncResultWrapper<DiscoveryResult>()
```
sjka: You could use a simple `Map` here instead. Saves a lot of `discoveryListenerResult.getWrappedObject()`s later on and does not have any additional value.

bundles/config/org.eclipse.smarthome.config.discovery.test/src/test/groovy/org/eclipse/smarthome/config/setup/test/discovery/DiscoveryServiceRegistryOSGITest.groovy
```diff
@@ -440,5 +440,55 @@ class DiscoveryServiceRegistryOSGITest extends OSGiTest {
         assertNotNull extendedDiscoveryServiceMock.discoveryServiceCallback.getExistingDiscoveryResult(thingUID)
     }
 
+    @Test
+    void 'assert that updated discovery result is stored properly in cache'() {
+        ThingUID thingUID = new ThingUID(EXTENDED_BINDING_ID, EXTENDED_THING_TYPE, "foo")
```
sjka: You might also want to define a ThingTypeUID here, so you don't have to create it several times (makes it easier to read)

bundles/config/org.eclipse.smarthome.config.discovery.test/src/test/groovy/org/eclipse/smarthome/config/setup/test/discovery/DiscoveryServiceRegistryOSGITest.groovy
```diff
@@ -440,5 +440,55 @@ class DiscoveryServiceRegistryOSGITest extends OSGiTest {
...
+        
+        
+        AsyncResultWrapper<DiscoveryResult> discoveryListenerResult = new AsyncResultWrapper<DiscoveryResult>()
+        DiscoveryListener discoveryListenerMock = [
+            thingDiscovered: { DiscoveryService source, DiscoveryResult result ->
+                discoveryListenerResult.set result
+
```
sjka: empty line

bundles/config/org.eclipse.smarthome.config.discovery.test/src/test/groovy/org/eclipse/smarthome/config/setup/test/discovery/DiscoveryServiceRegistryOSGITest.groovy
```diff
@@ -440,5 +440,55 @@ class DiscoveryServiceRegistryOSGITest extends OSGiTest {
...
+       
+        discoveryListenerResult.set null;
+        
+        Map<String, Object> props = new HashMap<>();
+        props.put("ip", "0.0.0.0");
+        
+        extendedDiscoveryServiceMock.setDiscoveryProperties(props);
+        
+        discoveryFinished = false;
+        discoveryServiceRegistry.startScan(new ThingTypeUID(EXTENDED_BINDING_ID, EXTENDED_THING_TYPE), [
```
sjka: factor the `ScanListener` out, so you can reuse the one from above (easier to read)

bundles/config/org.eclipse.smarthome.config.discovery.test/src/test/groovy/org/eclipse/smarthome/config/setup/test/discovery/DiscoveryServiceRegistryOSGITest.groovy
```diff
@@ -440,5 +440,55 @@ class DiscoveryServiceRegistryOSGITest extends OSGiTest {
...
+                []
+            }
+        ] as DiscoveryListener
+
+        boolean discoveryFinished = false;
+        discoveryServiceRegistry.startScan(new ThingTypeUID(EXTENDED_BINDING_ID, EXTENDED_THING_TYPE), [
+            onFinished: { discoveryFinished = true },
+            onErrorOccurred: { }
+        ] as ScanListener);
+        waitForAssert ({ assertTrue(discoveryFinished) }, 2000 );
```
sjka: please don't pass any custom timeouts but stick to the defaults here. I know at least to CI servers where this is going to fail occasionally...! 😉 

(same for all the timeouts below)

bundles/config/org.eclipse.smarthome.config.discovery.test/src/test/groovy/org/eclipse/smarthome/config/setup/test/discovery/DiscoveryServiceRegistryOSGITest.groovy
```diff
@@ -440,5 +440,55 @@ class DiscoveryServiceRegistryOSGITest extends OSGiTest {
...
+        
+        discoveryServiceRegistry.addDiscoveryListener(discoveryListenerMock);
+        waitForAssert ({ assertNotNull(discoveryListenerResult.getWrappedObject()) }, 2000 );
+        
+        assertEquals new HashMap<>(), discoveryListenerResult.getWrappedObject().getProperties();
```
sjka: did you know that `[:]` is a short notation for an empty map in Groovy?

bundles/config/org.eclipse.smarthome.config.discovery.test/src/test/groovy/org/eclipse/smarthome/config/setup/test/discovery/DiscoveryServiceRegistryOSGITest.groovy
```diff
@@ -440,5 +440,55 @@ class DiscoveryServiceRegistryOSGITest extends OSGiTest {
...
+        
+        discoveryServiceRegistry.removeDiscoveryListener(discoveryListenerMock);
+       
+        discoveryListenerResult.set null;
+        
+        Map<String, Object> props = new HashMap<>();
```
sjka: works here too: `def props = ["ip":"0.0.0.0"]`

bundles/config/org.eclipse.smarthome.config.discovery.test/src/test/groovy/org/eclipse/smarthome/config/setup/test/discovery/ExtendedDiscoveryServiceMock.groovy
```diff
@@ -35,6 +37,11 @@ class ExtendedDiscoveryServiceMock extends DiscoveryServiceMock implements Exten
 
     @Override
     public void startScan() {
-        thingDiscovered(new DiscoveryResultImpl(new ThingUID(thingType, "foo"), null, null, null, null, DEFAULT_TTL))
+        thingDiscovered(new DiscoveryResultImpl(new ThingUID(thingType, "foo"), null, this.discoveryProperties, null, null, DEFAULT_TTL))
```
sjka: `this.` is superfluous here

bundles/config/org.eclipse.smarthome.config.discovery.test/src/test/groovy/org/eclipse/smarthome/config/setup/test/discovery/ExtendedDiscoveryServiceMock.groovy
```diff
@@ -35,6 +37,11 @@ class ExtendedDiscoveryServiceMock extends DiscoveryServiceMock implements Exten
...
     }
+    
+    public void setDiscoveryProperties(Map<String, Object> _discoveryProperties) {
+        this.discoveryProperties = _discoveryProperties;
```
sjka: we generally don't use c++ best practices here. Simply name it `discoveryProperties`

ivivanov-bg:
@sjka I think I addressed your comments.

About the IDE - I use Eclipse - the formatting looks fine on my end but just in case  - I applied the ESH formatter once again.

About the other question - Yes - I tried the tests without the change to `DiscoveryServiceRegistryImpl` and I noticed very strange behavior of groovy as the tests were passing even without the fix.

Basically - when I print the content of the `cachedResults` map in the `DiscoveryServiceRegistryImpl` - it behaves as expected - the same element is NOT added to the map on the second call.

However - if I print the content of the very same map inside the groovy test - the content of the map is replaced on the second discovery.

This is IMHO totally strange and questions the correctness of all groovy tests. Can you please explain me why is this happening and is this expected behavior?

Regards,
Ivo

sjka:
> This is IMHO totally strange and questions the correctness of all groovy tests. Can you please explain me why is this happening and is this expected behavior?

No, the "problem" is not groovy - what you are seeing here in your test-case is a feature from the PersistentInbox: It [synchronizes](https://github.com/eclipse/smarthome/blob/master/bundles/config/org.eclipse.smarthome.config.discovery/src/main/java/org/eclipse/smarthome/config/discovery/internal/PersistentInbox.java#L186) newly discovered information with previous results.

The `PersistentInbox` is also a discovery listener. So it depends on the order: if yours is called after the inbox all *seems* to be good. 

The fact that the inbox manipulates the original object can be considered as yet another bug which needs to be fixed.

ivivanov-bg:
@sjka, @kaikreuzer - If you don't have more comments - I guess this could be merged

sjka:
> If you don't have more comments

How about addressing my previous comment? Your test is correct itself, but the setup invalidates its result, as I explained above.

ivivanov-bg:
Done

sjka:
lgtm

thanks, @ivivanov-bg !

Issue #1896, opened 2016-07-22
Avoid sending of events and notification of listeners during startup / shutdown of ESH


tomhoefer:
Hi all,

in our project we have a lot of event subscribers and registry change listeners implemented which are called during startup / shutdown of ESH as a matter of course. In shutdown phase these services will update their model accordingly which results in the problem that the model cannot be re-built after next startup (because the subscribers / listeners have been assumed that the item / thing / link has been really deleted). We should distinguish between event sending / listener notification for adding / removal of items / things / links **during framework startup / shutdown phase and normal runtime.** 

For this reason I would like to disable in our project that events are sent / listeners are notified during startup / shutdown. I could imagine two ways to implement this:

1. Providing a ```RuntimeStateService``` that can be requested in order to get the information if the runtime is started. So for the beginning the service would only consist of a single operation ```boolean : isStarted()``` and it will be injected as dynamic dependency into the registries (things, items, links, rules). Then I would skip sending events / notification if the service is present and the runtime / framework is not fully started.

2. Each service that requires the runtime state has to implement a ```RuntimeStateListener``` interface which is tracked by a central ```RuntimeStateService``` to be provided by solutions based on ESH. As soon as the runtime state has changed to started the service will inform all listeners about this. Once the runtime left the started state again all listeners are informed about this. For the beginning I will implement the new RuntimeStateListener interface by the AbstractRegistry (concrete registries can decide if the runtime state listener is to be provided as a service).

I think that 2 is the only valid option to implement this requirement. In 1 the runtime state service could unregister too early so that the events are sent / listeners are notified again.

What do you think?

maggu2810:
How do you want to decide that the framework is started or shut down?
If I update one Eclipse SmartHome bundle that triggers some restarts (deactivate, activate) of bundles and / or services. Which one is responsible for a whole "startup" / "shutdown" state?

Isn't this product specific which services needs to be available to signal "all present" / "normal runtime"?

tomhoefer:
In our product we can rely on our OSGi runtime that the framework is started. For solutions running e.g. on Equinox I thought to use a framework listener and listen on the ```FrameworkEvent.STARTED```

maggu2810:
I have assumed you are talking of the Startup and Shutdown phase of the Eclipse SmartHome framework. But you refer to the start and stop of the OSGi Framework. Correct?

So, there are some options, using a FrameworkListener, using a SynchronousBundleListener for bundle 0 to handle on the stopping event, ...

[...]

PR #2743, opened 2017-01-04
[RFC] introduce CREATED and DELETED events for Things and Items

sjka:
in order to allow distinguishing ADDED/REMOVED (which might also be caused by providers being added/removed , e.g. during startup/shutdown) from real CREATION and DELETION.

We had several discussions by now related to this topic (e.g. #1896). While in a dynamic system it is really hard to tell if startup/shudown is "finished", this PR is an attempt to ease the pain. I'm open for better suggestions, of course.

So, WDYT? 

### DIFF
--- a/bundles/core/org.eclipse.smarthome.core.test/src/test/groovy/org/eclipse/smarthome/core/internal/ItemRegistryOSGiTest.groovy
+++ b/bundles/core/org.eclipse.smarthome.core.test/src/test/groovy/org/eclipse/smarthome/core/internal/ItemRegistryOSGiTest.groovy
@@ -21,12 +21,17 @@ import org.eclipse.smarthome.core.items.ItemProvider
 import org.eclipse.smarthome.core.items.ItemRegistry
 import org.eclipse.smarthome.core.items.ItemRegistryChangeListener
 import org.eclipse.smarthome.core.items.ItemsChangeListener
+import org.eclipse.smarthome.core.items.ManagedItemProvider
+import org.eclipse.smarthome.core.items.events.AbstractItemRegistryEvent
 import org.eclipse.smarthome.core.items.events.ItemAddedEvent
+import org.eclipse.smarthome.core.items.events.ItemCreatedEvent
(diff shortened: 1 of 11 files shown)


kaikreuzer:
I agree that we will need to address the issue in some way, I nonetheless have some comments worth to discuss:

1. Besides Things and Items, we also have added/removed events for Links, Rules and Inbox. I think we should address all of them at once or none.
2. It does not necessarily solve startup / synchronization issues: A new "created" event seems to imply to consumers that they are informed about newly created entities. But this won't be the case for anything that has changed while the system was down - think e.g. of a restored backup or simply DSL files being changed. I assume we will still only send added events on a startup, even if there are new entities that didn't exist before.
3. For every created event, we also have an added event. While this keeps backward compatibility, it somehow looks weird as a first impression and might confuse people.
4. Currently, after a system startup, the semantic of "added" events is more or less what "created" now intends to express. So actually, every consumer should change his code to rather react on created instead of added. Is then there still any use case left for the added events? And if not: Why is there then a need for created events...?

afuechsel:
@kaikreuzer now you have confused me.... 

1. I agree
2. Did I get your point correctly, that we cannot distinguish at startup, whether a new entity has been created and will only send "added" events? 
3. It's kind of symmetric: each "created" will be followed by an "added", but "added" could also be sent without a previous "created" - vs. each "deleted" will have a "removed" before, but "removed" could also be sent without beeing followed by "deleted"
4. May be also only to have it symmetric

kaikreuzer:
2. correct.

sjka:
I agree with 1., disagree with 3. but in the end I think your 2. point is the most important one: With the introduction of new event types we indeed would imply a pretty strong guarantee that entities ARE REALLY created/deleted. But you are right, this approach actually is not trying hard enough and would fail in such scenarios.

I'm not sure though if it is possible (with reasonable effort) to do so. In the end, the use-cases I had in mind would have required such a strong guarantee, therefore I better withdraw this PR and think about another solution.

Issue #2702, opened 2016-12-26
Paper UI Enhancement: add filter and/or grouping options to pages Things and Inbox


MHerbst:
The new search and filter options are really a great usability improvement. 
But in the case that the Inbox contains lots of things resp. the list of things is quite large I could imagine the following enhancements:

- Inbox:
  - alphabetic order of discovered things
  - search function
  - filter/group by binding type
- Things:
  - filter/group by binding type
- All pages with lists
  - on optional "compact view" with smaller images and maybe less text.

kaikreuzer:
Sounds all like useful enhancements to me.
@aounhaider1 It probably makes sense to split it up in several PRs. Feel free to decide, what might be a quick win (like sorting inbox entries) and which might be a bigger refactoring (e.g. introducing filtering on other lists than items).

kaikreuzer:
@aounhaider1 I noticed that the filters on Inbox and Things only check labels, but not ids, while on items and extensions, also ids are taken into account. I would vote for checking ids in Inbox and Things as well, wdyt?

aounhaider1:
Yes, makes sense. I'll add that.

PR #2747, opened 2017-01-04
Refactored filters from items for inbox and things

aounhaider1:
Implements https://github.com/eclipse/smarthome/issues/2702 except for last point.


kaikreuzer:
Thanks. A few comments from my side, this is how the inbox appears:

screen shot 2017-01-05 at 15 07 18

- The "search" field is missing here (while it is there for items & things)
- The filter box should say "filter by binding" (i.e. remove the word "type") (both on Inbox and Things list)
- The selection box should only list bindings for which entries exist (both on Inbox and Things list)
- There is no "group by binding" as @MHerbst had suggested. Not sure if we really want that though, as it will conflict with the alphabetical sorting.
- Quite often, the selection box shows only one entry ("hue Binding") after setting and clearing the filter. A page reload makes it work correctly again.

aounhaider1:
Updated accordingly.

Issue #2649, opened 2016-12-13
Weather binding


lolodomo:
We don't have an equivalent of the OH 1.x weather binding in Eclipse smarthome or OH2. We only have  the Yahoo weather binding but it provides apparently less data than the 1.x binding.
Any reason for that ?
Any work in progress ?
Any problem with the proposal of a general weather binding in Eclipse smarthome with several possible providers that will be a migration of the 1.x binding ?
I am a candidate to work on this new binding.

lolodomo:
Here is my proposal:
* Bridges:
  * weather:geoLocation:locationId with config parameters: latitude, longitude
  * weather:yahooLocation:locationId with config parameter: woeid
* Things:
  * weather:yahooData:locationId:id with config parameter: updateInterval
  * weather:hamData:locationId:id with config parameters: clientId, clientSecret, updateInterval
  * weather:providerData:locationId:id with config parameters: provider, apikey, updateInterval, language
* Binding configuration parameters: speedUnit, temperatureUnit, rainUnit, snowUnit, pressureUnit, lenghtUnit
* Channels:
  * channel groups atmosphere_today, atmosphere_forecast1, atmosphere_forecast2, ... with channels humidity, visibility, pressure, pressureTrend, ozone, uvIndex
  * channel groups clouds_today, clouds_forecast1, clouds_forecast2, ... with channel percent
  * channel groups condition_today, condition_forecast1, condition_forecast2, ... with channels text, observationTime, id, lastUpdate, commonId
  * channel groups precipitation_today, precipitation_forecast1, precipitation_forecast2, ... with channels rain, snow, probability, total
  * channel groups temperature_today, temperature_forecast1, temperature_forecast2, ... with channels current, min, max, feel, dewpoint, minMax
  * channel groups wind_today, wind_forecast1, wind_forecast2, ... with channels speed, direction, degree, gust, chill
* General idea is to re-use what can be re-used from the 1.x binding, that is mainly all provider data retrieval and internal data storing
* what would be lost (at least in a first time)
  * round/scale for numeric values
  * WEB view
* this binding will entirely cover the Yahoo Weather binding

Any comment ?

watou:
I think each weather provider should have its own Bridge Thing handler implementation, hiding all provider specific details there, but feed data to a common set of Things like Conditions, Forecast, etc.  The binding itself should have no configuration.  The YahooWeather binding was meant mainly as a code example, iirc.

kaikreuzer:
[...]

PR #1857, opened 2016-07-13
Issue #601 : Introduce an optional unit in Number item, now that CQ h…

clinique:
…as been approved.

Not sure I included the libraries in the right way (I suspect it should be pushed to the target platform definition), but well here it is at last.

Signed-off-by: Gaël L'hopital glhopital@gmail.com

PR #2748, opened 2017-01-04
Weather Underground binding: initial contribution

lolodomo:

### DIFF
--- a/extensions/binding/org.eclipse.smarthome.binding.weatherunderground/.classpath
+++ b/extensions/binding/org.eclipse.smarthome.binding.weatherunderground/.classpath
@@ -1,0 +1,7 @@
+<?xml version="1.0" encoding="UTF-8"?>
+<classpath>
+	<classpathentry kind="con" path="org.eclipse.jdt.launching.JRE_CONTAINER/org.eclipse.jdt.internal.debug.ui.launcher.StandardVMType/JavaSE-1.8"/>
+	<classpathentry kind="con" path="org.eclipse.pde.core.requiredPlugins"/>
+	<classpathentry kind="src" path="src/main/java"/>
+	<classpathentry kind="output" path="target/classes"/>
+</classpath>
--- a/extensions/binding/org.eclipse.smarthome.binding.weatherunderground/ESH-INF/thing/thing-types.xml
+++ b/extensions/binding/org.eclipse.smarthome.binding.weatherunderground/ESH-INF/thing/thing-types.xml
@@ -1,0 +1,646 @@
+<?xml version="1.0" encoding="UTF-8"?>
+<thing:thing-descriptions bindingId="weatherunderground" xmlns:xsi="http://www.w3.org/2001/XMLSchema-instance"
+	xmlns:thing="http://eclipse.org/smarthome/schemas/thing-description/v1.0.0"
+	xsi:schemaLocation="http://eclipse.org/smarthome/schemas/thing-description/v1.0.0 http://eclipse.org/smarthome/schemas/thing-description-1.0.0.xsd">
+
+	<!-- WeatherUnderground Binding -->
+	<thing-type id="weather">
+		<label>Weather Information</label>
+		<description>Provides various weather data from the Weather Underground service</description>
+
+		<channel-groups>
+			<channel-group id="current" typeId="current" />
+			<channel-group id="forecastToday" typeId="forecast">
+				<label>Weather Forecast Today</label>
+				<description>This is the weather forecast for today</description>
+			</channel-group>
+			<channel-group id="forecastTomorrow" typeId="forecast">
+				<label>Weather Forecast Tomorrow</label>
+				<description>This is the weather forecast for tomorrow</description>
+			</channel-group>
+			<channel-group id="forecastDay2" typeId="forecast">
+				<label>Weather Forecast Day 2</label>
+				<description>This is the weather forecast in two days</description>
+			</channel-group>
+			<channel-group id="forecastDay3" typeId="forecast">
+				<label>Weather Forecast Day 3</label>
+				<description>This is the weather forecast in three days</description>
+			</channel-group>
+			<channel-group id="forecastDay4" typeId="forecast">
+				<label>Weather Forecast Day 4</label>
+				<description>This is the weather forecast in four days</description>
+			</channel-group>
+			<channel-group id="forecastDay5" typeId="forecast">
+				<label>Weather Forecast Day 5</label>
+				<description>This is the weather forecast in five days</description>
+			</channel-group>
+			<channel-group id="forecastDay6" typeId="forecast">
+				<label>Weather Forecast Day 6</label>
+				<description>This is the weather forecast in six days</description>
+			</channel-group>
+			<channel-group id="forecastDay7" typeId="forecast">
+				<label>Weather Forecast Day 7</label>
+				<description>This is the weather forecast in seven days</description>
+			</channel-group>
+			<channel-group id="forecastDay8" typeId="forecast">
+				<label>Weather Forecast Day 8</label>
+				<description>This is the weather forecast in eight days</description>
+			</channel-group>
+			<channel-group id="forecastDay9" typeId="forecast">
+				<label>Weather Forecast Day 9</label>
+				<description>This is the weather forecast in nine days</description>
+			</channel-group>
+		</channel-groups>
+
+		<config-description>
+			<parameter name="apikey" type="text" required="true">
+				<context>password</context>
+				<label>API Key</label>
+				<description>API key to access the Weather Underground service</description>
+			</parameter>
+			<parameter name="location" type="text" required="true">
+				<label>Location of Weather Information</label>
+				<description>Multiple syntaxes are supported. Please read the binding documentation for more information</description>
+			</parameter>
+			<parameter name="language" type="text" required="false">
+				<label>Language</label>
+				<description>Language to be used by the Weather Underground service</description>
+				<options>
+					<option value="AF">Afrikaans</option>
+					<option value="AL">Albanian</option>
+					<option value="AR">Arabic</option>
+					<option value="HY">Armenian</option>
+					<option value="AZ">Azerbaijan</option>
+					<option value="EU">Basque</option>
+					<option value="BY">Belarusian</option>
+					<option value="BU">Bulgarian</option>
+					<option value="LI">British English</option>
+					<option value="MY">Burmese</option>
+					<option value="CA">Catalan</option>
+					<option value="CN">Chinese - Simplified</option>
+					<option value="TW">Chinese - Traditional</option>
+					<option value="CR">Croatian</option>
+					<option value="CZ">Czech</option>
+					<option value="DK">Danish</option>
+					<option value="DV">Dhivehi</option>
+					<option value="NL">Dutch</option>
+					<option value="EN">English</option>
+					<option value="EO">Esperanto</option>
+					<option value="ET">Estonian</option>
+					<option value="FA">Farsi</option>
+					<option value="FI">Finnish</option>
+					<option value="FR">French</option>
+					<option value="FC">French Canadian</option>
+					<option value="GZ">Galician</option>
+					<option value="DL">German</option>
+					<option value="KA">Georgian</option>
+					<option value="GR">Greek</option>
+					<option value="GU">Gujarati</option>
+					<option value="HT">Haitian Creole</option>
+					<option value="IL">Hebrew</option>
+					<option value="HI">Hindi</option>
+					<option value="HU">Hungarian</option>
+					<option value="IS">Icelandic</option>
+					<option value="IO">Ido</option>
+					<option value="ID">Indonesian</option>
+					<option value="IR">Irish Gaelic</option>
+					<option value="IT">Italian</option>
+					<option value="JP">Japanese</option>
+					<option value="JW">Javanese</option>
+					<option value="KM">Khmer</option>
+					<option value="KR">Korean</option>
+					<option value="KU">Kurdish</option>
+					<option value="LA">Latin</option>
+					<option value="LV">Latvian</option>
+					<option value="LT">Lithuanian</option>
+					<option value="ND">Low German</option>
+					<option value="MK">Macedonian</option>
+					<option value="MT">Maltese</option>
+					<option value="GM">Mandinka</option>
+					<option value="MI">Maori</option>
+					<option value="MR">Marathi</option>
+					<option value="MN">Mongolian</option>
+					<option value="NO">Norwegian</option>
+					<option value="OC">Occitan</option>
+					<option value="PS">Pashto</option>
+					<option value="GN">Plautdietsch</option>
+					<option value="PL">Polish</option>
+					<option value="BR">Portuguese</option>
+					<option value="PA">Punjabi</option>
+					<option value="RO">Romanian</option>
+					<option value="RU">Russian</option>
+					<option value="SR">Serbian</option>
+					<option value="SK">Slovak</option>
+					<option value="SL">Slovenian</option>
+					<option value="SP">Spanish</option>
+					<option value="SI">Swahili</option>
+					<option value="SW">Swedish</option>
+					<option value="CH">Swiss</option>
+					<option value="TL">Tagalog</option>
+					<option value="TT">Tatarish</option>
+					<option value="TH">Thai</option>
+					<option value="TR">Turkish</option>
+					<option value="TK">Turkmen</option>
+					<option value="UA">Ukrainian</option>
+					<option value="UZ">Uzbek</option>
+					<option value="VU">Vietnamese</option>
+					<option value="CY">Welsh</option>
+					<option value="SN">Wolof</option>
+					<option value="JI">Yiddish - transliterated</option>
+					<option value="YI">Yiddish - unicode</option>
+				</options>
+			</parameter>
+			<parameter name="refresh" type="integer" min="5" required="false">
+				<label>Refresh interval</label>
+				<description>Specifies the refresh interval in minutes.</description>
+				<default>30</default>
+			</parameter>
+		</config-description>
+	</thing-type>
+
+	<channel-group-type id="current">
+		<label>Current Weather</label>
+		<description>This is the current weather</description>
+		<channels>
+			<channel id="location" typeId="location" />
+			<channel id="stationId" typeId="stationId" />
+			<channel id="observationTime" typeId="observationTime" />
+			<channel id="conditions" typeId="currentConditions" />
+			<channel id="temperature" typeId="temperature" />
+			<channel id="relativeHumidity" typeId="relativeHumidity" />
+			<channel id="windDirection" typeId="windDirection" />
+			<channel id="windDirectionDegrees" typeId="windDirection-degrees" />
+			<channel id="windSpeed" typeId="windSpeed" />
+			<channel id="windGust" typeId="windGust" />
+			<channel id="pressure" typeId="pressure" />
+			<channel id="dewPoint" typeId="dewPoint" />
+			<channel id="heatIndex" typeId="heatIndex" />
+			<channel id="windChill" typeId="windChill" />
+			<channel id="feelingTemperature" typeId="feelingTemperature" />
+			<channel id="visibility" typeId="visibility" />
+			<channel id="solarRadiation" typeId="solarRadiation" />
+			<channel id="UVIndex" typeId="UVIndex" />
+			<channel id="precipitationDay" typeId="rainDay" />
+			<channel id="precipitationHour" typeId="rainHour" />
+			<channel id="icon" typeId="icon" />
+		</channels>
+	</channel-group-type>
+
+	<channel-group-type id="forecast">
+		<label>Weather Forecast</label>
+		<description>This is the weather forecast</description>
+		<channels>
+			<channel id="forecastTime" typeId="forecastTime" />
+			<channel id="conditions" typeId="forecastConditions" />
+			<channel id="minTemperature" typeId="minTemperature" />
+			<channel id="maxTemperature" typeId="maxTemperature" />
+			<channel id="relativeHumidity" typeId="relativeHumidity" />
+			<channel id="probaPrecipitation" typeId="probaPrecipitation" />
+			<channel id="precipitationDay" typeId="rainDay" />
+			<channel id="snow" typeId="snow" />
+			<channel id="maxWindDirection" typeId="maxWindDirection" />
+			<channel id="maxWindDirectionDegrees" typeId="maxWindDirection-degrees" />
+			<channel id="maxWindSpeed" typeId="maxWindSpeed" />
+			<channel id="averageWindDirection" typeId="averageWindDirection" />
+			<channel id="averageWindDirectionDegrees" typeId="averageWindDirection-degrees" />
+			<channel id="averageWindSpeed" typeId="averageWindSpeed" />
+			<channel id="icon" typeId="icon" />
+		</channels>
+	</channel-group-type>
+
+	<channel-type id="location" advanced="true">
+		<item-type>String</item-type>
+		<label>Observation Location</label>
+		<description>Weather observation location</description>
+		<state readOnly="true" pattern="%s"></state>
+	</channel-type>
+
+	<channel-type id="stationId" advanced="true">
+		<item-type>String</item-type>
+		<label>Station Id</label>
+		<description>Weather station identifier</description>
+		<state readOnly="true" pattern="%s"></state>
+	</channel-type>
+
+	<channel-type id="observationTime" advanced="true">
+		<item-type>DateTime</item-type>
+		<label>Observation Time</label>
+		<description>Observation date and time</description>
+		<state readOnly="true"></state>
+	</channel-type>
+
+	<channel-type id="forecastTime" advanced="true">
+		<item-type>DateTime</item-type>
+		<label>Forecast Time</label>
+		<description>Forecast date and time</description>
+		<state readOnly="true"></state>
+	</channel-type>
+
+	<channel-type id="currentConditions">
+		<item-type>String</item-type>
+		<label>Current Conditions</label>
+		<description>Weather current conditions</description>
+		<state readOnly="true" pattern="%s"></state>
+	</channel-type>
+
+	<channel-type id="forecastConditions">
+		<item-type>String</item-type>
+		<label>Forecast Conditions</label>
+		<description>Weather forecast conditions</description>
+		<state readOnly="true" pattern="%s"></state>
+	</channel-type>
+
+	<channel-type id="temperature">
+		<item-type>Number</item-type>
+		<label>Temperature</label>
+		<description>Temperature</description>
+		<category>Temperature</category>
+		<state readOnly="true" pattern="%.1f" />
+		<config-description>
+			<parameter name="SourceUnit" type="text" required="true">
+				<label>Temperature Source Unit</label>
+				<description>Select the temperature unit provided by the Weather Underground service</description>
+				<options>
+					<option value="C">Degree Celsius</option>
+					<option value="F">Degree Fahrenheit</option>
+				</options>
+				<default>C</default>
+			</parameter>
+		</config-description>
+	</channel-type>
+
+	<channel-type id="minTemperature">
+		<item-type>Number</item-type>
+		<label>Minimum Temperature</label>
+		<description>Minimum temperature</description>
+		<category>Temperature</category>
+		<state readOnly="true" pattern="%.1f" />
+		<config-description>
+			<parameter name="SourceUnit" type="text" required="true">
+				<label>Minimum Temperature Source Unit</label>
+				<description>Select the minimum temperature unit provided by the Weather Underground service</description>
+				<options>
+					<option value="C">Degree Celsius</option>
+					<option value="F">Degree Fahrenheit</option>
+				</options>
+				<default>C</default>
+			</parameter>
+		</config-description>
+	</channel-type>
+
+	<channel-type id="maxTemperature">
+		<item-type>Number</item-type>
+		<label>Maximum Temperature</label>
+		<description>Maximum temperature</description>
+		<category>Temperature</category>
+		<state readOnly="true" pattern="%.1f" />
+		<config-description>
+			<parameter name="SourceUnit" type="text" required="true">
+				<label>Maximum Temperature Source Unit</label>
+				<description>Select the maximum temperature unit provided by the Weather Underground service</description>
+				<options>
+					<option value="C">Degree Celsius</option>
+					<option value="F">Degree Fahrenheit</option>
+				</options>
+				<default>C</default>
+			</parameter>
+		</config-description>
+	</channel-type>
+
+	<channel-type id="dewPoint" advanced="true">
+		<item-type>Number</item-type>
+		<label>Dew Point Temperature</label>
+		<description>Dew Point temperature</description>
+		<category>Temperature</category>
+		<state readOnly="true" pattern="%.1f" />
+		<config-description>
+			<parameter name="SourceUnit" type="text" required="true">
+				<label>Dew Point Temperature Source Unit</label>
+				<description>Select the dew Point temperature unit provided by the Weather Underground service</description>
+				<options>
+					<option value="C">Degree Celsius</option>
+					<option value="F">Degree Fahrenheit</option>
+				</options>
+				<default>C</default>
+			</parameter>
+		</config-description>
+	</channel-type>
+
+	<channel-type id="heatIndex" advanced="true">
+		<item-type>Number</item-type>
+		<label>Heat Index</label>
+		<description>Heat index</description>
+		<category>Temperature</category>
+		<state readOnly="true" pattern="%.1f" />
+		<config-description>
+			<parameter name="SourceUnit" type="text" required="true">
+				<label>Heat Index Source Unit</label>
+				<description>Select the heat index unit provided by the Weather Underground service</description>
+				<options>
+					<option value="C">Degree Celsius</option>
+					<option value="F">Degree Fahrenheit</option>
+				</options>
+				<default>C</default>
+			</parameter>
+		</config-description>
+	</channel-type>
+
+	<channel-type id="windChill" advanced="true">
+		<item-type>Number</item-type>
+		<label>Wind Chill Temperature</label>
+		<description>Wind chill temperature</description>
+		<category>Temperature</category>
+		<state readOnly="true" pattern="%.1f" />
+		<config-description>
+			<parameter name="SourceUnit" type="text" required="true">
+				<label>Wind Chill Temperature Source Unit</label>
+				<description>Select the wind chill temperature unit provided by the Weather Underground service</description>
+				<options>
+					<option value="C">Degree Celsius</option>
+					<option value="F">Degree Fahrenheit</option>
+				</options>
+				<default>C</default>
+			</parameter>
+		</config-description>
+	</channel-type>
+
+	<channel-type id="feelingTemperature" advanced="true">
+		<item-type>Number</item-type>
+		<label>Feeling Temperature</label>
+		<description>Feeling temperature</description>
+		<category>Temperature</category>
+		<state readOnly="true" pattern="%.1f" />
+		<config-description>
+			<parameter name="SourceUnit" type="text" required="true">
+				<label>Feeling Temperature Source Unit</label>
+				<description>Select the Feeling temperature unit provided by the Weather Underground service</description>
+				<options>
+					<option value="C">Degree Celsius</option>
+					<option value="F">Degree Fahrenheit</option>
+				</options>
+				<default>C</default>
+			</parameter>
+		</config-description>
+	</channel-type>
+
+	<channel-type id="relativeHumidity">
+		<item-type>Number</item-type>
+		<label>Relative Humidity</label>
+		<description>Relative humidity in %</description>
+		<category>Humidity</category>
+		<state readOnly="true" min="0" max="100" pattern="%d %%">
+		</state>
+	</channel-type>
+
+	<channel-type id="windDirection">
+		<item-type>String</item-type>
+		<label>Wind Direction</label>
+		<description>Wind direction</description>
+		<state readOnly="true" pattern="%s">
+		</state>
+	</channel-type>
+
+	<channel-type id="maxWindDirection" advanced="true">
+		<item-type>String</item-type>
+		<label>Maximum Wind Direction</label>
+		<description>Maximum wind direction</description>
+		<state readOnly="true" pattern="%s">
+		</state>
+	</channel-type>
+
+	<channel-type id="averageWindDirection" advanced="true">
+		<item-type>String</item-type>
+		<label>Average Wind Direction</label>
+		<description>Average wind direction</description>
+		<state readOnly="true" pattern="%s">
+		</state>
+	</channel-type>
+
+	<channel-type id="windDirection-degrees" advanced="true">
+		<item-type>Number</item-type>
+		<label>Wind Direction (°)</label>
+		<description>Wind direction in degrees</description>
+		<category>Wind</category>
+		<state readOnly="true" min="0" max="360" pattern="%.0f °">
+		</state>
+	</channel-type>
+
+	<channel-type id="maxWindDirection-degrees" advanced="true">
+		<item-type>Number</item-type>
+		<label>Maximum Wind Direction (°)</label>
+		<description>Maximum wind direction in degrees</description>
+		<category>Wind</category>
+		<state readOnly="true" min="0" max="360" pattern="%.0f °">
+		</state>
+	</channel-type>
+
+	<channel-type id="averageWindDirection-degrees" advanced="true">
+		<item-type>Number</item-type>
+		<label>Average Wind Direction (°)</label>
+		<description>Average wind direction in degrees</description>
+		<category>Wind</category>
+		<state readOnly="true" min="0" max="360" pattern="%.0f °">
+		</state>
+	</channel-type>
+
+	<channel-type id="windSpeed">
+		<item-type>Number</item-type>
+		<label>Wind Speed</label>
+		<description>Wind speed</description>
+		<category>Wind</category>
+		<state readOnly="true" pattern="%.1f" />
+		<config-description>
+			<parameter name="SourceUnit" type="text" required="true">
+				<label>Wind Speed Source Unit</label>
+				<description>Select the wind speed unit provided by the Weather Underground service</description>
+				<options>
+					<option value="kmh">Kilometers per hour</option>
+					<option value="mph">Miles per hour</option>
+				</options>
+				<default>kmh</default>
+			</parameter>
+		</config-description>
+	</channel-type>
+
+	<channel-type id="maxWindSpeed">
+		<item-type>Number</item-type>
+		<label>Maximum Wind Speed</label>
+		<description>Maximum wind speed</description>
+		<category>Wind</category>
+		<state readOnly="true" pattern="%.1f" />
+		<config-description>
+			<parameter name="SourceUnit" type="text" required="true">
+				<label>Maximum Wind Speed Source Unit</label>
+				<description>Select the maximum wind speed unit provided by the Weather Underground service</description>
+				<options>
+					<option value="kmh">Kilometers per hour</option>
+					<option value="mph">Miles per hour</option>
+				</options>
+				<default>kmh</default>
+			</parameter>
+		</config-description>
+	</channel-type>
+
+	<channel-type id="averageWindSpeed">
+		<item-type>Number</item-type>
+		<label>Average Wind Speed</label>
+		<description>Average wind speed</description>
+		<category>Wind</category>
+		<state readOnly="true" pattern="%.1f" />
+		<config-description>
+			<parameter name="SourceUnit" type="text" required="true">
+				<label>Average Wind Speed Source Unit</label>
+				<description>Select the average wind speed unit provided by the Weather Underground service</description>
+				<options>
+					<option value="kmh">Kilometers per hour</option>
+					<option value="mph">Miles per hour</option>
+				</options>
+				<default>kmh</default>
+			</parameter>
+		</config-description>
+	</channel-type>
+
+	<channel-type id="windGust">
+		<item-type>Number</item-type>
+		<label>Wind Gust</label>
+		<description>Wind gust</description>
+		<category>Wind</category>
+		<state readOnly="true" pattern="%.1f" />
+		<config-description>
+			<parameter name="SourceUnit" type="text" required="true">
+				<label>Wind Gust Source Unit</label>
+				<description>Select the wind Gust unit provided by the Weather Underground service</description>
+				<options>
+					<option value="kmh">Kilometers per hour</option>
+					<option value="mph">Miles per hour</option>
+				</options>
+				<default>kmh</default>
+			</parameter>
+		</config-description>
+	</channel-type>
+
+	<channel-type id="pressure">
+		<item-type>Number</item-type>
+		<label>Pressure</label>
+		<description>Pressure</description>
+		<category>Pressure</category>
+		<state readOnly="true" pattern="%.3f" />
+		<config-description>
+			<parameter name="SourceUnit" type="text" required="true">
+				<label>Pressure Source Unit</label>
+				<description>Select the pressure unit provided by the Weather Underground service</description>
+				<options>
+					<option value="hPa">Hectopascal</option>
+					<option value="inHg">Inch of mercury</option>
+				</options>
+				<default>hPa</default>
+			</parameter>
+		</config-description>
+	</channel-type>
+
+	<channel-type id="visibility" advanced="true">
+		<item-type>Number</item-type>
+		<label>Visibility</label>
+		<description>Visibility</description>
+		<state readOnly="true" pattern="%.1f" />
+		<config-description>
+			<parameter name="SourceUnit" type="text" required="true">
+				<label>Visibility Source Unit</label>
+				<description>Select the visibility unit provided by the Weather Underground service</description>
+				<options>
+					<option value="km">Kilometers</option>
+					<option value="mi">Miles</option>
+				</options>
+				<default>km</default>
+			</parameter>
+		</config-description>
+	</channel-type>
+
+	<channel-type id="solarRadiation" advanced="true">
+		<item-type>Number</item-type>
+		<label>Solar Radiation</label>
+		<description>Solar radiation in W/m2</description>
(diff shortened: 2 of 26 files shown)


lolodomo:
You can download [the jar file for testing Weather Underground binding](http://lg.hc.free.fr/openHAB/org.eclipse.smarthome.binding.weatherunderground-0.9.0-SNAPSHOT.jar)

mhilbush:
@lolodomo  I haven't been able to spend much time testing this.  One usability thing I noticed...  The Location configuration label and description are not very clear.  The binding's Location label conflicts with the existing Thing Location field, which is a bit confusing.  And, the Location description could be a little more descriptive of what's expected in the field.  While I do realize the README gives plenty of examples, of valid Locations, modifying the description may help reduce some head scratching.

lolodomo:
@mhilbush : what do you propose for the label and the description of this configuration setting ?

mhilbush:
> what do you propose for the label and the description

For label, if `Location of Weather Information` is not too long, this would be more descriptive.

The Yahoo weather binding references the Wiki in the description.  If that approach is acceptable, you could follow a similar approach.

Something like: 
`Multiple syntaxes are supported. See documentation at http://path-to-wunderground-readme/`

WDYT

AmpacheUser:
Can you add Solar Radiation to this Binding? I want to use it for some automating of lights.

lolodomo:
@AmpacheUser : I can add a channel for Solar Radiation. Any idea what is the range and the unit for such a value ?

AmpacheUser:
@lolodomo The Units are Watts/sq. m (W/m2)
The Range is  from about 0.00 to 2000.00
And the Precision is out to 2 decimal places.

lolodomo:
@AmpacheUser @mhilbush : I will provide a new version tomorrow that will take into account your requests or remarks.

lolodomo:
Updated binding:

- Add a channel for solar radiation
- Add a channel for the icon URL representing the weather current conditions
- Add a channel for the icon URL representing the weather forecast conditions
- Update the label and description for the "Location" configuration setting

AmpacheUser:
@lolodomo Thanks for the updated Version.
Unfortunately the Solar Radiation is not working for me.
I get the following error:
[WARN ] [nd.handler.WeatherUndergroundHandler] - Update channel current#solarradiation: Can't get value: org.eclipse.smarthome.binding.weatherunderground.internal.json.WeatherUndergroundJsonCurrent.getSolarradiation()

Also, there is a typo in the Channel name for Solar Radiation it is Spell like Solar Radation
Not sure if they are related.
I know that error is not very specific, if I can help get a better one let me know

lolodomo:
@AmpacheUser : how did you create the channel ?
In papaer UI, I see a channel with label "Solar Radiation" with a description "Solar radiation in W/m2" and  is my case an ID "weatherunderground:weather:maison:current#solarRadiation" and not "weatherunderground:weather:maison:current#solarradiation".
Take a look to my updated files and you will see no typo and correct channel name.
Maybe you have to delete your thing and create it again.

AmpacheUser:
@lolodomo Working now, thank you!

extensions/binding/org.eclipse.smarthome.binding.weatherunderground/src/main/java/org/eclipse/smarthome/binding/weatherunderground/internal/json/WeatherUndergroundJsonUtils.java
```diff
@@ -92,7 +92,8 @@ public static Calendar convertToCalendar(String value) {
      */
     public static Integer convertToInteger(String value) {
         Integer result = null;
-        if (value != null && !value.isEmpty() && !value.equalsIgnoreCase("N/A") && !value.equalsIgnoreCase("NA")) {
+        if (value != null && !value.isEmpty() && !value.equalsIgnoreCase("N/A") && !value.equalsIgnoreCase("NA")
+                && !value.equals("-") && !value.equals("--")) {
```
hakan42: Please change this to if (!(value == null || value.isEmpty() || value.equalsIgnoreCase(....)), this makes it better to read.

extensions/binding/org.eclipse.smarthome.binding.weatherunderground/src/main/java/org/eclipse/smarthome/binding/weatherunderground/internal/json/WeatherUndergroundJsonUtils.java
```diff
@@ -112,7 +113,8 @@ public static Integer convertToInteger(String value) {
      */
     public static BigDecimal convertToBigDecimal(String value) {
         BigDecimal result = null;
-        if (value != null && !value.isEmpty() && !value.equalsIgnoreCase("N/A") && !value.equalsIgnoreCase("NA")) {
+        if (value != null && !value.isEmpty() && !value.equalsIgnoreCase("N/A") && !value.equalsIgnoreCase("NA")
+                && !value.equals("-") && !value.equals("--")) {
```
hakan42: Please change this to if (!(value == null || value.isEmpty() || value.equalsIgnoreCase(....)), this makes it better to read.

hmerk:
@lolodomo could you please add channels for the commonId, as this value is used in Rotini for showing icons.

lolodomo:
@hmerk : I would suggest to discuss about that in https://github.com/eclipse/smarthome/issues/2649 as it would concern several bindings, in particular the Yahoo Weather binding already included in the distribution. We already started a discussion about the icons but did not finish it.

hmerk:
@lolodomo Thanks, will do, but don't really understand why. I was just asking for implementing a channel type, whose pendant already existed in 1.x weather binding. Comparing the two, there are some more channels missing, that might be usefull for our users.

lolodomo:
@hmerk : note that my binding is not a replacement of the 1.x weather binding. It is a binding specific to Weather Underground service. So I will provide channels only for data that are provided by this service. Most of WU data are already linked to channels in the binding; few of them are not, I could add channels if necessary later.

hmerk:
@lolodomo Absolutely understood and thats what I was asking for, adding channels whose data is provided by the WU.

lolodomo:
@hmerk : let me know what data you miss. All fields returned by WU that I don't use (don't link to a channel) are in comments in my Json objects.
After a quick look, I see only two potential data that could be added:
* precipitations in 1 hour (is it the last hour ?)
* heat index (any idea what it is ? what range of values ?)

hmerk:
@lolodomo : I just had alook into the 1.x weather Binding and found out, that commonId is not API based, but generated through a mapping. Therefore this missing information is not to be implemented as a channel. Sorry for bothering.

pinkfish:
Is there something specific that is blocking this merge into the main line btw?

extensions/binding/org.eclipse.smarthome.binding.weatherunderground/README.md
```diff
@@ -0,0 +1,125 @@
...
+| Current | location | String | Weather observation location |
+| Current | stationId | String | Weather station identifier |
+| Current | observationTime | DateTime | Observation date and time |
+| Current | conditions | String | Weather conditions |
+| Current | temperatureC | Number | Temperature in degrees Celsius |
+| Current | temperatureF | Number | Temperature in degrees Fahrenheit |
+| Current | relativeHumidity | Number | Relative humidity in % |
+| Current | windDirection | String | Wind direction |
+| Current | windDirectionDegrees | Number | Wind direction in degrees |
+| Current | windSpeedKmh | Number | Wind speed in km/h |
```
pinkfish: I really think we shuld just do one unit here.  Do everything in km/h or Celsius and do the conversion on the edges.  This is quite spammy with all these bonus channels :)

extensions/binding/org.eclipse.smarthome.binding.weatherunderground/src/main/java/org/eclipse/smarthome/binding/weatherunderground/handler/WeatherUndergroundHandler.java
```diff
@@ -0,0 +1,325 @@
...
+        logger.debug("config location = {}", config.location);
+        logger.debug("config language = {}", config.language);
+        logger.debug("config refresh = {}", config.refresh);
+
+        boolean validConfig = true;
+
+        if (StringUtils.trimToNull(config.apikey) == null) {
+            logger.error("Parameter 'apikey' is mandatory and must be configured, disabling thing '{}'",
+                    getThing().getUID());
+            validConfig = false;
```
pinkfish: When debugging this stuff myself I found it really useful to do a updateStatus with a text string on errors that says why the config failed.  Then just looking at the page with the binding on it I can see what went wrong, a lot easier for the user too so they don't need to dig through log files to find the issue.

extensions/binding/org.eclipse.smarthome.binding.weatherunderground/src/main/java/org/eclipse/smarthome/binding/weatherunderground/handler/WeatherUndergroundHandler.java
```diff
@@ -0,0 +1,325 @@
...
+        }
+
+        // Update the thing status
+        if (resultOk) {
+            updateStatus(ThingStatus.ONLINE);
+        } else {
+            updateStatus(ThingStatus.OFFLINE, ThingStatusDetail.OFFLINE.COMMUNICATION_ERROR, errorMsg);
+        }
+
+        return resultOk ? result : null;
```
pinkfish: Why put this into a global?  You only need it when updating the channel data right?

extensions/binding/org.eclipse.smarthome.binding.weatherunderground/README.md
maggu2810: I agree that one temperature channel should be enough. Once the unit PR is applied we could consider about a generic way to convert between Celsius, Fahrenheit, Kelvin, ...

maggu2810:
> Is there something specific that is blocking this merge into the main line btw?

As usual the lack of time and resources.

extensions/binding/org.eclipse.smarthome.binding.weatherunderground/README.md
lolodomo: But the service is providing the values in several units, I would like to keep an access to the provider raw value rather than than considering a conversion done by ESH.
But when conversion unit feature is implemented in ESH, I will change "bonus" channels into advanced channels.

extensions/binding/org.eclipse.smarthome.binding.weatherunderground/src/main/java/org/eclipse/smarthome/binding/weatherunderground/handler/WeatherUndergroundHandler.java
lolodomo: Looks acceptable ;) I will improuve that.

extensions/binding/org.eclipse.smarthome.binding.weatherunderground/src/main/java/org/eclipse/smarthome/binding/weatherunderground/handler/WeatherUndergroundHandler.java
lolodomo: The reason is that we have to consider the case of a REFRESH channel command in addition to regular refresh process. When receiving a REFRESH command for a channel, we use the result of the last request being saved in the class member.

lolodomo:
I can't re-request the data at each channel refresh command received because the number of allowed WU requests is limited. At OH startup, the binding is receiving such a REFRESH command for each channel !
By the way, it is not a problem because the weather data are not changing very often.

lolodomo:
For the unit subject, I will wait and see whether ESH conversion mechanism is able to deliver the same value as WU itself.

pinkfish:
BTW is there an issue on the ESH conversion mechanism that is tracking the
progress or the plan?

lolodomo:
Yes, PR 1857.

extensions/binding/org.eclipse.smarthome.binding.weatherunderground/README.md
maggu2810: @kaikreuzer What do you think about that? Should a binding provide channels with the same information but different units? A few month ago, I has been sure, that this does not fit into the ESH concept. But we already provided images as URL string and images now. So, I would be interested about your opinion.

lolodomo: Please remember that unit management was not (and is still not( implemented in ESH when I implemented this binding.

kaikreuzer: I agree that there should be only a single "windSpeed" channel. We could actually select the desired unit through a channel configuration property - this solution would even be independent of the unit support on number items.

lolodomo: @kaikreuzer : interesting. I was not aware that it was possible to set configuration parameters on channel types !
Is it something supported by Paper UI ? I imagine you have to set parameters before linking to an item ?
Are there existing bindings using this feature ?
I will implement it because it can reduce almost by 2 the number of my channels, and as you said that is something independent of the unit support on number items.
The problem with this solution is that the user will have to choose/set the parameter (unit) for almost all channels.

kaikreuzer: > Is it something supported by Paper UI ?

Yes, you will find an "edit" button on the right of the channels of a thing.

> I imagine you have to set parameters before linking to an item ?

No, you can change the config at any time.

> Are there existing bindings using this feature ?

Yes, see e.g. [here](https://github.com/eclipse/smarthome/blob/master/extensions/binding/org.eclipse.smarthome.binding.lifx/ESH-INF/thing/thing-types.xml#L83-L89).

lolodomo: Ok, thank you. We have to set a default value.
I discovered that there is one in NTP binding too (the default date format).
So I will add a provider unit configuration property on my channels.

lolodomo: @kaikreuzer : how can I select the vhannel configuration property in DSL (items file) ? What is the syntax ?

lolodomo: https://eclipse.org/smarthome/documentation/features/dsl.html

It explains how to add channels in the things file using an already defined channel type from the binding and how setting the configuration property. But is it possible to override a channel defined by the binding ?

kaikreuzer: Yes, afaik, if you use the same channel id, which is already defined on a thing type, it will simply override the definition, i.e. your channel configuration should be considered.

lolodomo: Tested and it works.

lolodomo:
New major update of my PR to take into account all your comments:
* new configuration property to select the source unit added to several channels. It voids having one channel defined for each unit.
* new channel for get the heat index
* new channel for get the rain fall during the last hour
* detailed error message added when setting the thing OFFLINE due to a wrong thing configuration. The message tells you in UI what's wrong in the thing configuration.
* detailed error message updated when setting the thing OFFLINE due to a communication error.
* logs level updated to avoid WARN and ERROR levels
* channels for forecast data are now in an advanced group types

Please not that lots of channel id have been changed.

lolodomo:
You can download [the jar file for testing this Weather Underground binding](http://lg.hc.free.fr/openHAB/org.eclipse.smarthome.binding.weatherunderground-0.9.0-SNAPSHOT.jar)

I will probably push this version in the IoT marketplace.

lolodomo:
Documentation finished.

extensions/binding/org.eclipse.smarthome.binding.weatherunderground/ESH-INF/thing/thing-types.xml
```diff
@@ -0,0 +1,429 @@
...
+	xsi:schemaLocation="http://eclipse.org/smarthome/schemas/thing-description/v1.0.0 http://eclipse.org/smarthome/schemas/thing-description-1.0.0.xsd">
+
+	<!-- WeatherUnderground Binding -->
+	<thing-type id="weather">
+		<label>Weather Information</label>
+		<description>Provides various weather data from the Weather Underground service</description>
+
+		<channel-groups>
+			<channel-group id="current" typeId="current" />
+			<channel-group id="forecastToday" typeId="forecast" />
```
sjka: IMHO it would make sense to also give them a label and description _here_, so users can better distinguish them. Otherwise you will see "Weather Forecast" for all of them and can only tell by looking an the ChannelUIDs for which day they actually are.

extensions/binding/org.eclipse.smarthome.binding.weatherunderground/.classpath
```diff
@@ -0,0 +1,7 @@
+<?xml version="1.0" encoding="UTF-8"?>
+<classpath>
+	<classpathentry kind="con" path="org.eclipse.jdt.launching.JRE_CONTAINER/org.eclipse.jdt.internal.debug.ui.launcher.StandardVMType/JavaSE-1.7"/>
```
sjka: You might want to update this to JavaSE-1.8, now that it is allowed to 😄

extensions/binding/org.eclipse.smarthome.binding.weatherunderground/src/main/java/org/eclipse/smarthome/binding/weatherunderground/handler/WeatherUndergroundHandler.java
```diff
@@ -0,0 +1,338 @@
...
+
+/**
+ * The {@link WeatherUndergroundHandler} is responsible for handling the
+ * weather things created to use the Weather Underground Service.
+ *
+ * @author Laurent Garnier - Initial contribution
+ */
+public class WeatherUndergroundHandler extends BaseThingHandler {
+
+    private Logger logger = LoggerFactory.getLogger(WeatherUndergroundHandler.class);
```
sjka: would make sense to declare it `final`

lolodomo:
@sjka :  thank you for your review, I will take into account your remarks.

For my understanding, why is there a need for IP review at Eclipse ?
I thought it was required only when using a new library.

sjka:
> For my understanding, why is there a need for IP review at Eclipse ?
> I thought it was required only when using a new library.

Not only. Also, contributions with >1000 lines of code have to be reviewed. There is a pretty good overview slide of the IP process [here](https://eclipse.org/legal/EclipseLegalProcessPoster.pdf), if you are interested. It normally doesn't take too long, just a couple of days usually.

extensions/binding/org.eclipse.smarthome.binding.weatherunderground/src/main/java/org/eclipse/smarthome/binding/weatherunderground/handler/WeatherUndergroundHandler.java
lolodomo: Ok

extensions/binding/org.eclipse.smarthome.binding.weatherunderground/src/main/java/org/eclipse/smarthome/binding/weatherunderground/handler/WeatherUndergroundHandler.java
```diff
@@ -0,0 +1,338 @@
...
+    public WeatherUndergroundHandler(Thing thing) {
+        super(thing);
+
+        gson = new Gson();
+    }
+
+    @Override
+    public void initialize() {
+        logger.debug("Initializing WeatherUnderground handler.");
+        super.initialize();
```
lolodomo: Right. Changed.

extensions/binding/org.eclipse.smarthome.binding.weatherunderground/src/main/java/org/eclipse/smarthome/binding/weatherunderground/handler/WeatherUndergroundHandler.java
```diff
@@ -0,0 +1,338 @@
...
+            refreshJob = null;
+        }
+    }
+
+    @Override
+    public void handleCommand(ChannelUID channelUID, Command command) {
+        if (command instanceof RefreshType) {
+            updateChannel(channelUID.getId());
+        } else {
+            logger.info("The Weather Underground binding is a read-only binding and can not handle command {}",
```
lolodomo: Fixed

extensions/binding/org.eclipse.smarthome.binding.weatherunderground/ESH-INF/thing/thing-types.xml
lolodomo: Label and description added.

extensions/binding/org.eclipse.smarthome.binding.weatherunderground/ESH-INF/thing/thing-types.xml
```diff
@@ -0,0 +1,429 @@
...
+			<channel id="windGust" typeId="windSpeed">
+				<label>Wind Gust</label>
+				<description>Wind gust</description>
+			</channel>
+			<channel id="pressure" typeId="pressure" />
+			<channel id="dewPoint" typeId="temperature">
+				<label>Dew Point Temperature</label>
+				<description>Dew Point temperature</description>
+			</channel>
+            <channel id="heatIndex" typeId="temperature">
```
lolodomo: Fixed

extensions/binding/org.eclipse.smarthome.binding.weatherunderground/src/main/java/org/eclipse/smarthome/binding/weatherunderground/handler/WeatherUndergroundHandler.java
```diff
@@ -0,0 +1,338 @@
...
+
+        try {
+
+            // Build a valid URL for the Weather Underground service using
+            // the requested features and the thing configuration settings
+            WeatherUndergroundConfiguration config = getConfigAs(WeatherUndergroundConfiguration.class);
+
+            String urlStr = URL.replace("%APIKEY%", StringUtils.trimToEmpty(config.apikey));
+
+            String featuresStr = "";
```
lolodomo: Cool. Didn't know this method.

extensions/binding/org.eclipse.smarthome.binding.weatherunderground/src/main/java/org/eclipse/smarthome/binding/weatherunderground/internal/json/WeatherUndergroundJsonCurrent.java
```diff
@@ -0,0 +1,537 @@
...
+/**
+ * The {@link WeatherUndergroundJsonCurrent} is the Java class used
+ * to map the entry "current_observation" from the JSON response to a Weather
+ * Underground request.
+ *
+ * @author Laurent Garnier - Initial contribution
+ */
+public class WeatherUndergroundJsonCurrent {
+
+    // private Object image;
```
lolodomo: Comment added.
Your assumption was correct.

extensions/binding/org.eclipse.smarthome.binding.weatherunderground/src/main/java/org/eclipse/smarthome/binding/weatherunderground/internal/json/WeatherUndergroundJsonUtils.java
```diff
@@ -0,0 +1,122 @@
...
+import org.slf4j.LoggerFactory;
+
+/**
+ * The {@link WeatherUndergroundJsonUtils} class contains utilities methods.
+ *
+ * @author Laurent Garnier - Initial contribution
+ */
+public class WeatherUndergroundJsonUtils {
+
+    private static Logger logger = LoggerFactory.getLogger(WeatherUndergroundJsonUtils.class);
```
lolodomo: Ok, suppressed.

extensions/binding/org.eclipse.smarthome.binding.weatherunderground/src/main/java/org/eclipse/smarthome/binding/weatherunderground/internal/json/WeatherUndergroundJsonUtils.java
```diff
@@ -0,0 +1,122 @@
...
+     *         or null in case of conversion error
+     */
+    public static Calendar convertToCalendar(String value) {
+        Calendar result = null;
+        if (value != null && !value.isEmpty() && !value.equalsIgnoreCase("N/A") && !value.equalsIgnoreCase("NA")) {
+            try {
+                result = Calendar.getInstance();
+                result.setTimeInMillis(Long.valueOf(value) * 1000);
+            } catch (NumberFormatException e) {
+                logger.debug("Cannot convert {} to Calendar", value);
```
lolodomo: Done

extensions/binding/org.eclipse.smarthome.binding.weatherunderground/src/main/java/org/eclipse/smarthome/binding/weatherunderground/internal/json/WeatherUndergroundJsonUtils.java
```diff
@@ -0,0 +1,122 @@
...
+     * @return the Integer object representing the value or null in case of conversion error
+     */
+    public static Integer convertToInteger(String value) {
+        Integer result = null;
+        if (value != null && !value.isEmpty() && !value.equalsIgnoreCase("N/A") && !value.equalsIgnoreCase("NA")
+                && !value.equals("-") && !value.equals("--")) {
+            try {
+                result = Integer.valueOf(value.trim());
+            } catch (NumberFormatException e) {
+                logger.debug("Cannot convert {} to Integer", value);
```
lolodomo: Done

extensions/binding/org.eclipse.smarthome.binding.weatherunderground/src/main/java/org/eclipse/smarthome/binding/weatherunderground/internal/json/WeatherUndergroundJsonUtils.java
```diff
@@ -0,0 +1,122 @@
...
+     * Convert a string representing an Epoch value into a Calendar object
+     *
+     * @param value the Epoch value as a string
+     *
+     * @return the Calendar object representing the date and time of the Epoch
+     *         or null in case of conversion error
+     */
+    public static Calendar convertToCalendar(String value) {
+        Calendar result = null;
+        if (value != null && !value.isEmpty() && !value.equalsIgnoreCase("N/A") && !value.equalsIgnoreCase("NA")) {
```
lolodomo: Done

extensions/binding/org.eclipse.smarthome.binding.weatherunderground/.classpath
lolodomo: Shall I change `Bundle-RequiredExecutionEnvironment: JavaSE-1.7` to Bundle`-RequiredExecutionEnvironment: JavaSE-1.8` in MANIFEST.MF ?

extensions/binding/org.eclipse.smarthome.binding.weatherunderground/src/main/java/org/eclipse/smarthome/binding/weatherunderground/handler/WeatherUndergroundHandler.java
```diff
@@ -0,0 +1,338 @@
...
+            } else {
+                urlStr = urlStr.replace("%SETTINGS%", "lang:" + lang.toUpperCase());
+            }
+
+            urlStr = urlStr.replace("%QUERY%", location);
+            logger.debug("URL = {}", urlStr);
+
+            // Run the HTTP request and get the JSON response from Weather Underground
+            URL url = new URL(urlStr);
+            URLConnection connection = url.openConnection();
```
lolodomo: Changed.
Isn't it too much to use a 30 seconds timeout ?

lolodomo:
@sjka : review comments taken into account. I have not squashed to let you see what was changed.

I have 2 remaining questions, one relative to a change to do or not in the MANIFEST.MF file and one relative to the 30 seconds timeout for the HTTP request.

extensions/binding/org.eclipse.smarthome.binding.weatherunderground/src/main/java/org/eclipse/smarthome/binding/weatherunderground/handler/WeatherUndergroundHandler.java
sjka: Keep in mind: It's the timeout for the whole request/response conversation. I'd rather set it too high than a little too low. The important point is to _have_ a timeout...

To be honest, I didn't spend too many thoughts for the concrete number, just consider it as an example and chose whatever sound reasonable to you.

lolodomo:
Any answer to my question about the manifest file ?
If not, I will squash and we will can then go with CQ.

sjka:
Sorry, I thought I did, but apparently I did not sent it. My apologies!
Yes, please update the manifest as well.

extensions/binding/org.eclipse.smarthome.binding.weatherunderground/src/main/java/org/eclipse/smarthome/binding/weatherunderground/handler/WeatherUndergroundHandler.java
```diff
@@ -0,0 +1,325 @@
...
+            refreshJob = null;
+        }
+    }
+
+    @Override
+    public void handleCommand(ChannelUID channelUID, Command command) {
+        if (command instanceof RefreshType) {
+            updateChannel(channelUID.getId());
+        } else {
+            logger.info("The Weather Underground binding is a read-only binding and cannot handle command {}", command);
```
kaikreuzer: Reduce to debug as maybe another binding is linked to the same item and can handle the command - so your message might be noise.

extensions/binding/org.eclipse.smarthome.binding.weatherunderground/src/main/java/org/eclipse/smarthome/binding/weatherunderground/handler/WeatherUndergroundHandler.java
```diff
@@ -0,0 +1,325 @@
...
+                        resultOk = false;
+                        errorDetail = "missing forecast sub-object";
+                    }
+                }
+            }
+
+            if (!resultOk) {
+                error = "Error in Weather Underground response";
+            }
+
```
kaikreuzer: remove empty line

extensions/binding/pom.xml
```diff
@@ -25,6 +25,7 @@
     <module>org.eclipse.smarthome.binding.ntp</module>
     <module>org.eclipse.smarthome.binding.ntp.test</module>
     <module>org.eclipse.smarthome.binding.sonos</module>
+    <module>org.eclipse.smarthome.binding.weatherunderground</module>
```
kaikreuzer: Could you please also add the binding to the [p2 feature](https://github.com/eclipse/smarthome/blob/master/features/org.eclipse.smarthome.feature.runtime.binding/feature.xml) and as well to the [Karaf feature](https://github.com/eclipse/smarthome/blob/master/features/karaf/esh-ext/src/main/feature/feature.xml)?

extensions/binding/org.eclipse.smarthome.binding.weatherunderground/ESH-INF/thing/thing-types.xml
```diff
@@ -0,0 +1,459 @@
...
+
+		<config-description>
+			<parameter name="apikey" type="text" required="true">
+				<context>password</context>
+				<label>API Key</label>
+				<description>API key to access the Weather Underground service</description>
+			</parameter>
+			<parameter name="location" type="text" required="true">
+				<label>Location of Weather Information</label>
+				<description>Multiple syntaxes are supported. See documentation at http://docs.openhab.org/addons/bindings/weatherunderground/readme.html</description>
```
lolodomo: Fixed.

extensions/binding/org.eclipse.smarthome.binding.weatherunderground/ESH-INF/thing/thing-types.xml
```diff
@@ -0,0 +1,459 @@
...
+			</parameter>
+			<parameter name="language" type="text" required="false">
+				<label>Language</label>
+				<description>Language to be considered by the Weather Underground service</description>
```
lolodomo: Used

extensions/binding/org.eclipse.smarthome.binding.weatherunderground/ESH-INF/thing/thing-types.xml
```diff
@@ -0,0 +1,459 @@
...
+					<option value="TK">Turkmen</option>
+					<option value="UA">Ukrainian</option>
+					<option value="UZ">Uzbek</option>
+					<option value="VU">Vietnamese</option>
+					<option value="CY">Welsh</option>
+					<option value="SN">Wolof</option>
+					<option value="JI">Yiddish - transliterated</option>
+					<option value="YI">Yiddish - unicode</option>
+				</options>
+				<default>EN</default>
```
lolodomo: It requires that I define a reference to org.eclipse.smarthome.core.i18n.LocaleProvider in my OSGI ?

extensions/binding/org.eclipse.smarthome.binding.weatherunderground/ESH-INF/thing/thing-types.xml
```diff
@@ -0,0 +1,459 @@
...
+
+	<channel-type id="temperature">
+		<item-type>Number</item-type>
+		<label>Temperature</label>
+		<description>Temperature</description>
+		<category>Temperature</category>
+		<state readOnly="true" pattern="%.1f" />
+		<config-description>
+			<parameter name="SourceUnit" type="text" required="true">
+				<label>Source Temperature Unit</label>
```
lolodomo: By "source", I mean I ask the binding to use as source of data the °C or °F value provided by WU.
So currently, while unit management is not yet part of ESH, the binding is sending state update using the unit defined by the source configuration.
Later, when unit management is really part of ESH, I think the idea would be to create a Quantity as state, using the source setting, but then the user could choose to display °C or °F using the conversion mechanism of the new unit management.

extensions/binding/org.eclipse.smarthome.binding.weatherunderground/ESH-INF/thing/thing-types.xml
```diff
@@ -0,0 +1,459 @@
...
+
+	<channel-type id="probaPrecipitation">
+		<item-type>Number</item-type>
+		<label>Precipitation Probability</label>
+		<description>Probability of precipitation in %</description>
+		<state readOnly="true" min="0" max="100" pattern="%d %%">
+		</state>
+	</channel-type>
+
+	<channel-type id="iconUrl" advanced="true">
```
lolodomo: When implementing this binding, I think I have not yet added support for image element in Basic UI and Classic UI. It makes no sense at this time to use Image rather than URL.
The discussion relative to weather icons is still opened (in the weather general issue). Waiting for a conclusion, I just implemented an image.
I am now moving from string URL to Image.

extensions/binding/org.eclipse.smarthome.binding.weatherunderground/META-INF/MANIFEST.MF
```diff
@@ -0,0 +1,26 @@
+Manifest-Version: 1.0
+Bundle-ManifestVersion: 2
+Bundle-Name: WeatherUnderground Binding
+Bundle-SymbolicName: org.eclipse.smarthome.binding.weatherunderground;singleton:=true
+Bundle-Vendor: Eclipse.org/SmartHome
+Bundle-Version: 0.9.0.qualifier
+Bundle-RequiredExecutionEnvironment: JavaSE-1.7
```
lolodomo: Fixed

extensions/binding/org.eclipse.smarthome.binding.weatherunderground/README.md
```diff
@@ -0,0 +1,168 @@
...
+layout: documentation
+---
+
+{% include base.html %}
+
+# WeatherUnderground Binding
+
+This binding uses the [Weather Underground service](https://www.wunderground.com/weather/api/) for providing weather information for any location worldwide.
+
+The Weather Underground API is provided by The Weather Underground, LLC (WUL) free of charge but there is a daily limit and minute rate limit to the number of requests that can be made to the API for free. You may exceed this threshold only if you are or become a fee paying subscriber. WUL will monitor your daily usage of the API to determine if you have exceeded the free-use threshold by using an API key. By using this binding, you confirm that you agree with the [Weather Underground API terms and conditions of use](https://www.wunderground.com/weather/api/d/terms.html).
```
lolodomo: Changed.

extensions/binding/org.eclipse.smarthome.binding.weatherunderground/src/main/java/org/eclipse/smarthome/binding/weatherunderground/handler/WeatherUndergroundHandler.java
```diff
@@ -0,0 +1,325 @@
...
+            }
+        }
+        if (config.refresh != null && config.refresh < 5) {
+            errors += " Parameter 'refresh' must be at least 5 minutes.";
+            validConfig = false;
+        }
+        errors = errors.trim();
+
+        if (validConfig) {
+            updateStatus(ThingStatus.ONLINE);
```
lolodomo: Ok, I set it to UNKNOWN.
It is important to have a status early to avoid having the thing remaining in INITIALIZING state.

extensions/binding/org.eclipse.smarthome.binding.weatherunderground/src/main/java/org/eclipse/smarthome/binding/weatherunderground/handler/WeatherUndergroundHandler.java
```diff
@@ -0,0 +1,325 @@
...
+            startAutomaticRefresh();
+        } else {
+            logger.info("Disabling thing '{}': {}", getThing().getUID(), errors);
```
lolodomo: Changed to debug.

extensions/binding/org.eclipse.smarthome.binding.weatherunderground/src/main/java/org/eclipse/smarthome/binding/weatherunderground/handler/WeatherUndergroundHandler.java
lolodomo: Changed to debug.

extensions/binding/org.eclipse.smarthome.binding.weatherunderground/src/main/java/org/eclipse/smarthome/binding/weatherunderground/handler/WeatherUndergroundHandler.java
lolodomo: Done

extensions/binding/org.eclipse.smarthome.binding.weatherunderground/ESH-INF/thing/thing-types.xml
```diff
@@ -0,0 +1,459 @@
...
+			<channel id="location" typeId="location" />
+			<channel id="stationId" typeId="stationId" />
+			<channel id="observationTime" typeId="time" />
+			<channel id="conditions" typeId="conditions" />
+			<channel id="temperature" typeId="temperature" />
+			<channel id="relativeHumidity" typeId="relativeHumidity" />
+			<channel id="windDirection" typeId="windDirection" />
+			<channel id="windDirectionDegrees" typeId="windDirection-degrees" />
+			<channel id="windSpeed" typeId="windSpeed" />
+			<channel id="windGust" typeId="windSpeed">
```
lolodomo: Lots of work but now changed.

extensions/binding/org.eclipse.smarthome.binding.weatherunderground/ESH-INF/thing/thing-types.xml
```diff
@@ -0,0 +1,459 @@
...
+		<description>Weather conditions</description>
+		<state readOnly="true" pattern="%s"></state>
+	</channel-type>
+
+	<channel-type id="temperature">
+		<item-type>Number</item-type>
+		<label>Temperature</label>
+		<description>Temperature</description>
+		<category>Temperature</category>
+		<state readOnly="true" pattern="%.1f" />
```
lolodomo: You're right for the current missing unit in the pattern.
I will update that later when unit management is really available in ESH and we have a concrete example showing to me how to solve this "problem".

Whatever the unit set for the overall system, I really would like to keep for this binding the ability for the user to select the source unit. If the output unit and the source unti are different, there would be an automatic conversion done by the binding using the new ESH unit system.

extensions/binding/pom.xml
lolodomo: Done

extensions/binding/org.eclipse.smarthome.binding.weatherunderground/ESH-INF/thing/thing-types.xml
lolodomo: Ok, I see how it is done by the NTP binding.

lolodomo: Can I have access to the different options from the binding ? How ?

extensions/binding/org.eclipse.smarthome.binding.weatherunderground/src/main/java/org/eclipse/smarthome/binding/weatherunderground/handler/WeatherUndergroundHandler.java
```diff
@@ -0,0 +1,325 @@
...
+                    try {
+                        // Request new weather data to the Weather Underground service
+                        updateWeatherData();
+
+                        // Update all channels from the updated weather data
+                        for (Channel channel : getThing().getChannels()) {
+                            updateChannel(channel.getUID().getId());
+                        }
+                    } catch (Exception e) {
+                        logger.debug("Exception occurred during execution: {}", e.getMessage());
```
lolodomo: Such catch exception is required to avoid stopping the task scheduling. We have a general issue opened for a better handling of exceptions in scheduled tasks.

The change of status is already done in updateWeatherData(). There is already some catch exceptions in updateWeatherData(), so when an exception is catched here, it should be a kind of bug. I have added the stack trace.

lolodomo:
A new version is available, taking into account almost all @kaikreuzer comments.

What is not yet taken into account is the unit management. I have done no change. When the new unit management will be available in ESH and used as example in the Yahoo Weather binding, I will consider an update of the binding.

We have now channels "icon" linked to image items. But the general discussion about weather icons is still opened. When a solution is found and acknowledged by everybody for all weather bindings, I will update the binding.

"language" configuration setting has now no default value, the language from the system locale is now used if the configuration setting is not set. I have not checked in the code that the system locale language is matching one of the accepted language value by WU.

The documentation has been updated.

lolodomo:
Little change applied for image download.

lolodomo:
When PR #3457 is merged, I will update the binding.

extensions/binding/org.eclipse.smarthome.binding.weatherunderground/src/main/java/org/eclipse/smarthome/binding/weatherunderground/handler/WeatherUndergroundHandler.java
sjka: Why that?

If initializing actually takes let's say a minute, than it is okay to stay in INITIALIZING status for that time. That's how it is. It is just important to return from the `initialize()` method as quickly as possible.

With UNKNOWN you signal that the handle is completely initialized and ready work, i.e. accept and handle commands.

extensions/binding/org.eclipse.smarthome.binding.weatherunderground/ESH-INF/thing/thing-types.xml
sjka: They are part of the ConfigDescriptions which are not directly referenced by the configuration object, therefore I'm not aware of an "official" way to get these.
May I ask for the use-case? Because normally it's the other way round: The XML files simply describe what the ThingHandler expects or takes for granted.

extensions/binding/org.eclipse.smarthome.binding.weatherunderground/src/main/java/org/eclipse/smarthome/binding/weatherunderground/handler/WeatherUndergroundHandler.java
lolodomo: We have in the pastr a lot of major problems with bindings remaining for ever in INITIALIZING state. It was due to the fact the thing handler status was set TO ONLINE or OFFLINE too late.
Setting an initial status very early and having code that can take time moved from initialize() to a scheduled task succeeded avoiding the INITIALIZING issue.

In the WU binding, the first retrieval of data is run in a task scheduled by the initialize() because it can take some time.

In case there was a major enhancement in the startup management of bindings, I can try to suppress this early status set and check again how the binding behaves. But it is something that takes time, not easy because it depends on the order of bindings startup; ...

extensions/binding/org.eclipse.smarthome.binding.weatherunderground/ESH-INF/thing/thing-types.xml
lolodomo: The use case was described by @kaikreuzer just before. He suggested that if the setting is not set, we use the language from system locale. In this case, it would be better to check that the value is one of the accepted value by WU.
Of course, a solution would be to define in the code a set with all the accepted values. The information will just be in two places.

extensions/binding/org.eclipse.smarthome.binding.weatherunderground/src/main/java/org/eclipse/smarthome/binding/weatherunderground/handler/WeatherUndergroundHandler.java
sjka: You are mixing up two things here.

We indeed had some issues with the startup when we first introduced the asynchronous parsing of the XMLs. And if there still are bugs in the framework, we should raise them and fix them there, not in the bindings.

There also were a couple of buggy bindings, not setting a status at all under some circumstances. Then these bindings should be fixed. 

The [contract](https://www.eclipse.org/smarthome/documentation/concepts/things.html) is pretty clear: INITIALIZING may take as long as it needs, UNKNOWN means initialization is complete and the handler is ready to work. The framework is relying on that!

lolodomo: I will suppress the initial status set and see what happens.

maggu2810: > The [contract](https://www.eclipse.org/smarthome/documentation/concepts/things.html) is

@sjka It seems that the table format is not correct.
The rows between INITIALIZING and OFFLINE seems to be incorrect.

Perhaps it is already fixed but not regenerated...
If it still persists, we should create an issue.

screenshot from 2017-05-18 22-58-52

sjka: @maggu I just updated the website and checked: the table looks okay now.

lolodomo:
I just pushed a new version
* fixing conflicts
* using the new downloadImage image from HttpUtil for the icon image
* suppressing the import of package org.apache.commons.io
* fixing localeProvider the same as PR #3717 for NTP binding

extensions/binding/org.eclipse.smarthome.binding.weatherunderground/src/main/java/org/eclipse/smarthome/binding/weatherunderground/handler/WeatherUndergroundHandler.java
lolodomo: @sjka : I suppressed the initial set of state to UNKNOWN. Are you sure there will be no issue in case the handler is disposed before the status is set to a value ?

maggu2810: I assume you need to ensure that the stop is not scheduled if the thing handler has been delayed.

sjka:
Thanks for addressing all the comments. So apart from the units topic it looks like the ball is in our field now. And while #1857 is still being in process, it somehow hurts to let this binding hanging around here for so long. The adaption to the units supports supposably will lead to some incompatible changes within this binding. As it might do for others. 

In any case, the adaption to the units stuff can be done in a separate PR, no matter when this is going to be merged. So my proposal would be to move forward with this PR and adapt it as soon as the quantity types are available, unless anybody objects.

I will therefore create a CQ for this contribution in order to get the merge approval as soon as possible. Therefore we need to "freeze" _this_ PR, please don't push anything anymore.

lolodomo:
What a good mews, thank you.
I was asking myself how many months it will take to have this binding finaly merged ;)

sjka:
For the record, I have created
* CQ 13829 for general approval to use the Weather Underground service
* CQ 13828 for this PR in particular

sjka:
Good news: the CQ was approved!

sjka:
Both CQs are approved, to be precise

sjka:
FTR: I have created #3852 to also add it to the (karaf) features and include it in the docu menu

MacGeyer:
Where can the latest .jar be downloaded ?

lolodomo:
The binding is now included on all OH snapshots. No need for a specific jar.

martinvw:
Hello @lolodomo , should it be removed from the marketplace?

image

lolodomo:
I will delete the version in the marketplace as soon as openHAB version 2.2 including the binding is released. Until this time, the version in the marketplace is necessary for users running openHAB version 2.1.

martinvw:
Indeed, didn't think of that 😨 

Great binding btw! Thanks for building it!

PR #2751, opened 2017-01-04
split Karaf feature into core and ext features

kaikreuzer:
All feature names stay the same, the only change here is to have two separate Maven artifacts for the ESH features. Everything under "extensions" has been moved to a new esh-ext artifact.

This makes it easier to provide updates of core and ext independent of each other - as discussed recently, this will help decoupling extensions from the core development.

I have also changed the artifactIds to not be called "karaf" anymore, but rather "esh-", because these names also appear in a Karaf-based solution.


kaikreuzer:
yes, I thought about that option as well. If you prefer it, I will change the PR accordingly.

kaikreuzer:
done.

features/karaf/esh-core/src/main/history/dependencies.xml
```diff
@@ -0,0 +1,160 @@
+<?xml version="1.0" encoding="UTF-8" standalone="yes"?>
```
maggu2810: Please change `features/karaf*/src/main/history` to `features/karaf/*/src/main/history` in `.gitignore` and don't add files that are placed in that directory.

features/karaf/esh-ext/src/main/history/dependencies.xml
```diff
@@ -0,0 +1,63 @@
+<?xml version="1.0" encoding="UTF-8" standalone="yes"?>
```
maggu2810: See above

features/karaf/esh-tp/src/main/history/dependencies.xml
```diff
@@ -0,0 +1,113 @@
+<?xml version="1.0" encoding="UTF-8" standalone="yes"?>
```
maggu2810: See above

features/karaf/pom.xml
```diff
@@ -1,35 +1,25 @@
...
   <parent>
     <groupId>org.eclipse.smarthome</groupId>
     <artifactId>features</artifactId>
     <version>0.9.0-SNAPSHOT</version>
   </parent>
 
-  <artifactId>karaf</artifactId>
-  <packaging>feature</packaging>
+  <modelVersion>4.0.0</modelVersion>
+  <groupId>org.eclipse.smarthome</groupId>
```
maggu2810: If the groupId is equal to the parent one, there is no need to declare it explicitly (it is also okay to keep it, but not necessary).

kaikreuzer:
Sorry, I should have paid more attention. Hope everything is fixed now.

PR #2759, opened 2017-01-05
obey thing specific configuration descriptions

### LINKED ISSUE #2698: Config parameters normalisation not working for thing parameters
I'm seeing some parameters being stored in the jsondb as a STRING when they are defined as INTEGER.

My guess is that this is happening in the normalisation code. The parameters in question are in the thing config descriptions - not the thing-type.

This is more obvious when using HABmin since HABmin only sends updates to parameters that have changed - if I don't change these parameters, then nothing is sent in the REST interface, but these parameters are being changed to STRING. PaperUI is sending the parameter as a number, and it's getting stored as 12.0 in json which is still wrong, but it does at least work in this instance.

I assume that whatever code does the normalisation is using the config descriptor for the thing-type, and not the thing itself. Possibly [here](https://github.com/eclipse/smarthome/blob/48e405fb3f1e60b56333e26bd42021bd5cfc1bc8/bundles/io/org.eclipse.smarthome.io.rest.core/src/main/java/org/eclipse/smarthome/io/rest/core/thing/ThingResource.java#L671).

Having updated the zwave binding to allow [textual thing definitions](https://github.com/openhab/org.openhab.binding.zwave/issues/153), the binding no longer functions if parameters are updated with the UI.

sjka:
...for normalization of configuration parameters.

fixes #2698

### DIFF
--- a/bundles/config/org.eclipse.smarthome.config.core.test/src/test/groovy/org/eclipse/smarthome/config/core/ConfigUtilTest.groovy
+++ b/bundles/config/org.eclipse.smarthome.config.core.test/src/test/groovy/org/eclipse/smarthome/config/core/ConfigUtilTest.groovy
@@ -1,0 +1,45 @@
+/**
+ * Copyright (c) 2014-2016 by the respective copyright holders.
+ * All rights reserved. This program and the accompanying materials
+ * are made available under the terms of the Eclipse Public License v1.0
+ * which accompanies this distribution, and is available at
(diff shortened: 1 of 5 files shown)


sjka:
rebased to get a valid feedback from TravisCI

cdjackson:
I may not get to look at this for a week or two as I’m on holiday - feel free to ping me after the 25th if I haven’t got back to it...

PR #2762, opened 2017-01-05
allow overriding of channel configurations within channel groups

sjka:
...as so far such channels could not be referenced in the DSL
('#' was not allowed as part of the channel id)

fixes openhab/openhab2-addons#1264

### DIFF
--- a/bundles/model/org.eclipse.smarthome.model.thing/src/org/eclipse/smarthome/model/thing/Thing.xtext
+++ b/bundles/model/org.eclipse.smarthome.model.thing/src/org/eclipse/smarthome/model/thing/Thing.xtext
@@ -49,7 +49,7 @@ ModelThing:
 ;
 
 ModelChannel:
-	(((channelKind= ('State' | 'Trigger'))? type=ModelItemType) | 'Type' channelType=UID_SEGMENT) ':' id=UID_SEGMENT
+	(((channelKind= ('State' | 'Trigger'))? type=ModelItemType) | 'Type' channelType=UID_SEGMENT) ':' id=CHANNEL_ID
     (label=STRING)?
 	('['
 		properties+=ModelProperty (',' properties+=ModelProperty)*
@@ -62,6 +62,10 @@ ModelItemType :
 
 ModelProperty:
 	key=ID '=' value=ValueType
+;
+
+CHANNEL_ID:
+    UID_SEGMENT ('#' UID_SEGMENT)?
 ;
 
 UID:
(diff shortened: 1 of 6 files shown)


sjka:
rebased to get a valid feedback from TravisCI

bundles/model/org.eclipse.smarthome.model.thing/src/org/eclipse/smarthome/model/thing/Thing.xtext
```diff
@@ -64,6 +64,10 @@ ModelProperty:
 	key=ID '=' value=ValueType
 ;
 
+CHANNEL_UID:
```
kaikreuzer: The correct name would actually be CHANNEL_ID, not CHANNEL_UID here.

sjka:
Very good point, thanks! I have adapted the grammar accordingly.

gerrieg:
@sjka 
How should this work?
I want to set the delay of the range event in the astro binding with
```
astro:sun:home [ geolocation=".....", interval=60] {
        Channels:
                String : rise#event [
                        offset="3"
                ]
}
```

The result is:
bildschirmfoto 2017-01-11 um 17 15 28

sjka:
If you want to override a typed channel, you'd also have to reference that type, i.e.

```
astro:sun:home [ geolocation="...", interval=60] {
        Channels:
                Type rangeEvent : rise#event [
                        offset="3"
                ]
}
```

itheiss:
@sjka I have to specify the offset without quotes otherwise the value is not shown in Paper UI.

In my first pass I had this in my configuration because I thought this is a trigger channel. With the following config the value was correctly set in Paper UI. If it worked I am not 100% sure because I wasn't at home and couldn't find any hints in the logs.

```
astro:sun:home    [ geolocation="XX.XXXXXX,-XXX.XXXXXX", interval=60 ] {
	Channels:
	Trigger String : set#event [
		offset=-90
	]
}
```

gerrieg:
OK, thanks, i got it!

sjka:
> I have to specify the offset without quotes otherwise the value is not shown in Paper UI.

Very good point indeed, I didn't spot this

> I thought this is a trigger channel

Indeed, that is correct. It became a trigger channel about four weeks ago (which is great, btw...). However, with the `Type rangeEvent` you simply refer to whatever is defined by the binding in the ThingType. You not only get the type correct automatically, but also have the proper labels and descriptions.

It's good to see that you got it working.

lolodomo:
@sjka : is it possible to update in DSL the channel configuration of a binding channel defined inbinding XML ?

lolodomo:
I found the answer by testing: yes it works.

sjka:
okay 😄

PR #2766, opened 2017-01-06
fixed encoding in automation integration tests

### LINKED ISSUE #2760: [Automation] encoding of property files in automation integration tests
With #2737 we now stricly support ISO-8859-1 encoded property files only (default in Java).
As it seems, the org.eclipse.smarthome.automation.integration.test bundle contains some jar files which do not follow this standard:

See also https://hudson.eclipse.org/smarthome/job/SmartHomeDistribution-Nightly/1563/

Could you please have a look?

sjka:
😞 

fixes #2760 


kaikreuzer:
Many thanks!

Issue #2371, opened 2016-10-30
Item Label using only Map Transformation fails


J-N-K:
An item defined in a .items file with a label using a MAP transformation

`String         Feiertag_Name       "[MAP(holidays_de.map):%s]"`

leads to 

`2016-10-30 13:07:24.043 [WARN ] [.core.transform.TransformationHelper] - Cannot get service reference for transformation service of type MMAP
2016-10-30 13:07:24.044 [WARN ] [ui.internal.items.ItemUIRegistryImpl] - couldn't transform value in label because transformationService of type 'MMAP' is unavailable`

(note that the M is doubled to MMAP) while

`String         Feiertag_Name       "Test [MAP(holidays_de.map):%s]"`

works fine. (OH 2.0.0b4)

ThomDietrich:
Confirmed with latest snapshot.

bdleedy:
Workaround is a space in front of [, but I can confirm this issue.

ThomDietrich:
An item should always have a label. If you want your sitemap to not show a label, define `label=""` for the corresponding element there.

Maintainers: This is a bug. It should be either supported or forbidden to have no label defined for an item. The "MMAP" error is clearly a bug in the evaluation algorithm.

bdleedy:
If I take a String in items and then in the sitemap have Text item=thatstring, the label for the sitemap element is on the left and the label for the item is on the right. That doesn't blank the items label.

ThomDietrich:
I believe you got me wrong?

```
# item
String KodiTitle "Episode/Movie Title [%s]" (gKodi) {xbmc="<[#livingRoom|Player.Title]"}

#sitemap
Text item=KodiTitle label="" icon="kodilogo"
```

bdleedy:
Yeah, that's exactly what I'm talking about. So with that config, nothing would be on the left but "Episode..." would be on the right.

ThomDietrich:
Sorry, of course the value placeholder has to be in there. With the corrected example above:

![grafik](https://cloud.githubusercontent.com/assets/2870104/20990589/dcf59370-bcd9-11e6-9a12-68e9720c8802.png)

ThomDietrich:
Thank you @maggu2810

PR #2774, opened 2017-01-07
fix transformation in empty label

J-N-K:
issue #2371

Comment: The problem seems to be that label.indexOf is evaluated before label.trim() has been executed. In case of an empty label a space might occur before the opening square bracket. That is trimmed, but the index is 1 where it should be 0. This has been fixed by moving the .trim() to a seperate statement.


J-N-K:
Travis failing is strange. It fails in Hue binding tests, I don't think this is in any way related to this change. Plus Travis succeeds in my branch.

PR #2780, opened 2017-01-08
removed dependency="true" from features

kaikreuzer:
This setting caused the bundles not being added to KAR files, but we should not expect these bundles being available to Karaf-based solutions, so they need to be a part of our features.

### DIFF
--- a/features/karaf/esh-ext/src/main/feature/feature.xml
+++ b/features/karaf/esh-ext/src/main/feature/feature.xml
@@ -68,9 +68,9 @@
   <feature name="esh-transform-jsonpath" description="JSONPath Transformation" version="${project.version}">
     <feature>esh-base</feature>
     <bundle start-level="75">mvn:org.eclipse.smarthome.transform/org.eclipse.smarthome.transform.jsonpath/${project.version}</bundle>
-    <bundle dependency="true">mvn:com.jayway.jsonpath/json-path/2.0.0</bundle>
-    <bundle dependency="true">mvn:net.minidev/asm/1.0.2</bundle>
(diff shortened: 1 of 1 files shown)


maggu2810:
dependency=true has been choosen by intention because e.g. we don't need to install mvn:net.minidev/asm/1.0.2 of already all dependencies has been resolved. A dependency could already be resolved because the product installs another bundle that suffers all the OSGi requirements.
So, if a product already installs another version or another bundle, there is no need to also install that one.

maggu2810:
If you want to generate a kar that contains also that stuff I assume you can set the flag to ignore the dependency attribute: https://github.com/apache/karaf/blob/90fc018/tooling/karaf-maven-plugin/src/main/java/org/apache/karaf/tooling/KarMojo.java#L94

kaikreuzer:
Thanks for the hint, this seems to work.

> So, if a product already installs another version or another bundle, there is no need to also install that one.

Hm, I am actually not so sure about that (as we are not using package version constraints), but I think I get the theory and am therefore fine with closing this PR again.

maggu2810:
FYI we already use it in the TP for the two different guava version.
* mvn:de.maggu2810.p2redist/com.google.guava/10.0.1.v201203051515
* mvn:com.google.guava/guava/18.0

If esh-tp-xtext is installed and esh-kat.cpy-swagger-jersey2-1.5.8 is not installed, only the 10.0.1 is installed.
If esh-tp-xtext is installed and esh-kat.cpy-swagger-jersey2-1.5.8 is installed, only the 18.0 is installed.

PR #2783, opened 2017-01-09
caching calls to the Yahoo Weather API

### LINKED ISSUE #2779: [yahooweather] Changing the refresh interval in Paper UI leads to message about an invalid WOEID
This maybe a general problem in Paper UI but right now I only saw it in the Yahoo Binding.

I have edited the thing configuration and modified the refresh interval (from 60 to 67):
image

After I pressed the save button I got the following messages in the Karaf console:
```
21:59:12.364 [INFO ] [marthome.event.ConfigStatusInfoEvent] - ConfigStatusInfo [configStatusMessages=[]]
21:59:12.365 [INFO ] [marthome.event.ConfigStatusInfoEvent] - ConfigStatusInfo [configStatusMessages=[ConfigStatusMessage [parameterName=location, type=ERROR, messageKey=null, arguments=null, message=Der Ort f³r die WOEID 641.311 konnte nicht gefunden werden., statusCode=null]]]
```
Please note the decimal point in the WOEID.

This is not really a problem because the binding works as before. But something happend with the content of the WOEID that is probably not desired.

sjka:
...for a few seconds when using an idential query
in order to avoid hitting their limits when changing
the configuration.

fixes #2779

### DIFF
--- a/extensions/binding/org.eclipse.smarthome.binding.yahooweather/src/main/java/org/eclipse/smarthome/binding/yahooweather/handler/YahooWeatherHandler.java
+++ b/extensions/binding/org.eclipse.smarthome.binding.yahooweather/src/main/java/org/eclipse/smarthome/binding/yahooweather/handler/YahooWeatherHandler.java
@@ -16,6 +16,7 @@ import java.net.URL;
 import java.net.URLConnection;
 import java.util.ArrayList;
 import java.util.Collection;
+import java.util.concurrent.ExecutionException;
 import java.util.concurrent.ScheduledFuture;
 import java.util.concurrent.TimeUnit;
 
@@ -35,6 +36,10 @@ import org.eclipse.smarthome.core.types.State;
 import org.eclipse.smarthome.core.types.UnDefType;
 import org.slf4j.Logger;
 import org.slf4j.LoggerFactory;
+
+import com.google.common.cache.Cache;
+import com.google.common.cache.CacheBuilder;
+import com.google.common.cache.CacheLoader;
 
 /**
  * The {@link YahooWeatherHandler} is responsible for handling commands, which are
@@ -139,11 +144,12 @@ public class YahooWeatherHandler extends ConfigStatusThingHandler {
         Collection<ConfigStatusMessage> configStatus = new ArrayList<>();
 
         try {
-            String locationData = getWeatherData("SELECT location FROM weather.forecast WHERE woeid = " + location);
+            String locationData = getWeatherData(
+                    "SELECT location FROM weather.forecast WHERE woeid = " + location.toPlainString());
             String city = getValue(locationData, "location", "city");
             if (city == null) {
                 configStatus.add(ConfigStatusMessage.Builder.error(LOCATION_PARAM)
-                        .withMessageKeySuffix("location-not-found").withArguments(location).build());
+                        .withMessageKeySuffix("location-not-found").withArguments(location.toPlainString()).build());
             }
         } catch (IOException e) {
             logger.debug("Communication error occurred while getting Yahoo weather information.", e);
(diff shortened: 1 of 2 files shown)


extensions/binding/org.eclipse.smarthome.binding.yahooweather/src/main/java/org/eclipse/smarthome/binding/yahooweather/handler/YahooWeatherHandler.java
```diff
@@ -166,15 +173,27 @@ private synchronized boolean updateWeatherData() {
...
+                @Override
+                public String load(String query) throws Exception {
+                    try {
+                        URL url = new URL("https://query.yahooapis.com/v1/public/yql?format=json&q="
+                                + query.replaceAll(" ", "%20").replaceAll("'", "%27"));
+                        URLConnection connection = url.openConnection();
+                        return IOUtils.toString(connection.getInputStream());
+                    } catch (MalformedURLException e) {
+                        logger.debug("Constructed query url '{}' is not valid: {}", query, e.getMessage());
+                        return null;
```
sjka: Just notices that they don't like `null` here, so this sucks.

Hang on a second...

extensions/binding/org.eclipse.smarthome.binding.yahooweather/src/main/java/org/eclipse/smarthome/binding/yahooweather/handler/YahooWeatherHandler.java
```diff
@@ -36,6 +37,10 @@
 import org.slf4j.Logger;
 import org.slf4j.LoggerFactory;
 
+import com.google.common.cache.Cache;
+import com.google.common.cache.CacheBuilder;
```
kaikreuzer: Ouch! Sooner or later it had to come after us. Different Guava lib versions are incompatible with each other and it seems that the one in openHAB has issues with the CacheBuilder, see https://github.com/openhab/openhab/issues/4987. Not sure how to solve this, though?

sjka:
Well, there are two ways to solve it:
* ESH defines which Guava version should be used, and than that's the one. Fix openHAB.
* We simply cant' rely on Guava. I.e. we should get rid of it completely.

Short-term I can simply work around it by not using Guava here in this case. Was convenient but would work without it equally well.

kaikreuzer:
> ESH defines which Guava version should be used, and than that's the one. Fix openHAB.

Note quite right. Guava is not marked as a singleton bundle, i.e. there can be multiple versions. As ESH suggests the use of the jax-rs osgi connector + jersey, if REST API is to be used, there is also the requirement to have a more recent Guava version in place at the same time.
So the correct fix would be a version constraint in the Yahoo weather binding and ask solutions to bundle two Guava versions.

> We simply cant' rely on Guava. I.e. we should get rid of it completely.

This is probably the safest solution.

But there is a third option: Upgrade our general Guava requirement from 10.1 to something more recent, as it has been discussed elsewhere already. But this won't solve the issue in general, that is clear.

sjka:
> So the correct fix would be a version constraint in the Yahoo weather binding and ask solutions to bundle two Guava versions.

Indeed. I just noticed that openHAB currently packages only Guava 18.0.

> Upgrade our general Guava requirement from 10.1 to something more recent, as it has been discussed elsewhere already.

Oh yes, I'd love to! It was discussed here: https://www.eclipse.org/forums/index.php/t/1065860/

The strongest point against it was the increase in size, which remains valid - 1.5M to 2.2M. However, forcing solutions to bundle _two_ versions is even worse wrt size. 

Realizing that I don't know any solution which packages Version 10 anyway I have created #2801.

Issue #2764, opened 2017-01-05
Add additional contexts for config descriptions


kaikreuzer:
We currently have [this list of supported contexts](https://www.eclipse.org/smarthome/documentation/development/bindings/xml-reference.html#supported-contexts) (actively supported in the Paper UI).

Looking at rule modules, we might want to cover some more of the ESH entities:

- The RuleEnableModule requires the selection of a rule. A "Rule" context would be nice for it.
- The ChannelEventTrigger requires a trigger channel as a parameter. It would be nice to be able to use "Channel" as a context with a filter criteria "kind" with possible values "STATE" and "TRIGGER". Note that we do not have a /rest/channels endpoint, so the UI would have to traverse through the Things. Is that a problem, @aounhaider1?

danchom:
LGTM.

PR #2791, opened 2017-01-10
Support Rule context

aounhaider1:
fixes https://github.com/eclipse/smarthome/issues/2764

### DIFF
--- a/bundles/automation/org.eclipse.smarthome.automation.module.core/ESH-INF/automation/moduletypes/RuleEnableModuleType.json
+++ b/bundles/automation/org.eclipse.smarthome.automation.module.core/ESH-INF/automation/moduletypes/RuleEnableModuleType.json
@@ -28,7 +28,8 @@
 					"label": "Rules UIDs.",
 					"description": "Specifies the rules that should be enabled or disabled.",
 					"required": true,
-					"multiple": true
+                    "multiple": true,
+					"context":"rule"
 				}
 			]
 		}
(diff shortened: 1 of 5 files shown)


kaikreuzer:
Could you please also adapt https://github.com/eclipse/smarthome/blob/master/docs/documentation/development/bindings/xml-reference.md#supported-contexts?

And add the filter to the "Channel Trigger" definition and the rule context to the Rule Enablement Action, so that both use cases described in the issue are directly working?

aounhaider1:
Updated. I added the the new contexts to "Supported Contexts" definition.

docs/documentation/development/bindings/xml-reference.md
```diff
@@ -127,6 +129,14 @@ In the case of above filter only those items will be shown that satisfy the filt
 
 ```
 (type=Switch OR type=Dimmer) AND (tag=Light OR tag=Heating) 
+
+```
+Similarly, the <strong>Channel</strong> context can contain criteria to filter channels based on <strong>kind</strong> field. The value of <strong>kind</strong> can either be STATE or TRIGGER. For Example:
```
maggu2810: For example (lower case "e")

kaikreuzer:
Sorry, both modules are still not working for me. :-( Tested with Chrome and Safari.

You also missed to address this comment:

> And add the filter to the "Channel Trigger" definition and the rule context to the Rule Enablement Action, so that both use cases described in the issue are directly working?

I would think that if you do this, you will have the same results as I have.

bundles/automation/org.eclipse.smarthome.automation.module.core/ESH-INF/automation/moduletypes/RuleEnableModuleType.json
```diff
@@ -28,7 +28,8 @@
 					"label": "Rules UIDs.",
 					"description": "Specifies the rules that should be enabled or disabled.",
 					"required": true,
-					"multiple": true
+                    "multiple": true,
```
kaikreuzer: could you replace spaces by tabs?

PR #2796, opened 2017-01-11
added org.joda.time as an implicit import to scripts and rules

### LINKED ISSUE #2793: Explicit import of joda DateTime required
Some date manipulations need explicit import of org.joda.time.DateTime.
Expected behaviour is that this is not required.

```
/* test.rules */

import org.joda.time.DateTime

rule "TestNow"
when
	Time cron "*/10 * * * * ?"
then

	var DateTime dt1 = now
	var DateTime dt2 = now.withTimeAtStartOfDay

	logInfo("TestNow", "Now: " + dt1) // raises error without import
	
	if (dt1 == dt2) { // raises error without import
		logInfo("TestNow", "EQ")
	} else {
		logInfo(TestNow", "DIFF")
	}

end
```

Please see this forum thread:
https://community.openhab.org/t/timer-and-date-calculations-in-rules-need-help/19843/12?u=rtvb

kaikreuzer:
fixes #2793


maggu2810:
Do we change that again after Eclipse SmartHome has been migrated to Java 8?

kaikreuzer:
Very likely yes - and assuming java.util.DateTime is compatible (enough), users would not even have to adapt there rules then.

PR #2799, opened 2017-01-11
BasicUI: check icons existence and replace with blank when needed

resetnow:

### DIFF
--- a/extensions/binding/org.eclipse.smarthome.binding.sonos/src/main/java/org/eclipse/smarthome/binding/sonos/handler/ZonePlayerHandler.java
+++ b/extensions/binding/org.eclipse.smarthome.binding.sonos/src/main/java/org/eclipse/smarthome/binding/sonos/handler/ZonePlayerHandler.java
@@ -920,38 +920,41 @@ public class ZonePlayerHandler extends BaseThingHandler implements UpnpIOPartici
             boolean opmlUrlSucceeded = false;
             if (opmlUrl != null) {
                 String stationID = StringUtils.substringBetween(currentURI, ":s", "?sid");
-                String url = opmlUrl;
-                url = StringUtils.replace(url, "%id", stationID);
-                url = StringUtils.replace(url, "%serial", getMACAddress());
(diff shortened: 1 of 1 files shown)


resetnow:
This is the least sketchy implementation I could come up with, would gladly hear any suggestions.

extensions/ui/org.eclipse.smarthome.ui.basic/src/main/java/org/eclipse/smarthome/ui/basic/internal/WebAppIconProvider.java
```diff
@@ -0,0 +1,61 @@
...
+ * Helper class, provides icon names
+ * for sitemap widgets.
+ *
+ * @author Vlad Ivanov - Initial contribution
+ */
+public class WebAppIconProvider {
+    private static final String DEFAULT_ICON_SET = "classic";
+
+    private static List<IconProvider> iconProviders = new ArrayList<>();
+    private static ItemUIRegistry itemUIRegistry;
```
resetnow: shouldn't be static, will fix it a bit later

kaikreuzer:
Err, just to summarize: You have implemented your own icon provider, which acts as a proxy for the normal ones?

This looks really a bit nasty... Will have to think about it and come up with a better suggestion - but now it is too late at night already and I cannot think clearly anymore :-/

resetnow:
Right — it's a proxy, so might as well name it so. Another option is to move this functionality into AbstractWidgetRenderer, but then a reference would need to be added to OSGi description file for every single renderer class.

resetnow:
@kaikreuzer just checking — did you have a time to look for a better solution?

kaikreuzer:
Not yet, sorry - will come back to you on this soon!

kaikreuzer:
@resetnow Sorry for the long silence. Taking a step back, I would actually first like to understand how it comes that there are situations with missing icons. I suggest to do this discussion at https://github.com/eclipse/smarthome/issues/2994.

PR #2814, opened 2017-01-12
refactored all automation modules

kaikreuzer:
* adapted labels to conform to the new "when...then...but only if" scheme
* adapted descriptions to follow a common style
* added prefixes for module type uids to prevent clashes
* sorted generic (hidden) module type definitions into their own json files
* removed (duplicate) item state conditions that relied on an event as input; kept the item state condition that uses the item registry
* added the possibility to optionally choose an audio sink to the media modules
* renamed BasicModuleHandlerFactory to CoreModuleHandlerFactory
* implemented dedicated handlers for Item Command/StateUpdate/StateChange triggers


kaikreuzer:
Hey @danchom,

I tried to clean up the existing module types and it became quite some refactoring in the end.
Note that I have changed existing module type uids, so this is clearly a breaking change for any existing rules - but I think it was urgently time to switch to using prefixes (afair, we had discussed doing so a long long time ago), because otherwise, we will easily risk name conflicts in the future. So I think the sooner we change it, the better.

danchom:
I think it is better when each module type is placed in separate json file, named as containing module type. In this way it will be easy to see and find available module types.
For me I don't see the need to use "core." prefix for the core module types. At all the module types are not so big number and I'm sure if someone want to create a  new module type it will check  first the existing once and will not duplicate their names. The other problem with such prefixes is that these module types can be packed in bundles which does not have the prefix in their names then the name for these module types will not be suitable.

kaikreuzer:
> I think it is better when each module type is placed in separate json file, named as containing module type. In this way it will be easy to see and find available module types.

Note that you designed the json files specifically in a way that they can have multiple definitions inside.
With this PR I tried to group them in a way that it is easy to find what you are looking for - it definitely is an improvement over the current files. If you think that it should be split even more, I would suggest you do that in a follow-up PR.

> At all the module types are not so big number

At the moment, but depending on the solution, this can change quickly.

> I'm sure if someone want to create a new module type it will check first the existing once

This is exactly what I want to avoid. Remember that we are designing an open and extensible framework here. Our guideline should be that everybody prefixes his types with his own identifier, through which he can be sure that any potential clash is avoided. Asking developers to check what others (not only the core framework, but ANY potential extension from 3rd parties) have already used, is simply not acceptable.

> these module types can be packed in bundles which does not have the prefix in bundle's name

There is no requirement that the prefix has to match the bundle name - for the existing ones, this simply seemed to be the most appropriate choice.

danchom:
>Note that you designed the json files specifically in a way that they can have multiple definitions inside.
Agree, may we have to fix this in the near future.

LGTM.

kaikreuzer:
Feel free to merge then.

maggu2810:
> Note that I have changed existing module type uids, so this is clearly a breaking change for any existing rules

:scream:

PR #2822, opened 2017-01-13
add missing dependencies to designer

### LINKED ISSUE #2813: SmartHome Designer fails to start
running Java 8 v111, and eclipse smarthome v8 or v9 unzipped in the applications folder.
I see the splash screen and then an error is thrown. The log says this:

```
!SESSION 2017-01-12 21:40:22.670 -----------------------------------------------
eclipse.buildId=unknown
java.version=1.8.0_111
java.vendor=Oracle Corporation
BootLoader constants: OS=macosx, ARCH=x86_64, WS=cocoa, NL=en_US
Command-line arguments:  -os macosx -ws cocoa -arch x86_64

!ENTRY org.eclipse.core.runtime 4 0 2017-01-12 21:40:23.533
!MESSAGE FrameworkEvent ERROR
!STACK 0
org.osgi.framework.BundleException: Could not resolve module: org.eclipse.core.runtime [2]
  Unresolved requirement: Require-Bundle: javax.annotation; bundle-version="1.1.0"; visibility:="reexport"; resolution:="optional"
  Unresolved requirement: Require-Bundle: javax.inject; bundle-version="1.0.0"; visibility:="reexport"; resolution:="optional"
  Unresolved requirement: Require-Bundle: org.eclipse.equinox.common; bundle-version="[3.6.100,4.0.0)"; visibility:="reexport"
```
Anybody ran into this too and solved it?

BR

sjka:
I just had a look at what's missing in order to let the designer start again without bundle resolution errors. This is what it took...

While some obvious bundles were missing, it certainly feels odd to add "half the internet". I can't believe that much had been missing. So if someone more designer-savvy could have a quick look this would be great.

Of course, as we are not separating api and runtime clearly, we draw lots of stuff in that would not be needed just to reference stuff within the DSLs, but that's another story and should not be addressed here. I'd rather like to focus on merely getting it running again as it is.

fixes #2813 

### DIFF
--- a/features/org.eclipse.smarthome.feature.dependencies.designer/feature.xml
+++ b/features/org.eclipse.smarthome.feature.dependencies.designer/feature.xml
@@ -490,4 +490,164 @@
          version="0.0.0"
          unpack="false"/>
 
+   <plugin
+         id="com.google.gson"
+         download-size="0"
+         install-size="0"
+         version="0.0.0"
+         unpack="false"/>
+
+   <plugin
+         id="org.eclipse.xtend.lib"
+         download-size="0"
+         install-size="0"
+         version="0.0.0"
+         unpack="false"/>
+
+   <plugin
+         id="org.eclipse.jetty.client"
+         download-size="0"
+         install-size="0"
+         version="0.0.0"
+         unpack="false"/>
+
+   <plugin
+         id="org.eclipse.emf.mwe.utils"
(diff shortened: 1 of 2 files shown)


kaikreuzer:
Awesome, this is really working! I failed finding all dependencies the last time that I had tried... Many thanks!

PR #2824, opened 2017-01-13
Fixes #2823 remove redundant NOT_INITILIZED evens and change event or…

### LINKED ISSUE #2823: Fix events and event order fired by the automation module
When a rule is added the first NOT_INITIALIZED event is redundant.
For example:
osgi> 14:28:16.652 INFO  s.event.RuleStatusInfoEvent[:43] - rule_3 updated: NOT_INITIALIZED
14:28:16.657 INFO  s.event.RuleStatusInfoEvent[:43] - rule_3 updated: IDLE
14:28:16.658 DEBUG o.e.s.c.c.r.AbstractManagedProvider[:62] - Added new element rule_3 to ManagedRuleProvider.
14:28:16.658 INFO  smarthome.event.RuleAddedEvent[:43] - Rule 'rule_3' has been added.

or when the rule can't be resolved, the following events are fired:
osgi> 14:28:16.652 INFO  s.event.RuleStatusInfoEvent[:43] - rule_3 updated: NOT_INITIALIZED
14:28:16.657 INFO  s.event.RuleStatusInfoEvent[:43] - rule_3 updated: NOT_INITIALIZED
14:28:16.658 DEBUG o.e.s.c.c.r.AbstractManagedProvider[:62] - Added new element rule_3 to ManagedRuleProvider.
14:28:16.658 INFO  smarthome.event.RuleAddedEvent[:43] - Rule 'rule_3' has been added.

1) The first one NOT_INITIALIZED must be skiped
2) Order of events should be:
- Rule 'rule_3' has been added.
- rule_3 updated: NOT_INITIALIZED
- rule_3 updated: IDLE

danchom:
…der.

### DIFF
--- a/bundles/automation/org.eclipse.smarthome.automation.core/src/main/java/org/eclipse/smarthome/automation/core/internal/RuleRegistryImpl.java
+++ b/bundles/automation/org.eclipse.smarthome.automation.core/src/main/java/org/eclipse/smarthome/automation/core/internal/RuleRegistryImpl.java
@@ -504,8 +504,8 @@ public class RuleRegistryImpl extends AbstractRegistry<Rule, String, RuleProvide
             String rUID = ruleEngine.getUniqueId();
             ruleWithUID = initRuleId(rUID, element);
         }
-        super.added(provider, ruleWithUID);
         postRuleAddedEvent(ruleWithUID);
+        super.added(provider, ruleWithUID);
         updateRuleByTemplate(provider, ruleWithUID);
     }
 
@@ -527,8 +527,8 @@ public class RuleRegistryImpl extends AbstractRegistry<Rule, String, RuleProvide
             ruleWithUID = initRuleId(rUID, element);
         }
         Rule resolvedRule = resolveRuleByTemplate(ruleWithUID);
-        super.updated(provider, oldElement, resolvedRule);
         postRuleUpdatedEvent(resolvedRule, oldElement);
+        super.updated(provider, oldElement, resolvedRule);
     }
 
     @Override
--- a/bundles/automation/org.eclipse.smarthome.automation.core/src/main/java/org/eclipse/smarthome/automation/core/internal/RuleRegistryImpl.java
+++ b/bundles/automation/org.eclipse.smarthome.automation.core/src/main/java/org/eclipse/smarthome/automation/core/internal/RuleRegistryImpl.java
@@ -65,11 +65,11 @@ import org.slf4j.LoggerFactory;
  * <ul>
  * <li>To check a Rule's status info, use the {@link #getStatusInfo(String)} method.</li>
  * <li>The status of a newly added Rule, or a Rule enabled with {@link #setEnabled(String, boolean)}, or an updated
- * Rule, is first set to {@link RuleStatus#NOT_INITIALIZED}.</li>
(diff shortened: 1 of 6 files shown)


bundles/automation/org.eclipse.smarthome.automation.core/src/main/java/org/eclipse/smarthome/automation/core/internal/RuleRegistryImpl.java
```diff
@@ -504,8 +504,8 @@ public void added(Provider<Rule> provider, Rule element) {
             String rUID = ruleEngine.getUniqueId();
             ruleWithUID = initRuleId(rUID, element);
         }
-        super.added(provider, ruleWithUID);
```
kaikreuzer: Is it safe to change this order? 
You are now sending the event, before the rule is available in the registry - why is this necessary?

danchom: It is not available in registry, but it is already stored in RuleManagedProvider. IMHO added/updated events should be before the state events of rule. Otherwise we can  receive a state events for rules which are not available in the registry (when the add method is failed)

kaikreuzer: Looking at the code, I would actually suggest to move the "post event" line to `notifyListenersAboutXXXXElement` - this is where events are sent in ItemRegistryImpl and ThingRegistryImpl, so doing the very same here would imho make sense, wdyt?

danchom: It will not be enough, because the status events are fired when the rule is added into RuleEngine, which is done in the onAddElement() method at the moment. The logic should be moved from onXXXElement() to  notifyListenersAboutXXXXElement(). I'll try to fix this.
BTW wdyt about removing redundant NOT_INITIALIZED event.
Up to now we have:
added -> not_initillized -> idle
the proposal is:
added -> idle 
or 
added -> not_initilized.

kaikreuzer: > BTW wdyt about removing redundant NOT_INITIALIZED event.

You are referring to [this suggetion](https://github.com/eclipse/smarthome/issues/2823), right? If so, please see my comment on that issue.

danchom:
Implemented:
added -> UNINITIALIZED(without event) -> INITIALIZING -> IDLE
added -> UNINITIALIZED(without event) -> INITIALIZING -> UNINITIALIZED (with details)
updated-> UNINITIALIZED(without event) -> INITIALIZING -> IDLE
updated ->UNINITIALIZED(without event) -> INITIALIZING -> UNINITIALIZED (with details)
enabled -> DISABLED ->UNINITIALIZED(without event) -> INITIALIZING -> IDLE
enabled -> DISABLED ->UNINITIALIZED(without event) -> INITIALIZING ->  UNINITIALIZED (with details)

danchom:
Could you please review this PR.

PR #2835, opened 2017-01-16
Bring chart servlet up to openHAB 1.8.3 look and functionality

### LINKED ISSUE #2829: Chart servlet legend opaque
It appears that the chart servlet in ESH is based on older code than what evolved in openHAB 1, particularly in the area of styling.  [This commit](https://github.com/openhab/openhab/commit/2bcc99b8420bb276937b6f8d1f6799e33d47615c#diff-08acc2b1a27385f3e3a783bf0a2ce50dL148) shows the opaque `Color.LIGHT_GRAY` was changed to use the alpha channel for transparency of the chart legend, which @cdjackson I think rightly thought looked nicer.  However, the ESH chart servlet [still uses the opaque grey](https://github.com/eclipse/smarthome/blob/master/bundles/ui/org.eclipse.smarthome.ui/src/main/java/org/eclipse/smarthome/ui/internal/chart/DefaultChartProvider.java#L142), so much of the chart data is needlessly obscured.

Is there an argument against bringing the ESH chart servlet up to the same look and capability that evolved in openHAB 1?  If not, I am happy to submit a PR.

watou:
Fixes #2829

Tested on latest OH2 unstable build on RPi2B.

newchart

XChart had to be upgraded beyond 2.2.1.  2.6.1 might need a Contribution questionnaire (CQ) to be submitted.  3.x was not used due to increased size without benefit and refactoring.

Signed-off-by: John Cocula <john@cocula.com>


kaikreuzer:
Many thanks, @watou!
I have created CQ 12530 for version 2.6.1 of XCharts.
Btw, did you also have a look at what XCharts 3.x might bring? This seems to be the latest release.

watou:
> what XCharts 3.x might bring? This seems to be the latest release.

3.x brings several new kinds of charts ([examples here](http://knowm.org/open-source/xchart/xchart-example-code/)), most of which don't make sense for the time-series plotting needed for the sitemap Chart widget and item persistence.  The API was heavily refactored to accommodate the new chart types.  However, whatever creative uses are emerging under HABPanel's charting could be addressed by XChart 3.x, but this would require a more expressive way of describing what kind of chart is needed than what is currently possible with the Chart sitemap widget.  Maybe a `type=` parameter to the Chart widget could be explored in a future PR, but this of course requires a careful review of objectives.

Issue #2820, opened 2017-01-13
limitToOptions is ignored by Paper UI


kaikreuzer:
I have a config parameter defined with provided options and limitToOptions=false:
```
      {  
         "name":"command",
         "type":"TEXT",
         "label":"Command",
         "description":"the command to be sent",
         "required":true,
         "limitToOptions":false,
         "options":[
            {
            "label": "ON",
            "value": "ON"
            },
            {
            "label": "OFF",
            "value": "OFF"
            },
            {
            "label": "OPEN",
            "value": "OPEN"
            },
            {
            "label": "CLOSED",
            "value": "CLOSED"
            },
            {
            "label": "UP",
            "value": "UP"
            },
            {
            "label": "DOWN",
            "value": "DOWN"
            }
          ]
      }
```
Nonetheless, I cannot enter any arbitrary value - I can only choose one of the given options:
<img width="288" alt="screen shot 2017-01-13 at 09 37 26" src="https://cloud.githubusercontent.com/assets/3244965/21923367/edc4bf1c-d973-11e6-8213-46c7d121a9be.png">

aounhaider1:
Thats because its not supported here. It is only supported in multiple widget.

kaikreuzer:
What do you mean? What is the difference of a "multiple widget" to a parameter with limitToOptions=false?

aounhaider1:
For multiple=false we are using default material md-select. It does not provide the option to add new values. On the other hand, the widget that we are using for multiple=true is our own custom widget, that supports adding new values and so on.

kaikreuzer:
So the bug is that the wrong widget type is used here? So I guess you can fix this.

kaikreuzer:
Sorry, there are still issues with that widget after #2841.

1. Set values (which are not in the option list) are not shown as the current value when opening a page:
![screen shot 2017-01-18 at 09 51 51](https://cloud.githubusercontent.com/assets/3244965/22057171/d15836e4-dd63-11e6-93ee-2ecb5e1a5074.png)

On this picture, the parameter is set to "1", but it is not shown, unless you open the drop-down.

2. The field value is grey, if it is an option from the list and black, if it is a custom entry. Imho, on an limitToOptions=false field, the value should always be a String that the user can type and edit and thus it should always be shown in black.

kaikreuzer:
Sorry, I am still not happy with this. It feels very unintuitive for the user when entering a custom value.

I would expect, he can enter a value and click "OK" to save it:
[...]

PR #2841, opened 2017-01-17
Support adding new options in non multiple dropdown

aounhaider1:
Before:
screen shot 2017-01-17 at 11 49 02

After:
screen shot 2017-01-17 at 10 48 51

The new option can be added just by typing and pressing enter or clicking on the green plus icon.
It still uses normal md-select for limitToOptions=true.
The items in the dropdown can now be selected by clicking on the row. (works in multi widget)


kaikreuzer:
I assume this fixes https://github.com/eclipse/smarthome/issues/2820.

kaikreuzer:
> I assume this fixes #2820.

Well, it does, but it shows now a bug in the opposite case. Not specifying limitToOptions should use "true" as a default, but I am still able to type in any value:

screen shot 2017-01-17 at 13 42 08

See the "item condition" of the rule engine.

When you adapt this, please note: As this parameter is NOT required, it should still be possible to set it to null (i.e. allow an empty selection).

kaikreuzer:
> Not specifying limitToOptions should use "true" as a default

This is what the documentation says. But then again, I found [this line in the code](https://github.com/eclipse/smarthome/blob/master/bundles/config/org.eclipse.smarthome.config.core/src/main/java/org/eclipse/smarthome/config/core/ConfigDescriptionParameter.java#L91), which contradicts. Changing the code to match the documentation will be a breaking change - I have created https://github.com/eclipse/smarthome/issues/2844 for discussion.

@aounhaider1 Please simply add limitToOptions=false for the above rule module parameter in order to fix the concrete issue for the moment.

kaikreuzer:
> As this parameter is NOT required, it should still be possible to set it to null (i.e. allow an empty selection).

Is this addressed? Doesn't see to work for me :-(

PR #2842, opened 2017-01-17
corrected serialisation of "default" property in config descriptions

### LINKED ISSUE #2819: Default value ignored by Paper UI
I have a config parameter defined as:
```
      {  
         "name":"operator",
         "type":"TEXT",
         "label":"Operator",
         "description":"the compare operator (one of =,<,>,!=)",
         "required":false,
         "options":[
            {
            "label": "=",
            "value": "="
            },
            {
            "label": "!=",
            "value": "!="
            },
            {
            "label": ">",
            "value": ">"
            },
            {
            "label": "<",
            "value": "<"
            }
          ],
          "default":"="
      }
```
I would expect the Paper UI to show "=" as a selection, but the field is left empty:

screen shot 2017-01-13 at 09 34 38

kaikreuzer:
fixes #2819

**NOTE**: Not yet meant to be merged - the Paper UI needs to be adapted at the same time.
@aounhaider1 Could you push further commits to this branch for that?

### DIFF
--- a/bundles/automation/org.eclipse.smarthome.automation.api.test/.classpath
+++ b/bundles/automation/org.eclipse.smarthome.automation.api.test/.classpath
@@ -1,0 +1,7 @@
+<?xml version="1.0" encoding="UTF-8"?>
+<classpath>
+	<classpathentry kind="con" path="org.eclipse.jdt.launching.JRE_CONTAINER/org.eclipse.jdt.internal.debug.ui.launcher.StandardVMType/JavaSE-1.8"/>
+	<classpathentry kind="con" path="org.eclipse.pde.core.requiredPlugins"/>
+	<classpathentry kind="src" path="src/main/java"/>
+	<classpathentry kind="output" path="target/classes"/>
+</classpath>
--- a/bundles/automation/org.eclipse.smarthome.automation.api.test/.project
+++ b/bundles/automation/org.eclipse.smarthome.automation.api.test/.project
@@ -1,0 +1,33 @@
+<?xml version="1.0" encoding="UTF-8"?>
+<projectDescription>
+	<name>org.eclipse.smarthome.automation.api.test</name>
+	<comment></comment>
+	<projects>
+	</projects>
+	<buildSpec>
+		<buildCommand>
+			<name>org.eclipse.jdt.core.javabuilder</name>
+			<arguments>
+			</arguments>
+		</buildCommand>
+		<buildCommand>
+			<name>org.eclipse.pde.ManifestBuilder</name>
+			<arguments>
+			</arguments>
+		</buildCommand>
+		<buildCommand>
+			<name>org.eclipse.pde.SchemaBuilder</name>
+			<arguments>
+			</arguments>
+		</buildCommand>
+		<buildCommand>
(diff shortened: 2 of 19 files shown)


aounhaider1:
Doesn't seem to work. The default field is now missing from the rest response.

kaikreuzer:
Could you debug why?

bundles/config/org.eclipse.smarthome.config.core/src/main/java/org/eclipse/smarthome/config/core/ConfigDescriptionParameter.java
```diff
@@ -81,6 +83,8 @@
     private String unitLabel;
 
     private String context;
+
+    @SerializedName("default")
```
kaikreuzer: Do you have any explanation why this is required? Are we some where serializing ConfigDescriptionParameter instead of ConfigDescriptionParameterDTO?

aounhaider1: Here while parsing from JSON:
https://github.com/eclipse/smarthome/blob/master/bundles/automation/org.eclipse.smarthome.automation.parser.gson/src/main/java/org/eclipse/smarthome/automation/parser/gson/internal/ModuleTypeGSONParser.java#L46

kaikreuzer: Interesting. I am actually not sure whether this should rather be considered a bug. The deserialization should imho produce DTOs, which we should then map the the internal classes.
@danchom, @maggu2810, what do you think?

maggu2810: IMHO we should use DTOs for (de)serialization over the REST interface.
All the time.
Without DTOs we could export internal stuff.

htreu: I second @maggu2810 suggestion. On the other hand we write ConfigurationParameter data to the JSONStorage which is done via the model classes. Should all import/export of JSON be done using DTOs? If we want to keep the JSONStorage in sync with the REST resources this would make sense. This would also mean to map to/from DTOs on all external borders and only use the model class internally.

@kaikreuzer @maggu2810 @sjka wdyt?

htreu:
I authored a new commit where both ModuleType and Rule are serialised for the REST API using DTOs, especially for the ConfigDescriptionParameter. In addition the ManagedRuleProvider now persists rules using the RuleDTO too.
There are other places left which do not yet serialise through DTOs. Especially the RuleTemplate holds another list of ConfigDescriptionParameter and is imported/exported by several implementations of RuleTemplateProvider.

Please comment your thoughts of how we should proceed here. The initial issue is fixed by this PR (besides the PaperUI having race conditions, addressed by another PR).

kaikreuzer:
Superseded by #3612.

PR #2850, opened 2017-01-18
make outputs available to scripts and provide access to the full context

### LINKED ISSUE #2849: Module outputs are not available in scripts
The values of outputs of other modules are made available to script, which is also directly supported by the rule editor:

screen shot 2017-01-18 at 10 15 23

When trying to access those outputs, an exception is thrown.

The problem is that in the rule context, they are made available as e.g. "1.newState", which is provided as a map 1->newState in within the rule execution context. It is probably rather a bad idea to define "1" as a variable name...

kaikreuzer:
outputs are now available with their unprefixed id. As this can be a problem if there are multiple modules with the same output ids, the full rule context is additionally made available. This makes the following options available:
```
print(newState)
print(context.get('1.newState'))
```

fixes #2849

### DIFF
--- a/bundles/automation/org.eclipse.smarthome.automation.module.script/ESH-INF/automation/moduletypes/ScriptTypes.json
+++ b/bundles/automation/org.eclipse.smarthome.automation.module.script/ESH-INF/automation/moduletypes/ScriptTypes.json
@@ -43,7 +43,8 @@
                         "label": "Javascript",
                         "value": "application/javascript"
                     }
-                ]
+                ],
+                "defaultValue":"application/javascript"
             },
             {  
                "name":"script",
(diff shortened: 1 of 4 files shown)


bundles/automation/org.eclipse.smarthome.automation.module.script/ESH-INF/automation/moduletypes/ScriptTypes.json
```diff
@@ -43,7 +43,8 @@
                         "label": "Javascript",
                         "value": "application/javascript"
                     }
-                ]
+                ],
+                "defaultValue":"application/javascript"
```
maggu2810: If `defaultValue` correct at this place?

kaikreuzer: Yes, at least as long as [this](https://github.com/eclipse/smarthome/issues/2819#issuecomment-272825332) isn't solved yet. "Value" will have to be removed with https://github.com/eclipse/smarthome/issues/2842 again.

sjka:
Funny thing... I finally wanted to set up this new rule engine at home and struggled hard getting a simple use case running. After two hours of debugging and learning far more about its internals than  intended (at least for tonight) I happened to find this PR. 

+1!!!!

danchom:
It is ok for me to be merged . I've comment it here #2849

PR #2853, opened 2017-01-18
Parse multiple default values

aounhaider1:


kaikreuzer:
Defaults should be provided as an array, so this PR won't be needed.

PR #2858, opened 2017-01-18
introduced new timeOfDay trigger and dayOfWeek condition

kaikreuzer:


bundles/automation/org.eclipse.smarthome.automation.module.timer/src/main/java/org/eclipse/smarthome/automation/module/timer/handler/TimeOfDayTriggerHandler.java
```diff
@@ -0,0 +1,80 @@
...
+import org.eclipse.smarthome.core.scheduler.Expression;
+import org.eclipse.smarthome.core.scheduler.ExpressionThreadPoolManager;
+import org.eclipse.smarthome.core.scheduler.ExpressionThreadPoolManager.ExpressionThreadPoolExecutor;
+import org.eclipse.smarthome.core.scheduler.RecurrenceExpression;
+import org.slf4j.Logger;
+import org.slf4j.LoggerFactory;
+
+/**
+ * This is an ModuleHandler implementation for Triggers which trigger the rule
+ * based on a cron expression. The cron expression can be set with the
```
maggu2810: Is the configuration HH:MM really a cron expression? I don't think so.

bundles/automation/org.eclipse.smarthome.automation.module.timer/src/main/java/org/eclipse/smarthome/automation/module/timer/handler/DayOfWeekConditionHandler.java
```diff
@@ -0,0 +1,76 @@
...
+                        break;
+                    case "FRI":
+                        days.add(6);
+                        break;
+                    case "SAT":
+                        days.add(7);
+                        break;
+                }
+            }
+            days.toArray(new Integer[days.size()]);
```
maggu2810: Does it make sense without assign the return value to anything?

bundles/automation/org.eclipse.smarthome.automation.module.timer/src/main/java/org/eclipse/smarthome/automation/module/timer/handler/DayOfWeekConditionHandler.java
```diff
@@ -0,0 +1,76 @@
...
+                        break;
+                    case "THU":
+                        days.add(5);
+                        break;
+                    case "FRI":
+                        days.add(6);
+                        break;
+                    case "SAT":
+                        days.add(7);
+                        break;
```
maggu2810: I would prefer to add a default case that logs a message that the string could not be interpret.

bundles/automation/org.eclipse.smarthome.automation.module.timer/src/main/java/org/eclipse/smarthome/automation/module/timer/handler/DayOfWeekConditionHandler.java
```diff
@@ -0,0 +1,76 @@
...
+
+    private static final String CFG_DAYS = "days";
+
+    private final ArrayList<Integer> days;
+
+    @SuppressWarnings("unchecked")
+    public DayOfWeekConditionHandler(Condition module) {
+        super(module);
+        try {
+            days = new ArrayList<>();
```
maggu2810: Shouldn't we use a set while parsing to grant that MON,MON will add the MON (the "1") only once?

bundles/automation/org.eclipse.smarthome.automation.module.timer/src/main/java/org/eclipse/smarthome/automation/module/timer/handler/TimeOfDayTriggerHandler.java
```diff
@@ -0,0 +1,80 @@
...
+    public static final String MODULE_CONTEXT_NAME = "MODULE";
+
+    private static final String CFG_TIME = "time";
+
+    private final ExpressionThreadPoolExecutor scheduler;
+    private final Expression expression;
+
+    public TimeOfDayTriggerHandler(Trigger module) {
+        super(module);
+        String time = (String) module.getConfiguration().get(CFG_TIME);
```
maggu2810: Above you suppress warning about unchecked casts.
Isn't this necessary here, too?
Okay, it is only a warning, so it is not really necessary, but if used above, why not using here?

bundles/automation/org.eclipse.smarthome.automation.module.timer/src/main/java/org/eclipse/smarthome/automation/module/timer/handler/TimeOfDayTriggerHandler.java
```diff
@@ -0,0 +1,80 @@
...
+        try {
+            String[] parts = time.split(":");
+
+            expression = new RecurrenceExpression("FREQ=DAILY;BYHOUR=" + parts[0] + ";BYMINUTE=" + parts[1]);
+        } catch (Exception e) {
+            throw new IllegalArgumentException("'time' parameter is not in valid format ' hh:mm'.", e);
```
maggu2810: Why is there a space between `'` and `hh`?

bundles/automation/org.eclipse.smarthome.automation.module.timer/src/main/java/org/eclipse/smarthome/automation/module/timer/handler/TimeOfDayTriggerHandler.java
```diff
@@ -0,0 +1,80 @@
...
+    private final Expression expression;
+
+    public TimeOfDayTriggerHandler(Trigger module) {
+        super(module);
+        String time = (String) module.getConfiguration().get(CFG_TIME);
+        try {
+            String[] parts = time.split(":");
+
+            expression = new RecurrenceExpression("FREQ=DAILY;BYHOUR=" + parts[0] + ";BYMINUTE=" + parts[1]);
+        } catch (Exception e) {
```
maggu2810: Should we really catch every Exception (perhaps only runtime once)?

bundles/automation/org.eclipse.smarthome.automation.module.timer/src/main/java/org/eclipse/smarthome/automation/module/timer/handler/TimeOfDayTriggerHandler.java
kaikreuzer: Obviously not.

bundles/automation/org.eclipse.smarthome.automation.module.timer/src/main/java/org/eclipse/smarthome/automation/module/timer/handler/TimeOfDayTriggerHandler.java
kaikreuzer: Eclipse doesn't show a warning here, don't know why. I'll change it to
```
String time = module.getConfiguration().get(CFG_TIME).toString();
```

kaikreuzer:
thanks for the timely review - all comments should be addressed now.

bundles/automation/org.eclipse.smarthome.automation.module.timer/src/main/java/org/eclipse/smarthome/automation/module/timer/handler/DayOfWeekConditionHandler.java
```diff
@@ -0,0 +1,80 @@
...
+                        days.add(5);
+                        break;
+                    case "FRI":
+                        days.add(6);
+                        break;
+                    case "SAT":
+                        days.add(7);
+                        break;
+                    default:
+                        logger.warn("Ignoring illegal weekday '{}'", day);
```
maggu2810: A `break` would be nice (I assume quality tools e.g. the work @svilenvul is doing) will complain about that).

bundles/automation/org.eclipse.smarthome.automation.module.timer/src/main/java/org/eclipse/smarthome/automation/module/timer/handler/DayOfWeekConditionHandler.java
```diff
@@ -0,0 +1,80 @@
...
+                }
+            }
+        } catch (Exception e) {
```
maggu2810: Same comment here as for the previous one, as every exception is a little bit much to catch. Also I don't know if it is a good practice to catch every (without filter) non-runtime exception into a runtime one.

kaikreuzer:
updated.

bundles/automation/org.eclipse.smarthome.automation.module.timer/src/main/java/org/eclipse/smarthome/automation/module/timer/handler/DayOfWeekConditionHandler.java
```diff
@@ -0,0 +1,81 @@
...
+
+    @SuppressWarnings("unchecked")
+    public DayOfWeekConditionHandler(Condition module) {
+        super(module);
+        try {
+            days = new HashSet<>();
+            for (String day : (Iterable<String>) module.getConfiguration().get(CFG_DAYS)) {
+                switch (day) {
+                    case "SUN":
+                        days.add(1);
```
sjka: As you are relying on `Calendar.getInstance().get(Calendar.DAY_OF_WEEK)` further down, wouldn't it make sense to also use e.g. `Calendar.SUNDAY` here instead of the int literals?

bundles/automation/org.eclipse.smarthome.automation.module.timer/src/main/java/org/eclipse/smarthome/automation/module/timer/handler/DayOfWeekConditionHandler.java
```diff
@@ -0,0 +1,81 @@
...
+    private final Set<Integer> days;
+
+    @SuppressWarnings("unchecked")
+    public DayOfWeekConditionHandler(Condition module) {
+        super(module);
+        try {
+            days = new HashSet<>();
+            for (String day : (Iterable<String>) module.getConfiguration().get(CFG_DAYS)) {
+                switch (day) {
+                    case "SUN":
```
sjka: is it case-sensitive by intention?

danchom:
LGTM

maggu2810:
As it looks good to @danchom (and all checks has been passed now) I merged it.

PR #2860, opened 2017-01-18
Tests for org.eclipse.smarthome.core.voice

m-memova:
Since they are groovy files a new test directory is created
and the old one is renamed.


svilenvul:
@kaikreuzer, @sjka 

Could you take a look at these tests?  

What has to be done in order to get it into mergeable state?

sjka:
Yes, sorry for the huge delay! 

> What has to be done in order to get it into mergeable state?

Without going into details yet, thankfully the implementation looks pretty clean. Nevertheless, there are y couple of things that need to be done:

* A rebase to the lastest master
* As with Java 8 we decided to not add any new groovy tests anymore, they have to be ported to Java first. Luckily they don't use any of the fancy groovy features anyway, so I hope this would be pretty much straight-forward. 
* Dropping of the assertion texts, so instead of writing
  ```java
  assertThat("The stream was not successfully processed", sink.isStreamProcessed(), is(true));
  ```
  the should be like
  ```
  assertThat(sink.isStreamProcessed(), is(true));
  ```
  remember that your are not obliged to using the matchers, i.e. in simple cases like this you may also directly use
  ```
  assertTrue(sink.isStreamProcessed());
  ```
* If that's done, we will have to create a CQ to get merge approval as the PR exceeds 1000 lines of code

So again my apologies for the huge delay and the extra-work it imposes on your side. But I think the move to Java still is for the better especially as we will have proper IDE support in Java.

VelinYordanov:
I migrated the voice tests to java and addressed @sjka 's comments.

@sjka : Unfortunately the build has failed. Can you please take a look?

sjka:
Kudos!
Sure - it's not related to your change, it seems download.eclipse.org still is a bit shaky after its outage. I restarted the build. Let's see...

sjka:
No it looks like Travis found something:

```
interpretSomeText(org.eclipse.smarthome.core.voice.voiceconsolecommandextension.InterpretCommandTest)  Time elapsed: 0.019 sec  <<< FAILURE!
java.lang.AssertionError: 
Expected: is "Interpreted text"
     but: was "Sorry, I didn't get that."
	at org.junit.Assert.assertThat(Assert.java:832)
	at org.eclipse.smarthome.core.voice.voiceconsolecommandextension.InterpretCommandTest.interpretSomeText(InterpretCommandTest.java:70)
```

VelinYordanov:
@sjka The problem should be fixed now. In some rare cases the Language Interpreters were collected in different order, which broke tests. Now  the stub interpreter is configured to be the default one, which will fix the tests.

VelinYordanov:
@sjka I have adapted the tests. Can you please take a look?

bundles/core/org.eclipse.smarthome.core.voice.test/src/test/java/org/eclipse/smarthome/core/voice/internal/AudioManagerStub.java
```diff
@@ -0,0 +1,97 @@
+package org.eclipse.smarthome.core.voice.internal;
```
sjka: Please add a license header here

bundles/core/org.eclipse.smarthome.core.voice.test/src/test/java/org/eclipse/smarthome/core/voice/javavoicemanager/VoiceManagerTest.java
```diff
@@ -0,0 +1,386 @@
...
+        config.put(CONFIG_DEFAULT_HLI, hliStub.getId());
+        config.put(CONFIG_DEFAULT_VOICE, voice.getUID());
+
+        ConfigurationAdmin configAdmin = super.getService(ConfigurationAdmin.class);
+        String pid = "org.eclipse.smarthome.voice";
+        Configuration configuration = configAdmin.getConfiguration(pid);
+        configuration.update(config);
+
+        // To allow the config admin to update the configuration.
+        Thread.sleep(100);
```
sjka: Please put the below two lines into a `waitForAssert(() -> {})`-Block instead. The fixed sleep is guaranteed to fail one day on some CI server...

bundles/core/org.eclipse.smarthome.core.voice.test/src/test/java/org/eclipse/smarthome/core/voice/voiceconsolecommandextension/InterpretCommandTest.java
```diff
@@ -0,0 +1,90 @@
...
+        config.put(CONFIG_DEFAULT_TTS, ttsService.getId());
+        config.put(CONFIG_DEFAULT_HLI, hliStub.getId());
+        config.put(CONFIG_DEFAULT_VOICE, voice.getUID());
+        ConfigurationAdmin configAdmin = super.getService(ConfigurationAdmin.class);
+        String pid = "org.eclipse.smarthome.voice";
+        Configuration configuration = configAdmin.getConfiguration(pid);
+        configuration.update(config);
+
+        // To allow the config admin to update the configuration.
+        Thread.sleep(100);
```
sjka: see above - try to find a way to detect if the configuration was applied successfully and put that into a `waitForAssert`-Block

bundles/core/org.eclipse.smarthome.core.voice.test/src/test/java/org/eclipse/smarthome/core/voice/voiceconsolecommandextension/VoiceConsoleCommandExtensionTest.java
```diff
@@ -0,0 +1,57 @@
...
+/**
+ * A base class for the classes testing the different commands of the {@link VoiceConsoleCommandExtension}.
+ * It takes care of the mocks, stubs and services that are used in all the extending classes.
+ *
+ * @author Mihaela Memova - initial contribution
+ *
+ * @author Velin Yordanov - migrated tests from groovy to java
+ *
+ */
+public class VoiceConsoleCommandExtensionTest extends JavaOSGiTest {
```
sjka: As it's a base class, make it abstract

sjka:
Thanks!
I'll create a CQ for this PR - therefore please don't change anything anymore 😄

sjka:
FTR: CQ 14038

kaikreuzer:
CQ has been approved!

kaikreuzer:
@VelinYordanov Our build now fails with
```
[ERROR] Code Analysis Tool has found: 
1 error(s)! 
2 warning(s) 
1 info(s)
[ERROR] org.eclipse.smarthome.core.voice.voiceconsolecommandextension.InterpretCommandTest.java:[90]
Call to org.eclipse.smarthome.core.voice.voiceconsolecommandextension.InterpretCommandTest.equals(String) in org.eclipse.smarthome.core.voice.voiceconsolecommandextension.InterpretCommandTest.lambda$2()
[WARNING] /jobs/genie.smarthome/SmartHomeDistribution-Nightly/workspace/bundles/core/org.eclipse.smarthome.core.voice.test/build.properties:[1]
Property  source.. in the build.properties file is missing value: src/test/resources
[WARNING] /jobs/genie.smarthome/SmartHomeDistribution-Nightly/workspace/bundles/core/org.eclipse.smarthome.core.voice.test/pom.xml:[29]
Avoid overriding a configuration inherited by the parent pom.
```
Could you create a PR to fix this issue? Thanks!

kaikreuzer:
@VelinYordanov Has already been fixed by #4241.

PR #2861, opened 2017-01-18
Multiselect show label

aounhaider1:
fixes issue reported here https://github.com/eclipse/smarthome/issues/2820#issuecomment-273418735


kaikreuzer:
Thanks, the two issues are fixed, but the overall widget is imho still not satisfying, see https://github.com/eclipse/smarthome/issues/2820#issuecomment-273418735.

PR #2864, opened 2017-01-19
fixed inconsistent caching behavior in rule module handler factories

sjka:
...as I noticed some differences in the implemenations of internalCreate().
* some factories  where first trying to access the cache, some didn't
* some factories were writing to the cache, some didn't

Therefore I moved the cache access completely to the abstract base class,
allowing derived classes only to access it for read-only purposes.
By doing that, the cache is managed completely within the abstract
base class and this concern does not leak.


sjka:
rebased

danchom:
LGTM

PR #2865, opened 2017-01-19
add a service to allow reuse of localized thing type creation

maggu2810:
With this change the code that localize a thing type using the i18n stuff is moved from the XML implementation to a separate service so it could be reused by other code, too.


maggu2810:
all tests passed, so the wemo test is IMHO a little bit unstable or depends on some other stuff

PR #2866, opened 2017-01-19
Multiselect use input value if save clicked

aounhaider1:
Related comment:
https://github.com/eclipse/smarthome/issues/2820#issuecomment-273521507

Known issue: Multiselect does not show always on top of other content.


kaikreuzer:
Looks much better, thanks.

Two questions though:
1. For multiple=false, does the "+" sign now make any sense at all or shouldn't we better suppress it?
2. If I have a value selected and want to clear it, I try to select the empty line, but this is not possible. It only works if I press the selection box on the right side.

aounhaider1:
Updated accordingly.

PR #2867, opened 2017-01-19
added tests for time based module handlers

kaikreuzer:


danchom:
Missing test for triggering of time of date trigger
LGTM

kaikreuzer:
> Missing test for triggering of time of date trigger

I couldn't think of a way to do this without potentially causing the test to run about a minute.
If you have ideas, feel free to add such a test!

danchom:
Yes I agree with you that there is no other way. If the seconds was an optional configuration parameter to the TimeOfDayTrigger, i mean to accept time of day in form HH:mm or HH:mm:ss, then the test can stop for a few seconds. Anyway this must not be done only for the test, but I think it will useful to have ability to set timers with precise of seconds.

danchom:
LGTM

PR #2872, opened 2017-01-20
validate models before loading them

### LINKED ISSUE #2870: DSL file parsing errors should log a warning
Currently, DSL files that have a broken syntax that causes the parser to fail are half-parsed or silently ignored, see e.g. #2863 and #2580.

It would be helpful, if Xtext validation could be performed. Resources should be completely ignored, if they have syntax errors and a warning should be logged in those cases.

sjka:
with this change,

* there are two "layers" of validation
  1. errors when loading the resource. Usually these are syntax
     violations which irritate the parser.
     They will be logged as a WARNING and the DSL file will be
     ignored completely, i.e. all its content will not be loaded
     or removed if it was an update
  2. all kinds of other errors (i.e. violations of validation checks)
     will only be logged, but the DSL file will be loaded and
     processed nevertheless.

* validation will be done on a separate resource, in order to
  keep the original one intact in case its content needs to
  be removed because of syntactical errors.

* empty [] for channel properties are allowed now in the Thing DSL

fixes #2870


bundles/model/org.eclipse.smarthome.model.core/src/main/java/org/eclipse/smarthome/model/core/internal/ModelRepositoryImpl.java
```diff
@@ -202,6 +218,41 @@ private Resource getResource(String name) {
...
+
+                // Check for validation errors, but log them only
+                org.eclipse.emf.common.util.Diagnostic diagnostic = Diagnostician.INSTANCE
+                        .validate(resource.getContents().get(0));
+                for (org.eclipse.emf.common.util.Diagnostic d : diagnostic.getChildren()) {
+                    warnings.append(d.getMessage());
+                    warnings.append("\n");
+                }
+                if (warnings.length() > 0) {
+                    logger.warn("Validation error(s) in configuration model '{}':\n {}", name, warnings.toString());
```
maggu2810: Shouldn't toString() be called automatically by the logger framework? It make some difference if you call it yourself.
If you call toString yourself, it is executed all the time regardless of the used log level (so, also if the current logger is set to error).

bundles/model/org.eclipse.smarthome.model.core/src/main/java/org/eclipse/smarthome/model/core/internal/ModelRepositoryImpl.java
```diff
@@ -202,6 +218,41 @@ private Resource getResource(String name) {
         return resourceSet.getResource(URI.createURI(name), false);
     }
 
+    private String validateModel(String name, InputStream inputStream) throws IOException {
```
maggu2810: Could you add a JavaDoc that explains the return value and the thrown exception (warnings are logged, errors are returned, ...)?

bundles/model/org.eclipse.smarthome.model.core/src/main/java/org/eclipse/smarthome/model/core/internal/ModelRepositoryImpl.java
```diff
@@ -119,11 +133,13 @@ public boolean addOrRefreshModel(String name, InputStream inputStream) {
                     }
                     notifyListeners(name, EventType.MODIFIED);
                     return true;
-                } catch (IOException e) {
-                    logger.warn("Configuration model '" + name + "' cannot be parsed correctly!", e);
-                    resourceSet.getResources().remove(resource);
                 }
             }
+        } catch (IOException e) {
+            logger.warn("Configuration model '" + name + "' cannot be parsed correctly!", e);
```
maggu2810: Don't build strings in logging statements (I know it has been moved around only, but a good time to fix).

sjka:
thanks for the feedback. I have updated the PR.

Issue #997, opened 2016-02-08
Introduce a mechanism to easily register action method signatures


kaikreuzer:
The Designer of openHAB 1 had most bundles containing actions included and thus way able to resolve these actions in the rule&script editors (for auto-completion, syntax checks etc.).

It also included the possibility to add further bundles in an `addons` folder by the user, although this mechanism never worked well.

The ESH Designer only contains a few "core ESH" actions. As it is not feasible to automatically bundle action-addons with it, there should be a way to easily add such method signatures (the code itself is not required for the editors).

SuperOok:
What exactly is the problem that you cannot provide actions later by some bundle? Is it that the Xtext grammar needs to know everything in advance? Isn't this the use case of Xbase integration in the grammar? Or is it an issue with class loading? Maybe fragments or buddy policy help?
Maybe I do not get the problem right.

kaikreuzer:
The code that provides the action definitions for the scripts/rules can be found [here](https://github.com/eclipse/smarthome/blob/master/bundles/model/org.eclipse.smarthome.model.script/src/org/eclipse/smarthome/model/script/scoping/ScriptImplicitlyImportedTypes.java#L89).
As I said, in theory, putting the action bundles in the `addons` folder should make them being resolved and their services being available.
Feel free to figure out, why this isn't working as expected. A difficult thing to get right is probably that you need all dependencies of the action bundles being available as well - and as the Designer does not have all the bundles of the runtime, this can be hard. That's why I think we would need a new/better mechanism than relying on registered OSGi services here.

SuperOok:
Hmm... the Designer is an Eclipse RCP Application, so why don't we use Eclipse mechanisms? The Bundles could be explicit Eclipse Plugins and register themselves either via Eclipse extension point or OSGi Service.
For the dependencies a p2 Update Site could be provided such that you can install the addons with the standard Eclipse installation mechanism:
http://www.vogella.com/tutorials/EclipseP2Update/article.html#tutorial_p2ui

Eclipse would then resolve all the necessary dependencies if they are provided by any known update site.

kaikreuzer:
>  The Bundles could be explicit Eclipse Plugins

Well, they are explizit OSGi bundles and thus can be installed on RCP, yes. 
[...]

PR #2873, opened 2017-01-20
added missing model runtime bundles to designer

### LINKED ISSUE #2190: Item resolution broken in Designer
Items are not resolved in the Designer, i.e. when trying to use content completion e.g. in sitemaps, no proposals show up.

sjka:
...in order to fix the registry population with elements.

fixes #2190

### DIFF
--- a/features/org.eclipse.smarthome.feature.designer/feature.xml
+++ b/features/org.eclipse.smarthome.feature.designer/feature.xml
@@ -248,4 +248,46 @@
          version="0.0.0"
          unpack="false"/>
 
+   <plugin
+         id="org.eclipse.smarthome.model.item.runtime"
+         download-size="0"
+         install-size="0"
+         version="0.0.0"
+         unpack="false"/>
+
+   <plugin
+         id="org.eclipse.smarthome.model.persistence.runtime"
+         download-size="0"
+         install-size="0"
+         version="0.0.0"
+         unpack="false"/>
+
+   <plugin
+         id="org.eclipse.smarthome.model.rule.runtime"
+         download-size="0"
+         install-size="0"
+         version="0.0.0"
+         unpack="false"/>
+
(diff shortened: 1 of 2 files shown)


lolodomo:
That would be really cool to include this fix in final 2.0 ;)

kaikreuzer:
There is no openHAB 2.0 Designer!
And it won't really be any better than released version 0.8.0, will it?

lolodomo:
I am probably using release 0.8.0 but few actions like say or interpret are not recognized. I imagine they are in more recent releases ?
What would be cool is a designer version that would fully support the eclipse smarthome version used by final OH 2.

kaikreuzer:
Right. For this, we will "only" have to solve https://github.com/eclipse/smarthome/issues/997.
Note that many actions are defined in openHAB1 as add-ons, so the Designer naturally cannot know about them, which is a problem :-(

kaikreuzer:
Thanks, @sjka, I just tested this. I can indeed now see the items in the "Items" window and they are nicely resolved in persist files and sitemaps. BUT my rules still look like shown in https://github.com/eclipse/smarthome/issues/2190#issuecomment-253628627.
Do you have the same? If so, there is still some other "resolution" issue to be fixed...

lolodomo:
@kaikreuzer could you please thing releasing a new designer official version because version 0.8 is outdated and recent snapshots look ok.

kaikreuzer:
Snapshots aren't ok as mentioned in https://github.com/eclipse/smarthome/issues/2190#issuecomment-275096251.
And Designer cannot be released independently. It will be part of an ESH 0.9.0 release.

lolodomo:
Ok
At least, designer smapshot is now usable, that is already a first step.
0.8 seems to be icompatible with defining channels properties in thing.

PR #2876, opened 2017-01-20
Fixes #2442 Basic UI: valuecolor attribute not considered when the value is updated

### LINKED ISSUE #2442: Basic UI: valuecolor attribute not considered when the value is updated
I defined in my sitemap different colors for some of my temperature items.
In Basic UI, when I open a frame containing this item, the color is correctly set. With the time, remaining on the same frame, the value is updated but the color is not updated.
Example:
```
			Text item=TemperatureSalon valuecolor=[<18="red",<19.5="orange"]
```
Tested with snapshot 554.

wborn:
With this PR the Basic UI also updates the `valuecolor` and `labelcolor` of sitemap items based on SSE or long polling events.


maggu2810:
@resetnow Do you want to have a look at?

maggu2810:
Thanks @resetnow for review and thanks for the fix @wborn

Issue #595, opened 2015-12-03
Allow bindings to deactivate auto-update for channels


kaikreuzer:
migrated from Bugzilla [#481178](https://bugs.eclipse.org/bugs/show_bug.cgi?id=481178)
status NEW severity _normal_ in component _Core_ for _---_
Reported in version _unspecified_ on platform _PC_
Assigned to: Project Inbox

On 2015-10-31 19:43:07 -0400, Kai Kreuzer wrote:

> By default, the auto-update binding sets the status for a channel after seeing a command for it. The binding must be allowed to deactivate this feature for certain channels, so that it can have the state updates under its own control.

maggu2810:
> By default, the auto-update binding sets the status for a channel after seeing a command for it. The binding must be allowed to deactivate this feature for certain channels, so that it can have the state updates under its own control.

I am currently looking into this...

Please correct me if I am wrong, but the auto-update post an update for the item after receiving a command for it.
This could currently be disabled by implementations of `AutoUpdateBindingConfigProvider` by using the item name.

There could be different ways to implement that feature, so, without going into deeper details of the special implementations I assume the principle is equal.
The ItemChannelLinkRegistry needs to be used to get the channel UIDs of the linked channels.
Then we could take the binding ID and decide if an auto-update should be done (how it is done e.g. by provided services, a map that is filled by some code, ... does not matter now).

But what is the desired behaviour of:

* binding A provides thing A that contains channel `channel_A`
* binding B provides thing B that contains channel `channel_B`
* there is an item `multiLinkedItem`
* `channel_A` is linked to `multiLinkedItem`
* `channel_B` is linked to `multiLinkedItem`
* binding A wants to enable the auto update
* binding B wants to disable the auto update
* `multiLinkedItem` received a command (e.g. `SwitchType.ON`)

maggu2810:
At the moment, if there exists multiple `AutoUpdateBindingConfigProvider` for the same item the auto update is done if at least one enables the auto update.
Should we proceed the same way using binding IDs?

kaikreuzer:
> Then we could take the binding ID and decide if an auto-update should be done

Hm, I think the binding ID is much too coarse grained. The idea of this issue was to achieve a similar feature as in OH1, where a binding can decide for each item (here now channel), whether it wants to disable the auto-update as it itself will take care of reliable status updates after commands. This feature very much depends on the specific channel, so having just all on/off isn't imho helpful.

[...]

PR #2881, opened 2017-01-21
AutoBridge Binding - Enable a mechanism to control how Command and St…

kgoderis:
…ate updates are bridged between Channels that are bound to an Item

`Dimmer TestDimmer {channel="knx:generic:ip1:1_1_131:1_2_65", channel="knx:generic:ip1:1_1_131:1_2_66", channel="knx:generic:ip1:1_1_131:1_2_67",channel="knx:generic:ip1:1_1_131:1_2_69", autobridge="none"}`

Example : a KNX Dimmer is made up of a series of Channels, each linked to a KNX Group Address with a different DPT (1.001 to switch it via ON/OFF, 3.007 to dim it via INCREASE/DECREASE, or 5.001 to set the absolute value using DecimalTypes). Adding autobridge="none" will change the behaviour of the ThingManager: a Command or Update that is received through one of the defined Channels, will not be passed on to the other Channels of the TestDimmer. This is essential, as in the example, the Dimmer Itemcan be controlled with an OnOffType (GA 1/2/65) from the UI, but is also set by reading out the "switch" state on GA 1/2/66 (The KNX device is reporting its actual state on that Group AddresS). If bridging would be allowed, then switching on the Dimmer by sending an ON Command to the dimmer will cause it to switch on, but consequently, when the KNX device is sending back a status update on GA 1/2/66, it will cause a second message to the KNX device ((e.g. the ThingManager is pushing a ON StateUpdate to that given channel)) through GA 1/2/67 , as that is the Group Address to set the absolute value of the dimmer. This behaviour is to be avoided

AutoBridge can take 4 values:

"all" -> This is the current behaviour of the ThingManager : anything that is passed on a given Channel, is bridged to all other channels
"none" -> no bridging
"inter" -> bridging between channels that have a different binding ID. In the example above, imagine an additional channel that is linked to LIFX bulb, then the Commands between a "knx" channel and the "lifx" channel will be bridged, but not in between the "knx" channels
"intra" -> bridging will only happen between channels having the same binding ID. 

This PR is essential for https://github.com/openhab/openhab2-addons/pull/1438, certainly in the use case whereby Things and Channels are automatically defined and provided from the ETS4/5 knxproj files.


kgoderis:
@kaikreuzer Kai, any chance you can integrate this one rapidly? I would like to do a stability test of the KNX binding in my operational environment so that the KNX binding everyone is waiting for can be finished. Without this PR this is not possible.

kaikreuzer:
@kgoderis I am sick this week and have a pile of stuff to catch up with, sorry.
I also have the problem that I do not quite get the idea - or at least not the way you suggest to address it. 
The "BindingConfigProvider" interface is rather some legacy API, which we also try to avoid using for the autoupdate feature (https://github.com/eclipse/smarthome/issues/595). it also isn't at all compatible with configuration through UIs.
Furthermore, we strive for de-coupling - the ThingManager should only deal with item events and passing them on to the bindings. But not being a configurable broker, where you can filter events based on their origin (some specific binding) etc.

Maybe @maggu2810 and @sjka find some time to review and comment further.

maggu2810:
I am sick myself, too.
I assume I don't understand the auto bridge approach, will try to read your first comment again...

kgoderis:
@kaikreuzer @maggu2810 @sjka Ok - just to clarify a bit the issue at hand. Imagine a very simple Item configuration as follows:

Number BathroomKidsSetpoint <temperature> { channel="knx:generic:ip1:1_2_16:4_3_38", channel="knx:generic:ip1:1_2_16:4_3_39" }

This Item serves to control a thermostat. The first channel is to set the temperature of the thermostat. The second channel is used by the remote thermostat to send back the actual value of the thermostat.

1. The user sends a value to the Item through the GUI
2. The ThingManager delegates the value to both channels
3a. The Binding, in casu KNX, writes the value to the first channel (as it sets the temperature)
3b. The binding, when asked to write the value to the second binding, says "no thank you, this is a read-only KNX Group Address" (e.g. it does not have the "W" ETS flag). 
4. The thermostat receives the commands, sets the temperature, and sends back the actual set temperature on the KNX bus, using the Group Address 4/3/99 of the second channel. 
5. The binding receives that value, and updates the state of the second channel, and thus the Item gets updated
6. The ThingManager picks up the change of state of the bound Item
7. The ThingManager delegates the updated value to both channels
8. The binding writes the this value to  group address of the first channel, as in step 3a. It does so, because it is by definition stateless: it does not know that a command was sent x seconds before, so it dutifully sends the value back to the KNX bus
9. We are back at 4.

and we end up in an endless loop.

The problem is that there is no mechanism *at all* to track what was written/done/... in the past. in openHAB 1 we could nicely solve this because Items and the binding were directly connected to each other. However, this is not the case anymore as we have the concept of Channels sitting in between. The binding, when getting a delegation order from the ThingManager can not know where that is coming from: the GUI, a channel of another binding, another channel of the KNX binding,.... Yes, we do capture the source of these things in the StateEvents, but bindings do not get that level of information. 

This was just a simple solution. Just imagine a Dimmer. You know that most Dimmer actors in KNX you can send 3.007 DPT's (increase/decrease), 5.001 (set absolute value), 1.001 (switch on/off). That alone means at least 6 Group Addresses for writing/reading statuses. 

The only alternative solution I see for this kind of problem - if you do not want the ThingManager to be broker - is to pass on the source of the Command or State update to the binding. So that at least it knows how to handle it.

But then again, even if it would be the case, it would not work for the KNX binding when parsing ETS5 files. I mean, in ETS5 there is no mechanism to describe what bunch of Group Addresses make up that Dimmer. I know that intuitively you would say: "no, there is". Well, you *do* configure them because you/the user interpret what is on the screen, and you link Group Addresses to Communication Objects, but when parsing the ETS5 XML, that context is not there. It is just a bunch of Group Addresses. So, even if we would pass on the source to the binding, and because the user defines the Dimmer as an Item, it will be up to the user to tell that the bunch of channels that make up the Item definition, are in fact "belonging together", or not. I have use cases where I do not want the ThingManager to delegate to all channels, and I have use cases where I do want this behaviour specifically. 

The ThingManager works well when you have a single channel bound to an Item, or 2 or more channels from different bindings bound to the Item, but for protocols like KNX this does not work at all

I solved that in the first iteration of the KNX binding by defining more complex Things that *do* capture a given functionality like a Dimmer (e.g. you would define the 6 Group Addresses for that Thing, and the ThingHandler would keep track of things, because it knows that these 6 Group Addresses belong together), but the general vote was to go the other way, e.g. define as simple Things as possible (thus, a Thing = a Group Address). That is ok, but that means that the above problem needs a different solution...

tavalin:
Sorry to be nosy...

>1. The user sends a value to the Item through the GUI
> 2. The ThingManager delegates the value to both channels

Why would it do that if you only sent the command to the first channel?

kgoderis:
> Sorry to be nosy...
> 
> The user sends a value to the Item through the GUI
> The ThingManager delegates the value to both channels
> Why would it do that if you only sent the command to the first channel?

You send the command the Item, and it is the ThingManager that delegates to *all* channels that are bound to the Item. You do not have a choice at all.

tavalin:
Right, I didn't realise you had bound both channels to the same item. I've only dealt with items bound to one channel.

watou:
> The problem is that there is no mechanism at all to track what was written/done/... in the past. in openHAB 1 we could nicely solve this because Items and the binding were directly connected to each other. However, this is not the case anymore as we have the concept of Channels sitting in between.

This is one of a set of challenges that exist when bindings are not connected to items.  Being able to update an item of any type is also not possible without prior user configuration, compared to OH1 bindings (or at least I still haven't found out how).  If correct, a range of OH1 bindings cannot be implemented against current ESH APIs.

maggu2810:
@kgoderis Isn't there a mix between states and commands?
If you send a command to the item the handleCommand method of both linked channels is called.
The one handle the command - it send something the KNX bus, the other channel does nothing (missing W).
After that a change is received by the second channel from the KNX bus and it sends a state update.
But this state update will not trigger a handleCommand call of the first channel anymore.

What am I missing?

kgoderis:
> What am I missing?

Oh - that is due to the way - in this case - the KNX binding is working. It does not differentiate between States and Commands. When raw data is received from the KNX bus, it is transformed into a Type (State or Command), and then accordingly handleState/Command() is called. These methods are doing exactly the same in this case, in that sense I interchanged them in my explanation.

kgoderis:
@kaikreuzer @maggu2810 @sjka Just for the sake of pushing the intellectual boundaries a bit, I tried to conceive an alternative mechanism that does not depend on the ThingManager, but that is completely embedded in the Binding. In short, the binding keeps track of when it receives events from the runtime or bus (State or Command), and it tries to guess where these events are coming from (not always able to determine), and then uses time thresholds to determine if it has "seen an event" before or not. If so, it will be trashed. Does it work? yes sort of, but not in a deterministic way

Needless to say that the fact of incorporating a time threshold will be a source of misery, as the choice of the threshold is very deliberate. What do we consider a repetition? something that happened less than 500ms ago? but what if we face a huge KNX busload, then we will miss repetitions. And what about the use case where a user is clicking very rapidly on a "+" button in a GUI to change a value of an Item? then we will consider these to be repetitions, whilst they are not. The last thing we want to do is to bring in timer-based/time-based conditions is something that is not quite a real-time system.

kgoderis:
Obviously, if we could agree that Bindings can use the ItemChannelLinkRegistry then a lot of these issue would not occur, but it was decided that this is not allowed.

J-N-K:
Maybe extending `State` and `Command` with something like

```
private String source;

public void setSource(String source) { 
  this.source=source;
}

public String getSource(String source) {
  return source;
}
```

would help? If source is initialized with the `BINDING_ID` (or perhaps the full class name of the class creating the `State` or `Command`) the handler could use that information to discard it. This is fully backwards compatible as there will be no effect on all functions that simply don't care about that extra information. The only thing that must be made sure ist that the ThingManager keeps that information when passing the `State`/`Command` around.

kgoderis:
@J-N-K That is an option indeed, but I would rather encapsulate other information in the source field. Maybe something along the lines of "itemName:source" with source being the value that is stored in the ItemState(Change)Event. In that Event, the source field (not sure now the field is exactly named) is empty when the source is the runtime itself (e.g. GUI), or the channelUID when it came from a Channel (and thus in fact, the Binding). To be investigated......

kgoderis:
@kaikreuzer @maggu2810 @watou @J-N-K @sjka I have tested an alternative solution (based on a time mechanism that traces what is send to Channels) in my operational environment, and as predicted it fails miserably because of the variability of the time between observations of States/Commands send to Channels.

Given that I see no other viable alternative than the solution I proposed initially in this PR, this effectively means that the KNX binding can not be developed against the current ESH APIs. I will thus cease all development efforts

kgoderis:
Superseded by https://github.com/eclipse/smarthome/pull/4108

Issue #2601, opened 2016-12-06
[Hue Binding]: Change/Update the library underneath which actually talks to Hue API


ddudnik:
Currently used library (https://github.com/Q42/Jue) is not maintained any longer since December of 2014 and will become more and more outdated with respect to Hue API. It is important to deal with this issue to be able to support all the newest and future features of Hue API (one example of such a feature is Sensor API).

After a short discussion with @kaikreuzer (https://github.com/openhab/openhab2-addons/issues/1260), here are some ideas to start with:

1. Take over the Jue library and include the source as a part of Hue binding
2. Find another library (some research is required, see https://github.com/Q42/hue-libs)
3. Ask Philips to publish the source code of their SDK under a proper OSI-compliant license :)
4. .....

P.S. When decision is made, I volunteer myself for the task.

ddudnik:
Jue library is under MIT License, which I guess permits to take it right away and include into Hue Binding source code. 

@kaikreuzer Please correct me if I'm wrong.

kaikreuzer:
In theory you are right, but we very much avoid to mix licenses within the ESH source repository. Thus I prefer to have the code being contributed under EPL. I will go and check with the author whether this is an option for him.

kaikreuzer:
> Find another library (some research is required, see https://github.com/Q42/hue-libs)

All listed Java ones are dead since 2-4 years...

ddudnik:
@kaikreuzer Yeah, I also did a brief review of other existing Java libraries and it seems all of them are not maintained any longer, mostly because Philips provides an official Java SDK now. Makes sense, I guess, since most developers would not required some other alternative than using an official SDK which would work just fine for an Android app etc.

I was also looking at Jue library more closely, it seem to be the most documented and cleanly designed of all, so makes sense to proceed with it if the author is OK with this option.

kaikreuzer:
Good news: I reached the original author of the lib and he got me in contact with the company who is holding the copyright on it. They are ok with our plan to contribute the sources directly to ESH under EPL!

We should prepare a PR, which refactors the code accordingly and adds the sources instead of the jar. @ddudnik, did I get you right that you volunteered for that?

The sources that must be taken are in https://github.com/qivicon/Jue/commits/get-new-lights-with-serial-number (you see that we already added two commits on that in the past for fixes).
[...]

PR #2887, opened 2017-01-24
Include Jue library source as a part of Philips Hue binding (#2601)

ddudnik:
Source code of Jue library is included into Hue binding under `org.eclipse.smarthome.binding.hue.internal` package. Binary 'jue.jar' is removed and all code references are updated.
Issue #2601


maggu2810:
If we want to keep the history we need to remember to not "squash and merge".

I think we could merge it as it is, all further changes on the lib needs to be done later if we get it into this repo.

maggu2810:
I restarted the build because a thing test failed (don't think it is related to this PR).

kaikreuzer:
> If we want to keep the history we need to remember to not "squash and merge".

I actually doubt that we want/can keep the history in place.
We so far have a CQ for a specific version of the library (and not for ALL versions up to that point), so whatever code might have existed earlier in history has NOT been IP approved.

We will also require the copyright holder to approve this contribution with adding him to the commit message. We therefore should imho have a single commit, which both the copyright holder and the Eclipse IP team can review and approve.

maggu2810:
Sound reasable, okay.

kaikreuzer:
@ddudnik Please let me know, once you have updated the PR. I will then try to reach out to the copyright holder for their approval.

ddudnik:
@kaikreuzer I rebased on upstream/master and squashed all commits into one as requested.

kaikreuzer:
Thanks @ddudnik I am taking over regarding further IP clarifications. Give me a week or two for that before we can continue.

extensions/binding/org.eclipse.smarthome.binding.hue/src/main/java/org/eclipse/smarthome/binding/hue/internal/BridgeDiscovery.java
```diff
@@ -0,0 +1,147 @@
...
+
+import org.eclipse.smarthome.binding.hue.internal.HttpClient.Result;
+import org.eclipse.smarthome.binding.hue.internal.exceptions.ApiException;
+
+import com.google.gson.Gson;
+
+/**
+ * Various ways of discovering bridges.
+ *
+ * @author Alexander Overvoorde - initial contribution (as standalone Jue library)
```
kaikreuzer: Could you please replace this line (in all classes from the Jue library) by:
```
@author Q42, standalone Jue library (https://github.com/Q42/Jue)
```

ddudnik:
@kaikreuzer done, '@author' tag changed.

kaikreuzer:
Thanks @ddudnik. I have created CQ 12624, so now the ball is with the Eclipse IP team.

kaikreuzer:
Good news, we have check-in approval!
Thanks @ddudnik for your work and patience - now the way is clear to implement cool new features for hue :-)

ddudnik:
Glad to hear that!

Issue #2893, opened 2017-01-25
Remove Quartz lib dependency from cron trigger


danchom:
Implementation of GenericCronTrigger should use build in cron scheduler instead of Quartz implementation.

PR #2895, opened 2017-01-25
Implements #2893. Remove Quarz lib from cron timer trigger.

danchom:
Fix NPE in RuleEngine. Fix NPE in ExpressionThreadPoolManager.


kaikreuzer:
@danchom Did you see that the timer tests failed on Travis?

danchom:
Yes I saw, unfortunately maven build is run on my machine without problems. I've forgot to commit launch configuration, but I'm not sure if this is  the problem.

kaikreuzer:
Succeeded this time - maybe just some instability in the tests again.

bundles/core/org.eclipse.smarthome.core/src/main/java/org/eclipse/smarthome/core/scheduler/ExpressionThreadPoolManager.java
```diff
@@ -318,29 +318,30 @@ public boolean removeFutures(Runnable task) {
...
                 ArrayList<Future<?>> taskFutures = futures.get(task);
-                if (taskFutures.size() != 0) {
-                    logger.trace("Runnable '{}' has {} Futures to be removed", task.toString(), taskFutures.size());
-                    for (Future<?> future : taskFutures) {
-                        future.cancel(false);
-                        timestamps.remove(future);
-                        obsoleteFutures.add(future);
+                if (taskFutures != null) {
+                    if (taskFutures.size() != 0) {
+                        logger.trace("Runnable '{}' has {} Futures to be removed", task.toString(), taskFutures.size());
```
maggu2810: You should not call toString yourself. This is done by the logger if needed (now toString is called all the time regardless if log level is trace).
This should be changed.

danchom: I've just add check here
  if (taskFutures != null) {
because there was a NPE

maggu2810: Fixed by #2955

Issue #2871, opened 2017-01-20
[PaperUI] items view does not update


sjka:
When having http://localhost:8080/paperui/index.html#/configuration/items open and e.g. updating an item DSL (i.e. removing/adding items), the displayed items stay the same. A refresh of course fixes it, but I would have expected the same behavior as for things etc.

@aounhaider1 could you please have a look?

aounhaider1:
The runtime is not sending any item update events when DSL is changed. Strangely, the event is sent when I update an item through rest endpoint.

sjka:
Strange... I will check that.

sjka:
funny. we hit a "TODO" in GenericItemProvider 😄 

I have created #2896 for the events. Could it be that they are still not processed properly by the PaperUI?

kaikreuzer:
@sjka Is there still a problem in the Paper UI or is it working as expected with #2896?

sjka:
No - the item events now are sent by the framework (& received by the paper UI) but the list does not update dynamically. @aounhaider1 is having a look.

PR #2896, opened 2017-01-25
send item events for generic items

sjka:
...as so far updating the DSL did not cause events to be sent out.
This PR addresses the TODO in GenericItemProvider.

As a result, the ItemsChangeListener is not needed anymore and is
removed from the API.

relates to #2871


sjka:
saw it myself - will fix it

sjka:
okay, with including the latest fixes (by rebasing it), the build works now

kaikreuzer:
Seems as if the ItemsChangeListener is indeed some unused code.
Pity to remove it, it has been there since openHAB 0.1 ;-)

PR #2897, opened 2017-01-25
removal of SmartHome Designer (moved to separate repo)

maggu2810:
That is related to eclipse/smarthome-designer#1


kaikreuzer:
As we keep all bundles in the main repo, I think it would make sense to keep the file "SmartHome Designer (Mac).launch" as well - through this, we should still be able to launch the Designer from within the IDE.

maggu2810:
Okay, will add it again.

I don't care much about the  `SmartHome Designer (Mac).launch` as I cannot use it at all.

I assume, some time we would like to clean the target platform.
I assume also `SmartHome Designer (Mac).launch`  is using Luna or older stuff, because (a more recnet) Tycho complains about that one if is tried to build the RCP application.

```
Failed to execute goal org.eclipse.tycho:tycho-p2-publisher-plugin:0.25.0:publish-products (default-publish-products) on project org.eclipse.smarthome.designer.product: Detected Luna or older launcher feature org.eclipse.equinox.executable version 3.6.100.v20140603-1326.
```

maggu2810:
restored `SmartHome Designer (Mac)` launch configuration

kaikreuzer:
> I assume, some time we would like to clean the target platform.

Probably not before moving the bundles away - they anyhow require most of what we have in the tp.

> I assume also SmartHome Designer (Mac).launch is using Luna or older stuff, because (a more recent) Tycho complains about that one if is tried to build the RCP application.

I doubt that this is related to the launch config, because afaik, Tycho does not care about them, they are really only relevant for the IDE.

maggu2810:
Sure, Tycho does not care about the launch configuration.
What I want to state was that the product build by Tycho and the launch configuration requires Luna and older stuff and a TP cleanup will be related to both.

PR #2914, opened 2017-01-28
Use CONTAINER_NONE in MP3 AudioFormat in URLAudioStream

### LINKED ISSUE #2913: MP3 AudioFormat used by URLAudioStream has an incorrect container
AudioFormat.MP3 is defined as follows:

```
public static AudioFormat MP3 = new AudioFormat(AudioFormat.CONTAINER_NONE, AudioFormat.CODEC_MP3, null, null, null,
            null);
```

While URLAudioStream creates an AudioFormat defined as follows:

```
this.audioFormat = new AudioFormat(AudioFormat.CODEC_MP3, AudioFormat.CODEC_MP3, false, 16, null, null);
```

Based on the JavaDoc for AudioFormat.CONTAINER_NONE, I believe the bug is in URLAudioStream.

```
    /**
     * {@link AudioCodec} encoded data without any container header or footer,
     * e.g. MP3 is a non-container format
     */
    public static final String CONTAINER_NONE = "NONE";
```

The container on URLAudioStream's AudioFormat should be AudioFormat.CONTAINER_NONE. Due to this, AudioFormat::isCompatible(AudioFormat) does not work correctly for URLAudioStream when testing for MP3.

The fix would be this inside org.eclipse.smarthome.core.audio.URLAudioStream:

```
this.audioFormat = new AudioFormat(AudioFormat.CONTAINER_NONE, AudioFormat.CODEC_MP3, false, 16, null, null);
```

syphr42:
This fixes #2913 by changing the container on the MP3 AudioFormat created by URLAudioStream to "NONE".

Bug: https://github.com/eclipse/smarthome/issues/2913
Signed-off-by: Gregory Moyer <moyerg@syphr.com>

### DIFF
--- a/bundles/core/org.eclipse.smarthome.core.audio/src/main/java/org/eclipse/smarthome/core/audio/URLAudioStream.java
+++ b/bundles/core/org.eclipse.smarthome.core.audio/src/main/java/org/eclipse/smarthome/core/audio/URLAudioStream.java
@@ -46,7 +46,7 @@ public class URLAudioStream extends org.eclipse.smarthome.core.audio.AudioStream
             throw new IllegalArgumentException("url must not be null!");
         }
         this.url = url;
-        this.audioFormat = new AudioFormat(AudioFormat.CODEC_MP3, AudioFormat.CODEC_MP3, false, 16, null, null);
+        this.audioFormat = new AudioFormat(AudioFormat.CONTAINER_NONE, AudioFormat.CODEC_MP3, false, 16, null, null);
         this.inputStream = createInputStream();
     }
 


maggu2810:
@kaikreuzer Has strings been choosen by intention for container and codec (e.g. the core framework does not need to be touched to support other ones)?
Using an enum class for container and codec would prevent wrong usage (a codec for a container).
We could use e.g. `AudioFormat.Container.NONE` and `AudioFormat.Codec.MP3`

kaikreuzer:
@maggu2810 That hasn't been introduced by myself, but yes, I think Strings are ok to more easily allow extending the possible values (even outside of the core framework).

PR #2917, opened 2017-01-29
change loglevel for failed stream request

marcelrv:
Signed-off-by: Marcel Verpaalen <marcel@verpaalen.com>


kaikreuzer:
Why? Imho this is only information required while debugging. Not useful for the regular user.

marcelrv:
As it causes otherwise audio to fail silently, which I think isn't very nice.

kaikreuzer:
> As it causes otherwise audio to fail silently, which I think isn't very nice.

If this message would only happen if "audio fails silently"  and thus this is a bug that needs to be reported, I would agree. But in effect, this is just like a 404 on the HTTP server and those should not be logged here.

As I said, for debugging this line is helpful, but then it should be only enabled when debugging is done.

PR #2938, opened 2017-01-31
Fixed logging statement

kaikreuzer:
Fixes https://community.openhab.org/t/rules-not-working/20907/13


danchom:
I'm not sure if it is needed to create redundant an empty map. I've handle this NPE in RuleEngine.updateContext() method.

kaikreuzer:
Ah, ok, you have fixed that in https://github.com/eclipse/smarthome/pull/2895.
I removed that commit, so this PR is now only about fixing the logging statement.

Issue #2937, opened 2017-01-31
Incorrect types of Configuration properties passed to ModuleHandlers


vilchev:
At the moment if we declare configuration with an Integer type property
for example having a 'count' int type Configuration Property
...
        {
          "name": "count",
          "type": "INTEGER",
          "label": "Count",
          "description": "Count",
          "required": true
        }
....

When RuleEngine creates the ModuleHandlers the normalized configuration which is passed to the ModuleHandler contains the property 'count' but with incorrect type -> BigDecimal.

So in order the ModuleHandler to receive its property in the correct type that was declared in the Module Type it has to manually normalize that value itself(but first having the knowledge that the property will be BigDecimal so that avoiding the class cast exception).
The problem is even nastier when having the property type Integer with additional 'multiple=true' then Handler receives a List of BigDecimals.

danchom:
Config property normalization converts all digit types (integer and decimal) to BigDecimal. This is the reason why the module handler receive BigDecimal instead of integer. For me it is not correct too.
@Kai could you explain why the normalization does not convert config values into types defined in their meta-data?

kaikreuzer:
I do not see a bug here.
For both "decimal" and "integer", the internal data type is BigDecimal - you can rely on this. We might have to document that a bit more specifically though.

vilchev:
> I do not see a bug here.
 If a user have declared its property type to be Integer and it receives something else in the end imho it is a bug (If user is not in depth with Configuration impl there is 100% chance of class cast except in the Handler).
We have an API that declares 4 types for Configuration property: TEXT, INTEGER, DECIMAL, BOOLEAN.
I don't see a point having the two types INTEGER, DECIMAL if at the end you will receive a third one BigDecimal in the ModuleHandler? so we actually have TEXT, NUMBER, BOOLEAN at the end?

> For both "decimal" and "integer", the internal data type is BigDecimal - you can rely on this
So each Handler should modify its logic either to normalize the type to the correct one (which is declared in the module type) or to modify its remaining 'Handler logic' to work with BigDecimals..

sjka:
You are mixing up two things here. In the config description, you have the option to specify that a value is expected to be an [INTEGER](https://en.wikipedia.org/wiki/Integer) in the sense of a number without fractions. We explicitly do NOT refer to java types there (as you might have concluded from the TEXT already, as there is no such class in the java language).
[...]

PR #2946, opened 2017-02-01
Fix normalization of Number values. #2937

danchom:

### DIFF
--- a/bundles/config/org.eclipse.smarthome.config.core/src/main/java/org/eclipse/smarthome/config/core/ConfigUtil.java
+++ b/bundles/config/org.eclipse.smarthome.config.core/src/main/java/org/eclipse/smarthome/config/core/ConfigUtil.java
@@ -8,11 +8,15 @@
 package org.eclipse.smarthome.config.core;
 
 import java.math.BigDecimal;
+import java.util.ArrayList;
+import java.util.Collection;
 import java.util.HashMap;
 import java.util.List;
+import java.util.ListIterator;
 import java.util.Map;
 import java.util.Map.Entry;
 
+import org.eclipse.smarthome.config.core.ConfigDescriptionParameter.Type;
 import org.eclipse.smarthome.config.core.normalization.Normalizer;
 import org.eclipse.smarthome.config.core.normalization.NormalizerFactory;
 import org.eclipse.smarthome.config.core.validation.ConfigDescriptionValidator;
@@ -63,9 +67,49 @@ public class ConfigUtil {
         if (configDescriptionParameter != null) {
             Normalizer normalizer = NormalizerFactory.getNormalizer(configDescriptionParameter);
             return normalizer.normalize(value);
-        } else {
-            return value instanceof Double ? BigDecimal.valueOf((Double) value) : value;
+        } else if (value == null || value instanceof Boolean || value instanceof String
+                || value instanceof BigDecimal) {
+            return value;
+        } else if (value instanceof Number) {
+            return new BigDecimal(value.toString());
+        } else if (value instanceof Collection) {
+            return normalizeCollection((Collection) value);
         }
+        throw new IllegalArgumentException(
+                "Invalid type '{" + value.getClass().getCanonicalName() + "}' of configuration value!");
+    }
+
+    private static Collection normalizeCollection(Collection collection) {
+        if (collection.size() != 0) {
+            List list = new ArrayList(collection);
+            Type type = getType(list.get(0));
+            for (ListIterator it = list.listIterator(); it.hasNext();) {
+                Object value = it.next();
+                Type valueType = getType(value);
+                if (type != valueType) {
+                    throw new IllegalArgumentException(
+                            "Invalid configuration property. Heterogeneous collection value!");
+                } else {
+                    if (valueType == Type.DECIMAL && !(value instanceof BigDecimal)) {
+                        it.set((new BigDecimal(value.toString())));
+                    }
+                }
+            }
+            return list;
+        }
+        return collection;
+    }
+
+    private static Type getType(Object value) {
+        if (value instanceof String) {
+            return Type.TEXT;
+        } else if (value instanceof Boolean) {
+            return Type.BOOLEAN;
+        } else if (value instanceof Number) {
+            return Type.DECIMAL;
+        }
+        throw new IllegalArgumentException(
+                "Invalid type '{" + value.getClass().getCanonicalName() + "}' of configuration value!");
     }
 
     /**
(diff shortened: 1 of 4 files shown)


sjka:
As I already pointed out in https://github.com/eclipse/smarthome/issues/2937#issuecomment-276369762: You are right to propose this change and it certainly makes sense. However, it will break many existing bindings in a very painful kind of way as it is pretty much impossible to find all places with tooling.

maggu2810:
There are failing tests.

If addons or bindings are using the Configuration in a wrong way, shouldn't the binding be fixed, stabilised before using a new ESH stable build instead?
I think it is a valid fix for the framework, to force that the configuration should correct types.
Every fix in the framework could break third party stuff. So changes should be done with respect... but if the fix only ensures correct behaviour I would give it a go.

maggu2810:
```diff
diff --git a/bundles/config/org.eclipse.smarthome.config.core.test/src/test/groovy/org/eclipse/smarthome/config/core/test/ConfigurationTest.groovy b/bundles/config/org.eclipse.smarthome.config.core.test/src/test/groovy/org/eclipse/smarthome/config/core/test/ConfigurationTest.groovy
index bb2a21d50..cfcba2876 100644
--- a/bundles/config/org.eclipse.smarthome.config.core.test/src/test/groovy/org/eclipse/smarthome/config/core/test/ConfigurationTest.groovy
+++ b/bundles/config/org.eclipse.smarthome.config.core.test/src/test/groovy/org/eclipse/smarthome/config/core/test/ConfigurationTest.groovy
@@ -83,7 +83,7 @@ class ConfigurationTest {
 
         def configuration = new Configuration(orgProperties)
 
-        assertThat configuration.get("intField"), is(equalTo(1))
+        assertThat configuration.get("intField"), is(equalTo(BigDecimal.ONE))
 
         configuration.setProperties(newProperties)
 ```

**Comment by danchom** [comment]
Unfortunately it was not so easy just to normalize the Numbers. The configurations contains also list of strings, booleans and numbers. I've add normalization for them and fix some TC.
If you think that this PR is not deserved, fell free to reject it.

File: bundles/config/org.eclipse.smarthome.config.core/src/main/java/org/eclipse/smarthome/config/core/ConfigUtil.java [hunk 1]
```diff
@@ -63,9 +66,51 @@ public static Object normalizeType(Object value, ConfigDescriptionParameter conf
...
             return normalizer.normalize(value);
-        } else {
-            return value instanceof Double ? BigDecimal.valueOf((Double) value) : value;
+        } else if (value == null || value instanceof Boolean || value instanceof String
+                || value instanceof BigDecimal) {
+            return value;
+        } else if (value instanceof Number) {
+            return new BigDecimal(value.toString());
+        } else if (value instanceof Collection) {
+            return normalizeCollection((List) value);
```
maggu2810: If you check for `Collection` wouldn't it be better to work on the collection interface (and perhaps create a new list) instead of rely that the collection will always be a list on this place (if it is the case, then use `instanceof List`.

bundles/config/org.eclipse.smarthome.config.core/src/main/java/org/eclipse/smarthome/config/core/ConfigUtil.java
```diff
@@ -63,9 +66,51 @@ public static Object normalizeType(Object value, ConfigDescriptionParameter conf
...
         }
+        throw new IllegalArgumentException(
+                "Invalid type '{" + value.getClass().getCanonicalName() + "}' of configuration value!");
+    }
+
+    private static List normalizeCollection(List collection) {
```
maggu2810: See above, I think this should be changed to
`private static List normalizeCollection(Collection collection)`

bundles/config/org.eclipse.smarthome.config.core/src/main/java/org/eclipse/smarthome/config/core/ConfigUtil.java
```diff
@@ -63,9 +66,51 @@ public static Object normalizeType(Object value, ConfigDescriptionParameter conf
...
+        if (collection.size() == 0) {
+            return collection;
+        }
+        Type type = null;
```
maggu2810: You can initialize it once and skip the `type == null` in every loop.
```java
final Type type = getType(collection.iterator().next());
```

kaikreuzer:
> Every fix in the framework could break third party stuff. So changes should be done with respect... but if the fix only ensures correct behaviour I would give it a go.

I agree, if there is a bug, we should fix it.

bundles/config/org.eclipse.smarthome.config.core/src/main/java/org/eclipse/smarthome/config/core/ConfigUtil.java
```diff
@@ -63,9 +67,49 @@ public static Object normalizeType(Object value, ConfigDescriptionParameter conf
         if (configDescriptionParameter != null) {
             Normalizer normalizer = NormalizerFactory.getNormalizer(configDescriptionParameter);
             return normalizer.normalize(value);
-        } else {
-            return value instanceof Double ? BigDecimal.valueOf((Double) value) : value;
+        } else if (value == null || value instanceof Boolean || value instanceof String
+                || value instanceof BigDecimal) {
+            return value;
+        } else if (value instanceof Number) {
+            return new BigDecimal(value.toString());
```
kaikreuzer: Is the toString() always precise and ok regarding performance? Before, Doubles were directly wrapped into the BigDecimal as a Double (without doing a string serialization inbetween). Might it be useful to do more specific BigDecimal constructions depending on the value type?

danchom: The constructor new BigDecimal (double ) must not be used because  constructor gets all the imprecision of the double

It was written in its doc:
" 
....
<b>Notes:</b>
     * The results of this constructor can be somewhat unpredictable.
     * One might assume that writing {@code new BigDecimal(0.1)} in
     * Java creates a {@code BigDecimal} which is exactly equal to
     * 0.1 (an unscaled value of 1, with a scale of 1), but it is
     * actually equal to
     * 0.1000000000000000055511151231257827021181583404541015625.
     * This is because 0.1 cannot be represented exactly as a
     * {@code double} (or, for that matter, as a binary fraction of
     * any finite length).  Thus, the value that is being passed
     * <i>in</i> to the constructor is not exactly equal to 0.1,
     * appearances notwithstanding.
.....
"

You can try this code:

        double c = 47.48;
        BigDecimal b = new BigDecimal(c);
        System.out.println(b);
        // This prints 47.47999999999999687361196265555918216705322265625

kaikreuzer: ok, thanks, valid point.

Issue #2911, opened 2017-01-28
Advanced dynamically added channels are not hidden in Paper UI


wborn:
For #2663 I am dynamically adding advanced channels (`colorzone`, `temperaturezone`) to `colormzlight` Things. That works well, however Paper UI does not hide these `advanced` channels when configuring or controlling a Thing.

I did some research and to me it seems Paper UI uses the response from the `thing-types` REST service to determine if channels should be hidden. 

However the dynamic channel types (`colorzone`, `temperaturezone`) are missing from the response of http://localhost:8080/rest/thing-types/lifx:colormzlight :

```json
{
    "channels": [{
        "description": "Selects the color of the light",
        "id": "color",
        "label": "Color",
        "tags": ["Lighting"],
        "properties": {},
        "category": "ColorLight",
        "advanced": false,
        "typeUID": "lifx:color"
    }, {
        "description": "Sets the temperature of the light",
        "id": "temperature",
        "label": "Temperature",
        "tags": [],
        "properties": {},
        "category": "Temperature",
        "advanced": false,
        "typeUID": "lifx:temperature"
    }],
    "channelGroups": [],
    "configParameters": [{
        "label": "LIFX device ID",
        "name": "deviceId",
        "required": true,
        "type": "TEXT",
        "readOnly": false,
        "multiple": false,
        "advanced": false,
        "limitToOptions": true,
        "options": [],
        "filterCriteria": []
    }, {
        "defaultValue": "300",
        "description": "The time to fade to the new color value (in ms).",
        "label": "Fade time",
        "name": "fadetime",
        "required": false,
        "type": "INTEGER",
        "readOnly": false,
        "multiple": false,
        "advanced": false,
        "limitToOptions": true,
        "options": [],
        "filterCriteria": []
    }],
    "parameterGroups": [],
    "properties": {},
    "UID": "lifx:colormzlight",
    "label": "LIFX Color MultiZone Light",
    "listed": true,
    "supportedBridgeTypeUIDs": [],
    "bridge": false
}
```

The channel type definitions used can be found in this [channel.xml](https://github.com/wborn/smarthome/blob/2663-lifx-multizone-support/extensions/binding/org.eclipse.smarthome.binding.lifx/ESH-INF/thing/channel.xml)

The channels are then dynamically created in this [LifxLightHandler.java](
[...]

PR #2949, opened 2017-02-01
Use channelTypeUID for channel types comparison

aounhaider1:
fixes https://github.com/eclipse/smarthome/issues/2911

### DIFF
--- a/extensions/ui/org.eclipse.smarthome.ui.paper/web-src/js/services.js
+++ b/extensions/ui/org.eclipse.smarthome.ui.paper/web-src/js/services.js
@@ -559,37 +559,35 @@ angular.module('PaperUI.services', [ 'PaperUI.constants' ]).config(function($htt
             var self = this;
             self.thingType = thingType, self.channelTypes = channelTypes, self.channels = channels;
             return $.grep(channels, function(channel, i) {
-                var channelType = self.getChannelTypeById(self.thingType, self.channelTypes, channel.id);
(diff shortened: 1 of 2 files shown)


extensions/ui/org.eclipse.smarthome.ui.paper/web-src/js/services.js
```diff
@@ -559,45 +559,43 @@ angular.module('PaperUI.services', [ 'PaperUI.constants' ]).config(function($htt
...
                     }
-                } else {
-                    return;
                 }
             }
             if (channelTypes) {
                 var c = {}, c_i, c_l;
                 for (c_i = 0, c_l = channelTypes.length; c_i < c_l; ++c_i) {
                     c = channelTypes[c_i];
                     var id = c.UID.split(':', 2);
```
wborn: The `id` variable assignment statement can be removed because `id` is no longer used.

wborn:
Yes I can confirm this PR now fixes the issue. :-) Using the UID will also solve issues when multiple bindings define a channel type with the same ID.

aounhaider1:
Updated.

wborn:
Thanks! LGTM now.

pathec:
Hi,

When will the changes be available? I would like to test it with the Z-Way binding. After updating openHAB, these changes are not yet included, because the verification process is probably not yet completed.

kaikreuzer:
See https://github.com/eclipse/smarthome/wiki/ESH-used-in-openHAB.
I will do a new ESH stable some time this week.

pathec:
Ok. Thanks for the information.

PR #2952, opened 2017-02-01
Handle item add/update/remove events

aounhaider1:
fixes https://github.com/eclipse/smarthome/issues/2871

### DIFF
--- a/extensions/ui/org.eclipse.smarthome.ui.paper/web-src/js/controllers.control.js
+++ b/extensions/ui/org.eclipse.smarthome.ui.paper/web-src/js/controllers.control.js
@@ -299,9 +299,14 @@ angular.module('PaperUI.controllers.control', []).controller('ControlPageControl
         'Zoom' : {},
     }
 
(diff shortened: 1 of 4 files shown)


kaikreuzer:
@sjka For you to review as you entered the bug :-)

sjka:
Hmmm, I see some really strange behavior. At first, I thought it's not working at alle, but then I noticed it is somehow always "one behind". I.e. if I change the item label from test1 to test2, nothing happens. If I change it from test2 to test3, the UI shows "test2". Can you please check what's going on?

aounhaider1:
I could not reproduce this issue. There is obviously a delay of 2-4 seconds until the event arrives but paperUI is updating the item immediately after that. I've put the item update to next digest cycle, that should fix any weird issue regarding scope update.

triller-telekom:
I have tested it with Firefox, Chromium and Opera, all browsers automatically show an added or removed item immediately.

sjka:
indeed, must have been an issue with my project setup.
lgtm

PR #2953, opened 2017-02-01
Fixes #2663 [LIFX] Add support for MultiZone Light Control 

### LINKED ISSUE #2663: [LIFX] Add support for MultiZone Light Control
The new LIFX protocol adds [MultiZone Light Control](https://lan.developer.lifx.com/docs/multizone-light-control) functionality. This way you can change the color and color temperature per zone. The protocol allows for controlling at most 256 zones.

The [LIFX Z](http://www.lifx.com/products/lifx-z) is the first light to support this new functionality. Each strip has 8 zones consisting of 6 LEDs. At most 10 strips can be chained. Some LIFX Z lights may thus have up to 80 uniquely addressable zones. 

In the current implementation the LIFX Binding always changes the color and color temperature of all zones. It would be nice if the binding also supports changing individual zones.

Creating a new `colormzlight` thing type for this is logical. But what would be the best way to support so many channels? Perhaps using a [channel group](http://www.eclipse.org/smarthome/documentation/development/bindings/thing-definition.html#channel-groups) for both color and temperature that does not show up in the UI (i.e `advanced="true"`)? Can the channels groups be easily updated when strips are added/removed?

In the worst case, when some day a light is introduced that does have the maximum of 256 zones, there would be 256 color channels and 256 color temperature channels for a single light.

Another option could be to create an action that changes individual zones.

What are your thoughts on this matter @kaikreuzer, @kgoderis ?

wborn:
This PR adds support for LIFX MultiZone Light Control as follows:
* The [MultiZone Messages](https://lan.developer.lifx.com/docs/multizone-messages) have been implemented in the protocol package
* A `colormzlight` Thing type has been introduced for the [LIFX Z](http://www.lifx.com/products/lifx-z)
* Advanced dynamic channels `colorzone` and `temperaturezone` have been implemented for controlling zones
* A `HSBK` type and `HSBKField` have been introduced to simplify keeping track of and working with LIFX colors
* The `LifxLightState` now keeps track of light colors using a `HSBK[]` array
* The `LifxLightStateChanger` queues zone updates and uses a single `SetColorRequest` when all zones have the same color
* The documentation has been updated to include the new thing type and channel types

Other improvements include:
* Reduce CPU load by only updating channels on changes
* Reworked ESH-INF XML files to reduce duplication
* Fixed response packets being requested from lights because `Packet` did not set bits to 0 that were already set to 1. This further reduces network traffic.


wborn:
Would you like to review my changes @kgoderis ?

kgoderis:
Sure!

Question : interested in taking the lead over this binding going forward? You are better equipped that me and following up features more rapidly than I can do myself

wborn:
> Question : interested in taking the lead over this binding going forward? You are better equipped that me and following up features more rapidly than I can do myself 

Yes that is something that I am interested in. My sizable test environment will also be around for years to come. Even though the binding is already very stable and feature complete for me, I still have some additions/improvements in mind: i18n support, Thing properties and maybe some light effects.

kgoderis:
@wborn Ok then, that's decided then ;-) 
Extensive changes in this PR, I was not able to test them myself, but reading through the changes it looks ok to me.

FYI, I am in the process of migrating my operational environment to a OH-2.1 snapshot, and I just noticed that the lifx binding is causing quite a bit of CPU load. Anything that you are aware of/observed as well?

wborn:
> Ok then, that's decided then

Great! :-)

> FYI, I am in the process of migrating my operational environment to a OH-2.1 snapshot, and I just noticed that the lifx binding is causing quite a bit of CPU load. Anything that you are aware of/observed as well?

It is not something I ran into myself lately. When I profile the binding with JProfiler on my desktop, it shows that most CPU time is spent in the item registry when the Thing state is updated:

lifx-profiler

If I recall correctly a LIFX light Thing has always been updating its state on every update without checking if there was an actual change. Probably to keep the code simple.

Should each binding optimize state updates itself? The LIFX binding polls the state of each light every 3 seconds.

wborn:
@kgoderis the fix I've just pushed makes the binding update channels only on changes. On my RPi3 the CPU load has dropped significantly.

extensions/binding/org.eclipse.smarthome.binding.lifx/src/main/java/org/eclipse/smarthome/binding/lifx/handler/LifxLightHandler.java
```diff
@@ -164,22 +198,23 @@ public void initialize() {
...
+            lightStateChanger = new LifxLightStateChanger(macAddress, pendingLightState, communicationHandler, product,
                     fadeTime);
 
             communicationHandler.start();
             currentStateUpdater.start();
             onlineStateUpdater.start();
             lightStateChanger.start();
         } catch (Exception e) {
-            logger.debug("Error occured while initializing LIFX handler: " + e.getMessage(), e);
+            logger.debug("Error occurred while initializing LIFX handler: " + e.getMessage(), e);
```
maggu2810: As you already touched this logging statement, can you fix it and stop using string building (use parameters instead, so the logging framework will build the string only if logging is done).

extensions/binding/org.eclipse.smarthome.binding.lifx/src/main/java/org/eclipse/smarthome/binding/lifx/internal/LifxLightCommunicationHandler.java
```diff
@@ -164,7 +164,7 @@ public void start() {
             broadcastPacket(packet);
 
         } catch (Exception ex) {
-            logger.error("Error occured while initializing LIFX handler: " + ex.getMessage(), ex);
+            logger.error("Error occurred while initializing LIFX handler: " + ex.getMessage(), ex);
```
maggu2810: As you already touched this logging statement, can you fix it and stop using string building (use parameters instead, so the logging framework will build the string only if logging is done).

wborn:
@maggu2810  thanks for looking at the code. I've reworked the logging statements.

Can you tell me what conventions are used for logging stack traces with `log.error(..)` in SmartHome? If I recall correctly, a lot of code only logs the `ex.getMessage()` and not the exception. As developer that often results in logging that is insufficient for solving the problem. I've also seen it logging `null` when there is no exception message.

If the goal is that end users are not bothered with complex stack traces, I think something like the logging appender should take care of that and not developers writing code.

maggu2810:
I don't know about an official convention / guideline.
I personally prefer logging the whole stack trace and don't use `ex.getMessage()`.
The only valid reason for the message and not the stack trace is if I know that the error is handled correctly but I will log the message (because I know the message contains useful information). For example the Xtext based rule engine throws an exception with a more or less meaningful description if a rule execution fails. The one that calls execute logs the message. At this place I think the stack trace does not help anyone.

wborn:
I've squashed all commits into one and cherry picked this commit onto the latest master. Travis now also approves this PR.

extensions/binding/org.eclipse.smarthome.binding.lifx/src/main/java/org/eclipse/smarthome/binding/lifx/internal/protocol/ApplicationRequest.java
```diff
@@ -0,0 +1,54 @@
...
+    /**
+     * Gets the integer value of this application request.
+     *
+     * @return the integer value
+     */
+    public int getValue() {
+        return value;
+    }
+
+    public static ApplicationRequest fromValue(int value) {
```
maggu2810: IMHO the JavaDoc should state if the returned reference could be null or is guaranteed to be non-null.

extensions/binding/org.eclipse.smarthome.binding.lifx/src/main/java/org/eclipse/smarthome/binding/lifx/internal/protocol/GetColorZonesRequest.java
```diff
@@ -0,0 +1,81 @@
...
+        setAddressable(true);
+        setResponseRequired(true);
+    }
+
+    public GetColorZonesRequest(int index) {
+        this(index, index);
+    }
+
+    public GetColorZonesRequest(int startIndex, int endIndex) {
+        this();
```
maggu2810: No reason to call the default / no-arg constructor explicit.

extensions/binding/org.eclipse.smarthome.binding.lifx/src/main/java/org/eclipse/smarthome/binding/lifx/internal/protocol/Products.java
```diff
@@ -74,15 +91,24 @@ public boolean isMultiZone() {
         return multiZone;
     }
 
-    public static Products getProductFromProductID(long id) throws IllegalArgumentException {
+    public static Products getLikelyProduct(ThingTypeUID uid) {
```
maggu2810: Can you add a JavaDoc, that the return value if non-null all the time or a IllegalArgumentException is thrown?

extensions/binding/org.eclipse.smarthome.binding.lifx/src/main/java/org/eclipse/smarthome/binding/lifx/internal/protocol/Products.java
```diff
@@ -74,15 +91,24 @@ public boolean isMultiZone() {
...
+            }
+        }
 
-        for (Products c : Products.values()) {
-            if (c.productID == id) {
-                return c;
+        throw new IllegalArgumentException(uid + " is not a valid product thing type UID");
+    }
+
+    public static Products getProductFromProductID(long id) throws IllegalArgumentException {
```
maggu2810: Can you add a JavaDoc, that the return value if non-null all the time or a IllegalArgumentException is thrown?

extensions/binding/org.eclipse.smarthome.binding.lifx/src/main/java/org/eclipse/smarthome/binding/lifx/internal/protocol/SetColorRequest.java
```diff
@@ -68,13 +51,9 @@ public SetColorRequest() {
         setResponseRequired(true);
     }
 
-    public SetColorRequest(int hue, int saturation, int brightness, int kelvin, long fadeTime) {
+    public SetColorRequest(HSBK color, long fadeTime) {
         this();
```
maggu2810: Not your change, but as commented above, I don't think we need to call the default constructor explicit all the time.

extensions/binding/org.eclipse.smarthome.binding.lifx/src/main/java/org/eclipse/smarthome/binding/lifx/internal/protocol/SetColorZonesRequest.java
```diff
@@ -0,0 +1,121 @@
...
+    public SetColorZonesRequest(HSBK color, long fadeTime, ApplicationRequest apply) {
+        this(MIN_ZONE_INDEX, MAX_ZONE_INDEX, color, fadeTime, apply);
+    }
+
+    public SetColorZonesRequest(int index, HSBK color, long fadeTime, ApplicationRequest apply) {
+        this(index, index, color, fadeTime, apply);
+    }
+
+    public SetColorZonesRequest(int startIndex, int endIndex, HSBK color, long fadeTime, ApplicationRequest apply) {
+        this();
```
maggu2810: drop that line :wink:

wborn:
Thanks for having another look at this PR @maggu2810. I agree with your comments and will process them after Easter. It would be nice to get this merged because almost every commit on the LIFX binding causes merge conflicts. :-)

kaikreuzer:
> @kaikreuzer Do we need to create a CQ? Changes: +1,406 -491

Yes, I think we will need one once it is good to merge. Hopefully it should not take long to process, though.

extensions/binding/org.eclipse.smarthome.binding.lifx/src/main/java/org/eclipse/smarthome/binding/lifx/internal/LifxLightCurrentStateUpdater.java
```diff
@@ -123,39 +124,47 @@ private void sendLightStateRequests() {
...
+            } else if (packet instanceof StateMultiZoneResponse) {
+                handleMultiZoneStatus((StateMultiZoneResponse) packet);
+            }
+
+            if (currentLightState.isOnline() && !wasOnline) {
+                wasOnline = true;
+                logger.trace("{} : The light just went online, immediately polling the state of the light", macAsHex);
+                sendLightStateRequests();
+            }
+        } catch (Exception e) {
```
wborn: I've removed the catch because the lock is anyhow released in the finally clause. There were no checked exceptions to catch.

extensions/binding/org.eclipse.smarthome.binding.lifx/src/main/java/org/eclipse/smarthome/binding/lifx/internal/fields/HSBK.java
```diff
@@ -0,0 +1,142 @@
...
+public class HSBK {
+
+    private static final String DEFAULT_PROPERTY_NAME = "hsbk";
+
+    private int hue;
+    private int saturation;
+    private int brightness;
+    private int kelvin;
+
+    public HSBK(int hue, int saturation, int brightness, int kelvin) {
```
wborn: I've rewritten the copy constructor. If kept the `hsb` `temperature` constructor since it is also used in `LifxBindingConstants`. Rewriting that would again introduce some of the copy/paste code that I wanted to get rid of by doing all the conversions from ESH to LIFX colors in the `HSBK` class.

extensions/binding/org.eclipse.smarthome.binding.lifx/src/main/java/org/eclipse/smarthome/binding/lifx/internal/protocol/ApplicationRequest.java
wborn: Added.

extensions/binding/org.eclipse.smarthome.binding.lifx/src/main/java/org/eclipse/smarthome/binding/lifx/internal/protocol/Products.java
wborn: Added.

extensions/binding/org.eclipse.smarthome.binding.lifx/src/main/java/org/eclipse/smarthome/binding/lifx/internal/protocol/Products.java
wborn: Added.

extensions/binding/org.eclipse.smarthome.binding.lifx/src/main/java/org/eclipse/smarthome/binding/lifx/internal/protocol/GetColorZonesRequest.java
wborn: I think that the methods in `GetColorZonesRequest()` would no longer be called? That would result in `protocol = 0` in the parent `Packet` class. Especially the `addressable` protocol bit needs to be set in most packets. See the LIFX protocol [header description](https://lan.developer.lifx.com/docs/header-description).

A lot of the LIFX protocol objects have the same protocol flags being set in their default constructor. This means that the current default value is not properly chosen. I think a good default would be:
* tagged = false (only set to true for discovery packets)
* addressable = true (most messages are sent to a single light)
* responseRequired = false (these response packets do not contain current values, already set to false in `LifxLightStateChanger.createPendingPackets(Packet...)`

Because changing these defaults impacts all packets I would rather do that in another PR after more testing.

kaikreuzer:
Thanks @wborn, looks all good to me.
There seem to be two small comments from @maggu2810 still open: https://github.com/eclipse/smarthome/pull/2953/files#r111538313 and https://github.com/eclipse/smarthome/pull/2953/files#r111538327.

If you could remove these lines as well, it would be nice. Please squash your commits afterwards and I will then create a CQ for the IP team asap.

wborn:
> There seem to be two small comments from @maggu2810 still open: https://github.com/eclipse/smarthome/pull/2953/files#r111538313 and https://github.com/eclipse/smarthome/pull/2953/files#r111538327.

@kaikreuzer I haven't explicitly added a reply to those comments but the `SetColorRequest` and `SetColorZonesRequest` packets also have a default constructor that set the protocol flags. So `this()` is called for the same reasons as outlined in my other [comment about protocol flags](https://github.com/eclipse/smarthome/pull/2953#discussion_r112081379). If you are OK with this I will squash the commits for the CQ.

kaikreuzer:
Sorry, @wborn, didn't make the link to the other comments, right.
I'll leave the last word to @maggu2810, but feel free to already squash the commits in optimistic anticipation of his answer ;-)

maggu2810:
All fine from my side.

wborn:
Great! The commits have been squashed for the CQ now @kaikreuzer.

kaikreuzer:
Thanks! I have created CQ 13288 - waiting for approval now!

kaikreuzer:
@wborn Good news, the CQ has been approved!

wborn:
Excellent news indeed! @kaikreuzer :-)

Issue #2921, opened 2017-01-30
Automation: Rules cannot be distinguished by scope


ViToni:
Rules can be created in different scopes.
Uses cases might be
* system rules
* rules created by an installing party
* rules created by the user
* rules created (automatically) by a remote UI

Currently there is no means to distinguish these rules and to find rules for a given scope in an easy way.

A solution might be to add a `Scope` property to Rule.
Changes needed would be
* `Scope` property for Rule
* Change the REST API to enable transfer of `Scope` for Rule
* Change the REST API to enable search by `Scope` for Rule
* Changes to PaperUI to show (not modify) this information

A different approach might be to a reuse the existing tags property.
There is already a search functionality for Rules by tags.
This approach has the downside that tags are user visible and would be editable by the user.

danchom:
There is visibility property to differentiate rules. The scope property duplicate purpose of the visibility.

danchom:
Also you can use tags for this purpose.

ViToni:
Thae basic idea is, that eg. if there is some kind of alarm system, the setup for the system could create rules with the scope `Alarm`.
If the setup is run again it could check for the rules with scope `Alarm` and change or delete & recreate them if necessary. 

All `Rules` created by the user could have the scope `User`.

Or if the user would want to uses some special remote UI, this remote Ui could manage its own rules in an easy way.

The scope is something which could be user visisble but should not necessarily be user editable.
Tags are in my understanding a means for the user to manage/categorize his rules, therefore editable by the user.
The scope is intended for  remote UIs, special systems, etc.

The aforementioned `Alarm` system example could use tags and scope together, eg:
* Tags=[Security, Indoor, Living room, Windows]
Scope= [InstallingParty1]
* Tags=[Security, Indoor, Living room, Smoke Detector],
Scope= [HealthApp3]
* Tags=[Security, Indoor, Kitchen, Windows]
Scope= [InstallingParty1]
* Tags=[Security, Indoor, Kitchen, Smoke Detector]
Scope= [HealthApp3]
* Tags=[Security, Garden, Camera]
Scope= [SecurityProvider1]
* Tags=[Security, Garden, Motion Detector]
Scope= [SecurityProvider1]
* Tags=[Security, Indoor, Motion Detector]
Scope= [User]
* Tags=[Security, Garden, Motion Detector]
Scope= [User]

The scope links the originator/creator to the rule.

danchom:
@ViToni  thanks for explanation.
[...]

PR #2960, opened 2017-02-02
Added the Scope property to Rule.

ViToni:
* Changed RuleDTO and RuleDTOMapper to support scope property.
* Changed RuleResource to search for tags and scope.
* Changed tests to check for scope.

This PR solves https://github.com/eclipse/smarthome/issues/2921


danchom:
Changes in API has to be discussed first. I don't see a reason for this scope property. It duplicate functionality of rule's visibility property.

Issue #2923, opened 2017-01-30
[Paper UI] Optionally present extensions in a card layout for easy browsing


kaikreuzer:
Currently, the Paper UI shows extensions as a flat list (sorted by category).

This is a good solution for "self-explaining" extensions, such as bindings, where the label of the extension almost says it all.

Once we add rule templates and similar things as extensions, we should imho think of better ways of presenting them - the user won't search for a specific label, but rather do "window shopping", i.e. browsing through the catalog to see what might be interesting for him. With https://github.com/eclipse/smarthome/issues/2922, the extensions can provide more data to present them nicely. To make use of it, I would suggest a card layout similar to the "Control UI" and the user should be able to toggle between list and card layout.

sjka:
I have created #2928 for enhancing the Extension class, which can be used in order to implement this here.  When implementing the card layout, please give feedback in #2928 in case something does not work out as expected from a UI's perspective.

PR #2961, opened 2017-02-02
Added card layout to extensions page

aounhaider1:
fixes https://github.com/eclipse/smarthome/issues/2923

### DIFF
--- a/extensions/ui/org.eclipse.smarthome.ui.paper/web-src/css/components.css
+++ b/extensions/ui/org.eclipse.smarthome.ui.paper/web-src/css/components.css
@@ -343,6 +343,12 @@ md-slider.md-default-theme.hue .md-thumb:after {
     min-height: 60px;
 }
 
+.showCards .fab-item .actions {
+	text-align: right;
+	text-align: -webkit-right;
+	text-align: -moz-right;
+}
+
 .fab-item .circle {
 	border-radius: 50%;
 	width: 56px;
@@ -357,7 +363,7 @@ md-slider.md-default-theme.hue .md-thumb:after {
 .fab-item .rectangle {
     width: 83px;
     height: 56px;
-    background-color: #CCC;
+    background-color: transparent;
     position: absolute;
     text-align: center;
     font-size: 30px;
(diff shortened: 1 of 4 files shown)


kaikreuzer:
Thanks. Some observations:

- I would prefer the list/card buttons to be positioned at the very top, next to the reload button. At the moment, the selection is individual per tab, i.e. when switching to another category, I again see the list view. Furthermore, moving the buttons would safe screen space (no additional white space under the search field) and also allow the inclusion of "card-only" views at other places in the UI in future.
- When pressing the install button, it disappears and the card shrinks in size. It should instead show the "waiting rotation" as in the list view.

kaikreuzer:
Some more:
- If there is no documentation link, the button on the top right must not be shown
- Page switches back to list view upon a browser refresh
- Images must not be changes in their aspect ratio.
- As most logos are wider than high, I would suggest to reserve a 3:2 box for it, which usually fits quite well (see e.g. [here](http://www.openhab.org/technologies.html).
- There must not be a grey background, but if the image is transparent, it should be displayed transparently as well.
- If no logo is available, I would leave the space blank. The grey square with a letter inside looks rather ugly.

Just to show what I am talking about wrt the logo:
screen shot 2017-02-03 at 12 09 58

aounhaider1:
Updated accordingly.

kaikreuzer:
> - There must not be a grey background, but if the image is transparent, it should be displayed transparently as well.
> - If no logo is available, I would leave the space blank. The grey square with a letter inside looks rather ugly.

These two issues are still not solved (tested with Chrome and Safari). Does this work for you?

> - When pressing the install button, it disappears and the card shrinks in size. It should instead show the "waiting rotation" as in the list view.

The "waiting rotation" is shown on the left, while the button disappears on the right - not really what the user would expect. Is it different for you?

kaikreuzer:
Thanks, better.

> Images must not be changes in their aspect ratio.

Is this already addressed?

aounhaider1:
As you said, I reserved fixed space of 3:2 for logo. I can remove the fixed sizing but that means that images will take as much space as they want.

kaikreuzer:
Well, what I asked for is to reserver a fixed 3:2 space and fit in the image without changing its aspect ration. Just the same way as on http://www.openhab.org/technologies.html. Distorting logos is not a good idea, I hope you agree.

aounhaider1:
Should work now.

kaikreuzer:
Yes, but it seems to be aligned top-left. Would it be possible to center it within the 3:2 box?

aounhaider1:
Centered. :)

kaikreuzer:
Indeed :-(

screen shot 2017-02-09 at 11 04 18

Issue #2717, opened 2016-12-29
Thing definition: accept a label and a description for a channel group


lolodomo:
Like for the "channel-type", it should be possible to define an optional label and an optional description for the "channel-group-type".
The XSD has to be updated and Paper UI has to be updated to consider these "properties".

It would make channels in Paper UI clearer for certain bindings like the astro binding.
https://github.com/openhab/openhab2-addons/issues/1520

lolodomo:
It implies mainly:

1. adding a label and a description field to the class ChannelGroupDefinition
2. updating the XSD and the XML loading
3. updating in few places where ChannelGroupDefinition is used to handle properly the new fields
4. maybe updating Paper UI ?

Do I miss something ?

I can certainly take in charge the first 3 points.

Can I have an agreement for this change ?

kaikreuzer:
> it should be possible to define an optional label and an optional description for the "channel-group-type".

I assume that you mean "for the channel-group", since for the channel-group-type, this is already possible.

> Can I have an agreement for this change ?

Yes, I agree that this makes sense. As you mention, we have that situation in the Astro binding and also in https://github.com/openhab/openhab2-addons/issues/1252 we came across the same issue.

If you want to go for it, that would be great! In case you get stuck somewhere, let us know. I also do not know by heart which places need to be adapted, but your list above sounds quite good.

lolodomo:
Ok, I can try to update the DTD, update the XML loading and adjust the thing REST API.
To validate, my test case will consist in adding a label and a description in one of the astro channel grouop, and then check that they are returned when using the thing REST API.

Can you confirm that Paper UI uses the thing REST API to get label and description for the channel groups ? If it does, it should be fixed without any additional change in Paper UI.

kaikreuzer:
> Can you confirm that Paper UI uses the thing REST API to get label and description for the channel groups ?

I do not know. But in any case, @aounhaider1 can probably quickly adapt the Paper UI to get the information from the right place, once it is there.

lolodomo:
@kaikreuzer : I need help. I've done the changes but unfortunately when I try to start OH to test them from Eclipse IDE, I got this stupid error: "Variable references non-existent resource : ${project_loc:distro-resources}". Where is defined this variable ?
In VM arguments, I can see that this variable is used: "-Dopenhab.conf=${project_loc:distro-resources}/src/main/resources"
[...]

PR #2970, opened 2017-02-04
Label and description added for channel groups

lolodomo:
Implements #2717


lolodomo:
The XSD has been updated without changing the version. It remains compatible with all existing files, only new optional tags have been added.

The enhanced XSD needs to be uploaded into http://eclipse.org/smarthome/schemas/thing-description-1.0.0.xsd

PR #2974, opened 2017-02-05
State/Command Trigger expose event

smerschjohann:
This PR allows an action to access the underlying event in a more generic way. All triggers expose the same name: event. This can be especially useful in scripted actions.

### DIFF
--- a/bundles/automation/org.eclipse.smarthome.automation.module.core/ESH-INF/automation/moduletypes/ItemTriggers.json
+++ b/bundles/automation/org.eclipse.smarthome.automation.module.core/ESH-INF/automation/moduletypes/ItemTriggers.json
@@ -13,7 +13,7 @@
                "description":"the name of the item",
                "required":true
             },
-            {  
+            {
                "name":"command",
                "type":"TEXT",
                "label":"Command",
@@ -49,13 +49,20 @@
            }
          ],
          "outputs":[
-            {
+           {
                "name":"command",
                "type":"command",
                "description":"the received command",
                "label":"Command"
+           },
+           {
+               "name":"event",
+               "type":"org.eclipse.smarthome.core.events.Event",
+               "label":"Event",
+               "description":"The events which was sent.",
+               "reference":"event"
             }
-          ]
+         ]
       },
       {
          "uid":"core.ItemStateUpdateTrigger",
@@ -70,7 +77,7 @@
                "description":"the name of the item",
                "required":true
             },
-            {  
+            {
                "name":"state",
                "type":"TEXT",
                "label":"State",
@@ -107,10 +114,17 @@
          ],
          "outputs":[
             {
-               "name":"state",
-               "type":"state",
-               "description":"the item state",
-               "label":"State"
+                "name":"state",
+                "type":"state",
+                "description":"the item state",
+                "label":"State"
+            },
+            {
+                "name":"event",
+                "type":"org.eclipse.smarthome.core.events.Event",
+                "label":"Event",
+                "description":"The events which was sent.",
+                "reference":"event"
             }
           ]
       },
@@ -127,7 +141,41 @@
                "description":"the name of the item",
                "required":true
             },
-            {  
+            {
+               "name":"previousState",
+               "type":"TEXT",
+               "label":"Previous State",
+               "description":"the required previous state of the item",
+               "required":false,
+               "limitToOptions":false,
+               "options":[
+               {
+                  "label": "ON",
+                  "value": "ON"
+               },
+               {
+                  "label": "OFF",
+                  "value": "OFF"
+               },
+               {
+                  "label": "OPEN",
+                  "value": "OPEN"
+               },
+               {
+                  "label": "CLOSED",
+                  "value": "CLOSED"
+               },
+               {
+                  "label": "UP",
+                  "value": "UP"
+               },
+               {
+                  "label": "DOWN",
+                  "value": "DOWN"
+               }
+               ]
+           },
+           {
                "name":"state",
                "type":"TEXT",
                "label":"State",
@@ -174,6 +222,13 @@
                "type":"state",
                "description":"the old item state",
                "label":"Old State"
+            },
+            {
+               "name":"event",
+               "type":"org.eclipse.smarthome.core.events.Event",
+               "label":"Event",
+               "description":"The events which was sent.",
+               "reference":"event"
             }
          ]
       }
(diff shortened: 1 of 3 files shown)


bundles/automation/org.eclipse.smarthome.automation.module.core/ESH-INF/automation/moduletypes/ItemTriggers.json
```diff
@@ -49,13 +49,18 @@
...
+           {
                "name":"command",
                "type":"command",
                "description":"the received command",
                "label":"Command"
-            }
-          ]
+           },
+           {
+                "name":"event",
```
kaikreuzer: Note that we had such an output previously, see https://github.com/smerschjohann/smarthome/blob/bddea50c1c6df2dd06930abefec52bd86d323e1f/bundles/automation/org.eclipse.smarthome.automation.module.core/ESH-INF/automation/moduletypes/EventTriggersTypeDefinition.json#L36.

If you want to bring it back, I'd suggest to also add the label again and define the type as the FQCN and not merely as "Object".

bundles/automation/org.eclipse.smarthome.automation.module.core/ESH-INF/automation/moduletypes/ItemTriggers.json
```diff
@@ -107,10 +112,15 @@
...
-               "type":"state",
-               "description":"the item state",
-               "label":"State"
+                "name":"state",
+                "type":"state",
+                "description":"the item state",
+                "label":"State"
+            },
+            {
+                "name":"event",
```
kaikreuzer: same as above

bundles/automation/org.eclipse.smarthome.automation.module.core/ESH-INF/automation/moduletypes/ItemTriggers.json
```diff
@@ -128,6 +138,40 @@
                "required":true
             },
             {  
+               "name":"oldState",
```
kaikreuzer: oldState does not make sense on an ItemStateEvent, as there is no old value - if someone is interested in a change, he must use ItemStateChangedEvent.

bundles/automation/org.eclipse.smarthome.automation.module.core/ESH-INF/automation/moduletypes/ItemTriggers.json
```diff
@@ -174,6 +218,11 @@
                "type":"state",
                "description":"the old item state",
                "label":"Old State"
+            },
+            {
+                "name":"event",
```
kaikreuzer: same as above

bundles/automation/org.eclipse.smarthome.automation.module.core/src/main/java/org/eclipse/smarthome/automation/module/core/handler/ItemStateTriggerHandler.java
```diff
@@ -42,6 +42,7 @@
 
     private String itemName;
     private String state;
+    private String oldState;
```
kaikreuzer: as mentioned above: please do not add this field.

bundles/automation/org.eclipse.smarthome.automation.module.core/ESH-INF/automation/moduletypes/ItemTriggers.json
smerschjohann: @kaikreuzer: I added it for uid: "core.ItemStateChangeTrigger", so yes it is the ItemStateChange-Event

bundles/automation/org.eclipse.smarthome.automation.module.core/src/main/java/org/eclipse/smarthome/automation/module/core/handler/ItemStateTriggerHandler.java
smerschjohann: The class ItemStateTriggerHandler currently handles changed events as well as normal events, so I think this is correct. But you are right, I have added it at the ItemStateEvent as well, that is wrong.

smerschjohann:
@kaikreuzer I added the FQCN and removed the null check. Please see my other comments as well.

bundles/automation/org.eclipse.smarthome.automation.module.core/ESH-INF/automation/moduletypes/ItemTriggers.json
kaikreuzer: Right, so please remove it here, because this is the definition of the ItemStateUpdateTrigger.

bundles/automation/org.eclipse.smarthome.automation.module.core/src/main/java/org/eclipse/smarthome/automation/module/core/handler/ItemStateTriggerHandler.java
kaikreuzer: Yes, it already handles change events and it does so without the need of this field, so please remove it.

bundles/automation/org.eclipse.smarthome.automation.module.core/ESH-INF/automation/moduletypes/ItemTriggers.json
smerschjohann: No it is not, it is the core.ItemStateChangeTrigger. It is an input parameter to allow reducing the trigger to only trigger if the previous state was "oldState".

bundles/automation/org.eclipse.smarthome.automation.module.core/src/main/java/org/eclipse/smarthome/automation/module/core/handler/ItemStateTriggerHandler.java
smerschjohann: It is an extension to allow trigger if a certain previous state was given.

bundles/automation/org.eclipse.smarthome.automation.module.core/ESH-INF/automation/moduletypes/ItemTriggers.json
kaikreuzer: ah, sorry, Github had it collapsed in an unpleasant way...
Ok then - but please change the "label" as well as at the moment it is identical to the "state" parameter.

bundles/automation/org.eclipse.smarthome.automation.module.core/src/main/java/org/eclipse/smarthome/automation/module/core/handler/ItemStateTriggerHandler.java
kaikreuzer: ok, I was mislead as I thought this is about the module outputs - now I see that this is a configuration parameter.

smerschjohann:
@kaikreuzer Ok, I renamed it to previousState.

kaikreuzer:
> I renamed it to previousState.

Ok for me, but what I referred to was the "label", not the "name". Both parameters have "State" as a label now - better change it to "Current State" and "Previous State" or something like that.

kaikreuzer:
@smerschjohann If you could update the labels, so that they are not identical, we are imho good to merge.

smerschjohann:
@kaikreuzer done

kaikreuzer:
thanks!

Issue #2964, opened 2017-02-03
create a RunRuleAction


BenediktNiehues:
In order to directly execute a set of actions defined in another rule there should be an action module which supports this. This would enable the rule engine to support a kind of scene support. Scenes could be defined in a rule with only a set of actions, theses scenes then can be "activated" by callling runNow() on the rule directly or by an action within anoter rule.
The RunRuleAction needs as parameter a set of ruleUIDs which should be executed.

kaikreuzer:
Sounds useful to me. @danchom, any comments?
@smerschjohann How well would that fit to your work on the JSR233 rule support? Would it be easily possible to define trigger-/condition-less rules in e.g. Jython that are effectively merely a script to be executed as an action?

danchom:
>execute a set of actions defined in another rule 

1) A set of actions can be presented as one action throught the CompositeAction
2) if you wish one rule to start other rule, you can do it now with event action and event trigger
3) I think it is not right a rule action to execute actions of another rule. You just to workaround the entire rule engine. 
If you want to reuse module instances (Trigger and Actions) between rules, this is other question. At the moment it is not possible because we support only AND logical operation between Conditions. If you want to do an OR operation you have to duplicate triggers and actions and split the conditions in two rules. 
I think it would be usable to support ANY logical expression between Conditions. We can do it for example with a special (not required) configuration property containing LDAP expression between conditions.  What do you think?

kaikreuzer:
1. As a user, you can't.
2. How? There is nothing like a "TriggeredByAnActionTrigger".
3. This is not at all related to conditions here. It is about reusing "a readily-configured set of actions".

danchom:
1) right, at the moment. In the feature I think we should have a composite editor for users.  
2) one of actions executed by the rule can be EventAction (which throws event with topic=a). The other rule has to have a EventTrigger which is listening for events with topic=A
3) Right "This is not at all related to conditions here", but if we can do OR operation between conditions we can reuse "a readily-configured set of actions". 
As I said if you think it is useful I'm ok to be merged.

smerschjohann:
@kaikreuzer that would be possible. That would just mean that no (other) triggers have to be added to the rule.

But I think @danchom's solution to simply use topics is more generic and more useful. If there are tags for the rules, it would be cool to simply use those to trigger them as it allows to group the rules.

danchom:
[...]

PR #2983, opened 2017-02-07
RuleRegistry refuse to add Rules without triggers and actions if they…

### LINKED ISSUE #2982: Rules without triggers and actions can be created using REST
Perform POST REST call using rule without triggers or actions (null or empty) - rule is added in RuleRegistry but should not be. The expected behavior is error message: code = 400, message = "Creation of the rule is refused. Missing required parameter."

adimova:
… are not created by template

(Fixes #2982)


kaikreuzer:
This is pretty counter-productive wrt the discussion in https://github.com/eclipse/smarthome/issues/2964.

adimova:
this PR is refused because of #2964

PR #2989, opened 2017-02-09
AudioFormat need to implement hashCode 

afuechsel:
Since AudioFormat implements equals(), it needs to implement hasCode() as well. It is used e.g. in SonosAudioSink in a HashMap. 


maggu2810:
> implements equals(), it needs to implement hasCode() as well

You are right!

PR #2991, opened 2017-02-09
remove double-wrapped safe method calling from ThingManager

### LINKED ISSUE #2986: Handlers taking more than 5000ms to run
There are a growing number of problems showing up on the OH2 forum about handlers taking more than 5000ms - see -:

https://community.openhab.org/t/handler-takes-more-than-5000ms-for-processing-event/22158/15

This seems to affect a wide range of bindings so I would guess it could be framework related.

sjka:
...when initializing Things.

fixes #2986

### DIFF
--- a/bundles/core/org.eclipse.smarthome.core.thing/src/main/java/org/eclipse/smarthome/core/thing/internal/ThingManager.java
+++ b/bundles/core/org.eclipse.smarthome.core.thing/src/main/java/org/eclipse/smarthome/core/thing/internal/ThingManager.java
@@ -65,7 +65,6 @@ import org.eclipse.smarthome.core.thing.type.ThingTypeRegistry;
 import org.eclipse.smarthome.core.thing.util.ThingHandlerHelper;
 import org.eclipse.smarthome.core.types.Command;
 import org.eclipse.smarthome.core.types.State;
-import org.osgi.framework.Bundle;
 import org.osgi.service.component.ComponentContext;
 import org.osgi.service.event.EventHandler;
 import org.slf4j.Logger;
@@ -73,8 +72,6 @@ import org.slf4j.LoggerFactory;
 
 import com.google.common.collect.HashMultimap;
 import com.google.common.collect.ImmutableSet;
-import com.google.common.collect.LinkedListMultimap;
-import com.google.common.collect.Multimap;
 import com.google.common.collect.Multimaps;
 import com.google.common.collect.SetMultimap;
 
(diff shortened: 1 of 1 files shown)


bundles/core/org.eclipse.smarthome.core.thing/src/main/java/org/eclipse/smarthome/core/thing/internal/ThingManager.java
```diff
@@ -977,26 +967,8 @@ protected void addThingHandlerFactory(ThingHandlerFactory thingHandlerFactory) {
...
-                        registerHandler(thing, thingHandlerFactory);
-                        initializeHandler(thing);
-                        return null;
-                    }
-                });
-            } catch (TimeoutException e) {
-                logger.warn("Registering a handler for thing '{}' takes more than {}ms.", thing.getUID(),
-                        SafeMethodCaller.DEFAULT_TIMEOUT);
-            } catch (Exception ex) {
-                ThingStatusInfo statusInfo = buildStatusInfo(ThingStatus.UNINITIALIZED,
```
kaikreuzer: Hm, this method is also called by thingUpdated() and thingAdded() and both these methods do NOT catch any runtime exceptions - so the exception handling that you remove here isn't done anywhere else. Am I missing something?

sjka: both `registerHandler(...)` and `initializeHandler(...)` do exception handling themselves, also setting the thing status accordingly. What are you missing here?

kaikreuzer: Ok, then I missed looking down the code, I only looked up. I somehow expected that this code here was of any use before, but apparently it wasn't...

Just a question which I asked myself (but which is not directly related to this PR): What happens in the (unlikely, but possible) case that handler.initialize() throws an exception AFTER MORE than 5 seconds?

sjka:
> What happens in the (unlikely, but possible) case that handler.initialize() throws an exception AFTER MORE than 5 seconds?

Short answer: The (angry) user will reboot 😉  

Long answer: The thing is going to stay in INITIALIZING state. In the logs there will be a timeout entry and another hopefully meaningful error. We could (in a separate PR) think about setting the thing's state back to UNITIALIZED (HANDLER_INITIALIZING_ERROR: it took too long). But we got to be careful about preventing race conditions which could accidentally lead to INITIALIZING -> ONLINE -> UNINITIALIZED.

Issue #2926, opened 2017-01-30
Switch to Java 8


kaikreuzer:
As discussed earlier, we should upgrade the minimal Java requirement of ESH to become Java 8.

This includes
- [x] changing the Maven build (which also allows us to update to a more recent Tycho version)
- [x] Change Hudson build plan
- [x] updating the manifests
- [x] updating the documentation
- [x] addressing to-dos from https://github.com/eclipse/smarthome/wiki/Java-8-CP2-cleanup

kaikreuzer:
@maggu2810 I won't take any actions before https://github.com/eclipse/smarthome/pull/2897 is done, but when creating the new build in the new repo, you might already want to consider this change.

kaikreuzer:
Null annotations are in place now, so I would consider the Java8 upgrade and related actions being done.

PR #2992, opened 2017-02-09
Stop blaming handlers for being queued

sjka:
..as so far it could happen that a ThingHandler call through the SafeMethodCaller
timed out because it was still queued, so the Handler is not to "blame".

Now the SafeMethodCall returns silently in such a case, not re-throwing the TimeoutException anymore.

relates to #2986


maggu2810:
@sjka Shouldn't we create a new wrapper (wrapper2) that takes the current nanoTime calls the wrapper and takes the nanoTime again. If the diff of that nano time exceeds the limit the warning should be printed.

sjka:
absolutely right, I fully agree! I needed to be quick yesterday night so we have something which can go into the next stable build which Kai wanted to create asap. But I have the same in mind an actually prepared something already. But it's still on my disk only, need to create a PR for it as soon as I have completed the tests.

Another addition I have in mind would be to black-list a binding so we can make the rest survive longer. But I think we should postpone this until #2926 because we could base it on the standard executors then.

PR #2995, opened 2017-02-10
Create default icon for Number type items from Text icon, fixes #2994

### LINKED ISSUE #2994: Broken image in basic ui for number items without explicit icon
Any Number type item without an explicitly defined icon (in a text based .items file) will show a broken image in basic UI.

Until a final choice has been made what to do with these icons i will create a PR to create a number icon that is an alias for the text icon.

rtvb:
Fixes #2994: Broken image in basic ui for number items without explicit icon

Any Number type item without an explicitly defined icon (in a text based .items file) will show a broken image in Basic UI. Until a final choice has been made what to do with these (missing) icons, create an alias for the number icon from the text icon.

Signed-off-by: Robert van Bregt <robert@robertvanbregt.nl>

### DIFF
--- a/extensions/ui/iconset/org.eclipse.smarthome.ui.iconset.classic/icons/_iconcopy
+++ b/extensions/ui/iconset/org.eclipse.smarthome.ui.iconset.classic/icons/_iconcopy
@@ -9,6 +9,7 @@
 cp -f house.png             group.png
 cp -f house.svg             group.svg
 
+cp -f number.svg            text.svg
 
 # Alternative Types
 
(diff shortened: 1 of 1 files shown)


rtvb:
I don't mind manually changing the title of the PR itself, but I am not to experienced with Git, so don't know how to change those. If you point me to some info, I will try if absolutely necessary.

I will copy and paste same more info about the contents of the change from the issue to the PR.

kaikreuzer:
> @kaikreuzer Do you agree to introduce that svg for number as long as missing icons are not handled another way?

I am a bit puzzled. If no icon is specified, the UI should always request an icon of the name of the widget type (which would be "text" and not "number" here) - so I do not see the need to a number icon.

I guess we will have to check what causes the Basic UI to ask for a "number" icon here in the first place...

rtvb:
Resolved through #2994

PR #2996, opened 2017-02-10
added guidelines for logging

kaikreuzer:

### DIFF
--- a/docs/documentation/development/guidelines.md
+++ b/docs/documentation/development/guidelines.md
@@ -39,7 +39,7 @@ Note that this list also serves as a checklist for code reviews on pull requests
 
 1. Eclipse SmartHome requires at least JavaSE 7. Hence no features of Java 8 must be used within the code. To allow optimized JavaSE 8 runtimes, the set of Java packages to be used is furthermore restricted to [Compact Profile 2](http://www.oracle.com/technetwork/java/embedded/resources/tech/compact-profiles-overview-2157132.html).
 1. The minimum OSGi framework version supported is [OSGi R4.2](http://www.osgi.org/Download/Release4V42), no newer features must be used.
-1. For logging, slf4j (v1.7.2) is used with [non-static loggers](http://slf4j.org/faq.html#declared_static).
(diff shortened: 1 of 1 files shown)


kaikreuzer:
I think this is very much in line with how we handle logging (levels) in the project, but I felt the necessity for formally write this down.

docs/documentation/development/guidelines.md
```diff
@@ -51,3 +51,16 @@ Note that this list also serves as a checklist for code reviews on pull requests
...
 1. Bundles need to cleanly start and stop without throwing exceptions or malfunctioning. This can be tested by manually starting and stopping the bundle from the console (```stop <bundle-id>``` resp. ```start <bundle-id>```).
 1. Bundles must not require any substantial CPU time. Test this e.g. using "top" or VisualVM and compare CPU utilization with your bundle stopped vs. started.
+
+## E. Logging
+
+1. As we are in a dynamic OSGi environment, loggers should be [non-static](http://slf4j.org/faq.html#declared_static), when ever possible and have the name ```logger```.
+1. Parametrized logging must be used (instead of string concatenation).
+1. Where ever unchecked exceptions are caught and logged, the exception should be added as a last parameter to the logging. For checked exceptions, this is normally not recommended, unless it can be considered an error situation and the stacktrace would hold additional important information for the analysis.
```
maggu2810: as last parameter / argument without a `{}` placeholder in the string (don't know how to describe that correctly)

kaikreuzer: > don't know how to describe that correctly

Me neither, that's why I wrote it like that...

maggu2810: Hm, my comment was about adding that the for the exception / the last argument **no** placeholder (`{}`) should be added to the string...

kaikreuzer: I understand. As I didn't know how to formulate it, I didn't specifically mention it and hoped it is clear to anyone using slf4j anyhow. If you think something should be added, feel free to do so.

Issue #1654, opened 2016-06-12
Provide access to Thing status in textual (DSL) rules


kaikreuzer:
In rules it is often desirable to access the status of a Thing - currently, this information is not yet available.
We should add a function to access this information within the (old) Xtext DSL rules.

kceiw:
Hi @kaikreuzer . Can you point me how to start? I would like to implement it in spare time. I look at the document https://www.eclipse.org/smarthome/documentation. But still I am not sure how to do.

kaikreuzer:
Hi @kceiw Unfortunately, this isn't too trivial - you should have some understanding of Xtext/Xtend/Xbase.

FTR, this is the place where item names are made available and resolvable within scripts/rules:
https://github.com/eclipse/smarthome/blob/master/bundles/model/org.eclipse.smarthome.model.script/src/org/eclipse/smarthome/model/script/jvmmodel/ScriptJvmModelInferrer.xtend

And the classes in https://github.com/eclipse/smarthome/tree/master/bundles/model/org.eclipse.smarthome.model.script/src/org/eclipse/smarthome/model/script/scoping bring implicit java package imports, command and state types and the action methods.

Somewhere along these places the Things would need to be added. We probably should first discuss the syntax of what we want to achieve. Probably, adding an action could be the simplest solution in the end, although I think we would also want a new rule trigger type.

kceiw:
Thanks @kaikreuzer .
There is a big learning curve. I'm ramping up the code base and trying to understand ESH's design. It could take much time for me to work on this. Anybody with more experience is welcome to work on it. However, I'm still interested in working on it, for satisfy my own requirement.

I'm not quite understand about "the Things need to be added". My understanding is that the status is an attribute of the Thing. Why do we need to add Things?
And for me, I'm using OH1's syntax (is there a new one?). So something like
```
when
    Thing DoorSensor changed from ONLINE to DEAD
then
    ...
end

```

It seems more like a condition than an action. Do you mean that setting state DEAD is an action in the rule system, and that'll trigger whatever in between then and end to be executed? Please correct me if I misunderstand.

kceiw:
Hi @kaikreuzer,
After reading the code, I think we could
(assuming we're using the syntax "Thing DoorSensor changed from ONLINE to OFFLINE)
[...]

PR #3001, opened 2017-02-12
Provide a trigger for thing Online/Offline state in rule.

kceiw:
This is to fix the issue in https://github.com/eclipse/smarthome/issues/1654
With this change, the user can specify the triggers such as
"Thing <thingUID> received update" or "Thing <thingUID> changed" in a rule.
Please note that <thingUID> is a string instead of a ID. This is done so because thingUID contains the character ":". And Online is defined as Thing.getStatus() == ThingStatus.Online. Otherwise, it's offline. I've tested it with two simple rules that just output a log for "received update" and "changed" triggers.
This change is not try to expose thing status to script. That should be done separately.


kceiw:
After I spent a few hours fixing the failing test, I have to ask for help.
The failing test is


Running org.eclipse.smarthome.model.script.tests.scriptengine.ScriptEngineOSGiTest
Tests run: 1, Failures: 1, Errors: 0, Skipped: 0, Time elapsed: 0.744 sec <<< FAILURE! - in org.eclipse.smarthome.model.script.tests.scriptengine.ScriptEngineOSGiTest

testInterpreter(org.eclipse.smarthome.model.script.tests.scriptengine.ScriptEngineOSGiTest)  Time elapsed: 0.465 sec  <<< FAILURE!
java.lang.AssertionError: null

	at org.junit.Assert.fail(Assert.java:86)
	at org.junit.Assert.assertTrue(Assert.java:41)
	at org.junit.Assert.assertNotNull(Assert.java:621)
	at org.junit.Assert.assertNotNull(Assert.java:631)
	at org.junit.Assert$assertNotNull$0.callStatic(Unknown Source)
	at org.codehaus.groovy.runtime.callsite.CallSiteArray.defaultCallStatic(CallSiteArray.java:53)
	at org.codehaus.groovy.runtime.callsite.AbstractCallSite.callStatic(AbstractCallSite.java:157)
	at org.codehaus.groovy.runtime.callsite.AbstractCallSite.callStatic(AbstractCallSite.java:165)


In the test, it verifies the ScriptServiceUtil is not null but it is
	
	scriptServiceUtil = getService(ScriptServiceUtil)
	assertNotNull(scriptServiceUtil)

The only change in ScriptServiceUtil is 

    <reference bind="setThingRegistry" cardinality="1..1" interface="org.eclipse.smarthome.core.thing.ThingRegistry" name="ThingRegistry" policy="static" unbind="unsetThingRegistry"/>

My guess is that ThingRegistry isn't created some how. I tried to debug like

    def ThingRegistry thingRegistry = getService(ThingRegistry)

or

    def ThingHandlerFactory thingHandlerFactory = getService(ThingHandlerFactory) // ThingRegistry needs a ThingHandlerFactory

or

    def ThingHandler thinHandler = getService(ThingHandler) // ThingHandlerFactory needs a ThingHandler

All of them return zero. I am not quite familiar internal OSGI works. During my debugging, what I can tell is that those classes (ThingRegistry, ThingHandlerFactory, ThingHandler) are found (somewhere LoadClass is called and it is successful). But an instance isn't created (don't have this source code and don't understand what's going on). Thus it returns null.

Can anybody please shed a light on it?

kaikreuzer:
I'm no expert on this myself, but I assume that you are simply missing the bundle org.eclipse.smarthome.core.thing during your test executions, since this provides the ThingRegistry as a service. Note that you can add additional required dependencies [here](https://github.com/eclipse/smarthome/blob/6dd1df77136a70ec55da6746be05bf65b78ba15b/bundles/model/org.eclipse.smarthome.model.script.tests/pom.xml#L30-L76), which are then available during the test execution. Check also https://www.eclipse.org/smarthome/documentation/development/testing.html#groovy.

kceiw:
Thanks @kaikreuzer . That works. I added it to pom.xml and updated the pull request.

fishpark:
Cool, seems that you are making progress. When will this feature be available in the snapshot relaease?

kceiw:
I don't see any reviews yet and the maintainers don't approve it yet. I think that is the process to complete the pull request. Until then, I don't think it'll be in snapshot release soon.

fishpark:
Thanks for the info @kceiw. I can hardly wait to use this feature :-)

maggu2810:
> I don't see any reviews yet and the maintainers don't approve it yet. I think that is the process to complete the pull request. Until then, I don't think it'll be in snapshot release soon.

You are right 😉 
Please be patient.

bundles/core/org.eclipse.smarthome.core/src/main/java/org/eclipse/smarthome/core/library/types/OnlineOfflineType.java
```diff
@@ -0,0 +1,43 @@
...
+ * http://www.eclipse.org/legal/epl-v10.html
+ */
+package org.eclipse.smarthome.core.library.types;
+
+import org.eclipse.smarthome.core.library.internal.StateConverterUtil;
+import org.eclipse.smarthome.core.types.Convertible;
+import org.eclipse.smarthome.core.types.PrimitiveType;
+import org.eclipse.smarthome.core.types.State;
+
+public enum OnlineOfflineType implements PrimitiveType, State, Convertible {
```
kaikreuzer: ONLINE/OFFLINE must not be introduced as a new state - it is a "status" which only applies for Things, not for Items.

bundles/model/org.eclipse.smarthome.model.rule.runtime/META-INF/MANIFEST.MF
```diff
@@ -28,6 +28,7 @@ Import-Package: org.eclipse.smarthome.core.common.registry,
  org.quartz.spi,
  org.quartz.utils,
  org.slf4j
-Require-Bundle: org.eclipse.smarthome.model.rule
+Require-Bundle: org.eclipse.smarthome.model.rule,
+ org.eclipse.smarthome.core
```
kaikreuzer: please use Import-Package instead

bundles/model/org.eclipse.smarthome.model.rule/src/org/eclipse/smarthome/model/rule/Rules.xtext
```diff
@@ -72,6 +74,13 @@ SystemOnShutdownTrigger:
 	'System' 'shuts down'
 ;
 
+ThingStateUpdateEventTrigger:
+    'Thing' thing=STRING 'received update' (state=ValidState)?
```
kaikreuzer: The update is not a "ValidState", but ThingStatus (or even rather a ThingStatusInfo, check out the classes - it would need to be decided what syntax to use here in the DSL for it).

kceiw: How about exposing ThingStatusInfo to the rule and scripts.

How about this new syntax?
'Thing' thing=STRING 'received update' (state=ThingStatusInfo)?
ThingStatusInfo = STRING

bundles/core/org.eclipse.smarthome.core/src/main/java/org/eclipse/smarthome/core/library/types/OnlineOfflineType.java
kceiw: Please see the comment in Rules.xtext about ValidState and ThingStatusInfo

bundles/model/org.eclipse.smarthome.model.rule/src/org/eclipse/smarthome/model/rule/Rules.xtext
kaikreuzer: Instead of allowing strings (and requiring hyphens around them), I think we should better introduce dedicated keywords - it is a [well defined list](https://github.com/eclipse/smarthome/blob/3ae5d7b1b284f2bde6199bfe8dad950fd38cae39/bundles/core/org.eclipse.smarthome.core.thing/src/main/java/org/eclipse/smarthome/core/thing/ThingStatus.java#L18-L24), so this should not be hard.
Not sure if we would also have to make the ThingStatusDetail available here, which would complicate stuff - I'd therefore suggest to leave it out of scope and rather only make the ThingStatusInfo available within the "then" block, so that more detailed constraints can be implemented in there.

kceiw: That sounds good to me. So we agree on the syntax like "Thing thing=STRING received update (state=ThingStatusInfo)" And "ThingStatusInfo=... (the well defined list)"

kceiw:
Hi @kaikreuzer 
I create a list of thing status in the Rules.xtext. So it won't messed with item state. I use the string format of ThingStatus.

At the same time, I create a new action (ThingAction) for script. The user can use this action to get ThingStatusInfo in 'then' block. I cannot use dot notation like "item.state" because a thing uid contains ':' which make it an invalid identifier in the script. So I create the action to accept the thing uid and return ThingStatusInfo.

Can you look at that again and give me more feedback?

kceiw:
I tested with a rule. Verify that the rule is executed when the updated/changed events are fired. And getThingStatusInfo() returns the right ThingStatusInfo in 'then' script.

kaikreuzer:
Many thanks @kceiw, this sounds good and at a quick glance, the code looks very clean as well - incredible, how you dived into Xtext/Xbase/actions etc. 👍 
I didn't find yet the time to test it myself, but I trust you that all works as expected - so many thanks again and sorry for my always too slow feedback!

bundles/model/org.eclipse.smarthome.model.rule.runtime/src/org/eclipse/smarthome/model/rule/runtime/internal/engine/RuleTriggerManager.java
```diff
@@ -133,6 +140,11 @@ public RuleTriggerManager(Injector injector) {
             case TRIGGER:
                 result = Iterables.concat(triggerEventTriggeredRules.values());
                 break;
+            case THINGUPDATE:
+                result = Iterables.concat(thingUpdateEventTriggeredRules.values());
+                break;
+            case THINGCHANGE:
+                result = Iterables.concat(thingChangedEventTriggeredRules.values());
```
maggu2810: I assume a break is missing here!

bundles/model/org.eclipse.smarthome.model.rule.runtime/src/org/eclipse/smarthome/model/rule/runtime/internal/engine/RuleTriggerManager.java
```diff
@@ -470,6 +580,15 @@ public void removeRule(TriggerTypes type, Rule rule) {
                 timerEventTriggeredRules.remove(rule);
                 removeTimerRule(rule);
                 break;
+            case THINGUPDATE:
+                for (Set<Rule> rules : thingUpdateEventTriggeredRules.values()) {
+                    rules.remove(rule);
+                }
```
maggu2810: missing break

maggu2810:
@kceiw As this PR has already been merged, can you create a new one that adds a break / return in every case branch?

bundles/model/org.eclipse.smarthome.model.rule.runtime/src/org/eclipse/smarthome/model/rule/runtime/internal/engine/RuleTriggerManager.java
kceiw: Many thanks. I'll create a new PR.

bundles/model/org.eclipse.smarthome.model.rule.runtime/src/org/eclipse/smarthome/model/rule/runtime/internal/engine/RuleTriggerManager.java
kceiw: I'll create a new PR.

yo8192:
Hi,

This looks like what I need, thank you for your work on this.
Any chance you could point me to some documentation on how to use these new features (what is the syntax), or alternatively provide a few examples to get me started?

Thank you,
Thib.

kceiw:
@yo8192 Thanks. 
The documentation in this project doesn't seem to touch this area. I'm not quite sure where to put it. Please let me know if you know it.
At the same time, let me answer your question below.
The syntax for the trigger can be found in the file https://github.com/eclipse/smarthome/pull/3001/files#diff-8033c42c5df568f5b7a61e26b48c1d52
It'll look similar to the one for item. Instead of using item and item name, please use thing and thing uid.
If you want to receive update, you write it
`Thing <thing uid> received update`
If you want to know whenever there is a change
`Thing <thing uid> changed`

You can find <thing uid> from Paper UI. It's a string separate by ':'
Of course, you can put some optional states after 'received update' or 'changed as you do to item. The valid state for a thing can be found from the above file.

In your rule or script, you can get the thing state by using `getThingStatusInfo(thingUid)` (see https://github.com/kceiw/smarthome/blob/25b04e91c423901ba492c77888a3a4517a8765bd/bundles/model/org.eclipse.smarthome.model.script/src/org/eclipse/smarthome/model/script/actions/ThingAction.java). This function returns a `ThingStatusInfo` (https://github.com/eclipse/smarthome/blob/3ae5d7b1b284f2bde6199bfe8dad950fd38cae39/bundles/core/org.eclipse.smarthome.core.thing/src/main/java/org/eclipse/smarthome/core/thing/ThingStatusInfo.java) and you can get the state by calling `getStatus` on the ThingStatusInfo you get.
Note that if you get null for the ThingStatusInfo, usually that means the thing is removed.

kaikreuzer:
@kceiw You are right that the ESH documentation does not cover such features at all.
But it would be cool if you could do a contribution to the [openHAB rule documentation](https://github.com/openhab/openhab-docs/blob/gh-pages/configuration/rules-dsl.md#event-based-triggers) as it would fit in there perfectly!

PR #3002, opened 2017-02-13
Fixed small bugs found by findbugs

afuechsel:


bundles/config/org.eclipse.smarthome.config.discovery/src/main/java/org/eclipse/smarthome/config/discovery/internal/console/InboxConsoleCommandExtension.java
```diff
@@ -126,7 +126,8 @@ private void clearInboxEntries(Console console, List<DiscoveryResult> discoveryR
             DiscoveryResultFlag flag = discoveryResult.getFlag();
             ThingUID bridgeId = discoveryResult.getBridgeUID();
             Map<String, Object> properties = discoveryResult.getProperties();
-            console.println(String.format("REMOVED [%s]: %s [thingId=%s, bridgeId=%s, properties=%s]", flag.name(),
+            console.println(String.format("REMOVED [%s]: %s [label= %s,thingId=%s, bridgeId=%s, properties=%s]",
```
kaikreuzer: could you move the space before %s to the correct place (behind it)?

maggu2810:
That's the reason I use final for method arguments :wink:

PR #3005, opened 2017-02-13
fixed selection of proxy implementation (the code throws an CDNFE, wh…

kaikreuzer:
…ich is an Error and not an Exception)


watou:
oops sorry, I tested the 2.4 servlet fall-through by throwing the wrong thing.  This is the correct fix.

PR #1783, opened 2016-06-30
scriptable automation

smerschjohann:
This PR replaces https://github.com/eclipse/smarthome/pull/803, https://github.com/eclipse/smarthome/pull/806 and https://github.com/eclipse/smarthome/pull/813

It provides the ability to write Rules/ActionHandlers/Triggers directly from script files.

To accomplish this, this PR introduces:
1. A script loader that can load scripts from a directory and bundles
2. An import mechanism to allow bundles to provide script dependant classes for scripts. This means: scripts can import class instances via a default interface where the imported instances can have a reference to the scriptEngineId. This allows the unloading of Rules etc. if the script becomes unavailable. This extension mechanism is for example used to provide the following:
   
   ```
   ScriptExtension.importPreset("RuleSupport")
   ScriptExtension.importPreset("RuleSimple")
   ```
3. The RuleSupport and RuleSimple presets provide the ability for scripts to programmatically add and remove Rules and Modules to the ScriptEngine. By providing the ScriptedRuleProvider, these Rules will only be kept at runtime and will not be persisted on reboot except the Rule will be explicitly added by `addPermanent`. For an example script, see: [distribution/smarthome/conf/scripts/test.py](https://github.com/smerschjohann/smarthome/blob/automation-scriptable/distribution/smarthome/conf/scripts/test.py)

PR #3007, opened 2017-02-14
inject the rule UID to the rule context 'ctx'

maggu2810:
I would like to use an unique identifier in my rule scripts.
IMHO the rule UID could be a good one as it is unique per rule per definition.

What do you think about?

### DIFF
--- a/bundles/automation/org.eclipse.smarthome.automation.module.script/src/main/java/org/eclipse/smarthome/automation/module/script/internal/factory/ScriptModuleHandlerFactory.java
+++ b/bundles/automation/org.eclipse.smarthome.automation.module.script/src/main/java/org/eclipse/smarthome/automation/module/script/internal/factory/ScriptModuleHandlerFactory.java
@@ -53,7 +53,7 @@ public class ScriptModuleHandlerFactory extends BaseModuleHandlerFactory {
                 ScriptConditionHandler handler = new ScriptConditionHandler((Condition) module);
                 return handler;
             } else if (ScriptActionHandler.SCRIPT_ACTION_ID.equals(moduleTypeUID) && module instanceof Action) {
-                ScriptActionHandler handler = new ScriptActionHandler((Action) module);
+                ScriptActionHandler handler = new ScriptActionHandler((Action) module, ruleUID);
                 return handler;
             } else {
                 logger.error("The ModuleHandler is not supported: {}", moduleTypeUID);
--- a/bundles/automation/org.eclipse.smarthome.automation.module.script/src/main/java/org/eclipse/smarthome/automation/module/script/internal/handler/ScriptActionHandler.java
+++ b/bundles/automation/org.eclipse.smarthome.automation.module.script/src/main/java/org/eclipse/smarthome/automation/module/script/internal/handler/ScriptActionHandler.java
@@ -28,17 +28,20 @@ import org.slf4j.LoggerFactory;
  */
 public class ScriptActionHandler extends AbstractScriptModuleHandler<Action> implements ActionHandler {
 
-    private final Logger logger = LoggerFactory.getLogger(ScriptActionHandler.class);
-
     public static final String SCRIPT_ACTION_ID = "script.ScriptAction";
+
+    private final Logger logger = LoggerFactory.getLogger(ScriptActionHandler.class);
+    private final String ruleUid;
 
     /**
      * constructs a new ScriptActionHandler
      *
      * @param module
+     * @param ruleUid the UID of the rule this handler is used for
      */
-    public ScriptActionHandler(Action module) {
+    public ScriptActionHandler(Action module, final String ruleUid) {
         super(module);
+        this.ruleUid = ruleUid;
     }
 
     @Override
@@ -46,25 +49,19 @@ public class ScriptActionHandler extends AbstractScriptModuleHandler<Action> imp
     }
 
     @Override
-    public Map<String, Object> execute(Map<String, ?> context) {
-        Object type = module.getConfiguration().get(SCRIPT_TYPE);
-        Object script = module.getConfiguration().get(SCRIPT);
-        if (type instanceof String) {
-            if (script instanceof String) {
-                ScriptEngine engine = ScriptModuleActivator.getScriptEngine((String) type);
+    public Map<String, Object> execute(final Map<String, ?> context) {
+        Object tmp;
+        tmp = module.getConfiguration().get(SCRIPT_TYPE);
+        if (tmp instanceof String) {
+            final String type = (String) tmp;
+            tmp = module.getConfiguration().get(SCRIPT);
+            if (tmp instanceof String) {
+                final String script = (String) tmp;
+                final ScriptEngine engine = ScriptModuleActivator.getScriptEngine(type);
                 if (engine != null) {
-                    ScriptContext executionContext = getExecutionContext(engine, context);
-                    try {
-                        Object result = engine.eval((String) script, executionContext);
-                        HashMap<String, Object> resultMap = new HashMap<String, Object>();
-                        resultMap.put("result", result);
-                        return resultMap;
-                    } catch (ScriptException e) {
-                        logger.error("Script execution failed: {}", e.getMessage());
-                    }
+                    return execute(engine, new HashMap<>(context), type, script);
                 } else {
-                    logger.debug("No engine available for script type '{}' in action '{}'.",
-                            new Object[] { type, module.getId() });
+                    logger.debug("No engine available for script type '{}' in action '{}'.", type, module.getId());
                 }
             } else {
                 logger.debug("Script is missing in the configuration of action '{}'.", module.getId());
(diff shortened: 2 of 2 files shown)


kaikreuzer:
Could you elaborate a bit on your use case behind it?
Btw, I just had a discussion last night, where I wondered if it would make sense to have an "executionId" in the context, i.e. not only being able to identify the instance, but the individual executions.

maggu2810:
The current use case is a scoped element to add some timer related functions to be used in JavaScript.

This element implements the typical `setTimeout`, `setInterval`, `clearTimeout` and `clearInterval` methods.

IMHO the usage of that functions in scripts could be dangerous as you can have a lot of dangling, long running jobs in background.
So I also added functions that takes an "rule id" for timeout and interval.

```
setTimeoutForRule(ctx.ruleUID, magicfunc1, 1000);
setTimeoutForRule(ctx.ruleUID, magicfunc2, 1000);
setIntervalForRule(ctx.ruleUID, magicfunc3, 1000);
setIntervalForRule(ctx.ruleUID, magicfunc4, 1000);
```

The rule can call
```
clearTimeoutForRule(ctx.ruleUID);
clearIntervalForRule(ctx.ruleUID);
```
at the beginning and ensures there are no timers remaining from an old rule execution.

In this example if the ruleUID is used, the timer could also be cleaned up if the room is deleted (if the scoped element implementation cleans up dangling stuff on rule removed event).

But this is just one use case.
I assume there is nothing wrong to give readable access to the ruleUID.

An execution count for every rule could be used as an execution ID.

danchom:
I think the RuleEngine has to put data into context only when they are needed for all modules, otherwise we will end up with context full with junk. Isn't be a  better a place to put ruleUID into context in the script handler when isSatisfied() or executed() methods are called from rule engine? I can't guess a reason when the other handler will need the ruleUID. Each handler factory receives ruleuid when creates a handler and this ruleUID can be pass to it.
@Kai could you explain why this execution number is needed? If it is needed only for script it also can be control by the handler or handler factory. Also what is the meaning of this number. Is this a count from rule creation or count from restart of the system. In the fist case some storage has to be used, but this will slowdown rule execution.

maggu2810:
@danchom Okay for me, as it could be interested for scripts only. I changed the code.

Another question: Why do you use a undefined type for the value (`final Map<String, ?> context`)? Do you want to prevent that someone is adding something to the context?
If the context should contain whatever type, so every object, you should use Object and not `?`
If you use `?` you cannot add anything to the map, because the compiler does not know if e.g. String is a subtype of `?`

danchom:
@maggu2810 you are right. It should be final Map<String, object> context, and for me it is up to the handler to change the context. Up to now we don't have such case and this have to be changed:
...
public interface ActionHandler extends ModuleHandler {
    public Map<String, Object> execute(Map<String, Object> context);
}

public interface ConditionHandler extends ModuleHandler {
    public boolean isSatisfied(Map<String, Object> context);
}
@kaikreuzer what do you think?

maggu2810:
> I think yes. If each handler can put everything into context and even harm or manipulate the info set by the rule engine i.e outputs of calculated modules.

In this case you should an unmodifiable map.
But I still think we should change from `?` to `Object`.
Also if we create a copy we have to change ("unsafe" cast) from `?` to `Object`.

@danchom If you create a new comment, we get notified by mail with the content of your comment. If you change your comment, no mail is sent (so only your first text exist in the mail and there is no mail with the changed one). IMHO it is very commonly used that the content of the mails are read and not every one open the Github page to look if the comment has been changed.

maggu2810:
Changed the PR to work on a copy of the context.

danchom:
>But I still think we should change from ? to Object.

@kaikreuzer  what do you think about that?

kaikreuzer:
Hi, sorry for my late reply. It seems we have different discussions in here - let me try to answer them individually:

> I assume there is nothing wrong to give readable access to the ruleUID.

If I understand your use cases correctly, what you need is a unique string for a ScriptHandler instance. As every rule has a different instance, imho the hashCode could potentially serve the same purpose. In general, a handler should always make sure that anything that it has created/scheduled is cleaned up when the handler is disposed (what happens if a rule is deleted or modified). So do you really need the ruleUID?

> @Kai could you explain why this execution number is needed?

Don't get me wrong, I didn't ask for this feature; it was just some brain storming regarding the possibility to have a track record of which event has been caused by what exactly. The idea was to have a kind of "stacktrace" which could hold a unique execution id of a rule as a record. But let's not dive into this any further here, I don't have any real requirement for that right now.

> In this case you should an unmodifiable map.
> But I still think we should change from ? to Object.

I'd agree with both points. Or do I miss any argument that speaks for `?`, @danchom?

maggu2810:
> It seems we have different discussions in here

You are right.
Sorry about that, I will create a new issue for the generic type and the usage of unmodifiable map.
It is as usual, you start digging into a code base and ... 😉 

> (what happens if a rule is deleted or modified). So do you really need the ruleUID?

The scoped element that I provide and use already handles the cleanup.
It registers itself as a rule registry change listener for the rule registry.
In the resp. callbacks (e.g. removed) it clean ups all stuff that has been added by the resp. rule UID.
So, I would like to use the ruleUID and no hashCode as the ruleUID is really unique and a hash is not necessary unique.
What I use the rule UID for is perhaps a very special use case and perhaps not the best approach, but as long as there is nothing wrong to export the rule UID I would be happy to get it.

maggu2810:
Created two new issues for the offtopic discussion that could be continued and solved separately now.

kaikreuzer:
> It registers itself as a rule registry change listener for the rule registry.

This sounds pretty nasty to me. Imho, the handlers should really be as independent as possible. Instead of listening for events from the rule registry, the script handler should imho better let the script engine know that it needs to clean up things when dispose() is called. Wouldn't that be an option?

maggu2810:
I agree. Perhaps I should be more precise:

It is not the handler that is used / modified in its current behaviour.
I used a custom scoped element provider that provides an Java object instance.

maggu2810:
I agree. Perhaps I should be more precise:

It is not the handler that is used / modified in its current behaviour.
I used a custom scoped element provider that provides an Java object instance of a custom class.
That custom class instance is registered as a rule registry listener.

kaikreuzer:
Ok, understood. I don't want to block this PR, but I am still not feeling confident about it and I wonder if we should have some notification mechanism from handler dispose->script engine->scoped element providers. With your current solution I fear that we might have some leaks in case the rule engine is restarted and thus does not send expected events or in some other edge cases.

maggu2810:
> I fear that we might have some leaks in case the rule engine is restarted and thus does not send expected events or in some other edge cases.

As it is not part of ESH, I don't think you should fear anything 😉

kaikreuzer:
Well, if this is how we expect solutions to implement their specific requirements, I still fear having designed a bad (or incomplete) API :-)

maggu2810:
For the moment the implementation of mine is hidden for the one that uses it and could be changed with every update. So, I am fine (at the moment).

>I wonder if we should have some notification mechanism from handler dispose->script engine->scoped element providers

> Well, if this is how we expect solutions to implement their specific requirements

We could start a thread on the forum some time.

smerschjohann:
I don't think this is really the right way to do it. As I have understood, you want to implement something like a Scheduler for specific rules. With #1783 this will be possible, if someone implements this type of ScriptExtension (maybe even myself ;) ). This extension also solves the problem of script loading/unloading, there will be no dangling entries if done right.

Do you have any other reason to pass the ruleUID? If there are reasons, I would much more prefer that every Module has a reference to its ruleUID.

smerschjohann:
@kaikreuzer what do you think?

maggu2810:
> Do you have any other reason to pass the ruleUID?

At the moment only for stuff like this:
```JavaScript
log = function(str) {
    print(ctx.ruleUID + ": " + str);
}
```

smerschjohann:
> Ok, understood. I don't want to block this PR, but I am still not feeling confident about it and I wonder if we should have some notification mechanism from handler dispose->script engine->scoped element providers. With your current solution I fear that we might have some leaks in case the rule engine is restarted and thus does not send expected events or in some other edge cases.

Actually, I have changed the ScriptAction in my latest commit to support exactly that. The ScriptExtension(s) are always notified as soon as a script is removed. This is now done at the disposal of the handler.

kaikreuzer:
So @maggu2810, would you be ok if #1783 reverts this change here again (as it comes up with a better mechanism)?

maggu2810:
I am happy using a better implementation of the script engine and scoped element providers handling.
I also will change my timer implementation to use this one if it is ready.

But I don't see what is bad if a script could read its rule UID (also if it is used for logging only).
Is there any reason why we need to hide the rule UID?

kaikreuzer:
Well, I think @danchom gave the [argument above](https://github.com/eclipse/smarthome/pull/3007#issuecomment-279996860) - we should not get into a mood to fill the context with all kind of stuff, but we should really check carefully what is really needed.
If you argue that it is nice to have it available in scripts, I would say that there shouldn't be any difference for scripts and Java code and thus the ruleId should be made availabe to ALL handler code - and with this, we are back at your original implementation to have it added to the rule context.

maggu2810:
> we should not get into a mood to fill the context with all kind of stuff, but we should really check carefully what is really needed.

Agree.

> If you argue that it is nice to have it available in scripts

I find it nice to have it in scripts...
For logging at the moment, but also there we can argue that we should use for logging a scoped element that provides e.g. logging.{trace,debug,info,warn,error,fatal} methods and this one is responsible to prepend the message with the rule UID.

> I would say that there shouldn't be any difference for scripts and Java code and thus the ruleId should be made availabe to ALL handler code - and with this, we are back at your original implementation to have it added to the rule context.

Yes, this has been the change I started with.

We should not use my example above (what I am doing at the moment) to argue if it should be present or not, because this is only one use case (and a better solution for that use case does not matter for this discussion).

So, for me this points that need to be considered:
* don't pollute the context
* don't inject information into the context that should be hidden

If you feel that the rule UID inside the context is a bad thing because we should not be able to use it (if we are allowed to use it, I think it is no pollution), then I agree,we should remove it again.

kaikreuzer:
Ok, I'd say that your current implementation nicely shows that it IS possible for a handler to get hold of the ruleUID even without polluting the rule context. So I think that is probably the best way to go - when ever a handler implementation thinks it needs it, it can have it injected by its factory.
And for the Script*Handler it is fine for me to pass the ruleUID to the scripts as it indeed might be useful for them.

smerschjohann:
seems reasonable, I will add it to the scripted handlers as well.

Issue #3021, opened 2017-02-16
Automation: Context map shouldn't be modifiable


maggu2810:
With respect to the (offtopic) discussion at #3007 I would like to create a new issue.

The context map that is given to the handler should be not modifiable.
So we should use a unmodifiable map view that is provided by the Collections class.

danchom:
Implemented by #3028

Issue #3020, opened 2017-02-16
Automation: Context map should use Object values


maggu2810:
With respect to the (offtopic) discussion at #3007 I would like to create a new issue.

The values of the elements in the context map should be of type Object (so, a known type) and not `?`.

danchom:
Implemented by #3028

PR #3028, opened 2017-02-16
Implements #3020 and #3021. Context map should use Object values and …

danchom:
…must not be modifiable.

### DIFF
--- a/bundles/automation/org.eclipse.smarthome.automation.api/src/main/java/org/eclipse/smarthome/automation/handler/ActionHandler.java
+++ b/bundles/automation/org.eclipse.smarthome.automation.api/src/main/java/org/eclipse/smarthome/automation/handler/ActionHandler.java
@@ -26,10 +26,11 @@ public interface ActionHandler extends ModuleHandler {
      * The Method is called by the RuleEngine to execute a {@link Rule} {@link Action}.
      *
      *
-     * @param context contains action input values and snapshot of all module output values. The output ids are defined
+     * @param context is an unmodifiable map containing action input values and snapshot of output values of triggers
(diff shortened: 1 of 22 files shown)


bundles/automation/org.eclipse.smarthome.automation.api/src/main/java/org/eclipse/smarthome/automation/handler/ActionHandler.java
```diff
@@ -30,6 +30,6 @@
      *            in form: ModuleId.outputId
      * @return values map of values which must be set to outputs of the {@link Action} (may be null).
      */
-    public Map<String, Object> execute(Map<String, ?> context);
+    public Map<String, Object> execute(Map<String, Object> context);
```
maggu2810: Shouldn't we change the JavaDoc to inform the implementer that the context argument is unmodifiable?

bundles/automation/org.eclipse.smarthome.automation.api/src/main/java/org/eclipse/smarthome/automation/handler/ConditionHandler.java
```diff
@@ -29,6 +29,6 @@
      *            defined in form: ModuleId.outputId
      * @return true if Condition is satisfied, false otherwise
      */
-    public boolean isSatisfied(Map<String, ?> context);
+    public boolean isSatisfied(Map<String, Object> context);
```
maggu2810: Shouldn't we change the JavaDoc to inform the implementer that the context argument is unmodifiable?

danchom:
@maggu2810  it is done.

Issue #2984, opened 2017-02-08
ClassCastException in Automation REST


adimova:
java.lang.ClassCastException: org.eclipse.smarthome.automation.type.ActionType cannot be cast to org.eclipse.smarthome.automation.type.TriggerType
at org.eclipse.smarthome.automation.dto.TriggerTypeDTOMapper.map(TriggerTypeDTOMapper.java:36)
at org.eclipse.smarthome.automation.rest.internal.ModuleTypeResource.getAll(ModuleTypeResource.java:82)

adimova:
Possible solutions are:
1. Changing the code in ModuleTypeResource.getAll:
@GET
    @Produces(MediaType.APPLICATION_JSON)
    @ApiOperation(value = "Get all available module types.", response = ModuleTypeDTO.class, responseContainer = "List")
    @ApiResponses(value = { @ApiResponse(code = 200, message = "OK") })
    public Response getAll(
            @HeaderParam("Accept-Language") @ApiParam(value = "language") String language,
            @QueryParam("tags") @ApiParam(value = "tags for filtering", required = false) String tagList,
            @QueryParam("type") @ApiParam(value = "filtering by action, condition or trigger", required = false) String type) {
        final Locale locale = LocaleUtil.getLocale(language);
        final String[] tags = tagList != null ? tagList.split(",") : null;
        final List<ModuleTypeDTO> modules = new ArrayList<ModuleTypeDTO>();

        if (type == null || type.equals("trigger")) {
            modules.addAll(TriggerTypeDTOMapper.map(moduleTypeRegistry.getTriggers(locale)));
        }
        if (type == null || type.equals("condition")) {
            modules.addAll(ConditionTypeDTOMapper.map(moduleTypeRegistry.getConditions(locale)));
        }
        if (type == null || type.equals("condition")) {
            modules.addAll(ActionTypeDTOMapper.map(moduleTypeRegistry.getActions(locale)));
        }
        if (tags != null) {
            filterByTags(modules, Arrays.asList(tags));
        }
        return Response.ok(modules).build();
    }

    /**
     * @param modules
     * @param tagList
     */
    private void filterByTags(List<ModuleTypeDTO> modules, Collection<String> tagList) {
        for (Iterator<ModuleTypeDTO> it = modules.iterator(); it.hasNext();) {
            ModuleTypeDTO mt = it.next();
            Collection<String> mtTags = mt.tags;
            if (!mtTags.containsAll(tagList)) {
                it.remove();
[...]

PR #3034, opened 2017-02-17
Changing ModuleTypeRegistry methods: getTriggers, getConditions, getA…

adimova:
…ctions by adding tags parameters. Solves #2984


maggu2810:
Isn't there a lot of code duplication now?
Why not using one function e.g.
```Java
private <T extends ModuleType> Collection<T> filterByClass(Class<T> clazz, Collection<ModuleType> modules) {
    ...
}
```
that is used in every case?

Then all the functions should be one-liners e.g.
`return filterByClass(TriggerType.class, getByTags(locale, tags));`

maggu2810:
@adimova I created a PR in your repo, so you can have a look at. WDYT?

PR #3042, opened 2017-02-19
BasicUI: restore SSE subscription after connection loss, references #3031

### LINKED ISSUE #3031: Basic UI: Recover from connection issues
The items in the Basic UI are no longer updated when the SSE stream stops working due to a connection timeout. I've looked at the Basic UI code but it seems no error handling function is assigned to the EventSource `onerror` property.

There could be several reasons for a connection timeout (or other connection related issues):
* reverse proxy or firewall connection timeout settings
* unstable WiFi or cellular connections
* unstable/restarted network equipment (switch/router)
* suspended network connections due to battery saving

Reestablishing the event stream will not be enough because events might have been missed. So perhaps the page could reload itself when the connection can be reestablished?

resetnow:


resetnow:
Also adds a notification shown when SSE connection is lost: 

image

Once connection is restored, a new subscription is activated and the content of the current page is reloaded.

resetnow:
Also, it doesn't work in Chrome for me. For some reason, unless the server explicitly closes the connection, onerror handler isn't being called. So when I unplug network cable, Firefox closes the connection and Chrome doesn't — it's still active even in developer tools.

maggu2810:
There are conflicts now

resetnow:
Updated

extensions/ui/org.eclipse.smarthome.ui.basic/web-src/smarthome.js
```diff
@@ -1732,6 +1747,64 @@
...
+			smarthome.UI.navigate(smarthome.UI.page, false);
+		};
+
+		_t.connectionError = function() {
+			if (_t.suppressErrorsState) {
+				return;
+			}
+
+			var
+				notify = renderTemplate(o.notifyTemplateOffline, {});
```
maggu2810: I assume this line break shouldn't be present (the one between `var` and `notify`).

resetnow: It's like this in the whole file — may be a little awkward when there's only one variable declared, but my intention was to keep it consistent. It looks less weird with tab width = 4.

maggu2810: > It's like this in the whole file

You are right, I have had a look at the diff only.
All fine then...

maggu2810:
Thanks

Issue #3027, opened 2017-02-16
Add template browser to rule editor in Paper UI


kaikreuzer:
With #2924, there will be a new extension type "ruletemplate" available that allows users to easily share and use complex rule logic.
From a user's perspective it is awkward to first go to "extensions", browse for a template use, install it, then go to the rule editor and select "create new rule from template".

It would be much nicer to allow browsing available rule templates directly within the rule creation wizard. I would expect it to work as follows:

- IF there are extensions of type "ruletemplate" available, the "+" button in the rule menu should offer a third option "Rule from catalog" (besides "New rule" and "Rule from template").
- This should open a page not a dialog) of all "ruletemplate" extensions in card layout.
- The cards should NOT have the "Install/Uninstall" button, but the whole card should be selectable instead. If that isn't easily possible, a "Select" button could be rendered as well.
- If a card is selected, the UI should do a REST call to `extension/{id}/install`  and the process should continue the same way as for the "rule from template" flow (using the extension id as the template id).

PR #3046, opened 2017-02-20
Extensions page in rules

aounhaider1:
Not to be merged yet.
fixes https://github.com/eclipse/smarthome/issues/3027


aounhaider1:
Updated accordingly.

> Please only do the REST call if the extension is NOT YET installed as otherwise you won't get an "installed" event
In case the extension is installed, it now checks for the rule template and navigates to "rule from template" page if it exists. Shows a "Rule template could not be found" error if extension is installed but template does not exist.

kaikreuzer:
Thanks for the update.

> and navigates to "rule from template" page if it exists

It does so, if you install the template from the "extension" menu as well by clicking the "install" button. This should clearly NOT be the case. Could you please fix that?

aounhaider1:
Updated.

PR #3048, opened 2017-02-20
set tycho version to latest release

maggu2810:


kaikreuzer:
Let's see if this passes... I thought to remember that Tycho 0.24+ requires Java8 to run correctly?

maggu2810:
If my reading has been correct, Java 8 has been necessary for Tycho development (>=0.24), but could be used for projects still using Java 7...
As I am using JDK 8 I don't know exactly.

If the build fails, it is just a matter of reordering... (https://github.com/eclipse/smarthome/issues/2926)

maggu2810:
This is not clear to me (https://wiki.eclipse.org/Tycho/Release_Notes/0.25):

* As a consequence, Tycho now requires Java 8 to run
  * Note that you can still build against and run tests with other JDKs with maven toolchains

maggu2810:
Travis CI (which already uses Java 8) failed caused by a failing test, so not related to this changes.
But I close this PR and create a new one (a WIP one) to start with the Java 8 migration.

PR #3055, opened 2017-02-22
Display of image items in classic UI and basic UI

lolodomo:
RawType enhanced with a mandatory MIME type
RawType enhanced so that toFullString() returns a data URI

Paper UI updated to handle the updated item state for images
Basic UI and Classic UI enhanced to handle sitemap image element associated to an image item
Basic UI and Classic UI enhanced to handle a sitemap image element without a default valid url argument

Sitemap DSL: make optional url parameter for image element

Include ImageItem in default sitemap

Sonos binding enhanced to search for the cover art MIME type, either from the HTTP header or the stream data buffer


bundles/core/org.eclipse.smarthome.core/src/main/java/org/eclipse/smarthome/core/library/types/RawType.java
```diff
@@ -22,32 +22,45 @@
  */
 public class RawType implements PrimitiveType, State {
 
+    private static final String DEFAULT_MIME_TYPE = "image/jpeg";
```
watou: Should `RawType` be renamed `ImageType`, since this assumes it to be a JPEG image unless otherwise specified?

lolodomo:
Here are the changes:
* RawType has been enhanced with a new MIME type attribute set to "image/jpeg" by default.
* The method toFullString() for a RawType now returns a data URI.
* Paper UI, Classic UI and Basic UI use the new data URI extracted from the item state to display the image

What remains to do:
* Image is not updated in Basic UI without reloading the page
* Image channels are not included in the _default sitemap
* bindings using image items have to updated so that they provide an appropriate MIME type to the RawType state in case image/jpeg is not appropriate

bundles/core/org.eclipse.smarthome.core/src/main/java/org/eclipse/smarthome/core/library/types/RawType.java
lolodomo: RawType is currently only used for image channels I believe.

bundles/core/org.eclipse.smarthome.core/src/main/java/org/eclipse/smarthome/core/library/types/RawType.java
```diff
@@ -22,32 +22,45 @@
...
     @Override
     public String toString() {
-        return String.format("raw type: %d bytes", bytes.length);
+        return String.format("raw type (%s): %d bytes", mimeType, bytes.length);
     }
 
     @Override
     public String toFullString() {
-        return PortableBase64.getEncoder().encode(bytes);
+        return String.format("data:%s;base64,%s", mimeType, PortableBase64.getEncoder().encode(bytes));
```
watou: Same question as above: since this assumes HTML formatting, this is less a `RawType` and more specific to representing raw image data in a way understood in HTML.

bundles/core/org.eclipse.smarthome.core/src/main/java/org/eclipse/smarthome/core/library/types/RawType.java
watou: I also believe that is the case, but I am unsure if/how that answers my question.

bundles/core/org.eclipse.smarthome.core/src/main/java/org/eclipse/smarthome/core/library/types/RawType.java
lolodomo: I should admit you are right on that point.

bundles/core/org.eclipse.smarthome.core/src/main/java/org/eclipse/smarthome/core/library/types/RawType.java
lolodomo: We could create a `ImageType` class and keep `RawType` unchanged if you prefer ?
But it will require an update in all (only 5) bindings using RawType for image.

watou: If there is a potential need for moving generic MIME-typed blocks of data as item states, its name is probably fine as is.  But I don't think it's a good idea to have item states that are untyped, generic blobs that don't have semantics for some type (like all the other state types -- OnOff, Percent, Decimal, Point, etc.).  

Doing so turns the state into just data instead of an object, and there may also be more natural transport formats and protocols.  Example: you can't stream a surveillance camera this way, and you obviously don't want to hold its full contents in memory in an item state.  So the model currently used by RawType won't be able to provide practical usage of a range of other resource types.  The range of relevant things that are commonly rendered by client components and are addressable via HTTP URL is very wide and won't all fit in a RawType.  

Something like a `URIType` seems more appropriate, that can update a `LinkItem`, and `LinkItems` can be associated with various UI widgets; for example, show an image on every update, or send `URIType` commands to change radio stations, switch surveillance camera feeds, etc.  This lets more natural protocols and renderers do all the work.

That is my thinking of why it should be renamed `ImageType`, or preferably, dropped as a state type and a more natural solution considered,

maggu2810:
WRT to the your comments, I agree that we should consider about the correct type / type name.
A RawType holds data / bytes without any knowledge about its usage.
A data-uri and a default mime type does IMHO not fit well for this type.

Perhaps we should think about a new type or change the raw type name and implementation.

lolodomo:
In this case I will create an ImageType and use it for ImageItem and keep RawType as it was before my change.
Ok for everybody ? @kaikreuzer

kaikreuzer:
I agree that Image items might need some meta-info about their mime-type. The question is whether this should be part of the state or part of the item. So far, the design is to say in ImageItem that we are dealing with an image, where the RawType is simply the "pure data" - so if we would keep this idea, we should actually add some info to the ImageItem itself (or, as suggested earlier, to the StateDescription for it).

I think it depends on whether we expect the mime-type to be static for an item (so we can be sure that for a given item it is ALWAYS a png) or if we want to allow having different formats for every single state update (which can actually make it much more difficult at runtime as a renderer might be able to render the one state, but not the next anymore).

IF we want to go for adding the info to the state, I would also suggest to keep RawType in place and introduce an additional Type, which ImageItem could then additionally accept - this would give us full backward compatibility.

lolodomo:
If I take the example of Sonos, several image formats are possible so we cannot define a hardcoded format in the state description. I think the information has to be provided in the item state.

I agree with the backward compatibility. If RawType is used for the image item state, I will assume a JPEG format in Basic UI and Classic UI.
I have just a difficulty with Paper UI and the code to adjust to support both RawType and ImageType. @resetnow : any advice how I can combine the old code and the new code depending on the state class (RawType or ImageType) ?

kaikreuzer:
Hm, I am not yet convinced about a new "ImageType" - this is imho too specific. When we have other media, we might have to come up with "AudioType", "FontType" and whatever. Wouldn't a generic "DataType" be the better option, which simply has some mime-type?

Then again, this is very close to the RawType, so I wonder if we cannot adapt the RawType in a backward-compatible way, e.g. by introducing an optional mime-type only.

extensions/ui/org.eclipse.smarthome.ui.basic/src/main/java/org/eclipse/smarthome/ui/basic/internal/render/ImageRenderer.java
```diff
@@ -54,7 +56,9 @@ public boolean canRender(Widget w) {
 
         String sitemap = w.eResource().getURI().path();
 
-        String url = "../proxy?sitemap=" + sitemap + "&amp;widgetId=" + widgetId + "&amp;t=" + (new Date()).getTime();
+        State state = itemUIRegistry.getState(w);
+        String url = (state instanceof RawType) ? state.toFullString()
+                : ("../proxy?sitemap=" + sitemap + "&amp;widgetId=" + widgetId + "&amp;t=" + (new Date()).getTime());
```
resetnow: While this works for now, further support will also require adding classes to image element so it can be handled in javascript

lolodomo:
If we keep RawType with an additional and optional DataType, **I still need to update the output to produce a data URI**. This data URI could be without a MIME type but in this case, browsers will not be able to display the data. So yes, we can avoid to set a default MIME type in RawType but in this case, all bindings using RawType (5) have to be updated to provide a data type; if not images cannot be rendered by browsers.
But I am OK with this solution too.

lolodomo:
It looks difficult to oconverge to a solution that convince evryone.
Until the solution is clear, I put my work in standby.

resetnow:
Well in the end of the day the only people you need to convince are actual maintainers :)

maggu2810:
WDYT about:
* adding a mime-type (String) field to the RawType
* if the field is set it defined the mime type of the RawType's data
* if it is not set (null / empty) the mime type of the RawType is unknown
* we add (to some class, perhaps a new one) a static function that tries to identify the mime type (magic headers, ...), I assume there are implementations that could be used
* we add a static function that receives a mime-type string and a data byte array and returns a data-uri

Pseudo-Code:
```java
String mimeType = rawType.getMimeType();
if (mimeType == null || mimeType.isEmpty()) { // We could perhaps drop null and return an empty string if unset
    mimeType = MagicUtility.identifyMimeType(rawType.getData());
    if (mimeType == null || mimeType.isEmpty()) { // We could perhaps drop null and return an empty string if unset
        logger.warn("Cannot handle unknown mime type"); // jpeg should be identifiable
        return;
    }
}
final String dataUri = MagicUtility.createDataUri(mimeType, rawType.getData());
```

kaikreuzer:
@maggu2810 Sounds ok in general, but I think @lolodomo's point in https://github.com/eclipse/smarthome/pull/3055#issuecomment-281948319 was that the `toFullString()` method will have to be changed in an incompatible way to always return a dataURI.
I actually think that it would make sense to say that we ALWAYS use a dataURI as the serialization format, because it can be ugly for the clients to check whether it is one or not and do a different content handling. But this would mean that we always need to know the mime-type (as it is mandatory for the URI, or rather, text/plain would be wrongly assumed, if not provided), so it shouldn't be optional.

I am not against a breaking API change, if everyone agrees that this is the best solution and we need it.

maggu2810:
Okay for me

lolodomo:
So the only change to my PR would be to suppress the default mime type.

Additionally it will be necessary to update the bindings so tjat they provide a mime type.

lolodomo:
The search of the mime type could be triggered automatically when a RawType is created with data but no mime type. This could avoid changing the bindings.

maggu2810:
I would assume that the one that creates a new RawType instance should know about the mime type already. Also if the code does not know the mime type I would like to see the "magic" outside of the RawType itself.
If the RawType works around the missing mime type itself, the (already known) mime type will not be provided (I think so) in some implementations.
IMHO the current usage of that type will be very rare, so changing that five open source bindings should be okay.

lolodomo:
I have updated my PR.

One very good news: finally basic UI and classic UI are able to display the image even without a mime type set. Very cool.

One bad news: Paper UI does not display the image if no mime type is set (and of course display it when a mime type is set). I don't understand why Paper UI has a different behaviour while the same data URI is used. This problem is the reason why I initially thought the mime type was required.

Tested with Firefox.

lolodomo:
The difference is that Paper UI is ising angularJS to build the img Html tag. Maybe it is more restrictive.

bundles/core/org.eclipse.smarthome.core/src/main/java/org/eclipse/smarthome/core/library/types/RawType.java
```diff
@@ -23,31 +23,51 @@
...
 
     protected byte[] bytes;
+    protected String mimeType;
 
     public RawType() {
         this(new byte[0]);
     }
 
     public RawType(byte[] bytes) {
+        this(bytes, null);
```
maggu2810: As Kai already stated the mime type should not be optional.

bundles/core/org.eclipse.smarthome.core/src/main/java/org/eclipse/smarthome/core/library/types/RawType.java
```diff
@@ -23,31 +23,51 @@
...
+    }
+
+    public RawType(byte[] bytes, String mimeType) {
         this.bytes = bytes;
```
maggu2810: Perhaps we should throw an NPE if mimeType == null

maggu2810:
> But this would mean that we always need to know the mime-type (as it is mandatory for the URI, or rather, text/plain would be wrongly assumed, if not provided), so it shouldn't be optional.

I agree the mime type should be mandatory. If it is optional the one that use the RawType needs always to check if the mime type is set or not and do magic stuff if it is not set. Would be easier to force one.

kaikreuzer:
> Also if the code does not know the mime type I would like to see the "magic" outside of the RawType itself.

It depends. This "magic" might not be easy to implement and forcing every binding to implement it on their own might lead to duplicate code and is error-prone.

What is the situation for e.g. Sonos: As far as I understand, the binding itself has no idea of the mime-type, it really just passes the image from the Sonos speaker over to the UI. So from a binding developer perspective, it would be much nicer if he does not have to deal with any byte stream analysis himself, just to find the mime-type. Being able to pass the bytes to RawType and have THIS do the magic, would be much appreciated (while we can allow the constructor to throw an exception, if the mime-type cannot be determined). Alternatively, we could provide some helper class that can be used by bindings and that holds this logic.

watou:
> A RawType holds data / bytes without any knowledge about its usage. / A data-uri and a default mime type does IMHO not fit well for this type.  Perhaps we should think about a new type or change the raw type name and implementation.

> Being able to pass the bytes to RawType and have THIS do the magic,...

So RawType or a coupled helper class is to now guess which image type from the bytes?  I'm having trouble following the logic in these comments.

> IMHO the current usage of that type will be very rare, so changing that five open source bindings should be okay.

So renaming RawType is an option?

Also interested to hear responses to resetnow's comments:

> I'm still in doubt regarding passing images encoded in base64 strings. I haven't seen any product or solution using this for large images — it's usually for small icons, CSS backgrounds and so on. Imagine passing a capture from an HD webcam through a mobile connection. It's just technically ugly solution.

And imagine holding half a dozen HD webcam images in server memory at once.  Is there some reason that these real concerns are not responded to?

And in case my earlier comment was not intentionally ignored before it became inaccessible:

If there is a potential need for moving generic MIME-typed blocks of data as item states, its name is probably fine as is. But I don't think it's a good idea to have item states that are untyped, generic blobs that don't have semantics for some type (like all the other state types -- OnOff, Percent, Decimal, Point, etc.).

Doing so turns the state into just data instead of an object, and there may also be more natural transport formats and protocols. Example: you can't stream a surveillance camera this way, and you obviously don't want to hold its full contents in memory in an item state. So the model currently used by RawType won't be able to provide practical usage of a range of other resource types. The range of relevant things that are commonly rendered by client components and are addressable via HTTP URL is very wide and won't all fit in a RawType.

Something like a URIType seems more appropriate, that can update a LinkItem, and LinkItems can be associated with various UI widgets; for example, show an image on every update, or send URIType commands to change radio stations, switch surveillance camera feeds, etc. This lets more natural protocols and renderers do all the work.

That is my thinking of why it should be renamed ImageType, or preferably, dropped as a state type and a more natural solution considered,

lolodomo:
lol
New standby for me.

My intention was to add an helper class helping to identify a mime type from a URL or a byte array. If this is still relevant, let me advice in what existing or new package I should create this class.

maggu2810:
> It depends. This "magic" might not be easy to implement and forcing every binding to implement it on their own might lead to duplicate code and is error-prone.

Why does everyone need to implement it itself.
I already suggested:

> we add (to some class, perhaps a new one) a static function that tries to identify the mime type (magic headers, ...), I assume there are implementations that could be used

So, the one that should fit the correct type could use that static function.
We could also use that function on other places that are not related to the RawType, e.g. to find the correct content type for web stuff etc.

maggu2810:
@watou The RawType implements the State interface. IIRC every State must be fully serializable to a String and deserializable fomr a String.
This "full string" needs to be used for events.
If this is correct, I don't think a HD WebCam should provide all its stuff by this type at all.
If we would like to support this in the core, perhaps there should be "something" that is allowed to emit events without the full string.

Also I don't think a State does fit to a Stream. There is a state at a current point in time, but a stream ...

> If there is a potential need for moving generic MIME-typed blocks of data as item states, its name is probably fine as is. But I don't think it's a good idea to have item states that are untyped, generic blobs that don't have semantics for some type (like all the other state types -- OnOff, Percent, Decimal, Point, etc.).

If we use a mime-type, then there souldn't be any untyped anymore. Or do I miss something?

> That is my thinking of why it should be renamed ImageType, or preferably, dropped as a state type and a more natural solution considered,

A RawType with a mime-type could not be used for images only, perhaps it is currently the only use case.
IIRC the number of types (and perhaps items) should be kept small and types should be choosen to be more generic and not only for use case.

maggu2810:
> My intention was to add an helper class helping to identify a mime type from a URL or a byte array. If this is still relevant, let me advice in what existing or new package I should create this class.

Let's wait for some comments.
IMHO such a utility class that could be used by "whatever" is a nice helper.
And if the mime type is a mandatory attribute for the RawType constructor, the "caller" have the choose to use what he already knows or use that helper class.

No consumer does need to handle a RawType with an unset mime-type anymore.

watou:
> If this is correct, I don't think a HD WebCam should provide all its stuff by this type at all.

Exactly.  And what is the arbitrary limit on image size that should be documented in the spec when RawType cannot be used, and what is the alternative?  RawType for blobs of binary data is of a completely different sort from all other States, and I still do not understand why ESH needs to invent a new way of handling images (or PDF files, to use an earlier example) since HTTP is already there for the REST API and servlets, and all the problems doing so creates (holding an unbounded number of them in server memory, serializing/deserializing in non-native formats, no persistence support (and why would you want to anyway), guessing the type (or failing to), having bindings read them from their sources even when no client is viewing them, and likely many more).  RawType (or whatever it's named) still lacks a basic justification to exist, IMHO.

kaikreuzer:
> Why does everyone need to implement it itself. I already suggested

Sorry, I had missed that. So let's go for a helper class (could the RawType could offer a mime-less constructor that calls this helper...?)

@watou ImageItem and RawType have been introduced 1-2 years ago and they are in use by ESH solutions. We are not discussing their existence here, we are only discussing the need to add some meta-data to it.

For webcam support etc., we have valid solutions in place for sitemap-based UIs using the Image and Video widgets and the proxy servlet. This can be used, so there is no size issue.
The ImageItem has some further benefits as its state is treated as a current snapshot just like for any other item, so it is possible to persist, to send it through MQTT, to pass it as an output in a rule to e.g. send it via SMTP or whatever. And as this exists, it is just a logical consequence to make it possible to use it in the Basic/Classic UI (while it is supported in Paper UI since 1-2 years already). You don't have to love and use that feature, but I think it is fair to be there as an option.

maggu2810:
@watou I don't get your point. Isn't it up to the binding developer what item type etc. is choosen?
If I want to provide full HD images to someone that can have a look at but don't want to use a state, it is possible. I will loose stuff like persistence integration etc. but if I would like to use it for my binding, I can do it.
I will use a ProxyServlet in my binding that implements a reverse proxy and offers the data I want to offer (so, on request only). I can e.g. use a string item that publishes the URL to my servlet (different location for different target, so also changed and update events are working).
If I don't want to use a ProxyServlet (e.g. because if it is a public URL), I can do so.
If I want to provide small thumbnails as state, so my setup can persist all the thumbnails, I can do so.

Why do you complain about the RawType at all? Perhaps it does not fit well for PDFs, etc. -- I agree.
But does the RawType to be used for all "bigger" stuff or should we perhaps create another type / solution for such a use case?

watou:
> ImageItem and RawType have been introduced 1-2 years ago and they are in use by ESH solutions. We are not discussing their existence here, we are only discussing the need to add some meta-data to it.

I am recommending that RawType be deprecated in preference to a solution that does not bring along all the problems it introduces, so existing solutions can be adapted to the better solution over time.  Doing so removes the need to further evolve RawType to try to overcome all of its practical limitations.

> Why do you complain about the RawType at all?

Because it is an uphill battle with many disadvantages for use for images and many other media types that will be needed in the future.

> But does the RawType to be used for all "bigger" stuff or should we perhaps create another type / solution for such a use case?

I earlier suggested URIType and LinkItem, with clients completely free to render URL-addressable resources in a natural way.  The proxy servlet can be made to serve all URL-addressable resources with full control over security concerns.

kaikreuzer:
Sorry, I cannot see how you would be able to cover the [above mentioned](https://github.com/eclipse/smarthome/pull/3055#issuecomment-282259925) use cases with URIType or LinkItem. I think deprecating RawType will be a bigger discussion, so I would suggest to open a separate issue for that.
This PR here is really just a small adaption to the existing code base, which can be deprecated with the whole type, if we decide so, because we come up with something better.

watou:
> The ImageItem has some further benefits as its state is treated as a current snapshot just like for any other item, so it is possible to persist, to send it through MQTT, to pass it as an output in a rule to e.g. send it via SMTP or whatever. And as this exists, it is just a logical consequence to make it possible to use it in the Basic/Classic UI (while it is supported in Paper UI since 1-2 years already). You don't have to love and use that feature, but I think it is fair to be there as an option.

I struggle to envision these use cases in practical usage, but I acknowledge there is at least the potential that they can be used that way.  I suppose that if all of RawType's major limitations were well documented as a caution for binding authors, and that binding authors have a formal alternative for exposing resources for clients to render in a natural way (in addition to the StringType/StringItem approach that you (IIRC) disfavor), then that would address my concerns.

lolodomo:
Yes, for image, all bindings can propose the 2 options, ImageItem vs URL in StringItem.

kaikreuzer:
> Yes, for image, all bindings can propose the 2 options, ImageItem vs URL in StringItem.

 Correct, and I think most bindings also do so. 

Btw, talking about cameras, I would like to see some new specific concepts being added for them anyhow, which should also be able to deal with streams, quality levels, transcoding, etc. Maybe something similar to the audio framework that has been added last year. But that's a topic for the future ;-)

lolodomo:
So finally, the idea would be tpo add an helper method that will try to extract the data type from the data buffer; and have this method called inside the RawType constructor when null is provided as argument ? Can I simply put this static method directly inside RawType ?

Additionally, any binding handling an image URL should first call getContentType() on URLConnection to try identifying a MIME type.

maggu2810:
I would prefer to have a separate class that provides that function. It is currently used by RawType but identifyMimeType (or whatever name is used) has itself no relationship to RawType.

I would prefer to use `new RawType(data, MagicClass.identifyMimeType(data));` instead of doing it itself in the RawType class. At least it makes no difference at runtime if there is a `RawType(byte[] data)` constructor that calls the identify mime type function itself, but without that function the caller is pointed to that it should give an mime type. The RawType shouldn't call the magic itself but should be a container for the data and mime type information.

lolodomo:
I have now analyzed when the image is not updated in Basic UI. The code in Basic UI has to be updated but the main problem I have is that the sitemap event received by UI has a state field built on server side using item.getState().toString(). For an image (RawType), that means a text like "raw type (unknown):1234 bytes".

Would it be a problem if I change toString in RawType to return the data URI (same as toFullString() ) ? This will add more data in few REST API...

The alternative I have is to update the code to provide the full string only in PageChangeListener:constructSitemapEvents. This will imply to enhance EnrichedItemDTO map method with a new parameter requesting normal string or full string.

lolodomo:
@maggu2810 : I don't have the code to identity data type from a data buffer !!!

But this code is available in UROConnection. So every binding getting an image from an URL will be able to use different features from UROConnection to identify the MIME type and then provide the result to new Rawtype(data, type).

I will update the sonos binding code with this so that a MIME type is provided to RawType.

I have no immediate solution for bindings only getting a data buffer. It looks like some libraries exist doing this job.

lolodomo:
To change the sitemap syntax for the image element and make url= optional, please confirm that I just have to update this file and this line: https://github.com/eclipse/smarthome/blob/master/bundles/model/org.eclipse.smarthome.model.sitemap/src/org/eclipse/smarthome/model/Sitemap.xtext#L46

resetnow:
```diff
+('url=' url=STRING)?
-('url=' url=STRING)
```

Should do the trick

lolodomo:
That was my feeling too.

lolodomo:
I have updated my PR (don't know why Travis CI build failed).
It seems to work very well, Paper UI, Basic UI and Classic UI.
I updated my PR comment to give more detais aboutl what was changed.
I updated the Sonos binding to search for the MIME type and it seems to work well too: it first look for tyhe content type in the header and if not found, search it from the input stream. So RawType constructor is called with a MIME type when found (in practice, apparently always).

extensions/ui/org.eclipse.smarthome.ui.basic/web-src/smarthome.js
```diff
@@ -405,10 +405,21 @@
 		_t.image = parentNode.querySelector("img");
 		_t.updateInterval = parseInt(parentNode.getAttribute("data-update-interval"), 10);
 
-		_t.url = _t.image.getAttribute("src").replace(/&t=\d+/g, "");
```
resetnow: There is a refresh attribute, that's why `t=` parameter is stripped.

extensions/ui/org.eclipse.smarthome.ui.basic/web-src/smarthome.js
```diff
@@ -405,10 +405,21 @@
...
+		_t.url = parentNode.getAttribute("data-proxied-url");
+		_t.validUrl = parentNode.getAttribute("data-valid-url");
 
-		_t.setValuePrivate = function() {
-			_t.image.setAttribute("src", _t.url + "&t=" + Date.now());
+		_t.setValuePrivate = function(value, itemState) {
+			if (itemState.startsWith("data:")) {
+				// Image element associated to an item of type ImageItem
+				_t.image.setAttribute("src", itemState);
+			} else if ((itemState !== "UNDEF") || (_t.validUrl === "true")) {
```
resetnow: ` || (_t.validUrl)`

extensions/ui/org.eclipse.smarthome.ui.basic/web-src/smarthome.js
```diff
@@ -405,10 +405,21 @@
...
-		_t.url = _t.image.getAttribute("src").replace(/&t=\d+/g, "");
+		_t.url = parentNode.getAttribute("data-proxied-url");
+		_t.validUrl = parentNode.getAttribute("data-valid-url");
 
-		_t.setValuePrivate = function() {
-			_t.image.setAttribute("src", _t.url + "&t=" + Date.now());
+		_t.setValuePrivate = function(value, itemState) {
+			if (itemState.startsWith("data:")) {
+				// Image element associated to an item of type ImageItem
+				_t.image.setAttribute("src", itemState);
```
watou: I think this should be verified on Safari to make sure it works.  My recollection on earlier PRs is that it did not work.

bundles/core/org.eclipse.smarthome.core/src/main/java/org/eclipse/smarthome/core/library/types/RawType.java
```diff
@@ -23,31 +23,51 @@
...
+        if (idx2 > 5) {
+            mimeType = value.substring(5, idx2);
+        }
+        return new RawType(Base64.getDecoder().decode(value.substring(idx + 1)), mimeType);
     }
 
     @Override
     public String toString() {
-        return String.format("raw type: %d bytes", bytes.length);
+        return String.format("raw type (%s): %d bytes", (mimeType != null) ? mimeType : "unknown", bytes.length);
```
watou: `unknown` is not a valid media type. [below](https://github.com/eclipse/smarthome/pull/3055/files#diff-7a4c8bb7df8b8ef0705510067c5561f8R69) you have an empty string when `mimeType` is `null`, which according to [this](https://en.wikipedia.org/wiki/Data_URI_scheme#Syntax) is the same as `text/plain`.

extensions/ui/org.eclipse.smarthome.ui.basic/web-src/smarthome.js
```diff
@@ -405,10 +405,21 @@
 		_t.image = parentNode.querySelector("img");
 		_t.updateInterval = parseInt(parentNode.getAttribute("data-update-interval"), 10);
 
-		_t.url = _t.image.getAttribute("src").replace(/&t=\d+/g, "");
+		_t.url = parentNode.getAttribute("data-proxied-url");
+		_t.validUrl = parentNode.getAttribute("data-valid-url");
```
resetnow: `parentNode.getAttribute("data-valid-url") === "true"`

bundles/core/org.eclipse.smarthome.core/src/main/java/org/eclipse/smarthome/core/library/types/RawType.java
lolodomo: toString() is a simplified string representation of the object.
The data URI is provided by toFullString and in this case it does not use "unknown".

extensions/ui/org.eclipse.smarthome.ui.basic/web-src/smarthome.js
lolodomo: Please test it, I have no Apple device.

extensions/ui/org.eclipse.smarthome.ui.basic/web-src/smarthome.js
lolodomo: The refresh attribute is not put in the new attribute, but it is added each time it is expected.

extensions/ui/org.eclipse.smarthome.ui.basic/web-src/smarthome.js
lolodomo: Why not, I can change that as you like.

extensions/ui/org.eclipse.smarthome.ui.basic/web-src/smarthome.js
lolodomo: But a least verified in Firefox and Chrome.

resetnow: Neither do many of us. Just in case we can't find anyone with Safari or you need to debug it, https://appetize.io/ — you can use iPhone on the front page for some amount of time, also there is some free time if you register an account

lolodomo: In case it would not work with Safari, someone will fix the code in the future.
At least, it works well now with Firefox and Chrome, that is already a big improvment.

extensions/ui/org.eclipse.smarthome.ui.basic/web-src/smarthome.js
lolodomo: @resetnow : if you really think that it is better to include the refresh attribute in this new data atribute on render servlet side and then suppress it here on client side, I can do it.
For me, that is better like I did, but no problem to change it.

resetnow: I think your way is actually better, thank you for explaining.

lolodomo:
Just a summary about RawType and new MIME type: we have now to recommend to each developper to provide when possible a MIME type when using RawType. It is what I did in the Sonos binding.
If no MIME type is provided, in practice it is not a problem for Basic UI and Classic UI when using Firefox or Chrome. The image is correctly displayed even without any MIME type in the data URI. But Paper UI requires this MIME type to display the image.

So there will be no problem with Basic UI and Classic UI for all existing bindings but a problem with Paper UI (except with the uodated Sonos binding).

So it will be better, as soon as this PR is merged, to update the 4 other OH 2 bindings using RawType, just due to problem with Paper UI.

lolodomo:
I think we need to keep null as accepted MIME type until someone is able to provide a **reliable** utility method that can identify the MIME type from a data buffer.

But I would suggest to already mark the RawType constructor without the MIME type as depreciated. Is it ok ?

extensions/ui/org.eclipse.smarthome.ui.basic/web-src/smarthome.js
lolodomo: Done

extensions/ui/org.eclipse.smarthome.ui.basic/web-src/smarthome.js
lolodomo: Done

kaikreuzer:
> But I would suggest to already mark the RawType constructor without the MIME type as depreciated. Is it ok ?

Makes sense to me, yes.
I'll also mark this PR as APIBreaking, because the serialization of the RawType changes through it - so any potential consumer (outside of our repo) will need to be adapted as well.

@aounhaider1 Could you please have a look at the changes on the Paper UI and confirm that this is fine? Would you possibly know a way to make the Paper UI render the images correctly without a mime-type as well?

maggu2810:
I would prefer to use `application/octet-stream`.
It seems to be the correct MIME type if no MIME type is known.

http://www.rfc-editor.org/rfc/rfc2046.txt

E.g.

This could be found in the RFC if you replace TYPE by image, audio, video, ...
```
Unrecognized subtypes of "TYPE" should at a miniumum be treated as
   "application/octet-stream".  Implementations may optionally elect to
   pass subtypes of "image" that they do not specifically recognize to a
   secure and robust general-purpose image viewing application, if such
   an application is available.
```

```
The "octet-stream" subtype is used to indicate that a body contains arbitrary binary data.
```

```
MIME implementations must in general treat unrecognized subtypes of "message" as being equivalent to "application/octet-stream".
```

extensions/ui/org.eclipse.smarthome.ui.basic/web-src/smarthome.js
watou: > someone will fix the code in the future

If Safari still doesn't allow setting the src to a data URI, it probably can't be fixed in the code, but I can't think of an acceptable workaround and obviously testing will reveal the answer.  C'est la vie! 😄

lolodomo:
To be checked if it is correctly supported by browsers.

lolodomo:
I have now included ImageItem in default sitemap. Works in Basic UI and Classic UI.

lolodomo:
I just tried with an hardcoded type "application/octet-stream" in the Sonos binding and it makes no difference; Chrome and Firefox are able to display PNG and JPEG images having this MIME type (Basic UI and Classic UI). It is still not working in Paper UI (as a remember, Paper UI is using "ng-src" img attribute while Basic UI and Classic UI are using the "src" img attribute).

I have no idea whether Safari makes a difference.
.
So if you think it is better to use "application/octet-stream" as default MIME type rather than no MIME type, that is ok for me.

maggu2810:
> So if you think it is better to use "application/octet-stream" as default MIME type rather than no MIME type, that is ok for me.

As we changed the meaning of RawType from data only to data + mime type, there should never be a "no MIME type" situation. Perhaps an "unknown" one and then we should follow the RFC and use MIME type correctly.

aounhaider1:
PaperUI should work fine as long as there is information about the content type available (data:image/*;base64) in the item state. We can add it in case it is missing in the state but I was unable to test it as setting the item state to base64 image without content-type information leads to image state being not properly set. The response in this case is: "data:;base64,", i.e., the image data it not returned at all.

lolodomo:
Ok, I have done the final change ("application/octet-stream" as default MIME type for RawType) and squashed everything.
I think that is ready for a merge.

bundles/core/org.eclipse.smarthome.core/src/main/java/org/eclipse/smarthome/core/library/types/RawType.java
```diff
@@ -18,36 +18,59 @@
...
 
+    public String getMimeType() {
+        return mimeType;
+    }
+
     public static RawType valueOf(String value) {
-        return new RawType(Base64.getDecoder().decode(value));
+        int idx;
+        if (value == null || !value.startsWith("data:") || ((idx = value.indexOf(",")) < 0)) {
+            return new RawType();
```
maggu2810: Same here, you cannot create a RawType of the given data, why is this accepted without throwing an exception?
Need to check other types. Is this our common handling to ignore a correct "deserialization"?

maggu2810: I would prefer to handle it similar to the point type and throw exception if invalid data is fit.

bundles/core/org.eclipse.smarthome.core/src/main/java/org/eclipse/smarthome/core/library/types/RawType.java
```diff
@@ -18,36 +18,59 @@
...
+        }
+        String mimeType = DEFAULT_MIME_TYPE;
+        int idx2 = value.indexOf(";");
+        if (idx2 > 5) {
+            mimeType = value.substring(5, idx2);
+        }
```
maggu2810: We defined that the mime type is mandatory, so the RawType full string representation contains a mime type all the time. I don't think we should accept missing mime type here, too.

lolodomo:
@maggu2810 : we want to keep compatibility with all existing bindings that don't provide a MIME type.
The failback to the default MIME type is for that reason.

maggu2810:
@lolodomo Thanks for your work.

> we want to keep compatibility with all existing bindings that don't provide a MIME type. The failback to the default MIME type is for that reason.

That's the reason we kept the old constructions and marked them deprecated.

My understanding (and the API break label) has been that the (de)serialization changed and so I don't think we need to support old formats there.

I also don't see the reason we need to handle null or empty mime type in the constructor that takes a mime type argument, because it is a new constructor that could be used correctly.

But perhaps I am missing something here, need to do further reading of the changes if there is more time. Just had a quick look at the changes.

lolodomo:
You're right, the compatibility is done with the deprecated constructor.
The advantage of accepting null as MIME type and setting it to "application/octet-stream" automatically is more a facility for developer. If we reject null (or empty string), all developers should know that they have to provide "application/octet-stream" when they don't know the MIME type. (or they can call the deprecated constructor).
But ok, I will change again !

lolodomo:
New change to make the MIME type mandatory in RawType !
I hope that is now ok and can switch to something else ;)

kaikreuzer:
> The review of the web UI code is better done by people with a better knowledge about that code.
> Just looked at the non UI related changes.

@resetnow Could you please do a quick final review as well?

extensions/binding/org.eclipse.smarthome.binding.sonos/src/main/java/org/eclipse/smarthome/binding/sonos/handler/ZonePlayerHandler.java
```diff
@@ -659,8 +660,16 @@ protected void updateChannel(String channeldD) {
...
+                        String contentType = connection.getContentType();
+                        InputStream input = connection.getInputStream();
+                        if (contentType == null) {
+                            contentType = URLConnection.guessContentTypeFromStream(input);
+                            if (contentType == null) {
+                                contentType = RawType.DEFAULT_MIME_TYPE;
+                            }
+                        }
+                        newState = new RawType(IOUtils.toByteArray(input), contentType);
                         IOUtils.closeQuietly(input);
```
kaikreuzer: I guess this line should actually to into a "finally" block

extensions/binding/org.eclipse.smarthome.binding.sonos/src/main/java/org/eclipse/smarthome/binding/sonos/handler/ZonePlayerHandler.java
```diff
@@ -659,8 +660,16 @@ protected void updateChannel(String channeldD) {
                 url = getAlbumArtUrl();
                 if (url != null) {
                     try {
-                        InputStream input = url.openStream();
-                        newState = new RawType(IOUtils.toByteArray(input));
+                        URLConnection connection = url.openConnection();
+                        String contentType = connection.getContentType();
+                        InputStream input = connection.getInputStream();
+                        if (contentType == null) {
+                            contentType = URLConnection.guessContentTypeFromStream(input);
```
kaikreuzer: This method requires "an input stream that supports marks" - otherwise, an exception will be thrown (because the stream has to be reset).
You will have to catch this exception and potentially (?) close and re-create the input stream.
Did you test whether this method works at all for media content?

kaikreuzer:
> I've reviewed BasicUI code before

Ok, thanks @resetnow!

extensions/binding/org.eclipse.smarthome.binding.sonos/src/main/java/org/eclipse/smarthome/binding/sonos/handler/ZonePlayerHandler.java
lolodomo: Tested and working well.

extensions/binding/org.eclipse.smarthome.binding.sonos/src/main/java/org/eclipse/smarthome/binding/sonos/handler/ZonePlayerHandler.java
maggu2810: You are right or try-with-resources...

lolodomo:
Sonos code enhanced with a separate method to identify the media type + a close of the input in a finally block.
Tell me if a squash is required.

lolodomo:
Note that the build failed in Hue binding tests, something I have of course not changed in my PR.

maggu2810:
@kaikreuzer Your requested changes has been implemented.
So, I assume you approve this PR now.

PR #3063, opened 2017-02-23
Extend the AudioSink infrastructure

kgoderis:
Extend the AudioSink infrastructure (ConsoleCommand, Rules) with the concept of “grouped” sinks that provide a multi-room audio functionality that goes beyond the boundary of a given hardware implementation (e.g. Sonos).

The Thread related code could be factored out if such a nee would arise. I have experimented with various Executor services, including the ones already implemented in the ESH core (e.g. SafeCall), but they proved not to be sufficient to let the AudioSinks play in a synchronised-lie way. Only the CachedThreadPool is fast enough to start up threads in sync, and thus play sounds correctly.

### DIFF
--- a/bundles/core/org.eclipse.smarthome.core.audio/src/main/java/org/eclipse/smarthome/core/audio/internal/AudioConsoleCommandExtension.java
+++ b/bundles/core/org.eclipse.smarthome.core.audio/src/main/java/org/eclipse/smarthome/core/audio/internal/AudioConsoleCommandExtension.java
@@ -26,6 +26,7 @@ import org.eclipse.smarthome.io.console.extensions.AbstractConsoleCommandExtensi
  */
 public class AudioConsoleCommandExtension extends AbstractConsoleCommandExtension {
 
+    private static final String SUBCMD_GROUPS = "group";
     private static final String SUBCMD_PLAY = "play";
     private static final String SUBCMD_STREAM = "stream";
     private static final String SUBCMD_SOURCES = "sources";
@@ -40,12 +41,16 @@ public class AudioConsoleCommandExtension extends AbstractConsoleCommandExtensio
     @Override
     public List<String> getUsages() {
         return Arrays.asList(new String[] {
-                buildCommandUsage(SUBCMD_PLAY + " <sink> <filename>",
+                buildCommandUsage(SUBCMD_PLAY + " <sink or group> <filename>",
                         "plays a sound file from the sounds folder through the optionally specified audio sink(s)"),
-                buildCommandUsage(SUBCMD_STREAM + " <sink> <url>",
+                buildCommandUsage(SUBCMD_STREAM + " <sink or group> <url>",
                         "streams the sound from the url through the optionally specified audio sink(s)"),
                 buildCommandUsage(SUBCMD_SOURCES, "lists the audio sources"),
-                buildCommandUsage(SUBCMD_SINKS, "lists the audio sinks") });
+                buildCommandUsage(SUBCMD_SINKS, "lists the audio sinks"),
+                buildCommandUsage(SUBCMD_GROUPS + " create <name>", "create a group of sinks"),
+                buildCommandUsage(SUBCMD_GROUPS + " clear <name>", "remove a group of sinks"),
+                buildCommandUsage(SUBCMD_GROUPS + " add <group> <sink>", "add a sink to a group of sinks"),
+                buildCommandUsage(SUBCMD_GROUPS + " remove <group> <sink>", "remove a sink to a group of sinks") });
 
     }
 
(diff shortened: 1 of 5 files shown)


kgoderis:
@kaikreuzer Not sure why the ip-validation if faling here....

maggu2810:
> Not sure why the ip-validation is faling here....

The only reason I could think about it that you are using `Signed-Off-By` instead of the common `Signed-off-by`. I don't know if it is checked case sensitive. But as the "signed off by" is a concept that is also added to the Git clients itself (e.g. the "-s" argument for commit) it wouldn't be surprising.

kaikreuzer:
Not sure if I overlook it, but is there any mechanism how to define and persist the groups?

Also, it appears to me as if a "group of sinks" are nothing else then a new type of (virtual) audio sink.
So I would assume that it is perfectly possible to put all the code into a separate bundle, which would nicely help on keeping clean code and having it structured by concerns. Wdyt?

kgoderis:
> Not sure if I overlook it, but is there any mechanism how to define and persist the groups?

define, yes. persist no. I'm not sure how the persistence bit should be dealt with

> 
> Also, it appears to me as if a "group of sinks" are nothing else then a new type of (virtual) audio sink.

Yes, they are, for the time being, since no specific logic has been added to make them behave differently

> So I would assume that it is perfectly possible to put all the code into a separate bundle, which would nicely help on keeping clean code and having it structured by concerns. Wdyt?

mmhm... I am not sure. I had to change some methods in the console command line extension for example

bundles/core/org.eclipse.smarthome.core.audio/src/main/java/org/eclipse/smarthome/core/audio/internal/AudioManagerImpl.java
```diff
@@ -81,18 +96,45 @@ public void play(AudioStream audioStream) {
...
+                            play(audioStream, aSink);
+                            logger.debug("Playing {} on {} completed in {} ms", audioStream.toString(), aSink,
+                                    System.currentTimeMillis() - stamp);
+                        }
+                    });
+                }
             } else {
-                logger.warn("Failed playing audio stream '{}' as no audio sink was found.", audioStream.toString());
+                logger.warn("Failed playing audio stream '{}' as no audio sink or group was found.",
+                        audioStream.toString());
```
maggu2810: That has not been introduced by this PR but as you changed that line already you could fix the implicit `toString()` function call.

bundles/core/org.eclipse.smarthome.core.audio/src/main/java/org/eclipse/smarthome/core/audio/internal/AudioConsoleCommandExtension.java
```diff
@@ -75,6 +80,8 @@ public void execute(String[] args, Console console) {
                 case SUBCMD_SINKS:
                     listSinks(console);
                     return;
+                case SUBCMD_GROUPS:
+                    handleGroups(args, console);
```
maggu2810: This case should contain a break / return / throw at the end regardless if the default case contains a logic or not.

bundles/core/org.eclipse.smarthome.core.audio/src/main/java/org/eclipse/smarthome/core/audio/internal/AudioConsoleCommandExtension.java
```diff
@@ -83,6 +90,60 @@ public void execute(String[] args, Console console) {
         }
     }
 
+    private static final String SUBCMD_CREATE = "create";
+    private static final String SUBCMD_REMOVE = "remove";
+    private static final String SUBCMD_ADD = "add";
+    private static final String SUBCMD_CLEAR = "clear";
```
maggu2810: If we keep that code into that file we should move that static finals to the other ones.
And IMHO we should also use that values in the buildCommandUsage, so if we change the SUBCMD constant, the documentation are changed to (similar to the SUBCMD_GROUPS etc.).
Perhaps the constants should be renamed from SUBCMD_xxx to SUBCMD_GROUP_xxx.

kaikreuzer:
> define, yes. persist no.

How does "define" work apart from using the console commands?

> mmhm... I am not sure. I had to change some methods in the console command line extension for example

The new bundle can itself add command line extensions.

kgoderis:
> define, yes. persist no.
> How does "define" work apart from using the console commands?
> 

There is support to use it in Rules as well.

> mmhm... I am not sure. I had to change some methods in the console command line extension for example
> The new bundle can itself add command line extensions.

ok. Only add, or also replace? In order to minimise the syntax I changed to audio sink related console commands so that they can cope with the groups as well, instead of copying over the console commands specifically for groups.

triller-telekom:
@kgoderis This Pr is quite old, but I am curios if you still plan to work on it. It would be nice to have groups of audio sinks. But I agree with @kaikreuzer that we need a way to make them persistent. if you can only define them through a console command and have to do it each time you restart your system this becomes nasty. Are you using this PR in your home setup? If so, how do you deal with the persistence problem?

Or is this PR just meant to be a proof of concept?

kgoderis:
@triller-telekom I do not use in my own setup as I found a work-around for my situation. I have no immediate plans to work on it either.

kaikreuzer:
Thanks for the feedback - so let's close this PR as it is not in a working/mergeable state.

PR #3064, opened 2017-02-23
allow ModuleHandlerFactories to dynamically add and remove supported moduletypes

smerschjohann:
Related to https://github.com/eclipse/smarthome/pull/3060 and https://github.com/eclipse/smarthome/pull/1783

Allows a ModuleHandlerFactory to support an additional type at runtime.


kaikreuzer:
Please note that there are compilation errors on org.eclipse.smarthome.automation.module.media reported by Travis.

smerschjohann:
Effectively it is a provider of handlers.. It "provides" a factory for given moduletypes -the supported moduletypes can change.
I can also define default methods for that interface if it is too cumbersome to write those empty listeners if you don't need them.

> A factory should always be able to produce handlers for module types that are newly added

Really? So it is really designed that way that for each new moduletype all modulehandlerfactories are asked again if they NOW support this moduletype? If that is the case, the provided type list of the ModuleHandlerFactory should at least be changed to a Set<String> or even HashSet<String>.

Nonetheless, the ModuleType and the ModuleTypeHandler are decoupled. Is it such a good idea to couple them in such an implicit way? What if someone declares a moduletype in JSON and implements the handler in a script language? He will never be able to update that handler again.

kaikreuzer:
> Effectively it is a provider of handlers.

Well, there isn't such a thing like a HandlerFactoryProvider and if possible, I would also avoid introducing it.

> So it is really designed that way that for each new moduletype all modulehandlerfactories are asked again if they NOW support this moduletype?

Absolutely. If the RuleEngine wants to initialize a rule, it asks all available handler factories to create a suitable handler for the modules.

> What if someone declares a moduletype in JSON and implements the handler in a script language? He will never be able to update that handler again.

Well, he would have to re-initialize his rule (e.g. by editing it or by restarting the system). Btw, this is the very same for a handler that is written in Java. I think it is much worse if you could change a handler implementation while a rule is currently running.

smerschjohann:
> Well, he would have to re-initialize his rule (e.g. by editing it or by restarting the system). Btw, this is the very same for a handler that is written in Java. I think it is much worse if you could change a handler implementation while a rule is currently running.

Why? All rules are paused beforehand. Why is it different if anything else from a rule changes and a handler?

> Absolutely. If the RuleEngine wants to initialize a rule, it asks all available handler factories to create a suitable handler for the modules.

This is inefficient O(nm) and not the way it is currently done. The RuleEngine "caches" which factories can provide the given type. If we either remove that and change the current `Collection<String> getTypes()` to a `boolean isSupported(String moduleTypeUID)` or change the getTypes to a `HashSet<String> getTypes()` it would at least be O(n).

smerschjohann:
And again. ModuleType and ModuleHandler are different things for the RuleEngine, they are lousily coupled. If you want something else, then the current implementation does not fit that requirement at all.

You are allowed to change the ModuleType at runtime, but you are not allowed to change the handler without pretending to change the ModuleType and by doing so let the RuleEngine inefficiently iterating over all modulehandlerfactories. 

This problem does not really apply for a static java implementations. Here the only possibility is the removal of the jar file which leads to deregistration of the factory which will result in suspension of all rules using those moduletypes. The replacement jar will automatically restart the rules. This can effectively lead to a "change [of] a handler implementation while a rule is currently running". So you point is not valid for the current implementation as well.

smerschjohann:
As it is possible to change the handler for moduletypes added through the script, I can live with that.

Nonetheless, it would be good to target the mentioned inefficiency but that would be another PR. closed.

Issue #2924, opened 2017-01-30
Implement ExtensionService for Eclipse IoT Marketplace


kaikreuzer:
The Eclipse Foundation is preparing an [IoT Marketplace](https://bugs.eclipse.org/bugs/show_bug.cgi?id=506792), which can host extensions specifically for ESH.
For this, we should have an ExtensionService implementation that can sync the content of the Marketplace and provide installation/uninstallation functionality.

Some remarks:
- The service should be configurable to allow filtering of the Marketplace content (e.g. regarding license (e.g. only open-source), quality level (only stable, no betas), etc.)
- https://github.com/eclipse/smarthome/issues/2922, https://github.com/eclipse/smarthome/issues/2920 and #2923 must be in place for it as well
- For the start, I think it is ok to sync the full (raw) content from the Marketplace (configurable, default e.g. once a day).
- If the content becomes too much for the Paper UI, a first useful measure will be to use pagination on the REST API - but this can be a later step.

cdjackson:
Will this allow multiple binding versions to be incorporated? Eg can we have a stable, nightly and bleeding edge version so users can decide? 

Below you say 'no betas' - I'm not sure if this means that you would only allow stable then?  If the IOT marketplace can support multiple versions then I think this would solve the issue that many people are asking on the OH forum about running some snapshot bindings with stable runtime.

kaikreuzer:
> Will this allow multiple binding versions to be incorporated? Eg can we have a stable, nightly and bleeding edge version so users can decide?

Yes.

> Below you say 'no betas' - I'm not sure if this means that you would only allow stable then?

This was just an example. What I wanted to say with that is that it should be fully up to the solution to decide, which content from the Marketplace should be made available.

cdjackson:
Great - I think this would be a very welcome addition.

kgoderis:
+1

lolodomo:
To be closed ?

PR #3076, opened 2017-02-26
Initial contribution of an IoT Market extension service

kaikreuzer:
implements #2924

WORK IN PROGRESS - not yet ready for merging!


maggu2810:
I assume the file `bundles/automation/org.eclipse.smarthome.automation.module.script.rulesupport/.project` should not be part of that PR.

kaikreuzer:
thx, removed it.

extensions/extensionservice/org.eclipse.smarthome.extensionservice.marketplace/build.properties
```diff
@@ -0,0 +1,8 @@
+output.. = target/classes
+bin.includes = META-INF/,\
+               .,\
+               OSGI-INF/,\
+               about.html,\
+               OSGI-INF/MarketplaceExtensionService.xml,\
+               OSGI-INF/MarketplaceTemplateProvider.xml
```
maggu2810: The two lines above could be removed, the whole directory is already included.

extensions/extensionservice/org.eclipse.smarthome.extensionservice.marketplace/pom.xml
```diff
@@ -0,0 +1,16 @@
...
+  <parent>
+    <groupId>org.eclipse.smarthome.extensionservice</groupId>
+    <artifactId>pom</artifactId>
+    <version>0.9.0-SNAPSHOT</version>
+  </parent>
+
+  <artifactId>org.eclipse.smarthome.extensionservice.marketplace</artifactId>
+  <packaging>eclipse-plugin</packaging>
+
+  <name>Eclipse SmartHome Web Audio Support</name>
```
maggu2810: I would expect another name 😉

extensions/extensionservice/org.eclipse.smarthome.extensionservice.marketplace/src/main/java/org/eclipse/smarthome/extensionservice/marketplace/internal/MarketplaceExtension.java
```diff
@@ -0,0 +1,37 @@
...
+public class MarketplaceExtension extends Extension {
+
+    public MarketplaceExtension(String id, String type, String label, String version, String link, boolean installed,
+            String description, String backgroundColor, String imageLink, String downloadUrl) {
+        super(id, type, label, version, link, installed, description, backgroundColor, imageLink);
+        this.downloadUrl = downloadUrl;
+    }
+
+    // we mark it as volatile, so that it isn't serialized through GSON
+    private volatile String downloadUrl;
```
maggu2810: Isn't transient the correct keyword to disable serialization?

extensions/extensionservice/org.eclipse.smarthome.extensionservice.marketplace/src/main/java/org/eclipse/smarthome/extensionservice/marketplace/internal/MarketplaceExtensionService.java
```diff
@@ -0,0 +1,236 @@
...
+                    } catch (ParsingException e) {
+                        postFailureEvent(extensionId, "Template is not valid.");
+                        logger.error("Rule template from marketplace is invalid: {}", e.getMessage());
+                    } catch (IOException e) {
+                        postFailureEvent(extensionId, "Template cannot be downloaded.");
+                        logger.error("Rule template from marketplace cannot be downloaded: {}", e.getMessage());
+                    }
+                    break;
+                default:
+                    postFailureEvent(extensionId, "Id not known.");
```
maggu2810: A break (also for the last statement) would be nice.

maggu2810: In `uninstall` you return, so no post(Un)installedEvent is called. Do you want to call `return` here, too?

kaikreuzer:
@maggu2810 Code has not been cleaned up yet, I mainly pushed it as WIP for testing the new Paper UI features. But thanks anyway, I'll incorporate your feedback (and will let you know once the PR is really ready for review :-))

kaikreuzer:
@maggu2810 Ready for review now - it is still a basic implementation that only covers certain use cases, but I think it is a good starting point.
I tried to make the code clean, fully functional and robust, so it is not just a PoC.

extensions/extensionservice/org.eclipse.smarthome.extensionservice.marketplace/src/main/java/org/eclipse/smarthome/extensionservice/marketplace/internal/MarketplaceExtensionService.java
```diff
@@ -0,0 +1,404 @@
...
+        }
+    }
+
+    private String getTemplate(String urlString) throws IOException {
+        URL url = new URL(urlString);
+        return IOUtils.toString(url);
+    }
+
+    private void postInstalledEvent(String extensionId) {
+        if (eventPublisher != null) {
```
maggu2810: I assume this function must only be called if the service is activated.
EventPublisher is a mandatory reference, so no need to check for nullness.

extensions/extensionservice/org.eclipse.smarthome.extensionservice.marketplace/src/main/java/org/eclipse/smarthome/extensionservice/marketplace/internal/MarketplaceExtensionService.java
```diff
@@ -0,0 +1,404 @@
...
+            Event event = ExtensionEventFactory.createExtensionInstalledEvent(extensionId);
+            eventPublisher.post(event);
+        }
+    }
+
+    private void postUninstalledEvent(String extensionId) {
+        if (eventPublisher != null) {
```
maggu2810: I assume this function must only be called if the service is activated.
EventPublisher is a mandatory reference, so no need to check for nullness.

extensions/extensionservice/org.eclipse.smarthome.extensionservice.marketplace/src/main/java/org/eclipse/smarthome/extensionservice/marketplace/internal/MarketplaceExtensionService.java
```diff
@@ -0,0 +1,404 @@
...
+            Event event = ExtensionEventFactory.createExtensionUninstalledEvent(extensionId);
+            eventPublisher.post(event);
+        }
+    }
+
+    private void postFailureEvent(String extensionId, String msg) {
+        if (eventPublisher != null) {
```
maggu2810: I assume this function must only be called if the service is activated.
EventPublisher is a mandatory reference, so no need to check for nullness.

extensions/extensionservice/org.eclipse.smarthome.extensionservice.marketplace/src/main/java/org/eclipse/smarthome/extensionservice/marketplace/internal/MarketplaceExtensionService.java
```diff
@@ -0,0 +1,404 @@
...
+            }
+        }
+    }
+
+    @Override
+    public void uninstall(String extensionId) {
+        Extension ext = getExtension(extensionId, null);
+        if (ext != null && ext.isInstalled()) {
+            String type = getType(extensionId);
+            switch (type) {
```
maggu2810: Isn't this a potential NPE as getType may return null.

extensions/extensionservice/org.eclipse.smarthome.extensionservice.marketplace/src/main/java/org/eclipse/smarthome/extensionservice/marketplace/internal/MarketplaceExtensionService.java
```diff
@@ -0,0 +1,404 @@
...
+        return extensionTypes;
+    }
+
+    @Override
+    public void install(String extensionId) {
+        Extension ext = getExtension(extensionId, null);
+        if (ext instanceof MarketplaceExtension && !ext.isInstalled()) {
+            MarketplaceExtension mpExt = (MarketplaceExtension) ext;
+            String type = getType(extensionId);
+            switch (type) {
```
maggu2810: Isn't this a potential NPE as getType may return null.

extensions/extensionservice/org.eclipse.smarthome.extensionservice.marketplace/src/main/java/org/eclipse/smarthome/extensionservice/marketplace/internal/MarketplaceExtensionService.java
```diff
@@ -0,0 +1,404 @@
...
+        if (dataFile != null && dataFile.exists()) {
+            try (FileReader reader = new FileReader(dataFile)) {
+                LineIterator lineIterator = IOUtils.lineIterator(reader);
+                Map<String, Long> map = new HashMap<>();
+                while (lineIterator.hasNext()) {
+                    String line = lineIterator.nextLine();
+                    String[] parts = line.split(";");
+                    try {
+                        map.put(parts[0], Long.valueOf(parts[1]));
+                    } catch (Exception e) {
```
maggu2810: You should check for `parts.length >= 2` before instead of execute the code and catch exceptions.
So, the only thing to catch is the NumberFormatException.

extensions/extensionservice/org.eclipse.smarthome.extensionservice.marketplace/src/main/java/org/eclipse/smarthome/extensionservice/marketplace/internal/MarketplaceRuleTemplateProvider.java
```diff
@@ -0,0 +1,100 @@
...
+    @Override
+    protected String keyToString(String key) {
+        return key;
+    }
+
+    protected void setParser(Parser<RuleTemplate> parser) {
+        this.parser = parser;
+    }
+
+    protected void unseteParser(Parser<RuleTemplate> parser) {
```
maggu2810: unset without an "e" at the end

extensions/extensionservice/org.eclipse.smarthome.extensionservice.marketplace/src/main/java/org/eclipse/smarthome/extensionservice/marketplace/internal/model/Marketplace.java
```diff
@@ -0,0 +1,21 @@
...
+ *
+ * @author Kai Kreuzer - Initial contribution and API
+ */
+public class Marketplace {
+
+    /**
+     * The category of the marketplace
+     */
+    public Category[] categories;
+}
```
maggu2810: We should add a new line at the end of the file.
Without the new line this will lead in noisy diffs in Git (and some quality tools also prefer new lines).
We should change the IDE to do it for us.

extensions/extensionservice/org.eclipse.smarthome.extensionservice.marketplace/src/main/java/org/eclipse/smarthome/extensionservice/marketplace/internal/model/Node.java
```diff
@@ -0,0 +1,56 @@
...
+    public String version;
+
+    public String supporturl;
+
+    public String packagetypes;
+
+    public String packageformat;
+
+    public String updateurl;
+}
```
maggu2810: We should add a new line at the end of the file.
Without the new line this will lead in noisy diffs in Git (and some quality tools also prefer new lines).
We should change the IDE to do it for us.

extensions/extensionservice/org.eclipse.smarthome.extensionservice.marketplace/META-INF/MANIFEST.MF
```diff
@@ -0,0 +1,24 @@
+Manifest-Version: 1.0
+Bundle-ManifestVersion: 2
+Bundle-Name: Eclipse SmartHome IoT Marketplace Extension Service
+Bundle-SymbolicName: org.eclipse.smarthome.extensionservice.marketplace
```
kaikreuzer: I agree, we should make sure to use the singleton flag

extensions/extensionservice/org.eclipse.smarthome.extensionservice.marketplace/src/main/java/org/eclipse/smarthome/extensionservice/marketplace/internal/MarketplaceExtensionService.java
kaikreuzer: good catch

kaikreuzer:
updated

kaikreuzer:
Please refrain from a final review for the moment - I just noticed that I need some refactoring to also make it possible to use this extension service, if no automation component is installed...

maggu2810:
Ah, okay.
My plan has been to do a checkout and tests at the weekend.
Will wait for a ping.

kaikreuzer:
Ping!
I think the additional time spent was worthwhile - the code is imho much better structured (split into different classes) and it is much easier to add support for additional formats&types through additional bundles.

features/karaf/esh-ext/src/main/feature/feature.xml
```diff
@@ -55,6 +55,17 @@
     <bundle>mvn:org.eclipse.smarthome.binding/org.eclipse.smarthome.binding.yahooweather/${project.version}</bundle>
   </feature>
 
+  <feature name="esh-extensionservice-marketplace" version="${project.version}">
+    <feature>esh-base</feature>
+    <bundle>mvn:org.eclipse.smarthome.extensionservice/org.eclipse.smarthome.extensionservice.marketplace/${project.version}</bundle>
+  </feature>
+
+  <feature name="esh-extensionservice-marketplace-automation" version="${project.version}">
```
maggu2810: Does the feature verification complete?
Shouldn't marketplace-automation depend on the marketplace feature?

kaikreuzer: We do not have feature validation in place at ESH. You are right, it should depend on the marketplace feature.

kaikreuzer: Updated. Any further review comments?

maggu2810: > We do not have feature validation in place at ESH.

You are wrong!

It has been present from the beginning.

```
commit 8484160501a8eb8fc62f84eef82b44bbf2050877
Date:   Mon Apr 18 10:24:48 2016 +0200
```

Yourself moved it from `features/karaf-verify` to `features/karaf/verify` at the beginning of 2017.

Perhaps you remember that we talked on introduction of the Karaf feature if the verifiction should be enabled at build time to be executed automatically or only be in place to be triggered manually.

It's just about the line `<!--<phase>process-resources</phase>-->` in the POM file, if we bind the goal to a phase or not.

To trigger the verification manually, you can do it as usual using this command:

```
mvn -pl features/karaf/verify org.apache.karaf.tooling:karaf-maven-plugin:verify@verify-esh
```

maggu2810: > Updated. Any further review comments?

Haven't had any time for review, just looked at the features yesterday.

extensions/extensionservice/org.eclipse.smarthome.extensionservice.marketplace.automation/src/main/java/org/eclipse/smarthome/extensionservice/marketplace/automation/internal/MarketplaceRuleTemplateProvider.java
```diff
@@ -0,0 +1,100 @@
...
+import org.eclipse.smarthome.automation.template.RuleTemplateProvider;
+import org.eclipse.smarthome.core.common.registry.DefaultAbstractManagedProvider;
+import org.slf4j.Logger;
+import org.slf4j.LoggerFactory;
+
+/**
+ * This class provides local access to the market place content. Once started, it downloads the catalog and then makes
+ * its content available from memory.
+ *
+ * Note that there is no progressive/lazy browsing implemented yet, but the service downloads the whole catalog.
```
maggu2810: Is this comment this correct?
This file does not download anything, does it?

extensions/extensionservice/org.eclipse.smarthome.extensionservice.marketplace/src/main/java/org/eclipse/smarthome/extensionservice/marketplace/internal/BindingExtensionHandler.java
```diff
@@ -0,0 +1,151 @@
...
+import org.osgi.framework.BundleException;
+import org.slf4j.Logger;
+import org.slf4j.LoggerFactory;
+
+/**
+ * A {@link MarketplaceExtensionHandler} implementation, which handles bindings as jar files (OSGi bundles) and installs
+ * them through the standard OSGi bundle installation mechanism.
+ * The information, which installed bundle corresponds to which extension is written to a file in the bundle's data
+ * store. It is therefore wiped together with the bundles upon an OSGi "clean".
+ * We might want to move this class into a separate bundle in future, when we add support for further extension types.
```
maggu2810: Wouldn't it be already a good time as we have a different one for the automation one?

extensions/extensionservice/org.eclipse.smarthome.extensionservice.marketplace/src/main/java/org/eclipse/smarthome/extensionservice/marketplace/MarketplaceExtensionHandler.java
```diff
@@ -0,0 +1,58 @@
...
+package org.eclipse.smarthome.extensionservice.marketplace;
+
+/**
+ * This interface can be implemented by services that want to register as handlers for specific marketplace extension
+ * types and formats.
+ * In a system there should always only be exactly one handler responsible for a given type+format combination. If
+ * multiple handers support it, it is undefined which one will be called.
+ * This mechanism allows solutions to add support for specific formats (e.g. Karaf features) that are not supported by
+ * ESH out of the box.
+ * It also allows to decide which extension types are made available at all.
```
maggu2810: As there is only one handler allowed per type+format and every solution is allowed to add its own type+format handlers, shouldn't the type and format be defined by the handler itself instead adding global constants to the `MarketplaceExtension`?
So, moving
* `EXT_TYPE_BINDING` and `EXT_FORMAT_BUNDLE` to `BindingExtensionHandler`
* `EXT_TYPE_RULE_TEMPLATE` and `EXT_FORMAT_JSON` to `AutomationExtensionHandler`

maggu2810: Ah, okay, see the usage below...

extensions/extensionservice/org.eclipse.smarthome.extensionservice.marketplace/src/main/java/org/eclipse/smarthome/extensionservice/marketplace/internal/MarketplaceExtensionService.java
```diff
@@ -0,0 +1,303 @@
...
+
+    private void postUninstalledEvent(String extensionId) {
+        Event event = ExtensionEventFactory.createExtensionUninstalledEvent(extensionId);
+        eventPublisher.post(event);
+    }
+
+    private void postFailureEvent(String extensionId, String msg) {
+        Event event = ExtensionEventFactory.createExtensionFailureEvent(extensionId, msg);
+        eventPublisher.post(event);
+
```
maggu2810: Empty line can be removed

extensions/extensionservice/org.eclipse.smarthome.extensionservice.marketplace/src/main/java/org/eclipse/smarthome/extensionservice/marketplace/internal/MarketplaceExtensionService.java
```diff
@@ -0,0 +1,303 @@
...
+            case MP_PACKAGETYPE_RULE_TEMPLATE:
+                sb.append(MarketplaceExtension.EXT_TYPE_RULE_TEMPLATE).append("-");
+                break;
+            case MP_PACKAGETYPE_BINDING:
+                sb.append(MarketplaceExtension.EXT_TYPE_BINDING).append("-");
+                break;
+            default:
+                return null;
+        }
+        sb.append(node.id.replaceAll("[^a-zA-Z0-9_]", ""));
```
maggu2810: Isn't the extension ID defined this way (REST): `[a-zA-Z_0-9-:]*`
You added a colon (`:`) above but here you remove colons.

extensions/extensionservice/org.eclipse.smarthome.extensionservice.marketplace.automation/src/main/java/org/eclipse/smarthome/extensionservice/marketplace/automation/internal/MarketplaceRuleTemplateProvider.java
maggu2810: Moved to `MarketplaceProxy`, correct.
Change comment please

extensions/extensionservice/org.eclipse.smarthome.extensionservice.marketplace/src/main/java/org/eclipse/smarthome/extensionservice/marketplace/internal/MarketplaceProxy.java
```diff
@@ -0,0 +1,89 @@
...
+    private long refresh_interval = 3600;
+    private long retry_delay = 60;
+
+    /**
+     * Creates a new instance, which immediately schedules a synchronization with the marketplace content.
+     */
+    public MarketplaceProxy() {
+        try {
+            url = new URL(MP_URL);
+            ThreadPoolManager.getScheduledPool("marketplace").scheduleWithFixedDelay(() -> refresh(), 0,
```
maggu2810: At which place is that pool cleaned up again?
`getScheduledPool(...)` creates a new if it is not known and put it to a map.
I would expect the MarketProxy should remove ti is it is removed from the system.
But this seems to be a general problem in the `ThreadPoolManager`, isn't it?
It adds new Pools to its map and never cleans it up. Miss I something?

maggu2810:
We can now install and uninstall a bundle.
After a bundle is installed its state should be `INSTALLED`.
Who is responsible to start and stop the bundle?
Make it sense to have a service to add and remove bundles if we don't provide a service to start and stop (at least start) the installed bundles?

extensions/extensionservice/org.eclipse.smarthome.extensionservice.marketplace/src/main/java/org/eclipse/smarthome/extensionservice/marketplace/internal/MarketplaceExtensionService.java
kaikreuzer: We are adding the "marketplace:" prefix for the extension ids and hence the part AFTER that prefix should not contain any additional colons. What is converted here is the node id and this is (from what I have seen so far) always a numerical number anyhow, so this line is only a safeguard for the future.

extensions/extensionservice/org.eclipse.smarthome.extensionservice.marketplace/src/main/java/org/eclipse/smarthome/extensionservice/marketplace/MarketplaceExtensionHandler.java
kaikreuzer: The types are statically defined in the IoT marketplace - so if the marketplace is extended, we should also maintain the new types here at a central place as constants.

extensions/extensionservice/org.eclipse.smarthome.extensionservice.marketplace/src/main/java/org/eclipse/smarthome/extensionservice/marketplace/internal/BindingExtensionHandler.java
kaikreuzer: If ok for you, I would do that as a follow up PR.

extensions/extensionservice/org.eclipse.smarthome.extensionservice.marketplace/src/main/java/org/eclipse/smarthome/extensionservice/marketplace/internal/MarketplaceProxy.java
kaikreuzer: You are right - I definitely missed to cancel the scheduled job.
And as we do not need a thread pool but only a single executor, I am not using ThreadPoolManager anymore now.

kaikreuzer:
> Who is responsible to start and stop the bundle?

Normally, the bindings will start by themselve as they have "activate immediately" DS components.
But for this service, I think it makes sense to start bundles automatically as the user wants to add a functionality (and not just populate a local cache). I have therefore added the bundle start to the code. If anyone does not like that, we can also make that a configuration option in future.

@maggu2810 I have pushed my changes, I hope this is ok now. Would be happy to have it merged soon to have an initial version to start testing. The market will only launch officially by the end of March, so we can improve things that come up by follow up PRs.

extensions/extensionservice/org.eclipse.smarthome.extensionservice.marketplace/src/main/java/org/eclipse/smarthome/extensionservice/marketplace/internal/MarketplaceExtensionService.java
maggu2810: > We are adding the "marketplace:" prefix for the extension ids and hence the part AFTER that prefix should not contain any additional colons.

Ah, okay.

maggu2810:
> Normally, the bindings will start by themselve as they have "activate immediately" DS components.

Really?
IIRC the order for start is bundle start, bundle activator, bundle DS. For stop, stop deactivate all DS, deactivate activator, stop bundle.
Activate immediately does IMHO state "the DS should be activated if not used by any other service, too".

Do you think about a lazy activation, that result in a bundle start if the DS inspect the services?

Would be great, if you could give me further details.

extensions/extensionservice/org.eclipse.smarthome.extensionservice.marketplace/src/main/java/org/eclipse/smarthome/extensionservice/marketplace/internal/MarketplaceProxy.java
```diff
@@ -0,0 +1,99 @@
...
+    private ScheduledFuture<?> refreshJob;
+
+    /**
+     * Creates a new instance, which immediately schedules a synchronization with the marketplace content.
+     */
+    public MarketplaceProxy() {
+        try {
+            url = new URL(MP_URL);
+            this.refreshJob = Executors.newSingleThreadScheduledExecutor().scheduleWithFixedDelay(() -> refresh(), 0,
+                    refresh_interval, TimeUnit.SECONDS);
```
maggu2810: IIRC you should call shutdown after you added the job (https://docs.oracle.com/javase/7/docs/api/java/util/concurrent/ExecutorService.html#shutdown()).
If you don't use it, the executor does not "know" about, that it does not accept any further jobs.

```Java
final ScheduledExecutorService executor = Executors.newSingleThreadScheduledExecutor();
this.refreshJob = executor.scheduleWithFixedDelay(() -> refresh(), 0, refresh_interval, TimeUnit.SECONDS);
executor.shutdown();
```

extensions/extensionservice/org.eclipse.smarthome.extensionservice.marketplace/src/main/java/org/eclipse/smarthome/extensionservice/marketplace/internal/BindingExtensionHandler.java
maggu2810: Okay for me if the "follow up PR" is following :wink:

kaikreuzer:
> Do you think about a lazy activation, that result in a bundle start if the DS inspect the services?

Yes, that's what I thought about, you are right - anyhow, I have changed the code to always do a bundle start.

extensions/extensionservice/org.eclipse.smarthome.extensionservice.marketplace/src/main/java/org/eclipse/smarthome/extensionservice/marketplace/internal/MarketplaceProxy.java
kaikreuzer: Ok, just pushed a commit to properly shut it down

kaikreuzer:
> Okay for me if the "follow up PR" is following

Yes, I will have to, because I actually plan a different implementation on openHAB side later, which will use felix fileinstall instead of directly installing the bundles. So no risk that this follow-up PR is forgotten :-)

Issue #3073, opened 2017-02-26
Test failure(s): NTP: communication error


maggu2810:
Sometimes I get this error on compilation:

```text
-------------------------------------------------------
 T E S T S
-------------------------------------------------------
Running org.eclipse.smarthome.binding.ntp.test.NtpOSGiTest
Tests run: 16, Failures: 1, Errors: 0, Skipped: 2, Time elapsed: 22.053 sec <<< FAILURE! - in org.eclipse.smarthome.binding.ntp.test.NtpOSGiTest
the dateTime channel is updated with the right time zone(org.eclipse.smarthome.binding.ntp.test.NtpOSGiTest) skipped
time is refreshed when string channel is linked(org.eclipse.smarthome.binding.ntp.test.NtpOSGiTest)  Time elapsed: 0.261 sec
if no time zone is set in the configuration, the dateTime channel is updated with the default one(org.eclipse.smarthome.binding.ntp.test.NtpOSGiTest)  Time elapsed: 0.141 sec
time is refreshed when dateTime channel is linked(org.eclipse.smarthome.binding.ntp.test.NtpOSGiTest)  Time elapsed: 0.045 sec
the status of a thing with string channel is updated with communication error, when unknown host is set in the configuraion(org.eclipse.smarthome.binding.ntp.test.NtpOSGiTest)  Time elapsed: 10.169 sec  <<< FAILURE!
java.lang.AssertionError: The thing status was not communication error
Expected: is <COMMUNICATION_ERROR>
     but: was <NONE>
	at org.hamcrest.MatcherAssert.assertThat(MatcherAssert.java:20)
	at org.junit.Assert.assertThat(Assert.java:865)
	at org.junit.Assert$assertThat$0.callStatic(Unknown Source)
	at org.eclipse.smarthome.binding.ntp.test.NtpOSGiTest$_assertCommunicationError_closure8.doCall(NtpOSGiTest.groovy:516)
	at org.eclipse.smarthome.binding.ntp.test.NtpOSGiTest$_assertCommunicationError_closure8.doCall(NtpOSGiTest.groovy)
	at sun.reflect.GeneratedMethodAccessor9.invoke(Unknown Source)
	at sun.reflect.DelegatingMethodAccessorImpl.invoke(DelegatingMethodAccessorImpl.java:43)
	at java.lang.reflect.Method.invoke(Method.java:498)
	at org.codehaus.groovy.reflection.CachedMethod.invoke(CachedMethod.java:90)
[...]

PR #2087, opened 2016-08-31
Refactoring bridge/thing life cycle.

sbussweiler:
- A `BridgeHandler` of a bridge is initialized before `ThingHandler`s of its child things are initialized.
- A `BridgeHandler` is disposed after all `ThingHandler`s of its child things are disposed.
- A bridge handler (new interface `BridgeHandler`) is notified if a child handler is initialized/disposed.
- The  methods `ThingHandler.bridgeHandlerInitialized()` and `ThingHandler.bridgeHandlerDisposed()` are removed. This **breaks the API**.
-  **OSGi service registration** of thing/bidge handlers has been removed. 

Signed-off-by: Stefan Bussweiler s.bussweiler@external.telekom.de

PR #3078, opened 2017-02-26
Improve pure Java OSGi test support

maggu2810:
Groovy and the Java compiler creates both bytecode that is run by the JVM.
So, you can access Groovy classes in your pure Java class implementation and the other way around.

The Groovy based testing approach all is done via reflection (at least it seems so to me).
Also it is not very strict to types, you can access private stuff etc.

I created a JavaOSGiTest class in a new package that functionality is equal to the OSGiTest Groovy class.
The Java source folder cannot use the same package as the groovy source folder, so I have chosen a sub-package (and the name for easier identifying).

I migrated the NTP Binding test to a pure Java one.
I identified the usage of "internal" stuff of another bundle (core.thing) that has been replaced by "correct" code (at least "public" accessible stuff).

IMHO as we can use lambdas now, it is also very comfortable to use pure Java to write tests. And we can make of of the support of the Java compiler to identify wrong usage of arguments, return types, accessible stuff, unused private functions / members, ... Also the refactoring and "find usage" will work (IMHO) better as there is no reflection used.

---

It seems there is one Eclipse IDE global workspace setting that is still set to compiler compliance 1.7 (or not updated in my setup) but it could be quick fixed...


maggu2810:
JavaDoc fixes done.

I kept the NTP changes as @petarnv is fine with using pure Java for the NTP tests (https://github.com/eclipse/smarthome/issues/3073#issuecomment-283053958).

So, ready for review.

maggu2810:
@petarnv Would you like to review the NTP test changes?

petarnv:
@maggu2810, the test changes look good to me. The only thing I would suggest is changing the test methods names. They look a bit long to me, e.g. `ifNoTimeZoneIsSetInTheConfigurationTheStringChannelIsUpdatedWithTheDefaultOne`. That was fine in groovy, as it was a whole sentence, but in pure Java it is hard to understand.

Also, for the internal stuff. I guess you mean the usage of `org.eclipse.smarthome.binding.ntp.internal.NtpHandlerFactory` and `org.eclipse.smarthome.core.thing.internal.ThingManager`. This was introduced in this PR - https://github.com/eclipse/smarthome/pull/2087. After this PR was merged some of the other tests are also using the same internal stuff. Should they be changed to pure Java, too?

maggu2810:
> only thing I would suggest is changing the test methods names. They look a bit long to me

I agree. I don't want to change your (?) naming of the test cases, so I only changed the previous method names to camel case.
Could you create a PR for my osgi-test branch with your preferred names?

> Also, for the internal stuff. I guess you mean the usage of `org.eclipse.smarthome.binding.ntp.internal.NtpHandlerFactory` and `org.eclipse.smarthome.core.thing.internal.ThingManager`

I don't really care about the NTP handler factory because in some cases it is okay that a test for bundle_a access internal stuff of bundle_a. It depends on the test itself.
It has been about internals of core.thing (as you already mentioned).
That has not been necessary at all because you can get the handle of a thing by a public thing method already (I assume you identified the changes already).

> Should they be changed to pure Java, too?

I don't think so.
**IMHO**: I would prefer using pure Java tests only, but it has been allowed to use Groovy and I don't think this will be changed (regardless it would be something **I** would prefer). That is just a personal view.

My intention was to create a base class similar to the Groovy one to make writing pure Java OSGi tests as easy as the Groovy ones.
The failing NTP test on my machine and because it has only be one file has been the reason I used that for my testing.

So, let's concentrate on moving the NTP tests to a pure Java one as an demonstration and a first use of the pure Java OSGi base test class.
We shouldn't change any unrelated stuff (no other bugfixes in the test, no other tests, ...) in that PR.

petarnv:
@maggu2810, I've made a PR to your branch: https://github.com/maggu2810/smarthome/pull/5

petarnv:
But now I see another problem - when I run the maven build, there are no NTP tests found:
```
-------------------------------------------------------
 T E S T S
-------------------------------------------------------

Results :

Tests run: 0, Failures: 0, Errors: 0, Skipped: 0

[WARNING] No tests found.
[INFO] 
[INFO] --- maven-install-plugin:2.3.1:install (default-install) @ org.eclipse.smarthome.binding.ntp.test ---
[INFO] Installing C:\dev\repos\smarthome\extensions\binding\org.eclipse.smarthome.binding.ntp.test\target\org.eclipse.smarthome.binding.ntp.test-0.9.0-SNAPSHOT.jar to C:\Users\petar.valchev\.m2\repository\org\eclipse\smarthome\binding\org.eclipse.smarthome.binding.ntp.test\0.9.0-SNAPSHOT\org.eclipse.smarthome.binding.ntp.test-0.9.0-SNAPSHOT.jar
[INFO] Installing C:\dev\repos\smarthome\extensions\binding\org.eclipse.smarthome.binding.ntp.test\pom.xml to C:\Users\petar.valchev\.m2\repository\org\eclipse\smarthome\binding\org.eclipse.smarthome.binding.ntp.test\0.9.0-SNAPSHOT\org.eclipse.smarthome.binding.ntp.test-0.9.0-SNAPSHOT.pom
[INFO] Installing C:\dev\repos\smarthome\extensions\binding\org.eclipse.smarthome.binding.ntp.test\target\org.eclipse.smarthome.binding.ntp.test-0.9.0-SNAPSHOT-sources.jar to C:\Users\petar.valchev\.m2\repository\org\eclipse\smarthome\binding\org.eclipse.smarthome.binding.ntp.test\0.9.0-SNAPSHOT\org.eclipse.smarthome.binding.ntp.test-0.9.0-SNAPSHOT-sources.jar
[INFO] Installing C:\dev\repos\smarthome\extensions\binding\org.eclipse.smarthome.binding.ntp.test\target\p2content.xml to C:\Users\petar.valchev\.m2\repository\org\eclipse\smarthome\binding\org.eclipse.smarthome.binding.ntp.test\0.9.0-SNAPSHOT\org.eclipse.smarthome.binding.ntp.test-0.9.0-SNAPSHOT-p2metadata.xml
[INFO] Installing C:\dev\repos\smarthome\extensions\binding\org.eclipse.smarthome.binding.ntp.test\target\p2artifacts.xml to C:\Users\petar.valchev\.m2\repository\org\eclipse\smarthome\binding\org.eclipse.smarthome.binding.ntp.test\0.9.0-SNAPSHOT\org.eclipse.smarthome.binding.ntp.test-0.9.0-SNAPSHOT-p2artifacts.xml
[INFO] 
[INFO] --- tycho-p2-plugin:1.0.0:update-local-index (default-update-local-index) @ org.eclipse.smarthome.binding.ntp.test ---
[INFO]               
```                                                       
The problem is not with the changed names. I tried running it with the old names and the tests are still not found. I run them on Windows.
@maggu2810, Any idea what the problem is?

maggu2810:
> when I run the maven build, there are no NTP tests found

Whoops!
I will have a look at.

maggu2810:
Should be fixed now.
As we are using Tycho and not a plain Maven build, Eclipse IDE settings are respected on Maven build...

kaikreuzer:
Thanks @maggu2810, I very much like that as I also prefer Java over Groovy. I think with Java 8 and closures, some benefits of Groovy are indeed not that relevant anymore.

kaikreuzer:
FTR: I did a rebase merge, so that @petarnv's commit is kept in the history as well.

PR #3082, opened 2017-02-27
make auto-update configurable: support config by channel UID and binding ID

maggu2810:
Fixes: #595

As I worked on that implementation I identified an unstable test in the upstream code base. A PR has been created and should be merged in front of this one (#3080).

---

Changes that are not really related to this PR. The cleanups have been done "on the fly":

* Remove old bundle activator that only use case is to log a message on activation / deactivation. Such stuff should be used for debugging only.
* Dynamic references should be volatile
* Don't build strings in logging statements

Changes purely related to this PR:

* Create a new interface that is provided by auto-update implementation.
* The new interface offers two function to add and remove a configuration to enable / disable auto-update for a whole binding (by Binding ID)
* The new interface offers two function to add and remove a configuration to enable / disable auto-update for channels (by Channel UID)

For performance reasons and an easy to use implementation I prefer to provide a service.
If every binding that wants to disable / force-enable auto-update has to provide a service, we end up with services for configuration and I don't think it will perform.

The interface name has been chosen to identify an configuration interface for the auto update service. So, we can consider to add other configuration functions in the future (if we ever need one) without breaking an API or adding further services etc.

---

To disable or force enable the auto-update in a specific bundle, you only need a reference to the AutoUpdateConfigurator.

This could easily be injected to the ThingFactory because this one implements already an OSGi service and I think we agree that every binding has one. If there is really a binding without a thing factory, place it at some other location.

If you want to disable the auto-update for he whole binding the only thing to do is to call e.g. in the thing factory activate method:

```Java
autoUpdateConfigurator.addAutoUpdateByBindingIdConfig(Constants.BINDING_ID, false);
```

and to remove the configuration again e.g. on the thing factory deactivate method:

```Java
autoUpdateConfigurator.removeAutoUpdateByBindingIdConfig(Constants.BINDING_ID);
```

If you want to disable auto-update for a given Channel, you can use e.g. this one:

```Java
autoUpdateConfigurator.addAutoUpdateByChannelUidConfig(Constants.MAGIC_CHANNEL_UID, false);
```


maggu2810:
We are waiting for another solution to disable the auto-update for channels:
https://github.com/eclipse/smarthome/issues/595#issuecomment-283066201

Issue #2357, opened 2016-10-25
invalid point to ItemThingLinkRegistry


maggu2810:
The JavaDoc of the public non-static method `ItemChannelLinkRegistry.getBoundThings(String itemName)` statues that the method is deprecated and the method `ItemThingLinkRegistry#getLinkedThings(String)` should be used instead.

https://github.com/eclipse/smarthome/blob/d87b1532643a1fbcdebe338e934db8b8c955507c/bundles/core/org.eclipse.smarthome.core.thing/src/main/java/org/eclipse/smarthome/core/thing/link/ItemChannelLinkRegistry.java#L83

I wasn't aware of such a ItemThingLinkRegistry.

Also the `LinkEventOSGiTest.groovy` announces it adds also tests for `ItemThingLinkRegistry`.

For the tests it is easy, let's remove the comment that it adds test for that classes, too.

How should we change the comment of the `getBoundThings` method?
In the ESH source tree this method is called in tests of `ItemChannelLinkOSGiTest` only.

kaikreuzer:
As the JavaDocs says "Will be removed soon", we might simply remove it now?
The idea of having an item for a Thing in order to be able to define a label is deprecated, since the Thing has a label property itself by now - so I would think the method is obsolete.

PR #3084, opened 2017-02-27
remove wrong deprecation and ItemThingLinkRegistry references

maggu2810:
Fixes: https://github.com/eclipse/smarthome/issues/2357

### DIFF
--- a/bundles/core/org.eclipse.smarthome.core.thing/src/main/java/org/eclipse/smarthome/core/thing/link/ItemChannelLinkRegistry.java
+++ b/bundles/core/org.eclipse.smarthome.core.thing/src/main/java/org/eclipse/smarthome/core/thing/link/ItemChannelLinkRegistry.java
@@ -41,8 +41,7 @@ public class ItemChannelLinkRegistry extends AbstractLinkRegistry<ItemChannelLin
     /**
      * Returns a set of bound channels for the given item name.
      *
-     * @param itemName
-     *            item name
+     * @param itemName item name
      * @return set of bound channels for the given item name
      */
     public Set<ChannelUID> getBoundChannels(String itemName) {
@@ -93,15 +92,10 @@ public class ItemChannelLinkRegistry extends AbstractLinkRegistry<ItemChannelLin
     /**
      * Returns a set of bound things for the given item name.
      *
-     * @deprecated Will be removed soon. Use {@link ItemThingLinkRegistry#getLinkedThings(String)} instead.
-     *
-     * @param itemName
-     *            item name
+     * @param itemName item name
      * @return set of bound things for the given item name
      */
-    @Deprecated
     public Set<Thing> getBoundThings(String itemName) {
-
         Set<Thing> things = new HashSet<>();
         Collection<ChannelUID> boundChannels = getBoundChannels(itemName);
 
--- a/bundles/core/org.eclipse.smarthome.core.thing.test/src/test/groovy/org/eclipse/smarthome/core/thing/link/LinkEventOSGiTest.groovy
+++ b/bundles/core/org.eclipse.smarthome.core.thing.test/src/test/groovy/org/eclipse/smarthome/core/thing/link/LinkEventOSGiTest.groovy
@@ -22,7 +22,7 @@ import org.junit.Test
 import com.google.common.collect.Sets
 
 /**
- * Event Tests for {@link ItemChannelLinkRegistry} and {@link ItemThingLinkRegistry}.
+ * Event Tests for {@link ItemChannelLinkRegistry}.
  *
  * @author Dennis Nobel - Initial contribution
  */


afuechsel:
Please do not (yet) merge this PR, we need to replace the removed method beforehand.

bundles/core/org.eclipse.smarthome.core.thing/src/main/java/org/eclipse/smarthome/core/thing/link/ItemChannelLinkRegistry.java
```diff
@@ -90,31 +87,6 @@ public ItemChannelLink update(ItemChannelLink element) {
...
-    /**
-     * Returns a set of bound things for the given item name.
-     *
-     * @deprecated Will be removed soon. Use {@link ItemThingLinkRegistry#getLinkedThings(String)} instead.
-     *
-     * @param itemName
-     *            item name
-     * @return set of bound things for the given item name
-     */
-    @Deprecated
```
afuechsel: I think this is an error. ItemThingLinkRegistry#getLinkedThings(String) is not a replacement for this method. ItemThingLinkRegistry#getLinkedThings(String) returns a thing, that is linked to an item directly, like the old items that were previously created for each thing. But this method here returns a collection of things, that are connected via channels to this specific item.

For example, getBoundThings("hue_0210_001788141f1a_1_color") would return the thing UID ["hue:0210:001788141f1a:1"], but ItemThingLinkRegistry#getLinkedThings("hue_0210_001788141f1a_1_color") would return []. 

Only ItemThingLinkRegistry#getLinkedThings("hue_0210_001788141f1a_1") would return ["hue:0210:001788141f1a:1"] (requesting using the old ThingItem).

afuechsel: Therefore I would leave this method in here and remove the @Deprecated marker, which is wrong.

bundles/core/org.eclipse.smarthome.core.thing.test/src/test/groovy/org/eclipse/smarthome/core/thing/link/ItemChannelLinkOSGiTest.groovy
```diff
@@ -96,18 +96,4 @@ class ItemChannelLinkOSGiTest extends OSGiTest {
         def boundThings = itemChannelLinkRegistry.getBoundChannels("notExistingItem")
         assertThat boundThings.isEmpty(), is(true)
     }
-
-    @Test
```
afuechsel: Not to be removed. See my comment on the tested method.

bundles/core/org.eclipse.smarthome.core.thing/src/main/java/org/eclipse/smarthome/core/thing/link/ItemChannelLinkRegistry.java
maggu2810: You are right, the comment is not correct, but isn't that exactly what we talked about in the referenced issue (this was my initial question)?


The function is deprecated and now removed because there shouldn't be a bound thing for an item anymore (see Kai's response in the referenced issue).

afuechsel: Discussed with Kai already.  I think, this deprecation is an error. This method does nothing have to do with the ItemChannelLinkRegistry.

maggu2810: > Discussed with Kai already. I think, this deprecation is an error.

We still support item thing links?

> This method does nothing have to do with the ItemChannelLinkRegistry

If it is the wrong place and we support item thing links, then the function still needs to be removed from ItemChannelLinkRegistry... and you need to create an ItemChannelLink... or whatever.

kaikreuzer: > We still support item thing links?

No, we don't. This method is actually a shortcut for getting all linked channels and getting the set of things that these channels belong to. I personally don't have any use case for such a method, but I also wouldn't mind too much to keep it in place, if someone really needs it.

maggu2810: > No, this is not a link to things! It's only a method to find out, to which things an item belongs to (is connected to via channels). All the assumptions behind the deprecation are wrong.

@afuechsel updated

Issue #3051, opened 2017-02-22
Switch Groups changing to UNDEF


lopez1de:
With OH2 GA (might also occur with other versions) I do face an issue with Group:Switch:OR going to "UNDEF" instead of ON or OFF.

I do own some motion sensors (Homematic HM-Sen-MDIR-O-2) and "itemized" them as follows:

```
Group:Switch:OR(ON,OFF) gMotion "Motion detected"
Switch M1 "Motion Sensor #1" <motion> (gMotion) { channel="homematic:..." }
Switch M2 "Motion Sensor #2 " <motion> (gMotion)  { channel="homematic:..." }
```
However, the group's state only is ON or OFF if both sensors have the same state. If their state's differ, the group reports itself as UNDEF.

On motion, the event.log prints

```
[GroupItemStateChangedEvent] - gMotion changed from OFF to UNDEF through M1
[ItemStateChangedEvent     ] - M1 changed from OFF to ON
```
later:

```
[ItemStateChangedEvent     ] - M1 changed from ON to OFF
[GroupItemStateChangedEvent] - gMotion changed from UNDEF to OFF through M1
Source of Issue (issue reported by m0wlheldChristoph Dahlen):
https://community.openhab.org/t/issue-with-group-switch-or-going-to-undef/21403
```

My findings:
If all switches are OFF, the group is OFF
If only one or several are ON, then the group is UNDEF
If all are ON, the group is ON

###################
I checked another group of 30 switches. All have changed from NULL to OFF after starting OH. The group then switches from NULL to OFF too, when the last item of the group switches to OFF.
`gStickyUnreach changed from NULL to OFF through Win_Kind1_StickyUnreach`

So far so good.

But when I change one switch to ON, the group goes UNDEF for some reason.
```
2017-02-18 11:34:46.020 [ItemStateChangedEvent     ] - Heiz_WohnZ_StickyUnreach changed from OFF to ON
2017-02-18 11:34:46.023 [GroupItemStateChangedEvent] - gStickyUnreach changed from OFF to UNDEF through Heiz_WohnZ_StickyUnreach
```

kaikreuzer:
Clearly sounds like a bug to me (some regression as this definitely worked before) - thanks for reporting.

lopez1de:
[...]

PR #3101, opened 2017-03-02
Show channel description on long press

aounhaider1:
fixes issue reported here: https://github.com/eclipse/smarthome/issues/3051#issuecomment-283314125

### DIFF
--- a/extensions/ui/org.eclipse.smarthome.ui.paper/web-src/js/controllers.configuration.js
+++ b/extensions/ui/org.eclipse.smarthome.ui.paper/web-src/js/controllers.configuration.js
@@ -575,6 +575,20 @@ angular.module('PaperUI.controllers.configuration', [ 'PaperUI.constants' ]).con
     $scope.hasProperties = function(properties) {
         return util.hasProperties(properties);
     }
+    
+    $scope.showDescription=function(channel,channelType){
+        var popup = $mdDialog.alert({
+            title: channel.label?channel.label:channel.channelType?channel.channelType.label:channel.id,
+            textContent: channel.description?channel.description:channel.channelType?channel.channelType.description:'No description available',
+            ok: 'Close'
+          });
+
+          $mdDialog
+            .show( popup )
+            .finally(function() {
+              alert = undefined;
+            });
+    }
 
     $scope.$watch('thing.channels', function() {
         $scope.refreshChannels($scope.showAdvanced);
--- a/extensions/ui/org.eclipse.smarthome.ui.paper/web-src/js/controllers.configuration.js
+++ b/extensions/ui/org.eclipse.smarthome.ui.paper/web-src/js/controllers.configuration.js
@@ -575,19 +575,19 @@ angular.module('PaperUI.controllers.configuration', [ 'PaperUI.constants' ]).con
     $scope.hasProperties = function(properties) {
         return util.hasProperties(properties);
     }
-    
-    $scope.showDescription=function(channel,channelType){
-        var popup = $mdDialog.alert({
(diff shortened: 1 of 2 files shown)


extensions/ui/org.eclipse.smarthome.ui.paper/web-src/js/controllers.configuration.js
```diff
@@ -575,6 +575,20 @@ angular.module('PaperUI.controllers.configuration', [ 'PaperUI.constants' ]).con
     $scope.hasProperties = function(properties) {
         return util.hasProperties(properties);
     }
+    
+    $scope.showDescription=function(channel,channelType){
+        var popup = $mdDialog.alert({
+            title: channel.label?channel.label:channel.channelType?channel.channelType.label:channel.id,
+            textContent: channel.description?channel.description:channel.channelType?channel.channelType.description:'No description available',
```
kaikreuzer: Instead of showing "No description available", might it make more sense to only offer the long press on channels where there is a description available?

extensions/ui/org.eclipse.smarthome.ui.paper/web-src/js/controllers.configuration.js
```diff
@@ -575,6 +575,20 @@ angular.module('PaperUI.controllers.configuration', [ 'PaperUI.constants' ]).con
     $scope.hasProperties = function(properties) {
         return util.hasProperties(properties);
     }
+    
+    $scope.showDescription=function(channel,channelType){
```
kaikreuzer: add a space before the curly bracket?

extensions/ui/org.eclipse.smarthome.ui.paper/web-src/js/controllers.configuration.js
aounhaider1: Can be, but then user might keep on trying to open the dialog and think its broken?

kaikreuzer: Might be. But then again, the description is localized and thus should be the "no description" message, if possible (which it isn't ;-)).
At many other places, a long press does nothing, so I think it is an expected situation that "nothing happens" means "nothing available" and not "all broken" ;-)

aounhaider1:
Updated.

extensions/ui/org.eclipse.smarthome.ui.paper/web-src/partials/configuration.html
```diff
@@ -154,14 +154,12 @@ <h2 class="inline" ng-show="thing.channels.length > 0">Channels</h2>
...
-									<h3 ng-if="channel.label">{{channel.label}}</h3>
-									<h3 ng-if="!channel.label && channel.channelType.label">{{channel.channelType.label}}</h3>
+									<h3 long-press on-long-press="showDescription(channel)" ng-if="channel.label || channel.channelType.label">{{channel.label?channel.label:channel.channelType.label}}</h3>
 									<h3 ng-if="!channel.label && !channel.channelType.label">{{channel.id}}</h3>
 									<p>
 										{{thing.UID + ':' + channel.id}}<i ng-show="advancedMode" class="material-icons copy-clipboard" copyclipboard="{{thing.UID + ':' + channel.id}}">content_copy</i>
 									</p>
-									<p ng-if="channel.description" ng-bind-html="channel.description"></p>
-									<p ng-if="!channel.description && channel.channelType.description" ng-bind-html="channel.channelType.description"></p>
+                                    <p>{{channel.itemType}}</p>
```
kaikreuzer: something seems to be wrong with the indentation here.

Issue #3102, opened 2017-03-02
ConfigParam MIN and MAX value not compated properly


ivivanov-bg:
When a config parameter is of type decimal and has `min` and `max` value specified, the comparison is done in `org.eclipse.smarthome.config.core.validation.internal.TypeIntrospections.FloatIntrospection` as follows:

```java
@Override
boolean isMaxViolated(Object value, int max) {
    if (isBigDecimalInstance(value)) {
           return Float.compare(((BigDecimal) value).floatValue(), max) > 0;
    }
    return Float.compare((float) value, max) > 0;
}
```

The signature of the method expects the MAX value to be integer which causes precision issues.

maggu2810:
Let's assume #3103 fixes your issue. Reopen is it doesn't fix it.

PR #3103, opened 2017-03-02
Change max / min value types to Object (for better handling of floats)

ivivanov-bg:
(possible fix for #3102)

### DIFF
--- a/bundles/config/org.eclipse.smarthome.config.core/src/main/java/org/eclipse/smarthome/config/core/validation/internal/TypeIntrospections.java
+++ b/bundles/config/org.eclipse.smarthome.config.core/src/main/java/org/eclipse/smarthome/config/core/validation/internal/TypeIntrospections.java
@@ -76,7 +76,18 @@ final class TypeIntrospections {
          *
          * @return true, if the given value is less than the given min attribute, otherwise false
          */
-        abstract boolean isMinViolated(Object value, int min);
+        boolean isMinViolated(Object value, BigDecimal min) {
+            if (min == null) {
+                return false;
+            }
+            final BigDecimal bd;
+            if (isBigDecimalInstance(value)) {
+                bd = (BigDecimal) value;
+            } else {
+                bd = new BigDecimal(value.toString());
+            }
+            return bd.compareTo(min) < 0;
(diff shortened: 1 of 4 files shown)


kaikreuzer:
Could you provide some unit tests with it that actually fail on the old code and succeed with this fix?

ivivanov-bg:
done

maggu2810:
Wouldn't it be much simpler to create `BigDecimal` objects and use the `compareTo` method?

ivivanov-bg:
You mean for the tests ?

maggu2810:
No, I mean the min and max type of the method arguments in the TypeIntrospection class

ivivanov-bg:
Done

bundles/config/org.eclipse.smarthome.config.core/src/main/java/org/eclipse/smarthome/config/core/validation/internal/TypeIntrospections.java
```diff
@@ -161,19 +161,26 @@ boolean isAssignable(Object value) {
...
+        boolean isMinViolated(Object value, BigDecimal min) {
+            if (min == null) {
+                return false;
+            }
             if (isBigDecimalInstance(value)) {
-                return Float.compare(((BigDecimal) value).floatValue(), min) < 0;
+                return ((BigDecimal) value).compareTo(min) < 0;
             }
-            return Float.compare((float) value, min) < 0;
+            return Float.compare((float) value, Float.valueOf(min.toString())) < 0;
```
maggu2810: @ivivanov-bg WDYT about

```Java
final BigDecimal bd;
if (isBigDecimalInstance(value)) {
    bd = (BigDecimal) value;
} else {
    bd = new BigDecimal(value.toString());
}
return bd.compareTo(min) < 0;
```

at least it does not make a big difference. Personally I assume it could be simpler to read.

WDYT?

ivivanov-bg: I like it

Issue #3105, opened 2017-03-02
Automation Rest ModuleTypeResource does not return children of the Composites


adimova:
When request GET {CompositeModuleTypeResourceID} module type, the children are not returned in response. The reason is a missing implementation for Composite Module Type DTOs

PR #3106, opened 2017-03-02
Extends Module Type DTOs with Composites

adimova:

Solves #3105

### DIFF
--- a/bundles/automation/org.eclipse.smarthome.automation.api/src/main/java/org/eclipse/smarthome/automation/dto/ActionTypeDTOMapper.java
+++ b/bundles/automation/org.eclipse.smarthome.automation.api/src/main/java/org/eclipse/smarthome/automation/dto/ActionTypeDTOMapper.java
@@ -13,19 +13,23 @@ import java.util.Collection;
 import java.util.List;
 
 import org.eclipse.smarthome.automation.type.ActionType;
+import org.eclipse.smarthome.automation.type.CompositeActionType;
 
 /**
  * This is a utility class to convert between the respective object and its DTO.
  *
  * @author Markus Rathgeb - Initial contribution and API
+ * @author Ana Dimova - extends Action Module type DTOs with composites
  */
 public class ActionTypeDTOMapper extends ModuleTypeDTOMapper {
 
     public static ActionTypeDTO map(final ActionType actionType) {
-        final ActionTypeDTO actionTypeDto = new ActionTypeDTO();
-        fillProperties(actionType, actionTypeDto);
-        actionTypeDto.inputs = actionType.getInputs();
-        actionTypeDto.outputs = actionType.getOutputs();
+        return map(actionType, new ActionTypeDTO());
+    }
+
+    public static CompositeActionTypeDTO map(final CompositeActionType actionType) {
+        final CompositeActionTypeDTO actionTypeDto = map(actionType, new CompositeActionTypeDTO());
+        actionTypeDto.children = ActionDTOMapper.map(actionType.getChildren());
         return actionTypeDto;
     }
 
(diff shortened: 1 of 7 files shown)


bundles/automation/org.eclipse.smarthome.automation.api/src/main/java/org/eclipse/smarthome/automation/dto/ActionTypeDTOMapper.java
```diff
@@ -29,15 +31,28 @@ public static ActionTypeDTO map(final ActionType actionType) {
...
+                dtos.add(map((CompositeActionType) type));
+            } else {
+                dtos.add(map(type));
+            }
         }
         return dtos;
     }
 
-}
+}
```
maggu2810: Please add a new line after `}`. This will IMHO work better with Git (according to POSIX ever `line` ends with a new line character).

bundles/automation/org.eclipse.smarthome.automation.api/src/main/java/org/eclipse/smarthome/automation/dto/ActionTypeDTOMapper.java
```diff
@@ -29,15 +31,28 @@ public static ActionTypeDTO map(final ActionType actionType) {
         return actionTypeDto;
     }
 
+    public static CompositeActionTypeDTO map(final CompositeActionType actionType) {
```
maggu2810: If we ever add an attribute to the ActionType (same for condition and trigger), it is also added to the CompositeActionType because it is a derived class.
Shouldn't we use a function that handles the ActionType[DTO] mapping that could be called by the mapping of a CompositeActionTypeDTO, too?

e.g.

```Java
    private static <T extends ActionTypeDTO> T map(final ActionType actionType, final T actionTypeDto) {
        fillProperties(actionType, actionTypeDto);
        actionTypeDto.inputs = actionType.getInputs();
        actionTypeDto.outputs = actionType.getOutputs();
        return actionTypeDto;
    }

    public static ActionTypeDTO map(final ActionType actionType) {
        return map(actionType, new ActionTypeDTO());
    }

    public static CompositeActionTypeDTO map(final CompositeActionType actionType) {
        final CompositeActionTypeDTO actionTypeDto = map(actionType, new CompositeActionTypeDTO());
        actionTypeDto.children = ActionDTOMapper.map(actionType.getChildren());
        return actionTypeDto;
    }
```

WDYT?

bundles/automation/org.eclipse.smarthome.automation.api/src/main/java/org/eclipse/smarthome/automation/dto/ConditionTypeDTOMapper.java
```diff
@@ -28,15 +30,27 @@ public static ConditionTypeDTO map(final ConditionType conditionType) {
...
+                dtos.add(map((CompositeConditionType) type));
+            } else {
+                dtos.add(map(type));
+            }
         }
         return dtos;
     }
 
-}
+}
```
maggu2810: See new line comment above.

bundles/automation/org.eclipse.smarthome.automation.api/src/main/java/org/eclipse/smarthome/automation/dto/TriggerTypeDTOMapper.java
```diff
@@ -28,15 +30,27 @@ public static TriggerTypeDTO map(final TriggerType triggerType) {
...
+                dtos.add(map((CompositeTriggerType) type));
+            } else {
+                dtos.add(map(type));
+            }
         }
         return dtos;
     }
 
-}
+}
```
maggu2810: See newline comment above

Issue #2854, opened 2017-01-18
Hidden files cause log exceptions in automation file provider


kaikreuzer:
Just came across this in my log:

```
2017-01-18 12:24:17.362 [DEBUG] [.f.TemplateFileProviderWatcher:248  ] - Parser DS_Store not availablejava.lang.Exception: null
	at org.eclipse.smarthome.automation.internal.provider.file.AbstractFileProvider.importFile(AbstractFileProvider.java:248)
	at org.eclipse.smarthome.automation.internal.provider.file.AbstractFileProvider.importResources(AbstractFileProvider.java:144)
	at org.eclipse.smarthome.automation.internal.provider.file.AbstractFileProvider.importResources(AbstractFileProvider.java:138)
	at org.eclipse.smarthome.automation.internal.provider.file.WatchServiceUtil.initializeWatchService(WatchServiceUtil.java:40)
	at org.eclipse.smarthome.automation.internal.provider.file.TemplateFileProviderWatcher.initializeWatchService(TemplateFileProviderWatcher.java:22)
	at org.eclipse.smarthome.automation.internal.provider.file.AbstractFileProvider.modified(AbstractFileProvider.java:110)
	at org.eclipse.smarthome.automation.internal.provider.file.AbstractFileProvider.activate(AbstractFileProvider.java:86)
	at sun.reflect.NativeMethodAccessorImpl.invoke0(Native Method)
	at sun.reflect.NativeMethodAccessorImpl.invoke(NativeMethodAccessorImpl.java:62)
	at sun.reflect.DelegatingMethodAccessorImpl.invoke(DelegatingMethodAccessorImpl.java:43)
	at java.lang.reflect.Method.invoke(Method.java:498)
```

I would suggest that hidden files (i.e. files starting with a ".") are always ignored.

PR #3112, opened 2017-03-05
Ignore single hidden files as well

kaikreuzer:
related to #2854

### DIFF
--- a/bundles/automation/org.eclipse.smarthome.automation.provider.file/src/main/java/org/eclipse/smarthome/automation/internal/provider/file/AbstractFileProvider.java
+++ b/bundles/automation/org.eclipse.smarthome.automation.provider.file/src/main/java/org/eclipse/smarthome/automation/internal/provider/file/AbstractFileProvider.java
@@ -143,13 +143,15 @@ public abstract class AbstractFileProvider<E> implements Provider<E> {
                     }
                 }
             } else {
-                try {
-                    URL url = file.toURI().toURL();
-                    String parserType = getParserType(url);
-                    importFile(parserType, url);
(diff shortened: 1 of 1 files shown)


maggu2810:
The `AutomationWatchService` still uses `!path.getFileName().startsWith(".")`.
Shouldn't this be changed to use `isHidden()` method, too?
What is about e.g. a new directory entry for a hidden directory that contains non-hidden files?
If we ignore hidden files, I would expect we ignore a hidden directory, too.
WDYT?

adimova:
I agree that isHidden() should be used instead of .startsWith(".")
If new directory entry for a hidden directory that contains non-hidden files - IMHO the directory and all files in it should be ignored

To achieve these, the fix should be changed:

public void importResources(File file) {
        if (file.exists() && !file.isHidden()) {
            File[] files = file.listFiles();
            if (files != null) {
                for (File f : files) {
                    importResources(f);
                }
            } else {
                try {
                    URL url = file.toURI().toURL();
                    String parserType = getParserType(url);
                    importFile(parserType, url);
                } catch (MalformedURLException e) {
                    // can't happen for the 'file' protocol handler with a correctly formatted URI
                    logger.debug("Can't create a URL", e);
                }
            }
        }
    }

Also, the .startsWith(".") check should be removed from org.eclipse.smarthome.automation.internal.provider.file.AutomationWatchService.WatchQueueReader.processWatchEvent

adimova:
May I propose to fix this issue as part of the fix for #3129? This way we'll avoid the merging and conflict solving.

kaikreuzer:
> May I propose to fix this issue as part of the fix for #3129?

Of course.

PR #3119, opened 2017-03-06
Changes in org.eclipse.smarthome.core.audio

petarnv:
I have written tests about the `org.eclipse.smarthome.core.audio` bundle: https://github.com/eclipse/smarthome/pull/2687. I have found the following problems:

1. `createInputStream()` in URLAudioStream does not reuse an already opened connection, but opens a new one:
`openStream()` method calls `openConnection().getInputStream()`, which opens an entirely new connection and does not reuse the old one. Also, `openStream()` initiates HTTP GET request: http://stackoverflow.com/questions/2778312/how-does-java-url-openstream-work. For example, this may cause a problem when we use `oneTimeStreams` in the `AudioServlet`, because `oneTimeStreams` can be requested only a single time.
This problem can be solved, if we use `getInputStream()` method on the currently opened connection.

2. `getSourceIds()` in `AudioManagerImpl` does not return the ids of the sources correctly:
The method `getSourceIds()` should be iterating through sources, not sinks.

Signed-off-by: Petar Valchev petar.valchev@musala.com

### DIFF
--- a/bundles/core/org.eclipse.smarthome.core.audio/src/main/java/org/eclipse/smarthome/core/audio/URLAudioStream.java
+++ b/bundles/core/org.eclipse.smarthome.core.audio/src/main/java/org/eclipse/smarthome/core/audio/URLAudioStream.java
@@ -90,7 +90,10 @@ public class URLAudioStream extends org.eclipse.smarthome.core.audio.AudioStream
                 os.write(req.getBytes());
                 is = shoutCastSocket.getInputStream();
             } else {
-                is = streamUrl.openStream();
+                // getInputStream() method is more error-proof than openStream(),
+                // because openStream() does openConnection().getInputStream(),
+                // which opens a new connection and does not reuse the old one.
+                is = connection.getInputStream();
             }
             return is;
         } catch (MalformedURLException e) {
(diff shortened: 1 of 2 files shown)


petarnv:
Not sure why the ip-validation failed.

maggu2810:
The IP validation failed because of the missing angle brackets around the mail address in the Signed-off-by footer.
I assume you should also change your mail address in your Git settings to use lower case (upper case sounds a little bit noisy).

```    
Signed-off-by: Petar Valchev petar.valchev@musala.com
```

should be changed to:

```    
Signed-off-by: Petar Valchev <petar.valchev@musala.com>
```

```
author Petar Valchev <petar.valchev@MUSALA.COM> 1488799203 +0200
committer Petar Valchev <petar.valchev@MUSALA.COM> 1488799203 +0200
```

should be changed to:

```
author Petar Valchev <petar.valchev@musala.com> 1488799203 +0200
committer Petar Valchev <petar.valchev@musala.com> 1488799203 +0200
```

maggu2810:
The changes itself look good to me.

maggu2810:
Please use a correct Signed-off-by line next time.
Fixed it manually.

Thanks for the fix!

Issue #3115, opened 2017-03-05
Hue test: error in the Eclipse IDE


maggu2810:
There seems to be a problem of the Groovy Eclipse IDE plugin and the usage of full vs. integral builds.

# Steps to reproduce

## Full build

Trigger a build of the whole "Hue binding test" bundle

* Start the Eclipse IDE
* Open the Hue binding test
* Ensure "Project, Build autmatically" is activated
* Select "Project, Clean..." and choose `org.eclipse.smarthome.binding.hue.test` only

Check the "Problems" view.

There will be one error:

* Description

```
General error during class generation: nyi class org.eclipse.jdt.internal.compiler.lookup.UnresolvedReferenceBinding

java.lang.IllegalStateException: nyi class org.eclipse.jdt.internal.compiler.lookup.UnresolvedReferenceBinding
	at org.codehaus.jdt.groovy.internal.compiler.ast.JDTClassNodeBuilder.toRawType(JDTClassNodeBuilder.java:169)
	at org.codehaus.jdt.groovy.internal.compiler.ast.JDTClassNode.makeParameter(JDTClassNode.java:395)
	at org.codehaus.jdt.groovy.internal.compiler.ast.JDTClassNode.makeParameters(JDTClassNode.java:384)
	at org.codehaus.jdt.groovy.internal.compiler.ast.JDTClassNode.methodBindingToMethodNode(JDTClassNode.java:357)
	at org.codehaus.jdt.groovy.internal.compiler.ast.JDTClassNode.initializeMembers(JDTClassNode.java:310)
	at org.codehaus.jdt.groovy.internal.compiler.ast.JDTClassNode.initialize(JDTClassNode.java:230)
	at org.codehaus.jdt.groovy.internal.compiler.ast.JDTClassNode.lazyClassInit(JDTClassNode.java:161)
	at org.codehaus.groovy.ast.ClassNode.getUnresolvedSuperClass(ClassNode.java:1118)
	at org.codehaus.groovy.ast.ClassNode.getUnresolvedSuperClass(ClassNode.java:1108)
	at org.codehaus.groovy.ast.ClassNode.getSuperClass(ClassNode.java:1102)
	at org.codehaus.groovy.ast.ClassNode.isDerivedFrom(ClassNode.java:1040)
	at org.codehaus.groovy.classgen.asm.InvocationWriter.castToNonPrimitiveIfNecessary(InvocationWriter.java:863)
[...]

PR #3122, opened 2017-03-06
use pure Java for some Hue tests

maggu2810:
This change migrated some Groovy OSGi tests to pure Java ones.
In addition it tries to eliminate some usage of private members etc.

Fixes: https://github.com/eclipse/smarthome/issues/3115


maggu2810:
IMHO #3123 should be merged in front of this!

kaikreuzer:
I guess you have to rebase now...

maggu2810:
Done, let's wait for Travis CI

kaikreuzer:
This PR includes changes to the code under test. Is this to make it testable or just improvements that you came across when working on the tests?

maggu2810:
The current test access private members. As Groovy is using reflection for all that stuff it worked before but is a very bad test design (IMHO).
So, I changed the scope of some stuff to package or protected or added a getter / setter (with package or protected scope) if needed.
Do you see any other changes in the code under test that are not related to "visibility"?

kaikreuzer:
> but is a very bad test design (IMHO).

For me it is in contrast not really nice to change visibility of productive code just for testability... 
Changes like [this one](https://github.com/eclipse/smarthome/pull/3122/files#diff-c35c0ced4b8dc5fbe154c4ce75c85c5fR23) simply leave a bad feeling for me as there are many classes in this package which can all of a sudden access this field, although they are not supposed to do so.

> Do you see any other changes in the code under test that are not related to "visibility"?

No, I think all is about visibility.

maggu2810:
Really?
We should use reflection to access stuff in tests instead of using the common Java visibility keywords?
You complan about the changed visibility of internal(!) bundle stuff, that is already used for testing?
So, if you prefer to use reflection this is easy to change.

If you don't want to use reflection and you don't want to change visibility (also to internal stuff), then could you give me a hint what is your prefered solution?

kaikreuzer:
No, I don't prefer reflection.
I think you are right that it is mainly bad test design; actually, tests should use the visible parts of classes and not access their internals (nor should it be necessary to expose their internals JUST for the tests).
As you only refactored the tests to Java and that made these changes necessary, I am fine to merge it.

maggu2810:
I assume we agree both in the argumentation of each other.
I don't want to rewrite the tests itself only port it to pure Java to solve the error in the Eclipse IDE.
Perhaps we could improve the whole Groovy and test situation some time.

Issue #3132, opened 2017-03-10
Scheduler not working reliable for Cron Expressions


maggu2810:
Rules using a Cron trigger does not work as expected.

There are messages from the `ExpressionThreadPoolManager` (I assume for every rule one message):

```
2017-03-08 07:15:00,423 | Expression '0 15 07 ? * MON,TUE,WED,THU,FRI,SAT,SUN' has no future executions anymore
2017-03-08 07:15:00,474 | Expression '0 15 07 ? * MON,TUE,WED,THU,FRI,SAT,SUN' has no future executions anymore
2017-03-08 07:15:00,835 | Expression '0 15 07 ? * MON,TUE,WED,THU,FRI,SAT,SUN' has no future executions anymore
2017-03-08 07:15:00,849 | Expression '0 15 07 ? * MON,TUE,WED,THU,FRI,SAT,SUN' has no future executions anymore
2017-03-08 08:00:00,615 | Expression '0 00 08 ? * MON,TUE,WED,THU,FRI,SAT,SUN' has no future executions anymore
2017-03-08 19:15:00,372 | Expression '0 15 19 ? * MON,TUE,WED,THU,FRI,SAT,SUN' has no future executions anymore
2017-03-08 19:45:00,450 | Expression '0 45 19 ? * MON,TUE,WED,THU,FRI,SAT,SUN' has no future executions anymore
2017-03-08 20:00:00,166 | Expression '0 00 20 ? * MON,TUE,WED,THU,FRI,SAT,SUN' has no future executions anymore
2017-03-08 21:00:00,157 | Expression '0 00 21 ? * MON,TUE,WED,THU,FRI,SAT,SUN' has no future executions anymore
```

I assume it is a bug that there are no future executors for that expression.

@kgoderis Could you have a look at.
Is it similar to #2105 or #2746?

maggu2810:
I just had a look at the uptime of the system:
```
11:30:46 up 3 days, 22:04
```

and the date of the system
```
Fri Mar 10 CET 2017
```

So, it worked only one day this time.

maggu2810:
@kgoderis I have created a test using the current code base:

```
    @Test
    public void findNext() throws ParseException {
        final SimpleDateFormat sdf = new SimpleDateFormat("yyyy-MM-dd HH:mm:ss.SSS");

        final String[] expressions = new String[] { //
                "0 15 07 ? * MON,TUE,WED,THU,FRI,SAT,SUN", //
[...]

PR #3133, opened 2017-03-10
improve cron expression tests

maggu2810:
Improve the Cron Expression tests.

Related to: #3132


maggu2810:
This adds a test to ensure there could be next candidates found for some valid Cron expressions.

PR #3134, opened 2017-03-10
Fix https://github.com/eclipse/smarthome/issues/3132

kgoderis:
@maggu2810 Can you give this a go? The adjusted test is not yet in the repo, hence I could not do the automated test, but I have manually tested the expressions you reported as faulty and they yield the correct output over here


maggu2810:
Hi @kgoderis thank you for your changes, but this seems to break a currently existing test:

```
    @Test(expected = IllegalArgumentException.class)
    public void dayOfWeekAndMonth() throws ParseException {
        new CronExpression("* * * 1 * 1");
    }
```

I assume this test needs to be changed now.

maggu2810:
The other test(s) I added succeeded now. Thank you! 
Can you change the one currently existing test, so it pass?

kgoderis:
> I assume this test needs to be changed now.

or the code ;-) I will check this asap

maggu2810:
Thank you for fixing it such promptly.

Issue #3129, opened 2017-03-08
Reduce the Threads "Dir Watcher" count


adimova:
The idea is all automation file providers to share one watch service, that way will achieve the reducing the count of the threads and they become two: one thread "Dir Watcher"(this one processing watch events) and one that watch the directory.

This change has impacts in the same code as #3112

kaikreuzer:
Question: Would it make sense to extend that improvement to ALL watch services in ESH, i.e. only have a single thread in place that watches all registered directories? This might be possible to change AbstractWatchService in a way that it only constructs a single watch thread for all class instances.

maggu2810:
Sounds like a good plan.

adimova:
As intention it is a good idea, I'll try to implement it, seems it is possible and not a big change

danchom:
Merged #3144

PR #3144, opened 2017-03-13
Solves Reduce the Threads "Dir Watcher" count #3129

adimova:


bundles/config/org.eclipse.smarthome.config.dispatch/src/main/java/org/eclipse/smarthome/config/dispatch/internal/ConfigDispatcher.java
```diff
@@ -69,9 +66,23 @@
  *
  * @author Kai Kreuzer - Initial contribution and API
  * @author Petar Valchev - Added sort by modification time, when configuration files are read
+ * @author Ana Dimova - reduce to a single watch thread for all class instances
  */
 public class ConfigDispatcher extends AbstractWatchService {
 
+    private static String getPathToWatch() {
+        String progArg = System.getProperty(SERVICEDIR_PROG_ARGUMENT);
```
kaikreuzer: Could we actually replace this automation specific "servicedir" by `ConfigConstants.getConfigFolder()`, which is the ESH global definition of the configuration directory?

adimova: probably we could, but I actually do not understand how should I change it

adimova: Is this what you mean?
private static String getPathToWatch() {
            return ConfigConstants.getConfigFolder() + File.separator + SERVICES_FOLDER;
}
or
private static String getPathToWatch() {
            return ConfigConstants.getConfigFolder();
}

kaikreuzer: Err, sorry, forget my comment - I actually didn't notice that this is within the ConfigDispatcher class, I somehow thought it was related to automation... Never mind!

bundles/core/org.eclipse.smarthome.core/src/main/java/org/eclipse/smarthome/core/service/WatchQueueReader.java
```diff
@@ -0,0 +1,290 @@
...
+            boolean watchSubDirectories) {
+        try {
+            if (watchSubDirectories) {
+                // walk through all folders and follow symlinks
+                registerWithSubDirectories(watchService, toWatch);
+            } else {
+                registerDirectoryInternal(watchService, watchService.getWatchEventKinds(toWatch), toWatch);
+            }
+        } catch (IOException e) {
+            e.printStackTrace();
```
kaikreuzer: I assume this should be deleted

bundles/core/org.eclipse.smarthome.core/src/main/java/org/eclipse/smarthome/core/service/WatchQueueReader.java
```diff
@@ -0,0 +1,290 @@
...
+
+    private synchronized void registerDirectoryInternal(AbstractWatchService service, Kind<?>[] kinds, Path directory) {
+        if (watchService == null) {
+            try {
+                watchService = FileSystems.getDefault().newWatchService();
+                qr = new Thread(this, "Dir Watcher");
+                qr.start();
+            } catch (IOException e) {
+                e.printStackTrace();
+                logger.error("Cannot activate watch service", e.getMessage());
```
kaikreuzer: '{}' is missing in the string

PR #3147, opened 2017-03-15
added media services (audio/voice) to script scope

kaikreuzer:


bundles/automation/org.eclipse.smarthome.automation.module.media/src/main/java/org/eclipse/smarthome/automation/module/media/internal/MediaScriptScopeProvider.java
```diff
@@ -0,0 +1,57 @@
...
+
+    @Override
+    public synchronized Map<String, Object> getScopeElements() {
+        if (elements == null) {
+            elements = new HashMap<>();
+            // services
+            elements.put("audio", audioManager);
+            elements.put("voice", voiceManager);
+        }
+        return elements;
```
maggu2810: If the audio or voice manager reference is removed (service disappears and appears again) you are still using the old references.
I don't think you should create the map only once.

kaikreuzer: The policy is static, so it would work if elements would be set to null again in deactivate.
But as the code here is not expensive, I removed the internal cache completely.
Note that I had copied that from DefaultScriptScopeProvider, which I have fixed by clearing the cache in deactivate now.

kaikreuzer:
> both differ from the common one

... which does not apply (yet) to the automation component - but thanks, I use at least the same now.

PR is updated, thanks for your quick review.

bundles/automation/org.eclipse.smarthome.automation.module.media/src/main/java/org/eclipse/smarthome/automation/module/media/internal/MediaScriptScopeProvider.java
maggu2810: > so it would work if elements would be set to null again in deactivate.

Sure, but there is no deactivate that set elements to null

PR #3149, opened 2017-03-15
Basic UI: fix update of group label with count (fix #3141)

### LINKED ISSUE #3141: Group:Switch / Group:Contact: number of items not refreshed in UIs
If you have such item with a %d in the label, you can notice that the value is not refreshed in Classic UI or HABdroid except when the group state is updated, that is number updated to 0 or number leaving 0.
To be even clearer. If you count your opened windows and you have 4 opened. You display the group item in UI, 4 is displayed. Then you close one window. You can see that the number remains at 4 in UI. It is updated only if you refresh the UI frame.

That is probably because there is no event produced with the updated number when the group global state is unchanged.

lolodomo:

### DIFF
--- a/bundles/io/org.eclipse.smarthome.io.rest.sitemap/src/main/java/org/eclipse/smarthome/io/rest/sitemap/internal/PageChangeListener.java
+++ b/bundles/io/org.eclipse.smarthome.io.rest.sitemap/src/main/java/org/eclipse/smarthome/io/rest/sitemap/internal/PageChangeListener.java
@@ -141,7 +141,11 @@ public class PageChangeListener implements StateChangeListener {
 
     @Override
     public void stateChanged(Item item, State oldState, State newState) {
-        Set<SitemapEvent> events = constructSitemapEvents(item, oldState, newState, widgets);
+        // For all items except group, send an event only when the event state is changed.
+        if (item instanceof GroupItem) {
+            return;
+        }
+        Set<SitemapEvent> events = constructSitemapEvents(item, widgets);
         for (SitemapEvent event : events) {
             for (SitemapSubscriptionCallback callback : distinctCallbacks) {
                 callback.onEvent(event);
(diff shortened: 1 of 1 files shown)


bundles/io/org.eclipse.smarthome.io.rest.sitemap/src/main/java/org/eclipse/smarthome/io/rest/sitemap/internal/PageChangeListener.java
```diff
@@ -151,13 +155,25 @@ public void stateChanged(Item item, State oldState, State newState) {
...
+        }
+        Set<SitemapEvent> events = constructSitemapEvents(item, state, widgets);
+        for (SitemapEvent event : events) {
+            for (SitemapSubscriptionCallback callback : distinctCallbacks) {
+                callback.onEvent(event);
+            }
+        }
     }
 
-    private Set<SitemapEvent> constructSitemapEvents(Item item, State oldState, State newState, List<Widget> widgets) {
```
kaikreuzer: you removed oldState, because it isn't used at all, right?
Is newState actually used? It does not seem to be the case either...

lolodomo: The oldState parameter was not used and I can't provide it in case of "state updated". That is the reason why I suppressed it.
But you're rigjht, the newState parameter is not used too.

kaikreuzer: So could you please remove it as well as?

lolodomo: Will do it.

lolodomo:
Parameter supressed.

lolodomo:
Done.
I renamed the PR as the fix is mainly for Basic UI.

kaikreuzer:
Thanks!

PR #3152, opened 2017-03-17
added access to itemRegistry to scripts

kaikreuzer:
Working on some new use cases, I quickly noticed that only having access to item states in scripts isn't enough - similar as we make the ThingRegistry available, we also need the ItemRegistry.
Nonetheless, it can also be helpful to have a quick access to the states directly - so I have kept that in place.
Wrt the naming, it didn't make any sense to keep "items" as a key for the states, so I have changed this to "states", which is effectively a breaking change.

### DIFF
--- a/bundles/automation/org.eclipse.smarthome.automation.module.script.defaultscope/src/main/java/org/eclipse/smarthome/automation/module/script/defaultscope/internal/DefaultScriptScopeProvider.java
+++ b/bundles/automation/org.eclipse.smarthome.automation.module.script.defaultscope/src/main/java/org/eclipse/smarthome/automation/module/script/defaultscope/internal/DefaultScriptScopeProvider.java
@@ -150,7 +150,8 @@ public class DefaultScriptScopeProvider implements ScriptScopeProvider {
             elements.put("StringType", StringType.class);
 
             // services
-            elements.put("items", new ItemRegistryDelegate(itemRegistry));
+            elements.put("states", new ItemRegistryDelegate(itemRegistry));
+            elements.put("items", itemRegistry);
             elements.put("things", thingRegistry);
             elements.put("events", busEvent);
             elements.put("rules", ruleRegistry);


maggu2810:
Shouldn't we stop to inject the registries and other non-wrapped objects at all?
If we every change the API of such an object we could break all existing script that use a removed / changed function. Also I don't know if really every function of that objects make sense for script.
Wouldn't it be better to create wrapper objects and start with a script API? So we could transparent fix a changed API of non-script stuff by changing the code in the wrapper objects?

kaikreuzer:
> start with a script API?

Well, this depends on whether you would want to consider the registries a part of an official ESH API or not - in general, they are exported and made available. So just the same way they are available to Java, they could be ade available to Javascript? At least that should be the case for #1783, where whole handler implementations can be scripted.

For The ScriptAction/ScriptCondition itself, we could also think about a restricted API. But this also means that we have additional documentation and maintenance efforts.

maggu2810:
Isn't there a little difference between bundles, addons, bindings, ...?
OSGi bundles can use version semantic, bundles could be updated by the developer to fit to the respective ESH version etc.
Scripts / rules are written / created by the end user and should be stable as long as possible. IMHO also after a big refactor of the remaining core part etc.
I think there is a difference between updating bundles and change all existing rules...
I also don't think that rules could be updated easily by a migration logic.

So, IMHO a "normal" Java API break that involved this party bundle and a API break that breaks script "at runtime" is not the same. At least it cannot be fixed such easily on a "central" place.

kaikreuzer:
Would you want to make a suggestion for a scripthandler javascript API?

kceiw:
I'm new to the code base and the project. I may miss some point from your perspective. But here is my two cent.  Exposing ItemRegistry makes it harder to make changes to the interface in future. If ItemRegistry interface changes again, it'll break the existing scripts again. If we need to expose the item to script, how about we create a new action such as getItemByName() which is implemented by ItemRegistry. We can create some interface to use for script. So that we can keep the same interface to script but we can change the ItemRegistry interface.

PR #3154, opened 2017-03-17
Don't try to guess the mimetype from the content for Sonos.

kaikreuzer:
I noticed that Sonos has some bug, which makes it serve the audio stream instead of the artwork image, if there is no image available - which means that megabytes of data are transferred, just to find out that it isn't an image after all.

@lolodomo Fyi, please let me know, in case I miss anything here.

### DIFF
--- a/extensions/binding/org.eclipse.smarthome.binding.sonos/src/main/java/org/eclipse/smarthome/binding/sonos/handler/ZonePlayerHandler.java
+++ b/extensions/binding/org.eclipse.smarthome.binding.sonos/src/main/java/org/eclipse/smarthome/binding/sonos/handler/ZonePlayerHandler.java
@@ -556,13 +556,8 @@ public class ZonePlayerHandler extends BaseThingHandler implements UpnpIOPartici
             URLConnection connection = url.openConnection();
             contentType = connection.getContentType();
             logger.debug("Content type from headers: {}", contentType);
-            if (contentType == null) {
-                input = connection.getInputStream();
-                contentType = URLConnection.guessContentTypeFromStream(input);
-                logger.debug("Content type from data: {}", contentType);
-                if (contentType == null) {
-                    contentType = RawType.DEFAULT_MIME_TYPE;
-                }
+            if (contentType == null || contentType.isEmpty()) {
+                contentType = RawType.DEFAULT_MIME_TYPE;
             }
         } catch (IOException e) {
             logger.debug("Failed to identify content type from URL: {}", e.getMessage());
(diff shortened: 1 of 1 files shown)


lolodomo:
What's a pity to disable this code :(
Can we explain a use case where you got this problem ? It was when using what music service ?
Because all my tests were ok on my side (tunein, spotify, ...).

If we really have no choice that suppressing this feature, your code change looks to me.

kaikreuzer:
Does the `connection.getContentType();` not already provide the type for tunein etc.? So is this code really necessary for those services?
My use case was an mp3 file, which I played from my local device and which did not have any cover art included. The Sonos speaker then served this whole file (5MB) as the cover art url and I didn't see a way to determine that it wasn't an image.

lolodomo:
No, unfortunately, content-type is not often provided by Sonos in the headers. That was the reason for my additional code that fetch the content from the the data itself.
Your use case is one I forgot to test ! Nothing special in the URL that could help us to determine that it is not the URL of a cover art ?

kaikreuzer:
> Nothing special in the URL that could help us to determine that it is not the URL of a cover art ?

Unfortunately not. As I said, I would say this is actually a bug in Sonos, because it does not really make any sense to serve the audio stream as cover art...

lolodomo:
You might keep my code in comments for the day Sonos will fix this bug.

And as image item with your patch will mostly no more work in any UI, it will make sense to make the channel currentalbumart an advanced channel and the channel currentalbumarturl a not advanced channel.

lolodomo:
@kaikreuzer : what do you think about checking the content length first ? If too big, we could assume that it is not a cover art and so do not guess the content ttpe.

kaikreuzer:
I would have to check if there is a content-length header set - if it isn't, downloading the whole content only to figure out its size might not help us much ;-)
I'm on the road this week, so won't have time to test that the next days, but I am happy to leave this PR open if we find some good solution. I definitely do not want to break the image support for the tunein, spotify etc.

lolodomo:
Yes, only check the content-length header.
I will try to find the time to check myself and will report to you.

lolodomo:
I tried to add
```
            long contentLength = connection.getContentLengthLong();
            logger.debug("Content length from headers: {}", contentLength);
```
and using tunein and spotify I always get a content-length while I don't get a content-type.
So I would suggest to check content-length header when there is no content-type header. If content-length is present and not too big, then try to guess the content type from the data.

tunein service is apparently using PNG images and their byte size seems to be very low (less than 10 KB.
Spotify service is apparently using JPEG images and their byte size seems to be generally lower than 250 KB.
We could try with 500000 bytes as threshold to not guess the type from data.?

kaikreuzer:
Sounds good, let's go for it!
Could you create an according PR (since I am still not in a situation where I can test it)? I'll close this PR here.

lolodomo:
@kaikreuzer : I worked on this but finally was not able to reproduce your problem. If I play a track from my media lobrary that is without album cover art, the cover art channel is just updated to UNDEF and there is no big data download.

PR #3169, opened 2017-03-22
fix NPE in Configuration.toString()

### LINKED ISSUE #3167: NPE in Configuration.toString()
In the ThingHandlerFactory of my binding I log a Thing's configuration when I process a Thing in createHandler(). The thing types are initialized with static XMLs, and I have several configuration properties which do not have a default value. The current Configuration.toString() method fails with a NPE, since it attempts to read the class of a property's value, and such value is not set.

sjka:
fixes #3167

### DIFF
--- a/bundles/config/org.eclipse.smarthome.config.core/src/main/java/org/eclipse/smarthome/config/core/Configuration.java
+++ b/bundles/config/org.eclipse.smarthome.config.core/src/main/java/org/eclipse/smarthome/config/core/Configuration.java
@@ -221,7 +221,7 @@ public class Configuration {
                 sb.append(", ");
             }
             sb.append(String.format("{key=%s; type=%s; value=%s}", prop.getKey(),
-                    prop.getValue().getClass().getSimpleName(), prop.getValue()));
+                    prop.getValue() != null ? prop.getValue().getClass().getSimpleName() : "?", prop.getValue()));
         }
         sb.append("]");
         return sb.toString();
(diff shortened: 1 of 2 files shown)


bundles/config/org.eclipse.smarthome.config.core/src/main/java/org/eclipse/smarthome/config/core/Configuration.java
```diff
@@ -221,7 +221,7 @@ public String toString() {
                 sb.append(", ");
             }
             sb.append(String.format("{key=%s; type=%s; value=%s}", prop.getKey(),
-                    prop.getValue().getClass().getSimpleName(), prop.getValue()));
+                    prop.getValue() != null ? prop.getValue().getClass().getSimpleName() : "?", prop.getValue()));
```
maggu2810: I prefer to use a temp. variable instead of calling the method `getValue()` three times if only one call is necessary. :wink:

sjka: good point - why did you merge it then? 😉 
-> #3171

maggu2810: Because the code itself is okay and I have not been sure, if such stuff is a good reason to not approve the changes.

PR #3170, opened 2017-03-22
Fix circular service reference

### LINKED ISSUE #3150: Circular reference while initializing StateDescriptionProvider
Hi,

we're having an issue, that sometimes(1 out of 4) at startup a circular reference occurs while initializing StateDescriptionProvider. It seems that this circular reference is happening between ItemChannelLinkRegistry and ItemRegistry. We tried to change some configurations, but couldn't solve it.

That's what we tried:

- We used different esh versions: 0.9.0.b4 and 0.9.0-SNAPSHOT
- We used different karaf versions: 4.0.4, 4.0.8, 4.1.0
- We tried Felix aswell as Equinox (you can find the stacktraces attached for both [karaf: 4.0.8 and esh: 0.9.0-SNAPSHOT])
- We looked into the configuration files of openhab and tried to adjust ours to the openhab configurations

We saw the #2413 issue but we're not sure if it is related here. We are using automation bundles, but the error occurs in esh core.

If you need more information, just tell.

**Stacktraces:**
- [circular_reference_equinox.txt](https://github.com/eclipse/smarthome/files/847929/circular_reference_equinox.txt)
- [resulted_in_a_cycle.txt](https://github.com/eclipse/smarthome/files/847939/resulted_in_a_cycle.txt)

sjka:
As it turned out there is the following "circle":

  StateDescriptionProvider
  -> ItemChannelLinkRegistry
     -> ItemRegistry

The ItemRegistryImpl opens a ServiceTracker to StateDescriptionProvider on activate().
As this last reference is not managed by DS it is not handled gracefully.

Therefore with this change I added dynamic service references into the ItemRegistry via DS.
As a side-effect I also introduced a callback interface for the items, so that the
actual service instances don't need to be injected into the items anymore, but the items
can get the most recent list via the StateDescriptionProviderAccessor interface.

fixes #3150

### DIFF
--- a/bundles/core/org.eclipse.smarthome.core/src/main/java/org/eclipse/smarthome/core/internal/items/ItemRegistryImpl.java
+++ b/bundles/core/org.eclipse.smarthome.core/src/main/java/org/eclipse/smarthome/core/internal/items/ItemRegistryImpl.java
@@ -13,7 +13,6 @@ import java.util.Collections;
 import java.util.Comparator;
 import java.util.List;
 import java.util.Map;
-import java.util.concurrent.ConcurrentHashMap;
 
 import org.eclipse.smarthome.core.common.registry.AbstractRegistry;
 import org.eclipse.smarthome.core.common.registry.Provider;
@@ -29,11 +28,7 @@ import org.eclipse.smarthome.core.items.ItemUtil;
 import org.eclipse.smarthome.core.items.ManagedItemProvider;
 import org.eclipse.smarthome.core.items.events.ItemEventFactory;
 import org.eclipse.smarthome.core.types.StateDescriptionProvider;
-import org.osgi.framework.BundleContext;
-import org.osgi.framework.Constants;
-import org.osgi.framework.ServiceReference;
 import org.osgi.service.component.ComponentContext;
-import org.osgi.util.tracker.ServiceTracker;
 import org.slf4j.Logger;
 import org.slf4j.LoggerFactory;
 
@@ -51,12 +46,8 @@ public class ItemRegistryImpl extends AbstractRegistry<Item, String, ItemProvide
 
     private final Logger logger = LoggerFactory.getLogger(ItemRegistryImpl.class);
 
-    private StateDescriptionProviderTracker stateDescriptionProviderTracker;
-
     private List<StateDescriptionProvider> stateDescriptionProviders = Collections
             .synchronizedList(new ArrayList<StateDescriptionProvider>());
-
-    private Map<String, Integer> stateDescriptionProviderRanking = new ConcurrentHashMap<>();
 
     public ItemRegistryImpl() {
         super(ItemProvider.class);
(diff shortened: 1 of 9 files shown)


maggu2810:
I didn't get the point why we need the `StateDescriptionProviderAccessor` now.

> As a side-effect I also introduced a callback interface for the items, so that the
actual service instances don't need to be injected into the items anymore, but the items
can get the most recent list via the StateDescriptionProviderAccessor interface.

At the moment the items have a reference to the list of the StateDescriptionProviders that is maintained by the ItemRegistryImpl.

The only difference I see now is that you make the list non-modifiable by the items (but this would be possible using an unmodifiable view, too).

But IMHO as long as the items have used the reference to the list, it is the most recent one, too.
What am I missing?

sjka:
No, it's not related. On a second thought I should have separated it into two PRs.

The reasoning for introducing the accessor was that if we spread the service instances everywhere (i.e. into all items), it is really hard to ensure nobody keeps any stale references anywhere, resulting in troubles when releasing the classloader on bundle update. So I thought in this way it is easier to stay in control of what happens to the service references. 

However, I can move that out of this PR easily, if you like.

sjka:
done - I took it out of this PR

maggu2810:
> The reasoning for introducing the accessor was that if we spread the service instances everywhere (i.e. into all items), it is really hard to ensure nobody keeps any stale references anywhere, resulting in troubles when releasing the classloader on bundle update. So I thought in this way it is easier to stay in control of what happens to the service references.

Regardless you took it out, but perhaps you can explain me:

Previously the reference to a `List<StateDescriptionProvider>` (always the same list) has been injected to the items.
After your change a reference to an accessor (always the same reference) that returns that list has been injected to the items.

I assume I do not get the real difference wrt the classloader etc.

bundles/core/org.eclipse.smarthome.core/src/main/java/org/eclipse/smarthome/core/internal/items/ItemRegistryImpl.java
```diff
@@ -342,63 +333,36 @@ protected void notifyListenersAboutUpdatedElement(Item oldElement, Item element)
...
             for (Item item : getItems()) {
                 ((GenericItem) item).setStateDescriptionProviders(stateDescriptionProviders);
             }
         }
     }
 
+    protected void removeStateDescriptionProvider(StateDescriptionProvider provider) {
+        stateDescriptionProviders.remove(provider);
+        for (Item item : getItems()) {
+            ((GenericItem) item).setStateDescriptionProviders(stateDescriptionProviders);
```
maggu2810: The list that has been set previously is still the same reference.
We are using a setXXX method, so the calling method should use the given reference.
This is still the same as before.

Didn't you think that a notification that the content of a collection has been changed differs from the collection has been changed?

sjka: Not sure if it would make a difference for Items, as they would not react on that themselves anyway. 

However, it's the way it was before this change, I didn't touch that (apart from moving it from the tracker to the DS injection method...)

PR #3172, opened 2017-03-22
fix a typo in firmware console command

Mixajlo:


Mixajlo:
Well, that is what happens when you are in a hurry :D

PR #3173, opened 2017-03-22
Fix template resolution when template appears after rule creation.

danchom:


bundles/automation/org.eclipse.smarthome.automation.provider.file/src/main/java/org/eclipse/smarthome/automation/internal/provider/file/AbstractFileProvider.java
```diff
@@ -259,9 +259,10 @@ protected void updateProvidedObjectsHolder(URL url, Set<E> providedObjects) {
             List<String> uids = new ArrayList<String>();
             for (E providedObject : providedObjects) {
                 String uid = getUID(providedObject);
-                notifyListeners(providedObjectsHolder.get(uid), providedObject);
+                E oldProvidedObject = providedObjectsHolder.get(uid);
```
maggu2810: the put method of a map returns the old object, so why calling `get` in front of if we notify the listeners after the put call now?

```Java
final String uid = getUID(providedObject);
uids.add(uid);
final E oldProvidedObject = providedObjectsHolder.put(uid, providedObject);
notifyListeners(oldProvidedObject, providedObject);
```

danchom: . Yes, thanks maggu2810

PR #3178, opened 2017-03-23
check the existence of the subscription

kaikreuzer:

### DIFF
--- a/extensions/ui/org.eclipse.smarthome.ui.basic/src/main/java/org/eclipse/smarthome/ui/basic/internal/servlet/WebAppServlet.java
+++ b/extensions/ui/org.eclipse.smarthome.ui.basic/src/main/java/org/eclipse/smarthome/ui/basic/internal/servlet/WebAppServlet.java
@@ -147,7 +147,7 @@ public class WebAppServlet extends BaseServlet {
                     if (subscriptions.exists(subscriptionId)) {
                         subscriptions.setPageId(subscriptionId, sitemap.getName(), sitemapName);
                     } else {
-                        logger.warn("Basic UI requested a non-existing event subscription id ({})", subscriptionId);
+                        logger.debug("Basic UI requested a non-existing event subscription id ({})", subscriptionId);
                     }
                 }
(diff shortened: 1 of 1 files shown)


extensions/ui/org.eclipse.smarthome.ui.basic/src/main/java/org/eclipse/smarthome/ui/basic/internal/servlet/WebAppServlet.java
```diff
@@ -144,7 +144,11 @@ public void service(ServletRequest req, ServletResponse res) throws ServletExcep
             if (widgetId == null || widgetId.isEmpty() || widgetId.equals(sitemapName)) {
                 // we are at the homepage, so we render the children of the sitemap root node
                 if (subscriptionId != null) {
-                    subscriptions.setPageId(subscriptionId, sitemap.getName(), sitemapName);
+                    if (subscriptions.exists(subscriptionId)) {
+                        subscriptions.setPageId(subscriptionId, sitemap.getName(), sitemapName);
+                    } else {
+                        logger.warn("Basic UI requested a non-existing event subscription id ({})", subscriptionId);
```
maggu2810: Should we really fill the log with a warning because a user of the Basic UI requests a subscription ID that does not exist?
As it is a web page I assume a incorrect request is responded (logical) correctly without a message in the log.

kaikreuzer: Note that I only copied&pasted this from https://github.com/kaikreuzer/smarthome/blob/a2a238c39b9128a41041b67f6ac2356bdbaabc58/extensions/ui/org.eclipse.smarthome.ui.basic/src/main/java/org/eclipse/smarthome/ui/basic/internal/servlet/WebAppServlet.java#L158-L162, there the very same is done. But yes, I can reduce both places to debug logging.

kaikreuzer:
updated PR.

PR #3180, opened 2017-03-23
Feature/rule scope 5942

ViToni:
Added scope to Rule and search capability via REST.


bundles/automation/org.eclipse.smarthome.automation.api/src/main/java/org/eclipse/smarthome/automation/Rule.java
```diff
@@ -101,6 +104,25 @@ public String getUID() {
     }
 
     /**
+     * This method is used for getting the scope of the Rule. This property is set by the RuleEngine when
+     * the {@link Rule} is added. It's an optional property.
+     *
+     * @return scope of this {@link Rule}, an empty {@link String} if none is found
```
maggu2810: You return null that is different from an empty string.
So, do you want that the caller need to check for null and non empty or does it make sense to return a non-null string all the time (empty as you written in the JavaDoc).

bundles/automation/org.eclipse.smarthome.automation.api/src/main/java/org/eclipse/smarthome/automation/RuleRegistry.java
```diff
@@ -115,4 +129,37 @@
      */
     public void runNow(String ruleUID);
 
+    /**
+     * Creates a {@link Predicate} which can be used to filter {@link Rule}s for a given scope or {@code null} scope.
+     * @param scope to search for
+     * @return created {@link Predicate}
+     */
+    public static Predicate<Rule> hasScope(String scope) {
+        if (null == scope) {
```
maggu2810: Impl. depends on the already comment method above. The JavaDoc states that an empty string (non-null) is returned...

ViToni: Good catch. I changed the implementation but forgot to change the JavaDoc.
Will clean it up and create a new PR.

bundles/automation/org.eclipse.smarthome.automation.core/src/main/java/org/eclipse/smarthome/automation/core/internal/RuleRegistryImpl.java
```diff
@@ -15,6 +15,7 @@
 import java.util.Map;
 import java.util.Map.Entry;
 import java.util.Set;
+import java.util.function.Predicate;
```
danchom: Please don't use 1.8 features, because this is a breakable change and the rule engine can't run on jdk  1.7 platforms.

bundles/automation/org.eclipse.smarthome.automation.core/src/main/java/org/eclipse/smarthome/automation/core/internal/RuleRegistryImpl.java
```diff
@@ -375,6 +377,20 @@ public Rule get(String key) {
     }
 
     @Override
+    public Collection<Rule> get(Predicate<Rule> predicate) {
+        Collection<Rule> result = new LinkedList<Rule>();
+
+        for (Collection<Rule> rules : elementMap.values()) {
+            // filter all matching rules and create copies
+            rules.stream()
```
danchom: This is incompatible change.

bundles/automation/org.eclipse.smarthome.automation.rest/src/main/java/org/eclipse/smarthome/automation/rest/internal/RuleResource.java
```diff
@@ -85,11 +90,66 @@ protected void unsetRuleRegistry(RuleRegistry ruleRegistry) {
...
+            } else {
+                // then this one is not null
+                predicate = tagPredicate;
+            }
+
+            Collection<EnrichedRuleDTO> enrichedRules = new ArrayList<>();
+
+            // add all matching rules and do the enrichment
+            ruleRegistry.get(predicate).stream()
+                .forEach(r -> enrichedRules.add(
```
danchom: please don't use steams and lambda expressions

bundles/automation/org.eclipse.smarthome.automation.api/src/main/java/org/eclipse/smarthome/automation/RuleRegistry.java
```diff
@@ -115,4 +129,70 @@
      */
     public void runNow(String ruleUID);
 
+    /**
+     * Creates a {@link Predicate} which can be used to filter {@link Rule}s for a given scope or {@code null} scope.
+     * @param scope to search for
+     * @return created {@link Predicate}
+     */
+    public static Predicate<Rule> hasScope(String scope) {
```
danchom: Impl and should not be part of the interface. Also I'm not sure that the scope is characteristic of the rule. It is up to your business logic.

bundles/automation/org.eclipse.smarthome.automation.api/src/main/java/org/eclipse/smarthome/automation/RuleRegistry.java
```diff
@@ -115,4 +129,70 @@
...
+        }
+    }
+
+    /**
+     * Creates a {@link Predicate} which can be used to filter {@link Rule}s for any of the given scopes and even {@code null} scope.
+     * @param scopes to search for
+     * @return created {@link Predicate}
+     */
+    public static Predicate<Rule> hasAnyScope(String... scopes) {
+        HashSet<String> scopeSet = new HashSet<String>(scopes.length);
```
danchom: Implementation. This can be impl by calling getAll(Predicate)  and should not be part of the interface. It is related to your business logic.

bundles/automation/org.eclipse.smarthome.automation.api/src/main/java/org/eclipse/smarthome/automation/RuleRegistry.java
```diff
@@ -115,4 +129,70 @@
...
+    }
+
+    /**
+     * Creates a {@link Predicate} which can be used to filter {@link Rule}s for a all given tags or {@link Rule}s without tags.
+     * All given tags must match, (the matched {@code Rule} might contain more).
+     * @param tags to search for
+     * @return created {@link Predicate}
+     */
+    public static Predicate<Rule> hasAllTags(Collection<String> tags) {
+        if (null == tags || tags.isEmpty()) {
```
danchom: Same as above. What is the difference with the getByTags(String...) method ?

bundles/automation/org.eclipse.smarthome.automation.api/src/main/java/org/eclipse/smarthome/automation/RuleRegistry.java
```diff
@@ -115,4 +129,70 @@
...
+        }
+    }
+
+    /**
+     * Creates a {@link Predicate} which can be used to filter {@link Rule}s for a any of given tags or {@link Rule}s without tags.
+     * @param tags to search for
+     * @return created {@link Predicate}
+     */
+    public static Predicate<Rule> hasAnyTags(Collection<String> tags) {
+        if (null == tags || tags.isEmpty()) {
```
danchom: Same as above. IMHO the interface should not be filled with utility methods. This functionality can be achieved by multiple calling getByTag() method or again you can use getAll(Predicate) method in your application.

bundles/automation/org.eclipse.smarthome.automation.api/src/main/java/org/eclipse/smarthome/automation/Rule.java
```diff
@@ -101,6 +104,30 @@ public String getUID() {
...
+     * the {@link Rule} is added or by the creating party. It's an optional property.
+     * <br/><br/>Implementation note
+     *    <br/>The scope is part of the {@code UID} and the prefix thereof.
+     *    <br/>If the UID does not contain a {@link SCOPE_DELIMITER} {@code null} will be returned.
+     *    <br/>If the UID does contain a {@link SCOPE_DELIMITER} the prefix until the first occurrence will be returned.
+     *    <br/>If the prefix would have a zero length {@code null} will be returned.
+     *
+     * @return scope of this {@link Rule}, or {@code null} if no scope or an empty scope is found
+    */
+    public String getScope() {
```
danchom: I think the scope is not part of the rule's features. It is up to your business logic how to treat uid of rule. What happen if  I don't like to divide my rule into scopes or I want to divide  them by groups or by color, in that case should I use your getScope() method to retrieve my rules?  I think we have agreed to add a method in ruleRegistry.getAll(Predicate) which can be used by your business logic and get scopes, groups and etc.

bundles/automation/org.eclipse.smarthome.automation.api/src/main/java/org/eclipse/smarthome/automation/Rule.java
```diff
@@ -32,9 +32,12 @@
  * @author Yordan Mihaylov - Initial Contribution
  * @author Ana Dimova - Initial Contribution
  * @author Vasil Ilchev - Initial Contribution
+ * @author Victor Toni - Add support for scope prefix
  */
 public class Rule {
 
+    public static final char SCOPE_DELIMITER = ':';
```
danchom: This constant is related to your business logic.

bundles/automation/org.eclipse.smarthome.automation.api/src/main/java/org/eclipse/smarthome/automation/RuleRegistry.java
```diff
@@ -44,10 +49,19 @@
...
  */
 public interface RuleRegistry extends Registry<Rule, String> {
 
     /**
+     * This method is used to get collection of {@link Rule}s which have the same {@link Predicate}.
+     *
+     * @param predicate to check the rules against
+     * @return collection of {@link Rule}s having matching predicate.
+     */
+    public Collection<Rule> get(Predicate<Rule> predicate);
```
ViToni: > Get method returns a single element. The name of the method should be getAll().
I will change the method name to reflect this convention.

Should the methods such as `getByTag()` and `getByTags()` be changed accordingly to `getAllByTag()` and `getAllByTags()` to support this nomenclature?

bundles/automation/org.eclipse.smarthome.automation.api/src/main/java/org/eclipse/smarthome/automation/RuleRegistry.java
ViToni: I can extract these to a dedicated class although IHMO they make the common `Predicate`s for `Rule` search readily available and because of their static nature they don't put any burden on the developer who wants to implement his own `RuleRegistry`...

bundles/automation/org.eclipse.smarthome.automation.api/src/main/java/org/eclipse/smarthome/automation/RuleRegistry.java
ViToni: Same as above.

bundles/automation/org.eclipse.smarthome.automation.api/src/main/java/org/eclipse/smarthome/automation/RuleRegistry.java
ViToni: The difference is that `Predicate`s can be combined and are way more powerful e.g. the current API can't be easily negated.

Example:
`getAll(hasAllTags(someTags).negate().or(hasScope(myScope)))`
Currently it's not possible to get the same result set using the current APi of `getBy...()`.
The Predicate is easy to write and one does not have to get all `Rule`s using `getAll()` and then do all the filtering manually.

bundles/automation/org.eclipse.smarthome.automation.api/src/main/java/org/eclipse/smarthome/automation/RuleRegistry.java
danchom: I don't think so, the get() method accepts uid as parameter and returns corresponding element (there is a 1:1 relation between parameter and the element). On the other hand the  getAll return collection of elements, all the registries in ESH implement this concept. 
Tags is something that group elements and for me it is normal to return the collection.

bundles/automation/org.eclipse.smarthome.automation.api/src/main/java/org/eclipse/smarthome/automation/Rule.java
ViToni: My understanding was that the scope wouldn't be a field of its own, so that persistence and other pats of the code don't have to be changed.
However it would be very useful to have the `getScope()` within the `Rule` class so that it can be used as part of the API.
For same reason it would be great to include the scope within the ID creation of the `RuleEngine` so that `Rule`s created by the RuleEngine can be identified as such.
The idea is to have `getScope()` in the API because otherwise every client would have to implement it itself.
In contrast to tags this property ("scope") cannot be edited by the user and therefor is a strong indicator "who" created the rule and can be used by specialzed clients of the API.

danchom: >However it would be very useful to have the getScope() within the Rule class

It will be useful if you want to divide rules into scopes i.e. in the DTAG case where you have a lot rule providers, but in case single provider the scopes can be useless.  Also in this implementation it is not possible for a rule to belong to more then one scope.
Also the uid is not required parameter when a new rule is added, there is not a problem to add a prefix for these rules, but I don't see practical reason why to do that. Why do you have to know which rules was created by specified uids and which rules are added with generated uid?
@kaikreuzer what do you think about this scope functionality of rule? Is it needed?

bundles/automation/org.eclipse.smarthome.automation.api/src/main/java/org/eclipse/smarthome/automation/RuleRegistry.java
danchom: This predicate make sense only if the rules have associated scope functionality.

bundles/automation/org.eclipse.smarthome.automation.api/src/main/java/org/eclipse/smarthome/automation/RuleRegistry.java
danchom: Agree that it is more powerful, but we duplicate functionality and removing the getByTags will be breakable change.

bundles/automation/org.eclipse.smarthome.automation.api/src/main/java/org/eclipse/smarthome/automation/Rule.java
kaikreuzer: > what do you think about this scope functionality of rule? Is it needed?

My understanding was that we agreed that we do not introduce the notion of "scope", but only implicitly induce it through a potential prefix/namespace within the rule UID.

danchom: I think the same.

bundles/automation/org.eclipse.smarthome.automation.rest/META-INF/MANIFEST.MF
```diff
@@ -5,10 +5,12 @@ Bundle-SymbolicName: org.eclipse.smarthome.automation.rest
 Bundle-Version: 0.9.0.qualifier
 Bundle-Vendor: Eclipse.org/SmartHome
 Bundle-RequiredExecutionEnvironment: JavaSE-1.8
-Import-Package: io.swagger.annotations;resolution:=optional,
+Import-Package: com.google.common.base,
+ io.swagger.annotations;resolution:=optional,
  javax.ws.rs,
  javax.ws.rs.core,
  org.eclipse.smarthome.automation,
+ org.eclipse.smarthome.automation.core.internal,
```
danchom: You can't import internal packages

bundles/automation/org.eclipse.smarthome.automation.core/src/main/java/org/eclipse/smarthome/automation/core/internal/RuleUtils.java
```diff
@@ -95,4 +100,31 @@ public static Rule getRuleCopy(Rule r) {
...
+     * the {@link Rule} is added or by the creating party. It's an optional property.
+     * <br/><br/>Implementation note
+     *    <br/>The scope is part of the {@code UID} and the prefix thereof.
+     *    <br/>If the UID does not contain a {@link SCOPE_DELIMITER} {@code null} will be returned.
+     *    <br/>If the UID does contain a {@link SCOPE_DELIMITER} the prefix until the first occurrence will be returned.
+     *    <br/>If the prefix would have a zero length {@code null} will be returned.
+     *
+     * @return scope of this {@link Rule}, or {@code null} if no scope or an empty scope is found
+    */
+    public static String getScope(Rule rule) {
```
danchom: This method is used by the RulePredicate class and may be its place is there.

bundles/automation/org.eclipse.smarthome.automation.api/META-INF/MANIFEST.MF
```diff
@@ -5,6 +5,7 @@ Bundle-SymbolicName: org.eclipse.smarthome.automation.api
 Bundle-Version: 0.9.0.qualifier
 Bundle-Vendor: Eclipse.org/SmartHome
 Import-Package: 
+ com.google.common.base,
```
kaikreuzer: @danchom I thought that you also wanted to avoid using Guava within the automation component, has this changed?

bundles/automation/org.eclipse.smarthome.automation.api/src/main/java/org/eclipse/smarthome/automation/RuleRegistry.java
```diff
@@ -44,10 +46,19 @@
...
  */
 public interface RuleRegistry extends Registry<Rule, String> {
 
     /**
+     * This method is used to get a collection of {@link Rule}s which have the same {@link Predicate}.
+     *
+     * @param predicate to check the rules against
+     * @return collection of {@link Rule}s having matching predicate.
+     */
+    public Collection<Rule> getAll(Predicate<Rule> predicate);
```
kaikreuzer: We must not create a dependency on Guava within our own API - this is very ugly.
We should also try to define a common scheme for doing filtered requests across our Registries (Items, Things, Rules, etc.). Currently, we have var args in place on the ItemRegistry and we had already discussed to introduce a common "Filter" mechanism.

bundles/automation/org.eclipse.smarthome.automation.core/src/main/java/org/eclipse/smarthome/automation/core/util/RulePredicates.java
```diff
@@ -0,0 +1,172 @@
...
+ * @author Victor Toni - initial contribution
+ * @author Victor Toni - provide default predicates for scope and tags
+ *
+ */
+public class RulePredicates {
+
+    /**
+     * Constant defining separator between module UID and output name.
+     */
+    public static final char SCOPE_DELIMITER = ':';
```
kaikreuzer: I thought we said that we do not introduce any notion of a "scope" anywhere - instead, I would simply keep calling it prefix or namespace.

bundles/automation/org.eclipse.smarthome.automation.api/META-INF/MANIFEST.MF
danchom: Now we have accepted guava, because otherwise it is hard to synchronize sources.

bundles/automation/org.eclipse.smarthome.automation.api/src/main/java/org/eclipse/smarthome/automation/RuleRegistry.java
danchom: I'm agree that there should be  a common filtering mechanism,
I think ViToni has choosen to use guava interfaces, because I've said that there was a jdk (i.e. skelmir) which currently does not supported jdk1.8. But you are right the dependence to guava is not a good idea.

bundles/automation/org.eclipse.smarthome.automation.api/META-INF/MANIFEST.MF
ViToni: Removed dependency to Guava.

bundles/automation/org.eclipse.smarthome.automation.api/src/main/java/org/eclipse/smarthome/automation/RuleRegistry.java
ViToni: Removed usage of predicates for this PR.

bundles/automation/org.eclipse.smarthome.automation.core/src/main/java/org/eclipse/smarthome/automation/core/util/RulePredicates.java
ViToni: It's called now PREFIX_SEPARATOR.

ViToni:
The PR has been changed to NOT include Gauva anymore.
Further the "Scope" has been renamed to "Prefix".
I would like to keep `org.eclipse.smarthome.automation.core.util.RulePrefix` as it provides `getPrefix(Rule)` which otherwise would have to be implemented by every other user.

bundles/automation/org.eclipse.smarthome.automation.core/src/main/java/org/eclipse/smarthome/automation/core/util/RulePredicates.java
sjka: I know you have been encouraged to not break the API preveiously and therefore don't make the Module's id become a UID. Although this is what it now effectively is... 
Nevertheless, why don't you at least refer to UID.SEPARATOR here and define your own constant instead?

kaikreuzer:
@ViToni Is this PR still relevant after having introduced the `stream()` method on the RuleRegistry?
The only interesting part might be to introduce some RulePredicates, which could also offer checking for a certain prefix, right?

ViToni:
I queued this PR until stream() was available. Will update the PR accordingly to include common Predicates.

bundles/automation/org.eclipse.smarthome.automation.core/src/main/java/org/eclipse/smarthome/automation/core/util/RulePredicates.java
ViToni: Changed the PR to use UID.SEPARATOR as reference.

ViToni:
@kaikreuzer Could you please check again. Hopefully all comments are considered now.

maggu2810:
This change breaks the Karaf feature verification:

```
        Caused by: org.osgi.service.resolver.ResolutionException: Unable to resolve root: missing requirement [root] osgi.identity; osgi.identity=esh-automation-rest; type=karaf.feature; version=0.9.0.SNAPSHOT; filter:="(&(osgi.identity=esh-automation-rest)(type=karaf.feature)(version>=0.9.0.SNAPSHOT))" [caused by: Unable to resolve esh-automation-rest/0.9.0.SNAPSHOT: missing requirement [esh-automation-rest/0.9.0.SNAPSHOT] osgi.identity; osgi.identity=org.eclipse.smarthome.automation.rest; type=osgi.bundle; version="[0.9.0.201706191106,0.9.0.201706191106]"; resolution:=mandatory [caused by: Unable to resolve org.eclipse.smarthome.automation.rest/0.9.0.201706191106: missing requirement [org.eclipse.smarthome.automation.rest/0.9.0.201706191106] osgi.wiring.package; filter:="(osgi.wiring.package=org.eclipse.smarthome.automation.core.util)"]]
                at org.apache.felix.resolver.ResolutionError.toException(ResolutionError.java:42)
```

kaikreuzer:
Sorry, I shouldn't have merged this - I also didn't notice the dependency to core.thing, which is a no-go... Will see if I can provide fixes for that.

maggu2810:
>I also didn't notice the dependency to core.thing, which is a no-go... 

We need to change that line:

```Java
public static final String PREFIX_SEPARATOR = UID.SEPARATOR;
```

and add the separator string instead of a reference.

kaikreuzer:
just doing it...

Issue #1560, opened 2016-05-24
Add syntax highlighting for Javascripts in script module


kaikreuzer:
It would be nice to have syntax highlighting for Javascript in the text area of the script module.

It seems that such stuff exists for angular, see http://ngmodules.org/modules/angular-highlightjs.

aounhaider1:
This library will not work in our case. It does the normal syntax highlighting in elements like divs. A textarea does not allow text formatting as such, so to implement this we need to explore some other available editor options.

lewie:
I often use this powerfull  and fast embeddable code editor: https://ace.c9.io/
A full featured example with 110 languages: https://ace.c9.io/build/kitchen-sink.html

lewie:
@aounhaider1, @kaikreuzer, 
we could write highlighters for OH specific Syntax like things, channels, items, sitemaps and xText Rules etc. too.

Don't know if this BSD-license does conflict with OH2-license in front-end https://github.com/ajaxorg/ace/blob/master/LICENSE

kaikreuzer:
BSD is compatible with EPL, so this should be no issue.

htreu:
An alternative to [ace](https://ace.c9.io/) might be [tern.js](http://ternjs.net/).

PR #3181, opened 2017-03-23
Script editor

aounhaider1:
Requires CQ for https://github.com/isagalaev/highlight.js/.
fixes https://github.com/eclipse/smarthome/issues/1560


htreu:
please improve on the following topics:
- the script code is not stored in the rule anymore.
- text area should have the same look/feel like before. grey & white borders are very large and the input line is not easily visible.
- highlighting works only on focus lost. highlighting should occur while you type.

aounhaider1:
```the script code is not stored in the rule anymore.````
Just checked, seems to be working.

```text area should have the same look/feel like before. grey & white borders are very large and the input line is not easily visible.```
Updated.

```highlighting works only on focus lost. highlighting should occur while you type.````
This is currently not achievable with highlight.js. Note that highlight.js is not a highlighting library for text areas or input fields. It is meant for divs etc. But I used it because of its light size. If we really need a complete text editor than we have to use some other highlighting library. The lightest one, that I know of, is around 200 KB (highlight.js is 12KB).

htreu:
Thanks for the fixes.
However there are still major issues when using highlight.js here:
- highlight.js will paste the formatted script code back into the input DIV and does not escape HTML tags (e.g. `let myVar = '<p>my paragraph<p>'`). This will totally mess up the editor and may even allow for script injection when reading a template from file.

My suggestion would be to look out for an alternative JS code editor with acceptable footprint.

htreu:
Maybe we can provide the xtended script editor as an option and load the specific JS editor from a cdn instead of providing it with the rule engine bundle. what do you think @aounhaider1?

kaikreuzer:
Might indeed be an option to download it remotely only when using it (and have it as a works-with dependency).
But this will be a new PR, so I suggest to close this one as highlight.js does not seem to be appropriate for our needs.

kaikreuzer:
Discussion should be continued on https://github.com/eclipse/smarthome/issues/1560.

Issue #3195, opened 2017-03-31
Problem with UID duplication when importing automation resources


adimova:
Currently the AbstractRegistry class does not check if the key of a newly added object is unique for the registry. To ensure ID uniqueness, the providers have to perform this check before adding the object in the registry. This is implemented in the bundle resource provider, but not in the file resource provider. As a result:
- If you install a bundle with some automation resources, and later import the same resources from the file system, you'll get automation objects with duplicate IDs. 
- If you first import the resources from the file system, and then install them in a bundle, the resources from the bundle will not be imported, and errors will be logged

danchom:
I think this check should be done in the registry, not in the provider. The provider should not have dependency to the registry. Also there is a problem if a third party provider does not do such check and the registry will end again with two module types with the same uids. 
The other option is different providers to use different prefixes and when it provides a new module type / template it has to check if the created object has uids implementing prefix convention of the provider. But again some providers may not implement the convention.

danchom:
@kaikreuzer what do you think about that?

kaikreuzer:
Sorry @danchom, I failed to read this due to holidays/Easter season at that time...

Yes, I fully agree - the check should be within the registry as otherwise we have cyclic dependencies. Using prefixes is imho fine as a measure to avoid clashes; we already introduced "core." prefixes on module types, other providers could do it similarly.

lewie:
I have tested JavaScript Rules and have the effect, that sometimes a deleted Script was not deleted from "openHAB registry" after it was deleted out of script files. It then continues until I restart openHAB again.

I have suspected the error in my JavaScript, but that fits rather to this issue?

Because: [The same effect also appears with jython scripts.](https://community.openhab.org/t/openhab-2-jython-support/27777/19)

kaikreuzer:
@lewie I doubt that this is related, because the JSR233 rules are not "imported" but provided through the ScriptedRuleProvider.

PR #3196, opened 2017-03-31
Prevent duplicated IDs of Module Types and Templates provided by diff…

adimova:
…erent providers

solves #3195

### DIFF
--- a/bundles/automation/org.eclipse.smarthome.automation.provider.file/src/main/java/org/eclipse/smarthome/automation/internal/provider/file/AbstractFileProvider.java
+++ b/bundles/automation/org.eclipse.smarthome.automation.provider.file/src/main/java/org/eclipse/smarthome/automation/internal/provider/file/AbstractFileProvider.java
@@ -256,9 +256,21 @@ public abstract class AbstractFileProvider<E> implements Provider<E> {
 
     protected void updateProvidedObjectsHolder(URL url, Set<E> providedObjects) {
         if (providedObjects != null && !providedObjects.isEmpty()) {
+            List<String> previousPortfolio = providerPortfolio.get(url);
             List<String> uids = new ArrayList<String>();
             for (E providedObject : providedObjects) {
                 String uid = getUID(providedObject);
+                if (providedObjectsHolder.get(uid) == null) {
+                    if (checkExistence(uid)) {
+                        logger.warn("{} with UID \"{}\" already exists! Failed to create a second with the same UID!",
+                                providedObject.getClass().getName(), uid, new IllegalArgumentException());
+                        continue;
+                    }
+                } else if (previousPortfolio == null || !previousPortfolio.contains(uid)) {
+                    logger.warn("{} with UID \"{}\" already exists! Failed to create a second with the same UID!",
+                            providedObject.getClass().getName(), uid, new IllegalArgumentException());
+                    continue;
+                }
                 uids.add(uid);
                 E oldProvidedObject = providedObjectsHolder.put(uid, providedObject);
                 notifyListeners(oldProvidedObject, providedObject);
@@ -296,6 +308,8 @@ public abstract class AbstractFileProvider<E> implements Provider<E> {
             }
         }
     }
+
+    protected abstract boolean checkExistence(String uid);
 
     protected abstract String getUID(E providedObject);
 
--- a/bundles/automation/org.eclipse.smarthome.automation.provider.file/src/main/java/org/eclipse/smarthome/automation/internal/provider/file/ModuleTypeFileProvider.java
+++ b/bundles/automation/org.eclipse.smarthome.automation.provider.file/src/main/java/org/eclipse/smarthome/automation/internal/provider/file/ModuleTypeFileProvider.java
@@ -14,6 +14,7 @@ import java.util.Locale;
 
 import org.eclipse.smarthome.automation.type.ModuleType;
 import org.eclipse.smarthome.automation.type.ModuleTypeProvider;
+import org.eclipse.smarthome.automation.type.ModuleTypeRegistry;
 
 /**
  * This class is implementation of {@link ModuleTypeProvider}. It extends functionality of {@link AbstractFileProvider}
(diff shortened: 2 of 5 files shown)


bundles/automation/org.eclipse.smarthome.automation.provider.file/src/main/java/org/eclipse/smarthome/automation/internal/provider/file/AbstractFileProvider.java
```diff
@@ -256,9 +256,19 @@ protected void importFile(String parserType, URL url) {
...
         if (providedObjects != null && !providedObjects.isEmpty()) {
+            List<String> previousPortfolio = providerPortfolio.get(url);
             List<String> uids = new ArrayList<String>();
             for (E providedObject : providedObjects) {
                 String uid = getUID(providedObject);
+                if (providedObjectsHolder.get(uid) == null) {
+                    if (checkExistence(uid)) {
+                        continue;
+                    }
+                } else if (previousPortfolio == null || !previousPortfolio.contains(uid)) {
```
kaikreuzer: Either this is a bug or the code is hard to read - wouldn't you have a problem IF the previousPortfolio already contains this uid?

bundles/automation/org.eclipse.smarthome.automation.provider.file/src/main/java/org/eclipse/smarthome/automation/internal/provider/file/AbstractFileProvider.java
```diff
@@ -256,9 +256,19 @@ protected void importFile(String parserType, URL url) {
...
+                    logger.error("{} with UID \"{}\" already exists! Failed to create a second with the same UID!",
```
kaikreuzer: In what kind of situation can we get here? If we have multiple files that contain entities with the same uid?
"error" logging should only be used in irrecoverable situations - I would think that "warning" should be enough here to tell the user that he has some problematic files and thus the second entity is simply ignored.

bundles/automation/org.eclipse.smarthome.automation.provider.file/src/main/java/org/eclipse/smarthome/automation/internal/provider/file/ModuleTypeFileProvider.java
```diff
@@ -49,8 +60,18 @@ protected String getUID(ModuleType providedObject) {
...
-            return Collections.<T>emptyList();
+            return Collections.<T> emptyList();
         }
         return (Collection<T>) new LinkedList<ModuleType>(values);
     }
+
+    @Override
+    protected boolean checkExistence(String uid) {
+        if (moduleTypeRegistry.get(uid) != null) {
+            logger.error("Module Type with UID \"{}\" already exists! Failed to create a second with the same UID!",
```
kaikreuzer: Having an error logging here is very unexpected. A method `checkExistence` should simply return true or false and not have any side effects.

bundles/automation/org.eclipse.smarthome.automation.provider.file/src/main/java/org/eclipse/smarthome/automation/internal/provider/file/TemplateFileProvider.java
```diff
@@ -53,4 +58,30 @@ public RuleTemplate getTemplate(String UID, Locale locale) {
...
+     * UIDs before these objects to be added in the system.
+     *
+     * @param uid UID of the newly created {@link Template}, which to be checked.
+     * @return {@code true} if {@link Template} with the same UID exists or {@code false} in the opposite
+     *         case.
+     */
+    @Override
+    protected boolean checkExistence(String uid) {
+        if (templateRegistry != null && templateRegistry.get(uid) != null) {
+            logger.error("Rule Template with UID \"{}\" already exists! Failed to create a second with the same UID!",
```
kaikreuzer: Having an error logging here is very unexpected. A method `checkExistence` should simply return true or false and not have any side effects.

bundles/automation/org.eclipse.smarthome.automation.provider.file/src/main/java/org/eclipse/smarthome/automation/internal/provider/file/AbstractFileProvider.java
adimova: This verification tells us that an object with such an identifier already exists, but has been added by another file, which could be a problem. IF the previousPortfolio already contains this uid, the object will simply be updated

bundles/automation/org.eclipse.smarthome.automation.provider.file/src/main/java/org/eclipse/smarthome/automation/internal/provider/file/AbstractFileProvider.java
adimova: ok

bundles/automation/org.eclipse.smarthome.automation.provider.file/src/main/java/org/eclipse/smarthome/automation/internal/provider/file/ModuleTypeFileProvider.java
adimova: Will be moved here:
if (checkExistence(uid)) {
      logger.error("Module Type with UID \"{}\" already exists! Failed to create a second with the same UID!"...
      continue;
}

bundles/automation/org.eclipse.smarthome.automation.provider.file/src/main/java/org/eclipse/smarthome/automation/internal/provider/file/TemplateFileProvider.java
adimova: Will be moved here:
if (checkExistence(uid)) {
logger.error("Module Type with UID "{}" already exists! Failed to create a second with the same UID!"...
continue;
}

kaikreuzer:
Sorry, I just took a step back here and think the overall approach of this PR is wrong.
If I am not mistaken, this would add a cyclic dependency between ModuleTypeFileProvider and ModuleTypeRegistry as well as the TemplateFileProvider and the TemplateRegistry. Or am I missing something?

adimova:
#this would add a cyclic dependency between ModuleTypeFileProvider and ModuleTypeRegistry as well as the TemplateFileProvider and the TemplateRegistry

yes, this is correct, see @danchom comment on  #3195

kaikreuzer:
Thanks for the pointer, @adimova - so I assume we should close this PR and maybe just do some hardening/safeguards in the registries?

adimova:
> just do some hardening/safeguards in the registries?
I agree with that

adimova:
This PR is replaced by #3670

Issue #2555, opened 2016-11-29
Things should automatically inherit channels from their (current) thing type


kaikreuzer:
I know that this has been discussed already, but I couldn't find an according issue for it, so I am entering one:

There is currently a mismatch in how Things are created that are read from a) the storage and b) from DSL.

In the DSL case, the Thing is constructed based on its type (and the channels declared within this type). Channels that are specified in the DSL are additionally added to it.

The ManagedThingProvider in contrast stores the Thing including all its channels and reads it back from the storage in the same way. As a result, channels that are newly introduced in a thing type do not end up on existing things. They have to be deleted and newly created in order to show the new features.

Imho we should change the ManagedThingProvider to work similar as the GenericThingProvider: It should only store what is not already defined on the thing type and always make sure that a loaded Thing has all channels that are defined on the current version of its type.

sjka:
Fully makes sense from my point of view. However, the implementation is going to be a little bit tricky as now the ManagedThingProvider will also have to wait until all XMLs have been processed, i.e. the ThingTypes are available. Just wanted to point out this limitation which we would newly introduce.

kaikreuzer:
Alternatively, could we have a service that updates Things in the registry dynamically when a thing type appears? That way, it could be the same mechanism (and implementation) for Managed+Generic and we would not even block the creation of Things in case no type information is available.

maggu2810:
> Imho we should change the ManagedThingProvider to work similar as the GenericThingProvider: It should only store what is not already defined on the thing type and always make sure that a loaded Thing has all channels that are defined on the current version of its type.

Make sense for me, too.

> is going to be a little bit tricky as now the ManagedThingProvider will also have to wait until all XMLs have been processed

Wouldn't it be make sense to wait for all the thing handling (and perhaps other stuff)  until all XMLs of a binding has been read?
This would also solve stuff #1550

cdjackson:
I've just seen this issue. This is a major concern with ZWave since thing types change reasonably often and there needs to be a way to update the definitions without deleting all the things and adding them back again each time.

In #4958 I proposed a different solution to this - to version the thing types and allow people to manually update (ie the UI tells them there is a new definition available, and they click a button in the UI to update to the latest version).

[...]

PR #3205, opened 2017-04-05
prevent bindings from removing things

### LINKED ISSUE #3199: Bindings should not be able to remove Things without user approval
Way back when we first discussed the introduction of [Thing Status](https://www.eclipse.org/smarthome/documentation/concepts/things.html#thing-status) we argued that a binding should only remove things upon user request. The [Status Transitions](https://www.eclipse.org/smarthome/documentation/concepts/things.html#status-transitions) visualize this by having the binding react to the status `REMOVING` and confirm the removal by setting the status to `REMOVED` upon which ESH deletes the thing itself.

But it appears that `REMOVING` is not a necessary precondition to `REMOVED` right now. We currently have bindings that set the status of a thing to `REMOVED` when the APIs the bindings use don't list a particular thing anymore. This leads to ESH cleaning out the thing and its references and the user doesn't (easily) realize what happened because the thing is just gone.

I see two options to get back to the idea of bindings not removing things without the user's permission or knowledge:

1. ESH may only clean up things when entering status `REMOVED` from `REMOVING`, otherwise preserving things in status `REMOVED` (as I believe was the original intention)
2. ESH introduced a new state `VANISHED` (or a new status detail, e.g. `OFFLINE:VANISHED`) to denote the situation where a thing no longer "physically" exists but is still mapped in ESH. Additionally ESH should make sure that `REMOVED` can't be set unless the status is `REMOVING`.

sjka:
...and making the documentation more explicit about this.

So far, bindings simply could set things to REMOVED at any time so they got
dropped from the database. Now it is ensured they can only do so after the
framework set it to REMOVING, i.e. they only are allowed to indicate that
the handleRemove() is finished, but cannot trigger the process anymore.

fixes #3199

### DIFF
--- a/bundles/core/org.eclipse.smarthome.core.thing/src/main/java/org/eclipse/smarthome/core/thing/internal/ThingManager.java
+++ b/bundles/core/org.eclipse.smarthome.core.thing/src/main/java/org/eclipse/smarthome/core/thing/internal/ThingManager.java
@@ -138,9 +138,9 @@ public class ThingManager extends AbstractItemEventSubscriber implements ThingTr
         @Override
         public void statusUpdated(Thing thing, ThingStatusInfo statusInfo) {
             // note: all provoked operations based on a status update should be executed asynchronously!
-            ensureValidStatus(statusInfo.getStatus());
-
             ThingStatusInfo oldStatusInfo = thing.getStatusInfo();
+            ensureValidStatus(oldStatusInfo.getStatus(), statusInfo.getStatus());
+
             if (ThingStatus.REMOVING.equals(oldStatusInfo.getStatus())
                     && !ThingStatus.REMOVED.equals(statusInfo.getStatus())) {
                 // only allow REMOVING -> REMOVED transition and ignore all other state changes
@@ -170,12 +170,17 @@ public class ThingManager extends AbstractItemEventSubscriber implements ThingTr
             }
         }
 
-        private void ensureValidStatus(ThingStatus status) {
-            if (!(ThingStatus.UNKNOWN.equals(status) || ThingStatus.ONLINE.equals(status)
-                    || ThingStatus.OFFLINE.equals(status) || ThingStatus.REMOVED.equals(status))) {
+        private void ensureValidStatus(ThingStatus oldStatus, ThingStatus newStatus) {
+            if (!(ThingStatus.UNKNOWN.equals(newStatus) || ThingStatus.ONLINE.equals(newStatus)
+                    || ThingStatus.OFFLINE.equals(newStatus) || ThingStatus.REMOVED.equals(newStatus))) {
                 throw new IllegalArgumentException(
-                        MessageFormat.format("Illegal status {0}. Bindings only may set {1}, {2}, {3} or {4}.", status,
+                        MessageFormat.format("Illegal status {}. Bindings only may set {}, {}, {} or {}.", newStatus,
                                 ThingStatus.UNKNOWN, ThingStatus.ONLINE, ThingStatus.OFFLINE, ThingStatus.REMOVED));
+            }
+            if (ThingStatus.REMOVED.equals(newStatus) && !ThingStatus.REMOVING.equals(oldStatus)) {
+                throw new IllegalArgumentException(
+                        MessageFormat.format("Illegal status {}. The thing was in state {} and not in {}", newStatus,
+                                oldStatus, ThingStatus.REMOVING));
             }
         }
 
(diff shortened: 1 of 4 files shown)


bundles/core/org.eclipse.smarthome.core.thing/src/main/java/org/eclipse/smarthome/core/thing/internal/ThingManager.java
```diff
@@ -170,12 +170,17 @@ public void statusUpdated(Thing thing, ThingStatusInfo statusInfo) {
...
         }
 
-        private void ensureValidStatus(ThingStatus status) {
-            if (!(ThingStatus.UNKNOWN.equals(status) || ThingStatus.ONLINE.equals(status)
-                    || ThingStatus.OFFLINE.equals(status) || ThingStatus.REMOVED.equals(status))) {
+        private void ensureValidStatus(ThingStatus oldStatus, ThingStatus newStatus) {
+            if (!(ThingStatus.UNKNOWN.equals(newStatus) || ThingStatus.ONLINE.equals(newStatus)
+                    || ThingStatus.OFFLINE.equals(newStatus) || ThingStatus.REMOVED.equals(newStatus))) {
+                throw new IllegalArgumentException(MessageFormat.format(
+                        "Illegal status {0}. Bindings only may set {1}, {2}, {3} or {4}.", newStatus,
```
maggu2810: Why using indices at all if the order is iterative?
Doesn't it add workload to the next one that would like to add another one in front of the others?

bundles/core/org.eclipse.smarthome.core.thing/src/main/java/org/eclipse/smarthome/core/thing/internal/ThingManager.java
```diff
@@ -170,12 +170,17 @@ public void statusUpdated(Thing thing, ThingStatusInfo statusInfo) {
...
+                        ThingStatus.UNKNOWN, ThingStatus.ONLINE, ThingStatus.OFFLINE, ThingStatus.REMOVED));
+            }
+            if (ThingStatus.REMOVED.equals(newStatus) && !ThingStatus.REMOVING.equals(oldStatus)) {
                 throw new IllegalArgumentException(
-                        MessageFormat.format("Illegal status {0}. Bindings only may set {1}, {2}, {3} or {4}.", status,
-                                ThingStatus.UNKNOWN, ThingStatus.ONLINE, ThingStatus.OFFLINE, ThingStatus.REMOVED));
+                        MessageFormat.format("Illegal status {0}. The thing was in state {1} and not in {2}", newStatus,
+                                oldStatus, ThingStatus.REMOVING));
```
maggu2810: Why using indices at all if the order is iterative?
Doesn't it add workload to the next one that would like to add another one in front of the others?

sjka:
Really good point, I didn't actually think about it when moving the code. 
PR is updated.

maggu2810:
Thanks

cdjackson:
If I understand this, it's not possible for a bundle to request that a thing is removed? Is that correct?

I would suggest that bridges at least should be able to request that a thing under its control is removed. If this isn't available, I think there will be quite some confusion about removing devices where a device is removed from a bridge. It will result in a two stage removal process where the device has to be removed from the bridge, and then the thing deleted.

sjka:
> it's not possible for a bundle to request that a thing is removed?

Yes, correct for the ThingHandler. This was not changed with this PR though, it was like that before. 

Generally it's a bad idea to mess with Things, even for a bridge. Bindings don't create Things magically (they put them into the inbox) and they should not delete them without the user's consent. 

Nevertheless, I agree that as of today there is no suitable way for a bridge to indicate that it _knows_ a device is gone. How about adding another ThingStatusDetail, it can be set to OFFLINE/"DEVICE_IS_GONE" (for the lack of a better name right now...)? Of course, if a solution doesn't want to bother the users with Thing management, there could be a service cleaing up those Things in the same way as it would "auto-approve" Things from the Inbox. 

Does that make sense for you? Would you create an issue for it then?

cdjackson:
> they should not delete them without the user's consent.

I don't disagree,  but we now have a situation where a user has to "consent" twice and this is confusing...
The problem here is that if a user requests a thing to be removed through the bridge, then there's no way for the bridge to actually remove the thing.

A concrete case -:

In ZWave, to exclude a device, you put the bridge into exclude mode, then go to the device and click a button. The bridge then gets a message saying "Node 4 is removed". The user is now confused as to why Node 4 (ie the thing) is still there, and they now have to go and remove the thing as well. Currently, when the bridge gets this message, it marks the thing as removed.

It's not normal for the binding to tell the network to remove a specific node (except in failure case) - just to put the zwave network into exclusion mode, and then to remove the node when you click the button. So, we couldn't have (for example) an implementation where the user simply deleted the thing, and it was excluded from the network as there's no guarantee that the thing (ie zwave device) that they deleted will be excluded.

One other scenario - at the moment, there's no easy way to update a things "configuration" (not parameters, but channels etc) when the XML is updated for the thing-type. Currently I tell people to delete the thing and add it back again in order to update the definition - if we linked "delete the thing" directly to removing the device from the network we would have another problem...

> Does that make sense for you? Would you create an issue for it then?

I think your approach sounds ok - have a device_is_gone state and allow an auto approval to have it removed. I wonder if the implementation shouldn't check that this can only be done by the things bridge (if it has a bridge) - but that's a detail.

I'll raise an issue - thanks.

maggu2810:
> One other scenario - at the moment, there's no easy way to update a things "configuration" (not parameters, but channels etc) when the XML is updated for the thing-type. Currently I tell people to delete the thing and add it back again

IIRC there is already an issue (or perhaps only a comment in an issue) to store only stuff that is changed and so new sections in the thing type are considered automatically.

maggu2810:
See #2555

PR #3211, opened 2017-04-06
Refactor thing properties display in PaperUI

cdjackson:
The primary change here is to split the thing properties table into a separate partial so that it can be overridden in a fragment. I've also modified the styling of the properties slightly as I personally think it looks a bit better, but if y'all disagree I can always change it back.

### DIFF
--- a/extensions/ui/org.eclipse.smarthome.ui.paper/web-src/css/components.css
+++ b/extensions/ui/org.eclipse.smarthome.ui.paper/web-src/css/components.css
@@ -857,7 +857,7 @@ md-input-container .md-errors-spacer {
 }
 
 .thing-properties {
-	margin-left: -5px;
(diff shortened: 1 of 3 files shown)


maggu2810:
Would be great to have "before" and "after" screenshots so we can discuss without the need to build and check the UI.

cdjackson:
The original...
image


And the proposal...
image

I've basically just got rid of the vertical lines in the table, and made them extend uniformly across the page (uniformly, as in it's the same for every thing).

kaikreuzer:
Looks much nicer indeed :-)

maggu2810:
Okay for me

maggu2810:
@aounhaider1 Would you like to review this changes?

aounhaider1:
Looks good.

PR #3215, opened 2017-04-07
Fix PaperUI to allow editing of thing parameters

cdjackson:
This updates PaperUI to allow editing of thing configuration parameters which is currently not possible.

### DIFF
--- a/extensions/ui/org.eclipse.smarthome.ui.paper/web-src/js/controllers.configuration.js
+++ b/extensions/ui/org.eclipse.smarthome.ui.paper/web-src/js/controllers.configuration.js
@@ -695,7 +695,7 @@ angular.module('PaperUI.controllers.configuration', [ 'PaperUI.constants' ]).con
     $scope.unlink = function() {
         $mdDialog.hide();
     }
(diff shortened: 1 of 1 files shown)


maggu2810:
@aounhaider1 Can you have a look at?

aounhaider1:
I don't understand what problem is this PR fixing. PaperUI already supports editing of thing configuration parameters. The only difference I see is that now instead of getting config description of parameters from thingType, this PR is getting it from config-descriptions. @cdjackson Can you please explain?

cdjackson:
Paper UI currently only allows editing of configuration that is defined in the thing-type - not the thing. This allows editing of configuration that is defined in the thing as well as the thing type.

Basically, as it currently stands, PaperUI doesn’t show any configuration of parameters that are defined dynamically in a ConfigProvider.

cdjackson:
To add - the important point here is not that we’re getting the config from the other REST endpoint - but we’re defining this with “thing” scope - not “thing-type” which is the default if you go through the thing endpoint.

aounhaider1:
Ok. The rest looks fine.

PR #2670, opened 2016-12-18
[WIP] add bluetooth transport and BLE binding

maggu2810:
This changes are based on a PR based by Chris Jackson.

Related to: https://github.com/eclipse/smarthome/pull/551

PR #3216, opened 2017-04-08
D-Bus: gh-pages

maggu2810:
Related to: https://github.com/eclipse/smarthome/pull/2670#issuecomment-289378352

I used a new major version for the respective bundles to signal that the API could been changed and does not need to be backward compatible.
I also used an alpha classifier to signal that the API could still change.

I realized the current dbus transport bundles states that the bundled dbus java uses the MIT license.
That is not correct, it "is licensed to you under your choice of the Academic Free License version 2.1, or the GNU Lesser/Library General Public License version 2." (see: https://cgit.freedesktop.org/dbus/dbus-java/tree/COPYING?id=2.7)

I cannot login to IPZilla anymore, so I cannot check the respective CQ.


sjka:
> I cannot login to IPZilla anymore, so I cannot check the respective CQ.

```
dbus-java Version: 2.7
Expat License (MIT Style)
```

kaikreuzer:
@maggu2810 As it stands for the moment, on the Eclipse Foundation is allowed to distribute the unmodified dbus-java library under the Expat License (MIT Style) and we had to add the about.html in io.transport.dbus accordingly.
If we really require patches to this library, we will have to discuss the way forward with the IP team.
For the time being, I would hope to get BLE working even with the unmodified version and thus stick to the existing dbus bundles.

PR #3217, opened 2017-04-08
dbus transport: unbundle libraries

maggu2810:
Prerequisite: #3216

There is currently no function of the bundle itself.
To be prepared for future utility classes, let's keep the bundle itself.


kaikreuzer:
Closing due to https://github.com/eclipse/smarthome/pull/3216#issuecomment-294830334.

PR #3221, opened 2017-04-08
PaperUI: Add ESH icon for browser use

cdjackson:
This adds the ESH icon so the browser can use it for icons in tabs/favourites etc.


cdjackson:
And for the more graphically minded, here's a picture ;)

Before-:
image

After-:
image

PR #3222, opened 2017-04-08
retain order of channels in ChannelTypeResource

### LINKED ISSUE #3220: Order of channel types returned by a binding gets lost in class ChannelTypeResource
If a binding implements the method `ChannelTypeProvider#getChannelTypes` to return a dynamically created list of channels types the channel order gets lost and the channels appear in Paper UI in a "random" order.

This happens because the returned channel types are copied to a `HashSet` in method  `ChannelTypeResource#convertToChannelTypeDTOs`.

sjka:
...as so far convertToChannelTypeDTOs() used a set with no obvious reason,
hence srambling them up.

fixes #3220

### DIFF
--- a/bundles/io/org.eclipse.smarthome.io.rest.core/src/main/java/org/eclipse/smarthome/io/rest/core/channel/ChannelTypeResource.java
+++ b/bundles/io/org.eclipse.smarthome.io.rest.core/src/main/java/org/eclipse/smarthome/io/rest/core/channel/ChannelTypeResource.java
@@ -8,10 +8,8 @@
 package org.eclipse.smarthome.io.rest.core.channel;
 
 import java.util.ArrayList;
-import java.util.HashSet;
 import java.util.List;
 import java.util.Locale;
-import java.util.Set;
 
 import javax.annotation.security.RolesAllowed;
 import javax.ws.rs.GET;
@@ -82,7 +80,7 @@ public class ChannelTypeResource implements SatisfiableRESTResource {
     public Response getAll(
             @HeaderParam(HttpHeaders.ACCEPT_LANGUAGE) @ApiParam(value = HttpHeaders.ACCEPT_LANGUAGE) String language) {
         Locale locale = LocaleUtil.getLocale(language);
-        Set<ChannelTypeDTO> channelTypeDTOs = convertToChannelTypeDTOs(channelTypeRegistry.getChannelTypes(locale),
+        List<ChannelTypeDTO> channelTypeDTOs = convertToChannelTypeDTOs(channelTypeRegistry.getChannelTypes(locale),
                 locale);
         return Response.ok(channelTypeDTOs).build();
     }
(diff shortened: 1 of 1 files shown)


htreu:
.
Simon Kaufmann <notifications@github.com> schrieb am Sa. 8. Apr. 2017 um
23:19:

sjka:
Closing and reopening in order to trigger another travis build - the failure seems unrelated, but still can't make any sense out of it...

bundles/io/org.eclipse.smarthome.io.rest.core/src/main/java/org/eclipse/smarthome/io/rest/core/channel/ChannelTypeResource.java
```diff
@@ -137,8 +129,8 @@ private ChannelTypeDTO convertToChannelTypeDTO(ChannelType channelType, Locale l
                 parameterGroups, channelType.getState(), channelType.getTags(), channelType.isAdvanced());
     }
 
-    private Set<ChannelTypeDTO> convertToChannelTypeDTOs(List<ChannelType> channelTypes, Locale locale) {
-        Set<ChannelTypeDTO> channelTypeDTOs = new HashSet<>();
+    private List<ChannelTypeDTO> convertToChannelTypeDTOs(List<ChannelType> channelTypes, Locale locale) {
+        List<ChannelTypeDTO> channelTypeDTOs = new ArrayList<>();
```
htreu: is uniqueness of `channelTypes` guaranteed? The former HashSet would have removed duplicates.

sjka: You are right in saying that it's now not guaranteed in the REST resource anymore. Nevertheless, it is using what it gets from `channelTypeRegistry.getChannelTypes(locale)`. I'm not sure right now whether it is explicitly handled there as of today, but that would be the right place to do so.

htreu: Looks like `ChannelTypeDTO` has no own hasCode or equals method. So the Set would have had a hard time finding duplicates in the first place.
How likely is a setup where `channelTypeRegistry.getChannelTypes(locale)` returns duplicate channelTypes?

sjka: You would need to have some other severe issues in one of the providers. And as I said: This is nothing that should be handled in the REST API!

htreu: agreed.

maggu2810: Why not adding a size hint?

sjka: Thought it might become boring for the JVM otherwise 😉 

Thanks for spotting this, I fixed it.

PR #3229, opened 2017-04-11
enable feature verification

maggu2810:
Prerequisite: #3226


maggu2810:
To run the verification we need to resolve all artifacts used in the features at build time.
Not all artifacts are available using Maven Central and the Eclipse repository only.

I will create a fix for the broken build soon.

PR #3230, opened 2017-04-11
MDNSDiscovery and HueNupnpDiscovery must be started asynchronous

afuechsel:

### DIFF
--- a/bundles/io/org.eclipse.smarthome.io.transport.mdns/src/main/java/org/eclipse/smarthome/io/transport/mdns/discovery/MDNSDiscoveryService.java
+++ b/bundles/io/org.eclipse.smarthome.io.transport.mdns/src/main/java/org/eclipse/smarthome/io/transport/mdns/discovery/MDNSDiscoveryService.java
@@ -26,12 +26,12 @@ import org.slf4j.Logger;
 import org.slf4j.LoggerFactory;
 
 /**
- * This is a {@link DiscoveryService} implementation, which can find mDNS services in the network.
- * Support for further devices can be added by implementing and registering a {@link MDNSDiscoveryParticipant}.
+ * This is a {@link DiscoveryService} implementation, which can find mDNS services in the network. Support for further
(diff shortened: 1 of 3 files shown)


bundles/io/org.eclipse.smarthome.io.transport.mdns/src/main/java/org/eclipse/smarthome/io/transport/mdns/discovery/MDNSDiscoveryService.java
```diff
@@ -62,12 +62,6 @@ public void unsetMDNSClient(MDNSClient mdnsClient) {
 
     @Override
     protected void startBackgroundDiscovery() {
-        scheduler.schedule(new Runnable() {
```
kaikreuzer: Why doesn't the background discovery do an initial scan anymore?

afuechsel: AFAIK the scan is running in the background (like in UPnPDiscovery), we only need to register listeners. MdnsClient.list() would not initiate a scan but just collect the service descriptors of all different listed services already found. Why should we do that?

kaikreuzer: The initial scan imho makes a lot of sense as you otherwise simply miss all the devices in the network that are already there. It is nothing expensive and it will bring the best result. "background discovery" simply means that the service does its best to find devices by itself, without waiting for the user to trigger a scan.

> MdnsClient.list() would not initiate a scan

That's why the code executes a `startScan();`.

afuechsel: After several discussions I would agree. Added startScan() to startBackgroundDiscovery().

PR #3234, opened 2017-04-12
Closes #2718 - Add createTimerWithArgument to ESH

### LINKED ISSUE #2718: take over createTimerWithArgument functionality from openhab
In Openhab, ScriptExecution.java was updated like [this](https://github.com/openhab/openhab/pull/1436/files) in 2014 to allow an extension to the createTimer syntax that allows a parameter to be supplied to the timer function. This was apparently not fed back to the equivalent smarthome source [file](https://github.com/eclipse/smarthome/blob/master/bundles/model/org.eclipse.smarthome.model.script/src/org/eclipse/smarthome/model/script/actions/ScriptExecution.java#L83). The background to the issue are shown [here](https://github.com/openhab/openhab/pull/1436)

DigitalBites:
Ported from PR 1436 on openhab1-addons.This will add the createTimerWithArgument to be used in ESH and downstream in OH2.

@berndpfrommer - your code lives on!

### DIFF
--- a/bundles/model/org.eclipse.smarthome.model.script/src/org/eclipse/smarthome/model/script/actions/ScriptExecution.java
+++ b/bundles/model/org.eclipse.smarthome.model.script/src/org/eclipse/smarthome/model/script/actions/ScriptExecution.java
@@ -20,6 +20,7 @@ import org.eclipse.smarthome.model.script.internal.actions.TimerExecutionJob;
 import org.eclipse.smarthome.model.script.internal.actions.TimerImpl;
 import org.eclipse.xtext.xbase.XExpression;
 import org.eclipse.xtext.xbase.lib.Procedures.Procedure0;
+import org.eclipse.xtext.xbase.lib.Procedures.Procedure1;
 import org.joda.time.base.AbstractInstant;
 import org.quartz.JobDataMap;
 import org.quartz.JobDetail;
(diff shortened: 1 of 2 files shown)


maggu2810:
Don't we need an also-by footer for berndpfrommer as he has written the code?

DigitalBites:
He contributed the version in OH1 which I heavily borrowed from :) Whatever should be done here I'm fine with. I always try to give credit where credit is due.

maggu2810:
See e.g.: https://wiki.eclipse.org/Development_Resources/Contributing_via_Git#The_Commit_Record

```
You can specify additional authors using one or more "Also-by" entries in the message footer.
```

@kaikreuzer What do you think about? I assume we normally add the original author in the commit message. Correct?

kaikreuzer:
>  I assume we normally add the original author in the commit message. Correct?

Correct, we definitely need this! So @DigitalBites, please add an according "Also-By" line in the commit message - thanks!

DigitalBites:
@kaikreuzer done!

kaikreuzer:
Thanks @DigitalBites - but "berndpfrommer@github" does not look like a valid e-mail address - please note that we will actually also need an ECA from Bernd, if he didn't yet sign any.

kaikreuzer:
I have just checked - unfortunately, @berndpfrommer does not yet seem to have an ECA :-(
Bernd, would you be ok to create an Eclipse account (if you don't have one yet) and [sign the ECA](https://www.eclipse.org/legal/ECA.php)? This would be the cleanest and best way forward for us to accept this contribution.

berndpfrommer:
It's great to see how respectful you are with respect to code ownership, but this is such a trivial piece of code, I really don't care about it. I hereby allow anybody to use whatever I may have contributed in whatever way they want. If it helps with the licensing, please remove me from the credits.

And sorry, I don't want to sign anything related to eclipse, nor get an account there. I deeply loathe the eclipse environment which I consider bloated and buggy. The fact that OH2 seems to have "eclipse" written all over it even more so than OH1 is the number one reason why I am still only working on OH1. (Why a home automation framework should be tied to an IDE has always escaped me, but I'm sure there are many good reasons).

kaikreuzer:
@berndpfrommer You are suffering a very common misconception here: When talking about "Eclipse", we do not by any means speak about the IDE, but about the Eclipse Foundation, which is the place that hosts projects (MANY not IDE related at all) and defines their governance - just like e.g. the Apache Foundation (were "Apache" is also not their HTTP server) or other organisations.
Thus OH2 is not tight in any way closer to the Eclipse IDE than OH1 - actually rather contrary, as it uses Apache Karaf as one of its core runtime components.

Being respectful with code ownership and intellectual property is an essential part of any well-managed open source project. That's why CLAs are a very common practice throughout the open source communities. If you do not want to sign the CLA with Eclipse, we will have to find some other way to get this contribution into the project.

berndpfrommer:
You shouldn't waste any more time on this. I got an eclipse account and signed the CLA, done.

kaikreuzer:
Awesome - many thanks @berndpfrommer! 🌷

PR #3236, opened 2017-04-12
Xml with ds

sjka:
When trying to find a solution for #1896, the first thing I noticed is the fact that the XML stuff related services are not registered using DS, but "manually" in the activators, using the awkward ServiceBinder class for hand-made dependency injection. I see absolutely no reason why this is not done via DS.

Hence I refactored these services and turned them into DS components. 

This however is a pretty tedious task, so I tried to find ways to avoid having to write ~10 DS XML files manually. And indeed, since Eclipse Neon there is a way of using DS Annotations in PDE. The beauty of the DS Annotations is that they are not RUNTIME annotations. So we only need to configure our tooling to generate the XML files. It still runs on OSGi 4.2, although annotations were introduced in OSGi 4.3. The runtime simply does not care. I have boiled down the details into a small sample here: https://github.com/sjka/de.sjka.sample.annotations. 

If you have any objections I could easily remove them again and add the (once generated 😄 ) XML files instead.


kaikreuzer:
Hey, that's pretty cool!
@maggu2810 What do you think about it? It seems to be a perfect way to use DS annotations without the need to switch to bnd for this (which would be a much bigger change).

bundles/pom.xml
```diff
@@ -43,7 +43,34 @@
...
+        </executions>
+      </plugin>      
     </plugins>
   </build>
 
+  <dependencies>
+    <dependency>
+      <groupId>org.apache.felix</groupId>
+      <artifactId>org.apache.felix.scr.ds-annotations</artifactId>
+      <version>1.2.8</version>
```
maggu2810: Please use a property for that version, so we don't need to sync `bundles/pom.xml` and `pom.xml`.

bundles/pom.xml
```diff
@@ -43,7 +43,34 @@
...
+          </execution>
+        </executions>
+      </plugin>      
     </plugins>
   </build>
 
+  <dependencies>
+    <dependency>
+      <groupId>org.apache.felix</groupId>
+      <artifactId>org.apache.felix.scr.ds-annotations</artifactId>
```
maggu2810: I don't think eclipse is using the old Apache Felix SCR DS-Annotations.
Can you check which one eclipse is using (the official 6.0.0 ones?)?

bundles/pom.xml
```diff
@@ -43,7 +43,34 @@
           </execution>
         </executions>
       </plugin>
+      <plugin>
+        <groupId>org.apache.felix</groupId>
+        <artifactId>maven-scr-plugin</artifactId>
+        <version>1.24.0</version>
```
maggu2810: Don't we define the plugin versions in the pluginManagement section of the root pom? So there is one source and different places don't use different stuff?

pom.xml
```diff
@@ -148,6 +148,15 @@
           <groupId>${tycho-groupid}</groupId>
           <artifactId>tycho-compiler-plugin</artifactId>
           <version>${tycho-version}</version>
+          <configuration>
+            <extraClasspathElements>
+              <extraClasspathElement>
+                <groupId>org.apache.felix</groupId>
+                <artifactId>org.apache.felix.scr.ds-annotations</artifactId>
+                <version>1.2.8</version>
```
maggu2810: Please use a property for that version, so we don't need to sync bundles/pom.xml and pom.xml.

bundles/pom.xml
sjka: Not sure if I got this question right, but did you notice the "ds-" there?
 `org.apache.felix.scr.annotations` are "the old" felix ones, while `org.apache.felix.scr.ds-annotations` actually includes the official OSGi ones (see also [here](http://felix.apache.org/documentation/subprojects/apache-felix-maven-scr-plugin/apache-felix-maven-scr-plugin-use.html)). Or did I miss anything?

sjka: PS: Just for clarification - this dependency brings the DS annotation processor (in this case for the standard OSGi annotations) into the build. It does NOT contain the DS Annotations themselves. They get resolved via the usual target platform by having the optional package dependency to `org.osgi.service.component.annotations`.

kaikreuzer:
I have created CQ 13290 for the new build-only dependencies to org.apache.felix.scr.ds-annotations and maven-scr-plugin.

kaikreuzer:
CQ has been approved, we can merge!

PR #3237, opened 2017-04-12
PaperUI Updated UI for bindings

cdjackson:
This proposes an update to the bindings list - mainly to better display the supported things which is unusable at the moment where we have a binding with many things (eg ZWave has around 450 at the moment).

Binding page essentially untouched -:
image

Clicking on the MORE button now brings up a full screen rather than then 1/3rd screen dialog -:

image

The SUPPORTED THINGS is now a filtered list -:

image

And the description is now in a collapsable box -:
image


sjka:
Indeed, it looks much cleaner with huge bindings. 
@aounhaider1 @htreu any of you willing to review?

maggu2810:
The screenshots looks good to me.

kaikreuzer:
+1

htreu:
I will take a look.

extensions/ui/org.eclipse.smarthome.ui.paper/web-src/js/controllers.configuration.bindings.js
```diff
@@ -0,0 +1,82 @@
...
+                }
+                return false
+            }
+            return true;
+        }
+    }
+
+    var bindingId = $scope.path[3];
+
+    $scope.searchText = "";
```
htreu: "" should be single quotes. this is my only comment for the code :-)

extensions/ui/org.eclipse.smarthome.ui.paper/web-src/partials/configuration.bindings.detail.html
```diff
@@ -0,0 +1,47 @@
...
+					<hr>
+				</div>
+			</div>
+		</div>
+
+	</div>
+
+
+
+
```
htreu: you may drop these empty lines.

extensions/ui/org.eclipse.smarthome.ui.paper/web-src/js/controllers.configuration.bindings.js
cdjackson: Why? Is this a project preference? If so, I guess this should be changed everywhere else in the project as I was just following the style already used given it doesn't matter either way.

htreu: I guess there is no "hard" (or 'hard' ;-)) argument for single quotes. here are mine:
* personal preference - I like single over double quotes so you can be sure to be in the js part of the project.
* standard syntax for JSON is double quoting. So even if we don´t do JSON by hand using single quotes for strings will let you type `var json = '{"key":"value"}'` w/o escaping.
* single quotes are more dominant in the js scene as facebook, google, node, npm and others prefer this style.

Regarding project preference: The is no linter in place yet so there is room for discussion. I would live to see this linted and chnage the project to use single quotes in the future.

maggu2810: @htreu I get your arguments.
Currently there is no rule about single vs. double quotes.
If we agree that single quotes should be used for JS we need to change every usage of double quotes in the current JS code base and create a separate PR.

So my suggestion:
* leave it to the author to use single or double as long as there is no rule
* create a issue or a mail on the dev list to discuss if we should add a section to our rules
* ...

WDYT?

htreu: @maggu2810 I do agree.

> leave it to the author to use single or double as long as there is no rule
the initial "should" falls back to optional then :-) its up to you @cdjackson.

> create a issue or a mail on the dev list to discuss if we should add a section to our rules
I will think of a set of javaScript rules which may apply to the PaperUI and post my ideas. We can continue discussion from there.

kaikreuzer:
Considering https://github.com/eclipse/smarthome/pull/3293#issuecomment-297421761, I would actually suggest to replace the "supported things" tab by an expansion panel.

cdjackson:
This will end up with nested expansion panes as this is done in the things list already. I decided to make the things list expandable as there are around 500 things in the zwave binding.

Maybe I'll look at more refactoring to make sure this is properly separated then it's easier to simply replace a single html file to use tabs or expansion panels (this is what is done in the things page, but I don't think I quite went that far here if I remember correctly).

cdjackson:
Updated to use expansion panels. This might require a few tweaks, but it would be good to get comments and get this merged if possible.

It pulls in a new package for the expansion panel component - this might need a CQ?
https://github.com/B-3PO/angular-material-expansion-panel

I've written this such that there's a template used for each panel, and an overall template for the container. I think this is a good design model which makes the code easy to maintain, and provides a good 'resolution' for overrides.

image

image

htreu:
@kaikreuzer do you approve the expansion panel look & feel? as @cdjackson mentioned we would then need a CQ for https://github.com/B-3PO/angular-material-expansion-panel

kaikreuzer:
> The "auto-collapse" feature feels a little strange.

I didn't try it out, but this sounds strange to me as well. My expectation would be that the "Overview" is open by default when entering the page and that it stays expanded if I expand the "supported types" panel. If this is how it works, it lgtm.

One further remark: The "Author" and "Description" font is HUGE - I would rather expect them to be sub-titles next to the real content. Maybe this can be adjusted as well?

htreu:
> One further remark: The "Author" and "Description" font is HUGE - I would rather expect them to be sub-titles next to the real content. Maybe this can be adjusted as well?

I would suggest the `<h5>` since it is the same font-size as the content text but bold.

cdjackson:
The reason I used the auto collapse is again for large bindings - the overview might be quite long, and thing types might also be VERY long. I can change it (I think), but the point of this was to make the UI more friendly and usable for larger bindings and it’s really a very common design. The issue for me was in a binding with 500+ things, this was pretty messy. Additionally, I see more panels being added in future (e.g. for configuration).

There was meant to be an option called “sticky” which meant the headers of each panel stayed at the top of the section when scrolling. This seemed like a good compromise to me, but it didn’t seem to work. I didn’t look into why - maybe there’s some other classes being employed in PaperUI that are screwing with this component.

At the end of the day it’s a personal preference, but my goal was to make this more usable and IMHO having tried both on larger bindings, this is the case. If you prefer, I can look at changing it.

I will change the titles to use md-caption unless you really prefer h5?

cdjackson:
I think I see the reason that sticky doesn’t work - it’s working, but it’s underneath the title bar I think. I’ll do an update to fix this and hopefully it should be good to go.

cdjackson:
I've not been able to work out why sticky doesn't work - it always wants to stick at the top of the page which means it's hidden under the title bar. It's using the mdSticky directive and there seems to be little info on it.

Once the CQ is done I'll maybe take another look at this, but otherwise this PR should be good to go.

kaikreuzer:
>  the overview might be quite long

The overview should be an overview and thus not be "quite long". Actually in 98% of the cases that I am aware of it is a single sentence. If you also want to cover the other 2%, we could think of checking the length of the description and collapse it, if it is longer than a thousand chars.

> as @cdjackson mentioned we would then need a CQ for https://github.com/B-3PO/angular-material-expansion-panel

I have created CQ 13437 for version 0.7.2 of the library.

kaikreuzer:
@cdjackson Good news, the CQ has been approved, so we can go ahead with that library!

cdjackson:
Ok, thanks. I think I only need to resolve the conflicts and this should be good to go.

extensions/ui/org.eclipse.smarthome.ui.paper/gulpfile.js
```diff
@@ -63,6 +64,7 @@ var paths = {
         './node_modules/masonry-layout/dist/masonry.pkgd.min.js',
         './node_modules/sprintf-js/dist/sprintf.min.js',
         './node_modules/bootstrap/dist/js/bootstrap.min.js',
+        './node_modules/angular-material-expansion-panel/dist/md-expansion-panel.min.js',
```
aounhaider1: Why do you need this here if you have already added the file to angular-bundle? Adding it to one place should be sufficient.

extensions/ui/org.eclipse.smarthome.ui.paper/web-src/index.html
```diff
@@ -72,6 +73,7 @@ <h1 flex="60">
 	<!-- TODO: Create custom variant to save space -->
 	<script src="js/tinycolor.js"></script>
 	<script src="js/masonry.pkgd.min.js"></script>
+	<script src="js/md-expansion-panel.min.js"></script>
```
aounhaider1: You can remove this if you add the md-expansion-panel to angular-bundle.

extensions/ui/org.eclipse.smarthome.ui.paper/web-src/partials/configuration.binding.thingtypes.html
```diff
@@ -0,0 +1,27 @@
...
+	<i ng-click="searchText = undefined" ng-class="{'vhidden': !searchText}" ng-style="{'font-size': '24px'}" class="material-icons">close</i>
+</div>
+
+<div class="container">
+	<div class="thing-type list-item" ng-repeat="thingType in binding.thingTypes | filter:filterItems(['uid','label'], searchText)">
+		<div ng-init="thingType.collapsed=true" class="thing container">
+			<div>
+				<h3>{{thingType.label}}</h3>
+				<div ng-if="thingType.description != null && thingType.description.length > 0">
+					<i ng-show="!thingType.collapsed" ng-click="thingType.collapsed=true" style="font-size: 22px; float: right;" class="md-raised material-icons">unfold_less</i> <i ng-show="thingType.collapsed" ng-click="thingType.collapsed=false" style="font-size: 22px; float: right; color: grey;" class="md-raised material-icons">unfold_more</i>
```
aounhaider1: If possible use the pointer cursor. It is already in the css styles so adding .pointer to the icon class will sufficient

aounhaider1: Add a little margin to the right. The icon is too far on the right side.

kaikreuzer:
@cdjackson You introduced a merge commit in your branch - better rebase/squash your stuff once you have incorporated the review feedback.

cdjackson:
> You introduced a merge commit in your branch

I just used the github 'resolve merge errors' button - I thought you used the squash feature when merging the PR and it didn't matter now if there were merge commits? Is that not correct? (just for my understanding).

kaikreuzer:
It left a commit in your branch which is not signed off and in such cases I am not allowed to do a squashing.
In any case, merge commits are always ugly - you should always better rebase your branch.

kaikreuzer:
@cdjackson Do you still have this on your list?

cdjackson:
Yes - I'm looking at it at the moment.

cdjackson:
@htreu thanks. Should be good to go. The test failure doesn't look to be related...

extensions/ui/org.eclipse.smarthome.ui.paper/web-src/partials/configuration.binding.html
```diff
@@ -0,0 +1,50 @@
+<section id="main" class="white-bg">
+	<div class="header-toolbar">
+		<md-button class="md-icon-button" aria-label="Refresh" ng-click="refresh()" aria-label="Refresh"> <md-tooltip>Refresh</md-tooltip> <i class="material-icons">refresh</i> </md-button>
```
htreu: The `BindingDetailController` does not define a `refresh` method so this button actually does nothing. You might remove it (my suggestion) or implement the refresh method to reload the binding.

extensions/ui/org.eclipse.smarthome.ui.paper/web-src/js/controllers.configuration.bindings.js
```diff
@@ -0,0 +1,87 @@
+angular.module('PaperUI.controllers.configuration.bindings', [ 'ngRoute' ]).config([ '$routeProvider', function($routeProvider) {
+    // Configure the routes for this controller
+    $routeProvider.when('/configuration/bindings', {
```
cdjackson: It makes it more modular. It means that the definition of routes sits with the implementation of the code. If someone adds more classes to implement specific additions) then it's all defined in one place.

In my opinion, it's much nicer to keep the code and it's configuration together rather than scattering it across multiple files.

cdjackson:
> Unfortunately the test failure seems to be related.

But the tests haven't even been run. The failure was in the models bundle and PaperUI isn't even being compiled. Or am I reading the report wrong?

Since this PR only changes PaperUI and not the models bundle, I'm not sure how it's related - maybe you have a better view than me, or maybe I misinterpret the report?

cdjackson:
So I closed and re-opened this to kick off the tests again, and now it's passing, so I don't think it's at all related to this PR given the PaperUI bundle was never compiled in the original failure.

extensions/ui/org.eclipse.smarthome.ui.paper/tests/configuration.binding.test.js
```diff
@@ -0,0 +1,71 @@
+describe('module PaperUI.controllers.configuration.bindings', function() {
+    beforeEach(function() {
+        module('PaperUI');
+    });
+    var restConfig;
```
cdjackson: Are you sure? Doesn't this run before all tests so we don't need to explicitly call it before the test, right?

htreu:
@cdjackson my bad. I was talking about the JavaScript tests. I didnt even had a look at travis. I will check again.

extensions/ui/org.eclipse.smarthome.ui.paper/tests/configuration.binding.test.js
htreu: I´m only talking about `var restConfig` since it is only assigned in the nested `beforeEach` block in line 28. So the definition can be done in line 24 which will narrow down the scope of `restConfig`.

extensions/ui/org.eclipse.smarthome.ui.paper/web-src/js/controllers.configuration.bindings.js
htreu: I would make the argument the other way round: Defining the routes with the controller code binds the controller to the specific URL. You can not use the same controller (code) in a different situation behind another route. Imho the routes should be defined by the application which ties all the controllers together.

cdjackson:
I thought the js tests were passing?

extensions/ui/org.eclipse.smarthome.ui.paper/web-src/js/controllers.configuration.bindings.js
cdjackson: Ok, you will make this difficult I think to add routes. On other projects I've worked on including the angular UI project, the routea are with the implementation controller. 
If you want all routes defined in the init block of the app controller then it will not be possible to add fragments with new conteollers since this block can't be easily changed.

htreu: Thats a good argument, lets see how it feels when we start changing the other controller implementations too. thanks for the discussion!

htreu:
Yes, its good to go. Do you want to change https://github.com/eclipse/smarthome/pull/3237#discussion_r122695204 before the merge?

cdjackson:
Yep - I'll take a look at this tonight. Thanks for the review.

cdjackson:
I've updated the tests. Let me know if you have further comments.

kaikreuzer:
Latest screenshot:

screen shot

cdjackson:
> "Binding Overview" title is imho irrelevant, I wouldn't even put that part into a expandable - it should simply be the header of the page.

Sorry - can you explain what this means? You want this whole panel removed (ie the author and description)? I think not, but I don't know what you want instead of the "Binding Overview"?

> Superfluous "Author" and "Description" labels that are not even using a different font/style then the rest

So you want to just have the author and description without the labels? These heading DO use different classes - they use the recommended md-caption class to define a caption. Maybe the rendering in PaperUI doesn't distinguish well, but that's another issue.

> The alignment looks broken.

All the different classes are interacting with each other so this is actually not simple :(


> The search field takes a lot of space and is rather annoying especially if there is just one thing as in the example.

What do you suggest? I could maybe remove it is the number of thing types is below (let's say) 10?

> Altogether, I would hope for improvements in UI design

Sorry - I thought it was agreed that this change to the expandable panel was considered an improvement. I'm happy to make changes, but don't want to waste my time if this isn't heading in the right direction for you

> P.S.: I haven't yet checked how a configurable binding is actually rendered...

It's currently not changed at all - it still uses the dialogue at the moment. I would propose to move it into a panel at some stage, but earlier you asked for small steps in these changes so I didn't yet do this.

kaikreuzer:
> I thought it was agreed that this change to the expandable panel was considered an improvement. I'm happy to make changes, but don't want to waste my time if this isn't heading in the right direction for you

Hey @cdjackson, sorry, I didn't mean to sound harsh or anything!
It definitely goes in the right direction and non of your efforts were a waste of time (so please allow me to re-open this PR). I mainly wish some visual/wording improvements on it, that's all - I'll try to come up with some concrete changes that we can then discuss whether you agree with them!

cdjackson:
Oops - sorry, I hadn’t meant to close the PR - just post the questions…

PR #3240, opened 2017-04-13
RunRule with passing context

BenediktNiehues:


sjka:
@danchom would you want to review this PR?

bundles/automation/org.eclipse.smarthome.automation.core/src/main/java/org/eclipse/smarthome/automation/core/internal/RuleEngine.java
```diff
@@ -860,6 +860,9 @@ protected void runNow(String ruleUID, boolean considerConditions) {
 
         try {
             clearContext(rule);
+            if (context!=null && !context.isEmpty()){
```
maggu2810: This looks like your formatter does not work. There should be a space in front of and one after the `!=`

PR #3242, opened 2017-04-13
remove package dependency to o.e.core.runtime

sjka:
...as it seems that it is not needed and makes it quite tricky to run it on
other frameworks than equinox.

Does anybody (@kaikreuzer?) remember why it's in there since the beginning? I simply couldn't figure out the reason...


teichsta:
i can confirm that the distribution (concierge) work well without this dependency

maggu2810:
> Does anybody remember why it's in there since the beginning? 

I don't know.

> makes it quite tricky to run it on other frameworks than equinox

"tricky"? Really? :wink: 

If all is working without this import, I am happy to merge this PR (the removal of unused imports is very much appreciated).

Let's wait for @kaikreuzer if he can give us some information why that import is present.

kaikreuzer:
This dependency has been [introduced exactly 5 years ago](https://github.com/openhab/openhab1-addons/commit/b0c21f24c9815226fbe3d027f310be3ddeb134f1). As far as I can see, it was only required because the code at that time used `org.eclipse.core.runtime.ListenerList`. As this is not the case anymore, I would assume we can safely remove it.

PR #3245, opened 2017-04-14
Remove limitation that thing config can only be updated if thing is managed

cdjackson:
This proposed to remove the restriction that thing configuration parameters can not be updated if the thing is not managed. Note that the limitation still remains for the thing itself.

Justification/use case for this: Even if a thing is not managed - ie it's configured through text files - the user may want to update configuration for the duration of the session. Currently this is impossible - configuration of the thing must be undertaken through the text file, which means the thing is reinitialised (removed, and re-added).

I don't think this has any bad side effects - I guess there's a philosophical question of "is this a good idea" in general, but I believe the use case above is valid and this is currently an issue in OH2.

### DIFF
--- a/bundles/io/org.eclipse.smarthome.io.rest.core/src/main/java/org/eclipse/smarthome/io/rest/core/thing/ThingResource.java
+++ b/bundles/io/org.eclipse.smarthome.io.rest.core/src/main/java/org/eclipse/smarthome/io/rest/core/thing/ThingResource.java
@@ -447,12 +447,11 @@ public class ThingResource implements SatisfiableRESTResource {
     @ApiOperation(value = "Updates thing's configuration.")
     @ApiResponses(value = { @ApiResponse(code = 200, message = "OK"),
             @ApiResponse(code = 400, message = "Configuration of the thing is not valid."),
-            @ApiResponse(code = 404, message = "Thing not found"),
-            @ApiResponse(code = 409, message = "Thing could not be updated. Maybe it is not managed.") })
+            @ApiResponse(code = 404, message = "Thing not found") })
     public Response updateConfiguration(@HeaderParam(HttpHeaders.ACCEPT_LANGUAGE) String language,
             @PathParam("thingUID") @ApiParam(value = "thing") String thingUID,
             @ApiParam(value = "configuration parameters") Map<String, Object> configurationParameters)
-                    throws IOException {
+            throws IOException {
         final Locale locale = LocaleUtil.getLocale(language);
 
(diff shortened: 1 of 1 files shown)


kaikreuzer:
As already mentioned by @sjka on another thread, allowing the user to change the configuration, just to notice that everything is lost after a restart simply does not make sense as it is VERY unexpected.

Although you do not specifically describe it, I assume that your use case is the device management here - so you are not talking about handler configuration that you want to change, but you want to give the user a chance to send some new configuration value to the device (which would then anyhow persist it by itself). If my assumption is right, I guess we are back at the discussion about how to [introduce a device management infrastructure](https://www.eclipse.org/forums/index.php?t=msg&th=1079343) - we should really keep those two very distinct kinds of "configuration" separate.

cdjackson:
I do agree that this MIGHT be unexpected to have the config disappear after a restart - but if someone is configuring their system manually using text files, then MAYBE they should understand this? I did consider looking at adding a flag into the thing to indicate to the UI if the thing is managed or not - I think this would be useful as then the UI could present the information a bit more usefully. Currently the user just gets ERROR CONFLICT - this is also unexpected and doesn’t really mean anything either.

So, for me, I would propose to solve this with a better presentation in the UI that implements this PR, but also implements a flag in the thing indicating that the thing is not managed, and the UI can then present the user with a “Config updated, but warning, this will not be persisted through a restart”.

However, as discussed in the OH forum, you are right in that this particular issue came up for configuring a device, so yes, we are back to that discussion ;). However, I do believe that this restriction should be removed to allow users to temporarily reconfigure their things as this is not possible within the current system without reinitialising the thing handler (right?).

cdjackson:
I've created a proposal to add a "managed" flag to the thing DTO (See #3289). This should allow a UI to properly present the interface to the user so that there is no confusion and the UI can properly inform users.

I think that these two PRs would allow a more joined up UX.

lolodomo:
Now that we have the new flag, I think we could allow the update of thing configuration if Paper UI clearly notice the user that the change will be not persisted.

kaikreuzer:
> I think we could allow the update of thing configuration if Paper UI clearly notice the user that the change will be not persisted.

No, I don't agree - what we really need is a way to allow posting device configuration, which does not at all require a local persistence (as it is not a handler configuration). There are other substantial differences in the requirements of such a device management infrastructure, that's why I suggest to continue the discussion on https://github.com/eclipse/smarthome/issues/3484 and come up with a clean architecture for this feature.

PR #3193, opened 2017-03-30
PaperUI: Install extensions by dropping url

aounhaider1:
Allows installing extensions by dropping url from eclipse marketplace. Should be merged after support is added to server side for installing extensions from url.

Signed-off-by: Aoun Bukhari <bukhari@itemis.de>

PR #3258, opened 2017-04-18
Support installing extensions from eclipse marketplace URL

### LINKED ISSUE #3232: Support installing extensions from eclipse marketplace URL
We already can install extensions and rule templates by id in ESH using rest endpoint. Having another rest endpoint, that supports installation of extensions directly from the marketplace URL, can be a useful feature IMO.

htreu:
Paper UI changes to support installing extensions by URL drag-n-drop:

* Install extension by dropping url
* Removes unused code + passes correct url

Also-by: Aoun Bukhari <bukhari@itemis.de>

superseds PR #3193 
fixes #3232


extensions/ui/org.eclipse.smarthome.ui.paper/web-src/partials/extensions.html
```diff
@@ -1,4 +1,4 @@
-<section id="main" class="white-bg extensions header-tabs">
+<section id="main" class="white-bg extensions header-tabs" droppable>
```
hakan42: should the "droppable" not be part of the class?

htreu: the attribute droppable is a directive defined in `extensions/ui/org.eclipse.smarthome.ui.paper/web-src/js/controllers.extension.js`. The directive adds drag-n-drop support to the whole `<section>` element.

htreu:
This PR needs review at least on the backend side.

bundles/io/org.eclipse.smarthome.io.rest.core/src/main/java/org/eclipse/smarthome/io/rest/core/extensions/ExtensionResource.java
```diff
@@ -149,6 +153,25 @@ public void run() {
...
+    @ApiOperation(value = "Installs the extension from the given URL.")
+    @ApiResponses(value = { @ApiResponse(code = 200, message = "OK"),
+            @ApiResponse(code = 400, message = "The given URL is malformed or not valid.") })
+    public Response installExtensionByURL(
+            final @PathParam("url") @ApiParam(value = "extension install URL", required = true) String url) {
+        try {
+            URI extensionURI = new URI(url);
+            String extensionId = getExtensionId(extensionURI);
+            installExtension(extensionId);
+        } catch (Exception e) {
```
maggu2810: Which exception could be thrown by the above code? Also non-runtime ones?

htreu: yes, `new URI(url)` will throw an URISyntaxException in case the given string violates RFC 2396. This was meant to be reported by `given URL is malformed`.

maggu2810: So, why catching all exceptions and not the one (`URISyntaxException`) that is thrown (IMHO it is a little bit a difference between checked and unchecked ones, if you want to ensure the caller does never see a strack trace, you could catch all uncecked one, but checked one should be catched as needed only)?

There are exceptions that needs to be handled differently (e.g. InterruptedException) in the most cases. If you catch all exceptions this part of code never gets identified by the IDE to be changed if in the call tree other checked exceptions are added.

I assume this is not only a personal point of view but a "common quality topic", see e.g. http://checkstyle.sourceforge.net/config_coding.html#IllegalCatch

htreu:
* Bugs fixed
* Unit tests with Mockito

bundles/io/org.eclipse.smarthome.io.rest.core/src/main/java/org/eclipse/smarthome/io/rest/core/extensions/ExtensionResource.java
htreu: Yes, good point. There is only one other point where an IllegalArgumentException is thrown (see ExtensionResource#getExtensionId) when no extensionId could be parsed from any ExtensionService. I will change the try/catch to specifically deal with these exceptions.

htreu:
@kaikreuzer do we need an extra CQ for this line (new dependency to _org.hamcrest_)? https://github.com/eclipse/smarthome/pull/3258/commits/2ba1b1b1209f30517b9e313dab32cc9a154e1665#diff-1808908fa4157aaaf26cd167b91776e9R58

kaikreuzer:
Yes, I think so - I have created CQ 13333 for it.

kaikreuzer:
CQ 13333 has been approved.

targetplatform/smarthome.target
```diff
@@ -55,6 +55,7 @@
 <unit id="org.codehaus.jackson.mapper.source" version="1.6.0.v20101005-0925"/>
 <unit id="org.hamcrest.core" version="1.3.0.v201303031735"/>
 <unit id="org.hamcrest.core.source" version="1.3.0.v201303031735"/>
+<unit id="org.hamcrest" version="1.1.0.v20090501071000"/>
```
kaikreuzer: When doing changes to this file, you need to increase the sequenceNumber in line 2 above.

kaikreuzer:
Marked as API breaking as it introduces a new method on the ExtensionService interface.

PR #3262, opened 2017-04-18
Avoid having no value displayed in UIs for auto-created string items (fix #3248)

### LINKED ISSUE #3248: String Item auto creation: missing [%s] in label => no display in UI
When you use channel auto linking, string items are created with a default label that does not contain a "[%s]" to display the item value.
In case the linked channel has no pattern defined, the result is that the item value is not displayed un UI when opening a page containing such a string item !
When the channel value is updated, Basic UI then updates the item value.

If I add a pattern "%s" to the channel, of course the problem is solved. But I suppose it is not mandatory to set a pattern on a "string" channel. The Sonos binding defines no pattern for channels for example.

If the item had a label with "[%s]" in its label, the problem would be solved.

So would it be ok to change automatic creation of string items in such a way ?

PS: I have not tested with channels linked to number items; maybe the problem is identical.

lolodomo:

### DIFF
--- a/bundles/ui/org.eclipse.smarthome.ui/src/main/java/org/eclipse/smarthome/ui/internal/items/ItemUIRegistryImpl.java
+++ b/bundles/ui/org.eclipse.smarthome.ui/src/main/java/org/eclipse/smarthome/ui/internal/items/ItemUIRegistryImpl.java
@@ -291,6 +291,9 @@ public class ItemUIRegistryImpl implements ItemUIRegistry {
                         if (pattern != null) {
                             label = label + " [" + pattern + "]";
                         }
+                    } else if (item instanceof StringItem) {
+                        // We consider a default pattern for string items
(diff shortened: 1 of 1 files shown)


kaikreuzer:
Hm, I think we should have a close look if this does not have any unexpected side-effects. I actually just came across a bug where there are already "too many" states shown, see the first floor on our demo sitemap: 

screen shot 2017-04-19 at 09 07 17

So my point is that there are many valid use cases where we do not want the state to be provided/displayed; we should make sure that this is not broken.

lolodomo:
I think we should rather fix the auto creation of items when auto link mode is enabled.
So I close this PR.

lolodomo:
@kaikreuzer : please note that the problem you mentionned has nothing to do with my PR. I just installed snapshot 889 in my production environment and I can reproduce the problem in Basic UI. I open a new issue.

PR #3263, opened 2017-04-18
Sonos binding: don't download too big or unknown sized cover arts

lolodomo:
If the cover art size is unknown or bigger than 500000 bytes, it is not downloaded and the image channel is set to UNDEF.


lolodomo:
@kaikreuzer : I suddenly thought that your use case with an audio file provided as cover art URL may be normal when the covert art is encoded inside the audio file. I will try to test this case.
My PR will already handle this case. But I could add another test, checking that the MIME type starts with "image".

kaikreuzer:
> when the covert art is encoded inside the audio file

That might indeed be the case - I don't have my setup available to check, if this was the case, so if you could try that out, it would be awesome!

kaikreuzer:
@lolodomo Did you manage to do some further testing/analysis on this?

lolodomo:
I have not found such files in my personal library. So I was not able to reproduce your problem.
My change is working, I tested it by setting a small threshold.

Issue #3223, opened 2017-04-10
[Test Failure] FSInternetRadioHandlerOSGiTest


sjka:
In a [Travis CI build](https://travis-ci.org/eclipse/smarthome/builds/220120289) of #3222, there was a weird problem in the `verify the volume is updated through the CHANNEL_VOLUME_PERCENT` of `FSInternetRadioHandlerOSGiTest`:

```
verify the volume is updated through the CHANNEL_VOLUME_PERCENT(org.eclipse.smarthome.binding.fsinternetradio.test.FSInternetRadioHandlerOSGiTest)  Time elapsed: 11.163 sec  <<< FAILURE!
java.lang.AssertionError: The item's state was not updated correctly in attempt to use the INCREASE command
Expected: is <1>
     but: was <2>
	at org.hamcrest.MatcherAssert.assertThat(MatcherAssert.java:20)
	at org.junit.Assert.assertThat(Assert.java:865)
	at org.junit.Assert$assertThat$0.callStatic(Unknown Source)
	at org.eclipse.smarthome.binding.fsinternetradio.test.FSInternetRadioHandlerOSGiTest$_testChannelWithINCREASEAndDECREASECommands_closure24.doCall(FSInternetRadioHandlerOSGiTest.groovy:498)
	at org.eclipse.smarthome.binding.fsinternetradio.test.FSInternetRadioHandlerOSGiTest$_testChannelWithINCREASEAndDECREASECommands_closure24.doCall(FSInternetRadioHandlerOSGiTest.groovy)
	at sun.reflect.GeneratedMethodAccessor12.invoke(Unknown Source)
	at sun.reflect.DelegatingMethodAccessorImpl.invoke(DelegatingMethodAccessorImpl.java:43)
	at java.lang.reflect.Method.invoke(Method.java:498)
	at org.codehaus.groovy.reflection.CachedMethod.invoke(CachedMethod.java:90)
	at groovy.lang.MetaMethod.doMethodInvoke(MetaMethod.java:324)
	at org.codehaus.groovy.runtime.metaclass.ClosureMetaClass.invokeMethod(ClosureMetaClass.java:278)
	at groovy.lang.MetaClassImpl.invokeMethod(MetaClassImpl.java:1016)
	at org.codehaus.groovy.runtime.callsite.PogoMetaClassSite.call(PogoMetaClassSite.java:39)
[...]

PR #3266, opened 2017-04-19
Fix unstable FSInternerRadioHandler tests

m-memova:
See: https://github.com/eclipse/smarthome/issues/3223

The code logic in the volume tests is simplified resulting in some new test cases.
Two of the methods are synchronized on the channel uid. That was neccessary because of some thread issues leading to inconsistent behaviour of the tests.


maggu2810:
Thank you!

PR #3268, opened 2017-04-19
New Feature: Provide access to channels via OSGI console

triller-telekom:
Supports only trigger channels at the moment


bundles/core/org.eclipse.smarthome.core.thing/src/main/java/org/eclipse/smarthome/core/thing/internal/console/ChannelConsoleCommandExtension.java
```diff
@@ -0,0 +1,76 @@
...
+import org.eclipse.smarthome.io.console.Console;
+import org.eclipse.smarthome.io.console.extensions.AbstractConsoleCommandExtension;
+
+/**
+ * Console command extension to access channels
+ *
+ * @author Stefan Triller - Initial contribution
+ *
+ */
+public class ChannelConsoleCommandExtension extends AbstractConsoleCommandExtension {
```
kaikreuzer: As SUBCMD_TRIGGER is "trigger", shouldn't it be rather a TriggerCommandExtension?

triller-telekom: This command extension was a starting point for other commands that can be implemented for channels, hence the subcommand trigger, which is implemented now. In the future others can use the class to add commands.

maggu2810: ... and the command itself is "channels" as shown in the super call at the default constructor. Correct?

triller-telekom: You are right "channel" makes more sense, since we only trigger ONE channel with ONE event, so I changed it to "channel" instead of "channels" and also corrected the usage message to show "event" instead of "command"

maggu2810: It seems you misinterpret my comment.
My intention was to state that the class name `ChannelConsoleCommandExtension` is (at least for me) okay, because the class handles a command `channel(s)` that itself is able to support multiple sub-commands. that `trigger` is the only sub command of the `channel(s)` command should not matter.

triller-telekom: I also briefly discussed this with kai this morning and he said that the only useful (non-debugging) feature   for channels is to trigger it. Thats why I thought I rename the class to `TriggerChannelConsoleCommandExtension`, since I only use it for triggering an event on a channel and there will not be any further subcommand in the future.

kaikreuzer: >  the class handles a command channel(s) that itself is able to support multiple sub-commands

@maggu2810 Would you see any other relevant sub-commands? Sending state updates and commands is something that should be done via items only (so that all "intermediate" logic within the framework is always respected). The other stuff you do with channels is to link them, to add or remove them from a thing, etc., but all of that is covered already in other commands.

maggu2810: > and there will not be any further subcommand in the future

Hm, okay. Then I don't see any use case of a sub command at all. If there will be never any other sub command for channels, then we could also use a command e.g. `channel-trigger` without any subcommand.
But I don't care much.
Whatever you want guys :wink:

kaikreuzer: >  the class name ChannelConsoleCommandExtension is okay

With the current setup, which results in commands like "channel trigger ...." I agree that this class name is actually correct.

kaikreuzer: > Whatever you want guys 😉

Just also discussed with @sjka and we came to the conclusion that it probably makes most sense to move this sub-command to things, so that we end up with "`smarthome:things trigger <channelUID> [<event>]`".

maggu2810: > probably makes most sense to move this sub-command to things

Ok for me

bundles/core/org.eclipse.smarthome.core.thing/src/main/java/org/eclipse/smarthome/core/thing/internal/console/ThingConsoleCommandExtension.java
```diff
@@ -26,16 +29,19 @@
...
  * @author Dennis Nobel - Initial contribution
  * @author Thomas Höfer - Added localization of thing status
+ * @author Stefan Triller - Added trigger channel command
  */
 public class ThingConsoleCommandExtension extends AbstractConsoleCommandExtension {
 
     private static final String SUBCMD_LIST = "list";
     private static final String SUBCMD_CLEAR = "clear";
     private static final String SUBCMD_REMOVE = "remove";
+    private static final String SUBCMD_TRIGGER_CHANNEL = "trigger-channel";
```
kaikreuzer: as mentioned in https://github.com/eclipse/smarthome/pull/3268#discussion_r112394362, I would think that "trigger" would suffice here.

PR #3271, opened 2017-04-20
Fix renaming trigger-channel compilation error

triller-telekom:


kaikreuzer:
Seems I was to quick with the merging...

PR #3273, opened 2017-04-20
added the model.core dependency to o.e.core.runtime again

sjka:
...as we found out the reason why it was in there: the MWE2 workflow
needs it during generation of the models. Therefore I made it optional,
so that it still works on concierge etc.

relates to #3242


kaikreuzer:
Ah, that's why I failed to run the code generator recently :-)

PR #3274, opened 2017-04-20
Introduce function NONE for item groups

### LINKED ISSUE #3161: Group Item Does Not Have Valid Defaults For Group:Dimmer and Group:Switch
The Item syntax in `Items.xtext` clearly indicates that the `ModelGroupFunction` is optional, even if the `ModelItemType` is specified. However, at least with `Group:Dimmer` and `Group:Switch` (without a `ModelGroupFunction`), declarations like
```
Group:Dimmer   bedroom_dimmers
Group:Switch   bedroom_switches
```
result in an error 
```
[ERROR] [arthome.core.items.dto.ItemDTOMapper] - Group function 'AND' requires two arguments. Using Equality instead.
```
This is in conflict with the openHAB documentation at http://docs.openhab.org/configuration/items.html#groups

> By default, if no function is provided to the group, the Group uses OR.

Pretty clearly, that's wrong, between the error message, and it not working.

Expected Behavior:

An appropriate default would be implemented for the impacted `ModelItemTypes`

---

The reason to support a group qualified by `ModelItemType` is at least three-fold:
* Clarify what can and what shouldn't be added to the group
* Provide at least Designer, if not run-time protection of adding the "wrong" kind of items to the group
* Narrow choices within Designer's pickers

I'm not terribly concerned as to _what_ the default is. I can always specify it if I'm using the group for more than just collecting it's elements for ease of management or for iterating over them.

See further https://community.openhab.org/t/items-bug-group-function-and-requires-two-arguments-using-equality-instead/25268

triller-telekom:
Since ModelGroupFunction is defined as enum in xtext there will always be a function assigned to the group, to check if there is no function we can now use the new element NONE

Fixes #3161

### DIFF
--- a/bundles/model/org.eclipse.smarthome.model.item/src/org/eclipse/smarthome/model/Items.xtext
+++ b/bundles/model/org.eclipse.smarthome.model.item/src/org/eclipse/smarthome/model/Items.xtext
@@ -27,7 +27,7 @@ ModelGroupItem :
 ;
 
 enum ModelGroupFunction :
-	AND='AND' | OR='OR' | NAND='NAND' | NOR='NOR' | AVG='AVG' | SUM='SUM' | MAX='MAX' | MIN='MIN' | COUNT='COUNT'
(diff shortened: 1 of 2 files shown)


kaikreuzer:
Cool, thanks for figuring this out!

PR #3277, opened 2017-04-20
Print warning instead of error if config file could not be read

### LINKED ISSUE #3192: Configuration reload fails two times before successful reload
Hi,

when I edit and save a config file inside the conf folder I see two IOExceptions before the reload is successful.

```java
2017-03-30 13:59:58.673 [ERROR] [.core.internal.folder.FolderObserver] - Error handling update of file '/var/packages/openHAB2/target/src/conf/sitemaps/admin.sitemap': File '/var/packages/openHAB2/target/src/con
f/sitemaps/admin.sitemap' cannot be read.
java.io.IOException: File '/var/packages/openHAB2/target/src/conf/sitemaps/admin.sitemap' cannot be read
    at org.apache.commons.io.FileUtils.openInputStream(FileUtils.java:251)[30:org.apache.commons.io:2.2.0]
    at org.eclipse.smarthome.model.core.internal.folder.FolderObserver.checkFile(FolderObserver.java:239)[123:org.eclipse.smarthome.model.core:0.9.0.201703201701]
    at org.eclipse.smarthome.model.core.internal.folder.FolderObserver.processWatchEvent(FolderObserver.java:308)[123:org.eclipse.smarthome.model.core:0.9.0.201703201701]
    at org.eclipse.smarthome.core.service.WatchQueueReader.run(WatchQueueReader.java:209)[99:org.eclipse.smarthome.core:0.9.0.201703201701]
    at java.lang.Thread.run(Thread.java:745)[:1.8.0_121]
2017-03-30 13:59:58.675 [ERROR] [.core.internal.folder.FolderObserver] - Error handling update of file '/var/packages/openHAB2/target/src/conf/sitemaps/admin.sitemap': File '/var/packages/openHAB2/target/src/con
f/sitemaps/admin.sitemap' cannot be read.
java.io.IOException: File '/var/packages/openHAB2/target/src/conf/sitemaps/admin.sitemap' cannot be read
    at org.apache.commons.io.FileUtils.openInputStream(FileUtils.java:251)[30:org.apache.commons.io:2.2.0]
    at org.eclipse.smarthome.model.core.internal.folder.FolderObserver.checkFile(FolderObserver.java:239)[123:org.eclipse.smarthome.model.core:0.9.0.201703201701]
    at org.eclipse.smarthome.model.core.internal.folder.FolderObserver.processWatchEvent(FolderObserver.java:308)[123:org.eclipse.smarthome.model.core:0.9.0.201703201701]
    at org.eclipse.smarthome.core.service.WatchQueueReader.run(WatchQueueReader.java:209)[99:org.eclipse.smarthome.core:0.9.0.201703201701]
    at java.lang.Thread.run(Thread.java:745)[:1.8.0_121]
2017-03-30 13:59:58.971 [INFO ] [el.core.internal.ModelRepositoryImpl] - Loading model 'admin.sitemap'
2017-03-30 13:59:58.988 [INFO ] [el.core.internal.ModelRepositoryImpl] - Refreshing model 'admin.sitemap'
```
I am editing the files directly from command line using vi.

Regards
Ingo

triller-telekom:
Fixes #3192 Well maybe not a real fix, but all we can do about it. We omit the stacktrace and just print a warning instead of an error.

### DIFF
--- a/bundles/model/org.eclipse.smarthome.model.core/src/main/java/org/eclipse/smarthome/model/core/internal/folder/FolderObserver.java
+++ b/bundles/model/org.eclipse.smarthome.model.core/src/main/java/org/eclipse/smarthome/model/core/internal/folder/FolderObserver.java
@@ -238,6 +238,9 @@ public class FolderObserver extends AbstractWatchService implements ManagedServi
                         if (parsers.contains(getExtension(file.getName()))) {
                             try (FileInputStream inputStream = FileUtils.openInputStream(file)) {
                                 modelRepo.addOrRefreshModel(file.getName(), inputStream);
+                            } catch (Exception e) {
+                                LoggerFactory.getLogger(FolderObserver.class)
+                                        .warn("Error while opening file during update: {}", file.getAbsolutePath());
                             }
                         } else {
                             ignoredFiles.add(file);
--- a/bundles/model/org.eclipse.smarthome.model.core/src/main/java/org/eclipse/smarthome/model/core/internal/folder/FolderObserver.java
+++ b/bundles/model/org.eclipse.smarthome.model.core/src/main/java/org/eclipse/smarthome/model/core/internal/folder/FolderObserver.java
@@ -12,6 +12,7 @@ import static java.nio.file.StandardWatchEventKinds.*;
 import java.io.File;
 import java.io.FileInputStream;
 import java.io.FilenameFilter;
+import java.io.IOException;
 import java.nio.file.Path;
 import java.nio.file.WatchEvent;
 import java.nio.file.WatchEvent.Kind;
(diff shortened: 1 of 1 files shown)


bundles/model/org.eclipse.smarthome.model.core/src/main/java/org/eclipse/smarthome/model/core/internal/folder/FolderObserver.java
```diff
@@ -238,6 +238,9 @@ private static void checkFile(final ModelRepository modelRepo, final File file,
                         if (parsers.contains(getExtension(file.getName()))) {
                             try (FileInputStream inputStream = FileUtils.openInputStream(file)) {
                                 modelRepo.addOrRefreshModel(file.getName(), inputStream);
+                            } catch (Exception e) {
```
sjka: with this, you are catching away all exceptions from [below](https://github.com/eclipse/smarthome/pull/3277/files#diff-06b195932ce2dfc1602326f23f1a79d2R252) and shield them with a very generic error log. Maybe it would make sense to keep the original catch block, reduce the severity to warn and avoid printing the stacktrace there. Otherwise I'd suggest to simply add a catch of IOException to the existing catch block and handle only those like that.

triller-telekom: I rather go with your last option. I didn't want to change the original error to warn, but handle this specific case differently. That's why I added the catch to the existing "inner try".
Your suggestion with catching an IOException  at the outer catch makes sense to me as it keeps the original error message and just prints warnings in the case of an IO error.

Issue #3270, opened 2017-04-20
Some issues with the archetype


martinvw:
There are some issues with the archetype, I'm planing on fixing them, but if there are others please let me know:

- I see that the Handler example contains a mix of tabs and spaces
- I see that the Handler example contains an outdated copyright header
- The xml files in the ESH-INF directory are not indented with tabs but with spaces, I believe the XML files should be indented with tabs.
- the build.properties is missing the about.html
- the about.html file is missing

kaikreuzer:
> the build.properties is missing the about.html

Please note that the about.html needs to be configurable as the [openHAB 2 script](https://github.com/openhab/openhab2-addons/blob/master/addons/binding/create_openhab_binding_skeleton.sh) requires a different version than the ESH one.

martinvw:
Thanks for the reminder!

martinvw:
How are you about returning null, I would really like the `HandlerFactory.createHandler` to return an exception instead of null. Why?

http://www.yegor256.com/2014/05/13/why-null-is-bad.html

Another reason is that the case always means a programming error, because the framework had the responsibility to check `supportsThingType` and if that returns `true`  while `createHandler` can't handle it we have an issue.

martinvw:
Should the `about.html` contain the (dynamic) current date?

martinvw:
Intellij marks this part of `pom.xml` unneeded, because it's inherited.
```xml
<groupId>org.openhab.binding</groupId>
``` 

Can I remove it

kaikreuzer:
> How are you about returning null, I would really like the HandlerFactory.createHandler to return an exception instead of null.

I would move this discussion out from this issue since if we decide to change the current implementation, we should also address existing handler factories accordingly.

> The date in the about.html files must be changed to the current.

Where did you find this sentence? For the [Eclipse about.html](https://www.eclipse.org/legal/epl/about.php), it is actually June 5, 2006 as a fixed date. Afaik, it is the date of this about text has been last modified. Only the "Third-Party Content About File Template" asks people to actually update the date (as they also modify the content of the file).

> Can I remove it

yes

martinvw:
Makes sense to move out the discussion :-)

about the date:

[openHAB B.2](http://docs.openhab.org/developers/development/guidelines.html#b-osgi-bundles)

[...]

PR #3278, opened 2017-04-20
Worked on a set of minor issues in the archetype

martinvw:
Worked on a set of minor issues in the archetype:

- XML files should be formatted with tabs
- Files should have the proper copyright headers
- Java files should be formatted with spaces and follow the formatter guidelines
- Loggers should be ~~static~~ *final*
- The `BINDING_ID` is only used in this class in all (most) all bindings make it private by default
- Include the about.html in the `build.properties`
- Escape the h2-headers in the README and add it to be exported
- Include the cfg example in the generated binding
- Include the about.html in the binding (have to find out a way to override it for openHAB, maybe just a cp in the skeleton script?)
- Removed inherited `groupId` from `pom.xml`
- Also made the year of the copyright dynamic
- Made the date of the about.html the current
- Added a newline at the end of the about.html

### DIFF
--- a/tools/archetype/binding/src/main/resources/archetype-resources/ESH-INF/binding/binding.xml
+++ b/tools/archetype/binding/src/main/resources/archetype-resources/ESH-INF/binding/binding.xml
@@ -1,11 +1,11 @@
 <?xml version="1.0" encoding="UTF-8"?>
 <binding:binding id="${bindingId}"
-        xmlns:xsi="http://www.w3.org/2001/XMLSchema-instance"
-        xmlns:binding="http://eclipse.org/smarthome/schemas/binding/v1.0.0"
-        xsi:schemaLocation="http://eclipse.org/smarthome/schemas/binding/v1.0.0 http://eclipse.org/smarthome/schemas/binding-1.0.0.xsd">
+				 xmlns:xsi="http://www.w3.org/2001/XMLSchema-instance"
+				 xmlns:binding="http://eclipse.org/smarthome/schemas/binding/v1.0.0"
+				 xsi:schemaLocation="http://eclipse.org/smarthome/schemas/binding/v1.0.0 http://eclipse.org/smarthome/schemas/binding-1.0.0.xsd">
 
-    <name>${bindingIdCamelCase} Binding</name>
-    <description>This is the binding for ${bindingIdCamelCase}.</description>
-    <author>${author}</author>
+	<name>${bindingIdCamelCase} Binding</name>
+	<description>This is the binding for ${bindingIdCamelCase}.</description>
+	<author>${author}</author>
 
 </binding:binding>
--- a/tools/archetype/binding/src/main/resources/archetype-resources/ESH-INF/i18n/__bindingId___xx_XX.properties
+++ b/tools/archetype/binding/src/main/resources/archetype-resources/ESH-INF/i18n/__bindingId___xx_XX.properties
@@ -1,3 +1,5 @@
+# FIXME: please substitute the xx_XX with a proper locale, ie. de_DE
+# FIXME: please do not add the file to the repo if you add or change no content
 # binding
 binding.${bindingId}.name = <Your localized Binding name>
 binding.${bindingId}.description = <Your localized Binding description>
--- a/tools/archetype/binding/src/main/resources/archetype-resources/ESH-INF/thing/thing-types.xml
+++ b/tools/archetype/binding/src/main/resources/archetype-resources/ESH-INF/thing/thing-types.xml
@@ -1,24 +1,24 @@
 <?xml version="1.0" encoding="UTF-8"?>
 <thing:thing-descriptions bindingId="${bindingId}"
-        xmlns:xsi="http://www.w3.org/2001/XMLSchema-instance"
-        xmlns:thing="http://eclipse.org/smarthome/schemas/thing-description/v1.0.0"
-        xsi:schemaLocation="http://eclipse.org/smarthome/schemas/thing-description/v1.0.0 http://eclipse.org/smarthome/schemas/thing-description-1.0.0.xsd">
+						  xmlns:xsi="http://www.w3.org/2001/XMLSchema-instance"
+						  xmlns:thing="http://eclipse.org/smarthome/schemas/thing-description/v1.0.0"
+						  xsi:schemaLocation="http://eclipse.org/smarthome/schemas/thing-description/v1.0.0 http://eclipse.org/smarthome/schemas/thing-description-1.0.0.xsd">
 
-    <!-- Sample Thing Type -->
-    <thing-type id="sample">
-        <label>${bindingIdCamelCase} Binding Thing</label>
-        <description>Sample thing for ${bindingIdCamelCase} Binding</description>
+	<!-- Sample Thing Type -->
+	<thing-type id="sample">
+		<label>${bindingIdCamelCase} Binding Thing</label>
+		<description>Sample thing for ${bindingIdCamelCase} Binding</description>
 
-        <channels>
-            <channel id="channel1" typeId="sample-channel"/>
-        </channels>
-    </thing-type>
+		<channels>
+			<channel id="channel1" typeId="sample-channel"/>
+		</channels>
+	</thing-type>
 
-    <!-- Sample Channel Type -->
-    <channel-type id="sample-channel">
-        <item-type>${bindingId}Item</item-type>
-        <label>${bindingIdCamelCase} Binding Channel</label>
-        <description>Sample channel for ${bindingIdCamelCase} Binding</description>
-    </channel-type>
+	<!-- Sample Channel Type -->
+	<channel-type id="sample-channel">
+		<item-type>${bindingId}Item</item-type>
+		<label>${bindingIdCamelCase} Binding Channel</label>
+		<description>Sample channel for ${bindingIdCamelCase} Binding</description>
+	</channel-type>
 
 </thing:thing-descriptions>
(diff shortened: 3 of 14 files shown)


martinvw:
See #3270

maggu2810:
> Loggers should be static

Loggers should be non-static if possible.

tools/archetype/binding/src/main/resources/archetype-resources/src/main/java/handler/__bindingIdCamelCase__Handler.java
```diff
@@ -20,28 +23,28 @@
...
  * The {@link ${bindingIdCamelCase}Handler} is responsible for handling commands, which are
  * sent to one of the channels.
- * 
+ *
  * @author ${author} - Initial contribution
  */
 public class ${bindingIdCamelCase}Handler extends BaseThingHandler {
 
-    private Logger logger = LoggerFactory.getLogger(${bindingIdCamelCase}Handler.class);
+    private final Logger logger = LoggerFactory.getLogger(${bindingIdCamelCase}Handler.class);
```
maggu2810: We could think about using `LoggerFactory.getLogger(getClass());`

The main difference is if you create a sub-class.

martinvw:
Sorry, I meant should be final, as you can see from my change.

Since loggers should imho be private, the only thing getClass protects against is copy paste errors. My feeling is that a drawback is that it actually logs from the runtime class instead of the file in which the logging occurs, so it might be harder to configure and track back your log statement to the real file it came from.

tools/archetype/binding/src/main/resources/archetype-resources/src/main/java/handler/__bindingIdCamelCase__Handler.java
martinvw: As you can see at the FAQ it explicitly not mentioned, but instead they mention some other way to get the real calling class and not the runtime type:

https://www.slf4j.org/faq.html#declaration_pattern

martinvw: A quick test with the same logging statement in the same file with either `${classname}.class` or  `getClass()` and the logging statement is in the parent class (`GreetingController`) while the class actually is an `AlsoGreetingController`
```
2017-04-21 17:15:26.822  WARN 24244 --- [nio-8090-exec-1] n.m.d.rest.backend.GreetingController    : Name should be longer than 4 characters.
2017-04-21 17:15:26.823  WARN 24244 --- [nio-8090-exec-1] n.m.d.r.backend.AlsoGreetingController   : Name should be longer than 4 characters.
```

So the getClass() does not point out the correct source of the logging statement, imho it should never be used :-)

martinvw:
@sjka can you maybe review this one?

tools/archetype/binding/src/main/resources/archetype-resources/src/main/java/handler/__bindingIdCamelCase__Handler.java
maggu2810: @martinvw Thanks for the links but I assume I know that stuff already :wink: 
If written "we could think about", not that you should change it.

> imho it should never be used

Hm, currently I use it if I have an abstract base class that is used by a lot of sub class and I want to enable the logging of code that resides in the base class just for one sub class.
At the moment I don't see an anti-pattern or "never use it" but perhaps a "normally not needed".

If you have an idea how to implement the my example given above "better", feel free to send me a mail (I will be interested in). I assume there is no need to discuss that "off-topic" here.

tools/archetype/binding/src/main/resources/archetype-resources/ESH-INF/binding/binding.xml
```diff
@@ -1,11 +1,11 @@
 <?xml version="1.0" encoding="UTF-8"?>
```
sjka: To my understanding, according to https://www.eclipse.org/legal/ _all_ distributed (source) files _must_ have have a license header. So this file should have one too, or am I missing anything? (I know, we need to check the existing codebase too....)

tools/archetype/binding/src/main/resources/archetype-resources/src/main/java/handler/__bindingIdCamelCase__Handler.java
martinvw: Thanks for your time! Your usecase sounds valid, although it is really an exception, lets rephrase my statement ;-)

> In generally it should not be used, because its makes it less clear where the log statements are actually originating from. If we want to guard against copy-paste errors lets follow the manual and use: `MethodHandles.lookup().lookupClass()`

Did you have any other review comments?

tools/archetype/binding/src/main/resources/archetype-resources/ESH-INF/thing/thing-types.xml
```diff
@@ -1,24 +1,24 @@
 <?xml version="1.0" encoding="UTF-8"?>
```
sjka: Missing license header? (see above)

tools/archetype/binding/src/main/resources/archetype-resources/ESH-INF/i18n/__bindingId___xx_XX.properties
```diff
@@ -1,3 +1,5 @@
+# FIXME: please substitute the xx_XX with a proper locale, ie. de_DE
+# FIXME: please do not add the file to the repo if you add or change no content
```
sjka: 👍

tools/archetype/binding/src/main/resources/archetype-resources/OSGI-INF/__bindingIdCamelCase__HandlerFactory.xml
```diff
@@ -1,7 +1,8 @@
 <?xml version="1.0" encoding="UTF-8"?>
 <!--
 
-    Copyright (c) 2014-2017 by the respective copyright holders.
+    Copyright (c) 2010-2017 by the respective copyright holders.
```
martinvw: Yes, I forgot it :-S

martinvw:
I'll fix the missing dynamic end-year asap when I'm at my home PC, if others have an opionon about adding the license header to the other files as well, please let me know.

martinvw:
@sjka @maggu2810 @kaikreuzer I fixed the missing dynamic value, please let me know if any other changes are required.

sjka:
Okay from my side. And thanks for taking care of these improvements btw!!

Regarding the additional license headers: We don't need to discuss this here as all existing bindings would need to be fixed too. It actually makes sense to separate it.

So: LGTM

tools/archetype/binding/src/main/resources/archetype-resources/src/main/java/internal/__bindingIdCamelCase__HandlerFactory.java
```diff
@@ -19,23 +22,22 @@
...
+ * The {@link ${bindingIdCamelCase}HandlerFactory} is responsible for creating things and thing
  * handlers.
- * 
+ *
  * @author ${author} - Initial contribution
  */
 public class ${bindingIdCamelCase}HandlerFactory extends BaseThingHandlerFactory {
-    
+
     private final static Set<ThingTypeUID> SUPPORTED_THING_TYPES_UIDS = Collections.singleton(THING_TYPE_SAMPLE);
```
cweitkamp: Just a very little to point out: Should be `static final` instead.

martinvw: pitty, that I missed it, it has been merged so it would be great if you could open a new pr for this one :-)

cweitkamp: Never mind. I will change it and create a PR.

kaikreuzer:
@martinvw I just saw that this PR "obfuscated" the page https://github.com/eclipse/smarthome/blob/master/tools/archetype/binding/src/main/resources/archetype-resources/README.md a bit by replacing the # signs.

The issue is that this page is linked to from https://www.eclipse.org/smarthome/documentation/development/bindings/docs.html - so I cannot easily point people to our template anymore now. Any suggestion how to resolve this?

martinvw:
We needed them to be able to replace the placeholders. 

The alternative would be not to replace any placeholders, because right now it's only the binding id which is dynamic. 

We should disable the `filtered` for a new fileset containing the README and exclude the README from the other fileset.

martinvw:
@kaikreuzer ^^ see above

kaikreuzer:
> Include the cfg example in the generated binding

Sorry, I didn't notice that when this was merged.
What exactly is "the cfg example"? Imho, no such config should be needed and I definitely don't think that we should introduce a cfg folder in the root of the bundle source project. Could we remove that again?

martinvw:
>What exactly is "the cfg example"?

https://github.com/eclipse/smarthome/commit/dc67633758236dc3ecd8a0ee096eb6f8f528bea3#diff-e322128541de6daa620b450b1b34add2

It was already there but not actually packaged along, I assumed somebody else (you in 2015 ;-) ) decided that it should be needed. 

Just to be sure we don't need another rework on this PR, can you review all files to be there on purpose in the correct format:

https://github.com/eclipse/smarthome/tree/master/tools/archetype/binding/src/main/resources/archetype-resources

kaikreuzer:
I found quite some issues with the generated unit test fragment... Will have to find the time to come up with a PR to address those :-/

martinvw:
I'm not feeling attached to "the cfg example" I will open a PR tonight to remove it from the artifact

martinvw:
For the record, @kaikreuzer fixed it with #3666

PR #3069, opened 2017-02-24
correctly handle hidden files

kaikreuzer:
fixes #2854
#2855 didn't work in cases where a file was part of a subdirectory (since getName also returns the folders).

Signed-off-by: Kai Kreuzer <kai@openhab.org>

PR #3286, opened 2017-04-21
Ignore hidden configuration files

### LINKED ISSUE #2100: Prevent hidden files from being parsed
When using Mac OS X to edit Openhab files through an SMB share, files with a leading ._ are created by Mac OS X to store file information about the actual file in.

**Example:**
Create/edit a file called "test.item" using OS X finder application  and Mac OS X also creates a file called "._test.item".

Although these ._ files do not contain any valid information for Openhab, they are being parsed causing problems in mainly the Openhab rule engine.
This leads to errors in the Openhab log like:

2016-09-02 09:54:45.764 [ERROR] [.script.engine.ScriptExecutionThread] - Rule 'Pantry light motion': Script interpreter couldn't be obtain

2016-09-01 15:47:55.026 [WARN ] [pse.smarthome.core.items.GenericItem] - failed notifying listener '[org.eclipse.smarthome.model.rule.runtime.internal.engine.RuleEngineImpl@1ec05f5, Home_State, null]' about state update of item java.lang.NullPointerException: {}

Perhaps preventing the ._ files to be parsed would help in preventing issues caused by these files?

triller-telekom:
Fixes #2100

### DIFF
--- a/bundles/model/org.eclipse.smarthome.model.core/src/main/java/org/eclipse/smarthome/model/core/internal/folder/FolderObserver.java
+++ b/bundles/model/org.eclipse.smarthome.model.core/src/main/java/org/eclipse/smarthome/model/core/internal/folder/FolderObserver.java
@@ -311,7 +311,7 @@ public class FolderObserver extends AbstractWatchService implements ManagedServi
     @Override
     protected void processWatchEvent(WatchEvent<?> event, Kind<?> kind, Path path) {
         File toCheck = getFileByFileExtMap(folderFileExtMap, path.getFileName().toString());
-        if (toCheck != null) {
+        if (toCheck != null && !toCheck.isHidden()) {
             checkFile(modelRepo, toCheck, kind);
         }
     }
--- a/bundles/model/org.eclipse.smarthome.model.core/src/main/java/org/eclipse/smarthome/model/core/internal/folder/FolderObserver.java
+++ b/bundles/model/org.eclipse.smarthome.model.core/src/main/java/org/eclipse/smarthome/model/core/internal/folder/FolderObserver.java
@@ -236,7 +236,8 @@ public class FolderObserver extends AbstractWatchService implements ManagedServi
             try {
                 synchronized (FolderObserver.class) {
                     if ((kind == ENTRY_CREATE || kind == ENTRY_MODIFY)) {
-                        if (parsers.contains(getExtension(file.getName()))) {
+                        // we omit parsing of hidden files possibly created by editors or operating systems
+                        if (parsers.contains(getExtension(file.getName())) && !file.getName().startsWith(".")) {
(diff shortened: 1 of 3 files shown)


sjka:
Makes sense. What about poor [FolderObserverTest](https://github.com/eclipse/smarthome/blob/master/bundles/model/org.eclipse.smarthome.model.core.test/src/test/groovy/org/eclipse/smarthome/model/core/internal/folder/test/FolderObserverTest.groovy) though?

bundles/model/org.eclipse.smarthome.model.core/src/main/java/org/eclipse/smarthome/model/core/internal/folder/FolderObserver.java
```diff
@@ -236,7 +236,8 @@ private static void checkFile(final ModelRepository modelRepo, final File file,
             try {
                 synchronized (FolderObserver.class) {
                     if ((kind == ENTRY_CREATE || kind == ENTRY_MODIFY)) {
-                        if (parsers.contains(getExtension(file.getName()))) {
+                        // we omit parsing of hidden files possibly created by editors or operating systems
+                        if (parsers.contains(getExtension(file.getName())) && !file.getName().startsWith(".")) {
```
kaikreuzer: This might not be the best approach.
Does this issue actually still exist? Please note https://github.com/eclipse/smarthome/pull/3069/files, https://github.com/eclipse/smarthome/pull/3112/files and https://github.com/eclipse/smarthome/pull/3148/files, which were about the same/similar issue.

triller-telekom: Indeed this issue still exists: Before I added my change, the FolderObserver parsed my .test.items file and the items in there where added to the model and showed up in the paperui

sjka: Just one more thing that crossed my mind: Why don't you simply use `file.isHidden()`? 
Then it would not only work on operating systems, but also on Windows 😉

kaikreuzer: > Why don't you simply use file.isHidden()?

This was actually what I meant by "This might not be the best approach."

kaikreuzer: But shouldn't it be already prevented that a hidden file is passed to checkFile() in the first place?

sjka:
> What about poor FolderObserverTest though?

My question remains valid: Could you please provide a test that verifies the desired behavior actually works as expected?

PR #3288, opened 2017-04-21
Update return to 500 following config update error

cdjackson:
getThingNotFoundResponse doesn't seem an appropriate response since we already checked that the thing is found a few lines earlier. I had an error in a handler that gave thing not found and it makes it hard to trace.

I suggest to change to 500 "Internal Server Error".


bundles/io/org.eclipse.smarthome.io.rest.core/src/main/java/org/eclipse/smarthome/io/rest/core/thing/ThingResource.java
```diff
@@ -493,7 +494,7 @@ public Response updateConfiguration(@HeaderParam(HttpHeaders.ACCEPT_LANGUAGE) St
         } catch (IllegalArgumentException ex) {
             logger.info("Received HTTP PUT request for update config at '{}' for the unknown thing '{}'.",
                     uriInfo.getPath(), thingUID);
-            return getThingNotFoundResponse(thingUID);
+            return Response.status(Status.INTERNAL_SERVER_ERROR).build();
```
tomhoefer: I think the logged message should be updated as well

cdjackson:
Good point - updated...

kaikreuzer:
Hm, a server 500 error should never be an expected return code - it should only happen, if there is a real bug on the server side and in this case it should be reported as ERROR in the log as well.

Do I understand it correctly that this situation ONLY occurs if there is a real bug in an implementation? Or could it also occur in situations where wrong parameters are passed in the REST call?

cdjackson:
It's catching an exception from the handler - currently only IllegalArgumentException but I did wonder if we shouldn't catch other exceptions here and report the error back to the user?

kaikreuzer:
Yes, we should probably catch Exception in general.
I am still wondering though if we should document the 500 in Swagger - after all, this applies to ALL bugs in REST endpoints, they will always result in a 500; but this is not really a part of the API.

cdjackson:
Ok - fair point. So how about I change the exception to catch all and remove the swagger doc?

kaikreuzer:
Sounds good to me.

cdjackson:
Updated as discussed.

bundles/io/org.eclipse.smarthome.io.rest.core/src/main/java/org/eclipse/smarthome/io/rest/core/thing/ThingResource.java
```diff
@@ -490,10 +490,10 @@ public Response updateConfiguration(@HeaderParam(HttpHeaders.ACCEPT_LANGUAGE) St
             logger.debug("Config description validation exception occurred for thingUID {} - Messages: {}", thingUID,
                     ex.getValidationMessages());
             return Response.status(Status.BAD_REQUEST).entity(ex.getValidationMessages(locale)).build();
-        } catch (IllegalArgumentException ex) {
-            logger.info("Received HTTP PUT request for update config at '{}' for the unknown thing '{}'.",
-                    uriInfo.getPath(), thingUID);
-            return getThingNotFoundResponse(thingUID);
+        } catch (Exception ex) {
+            logger.info("Exception during HTTP PUT request for update config at '{}' for thing '{}': {}",
```
kaikreuzer: As suggested, better use error logging to clearly indicate that we stumbled across an implementation bug.

bundles/io/org.eclipse.smarthome.io.rest.core/src/main/java/org/eclipse/smarthome/io/rest/core/thing/ThingResource.java
```diff
@@ -490,10 +490,10 @@ public Response updateConfiguration(@HeaderParam(HttpHeaders.ACCEPT_LANGUAGE) St
...
+                    uriInfo.getPath(), thingUID, ex.getMessage());
+            return Response.status(Status.INTERNAL_SERVER_ERROR).build();
```
kaikreuzer: Might it be worthwhile to return a JSONResponse that also passes the exception message to the client?

cdjackson:
Ok, I'll fix these two issues after the other PR is merged as there will be merge conflicts that will need resolving.

kaikreuzer:
"The other" is merged :-)

cdjackson:
Updated...

kaikreuzer:
Thanks!

bundles/io/org.eclipse.smarthome.io.rest.core/src/main/java/org/eclipse/smarthome/io/rest/core/thing/ThingResource.java
```diff
@@ -490,10 +490,10 @@ public Response updateConfiguration(@HeaderParam(HttpHeaders.ACCEPT_LANGUAGE) St
             logger.debug("Config description validation exception occurred for thingUID {} - Messages: {}", thingUID,
                     ex.getValidationMessages());
             return Response.status(Status.BAD_REQUEST).entity(ex.getValidationMessages(locale)).build();
-        } catch (IllegalArgumentException ex) {
-            logger.info("Received HTTP PUT request for update config at '{}' for the unknown thing '{}'.",
-                    uriInfo.getPath(), thingUID);
-            return getThingNotFoundResponse(thingUID);
+        } catch (Exception ex) {
```
maggu2810: I am still not a fan about catching every exception. Every unchecked one is okay for such cases but to catch also every checked one is IMHO broken.

maggu2810: Thinking a little bit more about the location, I assume for a rest resource it is okay... Nothing can be done in this case.

PR #3289, opened 2017-04-22
Add 'editable' flag to enrichedThingDto

cdjackson:
This adds a 'managed' flag to the enrichedThingDTO. This allows UIs to know if a thing is managed or not, and provide an appropriate UI.
Currently, users are left to guess what responses might mean when they try and send updates to the system for unmanaged things, and to me at least, it makes more sense for the UI to be able to present an appropriate interface.

### DIFF
--- a/bundles/io/org.eclipse.smarthome.io.rest.core/src/main/java/org/eclipse/smarthome/io/rest/core/thing/EnrichedThingDTO.java
+++ b/bundles/io/org.eclipse.smarthome.io.rest.core/src/main/java/org/eclipse/smarthome/io/rest/core/thing/EnrichedThingDTO.java
@@ -21,14 +21,25 @@ import org.eclipse.smarthome.core.thing.dto.ThingDTO;
  *
  * @author Dennis Nobel - Initial contribution
  * @author Kai Kreuzer - Removed links and items
+ * @author Chris Jackson - Added 'editable' flag
  *
  */
 public class EnrichedThingDTO extends ThingDTO {
 
     public ThingStatusInfo statusInfo;
+    public boolean editable;
     // public List<EnrichedChannelDTO> channels;
 
-    public EnrichedThingDTO(ThingDTO thingDTO, ThingStatusInfo statusInfo, Map<String, Set<String>> linkedItemsMap) {
+    /**
+     * Creates an enriched thing data transfer object.
+     *
+     * @param thingDTO the base {@link ThingDTO}
+     * @param statusInfo {@link ThingStatusInfo} for this thing
+     * @param linkedItemsMap a map of linked items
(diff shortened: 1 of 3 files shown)


sjka:
I see that it would make sense for UIs to know in advance. I'm just not sure about the naming. Of course, the corresponding providers have a "Managed" in their names, but I still think this is kind of confusing.

In the end, the UI wants to know whether it can write any changes or not, so maybe `writable=true` or `read-only=false` would be better to the point of what this actually is about...?

kaikreuzer:
I am also for a name change - how about "editable"?

kaikreuzer:
Btw, @cdjackson, if you could introduce the very same on the EnrichedItemDTO as well, it would be awesome :-)

cdjackson:
I chose managed as this is how the error messages are currently written and I was trying to be consistent - eg the REST docs say -"CONFLICT, Thing could not be deleted because it's not managed.".  Personally, I'm happy with either, but I think we should be consistent and saying editable in one place, and then giving errors saying it's not managed is IMHO also confusing.

kaikreuzer:
Fully agree. I would hence also adapt the error message (which should actually not come up anymore, if we correctly adapt the UI to only try to edit entities that actually are editable.

maggu2810:
I agree with a different naming, too.

If we think about differ between thing handler configuration and thing configuration, should we also consider a difference between thing is editable and thing handler is editable?
Currently we talking about a configurable thing handler configuration.
@cdjackson Also point about a thing configuration that should be editable if the thing handler is unmanaged...

cdjackson:
Ok, so I'll change this to use the word "Editable" and will also adapt the REST doc messages (and any other messages along these lines).

cdjackson:
Updated to editable...

kaikreuzer:
> If we think about differ between thing handler configuration and thing configuration, should we also consider a difference between thing is editable and thing handler is editable?

@maggu2810 I think this is a different discussion for now, which should be continued on https://github.com/eclipse/smarthome/pull/3245#issuecomment-294284367.

maggu2810:
> @maggu2810 I think this is a different discussion for now, which should be continued on #3245 (comment).

Why? It is not about a discussion how it should be solved etc. It is about the name you have been chosen now. If we change editable to editable-handler or editable-thing later we break the API. We had have the change to chose one that could stay whatever we discuss in #3245

kaikreuzer:
> It is about the name you have been chosen now.

Ok, I see. Nonetheless I think that the "editable" is correct here as it describes the entity that it is part of - i.e. it means that you can PUT an updated version of this entity through the REST API.
The "deviceconfig-editable" won't apply to the Thing, but to its configuration (or even rather the new concept we will come up with).

PR #3206, opened 2017-04-05
Firmware update

htreu:
This provides basic functionality for firmware handling:
- REST endpoints for querying, listing and putting firmware to things
- PaperUI components to display firmware status, available updates, update progress

fixes #2606
supersedes  PR #3164

PR #3293, opened 2017-04-22
WIP: PaperUI - New thing-editor

cdjackson:
This is an early PR for visibility and comment...

This refactors the thing editor pages - currently it's only a partial change toward what I would like to do. I think the current implementation is very messy - it pops up a number of dialogs, changes screens and just feels inconsistent with the rest of the UI, and very clunky.

I've tried to consolidate the thing interface and make it more consistent with the rest of the UI using tabs.

Note that this pulls in a new library (https://github.com/betsol/angular-input-modified). This is used in HABmin to detect changes in forms and allows an easy implementation whereby we can detect changes to the configuration, and send updates only for data that has changed. 

Images....

Things list remains the same -:
image

Selecting a thing brings up a new overview page. This displays information on the thing/thing-type (basically the same as the previous thing view page). Note the toolbar has changed, and the floating action button is gone (mainly because it doesn't fit with tabs).

image

Properties is now a separate tab, but otherwise the same look as per my changes a week or so back. The previous "click to show properties" button IMHO was clunky and not inline with the feel of the UI -:

image

Channels is on a different tab, but otherwise currently the same. I do think this needs a significant overhaul though as the UI currently pops up loads of dialogs here and it (IMHO) doesn't feel nice -:

image

Likewise configuration - basically the same, but in a separate tab -:

image

The "SAVE" button in the toolbar is only active when configuration changes -:

image

I wouldn't bother reviewing the code - I'd like comments on the interface first and if it's acceptable I'll tidy things up.


htreu:
Hi Chris,
here are some comments on your new design:

* Compared to the current design I'm strongly missing the "Edit" button hovering over the menu/thing pane. From a UX perspective the last tab "Configuration" for editing the Thing does not feel right. Using the current edit button and overlaying a paper sheet for editing, then accepting the changes and displaying the Thing again is much more closely to my expectations.
* The Thing view deals with one entity of a Thing. Spreading this single entity across different tabs mixes the notion of displaying different _aspects_ of the entity vs. _navigating_ through different categories (like in the "Services" view).
* I agree with you about the clunkiness of the `click to show properties` button. Maybe we can introduce [expansion panels](https://material.io/guidelines/components/expansion-panels.html) to hide large tables in the inline Thing display. If you take a look at the expansion panels I like the separate cards a lot. This might also be an alternative to separate information in the inline display.
* In addition we will introduce firmware support for things with PR #3206. This will also introduce information which might be hidden in an expansion panel.

What do you think?

cdjackson:
Ok, I personally don't like the current system. I think popping up dialogs feels horrible - there are multiple levels of dialog and it just feels wrong to me and IMHO it doesnt feel nice on mobile systems. Eg. To link a channel, you click on the link button, this pops up a dialog asking you to link an item which pops up another dialog to ask you to create an item.

Personally, I like the feel that HABmin provides which uses extension panels for the thing editor, but I was trying to work this into the feel of PaperUI with the Tabs.

It's up to you - if this isn't something that's desirable then I'll withdraw this PR. My secondary hope was to be able to refactor PaperUI to make it more configurable so it was possible for other implementations to lever off this work - what do you think?

htreu:
I guess we can split the discussion into three topics.
* the channel linking in the current ui with popup for item selection and 2nd popup for new item:
  * I second your statements: this needs UX overhaul.
* the new design for a single Thing entity using multiple tabs:
  * all of my comments above refer to this approach. Using tabs is not generally a bad thing. I was just concerned about using it for this single entity. Tabs in general are more of a navigation component then an information group element.
* refactoring PaperUI to separate controllers and partials into components like you did in the `initial work` commit.
  * imho this is also a good thing to do.

cdjackson:
Have you looked at HABmin and how it works for thing configuration? If not I can maybe provide some screenshots to walk through the process etc to aid discussion. (this covers the first two points).

For the last point I will comment further later as I have an email at home and will write this as a separate issue tonight.

kaikreuzer:
For point (2), I would agree with @htreu that we should follow UX of Android (as the main "user" of Material Design) as closely as possible and not trying to come up with our own concepts. And from what I see from a UX pov, tabs are not the way to go when dealing with an "entity" - they are rather a concept for sorting and categorizing lists, just as we do e.g. on the extensions menu.

cdjackson:
I'm fine with using the Material guidelines and I don't think the material guidelines states not to use tabs so I don't think this proposal breaks it at all (unlike the current UI which I would say is not designed within the guide). 

From the material guidelines -:

* Tabs enable content organization at a high level, such as switching between views, data sets, or functional aspects of an app.
* Use dialogs sparingly because they are interruptive. It states that dialogs should be used for alerts, simple menus and confirmations and this is the one that I really don't like (ie the extensive use of dialogs in the current UI)


Tonight I will post a set of images of HABmin and how it looks there. It uses expansion panels rather than dialogs or tabs. We can discuss further then...

kaikreuzer:
I think we all are on the same side regarding the dialogs, this is @htreu's point (1) above.
[Expansion panels](https://material.io/guidelines/components/expansion-panels.html) also seem to be the best choice for me and I would clearly prefer them over the use of tabs for the Thing details.

cdjackson:
As mentioned previous, here's an overview of the thing editor flow in HABmin for information and ideas - I don't mean that PaperUI should look like this, just that maybe some ideas could be taken instead of the more layered approach that is currently implemented -:

It's based on expansion panels - on starting, you get the overview/description panel open -:

image

Scrolling down you see the other panels -:

image

This is a ZWave device - in HABmin the different configuration groups are split into different panels to make it clearer rather than having (potentially) hundreds of configuration parameters in a single configuration tab.

Configuration panels look as follows - this would be the same as PaperUI currently does, but I've added this here for completeness -:

image

The Save and Cancel buttons are only highlighted if there are changes made to the configuration, so it's much more obvious if you make a change -:
image

Channel configuration panel lists all the channels. There are buttons provided to add a link, or add a new item. If the channel has configuration, it's also displayed at this point (but I don't have any channels with configuration to show right now).
image

To add a new item link, there is a dialog that pops up - not disimilar to the "add new item" box in PaperUI at the moment -:
image

If you just want to link a channel to an item, then click the link button and it asks what the item is -:
image

There's only ever one level of dialogs open - not a dialog opening another dialog...


At the bottom, the "Attributes" tab displays the thing properties -:
image

kaikreuzer:
@cdjackson It sounds as if we agree on https://github.com/eclipse/smarthome/pull/3293#issuecomment-297421761 then.
I would actually suggest to do independent improvements in separate issues/PRs - this will help keeping discussions focused and allows us to quickly merge stuff that is agreed upon.

cdjackson:
> @cdjackson It sounds as if we agree on #3293 (comment) then.

Ok, so to be clear I'll swap this around to use expansion panels rather than tabs? Right?


> I would actually suggest to do independent improvements in separate issues/PR

It was always my plan to do this progressively - hence why it's not included in this PR ;)

htreu:
> Ok, so to be clear I'll swap this around to use expansion panels rather than tabs? Right?

It would be awesome to see expansion panels here, just like you did with the new bindings view #3237. Again `angular-material-expansion-panel` would need a CQ.

cdjackson:
Yes - I'll update this PR after the other is merged as I want to make sure the CQ is done first.  The way I've structured the code is very simple to change this now.

htreu:
Hi @cdjackson, do you still plan to refactor the things view with this PR? As for the popup dialog to add a new channel link, it got actually more complex just recently with #5800.

kaikreuzer:
As this is highly out of sync with the code base by now, I'd suggest to close this PR.

PR #3298, opened 2017-04-24
changing waitForAssert to an absolute timeout

sjka:
...as so far the timout was the aggregated sleep time, which
led to quite long-running tests in case of nested waitForAssert
blocks.


maggu2810:
Good catch.
Can you change that to use nanoTime?

sjka:
> Good catch.

Well, I didn't even need to keep my hands open. It just hit me 😄 

> Can you change that to use nanoTime?

Done. Good point!

maggu2810:
Does this change break the build?
https://hudson.eclipse.org/smarthome/job/SmartHomeDistribution-Nightly/1782/consoleText

sjka:
At least I was not aware of this. I will have a look.

sjka:
If my guess is right, then #3300 will fix it.

PR #3300, opened 2017-04-24
Fix config dispatch test

sjka:
Usually waitForAssert is the magic healing potion for tests.
But ocasionally less is more.


sjka:
Please wait for Travis to finish before merging it, as I need a vote from a slow CI server

maggu2810:
CI failed

sjka:
Thanks for the notice. In the meantime I think I found the root-cause: An embarrassing little mistake in unit conversion AND a nice little int overflow. Just running the tests locally, will update this PR in a few minutes.

maggu2810:
> An embarrassing little mistake in unit conversion AND a nice little int overflow.

Ah, yes :wink: 

You should calculate the timeout (multiplication) only once at the beginning of the function.
Perhaps we should use:
```
final long timeoutNs = TimeUnit.MILLISECONDS.toNanos(timeout);
```

sjka:
> Perhaps we should use [...]

Now I'm wondering on which planet I've been living for the last 13 years. Nice hint, thanks.

Issue #3297, opened 2017-04-24
UpDownType should include a neutral position


n-kristev:
I am currently working on a binding, and one of the devices is a blinds actuator. The `RollershutterItem`  class is the default class used for blinds, and it is using the `UpDownType` for its current blinds direction state.

However, there is a problem: a blinds actuator actually has 3 states (not moving, up, down), while the `UpDownType` only has `UP` and `DOWN.` At the moment I am working around this by having the Not Moving state set as  `UnDefType.UNDEF`. However, this introduces an element of incertanity (why is the state undefined - an error or normal operation?) and additionally confuses the UI, as the `PercentType` component of `RollershutterItem` is updated correctly with the current position of the blinds, so the users sees something like "UNDEF/XX percent". 

For me it would make sense to add a new state - `NOT_MOVING` - to the posible states for `UpDownType`. What do you think?

maggu2810:
The rollershutter item supports this types:
* PercentType
* UpDownType
* UnDefType

and this commands (+ RefreshType):
* UpDownType
* StopMoveType
* PercentType

If we look at the `setState` function we will see that every state that is set and is an instance of a convertible one, will be transformed to a percent type.
So, it seems if you set "up" or "down" as a state to signal the rollershutter is currently moving, the state is already converted to 0 for up and 100 for down.
So, `up` as a command is meant to move to position `0` and if you set this state, the user "see" that the rollershutter is at position zero...

There is also the stop move command. Stop makes sense if you know that the rollershutter is currently moving. As the state does not signal "I am moving up" or "I am moving down", this makes only sense if you "see" the real rollershutter is moving.
To send "move" as a command is perhaps not really useful without direction (except you define that the direction toggles).

@n-kristev After reading this I don't know how the item should be correctly used (wrt. to up down stop move).

n-kristev:
@maggu2810 So it would like like the problem is that the shutter item state is ambiguous as it is currently defined. On one hand it shows the current position of the blinds, and on the other it also shows the direction they are moving. 
It would seem like the state should be made to some kind of a compound type, which includes movement status (up/down/static), and current position.

kaikreuzer:
The state simply describes the current position and not any transition. We also do not have information for Dimmers, whether they are currently being turned on or off or for Temperature values whether they are going up or down.
If you really need additional information about movement direction (and maybe even speed?), I would think that this should be modelled in additional channels.

n-kristev:
[...]

PR #3301, opened 2017-04-24
Comment on UpDownType usage in Rollershutter item (#3297)

n-kristev:
Signed off by: Nikolay Kristev (n.kristev@prosyst.com)

### DIFF
--- a/docs/documentation/concepts/items.md
+++ b/docs/documentation/concepts/items.md
@@ -69,3 +69,10 @@ HSB string values consist of three comma-separated values for hue (0-360°), sat
 | StopMoveType | `STOP`, `MOVE` |
 | UpDownType | `UP`, `DOWN` |
 
+## Additional notes
+
+### Using `UpDownType` states and commands in Rollershutter items
+
+Rollershutter items accept `UpDownType` command type and have an `UpDownType` state. It is important to note that the `UpDownType` state is not a directional indicator,
+that is, it does not indicate that the shutters are currently moving up or down. It is used only as a shorthand to indicate that the shutters are currently in their top(`UP`) or bottom(`DOWN`)
+positions. Respectively, the `UpDownType` command type is used to send a command to the shutters to move to their top or bottom positions.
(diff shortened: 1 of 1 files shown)


bundles/core/org.eclipse.smarthome.core/src/main/java/org/eclipse/smarthome/core/library/items/RollershutterItem.java
```diff
@@ -37,6 +37,11 @@
 
     static {
         acceptedDataTypes.add(PercentType.class);
+		/*
```
sjka: Honestly, I'm not really sure whether this is the right place for such an explanation. It's not even shown in the JavaDoc for RollershutterItem. Somebody would need to look inside the sources and scroll to the right position in order to find it. 

IMHO the [items documentation](https://github.com/eclipse/smarthome/blob/master/docs/documentation/concepts/items.md) would be the much better place for such explanations.

n-kristev:
Done, moved the comment to items.md

maggu2810:
@sjka I assume your change request is now fulfilled.

maggu2810:
@n-kristev There is a missing ip-validation. Can you squash your commits?

kaikreuzer:
@maggu2810 I am not really happy with that addition - it adds some very specific details on RollershutterItem on an overview page where all item types are only briefly described. That imho brings some bad proportions to the page. 

We should imho better document in general how commands vs. state updates in items work and how the different item types work - I wouldn't treat Rollershutter as some prominent exception here.

n-kristev:
@kaikreuzer OK, in this case I can change the note to something more general (if you want to keep it on the same page), and add Rollershutter and Dimmer as an example - in Dimmer's case, it is ON/OFF and the level which have a similar relation as the shutter UP/DOWN and level.

n-kristev:
@maggu2810 Done, commits are squished and the IP validation passes now.

n-kristev:
@kaikreuzer @sjka I have updated the note to be more generic re. item types.

sjka:
Fully makes sense now, I like how this turned out, thanks!

kaikreuzer:
@n-kristev Many thanks and sorry for my sluggish responsiveness...

PR #3303, opened 2017-04-24
Astro cleanups

maggu2810:


maggu2810:
@kaikreuzer The naming scheme is about the artifact name (not bundle name) and all starts with "Eclipse SmartHome"

kaikreuzer:
Thanks!

Issue #3281, opened 2017-04-21
Intropduction of a common filtering API


ViToni:
While working on the Automation part of ESH (https://github.com/eclipse/smarthome/issues/2921) the idea came up to introduce a filtering approach to the RuleRegistry (see https://github.com/eclipse/smarthome/issues/2921#issuecomment-285375056).

However a change such as a filtering API might be better provided in a more general way and in a central place so it's available project wide and not only in the automation part.

- The idea to use Java8 [`Predicate`](https://docs.oracle.com/javase/8/docs/api/java/util/function/Predicate.html)s seems out of scope as parts of ESH must be available in Java 1.5 environments.
- Another approach would have been to make use of Google guava library with its [`Predicate`](https://github.com/google/guava/blob/master/guava/src/com/google/common/base/Predicate.java)s but this would introduce a dependency to `guava` in the public API.
- So it might be reasonable to implement an independent solution for ESH.

The questions are:
1. What would be a good solution for ESH?
2. Where would be the place for such a filtering API?

sjka:
> The idea to use Java8 Predicates seems out of scope as parts of ESH must be available in Java 1.5 environments.

Why don't these "parts" then bring their own `Predicate`? I would prefer not re-inventing yet again another wheel...

kaikreuzer:
> Why don't these "parts" then bring their own Predicate? 

If we add a new method to `Registry`, we cannot make a difference between the "parts" - it should be the same API for them all. And introducing our own `Filter` class could also directly refer to the types of the Registry and offer ESH specific methods that we want to allow for such a filter.

maggu2810:
I would prefer also the Predicate of Java 1.8 but I see the point that it cannot be used for the automation because of its limitation to Java 1.5

@kaikreuzer You pointed out an interesting topic. Ever base class or every API of the core that is used by automation stuff is limited to the 1.5 feature set, too.

danchom:
The Automation is not 1.5 compatible from a long time ago.  As I remember @kaikreuzer said that we can't obligate contributors to write 1.5 compatible code when the ESH is 1.7 compatible and we agreed with him. At the moment the automation module is 1.7 compatible. Now the situation is the same. If we use the 1.8 Predicate we will lose the support of some jvms as skelmir, but on the other hand, there are a lot of other interfaces in the java.util.function which can be used with lambdas and I think it is not good idea to duplicate all of them. So I'm agree that the 1.8 predicate can be used.

[...]

PR #3307, opened 2017-04-25
Fix #3099 resolving of configuration references

### LINKED ISSUE #3099: Template resolving discussion.
I open this issue to continue discussion from #3050 - about template resolving.
 
@kaikreuzer 's comment:

> I think the PR#3097 will fix the problem. 

Thanks.

> But IMHO this should be a temporary change.

Yes, so we should continue discussing a proper solution. I did a suggestion above already:

> We could also try to dynamically determine this: If a module is marked as HIDDEN, we could keep the placeholder, while any module that is meant to be displayed in the UI should also have "real" config values and no placeholder anymore.

This would allow us to do "black box templates", i.e. create rules from templates, which do not expose their inner workings, but merely offer their configuration to the user. The hidden, internal modules would still have the placeholders, which are only replaced for the RuntimeRule.
On the other hand, it also allows templates as a one-time skeleton for a complex rule, which is created in all its glory and left for further customization to the user - he might change modules, their configuration, etc and thus there must be no placeholders in the module configuration anymore.

What do you think about that?

danchom:

### DIFF
--- a/bundles/automation/org.eclipse.smarthome.automation.integration.test/org.eclipse.smarthome.automation.integration.test.launch
+++ b/bundles/automation/org.eclipse.smarthome.automation.integration.test/org.eclipse.smarthome.automation.integration.test.launch
@@ -26,7 +26,7 @@
 <stringAttribute key="org.eclipse.jdt.junit.TESTNAME" value=""/>
 <stringAttribute key="org.eclipse.jdt.junit.TEST_KIND" value="org.eclipse.jdt.junit.loader.junit4"/>
 <booleanAttribute key="org.eclipse.jdt.launching.ATTR_USE_START_ON_FIRST_THREAD" value="true"/>
-<stringAttribute key="org.eclipse.jdt.launching.JRE_CONTAINER" value="org.eclipse.jdt.launching.JRE_CONTAINER/org.eclipse.jdt.internal.debug.ui.launcher.StandardVMType/JavaSE-1.7"/>
+<stringAttribute key="org.eclipse.jdt.launching.JRE_CONTAINER" value="org.eclipse.jdt.launching.JRE_CONTAINER/org.eclipse.jdt.internal.debug.ui.launcher.StandardVMType/JavaSE-1.8"/>
 <stringAttribute key="org.eclipse.jdt.launching.MAIN_TYPE" value=""/>
 <stringAttribute key="org.eclipse.jdt.launching.PROGRAM_ARGUMENTS" value="-os ${target.os} -ws ${target.ws} -arch ${target.arch} -nl ${target.nl} -consoleLog"/>
 <stringAttribute key="org.eclipse.jdt.launching.PROJECT_ATTR" value="org.eclipse.smarthome.automation.integration.test"/>
--- a/bundles/automation/org.eclipse.smarthome.automation.core.test/src/test/java/org/eclipse/smarthome/automation/core/internal/RuleEngineTest.java
+++ b/bundles/automation/org.eclipse.smarthome.automation.core.test/src/test/java/org/eclipse/smarthome/automation/core/internal/RuleEngineTest.java
@@ -191,7 +191,7 @@ public class RuleEngineTest {
         List<ConfigDescriptionParameter> rule4cfgD = rule4Get.getConfigurationDescriptions();
         Assert.assertNotNull("Rule configuration is null", rule4cfg);
         Assert.assertTrue("Missing config property in rule copy", rule4cfg.containsKey("config1"));
-        Assert.assertEquals("Wrong config value", new BigDecimal(5), rule4cfg.get("config1"));
+        Assert.assertEquals("Wrong config value", 5, rule4cfg.get("config1"));
 
         Assert.assertNotNull("Rule configuration description is null", rule4cfgD);
         Assert.assertEquals("Missing config description in rule copy", 1, rule4cfgD.size());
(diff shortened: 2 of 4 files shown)


bundles/automation/org.eclipse.smarthome.automation.integration.test/org.eclipse.smarthome.automation.integration.test.launch
```diff
@@ -26,7 +26,7 @@
 <stringAttribute key="org.eclipse.jdt.junit.TESTNAME" value=""/>
 <stringAttribute key="org.eclipse.jdt.junit.TEST_KIND" value="org.eclipse.jdt.junit.loader.junit4"/>
 <booleanAttribute key="org.eclipse.jdt.launching.ATTR_USE_START_ON_FIRST_THREAD" value="true"/>
-<stringAttribute key="org.eclipse.jdt.launching.JRE_CONTAINER" value="org.eclipse.jdt.launching.JRE_CONTAINER/org.eclipse.jdt.internal.debug.ui.launcher.StandardVMType/JavaSE-1.7"/>
+<stringAttribute key="org.eclipse.jdt.launching.JRE_CONTAINER" value="org.eclipse.jdt.launching.JRE_CONTAINER/org.eclipse.jdt.internal.debug.ui.launcher.StandardVMType/JavaSE-1.8"/>
```
maggu2810: One question about that change: The automation stuff uses 1.5, so why not using the correct 1.5 settings?

danchom: Actually the automation sources are 1.7 compatible because the ESH was 1.7. When the ESH goes to jdk 1.8 we have decided with Kai that they can be 1.8 compact 2 (compact 2 because the GSON requires compact2). We are trying to be close to 1.7 as match as possible to have opportunity to support jvm such as skelmir. 
I've changed required jvm because the test can be started. It depends from bundles which required 1,8 jvm and there are a lot of such exceptions: 

```
org.osgi.framework.BundleException: Could not resolve module: org.eclipse.smarthome.automation.providers [5]
  Unresolved requirement: Import-Package: org.eclipse.smarthome.automation
    -> Export-Package: org.eclipse.smarthome.automation; bundle-version="0.9.0.qualifier"; bundle-symbolic-name="org.eclipse.smarthome.automation.api"; version="0.0.0"
       org.eclipse.smarthome.automation.api [48]
         Unresolved requirement: Import-Package: org.eclipse.smarthome.config.core
           -> Export-Package: org.eclipse.smarthome.config.core; bundle-version="0.9.0.qualifier"; bundle-symbolic-name="org.eclipse.smarthome.config.core"; version="0.0.0"
              org.eclipse.smarthome.config.core [3]
                Unresolved requirement: Import-Package: org.eclipse.smarthome.core.common
                  -> Export-Package: org.eclipse.smarthome.core.common; bundle-version="0.9.0.qualifier"; bundle-symbolic-name="org.eclipse.smarthome.core"; version="0.0.0"
                     org.eclipse.smarthome.core [8]
                       Unresolved requirement: Require-Capability: osgi.ee; filter:="(&(osgi.ee=JavaSE)(version=1.8))"

	at org.eclipse.osgi.container.Module.start(Module.java:434)
	at org.eclipse.osgi.container.ModuleContainer$ContainerStartLevel.incStartLevel(ModuleContainer.java:1582)
	at org.eclipse.osgi.container.ModuleContainer$ContainerStartLevel.incStartLevel(ModuleContainer.java:1562)
	at org.eclipse.osgi.container.ModuleContainer$ContainerStartLevel.doContainerStartLevel(ModuleContainer.java:1533)
	at org.eclipse.osgi.container.ModuleContainer$ContainerStartLevel.dispatchEvent(ModuleContainer.java:1476)
	at org.eclipse.osgi.container.ModuleContainer$ContainerStartLevel.dispatchEvent(ModuleContainer.java:1)
	at org.eclipse.osgi.framework.eventmgr.EventManager.dispatchEvent(EventManager.java:230)
	at org.eclipse.osgi.framework.eventmgr.EventManager$EventThread.run(EventManager.java:340)

maggu2810: > I've changed required jvm because the test can be started. It depends from bundles which required 1,8 jvm and there are a lot of such exceptions:

Okay, all fine.

> When the ESH goes to jdk 1.8 we have decided with Kai that they can be 1.8 compact 2 (compact 2 because the GSON requires compact2). 

I don't get the point how does this fit to that comment:

> The idea to use Java8 Predicates seems out of scope as parts of ESH must be available in Java 1.5 environments.

https://github.com/eclipse/smarthome/issues/3281#issue-223322632

danchom: I've comment in #3281. Actually the Automation module must be some compact of 1.8.

maggu2810: Thanks for clarification

danchom:
This PR fixes following problems:
1) Move configuration resolving from rule engine to the rule registry. In that way the registry will contain only resolved rules. 
2) It resolves configuration links for all rules containing configuration, not only for rule's created from template.
3) Fix normalization of rule's configuration.  It must be applied before resolving of modules configurations,  otherwise template's default values are not applied to the module configuration.

danchom:
It is obvious that with the current api there is no clean approach for rule configuration topic. After internal discussion we have agree for that:

this command invocation ruleRegistry.update(ruleRegistry.get(rule)) must not change the rule
it is not good idea a property value to have possibility to contain different value types (i.e. Integer value or String reference to Integer value)
When we want to extract module properties as rule configuration property, for easy rule configuration, this can be created by template with this properties. Actually the templates are for this.
For these reasons we refuse of rule configuration and I think this issue and PR #3307 can be closed.
@kaikreuzer what do you think?

PR #3309, opened 2017-04-25
Changed 'final static' to 'static final' in archetype

cweitkamp:
Related to PR [#3278](https://github.com/eclipse/smarthome/pull/3278) (Worked on a set of minor issues in the archetype).

Signed-off-by: Christoph Weitkamp github@christophweitkamp.de

### DIFF
--- a/tools/archetype/binding/src/main/resources/archetype-resources/src/main/java/internal/__bindingIdCamelCase__HandlerFactory.java
+++ b/tools/archetype/binding/src/main/resources/archetype-resources/src/main/java/internal/__bindingIdCamelCase__HandlerFactory.java
@@ -29,7 +29,7 @@ import org.eclipse.smarthome.core.thing.binding.ThingHandler;
  */
 public class ${bindingIdCamelCase}HandlerFactory extends BaseThingHandlerFactory {
 
-    private final static Set<ThingTypeUID> SUPPORTED_THING_TYPES_UIDS = Collections.singleton(THING_TYPE_SAMPLE);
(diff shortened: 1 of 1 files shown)


kaikreuzer:
Thanks @cweitkamp! It seems you do not yet have signed an [ECA with the Eclipse Foundation](https://www.eclipse.org/legal/ecafaq.php), right? Would be wonderful, if you could do so!

cweitkamp:
@kaikreuzer Yes, that's right. I will look into that.

cweitkamp:
ECA is signed. Squashed my commit to retrigger validations.

kaikreuzer:
Many thanks, the ECA looks good now!

tools/archetype/binding/src/main/resources/archetype-resources/ESH-INF/binding/binding.xml
```diff
@@ -1,11 +1,11 @@
 <?xml version="1.0" encoding="UTF-8"?>
 <binding:binding id="${bindingId}"
-				 xmlns:xsi="http://www.w3.org/2001/XMLSchema-instance"
```
kaikreuzer: Did you change the tabs to spaces on purpose in this file? Afaik, our auto-formatter uses tabs in XML files and hence we should also use the same in the initially generated files.

cweitkamp: No, no special purpose. Just wanted to retrigger the validations. I will revert those changes.

PR #3312, opened 2017-04-25
Add i18n support to LIFX

wborn:
This PR adds i18n support to the LIFX Binding with English and Dutch translations.
I've also corrected 2016 to 2017 in some copyright headers.


kaikreuzer:
I am not quite clear about why you need the ChannelFactory at all - I would have assumed that our i18n infrastructure also supports the localization of channel (types) out of the box?

See e.g. the [Yahoo binding, which also has localized channels](https://github.com/eclipse/smarthome/blob/master/extensions/binding/org.eclipse.smarthome.binding.yahooweather/ESH-INF/i18n/yahooweather_de.properties#L15-L21).

Am I missing something?

wborn:
@kaikreuzer  The ChannelFactory is only used for the dynamic channels to get the correct zone index in the channel labels and descriptions.

The complexity could be reduced by removing the ChannelFactory but then all these channels will have the same labels and descriptions. 

That may not necessarily be an issue because they are advanced channels and users can still identify them with the ChannelUID.

kaikreuzer:
Ah, that's what I had missed then - thanks for the clarification, makes sense!

PR #3314, opened 2017-04-26
Corrected and defined behavior of group functions

### LINKED ISSUE #3051: Switch Groups changing to UNDEF

triller-telekom:
A plain group without type does NOT have a function now.
A group with a type uses the EQUALITY function by default
A group with type and function uses the specified function

Fixes #3051

### DIFF
--- a/bundles/core/org.eclipse.smarthome.core.test/src/test/groovy/org/eclipse/smarthome/core/items/GroupItemTest.groovy
+++ b/bundles/core/org.eclipse.smarthome.core.test/src/test/groovy/org/eclipse/smarthome/core/items/GroupItemTest.groovy
@@ -16,8 +16,11 @@ import org.eclipse.smarthome.core.items.events.GroupItemStateChangedEvent
 import org.eclipse.smarthome.core.items.events.ItemEventFactory
 import org.eclipse.smarthome.core.library.items.NumberItem
 import org.eclipse.smarthome.core.library.items.SwitchItem
+import org.eclipse.smarthome.core.library.types.ArithmeticGroupFunction
+import org.eclipse.smarthome.core.library.types.OnOffType
 import org.eclipse.smarthome.core.library.types.RawType
 import org.eclipse.smarthome.core.types.RefreshType
+import org.eclipse.smarthome.core.types.UnDefType
 import org.eclipse.smarthome.test.OSGiTest
 import org.junit.Assert
 import org.junit.Before
@@ -29,8 +32,12 @@ import org.junit.Test
  *
  * @author Dennis Nobel - Initial contribution
  * @author Christoph Knauf - event tests
+ * @author Stefan Triller - tests for group status with and without functions
  */
 class GroupItemTest extends OSGiTest {
+
+    /** Time to sleep when a file is created/modified/deleted, so the event can be handled */
+    private final static int WAIT_EVENT_TO_BE_HANDLED = 1000
 
     List<Event> events = []
     EventPublisher publisher
@@ -86,7 +93,7 @@ class GroupItemTest extends OSGiTest {
     @Test
     void 'assert that group item posts events for changes correctly' (){
         events.clear()
-        GroupItem groupItem = new GroupItem("root")
+        GroupItem groupItem = new GroupItem("root", new SwitchItem("mySwitch"), new GroupFunction.Equality())
         def member = new TestItem("member1")
         groupItem.addMember(member)
         groupItem.setEventPublisher(publisher)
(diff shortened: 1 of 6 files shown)


bundles/core/org.eclipse.smarthome.core/src/main/java/org/eclipse/smarthome/core/items/GroupItem.java
```diff
@@ -36,11 +36,12 @@
     protected GroupFunction function;
 
     public GroupItem(String name) {
-        this(name, null);
+        this(name, null, null);
     }
 
     public GroupItem(String name, GenericItem baseItem) {
-        this(name, baseItem, new GroupFunction.Equality());
+        // this(name, baseItem, new GroupFunction.Equality());
```
sjka: ??

bundles/core/org.eclipse.smarthome.core/src/main/java/org/eclipse/smarthome/core/items/GroupItem.java
```diff
@@ -286,7 +287,9 @@ public void stateChanged(Item item, State oldState, State newState) {
     @Override
     public void stateUpdated(Item item, State state) {
         State oldState = this.state;
-        setState(function.calculate(members));
+        if (null != function) {
```
sjka: I know, it's a matter of personal taste, but so far we have them the other way round pretty consistently.

triller-telekom:
A group without a function still has a state, but its members do not affect the status of the group. For example: Consider a group with 2 switches swA and swB and I do not have any type or function set. i.e. the items file looks like

```
Group mySwitches
Switch swA (mySwitches)
Switch swB (mySwitches)
```

Then if i trigger `swA` or `swB` the status of `mySwitches` does NOT change, i.e. it stays `NULL`. But if I have a sitemap with:

```
Switch item=mySwitches
```

and I trigger that "group switch", then the status of the group becomes `ON`. And if I trigger that switch again it becomes `OFF`.

bundles/core/org.eclipse.smarthome.core.test/src/test/groovy/org/eclipse/smarthome/core/items/GroupItemTest.groovy
```diff
@@ -117,4 +124,151 @@ class GroupItemTest extends OSGiTest {
...
+        assertTrue change.getItemName().equals(groupItem.getName())
+
+        assertTrue change.getOldItemState().equals(UnDefType.NULL)
+        assertTrue change.getItemState().equals(OnOffType.ON)
+
+        events.clear();
+
+        sw2.setState(OnOffType.ON);
+
+        events.clear();
```
sjka: as there is no event expected I wouldn't clear `events` here

bundles/core/org.eclipse.smarthome.core.test/src/test/groovy/org/eclipse/smarthome/core/items/GroupItemTest.groovy
```diff
@@ -117,4 +124,151 @@ class GroupItemTest extends OSGiTest {
...
+
+        sw2.setState(UnDefType.UNDEF);
+
+        //wait to see that the event doesn't fire
+        sleep(WAIT_EVENT_TO_BE_HANDLED)
+
+        waitForAssert {
```
sjka: no need to waitForAssert here

bundles/core/org.eclipse.smarthome.core.test/src/test/groovy/org/eclipse/smarthome/core/items/GroupItemTest.groovy
```diff
@@ -117,4 +124,151 @@ class GroupItemTest extends OSGiTest {
...
+        groupItem.setEventPublisher(publisher)
+        def oldGroupState = groupItem.getState()
+
+        //State changes -> NO change event should be fired
+        member.setState(new RawType())
+
+        //wait to see that the event doesn't fire
+        sleep(WAIT_EVENT_TO_BE_HANDLED)
+
+        waitForAssert {
```
sjka: no need to waitForAssert here

triller-telekom:
Just did these minor changes on the tests.

PR #3328, opened 2017-04-28
Fix unglify angular code for PaperUI

htreu:
do not uglify names in angular code to leave DI intact


kaikreuzer:
Thanks, this fixes it for me.

PR #3332, opened 2017-04-28
Sitemap Events provide item states in a format that matches map entries

### LINKED ISSUE #866: Hue binding: dimmer and switch items for brightness should move together
For a Philips Living Colors Bloom light, I linked manually the brightness channel to a switch item and to a dimmer item.
When I set the dimmer to 0 (using the UI), the switch is not switch OFF and remains ON.
PS: note that it works well in openHAB 1.x.

triller-telekom:
BasicUI converted an item state to a switch On/Off type, but ClassicUI did not.
ItemUiRegistry now takes care of such a state conversions for Widgets of type Switch and Slider.

Fixes #866


bundles/ui/org.eclipse.smarthome.ui/META-INF/MANIFEST.MF
```diff
@@ -32,6 +32,7 @@ Import-Package: javax.imageio,
  org.eclipse.smarthome.model.core,
  org.eclipse.smarthome.model.items,
  org.eclipse.smarthome.model.sitemap,
+ org.eclipse.smarthome.model.sitemap.impl,
```
sjka: err...?

bundles/ui/org.eclipse.smarthome.ui/src/main/java/org/eclipse/smarthome/ui/internal/items/ItemUIRegistryImpl.java
```diff
@@ -72,6 +76,7 @@
  *
  * @author Kai Kreuzer - Initial contribution and API
  * @author Chris Jackson
+ * @author Stefan Triller - convertState(Widget, State)
```
sjka: please state here _what_ you did

bundles/ui/org.eclipse.smarthome.ui/src/main/java/org/eclipse/smarthome/ui/internal/items/ItemUIRegistryImpl.java
```diff
@@ -476,6 +481,34 @@ public State getState(Widget w) {
...
+     * Converts an item state to the type the widget supports (if possible)
+     *
+     * @param w - Widget in sitemap that shows the state
+     * @param s - State of the item
+     * @return the converted state or the original if conversion was not possible
+     */
+    private State convertState(Widget w, State s) {
+        State returnState = null;
+
+        if (w instanceof SwitchImpl) {
```
sjka: please check for the interfaces here, not the implementations.

bundles/ui/org.eclipse.smarthome.ui/src/main/java/org/eclipse/smarthome/ui/internal/items/ItemUIRegistryImpl.java
```diff
@@ -476,6 +481,34 @@ public State getState(Widget w) {
...
+            if (s instanceof PercentType) {
+                returnState = ((PercentType) s).as(OnOffType.class);
+            }
+        } else if (w instanceof SliderImpl) {
+            if (s instanceof PercentType) { // catches also HSBType etc
+                returnState = ((PercentType) s).as(PercentType.class);
+            }
+        }
```
sjka: Are these the only widget types that potentially need conversion? How about e.g. the Colorpicker?

bundles/ui/org.eclipse.smarthome.ui/src/main/java/org/eclipse/smarthome/ui/internal/items/ItemUIRegistryImpl.java
```diff
@@ -476,6 +481,34 @@ public State getState(Widget w) {
...
+     *
+     * @param w - Widget in sitemap that shows the state
+     * @param s - State of the item
+     * @return the converted state or the original if conversion was not possible
+     */
+    private State convertState(Widget w, State s) {
+        State returnState = null;
+
+        if (w instanceof SwitchImpl) {
+            if (s instanceof PercentType) {
```
sjka: you don't need to check for a specific type here - just use `s instanceof Convertible` so we keep the knowledge of what can be converted to what in a single place.

bundles/ui/org.eclipse.smarthome.ui/META-INF/MANIFEST.MF
triller-telekom: Indeed, somehow slipped from testing into my commit...

bundles/ui/org.eclipse.smarthome.ui/src/main/java/org/eclipse/smarthome/ui/internal/items/ItemUIRegistryImpl.java
```diff
@@ -490,12 +489,12 @@ public State getState(Widget w) {
     private State convertState(Widget w, State s) {
         State returnState = null;
 
-        if (w instanceof SwitchImpl) {
-            if (s instanceof PercentType) {
+        if (w instanceof Switch) {
+            if (s instanceof Convertible) {
```
lolodomo: Why not just: `returnState = s.getStateAs(OnOffType.class);`
This call is doing the test with Covertible.

lolodomo:
I have not tested this PR but it makes really sense and will help all UIs.

bundles/ui/org.eclipse.smarthome.ui/src/main/java/org/eclipse/smarthome/ui/internal/items/ItemUIRegistryImpl.java
```diff
@@ -476,6 +480,34 @@ public State getState(Widget w) {
     }
 
     /**
+     * Converts an item state to the type the widget supports (if possible)
+     *
+     * @param w - Widget in sitemap that shows the state
```
maggu2810: Can you remove the "-" between the parameter name and the comment?

bundles/ui/org.eclipse.smarthome.ui/src/main/java/org/eclipse/smarthome/ui/internal/items/ItemUIRegistryImpl.java
lolodomo: You're right, Colorpicker is expecting a HSBType, so that might be a good idea to try a conversion.

lolodomo: For colopicker, a OFF or 0 could be converted to (0,0,0) but how will be converted a ON or a value > 0 ?

triller-telekom: A `Colorpicker` only makes sense for a `Color` item and this does have a HSBType state anyway. Why would anyone want to attach a `ColorPicker` to a `Switch` or `Dimmer` item for example? What would be the use case? I don't see any, which is why I added no conversion for the `ColorPicker`.

lolodomo: You're right.
The real usage is more the reverse, that is using a switch or a dimmer for a color item for example and it is what ypu have covered.

lolodomo:
LGTM

sjka:
To mee to. Thanks!

PR #3343, opened 2017-04-30
Basic UI: display value for switch element when requested

### LINKED ISSUE #3340: Basic UI: missing value for switch element
Either Classic UI or HABdroid are displaying for sitemap switch element the value (when defined in the item label) in addition to the switch control. Basic UI just displays the switch control without any value. I think this value should be displayed.

I prefer how it is displayed by Classic UI rather than HABdroid, that is at right just in front of the switch control.

lolodomo:
Fix #3340 

Handle label and value color


lolodomo:
@resetnow : can you review please ?

lolodomo:
Please wait before merging, I forgot to consider value color.

lolodomo:
Now ready for review.

extensions/ui/org.eclipse.smarthome.ui.basic/snippets/switch.html
```diff
@@ -2,13 +2,17 @@
 	<span class="mdl-form__icon">
 		<img data-icon="%category%" src="../icon/%category%?state=%state%&format=%icon_type%" />
 	</span>
-	<div class="mdl-form__label">
+	<span %labelstyle% class="mdl-form__label">
 		%label%
-	</div>
+	</span>
+	<span %valuestyle% class="mdl-form__value mdl-form__value--group">
```
resetnow: `mdl-form__value--group` should be `mdl-form__value--switch`

And the value needs some padding:

```less
// inside .mdl-form__value
&--switch {
	padding: 0 10px 0 10px;
}
```

extensions/ui/org.eclipse.smarthome.ui.basic/web-src/smarthome.js
```diff
@@ -1192,21 +1192,32 @@
...
 			var
 				val = itemState === "ON";
 
-			_t.input.checked = val;
-
-			if (val) {
+			if (_t.input.checked !== val) {
+				_t.input.checked = val;
+				if (val) {
 				_t.parentNode.MaterialSwitch.on();
```
resetnow: Please add missing indentation level

extensions/ui/org.eclipse.smarthome.ui.basic/web-src/smarthome.js
```diff
@@ -1192,21 +1192,32 @@
...
-			} else {
+				} else {
 				_t.parentNode.MaterialSwitch.off();
```
resetnow: Same

extensions/ui/org.eclipse.smarthome.ui.basic/web-src/smarthome.js
lolodomo: I have big difficulties with indentation. Eclipse, even after updating the settings is adding a mix of tabs and spaces. If I use notepad to edit the file, after saving, it looks like the beginning of the file is updated !
Is there a reason to use tabs instead of spaces in JS file while we are using spaces for Java files ?
What setting did you change in Eclipse ?
I spent almost more time trying to have a correct indentation than working on the fix !!!

extensions/ui/org.eclipse.smarthome.ui.basic/snippets/switch.html
lolodomo: Ok, makes sense. I am starting to understand the philosophy with CSS and less.
It looks like the font looked less bold than for a normal text item or maybe the color was less black. Any hint on this ?

resetnow: All `.mdl-form__value` elements have class `font-weight: 700;` style. For me it looks just like any other value.

extensions/ui/org.eclipse.smarthome.ui.basic/web-src/smarthome.js
resetnow: There should be a setting in Preferences somewhere in Text Editor (for web), but it doesn't work for me either, so I just use gedit or Sublime Text (the latter can also automatically check code with eslint)

resetnow: If you use Notepad++, save as UTF-8 (without BOM) instead of UTF-8. Also, you can stage files partially in git.

lolodomo:
Done.
As a bonus, the font looks good now.

lolodomo:
@kaikreuzer : I would appreciate if you could merge this PR quickly because I have in the pipe another fix for Basic UI fixing two issues relative to slection element in sitemap.
I prefer to not include the other fixes inside this PR as @resetnow already approved it.

sjka:
works for me too. Thanks!

lolodomo:
Wonderful.

PR #3350, opened 2017-05-02
Yahoo improvements

kaikreuzer:
Do not set Thing to ONLINE before it is fully initialized.


sjka:
as just discussed: the location param _should_ not be null: it's a mandatory config parameter, therefore it's not initializable (yet)...

kaikreuzer:
Updated PR.

PR #3353, opened 2017-05-02
Classic UI: fix icon for selection and switch elements with mappings …

lolodomo:
…in sitemap


sjka:
Thanks!

PR #3360, opened 2017-05-03
Rearrange cards on search text change

aounhaider1:
Before:
screen shot 2017-05-03 at 17 28 49

After:
screen shot 2017-05-03 at 17 28 24


aounhaider1:
Updated

htreu:
Code looks good to me.
But there is still a bug when searching/filtering extensions: The `X` which should reset the filter and display all available extensions again does not work. When filtering bindings by `knx`, then clicking the `X` the filtered bindings remain but the search text is gone. After that the search does not react to changes in the search text field. Reloading the page is the only way to recover from this.

Please have another look @aounhaider1.

aounhaider1:
@htreu should work now.

htreu:
@kaikreuzer its not essential here to wait for travis or the eclipse infrastructure... This is ready to get merged.

PR #3361, opened 2017-05-03
fix scope in automation script rule support

sjka:
I observed that the scope has been empty in JS rules without this change. However, as there is no documentation so far, I'm not 100% sure whether I really got the idea right. 
Therefore, @smerschjohann, it would be good if you could have a quick look at this too...?

### DIFF
--- a/bundles/automation/org.eclipse.smarthome.automation.module.script.rulesupport/src/main/java/org/eclipse/smarthome/automation/module/script/rulesupport/internal/RuleSupportScriptExtension.java
+++ b/bundles/automation/org.eclipse.smarthome.automation.module.script.rulesupport/src/main/java/org/eclipse/smarthome/automation/module/script/rulesupport/internal/RuleSupportScriptExtension.java
@@ -9,7 +9,6 @@ package org.eclipse.smarthome.automation.module.script.rulesupport.internal;
 
 import java.util.Arrays;
 import java.util.Collection;
-import java.util.Collections;
 import java.util.HashMap;
 import java.util.HashSet;
 import java.util.Map;
@@ -117,7 +116,7 @@ public class RuleSupportScriptExtension implements ScriptExtensionProvider {
 
     @Override
     public Collection<String> getDefaultPresets() {
-        return Collections.emptyList();
+        return presets.keySet();
     }
 
     @Override


kaikreuzer:
So let's assume @smerschjohann's approval and merge. We can at any time roll-back if necessary :-)

smerschjohann:
Ah, sorry. Well, it was meant to be imported using se.importPreset("RuleSupport"), now it will be always imported.

We could leave it like this for now. But I thought it would be a good idea to leave the default scope as clean as possible.

sjka:
OIC. The missing piece was the

```
se.importPreset("RuleSupport")
```
I simply was not aware of it.

> But I thought it would be a good idea to leave the default scope as clean as possible.

I second that! 

Plus: It would give us something like an "API versioning capability". So we could later on introduce a "RuleSupport2" scope in parallel, without necessarily breaking all existing scripts immediately. I like that. 

I will happily revert my change then.

lewie:
Some Examples: https://github.com/lewie/openhab2-javascript/blob/master/jslib/JSRule.js

sjka:
Thanks, @lewie. I found it before, was a great help!
However, I have been struggling with this:

```
ReferenceError: "ScriptExtension" is not defined
```

and simply didn't look at the right place. When you change `ScriptExtension` to `scriptExtension` or `se` it works. Should have looked into [ScriptEngineManagerImpl](https://github.com/eclipse/smarthome/blob/7911cb21a2330a31884ae7ffd470245029377612/bundles/automation/org.eclipse.smarthome.automation.module.script/src/main/java/org/eclipse/smarthome/automation/module/script/internal/ScriptEngineManagerImpl.java#L81) before...

sjka:
Btw, I created https://github.com/eclipse/smarthome/wiki/Scripted-Rule-Support until we have a proper docu. Especially the required VM argument hit me badly. Any further hints are welcome.

And a good user documentation would be awesome...! 😉

lewie:
@sjka, was not aware about the shortcuts *"scriptExtension"* and "se" too. I imported *ScriptExecution* directly: https://github.com/lewie/openhab2-javascript/blob/master/jslib/helper.js#L6
Thanks, will change this.

As I know, in a java 9 environment *-Dorg.osgi.framework.bundle.parent=ext* no longer is needed. They will change preferred class-loader to "ext" by default.

sjka:
Ah, that's why... This is what happens when not reading examples fully. So we both learned something new.

> As I know, in a java 9 environment -Dorg.osgi.framework.bundle.parent=ext no longer is needed. 

I'm totally excited that we now finally can use Java 8 😉

lewie:
Yes, lesson learned. 👍 

>I'm totally excited that we now finally can use Java 8

Absolutely agree!!!!
I just wanted to say that this parameter does not have to hit you badly because it is directed to the future. If it is that far, then we already have experience with this classloader strategy and its side effects. 😉

lewie:
@sjka, @smerschjohann, don't know why! But I do not need to set:

>-Dorg.osgi.framework.bundle.parent=ext

Don't need it neither in windows nor in linux, running JavaScript Rules! 

Please, can anybody verify this?

sjka:
It's indeed not required for javascript support to work at all. However, apparently without this parameter there are two classloaders behaving a little uncooperative:

> TypeError: Can not find a common class loader for ScriptObject and org.eclipse.smarthome.automation.module.script.rulesupport.shared.simple.SimpleRule. in <eval> at line number 6

Issue #3305, opened 2017-04-25
Move groovy unit tests to JUnit, mockito & hamcrest


htreu:
In addition to the move to Java 8 with #2926 we want to move the existing groovy unit tests to a pure Java setup with JUnit, [mockito](http://site.mockito.org/) and [hamcrest](http://hamcrest.org/JavaHamcrest/). By getting rid of groovy we want to streamline the tooling and reduce complexity.

Steps to take:
* introduce OSGi dependencies for mockito & hamcrest -> will be introduced by #3258.
* adopt the [archetype](https://github.com/eclipse/smarthome/tree/master/tools/archetype/binding.test) creating the OH2 binding test fragment to provide a mockito/hamcrest template test.
* adopt the documentation for OH2 binding tests [here](https://www.eclipse.org/smarthome/documentation/development/testing.html).
* refactor existing groovy tests to use the new test infrastructure.

kaikreuzer:
@maggu2810 I assume that plan gets your approval?

maggu2810:
Sure :wink:  :+1:

svilenvul:
> In addition to the move to Java 8 with #2926 we want to move the existing groovy unit tests to a pure Java setup with JUnit, mockito and hamcrest.

This is great! Some of these tests actually doesn't require an OSGi runtime anymore, but Tycho Surefire plugin still launches one for them. Eclipse test plugins can be executed by the Maven Surefire plugin as pure unit tests.

sjka:
Just an update on this:

```bash
➜  find . -name "*.groovy" | grep "/src/" | wc -l                                                                                                                                                               
     100
```

Exactly 100 left 😄

maggu2810:
Still some work left.

```
$ find . -name "*.groovy" | grep "/src/" | wc -l
96
```

```
=== automation ===

./bundles/automation/org.eclipse.smarthome.automation.event.test/src/test/groovy/org/eclipse/smarthome/automation/event/RuleEventTest.groovy

./bundles/automation/org.eclipse.smarthome.automation.integration.test/src/test/groovy/org/eclipse/smarthome/automation/integration/test/AutomationIntegrationJsonTest.groovy
./bundles/automation/org.eclipse.smarthome.automation.integration.test/src/test/groovy/org/eclipse/smarthome/automation/integration/test/AutomationIntegrationTest.groovy
[...]

PR #3363, opened 2017-05-04
Refactor the binding archetypes

htreu:
This addresses #3305:

* remove all groovy dependencies
* add Mockito dependency
* migrate tests from groovy to java
* add new Mockito base test
* add launch config test project
* fix binding creation scripts (remove unsupported catalog property)


tools/archetype/binding.test/src/main/resources/archetype-resources/__artifactId__.launch
```diff
@@ -0,0 +1,43 @@
+<?xml version="1.0" encoding="UTF-8" standalone="no"?>
```
sjka: there used to be a list of "selected target plugins" - they were important, as right now the OSGi tests fail in the IDE. The symptom is that the bundleContext is `null`, which is because the bundle under test is not started, and tons of others also aren't.

tools/archetype/binding.test/src/main/resources/archetype-resources/src/test/java/internal/__bindingIdCamelCase__InternalTest.java
```diff
@@ -1,5 +1,8 @@
+#set( $dt = $package.getClass().forName("java.util.Date").newInstance() )
```
sjka: I'm not wondering whether we really still need this test class at all. The HandlerTest above actually is the much better example for a plain JUnit test.

tools/archetype/binding.test/src/main/resources/archetype-resources/src/test/java/__bindingIdCamelCase__HandlerTest.java
```diff
@@ -0,0 +1,54 @@
...
+import org.junit.Before;
+import org.junit.Test;
+import org.mockito.Mockito;
+
+/**
+* Tests cases for {@link  ${bindingIdCamelCase}Handler}. The tests provide mocks for supporting entities using Mockito.
+*
+* @author  ${author} - Initial contribution
+*/
+public class ${bindingIdCamelCase}HandlerTest {
```
sjka: 👍  good idea to have this test!

tools/archetype/binding/src/main/resources/META-INF/maven/archetype-metadata.xml
```diff
@@ -11,21 +10,48 @@
...
-        <requiredProperty key="bindingIdCamelCase">
-		</requiredProperty>        
 	</requiredProperties>
 
 	<fileSets>
 		<fileSet filtered="true" packaged="false" encoding="UTF-8">
 			<directory></directory>
 			<includes>
-				<include>**/*.*</include>
+				<include>cfg/*.*</include>
```
sjka: is cfg/*.* filtered twice for any reasons (see below)? Otherwise I'd remove his block here.

tools/archetype/binding.test/src/main/resources/archetype-resources/src/test/java/internal/__bindingIdCamelCase__InternalTest.java
htreu: agreed. this will even slim line the whole template.

htreu:
The last commit adds documentation for the new test setup. It also extends the test template by some more Mockito stuff.

@sjka this is ready to get merged.

tools/archetype/binding.test/src/main/resources/archetype-resources/META-INF/MANIFEST.MF
```diff
@@ -7,22 +7,18 @@ Bundle-Version: ${version.replaceAll("-SNAPSHOT", ".qualifier")}
...
  org.eclipse.smarthome.core.thing.util,
- org.eclipse.smarthome.test,
  org.eclipse.smarthome.test.storage,
+ org.eclipse.smarthome.test.java,
  org.hamcrest;core=split,
+ org.mockito,
  org.junit;version="4.0.0",
  org.osgi.service.device,
  org.osgi.framework
+Require-Bundle: org.junit,org.mockito,org.hamcrest
```
martinvw: @htreu @sjka @kaikreuzer is it expected that thuis creaped in?

The manifest must not contain any `Require-Bundle` entries. Instead, `Import-Package` must be used. [[B.4](http://docs.openhab.org/developers/development/guidelines.html#b-osgi-bundles)]

This means you also have to reference specific packages in the `Import-Package` instead of the top-level packages needed for `Require-Bundle`

htreu: Hi @martinvw, unfortunately we need those in the test bundles: The package org.hamcrest is a split package but is not cleanly defined as such. The only way we got this to work was to provide the specific bundles. This way dependency resolving is able to find the whole package.

I´m sorry for this broad explanation as I don´t fully understand the dependency resolution in the OSGi context. Since this is only for the test bundles I guess we can live with it.

kaikreuzer: > The manifest must not contain any Require-Bundle entries

This is indeed only for "real" bundles at runtime - we allow exceptions for test fragments (also check the static analysis tool).

martinvw: Is it only needed for hamcrest, or do the other two need to be included there as well?

Maybe we should update the guidelines of both projects as well because our archetype does not follow the guidelines anymore. And in my experience you should not in general enforce lower standards on your tests, that way it will (slowly) become a mess.

htreu: > Is it only needed for hamcrest, or do the other two need to be included there as well?

From my experience with these libs all three are mandatory to make this work. And from my memory its not only about hamcrest but mockito, too.
The error when removing one or the other from the Required-Bundles is something like:
```
Access restriction: The method 'OngoingStubbing<Bridge>.thenReturn(Bridge)' is not API (restriction on required library '/Users/henning/work/smarthome-master2/ws/.metadata/.plugins/org.eclipse.pde.core/.bundle_pool/plugins/org.mockito_1.9.5.v201311280930.jar')
```

PR #3365, opened 2017-05-04
OSGI Console: ability to trigger channel with null event

triller-telekom:


sjka:
Thanks!

PR #3366, opened 2017-05-04
Trigger rules without specific event on any event

### LINKED ISSUE #3362: Rule is not executed when no event name is passed
Environment / device: openHAB 2.0 Build #905 / RaspberryPI 3B with openHABian

Steps to reproduce:
1. Create a new rule responding to the event channel, e.g.
```
rule "Tide Dash button pressed"
when
    Channel "amazondashbutton:dashbutton:Tide:press" triggered
then
    sendNotification("some@email.com", "Amazon button pressed!")
    sendCommand(Vacuum_Dock, ON)
end
```
or
```
 rule "Xiaomi Switch"
 when
     Channel "mihome:sensor_switch:<ID>:button" triggered
 then
     var actionName = receivedEvent.getEvent()
     switch(actionName) {
         case "CLICK": {
             <ACTION>
         }
         case "DOUBLE_CLICK": {
             <ACTION>
         }
         case "LONG_CLICK_PRESS": {
             <ACTION>
         }
         case "LONG_CLICK_RELEASE": {
             <ACTION>
         }
     }
 end
```
2. Try to trigger the events and see the results

Expected result: The rule is triggered and performed actions should be logged in the console.

Actual result: The rule doesn't trigger, nothing is being logged.

Note that both rules worked for me before I updated the distro to the latest one.

When I pass event name to the `when` section, e.g.
```
Channel "mihome:sensor_switch:<ID>:button" triggered CLICK
```
everything works properly as expected.

triller-telekom:
Fixes #3362

### DIFF
--- a/bundles/model/org.eclipse.smarthome.model.rule.runtime/src/org/eclipse/smarthome/model/rule/runtime/internal/engine/RuleTriggerManager.java
+++ b/bundles/model/org.eclipse.smarthome.model.rule.runtime/src/org/eclipse/smarthome/model/rule/runtime/internal/engine/RuleTriggerManager.java
@@ -217,8 +217,12 @@ public class RuleTriggerManager {
                     for (EventTrigger t : rule.getEventtrigger()) {
                         if (t instanceof EventEmittedTrigger) {
                             EventEmittedTrigger et = (EventEmittedTrigger) t;
-                            if (et.getTrigger() != null) {
-                                if (et.getChannel().equals(channel) && et.getTrigger().equals(event)) {
+
+                            if (et.getChannel().equals(channel)) {
+                                if (et.getTrigger() == null) {
+                                    // if the rule does not have a specific event , execute it on any event
+                                    result.add(rule);
+                                } else if ((et.getTrigger() != null && et.getTrigger().equals(event))) {
                                     result.add(rule);
                                 }
                             }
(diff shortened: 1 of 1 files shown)


bundles/model/org.eclipse.smarthome.model.rule.runtime/src/org/eclipse/smarthome/model/rule/runtime/internal/engine/RuleTriggerManager.java
```diff
@@ -217,8 +217,12 @@ public RuleTriggerManager(Injector injector) {
...
                         if (t instanceof EventEmittedTrigger) {
                             EventEmittedTrigger et = (EventEmittedTrigger) t;
-                            if (et.getTrigger() != null) {
-                                if (et.getChannel().equals(channel) && et.getTrigger().equals(event)) {
+
+                            if (et.getChannel().equals(channel)) {
+                                if (et.getTrigger() == null) {
+                                    // if the rule does not have a specific event , execute it on any event
+                                    result.add(rule);
+                                } else if ((et.getTrigger() != null && et.getTrigger().equals(event))) {
```
sjka: The first part is redundant: you already _know_ that `et.getTrigger()` is not `null`.

Anyway, why so complicated? Wouldn't a 

```java 
if (et.getChannel().equals(channel) && (et.getTrigger() == null || et.getTrigger().equals(event))) {
```
have done the job nicely without duplicating the (admittedly trivial) inner block?

triller-telekom:
Indeed, i simplified the nested if as you suggested. I was thinking too complicated ;)

PR #3371, opened 2017-05-04
Add kickstart instructions for Paper UI contributors

timonreinhard:
Allow first-time frontend contributors to get started quickly by
leveraging `npm` for the frontend build, thus avoiding a global `gulp`
installation.

Also adds a missing peer dependency to support npm >= 3.

### DIFF
--- a/extensions/ui/org.eclipse.smarthome.ui.paper/pom.xml
+++ b/extensions/ui/org.eclipse.smarthome.ui.paper/pom.xml
@@ -59,10 +59,10 @@
           <execution>
             <id>gulp build</id>
             <goals>
-              <goal>gulp</goal>
+              <goal>npm</goal>
             </goals>
             <configuration>
-              <arguments>default</arguments>
+              <arguments>run build</arguments>
             </configuration>
           </execution>
 
--- a/docs/documentation/development/notes.md
+++ b/docs/documentation/development/notes.md
@@ -28,15 +28,15 @@ PaperUI development (JS/HTML)
 The initial setup of PaperUI can be done in two ways:
 
 1. By running the maven build.
-2. Running an 'npm install' followed by 'gulp' build.
+2. Running an `npm install` followed by `npm run build`.
 
 The JS/HTML files of PaperUI are packaged using the 'Gulp' plugin. There are two important gulp tasks needed for development. For normal development you need to run the following command on console:
 
-    gulp serve --development
+    npm start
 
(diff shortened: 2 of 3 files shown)


sjka:
Thanks!

For jasmin-core, there already has been filed CQ 13290.
See also #3206 which also added it.

timonreinhard:
Ok, I guess I'll rebase against master once #3206 has been merged and leave this open meanwhile, right?

sjka:
Yes, makes sense.

kaikreuzer:
@timonreinhard #3206 has just been merged!

timonreinhard:
Rebased against master

extensions/ui/org.eclipse.smarthome.ui.paper/README.md
```diff
@@ -18,6 +18,18 @@ The following features are implemented:
 * Configuration of services
 * Event support for item state updates, Thing status updates and new inbox entries
 
+## Development
```
kaikreuzer: Sorry, I didn't look at this earlier.
Please note that the README.md is the end user documentation and not meant for developers - you should thus add this section rather to https://github.com/eclipse/smarthome/blob/master/docs/documentation/development/notes.md#paperui-development-jshtml.

timonreinhard: Thanks for pointing out. Actually I totally missed that documentation, which was the reason for my PR in the first place. How about linking those documents?

timonreinhard:
I just moved the docs and updated the commit message to better reflect the actual changes.

extensions/ui/org.eclipse.smarthome.ui.paper/pom.xml
```diff
@@ -59,10 +59,10 @@
           <execution>
             <id>gulp build</id>
```
kaikreuzer: This id seems to be slightly outdated, would you agree?

timonreinhard: I couldn't decide to rename it as actually it's still the "gulp build" that is run and I felt uncertain if that follows any convention. That said, I'd be happy to rename that to `frontend build` or whatever you prefer.

kaikreuzer: ok, if it still is gulp after all, let's simply keep it - I do not really care much :-)

Issue #3367, opened 2017-05-04
Label returned by UI backend: when including the state value ?


lolodomo:
https://github.com/eclipse/smarthome/blob/master/bundles/ui/org.eclipse.smarthome.ui/src/main/java/org/eclipse/smarthome/ui/internal/items/ItemUIRegistryImpl.java#L290

The label returned by UI backend will first use, if defined, the pattern defined in the item label. If not defined, it will look for a pattern in channel state description. If found, this pattern will be used to include the state as value in the label. If not, the value will not be included in the label.

This leads to a situation where the user cannot decide to not have the value as the channel has precedence.

At the reverse, the current code has of course an interest, that is a value will be displayed for such channel, even if the user did not care of the format in the label.

So 2 questions:
* when using the pattern defined in the state description ?
* how to prevent having any value displayed ?

lolodomo:
Issue #3248 has a direct link to this subject.
Issue #3294 has a partial link to this subject.

lolodomo:
Here is one proposal: we could decide that looking for a pattern in the channel state description will occur only if the item label contains "[]".
As a result:
* item label "text" will lead to "text" whatever the channel state description
* item label "text [%.1f"] will lead to "text [21.2]" whatever the channel state description
* item label "text []" will lead to "text" if there is no pattern in the channel state description
* item label "text []" will lead to "text [21.23]" if there is a pattern "%.2f" in the channel state description

Then for automatic creation of items, it will make sense to create label ending with " []".

Please comment.

sjka:
I see your point.

> Then for automatic creation of items (auto-link mode of Paper UI), it will make sense to create item with label ending with " []".

The channel defines such a format and from my point of view the "default" behavior should be that it gets used. If the user has to add empty brackets just to find out if maybe some default is hidden somewhere sounds a bit surprising to me.

How about turning it the other way round? By added empty `[]` you would override the default format, ending up with none (generously ignoring the difference between "empty" and "not set").

lolodomo:
I fully agree with you, that is better to do it in your way. The other advantage is that we have nothing to change in the automatic creation of items.

lolodomo:
@sjka : what would be your suggestion for #3248 ?

sjka:
fixed with #3372

PR #3372, opened 2017-05-04
UI: "[]" at end of label to disable the display of value

lolodomo:
Implements #3367

### DIFF
--- a/bundles/ui/org.eclipse.smarthome.ui.test/src/test/java/org/eclipse/smarthome/ui/internal/items/ItemUIRegistryImplTest.java
+++ b/bundles/ui/org.eclipse.smarthome.ui.test/src/test/java/org/eclipse/smarthome/ui/internal/items/ItemUIRegistryImplTest.java
@@ -24,6 +24,7 @@ import org.eclipse.smarthome.core.library.types.OnOffType;
 import org.eclipse.smarthome.core.library.types.PercentType;
 import org.eclipse.smarthome.core.library.types.StringType;
 import org.eclipse.smarthome.core.types.State;
+import org.eclipse.smarthome.core.types.StateDescription;
 import org.eclipse.smarthome.core.types.UnDefType;
 import org.eclipse.smarthome.model.sitemap.Sitemap;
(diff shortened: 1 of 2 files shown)


lolodomo:
I think I tested all combinations, label on sitemap element or only on item, format not set, format set and format empty.
I have not changed the order of priority but just added the empty format "[]".
The highest priority is the format in the label of the sitemap element; then it is the format in the item element and finally it is the format on the channel state description.

sjka:
Looks like it would do what we think it should, on a first glance. 

However, we have this wonderful [ItemUIRegistryImplTest](https://github.com/eclipse/smarthome/blob/8ba0e7cdd141595e5d7fbe8a3d7716ce63cba799/bundles/ui/org.eclipse.smarthome.ui.test/src/test/java/org/eclipse/smarthome/ui/internal/items/ItemUIRegistryImplTest.java) in the corresponding test project. I'd love to see your assumptions (similar like [here](https://github.com/eclipse/smarthome/issues/3367#issuecomment-299148973), just now the other way round) being manifested there as test case(s). Just to be sure it really does what we think and to prevent somebody breaking it in future accidentally...

lolodomo:
Independently of my change, there are missing test cases. I will add 2 new tests.

lolodomo:
Done.

sjka:
That looks good, thanks!

However, isn't the test-case for the purpose of this change missing? (i.e. pattern in state description, overriding it with `[]` to nothing)

lolodomo:
In fact, the code ignores the state description in this case. So don't matter.  But I agree it makes sense for the test to define a state description with pattern to show that it is ignored. I will update my test case and squash everything.

lolodomo:
@sjka : test updated and everything squashed.

sjka:
Thanks!

> [...] to show that it is ignored

From my point of view the primary goal of such tests are making sure nobody breaks it again accidentally. So it's about regressions. The more adoption this project gets, the more we should try to not break stuff all the time.

lolodomo:
Thank you for the merge.

Remains to update the documentation to explain the new option.

PR #3380, opened 2017-05-05
Basic UI: update the value of a text element only when expected

### LINKED ISSUE #3294: Basic UI: text element updated when not expected
When receiving a SSE event, Basic UI frontend takes the value from the label when the label contains something betwen []; if not, the value is the state value itself.
https://github.com/eclipse/smarthome/blob/master/extensions/ui/org.eclipse.smarthome.ui.basic/web-src/smarthome.js#L1624
But if the backend does not provide something between [] it might be because we don"t want a value displayed. That is the case for example for a string item without [%s] in label (and a channel without pattern / state description).
I am sure there was a reason for that piece of code in frontend but I don't find it and I don't understand why the frontend overrides what is provided by the backend.

lolodomo:
Fix #3294


lolodomo:
@resetnow : another one for you ;)

resetnow:
I'll review it after work but for now — maybe we should move `%has_value%` string replacement to preprocessSnippet?

lolodomo:
You're, I make the change.

extensions/ui/org.eclipse.smarthome.ui.basic/web-src/smarthome.js
```diff
@@ -1225,7 +1229,7 @@
 				}
 			}
 
-			if (_t.hasValue) {
+			if (_t.valueNode !== null && _t.hasValue) {
```
resetnow: Can valueNode ever be null here? It's always present in the markup

lolodomo:
@resetnow : yes, that is the issue I am fixing here. In fact, it only happened when there was no pattern defined in the text element in the sitemap, no pattern defined for the item label and no pattern defined by the channel. That is not the most common case ;)
It was the case for example when using auto-link mode (no pattern defined in the item label) for all string channels of the Sonos binding (generally a binding is not defining a pattern for a string channel).

extensions/ui/org.eclipse.smarthome.ui.basic/web-src/smarthome.js
lolodomo: We have the same test in ControlLink and ControlSelection.
Do you think we should suppress this test in the 3 cases ?

resetnow: Probably yes — I can't imagine a situation where this variable would be null

lolodomo: Changed. Still working.

kaikreuzer:
@lolodomo Could you please rebase this PR?

lolodomo:
Rebase done.

kaikreuzer:
Thanks!

PR #3383, opened 2017-05-05
Improve REST API and enable swagger on ESH

Rallymen007:
Noticed when using OpenHAB that swagger's generated description file, swagger.{yaml/json}, is missing schemas for most API methods. 
For example, here's a snippet for GetItemData
`"/items/{itemname}": {
            "get": {
                "tags": [
                    "items"
                ],
                "summary": "Gets a single item.",
                "description": "",
                "operationId": "getItemData",
                "produces": [
                    "application/json"
                ],
                "parameters": [
                    {
                        "name": "Accept-Language",
                        "in": "header",
                        "description": "language",
                        "required": false,
                        "type": "string"
                    },
                    {
                        "name": "itemname",
                        "in": "path",
                        "description": "item name",
                        "required": true,
                        "type": "string",
                        "pattern": "[a-zA-Z_0-9]*"
                    }
                ],
                "responses": {
                    "200": {
                        "description": "OK"
                    },
                    "404": {
                        "description": "Item not found"
                    }
                }
            },`
Notice that there's no schema in the responses block.

Now, this is a minor issue if swagger is only used to document the API via swagger-ui, as the missing return schema won't stop the UI from working properly and executing the interactive examples. But should you try to leverage [swagger-codegen](https://github.com/swagger-api/swagger-codegen) to generate clients to use the OH/ESH REST interface you'll find that all methods have a void return type.

This PR aims at fixing this issue. I've noticed most API methods have a response type definition in the ApiOperation annotation, but it is overriden by ApiResponses, so for those I've simply copied the already defined data.

I've also edited the ESH launch configuration to start the swagger related bundles, so swagger.json is also available for testing in ESH. The edited smarthome.cfg introduces the configuration parameters for these bundles, inspired from those found in OH2, with the addition of the required version parameter (that should be added to OH2 as well).


kaikreuzer:
Great, thanks a lot, @Rallymen007!
Please note though that in order to be allowed to accept your PR we have to ask you to [sign an ECA](https://www.eclipse.org/legal/ecafaq.php) and have a "signed-off-by" footer on your commit message. It would be wonderful if you could do so!

Rallymen007:
After a bit of fighting with git I've added the signed-off-by footer and got an ECA. Don't know why the ip-validation check doesn't validate the footer though.

kaikreuzer:
@Rallymen007 I think the problem is that this PR (your branch) contains 5 commits and only 3 of them show a signed-off-by footer. Could you [squash your commits into a single one](https://github.com/ginatrapani/todo.txt-android/wiki/Squash-All-Commits-Related-to-a-Single-Issue-into-a-Single-Commit)?

Rallymen007:
Good find! I think that worked 😄

Issue #3376, opened 2017-05-04
LIFX binding causes endless loop when network disappears


kaikreuzer:
Regularly when doing demos, leaving the runtime active, but shutting down the network (or switching the Wifi), ESH hangs in an endless loop in the LIFX binding, causing 100% CPU load:

```
21:41:43.523 [ERROR] [ternal.LifxLightCommunicationHandler] - An exception occurred while writing data : 'Can't assign requested address'
21:41:43.523 [ERROR] [ternal.LifxLightCommunicationHandler] - An exception occurred while writing data : 'Can't assign requested address'
21:41:43.523 [ERROR] [ternal.LifxLightCommunicationHandler] - An exception occurred while writing data : 'Can't assign requested address'
21:41:43.523 [ERROR] [ternal.LifxLightCommunicationHandler] - An exception occurred while writing data : 'Can't assign requested address'
21:41:43.523 [ERROR] [ternal.LifxLightCommunicationHandler] - An exception occurred while writing data : 'Can't assign requested address'
21:41:43.523 [ERROR] [ternal.LifxLightCommunicationHandler] - An exception occurred while writing data : 'Can't assign requested address'
21:41:43.523 [ERROR] [ternal.LifxLightCommunicationHandler] - An exception occurred while writing data : 'Can't assign requested address'
21:41:43.523 [ERROR] [ternal.LifxLightCommunicationHandler] - An exception occurred while writing data : 'Can't assign requested address'
21:41:43.523 [ERROR] [ternal.LifxLightCommunicationHandler] - An exception occurred while writing data : 'Can't assign requested address'
21:41:43.523 [ERROR] [ternal.LifxLightCommunicationHandler] - An exception occurred while writing data : 'Can't assign requested address'
21:41:43.523 [ERROR] [ternal.LifxLightCommunicationHandler] - An exception occurred while writing data : 'Can't assign requested address'
21:41:43.523 [ERROR] [ternal.LifxLightCommunicationHandler] - An exception occurred while writing data : 'Can't assign requested address'
```
@wborn is there an easy way to fix this?

wborn:
I haven't seen this myself but I'll check if I can reproduce it by disabling my network.

wborn:
@kaikreuzer Yes I can reproduce it an already have a fix by changing the exception handling. There are some loops that just won't quit unless the right result has been achieved.

[...]

PR #3384, opened 2017-05-05
Add LIFX signalstrength channel, improve online/offline detection and exception handling

wborn:
This PR adds a new `signalstrength` channel to all LIFX light Things. The channel contains the received signal strength at a LIFX light. It reuses the `system.signal-strength` typeId and is based on the `signal` value in the [StateWifiInfo](https://lan.developer.lifx.com/docs/device-messages#section-statewifiinfo-17) packet. The binding will only poll for this information when the `signalstrength` channel is linked to an item.

The lights can also send a [StateHostInfo](https://lan.developer.lifx.com/docs/device-messages#section-statehostinfo-13) packet but for some reason it always has a `signal` value of 0. The binding had mixed up the TYPE values in `GetHostInfoRequest` and `StateHostInfoResponse` so I've corrected that.

Furthermore the light online/offline detection has been improved. The `LifxLightOnlineStateUpdater` now piggybacks on all received packets. It assumes the light is online for as long as some packets are received (due to light state polling). It starts its own polling when no packets have been received for a while. This makes online/offline detection more stable while reducing network traffic at the same time.

Finally the exception handling has been improved for #3376. The binding no longer endlessly loops when packets are sent and exceptions occur. Instead is now sets the light offline and checks once every 15 seconds if a connection to the light can be made. The binding now logs such exceptions using debug level.

### DIFF
--- a/extensions/binding/org.eclipse.smarthome.binding.lifx/ESH-INF/thing/colorirlight.xml
+++ b/extensions/binding/org.eclipse.smarthome.binding.lifx/ESH-INF/thing/colorirlight.xml
@@ -9,6 +9,7 @@
             <channel id="color" typeId="color" />
             <channel id="temperature" typeId="temperature" />
             <channel id="infrared" typeId="infrared" />
+            <channel id="signalstrength" typeId="system.signal-strength" />
         </channels>
         <config-description-ref uri="thing-type:lifx:light" />
     </thing-type>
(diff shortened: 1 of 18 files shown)


wborn:
Thanks for merging this. Hopefully it also allows you to demo the CPU friendliness. :-)

PR #3386, opened 2017-05-06
[ntp] Increase timeout & change loglevel

### LINKED ISSUE #2174: [ntp] Is timeout on call to gettime too short?
@marcelrv 

I'm seeing these warnings pretty regularly in my log file.  I expect it might be because the timeout value, `NTP_TIMEOUT` in NtpHandler, is set pretty short at 5000 ms.  I see this using a variety of different NTP pools.

```

2016-09-13 21:38:39.895 [WARN ] [thome.binding.ntp.handler.NtpHandler] - ntp:ntp:local couldn't establish network connection [host '0.pool.ntp.org'] -> returning current sytem time instead.
2016-09-14 01:15:39.892 [WARN ] [thome.binding.ntp.handler.NtpHandler] - ntp:ntp:local couldn't establish network connection [host '0.pool.ntp.org'] -> returning current sytem time instead.

```

marcelrv:
closes #2174

Signed-off-by: Marcel Verpaalen <marcel@verpaalen.com>


extensions/binding/org.eclipse.smarthome.binding.ntp/src/main/java/org/eclipse/smarthome/binding/ntp/handler/NtpHandler.java
```diff
@@ -227,12 +227,12 @@ public long getTime(String hostname) {
         } catch (UnknownHostException uhe) {
             String msg = getThing().getUID().toString() + " the given hostname '" + hostname
                     + "' of the timeserver is unknown -> returning current sytem time instead.";
-            logger.warn(msg);
+            logger.info(msg);
```
kaikreuzer: I think we should go directly for debug level as the message is anyhow logged as a thing status update.

extensions/binding/org.eclipse.smarthome.binding.ntp/src/main/java/org/eclipse/smarthome/binding/ntp/handler/NtpHandler.java
```diff
@@ -56,7 +56,7 @@
     private Logger logger = LoggerFactory.getLogger(NtpHandler.class);
 
     /** timeout for requests to the NTP server */
-    private static final int NTP_TIMEOUT = 10000;
+    private static final int NTP_TIMEOUT = 30000;
```
kaikreuzer: If the server really takes 30 secs to respond, I wonder what value the delivered time still has :-/
I would actually think that everything that takes longer than 10 secs is indeed an issue - maybe the user should look for a different ntp server in such a situation.

extensions/binding/org.eclipse.smarthome.binding.ntp/src/main/java/org/eclipse/smarthome/binding/ntp/handler/NtpHandler.java
```diff
@@ -227,12 +227,12 @@ public long getTime(String hostname) {
...
             String msg = getThing().getUID().toString() + " the given hostname '" + hostname
                     + "' of the timeserver is unknown -> returning current sytem time instead.";
-            logger.warn(msg);
+            logger.info(msg);
             updateStatus(ThingStatus.ONLINE, ThingStatusDetail.COMMUNICATION_ERROR, msg);
         } catch (IOException ioe) {
             String msg = getThing().getUID().toString() + " couldn't establish network connection [host '" + hostname
                     + "'] -> returning current sytem time instead.";
-            logger.warn(msg);
+            logger.info(msg);
```
kaikreuzer: dito

kaikreuzer:
It seems you also need to adapt the test cases:
```
Results :
Failed tests: 
  NtpOSGiTest.testDateTimeChannelWithUnknownHost:344->assertCommunicationError:485->initialize:394->JavaOSGiTest.waitForAssert:340->JavaOSGiTest.waitForAssert:388->lambda$0:397 
Expected: is <ONLINE>
     but: was <OFFLINE>
  NtpOSGiTest.testStringChannelWithUnknownHost:349->assertCommunicationError:487->initialize:394->JavaOSGiTest.waitForAssert:340->JavaOSGiTest.waitForAssert:388->lambda$0:397 
Expected: is <ONLINE>
     but: was <OFFLINE>
Tests run: 16, Failures: 2, Errors: 0, Skipped: 2
```

extensions/binding/org.eclipse.smarthome.binding.ntp/src/main/java/org/eclipse/smarthome/binding/ntp/handler/NtpHandler.java
```diff
@@ -227,12 +227,12 @@ public long getTime(String hostname) {
...
-            logger.warn(msg);
-            updateStatus(ThingStatus.ONLINE, ThingStatusDetail.COMMUNICATION_ERROR, msg);
+            logger.debug(msg);
+            updateStatus(ThingStatus.OFFLINE, ThingStatusDetail.COMMUNICATION_ERROR, msg);
         } catch (IOException ioe) {
             String msg = getThing().getUID().toString() + " couldn't establish network connection [host '" + hostname
                     + "'] -> returning current sytem time instead.";
-            logger.warn(msg);
+            logger.debug(msg);
             updateStatus(ThingStatus.ONLINE, ThingStatusDetail.COMMUNICATION_ERROR, msg);
```
sjka: As you changed the thing status to OFFLINE above, shouldn't it be OFFLINE here too?

extensions/binding/org.eclipse.smarthome.binding.ntp/src/main/java/org/eclipse/smarthome/binding/ntp/handler/NtpHandler.java
```diff
@@ -227,12 +227,12 @@ public long getTime(String hostname) {
         } catch (UnknownHostException uhe) {
             String msg = getThing().getUID().toString() + " the given hostname '" + hostname
                     + "' of the timeserver is unknown -> returning current sytem time instead.";
-            logger.warn(msg);
-            updateStatus(ThingStatus.ONLINE, ThingStatusDetail.COMMUNICATION_ERROR, msg);
+            logger.debug(msg);
+            updateStatus(ThingStatus.OFFLINE, ThingStatusDetail.COMMUNICATION_ERROR, msg);
```
sjka: Did you change this by intention? Then why did you leave it untouched for an IOException? The result effectively is the same, so I think it should be OFFLINE then too.

As the JavaDoc above of this method currently states, this handler is implemented in a way that it fall back to the current system time if no connection could be established. IMHO this is awkward, so I'd second your change to switch it to OFFLINE/COMMUNICATION_ERROR - but do it consistently then.

kaikreuzer:
@marcelrv Did you see @sjka's comments?

kaikreuzer:
@marcelrv ping!

extensions/binding/org.eclipse.smarthome.binding.ntp/src/main/java/org/eclipse/smarthome/binding/ntp/handler/NtpHandler.java
marcelrv: @sjka yes, I think that may be better. 
When I developed it, I was not sure if state updates would come through if it is set to offline.
But I think the binding can still provide state updates when online, so indeed better.

marcelrv:
@kaikreuzer totally missed I still had open PR here..
made update to address  @sjka comments.

PR #3388, opened 2017-05-07
Icons for 'battery-on' and 'battery-off' states

cweitkamp:
The documentation for [channel definitions](https://www.eclipse.org/smarthome/documentation/development/bindings/thing-definition.html#channels) proposes the use of systemwide reusabel channels. A default channel for `system.low-battery` exists which has the item type `Switch`. But there doesn't exist a corresponding icon for the states ON/OFF.

I suggest to use battery-10 for ON
ON

and battery-90 for OFF.
OFF


kaikreuzer:
Sounds ok to me. But please note that the file names should be all lowercase, e.g. battery-on.svg. Could you adapt this please?

cweitkamp:
Yes of course. No problem.

PR #3394, opened 2017-05-08
Re-added accidentially removed constructor

triller-telekom:


bundles/core/org.eclipse.smarthome.core/src/main/java/org/eclipse/smarthome/core/items/GroupItem.java
```diff
@@ -37,16 +37,22 @@
...
      * Creates a plain GroupItem
-     * 
+     *
      * @param name name of the group
      */
     public GroupItem(String name) {
         this(name, null, null);
     }
 
+    // Apparently OH1 Compatibility Layer needs this constructor
```
kaikreuzer: remove this comment - the previous PR broke OUR API and we have no idea who might be using it already - the openHAB compat layer is merely one example.

PR #3395, opened 2017-05-08
Soft fallback to plain GroupItem instead of IllegalArgumentException

triller-telekom:


bundles/core/org.eclipse.smarthome.core/src/main/java/org/eclipse/smarthome/core/items/GroupItem.java
```diff
@@ -59,8 +59,10 @@ public GroupItem(String name, GenericItem baseItem) {
     public GroupItem(String name, GenericItem baseItem, GroupFunction function) {
         super(TYPE, name);
 
-        if ((baseItem != null && function == null) || (baseItem == null && function != null)) {
-            throw new IllegalArgumentException("baseItem AND function have to be passed as arguments");
+        // we only allow GroupItem with BOTH, baseItem AND function set, or NONE of them set
+        if (baseItem == null || function == null) {
+            baseItem = null;
+            function = null;
```
htreu: please don´t reassign method parameters. use `this.baseItem`instead and the `else`case for the other assignement.

Issue #3377, opened 2017-05-04
[Test failures] FSInternetRadioHandlerOSGiTest again


sjka:
Travis hit this in (completely unrelated #3371):

```
verify OFFLINE Thing-status and NULL Item-state when the HTTP response cannot be parsed correctly(org.eclipse.smarthome.binding.fsinternetradio.test.FSInternetRadioHandlerOSGiTest)  Time elapsed: 10.605 sec  <<< FAILURE!
java.lang.AssertionError: The item's state was not updated correctly when the HTTP response cannot be parsed
Expected: is <NULL>
     but: was <0>
	at org.hamcrest.MatcherAssert.assertThat(MatcherAssert.java:20)
	at org.junit.Assert.assertThat(Assert.java:865)
	at org.junit.Assert$assertThat$0.callStatic(Unknown Source)
	at org.eclipse.smarthome.binding.fsinternetradio.test.FSInternetRadioHandlerOSGiTest$_verify_OFFLINE_Thing-status_and_NULL_Item-state_when_the_HTTP_response_cannot_be_parsed_correctly_closure4.doCall(FSInternetRadioHandlerOSGiTest.groovy:200)
	at org.eclipse.smarthome.binding.fsinternetradio.test.FSInternetRadioHandlerOSGiTest$_verify_OFFLINE_Thing-status_and_NULL_Item-state_when_the_HTTP_response_cannot_be_parsed_correctly_closure4.doCall(FSInternetRadioHandlerOSGiTest.groovy)
	at sun.reflect.GeneratedMethodAccessor31.invoke(Unknown Source)
	at sun.reflect.DelegatingMethodAccessorImpl.invoke(DelegatingMethodAccessorImpl.java:43)
	at java.lang.reflect.Method.invoke(Method.java:498)
	at org.codehaus.groovy.reflection.CachedMethod.invoke(CachedMethod.java:90)
	at groovy.lang.MetaMethod.doMethodInvoke(MetaMethod.java:324)
	at org.codehaus.groovy.runtime.metaclass.ClosureMetaClass.invokeMethod(ClosureMetaClass.java:278)
	at groovy.lang.MetaClassImpl.invokeMethod(MetaClassImpl.java:1016)
	at org.codehaus.groovy.runtime.callsite.PogoMetaClassSite.call(PogoMetaClassSite.java:39)
	at org.codehaus.groovy.runtime.callsite.CallSiteArray.defaultCall(CallSiteArray.java:45)
[...]

PR #3398, opened 2017-05-08
Incorrect assertions are removed from FSInternetRadioHandlerOSGiTest.

m-memova:
All 'assertThat item.getState() is UnDefType.NULL' are removed since they cannot always be true.

see: https://github.com/eclipse/smarthome/issues/3377

Some small adaptions are made.

### DIFF
--- a/extensions/binding/org.eclipse.smarthome.binding.fsinternetradio.test/src/test/groovy/org/eclipse/smarthome/binding/fsinternetradio/test/FSInternetRadioHandlerOSGiTest.groovy
+++ b/extensions/binding/org.eclipse.smarthome.binding.fsinternetradio.test/src/test/groovy/org/eclipse/smarthome/binding/fsinternetradio/test/FSInternetRadioHandlerOSGiTest.groovy
@@ -10,12 +10,11 @@ package org.eclipse.smarthome.binding.fsinternetradio.test
 import static org.hamcrest.CoreMatchers.*
 import static org.junit.Assert.*
 
-import org.apache.commons.lang.StringUtils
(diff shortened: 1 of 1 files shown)


extensions/binding/org.eclipse.smarthome.binding.fsinternetradio.test/src/test/groovy/org/eclipse/smarthome/binding/fsinternetradio/test/FSInternetRadioHandlerOSGiTest.groovy
```diff
@@ -155,50 +166,38 @@ public class FSInternetRadioHandlerOSGiTest extends OSGiTest{
...
 
         Thing radioThing = initializeRadioThing(DEFAULT_COMPLETE_CONFIGURATION)
-        testRadioThingWithConfiguration(radioThing)
+        testRadioThingConsideringConfiguration(radioThing)
 
         // turn-on the radio
         turnTheRadioOn(radioThing)
 
+        ChannelUID modeChannelUID = radioThing.getChannel(modeChannelID).getUID()
+        Item modeTestItem = initializeItem(modeChannelUID, "mode", acceptedItemType)
```
htreu: Is it necessary to initialise the item here? there is no assertion on the resulting item.

m-memova: You are right, I don't need the item. That line is deleted and the PR is updated.

Issue #3408, opened 2017-05-09
ItemEventFactory has to use toFullString for GroupItemStateChangeEvent


BenediktNiehues:
When creating GroupItemStateChangeEvent the ItemEventFactory uses toString() on the new and old state which would end up in an IllegalArgumentException during deserialization when the state is a RawType.

lolodomo:
PR is merged. Issue to be closed.

sjka:
Ah, wrong syntax for the auto-close... Thanks for the hint!

PR #3409, opened 2017-05-09
use toFullString when creating GroupItemStateChangeEvent

BenediktNiehues:
fix for #3408


sjka:
Could you please also adapt [ItemEventFactoryTest](https://github.com/eclipse/smarthome/blob/3ae5d7b1b284f2bde6199bfe8dad950fd38cae39/bundles/core/org.eclipse.smarthome.core.test/src/test/groovy/org/eclipse/smarthome/core/items/events/ItemEventFactoryTest.groovy#L60)? (And maybe even make it _really_ test this behavior while you're there?  Although I have to admit that the culpable negligence is not on your side...)

BenediktNiehues:
@sjka I added a test...

sjka:
Thanks. Just one question, as this is meant to be a regression test:

When just using the test case from your PR, but leaving ItemEventFactory.java as it was before - what would you expect to happen when running ItemEventFactoryTest? 😉

BenediktNiehues:
@sjka in factory.createEvent I expect an IllegalArgumentException and the test will fail.

sjka:
Thanks

Issue #601, opened 2015-12-03
Introduce an optional unit in Number item


kaikreuzer:
migrated from Bugzilla [#434006](https://bugs.eclipse.org/bugs/show_bug.cgi?id=434006)
status ASSIGNED severity _enhancement_ in component _Core_ for _---_
Reported in version _unspecified_ on platform _All_
Assigned to: gael lhopital

On 2014-05-02 11:02:11 -0400, Kai Kreuzer wrote:

> To facilitate type determination, value formatting and calculations, an optional unit property should be introduced to number items. This could e.g. make use of Eclipse UOMo.

On 2014-10-16 03:16:42 -0400, Kai Kreuzer wrote:

> **\* Bug 447299 has been marked as a duplicate of this bug. ***

On 2015-03-19 04:02:29 -0400, gael lhopital wrote:

> (In reply to Kai Kreuzer from comment # 0)
> 
> > To facilitate type determination, value formatting and calculations, an
> > optional unit property should be introduced to number items. This could e.g.
> > make use of Eclipse UOMo.
> 
> I'm ok to work on this point. I had a first look at UOMo and this introduces a dependency an issue with the package com.ibm.icu (version 1.50 needed while 1.52 is present). On the other hand, I found the JScience package that relies also on JSR-275 spec. Is it possible to choose go with this one in an eclipse project ?
> 
> I started my tests by adding a unit member in DecimalType, but wondering what would be the best approach  : adding unit capability to DecimalType, subclassing DecimalType to something like MeasureType, or putting unit in NumberItem ?

On 2015-03-20 11:51:17 -0400, Kai Kreuzer wrote:

> Gael,
> 
> > I found the JScience package that relies also on JSR-275 spec. Is it possible to choose go with this one in an eclipse project ?
> 
> JSR-275 has never made it to be an official Java-API. Its "living" successor is JSR-363 (https://www.jcp.org/en/jsr/detail?id=363) and we were already in contact with the spec lead about it and filed a CQ for the Eclipse IP team about it. You can find its current reference implementation at https://github.com/unitsofmeasurement/unit-api.
> I am not sure though, how it feels from a developer point of view in comparison to JSR-275. From your implementation it seems that JSR-275 is quite straight forward and useful. Could you check if JSR-363 is similar or if it is lacking anything you would expect?
> 
> > I started my tests by adding a unit member in DecimalType, but wondering what would be the best approach
> 
[...]

PR #3410, opened 2017-05-09
add measurement feature to target platform

maggu2810:
Related to: https://github.com/eclipse/smarthome/pull/1857#issuecomment-299477230


clinique:
Could you please upgrade uom-se to 1.0.4 in target platform ?
We currently have the issue referenced [here](https://github.com/unitsofmeasurement/uom-se/issues/139)

kaikreuzer:
We cannot easily upgrade, we need a CQ process for every change.
Note that there is already a 1.0.6 version available.
If it isn't blocking, I would leave it as it is for the moment - once we merge #601 and start using it, we can then upgrade to the latest available uom-se version.

clinique:
Sure, it can wait - it's a reminder for when the time comes.

PR #2258, opened 2016-10-05
Add CRITICAL flag to config parameters

cdjackson:
This adds a `critical` flag to config parameters. This flag should indicate that a parameter is considered critical, or dangerous in that it could damage the system if used incorrectly. A GUI might provide an "Are you sure" prompt when a user changes such a parameter to provide a degree of protection against inadvertent misuse.

(note this isn't fully tested - I wanted to get feedback first)

Signed-off-by: Chris Jackson chris@cd-jackson.com

PR #3415, opened 2017-05-10
fix verify in config description parameter localization

sjka:
relates to #2258


sjka:
eclipse 4.4 repository apparently is fixed, so Travis is happy again.

PR #3418, opened 2017-05-10
simplified code by using normal constants instead of an enum

kaikreuzer:
See https://github.com/openhab/openhab2-addons/pull/2009/files#r111239598


kaikreuzer:
Tests are passing now :-)

PR #3419, opened 2017-05-10
added infos on the expected granularity of channel types

kaikreuzer:

### DIFF
--- a/docs/documentation/development/bindings/thing-definition.md
+++ b/docs/documentation/development/bindings/thing-definition.md
@@ -37,7 +37,16 @@ Thing types are listed by default, unless specified otherwise. Hiding thing type
 
 ## Channels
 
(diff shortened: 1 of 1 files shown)


cdjackson:
> If a Thing measures two temperature values, one for indoor and one for outdoor, this should be modelled as two different channel types.
> Overriding labels of a channel type must only be done if the very same functionality is offered multiple times, e.g. having an actuator with 5 relais, which each is a simple "switch", but you want to individually name the channels (1-5).

I'm a bit unclear what the difference is here. Why do we need to have 2 channels for 2 different temperatures, but it's ok to have a single channel and override the label for switches. Isn't this the same? The fact that I put one temperature sensor inside, and another outside, or in a different room, is this any different to having a power strip with 5 relays and I happen to run one into the loft and one outside?

To me, two temperature sensors do offer exactly the same functionality - they just happen to be in different locations so I don't see the need to make this distinction, but I might be missing the reason?

kaikreuzer:
It depends on your device. If you have a device that offers two temperature sensors, then you are right, they can be the same channel type and it is up to the user what he measures with them.
But if the device specifically offers one indoor and one outdoor temperature, the device itself already makes a difference in their semantic - that's why we should keep this differentiation in place within the ESH modelling.

PR #3420, opened 2017-05-11
Added ConsoleCommand for adding/removing tags to/from items

### LINKED ISSUE #1244: Missing option to add/edit Tags via console
Due to the beginning of testing OH2 which is based on ESH one options are missing.

With changing from OH1 to OH2 and the ESH-framework there are now Tags, that can be used in .item-files. 1st seen were the "Homegropups" in demo.items-file: 

`Group Rooms    "Rooms"     <groundfloor>   [ "home-group" ]`

With the integration of other features there grow more Tags like
`[ "homekit:Lightbulb" ]`

At the moment there is no other  way to assign this Tags to items even with an .items-file, not via PaperUI, via HABmin or via console. For the REST there is still an Issue here: https://github.com/eclipse/smarthome/issues/1122

triller-telekom:
Fixes #1244

### DIFF
--- a/bundles/io/org.eclipse.smarthome.io.console/src/main/java/org/eclipse/smarthome/io/console/internal/extension/ItemConsoleCommandExtension.java
+++ b/bundles/io/org.eclipse.smarthome.io.console/src/main/java/org/eclipse/smarthome/io/console/internal/extension/ItemConsoleCommandExtension.java
@@ -33,7 +33,7 @@ public class ItemConsoleCommandExtension extends AbstractConsoleCommandExtension
     private static final String SUBCMD_CLEAR = "clear";
     private static final String SUBCMD_REMOVE = "remove";
     private static final String SUBCMD_ADDTAG = "addTag";
-    private static final String SUBCMD_REMTAG = "remTag";
+    private static final String SUBCMD_REMTAG = "rmTag";
 
     private ItemRegistry itemRegistry;
 
@@ -99,24 +99,6 @@ public class ItemConsoleCommandExtension extends AbstractConsoleCommandExtension
         }
     }
 
-    private void removeTag(String itemName, String tag, Console console) {
-        Item item = itemRegistry.get(itemName);
-        if (item instanceof GenericItem) {
-            GenericItem gItem = (GenericItem) item;
-            gItem.removeTag(tag);
-            Item oldItem = itemRegistry.update(gItem);
-            if (oldItem == null) {
-                console.println("Error: Cannot remove tag " + tag + " from item " + itemName
-                        + " because this item does not belong to a ManagedProvider");
-            } else {
-                console.println("Successfully removed tag " + tag + " from item " + itemName);
-            }
-        } else {
-            console.println(
-                    "Error: Cannot remove tag " + tag + " from item " + itemName + " because it is not a GenericItem");
-        }
-    }
-
     private void addTag(String itemName, String tag, Console console) {
         Item item = itemRegistry.get(itemName);
         if (item instanceof GenericItem) {
@@ -132,6 +114,24 @@ public class ItemConsoleCommandExtension extends AbstractConsoleCommandExtension
         } else {
             console.println(
                     "Error: Cannot add tag " + tag + " to item " + itemName + " because it is not a GenericItem");
+        }
+    }
+
+    private void removeTag(String itemName, String tag, Console console) {
+        Item item = itemRegistry.get(itemName);
+        if (item instanceof GenericItem) {
+            GenericItem gItem = (GenericItem) item;
+            gItem.removeTag(tag);
+            Item oldItem = itemRegistry.update(gItem);
+            if (oldItem == null) {
+                console.println("Error: Cannot remove tag " + tag + " from item " + itemName
+                        + " because this item does not belong to a ManagedProvider");
+            } else {
+                console.println("Successfully removed tag " + tag + " from item " + itemName);
+            }
+        } else {
+            console.println(
+                    "Error: Cannot remove tag " + tag + " from item " + itemName + " because it is not a GenericItem");
         }
     }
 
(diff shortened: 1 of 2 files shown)


bundles/io/org.eclipse.smarthome.io.console/src/main/java/org/eclipse/smarthome/io/console/internal/extension/ItemConsoleCommandExtension.java
```diff
@@ -31,6 +32,8 @@
     private static final String SUBCMD_LIST = "list";
     private static final String SUBCMD_CLEAR = "clear";
     private static final String SUBCMD_REMOVE = "remove";
+    private static final String SUBCMD_ADDTAG = "addTag";
+    private static final String SUBCMD_REMTAG = "remTag";
```
maggu2810: Hm, what do you think about "rmTag"? It is more like the Unix command "rm" to "remove" something.

bundles/io/org.eclipse.smarthome.io.console/src/main/java/org/eclipse/smarthome/io/console/internal/extension/ItemConsoleCommandExtension.java
```diff
@@ -78,6 +99,42 @@ public void execute(String[] args, Console console) {
...
+            } else {
+                console.println("Successfully removed tag " + tag + " from item " + itemName);
+            }
+        } else {
+            console.println(
+                    "Error: Cannot remove tag " + tag + " from item " + itemName + " because it is not a GenericItem");
+        }
+    }
+
+    private void addTag(String itemName, String tag, Console console) {
```
maggu2810: It is more a personal preference, but can we move the add method above the remove method? :wink:

bundles/io/org.eclipse.smarthome.io.console/src/main/java/org/eclipse/smarthome/io/console/internal/extension/ItemConsoleCommandExtension.java
```diff
@@ -78,6 +99,42 @@ public void execute(String[] args, Console console) {
...
+        Item item = itemRegistry.get(itemName);
+        if (item instanceof GenericItem) {
+            GenericItem gItem = (GenericItem) item;
+            gItem.addTag(tag);
```
maggu2810: The add and remove function are identical except the remove(d) / add(ed) string and the addTag / removeTag function.
As we are using Java 8 now, we can use method references, what do you think about moving the most part to a generic function that takes additional the method reference (`void ...(String tag)`) and two string (e.g. "add", "added" vs. "remove", "removed")?
We could leave it separated, but perhaps it could remove redundant code.

triller-telekom: If I want to pass `GenericItem::removeTag` and `GenericItem::addTag`to a method that would handle adding and removing of tags the compiler complains about that `GenericItem` is not a functional interface, which is correct since it has more than one method. Thats why I left it as 2 methods for now, but changed the order. 

But maybe I completely misunderstood how how you wanted to refactor the code to use method references... :confused:

maggu2810: You don't need to use a function interface, you can also use a method directly as an argument to point the functional interface.

Let's me add an example (for static and non-static functions):

```java
import java.util.function.Function;

public class Test {

    public <T, R> R executeFunction(final Function<T, R> func, final T argument) {
        return func.apply(argument);
    }

    public String stringifyInteger(final Integer number) {
        return String.valueOf(number);
    }

    public String stringifyDouble(final Double number) {
        return String.valueOf(number);
    }

    public static String stringifyLong(final Long number) {
        return String.valueOf(number);
    }

    public static void main(final String[] args) {
        final Test test = new Test();

        final String stringifiedInteger = test.executeFunction(test::stringifyInteger, 1);
        final String stringifiedDouble = test.executeFunction(test::stringifyDouble, 1.2);
        final String stringifiedLong = test.executeFunction(Test::stringifyLong, 3L);
    }

}
```

triller-telekom: Thank you very much for this small example. This was very helpful to me.

Instead of `Function` i needed a `Consumer` because I am calling a `void` method, but otherwise it works as you described.

triller-telekom:
One more comment on this PR, independent of method references...

If I add a tag to an item which is defined via an xtext `.items` file, then I have to fetch the item from the registry and get a reference to it. On this reference I can call `addTag(String)`.

Later I tell the registry to `update(item)` itself and in this case it fails because the item doesn't belong to a `ManagedProvider`. So far so good. The problem now is that the item that I have i a reference to the existing one in the registry and thus the `addTag(String)` call has added the String to it. So even though I get an error on the update, the tag exists on the item until the next restart, where the item will be restored from the items file and thus looses the tag again.

Can we leave it like that? I mean someone who operates on the console should be aware that he does an operation on the "open heart" of the system and also gets the error message about that his item is not part of a `ManagedProvider`. But still this is somehow not nice.

I could clone the item before doing the `update(item)` call on the registry but then I have to replace the reference inside the registry in case it belongs to a `ManagedProvider`. Also such a `clone` method doesn't exist (yet) for `GenericItem`.

triller-telekom:
Ok, I have found a solution: I added a dependency to the `ManagedItemProvider` analog to the `ThingResource` which has a dependency to the `ManagedThingProvider`. So I can ask if the item is managed before I actually change it.

PR #3423, opened 2017-05-11
OSGi-ified waiting for xml processing

sjka:
this PR is a first step accounting to #1896.

Its main objective is to get rid of the BundleProcessorVetoManager and replace it with ReadyMarker service registrations per bundle. This results in much cleaner and less entangled code (as you can see, this actually is a "negative" contribution).

While implementing this, I had to visit some places which I couldn't touch without cleaning them up properly. So here is a brief summary of what happened:

* added a ready marker service interface
* registering ready markers for xml thing types
  * simplified the code by eliminating the AbstractAsyncBundleProcessor
  * Xml based providers register a ready marker
  * Dedicated ready markers are registered for each binding's thing types
* Using a thread pool for bundle processing instead of spawning an own thread
* ThingManager and GenericThingProvider listening to the ready marker registrations.
* dumped ResolvedBundleTracker in favor of simply using the plain OSGi BundleTracker
* factored out common aspects of all xml based providers into AbstractXmlProvider
* Split XmlChannelTypeProvider and XmlChannelGroupTypeProvider


maggu2810:
I am very interested in the work you have done.
Hopefully I will find some time next week to have a look at your changes.

bundles/config/org.eclipse.smarthome.config.xml/src/main/java/org/eclipse/smarthome/config/xml/AbstractXmlBasedProvider.java
```diff
@@ -0,0 +1,259 @@
...
+ *
+ * @author Simon Kaufmann - initial contribution, factored out of subclasses
+ *
+ * @param <T_ID> the key type, e.g. ThingTypeUID, ChannelUID, URI,...
+ * @param <T_OBJECT> the object type, e.g. ThingType, ChannelType, ConfigDescription,...
+ */
+public abstract class AbstractXmlBasedProvider<T_ID, T_OBJECT> {
+
+    private class LocalizedKey {
+        public T_ID uid;
```
maggu2810: As this is a generic and also an URI could be an ID (refer to your JavaDoc) we should name the variable `id` instead of `uid`.

bundles/config/org.eclipse.smarthome.config.xml/src/main/java/org/eclipse/smarthome/config/xml/AbstractXmlBasedProvider.java
```diff
@@ -0,0 +1,259 @@
...
+        public String locale;
```
maggu2810: A variable that is set only once by the constructor should be marked final (IMHO)

bundles/config/org.eclipse.smarthome.config.xml/src/main/java/org/eclipse/smarthome/config/xml/AbstractXmlBasedProvider.java
```diff
@@ -0,0 +1,259 @@
...
+
+        public LocalizedKey(T_ID uid, String locale) {
```
maggu2810: `uid` => `id`

bundles/config/org.eclipse.smarthome.config.xml/src/main/java/org/eclipse/smarthome/config/xml/AbstractXmlBasedProvider.java
```diff
@@ -0,0 +1,259 @@
...
+            }
+            if (getClass() != obj.getClass()) {
+                return false;
+            }
+            @SuppressWarnings("unchecked")
+            LocalizedKey other = (LocalizedKey) obj;
+            if (!getOuterType().equals(other.getOuterType())) {
+                return false;
+            }
+            if (locale == null) {
```
maggu2810: Why not using `Objects.equal(local, other.locale)`?

bundles/config/org.eclipse.smarthome.config.xml/src/main/java/org/eclipse/smarthome/config/xml/AbstractXmlBasedProvider.java
```diff
@@ -0,0 +1,259 @@
...
+                if (other.locale != null) {
+                    return false;
+                }
+            } else if (!locale.equals(other.locale)) {
+                return false;
+            }
+            if (uid == null) {
```
maggu2810: Same here, `Objects.equal` should be simpler instead of all the null checks...

bundles/config/org.eclipse.smarthome.config.xml/src/main/java/org/eclipse/smarthome/config/xml/AbstractXmlBasedProvider.java
```diff
@@ -0,0 +1,259 @@
...
+            }
+            if (obj == null) {
+                return false;
+            }
+            if (getClass() != obj.getClass()) {
+                return false;
+            }
+            @SuppressWarnings("unchecked")
+            LocalizedKey other = (LocalizedKey) obj;
+            if (!getOuterType().equals(other.getOuterType())) {
```
maggu2810: The class is private and as fair as I could see an instance is only created internally. Do you really need to check if the key itself is owned by different objects?

bundles/config/org.eclipse.smarthome.config.xml/src/main/java/org/eclipse/smarthome/config/xml/AbstractXmlBasedProvider.java
```diff
@@ -0,0 +1,259 @@
...
+ * Common base class for XML based providers.
+ *
+ * @author Simon Kaufmann - initial contribution, factored out of subclasses
+ *
+ * @param <T_ID> the key type, e.g. ThingTypeUID, ChannelUID, URI,...
+ * @param <T_OBJECT> the object type, e.g. ThingType, ChannelType, ConfigDescription,...
+ */
+public abstract class AbstractXmlBasedProvider<T_ID, T_OBJECT> {
+
+    private class LocalizedKey {
```
maggu2810: WRT to the `getOuterType` comment on `equals` we could perhaps make this a static class.

bundles/config/org.eclipse.smarthome.config.xml/src/main/java/org/eclipse/smarthome/config/xml/AbstractXmlBasedProvider.java
```diff
@@ -0,0 +1,259 @@
...
+                if (other.uid != null) {
+                    return false;
+                }
+            } else if (!uid.equals(other.uid)) {
+                return false;
+            }
+            return true;
+        }
+
+        private AbstractXmlBasedProvider<T_ID, T_OBJECT> getOuterType() {
```
maggu2810: WRT to the `getOuterType` comment on `equals` we could perhaps drop this method (and this will allow us to make the class static).

bundles/config/org.eclipse.smarthome.config.xml/src/main/java/org/eclipse/smarthome/config/xml/AbstractXmlBasedProvider.java
```diff
@@ -0,0 +1,259 @@
...
+     *
+     * @param object the object to return the identifier for
+     * @return the identifier. Must not be <code>null</code>
+     */
+    protected abstract T_ID getIndentifier(T_OBJECT object);
+
+    /**
+     * Create a translated/localized copy of the given object.
+     *
+     * @param bundle the bundle where the config description with its translations originates from
```
maggu2810: Is `config description` correct?

bundles/config/org.eclipse.smarthome.config.xml/src/main/java/org/eclipse/smarthome/config/xml/AbstractXmlBasedProvider.java
```diff
@@ -0,0 +1,259 @@
...
+    /**
+     * Adds an object to the internal list associated with the specified module.
+     * <p>
+     * This method returns silently, if any of the parameters is {@code null}.
+     *
+     * @param bundle the module to which the object is to be added
+     * @param object the object to be added
+     */
+    public final synchronized void add(Bundle bundle, T_OBJECT object) {
+        if (object == null) {
```
maggu2810: Wouldn't it be better (with respect to removal of redundant code) to call `addAll(bundle, Collections.singletonList(object))`

bundles/config/org.eclipse.smarthome.config.xml/src/main/java/org/eclipse/smarthome/config/xml/AbstractXmlBasedProvider.java
```diff
@@ -0,0 +1,259 @@
...
+     * @param key the key which identifies the object
+     * @param locale the locale
+     * @return the object if found, <code>null</code> otherwise
+     */
+    protected final T_OBJECT get(T_ID key, Locale locale) {
+        Collection<Entry<Bundle, List<T_OBJECT>>> objectList = bundleObjectMap.entrySet();
+        if (objectList == null) {
+            return null;
+        }
+        for (Entry<Bundle, List<T_OBJECT>> objects : objectList) {
```
maggu2810: Why not using `... : bundleObjectsMap.entrySet()) {`?

entrySet shouldn't return null but an empty entry set.

bundles/config/org.eclipse.smarthome.config.xml/src/main/java/org/eclipse/smarthome/config/xml/AbstractXmlBasedProvider.java
```diff
@@ -0,0 +1,259 @@
...
+    }
+
+    /**
+     * Gets all available objects.
+     *
+     * @param locale the locale
+     * @return a collection containing all available objects. Never <code>null</code>
+     */
+    protected final synchronized Collection<T_OBJECT> getAll(Locale locale) {
+        List<T_OBJECT> ret = new ArrayList<>(10);
```
maggu2810: Wouldn't it be better to use an `LinkedList` to avoid a lot of reallocations?

bundles/config/org.eclipse.smarthome.config.xml/src/main/java/org/eclipse/smarthome/config/xml/AbstractXmlBasedProvider.java
```diff
@@ -0,0 +1,259 @@
...
+     * <p>
+     * This method returns silently if the module is {@code null}.
+     *
+     * @param bundle the module for which all associated Thing types to be removed
+     */
+    public final synchronized void removeAll(Bundle bundle) {
+        if (bundle == null) {
+            return;
+        }
+        List<T_OBJECT> thingTypes = bundleObjectMap.remove(bundle);
```
maggu2810: Why `thingTypes`? Shouldn't we use `objects`?

bundles/config/org.eclipse.smarthome.config.xml/src/main/java/org/eclipse/smarthome/config/xml/AbstractXmlBasedProvider.java
```diff
@@ -0,0 +1,259 @@
...
+        }
+    }
+
+    private void removeCachedEntries(List<T_OBJECT> objects) {
+        for (T_OBJECT object : objects) {
+            removeCachedEntries(object);
+        }
+    }
+
+    private void removeCachedEntries(T_OBJECT thingType) {
```
maggu2810: Why `thingTypes`? Shouldn't we use `objects`?

bundles/config/org.eclipse.smarthome.config.xml/src/main/java/org/eclipse/smarthome/config/xml/osgi/XmlDocumentBundleTracker.java
```diff
@@ -34,27 +48,26 @@
  * @author Michael Grammling - Initial Contribution
  * @author Benedikt Niehues - Changed resource handling so that resources can be
  *         patched by fragments.
+ * @author Simon Kaufmann - Tracking of remaining bundles
  *
  * @param <T>
  *            the result type of the conversion
  */
+@SuppressWarnings({ "unchecked", "rawtypes" })
```
maggu2810: Do we really suppress all that warnings for the whole class?

maggu2810: Using this changes we don't need to supress warnings (perhaps `?` could be changed to some other type, I just used a q'n'd fix).

```diff
diff --git a/bundles/config/org.eclipse.smarthome.config.xml/src/main/java/org/eclipse/smarthome/config/xml/osgi/XmlDocumentBundleTracker.java b/bundles/config/org.eclipse.smarthome.config.xml/src/main/java/org/eclipse/smarthome/config/xml/osgi/XmlDocumentBundleTracker.java
index c9a97abb6..cc04466d3 100644
--- a/bundles/config/org.eclipse.smarthome.config.xml/src/main/java/org/eclipse/smarthome/config/xml/osgi/XmlDocumentBundleTracker.java
+++ b/bundles/config/org.eclipse.smarthome.config.xml/src/main/java/org/eclipse/smarthome/config/xml/osgi/XmlDocumentBundleTracker.java
@@ -53,7 +53,6 @@ import org.slf4j.LoggerFactory;
  * @param <T>
  *            the result type of the conversion
  */
-@SuppressWarnings({ "unchecked", "rawtypes" })
 public class XmlDocumentBundleTracker<T> extends BundleTracker<Bundle> {
 
     private final Logger logger = LoggerFactory.getLogger(XmlDocumentBundleTracker.class);
@@ -64,10 +63,10 @@ public class XmlDocumentBundleTracker<T> extends BundleTracker<Bundle> {
     private final Map<Bundle, XmlDocumentProvider<T>> bundleDocumentProviderMap = new ConcurrentHashMap<>();
     private final Map<Bundle, ScheduledFuture<?>> queue = new ConcurrentHashMap<>();
     private final Set<Bundle> finishedBundles = new CopyOnWriteArraySet<>();
-    private final Map<String, ServiceRegistration> bundleReadyMarkerRegistrations = new ConcurrentHashMap<>();
+    private final Map<String, ServiceRegistration<?>> bundleReadyMarkerRegistrations = new ConcurrentHashMap<>();
     private final String readyMarkerKey;
 
-    private BundleTracker relevantBundlesTracker;
+    private BundleTracker<?> relevantBundlesTracker;
 
     /**
      * Creates a new instance of this class with the specified parameters.
@@ -125,7 +124,7 @@ public class XmlDocumentBundleTracker<T> extends BundleTracker<Bundle> {
 
     @Override
     public final synchronized void open() {
-        relevantBundlesTracker = new BundleTracker(context,
+        relevantBundlesTracker = new BundleTracker<Object>(context,
                 Bundle.RESOLVED | Bundle.STARTING | Bundle.STOPPING | Bundle.ACTIVE, null) {
             @Override
             public Object addingBundle(Bundle bundle, BundleEvent event) {
@@ -342,21 +341,21 @@ public class XmlDocumentBundleTracker<T> extends BundleTracker<Bundle> {
     private void registerReadyMarker(Bundle bundle) {
         String bsn = bundle.getSymbolicName();
         if (!bundleReadyMarkerRegistrations.containsKey(bsn)) {
-            ServiceRegistration reg = ReadyUtil.markAsReady(context, readyMarkerKey, bsn);
+            ServiceRegistration<?> reg = ReadyUtil.markAsReady(context, readyMarkerKey, bsn);
             bundleReadyMarkerRegistrations.put(bsn, reg);
         }
     }
 
     private void unregisterReadyMarker(Bundle bundle) {
         String bsn = bundle.getSymbolicName();
-        ServiceRegistration reg = bundleReadyMarkerRegistrations.remove(bsn);
+        ServiceRegistration<?> reg = bundleReadyMarkerRegistrations.remove(bsn);
         if (reg != null) {
             reg.unregister();
         }
     }
 
     private void unregisterReadyMarkers() {
-        for (ServiceRegistration reg : bundleReadyMarkerRegistrations.values()) {
+        for (ServiceRegistration<?> reg : bundleReadyMarkerRegistrations.values()) {
             reg.unregister();
         }
         bundleReadyMarkerRegistrations.clear();
```

sjka: > Using this changes we don't need to supress warnings

Right. That was my initial thought too. 
And then I realized that it won't run on OSGi 4.2 anymore (which doesn't have generics in its signatures) and I end up having some very-not-so-amused colleagues 😉 
But I agree, it sucks!

> Do we really suppress all that warnings for the whole class?

You are right, it's safer to do that on variable level. Will change that.

sjka:
Thanks for the review and valuable feedback - I will fix it of course. 
Frustrating to see how many glitches I have missed to spot in the code which I didn't delete 😉

bundles/config/org.eclipse.smarthome.config.xml/src/main/java/org/eclipse/smarthome/config/xml/AbstractXmlBasedProvider.java
sjka: Because of https://bugs.eclipse.org/bugs/show_bug.cgi?id=424214 😉

sjka:
addresses the comments so far & rebased

bundles/config/org.eclipse.smarthome.config.xml/src/main/java/org/eclipse/smarthome/config/xml/AbstractXmlBasedProvider.java
```diff
@@ -0,0 +1,231 @@
...
+     */
+    public final synchronized void addAll(Bundle bundle, Collection<T_OBJECT> objectList) {
+        if (objectList == null || objectList.isEmpty()) {
+            return;
+        }
+        List<T_OBJECT> objects = acquireObjects(bundle);
+        if (objects == null) {
+            return;
+        }
+        for (T_OBJECT object : objectList) {
```
maggu2810: You are using an CopyOnWriteArrayList, but also for any other collection, I think it would be "better" to use `objects.addAll(objectList)` (so the collection implementation can perform an optimized algorithm to add multiple objects) and only removeCachedEntries on the for-each-loop.

bundles/core/org.eclipse.smarthome.core.thing.xml/src/main/java/org/eclipse/smarthome/core/thing/xml/internal/XmlThingTypeProvider.java
```diff
@@ -44,239 +41,44 @@
...
-            }
-            return true;
-        }
-
-        private XmlThingTypeProvider getOuterType() {
-            return XmlThingTypeProvider.this;
-        }
-
-    }
+public class XmlThingTypeProvider extends AbstractXmlBasedProvider<ThingTypeUID, ThingType>
```
maggu2810: There is now
* ConfigXmlConfigDescriptionProvider | esh.scope=core.xml.config
* XmlChannelGroupTypeProvider | esh.scope=core.xml.channelGroups
* XmlChannelTypeProvider | esh.scope=core.xml.channels
* XmlThingTypeProvider | esh.scope=core.xml

To be consistent I would expect the esh.scope should be equal to core.xml.things for the XmlThingTypeProvider (to fir to channel type and channel group type provider).

Why do we hide the "type" at all?

bundles/core/org.eclipse.smarthome.core.thing.xml/src/main/java/org/eclipse/smarthome/core/thing/xml/internal/XmlThingTypeProvider.java
```diff
@@ -44,239 +41,44 @@
...
-                thingTypeReader, thingTypeProviderFactory);
+        thingTypeTracker = new XmlDocumentBundleTracker<List<?>>(componentContext.getBundleContext(), XML_DIRECTORY,
+                thingTypeReader, this, ReadyMarker.XML_THING_TYPE);
         thingTypeTracker.open();
-
     }
 
     @Deactivate
-    protected void deactivate(ComponentContext context) {
+    protected void deactivate(ComponentContext componentContext) {
```
maggu2810: I assume we can remove the unused function argument at all.

bundles/core/org.eclipse.smarthome.core.thing.xml/src/main/java/org/eclipse/smarthome/core/thing/xml/internal/XmlThingTypeProvider.java
```diff
@@ -44,239 +41,44 @@
...
     private AbstractXmlConfigDescriptionProvider configDescriptionProvider;
-
     private XmlChannelTypeProvider channelTypeProvider;
+    private XmlChannelGroupTypeProvider channelGroupTypeProvider;
 
     private XmlDocumentBundleTracker<List<?>> thingTypeTracker;
 
     @Activate
-    protected void activate(ComponentContext context) {
+    protected void activate(ComponentContext componentContext) {
```
maggu2810: Why not using `BundleContext` instead of `ComponentContext` if we are interested on the bundle context only?

bundles/core/org.eclipse.smarthome.core.thing.xml/src/main/java/org/eclipse/smarthome/core/thing/xml/internal/XmlThingTypeProvider.java
```diff
@@ -299,13 +101,41 @@ public void unsetConfigDescriptionProvider(ConfigDescriptionProvider configDescr
         this.configDescriptionProvider = null;
     }
 
-    @Reference(target = "(esh.scope=core.xml)")
+    @Reference(target = "(esh.scope=core.xml.channels)")
```
maggu2810: I assume annotated fields does not work with the old DS we need to support. Correct?

bundles/core/org.eclipse.smarthome.core.thing/src/main/java/org/eclipse/smarthome/core/thing/internal/ThingManager.java
```diff
@@ -997,25 +987,60 @@ protected void activate(ComponentContext componentContext) {
...
+                        registerAndInitializeHandler(thing, thingHandlerFactory);
+                    } else {
+                        logger.debug("Thing handler for thing '{}' already registered", thing.getUID());
+                    }
+                }
+            });
+        });
+    }
+
+    private String getBundleName(ThingHandlerFactory it) {
```
maggu2810: Why does the argument use the name "it"? It doesn't matter, but IMHO uncommon to use "iterator" (??) for an variable name of an argument.

bundles/core/org.eclipse.smarthome.core/src/main/java/org/eclipse/smarthome/core/internal/i18n/ResourceBundleTracker.java
```diff
@@ -30,16 +31,15 @@
  * @author Markus Rathgeb - Add locale provider support
  * @author Ana Dimova - fragments support
  */
-@SuppressWarnings("deprecation")
-public class ResourceBundleTracker extends ResolvedBundleTracker {
+@SuppressWarnings({ "deprecation", "rawtypes", "unchecked" })
```
maggu2810: IIRC we already discussed the "omit supress warnings for the whole class if possible" already.

Isn't it possible here?

bundles/core/org.eclipse.smarthome.core/src/main/java/org/eclipse/smarthome/core/service/ReadyUtil.java
```diff
@@ -0,0 +1,44 @@
...
+import org.osgi.framework.ServiceRegistration;
+
+/**
+ * Utility class to easily register {@link ReadyMarker}s programmatically.
+ *
+ * @author Simon Kaufmann - initial contribution and API
+ *
+ */
+@SuppressWarnings({ "rawtypes", "unchecked" })
+public class ReadyUtil {
```
maggu2810: IIRC we already discussed the "omit supress warnings for the whole class if possible" already.

Isn't it possible here?

maggu2810:
@sjka You don't wait for me, do you?

kaikreuzer:
And for me - I planned to review and test today as well.

maggu2810:
> And for me - I planned to review and test today as well.

What I want to say has been:
@sjka As it seems this PR has been stalled, I want to ensure you do not wait for any further comments from my side. My review has been finished 22 days ago.

sjka:
No, I was just waiting for @kaikreuzer's feedback

bundles/config/org.eclipse.smarthome.config.xml/src/main/java/org/eclipse/smarthome/config/xml/AbstractXmlBasedProvider.java
```diff
@@ -0,0 +1,231 @@
...
+import java.util.concurrent.CopyOnWriteArrayList;
+
+import org.osgi.framework.Bundle;
+
+/**
+ * Common base class for XML based providers.
+ *
+ * @author Simon Kaufmann - initial contribution, factored out of subclasses
+ *
+ * @param <T_ID> the key type, e.g. ThingTypeUID, ChannelUID, URI,...
```
kaikreuzer: you might want to refactor this to use `Identifiable` now

bundles/config/org.eclipse.smarthome.config.xml/src/main/java/org/eclipse/smarthome/config/xml/AbstractXmlBasedProvider.java
```diff
@@ -0,0 +1,231 @@
...
+
+    /**
+     * Gets the object with the given key.
+     *
+     * @param key the key which identifies the object
+     * @param locale the locale
+     * @return the object if found, <code>null</code> otherwise
+     */
+    protected final T_OBJECT get(T_ID key, Locale locale) {
+        for (Entry<Bundle, List<T_OBJECT>> objects : bundleObjectMap.entrySet()) {
```
kaikreuzer: this might actually be far more elegant using lambdas - that might also be a consideration for some of the other methods here.

bundles/config/org.eclipse.smarthome.config.xml/src/main/java/org/eclipse/smarthome/config/xml/AbstractXmlBasedProvider.java
```diff
@@ -0,0 +1,231 @@
...
+    private final Map<Bundle, List<T_OBJECT>> bundleObjectMap = new ConcurrentHashMap<>();
+    private final Map<LocalizedKey, T_OBJECT> localizedObjectCache = new ConcurrentHashMap<>();
+
+    /**
+     * Return the unique identifier of the given object (e.g. UID, URI,...)
+     *
+     * @param object the object to return the identifier for
+     * @return the identifier. Must not be <code>null</code>
+     */
+    protected abstract T_ID getIndentifier(T_OBJECT object);
```
kaikreuzer: this will probably be replaced by getUID from `Identifiable`, but if not, you might want to fix the typo (Indent vs. Ident).

bundles/config/org.eclipse.smarthome.config.xml/src/main/java/org/eclipse/smarthome/config/xml/osgi/XmlDocumentBundleTracker.java
```diff
@@ -69,174 +83,297 @@
...
+        unregisterReadyMarker(bundle);
+    }
+
+    /**
+     * This method creates a list where all resources are contained
+     * except the ones from the host bundle which are also contained in
+     * a fragment. So the fragment bundle resources can override the
+     * host bundles resources.
+     *
+     * @param xmlDocumentPaths
```
kaikreuzer: did you miss adding some content here?

bundles/config/org.eclipse.smarthome.config.xml/src/main/java/org/eclipse/smarthome/config/xml/osgi/XmlDocumentBundleTracker.java
```diff
@@ -69,174 +83,297 @@
...
+    private void finishBundle(Bundle bundle) {
+        queue.remove(bundle);
+        finishedBundles.add(bundle);
+        registerReadyMarker(bundle);
+
+        Set<Bundle> remainingBundles = getRemainingBundles();
+        if (!remainingBundles.isEmpty()) {
+            logger.trace("Remaining bundles with {}: {}", xmlDirectory, remainingBundles);
+        } else {
+            logger.debug("Finished loading bundles with {}", xmlDirectory);
```
kaikreuzer: I seem to get one of these messages for every bundle in my installation...
Not really sure if we are at all expected to hit this line so often (if so, better reduce to trace) or if there is some issue in the processing and we should not run out of remainingBundles at startup continuously.

bundles/core/org.eclipse.smarthome.core.thing.xml/src/main/java/org/eclipse/smarthome/core/thing/xml/internal/XmlThingTypeProvider.java
kaikreuzer: This should simply create a "target" property in the generated DS XML and these properties are already supported in DS1.1/OSGi4.2, see specification chapter 112.3.4.

bundles/core/org.eclipse.smarthome.core.thing/src/main/java/org/eclipse/smarthome/core/thing/internal/ThingManager.java
```diff
@@ -997,25 +987,60 @@ protected void activate(ComponentContext componentContext) {
...
     private void registerAndInitializeHandler(final Thing thing, final ThingHandlerFactory thingHandlerFactory) {
         if (thingHandlerFactory != null) {
-            initializationVetoManager.applyActionFor(thingHandlerFactory.getClass(), thing);
+            String bsn = getBundleName(thingHandlerFactory);
+            if (loadedXmlThingTypes.contains(bsn)) {
+                registerHandler(thing, thingHandlerFactory);
+                initializeHandler(thing);
+            } else {
+                logger.debug(
+                        "Not registering a handler at this point. The thing types of bundle {} are not fully loaded yet.",
```
kaikreuzer: was that the auto formatter or you yourself?

bundles/core/org.eclipse.smarthome.core/src/main/java/org/eclipse/smarthome/core/service/ReadyMarker.java
```diff
@@ -0,0 +1,22 @@
...
+
+/**
+ * Marker service in order to denote that loading of a specific component has completed and it is ready to be used.
+ *
+ * @author Simon Kaufmann - initial contribution and API
+ *
+ */
+public interface ReadyMarker {
+
+    public final static String XML_THING_TYPE = "esh.xmlThingTypes";
```
kaikreuzer: I am not completely happy to define those constants in smarthome.core as this bundle in theory does not know anything about the existence of the xml bundles.

bundles/core/org.eclipse.smarthome.core/src/main/java/org/eclipse/smarthome/core/internal/i18n/ResourceBundleTracker.java
sjka: not for "rawtypes", as it complains about

```java
public class ResourceBundleTracker extends BundleTracker
```
so whatever you like better: living with the warning or with the too generous suppression, I'm fine with both

maggu2810: The only "solution" I found has been https://github.com/maggu2810/smarthome/commit/b7785d751d10e9f0fb1444a3ebfa194f206d074b which combines the suppress warnings to the sections that really need it.
But if it is necessary to create separate classes... I assume you will tell me, that it isn't :wink:

bundles/core/org.eclipse.smarthome.core.thing/src/main/java/org/eclipse/smarthome/core/thing/internal/ThingManager.java
sjka: it was him (of course)

bundles/config/org.eclipse.smarthome.config.xml/src/main/java/org/eclipse/smarthome/config/xml/AbstractXmlBasedProvider.java
sjka: done

bundles/config/org.eclipse.smarthome.config.xml/src/main/java/org/eclipse/smarthome/config/xml/AbstractXmlBasedProvider.java
sjka: I couldn't get it to look more elegant, so I rather left it as it use. At least it is readable.

kaikreuzer:
Thanks for updates, looks all good to me now!

@maggu2810 Feel free to check if your comments have been addressed as expected as well - if so, please merge :-)

bundles/core/org.eclipse.smarthome.core/src/main/java/org/eclipse/smarthome/core/internal/i18n/ResourceBundleTracker.java
sjka: Yes, it really feels wrong to me to create a separate class just for this specific reason 😄

lolodomo:
@sjka  : could you explain what issues your PR will solve ?
Does it solve all issues relative to rules at startup ?

sjka:
No, it doesn't. I is just a major cleanup within the core XML processing. 
Nevertheless, it is a prerequisite for being able to solve the rules startup issues at all...

bundles/config/org.eclipse.smarthome.config.core/src/main/java/org/eclipse/smarthome/config/core/ConfigDescription.java
```diff
@@ -113,7 +115,8 @@ public ConfigDescription(URI uri, List<ConfigDescriptionParameter> parameters,
      *
      * @return the URI of this description (not null)
      */
-    public URI getURI() {
+    @Override
+    public URI getUID() {
```
cweitkamp: @sjka I just found this one here, which leads to some broken bindings in OH2 repo (see [Travis failure here](https://travis-ci.org/openhab/openhab2-addons/builds/254991394?utm_source=github_status&utm_medium=notification)). Seems to be an APIBraking, isn't it?

kaikreuzer: I can confirm that this change breaks the openHAB build: https://openhab.ci.cloudbees.com/job/openHAB2-Bundles/99/console

kaikreuzer: Should hopefully be fixed with https://github.com/openhab/openhab2-addons/pull/2476.

sjka: Thanks!
By the way, the "root cause" was the introduction of the identifiable interface in #3762.

cweitkamp: :+1:

PR #3424, opened 2017-05-11
Made linked items list responsive

aounhaider1:
Label of linked items overlaps with the icon on small screens.

Before:
screen shot 2017-05-11 at 09 25 53

After:
screen shot 2017-05-11 at 13 40 32

### DIFF
--- a/extensions/ui/org.eclipse.smarthome.ui.paper/web-src/partials/configuration.html
+++ b/extensions/ui/org.eclipse.smarthome.ui.paper/web-src/partials/configuration.html
@@ -171,10 +171,10 @@
 										</button>
 									</h3>
 									<div ng-repeat="item in channel.items | orderBy" class="row cursor-pointer " ng-class="{'borderhidden':$last}" ng-click="navigateTo('item/edit/' + item.name)">
-										<div class="col-xs-1">
+										<div class="col-lg-1 col-md-2 col-xs-3">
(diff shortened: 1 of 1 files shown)


sjka:
Right, it does not overlap anymore. 

However, when shrinking the window, the space between the icon and the text first grows and then shrinks again:

bildschirmfoto 2017-05-11 um 14 23 13
bildschirmfoto 2017-05-11 um 14 23 30
bildschirmfoto 2017-05-11 um 14 23 43

Is that by intention?

aounhaider1:
Its rather by limitation.

htreu:
@aounhaider1 please have a look at my take on this:

screen shot 2017-05-12 at 11 53 13

It sets `padding-left: 0` on the parent `div` which will move the bullet to the left most position. Then you only have to define `col-xs-2` for small screens and `col-lg-1` for the rest. This way it will not bounce back and forth.

wdyt?

aounhaider1:
Looks better because of less grid classes. But label is too close to icon on small devices. I will go with 'xs-3' and 'sm-2'.
screen shot 2017-05-12 at 12 25 35

Moreover, the idea is to make sure that UI is responsive and consistent on devices with different screen sizes. In MOST cases, the screen size of a device does not change and 'bouncing' should not be a problem IMHO.

PR #3425, opened 2017-05-11
[Housekeeping] cleaning up some System.???.println

sjka:
...that I found throughout the code base.

Some were left in test cases from previous bug-huntings hence I simply deleted them. Others slipped through, so I turned them into log statements.


maggu2810:
Thank you

PR #3426, opened 2017-05-11
Console command things lists things again if no argument is passed

triller-telekom:
Be consistent with items and links command again


triller-telekom:
Obsolete, tests will be fixed

PR #3429, opened 2017-05-11
Only restore backup if there indeed was an unreadable file.

### LINKED ISSUE #3428: Noisy logging of JsonStorage on a fresh installation
A fresh installation welcomes the user with this logging:

```
15:12:10.463 [INFO ] [e.smarthome.storage.json.JsonStorage] - Json storage file at '/Users/kai/Downloads/oh/userdata/jsondb/org.eclipse.smarthome.core.items.Item.json' was not read.
15:12:10.600 [INFO ] [e.smarthome.storage.json.JsonStorage] - Json storage file at '/Users/kai/Downloads/oh/userdata/jsondb/org.eclipse.smarthome.core.thing.link.ItemChannelLink.json' was not read.
15:12:10.604 [INFO ] [e.smarthome.storage.json.JsonStorage] - Json storage file at '/Users/kai/Downloads/oh/userdata/jsondb/org.eclipse.smarthome.core.thing.link.ItemThingLink.json' was not read.
15:12:10.606 [INFO ] [e.smarthome.storage.json.JsonStorage] - Json storage file at '/Users/kai/Downloads/oh/userdata/jsondb/org.eclipse.smarthome.core.thing.Thing.json' was not read.
15:12:11.081 [INFO ] [e.smarthome.storage.json.JsonStorage] - Json storage file at '/Users/kai/Downloads/oh/userdata/jsondb/org.eclipse.smarthome.config.discovery.DiscoveryResult.json' was not read.
```
It is perfectly valid that there aren't yet any database files, so I consider this a bug.

kaikreuzer:
Fixes #3428.

Imho backup files must only be used to restore a database, if there is indeed a corrupt file. If no database file exists, no backup should be used.

### DIFF
--- a/bundles/storage/org.eclipse.smarthome.storage.json/src/main/java/org/eclipse/smarthome/storage/json/JsonStorage.java
+++ b/bundles/storage/org.eclipse.smarthome.storage.json/src/main/java/org/eclipse/smarthome/storage/json/JsonStorage.java
@@ -87,7 +87,13 @@ public class JsonStorage<T> implements Storage<T> {
 
         // If there was an error reading the file, then try one of the backup files
         if (inputMap == null) {
-            logger.info("Json storage file at '{}' was not read.", file.getAbsolutePath());
+            if (file.exists()) {
+                logger.info("Json storage file at '{}' seems to be corrupt - checking for a backup.",
+                        file.getAbsolutePath());
+            } else {
+                logger.debug("Json storage file at '{}' does not exist - checking for a backup.",
+                        file.getAbsolutePath());
+            }
             for (int cnt = 1; cnt <= maxBackupFiles; cnt++) {
                 File backupFile = getBackupFile(cnt);
                 if (backupFile == null) {
--- a/bundles/storage/org.eclipse.smarthome.storage.json/src/main/java/org/eclipse/smarthome/storage/json/JsonStorage.java
+++ b/bundles/storage/org.eclipse.smarthome.storage.json/src/main/java/org/eclipse/smarthome/storage/json/JsonStorage.java
@@ -274,7 +274,6 @@ public class JsonStorage<T> implements Storage<T> {
 
             try (FileOutputStream outputStream = new FileOutputStream(file, false);) {
                 outputStream.write(s.getBytes());
-                outputStream.close();
             } catch (Exception e) {
                 logger.error("Error writing JsonDB to {}. Cause {}.", file.getPath(), e.getMessage());
             }


maggu2810:
WRT to the file system handling that could result in missing files I think if there is a file missing but an update exists it is okay to use the backup.
If there is no database and no backup, I assume we don't need a log message at all.

philomatic:
@maggu2810  I agree.

Kai, how to handle this?

kaikreuzer:
> WRT to the file system handling that could result in missing files

Do you have an example how this could happen unintentionally?
I would assume that we only have to deal with corrupted files, i.e. writing is interrupted by a power outage.
If the file is missing, I would assume that it has been intentionally deleted by the user - and in this case, I think it is very unexpected if some old backup is automatically restored.

philomatic:
See my comment in #3428.

And yes, you're exactly describing the scenario I have:

File is corrupted by writing during power outage,
filesystem deletes corrupted files (ext4 on a Raspberry Pi),
expected behavior should be to restore old backup (if existent).

This is imho a better handling than loosing all data of created things etc.

kaikreuzer:
> filesystem deletes corrupted files

Really? That sounds pretty weird to me... Just out of interest, do you have any reference to that behavior (assuming it is a feature of ext4 and not a bug)?

kaikreuzer:
I have updated the PR.

maggu2810:
```
        synchronized (map) {
            // Rename the file for backup
            File rename = new File(file.getParent() + File.separator + BACKUP_EXTENSION,
                    System.currentTimeMillis() + SEPARATOR + file.getName());
            file.renameTo(rename);

            try (FileOutputStream outputStream = new FileOutputStream(file, false);) {
                outputStream.write(s.getBytes());
                outputStream.close();
            } catch (Exception e) {
                logger.error("Error writing JsonDB to {}. Cause {}.", file.getPath(), e.getMessage());
            }

            deferredSince = 0;
        }
```

The current database file is renamed, after that a new file is created.
A power loss between that instructions will lead to a missing file.
Do you agree?

Perhaps it would be better to copy the file, truncate the existing file to a size of zero and then fill it with new content. So we could ensure that the directory entry is not removed and an existing file (entry in the parent / directory) will never be removed again (only changed metadata like the size).

maggu2810:
Reading the above code I am a little bit irritated why there is a close instruction at the end of the try-with-resources block... @cdjackson What's the reason for the explicit close call? Isn't that exactly the reason to use try-with-resources to do it implicit on leaving the scope?

cdjackson:
Yes - you are right that this close is not needed. It wasn't added to irritate you ;)

kaikreuzer:
I have removed the "close".

If @philomatic's claim that ext4 might delete corrupted files is correct, I would like to leave the PR as it is now. If someone else wants to also improve the file writing mechanism, feel free to come up with another PR.

philomatic:
Well, it doesn't delete them in the strict sense. A power failure during write actions in ext3/ext4 can (!) lead to a corrupted/unlinked inode. Which means the contents are still there, but the FS doesn't know anymore to which location on the disk it belongs.
That is, in my understanding as I'm not a Linux expert, how the lost+found folder from fsck works, see also:

https://en.wikipedia.org/wiki/Fsck
https://en.wikipedia.org/wiki/Lost_and_found#Computing
http://www.tldp.org/LDP/Linux-Filesystem-Hierarchy/html/lostfound.html

I know that behavior from the angstrom distribution in combination with broken flash memory on the BeagleBone Black. I've had there some similar problems with bundleFiles in karaf, which could partially recovered by fsck (which I've tried just out of curiosity as it doesn't fix the problem ofc).

This time, I've tested the behavior by forcing continuous database commits during a power failure, which resulted in missing files.
However, I didn't run fsck though and have to admit:
@maggu2810 's explanation sounds much more reasonable to me.
I've never checked the commit code as I didn't think this could be the problem, thanks @maggu2810 !

tl;dr
Yes, that can happen on Linux with ext3/ext4, that's why we have fsck and lost+found.
But it's more likely that the commitDatabase() method is the problem.
So, I'm fine with this PR now :)

sjka:
I still think that we should fix the dabase's behavior as @maggu2810 described [here](https://github.com/eclipse/smarthome/pull/3429#issuecomment-300801580). 

For the meantime it looks to me like this PR reached a consensus.

sjka:
I created #3433 to track the follow-up.

PR #3434, opened 2017-05-12
BasicUI: Show value on slider if format is set

### LINKED ISSUE #3369: Basic UI: display value for dimmer items
https://community.openhab.org/t/basicui-and-slider-item-no-values-visable/11406

The wish is to optionally display the value for a dimmer item, when a format is set in the item label.

Classic UI is doing that, not Basic UI.

triller-telekom:
Fixes #3369


lolodomo:
@triller-telekom : you are handling the label in JS while the value (from label) is already provided in the "value" parameter of the function setValuePrivate. You defined %labelValue% while %value% already exists an is set by the expected value. In my opinion, you made all this more complex than it should be.
I had the intention to provide a fix later today for this with as a bonus the handling of labelcolor and valuecolor. Can I propose my own version of this enhancement in replacement of yours ? Or can you please simplify your PR looking at how it is handled for example with the switch ?

lolodomo:
We have to use %value%, %has_value%, %labelStyle% and %valueStyle% and so add them in the slider snippet.
There is nothing to change in the renderer (Java part).
In the JavaScript, there is no need for an additional parameter in the function setValuePrivate but just use the parameter "value" to set the new object (only when "ahs_value" is true.

lolodomo:
And it will make sense to update the CSS for the new value object.

triller-telekom:
@lolodomo : Thank you for your comments. I am pretty new to ESH and also the GUI/Javascript stuff was a good learning exercise for me.

I suggest to proceed in whatever way is more convenient for you since you are more experienced in this area. If you want to propose an easier implementation and already know how to do it then feel free to do so. Then I will delete my PR in favor to yours.

lolodomo:
Alternative PR: #3440

lolodomo:
@triller-telekom : my alternative PR was approved, please close your PR.

PR #3436, opened 2017-05-12
Remove SatisfiableRESTResource interface in favor for Java 8 default …

triller-telekom:
…interface methods

Part of Java 8 cleanup: https://github.com/eclipse/smarthome/pull/2320#issuecomment-253907800


bundles/io/org.eclipse.smarthome.io.rest/src/main/java/org/eclipse/smarthome/io/rest/RESTResource.java
```diff
@@ -11,7 +11,17 @@
...
+ * @author Stefan Triller - Added default implementation for isSatisfied
  */
 public interface RESTResource {
 
+    /**
+     * Method used to determine availability of a RESTResource
+     *
+     * @return true if this RESTResource is ready to process requests, false otherwise.
+     */
+    default public boolean isSatisfied() {
```
maggu2810:
```
It is permitted, but discouraged as a matter of style, to redundantly specify the public and/or abstract modifier for a method declared in an interface.
```
http://docs.oracle.com/javase/specs/jls/se7/html/jls-9.html#jls-9.4

kaikreuzer:
Note to implementors: All code using `SatisfiableRESTResource` needs to be refactored.

BenediktNiehues:
@triller-telekom seems that you missed to adapt/remove the component descriptor satisfiableresourcefilter.xml in o.e.s.io.rest

triller-telekom:
Thx for pointing this out, I created a new pr to remove it: #3452

Issue #1700, opened 2016-06-19
Loop(?) in PaperUI "Control" page


MHerbst:
I have used the PaperUI to addseveral things to my OpenHAB 2 configuration (Homematic, Astro, Yamaha Receiver). At some time the "Control" page stopped working correctly. In Chrome I could see that the refresh function was activated permanently before it "crashed" with a Chrome error. 
On Firefox I got this error message after some seconds:
![paperuicontrol](https://cloud.githubusercontent.com/assets/616792/16178074/38dae810-363f-11e6-98ba-c54127332958.PNG)
Perhabs the rendering of the page took too long and the refresh was triggered before it was rendered completely.

maggu2810:
The rendering of the control page takes some time. I don't know if it has been faster before but also for only a few things it needs (IMHO) to much time.

kaikreuzer:
@aounhaider1 Could you please have a look?

kaikreuzer:
>  I don't know if it has been faster before but also for only a few things it needs (IMHO) to much time.

We just tested it against an older version - and no, it wasn't any faster in earlier versions.

The problem is that a rather complex HTML is constructed dynamically for this page - there isn't really a quick fix for this. The best workaround is to not have too many things on a page) - with https://github.com/eclipse/smarthome/issues/1083, we will have the Things on multiple tabs, which will ease the situation.

maggu2810:
For me it has been the combination of the (unused but active) Z-Wave binding in the runtime and the circumstance that the Paper UI loads all thing types "/rest/thing-types".
The response content length is 1,579,263 Bytes

markusjw:
I just wanted to raise the same issue.
The effect is reproducible when you have a Number item (usually defined via a thing channel) with a high maximum value, combined with a small step value.
An example would be an energy counter item which can reach high numbers, so that such item definitions actually exist.
If there are several of these items, the page script can take minutes to load. It seems to load eventually, so it might be rather a very long than an infinite loop.

I suppose that the slider control is not designed to create a high number of discrete value stops.

markusjw:
@aounhaider1 Does this information help to find out more about the problem?
I hope there is a solution because right now the Control UI is just not usable for things with wide range Number items.

markusjw:
Besides the fact that the web UI should not block, there is a general question:
Should a slider be shown for values with large number ranges? In these cases, the slider is no longer usable for precise inputs.
@kaikreuzer What do you think?

[...]

PR #3437, opened 2017-05-12
Removes slider from number item on control page

aounhaider1:
as discussed here https://github.com/eclipse/smarthome/issues/1700#issuecomment-289928566

### DIFF
--- a/extensions/ui/org.eclipse.smarthome.ui.paper/web-src/js/controllers.control.js
+++ b/extensions/ui/org.eclipse.smarthome.ui.paper/web-src/js/controllers.control.js
@@ -244,7 +244,7 @@ angular.module('PaperUI.controllers.control', []).controller('ControlPageControl
         if (!item.stateDescription || isNaN(item.stateDescription.minimum)) {
             return '';
         } else if (!item.stateDescription.pattern) {
-            return item.stateDescription.minimum;
+            return ''+item.stateDescription.minimum;
         } else {
             return sprintf(item.stateDescription.pattern, item.stateDescription.minimum);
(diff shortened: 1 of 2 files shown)


htreu:
Looks good, works as described. @aounhaider1 do you mind adding `min` and `max` attributes to the input field? This way at least the stepper buttons will not allow input of values outside the given range.

aounhaider1:
Updated

htreu:
Its awesome to have the validation build right in with the `ng-min` and `ng-max` attributes. Nice! But the steppers will allow the input of values beyond the valid range. With additional `min` & `max` attributes this can be avoided.
Another thing: When angular reports the value as invalid its still possible to "commit" the value which will result in a `400 - Bad Request` response. Is it possible to deactivate the commit button on invalid data? See here:

screen shot 2017-05-15 at 13 56 29

The valid range is [-80;12]. The value 100 is clearly reported as being invalid, see attribute `aria-invalid="true"`

aounhaider1:
I've added min, max validation to number and rollershutter widget. But I think we should have decent editing option on control page as well (with proper error message and so on). However, we have a limited space for items so maybe we can introduce an expandable area and show the input field in that area. Wdyt?

htreu:
Imho the current editing function is fine. I´m just talking about the submit button when ng-min or gn-max detected an invalid value:

image

can't we disable the button then?

aounhaider1:
It should be working with the last commit. :)

htreu:
ready to get merged.

sjka:
Thanks!

PR #3438, opened 2017-05-12
Remove Convertible interface

triller-telekom:
In Java 8 we can now use a default implementation of as() on the interface State

Part of the Java 8 cleanup tasks: https://github.com/eclipse/smarthome/wiki/Java-8-CP2-cleanup


bundles/core/org.eclipse.smarthome.core/src/main/java/org/eclipse/smarthome/core/types/State.java
```diff
@@ -14,4 +14,19 @@
...
 public interface State extends Type {
+
+    /**
+     * Convert this {@link State}'s value into another type
+     *
+     * @param target the desired {@link State} type
+     * @return the {@link State}'s value in the given type's representation, or <code>null</code> if the conversion was
+     *         not possible
+     */
+    default public State as(Class<? extends State> target) {
```
maggu2810:
```
It is permitted, but discouraged as a matter of style, to redundantly specify the public and/or abstract modifier for a method declared in an interface.
```
http://docs.oracle.com/javase/specs/jls/se7/html/jls-9.html#jls-9.4

bundles/core/org.eclipse.smarthome.core/src/main/java/org/eclipse/smarthome/core/library/types/PercentType.java
```diff
@@ -81,7 +80,12 @@ public State as(Class<? extends State> target) {
         } else if (target == HSBType.class) {
             return new HSBType(DecimalType.ZERO, PercentType.ZERO, this);
         } else {
-            return StateConverterUtil.defaultConversion(this, target);
+            // if the target is the same type return, otherwise give up
```
sjka: dito

bundles/core/org.eclipse.smarthome.core/src/main/java/org/eclipse/smarthome/core/library/types/HSBType.java
```diff
@@ -269,7 +267,12 @@ public State as(Class<? extends State> target) {
         } else if (target == PercentType.class) {
             return new PercentType(getBrightness().toBigDecimal());
         } else {
-            return StateConverterUtil.defaultConversion(this, target);
+            // if the target is the same type return, otherwise give up
```
sjka: could you please define a `protected` method in `NumberType` which there delegates to `State.super.as(target)`? Then you don't need to repeat his logic twice.

triller-telekom: In `DecimalType` you mean? Because DecimalType is the super class of `PercentType`, which is the super class of `HSBType`. So the protected method in `DecimalType` can delegate it to state?

bundles/core/org.eclipse.smarthome.core/src/main/java/org/eclipse/smarthome/core/library/types/HSBType.java
```diff
@@ -269,7 +267,8 @@ public State as(Class<? extends State> target) {
         } else if (target == PercentType.class) {
             return new PercentType(getBrightness().toBigDecimal());
         } else {
-            return StateConverterUtil.defaultConversion(this, target);
+            // // if the target is the same type return, otherwise give up
```
sjka: what is this? A super-comment? 😉 
Just drop it completely...

bundles/core/org.eclipse.smarthome.core/src/main/java/org/eclipse/smarthome/core/library/types/DecimalType.java
```diff
@@ -143,6 +141,10 @@ public long longValue() {
         return value.longValue();
     }
 
+    protected State delegateStateAs(Class<? extends State> target) {
```
sjka: although technically correct, "delegate" somehow tends to mislead the kind read to expect the [delegate pattern](https://en.wikipedia.org/wiki/Delegation_pattern) here. How about "defaultConversion" or something similar?

bundles/core/org.eclipse.smarthome.core/src/main/java/org/eclipse/smarthome/core/library/types/DecimalType.java
```diff
@@ -169,7 +171,7 @@ public State as(Class<? extends State> target) {
             return new HSBType(DecimalType.ZERO, PercentType.ZERO,
                     new PercentType(this.toBigDecimal().multiply(BigDecimal.valueOf(100))));
         } else {
-            return StateConverterUtil.defaultConversion(this, target);
+            return State.super.as(target);
```
sjka: How about using your brand-new method here to?

bundles/core/org.eclipse.smarthome.core/src/main/java/org/eclipse/smarthome/core/library/types/DecimalType.java
sjka: PS: I know, I started with this term. Mea culpa!

triller-telekom: ;-)

PR #3439, opened 2017-05-12
Write jsondb backup without renaming original file.

### LINKED ISSUE #3433: jsonDB: risk of missing files
When writing to the jsonDB, it first moves the current file out of the way and then creates a new one at the same place. This can lead to missing files, when crashing the system at the right point in time. See https://github.com/eclipse/smarthome/pull/3429#issuecomment-300801580.

cdjackson:
This changes the way the backup is written. The backup is now written at the same time as the data is stored, rather than storing a single file, and then having to rename it later. The main database file is overwritten so should not ever be removed unless the underlying filesystem is doing this (in which case there is probably nothing that can be done at Java level?).

This avoids the situation where there is no file for any period of time, and this should ensure the backup is read if the main file is corrupted due to a crash (etc).

Writing the backup at the same time as the main file avoids a (slower) read->write which would be required to copy the original file if it was done later.

This has another (arguable) advantage in that there are now two copies of the latest file, so two chances to load with the latest data.

Fixes #3433

### DIFF
--- a/bundles/storage/org.eclipse.smarthome.storage.json/src/main/java/org/eclipse/smarthome/storage/json/JsonStorage.java
+++ b/bundles/storage/org.eclipse.smarthome.storage.json/src/main/java/org/eclipse/smarthome/storage/json/JsonStorage.java
@@ -260,6 +260,14 @@ public class JsonStorage<T> implements Storage<T> {
                 fileTimes.get(fileTimes.size() - age) + SEPARATOR + file.getName());
     }
 
+    private void writeDatabaseFile(File dataFile, String data) {
+        try (FileOutputStream outputStream = new FileOutputStream(dataFile, false)) {
+            outputStream.write(data.getBytes());
+        } catch (Exception e) {
+            logger.error("Error writing JsonDB to {}. Cause {}.", dataFile.getPath(), e.getMessage());
+        }
+    }
+
     /**
      * Write out any outstanding data.
      * <p>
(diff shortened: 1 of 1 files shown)


philomatic:
Looks nice to me :)

One thing worth to mention:

If writing of database fails (maybe due to an unwanted fs lock or s.th. else), we still would try to write a backup file?
This could lead to a state where the backup is newer than the current database (file).
This seems odd to me, however not a real problem (and should be a rare case).

cdjackson:
Why does it matter if the backup is newer than the master. In the case that the master is not written for any reason you are already in a position where the database will be out of sync. If in addition there's a restart and the master is found to be corrupt, isn't it better to read the latest data rather than data you know is wrong?

philomatic:
Agreed, it actually doesn't matter why the db couldn't be (over)written (data corruption or fs problems). 
It will be out of sync one way or another.

If there's a restart and master will be loaded (assume not corrupted), a "newer" backup won't harm anything.

bundles/storage/org.eclipse.smarthome.storage.json/src/main/java/org/eclipse/smarthome/storage/json/JsonStorage.java
```diff
@@ -261,23 +261,33 @@ private File getBackupFile(int age) {
...
             try (FileOutputStream outputStream = new FileOutputStream(file, false);) {
                 outputStream.write(s.getBytes());
             } catch (Exception e) {
                 logger.error("Error writing JsonDB to {}. Cause {}.", file.getPath(), e.getMessage());
             }
 
+            // And also write the backup
+            File backup = new File(file.getParent() + File.separator + BACKUP_EXTENSION,
+                    System.currentTimeMillis() + SEPARATOR + file.getName());
+            try (FileOutputStream outputStream = new FileOutputStream(backup, false);) {
```
sjka: As these lines for actually writing out the file are identical for `file` and `backup` (apart of this parameter, of course), could you please factor them out?

cdjackson: Done.

PR #3440, opened 2017-05-12
Basic UI: slider with value

### LINKED ISSUE #3369: Basic UI: display value for dimmer items

lolodomo:
Fix #3369
Add support for label and value color


lolodomo:
Testing my change, I discovered that sometimes the value is not updated or even wrong. It does not happen when dragging the slider but only when clicking on the slider. It is not systematic. I really don't think that the problem is due to my change.There are already existing issues relative to slider so I imagine I just encountered one of them.

lolodomo:
@resetnow : can you review please ?

extensions/ui/org.eclipse.smarthome.ui.basic/web-src/smarthome.less
```diff
@@ -371,6 +371,9 @@
 		&--switch {
```
resetnow: Please combine the selectors:
```less
&--switch, &--slider {
...
```

extensions/ui/org.eclipse.smarthome.ui.basic/web-src/smarthome.less
```diff
@@ -371,6 +371,9 @@
...
 			padding: 0 10px 0 10px;
 		}
+		&--slider {
+			padding: 0 10px 0 10px;
+		}
```
resetnow: This can be removed.

lolodomo:
Less file updated. Thank you again @resetnow for your quick review.
Ready for merge.

Issue #2784, opened 2017-01-09
Basic UI: Displaying broken image when no icon is defined.


chriscolden:
I'm getting a broken image displayed in the browser when no icon is defined. Classic UI doesn't add the img tags when no icon is defined.

kaikreuzer:
@resetnow Could you check?

resetnow:
I thought about that, looks ugly. Would it be better to replace icon to `none` if there is no matching icon like this:

<img src="https://cloud.githubusercontent.com/assets/2057191/21773122/4e1ec364-d6a7-11e6-839d-6d70fb915fe8.png" width="400">

Or remove the icon completely:

<img src="https://cloud.githubusercontent.com/assets/2057191/21773104/3c767986-d6a7-11e6-8d40-884fae60a4c5.png" width="400">

?

kaikreuzer:
> Would it be better to replace icon to none if there is no matching icon like this

Yes, definitely, this is also how the Classic UI renders it.

ThomDietrich:
> Would it be better to replace icon to none if there is no matching icon like this

I would also agree

lolodomo:
@resetnow: what would be the best solution to display the none.png image and how to know if we need to use the none.png file ?
I took a look to the code and I saw that the img src tag for the icon in snippets is currently a pattern in which we replace different parts (category, state, ...). Of course, similar code in the JavaScript code to refresh the icon.
Do we need to change all snippets to have the ability to set fully the value of the img src ?
And should we consider a specific "none" category when the comuted category is "number" ?

resetnow:
@lolodomo there is a PR which illustrates the basic idea for that, but it was way too hacky so it wasn't merged https://github.com/eclipse/smarthome/pull/2799

It was meant to check the icon existence server-side.

resetnow:
Just noticed I deleted the branch. There was a class that watched icon providers via OSGI, and there was a method getIcon, which would return "none" if there is no icon. Basically it was a proxy for all the icon providers.

lolodomo:
Ok, I think that the real subject is the issue #2994 and the need to define a default icon in the iconset for any number item.
Do you agree that the problem mentioned in the current issue concerns only number items with no icon set ?
Using my sitemap, I noticed that problem only in this case.

resetnow:
I too only have noticed it with numbers but I think a more future-proof solution would iclude other cases.

lolodomo:
[...]

PR #3441, opened 2017-05-12
Icon servlet: failback to category "none" when the requested icon cat…

lolodomo:
…egory is provided by no icon provider

Fix issue #2784

### DIFF
--- a/bundles/ui/org.eclipse.smarthome.ui.icon/src/main/java/org/eclipse/smarthome/ui/icon/internal/IconServlet.java
+++ b/bundles/ui/org.eclipse.smarthome.ui.icon/src/main/java/org/eclipse/smarthome/ui/icon/internal/IconServlet.java
@@ -117,14 +117,14 @@ public class IconServlet extends HttpServlet {
         String state = getState(req);
         String iconSetId = getIconSetId(req);
 
-        IconProvider topProvider = null;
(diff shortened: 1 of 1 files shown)


bundles/ui/org.eclipse.smarthome.ui.icon/src/main/java/org/eclipse/smarthome/ui/icon/internal/IconServlet.java
```diff
@@ -126,6 +126,20 @@ protected void doGet(HttpServletRequest req, HttpServletResponse resp) throws Se
                 topProvider = provider;
             }
         }
+        if (topProvider == null) {
+            logger.debug(
+                    "Requested icon category {} provided by no icon provider; try to failback to category \"none\"",
+                    category);
+            // Try to failback to category "none"
+            category = "none";
+            for (IconProvider provider : iconProvider) {
```
kaikreuzer: You are duplicating the logic here - it would probably be much cleaner to extract that into a method, wdyt?

lolodomo: You're right. I will update my PR.

lolodomo:
Ok, done.

PR #3442, opened 2017-05-12
Add OnOffType as accepted Type for PlayerItem

pfink:
It could be very useful to also accept On/Off for PlayerItem (should be equivalent to Play/Pause similar to DimmerItem).

Signed-off-by: Patrick Fink <mail@pfink.de>


kaikreuzer:
No, sorry, I do not think this is useful - as you say, it is the same as Play/Pause and thus would duplicate the very same feature.
I guess your wish is rather to have an easy way to have a suitable widget for UIs - but this should hence be addressed on the sitemap/UI code.

pfink:
@kaikreuzer: It's not only on the UI-level, but also on the logical level. What if I would like to have rule that turns every device within a group off? Instead of having one command for all, I always have to implement special handling for Players and repeat myself. Within sitemaps, when using rules, ...

PR #3443, opened 2017-05-13
Classic UI: icon for group depending on state

lolodomo:

### DIFF
--- a/extensions/ui/org.eclipse.smarthome.ui.classic/snippets/group.html
+++ b/extensions/ui/org.eclipse.smarthome.ui.classic/snippets/group.html
@@ -1,6 +1,6 @@
 <li class="iGroup">
     <a href="#_%id%" onclick="AsyncLoad('%id%')">
-        <img src="../icon/%category%.%format%" width=29 height=29 />
+        <img src="../icon/%category%?format=%format%&state=%state%" width=29 height=29 />
         %label%
     </a>
 </li>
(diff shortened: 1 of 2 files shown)


lolodomo:
@resetnow : even if it is a fix for Classic UI, maybe you can have a look to this little code change ?

extensions/ui/org.eclipse.smarthome.ui.classic/snippets/group.html
```diff
@@ -1,6 +1,6 @@
 <li class="iGroup">
     <a href="#_%id%" onclick="AsyncLoad('%id%')">
-        <img src="../icon/%category%.%format%" width=29 height=29 />
+        <img src="../icon/%category%?format=%format%&state=%state%" width=29 height=29 />
```
kaikreuzer: Does this correctly work if there is no state (which is probably the normal situation for Groups)?

lolodomo: Normally "NULL" should be used and it should work but this should be tested.

lolodomo: As it is working well in Basic UI with such URL, I am very confident.

kaikreuzer: ok.

Issue #3323, opened 2017-04-27
New or existing properties not updated by PersistentInbox


wborn:
The PersistentInbox synchronizes existing configurations using new discovery results with the `synchronizeConfiguration(..)` method. It does not add or update Thing properties using new discovery results. I do not know if this is an issue/simply not implemented or some thoughful design decision. 

It would be helpful if it did add or modify properties because otherwise this logic gets duplicated to Things. An example of such a property would be a firmware version property.

PR #3447, opened 2017-05-13
LIFX improvements

wborn:
* Update light properties when Thing goes online which fixes:
  * Properties not always updated for lights defined in .things files
  * Existing light Things which do not get new/updated properties from discovery (#3323)
* Don't power on IR lights when changing infrared channel brightness
* Reuse the default thing ScheduledExecutorService
* Use final Logger and ReentrantLock objects

### DIFF
--- a/extensions/binding/org.eclipse.smarthome.binding.lifx/src/main/java/org/eclipse/smarthome/binding/lifx/LifxBindingConstants.java
+++ b/extensions/binding/org.eclipse.smarthome.binding.lifx/src/main/java/org/eclipse/smarthome/binding/lifx/LifxBindingConstants.java
@@ -29,7 +29,6 @@ import com.google.common.collect.Sets;
 public class LifxBindingConstants {
 
     public static final String BINDING_ID = "lifx";
-    public static final String THREADPOOL_NAME = "lifx";
 
     // The LIFX LAN Protocol Specification states that lights can process up to 20 messages per second, not more.
     public static final long PACKET_INTERVAL = 50;
--- a/extensions/binding/org.eclipse.smarthome.binding.lifx/src/main/java/org/eclipse/smarthome/binding/lifx/handler/LifxLightHandler.java
+++ b/extensions/binding/org.eclipse.smarthome.binding.lifx/src/main/java/org/eclipse/smarthome/binding/lifx/handler/LifxLightHandler.java
@@ -22,10 +22,12 @@ import org.eclipse.smarthome.binding.lifx.internal.LifxChannelFactory;
 import org.eclipse.smarthome.binding.lifx.internal.LifxLightCommunicationHandler;
 import org.eclipse.smarthome.binding.lifx.internal.LifxLightCurrentStateUpdater;
 import org.eclipse.smarthome.binding.lifx.internal.LifxLightOnlineStateUpdater;
+import org.eclipse.smarthome.binding.lifx.internal.LifxLightPropertiesUpdater;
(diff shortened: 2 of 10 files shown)


kaikreuzer:
> Don't power on IR lights when changing infrared channel brightness

Can you tell me more details about this change?
Why should the IR lights not go on if I send a certain brightness value for them?

wborn:
After some testing it showed that the [power level](https://lan.developer.lifx.com/v2.0/docs/device-messages#section-power-level) does not control if infrared light is on or off. It only controls the human visible light. I.e. it is analogue to pressing the power/standby button at the bottom of the [color wheel in the LIFX App](https://support.lifx.com/hc/en-us/articles/115001172283-LIFX-Plus-IR). 

So when you use the App and press this button to power off the light, it disables the human visible light and enables the IR light.

Now if you would use the binding with an empty `powerOnBrightness` value, and send an OFF command, the binding powers off the light (similar to pressing standby button in the App, not setting brightness to 0%).  The light then turns on the IR light. But if you then start changing the `infrared` channel state, the binding would again power on the human visible light.

kaikreuzer:
> The light then turns on the IR light.

Do we want this behavior? As we have the color channel for the visible light and the IR channel for the IR light, I would say these should behave totally isolated from each other. Therefore, I would say that IR should stay off (or on or whatever state it has), if the visible brightness is turned off.

>  the binding would again power on the human visible light.

Ok, but you comment spoke about "Don't power on IR lights" and not about powering on the visible light.

wborn:
> Do we want this behavior?

There is some "smart" logic in the light firmware which determines if it is useful to enable IR LEDs. See [this](https://community.lifx.com/t/new-lifx-with-ir-help/2037) LIFX community discussion:

* _The bulb does not try and work out when it is dark or not. Once an IR level is set the IR LEDs will light up whenever the visible light channels go under a self determined threshold. This threshold is controlled by the bulb and looks at heat and power consumption of that bulb_

There is currently no setting in the App or protocol packet that can enable/disable this "smart" logic. So total isolation of IR and visible light is currently not possible.

The fix in this PR does further isolate both channels.

kaikreuzer:
Well, this smart behavior makes sense as it only kicks in if "an IR level is set", i.e. the IR channel is not at 0. This feature simply assumes that it can turn off IR if there is enough visible light anyhow. I would assume that the IR channel does not change its state in this situation (or does it)?

But after all our discussion, I still do not understand what "Don't power on IR lights when changing infrared channel brightness" means. Can you at all turn on IR if the visible light is bright enough?

wborn:
> I would assume that the IR channel does not change its state in this situation (or does it)?

That's right. The IR channel is a setting which determines the IR LED brightness when the smart behavior determines that the IR LEDs should be enabled. The channel state does not change when IR LEDs are disabled due to enough visible light. 

> I still do not understand what "Don't power on IR lights when changing infrared channel brightness" means. Can you at all turn on IR if the visible light is bright enough?

In retrospect a better comment would have been "Don't power on the visible light of IR capable lights when the infrared channel state is changed". Because of the smart behavior it is currently not possible to force IR LEDs being activated when the visible light is bright enough.

kaikreuzer:
> Don't power on the visible light of IR capable lights when the infrared channel state is changed

Ah, that makes so much more sense :-)

wborn:
Thank you! Yes the code was well tested. I'll do a typo PR later. :-)

kaikreuzer:
> I'll do a typo PR later. :-)

Just include it in your next major rework of the binding - no need for a separate PR!

Issue #3450, opened 2017-05-14
Voice: provide feedback after keyword spotting using a Switch item


ghys:
When using the `DialogProcessor`, there is no way to tell when the keyword (or "magic word") has been spotted by the KS service and the system is now awaiting an command to interpret from the STT service.

I would like to propose a way of providing some feedback to the user in the following manner:

1. Adding a new config parameter for the voice bundle accepting the name of a Switch item;
2. Having the `DialogProcessor` send an ON command to this Switch item, if provided, when the keyword has been spotted - i.e. when its `processing` field changes to `true`;
3. Similarly, send an OFF command when a STT event was received, signifying the end of the "listening phase" (with a command sucessfully interpreted or not) - i.e. the `processing` field changes back to `false`.

![image](https://cloud.githubusercontent.com/assets/2004147/26037569/29684af2-38f6-11e7-9ee3-315347c4851f.png)

In this way, the nature of the feedback is left to the preference of the user - turning on a light, playing a sound (with a rule), etc.

PR #3451, opened 2017-05-14
Voice: send feedback to an item when listening for a command

ghys:
Implments #3450.

Signed-off-by: Yannick Schaus <eclipse@schaus.net>

### DIFF
--- a/bundles/core/org.eclipse.smarthome.core.voice/src/main/java/org/eclipse/smarthome/core/voice/internal/DialogProcessor.java
+++ b/bundles/core/org.eclipse.smarthome.core.voice/src/main/java/org/eclipse/smarthome/core/voice/internal/DialogProcessor.java
@@ -96,8 +96,10 @@ public class DialogProcessor implements KSListener, STTListener {
     public void start() {
         try {
             ks.spot(this, source.getInputStream(format), locale, this.keyword);
-        } catch (KSException | AudioException e) {
-            logger.error("Encountered error calling spot: {}", e);
+        } catch (KSException e) {
+            logger.error("Encountered error calling spot: {}", e.getMessage());
+        } catch (AudioException e) {
+            logger.error("Error creating the audio stream", e);
         }
     }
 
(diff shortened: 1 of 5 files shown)


kubawolanin:
Nice!
Is there any timeout in place, so the user knows when the Listening switch goes back to `OFF`?

bundles/core/org.eclipse.smarthome.core.voice/src/main/java/org/eclipse/smarthome/core/voice/internal/DialogProcessor.java
```diff
@@ -81,28 +88,42 @@ public DialogProcessor(KSService ks, STTService stt, TTSService tts, HumanLangua
...
-            processing = true;
             this.isSTTServerAborting = false;
             if (ksEvent instanceof KSpottedEvent) {
+                toggleProcessing(true);
                 if (stt != null) {
                     try {
                         this.sttServiceHandle = stt.recognize(this, source.getInputStream(format), this.locale,
                                 new HashSet<String>());
                     } catch (STTException | AudioException e) {
+                        logger.error("Error during recognition", e);
```
kaikreuzer: Is this really a "software bug" situation which requires error logging and a stack trace?

ghys: Actually I had initially an exception coming from the `source.getInputStream(format)` call (wrong capture frequency IIRC). So I only had the message "Error creating the audio input stream" without the cause. Maybe it makes sense to only log an `AudioException`?

ghys:
> Is there any timeout in place, so the user knows when the Listening switch goes back to OFF?

I thought it was better not to handle a timeout here but let the `STTService` do it itself. It simply needs to send a `RecognitionStopEvent` and the switch will be set back to OFF (meaning it's in the keyword spotting state again).

bundles/core/org.eclipse.smarthome.core.voice/src/main/java/org/eclipse/smarthome/core/voice/internal/DialogProcessor.java
kaikreuzer: > Maybe it makes sense to only log an AudioException?

Yes, I think so, as it is a unchecked exception, while STTException is a checked one.

kaikreuzer:
@ghys Could you update this PR regarding the logging/exception handling? We would be good to merge then.

ghys:
@kaikreuzer sure (was AFK for a few days) - done!

PR #3457, opened 2017-05-15
Utils to download WEB image

### LINKED ISSUE #3448: Add util methods in ESH core to dowload WEB image
It was already discussed when implementing support of image items in UIs through rawType, we need some help methods that could be used by any binding implementing image channel.
In particular, we need a method returning a rawType from an URL.
Some code could be extracted from the sonos binding. I will propose something.

lolodomo:
Sonos binding: download of cover art updated

Fix #3448 

### DIFF
--- a/bundles/io/org.eclipse.smarthome.io.net/META-INF/MANIFEST.MF
+++ b/bundles/io/org.eclipse.smarthome.io.net/META-INF/MANIFEST.MF
@@ -15,6 +15,7 @@ Import-Package: org.apache.commons.exec,
  org.eclipse.jetty.util,
  org.eclipse.jetty.util.component,
  org.eclipse.jetty.util.ssl,
+ org.eclipse.smarthome.core.library.types,
  org.eclipse.smarthome.io.net.exec,
  org.eclipse.smarthome.io.net.http,
  org.slf4j
--- a/bundles/io/org.eclipse.smarthome.io.net/src/main/java/org/eclipse/smarthome/io/net/http/HttpUtil.java
+++ b/bundles/io/org.eclipse.smarthome.io.net/src/main/java/org/eclipse/smarthome/io/net/http/HttpUtil.java
@@ -7,12 +7,15 @@
  */
 package org.eclipse.smarthome.io.net.http;
 
+import java.io.ByteArrayInputStream;
 import java.io.IOException;
 import java.io.InputStream;
+import java.io.UnsupportedEncodingException;
 import java.net.MalformedURLException;
 import java.net.URI;
 import java.net.URISyntaxException;
 import java.net.URL;
+import java.net.URLConnection;
 import java.util.List;
 import java.util.Properties;
 import java.util.concurrent.TimeUnit;
@@ -33,6 +36,7 @@ import org.eclipse.jetty.http.HttpStatus;
 import org.eclipse.jetty.util.B64Code;
 import org.eclipse.jetty.util.StringUtil;
 import org.eclipse.jetty.util.ssl.SslContextFactory;
+import org.eclipse.smarthome.core.library.types.RawType;
 import org.slf4j.Logger;
 import org.slf4j.LoggerFactory;
 
@@ -47,7 +51,15 @@ public class HttpUtil {
 
     private static Logger logger = LoggerFactory.getLogger(HttpUtil.class);
 
-    private static HttpClient client = new HttpClient(new SslContextFactory());
+    private static final HttpClient CLIENT = new HttpClient(new SslContextFactory());
+
+    private static class ProxyParams {
+        public String proxyHost = null;
+        public int proxyPort = 80;
+        public String proxyUser = null;
+        public String proxyPassword = null;
+        public String nonProxyHosts = null;
+    }
 
     /**
      * Executes the given <code>url</code> with the given <code>httpMethod</code>.
@@ -104,32 +116,11 @@ public class HttpUtil {
      */
     public static String executeUrl(String httpMethod, String url, Properties httpHeaders, InputStream content,
             String contentType, int timeout) throws IOException {
-        String proxySet = System.getProperty("http.proxySet");
 
-        String proxyHost = null;
-        int proxyPort = 80;
-        String proxyUser = null;
-        String proxyPassword = null;
-        String nonProxyHosts = null;
+        final ProxyParams proxyParams = prepareProxyParams();
 
-        if ("true".equalsIgnoreCase(proxySet)) {
-            proxyHost = System.getProperty("http.proxyHost");
-            String proxyPortString = System.getProperty("http.proxyPort");
-            if (StringUtils.isNotBlank(proxyPortString)) {
-                try {
-                    proxyPort = Integer.valueOf(proxyPortString);
-                } catch (NumberFormatException e) {
-                    logger.warn("'{}' is not a valid proxy port - using port 80 instead");
-                }
-            }
-            proxyUser = System.getProperty("http.proxyUser");
-            proxyPassword = System.getProperty("http.proxyPassword");
-            nonProxyHosts = System.getProperty("http.nonProxyHosts");
-        }
-
-        return executeUrl(httpMethod, url, httpHeaders, content, contentType, timeout, proxyHost, proxyPort, proxyUser,
-                proxyPassword, nonProxyHosts);
-
+        return executeUrl(httpMethod, url, httpHeaders, content, contentType, timeout, proxyParams.proxyHost,
+                proxyParams.proxyPort, proxyParams.proxyUser, proxyParams.proxyPassword, proxyParams.nonProxyHosts);
     }
 
     /**
@@ -147,6 +138,7 @@ public class HttpUtil {
      * @param proxyUser the username to authenticate with the proxy
      * @param proxyPassword the password to authenticate with the proxy
      * @param nonProxyHosts the hosts that won't be routed through the proxy
+     *
      * @return the response body or <code>NULL</code> when the request went wrong
      * @throws IOException when the request execution failed, timed out or it was interrupted
      */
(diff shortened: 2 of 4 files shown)


bundles/core/org.eclipse.smarthome.core/src/main/java/org/eclipse/smarthome/core/net/WebDataDownloadUtil.java
```diff
@@ -0,0 +1,148 @@
...
+     *         type or the data size is too big
+     */
+    public static RawType downloadData(URL url, String contentTypeRegex, boolean scanTypeInContent,
+            long maxContentLength) {
+
+        String contentType = RawType.DEFAULT_MIME_TYPE;
+        if (contentTypeRegex != null) {
+            contentType = WebDataDownloadUtil.getContentType(url, scanTypeInContent, maxContentLength);
+            if (contentType == null) {
+                LoggerFactory.getLogger(WebDataDownloadUtil.class)
```
hakan42: Could you create the Logger object only once?

lolodomo:
For static methods, a previous PR review requested that I did not declare it globally but like that. So what is the best choice ?

hakan42:
Ah, sorry, I did not notice that you had a static utility class here.

Honestly, I would avoid creating static helper classes at all but this might be my own preference 😄

bundles/core/org.eclipse.smarthome.core/src/main/java/org/eclipse/smarthome/core/net/WebDataDownloadUtil.java
```diff
@@ -0,0 +1,148 @@
...
+                        "Unexpected content type \"{}\" from URL {} ; download aborted", contentType,
+                        url.toExternalForm());
+                return null;
+            }
+        }
+
+        RawType rawData;
+        InputStream input = null;
+        try {
+            input = url.openStream();
```
sjka: Hard to get it right, just think about proxies, handling all possible error situations, not leaking any resources... Therefore please use the available http client (jetty).

bundles/core/org.eclipse.smarthome.core/src/main/java/org/eclipse/smarthome/core/net/WebDataDownloadUtil.java
```diff
@@ -0,0 +1,148 @@
...
+     *            data size
+     * @return a RawType object containing the downloaded data, null if the content type does not match the expected
+     *         type or the data size is too big
+     */
+    public static RawType downloadData(URL url, String contentTypeRegex, boolean scanTypeInContent,
+            long maxContentLength) {
+
+        String contentType = RawType.DEFAULT_MIME_TYPE;
+        if (contentTypeRegex != null) {
+            contentType = WebDataDownloadUtil.getContentType(url, scanTypeInContent, maxContentLength);
```
sjka: with `getContentType(...)` you already open a connection, just to open another one later on to download the content. Does not sound very efficient to me...

bundles/core/org.eclipse.smarthome.core/src/main/java/org/eclipse/smarthome/core/net/WebDataDownloadUtil.java
```diff
@@ -0,0 +1,148 @@
...
+     * @param scanTypeInContent true to allow the scan of data to determine the content type if not found in the headers
+     * @param maxContentLength the maximum data size in bytes to trigger the search; any negative value to ignore the
+     *            data size
+     * @return the MIME type of the content, null if the content type could not be found or the data size is too big
+     */
+    public static String getContentType(URL url, boolean scanTypeInContent, long maxContentLength) {
+        String contentType = null;
+        InputStream input = null;
+        try {
+            URLConnection connection = url.openConnection();
```
sjka: Again, URLConnection is _not_ our friend.

bundles/core/org.eclipse.smarthome.core/src/main/java/org/eclipse/smarthome/core/net/WebDataDownloadUtil.java
lolodomo: Yes, that's true. It is necessary when you scan the content to determine the content type. But when this scan is not required, I agree that we should avoid opening the connection twice. That means having only one function.

lolodomo:
@sjka : I am ok with your location proposal. I should have thought about io.

Regarding use of URLConnection, is it your opinion to not use it or a real guideline for the project ? Because it is already used in lots of bindings and even in the ESH core in URLAudioStream ?
@kaikreuzer : a commetn about URLConnection ?

kaikreuzer:
> Regarding use of URLConnection, is it your opinion to not use it or a real guideline for the project ? 

At least for bindings, this is the official guideline, see https://www.eclipse.org/smarthome/documentation/development/bindings/dependencies.html#for-http-operations.
Note that the HttpUtil is using Jetty, so if you are going to share a lot of code with this class, it will be your natural choice anyhow :-)

> it is already used in lots of bindings

These should be ideally adapted.

> even in the ESH core in URLAudioStream

Maybe also worth a look whether this should be changed.

lolodomo:
Guys, the interest for URLConnection is in particular its method guessContentTypeFromStream.

htreu:
@lolodomo you may still use `URLConnection#guessContentTypeFromStream` even when your code moved to o.e.sh.io.net and is refactored to use the HttpClient from jetty. The jetty client will give you a byte[] which you can feed easily into an ByteArrayInputStream and let URLConnection guess the content type. This way you dont have to connect twice and use the data at hand.

htreu:
HI @lolodomo, in addition to the comment above I wanted to make a suggestion on where to move your code (as @sjka suggested, the o.e.sh.io.net package is suitable): The [o.e.sh.net.io.http.HttpUtil](https://github.com/eclipse/smarthome/blob/master/bundles/io/org.eclipse.smarthome.io.net/src/main/java/org/eclipse/smarthome/io/net/http/HttpUtil.java) already has the logic to handle web connections including proxies and deal with different error situations. How about including your code into this class and enabling it to provide images alongside the current String values? In line [226](https://github.com/eclipse/smarthome/blob/master/bundles/io/org.eclipse.smarthome.io.net/src/main/java/org/eclipse/smarthome/io/net/http/HttpUtil.java#L226) it already gives you the content in raw bytes which are even suitable for the URLConnection#guessContentTypeFromStream (see above).
This way we are also able to use the new method in rules an assign image data to an ImageItem. What do you think?

lolodomo:
I think in most cases we don't want to download all the media just to identify its content type. Or I should if the content type cannot be identified, we want to bypass the full download of all the media. And I hope that guessContentTypeFromStream is enough clever to do it just by downloading the very beginning of the data.

htreu:
Yes, you are right. Having the full byte[] and then throwing it all away is not clever. However the jetty client alternatively allows to stream the response, too. Have a look at the client docu here (the very last paragraph): https://www.eclipse.org/jetty/documentation/9.4.x/http-client-api.html

lolodomo:
Ok, I have a new version working, based only on Jetty HTTP client. URLConnection is now used only for its static method guessContentTypeFromStream applied to the data buffer.
The only thing I don't like is that the full data is systematically downloaded even if finally not used in certain cases (file too big or unexpected content type for example).

What do you thing about the idea to have a first HEAD request to control media size and content type ? It will finally lead to 2 requests, one HEAD request + one GET request.

bundles/io/org.eclipse.smarthome.io.net/src/main/java/org/eclipse/smarthome/io/net/http/HttpUtil.java
```diff
@@ -319,4 +361,127 @@ private static void startHttpClient(HttpClient client) {
...
+            input = new ByteArrayInputStream(data);
+            contentType = URLConnection.guessContentTypeFromStream(input);
+            if (contentType != null && contentType.isEmpty()) {
+                contentType = null;
+            }
+        } catch (IOException e) {
+            logger.debug("Failed to determine content type: {}", e.getMessage());
+            contentType = null;
+        } finally {
+            IOUtils.closeQuietly(input);
```
kaikreuzer: If you used try-with-resources, you would not need IOUtils at all:
```
try(ByteArrayInputStream input = new ByteArrayInputStream(data)) {
   ...
}
```

bundles/io/org.eclipse.smarthome.io.net/META-INF/MANIFEST.MF
```diff
@@ -6,6 +6,7 @@ Bundle-Version: 0.9.0.qualifier
 Bundle-Vendor: Eclipse.org/SmartHome
 Bundle-RequiredExecutionEnvironment: JavaSE-1.8
 Import-Package: org.apache.commons.exec,
+ org.apache.commons.io,
```
kaikreuzer: you should be able to remove this dependency again, if you follow my suggestion below.

bundles/io/org.eclipse.smarthome.io.net/src/main/java/org/eclipse/smarthome/io/net/http/HttpUtil.java
lolodomo: Sorry, I don't understand your suggestion. IOUtils is only used to close quietly the stream.
I can suppress IOUtils by using a try/catch around a close call.
The current try/catch is for the call to the URLConnection method.

kaikreuzer: Just do it like the snippet I provided above. See details about try-with-resources at https://docs.oracle.com/javase/tutorial/essential/exceptions/tryResourceClose.html.

maggu2810: I assume you know that a resource is closed automatically by using try-with-resources.
In this case the `close` method of the `ByteArrayInputStream` is allowed to throw an `IOException` that need to be catch. So, you need an additional catch block, too (I assume you pointed to that topic -- correct?). But there is nothing we can do if the close fails at all (or are there any suggestions).

So, perhaps @kaikreuzer wanted to use a similar code as the following one, to get rid of the dependency.

```Java
    /**
     * Determine the content type from the data.
     *
     * @param data the data as a buffer of bytes
     * @return the MIME type of the content, null if the content type could not be found
     */
    public static String guessContentTypeFromData(byte[] data) {
        String contentType = null;
        try (final ByteArrayInputStream input = new ByteArrayInputStream(data)) {
            try {
                contentType = URLConnection.guessContentTypeFromStream(input);
                if (contentType != null && contentType.isEmpty()) {
                    contentType = null;
                }
            } catch (final IOException e) {
                logger.debug("Failed to determine content type: {}", e.getMessage());
            }
        } catch (final IOException ex) {
            // Error on closing input stream -- nothing we can do here.
        }
        return contentType;
    }
```

bundles/io/org.eclipse.smarthome.io.net/src/main/java/org/eclipse/smarthome/io/net/http/HttpUtil.java
```diff
@@ -49,6 +54,12 @@
 
     private static HttpClient client = new HttpClient(new SslContextFactory());
 
+    private static String proxyHost = null;
+    private static int proxyPort = 80;
+    private static String proxyUser = null;
+    private static String proxyPassword = null;
+    private static String nonProxyHosts = null;
```
maggu2810: I would prefer we change the code to not use such global variables that will be used by different invokation.

bundles/io/org.eclipse.smarthome.io.net/src/main/java/org/eclipse/smarthome/io/net/http/HttpUtil.java
```diff
@@ -243,6 +264,27 @@ public boolean matches(String type, URI uri, String realm) {
...
+    private static void prepareProxyParams() {
+        String proxySet = System.getProperty("http.proxySet");
+        if ("true".equalsIgnoreCase(proxySet)) {
+            proxyHost = System.getProperty("http.proxyHost");
+            String proxyPortString = System.getProperty("http.proxyPort");
+            if (StringUtils.isNotBlank(proxyPortString)) {
+                try {
+                    proxyPort = Integer.valueOf(proxyPortString);
+                } catch (NumberFormatException e) {
+                    logger.warn("'{}' is not a valid proxy port - using port 80 instead");
```
maggu2810: Parameter missing (@lolodomo I will send you a PR with a small change -- you can have then a look at -- it will contain this fix.).

maggu2810: https://github.com/lolodomo/smarthome/pull/1

bundles/io/org.eclipse.smarthome.io.net/src/main/java/org/eclipse/smarthome/io/net/http/HttpUtil.java
```diff
@@ -147,13 +137,52 @@ public static String executeUrl(String httpMethod, String url, Properties httpHe
...
             String proxyPassword, String nonProxyHosts) throws IOException {
 
+        ContentResponse response = executeUrlAndGetReponse(httpMethod, url, httpHeaders, content, contentType, timeout,
+                proxyHost, proxyPort, proxyUser, proxyPassword, nonProxyHosts);
+        String encoding = response.getEncoding() != null ? response.getEncoding().replaceAll("\"", "").trim() : "UTF-8";
+        String responseBody;
+        try {
+            responseBody = new String(response.getContent(), encoding);
+            if (!responseBody.isEmpty()) {
+                logger.trace(responseBody);
```
maggu2810: Quality tools will complain about that statement because the string is not constant.

bundles/io/org.eclipse.smarthome.io.net/src/main/java/org/eclipse/smarthome/io/net/http/HttpUtil.java
htreu: I noticed that too. In this case the global statics will only be filled by also globally defined system properties. Otherwise I would prefer HttpUtils not to be static which is API breaking...

maggu2810: System properties could change and this will not reflected correctly.
See comment at lolodomo#1

htreu: agreed.

lolodomo:
All changes done. Thanks to @maggu2810 for its code contribution.

kaikreuzer:
Thanks @lolodomo!

kaikreuzer:
I'll do another stable ESH build this afternoon, so that this feature becomes available to openhab2-addons as well.

PR #3464, opened 2017-05-16
New API function to retrieve members of a GroupItem

### LINKED ISSUE #586: Get all members including group items
migrated from Bugzilla [#465201](https://bugs.eclipse.org/bugs/show_bug.cgi?id=465201)
status NEW severity _enhancement_ in component _Core_ for _---_
Reported in version _unspecified_ on platform _PC_
Assigned to: Project Inbox

On 2015-04-22 10:11:04 -0400, Smart Home wrote:

> Currently getAllMembers() returns all members of group item excluding group items.
> Can it be extended (new function?) to support returning all memebers including group items?

On 2015-04-27 09:09:45 -0400, Dennis Nobel wrote:

> We can not change the existing method because of backward-compatibility, but we can introduce another method. Do you need in in the Java API or would the REST API be sufficient?

On 2015-04-28 15:29:29 -0400, Smart Home wrote:

> I need it in the Java API as rest API is not supported with my application.
> 
> Do you plan to add something like predicates in the future?

On 2015-04-28 16:14:18 -0400, Dennis Nobel wrote:

> I think this is related to the discussion in the forum: https://www.eclipse.org/forums/index.php/m/1692318/?srch=guava#msg_1692318 ?
> 
> I think the concept of predicates can be supported, but we need to decide wether we want to have Guava dependencies or we want to define our own interfaces.

On 2015-04-28 16:30:02 -0400, Smart Home wrote:

> Maybe we can start by defining the same interface Guava have for predicates and in the future replace it if Guava will be used. 
> 
> Starting using predicates now will save the need to add different functions of filtering and will be easy to replace by official predicates in the future.

On 2015-04-29 12:35:42 -0400, Dennis Nobel wrote:

> > Maybe we can start by defining the same interface Guava have for predicates and in the future replace it if Guava will be used. 
> 
> This would break the API, if we later change to Guava.
> 
> Guava is already used in the internal implementation of Eclipse SmartHome, but it is not yet part of the API. As we tend to have a clean API without dependencies to third party libs, we should better define our own interface and see how it can be combined with Guava.

triller-telekom:
With Java 8 we can now use streams and their filter functions, so this method can retrieve any members matching a given filter

Fixes #586

### DIFF
--- a/bundles/core/org.eclipse.smarthome.core.test/src/test/groovy/org/eclipse/smarthome/core/items/GroupItemTest.groovy
+++ b/bundles/core/org.eclipse.smarthome.core.test/src/test/groovy/org/eclipse/smarthome/core/items/GroupItemTest.groovy
@@ -10,6 +10,8 @@ package org.eclipse.smarthome.core.items
 import static org.hamcrest.CoreMatchers.*
 import static org.junit.Assert.*
 
+import java.util.function.Predicate
+
 import org.eclipse.smarthome.core.events.Event
 import org.eclipse.smarthome.core.events.EventPublisher
 import org.eclipse.smarthome.core.items.events.GroupItemStateChangedEvent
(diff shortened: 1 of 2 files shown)


bundles/core/org.eclipse.smarthome.core.test/src/test/groovy/org/eclipse/smarthome/core/items/GroupItemTest.groovy
```diff
@@ -92,6 +94,55 @@ class GroupItemTest extends OSGiTest {
...
+                    @Override
+                    public boolean test(Item t) {
+                        if(t instanceof GroupItem) {
+                            return true;
+                        }
+                        return false;
+                    }
+                };
+
+        Set<Item> members = rootGroupItem.getMembers(groupItemsInMembers)
```
triller-telekom: Just figured out they do, but they have a slightly different syntax. Works now.

bundles/core/org.eclipse.smarthome.core/src/main/java/org/eclipse/smarthome/core/items/GroupItem.java
```diff
@@ -109,22 +111,36 @@ public GroupFunction getFunction() {
...
+        Set<Item> allMembers = getMembers((Item i) -> !(i instanceof GroupItem));
         return ImmutableSet.copyOf(allMembers);
     }
 
     private void collectMembers(Set<Item> allMembers, Set<Item> members) {
         for (Item member : members) {
             if (member instanceof GroupItem) {
                 collectMembers(allMembers, ((GroupItem) member).members);
+                allMembers.add(member);
+
```
maggu2810: Have you added this empty line by intention?

bundles/core/org.eclipse.smarthome.core/src/main/java/org/eclipse/smarthome/core/items/GroupItem.java
```diff
@@ -109,22 +111,36 @@ public GroupFunction getFunction() {
      * @return all members of this and all contained {@link GroupItem}s
      */
     public Set<Item> getAllMembers() {
-        Set<Item> allMembers = new HashSet<Item>();
-        collectMembers(allMembers, members);
+        Set<Item> allMembers = getMembers((Item i) -> !(i instanceof GroupItem));
```
maggu2810: Now we can also use a one liner...

WRT: http://stackoverflow.com/a/31734226/7546824

bundles/core/org.eclipse.smarthome.core/src/main/java/org/eclipse/smarthome/core/items/GroupItem.java
triller-telekom: No, this unfortunately just happens to me sometimes. I create them to get a better overview while coding and moving lines around and then forget to clean them up afterwards. I have removed it now.

bundles/core/org.eclipse.smarthome.core/src/main/java/org/eclipse/smarthome/core/items/GroupItem.java
triller-telekom: Indeed, I have removed the creation of the local variable there.

triller-telekom:
Arrrg travis, c'mon you can download the bundles... try harder...

kaikreuzer:
restarted Travis...

PR #3467, opened 2017-05-16
An Interface method should not allow for throwing a generic Exception

### LINKED ISSUE #2396: MQTT Transport: Excpetions
I recently had a look at "illegal catches" in my code using Checkstyle: http://checkstyle.sourceforge.net/config_coding.html#IllegalCatch

So I noticed that the MqttSenderChannel provides an interface method publish that states that this method throws "Exception".
It seems very uncommon to use such the class "Exception" in an API.

Shouldn't this be changed on the interface to an IOException or something similar?

triller-telekom:
Fixes #2396

### DIFF
--- a/bundles/io/org.eclipse.smarthome.io.transport.mqtt/src/main/java/org/eclipse/smarthome/io/transport/mqtt/MqttSenderChannel.java
+++ b/bundles/io/org.eclipse.smarthome.io.transport.mqtt/src/main/java/org/eclipse/smarthome/io/transport/mqtt/MqttSenderChannel.java
@@ -7,6 +7,8 @@
  */
 package org.eclipse.smarthome.io.transport.mqtt;
 
+import java.io.IOException;
+
 /**
  * Callback interface for sending a message to the MqttBrokerConnection.
  *
@@ -21,6 +23,6 @@ public interface MqttSenderChannel {
      * @param message message payload.
      * @throws Exception if an error occurs during sending.
      */
-    public void publish(String topic, byte[] message) throws Exception;
+    public void publish(String topic, byte[] message) throws IOException;
 
 }
--- a/bundles/io/org.eclipse.smarthome.io.transport.mqtt/src/main/java/org/eclipse/smarthome/io/transport/mqtt/MqttSenderChannel.java
+++ b/bundles/io/org.eclipse.smarthome.io.transport.mqtt/src/main/java/org/eclipse/smarthome/io/transport/mqtt/MqttSenderChannel.java
@@ -21,7 +21,7 @@ public interface MqttSenderChannel {
      *
      * @param topic Topic to publish the message to.
      * @param message message payload.
-     * @throws Exception if an error occurs during sending.
+     * @throws IOException if an error occurs during sending.
      */
     public void publish(String topic, byte[] message) throws IOException;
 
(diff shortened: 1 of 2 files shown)


bundles/io/org.eclipse.smarthome.io.transport.mqtt/src/main/java/org/eclipse/smarthome/io/transport/mqtt/internal/MqttBrokerConnection.java
```diff
@@ -457,19 +458,22 @@ public void publish(String topic, byte[] payload) throws Exception {
...
+                        // wait for publish confirmation
+                        deliveryToken.waitForCompletion(10000);
+                        if (!deliveryToken.isComplete()) {
+                            logger.error(
+                                    "Did not receive completion message within timeout limit whilst publishing to topic '{}'",
+                                    topic);
+                        }
                     }
+                } catch (MqttException e) {
+                    logger.error("Could not publish message to topic {}", topic);
```
kaikreuzer: Does this comply with our [logging guidelines](https://www.eclipse.org/smarthome/documentation/development/guidelines.html#e-logging)?
Not sure if it should be ERROR level. And if so, we might want the exception message in the log as well and if it is a system bug, there should be the stacktrace as well.

bundles/io/org.eclipse.smarthome.io.transport.mqtt/src/main/java/org/eclipse/smarthome/io/transport/mqtt/MqttSenderChannel.java
```diff
@@ -21,6 +23,6 @@
      * @param message message payload.
      * @throws Exception if an error occurs during sending.
      */
-    public void publish(String topic, byte[] message) throws Exception;
+    public void publish(String topic, byte[] message) throws IOException;
```
maggu2810: You need to change the `@throws` parameter, too.

bundles/io/org.eclipse.smarthome.io.transport.mqtt/src/main/java/org/eclipse/smarthome/io/transport/mqtt/internal/MqttBrokerConnection.java
maggu2810: Should we rely on the message at all? Why not using `, topic, e` and leave it up to the distributor / user if a stack trace is logged.

triller-telekom: I reduced it to trace but added the exception, because if publishing a message goes wrong this is probably only interesting for debug purposes and while debugging one is interested in the exception as well.

bundles/io/org.eclipse.smarthome.io.transport.mqtt/src/main/java/org/eclipse/smarthome/io/transport/mqtt/internal/MqttBrokerConnection.java
```diff
@@ -457,19 +458,22 @@ public void publish(String topic, byte[] payload) throws Exception {
...
+                        // wait for publish confirmation
+                        deliveryToken.waitForCompletion(10000);
+                        if (!deliveryToken.isComplete()) {
+                            logger.error(
+                                    "Did not receive completion message within timeout limit whilst publishing to topic '{}'",
+                                    topic);
+                        }
                     }
+                } catch (MqttException e) {
+                    logger.error("Could not publish message to topic {} {}", topic, e);
```
maggu2810: No placeholder for exceptions

triller-telekom:
I removed the placeholder.

sjka:
@maggu2810 assuming that your veto really only was about the placeholder in the log statement and this has been fixed I assume your consent that it can be merged...

maggu2810:
@sjka You are right

PR #3478, opened 2017-05-17
Add basic test classes for static import on content assist.

htreu:
Add the following types for static import on content assist:

* org.junit.Assert.*
* org.mockito.Mockito.*
* org.mockito.MockitoAnnotations.*
* org.hamcrest.CoreMatchers.*


sjka:
Thanks, always missed that (in Groovy though 😉 ).

Issue #3481, opened 2017-05-18
FolderObserverTest fails in project org.eclipse.smarthome.model.core.test


frhei:
Build of latest commit ea5b67372d8a7dd00c52e8d41c699e4a440a5d72 fails in test case FolderObserverTest of test project org.eclipse.smarthome.model.core.test:

```
Running org.eclipse.smarthome.model.core.internal.folder.test.FolderObserverTest
Tests run: 10, Failures: 0, Errors: 1, Skipped: 0, Time elapsed: 12.955 sec <<< FAILURE! - in org.eclipse.smarthome.model.core.internal.folder.test.FolderObserverTest
test the updating of the configuration with a non existing subdirectory(org.eclipse.smarthome.model.core.internal.folder.test.FolderObserverTest)  Time elapsed: 0.461 sec
test modification of file with valid extension in an existing subdirectory(org.eclipse.smarthome.model.core.internal.folder.test.FolderObserverTest)  Time elapsed: 2.147 sec
test that an exception does not kill the watcher(org.eclipse.smarthome.model.core.internal.folder.test.FolderObserverTest)  Time elapsed: 2.055 sec
test ignore hidden files with extension that is registered(org.eclipse.smarthome.model.core.internal.folder.test.FolderObserverTest)  Time elapsed: 1.234 sec  <<< ERROR!
groovy.lang.MissingPropertyException: No such property: set for class: java.nio.file.Files
        at groovy.lang.MetaClassImpl.invokeStaticMissingProperty(MetaClassImpl.java:996)
        at groovy.lang.MetaClassImpl.getProperty(MetaClassImpl.java:1851)
        at groovy.lang.MetaClassImpl.getProperty(MetaClassImpl.java:1827)
        at groovy.lang.MetaClassImpl.getProperty(MetaClassImpl.java:3690)
        at org.codehaus.groovy.runtime.callsite.ClassMetaClassGetPropertySite.getProperty(ClassMetaClassGetPropertySite.java:48)
        at org.codehaus.groovy.runtime.callsite.AbstractCallSite.callGetProperty(AbstractCallSite.java:227)
        at org.eclipse.smarthome.model.core.internal.folder.test.FolderObserverTest.test ignore hidden files with extension that is registered(FolderObserverTest.groovy:521)

[...]

PR #3490, opened 2017-05-18
fixed FolderObserverTest for hidden files on windows systems

frhei:
PR #3483 solves the root cause of issue #3481, so that the test 'test ignore hidden files with extension that is registered' can execute in general.
Unfortunately, the test still fails on windows systems, because the way of creating the file triggers the obervation mechanism and the assertion fails.
In windows a hidden file cannot be created with a single api call (at least not with the Java api).
The file must be created and afterwards it needs a filesystem property set for a file to be hidden.
But the initial creation already triggers the folder observer mechanism, therefore the following strategy can solve the problem:

- create a file  in an unobserved directory
- hide it
- afterwards move it to the observed directory

P.S.: Needed to change file FolderObserverTestsConfiguration.launch to be able to execute test in eclipse and debug.


kaikreuzer:
Thanks @frhei!
Please note that you need to have an ECA and sign-off your commits in order for us to be able to accept your PR. Please see https://wiki.eclipse.org/Development_Resources/Contributing_via_Git#Eclipse_Contributor_Agreement for details.

frhei:
Thanks for the hint!
I am pretty new to this topic and was not aware of the need having an signed ECA.
Meanwhile I did so.
What are the next steps? Do I have to re-commit? (Email-adresses of ECA and the one used in commit are different!)

sjka:
Thanks for your patience with the little bureaucracy we unfortunately have to apply 😄 

Yes, since the email address is different we need to ask you to please do so. Just  amend your commit and also don't forget to add a "signed-off-by" line (git actually does it for your with the `-s` argument). 

In order to change everything you first need to configure git to use the correct email in this repository with

```
git config user.email "email@address.com"
```

and then fix the existing commit by doing a 

```
git commit --amend -s --author "New Author Name <email@address.com>" 
```

then you should be all set and simply need to force-push the branch again.

frhei:
No worries and thanks for your support :)
Seems to have worked, so ready to go ahead...

sjka:
Indeed, all good now, thanks for the fix.
Congratulations to your first contribution!

PR #3494, opened 2017-05-19
remove "Type" postfix from state types in item events

### LINKED ISSUE #3282: Inconsistency SSE Type and REST State
**tl;dr**
As a generic UI I have no chance to find out, which type the current state of an DimmerItem has, until I receive an event-message via SSE with that information. Thats a problem at initial rendering because I cannot trigger that event-message.

**Details**
I hope this is an understanding problem at my side. I tried to find an issue related to this but it is not easy to search for that topic.

I have an Dimmer item defined, which state is `24`. Now I do a HTTP POST with a body `26`, then it has the new state `26`. Until here as expected.

Now I POST a `0.27` and my UI interprets it as 0.27%, but another UI interprets it as 27%.

If I retrieve the item's state via the REST API, I have no way of knowing the type of that state value - because that type is only provided via the event stream (SSE).

All I really need is an additional field in the JSON representation that corresponds to the event payload's type information.

Another thing here is that I cannot provide a type information with the REST call, but the SSE Events contain a Type (`PercentType`) after the state changes/updates. That type also never changes.

A solution would be to provide that type information on the Item-endpoint JSON (ItemResource) as sibling to the `state` property.
So that additional field in the JSON representation should correspond to the event payload's type information.

Examples:
```
$ curl -X POST -H 'Content-Type: text/plain' --data-binary 70 http://localhost:8080/rest/items/DemoRoom_Dimmer
→ { type: "PercentType", value: "70" }
```

```
$ curl -X POST -H 'Content-Type: text/plain' --data-binary 0.6 http://localhost:8080/rest/items/DemoRoom_Dimmer
→ { type: "PercentType", value: "0.6" }
```

```
curl -X POST -H 'Content-Type: text/plain' --data-binary ON http://localhost:8080/rest/items/DemoRoom_Dimmer
→ { type: "PercentType", value: "100" }
```

htreu:
This removes the "Type" postfix from state types in Item specific events. PaperUI event handling is adopted. From what I can see in the code the BasicUI is not affected by this change.

Fixes #3282.

### DIFF
--- a/bundles/core/org.eclipse.smarthome.core/src/main/java/org/eclipse/smarthome/core/items/events/ItemEventFactory.java
+++ b/bundles/core/org.eclipse.smarthome.core/src/main/java/org/eclipse/smarthome/core/items/events/ItemEventFactory.java
@@ -7,10 +7,12 @@
  */
 package org.eclipse.smarthome.core.items.events;
 
+import java.lang.reflect.InvocationTargetException;
 import java.lang.reflect.Method;
 import java.util.LinkedList;
 import java.util.List;
 
+import org.apache.commons.lang.StringUtils;
 import org.eclipse.smarthome.core.events.AbstractEventFactory;
 import org.eclipse.smarthome.core.events.Event;
 import org.eclipse.smarthome.core.items.Item;
@@ -32,6 +34,8 @@ import com.google.common.collect.Sets;
  * @author Stefan Bußweiler - Initial contribution
  */
 public class ItemEventFactory extends AbstractEventFactory {
+
+    private static final String TYPE_POSTFIX = "Type";
 
     private static final String CORE_LIBRARY_PACKAGE = "org.eclipse.smarthome.core.library.types.";
 
@@ -85,18 +89,12 @@ public class ItemEventFactory extends AbstractEventFactory {
         State state = getState(bean.getType(), bean.getValue());
         State oldState = getState(bean.getOldType(), bean.getOldValue());
         return new GroupItemStateChangedEvent(topic, payload, itemName, memberName, state, oldState);
-
     }
 
     private Event createCommandEvent(String topic, String payload, String source) {
         String itemName = getItemName(topic);
         ItemEventPayloadBean bean = deserializePayload(payload, ItemEventPayloadBean.class);
-        Command command = null;
-        try {
-            command = (Command) parse(bean.getType(), bean.getValue());
-        } catch (Exception e) {
-            throw new IllegalArgumentException("Parsing of item command event failed.", e);
-        }
+        Command command = parseType(bean.getType(), bean.getValue(), Command.class);
         return new ItemCommandEvent(topic, payload, itemName, command, source);
     }
 
@@ -116,13 +114,7 @@ public class ItemEventFactory extends AbstractEventFactory {
     }
 
     private State getState(String type, String value) {
-        State state = null;
-        try {
-            state = (State) parse(type, value);
-        } catch (Exception e) {
-            throw new IllegalArgumentException("Parsing of item state event failed.", e);
-        }
-        return state;
+        return parseType(type, value, State.class);
     }
 
     private String getItemName(String topic) {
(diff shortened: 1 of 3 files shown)


bundles/core/org.eclipse.smarthome.core/src/main/java/org/eclipse/smarthome/core/items/events/ItemEventFactory.java
```diff
@@ -93,7 +96,7 @@ private Event createCommandEvent(String topic, String payload, String source) {
         ItemEventPayloadBean bean = deserializePayload(payload, ItemEventPayloadBean.class);
         Command command = null;
         try {
-            command = (Command) parse(bean.getType(), bean.getValue());
+            command = (Command) parse(bean.getType() + TYPE_POSTFIX, bean.getValue());
```
kaikreuzer: Wait, you add the post fix here, because the internal implementation of parse requires it and in order to be able to pass it from the outside you change the parameter from typeName to simpleClassName? Why not keep the signature as is and deal with it only within the parse method?

maggu2810: I agree.

```java
final String simpleClassName = type + TYPE_POSTFIX;
```

as first statement in the `parse` method is perhaps a little bit "better".

htreu: My idea was to have `parse` as my internal API which parses class names to instances. I didnt want the method to know about the fact that we pass around truncated class names. To make that even more clear I changed the parameter to `simpleClassName`. This respects the open/closed principle in a way that we are open for Commands and States with different class name patterns and the method is closed for modification because it will always do what it says it does.
I see your point though, that we handle all class names the same in this case and parse may also be the one knowing about the truncation. wdyt?

maggu2810: > open/closed principle

We could use a `parseType` and `parseSimpleClassName` method then no one needs to be changes anymore. :wink:

htreu: tnx @maggu2810, I will try this in code and update the PR.

htreu:
I´m completely with you @maggu2810. Already thought about the Exception handling by myself... I will give it another shot tomorrow!

htreu:
I refactored the exception handling so that `parseSimpleClassName` has to deal with all the dirty stuff. `parseType` will only report if parsed type and desired type mismatch and the actual caller does not have to worry about exception handling. Since we are only parsing classes used in this very same context this should be "save".
One more thing: all subclasses of `State` and `Command` do define a public static method `valueOf(String)` but are not forced to do so by interface of superclass. This is okay for the moment but makes this part of class/value parsing weak somehow.

bundles/core/org.eclipse.smarthome.core/src/main/java/org/eclipse/smarthome/core/items/events/ItemEventFactory.java
```diff
@@ -141,16 +133,45 @@ private String getMemberName(String topic) {
         return topicElements[3];
     }
 
-    private Object parse(String typeName, String valueToParse) throws Exception {
-        if (typeName.equals(UnDefType.class.getSimpleName())) {
+    @SuppressWarnings("unchecked")
```
maggu2810: Why do you want to suppress unchecked warning?

IMHO it is much simpler to use the cast method of the given class as you already checked that is it assignable.

```java
    private <T> T parseType(final String typeName, final String valueToParse, final Class<T> desiredClass) {
        Object parsedObject = null;
        final String simpleClassName = typeName + TYPE_POSTFIX;
        parsedObject = parseSimpleClassName(simpleClassName, valueToParse);

        if (parsedObject == null || !desiredClass.isAssignableFrom(parsedObject.getClass())) {
            final String parsedObjectClassName = parsedObject != null ? parsedObject.getClass().getName()
                    : "<undefined>";
            throw new IllegalArgumentException("Error parsing simpleClasssName '" + simpleClassName + "' with value '"
                    + valueToParse + "'. Desired type was '" + desiredClass.getName() + "' but got '"
                    + parsedObjectClassName + "'.");
        } else {
            return desiredClass.cast(parsedObject);
        }
    }
```

htreu: nice, didn´t now of that method!

maggu2810:
I restarted Travis to get (hopefully) a successful integration build.

htreu:
Rebased and conflicts resolved.

maggu2810:
@sjka @kaikreuzer Okay to merge or should it be postponed a few days because of the API break?

kaikreuzer:
ok to merge.

Issue #2902, opened 2017-01-26
References to items in the Frame declaration, within a sitemap, block updates on icons!


sailorsail:
In a clean install of openHAB 2.0 Release Build, I've used the following configuration:
file.items
```csv
Switch Light_WCSmall_Ceiling 	"Light Ceiling WC Small" 				
Switch Light_WCBig_Ceiling 	"Light Ceiling WC Big" 				
Number Temp_WCBig 		"Temperature WC Big [%.2f °C]" 	
```
file.sitemap
```php
sitemap demo label="demo" {
	Text label="WC1" {
		Frame label="test" {
			Switch item=Light_WCBig_Ceiling
			}
		}
	Text label="WC2" {
		Frame item=Temp_WCBig {
			Switch item=Light_WCSmall_Ceiling
			}
		}
	}
```
When I toggle 'Light_WCBig_Ceiling', within “WC1” page, the update of the icon is performed; no refresh is necessary on the user interface (either iOS or Web Basic UI).
However if I toggle 'Light_WCBig_Ceiling', within “WC2” page, the update of the icon is not performed.
Nevertheless, if I refresh the page, navigate back, and in again, the change is then shown.
The only diference here is the use of item=Temp_WCBig in the Frame declaration.
Allegedly, the use of items in frame declarations is possible, but I wasn’t expecting the different behaviour. Shouldn’t the reaction to the toggle be consistent, or is this the expected behaviour?
Do you have any ideias about what's going on?

kaikreuzer:
Same question that I asked before: Did you identify this to be a problem with the Basic UI only? Does it work with the Classic UI and the native apps?

sailorsail:
[I did](https://community.openhab.org/t/do-references-to-items-on-frame-declaration-in-the-sitemap-block-updates-on-icons/21083/5), but I forgot to mention it here! Either Basic UI, Classic UI or native iOS present the same behaviour.

lolodomo:
I can confirm this bug, I was able to reproduce it.
Will try to analyze it.
Maybe a problem with produced SSE events ?

lolodomo:
In fact, all items in such a frame are simply not updated, even when the update is triggered outside.

I should mention some validation errors of my sitemap at OH startup when using such a frame:
```
2017-04-30 20:24:01.924 [ERROR] [xtext.validation.CompositeEValidator] - Error executing EValidator
[...]

PR #3495, opened 2017-05-20
Classic UI: fix dynamic widget visibility and dynamic update of the page title

### LINKED ISSUE #3176: Classic UI: page title containing item state is not updated
If you open a frame linked to an item, the title at the top of the page display the item label with including its state. If the item state is then changed, the title is not updated.
Exiting the frame and re-opening it then oif course displays the correct title.

### LINKED ISSUE #3488: Classic UI: update of widget visibility only when reloading the page
In Classic UI, there is no dynamic display of widgets depending on a visibility attribute.
The visibility attribute is considered only when the page is reloaded.

PS: Basic UI is doing that dynamically.

lolodomo:
Fix #3488
Fix #3176 

### DIFF
--- a/extensions/ui/org.eclipse.smarthome.ui.classic/src/main/java/org/eclipse/smarthome/ui/classic/internal/servlet/WebAppServlet.java
+++ b/extensions/ui/org.eclipse.smarthome.ui.classic/src/main/java/org/eclipse/smarthome/ui/classic/internal/servlet/WebAppServlet.java
@@ -19,8 +19,10 @@ import javax.servlet.ServletException;
 import javax.servlet.ServletRequest;
 import javax.servlet.ServletResponse;
 
+import org.eclipse.emf.common.util.BasicEList;
 import org.eclipse.emf.common.util.EList;
 import org.eclipse.smarthome.core.items.GenericItem;
+import org.eclipse.smarthome.core.items.GroupItem;
 import org.eclipse.smarthome.core.items.Item;
 import org.eclipse.smarthome.core.items.ItemNotFoundException;
 import org.eclipse.smarthome.core.items.StateChangeListener;
@@ -29,6 +31,7 @@ import org.eclipse.smarthome.model.sitemap.Frame;
 import org.eclipse.smarthome.model.sitemap.LinkableWidget;
 import org.eclipse.smarthome.model.sitemap.Sitemap;
 import org.eclipse.smarthome.model.sitemap.SitemapProvider;
+import org.eclipse.smarthome.model.sitemap.VisibilityRule;
 import org.eclipse.smarthome.model.sitemap.Widget;
 import org.eclipse.smarthome.ui.classic.internal.WebAppConfig;
 import org.eclipse.smarthome.ui.classic.internal.render.PageRenderer;
@@ -129,7 +132,7 @@ public class WebAppServlet extends BaseServlet {
             if (sitemap == null) {
                 throw new RenderException("Sitemap '" + sitemapName + "' could not be found");
             }
-            logger.debug("reading sitemap {}", sitemap.getName());
+            logger.debug("reading sitemap {} widgetId {} async {} poll {}", sitemap.getName(), widgetId, async, poll);
             if (widgetId == null || widgetId.isEmpty() || widgetId.equals("Home")) {
                 // we are at the homepage, so we render the children of the sitemap root node
                 String label = sitemap.getLabel() != null ? sitemap.getLabel() : sitemapName;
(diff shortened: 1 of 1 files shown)


lolodomo:
This PR also fixes the issue #2902 when using Classic UI.
Issue #2902 requires an additional fix for Basic UI.

extensions/ui/org.eclipse.smarthome.ui.classic/src/main/java/org/eclipse/smarthome/ui/classic/internal/servlet/WebAppServlet.java
```diff
@@ -221,27 +228,32 @@ private boolean waitForChanges(EList<Widget> widgets) {
...
                 }
             }
         }
         return items;
     }
 
+    private void addItemWithName(Set<GenericItem> items, String itemName) {
+        if (itemName != null) {
+            try {
+                Item item = itemRegistry.getItem(itemName);
```
triller-telekom: I know you just re-factored this code into another method, but I am asking myself why the `getItem()` method is called on the `itemRegistry` vs `itemUIregistry` in case of the basicui.

lolodomo: You're tight. I changed to itemUIregistry in Classic UI, it still works for my use cases to test the issues I am fixing with this PR. I have not done a full analysis of all possible consequences of this change from itemRegistry to itemUIregistry but I imagine it should be ok as it is with Basic UI.

lolodomo:
Comment taken into account + changes squashed.

extensions/ui/org.eclipse.smarthome.ui.classic/src/main/java/org/eclipse/smarthome/ui/classic/internal/servlet/WebAppServlet.java
```diff
@@ -275,7 +289,9 @@ public boolean hasChangeOccurred() {
          */
         @Override
         public void stateUpdated(Item item, State state) {
-            // ignore if the state did not change
+            if (item instanceof GroupItem) {
+                changed = true;
```
kaikreuzer: This looks pretty dangerous to me.
Note that the Classic UI intentionally had some constraints regarding dynamic updates as every update results in a full page refresh, which lets the screen flicker, scroll and makes the UI difficult to use, if updates are too frequent.
Especially GroupItems receive MANY state updates and thus they are prone to such negative effects.

@lolodomo & @triller-telekom Did you extensively test this on large sitemaps with lots of events? We really should make sure not to cause any major new issues here. Classic UI should really be in maintenance mode and only severe stuff should be fixed on it.

triller-telekom: I did not check this on a very large sitemap and I was not aware of these constraints. But they surely make sense because scrolling and screen flickering isn't nice.

@lolodomo: What did you experiments say when you tested it?

lolodomo: We are in Home Automation systems, I doubt users have setup with lots of event updates per second. That is not my case.
I think I recently read that for standard groups there is no more any state update. Can you please confirm that point ? If that is true, the change will only concern groups with a function. If that is not true, I should change the code to trigger a reload on state update only for groups having a function.

triller-telekom: Standard groups can still have a [state](https://github.com/eclipse/smarthome/blob/master/bundles/core/org.eclipse.smarthome.core.test/src/test/groovy/org/eclipse/smarthome/core/items/GroupItemTest.groovy#L298). However, if they do not have a function their state is not derived from their member items.

If you change the code to only refresh the site for groups with a function you will miss state changes on those without function.

lolodomo:
If you think I introduced a problem, I put in comments my fix relative to the handling of group state and then I switch to something else. At least this PR will fix the handling of visibility attribute and the update of page title for non group items.
Please let me know.

kaikreuzer:
> I think I recently read that for standard groups there is no more any state update. 

Yes, you might be right - so let's hope that this indeed does not have any "flickering" impact.

Issue #2963, opened 2017-02-03
[Persistence REST API] Java heap space OutOfMemoryError when querying more than 200.000 datapoints


bodiroga:
Hi guys!

I have being testing openHAB2.0 in a Rasperry Pi 3 Model B for the last few days using the Z-Wave binding and things have worked really great, congratulations to the whole team!

I'm using [this](http://www.vesternet.com/z-wave-aeon-labs-smart-switch-6-gen5) device to register the voltage levels at home and, because the device sends the data very frecuently, in a week I have collected more than 200.000 datapoints in the mySQL database. First I tried to graph the data through HABmin, but I couldn't see any value, so I decided to query the REST API directly and have a look at the log files. This is the HTTP GET command I'm using to call the persistence endpoint:

http://openhabianpi.local:8080/rest/persistence/items/fridge_voltage?endtime=2017-01-31T07:49:05.307Z&serviceId=mysql&starttime=2017-01-20T00:49:05.307Z

The result is a HTTP ERROR 500, Problem accessing /rest/persistence/items/fridge_voltage. Reason:

    Request failed.

And here you have the relevant part of the log file:

`2017-02-03 08:37:41.797 [DEBUG] [sql.internal.MysqlPersistenceService] - mySQL: query:SELECT Time, Value FROM Item1 WHERE TIME>'2017-01-20 01:49:05' AND TIME<'2017-01-31 08:49:05' ORDER BY Time ASC
2017-02-03 08:37:48.609 [DEBUG] [sql.internal.MysqlPersistenceService] - mySQL: query returned 231948 rows in 6813ms
2017-02-03 08:38:10.231 [WARN ] [eclipse.jetty.servlet.ServletHandler] - 
javax.servlet.ServletException: org.glassfish.jersey.server.ContainerException: java.lang.OutOfMemoryError: Java heap space
	at org.glassfish.jersey.servlet.WebComponent.serviceImpl(WebComponent.java:489)[155:org.glassfish.jersey.containers.jersey-container-servlet-core:2.22.2]
	at org.glassfish.jersey.servlet.WebComponent.service(WebComponent.java:427)[155:org.glassfish.jersey.containers.jersey-container-servlet-core:2.22.2]
	at org.glassfish.jersey.servlet.ServletContainer.service(ServletContainer.java:388)[155:org.glassfish.jersey.containers.jersey-container-servlet-core:2.22.2]
[...]

PR #3500, opened 2017-05-22
Provide JSON streaming for REST resources

htreu:
This provides two methods of streaming for REST resources:

* Entities passed to the `JSONResponse.createResponse` method are streamed to the client during serialisation. This way large collections wont be materialised in memory.
* Collections from *Provider or *Registry implementations which are exposed through the REST interface my also be streamed to the client. Using `#stream()` on those Collections and passing `new Stream2JSONInputStream(stream)` as entity to the ResponseBuilder will only serialise one entity at a time and stream the result to the client.

Both methods result in reduced memory usage when it comes to large collections.

Fixs #2963.


sjka:
...and Travis goes into lead! 
Looks related on the first glance: https://travis-ci.org/eclipse/smarthome/builds/234835079#L2109 
Can you check what's going on there?

htreu:
Its completely related and not so easy to fix. The tests heavily rely on a Response object being returned by the *Resource services. I will have to rewrite the tests which requires real understanding ;-)
Maybe we should rewrite the tests in Java? wdyt?

maggu2810:
> Maybe we should rewrite the tests in Java? wdyt?

:+1:

htreu:
Ready for review.

bundles/io/org.eclipse.smarthome.io.rest/src/main/java/org/eclipse/smarthome/io/rest/Stream2JSONInputStream.java
```diff
@@ -0,0 +1,100 @@
...
+ * nested collections JSON representation will be fully transformed into memory.
+ *
+ * @author Henning Treu - initial contribution
+ *
+ */
+public class Stream2JSONInputStream extends InputStream {
+
+    private final Iterator<String> iterator;
+
+    private InputStream jsonElementStream;
```
maggu2810: Shouldn't we implement the Closeable interface and ensure the input stream is closed if the close function is called (and ensure that the close function is called by the user of this class)?

Isn't there a resource leak if all the input stream are never closed?

bundles/io/org.eclipse.smarthome.io.rest/src/main/java/org/eclipse/smarthome/io/rest/JSONResponse.java
```diff
@@ -64,79 +102,70 @@ private static ResponseBuilder response(Response.Status status) {
...
-     */
-    public static Response createResponse(Response.Status status, Object entity, String errormessage) {
-        JsonElement ret;
-        if (status.getFamily() == Response.Status.Family.SUCCESSFUL) {
-            // create non-null JsonElement if null!=entity
-            ret = null != entity ? GSON.toJsonTree(entity) : null;
-        } else {
-            ret = createErrorJson(errormessage, status, entity, null);
-        }
+    private Response createResponse(Status status, Object entity) {
```
maggu2810: Don't we need to close `in` and `out` ever?

bundles/io/org.eclipse.smarthome.io.rest/src/main/java/org/eclipse/smarthome/io/rest/Stream2JSONInputStream.java
htreu: Yes, you are probably right. I will address this and update the PR.

bundles/io/org.eclipse.smarthome.io.rest/src/main/java/org/eclipse/smarthome/io/rest/JSONResponse.java
htreu: I will push an update which closes `out` when the writing thread finishes. The `in` stream will be GC'ed once the receiving side did consume the response. We can not actively close it while the receiver is reading from it. It will notice the close of `out` though and free up.

bundles/io/org.eclipse.smarthome.io.rest/src/main/java/org/eclipse/smarthome/io/rest/Stream2JSONInputStream.java
```diff
@@ -0,0 +1,116 @@
...
+    private boolean finished() {
+        return !firstIteratorElement && !iterator.hasNext();
+    }
+
+    private void closeJSONStream() throws IOException {
+        if (jsonElementStream != null) {
+            try {
+                jsonElementStream.close();
+            } catch (IOException e) {
+                throw new IOException("Error closing the underlying JSON stream.", e);
```
maggu2810: Do you really need to encapsulate the `IOException` into another `IOException` only to add a new message?
Shouldn't it be enough to have a look at the stack trace if someone needs to further identify error?

bundles/io/org.eclipse.smarthome.io.rest/src/main/java/org/eclipse/smarthome/io/rest/Stream2JSONInputStream.java
```diff
@@ -0,0 +1,116 @@
...
+        String entity = iterator.hasNext() ? iterator.next() : "";
+
+        String postfix = "";
+        if (!iterator.hasNext()) {
+            postfix = "]";
+        }
+
+        try {
+            closeJSONStream();
+            jsonElementStream = IOUtils.toInputStream(prefix + entity + postfix, StandardCharsets.UTF_8.name());
```
maggu2810: Shouldn't you use a separate try-catch-block for the closeJSONStream method call?

If the close fails there is nothing you can do (and IMHO) this is the reason why you ignore it silently.

If the `IOUtils.toInputStream` fails, then it should not be catched because it is a failure the caller needs to know.

bundles/io/org.eclipse.smarthome.io.rest/src/main/java/org/eclipse/smarthome/io/rest/Stream2JSONInputStream.java
```diff
@@ -0,0 +1,116 @@
...
+            result = jsonElementStream.read();
+        }
+
+        return result;
+    }
+
+    @Override
+    public void close() throws IOException {
+        closeJSONStream();
+        super.close();
```
maggu2810: `InputStream.close()` doesn't contain any implementation so you can drop that call (but it doesn't matter).
But I am pretty sure that the base class close will not get an implementation some time because it would break all previously working subclasses.

bundles/io/org.eclipse.smarthome.io.rest/src/main/java/org/eclipse/smarthome/io/rest/JSONResponse.java
```diff
@@ -64,79 +103,75 @@ private static ResponseBuilder response(Response.Status status) {
...
+                OutputStreamWriter outStreamWriter = new OutputStreamWriter(out);
+                try (JsonWriter jsonWriter = GSON.newJsonWriter(new BufferedWriter(outStreamWriter))) {
+                    if (entity != null) {
+                        GSON.toJson(entity, entity.getClass(), jsonWriter);
+                        jsonWriter.flush();
+                    }
+                } catch (IOException | JsonIOException e) {
+                    logger.error("Error streaming JSON through PipedInpuStream/PipedOutputStream: ", e);
+                } finally {
+                    IOUtils.closeQuietly(outStreamWriter);
```
maggu2810: Any reason not to use try-with-resources for `outStreamWriter`?
You don't need another try-block but can add it to the current one.

```Java
            public void run() {
                try (final OutputStreamWriter outStreamWriter = new OutputStreamWriter(out);
                        final JsonWriter jsonWriter = GSON.newJsonWriter(new BufferedWriter(outStreamWriter))) {
                    if (entity != null) {
                        GSON.toJson(entity, entity.getClass(), jsonWriter);
                        jsonWriter.flush();
                    }
                } catch (IOException | JsonIOException e) {
                    logger.error("Error streaming JSON through PipedInpuStream/PipedOutputStream: ", e);
                }
            }
```

bundles/io/org.eclipse.smarthome.io.rest/src/main/java/org/eclipse/smarthome/io/rest/JSONResponse.java
```diff
@@ -64,79 +103,75 @@ private static ResponseBuilder response(Response.Status status) {
...
+        ResponseBuilder rp = responseBuilder(status);
+        PipedOutputStream out = new PipedOutputStream();
+
+        try {
+            // we will not actively close the PipedInputStream since it is read by the receiving end
+            // and will be GC'ed once the response is consumed.
+            PipedInputStream in = new PipedInputStream(out);
+            rp.entity(in);
+        } catch (IOException e) {
+            throw new RuntimeException(e);
```
maggu2810: Perhaps I am wrong, but in this case out is still not closed. Correct?

Why not using a try-with-resources for out, too?

bundles/io/org.eclipse.smarthome.io.rest/src/main/java/org/eclipse/smarthome/io/rest/JSONResponse.java
```diff
@@ -64,79 +103,75 @@ private static ResponseBuilder response(Response.Status status) {
...
+                        GSON.toJson(entity, entity.getClass(), jsonWriter);
+                        jsonWriter.flush();
+                    }
+                } catch (IOException | JsonIOException e) {
+                    logger.error("Error streaming JSON through PipedInpuStream/PipedOutputStream: ", e);
+                } finally {
+                    IOUtils.closeQuietly(outStreamWriter);
+                }
+            }
+        }).start();
```
maggu2810: There is no `join()` cal ever and the thread is not a daemon thread (correct)?
Shouldn't you call `setDaemon(true)` to ensure an proper clean up?

maggu2810: It seems I mixed here two different contexts.

But I assume a daemon thread would be okay, because the JVM could exit if only such threads are running.

bundles/io/org.eclipse.smarthome.io.rest/src/main/java/org/eclipse/smarthome/io/rest/Stream2JSONInputStream.java
htreu: super call removed.

bundles/io/org.eclipse.smarthome.io.rest/src/main/java/org/eclipse/smarthome/io/rest/Stream2JSONInputStream.java
htreu: cleaned up.

bundles/io/org.eclipse.smarthome.io.rest/src/main/java/org/eclipse/smarthome/io/rest/Stream2JSONInputStream.java
htreu: cleaned up.

bundles/io/org.eclipse.smarthome.io.rest/src/main/java/org/eclipse/smarthome/io/rest/JSONResponse.java
htreu: nice, didnt know of that. done.

bundles/io/org.eclipse.smarthome.io.rest/src/main/java/org/eclipse/smarthome/io/rest/JSONResponse.java
htreu: Unfortunately try-with-resource would close `out` too early once the try-block ends. We need to make sure the writer thread is able to write data to `out` even when the `createResponse` method returns. Thats why we can only close it inside the writer thread.

bundles/io/org.eclipse.smarthome.io.rest/src/main/java/org/eclipse/smarthome/io/rest/JSONResponse.java
htreu: thread daemonized. done.

bundles/io/org.eclipse.smarthome.io.rest/src/main/java/org/eclipse/smarthome/io/rest/JSONResponse.java
htreu: But you are right: `out` would not be closed in case of an exception here. Reason for an exception here:
The `in` and `out` piped streams are already connected, which we will not encounter since both are freshly created.

maggu2810: If the error cannot occur, it would be nice to document it (and the reason why we don't need to close `out` here.

Another option (that is perhaps more easy to read) is to refactor the code to catch the two exit points. Successful exit - do not close `out`, exit by an error (throwable) - close `out`.

WDYT?

htreu: I dont really get what you mean by "catching the two exit points". I will update the PR and document in the code what I just wrote above and you can make another suggestion on the refactoring. I dont see it right now.

maggu2810: What I mean:
You can leave a function:
* if it reaches the end of the last line or by using a return statement (this has been named one of the two exit points above - the normal exit)
* if you (or another function you call) throw an exception (the second of the two exit points -- "by an error")

bundles/io/org.eclipse.smarthome.io.rest/src/main/java/org/eclipse/smarthome/io/rest/JSONResponse.java
htreu: I actually misread the code of BufferedWriter and had the impression it would not close its underlying writer. This is wrong, so the initial (internal) construct with only the JsonWriter being in the try-with-resource is sufficient here.

Updated.

bundles/io/org.eclipse.smarthome.io.rest/src/main/java/org/eclipse/smarthome/io/rest/JSONResponse.java
htreu: My problem is to refactor the method to make it more readable and more clear what happens to the `out` stream in both cases. The nested Thread is what´s in my way i guess, I just stared at the code for too long now ;-)
Maybe you can give an example?

bundles/io/org.eclipse.smarthome.io.rest/src/main/java/org/eclipse/smarthome/io/rest/Stream2JSONInputStream.java
```diff
@@ -0,0 +1,106 @@
...
+ */
+public class Stream2JSONInputStream extends InputStream {
+
+    private final Iterator<String> iterator;
+
+    private InputStream jsonElementStream;
+
+    private boolean firstIteratorElement;
+
+    private Gson GSON = new GsonBuilder().create();
```
kaikreuzer: As this isn't static, shouldn't it be lowercase?

bundles/io/org.eclipse.smarthome.io.rest/src/main/java/org/eclipse/smarthome/io/rest/JSONResponse.java
```diff
@@ -20,29 +27,60 @@
...
-    public static final String JSON_KEY_ERROR_MESSAGE = "message";
-    public static final String JSON_KEY_ERROR = "error";
-    public static final String JSON_KEY_HTTPCODE = "http-code";
 
-    // also dump stacktrace?
-    private final static boolean WITH_STACKTRACE = false;
+    private final Logger logger = LoggerFactory.getLogger(JSONResponse.class);
+
+    private static final JSONResponse INSTANCE = new JSONResponse();
+    private final Gson GSON = new GsonBuilder().setDateFormat(DateTimeType.DATE_PATTERN_WITH_TZ_AND_MS).create();
```
kaikreuzer: lower case?

bundles/io/org.eclipse.smarthome.io.rest/src/main/java/org/eclipse/smarthome/io/rest/JSONResponse.java
```diff
@@ -64,79 +102,79 @@ private static ResponseBuilder response(Response.Status status) {
...
+        } catch (IOException e) {
+            throw new RuntimeException(e);
         }
 
-        // configure response
-        ResponseBuilder rp = response(status);
-        if (null != ret) {
-            rp = rp.entity(GSON.toJson(ret));
-        }
+        Thread writerThread = new Thread(new Runnable() {
```
kaikreuzer: any reason not to use lambda syntax here?

htreu: no :-) expect an update `soon`.

kaikreuzer:
Very cool feature, thanks a lot!

PR #3501, opened 2017-05-22
[WIP][astro] Removed Quartz dependency and added ESH scheduler

### LINKED ISSUE #3416: Remove Quartz dependency from Astro binding
We are trying to make ESH fully Java8 CP2 compatible.
One part of this effort is to replace the use of the Quartz library (which requires a full JRE) by the new ESH scheduler implementation.
The Astro binding currently still relies on Quartz, so that it should be refactored.

cweitkamp:
- removed org.quartz dependencies
- added org.eclipse.smarthome.core.scheduler dependencies
- refactoring to simplify the code and ensure the business logic of the binding

Closes #3416


cweitkamp:
Just for everyone to notice: This is a first try. I think there are some lines of code which can be simplified. Additionally there is an open point (see [here](https://github.com/cweitkamp/smarthome/commit/9091bf20936727e0277dcf9318f0ae704c26f21f#commitcomment-22222972)) to discuss. I am currently running some tests in my test environment to proof funtionality.

cweitkamp:
@kaikreuzer @kgoderis I currently receive the following warnings if I try to run this binding.
```
2017-05-22 22:26:38.082 [WARN ] [cheduler.ExpressionThreadPoolManager] - Thread pool 'astro' rejected execution of class java.util.concurrent.ScheduledThreadPoolExecutor$ScheduledFutureTask
2017-05-22 22:26:38.119 [WARN ] [cheduler.ExpressionThreadPoolManager] - Thread pool 'astro' rejected execution of class java.util.concurrent.ScheduledThreadPoolExecutor$ScheduledFutureTask
```
If I understand [this (ThreadPoolManager.java#L37)](https://github.com/eclipse/smarthome/blob/master/bundles/core/org.eclipse.smarthome.core/src/main/java/org/eclipse/smarthome/core/common/ThreadPoolManager.java#L37) and [this ExpressionThreadPoolManager.java#L85](https://github.com/eclipse/smarthome/blob/master/bundles/core/org.eclipse.smarthome.core/src/main/java/org/eclipse/smarthome/core/scheduler/ExpressionThreadPoolManager.java#L85) lines of code correctly the pool size is bounded an I have to configure a `poolSize` for my new pool named `astro`. Am I right?

I found similar configuration in the `smarthome.cfg` file in this repo. Where should I put the configuration in my openHAB2 environment? In the `runtime.cfg` file?

Thanks for help.

extensions/binding/org.eclipse.smarthome.binding.astro/src/main/java/org/eclipse/smarthome/binding/astro/handler/AstroThingHandler.java
```diff
@@ -185,52 +179,46 @@ public void run() {
...
+                                dateExpression = new DateExpression(
+                                        AbstractBaseJob.ISO8601_FORMAT.format(start.getTime()));
+                                getScheduler().schedule(new PositionalJob(jobDataMap), dateExpression);
                                 logger.info("Scheduled astro job-positional with interval of {} seconds for thing {}",
                                         thingConfig.getInterval(), thingUid);
                             }
                         }
                     }
-                } catch (SchedulerException ex) {
+                } catch (Exception ex) {
```
maggu2810: Please don't catch `Exception`

extensions/binding/org.eclipse.smarthome.binding.astro/src/main/java/org/eclipse/smarthome/binding/astro/handler/AstroThingHandler.java
```diff
@@ -243,17 +231,12 @@ public void run() {
...
-                    String thingUid = getThing().getUID().toString();
-                    for (JobKey jobKey : quartzScheduler.getJobKeys(jobGroupEquals(thingUid))) {
-                        logger.debug("Deleting astro {} for thing '{}'", jobKey.getName(), thingUid);
-                        quartzScheduler.deleteJob(jobKey);
-                    }
-                } catch (SchedulerException ex) {
-                    logger.error("{}", ex.getMessage(), ex);
-                }
-            }
+            // String thingUid = getThing().getUID().toString();
```
sjka: Hoe about something along this (not really tested though...):

```java
for (Runnable runnable : getScheduler().getQueue()) {
  if (runnable instanceof AbstractBaseJob) {
    if (thingUid.equals(((AbstractBaseJob) runnable).getJobDataMap().get(AbstractBaseJob.KEY_THING_UID))) {
      getScheduler().remove(runnable);
    }
  }
}
```

You'd have to extend `AbstractBaseJob` to expose a `Collections.unmodifiableMap(jobDataMap)`, but I don't see a reason what it shouldn't.

extensions/binding/org.eclipse.smarthome.binding.astro/src/main/java/org/eclipse/smarthome/binding/astro/internal/job/AbstractBaseJob.java
```diff
@@ -8,41 +8,47 @@
...
 /**
  * Baseclass for all jobs with common methods.
  *
  * @author Gerhard Riegler - Initial contribution
+ * @author Christoph Weitkamp - Removed Quartz dependency
  */
-public abstract class AbstractBaseJob implements Job {
+public abstract class AbstractBaseJob implements Runnable {
     private final Logger logger = LoggerFactory.getLogger(AbstractBaseJob.class);
+    public static final SimpleDateFormat ISO8601_FORMAT = new SimpleDateFormat("yyyy-MM-dd'T'HH:mm:ssZ");
```
sjka: Attention: SDF is NOT thread-safe!

itheiss: @sjka Does it make sense to tackle the NOT thread-safe SDF problem in a more global manner?

As the ThreadLocal approach seams the best approach performance wise an own eclipse smarthome implementation makes the most sense to me. WDYT ?

extensions/binding/org.eclipse.smarthome.binding.astro/src/main/java/org/eclipse/smarthome/binding/astro/handler/AstroThingHandler.java
maggu2810: Without having an IDE open, does the scheduler return a view of the queue or a copy?
I assume the queue could be changed by the scheduler while iteration.

extensions/binding/org.eclipse.smarthome.binding.astro/src/main/java/org/eclipse/smarthome/binding/astro/handler/AstroThingHandler.java
cweitkamp: @maggu2810 Thanks for your review. I can change the code to catch each type of exception here. I have to catch six different types of exeptions here. That's the case why I use `Exception` in the first place.

extensions/binding/org.eclipse.smarthome.binding.astro/src/main/java/org/eclipse/smarthome/binding/astro/internal/job/AbstractBaseJob.java
sjka: I think the most global solution would be using [DateTimeFormatter](https://docs.oracle.com/javase/8/docs/api/java/time/format/DateTimeFormatter.html) instead, which was introduced with Java 8.

sjka:
Just a quick heads-up: Due to some bad timing we got another similar PR here #3503. It in addition contains some more cleanups. Ideally we somehow get the best of both 😉

extensions/binding/org.eclipse.smarthome.binding.astro/src/main/java/org/eclipse/smarthome/binding/astro/internal/job/AbstractBaseJob.java
cweitkamp: I maybe observed an issue with the `DateExpression` from ESH scheduler. Every schedule is executed more than once. I am going to file an issue for that if I can confirm my guess.

Currently I want to replace `DateExpression` completely with `CronExpression`.

cweitkamp:
@sjka I see. I am really sure we can find a way to combine the besst of both implementations.

sjka:
@cweitkamp Just for clarification - and to prevent duplicate efforts: Will you be working on merging #3503 into yours? 

Technically you can get it by using `git fetch origin pull/3503/head` and then you should be able to cherry-pick it into your branch. You'd only have to include a `Also-by: Amit Kumar Mondal <admin@amitinside.com>` line in your commits then right before your `Signed-off-by: [...]` line.

cweitkamp:
@sjka Thanks for your help. I will try to do like you recommended.

cweitkamp:
@sjka I looked at the changes of #3503. There are a lot of very good improvements beside some changes for which a had no solution yet. I don't really know where to start the merging and if there are parts of the code which we can use from my code, so I decides to give up this PR and let take @amitjoy the lead for this. Nice work.

kaikreuzer:
@cweitkamp Thanks for looking into #3503 and checking the differences. And thanks for suggesting to go ahead with #3503 instead of your PR (which proves your expertise as usually the own code is always considered to be superior to other's ;-)). 

So let's close this PR then.

PR #3507, opened 2017-05-23
Refactored CoreItemFactory

AlexanderKostadinov28:
### org.eclipse.smarthome.core.library.CoreItemFactory

> - Re-ordered all item type constants in alphabetical order
> - Changed `createItem` method to make use of a switch statement (using alphabetical order)
> - Fixed broken {@inheritDoc} links

### DIFF
--- a/bundles/core/org.eclipse.smarthome.core/src/main/java/org/eclipse/smarthome/core/library/CoreItemFactory.java
+++ b/bundles/core/org.eclipse.smarthome.core/src/main/java/org/eclipse/smarthome/core/library/CoreItemFactory.java
@@ -27,69 +27,64 @@ import org.eclipse.smarthome.core.library.items.SwitchItem;
  *
  * @author Thomas.Eichstaedt-Engelen
  * @author Kai Kreuzer
+ * @author Alexander Kostadinov
  */
 public class CoreItemFactory implements ItemFactory {
 
-    public static final String SWITCH = "Switch";
-    public static final String ROLLERSHUTTER = "Rollershutter";
-    public static final String CONTACT = "Contact";
-    public static final String STRING = "String";
-    public static final String NUMBER = "Number";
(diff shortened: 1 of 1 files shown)


bundles/core/org.eclipse.smarthome.core/src/main/java/org/eclipse/smarthome/core/library/CoreItemFactory.java
```diff
@@ -27,69 +27,62 @@
...
-            return new PlayerItem(itemName);
-        }
-        if (itemTypeName.equals(LOCATION)) {
-            return new LocationItem(itemName);
-        }
-        if (itemTypeName.equals(CALL)) {
-            return new CallItem(itemName);
-        } else {
-            return null;
+        if (itemTypeName != null) {
```
sjka: Good you thought about it! 
Please just invert this statement in order to reduce the nesting level.

AlexanderKostadinov28: Did you mean just changing the statement to 
`if (null != itemTypename)` or I got it wrong? Or you have the idea to do
```java
if (itemTypeName != null) {
    return null;
} 
```
before the switch and add a default `return null` to the switch?

sjka: The latter. Return directly before the switch in case it is null.
Sorry, that was not clear enough from my side.

AlexanderKostadinov28: Fixed. :)

sjka:
Thanks!

PR #3508, opened 2017-05-23
[Scheduler] Dropped usage of SimpleDateFormat for trace logging

### LINKED ISSUE #3505: [Scheduler] ArrayIndexOutOfBoundsException in scheduler
When trying it out #3501, I ran into this here:

```
14:08:37.864 ERROR o.e.s.b.a.h.AstroThingHandler[:234] - 853
java.lang.ArrayIndexOutOfBoundsException: 853
	at sun.util.calendar.BaseCalendar.getCalendarDateFromFixedDate(BaseCalendar.java:453)
	at java.util.GregorianCalendar.computeFields(GregorianCalendar.java:2397)
	at java.util.GregorianCalendar.computeFields(GregorianCalendar.java:2312)
	at java.util.Calendar.setTimeInMillis(Calendar.java:1804)
	at java.util.Calendar.setTime(Calendar.java:1770)
	at java.text.SimpleDateFormat.format(SimpleDateFormat.java:943)
	at java.text.SimpleDateFormat.format(SimpleDateFormat.java:936)
	at java.text.DateFormat.format(DateFormat.java:345)
	at org.eclipse.smarthome.core.scheduler.AbstractExpression.applyExpressionParts(AbstractExpression.java:202)
	at org.eclipse.smarthome.core.scheduler.AbstractExpression.parseExpression(AbstractExpression.java:176)
	at org.eclipse.smarthome.core.scheduler.AbstractExpression.parseExpression(AbstractExpression.java:144)
```

### LINKED ISSUE #3506: [Scheduler] Excessive use of SimpleDateFormat
When looking into [AbstractExpression.java](https://github.com/eclipse/smarthome/blob/3ae5d7b1b284f2bde6199bfe8dad950fd38cae39/bundles/core/org.eclipse.smarthome.core/src/main/java/org/eclipse/smarthome/core/scheduler/AbstractExpression.java), I noticed it is using `SimpleDateFormat` excessively, mainly for logging on trace level - regardless of whether logging is enabled for this level or not. This especially hurts as it causes failures (see #3505) and massively slows down the scheduler. 

Please be aware that it will executed all the time in constructs like this, no matter if something gets logged or not:

```java
for (Date aDate : getCandidates()) {
    logger.trace("Final candidate {} is {}", getCandidates().indexOf(aDate), sdf.format(aDate));
}
```

sjka:
...as it slows down the scheduler and was used in a dangeriously
wrt thread-safety.

fixes #3505
fixes #3506

### DIFF
--- a/bundles/core/org.eclipse.smarthome.core/src/main/java/org/eclipse/smarthome/core/scheduler/ExpressionThreadPoolManager.java
+++ b/bundles/core/org.eclipse.smarthome.core/src/main/java/org/eclipse/smarthome/core/scheduler/ExpressionThreadPoolManager.java
@@ -7,7 +7,6 @@
  */
 package org.eclipse.smarthome.core.scheduler;
 
-import java.text.SimpleDateFormat;
 import java.util.ArrayList;
 import java.util.Collections;
 import java.util.Date;
@@ -36,10 +35,6 @@ import org.slf4j.LoggerFactory;
  */
 public class ExpressionThreadPoolManager extends ThreadPoolManager {
 
-    private final static Logger logger = LoggerFactory.getLogger(ExpressionThreadPoolManager.class);
-
-    static SimpleDateFormat sdf = new SimpleDateFormat("yyyy-MM-dd HH:mm:ss.SSS");
-
     /**
      * Returns an instance of an expression-driven scheduled thread pool service. If it is the first request for the
      * given pool name, the instance is newly created.
@@ -47,7 +42,7 @@ public class ExpressionThreadPoolManager extends ThreadPoolManager {
      * @param poolName a short name used to identify the pool, e.g. "discovery"
      * @return an instance to use
      */
-    static public ExpressionThreadPoolExecutor getExpressionScheduledPool(String poolName) {
+    public static ExpressionThreadPoolExecutor getExpressionScheduledPool(String poolName) {
         ExecutorService pool = pools.get(poolName);
         if (pool == null) {
             synchronized (pools) {
@@ -59,8 +54,8 @@ public class ExpressionThreadPoolManager extends ThreadPoolManager {
                     ((ThreadPoolExecutor) pool).setKeepAliveTime(THREAD_TIMEOUT, TimeUnit.SECONDS);
                     ((ThreadPoolExecutor) pool).allowCoreThreadTimeOut(true);
                     pools.put(poolName, pool);
-                    logger.debug("Created an expression-drive scheduled thread pool '{}' of size {}",
-                            new Object[] { poolName, cfg });
+                    LoggerFactory.getLogger(ExpressionThreadPoolManager.class)
+                            .debug("Created an expression-drive scheduled thread pool '{}' of size {}", poolName, cfg);
                 }
             }
         }
@@ -73,6 +68,8 @@ public class ExpressionThreadPoolManager extends ThreadPoolManager {
 
     public static class ExpressionThreadPoolExecutor extends ScheduledThreadPoolExecutor {
 
+        private final Logger logger = LoggerFactory.getLogger(ExpressionThreadPoolExecutor.class);
+
         private Map<Expression, Runnable> scheduled = new ConcurrentHashMap<>();
         private Map<Runnable, ArrayList<Future<?>>> futures = Collections
                 .synchronizedMap(new HashMap<Runnable, ArrayList<Future<?>>>());
@@ -82,12 +79,14 @@ public class ExpressionThreadPoolManager extends ThreadPoolManager {
 
         public ExpressionThreadPoolExecutor(final String poolName, int corePoolSize) {
             this(poolName, corePoolSize, new NamedThreadFactory(poolName), new ThreadPoolExecutor.DiscardPolicy() {
+
+                private final Logger logger = LoggerFactory.getLogger(ThreadPoolExecutor.DiscardPolicy.class);
+
                 // The pool is bounded and rejections will happen during shutdown
                 @Override
                 public void rejectedExecution(Runnable runnable, ThreadPoolExecutor threadPoolExecutor) {
                     // Log and discard
-                    logger.warn("Thread pool '{}' rejected execution of {}",
-                            new Object[] { poolName, runnable.getClass() });
+                    logger.debug("Thread pool '{}' rejected execution of {}", poolName, runnable.getClass());
                     super.rejectedExecution(runnable, threadPoolExecutor);
                 }
             });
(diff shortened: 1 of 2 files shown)


bundles/core/org.eclipse.smarthome.core/src/main/java/org/eclipse/smarthome/core/scheduler/ExpressionThreadPoolManager.java
```diff
@@ -38,8 +37,6 @@
 
     private final static Logger logger = LoggerFactory.getLogger(ExpressionThreadPoolManager.class);
```
htreu: I know its not part of the PR but maybe you can change the logger to `privat final` right away.

sjka: I feared this would catch sb's eye 😉 
No, it's not _that_ easy, but will do so.

maggu2810: Then let's do it in another PR as this change is not related at all. Okay?

sjka:
done

bundles/core/org.eclipse.smarthome.core/src/main/java/org/eclipse/smarthome/core/scheduler/ExpressionThreadPoolManager.java
```diff
@@ -86,7 +84,8 @@ public ExpressionThreadPoolExecutor(final String poolName, int corePoolSize) {
                 @Override
                 public void rejectedExecution(Runnable runnable, ThreadPoolExecutor threadPoolExecutor) {
                     // Log and discard
-                    logger.warn("Thread pool '{}' rejected execution of {}",
+                    LoggerFactory.getLogger(ExpressionThreadPoolExecutor.class).warn(
```
maggu2810: As we create here a implementation of the ThreadPoolExecutor.DiscardPolicy class, shouldn't we add a logger to that implementation using the class it implements?

bundles/core/org.eclipse.smarthome.core/src/main/java/org/eclipse/smarthome/core/scheduler/ExpressionThreadPoolManager.java
```diff
@@ -86,7 +84,8 @@ public ExpressionThreadPoolExecutor(final String poolName, int corePoolSize) {
...
+                            "Thread pool '{}' rejected execution of {}",
                             new Object[] { poolName, runnable.getClass() });
```
maggu2810: Not introduced by your changes, but do we need to create an array of the two logger arguments?

bundles/core/org.eclipse.smarthome.core/src/main/java/org/eclipse/smarthome/core/scheduler/ExpressionThreadPoolManager.java
```diff
@@ -59,7 +54,8 @@ static public ExpressionThreadPoolExecutor getExpressionScheduledPool(String poo
                     ((ThreadPoolExecutor) pool).setKeepAliveTime(THREAD_TIMEOUT, TimeUnit.SECONDS);
                     ((ThreadPoolExecutor) pool).allowCoreThreadTimeOut(true);
                     pools.put(poolName, pool);
-                    logger.debug("Created an expression-drive scheduled thread pool '{}' of size {}",
+                    LoggerFactory.getLogger(ExpressionThreadPoolManager.class).debug(
```
maggu2810: Also not really part of this PR: If and only if you do another change again, WDYT about changing `static public` to `public static` to get the common keyword order?

bundles/core/org.eclipse.smarthome.core/src/main/java/org/eclipse/smarthome/core/scheduler/ExpressionThreadPoolManager.java
```diff
@@ -59,7 +54,8 @@ static public ExpressionThreadPoolExecutor getExpressionScheduledPool(String poo
...
+                            "Created an expression-drive scheduled thread pool '{}' of size {}",
                             new Object[] { poolName, cfg });
```
maggu2810: Also not really part of this PR: Why using an array for the arguments?

bundles/core/org.eclipse.smarthome.core/src/main/java/org/eclipse/smarthome/core/scheduler/ExpressionThreadPoolManager.java
sjka: done - plus I reduced the severity to DEBUG as the comment suggests this actually is expected to happen during shutdown.

bundles/core/org.eclipse.smarthome.core/src/main/java/org/eclipse/smarthome/core/scheduler/ExpressionThreadPoolManager.java
```diff
@@ -36,18 +35,14 @@
  */
 public class ExpressionThreadPoolManager extends ThreadPoolManager {
 
-    private final static Logger logger = LoggerFactory.getLogger(ExpressionThreadPoolManager.class);
```
amitjoy: Instead of using logger for every inner class, you could also do this:

```java
private final Logger logger = LoggerFactory.getLogger(MethodHandles.lookup().lookupClass());
```
This will be automatically resolved for every class instance that uses it.

martinvw: Are you really sure that this will give a different logger class when calling the same instance from the different locations, I cannot see how that would work out. The lookup is done only once when initializing.

amitjoy: I actually use this in inheritance pattern in which the logger is mentioned in interface or base (abstract) class so that the logger gets resolved for every child class instances. I completely overlooked in this scenario that it will be resolved only for the top-level class during instantiation and the same logger instance will be used by the inner classes.

martinvw: :-)

PR #3512, opened 2017-05-24
Revert JsonPath to 2.0.0, add source jars

htreu:
* Revert the JsonPath version to 2.0.0 until 2.1.0 gets approved, see discussion here: #3422 
* Add source jars for 3rd party libs.


maggu2810:
Thanks for this two changes. But one question:

Shouldn't we try to use the VCS as intended and split unrelated changes into separate commits?
Perhaps there is one, but I don't see the relationship between "add sources jars" and "downgrade JsonPath".
Or don't we care about the separation in this third-party gh-pages branch at all?

sjka:
The relation is a bit more subtle: the pom.xml previously was updated to jsonpath-2.2.0, but the repository never was updated. Neither was the target platform. So as soon as @htreu built the new repository, it would broken the build. That's why it had to be in a single PR.

maggu2810:
@sjka Okay, we have the upgrade / downgrade of the jsonpath artifact and the repository regeneration...

But how is this related to the "add `<source>true</source>` to the pom for every artifact to fill the repository with the sources" (I agree with that change, as we should provide the source artifact, btu I don't see the reason to mix it into one commit).

sjka:
correct, ideally jsonpath should have been reverted first. You are right.

htreu:
thanks for the discussion, I will keep this in mind the next time :-)

PR #3514, opened 2017-05-24
generate revision info into bundles

sjka:
This is just a suggestion to include the git commit ID in our bundles' binaries. 

The used maven plugin would generate a "git.properties" file for each bundle, containing the following information (I reduced it to only the most relevant entries, the plugin actually could generate many more it we let it...):

```
#Generated by Git-Commit-Id-Plugin
#Tue May 23 12:54:27 CEST 2017
git.dirty=true
git.commit.time=23.05.2017 @ 12\:53\:27 MESZ
git.commit.id.abbrev=259b030
git.commit.id=259b03051341aa9ccbfb2597228b61b192e07eb2
git.build.time=23.05.2017 @ 12\:54\:22 MESZ

```

This would give us some additional traceability, as build numbers etc. are nice for users (like e.g. in https://github.com/openhab/openhab-core/pull/149), but technically are unreliable as they are only usable as long as the build is retained on the build server. 

What do you think?

### DIFF
--- a/pom.xml
+++ b/pom.xml
@@ -141,6 +141,10 @@
       <plugin>
         <artifactId>maven-compiler-plugin</artifactId>
       </plugin>
+      <plugin>
+        <groupId>pl.project13.maven</groupId>
+        <artifactId>git-commit-id-plugin</artifactId>
+      </plugin>      
     </plugins>
 
     <pluginManagement>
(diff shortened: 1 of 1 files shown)


kaikreuzer:
I think it will require us to create a CQ...
So far I was happy with the build qualifier as this is directly visible on the bundles. The commit id imho won't be really helpful for most non-tech users.

maggu2810:
> I think it will require us to create a CQ...

Do you need a CQ for every tool that is involved in the build process?
Also if the output of the tool is only a properties file that is included in the artifact?

sjka:
Yes: https://wiki.eclipse.org/Development_Resources/IP/Test_and_Build_Dependencies

Anway, _if_ this is of interest, I will even volunteer to create the CQ.

hakan42:
Actually, I find this information pretty interesting, especially if the addons could have access to it too.

Perhaps if that information could be acessed via the karaf console and / or the Paper UI so we could "know" what version of a binding (which the user might have downloaded from ther Marketplace) is being used...

kaikreuzer:
@hakan42 The version of the binding is already included - this discussion is only about information for developers who have a clue what a commit id is and how to look it up.

htreu:
Just to clarify what the commit id is in this context: does it mean the specific bundle was last modified in some way in the commit given? Or does every bundle from a single maven build hold the same commit id (the HEAD pointer)?

sjka:
To my knowledge it is just the HEAD pointer.

maggu2810:
I don't know if this is really necessary.

If bundles are used somewhere I prefer versioned ones at all.
If you are using this to track the bundles at development you still don't know if the working copy contains local non-committed changes.

But I don't see any real "cons" against this information.
The one that does not need it, can ignore it (I assume the size of this small property file does not matter).
If it is useful for at least some peoples, let's add it.

sjka:
> you still don't know if the working copy contains local non-committed changes

Oh, you do: 

```
git.dirty=true
```

> If it is useful for at least some peoples, let's add it.

So far nobody really raised an interest on this (apart from the one that Kai correctly pointed back to the bundle version). As it looks like, I'd actually withdraw this as it currently looks like.

PR #3411, opened 2017-05-09
A group without a function cannot have a state

triller-telekom:
Fixes #1738

Signed-off-by: Stefan Triller <stefan.triller@telekom.de>

PR #3521, opened 2017-05-25
null is not an accepted type so convert it to an accepted type

### LINKED ISSUE #3520: NPEs when initializing items
The demo setup of openHAB now results in such exceptions in the log:

```
2017-05-25 17:28:58.990 [ERROR] [ore.internal.events.OSGiEventManager] - Dispatching/filtering event for subscriber 'org.eclipse.smarthome.core.events.EventSubscriber' failed: java.lang.IllegalArgumentException: The argument 'state' must not be null or empty.
java.util.concurrent.ExecutionException: java.lang.IllegalArgumentException: The argument 'state' must not be null or empty.
        at java.util.concurrent.FutureTask.report(FutureTask.java:122)[:1.8.0_112]
        at java.util.concurrent.FutureTask.get(FutureTask.java:206)[:1.8.0_112]
        at org.eclipse.smarthome.core.common.SafeMethodCaller.callAsynchronous(SafeMethodCaller.java:194)[99:org.eclipse.smarthome.core:0.9.0.201705251309]
        at org.eclipse.smarthome.core.common.SafeMethodCaller.call(SafeMethodCaller.java:83)[99:org.eclipse.smarthome.core:0.9.0.201705251309]
        at org.eclipse.smarthome.core.common.SafeMethodCaller.call(SafeMethodCaller.java:67)[99:org.eclipse.smarthome.core:0.9.0.201705251309]
        at org.eclipse.smarthome.core.internal.events.OSGiEventManager.dispatchESHEvent(OSGiEventManager.java:188)[99:org.eclipse.smarthome.core:0.9.0.201705251309]
        at org.eclipse.smarthome.core.internal.events.OSGiEventManager.handleEvent(OSGiEventManager.java:163)[99:org.eclipse.smarthome.core:0.9.0.201705251309]
        at org.eclipse.smarthome.core.internal.events.OSGiEventManager.handleEvent(OSGiEventManager.java:146)[99:org.eclipse.smarthome.core:0.9.0.201705251309]
        at org.apache.felix.eventadmin.impl.handler.EventHandlerProxy.sendEvent(EventHandlerProxy.java:415)[6:org.apache.karaf.services.eventadmin:4.0.8]
        at org.apache.felix.eventadmin.impl.tasks.HandlerTask.run(HandlerTask.java:90)[6:org.apache.karaf.services.eventadmin:4.0.8]
        at java.util.concurrent.Executors$RunnableAdapter.call(Executors.java:511)[:1.8.0_112]
        at java.util.concurrent.FutureTask.run(FutureTask.java:266)[:1.8.0_112]
        at java.util.concurrent.ThreadPoolExecutor.runWorker(ThreadPoolExecutor.java:1142)[:1.8.0_112]
        at java.util.concurrent.ThreadPoolExecutor$Worker.run(ThreadPoolExecutor.java:617)[:1.8.0_112]
        at java.lang.Thread.run(Thread.java:745)[:1.8.0_112]
Caused by: java.lang.IllegalArgumentException: The argument 'state' must not be null or empty.
        at com.google.common.base.Preconditions.checkArgument(Preconditions.java:122)[12:com.google.guava:18.0.0]
        at org.eclipse.smarthome.core.items.events.ItemEventFactory.assertValidArguments(ItemEventFactory.java:370)[99:org.eclipse.smarthome.core:0.9.0.201705251309]
        at org.eclipse.smarthome.core.items.events.ItemEventFactory.createStateChangedEvent(ItemEventFactory.java:272)[99:org.eclipse.smarthome.core:0.9.0.201705251309]
        at org.eclipse.smarthome.core.items.GenericItem.sendStateChangedEvent(GenericItem.java:215)[99:org.eclipse.smarthome.core:0.9.0.201705251309]
        at org.eclipse.smarthome.core.items.GenericItem.applyState(GenericItem.java:209)[99:org.eclipse.smarthome.core:0.9.0.201705251309]
        at org.eclipse.smarthome.core.library.items.DimmerItem.setState(DimmerItem.java:74)[99:org.eclipse.smarthome.core:0.9.0.201705251309]
        at org.eclipse.smarthome.core.internal.items.ItemUpdater.receiveUpdate(ItemUpdater.java:70)[99:org.eclipse.smarthome.core:0.9.0.201705251309]
        at org.eclipse.smarthome.core.items.events.AbstractItemEventSubscriber.receive(AbstractItemEventSubscriber.java:44)[99:org.eclipse.smarthome.core:0.9.0.201705251309]
        at org.eclipse.smarthome.core.internal.events.OSGiEventManager$1.call(OSGiEventManager.java:192)[99:org.eclipse.smarthome.core:0.9.0.201705251309]
        at org.eclipse.smarthome.core.internal.events.OSGiEventManager$1.call(OSGiEventManager.java:1)[99:org.eclipse.smarthome.core:0.9.0.201705251309]
        at org.eclipse.smarthome.core.common.SafeMethodCaller$CallableWrapper.call(SafeMethodCaller.java:181)[99:org.eclipse.smarthome.core:0.9.0.201705251309]
        ... 4 more
2017-05-25 17:28:59.064 [ERROR] [ore.internal.events.OSGiEventManager] - Dispatching/filtering event for subscriber 'org.eclipse.smarthome.core.events.EventSubscriber' failed: java.lang.NullPointerException
java.util.concurrent.ExecutionException: java.lang.NullPointerException
        at java.util.concurrent.FutureTask.report(FutureTask.java:122)[:1.8.0_112]
        at java.util.concurrent.FutureTask.get(FutureTask.java:206)[:1.8.0_112]
        at org.eclipse.smarthome.core.common.SafeMethodCaller.callAsynchronous(SafeMethodCaller.java:194)[99:org.eclipse.smarthome.core:0.9.0.201705251309]
        at org.eclipse.smarthome.core.common.SafeMethodCaller.call(SafeMethodCaller.java:83)[99:org.eclipse.smarthome.core:0.9.0.201705251309]
        at org.eclipse.smarthome.core.common.SafeMethodCaller.call(SafeMethodCaller.java:67)[99:org.eclipse.smarthome.core:0.9.0.201705251309]
        at org.eclipse.smarthome.core.internal.events.OSGiEventManager.dispatchESHEvent(OSGiEventManager.java:188)[99:org.eclipse.smarthome.core:0.9.0.201705251309]
        at org.eclipse.smarthome.core.internal.events.OSGiEventManager.handleEvent(OSGiEventManager.java:163)[99:org.eclipse.smarthome.core:0.9.0.201705251309]
        at org.eclipse.smarthome.core.internal.events.OSGiEventManager.handleEvent(OSGiEventManager.java:146)[99:org.eclipse.smarthome.core:0.9.0.201705251309]
        at org.apache.felix.eventadmin.impl.handler.EventHandlerProxy.sendEvent(EventHandlerProxy.java:415)[6:org.apache.karaf.services.eventadmin:4.0.8]
        at org.apache.felix.eventadmin.impl.tasks.HandlerTask.run(HandlerTask.java:90)[6:org.apache.karaf.services.eventadmin:4.0.8]
        at java.util.concurrent.Executors$RunnableAdapter.call(Executors.java:511)[:1.8.0_112]
        at java.util.concurrent.FutureTask.run(FutureTask.java:266)[:1.8.0_112]
        at java.util.concurrent.ThreadPoolExecutor.runWorker(ThreadPoolExecutor.java:1142)[:1.8.0_112]
        at java.util.concurrent.ThreadPoolExecutor$Worker.run(ThreadPoolExecutor.java:617)[:1.8.0_112]
        at java.lang.Thread.run(Thread.java:745)[:1.8.0_112]
Caused by: java.lang.NullPointerException
        at org.eclipse.smarthome.core.items.GenericItem.applyState(GenericItem.java:208)[99:org.eclipse.smarthome.core:0.9.0.201705251309]
        at org.eclipse.smarthome.core.library.items.DimmerItem.setState(DimmerItem.java:74)[99:org.eclipse.smarthome.core:0.9.0.201705251309]
        at org.eclipse.smarthome.core.internal.items.ItemUpdater.receiveUpdate(ItemUpdater.java:70)[99:org.eclipse.smarthome.core:0.9.0.201705251309]
        at org.eclipse.smarthome.core.items.events.AbstractItemEventSubscriber.receive(AbstractItemEventSubscriber.java:44)[99:org.eclipse.smarthome.core:0.9.0.201705251309]
        at org.eclipse.smarthome.core.internal.events.OSGiEventManager$1.call(OSGiEventManager.java:192)[99:org.eclipse.smarthome.core:0.9.0.201705251309]
        at org.eclipse.smarthome.core.internal.events.OSGiEventManager$1.call(OSGiEventManager.java:1)[99:org.eclipse.smarthome.core:0.9.0.201705251309]
        at org.eclipse.smarthome.core.common.SafeMethodCaller$CallableWrapper.call(SafeMethodCaller.java:181)[99:org.eclipse.smarthome.core:0.9.0.201705251309]
        ... 4 more
```

See also https://community.openhab.org/t/java-lang-nullpointerexception-every-few-seconds/28944

martinvw:
null is not an accepted type so convert it to an accepted type

Overcomes NPE's in a places where the state is broadcasted to

Fixes #3520

### DIFF
--- a/bundles/core/org.eclipse.smarthome.core/src/main/java/org/eclipse/smarthome/core/items/ItemUtil.java
+++ b/bundles/core/org.eclipse.smarthome.core/src/main/java/org/eclipse/smarthome/core/items/ItemUtil.java
@@ -7,7 +7,10 @@
  */
 package org.eclipse.smarthome.core.items;
 
+import org.apache.commons.lang.StringUtils;
 import org.eclipse.smarthome.core.types.State;
+import org.eclipse.smarthome.core.types.UnDefType;
+import org.slf4j.Logger;
 import org.slf4j.LoggerFactory;
 
 /**
@@ -17,8 +20,10 @@ import org.slf4j.LoggerFactory;
  *
  * @author Michael Grammling - Initial contribution and API
  * @author Simon Kaufmann - added type conversion
+ * @author Martin van Wingerden - when converting types convert null to UnDefType.NULL
  */
 public class ItemUtil {
+    private static final Logger LOGGER = LoggerFactory.getLogger(ItemUtil.class);
 
     /**
      * The constructor is private.
@@ -44,11 +49,7 @@ public class ItemUtil {
      * @return true if the specified name is a valid item name, otherwise false
      */
     public static boolean isValidItemName(final String itemName) {
-        if (itemName != null && !itemName.isEmpty()) {
-            return itemName.matches("[a-zA-Z0-9_]*");
-        }
-
-        return false;
+        return StringUtils.isNotEmpty(itemName) && itemName.matches("[a-zA-Z0-9_]*");
     }
 
     /**
@@ -75,14 +76,18 @@ public class ItemUtil {
         }
     }
 
-    public static State convertToAcceptedState(final State state, final Item item) {
+    public static State convertToAcceptedState(State state, Item item) {
+        if (state == null) {
+            LOGGER.error("A conversion of null was requested", new NullPointerException("state should not be null"));
+            return UnDefType.NULL;
+        }
+
         if (item != null && !isAccepted(item, state)) {
             for (Class<? extends State> acceptedType : item.getAcceptedDataTypes()) {
                 State convertedState = state.as(acceptedType);
                 if (convertedState != null) {
-                    LoggerFactory.getLogger(ItemUtil.class).debug("Converting {} '{}' to {} '{}' for item '{}'",
-                            state.getClass().getSimpleName(), state.toString(),
-                            convertedState.getClass().getSimpleName(), convertedState.toString(), item.getName());
+                    LOGGER.debug("Converting {} '{}' to {} '{}' for item '{}'", state.getClass().getSimpleName(), state,
+                            convertedState.getClass().getSimpleName(), convertedState, item.getName());
                     return convertedState;
                 }
             }


bundles/core/org.eclipse.smarthome.core/src/main/java/org/eclipse/smarthome/core/items/ItemUtil.java
```diff
@@ -75,14 +75,18 @@ public static void assertValidItemName(String itemName) throws IllegalArgumentEx
...
+        }
+
         if (item != null && !isAccepted(item, state)) {
             for (Class<? extends State> acceptedType : item.getAcceptedDataTypes()) {
                 State convertedState = state.as(acceptedType);
                 if (convertedState != null) {
                     LoggerFactory.getLogger(ItemUtil.class).debug("Converting {} '{}' to {} '{}' for item '{}'",
-                            state.getClass().getSimpleName(), state.toString(),
-                            convertedState.getClass().getSimpleName(), convertedState.toString(), item.getName());
+                            state.getClass().getSimpleName(), state, convertedState.getClass().getSimpleName(),
```
martinvw: You should not call toString on logging arguments, just pass them on as an object they will be converted by logging framework if the log level applies

bundles/core/org.eclipse.smarthome.core/src/main/java/org/eclipse/smarthome/core/items/ItemUtil.java
```diff
@@ -75,14 +75,18 @@ public static void assertValidItemName(String itemName) throws IllegalArgumentEx
...
+    public static State convertToAcceptedState(State state, Item item) {
+        if (state == null) {
+            return UnDefType.NULL;
+        }
+
         if (item != null && !isAccepted(item, state)) {
             for (Class<? extends State> acceptedType : item.getAcceptedDataTypes()) {
                 State convertedState = state.as(acceptedType);
                 if (convertedState != null) {
                     LoggerFactory.getLogger(ItemUtil.class).debug("Converting {} '{}' to {} '{}' for item '{}'",
```
martinvw: imho this is a real strange solution to overcome a static logger. Now you are going to re-fetch the logger from the factory each time you pass by this debug (!!!) statement which is really often for this specific line. The coding guidelines leave room to have static loggers if required.

bundles/core/org.eclipse.smarthome.core/src/main/java/org/eclipse/smarthome/core/items/ItemUtil.java
```diff
@@ -75,14 +75,18 @@ public static void assertValidItemName(String itemName) throws IllegalArgumentEx
         }
     }
 
-    public static State convertToAcceptedState(final State state, final Item item) {
+    public static State convertToAcceptedState(State state, Item item) {
+        if (state == null) {
```
martinvw: This is the core of the actual change :-)

bundles/core/org.eclipse.smarthome.core/src/main/java/org/eclipse/smarthome/core/items/ItemUtil.java
```diff
@@ -44,11 +47,8 @@ private ItemUtil() {
      * @return true if the specified name is a valid item name, otherwise false
      */
     public static boolean isValidItemName(final String itemName) {
-        if (itemName != null && !itemName.isEmpty()) {
-            return itemName.matches("[a-zA-Z0-9_]*");
-        }
+        return StringUtils.isNotEmpty(itemName) && itemName.matches("[a-zA-Z0-9_]*");
```
martinvw: Sorry could not leave it like that :-)

martinvw: Btw, do we think that itemName of 0 characters is valid....?

martinvw:
@kaikreuzer imho this should solve the problem which was mentioned in the issue #3520 how were you able to reproduce it?

And on the community forum: https://community.openhab.org/t/java-lang-nullpointerexception-every-few-seconds/28944

bundles/core/org.eclipse.smarthome.core/src/main/java/org/eclipse/smarthome/core/items/ItemUtil.java
sjka: 👍  and no, AFAIK it needs to be a least one character long

bundles/core/org.eclipse.smarthome.core/src/main/java/org/eclipse/smarthome/core/items/ItemUtil.java
sjka: I wonder how `null` actually gets in here. IMHO we should fix the root cause of it first. Still, it's valid to make it fail-safe here.

bundles/core/org.eclipse.smarthome.core/src/main/java/org/eclipse/smarthome/core/items/ItemUtil.java
sjka: 👍

sjka:
The main question remains: What is the root-cause? Items should initially have a `UnDefType.NULL` state, so something recently must have changed which breaks with this assumption. Any clues?

sjka:
GroupItems without a function maybe? Then it could have been https://github.com/eclipse/smarthome/pull/3411

kaikreuzer:
I think this is a likely a correct assumption...

martinvw:
If I follow up the call-stack, I arrive here:
ItemUtil.convertToAcceptedState
called from 
ThingManager line 128 thingHandlerCallback.stateUpdated
called from
BaseThingHandler.updateState

Why can't it be a binding which just call's:

`updateState(channelUID, null);`

bundles/core/org.eclipse.smarthome.core/src/main/java/org/eclipse/smarthome/core/items/ItemUtil.java
martinvw: > and no, AFAIK it needs to be a least one character long

@kaikreuzer agree? Should it be a separate PR?

sjka:
> Why can't it be a binding which just call's:
updateState(channelUID, null);

Because then (as bad as it is...) it would have failed with an NPE in 
[ItemUtil line 94](https://github.com/eclipse/smarthome/blob/69393152219b85ae0e9efffda9793271345b7625/bundles/core/org.eclipse.smarthome.core/src/main/java/org/eclipse/smarthome/core/items/ItemUtil.java#L94), if I'm not mistaken.

martinvw:
Well spotted, yes I agree on that, so my change most likely doesn't fix the issue

martinvw:
I have my sister visiting, but I'll add changes later to fix setting of the state of e group.

bundles/core/org.eclipse.smarthome.core/src/main/java/org/eclipse/smarthome/core/items/ItemUtil.java
kaikreuzer: Certainly agree. Yes, to be clean, you can do it in a separate PR.

bundles/core/org.eclipse.smarthome.core/src/main/java/org/eclipse/smarthome/core/items/ItemUtil.java
```diff
@@ -44,11 +47,8 @@ private ItemUtil() {
...
```
kaikreuzer: hey, you missed to remove this empty line :-)

martinvw:
I don't see it coming from the mentioned PR either (no call to setState which might contain `null`), what do you think about this one:
[GroupItem.java#L329](https://github.com/eclipse/smarthome/blob/master/bundles/core/org.eclipse.smarthome.core/src/main/java/org/eclipse/smarthome/core/items/GroupItem.java#L329)

martinvw:
> I don't see it coming from the mentioned PR either (no call to setState which might contain null), what do you think about this one: GroupItem.java#L329

Then I'm most likely missing something because my branch contains no function which can return null either. Can you point out the line where you think its actually assigning null to the `oldState`?

martinvw:
Another likely suspect, #3438

https://github.com/eclipse/smarthome/blob/master/bundles/core/org.eclipse.smarthome.core/src/main/java/org/eclipse/smarthome/core/types/State.java#L29

bundles/core/org.eclipse.smarthome.core/src/main/java/org/eclipse/smarthome/core/items/ItemUtil.java
martinvw: @kaikreuzer what is your view on this specific solution

kaikreuzer: static logger would definitely be ok for me here

martinvw:
@triller-telekom can you take a look at NPE in combination with the change and line to which I referred, do you think this might be the cause of this? Thanks!

kaikreuzer:
You are right, the issue is definitely in https://github.com/eclipse/smarthome/commit/69393152219b85ae0e9efffda9793271345b7625#diff-543ecd419447346e6a177db8638e3ad0R74 - the `as()` method is allowed to return null and if you feed in an UNDEF at that place, the state will (now) be set to null internally. This worked in the past as UNDEF wasn't a Convertable state and thus was set directly.

bundles/core/org.eclipse.smarthome.core/src/main/java/org/eclipse/smarthome/core/items/ItemUtil.java
kaikreuzer: @martinvw I think we should not silently swallow this situation, but log an error (incl. stacktrace) nonetheless to highlight that there is some bug somewhere in the code as we should never get here.

martinvw:
So do you mean an UNDEF is converted to null, should we change the default implementation in that case?

Or should UnDefType override that method with `return this;`

bundles/core/org.eclipse.smarthome.core/src/main/java/org/eclipse/smarthome/core/items/ItemUtil.java
martinvw: I just see that we just did a isNotEmpty check so only the regex looked misleading but nothing will change :-D

kaikreuzer:
Not sure how we best get back the behavior we had with the "Convertibles" - I'd leave it to @sjka & @triller-telekom to decide as they both worked on that.

Issue #3518, opened 2017-05-25
[Paper UI] Sliders for Dimmer items have lost % sign


kaikreuzer:
So far, the Paper UI showed the textual percent value for sliders. This does not seem to be the case anymore:

<img width="382" alt="screen shot 2017-05-25 at 16 18 00" src="https://cloud.githubusercontent.com/assets/3244965/26453947/d470e036-4165-11e7-8d89-65150f70e2b1.png">

@aounhaider1 & @htreu Any idea what might have caused this change?

aounhaider1:
It is caused by https://github.com/eclipse/smarthome/pull/3466 which solved https://github.com/eclipse/smarthome/issues/3455. I intentionally removed the manual formatting done in PaperUI. Now it only formats it according to the given pattern. But I think to accommodate what you said, we can check if the pattern is empty and can do manual formatting in that case. Wdyt?

kaikreuzer:
We only render the slider for Dimmer items which are by definition always % values - so yes, I would say if there is no formatting pattern, the % should always be displayed as the default.

PR #3525, opened 2017-05-26
Use % sign as default for dimmer items

aounhaider1:
fixes https://github.com/eclipse/smarthome/issues/3518


extensions/ui/org.eclipse.smarthome.ui.paper/web-src/js/utility.js
```diff
@@ -29,6 +29,9 @@ angular.module('PaperUI.services').factory('util', function($filter, dateTime) {
                     return $filter('date')(date, "dd.MM.yyyy HH:mm:ss");
                 }
             } else if (!item.stateDescription || !item.stateDescription.pattern) {
+                if (item.type && item.type.toUpperCase() == "DIMMER") {
+                    state += "%";
```
kaikreuzer: Question: Might we want to have a space in front of the %?

aounhaider1: There was no space there previously but I think it makes sense to have it. Updated.

PR #551, opened 2015-11-29
Initial commit of Bluetooth binding and Bluetooth transport bundles

cdjackson:
This commit is just to provide visibility and prompt any discussion about API etc - **it's not considered for merging**.

Development is still ongoing and the API is likely to change!

This PR currently includes two bundles - a bluetooth (BLE) binding, and a `io.transport.bluetooth` bundle defining an abstract transport API. 

The transport bundle defines the API and provides a service to allow implementations to register. The concept is that hardware vendors can extend the transport bundle to define the hardware specific implementation. Development is very advanced on an implementation for Linux using the BlueZ library and that will be added to this PR in the near future.

The transport bundle provides an API that is very similar to the Android Bluetooth library - I've tried to keep definitions and implementation characteristics the same where possible (and where I can work out what Android does!) to make it easier for documentation and development etc.

The binding is far from complete - it is being used to support my debugging at the moment. The API I would consider is largely complete for BLE services at least although the base classes will still be updated.

Signed-off-by: Chris Jackson chris@cd-jackson.com

PR #3531, opened 2017-05-29
Initial commit of Bluetooth binding and Bluetooth transport bundles

maggu2810:
The code base should be equal to: https://github.com/eclipse/smarthome/pull/551

* based on the current master with all conflicts solved
* cleaned up POM files
* fixed license headers
* Java is set to 1.8
* fixed classpath
* applied clean up rules
* re-formatted ESH-INF XML files
* renamed yeelight to yeelightblue

---

IMHO the code itself needs further work.

* don't use printStackTrace
* don't catch exceptions and ignore them
* apply changes to work with the current Bluez versions
* etc.

BUT it should be enough to create a CQ and merge it, so further improvements could be done using a central place and code base.

We should ensure that someone is working (again) on that cleanup.


cdjackson:
I agree - it would be good to get this merged.

I’m currently working on an alternative transport (i.e. not dbus) using the BlueGiga BLED112 dongle which should be usable universally. I’m not 100% how much work it is at the moment but I’ve written an interface to the dongle which is working and I think it’s doable without huge work. At the moment I’m trying to avoid changing the common transport bundle but if I succeed in getting this working then it will be much easier for me to look at this as I won’t need a linux machine.

One comment on this - the yeelight bundle I think conflicts with another bundle in OH2 for the wifi version. I refactored the bundle here to be yeelightblue - it might pay to think about that before this gets merged.

maggu2810:
I created #3533 for the failed integration build.

@cdjackson I don't care if the openHAB binding changes its name to yeelightwifi or we change the ESH binding to yeelightblue (perhaps `yeelight-wifi` and `yeelight-blue` would make sense, so both state the transport it is using).

Should I change yeelight to yeelightblue or yeelight-blue?

cdjackson:
I personally don’t mind. I used yeelightblue here, but yeelight-blue does look nicer ;).

kaikreuzer:
> Should I change yeelight to yeelightblue or yeelight-blue?

We so far do not use any hyphens in binding ids, so I would clearly vote for yeelightblue.

maggu2810:
renamed yeelight to yeelightblue

extensions/binding/org.eclipse.smarthome.binding.yeelightblue/src/main/java/org/eclipse/smarthome/binding/yeelightblue/internal/handler/YeelightThingHandler.java
```diff
@@ -0,0 +1,349 @@
...
+
+    public synchronized void resetSleepTimer() {
+        // Stop any existing timer
+        if (timerTask != null) {
+            timerTask.cancel();
+        }
+        timerTask = null;
+    }
+
+    public float[] RGBtoHSB(int r, int g, int b, float[] hsbvals) {
```
svilenvul: I understand that the purpose of this PR is to apply only the most important changes, but I just want to mention that this method seems to me redundant, as it is already implemented in ESH - see HSBType.fromRGB()

kaikreuzer: @svilenvul Or alternatively feel free to directly commit fixes/clean-ups on a personal branch in your fork and we could pull that into this PR before it is going to be merged.

maggu2810: So, when do we want to create a CQ? Doesn't every change of the code base in this PR result in a delay of the initial CQ?
Wouldn't it be better to create a CQ for the initial code base and then allow us all the other changes (as long as they consists of <1k LOC) to be merged immediately without a CQ?

kaikreuzer: I wanted to at least spend an hour this week to roughly go over the code to see if it is ok to create a CQ (and thus merge as is) or if we (I) should do any changes to it upfront. In the meantime, I think it is ok for @svilenvul to work on improvements in his own branch - those can then result in a follow-up PR once this one is merged, right?

kaikreuzer:
Having had a closer look at the code, I am not really too keen to merge it as is anymore.

Especially the Yeelight binding shows that there are huge issues in the architecture:

- It implements a discovery service that triggers general BT device discovery. Imho we should have a BTDiscoveryService with a Participant interface, similar to UPnP and mDNS.
- The whole code depends on the BluetoothAdapter, its correct configuration, etc and has to handle all potential issues. I think it would be much better to model the BluetoothAdapter as a Bridge - this would easily allow adding other Bridges like the BlueGiga dongle.
- Imho, we SHOULD after all have a "BLE Binding", which defines the bridge and provides the connectivity (i.e. parts what is in the transport.bluetooth.bluez) and additional bundles that then come with specific thing-types and handler implementations.

Additionally, the Yeelight binding is in a bad shape:
- The binding has switch/dimmer/color channels side by side, which is wrong.
- It defines channels like Temperature & Luminance, which are not reflected anywhere in the code.
- Its config parameters are all device management actions, which is something we do not really officially support (and where I do not even see how they are used).
- The unique ID is added as a property, although this needs to be a config parameter for the handler.

I tried to refactor the binding, but ended up with hardly any code left in place. As the binding isn't much code anyhow, I would suggest to completely remove it from the PR, so that it does not accidentally serve as a template for others.

For the other two transport bundles, I would create a CQ then, unless @cdjackson has again time for helping on that code, so that we could discuss potential refactorings wrt the BlueGiga dongle.

extensions/binding/org.eclipse.smarthome.binding.yeelightblue/src/main/java/org/eclipse/smarthome/binding/yeelightblue/internal/handler/YeelightThingHandler.java
svilenvul: Thanks for giving me green light to apply some changes, I have put this on my TODO list :)

kaikreuzer: > I have put this on my TODO list :)

Better refrain from it for the moment - I just had a call with Chris and I will write a small summary soon; we agreed on some substantial refactoring of the code :-/

cdjackson:
I would suggest to hold off for a couple of weeks while I investigate some changes - I think Kai will write something on this later.

extensions/binding/org.eclipse.smarthome.binding.yeelightblue/src/main/java/org/eclipse/smarthome/binding/yeelightblue/internal/handler/YeelightThingHandler.java
svilenvul: OK, I haven't read you comment below before :)

kaikreuzer:
@cdjackson I tried to put our discussion in a diagram:

ble

The general scheme is that we should not think of the "BLE Binding" to be a single bundle, but rather a feature (i.e. a set of bundles, not all of which need to be installed).

- The "base" BLE bundle is the central one that is always required.
  - It has the "ble" binding.xml
  - It contains an implementation of a discovery service (which requires the presence of a BLEBridge)
  - It defines a (yet to be designed) high level BLE-API interface for specific device support bundles as the current Android-like API is too low level.
- BLEBridge implementations can be added as separate bundles. 
  - They come with a bridge-type (which defines their required configuration parameters) and an according handler implementations
  - Their handlers implement the BLE-API interface.
  - We can have one bundle that brings a handler that uses BlueZ over DBUS and another bundle bringing support for BlueGiga USB dongle.
- Specific device support can as well be added via separate bundles.
  - These bundles register a BLEDiscoveryParticipant and thus can provide DiscoveryResults for specific devices.
 - They define a specific thing-type and come with a specific handler implementation.
 - Their handlers have access to the bridge, which by definition implements the BLE-API interface, which can hence be used by those handlers.

This should all nicely fit into the current framework architecture (actually, the binding API had been designed specifically in a way to support that, so I think it is time to actually use it that way :-)).
There is one issue that comes to my mind, though: All thing-types will have to reference all possible bridge-types in their XML, which is not nice. We should check whether it is possible to also allow some kind of placeholder, so that specific bundles do not need to be aware of all potential BLEBridge implementations.

Chris, does that match our conversation? Do you have anything to add?
If you have time to work on such refactorings, I would suggest to leave this PR open (after all) for now.
As @svilenvul is also eagerly waiting to get his hands into BLE, please let us discuss how the work could be possibly shared. Could e.g. @svilenvul already start with creating the suggested bundle structure?

cdjackson:
Yes - I think that reflects our conversation. I have started already to implement this and would hope to have something in a week or so with any luck - we can discuss more then as I have a feeling there may be a few issues that might pop up ;)

kaikreuzer:
If I am not mistaken, this is superseded by #3633 - unless anyone does not agree with the suggested refactoring of the architecture.

PR #3532, opened 2017-05-29
Refactor the HttpServiceUtil

htreu:
* refactor HttpServiceUtil (decrease indention level, remove @SuppressWarnings)
* add unit test


bundles/core/org.eclipse.smarthome.core.test/META-INF/MANIFEST.MF
```diff
@@ -18,4 +18,6 @@ Import-Package: groovy.lang,
  org.junit,
  org.junit.runner,
  org.junit.runners,
- org.osgi.service.cm
+ org.osgi.service.cm,
+ org.mockito
+Require-Bundle: org.junit,org.mockito,org.hamcrest
```
maggu2810: Why do we need Require-Bundle?

bundles/core/org.eclipse.smarthome.core/src/main/java/org/eclipse/smarthome/core/net/HttpServiceUtil.java
```diff
@@ -39,45 +39,44 @@ public static int getHttpServicePortSecure(final BundleContext bc) {
...
-                    } else {
-                        serviceRanking = (Integer) value;
-                    }
-                    if (serviceRanking >= candidate) {
-                        candidate = serviceRanking;
-                        port = servicePort;
-                    }
-                }
-            }
+            refs = bc.getAllServiceReferences("org.osgi.service.http.HttpService", null);
```
maggu2810: I don't think we need to initialize the refs array first, then fetch the correct one with catching the error (that should never be thrown) and at last evaluate -- if the error has been catched -- the null array...

If you want to move the code outside of the try-block then you can keep the final and remove the initialization with the null array. Just add a `return -1;` to the catch block.

bundles/core/org.eclipse.smarthome.core.test/src/test/java/org/eclipse/smarthome/core/net/HttpServiceUtilTest.java
```diff
@@ -0,0 +1,110 @@
...
+import org.osgi.framework.Constants;
+import org.osgi.framework.InvalidSyntaxException;
+import org.osgi.framework.ServiceReference;
+
+import com.google.common.collect.Lists;
+
+public class HttpServiceUtilTest {
+
+    private static final String ORG_OSGI_SERVICE_HTTP_SERVICE = "org.osgi.service.http.HttpService";
+    private final static String HTTP_PORT = "org.osgi.service.http.port";
```
amitjoy: To be consistent - `public static final`

bundles/core/org.eclipse.smarthome.core/src/main/java/org/eclipse/smarthome/core/net/HttpServiceUtil.java
```diff
@@ -39,45 +39,44 @@ public static int getHttpServicePortSecure(final BundleContext bc) {
     }
 
     // Utility method that could be used for non-secure and secure port.
-    @SuppressWarnings("rawtypes")
     private static int getHttpServicePortProperty(final BundleContext bc, final String propertyName) {
         Object value;
         int port = -1;
 
         // Try to find the port by using the service property (respect service ranking).
+        ServiceReference<?>[] refs = new ServiceReference[0];
```
amitjoy: As it is implementation specific, we can use `BundleContext#getServiceReferences(Class, String)` that returns `Collection<ServiceReference<SERVICE>>`

bundles/core/org.eclipse.smarthome.core/src/main/java/org/eclipse/smarthome/core/net/HttpServiceUtil.java
```diff
@@ -39,45 +39,44 @@ public static int getHttpServicePortSecure(final BundleContext bc) {
...
-                    if (serviceRanking >= candidate) {
-                        candidate = serviceRanking;
-                        port = servicePort;
-                    }
-                }
-            }
+            refs = bc.getAllServiceReferences("org.osgi.service.http.HttpService", null);
         } catch (final InvalidSyntaxException ex) {
             // This point of code should never be reached.
             final Logger logger = LoggerFactory.getLogger(HttpServiceUtil.class);
```
amitjoy: As you have already pointed that this code block will never be executed as there is no OSGi filter associated, we can leave it empty. What do you think?

bundles/core/org.eclipse.smarthome.core/src/main/java/org/eclipse/smarthome/core/net/HttpServiceUtil.java
sjka: OSGi 4.2!

bundles/core/org.eclipse.smarthome.core.test/src/test/java/org/eclipse/smarthome/core/net/HttpServiceUtilTest.java
sjka: `private` ...!

bundles/core/org.eclipse.smarthome.core/src/main/java/org/eclipse/smarthome/core/net/HttpServiceUtil.java
maggu2810: I don't like empty catch blocks (except we know that the behavior is "okay" and could be ignored). The `InvalidSyntaxException` should not be thrown ever, so I think it is okay to write a log message if it is thrown just to ensure to signal that there is something wrong.

bundles/core/org.eclipse.smarthome.core/src/main/java/org/eclipse/smarthome/core/net/HttpServiceUtil.java
maggu2810: See below

bundles/core/org.eclipse.smarthome.core.test/src/test/java/org/eclipse/smarthome/core/net/HttpServiceUtilTest.java
maggu2810: Sure, you are right :wink:

bundles/core/org.eclipse.smarthome.core.test/src/test/java/org/eclipse/smarthome/core/net/HttpServiceUtilTest.java
```diff
@@ -0,0 +1,110 @@
...
+    private final static String HTTP_PORT_SECURE = "org.osgi.service.http.port.secure";
```
maggu2810: Same here :wink:

bundles/core/org.eclipse.smarthome.core/src/main/java/org/eclipse/smarthome/core/net/HttpServiceUtil.java
amitjoy: Oh I see.

bundles/core/org.eclipse.smarthome.core/src/main/java/org/eclipse/smarthome/core/net/HttpServiceUtil.java
amitjoy: 👍

bundles/core/org.eclipse.smarthome.core/src/main/java/org/eclipse/smarthome/core/net/HttpServiceUtil.java
htreu: nice!

bundles/core/org.eclipse.smarthome.core.test/src/test/java/org/eclipse/smarthome/core/net/HttpServiceUtilTest.java
htreu: tnx guys. fixed.

bundles/core/org.eclipse.smarthome.core.test/META-INF/MANIFEST.MF
htreu: Require-Bundle is needed to resolve `org.hamcrest` correctly. Since it is a split package and none of the bundles providing parts of it (junit, mockito) do export the base package itself we have to hard reference the bundles which do here (this is at least my understanding and the only configuration I could get this to work). This is all required to have mockito available.

maggu2810: The only thing to do is to add "org.mockito.stubbing" to your package imports:

```diff
diff --git a/bundles/core/org.eclipse.smarthome.core.test/META-INF/MANIFEST.MF b/bundles/core/org.eclipse.smarthome.core.test/META-INF/MANIFEST.MF
index 889feb68c..b83e1b376 100644
--- a/bundles/core/org.eclipse.smarthome.core.test/META-INF/MANIFEST.MF
+++ b/bundles/core/org.eclipse.smarthome.core.test/META-INF/MANIFEST.MF
@@ -19,5 +19,5 @@ Import-Package: groovy.lang,
  org.junit.runner,
  org.junit.runners,
  org.osgi.service.cm,
- org.mockito
-Require-Bundle: org.junit,org.mockito,org.hamcrest
+ org.mockito,
+ org.mockito.stubbing
```

Or am I missing something. Can you give it a try?

htreu: Awesome, this works very well. Do you mind explaining _why_ it solves this issue?

htreu: This affects some other places too where mockito was introduced recently. I will push anothr PR addressing this.

maggu2810: You use `import static org.mockito.Mockito.*;` and call e.g. `when(...)`

The function you are calling is defined in `org.mockito.Mockito` as followed:

```java
    public static <T> OngoingStubbing<T> when(T methodCall) {
        return MOCKITO_CORE.when(methodCall);
    }
```

So, there is e.g. `org.mockito.stubbing.OngoingStubbing<T>` used.
The package `org.mockito.stubbing` is not part of the respective classloader, because there is no import section defined.
If you use Require-Bundle all part of the required bundle is available in by the classloader, but you refer a special bundle and not "any" bundle that provides that package.


> This affects some other places too where mockito was introduced recently. I will push another PR addressing this.

There should be no other place affected, IIRC "do not use Require-Bundle" is part of our coding guidelines :wink: 
So, if it has been used by new contributions, it would be great, if you could remove it again.

htreu: My explanation above was wrong in one point: I did introduce the Required-Bundle to fully support hamcrest (not mockito). I tried with one of the other affected bundles (org.eclipse.smarthome.io.rest.core.test) and there the change will lead to compile errors since hamcrest is not fully available. Here only org.hamcrest.CoreMatchers is used so no Required-Bundle is required.

htreu:
As travis tells, the proposed change in MANIFEST.MF fails to run the tests on the command line. Here is an excerpt from the test logfile:

```
!ENTRY org.eclipse.smarthome.core.test 2 0 2017-05-29 15:33:55.936
!MESSAGE Could not resolve module: org.eclipse.smarthome.core.test [91]
  Unresolved requirement: Import-Package: org.mockito
    -> Export-Package: org.mockito; bundle-symbolic-name="org.mockito"; bundle-version="1.9.5.v201311280930"; version="1.9.5"; uses:="org.mockito.stubbing,  org.hamcrest,  org.mocki
       org.mockito [94]
         Unresolved requirement: Import-Package: COM.jrockit.reflect; resolution:="optional"
         Unresolved requirement: Import-Package: jrockit.vm; resolution:="optional"
         Unresolved requirement: Import-Package: org.hamcrest; version="[1.0.0,2.0.0)"


!ENTRY org.mockito 2 0 2017-05-29 15:33:55.937
!MESSAGE Could not resolve module: org.mockito [94]
  Unresolved requirement: Import-Package: COM.jrockit.reflect; resolution:="optional"
  Unresolved requirement: Import-Package: jrockit.vm; resolution:="optional"
  Unresolved requirement: Import-Package: org.hamcrest; version="[1.0.0,2.0.0)"
```

I will once more update the PR and roll back to `Require-Bundle: org.junit,org.mockito,org.hamcrest
`.

maggu2810:
Okay, I was not aware that hamcrest is using split packages etc.

PR #3536, opened 2017-05-29
fix license headers

maggu2810:
As documented it would be good practice to run `mvn license:format` in front of a commit, so new files use the correct license headers.


maggu2810:
Should we exclude the `tools/archetype` directory or at least `tools/archetype/binding/src/main/resources/archetype-resources/OSGI-INF/__bindingIdCamelCase__HandlerFactory.xml`?

kaikreuzer:
> Should we exclude the tools/archetype directory

I guess so, yes.

maggu2810:
IMHO we should use "Rebase and merge" for this PR. If you have a look at the first commit, it clean up the include and exclude section first.
The second one applies the correct license header update itself.

kaikreuzer:
done, thanks.

PR #3549, opened 2017-05-31
Updated ItemRegistryImpl to make use of {@inheritDoc} notation

AlexanderKostadinov28:
Removed `{@inheritDoc}` and `(non-Javadoc) @see` notations throughout ESH completely (code style consistency).

### DIFF
--- a/bundles/model/org.eclipse.smarthome.model.item/src/org/eclipse/smarthome/model/item/internal/GenericItemProvider.java
+++ b/bundles/model/org.eclipse.smarthome.model.item/src/org/eclipse/smarthome/model/item/internal/GenericItemProvider.java
@@ -129,9 +129,6 @@ public class GenericItemProvider extends AbstractProvider<Item>
         }
     }
 
-    /**
-     * {@inheritDoc}
-     */
     @Override
     public Collection<Item> getAll() {
         List<Item> items = new ArrayList<Item>();
@@ -339,11 +336,6 @@ public class GenericItemProvider extends AbstractProvider<Item>
         }
     }
 
-    /**
-     * {@inheritDoc}
-     * <p>
-     * Dispatches all binding configs and fires all {@link ItemsChangeListener}s if {@code modelName} ends with "items".
-     */
     @Override
     public void modelChanged(String modelName, EventType type) {
         if (modelName.endsWith("items")) {
--- a/extensions/transform/org.eclipse.smarthome.transform.scale/src/main/java/org/eclipse/smarthome/transform/scale/internal/ScaleTransformationService.java
+++ b/extensions/transform/org.eclipse.smarthome.transform.scale/src/main/java/org/eclipse/smarthome/transform/scale/internal/ScaleTransformationService.java
@@ -38,18 +38,16 @@ public class ScaleTransformationService extends AbstractFileTransformationServic
     private static final Pattern LIMITS_PATTERN = Pattern.compile("(\\[|\\])(.*)\\.\\.(.*)(\\[|\\])");
 
     /**
-     * <p>
-     * Transforms the input <code>source</code> by matching searching the range where it fits
-     * i.e. [min..max]=value or ]min..max]=value
-     * </p>
+     * Performs transformation of the input <code>source</code>
+     * 
+     * The method transforms the input <code>source</code> by matching searching
+     * the range where it fits i.e. [min..max]=value or ]min..max]=value
      *
      * @param properties
      *            the list of properties defining all the available ranges
      * @param source
      *            the input to transform
-     *
-     * @{inheritDoc
-     *
+     * 
      */
     @Override
     protected String internalTransform(Map<Range, String> data, String source) throws TransformationException {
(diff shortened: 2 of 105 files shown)


maggu2810:
Reading the JavaDoc documentation (http://docs.oracle.com/javase/8/docs/technotes/tools/windows/javadoc.html) states that inheritDoc is already done implicit for overwritten methods.

So I would prefer not to add implicit stuff as it does not improve readability of the code at all.

maggu2810:
So, I agree with removing the JavaDoc for overwritten methods if it is equal to the one of the parent class but remove the use of inheritDoc.

AlexanderKostadinov28:
I agree then, but in such case we should remove the JavaDoc for all inherited methods from `ItemRegistry` (there are some left `non-JavaDoc @see` notations).

kaikreuzer:
> remove the use of inheritDoc

I agree and that is also the style we tried to apply throughout ESH so far.

AlexanderKostadinov28:
In such case, I can remove the {@inheritDoc} notation, as well as get rid of unnecessary javadoc notations for overriden methods throughout the whole ESH.

AlexanderKostadinov28:
Finally, I removed {@inheritDoc} and (non-Javadoc) @see notations throughout ESH completely for code style consistency. Significant number of the (non-Javadoc) @see notations were obsolete and mentioned wrong location.

bundles/model/org.eclipse.smarthome.model.item/src/org/eclipse/smarthome/model/item/internal/GenericItemProvider.java
```diff
@@ -340,8 +337,6 @@ private void internalDispatchBindings(BindingConfigReader reader, String modelNa
     }
 
     /**
-     * {@inheritDoc}
-     * <p>
```
maggu2810: I am not sure about this one.
Wasn't the previous approach to use the parent JavaDoc and additionally an sentence?
Isn't now the parent JavaDoc overwritten by the sentence?

I can't check it ATM but it feels like "parent + custom" before and "custom" now. WDYT?

extensions/transform/org.eclipse.smarthome.transform.scale/src/main/java/org/eclipse/smarthome/transform/scale/internal/ScaleTransformationService.java
```diff
@@ -47,9 +47,6 @@
      *            the list of properties defining all the available ranges
      * @param source
      *            the input to transform
-     *
```
maggu2810: This JavaDoc seems to be wrong formatted for me at all.

So, not really part of your "remove inheritDoc" changes, but I don't think a correct JavaDoc starts with a `<p>` at all.
Shouldn't it start with a short sentence and then perhaps a longer description.

Another todo that we should remember is the removal of all `</p>`. AFAIK there is only `<p>` defined for "correct" JavaDoc.

extensions/transform/org.eclipse.smarthome.transform.xslt/src/main/java/org/eclipse/smarthome/transform/xslt/internal/XsltTransformationService.java
```diff
@@ -46,9 +46,6 @@
      *            as well
      * @param source
      *            the input to transform
-     *
```
maggu2810: See above (don't start with `<p>`, remove `</p>`)

bundles/model/org.eclipse.smarthome.model.item/src/org/eclipse/smarthome/model/item/internal/GenericItemProvider.java
AlexanderKostadinov28: Actually, you are right - there is no JavaDoc in the parent but only in the inherited class.  So, I will update the PR and remove the JavaDoc from here and move it to ModelRepositoryChangeListener.

extensions/transform/org.eclipse.smarthome.transform.scale/src/main/java/org/eclipse/smarthome/transform/scale/internal/ScaleTransformationService.java
AlexanderKostadinov28: Here I will remove the <p> in the beginning, as well as edit the JavaDoc according to your comment. In general, in the upcoming days, I will create another PR for fixing all of the unclosed <p> tags in JavaDocs throughout ESH - this has to be done. :)

extensions/transform/org.eclipse.smarthome.transform.xslt/src/main/java/org/eclipse/smarthome/transform/xslt/internal/XsltTransformationService.java
AlexanderKostadinov28: I will do the same as above.

AlexanderKostadinov28:
@maggu2810 All of your latest comments are resolved now. As said, I will consider creating another PR for fixing the JavaDoc </p> usage.

Issue #2327, opened 2016-10-15
Example bundle for UI testing


kaikreuzer:
We have many different combinations of channel types, item types, read-only, state-options etc. and I see that these are a regular source for [regressions](https://github.com/eclipse/smarthome/issues/2326) in the Paper UI (and possibly others as well).

As automatic UI testing is challenging and we do not have anything for testing the visual appearence in place, I would suggest to create an example bundle, which has a large number of different channels, config descriptions, rule modules, etc. in place and which can be used to do manual testing. At the same time, it would be very valuable for reproducing bug reports as well.

@maggu2810, @aounhaider1, @sjka What do you think? If you like the idea, what would be a good place for it to live in the repo?

maggu2810:
I am fine with this suggestion.

I have already done it myself some time e.g. here if I need to test / fix the Paper UI myself:
https://github.com/maggu2810/smarthome/tree/binding-test/extensions/binding/org.eclipse.smarthome.binding.test

sjka:
I like the idea too. We actually have built something similar here for our purposes, like a "virtual" binding. It's used heavily for development and fulfills its purpose for the "standard" use-cases. 

However, I think it's also worth mentioning that the complexity gets enormous if you want to cover all dimensions completely and also use it for bug analysis. So we just had the idea to extend it in a way that it exposes its own REST resource where it can be more or less configured (e.g. "give me a ThingType with several channels where the ThingHandler takes ages to initialize"). That doesn't conflict with the idea though to have a static set of predefined types which you can rely on for testing. It's more like an extension to it.

I would be happy to have it here! extensions/binding/o.e.sh.binding.test ist surely the best obvious place we currently have. An alternative would be tests/binding/... to separate it from the "real" bindings, but on the other hand this folder would stay pretty much empty and we don't separate the other test fragments from their bundles either. I'm fine with @maggu2810's suggestion.

kaikreuzer:
> extensions/binding/o.e.sh.binding.test ist surely the best obvious place we currently have.

I am not sure if we should restrict it to "binding". As mentioned above, we could also have rule modules and other stuff (persistence services, voice features, etc.) in there - I would prefer to have all this in a single bundle and not ending up with 5 different bundles that cover different parts of the system.

htreu:
I will take a look at this and will propose a first version of such a bundle. stay tuned.

PR #3555, opened 2017-06-01
Initial commit of the MagicBundle - a UI test bundle

htreu:
The MagicBundle is a dummy bundle/binding implementation which provides virtual things and channels. These things can be simply set up and used to test the PaperUI. In addition the initial commit provides the MagicService which already has an extensive config description. This can also be used to test the service configuration in PaperUI.

This addresses #2327.


htreu:
@kaikreuzer agreed. will update.

bundles/test/org.eclipse.smarthome.magic.test/pom.xml
```diff
@@ -0,0 +1,126 @@
...
+
+  <modelVersion>4.0.0</modelVersion>
+
+  <parent>
+    <groupId>org.eclipse.smarthome.bundles</groupId>
+    <artifactId>test</artifactId>
+    <version>0.9.0-SNAPSHOT</version>
+  </parent>
+
+  <groupId>org.eclipse.smarthome.archetype</groupId>
```
maggu2810: I don't think this is correct.

PR #3558, opened 2017-06-01
Added CQ references and removed Californium sources

kaikreuzer:
For some reason, I failed to use the sources of californium from the p2 site (or rather they made it impossible to resolve the target platform), so I decided to remove them here. Any better ideas welcome.

screen shot 2017-06-01 at 12 27 43


sjka:
Couldn't figure out the reason either. 
Same as with why git always reports _all_  the binaries as changed. They are absolutely identical as per `diff`. 
Anyway, let's figure this later, first fix the repo again.

PR #3559, opened 2017-06-01
replaced printStackTrace by proper logging

kaikreuzer:


bundles/automation/org.eclipse.smarthome.automation.providers/src/main/java/org/eclipse/smarthome/automation/internal/core/provider/AbstractResourceBundleProvider.java
```diff
@@ -400,7 +400,7 @@ protected Bundle getBundle(String uid) {
                 try {
                     uri = new URI(prefix + ":" + uid + ".name");
                 } catch (URISyntaxException e) {
-                    e.printStackTrace();
+                    logger.error("Constructed invalid uri '{}'", prefix + ":" + uid + ".name", e);
```
sjka: Not that it hurts much in this place, but wouldn't the idea of slf4j be

```
logger.error("Constructed invalid uri '{}:{}.name'", prefix, uid, e);
```

kaikreuzer: You are clearly right. Updated.

PR #3565, opened 2017-06-02
Fixed all JavaDoc paragraphs throughout ESH.

AlexanderKostadinov28:
Throughout all java classes, all closing `</p>` tags in JavaDoc are removed and only the opening `<p>` are left, since this complies to the official Oracle rules. See [How to Write Doc Comments for the Javadoc Tool](http://www.oracle.com/technetwork/articles/java/index-137868.html), section "Writing Doc Comments".

PR suggested by @maggu2810 


kaikreuzer:
> I like it.

The PR is fully correct, what I don't like is the Javadoc style guide, which seems to come from an ancient time, when there was only HTML, but no XHTML defined yet...
But well, we obviously cannot do anything against that.

maggu2810:
Why should JavaDoc follow the rules of (X)HTML at all? :wink:

PR #3568, opened 2017-06-02
fix JavaDoc of the script extension provider

maggu2810:
* `importPreset` must not return a null object if we collow the
  callstack but should use an empty collection.
  The returned map is forarded to the `scopeValues` method of an engine
  provider. The engine provider calls `entrySet` on the given argument,
  so if someone uses currently null it should already result in a NPE.
  IMHO it is most of the time better to return an empty map / collection
  etc. instead of null.
* I don't know what is the best approach for `get` if the type is not
  supported. There is a `getTypes` method I assume that this function is
  called for "supported" types only. IMHO it is okay to throw an
  `IllegalArgumentException` if the `get` method is called with an
  unsupported type and we could manifest that a non-null object should
  be returned.

WDYT about the return values / exceptions?

I will change the commit (JavaDoc) if we agree how to handle the second
point.

### DIFF
--- a/bundles/automation/org.eclipse.smarthome.automation.module.script/src/main/java/org/eclipse/smarthome/automation/module/script/ScriptExtensionProvider.java
+++ b/bundles/automation/org.eclipse.smarthome.automation.module.script/src/main/java/org/eclipse/smarthome/automation/module/script/ScriptExtensionProvider.java
@@ -44,17 +44,20 @@ public interface ScriptExtensionProvider {
      * This method should return an Object of the given type. Note: get can be called multiple times in the scripts use
      * caching where appropriate.
      *
-     * @param scriptEngine the script engine instance requesting the given type
-     * @param type
-     * @return
+     * @param scriptIdentifier the identifier of the script that requests the given type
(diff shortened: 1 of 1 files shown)


maggu2810:
@smerschjohann As you created the file, can you have a look at please?

smerschjohann:
1) Yes, I agree. `null` is no valid output for this method. We could even use the the `@NonNull` annotation.

2) Yes, the method is only called for the supported types, so it is okay to throw here. So, if we want to restrict the interfaces a little bit, it is fine with me if we define it as non-null as well.

maggu2810:
Thank you for your reply, I will change the JavaDoc.

I am using nullable / not nullable annotations in my projects and the checker framework to check my code against potential NPEs etc. It is currently not supported by the ESH but I already created a Wiki entry (as a suggestion) that we could support such annotations.
I prefer the point of view "all non annotated type states 'non-null' and nullable is annotated" (checker framework nullness checker default)... but this is another story.

maggu2810:
@smerschjohann Sorry to bother you again, but would you be able to review the changes?

smerschjohann:
LGTM

PR #3572, opened 2017-06-03
Moved ExpiringCache from Yahoo Weather Binding to Eclipse SmartHome Core

### LINKED ISSUE #3302: Reusable cache util
As we have seen in eclipse/smarthome#2809 and now also in https://github.com/openhab/openhab2-addons/pull/1543, it would be useful to have little a utility to easily build a small cache. While guava offeres a nice one in some versions, it turned out it's not feasible for us to use because we'd need to update guava to a newer version (see #2801 why this is not a good idea). 
Therefore I suggest to (maybe) pimp the little [ExpiringCache](https://github.com/eclipse/smarthome/blob/3ae5d7b1b284f2bde6199bfe8dad950fd38cae39/extensions/binding/org.eclipse.smarthome.binding.yahooweather/src/main/java/org/eclipse/smarthome/binding/yahooweather/internal/ExpiringCache.java) from the yahoo weather binding a little and make it available as an API from the core.

cweitkamp:
* Moved `ExpiringCache` from Yahoo Weather Binding to Eclipse SmartHome Core
* Replaced Interface `LoadAction` with Interface `java.util.function.Supplier`
* Added `invalidate` an `refresh` methods
* Added `ExpiringCacheMap` to support handling multiple cache values
* Added `ExpiringCacheTest` and `ExpiringCacheMapTest`

Closes #3302 

Also-by: Simon Kaufmann <simon.kfm@googlemail.com>

### DIFF
--- a/bundles/core/org.eclipse.smarthome.core.test/src/test/java/org/eclipse/smarthome/core/cache/ExpiringCacheTest.java
+++ b/bundles/core/org.eclipse.smarthome.core.test/src/test/java/org/eclipse/smarthome/core/cache/ExpiringCacheTest.java
@@ -1,0 +1,85 @@
+/**
+ * Copyright (c) 2014-2017 by the respective copyright holders.
+ * All rights reserved. This program and the accompanying materials
+ * are made available under the terms of the Eclipse Public License v1.0
+ * which accompanies this distribution, and is available at
+ * http://www.eclipse.org/legal/epl-v10.html
+ */
+package org.eclipse.smarthome.core.cache;
+
+import static org.junit.Assert.*;
+
+import org.junit.Test;
+import org.slf4j.Logger;
+import org.slf4j.LoggerFactory;
+
+/**
+ * Test class for the {@link ExpiringCache} class.
+ *
+ * @author Christoph Weitkamp - Initial contribution.
+ */
+public class ExpiringCacheTest {
+
+    private final Logger logger = LoggerFactory.getLogger(ExpiringCacheTest.class);
+
+    @Test(expected = IllegalArgumentException.class)
+    public void testIllegalArgumentException1() throws IllegalArgumentException {
+        new ExpiringCache<String>(ExpiringCacheMapTest.CACHE_EXPIRY, null);
+    }
+
+    @Test
+    public void testGetValue() {
+        final ExpiringCache<String> single_cache = new ExpiringCache<String>(ExpiringCacheMapTest.CACHE_EXPIRY,
+                ExpiringCacheMapTest.CACHE_ACTION);
+
+        // use the same key twice
+        String value1 = single_cache.getValue();
+        assertNotNull(value1);
+        String value2 = single_cache.getValue();
+        assertNotNull(value2);
+        assertEquals(value1, value2);
+    }
+
+    @Test
+    public void testExpired() throws InterruptedException {
+        final ExpiringCache<String> single_cache = new ExpiringCache<String>(ExpiringCacheMapTest.CACHE_EXPIRY,
+                ExpiringCacheMapTest.CACHE_ACTION);
+
+        String value1 = single_cache.getValue();
+        assertFalse(single_cache.isExpired());
+
+        // wait until cache expires
+        Thread.sleep(ExpiringCacheMapTest.CACHE_EXPIRY + 100);
+        assertTrue(single_cache.isExpired());
+
+        String value2 = single_cache.getValue();
+        assertFalse(single_cache.isExpired());
+        assertNotEquals(value1, value2);
+    }
+
+    @Test
+    public void testInvalidate() {
+        final ExpiringCache<String> single_cache = new ExpiringCache<String>(ExpiringCacheMapTest.CACHE_EXPIRY,
+                ExpiringCacheMapTest.CACHE_ACTION);
+
+        String value1 = single_cache.getValue();
+
+        // invalidate item
+        single_cache.invalidateValue();
+
+        String value2 = single_cache.getValue();
+        assertNotEquals(value1, value2);
+    }
+
+    @Test
+    public void testRefresh() {
+        final ExpiringCache<String> single_cache = new ExpiringCache<String>(ExpiringCacheMapTest.CACHE_EXPIRY,
+                ExpiringCacheMapTest.CACHE_ACTION);
+
+        String value1 = single_cache.getValue();
+
+        // refresh item
+        String value2 = single_cache.refreshValue();
+        assertNotEquals(value1, value2);
+    }
+}
--- a/bundles/core/org.eclipse.smarthome.core/src/main/java/org/eclipse/smarthome/core/cache/ExpiringCache.java
+++ b/bundles/core/org.eclipse.smarthome.core/src/main/java/org/eclipse/smarthome/core/cache/ExpiringCache.java
@@ -1,0 +1,91 @@
+/**
+ * Copyright (c) 2014-2017 by the respective copyright holders.
+ * All rights reserved. This program and the accompanying materials
+ * are made available under the terms of the Eclipse Public License v1.0
+ * which accompanies this distribution, and is available at
+ * http://www.eclipse.org/legal/epl-v10.html
+ */
+package org.eclipse.smarthome.core.cache;
+
+import java.util.concurrent.TimeUnit;
+import java.util.function.Supplier;
+
+import org.slf4j.Logger;
+import org.slf4j.LoggerFactory;
+
+/**
+ * This is a simple expiring and reloading cache implementation.
+ *
+ * There must be provided an action in order to retrieve/calculate the value. This action will be called only if the
+ * answer from the last calculation is not valid anymore, i.e. if it is expired.
+ *
+ * @author Christoph Weitkamp - Initial contribution and API.
+ * 
+ * @param <V> the type of the value
+ */
+public class ExpiringCache<V> {
+
+    private final Logger logger = LoggerFactory.getLogger(ExpiringCache.class);
+
+    private final long expiry;
+    private final Supplier<V> action;
+    private V value;
+    private long expiresAt;
+
+    /**
+     * Create a new instance.
+     * 
+     * @param expiry the duration in milliseconds for how long the value stays valid
+     * @param action the action to retrieve/calculate the value
+     */
+    public ExpiringCache(long expiry, Supplier<V> action) {
+        if (action == null) {
+            throw new IllegalArgumentException("ExpiringCacheItem cannot be created as action is null.");
+        }
+
+        this.expiry = TimeUnit.MILLISECONDS.toNanos(expiry);
+        this.action = action;
+
+        invalidateValue();
+    }
+
+    /**
+     * Returns the value - possibly from the cache, if it is still valid.
+     * 
+     * @return the value
+     */
+    public synchronized V getValue() {
+        if (value == null || isExpired()) {
+            return refreshValue();
+        }
+        return value;
+    }
+
+    /**
+     * Invalidates the value in the cache.
+     */
+    public synchronized void invalidateValue() {
+        value = null;
+        expiresAt = 0;
+    }
+
+    /**
+     * Refreshes and returns the value in the cache.
+     * 
+     * @return the new value
+     */
+    public synchronized V refreshValue() {
+        value = action.get();
+        expiresAt = System.nanoTime() + expiry;
+        return value;
+    }
+
+    /**
+     * Checks if the value is expired.
+     * 
+     * @return true if the value is expired
+     */
+    public boolean isExpired() {
+        return expiresAt < System.nanoTime();
+    }
+}
--- a/bundles/core/org.eclipse.smarthome.core/src/main/java/org/eclipse/smarthome/core/cache/ExpiringCacheMap.java
+++ b/bundles/core/org.eclipse.smarthome.core/src/main/java/org/eclipse/smarthome/core/cache/ExpiringCacheMap.java
@@ -1,0 +1,181 @@
+/**
+ * Copyright (c) 2014-2017 by the respective copyright holders.
+ * All rights reserved. This program and the accompanying materials
+ * are made available under the terms of the Eclipse Public License v1.0
+ * which accompanies this distribution, and is available at
+ * http://www.eclipse.org/legal/epl-v10.html
+ */
+package org.eclipse.smarthome.core.cache;
+
+import java.util.Collection;
+import java.util.LinkedHashSet;
+import java.util.LinkedList;
+import java.util.Set;
+import java.util.concurrent.ConcurrentHashMap;
+import java.util.concurrent.ConcurrentMap;
+import java.util.function.Supplier;
+
+import org.slf4j.Logger;
+import org.slf4j.LoggerFactory;
+
+/**
+ * This is a simple expiring and reloading multiple key-value-pair cache implementation. The value expires after the
+ * specified duration has passed since the item was created, or the most recent replacement of the value.
+ *
+ * @author Christoph Weitkamp - Initial contribution and API.
+ *
+ * @param <K> the type of the key
+ * @param <V> the type of the value
+ */
+public class ExpiringCacheMap<K, V> {
+
+    private final Logger logger = LoggerFactory.getLogger(ExpiringCacheMap.class);
+
+    private final long expiry;
+    private final ConcurrentMap<K, ExpiringCache<V>> items;
+
+    /**
+     * Creates a new instance.
+     * 
+     * @param expiry the duration in milliseconds for how long the value stays valid
+     */
+    public ExpiringCacheMap(long expiry) {
+        this.expiry = expiry;
+        this.items = new ConcurrentHashMap<K, ExpiringCache<V>>();
+    }
+
+    /**
+     * Creates an {@link ExpiringCache} and adds it to the cache.
+     * 
+     * @param key the key with which the specified value is to be associated
+     * @param action the action for the item to be associated with the specified key to retrieve/calculate the value
+     */
+    public void put(K key, Supplier<V> action) {
+        put(key, new ExpiringCache<V>(expiry, action));
+    }
+
+    /**
+     * Adds an {@link ExpiringCache} to the cache.
+     * 
+     * @param key the key with which the specified value is to be associated
+     * @param item the item to be associated with the specified key
+     */
+    public void put(K key, ExpiringCache<V> item) {
+        if (key == null) {
+            throw new IllegalArgumentException("Item cannot be added as key is null.");
+        }
+        if (item == null) {
+            throw new IllegalArgumentException("Item cannot be added as item is null.");
+        }
+
+        items.put(key, item);
+    }
+
+    /**
+     * Removes the item associated with the given key from the cache.
+     * 
+     * @param key the key whose associated value is to be removed
+     */
+    public void remove(K key) {
+        items.remove(key);
+    }
+
+    /**
+     * Discards all items from the cache.
+     */
+    public void clear() {
+        items.clear();
+    }
+
+    /**
+     * Returns a set of all keys.
+     * 
+     * @return the set of all keys
+     */
+    public synchronized Set<K> keys() {
+        final Set<K> keys = new LinkedHashSet<K>();
+        for (final K key : items.keySet()) {
+            keys.add(key);
+        }
+        return keys;
+    }
+
+    /**
+     * Returns the value associated with the given key - possibly from the cache, if it is still valid.
+     * 
+     * @param key the key whose associated value is to be returned
+     * @return the value associated with the given key, or null if there is no cached value for the given key
+     */
+    public V get(K key) {
+        final ExpiringCache<V> item = items.get(key);
+        if (item == null) {
+            logger.debug("No item for key '{}' found", key);
+            return null;
+        } else {
+            return item.getValue();
+        }
+    }
+
+    /**
+     * Returns a collection of all values - possibly from the cache, if they are still valid.
+     * 
+     * @return the collection of all values
+     */
+    public synchronized Collection<V> values() {
+        final Collection<V> values = new LinkedList<V>();
+        for (final ExpiringCache<V> item : items.values()) {
+            values.add(item.getValue());
+        }
+        return values;
+    }
+
+    /**
+     * Invalidates the value associated with the given key in the cache.
+     * 
+     * @param key the key whose associated value is to be invalidated
+     */
+    public synchronized void invalidate(K key) {
+        final ExpiringCache<V> item = items.get(key);
+        if (item == null) {
+            logger.debug("No item for key '{}' found", key);
+        } else {
+            item.invalidateValue();
+        }
+    }
+
+    /**
+     * Invalidates all values in the cache.
+     */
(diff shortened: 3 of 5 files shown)


cweitkamp:
This is the initial version. Next steps / ideas:
- Replace `ExpiringCache` from Yahoo Weather Binding
- Switch from single key-value-pair to multiple key-value-pairs
- Replace Interface `LoadAction` with Interface `Function` ([java.util.function](https://docs.oracle.com/javase/8/docs/api/java/util/function/Function.html))

wdyt?

kaikreuzer:
Looks great, thanks!

> Replace Interface LoadAction with Interface Function (java.util.function)

As we are on Java8, that would imho make a lot of sense. As it changes the API of that feature, it might be best if your directly change it within this PR, so it becomes directly the API to use, wdyt?

> Replace ExpiringCache from Yahoo Weather Binding

Yep, should be a follow-up PR soon!

> Switch from single key-value-pair to multiple key-value-pairs

What exactly is your idea here?

bundles/core/org.eclipse.smarthome.core.test/src/test/java/org/eclipse/smarthome/core/cache/ExpiringCacheTest.java
```diff
@@ -0,0 +1,108 @@
...
+/**
+ * Test class for the {@link ExpiringCache} class.
+ *
+ * @author Christoph Weitkamp - Initial contribution.
+ */
+public class ExpiringCacheTest {
+
+    private final Logger logger = LoggerFactory.getLogger(ExpiringCacheTest.class);
+
+    private final int CACHE_EXPIRY = 2 * 1000; // 2s
```
kaikreuzer: add `static`

bundles/core/org.eclipse.smarthome.core.test/src/test/java/org/eclipse/smarthome/core/cache/ExpiringCacheTest.java
```diff
@@ -0,0 +1,108 @@
...
+
+    private final String TEST_QUERY = "TEST_QUERY";
```
kaikreuzer: add `static`

bundles/core/org.eclipse.smarthome.core.test/src/test/java/org/eclipse/smarthome/core/cache/ExpiringCacheTest.java
```diff
@@ -0,0 +1,108 @@
...
+
+    @Test(expected = IllegalArgumentException.class)
+    public void testConstructorIllegalArgumentException() throws IllegalArgumentException {
+        new ExpiringCache(0, null);
+    }
+
+    @Test(expected = IllegalArgumentException.class)
+    public void testGetIllegalArgumentException() throws IllegalArgumentException, IOException {
+        final ExpiringCache<String, String> CACHE = new ExpiringCache<String, String>(CACHE_EXPIRY,
```
kaikreuzer: as this isn't static, you should use lower case `cache`

bundles/core/org.eclipse.smarthome.core.test/src/test/java/org/eclipse/smarthome/core/cache/ExpiringCacheTest.java
```diff
@@ -0,0 +1,108 @@
...
+                        return RandomStringUtils.random(8);
+                    }
+                });
+
+        CACHE.get(null);
+    }
+
+    @Test
+    public void testGet() throws IOException {
+        final ExpiringCache<String, String> CACHE = new ExpiringCache<String, String>(CACHE_EXPIRY,
```
kaikreuzer: as this isn't static, you should use lower case `cache`

bundles/core/org.eclipse.smarthome.core.test/src/test/java/org/eclipse/smarthome/core/cache/ExpiringCacheTest.java
```diff
@@ -0,0 +1,108 @@
...
+        assertEquals(value1, value2);
+
+        // use a different query
+        value2 = CACHE.get("OTHER_TEST_QUERY");
+        assertNotEquals(value1, value2);
+    }
+
+    @Test
+    public void testClear() throws IOException {
+        final ExpiringCache<String, String> CACHE = new ExpiringCache<String, String>(CACHE_EXPIRY,
```
kaikreuzer: as this isn't static, you should use lower case `cache`

bundles/core/org.eclipse.smarthome.core.test/src/test/java/org/eclipse/smarthome/core/cache/ExpiringCacheTest.java
```diff
@@ -0,0 +1,108 @@
...
+        // clear cache
+        CACHE.clear();
+
+        String value2 = CACHE.get(TEST_QUERY);
+        assertNotEquals(value1, value2);
+    }
+
+    @Test
+    public void testExpired() throws InterruptedException, IOException {
+        final ExpiringCache<String, String> CACHE = new ExpiringCache<String, String>(CACHE_EXPIRY,
```
kaikreuzer: as this isn't static, you should use lower case `cache`

bundles/core/org.eclipse.smarthome.core/src/main/java/org/eclipse/smarthome/core/cache/ExpiringCache.java
```diff
@@ -0,0 +1,89 @@
...
+     */
+    public synchronized V get(Q key) throws IOException {
+        if (key == null) {
+            throw new IllegalArgumentException("Value cannot be retrieved/calculated as key is null.");
+        }
+
+        if (lastQuery == null || !lastQuery.equals(key) || isExpired()) {
+            lastValue = action.load(key);
+            lastQuery = key;
+            lastUpdate = System.currentTimeMillis();
```
kaikreuzer: How about [using nanoTime](http://efectivejava.blogspot.de/2013/09/why-to-use-systemnanotime-instead-of.html) to make sure we are not impacted by adjusted system clocks?

cweitkamp: I already realized it in push 1fc32d0.

cweitkamp:
> As we are on Java8, that would imho make a lot of sense. As it changes the API of that feature, it might be best if your directly change it within this PR, so it becomes directly the API to use, wdyt?

Yes of course. IMHO it simplifies the code. I already realized it in push 1fc32d0.

> What exactly is your idea here?

My first guess was to add a `Map` inside the `ExpiringCache` class to handle multiple key-value-pairs. This method needs to introduce a `ExpiringCacheItem` which stores the needed properties (`expiry`, `action`, `query`, etc.).

kaikreuzer:
> Yes of course. IMHO it simplifies the code. I already realized it in push 1fc32d0.

Thanks, makes it indeed nicer.

> handle multiple key-value-pairs

This is to avoid requiring people to setup multiple cache instances, right? I am not sure whether we really need it - I would postpone adding such a feature until we really have someone requesting it for a good reason.

One last question: You chose `smarthome.core.cache` as the package name, while I had suggested putting it into `smarthome.core.common`. I am a bit concerned that we might end up with too many different packages some time; for stuff that has no dependencies to other libs and consists out of a single class, I would think we could put them into `common`.
But this is probably rather a question to @maggu2810 and @sjka on how we want to structure the core API in general.

bundles/core/org.eclipse.smarthome.core/src/main/java/org/eclipse/smarthome/core/cache/ExpiringCache.java
```diff
@@ -0,0 +1,85 @@
...
+     * @param expiry duration in milliseconds for how long the value stays valid
+     * @param action the action to retrieve/calculate the value
+     */
+    public ExpiringCache(long expiry, Function<Q, V> action) {
+        if (action == null) {
+            throw new IllegalArgumentException(
+                    "ExpiringCache cannot be created as action is not an instance of Function.");
+        }
+
+        this.expiry = expiry * 1000000;
```
maggu2810: It is exactly the same, but to improve readability, perhaps we could consider to use `TimeUnit.MILLISECONDS .toNanos(expiry)`

bundles/core/org.eclipse.smarthome.core/src/main/java/org/eclipse/smarthome/core/cache/ExpiringCache.java
```diff
@@ -0,0 +1,85 @@
...
+    /**
+     * Create a new instance.
+     *
+     * @param expiry duration in milliseconds for how long the value stays valid
+     * @param action the action to retrieve/calculate the value
+     */
+    public ExpiringCache(long expiry, Function<Q, V> action) {
+        if (action == null) {
+            throw new IllegalArgumentException(
+                    "ExpiringCache cannot be created as action is not an instance of Function.");
```
maggu2810: Hm, why not log what the check is exactly about e.g. `ExpiringCache cannot be created using a null action`.

Sure, we could state that if action is a function (correct argument type), then the only possibility is a null one...

bundles/core/org.eclipse.smarthome.core/src/main/java/org/eclipse/smarthome/core/cache/ExpiringCache.java
```diff
@@ -0,0 +1,85 @@
...
+            lastUpdate = System.nanoTime();
+        }
+        return lastValue;
+    }
+
+    /**
+     * Discards the value in the cache.
+     */
+    public void clear() {
+        lastQuery = null;
```
maggu2810: Shouldn't this be synchronized using the same synchronization object as the `get` method?

maggu2810:
>> handle multiple key-value-pairs

> This is to avoid requiring people to setup multiple cache instances, right? I am not sure whether we really need it - I would postpone adding such a feature until we really have someone requesting it for a good reason.

The current implementation does not really make sense if you request all the time two keys from the cache. The key will differ on every request so no cache is used at all. This detail should be hidden by the implementation as I don't think a cache makes sense if the caller does need to know how to request values to get a good performance.
So I would prefer to add a better cache implementation.

maggu2810:
> One last question: You chose `smarthome.core.cache` as the package name, while I had suggested putting it into `smarthome.core.common`. I am a bit concerned that we might end up with too many different packages some time; for stuff that has no dependencies to other libs and consists out of a single class, I would think we could put them into `common`.

Why do you think that many different packages are a bad idea?
Do you see any performance impacts using different packages?

For me it makes sense to use `cache` in the namespace for all `cache` stuff. If there is (currently) only one class, it makes sense for me, too.

Some times we have packages like `common` or `utils` that contains a lot of unrelated classes and this could be split only by breaking all the imports used by consumers.
Perhaps we could use think about using `smarthome.core.utils.cache` or something similar. So the single class is still in a utils sub-package and grouped by "category" or "semantic".

bundles/core/org.eclipse.smarthome.core/src/main/java/org/eclipse/smarthome/core/cache/ExpiringCache.java
cweitkamp: I did that in my first version. Changed it back again.

cweitkamp:
> One last question: You chose `smarthome.core.cache` as the package name, while I had suggested putting it into `smarthome.core.common`.

@kaikreuzer To be fair: You also mentioned to 

> feel free to come up with any better suggestions

So I did. ;-)

I am going with @maggu2810 . I like to set up an easy to use structure and like grouping depending features together in one package. If that will be `smarthome.core.cache` or `smarthome.core.utils.cache` is up to you. Just let me know, I will change it.

> This is to avoid requiring people to setup multiple cache instances, right? I am not sure whether we really need it - I would postpone adding such a feature until we really have someone requesting it for a good reason.

Yes, that is one point. Without thinking long I have two bindings in my head which could benefit immediately. First is the OH1 [http binding](https://github.com/openhab/openhab1-addons/tree/master/bundles/binding/org.openhab.binding.http) which already provides its own methods to cache multiple requests, second is the OH1 [weather binding](https://github.com/openhab/openhab1-addons/tree/master/bundles/binding/org.openhab.binding.weather) which supports idependant weather providers.

> So I would prefer to add a better cache implementation.

I appended a very quick draft of my improved approach with multi key-value-pair I have in my head. As this approach is completely different from the initial implementation it changes the API in the same manner. But it may be the base for any other extension for any other caching method you could imagine. wdyt?
cache

maggu2810:
> I appended a very quick draft of my improved approach with multi key-value-pair I have in my head

If I understand your approach correctly it is the same as I have in mind after reading your PR.

Your current `ExpiringCache` implementation could be moved to an `ExpiringCacheItem` (I named it `ExpiringSingleCache` in my mind this morning) with a fixed key and so an action without argument (the return value fits to the value of the key as before) -- so `Supplier` instead of `Function`.

The `ExpiringCache` is one of the above object per key (e.g. `addItem(String key, Supplier<T> action)` that internally creates an instance of the above object using the expiring timeout as the whole `ExpiringCache` -- or `addItem(ExpringCacheItem cacheItem)` that fetches the informations from the object itself).

So, IMHO the real implementation could be discussed, which one is the best approach, but IMHO we should support the multi-key-value approach.

sjka:
First of all: Thanks for taking care of this, your ideas for improvements all fully make sense!

> I agree with @cweitkamp that we should support the multi-key-value approach

Me too! When I first implemented it, I simply wanted to keep it simple as the yahoo bindings simply didn't need it, so it just felt wrong to implement more than the absolute minimum. However, I do agree that other bindings would benefit from it, so I would support extending the API to support it.

> with a fixed key and so an action without argument (the return value fits to the value of the key as before) -- so Supplier instead of Function

This is a pretty fundamental design question: Will the cache contain related objects which share the way of they get refreshed (i.e. same function with the key as an input), or does the cache mainly focus on the expiry feature of completely unrelated stuff (i.e. ExpringCacheItem which only knows how to refresh itself)? I'd assume that if you have unrelated stuff it is also acceptable to have different instances of the ExpiryCache, as it does not carry a lot of overhead?

kaikreuzer:
> To be fair: You also mentioned to "feel free to come up with any better suggestions"
> So I did. ;-)

Absolutely right, I only wanted to have an explicit discussion about the decision, which has happened now and I am fully ok with @maggu2810 and your opinion.

> Will the cache contain related objects which share the way of they get refreshed 

As @cweitkamp mentions above, we will have such examples like e.g. http and weather binding - and also the [SleepIQ binding](https://github.com/openhab/openhab2-addons/pull/2009#discussion_r118821084) (correct, @syphr42?) will imho require it.

maggu2810:
>> with a fixed key and so an action without argument (the return value fits to the value of the key as before) -- so Supplier instead of Function

> This is a pretty fundamental design question: Will the cache contain related objects which share the way of they get refreshed (i.e. same function with the key as an input),

Really? Using lambda expression there are several ways to handle it.
We can use functions all the time and the called function could ignore the argument if it is used for one key only.
Or we can use actions in the cache item and wrap it by a lambda.
Now, you can argue that "you don't want to use lambdas if the user of a cache item wants to use the same function" or you can argue that "a cache item uses one supplier to refresh its value (of its single key)".

```java
cache.addItem("key1", () -> function("key1"));
cache.addItem("key2", () -> function("key2"));
```

cweitkamp:
Hey all,

I pushed an initial version of the multi-key-value approach to my personal fork. You may have a look at it if you want: [#8ed1bd8](https://github.com/cweitkamp/smarthome/commit/8ed1bd82888c19a045bb455f25a14c4e52aae530). Shall I push it to this PR or create a new PR for it?

The `ExpiringCacheItem` itself only takes care of the time it expires an calls the assigned action to refresh its value - at the moment it gets accessed. I don't think the `ExpiringCacheItem` needs to store any meta information on how it will be refreshed. The way how to refresh the value is completely put into the hands of the developer. So IMHO a lambda to retreive/calculate the value meets the requirements.

The `ExpiringCache` handles the mapping of keys to cached items. There are some methods for maintenance of the cache (e.g. if you want to clean up unnecessary `ExpiringCacheItem`).

The more I am working on it the more ideas for improvements come to my mind.

syphr42:
>> Will the cache contain related objects which share the way of they get refreshed

>As @cweitkamp mentions above, we will have such examples like e.g. http and weather binding - and also the SleepIQ binding (correct, @syphr42?) will imho require it.

The SleepIQ API is an all-at-once type of status retrieve, so there will likely be a single key for that use case with a single associated value. The refresh function will get everything in one shot.

kaikreuzer:
>  Shall I push it to this PR or create a new PR for it?

Afaics, this changes the API of the cache (and thus it does NOT support the "simple" usage that it offers in this PR here). A new (follow-up) PR would imho only make sense if it keeps existing functionality in place. Is there a chance to refactor the multi-key-support in a way that you can also still use the cache with the simple style, i.e. passing the function directly to the constructor and not having to call `add`?

If so, we could merge this PR here asap (it would e.g. fully suffice for @syphr42 already) and continue the discussion about multi-key-support on a new PR.

maggu2810:
> Afaics, this changes the API of the cache (and thus it does NOT support the "simple" usage that it offers in this PR here).

Support for multiple keys and only cache the last value of the last used key is not anything that needs to be supported at all.
Using one `ExpiringCacheItem` for one key and another `ExpiringCacheItem` for another key (so, not using the add function but the constructor) should be possible.

If we introduce a cache in the core API we should go with a good implementation from the beginning and not push something and change it later.

Why not force push the changes @cweitkamp already implemented and just apply further cosmetic changes if there are any needed.
Then we have a single key value cache item and a cache that holds multiple ones.

kaikreuzer:
> Why not force push the changes @cweitkamp already implemented and just apply further cosmetic changes if there are any needed.

Because it implies a more complicated API than is necessary for 80% of the use cases. Or do I miss something in how it is used?

cweitkamp:
I just pushed the newest version [#97eb0e3](https://github.com/cweitkamp/smarthome/commit/97eb0e3f959f8bd508b06fd8ce7f2526efa0fb5b) in my fork.

> Or do I miss something in how it is used?

For clarification:

The class `ExpiringCacheItem` provides - except for a few changes - the same functionality like the version implemented in the Yahoo binding. You can feel free to use it as single key-value-pair instance (see [unit test](https://github.com/cweitkamp/smarthome/commit/97eb0e3f959f8bd508b06fd8ce7f2526efa0fb5b#diff-c13bbb498166549857c631743a8d1ada)).

The class `ExpiringCache` is - roughly said - only a wrapper containing a Map for better maintenance of multiple key-value-pairs.

maggu2810:
> a more complicated API than is necessary for 80% of the use cases

I really don't see any complicated API in @cweitkamp's implementation

kaikreuzer:
> The class ExpiringCacheItem provides the same functionality like the version implemented in the Yahoo binding

Ah, ok, I see. My only issue is that I wouldn't ever expect to use a class called ExpiringCacheItem if I need a cache implementation for a value.

> The class ExpiringCache is only a wrapper containing a Map

Maybe we could improve the names then? How about `ExpiringCache` and `ExpiringCacheMap` (with a put instead of add method)?

maggu2810:
> Maybe we could improve the names then? How about ExpiringCache and ExpiringCacheMap (with a put instead of add method)?

Or `ExpiringSingleCache` and `ExpiringMultiCache`?

But also your ones are okay for me.

cweitkamp:
> How about ExpiringCache and ExpiringCacheMap (with a put instead of add method)?

I give this suggestion a try and pushed the multiple key-value-pair version into this PR.

bundles/core/org.eclipse.smarthome.core/src/main/java/org/eclipse/smarthome/core/cache/ExpiringCache.java
```diff
@@ -0,0 +1,91 @@
...
+    }
+
+    /**
+     * Returns the value - possibly from the cache, if it is still valid.
+     * 
+     * @return the value
+     */
+    public synchronized V getValue() {
+        if (value == null || isExpired()) {
+            refreshValue();
```
maggu2810: You can use `return refreshValue()`.

bundles/core/org.eclipse.smarthome.core/src/main/java/org/eclipse/smarthome/core/cache/ExpiringCacheMap.java
```diff
@@ -0,0 +1,185 @@
...
+ *
+ * @param <K> the type of the key
+ * @param <V> the type of the value
+ */
+public class ExpiringCacheMap<K, V> {
+
+    private final Logger logger = LoggerFactory.getLogger(ExpiringCacheMap.class);
+
+    private final long expiry;
+    protected ConcurrentMap<K, ExpiringCache<V>> items;
```
maggu2810: can be `final`

maggu2810: Why `protected` instead of `private`?

bundles/core/org.eclipse.smarthome.core/src/main/java/org/eclipse/smarthome/core/cache/ExpiringCacheMap.java
```diff
@@ -0,0 +1,185 @@
...
+        if (key == null) {
+            throw new IllegalArgumentException("Item cannot be added as key is null.");
+        }
+        if (item == null) {
+            throw new IllegalArgumentException("Item cannot be added as item is null.");
+        }
+
+        items.put(key, item);
+
+        return item.getValue();
```
maggu2810: I don't know if the `put` method should trigger a cache refresh (that could be slow).

bundles/core/org.eclipse.smarthome.core/src/main/java/org/eclipse/smarthome/core/cache/ExpiringCacheMap.java
cweitkamp: Was a superclass before, forget to change it.

cweitkamp:
@maggu2810 Thanks for your review, I applied your changes.

kaikreuzer:
@cweitkamp As a follow-up, it would probably make sense to adapt the YahooWeather binding to use this implementation now :-)

hakan42:
@kaikreuzer , We will need another stable build of ESH before we can use this in OH2 bindings, right? I would like to switch the senseBox binding as well...

kaikreuzer:
@hakan42 Absolutely correct and just done: https://hudson.eclipse.org/smarthome/job/SmartHomeDistribution-Stable/203/

So simply reload your openhab2.target platform and you should have it available!
@syphr42 Same for you: You can go ahead with it in the SleepIQ binding!

cweitkamp:
@maggu2810 @kaikreuzer No problem, you're welcome. Thank you, too.

> As a follow-up, it would probably make sense to adapt the YahooWeather binding to use this implementation now.

Of course. I like to do that.

PR #3583, opened 2017-06-04
Added a simple setup for a simple math transformations

martinvw:
Added a simple setup for a simple math transformations

To simpify things, a MINUS and DIVIDE could of course be added

See also 'discussion' at:

https://community.openhab.org/t/would-a-multiply-math-transformation-make-sense/29425/3


martinvw:
Hereby a jar to test:

[org.eclipse.smarthome.transform.simplemath-0.9.0-SNAPSHOT.jar.zip](https://github.com/eclipse/smarthome/files/1050710/org.eclipse.smarthome.transform.simplemath-0.9.0-SNAPSHOT.jar.zip)

I used it in the demo items like this and it worked, I could not format the value as well, because it seemed to conflict:

```
Group:Contact:OR(OPEN, CLOSED)  Windows     "Open windows [MULTIPLY(0.001):%d]"               <contact>
```

martinvw:
Did some changes now I can handle formatting as well:

```
Group:Contact:OR(OPEN, CLOSED)  Windows "Open windows [MULTIPLY(0.001):(%d watt)]" <contact>
```

[org.eclipse.smarthome.transform.simplemath-0.9.0-SNAPSHOT.jar.zip](https://github.com/eclipse/smarthome/files/1050737/org.eclipse.smarthome.transform.simplemath-0.9.0-SNAPSHOT.jar.zip)

martinvw:
Some small test failure with BigDecimals :-)

```
Failed tests: 
  MultiplyTransformationServiceTest.testTransformInsideString:32 expected:<10[] watt> but was:<10[.0] watt>
```

Working on it.

sjka:
If I understand it correctly, your main driver for this PR is to have a way to convert units. And of course I fully agree this is important and completely makes sense. However, I see to major drawbacks:

* It addresses the problem on the wrong level, i.e. will be respected in UIs only. The persistence layer stores the raw states, so you will have different values than in your database than what you see in the UIs. This is at least confusing.
* There is #601, which aims at teaching the framework to deal with units properly.

So while this certainly is a valid approach to ease the pain until #601 got resolved at least partially, I'm not so sure whether we should add it to ESH at this point in time where we are about to have something much better for solving this problem, or rather should wait for it and them implement unit transformations instead? What other use-cases apart from unit conversion are you aware of?

martinvw:
> It addresses the problem on the wrong level, i.e. will be respected in UIs only

I indeed realized already :-(

> What other use-cases apart from unit conversion are you aware of?

Correcting for sensor offset, but you would also want to that one state earlier.

martinvw:
@sjka maybe the community can not come up with other use cases, should we close it or leave it open for others, wdyt?

sjka:
> Correcting for sensor offset, but you would also want to that one state earlier.

As this is the only one we can come up with and we agree that it's the wrong level, I'd be with you that we should rather close it. It will remain here, so we can re-open it/pick it up anytime once another use-case pops up.

niebochod:
Another use case is to use it together with modbus binding where some devices store values that requires multiplication by an factor as shown on:
https://docs.openhab.org/addons/bindings/modbus1/readme.html
So far solution is JS transformation, but it's not very convenient.

PR #3587, opened 2017-06-06
Implemented unique geolocation for ESH installation

### LINKED ISSUE #1694: Unify geolocation
Several bindings require a [geolocation](https://en.wikipedia.org/wiki/Geolocation) to work. Some require the system's location (Astro) while others require any (nearby) location (Weather).

Instead of providing the same latitude and longitude (and altitude etc.) over and over again, I was thinking if this could not be set globally once for the entire installation.

I have come up with these options:
- Make these a setting that can be managed through Paper UI (much like locale)
  - manually enter lat/lon
  - compute at request through the browser (like many geo aware websites do these days)
- Provide it through some `.cfg` file
- Compute in real time (e.g. at system start) so that if I move my installation the new geolocation is automatically determined and available for use.
- Create a binding that determines this

The bindings or other parts of the system (Eclipse Smarthome or openHAB) could then optionally use this setting. For example,
- if set, force its use or
- use it as a default if not explicitly overwritten through the binding configuration
  What are your thoughts?

(This idea started as [a thread in the openHAB community forum](https://community.openhab.org/t/idea-unify-geolocation/11669))

triller-telekom:
Fixes #1694


bundles/core/org.eclipse.smarthome.core/src/main/java/org/eclipse/smarthome/core/i18n/LocaleProvider.java
```diff
@@ -25,4 +28,10 @@
      */
     Locale getLocale();
 
+    /**
+     * Provides access to the location of the installation
+     *
+     * @return location of the current installation
+     */
+    PointType getLocation();
```
maggu2810: If I understand your code correctly this method could also return `null`. You should add this information to the JavaDoc e.g. `@return location of the current installation or null if the location is not set`.

bundles/core/org.eclipse.smarthome.core/src/main/java/org/eclipse/smarthome/core/internal/i18n/LocaleProviderImpl.java
```diff
@@ -41,6 +45,11 @@ protected synchronized void modified(Map<String, Object> config) {
         final String script = (String) config.get(SCRIPT);
         final String region = (String) config.get(REGION);
         final String variant = (String) config.get(VARIANT);
+        final String location = (String) config.get(LOCATION);
+
+        if (location != null) {
+            this.location = PointType.valueOf(location);
```
maggu2810: The location is configured as a string. If the user provides an incorrect formatted string the `valueOf` method throw an `IllegalArgumentException`. What is the expectation of the modified handling? Shouldn't we unset the location (set to `null`) or keep the old value instead of stopping the modified function handling for all other values caused by the non-catch exception?

triller-telekom: I agree, indeed its a problem if the user provides a mal-formed String. I will catch the `IllegalArgumentException` and keep the old value or `null`, i.e. whatever was set before the wrongly entered information.

bundles/core/org.eclipse.smarthome.core/src/main/java/org/eclipse/smarthome/core/i18n/LocaleProvider.java
```diff
@@ -25,4 +28,10 @@
      */
     Locale getLocale();
 
+    /**
+     * Provides access to the location of the installation
+     *
+     * @return location of the current installation or null if the location is not set
+     */
+    PointType getLocation();
```
martinvw: Maybe I miss some context but it seems unexpected that a `LocaleProvider` in package `i18n` does something with a location instead of strictly with locals maybe some bigger refactoring is needed to make this actually a `ConfigurationProvider` in a more generic package.

martinvw: I see there was some discussion in the issue, but just my two cents.

kaikreuzer: `ConfigurationProvider` would definitely be too broad (that actually already exists as the OSGi Configuration Admin) and having a new LocationProvider for a single value might not be worth the effort, especially as Locale and Location seem to have something in common, if you look at the word, right? ;-)

martinvw: That is quite some fuzzy matching then ;-P

And imho it has nothing to do with `i18n`

kaikreuzer: Well, in most cases there is a close relation of the locale that you use and the location you are at. Actually, I also wish the feature to derive a sensible default locale from the initially set location.

triller-telekom: @martinvw I agree that `i18n` might be a bit misleading but our idea was to put it into the `LocaleProvider` so if the user sets up a location but no `locale` we could derive a `locale` based on the position. Such a logic can be implemented a lot easier than if we separate them into two services.

martinvw: It is just my opinion, I don't say they should not live together but the file and package name could be renamed to reflect their new goal and purpose. If all others do not agree on that, don't bother :-D

PS: note that travis is failing :-)

kaikreuzer: > but the file and package name could be renamed to reflect their new goal and purpose

Any suggestion what that could be...?

martinvw: Could have seen that one coming ;-)

And I do agree it is a though one*

## Concepts

Okay we have the following concepts:

**i18n**: 
> Internationalization is the design and development of a product, application or document content that enables easy localization for target audiences that vary in culture, region, or language.

[Source](https://www.w3.org/International/questions/qa-i18n)

**Locale** 
> A Locale object represents a specific geographical, political, or cultural region. 

[Source](https://docs.oracle.com/javase/7/docs/api/java/util/Locale.html)

**Location** ;-)
> An absolute location is designated using a specific pairing of latitude and longitude in a Cartesian coordinate grid

[Source](https://en.wikipedia.org/wiki/Location_(geography))

## First Option

Given that maybe providing a Location is a form of *i18n* but its not a locale, so for example

`i18n.InternationalizationProvider`

## Alternative

An alternative could be: **globalization** 

> Some companies, like IBM and Sun Microsystems, use the term "globalization", g11n, for the combination of internationalization and localization.[6] Also known as "glocalization" (a portmanteau of globalization and localization).

[source](https://en.wikipedia.org/wiki/Internationalization_and_localization)

globalization.GlobalizationProvider

\* Maybe that actually means they are not closely enough related, but I can understand this choice.

kaikreuzer: Ok, so I agree that i18n is the more general definition and thus a better fit than l10n.

Regarding `i18n.InternationalizationProvider`: Note that there IS already a `i18n.I18nProvider` - which is used everywhere for translating texts. This provider already has a dependency on `LocaleProvider`. I wonder if we should merge the code of `LocaleProvider` into the `I18nProvider` and hence also add our new location there?
I would prefer having these features grouped in a single service and not having i18n, l10n and location as three different services that all depend on one another.
@maggu2810 As you introduced the LocaleProvider at the time - what's your opinion here?

triller-telekom:
Regarding the Travis fail:
```
Running org.eclipse.smarthome.automation.module.script.ScriptRuleTest
Tests run: 1, Failures: 1, Errors: 0, Skipped: 0, Time elapsed: 10.958 sec <<< FAILURE! - in org.eclipse.smarthome.automation.module.script.ScriptRuleTest
testPredefinedRule(org.eclipse.smarthome.automation.module.script.ScriptRuleTest)  Time elapsed: 10.41 sec  <<< FAILURE!
java.lang.AssertionError: 
Expected: is <false>
     but: was <true>
	at org.junit.Assert.assertThat(Assert.java:832)
	at org.junit.Assert$assertThat.callStatic(Unknown Source)
	at org.eclipse.smarthome.automation.module.script.ScriptRuleTest$_testPredefinedRule_closure8.doCall(ScriptRuleTest.groovy:95)
```

This looks completely unrelated to this change. At the beginning of this test there is a check that the rule registry is empty, which fails
```
assertThat ruleRegistry.getAll().isEmpty(), is(false)
```

bundles/core/org.eclipse.smarthome.core/src/main/java/org/eclipse/smarthome/core/i18n/LocaleProvider.java
maggu2810: Normally I would prefer different services / providers for different stuff.

If we find a good name that combines Location and Locale I agree that this would be simpler to merge into one service instead of having two (as every service will provide only one function).

But I don't think we should merge `I18nProvider` and the `LocaleProvider` because the I18nProvider offers functions to localize text. The `LocaleProvider` itself holds a configuration.

kaikreuzer: > If we find a good name

Hm, I18N seemed to be the very best choice, but this is already taken :-/

> But I don't think we should merge I18nProvider and the LocaleProvider because the I18nProvider offers functions to localize text. The LocaleProvider itself holds a configuration.

Hm, maybe we will also offer services around locale&location later on in that service, like e.g. formatting numbers/strings according to a locale, getting a local currency or whatever...
Why should a configuration be necessarily be separated from functionality?

martinvw: Why not `I18nProvider` & `I18nConfigurationProvider` seems like the most clear way to separate the behavior from the configuration

maggu2810: > Why should a configuration be necessarily be separated from functionality?
￼
There is IMHO not a "it should" or "it must".

But if you need a location I would like an LocationProvider instance, if I need a locale I would like an LocationProvider intance and if I need to translate some stuff I would like an `I18nProvider`.

So, it is not about that the implementation could be done in one class but I don't need an implementation of a transaltion mechanism just because I would like to know the location or locale of a system.

IMHO that is the intention of interfaces to define a subset of stuff that should be combined and not to generate an interface that defines all the stuff that could ever be used or implemented by someone.

martinvw: > So, it is not about that the implementation could be done in one class but I don't need an implementation of a transaltion mechanism just because I would like to know the location or locale of a system.

That is in fact the core of it all. Maybe we should split up the interface in two parts (LocaleProvider && LocationProvider) but keep the implementation of the configuration of I18n related stuff together.

kaikreuzer: >  if I need to translate some stuff I would like an I18nProvider.

No, definitely not - you would like a TranslationProvider. "i18n" is a wide term, which covers locale (and possibly location). So having it in place, but ONLY providing localised (!) texts, is weird.

So the decision had already been taken a long time ago when the I18nProvider has been introduced: The "set of stuff that should be combined" is the functionality around i18n - that's why it is most logical to me to merge the implementations.

Another consideration has to be what is "configured together", i.e. what should have an entry in ConfigAdmin and also appear as a single service e.g. in the Paper UI Configuration->Services section. Here as well, I think it makes sense to have those settings together in an "i18n" section. Splitting it in smaller pieces will only make it harder to find stuff.

maggu2810: > No, definitely not - you would like a TranslationProvider. "i18n" is a wide term, which covers locale (and possibly location). So having it in place, but ONLY providing localised (!) texts, is weird.

I agree. `TranslationProvider` should be the name of the current `I18nProvider`.
But I think renaming the `i18nProvider` interface name should not be done in this PR at all.

> So the decision had already been taken a long time ago when the I18nProvider has been introduced

At that time there has been no location and no locale provider at all.

So, I don't see the real "argument" here. It is the same (but I don't think it is one) argument as:
There is currently a `I18nProvider` and a `LocaleProvider`, so the decision to keep them separated has been already taken.

martinvw: _Sorry for starting this discussion_ :-)

maggu2810: > Another consideration has to be what is "configured together", i.e. what should have an entry in ConfigAdmin and also appear as a single service e.g. in the Paper UI Configuration->Services section. Here as well, I think it makes sense to have those settings together in an "i18n" section. Splitting it in smaller pieces will only make it harder to find stuff.

Hm, isn't the configuration winded to a service.pid and the service pid is choosen by the implementation?
Can't we have different provider interfaces and one implementation that uses one configuration?

E.g. `I18nProviderImpl` implements the `LocaleProvider`, `TranslationProvider` and `LocationProvider` interface and is configured using the service pid `org.eclipse.smarthome.i18n`?

maggu2810: > Sorry for starting this discussion :-)

Hopefully we never stop to discuss the design of the software / framework.

kaikreuzer: > E.g. I18nProviderImpl implements the LocaleProvider, TranslationProvider and LocationProvider interface and is configured using the service pid org.eclipse.smarthome.i18n?

Sounds like a reasonable compromise. I am just afraid that we cannot easily change `I18nProvider` to `TranslationProvider` as it is used almost everywhere. Would you dare such a refactoring?

> Hopefully we never stop to discuss the design of the software / framework.

Small English lesson: "never stop to discuss" and "never stop discussing" have a very different meaning - and I hope you actually meant the second one!

maggu2810: > I am just afraid that we cannot easily change `I18nProvider` to `TranslationProvider` as it is used almost everywhere. Would you dare such a refactoring?

It has been pretty easy, see https://github.com/eclipse/smarthome/pull/3615
The local build succeeded (it took the most time).

> Small English lesson: "never stop to discuss" and "never stop discussing" have a very different meaning - and I hope you actually meant the second one!

Sorry, I need to learn ...

bundles/core/org.eclipse.smarthome.core/ESH-INF/config/i18n.xml
```diff
@@ -43,6 +43,14 @@
 			</description>
 			<advanced>true</advanced>
 		</parameter>
+		<parameter name="location" type="text" required="false">
+			<label>Location</label>
```
kaikreuzer: didn't we want to introduce a dedicated context here? It should then also be added at https://www.eclipse.org/smarthome/documentation/development/bindings/xml-reference.html#supported-contexts.

bundles/core/org.eclipse.smarthome.core/ESH-INF/config/i18n.xml
```diff
@@ -43,6 +43,14 @@
...
+			<description><![CDATA[
+		<p>Location of this installation.</p>
+		<p>Formated as "latitude,longitude,altitude"</p>
```
kaikreuzer: remove that part

bundles/core/org.eclipse.smarthome.core/src/main/java/org/eclipse/smarthome/core/internal/I18nProviderImpl.java
```diff
@@ -0,0 +1,162 @@
+package org.eclipse.smarthome.core.internal;
```
kaikreuzer: license header missing

bundles/core/org.eclipse.smarthome.core/src/main/java/org/eclipse/smarthome/core/internal/I18nProviderImpl.java
```diff
@@ -0,0 +1,162 @@
...
+ * and {@link LocationProvider} service interfaces.
+ * *
+ * <p>
+ * This implementation uses the i18n mechanism of Java ({@link ResourceBundle}) to translate a given key into text. The
+ * resources must be placed under the specific directory {@link LanguageResourceBundleManager#RESOURCE_DIRECTORY} within
+ * the certain modules. Each module is tracked in the platform by using the {@link ResourceBundleTracker} and managed by
+ * using one certain {@link LanguageResourceBundleManager} which is responsible for the translation.
+ * <p>
+ * <p>
+ * It reads a user defined configuration to set a locale and a location for this installation
```
kaikreuzer: missing full stop at the end

bundles/core/org.eclipse.smarthome.core/src/main/java/org/eclipse/smarthome/core/internal/I18nProviderImpl.java
```diff
@@ -0,0 +1,162 @@
...
+
+        try {
+            builder.setVariant(variant);
+        } catch (final RuntimeException ex) {
+            logger.warn("Variant ({}) is invalid. Skip it.", variant, ex);
+            return;
+        }
+
+        locale = builder.build();
+        logger.info("Locale set to {}", locale);
```
kaikreuzer: Also add the location here in the log message (at least when it is modified).

extensions/binding/org.eclipse.smarthome.binding.astro/src/main/java/org/eclipse/smarthome/binding/astro/discovery/AstroDiscoveryService.java
```diff
@@ -9,27 +9,28 @@
...
+import org.eclipse.smarthome.core.library.types.PointType;
 import org.eclipse.smarthome.core.thing.ThingTypeUID;
 import org.eclipse.smarthome.core.thing.ThingUID;
-import org.eclipse.smarthome.io.net.http.HttpUtil;
 import org.slf4j.Logger;
 import org.slf4j.LoggerFactory;
 
 /**
- * The {@link AstroDiscoveryService} tries to automatically discover the geolocation based on the internet IP address.
+ * The {@link AstroDiscoveryService} creates things based on configured location.
```
kaikreuzer: -> "on the configured location".

extensions/binding/org.eclipse.smarthome.binding.astro/src/main/java/org/eclipse/smarthome/binding/astro/discovery/AstroDiscoveryService.java
```diff
@@ -41,38 +42,33 @@ public AstroDiscoveryService() {
...
-            String lon = StringUtils.trim(StringUtils.substringBetween(result, "\"lon\":", "}"));
+        if (location == null) {
+            logger.debug("LocationProvider.getLocation() is not set -> Will not provide any discovery results");
+            return;
+        }
 
-            try {
-                double latitude = Double.parseDouble(lat);
-                double longitude = Double.parseDouble(lon);
+        ThingTypeUID sunType = new ThingTypeUID(BINDING_ID, SUN);
```
kaikreuzer: this already exists as 'AstroBindingConstants.THING_TYPE_SUN'

extensions/binding/org.eclipse.smarthome.binding.astro/src/main/java/org/eclipse/smarthome/binding/astro/discovery/AstroDiscoveryService.java
```diff
@@ -41,38 +42,33 @@ public AstroDiscoveryService() {
...
+        ThingTypeUID moonType = new ThingTypeUID(BINDING_ID, MOON);
```
kaikreuzer: this already exists as 'AstroBindingConstants.THING_TYPE_MOON'

extensions/binding/org.eclipse.smarthome.binding.astro/src/main/java/org/eclipse/smarthome/binding/astro/discovery/AstroDiscoveryService.java
```diff
@@ -41,38 +42,33 @@ public AstroDiscoveryService() {
...
 
-                logger.info("Evaluated Astro geolocation: latitude: {}, longitude: {}", latitude, longitude);
+        ThingUID sunThing = new ThingUID(sunType, LOCAL);
+        ThingUID moonThing = new ThingUID(moonType, LOCAL);
 
-                ThingTypeUID sunType = new ThingTypeUID(BINDING_ID, SUN);
-                ThingTypeUID moonType = new ThingTypeUID(BINDING_ID, MOON);
+        String propGeolocation = String.format("%s,%s", location.getLatitude(), location.getLongitude());
```
kaikreuzer: Why do you throw away the altitude?

triller-telekom:
If you tell me how to start the discovery service once a configuration parameter is changed I can implement the automatic discovery of new astro things based on the new location.

kaikreuzer:
What solutions come to your mind?

extensions/binding/org.eclipse.smarthome.binding.astro/src/main/java/org/eclipse/smarthome/binding/astro/discovery/AstroDiscoveryService.java
```diff
@@ -41,38 +43,31 @@ public AstroDiscoveryService() {
...
+        }
 
-                logger.info("Evaluated Astro geolocation: latitude: {}, longitude: {}", latitude, longitude);
+        ThingUID sunThing = new ThingUID(AstroBindingConstants.THING_TYPE_SUN, LOCAL);
+        ThingUID moonThing = new ThingUID(AstroBindingConstants.THING_TYPE_MOON, LOCAL);
 
-                ThingTypeUID sunType = new ThingTypeUID(BINDING_ID, SUN);
-                ThingTypeUID moonType = new ThingTypeUID(BINDING_ID, MOON);
+        String propGeolocation = String.format("%s,%s,%s", location.getLatitude(), location.getLongitude(),
+                location.getAltitude());
```
kaikreuzer: What, if altitude isn't set?

docs/documentation/development/bindings/xml-reference.md
```diff
@@ -128,7 +128,8 @@ Context is used to provide some semantic details about the parameter. The UIs us
   <tr><td>service</td><td>text</td><td>service name</td><td>custom input field</td></tr>
   <tr><td>channel</td><td>text</td><td>UID of a channel<br></td><td>custom input field</td></tr>
   <tr><td>rule</td><td>text</td><td>UID of a rule<br></td><td>custom input field</td></tr>
-
+  <tr><td>location</td><td>text</td><td>latitude,longitude,altitude<br></td><td>custom input field</td></tr>
```
kaikreuzer: this should support all variants that PointType supports (where altitude afair is an optional part).

bundles/core/org.eclipse.smarthome.core/src/main/java/org/eclipse/smarthome/core/internal/I18nProviderImpl.java
```diff
@@ -0,0 +1,178 @@
...
+
+        try {
+            builder.setVariant(variant);
+        } catch (final RuntimeException ex) {
+            logger.warn("Variant ({}) is invalid. Skip it.", variant, ex);
+            return;
+        }
+
+        locale = builder.build();
+        if (locationChanged) {
```
kaikreuzer: And what is if only location has changed, but not locale? This looks a bit unsymmetric to me here.

docs/documentation/development/bindings/xml-reference.md
```diff
@@ -128,7 +128,7 @@ Context is used to provide some semantic details about the parameter. The UIs us
   <tr><td>service</td><td>text</td><td>service name</td><td>custom input field</td></tr>
   <tr><td>channel</td><td>text</td><td>UID of a channel<br></td><td>custom input field</td></tr>
   <tr><td>rule</td><td>text</td><td>UID of a rule<br></td><td>custom input field</td></tr>
-  <tr><td>location</td><td>text</td><td>latitude,longitude,altitude<br></td><td>custom input field</td></tr>
+  <tr><td>location</td><td>text</td><td>latitude,longitude,altitude or latitude,longitude<br></td><td>custom input field</td></tr>
```
kaikreuzer: wouldn't it be nicer to just write `latitude,longitude[,altitude]`?

bundles/core/org.eclipse.smarthome.core.test/src/test/groovy/org/eclipse/smarthome/core/internal/i18n/LocationProviderOSGITest.groovy
```diff
@@ -0,0 +1,102 @@
...
+import org.junit.Before
+import org.junit.Test
+import org.osgi.service.cm.ConfigurationAdmin
+
+/**
+ * The {@link LocationProviderOSGiTest} tests the basic functionality of the {@link LocationProvider} OSGi service.
+ *
+ * @author Stefan Triller - Initial contribution
+ */
+class LocationProviderOSGiTest extends OSGiTest {
```
maggu2810: Why do we add new Groovy tests? Isn't Groovy already deprecated (for ESH)?
https://www.eclipse.org/smarthome/documentation/development/testing.html#groovy---deprecated

kaikreuzer: Sorry, didn't notice that - but you are certainly right - @triller-telekom could you migrate that to a Java test please?

PR #3589, opened 2017-06-06
Inform the user that a script module was successfully loaded

### LINKED ISSUE #3581: JSR223 Scripting: log discovered script engine factories.
@smerschjohann It would be useful to log (info level) the discovered script engine factories so that users can verify that they've installed and configured the related JSR223 libraries correctly. There was similar logging in openHAB1.

triller-telekom:
Fixes #3581

### DIFF
--- a/bundles/automation/org.eclipse.smarthome.automation.module.script/src/main/java/org/eclipse/smarthome/automation/module/script/internal/ScriptEngineManagerImpl.java
+++ b/bundles/automation/org.eclipse.smarthome.automation.module.script/src/main/java/org/eclipse/smarthome/automation/module/script/internal/ScriptEngineManagerImpl.java
@@ -84,6 +84,8 @@ public class ScriptEngineManagerImpl implements ScriptEngineManager {
                 ScriptExtensionManager.importDefaultPresets(engineProvider, engine, scriptIdentifier);
 
                 loadedScriptEngineInstances.put(scriptIdentifier, result);
+                logger.info("Successfully loaded ScriptEngine for extension {} with identifier {}", fileExtension,
+                        scriptIdentifier);
             } catch (Exception ex) {
                 logger.error("Error while creating ScriptEngine", ex);
                 removeScriptExtensions(scriptIdentifier);


maggu2810:
WRT #3588 shouldn't we use `debug` severity instead of `info`? Isn't it an expected log message?
INFO is most of the time the default log level. Does this result in a log message for every script that has been loaded?

kaikreuzer:
No, it only results in a one-time log message at startup for each script ENGINE, not each script.

triller-telekom:
Right, if it would have been per script, I would have chosen `debug` to not flood our log files. But once per engine should be fine since I agree with the author of the bug that setting up all libraries/classpaths etc. for another script language is not necessarily easy, so a message that shows the successful loading is really helpful.

maggu2810:
Hm, every time I edit a rule that exectues a script, I get a new such line.

```text
2017-06-06 17:32:43,919 | INFO  | tp1365536236-135 | ScriptEngineManagerImpl          | 181 - org.eclipse.smarthome.automation.module.script - 0.9.0.201706061452 | Successfully loaded ScriptEngine for extension application/javascript with identifier aea6753b-2f46-4dbf-9711-99103cbb9c1e
...
2017-06-06 17:33:28,389 | INFO  | tp1365536236-135 | ScriptEngineManagerImpl          | 181 - org.eclipse.smarthome.automation.module.script - 0.9.0.201706061452 | Successfully loaded ScriptEngine for extension application/javascript with identifier 8d1f644f-fb4b-46f6-88e1-65a1fc74f1ed
...
2017-06-06 17:34:20,439 | INFO  | tp1365536236-264 | ScriptEngineManagerImpl          | 181 - org.eclipse.smarthome.automation.module.script - 0.9.0.201706061452 | Successfully loaded ScriptEngine for extension application/javascript with identifier 03ea8d6e-4e48-4777-ae68-306b44dc69cc
```

maggu2810:
The first time of execution of a rule that contains a script I see this message for every rule:

```text
2017-06-06 17:32:43,919 | INFO  | tp1365536236-135 | ScriptEngineManagerImpl          | 181 - org.eclipse.smarthome.automation.module.script - 0.9.0.201706061452 | Successfully loaded ScriptEngine for extension application/javascript with identifier aea6753b-2f46-4dbf-9711-99103cbb9c1e
...
2017-06-06 17:33:28,389 | INFO  | tp1365536236-135 | ScriptEngineManagerImpl          | 181 - org.eclipse.smarthome.automation.module.script - 0.9.0.201706061452 | Successfully loaded ScriptEngine for extension application/javascript with identifier 8d1f644f-fb4b-46f6-88e1-65a1fc74f1ed
...
2017-06-06 17:34:20,439 | INFO  | tp1365536236-264 | ScriptEngineManagerImpl          | 181 - org.eclipse.smarthome.automation.module.script - 0.9.0.201706061452 | Successfully loaded ScriptEngine for extension application/javascript with identifier 03ea8d6e-4e48-4777-ae68-306b44dc69cc
```

kaikreuzer:
Clearly sounds like a bug in this PR then - I should better have tested it...

steve-bate:
When a novice user is configuring JSR223, there are at least a couple of common categories of mistakes they make (based on OH1 forum discussions). Either the classpath was not configured correctly and the engine factory can't be found or there is a problem with the script file (wrong directory, wrong extension, etc.). 

The OH2 implementation will silently ignore any scripts/files with unknown extensions. This is probably a good idea in many cases since there may be non-script files in the automation/jsr223 directory. However, the silence makes it difficult for the user diagnose why their script is not loading.

My local modification was to add a GenericScriptEngineFactory default constructor to log the engines discovered by the JSR223 ScriptEngineManager. This logging would be done independent of any scripts discovered later. One benefit of this approach is that the user can definitely know if the engine factory was successfully found on the classpath or not. If it was found and the script is still not loading, then they can look for file location or naming issues. Another benefit is that the GenericScriptEngineFactory is only created once so the information is logged only one time.

```java
public GenericScriptEngineFactory() {
       for (javax.script.ScriptEngineFactory f : engineManager.getEngineFactories()) {
            logger.info("Supported script engine: {}", f.getEngineName());
       }
 }
```

kaikreuzer:
@steve-bate Makes sense and you place seems to be the right one to add the logging statement.

triller-telekom:
Sorry I guess I was too quick with my fix yesterday :(

I have created a new PR with the fix suggested by @steve-bate. Thanks!

PR #3593, opened 2017-06-06
Provide better status information when bridge is missing

cdjackson:
Currently if there's an error with the bridge, it's things will have status HANDLER_MISSING which is wrong. This adds a BRIDGE_HANDLER_MISSING to provide a more accurate error message so we know where to look for the problem...

### DIFF
--- a/docs/documentation/concepts/things.md
+++ b/docs/documentation/concepts/things.md
@@ -47,11 +47,12 @@ The initial state of a thing is UNINITIALIZED. From UNINITIALIZED the thing goes
 A status is detailed further with a status detail object. The following table lists the different status details for each status:
 
 <table>
-<tr valign="top"><td rowspan="5">UNINITIALIZED</td><td>NONE</td><td>No further status details available.</td></tr>
+<tr valign="top"><td rowspan="6">UNINITIALIZED</td><td>NONE</td><td>No further status details available.</td></tr>
 <tr valign="top">                                  <td>HANDLER_MISSING_ERROR</td><td>The handler cannot be initialized, because the responsible binding is not available or started.</td></tr>
 <tr valign="top">                                  <td>HANDLER_REGISTERING_ERROR</td><td>The handler failed in the service registration phase.</td></tr>
 <tr valign="top">                                  <td>HANDLER_CONFIGURATION_PENDING</td><td>The handler is registered but can not be initialized caused by missing configuration parameters.</td></tr>
 <tr valign="top">                                  <td>HANDLER_INITIALIZING_ERROR</td><td>The handler failed in the initialization phase.</td></tr>
+<tr valign="top">                                  <td>BRIDGE_UNINITIALIZED</td><td>The bridge associated with this thing is not initialized.</td></tr>
 <tr valign="top"><td>INITIALIZING</td>             <td>NONE</td><td>No further status details available.</td></tr>
 <tr valign="top"><td>UNKNOWN</td>                  <td>NONE</td><td>No further status details available.</td></tr>
 <tr valign="top"><td rowspan="2">ONLINE</td>       <td>NONE</td><td>No further status details available.</td></tr>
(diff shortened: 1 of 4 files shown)


cdjackson:
I see there is a build failure - it doesn't seem to be linked to code...

sjka:
I see the need to have a better way to identify the "culprit". There two things I would like to point out:

1. HANDLER_MISSING technically is correct - it just says nothing about the reason.
2. Your suggested BRIDGE_HANDLER_MISSING state is not 100% precise - it might actually be there, but not completed the initialization yet. 

Have you considered adding a description together with HANDLER_MISSING instead?

cdjackson:
Yes, you're right - it could be changed to "missing or not initialised". Or instead of BRIDGE_HANDLER_MISSING_ERROR, we could just make it BRIDGE_HANDLER_ERROR.

I just wanted to make the error related to the bridge handler, and not the thing handler as that is totally wrong and wasted a bunch of time looking at this when there was actually a simple error in the bridge.

> Have you considered adding a description together with HANDLER_MISSING instead?

I didn't consider this - in my opinion the error should be linked to the bridge, but if you prefer I can change this back to HANDLER_MISSING_ERROR and just add a description to say "Bridge handler missing or not initialised"?

sjka:
So we have have something like 

```
UNINITIALIZED / HANDLER_MISSING_ERROR /  "Bridge handler missing or not initialized"
```

against something like

```
UNITITIALIZED / BRIDGE_UNINITIALIZED
```

(I took the freedom to refine it a little bit further)

> I just wanted to make the error related to the bridge handler, and not the thing handler as that is totally wrong [...]

I tend to agree - although HANDLER_MISSING technically is correct, it is not the whole truth and therefore misleading. So I personally would be okay with the second one, no strong opinion about it. Anybody some objections? @kaikreuzer? @maggu2810?

sjka:
(just added the APIBreaking label as a reminder that UIs might need to adapt...)

cdjackson:
At least PaperUI doesn’t care - it just prints the enum.

kaikreuzer:
> `UNITITIALIZED / BRIDGE_UNINITIALIZED`

Would be ok for me, if the typo isn't on purpose ;-)
Please do not forget to update [the docs](https://www.eclipse.org/smarthome/documentation/concepts/things.html) as well.

cdjackson:
Updated

docs/documentation/concepts/things.md
```diff
@@ -47,11 +47,12 @@ The initial state of a thing is UNINITIALIZED. From UNINITIALIZED the thing goes
...
 
 <table>
-<tr valign="top"><td rowspan="5">UNINITIALIZED</td><td>NONE</td><td>No further status details available.</td></tr>
+<tr valign="top"><td rowspan="6">UNINITIALIZED</td><td>NONE</td><td>No further status details available.</td></tr>
 <tr valign="top">                                  <td>HANDLER_MISSING_ERROR</td><td>The handler cannot be initialized, because the responsible binding is not available or started.</td></tr>
 <tr valign="top">                                  <td>HANDLER_REGISTERING_ERROR</td><td>The handler failed in the service registration phase.</td></tr>
 <tr valign="top">                                  <td>HANDLER_CONFIGURATION_PENDING</td><td>The handler is registered but can not be initialized caused by missing configuration parameters.</td></tr>
 <tr valign="top">                                  <td>HANDLER_INITIALIZING_ERROR</td><td>The handler failed in the initialization phase.</td></tr>
+<tr valign="top">                                  <td>BRIDGE_UNINITIALIZED</td><td>The brdige associated with this thing is not initialized.</td></tr>
```
sjka: brdige

cdjackson:
Engineers and spelling - never a good combination ;)

sjka:
Now that _we_ all are happy, I noticed that travis is not: 

```
Failed tests: 
  ThingManagerOSGiTest.ThingManager handles bridge-thing handler life cycle correctly:571->OSGiTest.waitForAssert:-1->OSGiTest.waitForAssert:191->OSGiTest.waitForAssert:220 
Expected: is <UNINITIALIZED (HANDLER_MISSING_ERROR)>
     but: was <UNINITIALIZED (BRIDGE_UNINITIALIZED)>
```

Looks like we have a pretty good ThingManagerOSGiTest, which you need to teach the "new correct" too.

maggu2810:
> Anybody some objections? @kaikreuzer? @maggu2810?

I am fine about adding a `BRIDGE_UNINITIALIZED` status detail.

cdjackson:
Updated and tests passing...

Issue #3562, opened 2017-06-01
ScriptServiceUtil not initialized yet on startup


sjka:
```
2017-05-31 20:09:57.118 [ERROR] [xbase.resource.BatchLinkableResource] - resolution of uriFragment '|::0.2.1.2.0.1.2.2.2.0.6.6.6::0::/1' failed.
java.lang.IllegalStateException: ScriptServiceUtil not initialized yet!
        at org.eclipse.smarthome.model.script.ScriptServiceUtil.getInstance(ScriptServiceUtil.java:88)
        at org.eclipse.smarthome.model.script.ScriptServiceUtil.getActionServices(ScriptServiceUtil.java:114)
        at org.eclipse.smarthome.model.script.internal.engine.ServiceTrackerActionServiceProvider.get(ServiceTrackerActionServiceProvider.java:21)
        at org.eclipse.smarthome.model.script.internal.engine.ServiceTrackerActionServiceProvider.get(ServiceTrackerActionServiceProvider.java:1)
        at org.eclipse.smarthome.model.script.scoping.ScriptImplicitlyImportedTypes.getActionClasses(ScriptImplicitlyImportedTypes.java:100)
        at org.eclipse.smarthome.model.script.scoping.ScriptImplicitlyImportedTypes.getStaticImportClasses(ScriptImplicitlyImportedTypes.java:94)
        at org.eclipse.xtext.xbase.scoping.batch.ImplicitlyImportedFeatures.getStaticImportClasses(ImplicitlyImportedFeatures.java:66)
        at org.eclipse.xtext.xbase.scoping.batch.XbaseBatchScopeProvider.newSession(XbaseBatchScopeProvider.java:112)
        at org.eclipse.xtext.xbase.typesystem.internal.DefaultReentrantTypeResolver.resolve(DefaultReentrantTypeResolver.java:162)
        at org.eclipse.xtext.xbase.typesystem.internal.DefaultReentrantTypeResolver.reentrantResolve(DefaultReentrantTypeResolver.java:139)
        at org.eclipse.xtext.xbase.typesystem.internal.CachingBatchTypeResolver$LazyResolvedTypes.resolveTypes(CachingBatchTypeResolver.java:80)
        at org.eclipse.xtext.xbase.typesystem.internal.CachingBatchTypeResolver$2.process(CachingBatchTypeResolver.java:57)
[...]

PR #3595, opened 2017-06-07
less statics in FolderObserver

sjka:
In the continued attempt to hunt down #3562 the static fields (and thereby also the methods) fell from grace. Ideally they should work nevertheless, but as ordering of the set/unset calls between two different instances is not guaranteed, this _could_ be a problem, potentially.

### DIFF
--- a/bundles/model/org.eclipse.smarthome.model.core/src/main/java/org/eclipse/smarthome/model/core/internal/folder/FolderObserver.java
+++ b/bundles/model/org.eclipse.smarthome.model.core/src/main/java/org/eclipse/smarthome/model/core/internal/folder/FolderObserver.java
@@ -64,10 +64,10 @@ public class FolderObserver extends AbstractWatchService implements ManagedServi
     private final Map<String, String[]> folderFileExtMap = new ConcurrentHashMap<String, String[]>();
 
     /* set of file extensions for which we have parsers already registered */
-    private static Set<String> parsers = new HashSet<>();
+    private final Set<String> parsers = new HashSet<>();
 
     /* set of files that have been ignored due to a missing parser */
-    private static Set<File> ignoredFiles = new HashSet<>();
+    private final Set<File> ignoredFiles = new HashSet<>();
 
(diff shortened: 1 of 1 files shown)


bundles/model/org.eclipse.smarthome.model.core/src/main/java/org/eclipse/smarthome/model/core/internal/folder/FolderObserver.java
```diff
@@ -289,10 +287,8 @@ private static File getFileByFileExtMap(Map<String, String[]> folderFileExtMap,
      *            the file name to get the {@link File} for
      * @return the corresponding {@link File}
      */
-    private static File getFile(String filename) {
-        File folder = new File(ConfigConstants.getConfigFolder() + File.separator + filename);
-
-        return folder;
+    private File getFile(String filename) {
+        return new File(ConfigConstants.getConfigFolder() + File.separator + filename);
```
maggu2810: Just FYI you could also use:

```Java
Paths.get(ConfigConstants.getConfigFolder(), filename).toFile();
```

Just if you need to use multiples `File.separator` some time this is simpler to write...

sjka: ...or 

```java
File(ConfigConstants.getConfigFolder(), filename)
```

Frankly said, I did not even closely look at it, just hit "Cmd+1" -> "inline" 😞

sjka: Let's clean that up separately

maggu2810: Sure, as written above my comment should be only "FYI", no need to change it.

PR #3597, opened 2017-06-07
Inform the user only once for each loaded script engine

triller-telekom:
Relates to #3589

### DIFF
--- a/bundles/automation/org.eclipse.smarthome.automation.module.script/src/main/java/org/eclipse/smarthome/automation/module/script/internal/GenericScriptEngineFactory.java
+++ b/bundles/automation/org.eclipse.smarthome.automation.module.script/src/main/java/org/eclipse/smarthome/automation/module/script/internal/GenericScriptEngineFactory.java
@@ -16,9 +16,19 @@ import javax.script.ScriptEngine;
 import javax.script.ScriptEngineManager;
 
 import org.eclipse.smarthome.automation.module.script.ScriptEngineFactory;
+import org.slf4j.Logger;
+import org.slf4j.LoggerFactory;
 
 public class GenericScriptEngineFactory implements ScriptEngineFactory {
     private ScriptEngineManager engineManager = new ScriptEngineManager();
+    private final Logger logger = LoggerFactory.getLogger(getClass());
+
+    public GenericScriptEngineFactory() {
+        for (javax.script.ScriptEngineFactory f : engineManager.getEngineFactories()) {
+            logger.info("Supported script engine: {} for language {} with mimetypes {}", f.getEngineName(),
+                    f.getLanguageName(), f.getMimeTypes());
+        }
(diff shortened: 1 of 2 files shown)


bundles/automation/org.eclipse.smarthome.automation.module.script/src/main/java/org/eclipse/smarthome/automation/module/script/internal/GenericScriptEngineFactory.java
```diff
@@ -16,9 +16,19 @@
...
+import org.slf4j.Logger;
+import org.slf4j.LoggerFactory;
 
 public class GenericScriptEngineFactory implements ScriptEngineFactory {
     private ScriptEngineManager engineManager = new ScriptEngineManager();
+    private final Logger logger = LoggerFactory.getLogger(getClass());
+
+    public GenericScriptEngineFactory() {
+        for (javax.script.ScriptEngineFactory f : engineManager.getEngineFactories()) {
+            logger.info("Supported script engine: {} for language {} with mimetypes {}", f.getEngineName(),
```
kaikreuzer: As it is INFO level that most end users will get to see in their main log, I would suggest to make it a bit more end user compatible, e.g.

> Activated scripting support for ECMAScript

We could add an additional debug log that then also lists all possible file extensions and mimetypes as well as the engine name itself.

triller-telekom: Done.

bundles/automation/org.eclipse.smarthome.automation.module.script/src/main/java/org/eclipse/smarthome/automation/module/script/internal/GenericScriptEngineFactory.java
```diff
@@ -16,9 +16,22 @@
...
+import org.slf4j.Logger;
+import org.slf4j.LoggerFactory;
 
 public class GenericScriptEngineFactory implements ScriptEngineFactory {
     private ScriptEngineManager engineManager = new ScriptEngineManager();
+    private final Logger logger = LoggerFactory.getLogger(getClass());
+
+    public GenericScriptEngineFactory() {
+        for (javax.script.ScriptEngineFactory f : engineManager.getEngineFactories()) {
+            logger.info("Activated scripting support for language {}", f.getLanguageName());
```
kaikreuzer: could you remove the word "language"? I think this is obsolete here.

triller-telekom: Done

PR #3598, opened 2017-06-07
[sonos] Exchanged file extensions to existing constants

cweitkamp:

### DIFF
--- a/extensions/binding/org.eclipse.smarthome.binding.sonos/src/main/java/org/eclipse/smarthome/binding/sonos/internal/SonosAudioSink.java
+++ b/extensions/binding/org.eclipse.smarthome.binding.sonos/src/main/java/org/eclipse/smarthome/binding/sonos/internal/SonosAudioSink.java
@@ -17,6 +17,7 @@ import org.eclipse.smarthome.core.audio.AudioFormat;
 import org.eclipse.smarthome.core.audio.AudioHTTPServer;
 import org.eclipse.smarthome.core.audio.AudioSink;
 import org.eclipse.smarthome.core.audio.AudioStream;
+import org.eclipse.smarthome.core.audio.FileAudioStream;
 import org.eclipse.smarthome.core.audio.FixedLengthAudioStream;
 import org.eclipse.smarthome.core.audio.URLAudioStream;
 import org.eclipse.smarthome.core.audio.UnsupportedAudioFormatException;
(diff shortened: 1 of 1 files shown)


extensions/binding/org.eclipse.smarthome.binding.sonos/src/main/java/org/eclipse/smarthome/binding/sonos/internal/SonosAudioSink.java
```diff
@@ -92,9 +93,9 @@ public void process(AudioStream audioStream) throws UnsupportedAudioFormatExcept
                         logger.warn("Sonos speaker '{}' is not initialized - status is {}", handler.getThing().getUID(),
                                 handler.getThing().getStatus());
                     } else if (AudioFormat.WAV.isCompatible(format)) {
-                        handler.playNotificationSoundURI(new StringType(url + ".wav"));
+                        handler.playNotificationSoundURI(new StringType(url + FileAudioStream.WAV_EXTENSION));
                     } else if (AudioFormat.MP3.isCompatible(format)) {
-                        handler.playNotificationSoundURI(new StringType(url + ".mp3"));
+                        handler.playNotificationSoundURI(new StringType(url + FileAudioStream.MP3_EXTENSION));
```
maggu2810: I really don't know, but isn't it a little bit unexpected that an "extension" constant contains the "filename-extension-separator" (the dot)?

cweitkamp: Hm ... probably not.
The constant(s) is/are used in the [getAudioFormat(File file)](https://github.com/eclipse/smarthome/blob/master/bundles/core/org.eclipse.smarthome.core.audio/src/main/java/org/eclipse/smarthome/core/audio/FileAudioStream.java#L47) method to determine the audio format. Maybe we should refactor those lines to compare file extensions using the [FilenameUtils.getExtension](http://commons.apache.org/proper/commons-io/javadocs/api-2.5/org/apache/commons/io/FilenameUtils.html#getExtension(java.lang.String)) method from Apache Commons IO. And additionally use `FilenameUtils.EXTENSION_SEPARATOR_STR` to concat again. wdyt?

cweitkamp: I mean ... probably _yes_. Which changes the statement of my reply.

kaikreuzer: Sure, this should be changed.
I'd prefer doing it without Apache commons though - the less dependencies we have, the better.

cweitkamp: PR #3837

PR #3599, opened 2017-06-07
additional logging for ScriptServiceUtil

sjka:
In another attempt to hunt down #3562 I would like to suggest some additional logging which we can happily remove again once the issue got resolved.


kaikreuzer:
And don't forget the removal later on ;-)

PR #3503, opened 2017-05-23
Astro-Binding compliance with ESH Scheduler

amitjoy:
1. Added New Simplified Job API
2. Adapted Previous Jobs to new API
3. Removed Quartz Dependencies
4. Added CRON Helper Tests
5. Compact Profile 2 Compliance
6. Code Refactorings

PR #3603, opened 2017-06-07
Overwrite getQueue on Scheduler

### LINKED ISSUE #3602: Return Queue of Runnables in ExpressionThreadPoolExecutor

triller-telekom:
Fixes #3602

### DIFF
--- a/bundles/core/org.eclipse.smarthome.core/src/main/java/org/eclipse/smarthome/core/scheduler/ExpressionThreadPoolManager.java
+++ b/bundles/core/org.eclipse.smarthome.core/src/main/java/org/eclipse/smarthome/core/scheduler/ExpressionThreadPoolManager.java
@@ -13,10 +13,12 @@ import java.util.Date;
 import java.util.HashMap;
 import java.util.List;
 import java.util.Map;
+import java.util.concurrent.BlockingQueue;
 import java.util.concurrent.CancellationException;
 import java.util.concurrent.ConcurrentHashMap;
 import java.util.concurrent.ExecutorService;
 import java.util.concurrent.Future;
+import java.util.concurrent.LinkedBlockingQueue;
 import java.util.concurrent.RejectedExecutionException;
 import java.util.concurrent.RejectedExecutionHandler;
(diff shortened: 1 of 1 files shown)


kaikreuzer:
@kgoderis Could you briefly review?

triller-telekom:
Some background on this change:
Inside the [AstroBinding using the new scheduler](https://github.com/eclipse/smarthome/pull/3503/files#diff-637f70cb0a947f264138d125132a0a99R253) we came across the problem that we want to check the content of the schedulers queue to cancel jobs only matching a certain `thing uuid`.

Without this change the `getQueue()` method will be the one from `ScheduledThreadPoolExecutor` and return objects of type `ScheduledFuture`.

There is no way to retrieve a `Runnable` which was scheduled in the scheduler from these `ScheduledFuture` objects.

amitjoy:
> There is no way to retrieve a Runnable which was scheduled in the scheduler from these ScheduledFuture objects.

Just in continuation. This is because the member map instance is neither exposed nor passed down to the `ScheduledThreadExecutor` of Java so that the scheduled `Runnable`s could have been extracted as is. That's why this change is integral for any scheduling operation and do some further operations on the scheduled `Runnable`s

kgoderis:
@amitjoy @triller-telekom The queue and map with runnables are not exposed on purpose. Rather than directly manipulating these members, you have to keep track of the Expression you schedule, and call ExpressionThreadPoolExecutor.remove(Expression expression) instead.

kgoderis:
> I guess the background of @kgoderis's veto is that if they get exposed, it becomes way too easy to accidentally tamper with other things'/handlers' jobs. I second that.

Right indeed, and beyond that, it is all about Expressions that get scheduled for execution, so, even as we would enhance the API, it should evolve around the Expressions themselves primarily.

triller-telekom:
Alright, I agree that having direct access to the queue which would also reveal access to other tasks that I didn't schedule isn't a good thing. So I discard this PR and users of the scheduler like the astro binding have to store all tasks they schedule.

PR #3606, opened 2017-06-07
add support for retrieving channel specific config descriptions

sjka:
...so that bindings have a chance to provide channel specific config options
via a ConfigOptionsProvider in the same way as it is possible for things already.

I therefore simply extended the already existing proxy to also handle the "channel" scheme.

The PaperUI will have to be adapted to actually use them (similar as #3215 which added this feature for things). 


kaikreuzer:
The class name isn't a perfect fit anymore, but considering that a Channel is a part of a Thing that should be ok...

PR #3609, opened 2017-06-08
[IDE Setup] abbreviate package names

sjka:
I just (re)discovered Eclipse's feature to abbreviate package names in the Package Explorer:

image

It lets us use even more space of today's tremendously large screens, so I though to suggest it as a general configuration. WDYT?


targetplatform/EclipseSmartHome.setup
```diff
@@ -455,6 +459,10 @@
           value="true"/>
       <setupTask
           xsi:type="setup:PreferenceTask"
+          key="/instance/org.eclipse.jdt.ui/org.eclipse.jdt.ui.pkgNameAbbreviationPatternForPackagesView"
+          value="org.eclipse.smarthome=o.e.sh"/>
```
maggu2810: I would also agree using `esh` as a abbreviation for `org.eclipse.smarthome`

kaikreuzer: I'd prefer `o.e.sh` as it still indicates that it is a package abbreviation (similar as logback does it in logfiles), while `esh` might confuse some people.

kaikreuzer: @maggu2810 If ok for you as well, feel free to merge!

PR #3615, opened 2017-06-08
rename I18nProvider to TranslationProvider

maggu2810:
WRT https://github.com/eclipse/smarthome/pull/3587#discussion_r120869757

### DIFF
--- a/bundles/automation/org.eclipse.smarthome.automation.providers/OSGI-INF/ModuleTypeProvider.xml
+++ b/bundles/automation/org.eclipse.smarthome.automation.providers/OSGI-INF/ModuleTypeProvider.xml
@@ -17,7 +17,7 @@
 
     <reference bind="addParser" cardinality="1..n" interface="org.eclipse.smarthome.automation.parser.Parser" name="Parser" policy="dynamic" unbind="removeParser" target="(parser.type=parser.module.type)"/>
     <reference bind="setModuleTypeRegistry" cardinality="1..1" interface="org.eclipse.smarthome.automation.type.ModuleTypeRegistry" name="ModuleTypeRegistry" policy="static" unbind="removeModuleTypeRegistry"/>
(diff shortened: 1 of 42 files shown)


kaikreuzer:
Did you check at how many places in the docs & tutorials this would need to be changed?

And do you have any idea of how much existing code out there in solutions will have to be adapted for this breaking change?

maggu2810:
> Did you check at how many places in the docs & tutorials this would need to be changed?

It should be changed in all places that are part of the repository.

> And do you have any idea of how much existing code out there in solutions will have to be adapted for this breaking change?

No, how should I?

maggu2810:
> Would you dare such a refactoring?

You asked me if I dare such a refactoring. And I dared it.

kaikreuzer:
You are a daring guy ;-)

Ok, I have checked the openhab2 repo and it doesn't seem to have any significant impact.
Also checked with DT colleagues which are fine with the renaming. So I'd hope that any other ESH consumer will also be able to cope with it easily.
But I added "API Breaking" label to highlight that.

kaikreuzer:
Nonetheless, reading https://github.com/eclipse/smarthome/blob/master/docs/documentation/features/internationalization.md#i18n-text-provider-api, I am wondering if "Translation" is really the right word. This service does not translate anything, but it provides texts in a given locale for a key. So it really is about l10n...

cdjackson:
If this change goes ahead please can we synchronise it so we don’t break bindings that use the current provider.

tomhoefer:
It is ok for me ... we will take care of the changes required by this PR once it has been merged

kaikreuzer:
> I am wondering if "Translation" is really the right word. This service does not translate anything, but it provides texts in a given locale for a key. 

Ok, as no one (including myself) has a better suggestion, let's keep it.

PR #3617, opened 2017-06-08
allow bind new storage service if it uses a higher service ranking

maggu2810:
If there are multiple storage service implementations used in the runtime and a component already uses a storage service reference shouldn't the storage service with the highest service ranking be used?
Without force the usage of the highest service ranking one we cannot ensure which one get's injected if a product uses multiple ones.

### DIFF
--- a/bundles/automation/org.eclipse.smarthome.automation.core/OSGI-INF/ManagedRuleProvider.xml
+++ b/bundles/automation/org.eclipse.smarthome.automation.core/OSGI-INF/ManagedRuleProvider.xml
@@ -10,14 +10,12 @@
 -->
 <scr:component xmlns:scr="http://www.osgi.org/xmlns/scr/v1.1.0" immediate="true" name="org.eclipse.smarthome.automation.managedruleprovider">
    <implementation class="org.eclipse.smarthome.automation.ManagedRuleProvider"/>
-   
+
    <service>
       <provide interface="org.eclipse.smarthome.automation.RuleProvider"/>
       <provide interface="org.eclipse.smarthome.automation.ManagedRuleProvider"/>
    </service>
-   
-   <reference bind="setStorageService" cardinality="1..1" 
-      interface="org.eclipse.smarthome.core.storage.StorageService" name="StorageService" 
-      policy="static" unbind="unsetStorageService"/>
-   
+
+<reference bind="setStorageService" cardinality="1..1" interface="org.eclipse.smarthome.core.storage.StorageService" name="StorageService" policy="dynamic" unbind="unsetStorageService"/>
+
 </scr:component>
--- a/bundles/config/org.eclipse.smarthome.config.discovery/src/main/java/org/eclipse/smarthome/config/discovery/internal/PersistentInbox.java
+++ b/bundles/config/org.eclipse.smarthome.config.discovery/src/main/java/org/eclipse/smarthome/config/discovery/internal/PersistentInbox.java
@@ -129,7 +129,8 @@ public final class PersistentInbox implements Inbox, DiscoveryListener, ThingReg
     private ManagedThingProvider managedThingProvider;
     private ThingTypeRegistry thingTypeRegistry;
     private ConfigDescriptionRegistry configDescRegistry;
-    private Storage<DiscoveryResult> discoveryResultStorage;
+    private StorageService storageService;
+    private volatile Storage<DiscoveryResult> discoveryResultStorage;
     private Map<DiscoveryResult, Class<?>> resultDiscovererMap = new ConcurrentHashMap<>();
     private ScheduledFuture<?> timeToLiveChecker;
     private EventPublisher eventPublisher;
@@ -603,14 +604,20 @@ public final class PersistentInbox implements Inbox, DiscoveryListener, ThingReg
         this.managedThingProvider = null;
     }
 
-    @Reference
-    protected void setStorageService(StorageService storageService) {
-        this.discoveryResultStorage = storageService.getStorage(DiscoveryResult.class.getName(),
-                this.getClass().getClassLoader());
+    @Reference(policy = ReferencePolicy.DYNAMIC)
+    protected void setStorageService(final StorageService storageService) {
+        if (this.storageService != storageService) {
+            this.storageService = storageService;
+            this.discoveryResultStorage = storageService.getStorage(DiscoveryResult.class.getName(),
+                    this.getClass().getClassLoader());
+        }
     }
 
-    protected void unsetStorageService(StorageService storageService) {
-        this.discoveryResultStorage = null;
+    protected void unsetStorageService(final StorageService storageService) {
+        if (this.storageService == storageService) {
+            this.storageService = null;
+            this.discoveryResultStorage = null;
+        }
     }
 
     @Reference(cardinality = ReferenceCardinality.OPTIONAL, policy = ReferencePolicy.DYNAMIC)
--- a/bundles/core/org.eclipse.smarthome.core.thing/src/main/java/org/eclipse/smarthome/core/thing/ManagedThingProvider.java
+++ b/bundles/core/org.eclipse.smarthome.core.thing/src/main/java/org/eclipse/smarthome/core/thing/ManagedThingProvider.java
@@ -11,6 +11,7 @@ import org.eclipse.smarthome.core.common.registry.DefaultAbstractManagedProvider
 import org.eclipse.smarthome.core.storage.StorageService;
 import org.osgi.service.component.annotations.Component;
 import org.osgi.service.component.annotations.Reference;
+import org.osgi.service.component.annotations.ReferencePolicy;
 
 /**
  * {@link ManagedThingProvider} is an OSGi service, that allows to add or remove
@@ -35,7 +36,7 @@ public class ManagedThingProvider extends DefaultAbstractManagedProvider<Thing, 
         return key.toString();
     }
 
-    @Reference
+    @Reference(policy = ReferencePolicy.DYNAMIC)
     @Override
     protected void setStorageService(StorageService storageService) {
         super.setStorageService(storageService);
--- a/bundles/core/org.eclipse.smarthome.core.thing/src/main/java/org/eclipse/smarthome/core/thing/link/ManagedItemChannelLinkProvider.java
+++ b/bundles/core/org.eclipse.smarthome.core.thing/src/main/java/org/eclipse/smarthome/core/thing/link/ManagedItemChannelLinkProvider.java
@@ -14,6 +14,7 @@ import org.eclipse.smarthome.core.storage.StorageService;
 import org.eclipse.smarthome.core.thing.ThingUID;
 import org.osgi.service.component.annotations.Component;
 import org.osgi.service.component.annotations.Reference;
+import org.osgi.service.component.annotations.ReferencePolicy;
 
 /**
  *
@@ -45,7 +46,7 @@ public class ManagedItemChannelLinkProvider extends DefaultAbstractManagedProvid
         }
     }
 
-    @Reference
+    @Reference(policy = ReferencePolicy.DYNAMIC)
     @Override
     protected void setStorageService(StorageService storageService) {
         super.setStorageService(storageService);
--- a/bundles/core/org.eclipse.smarthome.core/OSGI-INF/manageditemprovider.xml
+++ b/bundles/core/org.eclipse.smarthome.core/OSGI-INF/manageditemprovider.xml
@@ -10,17 +10,13 @@
 -->
 <scr:component xmlns:scr="http://www.osgi.org/xmlns/scr/v1.1.0" immediate="true" name="org.eclipse.smarthome.core.manageditemprovider">
    <implementation class="org.eclipse.smarthome.core.items.ManagedItemProvider"/>
-   
+
    <service>
       <provide interface="org.eclipse.smarthome.core.items.ItemProvider"/>
       <provide interface="org.eclipse.smarthome.core.items.ManagedItemProvider"/>
    </service>
-   
-   <reference bind="setStorageService" cardinality="1..1" 
-      interface="org.eclipse.smarthome.core.storage.StorageService" name="StorageService" 
-      policy="static" unbind="unsetStorageService"/>
-   <reference bind="addItemFactory" cardinality="0..n" 
-      interface="org.eclipse.smarthome.core.items.ItemFactory" name="ItemFactory" 
-      policy="dynamic" unbind="removeItemFactory"/>
-   
+
+   <reference bind="setStorageService" cardinality="1..1" interface="org.eclipse.smarthome.core.storage.StorageService" name="StorageService" policy="dynamic" unbind="unsetStorageService"/>
+   <reference bind="addItemFactory" cardinality="0..n" interface="org.eclipse.smarthome.core.items.ItemFactory" name="ItemFactory" policy="dynamic" unbind="removeItemFactory"/>
+
 </scr:component>
--- a/bundles/core/org.eclipse.smarthome.core/src/main/java/org/eclipse/smarthome/core/common/registry/AbstractManagedProvider.java
+++ b/bundles/core/org.eclipse.smarthome.core/src/main/java/org/eclipse/smarthome/core/common/registry/AbstractManagedProvider.java
@@ -42,7 +42,9 @@ import com.google.common.collect.ImmutableList;
 public abstract class AbstractManagedProvider<E extends Identifiable<K>, K, PE> extends AbstractProvider<E>
         implements ManagedProvider<E, K> {
 
-    private Storage<PE> storage;
+    private StorageService storageService;
+    private volatile Storage<PE> storage;
+
     protected final Logger logger = LoggerFactory.getLogger(AbstractManagedProvider.class);
 
     @Override
@@ -152,8 +154,18 @@ public abstract class AbstractManagedProvider<E extends Identifiable<K>, K, PE> 
      */
     protected abstract @NonNull String keyToString(@NonNull K key);
 
-    protected void setStorageService(StorageService storageService) {
-        this.storage = storageService.getStorage(getStorageName(), this.getClass().getClassLoader());
+    protected void setStorageService(final StorageService storageService) {
+        if (this.storageService != storageService) {
+            this.storageService = storageService;
+            storage = storageService.getStorage(getStorageName(), this.getClass().getClassLoader());
+        }
+    }
+
+    protected void unsetStorageService(final StorageService storageService) {
+        if (this.storageService == storageService) {
+            this.storageService = null;
+            this.storage = null;
+        }
     }
 
     /**
@@ -174,9 +186,5 @@ public abstract class AbstractManagedProvider<E extends Identifiable<K>, K, PE> 
      * @return persistable element
      */
     protected abstract PE toPersistableElement(E element);
-
-    protected void unsetStorageService(StorageService storageService) {
-        this.storage = null;
-    }
 
 }
--- a/extensions/extensionservice/org.eclipse.smarthome.extensionservice.marketplace.automation/OSGI-INF/MarketplaceTemplateProvider.xml
+++ b/extensions/extensionservice/org.eclipse.smarthome.extensionservice.marketplace.automation/OSGI-INF/MarketplaceTemplateProvider.xml
@@ -14,6 +14,6 @@
       <provide interface="org.eclipse.smarthome.extensionservice.marketplace.automation.internal.MarketplaceRuleTemplateProvider"/>
       <provide interface="org.eclipse.smarthome.automation.template.RuleTemplateProvider"/>
    </service>
-   <reference bind="setStorageService" cardinality="1..1" interface="org.eclipse.smarthome.core.storage.StorageService" name="StorageService" policy="static" unbind="unsetStorageService"/>
+   <reference bind="setStorageService" cardinality="1..1" interface="org.eclipse.smarthome.core.storage.StorageService" name="StorageService" policy="dynamic" unbind="unsetStorageService"/>
    <reference bind="setParser" cardinality="1..1" interface="org.eclipse.smarthome.automation.parser.Parser" name="Parser" policy="static" target="(&amp;(format=json)(parser.type=parser.template))" unbind="unsetParser"/>
 </scr:component>


maggu2810:
@kaikreuzer @sjka WDYT?

sjka:
Sounds risky. Imagine what happens if one is down for a while (due to an update of the bundle, different positions in the startup sequence, etc...).... Data might get written into one, and a moment later it's gone because a higher ranked one pops up, which or course does not know anything about the latest changes written to the other one. Or did I get something wrong?

maggu2810:
Hm, I don't think it is more or less risky.

If you use only one implementation it does not make any difference if static or dynamic is used.
If you use two implementations then you do it by intention and you would know which one is used. And the used service should be that one with the higher ranking.
If the rebind is applied as soon as a newer ranking is available it is IMHO "more transparent" as you know the time you install the new one. If you restart or refresh bundles or restart your whole distribution then you will only postpone the rebind. And then it is the same, data could be written to the one service and after an restart read from the other one (so, it doesn't make really a difference).

Last but not least, I would like that the distributor could create a StorageService using the highest service ranking and reuse code from other StorageService.
Let's consider this example:

```Java
@Component(service = { StorageService.class },
        immediate = true,
        property = {
            Constants.SERVICE_RANKING + ":Integer=" + Integer.MAX_VALUE
        })
public class MigrateStorageService implements StorageService {

    private final Logger logger = LoggerFactory.getLogger(MigrateStorageService.class);

    @Reference(service = StorageService.class, target = "(storage.format=...)")
    private StorageService ssNew;


    @Reference(service = StorageService.class, target = "(storage.format=...)")
    private StorageService ssOld;

    @Override
    public <T> Storage<T> getStorage(final String name) {
        logger.info("Get storage without explicit classloader: {}", name);
        return migrate(ssOld.getStorage(name), ssNew.getStorage(name));
    }

    @Override
    public <T> Storage<T> getStorage(final String name, final ClassLoader classLoader) {
        logger.info("Get storage with explicit classloader: {}", name);
        return migrate(ssOld.getStorage(name, classLoader), ssNew.getStorage(name, classLoader));
    }

    private <T> Storage<T> migrate(final Storage<T> storageOld, final Storage<T> storageNew) {
        final Collection<String> keys = storageOld.getKeys();
        for (final String key : keys) {
            final T value = storageOld.get(key);
            storageNew.put(key, value);
            storageOld.remove(key);
        }
        return storageNew;
    }

}
```

Using such an implementation it is very easy to switch between storage service implementations in distributions, if the distributor ever decides that another one should be used.

kaikreuzer:
Having an "extended" implementation in a system sounds like a valid use case (especially for the mentioned data migration use case).

Two questions though:

1. Does OSGi DS really automatically replace an injected service if a higher service rank property is provided? I faintly remember that only OSGi 4.3+ supports this, right?
2. What is the order of set/unset being called? Is it unset(old)->set(new)? If so, we would have the risk that storage (in `AbstractManagedProvider`) can be null for a moment, which we do not check yet. If it is set(new)->unset(old) (this is what I read from the specs), our unset implementation would set storage incorrectly to null, which is not what you expect.

maggu2810:
> Does OSGi DS really automatically replace an injected service if a higher service rank property is provided? I faintly remember that only OSGi 4.3+ supports this, right?

The compendium version 4.2 (IIRC this is the lowest supported one) only specified "With the dynamic policy, SCR can change the set of bound services without deactivating a component configuration."

There is no word about service ranking in 4.2 policy description.
The behavior of a better service respecting the Service Ranking is explicitly named in 4.3+

> What is the order of set/unset being called? Is it unset(old)->set(new)? If so, we would have the risk that storage (in AbstractManagedProvider) can be null for a moment, which we do not check yet. If it is set(new)->unset(old) (this is what I read from the specs), our unset implementation would set storage incorrectly to null, which is not what you expect.

The compendium version 4.2 (IIRC this is the lowest supported one) only specified that the bind and unbind method are called, not the order (I don't think the intention of "by calls to the bind and unbind meth-
ods" is to specify an order).

If the order would be really "set(new)->unset(old)" it could always use the latest set one and only set `null` in the unset method if `old == currentUsedOne`.

IIRC there are already services using the `dynamic` approach so I assume the one that introduced this approaches should know how it is handled by the OSGi framework. Correctly?

sjka:
> I don't think the intention of "by calls to the bind and unbind methods" is to specify an order
> If the order would be really "set(new)->unset(old)" it could always use the latest set one and only set null in the unset method if old == currentUsedOne.

That's correct, there is no order specified. However, apparently most (if not all) implementations set the new instance before removing the old one. So this check actually is required - either together with synchronization or with AtomicReferences (see also https://stackoverflow.com/a/30098048). We are doing it wrong everywhere.

sjka:
> Hm, I don't think it is more or less risky.

You are right. I agree, this change doesn't add to the risk. It _can_ be a bad idea to have two storage services, but this did not change with this PR. 

Nevertheless, the current implementations would need to be prepared for handling dependency replacements correctly (like discussed above). I'm pretty sure all of them currently simply nullify their fields. We might consider fixing this throughout the code base, at least for all _dynamic_ single references in a separate PR...?

kaikreuzer:
> We are doing it wrong everywhere.

Correct - but more or less by intention, because almost everybody does it this way, the code is much better readable AND: Because I have NEVER seen the described behavior in the wild (using OSGi 5/6 framework in openHAB since many years)... Once someone shows me that any implementation actually does it that way, I might give in to do a refactoring to that very ugly code in the setters/unsetters...

maggu2810:
> That's correct, there is no order specified. However, apparently most (if not all) implementations set the new instance before removing the old one. So this check actually is required - either together with synchronization or with AtomicReferences (see also https://stackoverflow.com/a/30098048).

Thanks a lot for the link. The comment has been written by Neil Bartlett and he should know it.

maggu2810:
And it really make sense to call `service.bind(refNew); service.unbind(refOld);` as it could be a mandatory dependency and there should be every time one set if the component is active.

sjka:
> The comment has been written by Neil Bartlett and he should know it.

Yepp, think so too...

> I might give in to do a refactoring to that very ugly code in the setters/unsetters

How about refactoring it to using DS annotations? Even less ugly code and synchronization handling for free... 😉

maggu2810:
> How about refactoring it to using DS annotations? Even less ugly code and synchronization handling for free... ￼

I prefer using annotations and DS XML generation but in this case I don't see how it should solve to check the instance on unset. How do you think that an annotation could help?

kaikreuzer:
How many uses of dynamic 1..1 dependencies do we have in ESH? I'd assume hardly any so far, since we have to work on 4.2, in which case the use of such dependencies should be avoided, right?
So in that light, I'd be ok with an according refactoring ;-)

maggu2810:
I tried to make the code for the storage handling more robust.

I used synchronization blocks for the set / unset method instead of an atomic reference because we need to handle the storage if the storage service changes and they need to be in sync.

We also need to use volatile (if we don't use atomic refs) because the variable itself will be written by another thread then it is used.

I don't put ever usage of `storage` into an synchronized block. The `volatile` variable will always use the latest set value (it only reads the reference) and it should be non-null by the OSGi service reference declaration (mandatory).

sjka:
> How many uses of dynamic 1..1 dependencies do we have in ESH? I'd assume hardly any so far, since we have to work on 4.2, in which case the use of such dependencies should be avoided, right?

```bash
grep -R "policy=\"dynamic\"" * | grep "\.\.1" | wc -l
``` 
gives 

* 65 dynamic 1..1 dependencies and another
* 64 dynamic 0..1 dependencies

So I guess we might have just been lucky so far, since openHAB is not using much of the dynamics (in the sense that it is not doing any bundle updates at runtime) or we are running on OSGi 4.2 stacks which don't care...

Anyway, back to this PR: It looks good to me - although strictly speaking I have to second Kai that this does not exactly _look_ good. @kaikreuzer any concerns?

maggu2810:
> I have to second Kai that this does not exactly look good.

What do you mean exactly by `this`? What do you guys complain.
Something the PR could break or some handling in general?

sjka:
No no, all good!
I was only referring to Kai's statement above about the "very ugly code in the setters/unsetters..." as opposed to a simple (but broken) `this.someField = null`  - sorry this joke was not clear enough.

kaikreuzer:
> sorry this joke was not clear enough.

I actually do not feel like joking here. Seriously, @maggu2810's code nicely shows the problem that I have with it: He had to judge atomic references vs. synchronization, use volatile and produce a couple of lines of highly sophisticated and perfect code JUST TO IMPLEMENT A SIMPLE SETTER.

So is this what the framework expects from every dummy user who implements a component? If so, there is something fundamentally broken in OSGi.

Will we as a consequence change the code of all other 65 occurrences to such code as well? And teach every contributor how and why to write such code in their contributions?

Btw, I am still waiting for someone giving me an example of which OSGi framework actually does it that way (and hence makes such code necessary).

@maggu2810 Please don't get me wrong, I am not blaming anything about your PR - but I hope you understand my worries about the level of complexity that OSGi seem to put here on its users.

maggu2810:
Why do you think there is something fundamental broken?

Isn't the volatile and atomic stuff "normal doing" using multiple threads?

If a variable is not marked volatile and is modified by one thread but used by another thread, isn't it known behavior that without using memory barriers you could run into the situation that the other thread never gets the new reference?

Bnd (e.g. the bnd-maven-plugin or maven-bundle-plugin) complains if you are using a dynamic reference but don't mark it volatile.

Also here (http://blog.vogella.com/2016/06/21/getting-started-with-osgi-declarative-services/) this statement could be found: http://blog.vogella.com/2016/06/21/getting-started-with-osgi-declarative-services/

> Btw, I am still waiting for someone giving me an example of which OSGi framework actually does it that way (and hence makes such code necessary).

There has been an article on enroute.osgi.org recently about that topic (but I think I should read them again): http://enroute.osgi.org/appnotes/concurrency.html

Kai, I don't understand which example you want to get.
Do we consider that the statement of Neil Bartlett and the article of Peter Kriens are wrong?

Should the OSGi framework change the threading mechanism, reference updates of the JVM in general?
Didn't show the first example in "Time is More Relative than You Think" already the problem in general? But this topic is at all not OSGi related but true for ever code used in a multi threaded environment.

maggu2810:
@kaikreuzer I understand that you would like to find a better / simpler way to write correct code. I don't know if there is any.

Perhabs @pkriens could join the discussion.

maggu2810:
@sjka Here is another link if you are interested in: http://njbartlett.name/files/osgibook_preview_20091217.pdf
See "11.6.5. Service Replacement"

pkriens:
Why do you use dynamic? It sounds like a rare event that you switch storage, not sure you need to make this dynamic? I would definitely not take this complexity until there is a very sound reason. With DS, you hardly notice the deactivation and activation and it tests your code to be a good citizen.

Second, use the annotations. That just saves you an enormous amount of grief. Especially with dynamic.

If you still want to use dynamic. The setter does not have to be protected. OSGi will always keep you updated of the latest copy so you can just set it. The unset must be protected because you can SCRs are allowed to set before unset. I.e. `set(a)`, `set(b)`, `unset(a)`. This was done to allow the service to have a value continuously.  Using the AtomicReference you can do this in one line.  However, if you need to do something on binding then the code you use looks ok for the unset. (As said, the set can just be an assignment.)

Last, but definitely not least, this design is very Eclipsish and notOSGish. Especially the use of a class loader is scary as hell in OSGi, you very easily create unmanaged dependencies. 

Therefore, if I had to do this from scratch I would register the proper storage service _instance_ instead of a factory. 

        @Designate(ocd=StorageImpl.Config.class, factory=true)
        @Component
        public class StorageImpl implements Storage {
               @interface Config {
                       String path();
                       String name();
                }
                Db db;

                @Activate
                void activate( Config config ) {
                       db = new Db( config.path());
                }
               ...
        }

This model is much more flexible and much easier to handle. Note that you can always make a specific DS component depend on a specific Storage service using the <refname>.target option. The given DS component always registers the name service property.

maggu2810:
@pkriens Thanks for joining this discussion.

> Why do you use dynamic? It sounds like a rare event that you switch storage, not sure you need to make this dynamic?

If the storage service with the higher service ranking is installed after another service is already bound, I don't see another option to grant that the higher service is used. Okay, someone could trigger a re-activation manually, but I would like that the highest ranking is used if one is present.

> Second, use the annotations. That just saves you an enormous amount of grief. Especially with dynamic.

I use annotations in all my non-Eclipse projects (just to note that I prefer the annotations in general).
How do you think it will make things easier in this case?
As the service itself is not really of interest but the an instance that is created by this variable, I thought need to synchronize the block. At least to grant that the storage service and the storage instance are kept in sync.

> If you still want to use dynamic. The setter does not have to be protected.

Just to ensure you know: We have to support OSGi 4.2 and so DS 1.1, so updating fields without using the function isn't supported.
But I assume you mean we could use a private method. Correct? that does not fit in this example as the class containing the `set` function is a base class of the component and so this requirement does not fit "Otherwise, if the method has the private access modifier, then access is permitted only if the method is declared in the component implementation class."

> Note that you can always make a specific DS component depend on a specific Storage service using the .target option. The given DS component always registers the name service property.

I has been told that `component.name` is a implementation detail used by Felix SCR and not part of the specification. But you are right, 112.6 of the 4.2 compendium states that SCR always adds that property.

pkriens:
Why do you prefer the higher ranking is used? I often hear this argument but it seems very fuzzy. Either the lower ranking is good enough or not. What changes when you have a higher ranking that makes you switch? There might be a requirement that only only one storage is available but that is a completely different issue.

BTW, the later DS's have a greedy option that will reactivate the component when a higher ranking service comes in. 

The biggest problem I see that you need to use an almost 10 year old spec combined with design patterns that do not fit the service model very well. This combination that you do need some ugly code sometimes.

maggu2810:
> Why do you prefer the higher ranking is used? I often hear this argument but it seems very fuzzy. Either the lower ranking is good enough or not. What changes when you have a higher ranking that makes you switch? There might be a requirement that only only one storage is available but that is a completely different issue.

The storage service is used by the framework to deserialize its "knowledge base" and restore / serialize it after a restart again (or at the time of usage) again.
There are two public storage services, one that uses MapDB internally another one that uses JSON in text files as storage format (and perhaps other storage services that are used by custom products).
If a product would like to switch from one storage service to another one (so, it is not really something that the user chooses but the distributor) there are several ways to handle the conversion of existing storage in the old format.
I assume the above one (https://github.com/eclipse/smarthome/pull/3617#issuecomment-307302031) is also valid. It can reuse the storage implementation itself. Using the service ranking the distributor can force that this storage service is used at runtime.
So, in this case not every service suffers the needs, as one format (the one the distributor chosen) should be used from time_a and all other data that has been previously written should be migrated transparently.

> BTW, the later DS's have a greedy option that will reactivate the component when a higher ranking service comes in.

> The biggest problem I see that you need to use an almost 10 year old spec combined with design patterns that do not fit the service model very well. This combination that you do need some ugly code sometimes.

You are right. But that's another topic. I assume a lot of people would prefer to change the project requirements. But as long as products / distributors have to use the 4.2 spec I don't think it will be changed.

> Second, use the annotations. That just saves you an enormous amount of grief. Especially with dynamic.

Just to continue this topic again (please correct me if I am wrong):
The usage of annotations could simplify the code and remove boilerplate constructs. I agree.
But it does not solve any concurrency at all. I still have to use volatile, atomics, synchronize, ...
Correct?

pkriens:
About the annotations, without the field support we added for them there is no difference. You're stuck.

Understanding the problem better, an even better design than the Storage service (i.e. without factory) is to use a push model. Instead of the component fetching its configuration, the DB could push the configuration into the component. Then conversions would be not a problem. A component could then not detect there were multiple storage services. This is more or less the mode OSGi Configuration Admin uses, it significantly simplifies components.

pkriens:
a) OSGi Design patterns, love to get more involved in that. So often see design patterns that are hard in OSGi while much (and i mean MUCH) simpler solutions exist. But agree, it is outside the scope of this discussion
b) The setter is always trivial. Only when you go dynamic you need to take the 'unset' after 'set' into account which makes it more tricky. Design patterns as used with the factory make this painful. Again, just make the references static and you never see any of those details. 
c) Static access has no concurrency issues unless you cause them yourself by creating threads in the activate method. Every service 'get' goes through a lock that ensures that activate is only called once. This lock properly synchronizes any thread that accesses the object through the service registry.

kaikreuzer:
> Only when you go dynamic you need to take the 'unset' after 'set' into account which makes it more tricky.

@pkriens What exactly means "after" here? Does DS make any statement regarding concurrency of the `set(a)`, `set(b)` and `unset(a)` calls? I would expect that they are done "after" each other, i.e. single threaded / synchronized by DS.

@maggu2810 IF this assumption is correct, I would claim that we neither need `volatile` nor the synchronized blocks in this PR, since we only allow DS to use the set/unset methods and the passed reference is not used anywhere else within our code. Would you agree?

amitjoy:
Just my two cents.

According to the app notes on OSGi concurrency [1], the DS strictly follows a happens before relationship among the DS states. As @kaikreuzer assumed, there will be strict ordering if we use static references. And the use of synchronization constructs, for example, `volatile` or `synchronized` blocks will not add any benefit. In case of dynamism, it is not defined clearly through. That is, `set(b)` can be invoked before `unset(a)` and that is why, we need to check if the service reference to be unbound is as same as the one bound previously. This is my understanding.

[1] - http://enroute.osgi.org/appnotes/concurrency.html

pkriens:
@kaikreuzer  DS cannot call unset before set but it can (in the case of dynamic) do

         set(a); set(b); unset(a) 

Purpose is to keep the service always with a proper service. Disadvantage is that you need to be careful not to throw away your new reference.

kaikreuzer:
@pkriens This is completely clear and that's why my question was: Isn't the "check to not accidentially throw away the new instance" the ONLY change that we actually require in this PR?

pkriens:
The code looks a bit cumbersome (AtomicReferences are easier) and there is volatile + lock on the object but I think there is no flaw in it.( I am not deep into this PR so I could miss the actual issue you refer to, you can contact me on @pkriens on Skype to discuss the finer details :-) )

kaikreuzer:
> volatile + lock on the object but I think there is no flaw in it

My point is that this complicates the code for no reason. My claim is that all is fully correct if these parts are removed. But let's wait for @maggu2810's opinion here.

maggu2810:
Isn't the topic "synchronized", "volatile", "atomics" already explained here:
https://github.com/eclipse/smarthome/pull/3617#issuecomment-307367283

kaikreuzer:
Maybe, but if so, I do not understand it.
Could you explain a scenario where something goes wrong if we hadn't volatile+synchronize (in this specific code)?

pkriens:
I feel a bit daft but if you remove the lock then you you will run into the 'unset kills new' problem?? It is a very small window but it can hit you

pkriens:
Just had a talk with BJ. I think we concluded that the interpretation  SHOULD be that there is never any overlap between bind/unbind operations. So in that case you can get rid of the synchronized but you must keep the volatile. However, SHOULD is not MUST. The spec is perfectly clear that 

    bind(a) bind(b) unbind(a) 

Is always non-overlapping. This seems to imply that all calls are non-overlapping but I am not a 100% sure in the case of bind(a) bind(b) unbind(a) bind(c). I've therefore always programmed defensively with AtomicReferences but it is very, very, likely not needed. However, should is should and not must.

The boilerplate is quite large because you the factory and the instance are two objects so atomic references don't work well. You could of course make a utility class (Binder) to simplify this case. A demo:

        public class Demo {
	   interface Foo {}
	   interface FooFactory { Foo make(); }
	
           public class Binder<F, I> {
		volatile  I	        instance;
		F			factory;

		public void bind(F f, I i) {
			I close;
			synchronized (this) {
				close = instance;
				this.instance = i;
				this.factory = f;
			}
			if (close != null) {
				destroy(close);
			}
		}
		public void unbind(F f) {
			I close;
			synchronized (this) {
				if (factory == f) {
					close = instance;
					this.instance = null;
				} else
					close = null;
			}
			if (close != null) {
				destroy(close);
			}
		}
		public void destroy(I instance) {}

		public I get() {
			return instance;
		}
	}

	Binder<FooFactory,Foo> binder = new Binder<>();

        void bind(FooFactory ff) {
		binder.bind(ff, ff.make());
	}

	void unbind(FooFactory ff) {
		binder.unbind(ff);
	}

	public void use() {
		if (binder.get() != null) {
			// ...
		}
	}
     }

kaikreuzer:
> So in that case you can get rid of the synchronized but you must keep the volatile. 

Ok, glad to hear that we agree that the synchronized can be removed.

So next up: The volatile. What is the argument why it is needed (not in general, but here in this specific PR)?

pkriens:
The thread that is used to bind/unbind is unlikely different from the thread that uses the set variable. Without volatile the variable change could be invisible or even parts of the contents of the objects could be invisible to the using thread.

And removing the synchronized is not guaranteed to work, just extremely unlikely to fail. I will file a bug to get this corner case clarified in R7. 

I think I recommend to create a Binder class utility for this factory case. Synchronize it as suggested and then when R7 comes out you can remove the synchronize. Better safe than sorry imho. No such utility class exists because I think the pattern to use a factory as service is not very common in OSGi.

kaikreuzer:
> Without volatile the variable change could be invisible 

Not for `storageService`as this is only used in set/unset. So imho no need here.
For `discoveryResultStorage` it could be invisible for a short moment in time - which simply means that we will "notice" the change a bit later only and keep using the old storage; which is not different to e.g. using a copy of the reference in our methods, so again, I do not see a scenario where this is an issue.

> or even parts of the contents of the objects could be invisible to the using thread.

Are you saying that without volatile, you can see completely broken objects that consist half of the one instance and the rest of another instance? I would be hugely surprised if that could happen in Java. Do you have any pointers?

> Better safe than sorry imho.

No, better avoid noise instead of implicitly implying to the reader that there is an important reason for doing it. Otherwise we could probably put 90% of our code in synchronized blocks, just to be safe...

sjka:
Maybe goes a little too far here (but hey, we are already anyway...).

Let me quickly sketch up what I understood should happen:

```
[DS-Thread] calls `setReference(Reference 1)`
[DS-Thread] calls `activate()`
[...]
[DS-Thread] calls `setReference(Reference 2)`
[DS-Thread] calls `unsetReference(Reference 1)`
[HTTP-Thread] calls `someMethodThatReliesOnReference()`
```

So I agree, we would need to guard the unset method against the set->set->unset problem. Fine. Apart from that, there is a theoretic risk that `someMethodThatReliesOnReference` still sees `Reference 1`,  correct? I have never seen this happening in real-life. And if so, I have to second @kaikreuzer here that this is no different than obtaining a local copy (remember, it's a 1..1 cardinality).

By the way, how is the above example different to 

```
[DS-Thread] calls `setReference(Reference 1)`
[DS-Thread] calls `activate()`
[HTTP-Thread] calls `someMethodThatReliesOnReference()`
```

? Isn't this the same situation, no matter if the reference is static or dynamic? 

So either an object which gets called by multiple threads must have _all_ of its member variables be declared `volatile`, or I still didn't understand the theory right. And then: how comes it works so very reliable without? My estimate would be that ~97% of the world-wide java code would be wrong in that aspect...?

I never experienced a real issue which I could track back to a non-volatile variable problem. Even in this project: we are dealing with several threads, scheduled jobs, HTTP request etc. On the other hand, a quick search only gave me 18 variables that are declared volatile. Which I'm pretty sure is by far not enough - only from a statistical point of view. So we certainly missed a lot.

Now, while I always _knew_  the theory, I yesterday tried playing around with `volatile` (or rather: non-volatile) a little. And I also knew that the JVM tries hard to optimize where a Thread gets run in order to cause the least trouble possible. However, what I was not aware of so far is _how hard_ it tries and _how good_ it is at it. What I want to say with it: I failed. I couldn't cause such an issue artificially.

Does anybody here have a working (in the sense of visibly broken) example?

sjka:
Oh, never mind, I finally found one: https://dzone.com/articles/java-volatile-keyword-0
Apparently keeping one thread really busy does the trick.

pkriens:
True, factory storageServiceas  does not require to be volatile but the instance discoveryResultStorage  does. There is **no** guarantee for a maximum time here. True, generally there is some barrier that synchronizes it but it is not hard to create a situation that never picks up the change. And volatile is quite cheap nowadays.

You're not the only one surprised by the behavior of seeing partially constructed objects. A famous example is the double checked locking problem that clearly demonstrates you could see a half constructed object. See the wikipedia for an explanation: https://en.wikipedia.org/wiki/Double-checked_locking (see the 3rd Java example which explains what is happening.) The https://www.cs.umd.edu/~pugh/java/memoryModel/jsr-133-faq.html which also discussed the erroneous double checked locking. And I agree, it is scary as hell if you read a lot of Java open source code. We're lucky that Java has catchable exceptions and that it most of the works. (When I first read about it I was paralysed writing Java code and it was scary to look at code I'd written before.)

For an example that demonstrates the invisible var: 

   public class CTest {
	boolean stop = false;
	@Test public void ctest() throws InterruptedException {
		Thread t = new Thread(() -> {
			while (!stop) {
			}
		});
		t.start();
		Thread.sleep(100);
		stop = true;
		t.join();
	}
    }

This test never ends when you run it outside a debugger. Turning stop to a volatile makes it end immediately.

Anyway. Putting 90% of your code in synchronized blocks is in itself a guarantee for a disaster, you have a deadlock as big as a house. Always try to avoid calling external code when synchronized.

Concurrency is really hard. Even after 40 years in the business I can spent hours looking at 45 lines. (I am today trying to use Alloy to _proof_ my theories are correct.) However, there are many ways to simplify life. Most important is to minimize shared state. DS is quite nice that if you use it statically things work quite well without synchronization. I always strongly recommend this model for this reason.

Once you go to dynamic mode the state that changes and is observable from multiple threads **must** be properly protected. There is no excuse not to do this because someone will trip over it and you will not be able to diagnose the problem. I do share your quest for keeping the code simple but I think it is paramount for software to handles these cases. It is better to be safe than sorry. It is already hard to enough to make reliable systems.

pkriens:
BTW, a lot of this is in http://enroute.osgi.org/appnotes/concurrency.html

sjka:
Okay, I see. Many thanks! 

Still, this example again only works because of the infinite loop that busily does nothing and thereby does not give the JVM/OS a chance to synchronize the memory. 

With respect to service injection, this simply is not the case here. In fact we actually need to do the exact opposite in OSGi by getting a local copy of the service reference, to make sure the variable does not change at a bad moment in time. So I agree that in situations where a thread is doing nothing than spendings its time in an uninterrupted loop, it would need to take precautions. However, in cases of such a Storage which gets used occasionally, i.e. every now and then (REST call, event trigger, scheduled job, etc), I does not look like the reference will get _never_ updated. And a tiny latency, as I said, won't cause any harm as this is what we need to protect ourselves against anyway.

maggu2810:
> Once you go to dynamic mode the state that changes and is observable from multiple threads must be properly protected. There is no excuse not to do this because someone will trip over it and you will not be able to diagnose the problem. I do share your quest for keeping the code simple but I think it is paramount for software to handles these cases. It is better to be safe than sorry. It is already hard to enough to make reliable systems.

I fully agree with @pkriens

> ? Isn't this the same situation, no matter if the reference is static or dynamic?

I asked it already and @pkriens replied to it here https://github.com/eclipse/smarthome/pull/3617#issuecomment-308117456 -- see (c).

If a static reference is used the service gets deactivated.
All services using a static reference to the deactivated service gets deactivated.
The first one gets a new reference injected and get's activated again.
All service using a static reference to the now activate service gets activated again.
As this actions are done with using synchronization, all is fine.

pkriens:
> Still, this example again only works because of the infinite loop that busily does nothing and thereby does not give the JVM/OS a chance to synchronize the memory.

I think this is phrased wrong. It is not about _being to busy_, it is about having no after/before relation with the code that modifies stop. Crossing a memory barrier will relatively become more and more expensive over time when you get more cores and rarer when the core's caches get bigger. And this is the only direction left for CPU vendors since we've maxed out on speed. Realize that without these caches your code would be thousands of times slower.

According to the Java memory model the processor has no reason to sync its core cache until it crosses a barrier. And note that it must be the _same_ memory barrier. I.e. you can have two threads that exchange data with proper synchronization that are not observed by a third thread that never passes that specific barrier.

And please do not forget it is not just the update of a reference, the chance of a partially constructed object is extremely rare but real. At billions of instructions per second on thousands of computers you're guaranteed to run into it. And these problems are incredible hard to test and even worse to diagnose.

I am a bit puzzled I actually have to defend this :-(

kaikreuzer:
> And I agree, it is scary as hell if you read a lot of Java open source code

Haha, but I hope that you know that it is often even more scary to read a lot of Java **closed** source code 😎 

So thanks everyone for the interesting discussion.

To summarize:
- synchronization can be removed from this PR
- volatile on storageService can be removed
- We should actually check ALL ESH code and add volatile to ANY field that is potentially accessed through different threads (what might be maaaany...). Is that my correct interpretation? Reading [this article](https://www.ibm.com/developerworks/java/library/j-jtp06197/), I am not fully convinced as it seems that volatile is mainly used if synchronization is important and otherwise it isn't really mandatory, even in multi-threaded environments...

pkriens:
@kaikreuzer The synchronized is actually officially necessary because the spec says nothing about it today, which will likely change. 

I think you're still too casual about this though. Ok, now I shut up.

kaikreuzer:
@pkriens The word you were looking for is pragmatic, not casual ;-)

kaikreuzer:
Sorry, @maggu2810 I lost this from sight due to vacation season. Would you agree with the summary in https://github.com/eclipse/smarthome/pull/3617#issuecomment-312294073? If so, please update the PR accordingly and we can finally merge it!

kaikreuzer:
@maggu2810 ping!

maggu2810:
Hi Kai,
I am pretty busy at the moment, so really no time for this topic. But it is still in my todo list (priority queue) :wink:

kaikreuzer:
Ok, no worries, leave your time :-)

maggu2810:
> The synchronized is actually officially necessary because the spec says nothing about it today, which will likely change.

So, removing volatile from the storage service should be the only change. Shouldn't it?

maggu2810:
If we don't like a sync object we could change the code to use synchronized set and unset methods:
```Java
protected synchonized void setStorageService
```

kaikreuzer:
> So, removing volatile from the storage service should be the only change. Shouldn't it?

Well, you need to read the statement between the lines. It should be actually interpreted as "The synchronized is actually practically not necessary and the spec will likely be change to make this explicitly clear.".

maggu2810:
removed sync

maggu2810:
resolved merge conflict caused by #4102

kaikreuzer:
Travis failed with a time-out, but the PR looks good, so let's merge it before the next merge conflicts occur.

Thanks and sorry again to have that open for so long!

PR #3618, opened 2017-06-08
add storage.format property to StorageService implementations

maggu2810:
This property could be consumed e.g. by a filter to get a specific storage service injected.


sjka:
Is it actually feasible to have both available at the same time? I see no usage which would be prepared for that situation, and also I doubt that the consumer of the StorageService should decide which one it _wants_ to use, I would rather expect the end user or "distro builder" to make this decision. One the other hand it does not hurt to have this property, of course. I'm fine with it, just curious for the use-case you have in mind...

maggu2810:
> One the other hand it does not hurt to have this property, of course.

Exactly :wink: 

> I'm fine with it, just curious for the use-case you have in mind...
> I would rather expect the end user or "distro builder" to make this decision.

It is exactly what I want to do. The distro builder make the decision but use a transparent migration from one StorageService to the other one (and so use the other implementations):
See example: https://github.com/eclipse/smarthome/pull/3617#issuecomment-307302031

PR #3628, opened 2017-06-09
[Yahoo] Changed use of caching utils to ESH ExpiringCacheMap

### LINKED ISSUE #3625: [Yahoo] Change used cache utils to ESH ExpiringCache
Follow up to PR #3572

cweitkamp:
* Removed individual caching utils
* Added reusable `YahooWeatherConnection` class

Closes #3625 


extensions/binding/org.eclipse.smarthome.binding.yahooweather/src/main/java/org/eclipse/smarthome/binding/yahooweather/internal/connection/YahooWeatherConnection.java
```diff
@@ -0,0 +1,47 @@
...
+    private final Logger logger = LoggerFactory.getLogger(YahooWeatherConnection.class);
+
+    private final static String WEBSERVICE_URL = "https://query.yahooapis.com/v1/public/yql?format=json";
+
+    public String getResponseFromQuery(String query) {
+        try {
+            final URL url = new URL(
+                    WEBSERVICE_URL + "&q=" + URLEncoder.encode(query, StandardCharsets.UTF_8.toString()));
+            final URLConnection connection = url.openConnection();
+            return IOUtils.toString(connection.getInputStream());
```
maggu2810: Do you need to close the input or output stream?

WRT:
Invoking the close() methods on the InputStream or OutputStream of an URLConnection after a request may free network resources associated with this instance, unless particular protocol specifications specify different behaviours for it.

hakan42: Kai suggested try-with-resources in my sensebox binding:

try (InputStream inputStream = connection.getInputStream()) {
  do-some-stuff
} catch (IOException ioe) {
  switch-binding-to-offline
}

cweitkamp: @hakan42 May I ask you to link where I can find your code? I think we should go the same way to solve this.

sjka: I'm well aware that you just moved the existing code, but as we are talking about it already... (Thanks for bearing with us 😉 )
Isn't hat a perfect use-case for `return HttpUtil.executeUrl(...)` instead of fiddling around with these nasty UrlConnections? Or did I miss anything?

cweitkamp: Of course it is. I am going to change that.

I just looked for a way to get rid of the `org.apache.commons.io.IOUtils` dependency and your suggestion seems to be the solution. Thanks for pointing me into that direction.

cweitkamp: @maggu2810 @sjka I just pushed my changes. The binding uses `org.eclipse.smarthome.io.net.http.HttpUtil` now.

kaikreuzer:
Many thanks @cweitkamp!

extensions/binding/org.eclipse.smarthome.binding.yahooweather/src/main/java/org/eclipse/smarthome/binding/yahooweather/handler/YahooWeatherHandler.java
```diff
@@ -55,6 +51,12 @@
 
     private final int MAX_DATA_AGE = 3 * 60 * 60 * 1000; // 3h
     private final int CACHE_EXPIRY = 10 * 1000; // 10s
+    private final ExpiringCacheMap<String, String> CACHE = new ExpiringCacheMap<String, String>(CACHE_EXPIRY);
```
martinvw: @cweitkamp is there a reason why all these non-final's are uppercase?

See also my comments at: https://github.com/openhab/openhab2-addons/pull/2361#discussion_r121472369

cweitkamp: @martinvw Good point. I think we failed to see that. I will fix it in a follow-up PR.

PR #3510, opened 2017-05-23
Category for Thing types

triller-telekom:
Added "category" to thing types and group channel types

Fixes #568

PR #3629, opened 2017-06-09
Added category to group channel types

triller-telekom:

### DIFF
--- a/bundles/core/org.eclipse.smarthome.core.thing.xml.test/src/test/resources/test-bundle-pool/ChannelTypesTest.bundle/ESH-INF/thing/thing-types.xml
+++ b/bundles/core/org.eclipse.smarthome.core.thing.xml.test/src/test/resources/test-bundle-pool/ChannelTypesTest.bundle/ESH-INF/thing/thing-types.xml
@@ -26,7 +26,7 @@
 	
 	<channel-group-type id="channelgroup">
 		<label>Channel group</label>
-		<category>TEMPERATURE</category>
+		<category>Temperature</category>
 		<channels>
 			<channel id="channel1" typeId="channel1" />
 			<channel id="channel2" typeId="channel1" />
--- a/bundles/core/org.eclipse.smarthome.core.thing.xml.test/src/test/resources/test-bundle-pool/ChannelTypesTest.bundle/ESH-INF/thing/thing-types.xml
+++ b/bundles/core/org.eclipse.smarthome.core.thing.xml.test/src/test/resources/test-bundle-pool/ChannelTypesTest.bundle/ESH-INF/thing/thing-types.xml
@@ -26,6 +26,7 @@
 	
 	<channel-group-type id="channelgroup">
 		<label>Channel group</label>
+		<category>TEMPERATURE</category>
 		<channels>
 			<channel id="channel1" typeId="channel1" />
 			<channel id="channel2" typeId="channel1" />
--- a/bundles/core/org.eclipse.smarthome.core.thing/src/main/java/org/eclipse/smarthome/core/thing/type/ChannelGroupType.java
+++ b/bundles/core/org.eclipse.smarthome.core.thing/src/main/java/org/eclipse/smarthome/core/thing/type/ChannelGroupType.java
@@ -24,6 +24,7 @@ public class ChannelGroupType extends AbstractDescriptionType {
 
     private final boolean advanced;
     private final List<ChannelDefinition> channelDefinitions;
+    private String category;
 
     /**
      * Creates a new instance of this class with the specified parameters.
@@ -45,11 +46,12 @@ public class ChannelGroupType extends AbstractDescriptionType {
      * @throws IllegalArgumentException if the UID is null, or the label is null or empty
      */
     public ChannelGroupType(ChannelGroupTypeUID uid, boolean advanced, String label, String description,
-            List<ChannelDefinition> channelDefinitions) throws IllegalArgumentException {
+            String category, List<ChannelDefinition> channelDefinitions) throws IllegalArgumentException {
 
         super(uid, label, description);
 
         this.advanced = advanced;
+        this.category = category;
 
         if (channelDefinitions != null) {
             this.channelDefinitions = Collections.unmodifiableList(channelDefinitions);
@@ -62,7 +64,7 @@ public class ChannelGroupType extends AbstractDescriptionType {
      * Returns {@code true} if this {@link ChannelGroupType} contains advanced functionalities
      * which should be typically not shown in the basic view of user interfaces,
      * otherwise {@code false}.
-     * 
+     *
      * @return true if this channel group contains advanced functionalities, otherwise false
      */
     public boolean isAdvanced() {
@@ -78,6 +80,10 @@ public class ChannelGroupType extends AbstractDescriptionType {
      */
     public List<ChannelDefinition> getChannelDefinitions() {
         return channelDefinitions;
+    }
+
+    public String getCategory() {
+        return this.category;
     }
 
     @Override
--- a/bundles/core/org.eclipse.smarthome.core.thing/src/main/java/org/eclipse/smarthome/core/thing/type/ChannelGroupType.java
+++ b/bundles/core/org.eclipse.smarthome.core.thing/src/main/java/org/eclipse/smarthome/core/thing/type/ChannelGroupType.java
@@ -48,7 +48,7 @@ public class ChannelGroupType extends AbstractDescriptionType {
      * @throws IllegalArgumentException if the UID is null, or the label is null or empty
      */
     public ChannelGroupType(ChannelGroupTypeUID uid, boolean advanced, String label, String description,
-            final String category, List<ChannelDefinition> channelDefinitions) throws IllegalArgumentException {
+            String category, List<ChannelDefinition> channelDefinitions) throws IllegalArgumentException {
 
         super(uid, label, description);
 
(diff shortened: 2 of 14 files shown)


bundles/core/org.eclipse.smarthome.core.thing/src/main/java/org/eclipse/smarthome/core/thing/type/ChannelGroupType.java
```diff
@@ -45,11 +46,12 @@
      * @throws IllegalArgumentException if the UID is null, or the label is null or empty
      */
     public ChannelGroupType(ChannelGroupTypeUID uid, boolean advanced, String label, String description,
-            List<ChannelDefinition> channelDefinitions) throws IllegalArgumentException {
+            String category, List<ChannelDefinition> channelDefinitions) throws IllegalArgumentException {
```
martinvw: Just a quick review comment, I saw it was marked wip but it caught my eyes, the category field should imho also be final and it is missing from the constructor its docblock.

bundles/core/org.eclipse.smarthome.core.thing.xml.test/src/test/resources/test-bundle-pool/ChannelTypesTest.bundle/ESH-INF/thing/thing-types.xml
```diff
@@ -26,6 +26,7 @@
 	
 	<channel-group-type id="channelgroup">
 		<label>Channel group</label>
+		<category>TEMPERATURE</category>
```
martinvw: Jus curious what is the preferred case here the PR contains both pure uppercase TEMPERATURE and more human readable Temperature

kaikreuzer: Human readable "Temperature" as it is defined in https://www.eclipse.org/smarthome/documentation/development/bindings/thing-definition.html#channel-categories!

docs/documentation/development/bindings/categories.md
```diff
@@ -0,0 +1,68 @@
...
+
+## Thing Categories
+
+The thing type definition allows to specify a category. User interfaces can parse this category to get an idea how to render this thing. A binding can classify each thing into one of the existing categories. The list of all predefined categories can be found in our categories overview:
+
+| Category        | Description                                          |
+|-----------------|------------------------------------------------------|
+
+
+## Channel Categories
```
kaikreuzer: As you have copied this list from an existing page, shouldn't you remove it from the old location (and merely link to the new page where appropriate)?

docs/documentation/development/bindings/thing-definition.md
```diff
@@ -231,12 +231,13 @@ Inside the thing types XML file channel groups can be defined like this:
 </thing-type>    
 ```
 
-The channel group type is defined on the same level as the thing types and channel types. The group type must have a label and an optional description. Moreover the list of contained channels must be specified:
+The channel group type is defined on the same level as the thing types and channel types. The group type must have a label, an optional description, and an optional category. Moreover the list of contained channels must be specified:
```
kaikreuzer: Could you make "category" a link to directly point to the category page?

docs/documentation/development/bindings/categories.md
triller-telekom: Done

docs/documentation/development/bindings/thing-definition.md
```diff
@@ -171,48 +171,7 @@ The user interface can use these values to render labels for values or to provid
 
 ### Channel Categories
```
kaikreuzer: As you have removed the content, I don't think that a headline still makes sense here. Rather check how the information can be put within the existing docs on this page (or any other).

kaikreuzer:
Would be nice if you could also rebase the branch, so that there are no conflicts anymore.

docs/documentation/development/bindings/thing-definition.md
```diff
@@ -169,51 +170,6 @@ Some channels might have only a limited and countable set of states. These state
 
 The user interface can use these values to render labels for values or to provide a selection of states, when the channel is writable. The option labels can also be localized.
 
-### Channel Categories
```
kaikreuzer: I just found a link to this section on https://www.eclipse.org/smarthome/documentation/features/icons.html (which is currently already broken) - I guess that needs to be changed as well. Did you check if there are any other references that need to be updated?

bundles/core/org.eclipse.smarthome.core.thing/src/main/java/org/eclipse/smarthome/core/thing/type/ChannelGroupType.java
```diff
@@ -39,17 +40,20 @@
...
+     * @param category the category of this channel group type, e.g. Temperature (could be null or empty)
+     *
      * @param channelDefinitions the channel definitions this channel group forms
      *            (could be null or empty)
      *
      * @throws IllegalArgumentException if the UID is null, or the label is null or empty
      */
     public ChannelGroupType(ChannelGroupTypeUID uid, boolean advanced, String label, String description,
-            List<ChannelDefinition> channelDefinitions) throws IllegalArgumentException {
+            final String category, List<ChannelDefinition> channelDefinitions) throws IllegalArgumentException {
```
kaikreuzer: Any special reason why you just added "final" here?

docs/documentation/features/icons.md
```diff
@@ -12,7 +12,7 @@ Bundle: `org.eclipse.smarthome.ui.icon`
 
 Eclipse SmartHome comes with a flexible infrastructure for handling icons that are to be used within user interfaces.
 This bunde registers a servlet under the url `/icon/`, which can then be easily queried for icons using GET requests of the form `/icon/<category>?state=<state>&format=[png|svg]&iconset=<iconsetid>`, where 
-- `category` is one from the [list of channel categories](../../development/bindings/thing-definition.html#channel-categories) or any custom category that might be used within the solution
+- `category` is one from the [list of channel categories](../../development/bindings/categories.html#channel-categories) or any custom category that might be used within the solution
```
kaikreuzer: Looking at why the link is broken today already, I would guess the correct link would be `../development/bindings/categories.html#channel-categories`.

bundles/core/org.eclipse.smarthome.core.thing/src/main/java/org/eclipse/smarthome/core/thing/type/ChannelGroupType.java
triller-telekom: [Martins comment](https://github.com/eclipse/smarthome/pull/3629#discussion_r121996976). because I will assign it to a final variable afterwards anyway.

kaikreuzer: Afaics, he mentioned the field, not the variable. Just like the fields for advanced and channelDefinitions are final, while the method parameters are not.

kaikreuzer:
Question: Does this PR need to be labelled "APIBreaking"? If so, do we all agree that it is worth it?

maggu2810:
The new `categories.md` already contains a section "Thing Categories" (no category is named, but stated that there is support for). Why do we add this section if there is currently no category support for things at all? Is this comes with a following PR, then this section could be added later.

I assume the the xml-reference file is missing in this PR.

Just to summarize some (well known) stuff:

* There are channels and channel-types. Channels are "instances" of channel-types. There is no category per channel but the category is set for /inherited from the channel type.
* There are channels-groups and channel-group-types. Channels groups are "instances" of channel-group-types. There is no category per channel group but the category is set for /inherited from the channel group type.
* A channel group type contains channels, not channel types.

What is the relationship between the categories?
If we define a channel-group-type that contains channels of channel-types using different categories?
Is this valid?

If you think about thing categories, will it be possible to mix different categories for thing, channel-group, channel, that does not make sense at all? Is it up to the one that defines that types?

maggu2810:
1 day ago
> If so, do we all agree that it is worth it?

3 hours ago
> merged commit 0c0487e into eclipse:master

Give me some time please :wink:

kaikreuzer:
No need to give you time, since you have missed https://github.com/eclipse/smarthome/pull/3629/commits/f1fcd000b6855cde513376ab413860d87d881fdf, which made the question outdated :-)

triller-telekom:
> I assume the the xml-reference file is missing in this PR.

@maggu2810 Why do you think this file is missing? It was part of this PR and I added the category in the example [here](https://github.com/eclipse/smarthome/blame/master/docs/documentation/development/bindings/xml-reference.md#L410) and to the text [here](https://github.com/eclipse/smarthome/blame/master/docs/documentation/development/bindings/xml-reference.md#L487)

kaikreuzer:
What about @maggu2810's other questions? Would it make sense to improve the documentation in that regard?

maggu2810:
>> I assume the the xml-reference file is missing in this PR.

> @maggu2810 Why do you think this file is missing?

Long days, short nights, lots of coffee, ... :wink: 
Sorry, missed it in the diff view, but you are right, it has been part of this PR.

triller-telekom:
The section "Thing Categories" in the file `categories.md` is there by accident because I had the two PRs, group channel categories and thing type categories in one PR at first and then split them. Shall I remove the empty "Thing Categories" section in a follow-up PR for now?

Regarding the other questions: I would say a mixing of different categories does not make sense, i.e. if you define a "group channel type", all its containing channels should be instances of channel types of the same category. Though in theory we have something like a "default category" on the "channel group type" and allow its containing channels to be of "any" type and if they do not define one, the type will be the default one.

But the more I think about these questions now, the more I think we should have discussed this PR more thoroughly before merging ;)

If we think of how the thing category and the channel categories, this should be up to the one that defines them. In my opinion a thing can be in thing type category A, where its channels can be in channel type category B and C for example. Maybe we should continue that discussion on the related PR #3510.

maggu2810:
> Shall I remove the empty "Thing Categories" section in a follow-up PR for now?

Hm, perhaps we could keep it if you plan to introduce it soon.

triller-telekom:
It's part of PR #3510

PR #3631, opened 2017-06-10
Basic UI: fix update of page title

lolodomo:


lolodomo:
For information, this PR only fixes the problem when event source (SSE ?) is used.

@resetnow : Is the long polling mode still relevant ? If it is, several things are wrong in this mode in my opinion. For example setValue is called like this: `w.setValue(value, value);` while value is the item state. The first parameter should be the value extracted from the label (meaning with a possible transformation), like we handle it in the SSE part.

lolodomo:
@resetnow : in the back, when the page is first loaded, a HTML escape is applied to the text used to set the title. Is it required ? Should I apply a similar escape of HTML characters in the front before calling smarthome.UI.setTitle ? If I have to, can you tell me if there is a common JavaScript function to do that ?

When an event is handled, there is no escape of HTML characters applied to the value extracted from the label. Would it be something missing ? Or is it a different case ?

resetnow:
>Should I apply a similar escape of HTML characters in the front before calling smarthome.UI.setTitle

I think yes. It does have some XSS potential. There is no such function in JS. I would use the one provided by Angular (https://stackoverflow.com/a/30558011/7303703). However, there are license implications. It's MIT licensed.

@kaikreuzer how would one go about including a part of Angular code here?

kaikreuzer:
> except for the comment below.

Hey, you added that part to your comment while I was merging already :-)

resetnow:
Yes, sorry about that, I'm using the code review feature in a wrong way. It's not critical and we should be fine as long as we agree on escaping solution. I'll provide a fix.

kaikreuzer:
> how would one go about including a part of Angular code here?

You mean copying and not "using as a lib"? We should try to avoid copying code - it would be better if you could rewrite the necessary parts...

resetnow:
We could use it as a separate file but I would still have to edit out the rest of the library as we don't need it. I don't think I could rewrite in a significantly different way.

resetnow:
On the other hand, we only need to replace &lt; and &gt; to avoid XSS since we don't use the data anywhere except in DOM text node, so I'll go with that for the time being.

resetnow:
#3655 now ready. Sorry again for the misleading comment.

@lolodomo regarding longpolling clients: the only browsers without SSE support with a somewhat significant market share are IE and Edge. Longpolling update does not provide formatted values anyway, — only item state is supplied — so I'm not even sure how that would look or work. I'm inclined to say that IE users called that on themselves but even if we tried to fix that, it would require a lot of work — every widget will have to keep its format string and format it when a new item state arrives. I'd say leave it as it is.

lolodomo:
@resetnow: we should have now a field named transformedState in the response that could certainly be used for non SSE browsers. I will try to improove that.

Issue #3607, opened 2017-06-07
RulesJvmModelInferrer - Duplicate field


martinvw:
After being busy setting everything up (via the Eclipse Smarthome Designer) I got a lot of warnings when saving:

```
2017-06-07 20:11:34.649 [WARN ] [.rule.jvmmodel.RulesJvmModelInferrer] - Duplicate field: 'Washing_Machine_Plug'. Ignoring 'org.eclipse.smarthome.core.library.items.SwitchItem'.
2017-06-07 20:11:34.649 [WARN ] [.rule.jvmmodel.RulesJvmModelInferrer] - Duplicate field: 'Washing_Machine_Plug_Power'. Ignoring 'org.eclipse.smarthome.core.library.items.NumberItem'.
2017-06-07 20:11:34.649 [WARN ] [.rule.jvmmodel.RulesJvmModelInferrer] - Duplicate field: 'Washing_Machine_Plug_Energy'. Ignoring 'org.eclipse.smarthome.core.library.items.NumberItem'.
2017-06-07 20:11:34.649 [WARN ] [.rule.jvmmodel.RulesJvmModelInferrer] - Duplicate field: 'Z_CV_Closet_Battery'. Ignoring 'org.eclipse.smarthome.core.library.items.NumberItem'.
2017-06-07 20:11:34.649 [WARN ] [.rule.jvmmodel.RulesJvmModelInferrer] - Duplicate field: 'Z_CV_Closet_Temp'. Ignoring 'org.eclipse.smarthome.core.library.items.NumberItem'.
2017-06-07 20:11:34.649 [WARN ] [.rule.jvmmodel.RulesJvmModelInferrer] - Duplicate field: 'Keuken_Buiten'. Ignoring 'org.eclipse.smarthome.core.library.items.SwitchItem'.
2017-06-07 20:11:34.649 [WARN ] [.rule.jvmmodel.RulesJvmModelInferrer] - Duplicate field: 'Coolcam_Plug_Energy_W'. Ignoring 'org.eclipse.smarthome.core.library.items.NumberItem'.
2017-06-07 20:11:34.649 [WARN ] [.rule.jvmmodel.RulesJvmModelInferrer] - Duplicate field: 'Coolcam_Plug_Energy_V'. Ignoring 'org.eclipse.smarthome.core.library.items.NumberItem'.
[...]

PR #3641, opened 2017-06-12
improve handling of links related to model-items

### LINKED ISSUE #2491: BaseThingHandler:channelLinked() called frequently / too much
In a simple setup with a single Item -> Channel definition in the DSL, the above method is called three times (for the given channel) at startup of the runtime, twice triggered by ThingManager.added() and once by ItemRegistry.allItemsChanged(). 

Expected behaviour would be that the method is called once for each time an Item definition is parsed, taking into account the fact that "temporary" items are created. 

In my use case the channelLinked() triggers a read request on a bus (e.g. KNX), which is not designed to be flooded in this way, especially when reading out the same value multiple times.

sjka:
* change the order of how the GenericItemProvider registers links and items
* enable the GenericItemChannelLinkProvider to handle deletions correctly
* call GenericItemChannelLinkProvider from within GenericItemProvider only
* enforce correct transaction handling in GenericItemChannelLinkProvider

fixes #2491
relates to #3607


sjka:
please note that this needs #3639 merged (or rather #3607 to be fixed) first

sjka:
Hmm, just realized that my previous comment was wrong - the ChannelItemProvider is taken out of the game because of the optimized order wrt link & item creation.

Issue #3569, opened 2017-06-02
[Astro PaperUI] Radiation wrong, unit missing, negative values


NorbertHD:
Astro in PaperUI has some errors. When the altitude is not set, the radiaton values are reasonable.
![rad](https://cloud.githubusercontent.com/assets/8472266/26731465/7879cc84-47b4-11e7-8eb0-c2a68607c232.png)

But if I set the altitude to e.g. 100 m the values are way to high and diffuse radiation is negativ.
![rad2](https://cloud.githubusercontent.com/assets/8472266/26731481/8a938216-47b4-11e7-9c84-5d5630da6370.png)

Also the unit specifier for duration is missing, it should be "min".
![unit](https://cloud.githubusercontent.com/assets/8472266/26731560/be985eec-47b4-11e7-8e6e-19692046eb49.png)

When there is no "evening night" a duration of -1 is shown.
![negative](https://cloud.githubusercontent.com/assets/8472266/26731623/f10cfc34-47b4-11e7-8fa1-20ca54247abb.png)

sjka:
@gerrieg this looks like something for the expert

gerrieg:
The radiation has been implemented by @clinique 

@clinique 
Can you please check this and when you're there, can you please fix the other two small issues too?  😉 Thanks!

sjka:
Oh, right... Thanks!

clinique:
@gerrieg , sure, I'll do

clinique:
Pushing a PR adressing first two bugs, but the third one is as designed.

NorbertHD:
@clinique Thank you very much.

I have 3 other small issues:

"Moon age" and "Moon illumination" also don't have units. They should have "d" and "%".

The label of the "Full moon" channel-type is wrong (New moon):
```
	<channel-type id="fullMoon">
		<item-type>DateTime</item-type>
!!!!		<label>New moon</label>
		<description>The DateTime for full moon</description>
		<state readOnly="true" pattern="%1$tF %1$tR" />
	</channel-type>
```

clinique:
Ok, I'll integrate these small issues in the same PR

clinique:
PR was already merged.

NorbertHD:
I don't think that the PR fixes the altitude problem.
The parenthesis around the expression "(altitude / 1000)" makes it now an integer division.
So all altitude values between 0 and 999 m have the same result of 0.
When you set the altitude to 1000 m or above you should see wrong radiation values again.

[...]

PR #3642, opened 2017-06-12
Correction to Issue #3569

clinique:
Signed-off-by: Gaël L'hopital <glhopital@gmail.com>


clinique:
Aargh, you where too fast Kai, I had two more small tweaks to put in it

PR #3643, opened 2017-06-12
Make IncreaseDecrease implement State and add to Dimmer

cdjackson:
This allows DimmerType to use IncreaseDecrease as a State which is needed for "valueless" commands - ie the user presses a dimmer and the dimmer sends up/down rather than an actual level.

https://github.com/openhab/org.openhab.binding.zwave/issues/598

### DIFF
--- a/bundles/core/org.eclipse.smarthome.core.test/src/test/java/org/eclipse/smarthome/core/library/items/DimmerItemTest.java
+++ b/bundles/core/org.eclipse.smarthome.core.test/src/test/java/org/eclipse/smarthome/core/library/items/DimmerItemTest.java
@@ -13,6 +13,7 @@ import java.math.BigDecimal;
 
 import org.eclipse.smarthome.core.items.Item;
 import org.eclipse.smarthome.core.library.types.DecimalType;
+import org.eclipse.smarthome.core.library.types.IncreaseDecreaseType;
 import org.eclipse.smarthome.core.library.types.PercentType;
 import org.eclipse.smarthome.core.types.State;
 import org.junit.Test;
@@ -66,6 +67,17 @@ public class DimmerItemTest {
         final DimmerItem item = createDimmerItem(new PercentType(origin));
         final BigDecimal result = getState(item, DecimalType.class);
         assertEquals(origin.divide(new BigDecimal(100)).compareTo(result), 0);
+    }
+
+    @Test
+    public void getAsPercentFromIncreaseDecrease() {
+        DimmerItem item = createDimmerItem(IncreaseDecreaseType.DECREASE);
+        BigDecimal result = getState(item, PercentType.class);
+        assertEquals(new BigDecimal(0).compareTo(result), 0);
+
+        item = createDimmerItem(IncreaseDecreaseType.INCREASE);
+        result = getState(item, PercentType.class);
+        assertEquals(new BigDecimal(100).compareTo(result), 0);
     }
 
 }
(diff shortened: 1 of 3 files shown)


kaikreuzer:
No, sorry - INCREASE and DECREASE are definitely not states, but only commands. A Dimmer item only has a percentage as a state. The state is a "snapshot" it does not describe a transition (like "increasing").

cdjackson:
But why is this allowed then in the Rollershutter. Isn't it exactly the same there - except it's UpDown instead of IncreaseDecrease. There it is allowed as a state as well as a command and this is also stated in the documentation.

This seems inconsitent and forces us to use Rollershutter instead of Dimmer.

maggu2810:
UP means it is fully up, to the state itself signals the current position (same for DOWN). It does not mean "movin up" or "moving down" but the rollershutter is fully up or fully down.

This differs from increase / decrease. You don't know the state of the dimmer if the state is "increase".

maggu2810:
See also https://github.com/eclipse/smarthome/issues/3297

cdjackson:
Ok, so how do we represent a state as moving? I think there needs to be some way to represent this as an absolute value is not always possible and not having this eliminates a group of devices from being integrated into ESH.

kaikreuzer:
> Ok, so how do we represent a state as moving?

We don't. What you ask for is not info about a state, but about a state transition. We do not have that notion in ESH and I do not see any straightforward way to introduce this.

> not having this eliminates a group of devices

Why?

cdjackson:
> What you ask for is not info about a state, but about a state transition.

No - I don't agree. A state might be "rotating left" or "rotating right" or "stopped" (or increasing / decreasing". This might not be an absolute value, but it's still a state. We have some devices that are not linked to an absolute value.

Two examples.

In ZWave, there are devices that do not provide a number (eg 23%) - they only provide an increasing and decreasing state. I could try and convert these to a number somehow, but it's false, and assumes an implementation (ie it assumes we have a fixed max/min value). It's also not simple in ESH to do this in the binding since the binding has no visibility of an item.

The Nuimo provides a rotation sensor - again this provides an indication of speed and direction (using an endlessly changing number).

Maybe you have an idea how these can be incorporated but it's not obvious to me. To me, these are states - just like the player states (play, reverse etc).

kaikreuzer:
>  there are devices that do not provide a number (eg 23%) 

So clearly, you should not use a Dimmer item for them. As a matter of fact, the only useful item types are Number or String and in that case I would go for a String item. And for this, you can add a state description to the channel to say that the state can only have the options "Increasing" and "Decreasing". We won't be able to add new item types (and state types) for all such situations, so String types with additional meta-information is the way to go.

cdjackson:
For me this seems a bit wrong. We are now saying that each binding should just make its own types - this means that UIs can't be expected to know how to handle them, and different bindings will implement differently.

I agree that we can't cover every possible thing that someone may want. But I think we should try and have some abstract types that are usable across a broad range of use cases, and this should be standard functionality that is handled the same for all bindings and UIs. If we tell bindings to generate their own types using strings then we will introduce dependencies.

Would you accept a proposal for a generic spinner item type - defined as (let's say) a value of +/- 100 to indicate direction and speed. 0 would mean stopped, negative numbers would be down, or anti-clockwise, or decreasing and positive number clockwise, up, increasing (etc).

jimduchek:
This commit is very helpful, but I would suggest that IncreaseDecrease type needs to hold a velocity, not simply 'increase' or 'decrease'.  I am, however, not quite sure the best way to abstract that.  Perhaps as simple as adding an unsigned integer to the type in addition to the increase/decrease states (thus making the velocity an optional thing)... or getting rid of the increase/decrease states and having a signed integer be the state.  Either way I would make no attempt to define what the velocity 'is' (i.e. percent-per-second or anything), but let the user interpret it as necessary.

dastrix80:
Hi All
has there been any progress on this? user control via gestures/clicks is fundamental to a usable openhab system so I think this request is extremely important.

PR #3648, opened 2017-06-13
fix map access in (old) rule engine

sjka:
The object types simply did not match!

### DIFF
--- a/bundles/model/org.eclipse.smarthome.model.rule.runtime/src/org/eclipse/smarthome/model/rule/runtime/internal/engine/RuleTriggerManager.java
+++ b/bundles/model/org.eclipse.smarthome.model.rule.runtime/src/org/eclipse/smarthome/model/rule/runtime/internal/engine/RuleTriggerManager.java
@@ -526,7 +526,7 @@ public class RuleTriggerManager {
                 rules.add(rule);
             } else if (t instanceof ThingStateUpdateEventTrigger) {
                 ThingStateUpdateEventTrigger tsuTrigger = (ThingStateUpdateEventTrigger) t;
-                Set<Rule> rules = thingUpdateEventTriggeredRules.get(tsuTrigger);
+                Set<Rule> rules = thingUpdateEventTriggeredRules.get(tsuTrigger.getThing());
                 if (rules == null) {
                     rules = new HashSet<Rule>();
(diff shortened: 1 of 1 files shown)


bundles/model/org.eclipse.smarthome.model.rule.runtime/src/org/eclipse/smarthome/model/rule/runtime/internal/engine/RuleTriggerManager.java
```diff
@@ -574,7 +574,9 @@ public void removeRule(TriggerTypes type, Rule rule) {
                 }
                 break;
             case TRIGGER:
-                triggerEventTriggeredRules.remove(rule);
+                for (Set<Rule> rules : triggerEventTriggeredRules.values()) {
+                    rules.remove(rule);
```
triller-telekom: Are you sure that you can remove an item from this set while you are iterating over it?

maggu2810: I don't see any problem here. He iterates over a collection of set of rules and removed something from the rules.
He does not remove something from the collection he iterates.

sjka: RTFJD:

> Returns a Collection view of the values contained in this map. The collection is backed by the map, so changes to the map are reflected in the collection, and vice-versa. If the map is modified while an iteration over the collection is in progress (except through the iterator's own remove operation), the results of the iteration are undefined. The collection supports element removal, which removes the corresponding mapping from the map, via the Iterator.remove, Collection.remove, removeAll, retainAll and clear operations. It does not support the add or addAll operations.

sjka: So I _could_ remove something from the map I iterate over, if I wanted. But I don't, as @maggu2810 correctly pointed out.

triller-telekom: Sorry my mistake, you are right.

PR #3650, opened 2017-06-13
fixed some happy little accidents

sjka:

### DIFF
--- a/bundles/automation/org.eclipse.smarthome.automation.module.script.rulesupport/src/main/java/org/eclipse/smarthome/automation/module/script/rulesupport/internal/delegates/SimpleConditionHandlerDelegate.java
+++ b/bundles/automation/org.eclipse.smarthome.automation.module.script.rulesupport/src/main/java/org/eclipse/smarthome/automation/module/script/rulesupport/internal/delegates/SimpleConditionHandlerDelegate.java
@@ -22,11 +22,11 @@ import org.eclipse.smarthome.automation.module.script.rulesupport.shared.simple.
  */
 public class SimpleConditionHandlerDelegate extends BaseModuleHandler<Condition> implements ConditionHandler {
 
-    private org.eclipse.smarthome.automation.module.script.rulesupport.shared.simple.SimpleConditionHandler conditionHandler;
+    private SimpleConditionHandler conditionHandler;
 
     public SimpleConditionHandlerDelegate(Condition condition, SimpleConditionHandler scriptedHandler) {
         super(condition);
-
+        this.conditionHandler = scriptedHandler;
         scriptedHandler.init(condition);
     }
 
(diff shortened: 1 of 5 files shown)


bundles/automation/org.eclipse.smarthome.automation.module.script.rulesupport/src/main/java/org/eclipse/smarthome/automation/module/script/rulesupport/internal/delegates/SimpleConditionHandlerDelegate.java
```diff
@@ -22,11 +22,11 @@
  */
 public class SimpleConditionHandlerDelegate extends BaseModuleHandler<Condition> implements ConditionHandler {
 
-    private org.eclipse.smarthome.automation.module.script.rulesupport.shared.simple.SimpleConditionHandler conditionHandler;
+    private SimpleConditionHandler conditionHandler;
 
     public SimpleConditionHandlerDelegate(Condition condition, SimpleConditionHandler scriptedHandler) {
         super(condition);
-
+        this.conditionHandler = scriptedHandler;
```
maggu2810: Using `final` on the declaration would help to identify such errors.

bundles/core/org.eclipse.smarthome.core.thing/src/main/java/org/eclipse/smarthome/core/thing/i18n/ThingStatusInfoI18nLocalizationService.java
```diff
@@ -87,7 +87,7 @@ protected void setTranslationProvider(TranslationProvider i18nProvider) {
     }
 
     protected void unsetTranslationProvider(TranslationProvider i18nProvider) {
-        i18nProvider = null;
+        this.i18nProvider = null;
```
maggu2810: Using `final` on the method argument would help to identify such errors.

sjka:
> Using final on the method argument would help to identify such errors.

let's split out this discussion -> #3651

maggu2810:
> let's split out this discussion -> #3651

Sure, I already approved it (regardless if `final` is used or not), only wait for Travis CI.
Thanks for opening a discussion.

bundles/core/org.eclipse.smarthome.core.thing/src/main/java/org/eclipse/smarthome/core/thing/i18n/ThingStatusInfoI18nLocalizationService.java
kaikreuzer: @maggu2810 Just by chance I noticed on that line that the search&replace from I18nProvider->TranslationProvider was only a partial success: It is a bit nasty that all variables are still called i18nProvider :-(

maggu2810: You are right, the variable names has been not replaced and so they are not really correct.
But this does not introduce that bug at all.

kaikreuzer: No, definitely unrelated to the bug, that's not what I wanted to convey!

PR #3659, opened 2017-06-14
avoid NPE if no service is found

kaikreuzer:


maggu2810:
Just because I am interested in: Do you reallize that NPE on a system running on OSGi 4.2?
I would expect that all versions greater then 4.3 would rely on an empty array.

[r4v42](https://osgi.org/javadoc/r4v42/org/osgi/framework/BundleContext.html#getServiceReferences(java.lang.String, java.lang.String))

```
An array of ServiceReference objects or null if no services are registered which satisfy the search.
```

[r4v43](https://osgi.org/javadoc/r4v43/core/org/osgi/framework/BundleContext.html#getServiceReferences(java.lang.Class, java.lang.String))

```
A collection of ServiceReference objects. May be empty if no services are registered which satisfy the search.
```

kaikreuzer:
@maggu2810 I came across this when implementing the Tradfr tests. It occurred while running the tests in the ESH IDE - and we are using there Equinox 3.10, which should be OSGi R5.

PR #3666, opened 2017-06-14
fixed a few issues in the archetypes

kaikreuzer:

### DIFF
--- a/tools/archetype/binding.test/src/main/resources/archetype-resources/src/test/java/__bindingIdCamelCase__OSGiTest.java
+++ b/tools/archetype/binding.test/src/main/resources/archetype-resources/src/test/java/__bindingIdCamelCase__OSGiTest.java
@@ -35,19 +35,26 @@ public class ${bindingIdCamelCase}OSGiTest extends JavaOSGiTest {
 
     private ManagedThingProvider managedThingProvider;
     private VolatileStorageService volatileStorageService = new VolatileStorageService();
-
+    private Bridge bridge;
+    
     @Before
     public void setUp() {
         registerService(volatileStorageService);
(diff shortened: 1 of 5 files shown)


martinvw:
Did you also test it? :-)

kaikreuzer:
Almost - I implemented those tests in the Tradfri binding, where they succeed.
And I had a hard time noticing the the old test never ever had a chance to succeed as we do not register handlers as OSGi services since quite a while now...

sjka:
The build failure is caused by 

> Connection to http://download.itemis.com refused

Ouch 😉

tools/archetype/binding.test/src/main/resources/archetype-resources/src/test/java/__bindingIdCamelCase__OSGiTest.java
```diff
@@ -35,19 +35,26 @@
...
     }
 
     @Test
     public void creationOf${bindingIdCamelCase}Handler() {
-        ${bindingIdCamelCase}Handler handler = getService(ThingHandler.class,${bindingIdCamelCase}Handler.class);
-
-        assertThat(handler, is(nullValue()));
+        assertThat(bridge.getHandler(), is(nullValue()));
+        managedThingProvider.add(bridge);
+        waitForAssert(() -> assertThat(bridge.getHandler(), notNullValue()));
```
sjka: you wrote `is(nullValue())` above, shouldn't it be `is(notNullValue())` here? Not sure if it actually  _is_ required, but imho it should be symmetrical at least.

kaikreuzer: ok, updated it accordingly.

PR #3667, opened 2017-06-14
configurable initial delay of ChannelItemProvider

sjka:
It's been a pain for tests (see https://hudson.eclipse.org/smarthome/job/SmartHomeDistribution-Nightly/1936), but I think it also cannot hurt to have it switched off configurable (on remains the default).

Alternatively we could add an `isInitialized()`method. I don't mind.

### DIFF
--- a/bundles/core/org.eclipse.smarthome.core.thing/src/main/java/org/eclipse/smarthome/core/thing/internal/ChannelItemProvider.java
+++ b/bundles/core/org.eclipse.smarthome.core.thing/src/main/java/org/eclipse/smarthome/core/thing/internal/ChannelItemProvider.java
@@ -146,9 +146,10 @@ public class ChannelItemProvider implements ItemProvider {
         }
 
         if (enabled) {
-            Executors.newSingleThreadExecutor().submit(new Runnable() {
-                @Override
-                public void run() {
+            boolean initialDelay = properties == null
+                    || !"false".equalsIgnoreCase((String) properties.get("initialDelay"));
+            if (initialDelay) {
+                Executors.newSingleThreadExecutor().submit(() -> {
                     // we wait until no further new links or items are announced in order to avoid creation of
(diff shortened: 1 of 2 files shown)


bundles/core/org.eclipse.smarthome.core.thing/src/main/java/org/eclipse/smarthome/core/thing/internal/ChannelItemProvider.java
```diff
@@ -175,6 +174,14 @@ public void run() {
         }
     }
 
+    private void initialize() {
+        logger.debug("Enabling channel item provider.");
+        initialized = true;
```
maggu2810: Isn't that the case we need to use volatile?

The variable initialized could be written by a single thread executor created and read by another thread.

bundles/core/org.eclipse.smarthome.core.thing/src/main/java/org/eclipse/smarthome/core/thing/internal/ChannelItemProvider.java
```diff
@@ -157,13 +158,11 @@ public void run() {
...
                     }
-                    logger.debug("Enabling channel item provider.");
-                    initialized = true;
-                    // simply call getAll() will create the items and notify all registered listeners automatically
-                    getAll();
-                    addRegistryChangeListeners();
-                }
-            });
+                    initialize();
+                });
```
maggu2810: Shouldn't the `shutdown` method be called, so queued task are executed and the scheduler could be cleaned up after all tasks has been executed?

maggu2810: https://github.com/eclipse/smarthome/pull/3691

PR #3668, opened 2017-06-14
Icon servlet: take care of the case + additional debug log

lolodomo:


lolodomo:
Please don't merge, I will change another little thing for a better handling of the case when the request has no state set.

lolodomo:
Change done. Ready for review.

PR #3669, opened 2017-06-14
Basic UI: improvements for non SSE browsers

lolodomo:


lolodomo:
@resetnow : please review my changes, there is a major fix for non SSE browsers when the page is updated. In addition to the item state, the value of an element is taken from the label, the same way as for SSE.

The update of the page title is only working partially for non SSE browsers because the update is not triggered when expected but only when an element of the page is updated. I will not try to fix that that it would require a change in the sitemap REST API.

lolodomo:
@resetnow: no time to review ?

resetnow:
Been a little busy, BRB

extensions/ui/org.eclipse.smarthome.ui.basic/web-src/smarthome.js
```diff
@@ -1755,15 +1774,32 @@
...
+						(label.indexOf("[") !== -1) &&
+						(label.indexOf("]") !== -1)
+					) {
+						var
+							pos = label.indexOf("[");
+
+						value = label.substr(
+							pos + 1,
+							label.lastIndexOf("]") - (pos + 1)
+						);
```
resetnow: Since value extraction code is now repeated 3 times (2 here and 1 in SSE listener) I think it makes sense to extract it in a function which would take formatted string `Temperature [23 °C]` and return either a value `23 °C` or a string itself. AbstractChangeListener is probably a good place to put it in.

lolodomo:
@resetnow : functions added.

PR #3670, opened 2017-06-15
Prevent automation resources UIDs duplication

adimova:
solves #3195

### DIFF
--- a/bundles/automation/org.eclipse.smarthome.automation.providers/META-INF/MANIFEST.MF
+++ b/bundles/automation/org.eclipse.smarthome.automation.providers/META-INF/MANIFEST.MF
@@ -12,6 +12,7 @@ Import-Package:
  org.eclipse.smarthome.config.core.i18n,
  org.eclipse.smarthome.core.common.registry,
  org.eclipse.smarthome.core.i18n,
+ org.eclipse.smarthome.core.storage,
  org.osgi.framework,
  org.osgi.service.packageadmin,
  org.osgi.util.tracker,
--- a/bundles/automation/org.eclipse.smarthome.automation.providers/META-INF/MANIFEST.MF
+++ b/bundles/automation/org.eclipse.smarthome.automation.providers/META-INF/MANIFEST.MF
@@ -12,7 +12,6 @@ Import-Package:
  org.eclipse.smarthome.config.core.i18n,
  org.eclipse.smarthome.core.common.registry,
  org.eclipse.smarthome.core.i18n,
- org.eclipse.smarthome.core.storage,
  org.osgi.framework,
  org.osgi.service.packageadmin,
  org.osgi.util.tracker,
--- a/bundles/automation/org.eclipse.smarthome.automation.providers/src/main/java/org/eclipse/smarthome/automation/internal/core/provider/AbstractResourceBundleProvider.java
+++ b/bundles/automation/org.eclipse.smarthome.automation.providers/src/main/java/org/eclipse/smarthome/automation/internal/core/provider/AbstractResourceBundleProvider.java
@@ -14,6 +14,7 @@ import java.io.InputStreamReader;
 import java.net.URI;
 import java.net.URISyntaxException;
 import java.net.URL;
+import java.nio.charset.StandardCharsets;
 import java.util.ArrayList;
 import java.util.Enumeration;
 import java.util.LinkedList;
@@ -459,7 +460,7 @@ public abstract class AbstractResourceBundleProvider<E> {
         InputStream is = null;
         try {
             is = url.openStream();
-            reader = new InputStreamReader(is);
+            reader = new InputStreamReader(is, StandardCharsets.UTF_8);
             return parser.parse(reader);
         } catch (ParsingException e) {
             logger.error(e.getLocalizedMessage(), e);
@@ -486,24 +487,21 @@ public abstract class AbstractResourceBundleProvider<E> {
     @SuppressWarnings("unchecked")
     protected void addNewProvidedObjects(List<String> newPortfolio, List<String> previousPortfolio,
             Set<E> parsedObjects) {
+        List<ProviderChangeListener<E>> snapshot = null;
+        synchronized (listeners) {
+            snapshot = new LinkedList<ProviderChangeListener<E>>(listeners);
+        }
         for (E parsedObject : parsedObjects) {
             String uid = getUID(parsedObject);
-            if (providedObjectsHolder.get(uid) == null) {
-                if (checkExistence(uid)) {
-                    continue;
-                }
-            } else if (previousPortfolio == null || !previousPortfolio.contains(uid)) {
-                logger.error("{} with UID \"{}\" already exists! Failed to create a second with the same UID!",
-                        parsedObject.getClass().getName(), uid, new IllegalArgumentException());
+            E oldelement = providedObjectsHolder.get(uid);
+            if (oldelement != null && !previousPortfolio.contains(uid)) {
+                logger.warn(
+                        String.format("%s with UID \"%s\" already exists! Failed to add a second with the same UID!",
+                                parsedObject.getClass().getName(), uid));
                 continue;
-            }
-            newPortfolio.add(uid);
-            E oldelement = providedObjectsHolder.put(uid, parsedObject);
-            if (listeners != null) {
-                List<ProviderChangeListener<E>> snapshot = null;
-                synchronized (listeners) {
-                    snapshot = new LinkedList<ProviderChangeListener<E>>(listeners);
-                }
+            } else {
+                newPortfolio.add(uid);
+                providedObjectsHolder.put(uid, parsedObject);
                 for (ProviderChangeListener<E> listener : snapshot) {
                     if (oldelement == null) {
                         listener.added((Provider<E>) this, parsedObject);
@@ -529,12 +527,6 @@ public abstract class AbstractResourceBundleProvider<E> {
             waitingProviders.remove(bundle);
         }
     }
-
-    /**
-     * @param uid
-     * @return
-     */
-    protected abstract boolean checkExistence(String uid);
 
     /**
      * @param parsedObject
--- a/bundles/automation/org.eclipse.smarthome.automation.providers/src/main/java/org/eclipse/smarthome/automation/internal/core/provider/AbstractResourceBundleProvider.java
+++ b/bundles/automation/org.eclipse.smarthome.automation.providers/src/main/java/org/eclipse/smarthome/automation/internal/core/provider/AbstractResourceBundleProvider.java
@@ -493,20 +493,19 @@ public abstract class AbstractResourceBundleProvider<E> {
         }
         for (E parsedObject : parsedObjects) {
             String uid = getUID(parsedObject);
-            E oldelement = providedObjectsHolder.get(uid);
-            if (oldelement != null && !previousPortfolio.contains(uid)) {
-                logger.warn(
-                        String.format("%s with UID \"%s\" already exists! Failed to add a second with the same UID!",
-                                parsedObject.getClass().getName(), uid));
+            E oldElement = providedObjectsHolder.get(uid);
+            if (oldElement != null && !previousPortfolio.contains(uid)) {
+                logger.warn("{} with UID '{}' already exists! Failed to add a second with the same UID!",
+                        parsedObject.getClass().getName(), uid);
                 continue;
             } else {
                 newPortfolio.add(uid);
                 providedObjectsHolder.put(uid, parsedObject);
                 for (ProviderChangeListener<E> listener : snapshot) {
-                    if (oldelement == null) {
+                    if (oldElement == null) {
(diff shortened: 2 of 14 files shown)


adimova:
> This is a lot of code change for a "simple prevention of duplicates". Is this really all merely related to that issue or should it be split into different PRs?

You're right, there are 2 separate optimizations in the commit, I'll post separate PRs for them.

> I am a bit concerned about throwing IllegalArgumentExceptions. From a provider's perspective, the arguments are not illegal as long as it itself does not have duplicates. It is imho therefore wrong to blame the provider implementation. Instead, I would expect a warning in the log that tells the system admin/user that something is screwed up in his installation that needs to be fixed. The framework should then simply ignore any additional entity with an already existing UID.

The IllegalArgumentExceptions is actually caught and logged in AbstractRegistry.addProvider & AbstractRegistry.added

> I'd also very much prefer to see the automation registries behave the very same way as the other ESH registries  in this regard, i.e. either all should be changed or none.

I'm not sure what should be done about it: do you expect me to make the same changes in all the other registries?

kaikreuzer:
> The IllegalArgumentExceptions is actually caught and logged in AbstractRegistry.addProvider & AbstractRegistry.added

Ok, you are right.

> do you expect me to make the same changes in all the other registries?

If we say that we want to avoid duplicates in the RuleRegistry, I think it indeed would be good to agree that any other Registry should have the same behavior - and I actually cannot think of a use case where we want to have duplicates in a registry. So rejecting elements with a duplicate id through a warning in the log should be fine. And with that it means you neither have to "change the other registries" nor do you have to implement anything special in the RuleRegistry - but simply directly adapt the code within `AbstractRegistry.addProvider` and `AbstractRegistry.added`. 
Does that make sense?

adimova:
ok

bundles/core/org.eclipse.smarthome.core/src/main/java/org/eclipse/smarthome/core/common/registry/AbstractRegistry.java
```diff
@@ -268,6 +279,15 @@ protected void addProvider(Provider<E> provider) {
...
     }
 
+    /**
+     * Returns the key for a given element
+     *
+     * @param element
+     *            element
+     * @return key (must not be null)
+     */
+    protected abstract K getKey(E element);
```
danchom: May be the getElementKey is better name, but I'm not sure if It isn't better to have RegistryElement interface which will unify getting the key and value of element

kaikreuzer: Hm, that's getting more complicated than I had hoped. Looping through all entries and figuring out the key for every element looks pretty inefficient to me.
The underlying problem is that our `Provider<E>` interface does not at all HAVE the notion of a key - this only exists for `ManagedProvider`.
So by definition, we do not have a key at hand for the provided entities (and we do not even know, if such a key exists). I think our best option is the equals method - if an entity has a key, this should be the one that is used in equals; for our `Rule` class, this is e.g. the case.
Wdyt?

adimova: The Registry interface has method get by the key - public E get(K key), so registries have the notion of a key, also the expectation is - this key to be unique

bundles/automation/org.eclipse.smarthome.automation.providers/src/main/java/org/eclipse/smarthome/automation/internal/core/provider/AbstractResourceBundleProvider.java
```diff
@@ -459,7 +459,7 @@ protected Bundle getBundle(String uid) {
         InputStream is = null;
         try {
             is = url.openStream();
-            reader = new InputStreamReader(is);
+            reader = new InputStreamReader(is, "UTF-8");
```
maggu2810:
```Java
reader = new InputStreamReader(is, StandardCharsets.UTF_8);
```

bundles/core/org.eclipse.smarthome.core/src/main/java/org/eclipse/smarthome/core/common/registry/AbstractRegistry.java
```diff
@@ -268,6 +279,15 @@ protected void addProvider(Provider<E> provider) {
         }
     }
 
+    /**
+     * Returns the key for a given element
+     *
+     * @param element
```
maggu2810: Can you remove the new line between the parameter name and the parameter documentation?

bundles/core/org.eclipse.smarthome.core/src/main/java/org/eclipse/smarthome/core/common/registry/AbstractRegistry.java
maggu2810: What about adding a `RegistryElement` interface as suggested by @danchom?

https://github.com/eclipse/smarthome/compare/master...maggu2810:registry-element?expand=1

maggu2810: Hm, it would be great if we could use `getUID()` for the key, because we would like a unique ID.
For the item itself it would be easy, the name will be the unique. All the other classes already provide a `getUID()` method.
BUT: The "unique" identifier for the link is provided by `getID()` and the stuff that is linked to an item will be returned by `getUID()`. This is a little bit confusing IMHO.

maggu2810: @kaikreuzer @sjka
WRT the comments above, I assume I found a "solution" for the element-key-topic that has very little influence about the existing API / code changes.
Please have a look at: maggu2810@9df148865fb46f3bf5cd6b399a1614a85705d706

If we really don't like an API breaking we could add a default implementation that returns the hashCode of the element, but as it is only about a lambda added to the constructor, I assume it is easy to adapt the proprietary registry classes.

adimova: @maggu2810 I like the way you think, but I have one remark - the name UniqueIdProvider is in conflict with Provider interface, could we change it to UniquelyIdentifiable or something like this?

maggu2810: > the name UniqueIdProvider is in conflict with Provider interface, could we change it to UniquelyIdentifiable or something like this?

Sure, I am happy to change the naming.
I will wait for that change as soon as I get a go. The change would  be then part of the PR if we agree.

sjka: > If we really don't like an API breaking

I'm not aware of anything outside our realm which would be broken by this change, so I'm okay with it.

kaikreuzer: I am not quite sure if I am overseeing everything yet, but wouldn't it also be an option to directly change the Provider to `Provider<K,E>` and thus not require the introduction of yet another interface?

adimova: If we parameterize the Provider, then we'll need to introduce a new method "K getKey(E element)" in the Provider interface and then the registries should ask the providers about the key for the element and the control over the key will become decentralized. Do we need such freedom? Also this will be larger change against this proposal.

kaikreuzer: Ah, sorry, I was completely mislead by the name %&$**Provider**, so just what @sjka already mentioned as the naming conflict.
Ok, then I agree that we should have a common interface on our entities. It seems that most people simply use an `Identifiable` interface, wouldn't that be fine here as well? Note that "UniquelyIdentifiable" might be rather misleading, because the ids are only unique within their own type, but not across all Identifiables!

adimova: I agree, let use an Identifiable interface

danchom: I like Identifiable name too.

maggu2810: Let's continue here: https://github.com/eclipse/smarthome/pull/3762

kaikreuzer: #3762 is merged, so @adimova, you can update this PR accordingly.

bundles/automation/org.eclipse.smarthome.automation.providers/META-INF/MANIFEST.MF
```diff
@@ -12,6 +12,7 @@ Import-Package:
  org.eclipse.smarthome.config.core.i18n,
  org.eclipse.smarthome.core.common.registry,
  org.eclipse.smarthome.core.i18n,
+ org.eclipse.smarthome.core.storage,
```
kaikreuzer: This isn't needed anymore, right?

bundles/automation/org.eclipse.smarthome.automation.providers/src/main/java/org/eclipse/smarthome/automation/internal/core/provider/AbstractResourceBundleProvider.java
```diff
@@ -486,24 +487,21 @@ protected Bundle getBundle(String uid) {
...
         for (E parsedObject : parsedObjects) {
             String uid = getUID(parsedObject);
-            if (providedObjectsHolder.get(uid) == null) {
-                if (checkExistence(uid)) {
-                    continue;
-                }
-            } else if (previousPortfolio == null || !previousPortfolio.contains(uid)) {
-                logger.error("{} with UID \"{}\" already exists! Failed to create a second with the same UID!",
-                        parsedObject.getClass().getName(), uid, new IllegalArgumentException());
+            E oldelement = providedObjectsHolder.get(uid);
```
kaikreuzer: Could you change the name to `oldElement`?

bundles/automation/org.eclipse.smarthome.automation.providers/src/main/java/org/eclipse/smarthome/automation/internal/core/provider/AbstractResourceBundleProvider.java
```diff
@@ -486,24 +487,21 @@ protected Bundle getBundle(String uid) {
...
+            if (oldelement != null && !previousPortfolio.contains(uid)) {
+                logger.warn(
```
kaikreuzer: Is there a special reason why you use `String.format` and not simply use parameterized logging?

bundles/core/org.eclipse.smarthome.core/src/main/java/org/eclipse/smarthome/core/common/registry/AbstractRegistry.java
```diff
@@ -118,6 +118,13 @@ public void added(Provider<E> provider, E element) {
         Collection<E> elements = elementMap.get(provider);
         if (elements != null) {
             try {
+                K uid = element.getUID();
+                if (uid != null && get(uid) != null) {
+                    logger.warn(String.format(
```
kaikreuzer: Is there a special reason why you use `String.format` and not simply use parameterized logging?

bundles/core/org.eclipse.smarthome.core/src/main/java/org/eclipse/smarthome/core/common/registry/AbstractRegistry.java
```diff
@@ -254,6 +261,13 @@ protected void addProvider(Provider<E> provider) {
             elementMap.put(provider, elements);
             for (E element : elementsOfProvider) {
                 try {
+                    K uid = element.getUID();
+                    if (uid != null && get(uid) != null) {
+                        logger.warn(String.format(
```
kaikreuzer: Is there a special reason why you use `String.format` and not simply use parameterized logging?

kaikreuzer:
@danchom I have addressed my review comments myself through [this commit](https://github.com/eclipse/smarthome/pull/3670/commits/9103afc2043fd62e8f320b971562bdff1755b863) - if they are ok to you, please feel free to merge!

Issue #3474, opened 2017-05-17
RuleStatus "Running" not sufficient


BenediktNiehues:
As user I would assume that if the Rule is in status "RUNNING", that currently the actions are performed. But it is also "RUNNING" if the conditions are evaluated. 
Maybe there should be StatusDetails like "EvaluatingConditions" and "ExecutingActions" accordingly.
Furthermore we have to think about the ruleStatus handling for long running (asynchronous) actions..

0nse:
> Furthermore we have to think about the ruleStatus handling for long running (asynchronous) actions..

To make this more clear: Assuming there were additional status `EVALUATING` and `EXECUTING`, `EXECUTING` would only be a briefly active status.

However, the actual, physical actions are likely to take longer until they have actually finished as they may depend on capabilities of some hardware. Thus, it would be great, if rules were `EXECUTING` for until the actual actions have been completed as opposed to the ESH actions having been fired.

danchom:
If we add more states and corresponding events for them this will slowdown the RuleEngine. Actually the running state is an internal state which does not permit to run a rule before ending of previous execution. Even I think we should not fire an event for this state because I don't see why it is interesting for the user. The user has to know if his rule is idle (resoled and ready for run)  or it is uninitialized (base on some problems defined in status details) and it can't run. Why the user has to be notified for each run? Yes this is useful for logging debug messages and check if the rules work as expected. 
>Thus, it would be great, if rules were EXECUTING for until the actual actions have been completed 

I think it is not good idea to block a rule until actual action is executing. What is going on with a new events coming from triggers during this period of time? Yes they are collected in a queue, which must not grow forever. Also some of queue events may want to cancel or change current operation immediately. For me the rule has to responds to the trigger events as quickly as possible and it must be  a realtime system.
Of course you can implement an action  handler which finish its execution when the actual action is executed.

0nse:
> we should not fire an event for this state because I don't see why it is interesting for the user

In the past, we witnessed several users complaining about not being able to see when rules execute. Of course, the tech-savvy ESH user is most likely aware of their rules and their effects. However, there are also users that do not have them all in mind and wonder about how the state of some devices has been achieved. That is why we were asked whether a user can inspect when which rules run/ran.

So, if ESH wants to target the casual user, I believe that this information can be of value for some.

> I think it is not good idea to block a rule until actual action is executing

I see your points and for now would agree.

danchom:
[...]

PR #3671, opened 2017-06-16
Implements #3474. Add triggered state when a rule is triggered.

danchom:
The running state appears only when actions are executed.

### DIFF
--- a/bundles/automation/org.eclipse.smarthome.automation.api/src/main/java/org/eclipse/smarthome/automation/RuleStatus.java
+++ b/bundles/automation/org.eclipse.smarthome.automation.api/src/main/java/org/eclipse/smarthome/automation/RuleStatus.java
@@ -1,5 +1,5 @@
 /**
- * Copyright (c) 1997, 2015 by ProSyst Software GmbH and others.
+ * Copyright (c) 2015, 2017 by Bosch Software Innovations GmbH and others.
  * All rights reserved. This program and the accompanying materials
  * are made available under the terms of the Eclipse Public License v1.0
  * which accompanies this distribution, and is available at
@@ -18,7 +18,8 @@ public enum RuleStatus {
     UNINITIALIZED(1),
     INITIALIZING(2),
     IDLE(3),
-    RUNNING(4);
+    TRIGGERED(4),
+    RUNNING(5);
 
     private final int value;
 
--- a/bundles/automation/org.eclipse.smarthome.automation.api/src/main/java/org/eclipse/smarthome/automation/RuleStatusInfo.java
+++ b/bundles/automation/org.eclipse.smarthome.automation.api/src/main/java/org/eclipse/smarthome/automation/RuleStatusInfo.java
@@ -1,11 +1,14 @@
 /**
- * Copyright (c) 1997, 2015 by ProSyst Software GmbH and others.
+ * Copyright (c) 2015, 2017 by Bosch Software Innovations GmbH and others.
  * All rights reserved. This program and the accompanying materials
  * are made available under the terms of the Eclipse Public License v1.0
  * which accompanies this distribution, and is available at
  * http://www.eclipse.org/legal/epl-v10.html
  */
 package org.eclipse.smarthome.automation;
+
+import java.util.Collections;
+import java.util.Map;
 
 /**
  * This class is used to present status of rule. The status consists out of three parts:
@@ -19,6 +22,18 @@ public class RuleStatusInfo {
     private RuleStatus status;
     private RuleStatusDetail statusDetail;
     private String description;
+    private Map<String, ?> properties;
+    private static final RuleStatusInfo DISABLED = new RuleStatusInfo(RuleStatus.DISABLED, RuleStatusDetail.NONE, null,
+            null);
+    private static final RuleStatusInfo UNINITIALIZED = new RuleStatusInfo(RuleStatus.UNINITIALIZED,
+            RuleStatusDetail.NONE, null, null);
+    private static final RuleStatusInfo INITIALIZING = new RuleStatusInfo(RuleStatus.INITIALIZING,
+            RuleStatusDetail.NONE, null, null);
+    private static final RuleStatusInfo IDLE = new RuleStatusInfo(RuleStatus.IDLE, RuleStatusDetail.NONE, null, null);
+    private static final RuleStatusInfo TRIGGERED = new RuleStatusInfo(RuleStatus.TRIGGERED, RuleStatusDetail.NONE,
+            null, null);
+    private static final RuleStatusInfo RUNNING = new RuleStatusInfo(RuleStatus.RUNNING, RuleStatusDetail.NONE, null,
+            null);
 
     /**
      * Default constructor for deserialization e.g. by Gson.
@@ -33,8 +48,27 @@ public class RuleStatusInfo {
      *
      * @throws IllegalArgumentException if status is null
      */
-    public RuleStatusInfo(RuleStatus status) throws IllegalArgumentException {
-        this(status, RuleStatusDetail.NONE);
+    public static RuleStatusInfo create(RuleStatus status) {
+        if (status == null) {
+            throw new IllegalArgumentException("Status must not be null.");
+        }
+        switch (status) {
+            case DISABLED:
+                return RuleStatusInfo.DISABLED;
+            case UNINITIALIZED:
+                return RuleStatusInfo.UNINITIALIZED;
+            case INITIALIZING:
+                return RuleStatusInfo.INITIALIZING;
+            case IDLE:
+                return RuleStatusInfo.IDLE;
+            case TRIGGERED:
+                return RuleStatusInfo.TRIGGERED;
+            case RUNNING:
+                return RuleStatusInfo.RUNNING;
+            default:
+                throw new IllegalArgumentException("Invalid status value.");
+        }
+
     }
 
     /**
@@ -45,8 +79,32 @@ public class RuleStatusInfo {
      *
      * @throws IllegalArgumentException if status or status detail is null
      */
-    public RuleStatusInfo(RuleStatus status, RuleStatusDetail statusDetail) throws IllegalArgumentException {
-        this(status, statusDetail, null);
+    public static RuleStatusInfo create(RuleStatus status, RuleStatusDetail statusDetail)
+            throws IllegalArgumentException {
+        if (statusDetail == null || statusDetail == statusDetail.NONE) {
+            return create(status);
+        } else {
+            return create(status, statusDetail, null);
+        }
+    }
+
+    public static RuleStatusInfo create(RuleStatus status, RuleStatusDetail statusDetail, String description) {
+        if (status == null) {
+            throw new IllegalArgumentException("Status must not be null.");
+        }
+        if (description == null && (statusDetail == null || statusDetail == statusDetail.NONE)) {
+            return create(status);
+        } else {
+            return new RuleStatusInfo(status, statusDetail, description, null);
+        }
+    }
+
+    public static RuleStatusInfo createTriggeredStatusInfo(Map<String, ?> properties) {
+        if (properties == null || properties.isEmpty()) {
+            return RuleStatusInfo.TRIGGERED;
+        } else {
+            return new RuleStatusInfo(RuleStatus.TRIGGERED, RuleStatusDetail.NONE, null, properties);
+        }
     }
 
     /**
@@ -55,11 +113,12 @@ public class RuleStatusInfo {
      * @param status the status (must not be null)
      * @param statusDetail the detail of the status (must not be null)
      * @param description the description of the status
+     * @param properties additional properties related to this status.
      *
      * @throws IllegalArgumentException if status or status detail is null
      */
-    public RuleStatusInfo(RuleStatus status, RuleStatusDetail statusDetail, String description)
-            throws IllegalArgumentException {
+    private RuleStatusInfo(RuleStatus status, RuleStatusDetail statusDetail, String description,
+            Map<String, ?> properties) throws IllegalArgumentException {
         if (status == null) {
             throw new IllegalArgumentException("Thing status must not be null");
         }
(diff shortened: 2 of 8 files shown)


danchom:
@kaikreuzer could you check this PR please

bundles/automation/org.eclipse.smarthome.automation.module.timer.test/src/test/groovy/org/eclipse/smarthome/automation/module/timer/internal/RuntimeRuleTest.groovy
```diff
@@ -133,7 +133,8 @@ class RuntimeRuleTest extends OSGiTest{
          * Check if auto update is working and
          * ensure that the lamp item state if OFF after this check
          */
-        logger.info("Check auto update");
+        logger.info("Check auto update")
```
sjka: this line break slipped in

bundles/automation/org.eclipse.smarthome.automation.api/src/main/java/org/eclipse/smarthome/automation/RuleStatusInfo.java
```diff
@@ -69,6 +74,7 @@ public RuleStatusInfo(RuleStatus status, RuleStatusDetail statusDetail, String d
         this.status = status;
         this.statusDetail = statusDetail;
         this.description = description;
+        this.properties = properties;
```
sjka: how do these properties relate to the triggered state?

danchom: The RuleStatusInfo of triggered state should contain output values of the trigger which was activated the rule. These values can be used by a CEP trigger based on the existing trigger type, also  they can be used for logging or remote tracking of automation engine.

danchom:
Test cases were added.

bundles/automation/org.eclipse.smarthome.automation.module.core.test/src/test/java/org/eclipse/smarthome/automation/module/core/internal/TriggeredRunningStateTest.java
```diff
@@ -0,0 +1,281 @@
...
+                        break;
+                    case IDLE:
+                        ruleStatusInfoEventFound = triggeredEventFound;
+                        break;
+                }
+            }
+        } while (!ruleStatusInfoEventFound && wait > 0);
+        Assert.assertTrue(ruleStatusInfoEventFound);
+        Assert.assertNotNull(myLampItemAction);
+        Assert.assertEquals(OnOffType.OFF, myLampItemAction.getState());
```
triller-telekom: Shouldn't the action lamp be on once the rule has fired?

bundles/automation/org.eclipse.smarthome.automation.api/src/main/java/org/eclipse/smarthome/automation/RuleStatus.java
```diff
@@ -18,7 +18,8 @@
     UNINITIALIZED(1),
     INITIALIZING(2),
     IDLE(3),
-    RUNNING(4);
+    RUNNING(4),
+    TRIGGERED(5);
```
triller-telekom: How about putting "TRIGGERED" before "RUNNING"? Because that is also how the lifecycle of a rule will be

bundles/automation/org.eclipse.smarthome.automation.module.core.test/src/test/java/org/eclipse/smarthome/automation/module/core/internal/TriggeredRunningStateTest.java
danchom: Actually no, because the condition is not satisfied (myLampCondition is OFF) and the rule is not executed.  Therefor the action myLampItemAction is not executed and the state is not changed.

danchom:
The PR is updated, it has optimization in rule engine synchronizations also the factory of RuleStatusInfo is added.

danchom:
ping

bundles/automation/org.eclipse.smarthome.automation.module.core.test/src/test/java/org/eclipse/smarthome/automation/module/core/internal/TriggeredRunningStateTest.java
triller-telekom: Ah I see, sorry I missed that condition...

bundles/automation/org.eclipse.smarthome.automation.api/src/main/java/org/eclipse/smarthome/automation/RuleStatusInfo.java
```diff
@@ -45,20 +75,44 @@ public RuleStatusInfo(RuleStatus status) throws IllegalArgumentException {
...
+    public static RuleStatusInfo create(RuleStatus status, RuleStatusDetail statusDetail) throws IllegalArgumentException {
+        if(statusDetail == null || statusDetail == statusDetail.NONE) {
+            return create(status);
+        }else {
+            return create(status, statusDetail, null);
+        }    
+    }
+
+    public static RuleStatusInfo create(RuleStatus status, RuleStatusDetail statusDetail, String description) {
+        if(status == null) {
```
kaikreuzer: @triller-telekom Instead of having all those IAEs, wouldn't that be a perfect case to make use of the new Nonnull annotations?

bundles/automation/org.eclipse.smarthome.automation.core/src/main/java/org/eclipse/smarthome/automation/core/internal/RuleEngine.java
```diff
@@ -116,40 +120,35 @@
...
      * not. The relation is rule's id to {@link Rule} object.
      */
-    private Map<String, RuntimeRule> rules;
+    private Map<String, RuntimeRule> rules = new ConcurrentHashMap<String, RuntimeRule>();
 
     /**
      * {@link Map} system module type to corresponding module handler factories.
      */
-    private Map<String, ModuleHandlerFactory> moduleHandlerFactories;
+    private Map<String, ModuleHandlerFactory> moduleHandlerFactories = new ConcurrentHashMap<>(31);
```
kaikreuzer: Any specific reason for 31?

bundles/automation/org.eclipse.smarthome.automation.api/src/main/java/org/eclipse/smarthome/automation/RuleStatusInfo.java
kaikreuzer: I will need a deeper look into this, but I am not convinced that such properties should be introduced here (as it brings it out of sync with the Thing status concept and also most likely breaks all our evening mechanism...).

bundles/automation/org.eclipse.smarthome.automation.core/src/main/java/org/eclipse/smarthome/automation/core/internal/RuleEngine.java
```diff
@@ -1,4 +1,8 @@
 /**
```
kaikreuzer: There are MANY changes in this class - is this really all related to this feature or does it include some further refactorings? In the latter case, I would prefer to have it split into separate PRs, which would make reviews much easier.

bundles/automation/org.eclipse.smarthome.automation.core/src/main/java/org/eclipse/smarthome/automation/core/internal/RuntimeRule.java
```diff
@@ -28,6 +33,7 @@
 public class RuntimeRule extends Rule {
 
     private Map<String, Module> moduleMap;
+    private RuleStatusInfo statusInfo;
```
kaikreuzer: This looks fishy. Afair, we so far kept the status of the rules in the rule engine, having it mixed up with the rule meta-data might not be the best idea...

bundles/automation/org.eclipse.smarthome.automation.core/src/main/java/org/eclipse/smarthome/automation/core/internal/RuleEngine.java
danchom: I've optimize synchronizations. Also I have remove status map from rule engine and put it into runtime rule. I don't see why we have to hold a rules' status into separate map and do unduly synchronization between runtime rules and status map. Each  runtime rule must have a status and I think it must be a field of it.

bundles/automation/org.eclipse.smarthome.automation.core/src/main/java/org/eclipse/smarthome/automation/core/internal/RuleEngine.java
danchom: No, I'll remove it.

bundles/automation/org.eclipse.smarthome.automation.core/src/main/java/org/eclipse/smarthome/automation/core/internal/RuntimeRule.java
danchom: Actually each runtime rule must always have a status. I think it should be part of it. In these case there is no need to synchronize maps of runtime rules and their statuses.

bundles/automation/org.eclipse.smarthome.automation.api/src/main/java/org/eclipse/smarthome/automation/RuleStatusInfo.java
```diff
@@ -19,6 +26,18 @@
     private RuleStatus status;
     private RuleStatusDetail statusDetail;
     private String description;
+    private Map<String, ?> properties;
+
```
danchom: I'll

bundles/automation/org.eclipse.smarthome.automation.api/src/main/java/org/eclipse/smarthome/automation/RuleStatus.java
```diff
@@ -1,4 +1,8 @@
 /**
+ * Copyright (c) 2017 by Bosch Software Innovations GmbH. All rights reserved,
```
kaikreuzer: Well, yes, it is against the agreed rules for the license headers in this project (or rather this component).
If I may repeat them: We use
```
 * Copyright (c) {DATE} {INITIAL COPYRIGHT OWNER} and others.
 * All rights reserved. This program and the accompanying materials
 * are made available under the terms of the Eclipse Public License v1.0
 * which accompanies this distribution, and is available at
 * http://www.eclipse.org/legal/epl-v10.html
```
where `{INITIAL COPYRIGHT OWNER}` is the party that created the file in the first place (in this case ProSyst) and "and others" includes every one else that has done subsequent changes to it (which imho includes successor companies such as Bosch SI).

We definitely must not have two different license headers on a file.
In any case, please remember that we plan to move to a completely [new style of license headers](https://bugs.eclipse.org/bugs/show_bug.cgi?id=514948) in future.

danchom:
We have a response form our legacy department and it says to change the Prosyst name with the Bosch SI one in the license header when the initial creator is Prosyst and Bosch have modified file after that. We also have to add the year when changing license header.
For example:

/**
 * Copyright (c) 2015 by ProSyst Software GmbH and others.
 * All rights reserved. This program and the accompanying materials
 * are made available under the terms of the Eclipse Public License v1.0
 * which accompanies this distribution, and is available at
 * http://www.eclipse.org/legal/epl-v10.html
 */

should become to:
/**
 * Copyright (c) 2015, 2017 by Bosch Software Innovations and others.
 * All rights reserved. This program and the accompanying materials
 * are made available under the terms of the Eclipse Public License v1.0
 * which accompanies this distribution, and is available at
 * http://www.eclipse.org/legal/epl-v10.html
 */

@kaikreuzer is this solution appropriate for you?

kaikreuzer:
Yes, thanks, that's absolutely fine for me.

danchom:
@kaikreuzer Could you please check this PR again.

danchom:
ping

bundles/automation/org.eclipse.smarthome.automation.api/src/main/java/org/eclipse/smarthome/automation/RuleStatusInfo.java
kaikreuzer: No, imho that is not "natural" at all, it is rather the "easy" way... I really think that this is no information that is in any way related to the "RuleStatus". It is some information that you want to send out as an event for specific use cases. Couldn't the eventing be done by a special type of CEPTrigger? Its handler implementation could use the EventAdmin and post there whatever it likes to - without having any negative impact on the general infrastructure.

bundles/automation/org.eclipse.smarthome.automation.core/src/main/java/org/eclipse/smarthome/automation/core/internal/RuntimeRule.java
kaikreuzer: Ok, if we consider the RuntimeRules an internal data structure for the rule engine, it should be fine.

bundles/automation/org.eclipse.smarthome.automation.api/src/main/java/org/eclipse/smarthome/automation/RuleStatusInfo.java
kaikreuzer: @adimova Could you follow-up on this? Is it ok for you to move the specific information to the trigger implementation and revert the addition of properties to the RuleStatusInfo from this PR?

adimova: Yes, that was my intention, to prepare new PR with "Add triggered state when a rule is triggered", 'cause this PR has merge conflicts, also we still should discuss CEP concept to get clean. So this PR can be closed.

kaikreuzer:
I guess the issue is that this PR had been created by @danchom and @adimova cannot easily take it over. So ok for me to close and have a new one created by @adimova.

Issue #3673, opened 2017-06-16
[HttpUtil] guessContentType for downloadData not sufficient for JPEG


andreasreinhardt:
`org.eclipse.smarthome.io.net.http.HttpUtil.downloadData(String, String, boolean, long)` uses `java.net.URLConnection.guessContentTypeFromStream(InputStream)` tries to determine the actual MIME type of content if no header was sent from the server. This implementation is not sufficient for JPEG (see [Wikipedia](https://en.wikipedia.org/wiki/Magic_number_%28programming%29#Magic_numbers_in_files), enumeration item JPEG), as it compares the first 3 bytes instead of only the first 2 bytes. In the special use case of this method, it is also possible to look into the first and the last 2 bytes as the data is already downloaded and available in memory.
This will fix [openhab2-addons/issues/2302](https://github.com/openhab/openhab2-addons/issues/2302)

kaikreuzer:
Could you come up with a PR to fix this?

andreasreinhardt:
I'm not sure, whether I am allowed to contribute.

andreasreinhardt:
I'm currently preparing a pull request, give me some more minutes.

triller-telekom:
Could you please create a PR with your fix?

kaikreuzer:
@triller-telekom Already there since 5 days and under discussion: https://github.com/eclipse/smarthome/pull/3674

triller-telekom:
THX, I expected a direct link to the PR, which you just created now :)

lolodomo:
This issue can be closed now.

PR #3674, opened 2017-06-16
Fix MIME type detection of JPEG files (#3673)

andreasreinhardt:
Signed-off-by: Andreas Reinhardt <andreas.reinhardt@novatec-gmbh.de>

### DIFF
--- a/bundles/io/org.eclipse.smarthome.io.net/src/main/java/org/eclipse/smarthome/io/net/http/HttpUtil.java
+++ b/bundles/io/org.eclipse.smarthome.io.net/src/main/java/org/eclipse/smarthome/io/net/http/HttpUtil.java
@@ -470,6 +470,12 @@ public class HttpUtil {
      */
     public static String guessContentTypeFromData(byte[] data) {
         String contentType = null;
+
+        // URLConnection.guessContentTypeFromStream(input) is not sufficient to detect all JPEG files
+        if (isJpeg(data)) {
+            return "image/jpeg";
+        }
+
         try (final ByteArrayInputStream input = new ByteArrayInputStream(data)) {
             try {
                 contentType = URLConnection.guessContentTypeFromStream(input);
@@ -483,6 +489,19 @@ public class HttpUtil {
             // Error on closing input stream -- nothing we can do here.
         }
         return contentType;
+    }
+
+    /**
+     * Check whether the content data is a JPEG file checking file start and end bytes.
+     * {@link URLConnection#guessContentTypeFromStream(InputStream)} is wrong for some JPEG files.
+     *
+     * @see https://en.wikipedia.org/wiki/JPEG#Syntax_and_structure
+     * @param data the data as buffer of bytes
+     * @return <code>true</code> if the content is a JPEG file, <code>false</code> otherwise
+     */
+    private static boolean isJpeg(byte[] data) {
+        return (data[0] == (byte) 0xFF && data[1] == (byte) 0xD8 && data[data.length - 2] == (byte) 0xFF
+                && data[data.length - 1] == (byte) 0xD9);
     }
 
 }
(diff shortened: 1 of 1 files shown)


bundles/io/org.eclipse.smarthome.io.net/src/main/java/org/eclipse/smarthome/io/net/http/HttpUtil.java
```diff
@@ -485,4 +491,17 @@ public static String guessContentTypeFromData(byte[] data) {
...
+    /**
+     * Check whether the content data is a JPEG file checking file start and end bytes.
+     * {@link URLConnection#guessContentTypeFromStream(InputStream)} is wrong for some JPEG files.
+     *
+     * @see https://en.wikipedia.org/wiki/JPEG#Syntax_and_structure
+     * @param data the data as buffer of bytes
+     * @return <code>true</code> if the content is a JPEG file, <code>false</code> otherwise
+     */
+    private static boolean isJpeg(byte[] data) {
+        return (data[0] == (byte) 0xFF && data[1] == (byte) 0xD8 && data[data.length - 2] == (byte) 0xFF
```
lolodomo: Please check that the data buffer has at least 2 bytes to avoid a NPE.

bundles/io/org.eclipse.smarthome.io.net/src/main/java/org/eclipse/smarthome/io/net/http/HttpUtil.java
```diff
@@ -485,4 +491,17 @@ public static String guessContentTypeFromData(byte[] data) {
         return contentType;
     }
 
+    /**
+     * Check whether the content data is a JPEG file checking file start and end bytes.
+     * {@link URLConnection#guessContentTypeFromStream(InputStream)} is wrong for some JPEG files.
+     *
+     * @see https://en.wikipedia.org/wiki/JPEG#Syntax_and_structure
```
sjka: To me, this reads like it would be safe to additionally check for `data[2] == (byte) 0xFF`. I currently can't find a counter example what it would be absolutely required, but just to be on the safe side...

kaikreuzer:
What I am wondering is that the code in URLConnection is actually [more complex for jpeg detection](http://hg.openjdk.java.net/jdk8u/jdk8u-dev/jdk/file/tip/src/share/classes/java/net/URLConnection.java#l1517) than yours.
And I cannot find any filed bug in the OpenJDK issue tracker that anybody ever claimed that their logic is wrong. Can you comment on these two concerns?

lolodomo:
On my side i can confirm that JPEG files from Spotify returned by Sonos are correctly identified by the guess method.

andreasreinhardt:
@kaikreuzer I think, the OpenJDK code is more complex because it does not want to do a wrong guess so it covers the most common cases to have a pattern that is longer than just 2 bytes.
I had a deeper look into the JPEG files that Kodi delivers, and the starting sequence is `FF D8 FF FE`, so the picture starts with a text comment, containing `Lavc57.64.101` and some non-readable chars. This might be a hint that libav is used to convert the picture internally in Kodi, because the original image, that Kodi caches, does not have this comment and instead starts with `FF D8 FF E1` and contains EXIF data.
I don't know why there is no filed bug in OpenJDK yet, perhaps this case is not very common.

sjka:
> What I am wondering is that the code in URLConnection is actually more complex for jpeg detection than yours.

Having had a deeper look into the implementations, I think the major difference comes from the fact that the JRE constraints itself to the first few bytes while we here first get the whole file so you can have a look at the start _and_ end. 

I have no idea why we are doing it - as long as whatever gets returned from the server is small, this should not matter much. If this is ever going to change, we might run unto troubles in small devices and would need to change the implementation, e.g. to using a [PushbackInputStream](https://docs.oracle.com/javase/8/docs/api/java/io/PushbackInputStream.html) to buffer only the first few bytes. 

So to me this is mainly is about the decision whether to use the full file for a safe guessing or living with some fuzziness, considering only the first three bytes and hoping for the best. I'd actually be okay with the latter in our scope here.

sjka:
The discussion here ceased. Therefore I'd say this PR is good to go. To my understanding the strategy introduced here is fully correct and actually more robust than what the JDK does because we (currently) are in the luxurious position of having the fully image at hand. 

I suggest to return to this if at any time we need to reduce the detection scope to the first few bytes only.

PR #3675, opened 2017-06-16
[tradfri] restart transport layer on error

### LINKED ISSUE #3672: [Trådfri] Gateway does not come back online after being unplugged
Steps to reproduce:
* Have a bridge with bulbs configured so they are ONLINE
* Unplug the power cable of the gateway
* Wait until the thing are OFFLINE
* Replug the power cable

I'd expect that the schedule job (every minute) would make it get back online. However, this job continues to run into CoAP errors because of response timeouts. I don't get why though. 

Maybe it would be easiest to wireshark it in order to find out whether it's the bridge not responding, or californium choking on anything.

sjka:
Restarting the transport layer causes all CoAP connections to create a new TCP connection with a new SSL handshake. While this for sure is required if the Gateway has been restarted, I'm not really sure whether it is the right way of doing it when using the Californium APIs. I would have expected californium to take care of such aspects and also doing it only when necessary. This is brute-force right now: Whenever there is an error, renew all sockets. 

So big disclaimer: I have no clue what I'm doing here, please help if you have a better idea!

fixes #3672


kaikreuzer:
Good point, I didn't put much consideration in the the dtls connector so far.
Let's go with that fix for the time being until someone else knows it any better :-)

kartben:
Just a thought but Californium has a fairly active mailing list, with lots of knowledgeable people there ;-)
https://dev.eclipse.org/mailman/listinfo/cf-dev

Issue #3678, opened 2017-06-16
Bug in parameter value normalization


kaikreuzer:
Just found that in my log using the most recent code:

```
17:27:57.440 [WARN ] [.smarthome.config.core.Configuration] - Could not set field value for field 'id': Can not set java.lang.Integer field org.eclipse.smarthome.binding.tradfri.DeviceConfig.id to java.lang.String
java.lang.IllegalArgumentException: Can not set java.lang.Integer field org.eclipse.smarthome.binding.tradfri.DeviceConfig.id to java.lang.String
	at sun.reflect.UnsafeFieldAccessorImpl.throwSetIllegalArgumentException(UnsafeFieldAccessorImpl.java:167)[:1.8.0_112]
	at sun.reflect.UnsafeFieldAccessorImpl.throwSetIllegalArgumentException(UnsafeFieldAccessorImpl.java:171)[:1.8.0_112]
	at sun.reflect.UnsafeObjectFieldAccessorImpl.set(UnsafeObjectFieldAccessorImpl.java:81)[:1.8.0_112]
	at java.lang.reflect.Field.set(Field.java:764)[:1.8.0_112]
	at org.apache.commons.lang.reflect.FieldUtils.writeField(FieldUtils.java:523)[31:org.apache.commons.lang:2.6.0]
	at org.apache.commons.lang.reflect.FieldUtils.writeField(FieldUtils.java:500)[31:org.apache.commons.lang:2.6.0]
	at org.apache.commons.lang.reflect.FieldUtils.writeField(FieldUtils.java:560)[31:org.apache.commons.lang:2.6.0]
	at org.eclipse.smarthome.config.core.Configuration.as(Configuration.java:96)[95:org.eclipse.smarthome.config.core:0.9.0.201706161358]
	at org.eclipse.smarthome.core.thing.binding.BaseThingHandler.getConfigAs(BaseThingHandler.java:248)[106:org.eclipse.smarthome.core.thing:0.9.0.201706161358]
	at org.eclipse.smarthome.binding.tradfri.handler.TradfriLightHandler.initialize(TradfriLightHandler.java:69)[184:org.eclipse.smarthome.binding.tradfri:0.9.0.201706161358]
	at org.eclipse.smarthome.core.thing.internal.ThingManager$9.call(ThingManager.java:741)[106:org.eclipse.smarthome.core.thing:0.9.0.201706161358]
[...]

PR #3679, opened 2017-06-16
fixed the type of the id parameter for Tradfri devices

kaikreuzer:
fixes https://github.com/eclipse/smarthome/issues/3678


kaikreuzer:
Good catch - updated!

Issue #3680, opened 2017-06-16
io.rest uses gson 2.5.0, although we so far only support gson 2.3.1


kaikreuzer:
I found this in my logs:

```
Exception in thread "Thread-65" java.lang.IllegalAccessError: tried to access method com.google.gson.Gson.newJsonWriter(Ljava/io/Writer;)Lcom/google/gson/stream/JsonWriter; from class org.eclipse.smarthome.io.rest.JSONResponse
        at org.eclipse.smarthome.io.rest.JSONResponse.lambda$0(JSONResponse.java:162)
        at java.lang.Thread.run(Thread.java:745)
```
A closer analysis shows that this method is only publicly available in gson 2.5, while my setup only has gson 2.3.1 available.

Interestingly, our smarthome.target resolves 2.2.4 AND 2.5.0, so the IDE picks up 2.5.0 and compiles against it.

As far as I remember, we wanted to be compatible with gson 2.3.1, but it all seems to be a bit messed up right now...

sjka:
it's probably the jax-rs adapter which brings the newest gson version. Looking into it, it's so dead-simple that we could bring our own instead easily and remove it from our target scope.

PR #3682, opened 2017-06-16
implement own Jersey GSON provider

sjka:
...in order to cleanup the GSON library version mess.

It's effectively 3 lines of trivial code that it takes.

related to #3680

### DIFF
--- a/bundles/io/org.eclipse.smarthome.io.rest.core/src/main/java/org/eclipse/smarthome/io/rest/core/internal/GsonProvider.java
+++ b/bundles/io/org.eclipse.smarthome.io.rest.core/src/main/java/org/eclipse/smarthome/io/rest/core/internal/GsonProvider.java
@@ -1,0 +1,81 @@
+/**
+ * Copyright (c) 2014-2017 by the respective copyright holders.
+ * All rights reserved. This program and the accompanying materials
+ * are made available under the terms of the Eclipse Public License v1.0
+ * which accompanies this distribution, and is available at
+ * http://www.eclipse.org/legal/epl-v10.html
+ */
+package org.eclipse.smarthome.io.rest.core.internal;
+
+import java.io.IOException;
+import java.io.InputStream;
+import java.io.InputStreamReader;
+import java.io.OutputStream;
+import java.lang.annotation.Annotation;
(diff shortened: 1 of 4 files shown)


bundles/io/org.eclipse.smarthome.io.rest.core/src/main/java/org/eclipse/smarthome/io/rest/core/internal/GsonProvider.java
```diff
@@ -0,0 +1,81 @@
...
+        }
+    }
+
+    @Override
+    public boolean isReadable(Class<?> type, Type genericType, Annotation[] annotations, MediaType mediaType) {
+        return true;
+    }
+
+    @Override
+    public T readFrom(Class<T> type, Type gnericType, Annotation[] annotations, MediaType mediaType,
```
kaikreuzer: typo: gnericType

maggu2810:
Doesn't this involve a change in the Karaf TP?
Now we are using the old and the new one, aren't we?
https://github.com/eclipse/smarthome/blob/f4f7001/features/karaf/esh-tp/src/main/feature/feature.xml#L131

kaikreuzer:
Good question - both might be there now.
Can we simply drop the esh-tp-jax-rs-provider-gson feature or is there any reason why we have to keep it?

sjka:
On my side it was just by mistake that I kept it. I don't know any good reasons.

sjka:
PS: in case anybody has some strong feelings about it, feel free to express them in #3696

kaikreuzer:
I am not sure if it might break https://github.com/openhab/openhab2-addons/pull/2009#discussion_r115700463, but if so, I will fix it on the openHAB side.

sjka:
> I am not sure if it might break openhab/openhab2-addons#2009 (comment), but if so, I will fix it on the openHAB side.

...and I don't get how it is working there. I can't find where it is actually used.

kaikreuzer:
It is used in the included jar...

sjka:
...by just hoping that `com.eclipsesource.jaxrs.provider.gson.GsonProvider` is on the classpath? Yes, then it is broken.

PR #3691, opened 2017-06-17
properly shutdown scheduler

maggu2810:


maggu2810:
But I wonder, why we need to use an executor service that starts a thread at all.
Why not using a thread ourselves?

```Java
                final Thread thread = new Thread(() -> {
                    // we wait until no further new links or items are announced in order to avoid creation of
                    // items which then must be removed again immediately.
                    while (lastUpdate > System.nanoTime() - TimeUnit.SECONDS.toNanos(2)) {
                        try {
                            Thread.sleep(100L);
                        } catch (InterruptedException e) {
                        }
                    }
                    initialize();
                });
                thread.setDaemon(true);
                thread.start();
```

WRT to the "volatile" story of initialized I also think that the implementation is not correct.

If modified is called multiple times, multiple "initialize" calls are submitted.

And another point:
If we use an executor service, why not an scheduled one, so we could postpone the execution two seconds in the future without using a while loop...

kaikreuzer:
> If we use an executor service, why not an scheduled one, so we could postpone the execution two seconds in the future without using a while loop...

I think this would indeed make most sense, yes.

bundles/core/org.eclipse.smarthome.core.thing/src/main/java/org/eclipse/smarthome/core/thing/internal/ChannelItemProvider.java
```diff
@@ -174,6 +165,18 @@ protected synchronized void modified(Map<String, Object> properties) {
         }
     }
 
+    private void delayedInitialize(ScheduledExecutorService executor) {
+        // we wait until no further new links or items are announced in order to avoid creation of
+        // items which then must be removed again immediately.
+        final long diff = System.nanoTime() - lastUpdate - TimeUnit.SECONDS.toNanos(2);
+        if (diff < 0) {
```
kaikreuzer: Hm, looking closer, I am afraid that logic won't work. We are waiting the two seconds to see if `lastUpdate`might change meanwhile, so we have to repeat this if statement over and over - at the moment, we would only check it once when "modified" is called and never again.
So I guess we have to stick to the separate thread after all (or re-schedule a job upon every update of `lastUpdate`).

maggu2810:
Applied the following changes:

* using a scheduled executor service and drop while-loop and thread sleep usage
* shutdown executor
* fix threading: lastUpdate needs to be in sync between the different threads

bundles/core/org.eclipse.smarthome.core.thing/src/main/java/org/eclipse/smarthome/core/thing/internal/ChannelItemProvider.java
maggu2810: Hm, isn't this what I am doing now?
The diff between the 2 seconds delay is calculated and if it is too short another evaluation is scheduled.

kaikreuzer: But you only calculate it once, don't you?

maggu2810: E.g.
* `delayedInitialize` is called fist
* calculate the diff
* The diff result (-1.000.000.000) indicates that there is still 1 second to a 2 second delay
* The method `delayedInitialize` is scheduled again with a delay of 1 second (`-diff`)
* After about 1 second the scheduler will execute the scheduled task `delayedInitialize` again
* calculate the diff
* the diff is now >= 0
* initialize is called

If diff is < 0 again (because `lastUpdate` has been updated), the `delayedInitialize` function is scheduled again.

So the diff should be calculated, the delta is used to schedule another diff calculation if necessary.

kaikreuzer: Ok, thanks - I somehow had missed that you schedule `delayedInitialize` itself again.

PR #3703, opened 2017-06-20
Fix the  java.lang.RuntimePermission in RuleEngineCallbackImpl #3702

### LINKED ISSUE #3702: Java runtime permission is required in RuleEngineCallbackImpl
When the automation is running in security mode, a java.lang.RuntimePermission is required when a rule is removed.
The problem comes from `RuleEngineCallbackImpl` because `shutdownNow()` is called on `ExecutorService` instance.
[eshStackTrace.txt](https://github.com/eclipse/smarthome/files/1087401/eshStackTrace.txt)

My suggestion is to shange the `dispose()` method to look like this:

    public void dispose() {
        synchronized (this) {
            AccessController.doPrivileged(new PrivilegedAction<Void>() {

                @Override
                public Void run() {
                    executor.shutdownNow();
                    return null;
                }
            });
            executor = null;
        }
    }

plamen-peev:
executor.shutdownNow() is now in privileged call

Signed-off-by: Plamen Veselinov Peev <p.peev@prosyst.bg>

### DIFF
--- a/bundles/automation/org.eclipse.smarthome.automation.core/src/main/java/org/eclipse/smarthome/automation/core/internal/RuleEngineCallbackImpl.java
+++ b/bundles/automation/org.eclipse.smarthome.automation.core/src/main/java/org/eclipse/smarthome/automation/core/internal/RuleEngineCallbackImpl.java
@@ -7,6 +7,8 @@
  */
 package org.eclipse.smarthome.automation.core.internal;
 
+import java.security.AccessController;
+import java.security.PrivilegedAction;
 import java.util.Map;
 import java.util.concurrent.ExecutorService;
 import java.util.concurrent.Executors;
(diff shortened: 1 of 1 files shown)


danchom:
Add description in PR message title.

kaikreuzer:
Thanks - could you maybe also use the much nicher Java 8 lambda syntax as in https://github.com/eclipse/smarthome/pull/3706/files?

danchom:
Are you (@kaikreuzer, @maggu2810, @sjka) have something against this PR or it can be merge?

maggu2810:
IMHO it could be merged.

kaikreuzer:
@plamen-peev After pushing changes, it is helpful if you could leave a comment on the issue that you have updated it, because otherwise we do not get any notification and thus are not aware that things are ready for being merged. Thanks!

PR #3704, opened 2017-06-20
Solving radiation calculation issues and small corrections in channel…

clinique:
… definitions.

Signed-off-by: Gaël L'hopital <glhopital@gmail.com>

### DIFF
--- a/extensions/binding/org.eclipse.smarthome.binding.astro/ESH-INF/thing/channels.xml
+++ b/extensions/binding/org.eclipse.smarthome.binding.astro/ESH-INF/thing/channels.xml
@@ -100,7 +100,7 @@
 		<item-type>Number</item-type>
 		<label>Duration</label>
 		<description>The duration time in minutes of the event</description>
-		<state readOnly="true" pattern="%.1f min"/>
+		<state readOnly="true" pattern="%d min"/>
 	</channel-type>
 
 	<channel-type id="rangeEvent">
@@ -313,7 +313,7 @@
 
 	<channel-type id="fullMoon">
 		<item-type>DateTime</item-type>
-		<label>New moon</label>
+		<label>Full moon</label>
 		<description>The DateTime for full moon</description>
 		<state readOnly="true" pattern="%1$tF %1$tR" />
 	</channel-type>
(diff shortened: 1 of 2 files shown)


extensions/binding/org.eclipse.smarthome.binding.astro/ESH-INF/thing/channels.xml
```diff
@@ -329,14 +329,14 @@
...
-		<state readOnly="true" />
+		<state readOnly="true" pattern="%d d" />
 	</channel-type>
 
 	<channel-type id="illumination">
 		<item-type>Number</item-type>
 		<label>Moon illumination</label>
 		<description>The illumination of the moon in %</description>
-		<state readOnly="true" />
+		<state readOnly="true" pattern="%.2f %%"  />
```
kaikreuzer: Do we really need that with 2 fractal digits precision?

clinique: Sure, even %d would be far enough

extensions/binding/org.eclipse.smarthome.binding.astro/ESH-INF/thing/channels.xml
```diff
@@ -329,14 +329,14 @@
...
-		<state readOnly="true" />
+		<state readOnly="true" pattern="%d d" />
 	</channel-type>
 
 	<channel-type id="illumination">
 		<item-type>Number</item-type>
 		<label>Moon illumination</label>
 		<description>The illumination of the moon in %</description>
-		<state readOnly="true" />
+		<state readOnly="true" pattern="%d %%"  />
```
kaikreuzer: Just to be sure: Can you confirm that the values are ALWAYS only integers? Otherwise this will lead to exceptions, because floats cannot be formatted with %d. If we deal with floats, better use '%.0f'. (same question for duration and moon age above)

NorbertHD: Duration is a long, moon age is integer and illumination is double.
So only illumination needs a change  to "%.0f %%".

kaikreuzer:
@clinique Please update accordingly, I'll then merge!

clinique:
Done - sorry for lasting on this.

PR #3706, opened 2017-06-20
added doPriviledged to executor shutdown

kaikreuzer:


maggu2810:
Travis failed because the feature verification failed.
The feature verification failed because of a download error of bintray.

As you don't introduce new dependencies to non-JVM stuff, I merge this one regardless of the build error.

PR #3708, opened 2017-06-20
Converted I18NProviderImpl Groovy tests to Java

triller-telekom:


bundles/test/org.eclipse.smarthome.test/src/main/java/org/eclipse/smarthome/test/java/JavaOSGiTest.java
```diff
@@ -64,7 +64,7 @@ public void bindBundleContext() {
      *
      * @return bundle context
      */
-    private BundleContext getBundleContext() {
+    public BundleContext getBundleContext() {
```
maggu2810: I would prefer if you change the bundleContext member from default to protected.
getBundleContext() is used to initialize the bundleContext variable only once (I know the description is not correct).

triller-telekom: I did not notice that `bundleContext` is already `protected`. I just saw that the method `getBundleContext` existsted. But I agree that its better to use the `protected` member variable. I have amended my commit to fix it.

maggu2810:
Thank you

PR #3712, opened 2017-06-20
Remove altitude parameter from astro binding

### LINKED ISSUE #3710: Remove separate altitude parameter from Astro binding
When approving my auto-discovered Sun thing, I got this in my log:

```
15:52:30.491 [INFO ] [smarthome.event.ThingAddedEvent     ] - Thing 'astro:sun:local' has been added.
15:52:30.493 [INFO ] [me.event.ThingStatusInfoChangedEvent] - 'astro:sun:local' changed from UNINITIALIZED to INITIALIZING
15:52:30.500 [ERROR] [ding.astro.handler.AstroThingHandler] - Astro parameters geolocation could not be split into latitude and longitude, disabling thing 'astro:sun:local'
15:52:30.501 [INFO ] [me.event.ThingStatusInfoChangedEvent] - 'astro:sun:local' changed from INITIALIZING to OFFLINE
```
The issue is that the discovery now adds "<lat>,<long>,<alt>" in the geoLocation config parameter of the Thing, while it expects the altitude to be filled into the dedicated "altitude" parameter instead.

As we will soon have a [location context](https://github.com/eclipse/smarthome/issues/3709), which also includes altitude information, I think it would make most sense to remove the "altitude" parameter of Astro things and rather mark the geoLocation parameter to be of context "location", which implicitly can hold an optional altitude value.

triller-telekom:
Fixes #3710

### DIFF
--- a/extensions/binding/org.eclipse.smarthome.binding.astro/ESH-INF/config/config.xml
+++ b/extensions/binding/org.eclipse.smarthome.binding.astro/ESH-INF/config/config.xml
@@ -6,12 +6,9 @@
 
 	<config-description uri="thing-type:astro:config">
 		<parameter name="geolocation" type="text" required="true">
+			<context>location</context>
 			<label>GeoLocation</label>
-			<description>The latitude and longitude separated with a comma (xx.xxxxxx,xx.xxxxxx)</description>
-		</parameter>
-		<parameter name="altitude" type="integer" required="false">
-			<label>Altitude</label>
-			<description>The altitude of the GeoLocation in meter above sea level.</description>
+			<description>The latitude, longitude and altitude separated with a comma (xx.xxxxxx,xx.xxxxxx,xx)</description>
 		</parameter>
 		<parameter name="interval" type="integer" min="1" max="86400">
 			<label>Interval</label>
--- a/extensions/binding/org.eclipse.smarthome.binding.astro/ESH-INF/config/config.xml
+++ b/extensions/binding/org.eclipse.smarthome.binding.astro/ESH-INF/config/config.xml
@@ -7,8 +7,8 @@
 	<config-description uri="thing-type:astro:config">
 		<parameter name="geolocation" type="text" required="true">
 			<context>location</context>
-			<label>GeoLocation</label>
-			<description>The latitude, longitude and altitude separated with a comma (xx.xxxxxx,xx.xxxxxx,xx)</description>
+			<label>Location</label>
+			<description>The latitude, longitude and altitude separated with a comma (<lat>,<long>,<alt>)</description>
 		</parameter>
 		<parameter name="interval" type="integer" min="1" max="86400">
 			<label>Interval</label>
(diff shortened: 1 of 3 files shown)


extensions/binding/org.eclipse.smarthome.binding.astro/ESH-INF/config/config.xml
```diff
@@ -6,12 +6,9 @@
 
 	<config-description uri="thing-type:astro:config">
 		<parameter name="geolocation" type="text" required="true">
+			<context>location</context>
 			<label>GeoLocation</label>
```
kaikreuzer: Could we fix the label as well? "GeoLocation" is not a word... Probably best to change it to "Location".

extensions/binding/org.eclipse.smarthome.binding.astro/ESH-INF/config/config.xml
```diff
@@ -6,12 +6,9 @@
...
-			<description>The latitude and longitude separated with a comma (xx.xxxxxx,xx.xxxxxx)</description>
-		</parameter>
-		<parameter name="altitude" type="integer" required="false">
-			<label>Altitude</label>
-			<description>The altitude of the GeoLocation in meter above sea level.</description>
+			<description>The latitude, longitude and altitude separated with a comma (xx.xxxxxx,xx.xxxxxx,xx)</description>
```
kaikreuzer: Assuming that we will soon have a custom widget, I would remove the "(xx.xxxxxx,xx.xxxxxx,xx)" part (or maybe change it to `(<lat>,<long>,<alt>)`.

extensions/binding/org.eclipse.smarthome.binding.astro/src/main/java/org/eclipse/smarthome/binding/astro/internal/config/AstroThingConfig.java
```diff
@@ -36,6 +36,10 @@ public void parseGeoLocation() {
         if (geoParts.length == 2) {
             latitude = toDouble(geoParts[0]);
             longitude = toDouble(geoParts[1]);
+        } else if (geoParts.length == 3) {
+            latitude = toDouble(geoParts[0]);
+            longitude = toDouble(geoParts[1]);
+            altitude = Integer.parseInt(geoParts[2]);
```
kaikreuzer: Why do you assume that it is an int? PointType uses BigDecimal for altitude, so you should support this.

triller-telekom: That wasn't me :-) The astro binding used `Integer` for the altitude already internaly, so I parsed that configuration value to an `Integer`. I have changed it now to `double` so its the same as lat and lon inside the binding. I am not sure whether the whole binding should make use of `BigDecimal` instead of `Double` though...

kaikreuzer:
Sorry, just noticed that you missed updating the readme: https://github.com/eclipse/smarthome/tree/master/extensions/binding/org.eclipse.smarthome.binding.astro#thing-configuration

triller-telekom:
Will be done in a separate PR together with the README adjustments for the new unique location

PR #3715, opened 2017-06-21
[RFC] Mqtt transport: Redesign the service to cover more usecases.

### LINKED ISSUE #3695: io.transport.mqtt: Too restrictive
## API is too restrictive
The io.transport.mqtt bundle is very restrictive in what functionality it exports. At the moment it allows: 
* (un)registerConnectionObserver: Notification about dis/connection to a **known** broker.
* (un)registerMessageConsumer: Subscribe to a topic on a known broker.
* (un)registerMessageProducer: Send data to a known broker.

It does not allow to alter or enumerate the brokers, only known ones can be accessed. It does not allow to access the underlying MqttBrokerConnection.

## Flaws in the current implementation
But access would be necessary for the following reasons:
* Increase the security by not using the **getVeryTrustingTrustManager** [MqttBrokerConnection.java#L385](https://github.com/eclipse/smarthome/blob/master/bundles/io/org.eclipse.smarthome.io.transport.mqtt/src/main/java/org/eclipse/smarthome/io/transport/mqtt/internal/MqttBrokerConnection.java#L385) but an actual ``X509TrustManager`` implementation. 
* Finer control of what happens if the connection to the broker is lost and how it could be reestablished.There is only a fixed interval timer right now. A smarter solution might react to disabled/enabled network interfaces, check for a connection faster/slower than now.

## Proposed changes
All changes are **fully backwards compatible** to existing configuration files and API consumers.
I'd like to make the following changes to the bundle and would like to collect some feedback first:
1. Make the ``getConnection`` method of **MqttService** public. Make the **MqttBrokerConnection** class public.
2. Allow to replace the `X509TrustManager`` in a MqttBrokerConnection.
3. Add means, to add / remove / enumerate **MqttBrokerConnection** of **MqttService**. This way a broker discovery binding could be written, which adds/removes brokers to the **MqttService** and binding writers can write discovery services for ESH things, that represent a specific topic on a matching broker. 
4. Remove MqttBrokerConnection::setAsync. Use IMqttActionListener to be asynchronously notified of success/fail of a MqttDeliveryToken and notify the producer asynchronously in return, instead of block / waiting for a response [MqttBrokerConnection.java#L467](https://github.com/eclipse/smarthome/blob/master/bundles/io/org.eclipse.smarthome.io.transport.mqtt/src/main/java/org/eclipse/smarthome/io/transport/mqtt/internal/MqttBrokerConnection.java#L467).
5. The question is, how much is backwards compatibility important. It is possible to archive everything above without removing anything of the old MqttService interface, but it adds code clutter.

### Easy configuration of Mqtt Brokers
I'm using some Mqtt sensors and actors myself and know that it can be tedious to configure a working, securely connected Mqtt broker. I've seen a lot of forum topics about this subject in the OpenHab community.
I therefore plan an OpenHab binding, that
* uses auto discovery (most brokers announce themselves via mDNS Service Discovery) and
* the PaperUI to configure broker connections.
* The binding would also offer to use an embedded broker (the [Moquette broker](https://github.com/andsel/moquette)).

The configured brokers will end up in the MqttService, available to be consumed by other bindings additionally to the textual configured brokers.

### Mqtt vendor binding
A vendor that uses mqtt as communication protocol with an edge device that runs a mqtt broker. The vendor can easily write an ESH binding with the `MqttBrokerConnection`, equipped with the necessary ssl secrets and pairing mechanisms. 

### Binding that implements things/channels based on Mqtt topics
An imaginary future binding that doesn't want to care about Mqtt broker configurations and possibly delegate this task to a core binding like in suggestion 1.
There exists for example the "Homie" Mqtt convention (https://git.io/homieiot) or the EasyESP OpenHab Mqtt protocol. A binding would only need to map those known topics to statically defined things and channels like:

| Mqtt Topic                             | -cmd | -state | ESH Thing    | Channel |
|---------------------------------------|-------|--------|------------|--------------|
| next_great_mqtt_iot_spec/doorbell | set | on | doorbell | state   |

With the proposed changes, it would simply enumerate all configured brokers and find out which brokers support the wanted topics and list the available things in the inbox.

## Discussion
Topic in the ESH forum: https://www.eclipse.org/forums/index.php/m/1766231/#msg_1766231

davidgraeff:
Fixes #3695 
This is a request for comments PR, discussion takes places in the linked issue.

* This will keep the 'old' MqttService API untouched for full backwards compatibility.
* MqttBrokerConnection is exported now.
* Improve comments all over the code.

MqttService:
* Allows to add / remove / enumerate broker connections, additionally to the existing textual configuration.
* A new interface allows listeners to be notified of added / removed broker connections.
* All proxy methods are deprecated now. The MqttBrokerConnection can be accessed directly.
* Add unit test

MqttWillAndTestament:
* Introduce an isValid() method and return false if the build method fromString is not provided with all necessary properties.

MqttSenderChannel:
* Is depreacted now. MqttBrokerConnection.publish() should be used instead.
* The reason: The interface does not allow to implement an async behaviour, because the publisher can not be informed of the success/failure of the operation.

MqttMessageConsumer:
* is deprecated now and replaced by a slimer MqttMessageSubscriber interface without a forced setTopic and event method.  

MqttBrokerConnection:
* Reconnection is no longer handled via a fixed 60s interval timer, but fully customizable.
  The default implementation (PeriodicReconnectPolicy) resembles the old behaviour.
* A SSL broker connection no longer uses a non-configurable "trust all and everything" SSL context provider, but an interface that have to provide the SSL context provider. A default implementation, again, resembles the old behaviour.
* Remove 'async' configuration option. All communication aspects (connect/disconnect/publish/subscribe) are handled in an async fashion now.
* Add unit test


bundles/io/org.eclipse.smarthome.io.transport.mqtt/src/main/java/org/eclipse/smarthome/io/transport/mqtt/MqttWillAndTestament.java
```diff
@@ -45,39 +45,39 @@ public static MqttWillAndTestament fromString(String string) {
...
     }
 
     /**
-     * @return the topic for the last will.
+     * Return true if the last will and testament object is valid.
      */
-    public String getTopic() {
-        return topic;
+    private boolean isValid() {
+        return !StringUtils.isBlank(topic) && payload.length > 0;
```
maggu2810: An 0 byte will message is valid
http://docs.oasis-open.org/mqtt/mqtt/v3.1.1/os/mqtt-v3.1.1-os.html#_Toc385349244

maggu2810: It's been some time since I read the spec and wrote a MQTT broker, so excuse me if it is not correct, but if IIRC:
If the retained flag is used for MQTT messages, a null message is used to "un"-retain a topic again.
So using this approach to signal a disconnect (remove a retained topic) is a valid approach.

Miss I something?

davidgraeff: True. isValid() should solely check the topic.

bundles/io/org.eclipse.smarthome.io.transport.mqtt/src/main/java/org/eclipse/smarthome/io/transport/mqtt/MqttService.java
```diff
@@ -29,211 +34,315 @@
...
+        // load broker configurations from configuration file
+        if (properties == null || properties.isEmpty()) {
+            return;
+        }
+
+        Map<String, Map<String, Object>> brokerConfigs = extractBrokerConfigurations(properties);
+
+        for (Map<String, Object> brokerConfig : brokerConfigs.values()) {
+            try {
+                addBrokerConnection(brokerConfig).start();
```
htreu: Already available connections to the named broker will be returned but updated config will not be taken into account. Should an updated configuration restart existing connections with updated parameters?

bundles/io/org.eclipse.smarthome.io.transport.mqtt/src/main/java/org/eclipse/smarthome/io/transport/mqtt/MqttService.java
```diff
@@ -29,211 +34,315 @@
...
+        }
+    }
+
+    /**
+     *
+     * @param connection
+     */
+    public void addBrokerConnection(MqttBrokerConnection connection) {
+        synchronized (brokerConnections) {
+            brokerConnections.put(connection.getName(), connection);
```
htreu: Is it guaranteed that externally created BrokerConnections also have an all lower case name? Otherwise you will not be able to get it back from the MqttService (see line 144).

bundles/io/org.eclipse.smarthome.io.transport.mqtt/src/main/java/org/eclipse/smarthome/io/transport/mqtt/MqttService.java
```diff
@@ -29,211 +34,315 @@
...
+     * @param connection The broker connection
      */
-    private MqttBrokerConnection getConnection(String brokerName) {
+    public void removeBrokerConnection(MqttBrokerConnection connection) {
         synchronized (brokerConnections) {
-            MqttBrokerConnection conn = brokerConnections.get(brokerName.toLowerCase());
-            if (conn == null) {
-                conn = new MqttBrokerConnection(brokerName);
-                brokerConnections.put(brokerName.toLowerCase(), conn);
+            brokerConnections.remove(connection.getName(), connection);
```
htreu: This again will only work for external connections if the name is not all lower case.
In addition you may check the return value and call the observers only if the removal succeeded.

bundles/io/org.eclipse.smarthome.io.transport.mqtt/src/main/java/org/eclipse/smarthome/io/transport/mqtt/MqttService.java
```diff
@@ -29,211 +34,315 @@
...
+     *
+     * @param brokerName The broker name
+     */
+    public void removeBrokerConnection(String brokerName) {
+        synchronized (brokerConnections) {
+            MqttBrokerConnection connection = brokerConnections.get(brokerName.toLowerCase());
+            if (connection == null) {
+                return;
+            }
+            brokerConnections.remove(brokerName);
```
htreu: here again `brokerName` may not be equal to `brokerName.toLowerCase()` in which case the observers get notified but the connection is still in the map. You may call `brokerConnections.remove(brokerName.toLowerCase(), connection)` here and again check the return value. If the remove fails you may log.debug to give a hint what happened.

bundles/io/org.eclipse.smarthome.io.transport.mqtt/src/main/java/org/eclipse/smarthome/io/transport/mqtt/MqttService.java
```diff
@@ -29,211 +34,315 @@
...
      * Register a new message consumer which can process messages received on
      *
+     * @deprecated
      * @param brokerName Name of the broker on which to listen for messages.
      * @param mqttMessageConsumer Consumer which will process any received message.
      */
+    @Deprecated
     public void registerMessageConsumer(String brokerName, MqttMessageConsumer mqttMessageConsumer) {
-
         mqttMessageConsumer.setEventPublisher(eventPublisher);
```
htreu: not part of your PR and deprecated but I guess this should be done in line 283 only when the consumer was added to the connection.

bundles/io/org.eclipse.smarthome.io.transport.mqtt/src/main/java/org/eclipse/smarthome/io/transport/mqtt/internal/MqttSenderChannelImpl.java
```diff
@@ -0,0 +1,32 @@
+package org.eclipse.smarthome.io.transport.mqtt.internal;
```
htreu: missing license text

bundles/io/org.eclipse.smarthome.io.transport.mqtt/src/main/java/org/eclipse/smarthome/io/transport/mqtt/reconnect/AbstractReconnectPolicy.java
```diff
@@ -0,0 +1,25 @@
+package org.eclipse.smarthome.io.transport.mqtt.reconnect;
```
htreu: missing license text

bundles/io/org.eclipse.smarthome.io.transport.mqtt/src/main/java/org/eclipse/smarthome/io/transport/mqtt/reconnect/PeriodicReconnectPolicy.java
```diff
@@ -0,0 +1,58 @@
+package org.eclipse.smarthome.io.transport.mqtt.reconnect;
```
htreu: missing license text

bundles/io/org.eclipse.smarthome.io.transport.mqtt/src/main/java/org/eclipse/smarthome/io/transport/sslcontext/AcceptAllCertifcatesSSLContext.java
```diff
@@ -0,0 +1,45 @@
+package org.eclipse.smarthome.io.transport.sslcontext;
```
htreu: missing license text

bundles/io/org.eclipse.smarthome.io.transport.mqtt/src/main/java/org/eclipse/smarthome/io/transport/sslcontext/SSLContextProvider.java
```diff
@@ -0,0 +1,8 @@
+package org.eclipse.smarthome.io.transport.sslcontext;
```
htreu: missing license text

bundles/io/org.eclipse.smarthome.io.transport.mqtt/src/main/java/org/eclipse/smarthome/io/transport/mqtt/MqttService.java
```diff
@@ -29,211 +34,315 @@
...
+     * service configuration documentation for a detailed description.
+     *
+     * @param brokerConnectionConfig The configuration key-value map.
+     * @return Returns the created broker connection.
+     * @throws ConfigurationException Most likely your provided name and url are invalid.
+     * @throws MqttException
+     */
+    public MqttBrokerConnection addBrokerConnection(Map<String, Object> brokerConnectionConfig)
+            throws ConfigurationException, MqttException {
+        String name = brokerConnectionConfig.get("name").toString();
```
htreu: `"name"` should be a constant.

bundles/io/org.eclipse.smarthome.io.transport.mqtt/src/main/java/org/eclipse/smarthome/io/transport/mqtt/MqttService.java
```diff
@@ -29,211 +34,315 @@
...
-            if (StringUtils.isBlank(value)) {
-                logger.trace("Property is empty: {}", key);
-                continue;
-            } else {
-                logger.trace("Processing property: {} = {}", key, value);
+            Map<String, Object> brokerConfig = configPerBroker.get(brokername);
+            if (brokerConfig == null) {
+                brokerConfig = new TreeMap<>();
+                configPerBroker.put(brokername, brokerConfig);
+                brokerConfig.put("name", brokername);
```
htreu: `"name"` should be a constant.

bundles/io/org.eclipse.smarthome.io.transport.mqtt/src/main/java/org/eclipse/smarthome/io/transport/mqtt/MqttService.java
```diff
@@ -29,211 +34,315 @@
...
+     * @param brokerConnectionConfig The configuration key-value map.
+     * @return Returns the created broker connection.
+     * @throws ConfigurationException Most likely your provided name and url are invalid.
+     * @throws MqttException
+     */
+    public MqttBrokerConnection addBrokerConnection(Map<String, Object> brokerConnectionConfig)
+            throws ConfigurationException, MqttException {
+        String name = brokerConnectionConfig.get("name").toString();
+        if (name == null) {
+            throw new ConfigurationException("MQTT Broker property '{}' should have the format 'broker.propertykey'");
```
htreu: there is no value for `{}` here. You can name the `name` property right away here.

bundles/io/org.eclipse.smarthome.io.transport.mqtt/src/main/java/org/eclipse/smarthome/io/transport/mqtt/MqttService.java
```diff
@@ -29,211 +34,315 @@
...
-        }
-
+    /**
+     * Transform the JDK 1.0 old-school, rotten, deprecated Dictionary content to something useful.
+     *
+     * @param properties An old-school, deprecated dictionary object with the service configuration.
+     * @return A 'list' of broker configurations as key-value maps. A configuration map at least contains a "name".
+     */
+    public Map<String, Map<String, Object>> extractBrokerConfigurations(Dictionary<String, ?> properties) {
+        Map<String, Map<String, Object>> configPerBroker = new TreeMap<String, Map<String, Object>>();
```
davidgraeff: I prefer TreeMaps over HashMaps if I expect only a few entries and if only insert and no remove operation are performed (in this case in average 3 configuration entries per broker are added) to reduce the memory footprint. But if ESH Guidelines prefer HashMaps, I'll change it.

bundles/io/org.eclipse.smarthome.io.transport.mqtt/src/main/java/org/eclipse/smarthome/io/transport/mqtt/MqttService.java
```diff
@@ -29,211 +34,315 @@
...
 
     /**
-     * Start service.
+     * Create broker connections based on the service configuration.
      */
+    @Override
+    public void updated(Dictionary<String, ?> properties) {
+        // load broker configurations from configuration file
+        if (properties == null || properties.isEmpty()) {
+            return;
```
davidgraeff: Good point, to be sure I will disconnect and discard all existing, textual configured broker connections

bundles/io/org.eclipse.smarthome.io.transport.mqtt/src/main/java/org/eclipse/smarthome/io/transport/mqtt/MqttService.java
davidgraeff: It is not guaranteed, I will force lowercase names here and add a boolean return value (true==successfully added, false==already existing).

bundles/io/org.eclipse.smarthome.io.transport.mqtt/src/main/java/org/eclipse/smarthome/io/transport/mqtt/MqttService.java
davidgraeff: I agree

davidgraeff:
I addressed all issues except the unit tests, but cannot build right now. My eclipse is not able to find the  org.eclipse.paho.client.mqttv3 library anymore in the target configuration since the last update.

bundles/io/org.eclipse.smarthome.io.transport.mqtt/src/main/java/org/eclipse/smarthome/io/transport/mqtt/MqttService.java
htreu: I dont think this is a ESH guideline and I also guess the memory difference is not that big. What do you think @sjka?

sjka: No, I don't think we have a strict written guideline about it.

Nevertheless, it is common practice in this project to use sorted collections such as TreeSet, TreeMap etc. only if the order of elements has a relevant meaning. I don't think this is the case here, is it?

davidgraeff:
I have added unit tests as well for MqttService and MqttBrokerConnection (except start/close).

bundles/io/org.eclipse.smarthome.io.transport.mqtt/src/main/java/org/eclipse/smarthome/io/transport/mqtt/MqttService.java
davidgraeff: The order is irrelevant. I will change it then. I was just concerned of the in average 13 unused entries in a hash map (with an initial default capacity of 16). But because we store string references only it might be just a few bytes per entry anyway.

davidgraeff:
I've got one question: Should we keep the deprecated methods and classes? If there are no consumers of this API and ESH hasn't reached 1.0 it might be a good idea to remove them.

bundles/io/org.eclipse.smarthome.io.transport.mqtt/src/main/java/org/eclipse/smarthome/io/transport/mqtt/MqttService.java
```diff
@@ -25,215 +31,375 @@
...
+     * Add a broker by a configuration key-value map. You need to provide at least a "name" and an "url".
+     * Additional properties are "user","pwd","qos","retain","lwt","keepAlive","clientId", please read the
+     * service configuration documentation for a detailed description.
+     *
+     * @param brokerConnectionConfig The configuration key-value map.
+     * @return Returns the created broker connection or null if there is already a connection with the same name.
+     * @throws ConfigurationException Most likely your provided name and url are invalid.
+     * @throws MqttException
+     */
+    public MqttBrokerConnection addBrokerConnection(Map<String, Object> brokerConnectionConfig)
```
htreu: What other types of values instead of `String` do you expect here? I would suggest to expect a Map<String, String> which would result in property retrieval like

```
String brokerURL = brokerConnectionConfig.get("url");
if (StringUtils.isBlank(brokerURL) {
    throw new ConfigurationException("MQTT Broker property 'url' is not provided");
}
```

Right now passing in a blank String would satisfy all the checks but will fail on connecting to the broker.
On another note this would also make the check in line 66 obsolete.

bundles/io/org.eclipse.smarthome.io.transport.mqtt/src/main/java/org/eclipse/smarthome/io/transport/mqtt/MqttService.java
```diff
@@ -25,215 +31,375 @@
...
+            // ignore the only non-broker property..
             if (key.equals("service.pid")) {
-                // ignore the only non-broker property..
                 continue;
             }
 
             String[] subkeys = key.split("\\.");
-            if (subkeys.length != 2) {
-                logger.debug("MQTT Broker property '{}' should have the format 'broker.propertykey'", key);
+            if (subkeys.length != 2 || StringUtils.isBlank(value)) {
```
htreu: Are blank values a problem? Only `name` and `url` are required. Other properties might be blank. Its okay to check it here but with a brokerConfig map being <String, String> and a public API allowing to put even blank Strings (see my comment on line 203) this check is obsolete.

bundles/io/org.eclipse.smarthome.io.transport.mqtt.test/src/test/java/org/eclipse/smarthome/io/transport/mqtt/test/MqttBrokerConnectionTests.java
```diff
@@ -0,0 +1,185 @@
...
+            MqttBrokerConnection b = new MqttBrokerConnection("name", "ssl://123.123.123.123", true);
+            assertFalse(a.isTextualConfiguredBroker());
+            assertTrue(b.isTextualConfiguredBroker());
+        } catch (ConfigurationException e) {
+            fail("Unexpected configuration exception: " + e.getMessage());
+        }
+        // Test an invalid URL
+        try {
+            @SuppressWarnings("unused")
+            MqttBrokerConnection ignored = new MqttBrokerConnection("name", "unsupported://123.123.123.123", false);
```
htreu: I guess this is a matter of taste: I would prefer separate tests for these assertions. You may then shorten the test to something like

```
@Test(expected = ConfigurationException.class)
public void unsupportedProtocolShouldThrowException() throws Exception {
    new MqttBrokerConnection("name", "unsupported://123.123.123.123", false);
}
```

bundles/io/org.eclipse.smarthome.io.transport.mqtt.test/src/test/java/org/eclipse/smarthome/io/transport/mqtt/test/MqttBrokerConnectionTests.java
```diff
@@ -0,0 +1,185 @@
...
+    }
+
+    @Test
+    public void messageConsumerTests() {
+        final String url = "tcp://123.123.123.123";
+        final String name = "TestName12@!";
+        MqttBrokerConnection a;
+        try {
+            a = new MqttBrokerConnection(name, url, false);
+        } catch (ConfigurationException e) {
```
htreu: You may declare the test to `throws Exception` and just call the constructor here. This would clean up the test and it will fail anyway if the exception occurs.

bundles/io/org.eclipse.smarthome.io.transport.mqtt.test/src/test/java/org/eclipse/smarthome/io/transport/mqtt/test/MqttBrokerConnectionTests.java
```diff
@@ -0,0 +1,185 @@
...
+            }
+
+            @Override
+            public String getTopic() {
+                return "topic";
+
+            }
+        };
+        try {
+            a.addConsumer(subscriber);
```
htreu: same as in line 71/72. Just call `addConsumer` and let the test fail due to the anyhow thrown exception.

bundles/io/org.eclipse.smarthome.io.transport.mqtt.test/src/test/java/org/eclipse/smarthome/io/transport/mqtt/test/MqttBrokerConnectionTests.java
```diff
@@ -0,0 +1,185 @@
...
+        isConnected = true;
+        MqttConnectionObserver connectionObserver = new MqttConnectionObserver() {
+            @Override
+            public void setConnected(boolean connected) {
+                isConnected = connected;
+            }
+        };
+        a.addConnectionObserver(connectionObserver);
+        // Adding a connection observer should immediately call its setConnected() method.
+        assertFalse(isConnected);
```
htreu: Just as an alternative path using a mock:

```
MqttConnectionObserver connectionObserver = mock(MqttConnectionObserver.class);
a.addConnectionObserver(connectionObserver);
verify(connectionObserver).setConnected(false);
```

wdyt?

bundles/io/org.eclipse.smarthome.io.transport.mqtt.test/src/test/java/org/eclipse/smarthome/io/transport/mqtt/test/MqttBrokerConnectionTests.java
```diff
@@ -0,0 +1,185 @@
...
+        final String url = "tcp://123.123.123.123";
+        final String name = "TestName12@!";
+        MqttBrokerConnection a;
+        try {
+            a = new MqttBrokerConnection(name, url, false);
+        } catch (ConfigurationException e) {
+            fail();
+            return;
+        }
+        assertTrue("URL getter", a.getUrl().equals(url));
```
htreu: Please use `assertEquals(url, a.getUrl());` or even better `assertThat(a.getUrl(), is(url));` here. Please have a look at the other assertions where semantically more appropriate methods are available.

bundles/io/org.eclipse.smarthome.io.transport.mqtt.test/src/test/java/org/eclipse/smarthome/io/transport/mqtt/test/MqttBrokerConnectionTests.java
```diff
@@ -0,0 +1,185 @@
...
+        assertEquals(80, a.getKeepAliveInterval());
+
+        assertFalse(a.isRetain());
+        a.setRetain(true);
+        assertTrue(a.isRetain());
+
+        assertNull(a.getLastWill());
+        assertNull(MqttWillAndTestament.fromString(""));
+        a.setLastWill(MqttWillAndTestament.fromString("topic:message:1:true"));
+        assertNotNull(a.getLastWill());
```
htreu: this is obsolete since an NPE in the next line would also fail the test.

bundles/io/org.eclipse.smarthome.io.transport.mqtt.test/src/test/java/org/eclipse/smarthome/io/transport/mqtt/test/MqttServiceTests.java
```diff
@@ -0,0 +1,117 @@
...
+
+        MqttBrokerConnection connection;
+        try {
+            connection = new MqttBrokerConnection("name", "tcp://123.123.123.123", false);
+        } catch (ConfigurationException c) {
+            fail("Couldn't create a MqttBrokerConnection object");
+            return;
+        }
+        assertTrue(service.addBrokerConnection(connection));
+        assertTrue(brokerExisting);
```
htreu: Just like in the `MqttBrokerConnectionTests` this would be a little shorter with a mock `MqttBrokersObserver` and `org.mockito.Mockito.vrify`.

bundles/io/org.eclipse.smarthome.io.transport.mqtt.test/src/test/java/org/eclipse/smarthome/io/transport/mqtt/test/MqttServiceTests.java
```diff
@@ -0,0 +1,117 @@
...
+
+        service.removeBrokerConnection(connection);
+        assertFalse(brokerExisting);
```
htreu: same as in line 59.

bundles/io/org.eclipse.smarthome.io.transport.mqtt.test/src/test/java/org/eclipse/smarthome/io/transport/mqtt/test/MqttServiceTests.java
```diff
@@ -0,0 +1,117 @@
...
+    // Tests extractBrokerConfigurations() and addBrokerConnection(map)
+    @Test
+    public void textualConfigurationTests() {
+        MqttService service = new MqttService();
+
+        Dictionary<String, String> properties = new Hashtable<>();
+        properties.put("bam.name", "brokername");
+        properties.put("bam.url", "tcp://123.123.123.123");
+        Map<String, Map<String, Object>> map = service.extractBrokerConfigurations(properties);
+        assertNotNull(map);
```
htreu: obsolete, will fail with NPE in the next line.

bundles/io/org.eclipse.smarthome.io.transport.mqtt.test/src/test/java/org/eclipse/smarthome/io/transport/mqtt/test/MqttServiceTests.java
```diff
@@ -0,0 +1,117 @@
...
+        assertEquals(map.size(), 1);
+        Map<String, Object> data = map.get("bam");
+        assertNotNull(data);
```
htreu: same as line 77.

bundles/io/org.eclipse.smarthome.io.transport.mqtt.test/src/test/java/org/eclipse/smarthome/io/transport/mqtt/test/MqttServiceTests.java
```diff
@@ -0,0 +1,117 @@
...
+        assertTrue("brokername".equals(data.get("name")));
+        assertTrue("tcp://123.123.123.123".equals(data.get("url")));
+
+        assertEquals(service.getAllBrokerConnections().size(), 0);
```
htreu: good. even better: `assertThat(service.getAllBrokerConnections(), hasSize(0));` with `hasSize` being an static import of `IsCollectionWithSize.hasSize`. 
All in all: matter of taste.

bundles/io/org.eclipse.smarthome.io.transport.mqtt.test/src/test/java/org/eclipse/smarthome/io/transport/mqtt/test/MqttServiceTests.java
```diff
@@ -0,0 +1,117 @@
...
+        try {
+            service.addBrokerConnection(data);
+        } catch (ConfigurationException | MqttException e) {
```
htreu: same as in the `MqttBrokerConnectionTests`. You dont want to test the Exception here. Just let it fail with its own Exception.

davidgraeff:
Thanks for the test suite review, your suggestions are quite helpful and I will adapt the tests accordingly. In the Openhab world, writing tests with eclipse doesn't work right now because of https://github.com/openhab/openhab-distro/issues/230 so I'm a little inexperienced in this area.

bundles/io/org.eclipse.smarthome.io.transport.mqtt/src/main/java/org/eclipse/smarthome/io/transport/mqtt/MqttService.java
davidgraeff: No they are not a problem therefore we don't fail here, but log the unexpected name.key=no_value triple as debug log level. Improper used "key=" lines should be commented out instead of keeping the parser busy, I reckon.

bundles/io/org.eclipse.smarthome.io.transport.mqtt/src/main/java/org/eclipse/smarthome/io/transport/mqtt/MqttService.java
davidgraeff: I agree and adapted the code

bundles/io/org.eclipse.smarthome.io.transport.mqtt.test/src/test/java/org/eclipse/smarthome/io/transport/mqtt/test/MqttServiceTests.java
```diff
@@ -19,75 +21,60 @@
...
 /**
  * Tests the MqttService class
  *
  * @author David Graeff - Initial contribution
  */
 public class MqttServiceTests {
-    boolean brokerExisting = false;
+    @Before
+    public void setUp() {
+        MockitoAnnotations.initMocks(this);
```
htreu: You can remove `MockitoAnnotations.initMocks(this)` here because you dont have annotated mocks. Your mock in line 41 is created instant. The line in setup is necessary if you have instance members which are mocks like

```
@Mock
private MqttBrokersObserver observer;
```

This way `observer` will always be present. You might go this way if you use your mock in more than one test. So ne need to change it here.

bundles/io/org.eclipse.smarthome.io.transport.mqtt.test/src/test/java/org/eclipse/smarthome/io/transport/mqtt/test/MqttBrokerConnectionTests.java
```diff
@@ -28,125 +29,77 @@
  * @author David Graeff - Initial contribution
  */
 public class MqttBrokerConnectionTests {
-    boolean isConnected = true;
-
     @Before
     public void setUp() {
         MockitoAnnotations.initMocks(this);
```
htreu: see my comment in MqttServiceTests line 36.

bundles/io/org.eclipse.smarthome.io.transport.mqtt.test/src/test/java/org/eclipse/smarthome/io/transport/mqtt/test/MqttBrokerConnectionTests.java
```diff
@@ -28,125 +29,77 @@
...
-
-            }
-        };
-        try {
-            a.addConsumer(subscriber);
-        } catch (MqttException e) {
-            fail("addConsumer failed");
-        }
+        MqttMessageSubscriber subscriber = mock(MqttMessageSubscriber.class);
+        when(subscriber.getTopic()).thenAnswer(i -> "topic");
```
htreu: Nice! alternative (no need for you to change it here):

`when(subscriber.getTopic()).thenReturn("topic");`

bundles/io/org.eclipse.smarthome.io.transport.mqtt.test/src/test/java/org/eclipse/smarthome/io/transport/mqtt/test/MqttServiceTests.java
htreu: `setUp` is obsolete then, too.

bundles/io/org.eclipse.smarthome.io.transport.mqtt.test/src/test/java/org/eclipse/smarthome/io/transport/mqtt/test/MqttServiceTests.java
```diff
@@ -19,75 +21,60 @@
...
-        try {
-            connection = new MqttBrokerConnection("name", "tcp://123.123.123.123", false);
-        } catch (ConfigurationException c) {
-            fail("Couldn't create a MqttBrokerConnection object");
-            return;
-        }
+        MqttBrokerConnection connection = new MqttBrokerConnection("name", "tcp://123.123.123.123", false);
         assertTrue(service.addBrokerConnection(connection));
-        assertTrue(brokerExisting);
+        verify(observer).brokerAdded(connection);
```
htreu: nice!

davidgraeff:
I changed all occurrences of TreeMap to HashMap and removed the not used setups of the test suits.

sjka:
Thanks for this contribution! It's a good work for the better of the MQTT transport.
As it exceed 1000 lines of code we unfortunately have to run it through the Eclipse IP process.  This usually takes just a few days, no too bad. The good news for you is that you normally don't need to do anything for it 😄 

So please do not change this PR anymore. If anything would be needed to be changed, then it needs to go through a separate PR. 

For the record: I have created CQ 13824.

sjka:
Good news, the CQ was approved!

sjka:
Thanks!

PR #3717, opened 2017-06-21
Remove unnecessary localeProviderHolder from NTP Binding

triller-telekom:


maggu2810:
I know the intention but it has been never necessary because the locale provider is a static mandatory dependency of the factory and so the thing handlers are re-created if another locale provider will be injected (factory deactivated -> unbind old locale provider -> bind new locale provider -> factory activated -> recreate thing handlers). Correct?

kaikreuzer:
> Correct?

I think so :-)

maggu2810:
And the intention has been (e.g. locale provider is inject using dynamic policy) that if the locale provider reference changes, the thing handler will use the new reference automatically. The holder will stay the same but holds another reference...
As stated above: not necessary at all.

Issue #3711, opened 2017-06-20
 [FS Internetradio] OutOfMemory Leaks with openHAB 2?


bytedealer:
I am using openHAB 2 on a DS214play (2.0.0.001-DSM6) and bought some Silvercrest SMRS 35 A1 and a SMRB 40 A1 with FS chip. Everything is starting and being controlled perfectly. But after a couple of hours, less than one day, there is an interruption and errors coming up.

First:

```
2017-06-18 14:09:48.268 [ERROR] [ore.internal.events.OSGiEventManager] - Dispatching/filtering event for subscriber 'org.eclipse.smarthome.core.events.EventSubscriber' failed: java.lang.OutOfMemoryError: unable to create new native thread
java.util.concurrent.ExecutionException: java.lang.OutOfMemoryError: unable to create new native thread
at java.util.concurrent.FutureTask.report(FutureTask.java:122)[:1.8.0_91]
at java.util.concurrent.FutureTask.get(FutureTask.java:206)[:1.8.0_91]
at org.eclipse.smarthome.core.common.SafeMethodCaller.callAsynchronous(SafeMethodCaller.java:188)[98:org.eclipse.smarthome.core:0.9.0.b4]
at org.eclipse.smarthome.core.common.SafeMethodCaller.call(SafeMethodCaller.java:81)[98:org.eclipse.smarthome.core:0.9.0.b4]
at org.eclipse.smarthome.core.common.SafeMethodCaller.call(SafeMethodCaller.java:65)[98:org.eclipse.smarthome.core:0.9.0.b4]
at org.eclipse.smarthome.core.internal.events.OSGiEventManager.dispatchESHEvent(OSGiEventManager.java:188)[98:org.eclipse.smarthome.core:0.9.0.b4]
at org.eclipse.smarthome.core.internal.events.OSGiEventManager.handleEvent(OSGiEventManager.java:163)[98:org.eclipse.smarthome.core:0.9.0.b4]
at org.eclipse.smarthome.core.internal.events.OSGiEventManager.handleEvent(OSGiEventManager.java:146)[98:org.eclipse.smarthome.core:0.9.0.b4]
at org.apache.felix.eventadmin.impl.handler.EventHandlerProxy.sendEvent(EventHandlerProxy.java:415)[6:org.apache.karaf.services.eventadmin:4.0.8]
[...]

PR #3720, opened 2017-06-21
Reuse same instance of HttpClient in FSInternetRadio

svilenvul:
Removed unused method.
Added test case.

Fixes: #3711

### DIFF
--- a/extensions/binding/org.eclipse.smarthome.binding.fsinternetradio.test/src/test/groovy/org/eclipse/smarthome/binding/fsinternetradio/test/FSInternetRadioHandlerOSGiTest.groovy
+++ b/extensions/binding/org.eclipse.smarthome.binding.fsinternetradio.test/src/test/groovy/org/eclipse/smarthome/binding/fsinternetradio/test/FSInternetRadioHandlerOSGiTest.groovy
@@ -10,11 +10,12 @@ package org.eclipse.smarthome.binding.fsinternetradio.test
 import static org.hamcrest.CoreMatchers.*
 import static org.junit.Assert.*
 
+import org.apache.commons.lang.StringUtils
+import org.eclipse.smarthome.binding.fsinternetradio.FSInternetRadioBindingConstants
+import org.eclipse.smarthome.binding.fsinternetradio.handler.FSInternetRadioHandler
+import org.eclipse.smarthome.binding.fsinternetradio.internal.FSInternetRadioHandlerFactory
+import org.eclipse.smarthome.binding.fsinternetradio.internal.radio.FrontierSiliconRadioConnection
 import org.eclipse.smarthome.config.core.Configuration
-import org.eclipse.smarthome.config.discovery.DiscoveryResult
-import org.eclipse.smarthome.config.discovery.DiscoveryService
-
-import org.eclipse.smarthome.core.items.GenericItem
 import org.eclipse.smarthome.core.items.Item
 import org.eclipse.smarthome.core.items.ItemRegistry
 import org.eclipse.smarthome.core.library.items.*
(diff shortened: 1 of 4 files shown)


extensions/binding/org.eclipse.smarthome.binding.fsinternetradio/src/main/java/org/eclipse/smarthome/binding/fsinternetradio/internal/radio/FrontierSiliconRadioConnection.java
```diff
@@ -50,7 +50,7 @@
     private String sessionId;
 
     /** http clients, store cookies, so it is kept in connection class. */
-    private HttpClient httpClient = null;
+    private static final HttpClient HTTP_CLIENT = new HttpClient();
```
kaikreuzer: Does it really have to be static? First of all, the javadoc does then not really fit anymore.
Secondly, I would expect that there is only one connection per radio - I see that you mentioned that new instances are created over and over again, but maybe this is then the issue that should be addressed?

svilenvul: No, it doesn't have to be. 

OK, in this case I will try to address the issue with the multiple instances with as little as possible changes in the handler.

svilenvul:
The problem was in this block being executed periodically:
```
                    try {
                        final FrontierSiliconRadio tmpRadio = new FrontierSiliconRadio(ip, port.intValue(), pin);
                        tmpRadio.login();
                        radio = tmpRadio; // login successful, now store radio in field

                        // Thing initialized. If done set status to ONLINE to indicate proper working.
                        updateStatus(ThingStatus.ONLINE);

                        // now update all channels!
                        updateRunnable.run();
                    } catch (Exception e) {
                        updateStatus(ThingStatus.OFFLINE, ThingStatusDetail.COMMUNICATION_ERROR, e.getMessage());
                    }
```

Changing it required some other modifications. Please take a look if you will approve them.

extensions/binding/org.eclipse.smarthome.binding.fsinternetradio.test/src/test/groovy/org/eclipse/smarthome/binding/fsinternetradio/test/FSInternetRadioHandlerOSGiTest.groovy
```diff
@@ -658,6 +647,32 @@ public class FSInternetRadioHandlerOSGiTest extends OSGiTest{
...
+    @Test
+    void 'verify error on login doesnt craete new threads' () {
+        def wrongPin = "5678"
+        // The default number of threads that are created by the Jetty HttpClient for each new client instance
+        def threadsCreatedByHttpClient = 8
+        Configuration config = createConfiguration(wrongPin, DEFAULT_CONFIG_PROPERTY_IP, DEFAULT_CONFIG_PROPERTY_PORT, DEFAULT_CONFIG_PROPERTY_REFRESH)
+
+        Thing radioThing = initializeRadioThing(config)
+
+        int threadsBeforeSleep = getThreadsCountByName("HttpClient")
```
sjka: This relies of some interna of another library, out of our control

extensions/binding/org.eclipse.smarthome.binding.fsinternetradio/src/main/java/org/eclipse/smarthome/binding/fsinternetradio/handler/FSInternetRadioHandler.java
```diff
@@ -31,29 +32,30 @@
...
  * @author Mihaela Memova - removed the unused boolean parameter, changed the check for the PIN
+ * @author Svilen Valkanov - changed handler initialization
  */
 public class FSInternetRadioHandler extends BaseThingHandler {
 
     private Logger logger = LoggerFactory.getLogger(FSInternetRadioHandler.class);
 
     private FrontierSiliconRadio radio;
 
+    private boolean isLoginSuccessful = false;
```
sjka: I'm just thinking of whether it would be better to expose radio->conn->isLoggedIn somehow nicely instead of maintaining our own state here. It'll always be a struggle to keep those two in sync.

sjka:
Thanks!

PR #3721, opened 2017-06-21
Fixed XML Errors

amitjoy:
The XML Comprised `<` and `>` symbols that interpreted the inline words as attributes.

### DIFF
--- a/extensions/binding/org.eclipse.smarthome.binding.astro/ESH-INF/config/config.xml
+++ b/extensions/binding/org.eclipse.smarthome.binding.astro/ESH-INF/config/config.xml
@@ -8,7 +8,7 @@
 		<parameter name="geolocation" type="text" required="true">
 			<context>location</context>
 			<label>Location</label>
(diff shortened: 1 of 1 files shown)


extensions/binding/org.eclipse.smarthome.binding.astro/ESH-INF/config/config.xml
```diff
@@ -8,7 +8,7 @@
 		<parameter name="geolocation" type="text" required="true">
 			<context>location</context>
 			<label>Location</label>
-			<description>The latitude, longitude and altitude separated with a comma (<lat>,<long>,<alt>)</description>
+			<description>The latitude, longitude and altitude separated with a comma ([lat],[long],[alt])</description>
```
kaikreuzer: No, square brackets usually define an optional part.
It should be `The latitude, longitude and optionally altitude separated with a comma (<lat>,<long>,[<alt>])`.
And of course, the diamond brakes need to be XML encoded as `&lt`; and `&gt;`.

PR #3724, opened 2017-06-22
fix concurrent modification of expressionParts in AbstractExpression

### LINKED ISSUE #3719: CME in AbstractExpression of Scheduler
Just found this in my log:

```
Exception in thread "ESH-astro-Monitor-1" java.util.ConcurrentModificationException
	at java.util.ArrayList.sort(ArrayList.java:1456)
	at java.util.Collections.sort(Collections.java:141)
	at org.eclipse.smarthome.core.scheduler.AbstractExpression.applyExpressionParts(AbstractExpression.java:195)
	at org.eclipse.smarthome.core.scheduler.AbstractExpression.parseExpression(AbstractExpression.java:173)
	at org.eclipse.smarthome.core.scheduler.AbstractExpression.parseExpression(AbstractExpression.java:141)
	at org.eclipse.smarthome.core.scheduler.AbstractExpression.setStartDate(AbstractExpression.java:96)
	at org.eclipse.smarthome.core.scheduler.CronExpression.setStartDate(CronExpression.java:188)
	at org.eclipse.smarthome.core.scheduler.AbstractExpression.getTimeAfter(AbstractExpression.java:265)
	at org.eclipse.smarthome.core.scheduler.ExpressionThreadPoolManager$ExpressionThreadPoolExecutor$1.run(ExpressionThreadPoolManager.java:166)
```

At a first glance, I have no idea how we can run in concurrent modifications here, but there must be some situations that need to be considered.

sjka:
fixes #3719

### DIFF
--- a/bundles/core/org.eclipse.smarthome.core/src/main/java/org/eclipse/smarthome/core/scheduler/AbstractExpression.java
+++ b/bundles/core/org.eclipse.smarthome.core/src/main/java/org/eclipse/smarthome/core/scheduler/AbstractExpression.java
@@ -12,6 +12,8 @@ import java.util.ArrayList;
 import java.util.Calendar;
 import java.util.Collections;
 import java.util.Date;
+import java.util.LinkedList;
+import java.util.List;
 import java.util.StringTokenizer;
 import java.util.TimeZone;
 
@@ -34,7 +36,7 @@ public abstract class AbstractExpression<E extends AbstractExpressionPart> imple
 
     private String expression;
     private String delimiters;
-    private ArrayList<E> expressionParts = new ArrayList<E>();
+    private List<E> expressionParts = Collections.emptyList();
 
     private boolean continueSearch;
     private ArrayList<Date> candidates = new ArrayList<Date>();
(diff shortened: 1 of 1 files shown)


sjka:
we probably have similar potential  fun with the `candidates` collections,  but this PR addresses the `expressionParts` aspect only.

bundles/core/org.eclipse.smarthome.core/src/main/java/org/eclipse/smarthome/core/scheduler/AbstractExpression.java
```diff
@@ -356,13 +359,13 @@ protected void clearCandidates() {
         this.candidates = null;
     }
 
-    public ArrayList<E> getExpressionParts() {
+    public List<E> getExpressionParts() {
         return expressionParts;
```
maggu2810: Should we return a immutable copy here so we detect other places where the internal expression list is changed without sync.?

sjka: ack! see the setter -> I realize this was surprising, so I will change it to make it unmodifiable here instead of keeping an unmodifiable collection all the time...

maggu2810: Ah, missed it. :1:

PR #3725, opened 2017-06-22
Astro Background discovery for location changes

triller-telekom:


extensions/binding/org.eclipse.smarthome.binding.astro/README.md
```diff
@@ -187,3 +189,7 @@ then
     ...
 end
 ```
+
+##Tipps
```
maggu2810: Shouldn't there be a space between `##` and `Things`

extensions/binding/org.eclipse.smarthome.binding.astro/src/main/java/org/eclipse/smarthome/binding/astro/discovery/AstroDiscoveryLocationChangedTask.java
```diff
@@ -0,0 +1,23 @@
...
+    private PointType previousLocation;
+
+    public AstroDiscoveryLocationChangedTask(AstroDiscoveryService astroDiscoveryService) {
+        this.astroDiscoveryService = astroDiscoveryService;
+    }
+
+    @Override
+    public void run() {
+        PointType currentLocation = astroDiscoveryService.getLocationProvider().getLocation();
+        if (currentLocation != previousLocation) {
```
maggu2810: Do you really want to compare the reference, regardless if another point type variable points to the same location? Shouldn't we use equals?

extensions/binding/org.eclipse.smarthome.binding.astro/src/main/java/org/eclipse/smarthome/binding/astro/discovery/AstroDiscoveryService.java
```diff
@@ -31,42 +36,77 @@
...
         }
-        thingDiscovered(DiscoveryResultBuilder.create(sunThing).withLabel("Local Sun")
+        thingDiscovered(DiscoveryResultBuilder.create(SUN_THING).withLabel("Local Sun")
                 .withProperty("geolocation", propGeolocation).build());
-        thingDiscovered(DiscoveryResultBuilder.create(moonThing).withLabel("Local Moon")
+        thingDiscovered(DiscoveryResultBuilder.create(MOON_THING).withLabel("Local Moon")
                 .withProperty("geolocation", propGeolocation).build());
     }
 
+    public LocationProvider getLocationProvider() {
```
maggu2810: We could also thing about adding the location provider to the constructor argument of `AstroDiscoveryLocationChangedTask` without introducing a new function.

extensions/binding/org.eclipse.smarthome.binding.astro/README.md
```diff
@@ -10,17 +10,19 @@ This binding supports two Things: Sun and Moon
 
 ## Discovery
 
-Discovery is not necessary, because all calculations are done within the binding.
+You can use the auto discovery if you have a location set in "Configuration - System - Regional Settings - Location". This will automatically create a "Local Sun" and a "Local Moon" item for this location.
```
kaikreuzer: Simply say "if you have a system location set". There might be many ways how to do that, it is not up to the binding to know how.

extensions/binding/org.eclipse.smarthome.binding.astro/README.md
```diff
@@ -10,17 +10,19 @@ This binding supports two Things: Sun and Moon
...
+If you change your system location the background discovery detects this change and applies it to the "Local Sun" and a "Local Moon" items automatically.
 
 ## Binding Configuration
 
 No binding configuration required.
 
 ## Thing Configuration
 
-A thing requires the geolocation (latitude, longitude) for which the calculation is done.
+A thing requires the geolocation (latitude, longitude, (altitude)) for which the calculation is done.
```
kaikreuzer: The comma should be optional as well. And as commented on the PR, I would prefer the syntax
```
(latitude,longitude[,altitude])
```
especially as we now have two levels of brackets.

extensions/binding/org.eclipse.smarthome.binding.astro/README.md
```diff
@@ -187,3 +189,7 @@ then
     ...
 end
 ```
+
+##Tipps
+
+Do not worry if for example the "astro dawn" is "-" in the place that you set for your item. The reason might be that you live in a northern country and it is summer, because then by definition the sun is not 18 degrees below the horizon in the morning. For details see: https://en.wikipedia.org/wiki/Dawn Also while reading this wikipedia article you might come to the conclusion that you want to use "civil dawn" anyway.
```
kaikreuzer: Please do every sentence on a new line for easier diffing.

extensions/binding/org.eclipse.smarthome.binding.astro/src/main/java/org/eclipse/smarthome/binding/astro/discovery/AstroDiscoveryLocationChangedTask.java
```diff
@@ -0,0 +1,23 @@
+package org.eclipse.smarthome.binding.astro.discovery;
```
kaikreuzer: license?

extensions/binding/org.eclipse.smarthome.binding.astro/src/main/java/org/eclipse/smarthome/binding/astro/discovery/AstroDiscoveryLocationChangedTask.java
triller-telekom: Oh damn... It was late and warm in the office yesterday... O_o

extensions/binding/org.eclipse.smarthome.binding.astro/src/main/java/org/eclipse/smarthome/binding/astro/discovery/AstroDiscoveryLocationChangedTask.java
```diff
@@ -0,0 +1,23 @@
...
+
+import org.eclipse.smarthome.core.library.types.PointType;
+
+public class AstroDiscoveryLocationChangedTask implements Runnable {
```
kaikreuzer: I wouldn't know why we would need this logic in its own public (and exported) class.

extensions/binding/org.eclipse.smarthome.binding.astro/src/main/java/org/eclipse/smarthome/binding/astro/discovery/AstroDiscoveryService.java
```diff
@@ -31,42 +36,77 @@
...
+        try {
+            expression = new CronExpression(
+                    CronHelper.createCronForRepeatEverySeconds(LOCATION_CHANGED_CHECK_INTERVAL));
+        } catch (ParseException e) {
+            return;
+        }
+        if (expression != null && backgroundJob == null) {
+            AstroDiscoveryLocationChangedTask job = new AstroDiscoveryLocationChangedTask(this);
+            scheduledExecutor.schedule(job, expression);
+            logger.info("Scheduled astro location-changed job every {} seconds", LOCATION_CHANGED_CHECK_INTERVAL);
```
kaikreuzer: debug is enough

extensions/binding/org.eclipse.smarthome.binding.astro/src/main/java/org/eclipse/smarthome/binding/astro/discovery/AstroDiscoveryService.java
kaikreuzer: Just completely remove `AstroDiscoveryLocationChangedTask` and use a lambda here. No need for this function either in that case.

extensions/binding/org.eclipse.smarthome.binding.astro/src/main/java/org/eclipse/smarthome/binding/astro/discovery/AstroDiscoveryService.java
```diff
@@ -31,39 +36,70 @@
...
             return;
         }
+        createResults(location);
+    }
+
+    @Override
+    protected void startBackgroundDiscovery() {
+        Expression expression = null;
+        try {
+            expression = new CronExpression(
```
kaikreuzer: Why do we need the complexity of a cron job here?

triller-telekom: @sjka suggested that we can make use of our scheduler. I thought its a good idea.

kaikreuzer: I doubt that this is what he meant.
Every discovery service already HAS a `scheduler` available. I see no reason why this should not be used here. Check the docs at https://www.eclipse.org/smarthome/documentation/development/bindings/discovery-services.html#background-discovery (and please improve them, if they weren't clear enough on that point!).

extensions/binding/org.eclipse.smarthome.binding.astro/src/main/java/org/eclipse/smarthome/binding/astro/discovery/AstroDiscoveryLocationChangedTask.java
triller-telekom: I am now using a lambda expression instead

triller-telekom:
`previousLocation` can also be `null`. And I can only proceed if I have a `currentLocation != null`, so I think my check is correct.

PS: can you make you comments next to the code next time please? its easier to reply then.

amitjoy:
@triller-telekom I actually commented at the source but I don't why it showed this way. I believe concurrent reviews and changes to the same files caused github to arrange it differently. Anyways thanks for noticing it.

extensions/binding/org.eclipse.smarthome.binding.astro/src/main/java/org/eclipse/smarthome/binding/astro/discovery/AstroDiscoveryService.java
```diff
@@ -31,39 +32,69 @@
...
+            }, 0, LOCATION_CHANGED_CHECK_INTERVAL, TimeUnit.SECONDS);
+            logger.debug("Scheduled astro location-changed job every {} seconds", LOCATION_CHANGED_CHECK_INTERVAL);
+        }
+    }
+
+    @Override
+    protected void stopBackgroundDiscovery() {
+        logger.debug("Stopping Astro device background discovery");
+        if (astroDiscoveryJob != null && !astroDiscoveryJob.isCancelled()) {
+            astroDiscoveryJob.cancel(true);
```
amitjoy: you can actually use the returned boolean value to nullify `astroDiscoveryJob` and print the log.

extensions/binding/org.eclipse.smarthome.binding.astro/src/main/java/org/eclipse/smarthome/binding/astro/discovery/AstroDiscoveryService.java
```diff
@@ -31,39 +32,69 @@
...
+    }
 
-        ThingUID sunThing = new ThingUID(AstroBindingConstants.THING_TYPE_SUN, LOCAL);
-        ThingUID moonThing = new ThingUID(AstroBindingConstants.THING_TYPE_MOON, LOCAL);
+    @Override
+    protected void startBackgroundDiscovery() {
+        if (astroDiscoveryJob == null) {
+            astroDiscoveryJob = scheduler.scheduleAtFixedRate(() -> {
+                PointType currentLocation = locationProvider.getLocation();
+                if ((currentLocation != null) && !currentLocation.equals(previousLocation)) {
```
amitjoy: As suggested, you can use `Objects#equals(Object, Object)` as well

extensions/binding/org.eclipse.smarthome.binding.astro/README.md
```diff
@@ -187,3 +189,7 @@ then
     ...
 end
 ```
+
+## Tipps
```
amitjoy: English: Tip German: Tipp 😉

extensions/binding/org.eclipse.smarthome.binding.astro/src/main/java/org/eclipse/smarthome/binding/astro/discovery/AstroDiscoveryService.java
```diff
@@ -31,39 +32,69 @@
 public class AstroDiscoveryService extends AbstractDiscoveryService {
     private final Logger logger = LoggerFactory.getLogger(AstroDiscoveryService.class);
     private static final int DISCOVER_TIMEOUT_SECONDS = 30;
+    private static final int LOCATION_CHANGED_CHECK_INTERVAL = 60;
     private LocationProvider locationProvider;
+    private ScheduledFuture<?> astroDiscoveryJob;
+    private PointType previousLocation;
+
+    private static ThingUID SUN_THING = new ThingUID(THING_TYPE_SUN, LOCAL);
+    private static ThingUID MOON_THING = new ThingUID(THING_TYPE_MOON, LOCAL);
```
amitjoy: Capitalized variables are primarily used for constant string literals.

sjka: ...so that okay - let's worry why they're not declared `final` here...?! 😉

amitjoy: You are right. That's also my concern 😉.

Issue #1702, opened 2016-06-20
MapDB storage backend


maggu2810:
Caused by the recent changes, that properties are transient and so no longer persisted (#1682) I have a look at the esh storage.mapdb bundle.

There is one storage database (storage.mapdb) and all the maps are created in this backend.
[Here](https://github.com/eclipse/smarthome/blob/bb0b5f6/bundles/storage/org.eclipse.smarthome.storage.mapdb/src/main/java/org/eclipse/smarthome/storage/mapdb/MapDbStorage.java#L47) a new storage service is created.

I do not find any methods that removes such a storage service again.
Is this by intention?
Shouldn't the [storage interface](https://github.com/eclipse/smarthome/blob/bb0b5f6a68aafbbae72b38d34412c93b0a6f8c86/bundles/core/org.eclipse.smarthome.core/src/main/java/org/eclipse/smarthome/core/storage/Storage.java#L19) provide a method to remove the storage.

If someone want at least remove the whole content of the storage (without removing it or rely that the db is such clever and remove empty ones itself) shouldn't we add a clear function?

The function getKeys returns a key set and if my reading is correct, that is not documented.
So a simple loop over the return value of getKeys and a remove will result in a CME, correct?
Should we change the getKeys of the mapdb storage to create a new collection that contains all keys?

kaikreuzer:
> properties are transient and so no longer persisted (#1682) 

This isn't fully correct - properties are still persisted, it is just that bindings themselves do not persist changes anymore.

> I do not find any methods that removes such a storage service again.
> Is this by intention?

I simply assume that nobody needed such a feature yet and it sounds dangerous ;-)

> shouldn't we add a clear function?

Feel free.

> The function getKeys returns a key set and if my reading is correct, that is not documented.

What do you mean by "not documented"? The interface has:

```
    /**
     * Gets all keys of this Storage.
     * 
     * @return the keys of this Storage
     */
    Collection<String> getKeys();
```
but I agree that this should actually rather be a Set than a Collection.

> Should we change the getKeys of the mapdb storage to create a new collection that contains all keys?

No idea what the MapDB implementation exactly returns upon a map.keySet(), but yes, it probably is safer to return a copy.

maggu2810:
> What do you mean by "not documented"? 

I would like to say, that it is not documented that the returned collection is a **view** and so it is dangerous to use the returned collection while changing the storage.

PR #3729, opened 2017-06-23
deletable storage, return set instead view

maggu2810:
Fixes: #1702

### DIFF
--- a/bundles/core/org.eclipse.smarthome.core/src/main/java/org/eclipse/smarthome/core/storage/DisposableStorage.java
+++ b/bundles/core/org.eclipse.smarthome.core/src/main/java/org/eclipse/smarthome/core/storage/DisposableStorage.java
@@ -1,0 +1,26 @@
+/**
+ * Copyright (c) 2014-2017 by the respective copyright holders.
+ * All rights reserved. This program and the accompanying materials
+ * are made available under the terms of the Eclipse Public License v1.0
+ * which accompanies this distribution, and is available at
+ * http://www.eclipse.org/legal/epl-v10.html
+ */
+package org.eclipse.smarthome.core.storage;
+
+/**
+ * A {@code Storage} that could be disposed.
+ *
+ * @author Markus Rathgeb - Initial Contribution and API
+ */
+public interface DisposableStorage<T> extends Storage<T> {
+
+    /**
+     * Dispose the storage.
+     *
+     * <p>
+     * This function could be called if the storage is not longer used (now and in future). The storage implementation
+     * will clean up / remove the storage (e.g. file, database, etc.).
+     * After this function has been called the storage must not be used anymore.
+     */
+    void dispose();
+}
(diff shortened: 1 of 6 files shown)


maggu2810:
I missed the disposable storage service, will add it tomorrow.

bundles/core/org.eclipse.smarthome.core/src/main/java/org/eclipse/smarthome/core/storage/DisposableStorage.java
```diff
@@ -0,0 +1,26 @@
...
+ * http://www.eclipse.org/legal/epl-v10.html
+ */
+package org.eclipse.smarthome.core.storage;
+
+/**
+ * A {@code Storage} that could be disposed.
+ *
+ * @author Markus Rathgeb - Initial Contribution and API
+ */
+public interface DisposableStorage<T> extends Storage<T> {
```
sjka: I'm not 100% sure whether this is the best name, because (at least to me) it is associated destroying UI elements /freeing memory etc. and not with something as dangerous and terminal as deleting the database. How about `DeletableStoryge` - weird word, not happy with this suggestion myself I have to admit though 😕

bundles/storage/org.eclipse.smarthome.storage.mapdb/src/main/java/org/eclipse/smarthome/storage/mapdb/MapDbStorage.java
```diff
@@ -69,7 +83,7 @@ public T get(String key) {
 
     @Override
     public Collection<String> getKeys() {
-        return map.keySet();
+        return new HashSet<>(map.keySet());
```
sjka: Good catch 👍

sjka:
Woudln't it make sense to also implement it in the JSON DB?

maggu2810:
> Woudln't it make sense to also implement it in the JSON DB?

I would like to create an interface for a storage and storage service that could be deleted / disposed.
The MapDB itself already provides that feature and so I use this as a reference implementation.
We could implement it for every other storage service, too -- so JSON DB, but IMHO it could be done in a separate PR if somebody needs this feature for JSON DB. So I would like to postpone the discussion "should the JSON backup be kept or deleted, too" etc.

sjka:
Good point, I didn't think of the discussions the JSON DB could lead to...

How about the name then? I won't fight for it, but while in natural language "to dispose of sth" is pretty clear to end up in a trash can, at least for me it is not in relation to objects and databases. But maybe I'm the only one who fought disposed widgets for too long... 😉

maggu2810:
I don't care to much.
Isn't dispose similar to "entsorgen" and isn't the storage, so the saved data disposed?
But of course it is also deleted. So, I am fine with `DeletableStorage`, too.

bundles/core/org.eclipse.smarthome.core/src/main/java/org/eclipse/smarthome/core/storage/DisposableStorage.java
kaikreuzer: As we are talking about filesystem/database operations here, I think the "usual" vocabulary is DELETE, while DISPOSE is rather used in memory related cases. So I would vote to go for a `DeletableStorage`.

maggu2810:
* added DeletableStorageService
* implemented review comments

PR #3730, opened 2017-06-23
PaperUI: implement location widget

### LINKED ISSUE #3709: [Paper UI] Support "location" context for config parameters
According [to the docs](https://github.com/eclipse/smarthome/pull/3587/files#diff-8e67ab95e48e5025e43751c8d5255b9eR131) we now have a "location" context for config parameter.

For this, it is expected that the Paper UI renders a dedicated widget. We already have an OSM widget for Location items, so it would make sense to use the same here. It must be possible to select a location from the map, but also to still manually enter lat,long,alt fields as numbers in input fields.

htreu:
* OpenLayers & OpenStreetMap based map widget
* Edit configuration parameters with context 'location'
* Control location items
* 2-way data binding: update location configuration/items by dragging on the map
* Replaces OpenStreetMap <iframe> integration

Fixes #3709.


sjka:
Pretty cool, thanks! 

Two thoughts from trying it out:

1. the visual pin does not really "point" to the right position - at least the map jumps to a different location when I drag the pin somewhere
2. the pin really resembles an ice cream cone. I like the idea of using the ESH logo with a down arrow below, but maybe a different color for the arrow or a little padding between arrow and logo could help.

htreu:
We encountered the pin-drag-map-jump during development, too. This is due to the fact that after drag not the tip of the arrow is the location you set but the tip of your mouse pointer. Since you drag the pointer by taking the "middle" this will result in an offset. Its not so nice but we didn´t find a smooth way to overcome this issue.

regarding the logo-location-pin: maybe we can ask for help on this one? Its more a try-and-error approach for non designers.

htreu:
I will give the pin another try and present the image here. maybe different color and padding will already do.

htreu:
What about:

screen shot 2017-06-30 at 17 16 29

htreu:
Updated:

screen shot 2017-07-01 at 09 29 06

kaikreuzer:
Nice! I hope color and logo can also be customized through a fragment, right? ☺️

htreu:
Since its a PNG I guess the color is fixed for this one. But I will check on the fragment thing and provide a OH version.

kaikreuzer:
That's what I wanted to hear 👍

sjka:
cool, thanks for the icon update. Looks much better IMHO!
I'll create a CQ tomorrow for OpenLayers then.

extensions/ui/org.eclipse.smarthome.ui.paper/pom.xml
```diff
@@ -29,8 +29,8 @@
         <version>1.0</version>
 
         <configuration>
-          <nodeVersion>v5.4.0</nodeVersion>
-          <npmVersion>3.3.12</npmVersion>
+          <nodeVersion>v8.1.2</nodeVersion>
+          <npmVersion>5.0.3</npmVersion>
```
hakan42: We are upgrading the Node and npm versions... Do we need seperate CQs for this?

extensions/ui/org.eclipse.smarthome.ui.paper/web-src/partials/directive.component.map.html
```diff
@@ -0,0 +1,3 @@
+<div class="map-container">
+	<div id="map" class="map"></div>
+</div>
```
hakan42: Please add empty line at then end of the file.

extensions/ui/org.eclipse.smarthome.ui.paper/web-src/partials/directive.control.location.html
```diff
@@ -0,0 +1,21 @@
...
+	<div ng-show="isReadOnly(item)" layout="row" layout-align="end center">
+		<h2 class="state" ng-if="isReadOnly(item)">{{formattedState}}</h2>
+	</div>
+	<div ng-show="!isReadOnly(item)" layout="row" layout-align="end center">
+		<h2 class="state edit" ng-if="!editMode" ng-click="editState()">{{formattedState}}</h2>
+		<md-input-container ng-show="editMode"> <input ng-model="item.state" aria-label={{item.label}}></input> </md-input-container>
+		<i class="md-button control-button material-icons" ng-show="editMode" ng-click="updateState()">check</i>
+	</div>
+</div>
+<div layout="row" item-state-dropdown layout-align="end center" ng-if="isOptionList(item) && !isReadOnly(item)"></div>
```
hakan42: Please add empty line at then end of the file.

extensions/ui/org.eclipse.smarthome.ui.paper/web-src/js/directives/directive.component.map.js
```diff
@@ -0,0 +1,208 @@
...
+    Drag.prototype.handleUpEvent = function() {
+        this.callback_(this.coordinate_);
+        this.coordinate_ = null;
+        this.feature_ = null;
+        return false;
+    };
+
+    function createMap(element, coordinateCallback) {
+        var mapLayer = new ol.layer.Tile({
+            source : new ol.source.OSM()
```
hakan42: Could be a nice improvement if the base layer could be configurable.

Although I prefer OSM whereever possible 😄

htreu: This is indeed a very nice idea. For OH I can see that base layer providers may be exchangeable. For ESH we should stick with OSM. Do you mind creating an issue (feature request) for this?

extensions/ui/org.eclipse.smarthome.ui.paper/pom.xml
htreu: Good point: @sjka or @maggu2810 what about this version bump of node and npm? We need it here to provide a stable and customised build of the OpenLayers library. This reduces file size from ~500KB to 230KB.

extensions/ui/org.eclipse.smarthome.ui.paper/web-src/js/directives/directive.component.map.js
hakan42: Done, 

* https://github.com/eclipse/smarthome/issues/3793
* https://github.com/openhab/openhab-core/issues/163

extensions/ui/org.eclipse.smarthome.ui.paper/web-src/partials/directive.component.map.html
htreu: done.

extensions/ui/org.eclipse.smarthome.ui.paper/web-src/partials/directive.control.location.html
htreu: done. There were other source files from this PR w/o line break at the last line. updated.

extensions/ui/org.eclipse.smarthome.ui.paper/pom.xml
sjka: I haven't had a look at the sources yet, so thanks for bringing this up that early.
Yes, we do need approval for those too 😭

hakan42: At least we don't need an approval to ask for an approval... 😭

sjka:
CQ 13827 for openlayers 4.2.0 submitted

extensions/ui/org.eclipse.smarthome.ui.paper/pom.xml
sjka: Just realized that these are build-time only and that there apparently no specific versions have to be given approval for.

htreu:
For licensing issues this PR will change its focus:
The framework will only provide the infrastructure to render maps for location display and inputs but not implement a specific map provider. Instead the implementing entity (like openHAB, qivicon, ...) must provide a fragment with a map provider snippet following the [OpenLayers](http://openlayers.org/) API. Based on this PR I will provide an OpenStreetMap implementation in openHAB in case the OSM system admins are okay with openHAB to provide OSM maps.

kaikreuzer:
Good news, the CQ has been approved!

Issue #3732, opened 2017-06-23
fsinternetradio test: Migrate from Groovy to pure Java


maggu2810:
I am working on this topic.
So, no need to change the Groovy tests at the moment.

PR #3733, opened 2017-06-23
fsinternetradio tests: migrate vom Groovy to pure Java

maggu2810:
Fixes: #3732


extensions/binding/org.eclipse.smarthome.binding.fsinternetradio.test/.classpath
```diff
@@ -1,10 +1,8 @@
 <?xml version="1.0" encoding="UTF-8"?>
 <classpath>
-	<classpathentry kind="src" output="target/test-classes" path="src/test/groovy"/>
+	<classpathentry kind="src" path="src/test/java"/>
 	<classpathentry kind="con" path="org.eclipse.jdt.launching.JRE_CONTAINER/org.eclipse.jdt.internal.debug.ui.launcher.StandardVMType/JavaSE-1.8"/>
 	<classpathentry kind="con" path="org.eclipse.pde.core.requiredPlugins"/>
-	<classpathentry exported="true" kind="con" path="GROOVY_SUPPORT"/>
-	<classpathentry exported="true" kind="con" path="GROOVY_DSL_SUPPORT"/>
 	<classpathentry kind="con" path="org.eclipse.m2e.MAVEN2_CLASSPATH_CONTAINER"/>
```
sjka: I know, it was there before - but just stumbled across while reading: as the project (correctly) does not have the m2e nature, isn't this here old garbage?

extensions/binding/org.eclipse.smarthome.binding.fsinternetradio.test/src/test/java/org/eclipse/smarthome/binding/fsinternetradio/test/FSInternetRadioDiscoveryParticipantOSGiTest.java
```diff
@@ -0,0 +1,196 @@
...
+
+    String RADIO_BINDING_ID = "fsinternetradio"; // taken from the binding.xml file
+    String RADIO_THING_TYPE_ID = "radio"; // taken from the thing-types.xml file
+    String DEFAULT_RADIO_THING_UID = String.format("%s:%s:%s", RADIO_BINDING_ID, RADIO_THING_TYPE_ID,
+            DEFAULT_RADIO_SERIAL_NUMBER);
+
+    @Before
+    public void setUp() {
+        discoveryParticipant = getService(UpnpDiscoveryParticipant.class, FSInternetRadioDiscoveryParticipant.class);
+        assertThat("The FSInternetRadioDiscoveryParticipant service cannot be found", discoveryParticipant,
```
sjka: I really don't like these string messages. They are completely agains the goal of the matchers which aim at producing readable code. To me, this reads like "assert That The FSInternetRadioDiscoveryParticipant service cannot be found" - which of course is the exact opposite of what should be asserted. Therefore IMHO we should drop them (everywhere - won't mention it again). wdyt?

extensions/binding/org.eclipse.smarthome.binding.fsinternetradio.test/src/test/java/org/eclipse/smarthome/binding/fsinternetradio/test/FSInternetRadioDiscoveryParticipantOSGiTest.java
```diff
@@ -0,0 +1,196 @@
...
+    }
+
+    /**
+     * Verify valid DiscoveryResult with FSInterntRadio device without base URL.
+     *
+     * @throws ValidationException
+     */
+    @Test
+    public void validDiscoveryResultIfWithoutBaseUrl() throws ValidationException {
+
```
sjka: empty line

extensions/binding/org.eclipse.smarthome.binding.fsinternetradio.test/src/test/java/org/eclipse/smarthome/binding/fsinternetradio/test/FSInternetRadioDiscoveryParticipantOSGiTest.java
```diff
@@ -0,0 +1,196 @@
...
+                result.getProperties().get(FSInternetRadioBindingConstants.PROPERTY_MANUFACTURER),
+                is(DEFAULT_RADIO_MANIFACTURER));
+        assertThat("A result with wrong model number was created for a FSInternetRadio device without base URL",
+                result.getProperties().get(FSInternetRadioBindingConstants.PROPERTY_MODEL),
+                is(DEFAULT_RADIO_MODEL_NUMBER));
+
+    }
+
+    private RemoteDevice createDefaultFSInternetRadioDevice(URL baseURL) throws ValidationException {
+
```
sjka: empty line

extensions/binding/org.eclipse.smarthome.binding.fsinternetradio.test/src/test/java/org/eclipse/smarthome/binding/fsinternetradio/test/FSInternetRadioHandlerOSGiTest.java
```diff
@@ -0,0 +1,969 @@
...
+
+    private void turnTheRadioOn(Thing radioThing) {
+        radioHandler.handleCommand(radioThing.getChannel(FSInternetRadioBindingConstants.CHANNEL_POWER).getUID(),
+                OnOffType.ON);
+
+        FrontierSiliconRadio radio;
+        try {
+            Field f = radioHandler.getClass().getDeclaredField("radio");
+            f.setAccessible(true);
+            radio = (FrontierSiliconRadio) f.get(radioHandler);
```
sjka: Okay here, of course - so this is more meant to be a mental note for us: maybe hassle-free private field access is worth a utility method in our base class?

extensions/binding/org.eclipse.smarthome.binding.fsinternetradio.test/src/test/java/org/eclipse/smarthome/binding/fsinternetradio/test/RadioServiceMock.java
```diff
@@ -0,0 +1,219 @@
...
+import org.eclipse.jetty.http.HttpStatus;
+import org.eclipse.smarthome.binding.fsinternetradio.internal.radio.FrontierSiliconRadioConstants;
+
+/**
+ * Radio service mock.
+ *
+ * @author Markus Rathgeb - Migrated from Groovy to pure Java, made more robust
+ *
+ */
+public class RadioServiceMock extends HttpServlet {
```
sjka: Maybe `RadioServiceDummy` in order to better distinguish it from Mockito-Mocks? Technically the name is correct, of course...

maggu2810:
Thanks for having a look at.
For the most comments I have to admit, that I first want to migrate the tests itself to pure Java.
I fixed the most visible stuff but go not into such detail. My intention has been, that the all improvement could be done later and this one should just drop Groovy.
I will comment inline.

extensions/binding/org.eclipse.smarthome.binding.fsinternetradio.test/.classpath
maggu2810: Could be removed IMHO, too.
I will remove it.

extensions/binding/org.eclipse.smarthome.binding.fsinternetradio.test/src/test/java/org/eclipse/smarthome/binding/fsinternetradio/test/FSInternetRadioDiscoveryParticipantOSGiTest.java
maggu2810: I agree, I don't think we need to add a custom string to the assertions at all.
If the assertion fails on a test, we need to look at the code to fix it.
And there is already "expected behavior" that is reported to not match.

extensions/binding/org.eclipse.smarthome.binding.fsinternetradio.test/src/test/java/org/eclipse/smarthome/binding/fsinternetradio/test/FSInternetRadioHandlerOSGiTest.java
maggu2810: I tried to remove all reflection used by the Groovy tests.
The only solution for this one has been to change the access level for the radio member to "no modifier" or protected.
If give me a go to change the access level we could drop that reflection.

extensions/binding/org.eclipse.smarthome.binding.fsinternetradio.test/src/test/java/org/eclipse/smarthome/binding/fsinternetradio/test/RadioServiceMock.java
maggu2810: Okay for me - I kept the old name, but will change it.

maggu2810:
@sjka All comments has been addressed.

sjka:
Nice, thanks!

PR #3736, opened 2017-06-23
[WIP] [WeMo Binding] Implement Dimmer Switch and update Readme for Coffe Ma…

hmerk:
…ker and Dimmer


extensions/binding/org.eclipse.smarthome.binding.wemo/src/main/java/org/eclipse/smarthome/binding/wemo/internal/http/WemoHttpCall.java
```diff
@@ -43,4 +43,34 @@ public static String executeCall(String wemoURL, String soapHeader, String conte
...
+    public static String getEventServiceXML(String wemoURL) {
+
+        try {
+
+            String eventServiceURL = wemoURL + "/eventservice.xml";
+
+            String wemoCallResponse = HttpUtil.executeUrl("GET", eventServiceURL, 2000);
+            return wemoCallResponse;
+
+        } catch (Exception e) {
```
maggu2810: I see, it has already be done above, but I don't think that catching `Exception` is a good thing.
Perhaps you could catch every `RuntimeExcpetion` and the checked exceptions that are really thrown.

extensions/binding/org.eclipse.smarthome.binding.wemo/src/main/java/org/eclipse/smarthome/binding/wemo/internal/http/WemoHttpCall.java
```diff
@@ -43,4 +43,34 @@ public static String executeCall(String wemoURL, String soapHeader, String conte
...
+    public static String getDeviceInfoServiceXML(String wemoURL) {
+
+        try {
+
+            String deviceInfoServiceURL = wemoURL + "/deviceinfoservice.xml";
+
+            String wemoCallResponse = HttpUtil.executeUrl("GET", deviceInfoServiceURL, 2000);
+            return wemoCallResponse;
+
+        } catch (Exception e) {
```
maggu2810: see above

sjka:
@hmerk did you see Markus' comments? Apart from these and a few excess empty lines in the handler (mostly at the beginning of methods) this would be good to go.

hmerk:
@sjka yes, saw the comment but did not find time for changes, especially, as we are trying to get more functionality into the Dimmer handler. I will change the title to WIP for now.

sjka:
Nah, all fine! I just wanted to make sure it didn't fell off the desk (I know how it is...). 
Do it when there is time, my question really was not intended to put any pressure on it.

hmerk:
No worries Simon, did not feel like putting preassure on me :-)

hmerk:
It seems I have messed up this PR with my latest commits, therefore I will close it, clean my workspace and submit a new PR .

PR #3739, opened 2017-06-24
improved readme and logging

kaikreuzer:

### DIFF
--- a/extensions/binding/org.eclipse.smarthome.binding.astro/src/main/java/org/eclipse/smarthome/binding/astro/discovery/AstroDiscoveryService.java
+++ b/extensions/binding/org.eclipse.smarthome.binding.astro/src/main/java/org/eclipse/smarthome/binding/astro/discovery/AstroDiscoveryService.java
@@ -62,10 +62,10 @@ public class AstroDiscoveryService extends AbstractDiscoveryService {
     @Override
     protected void startBackgroundDiscovery() {
         if (astroDiscoveryJob == null) {
-            astroDiscoveryJob = scheduler.scheduleAtFixedRate(() -> {
+            astroDiscoveryJob = scheduler.scheduleWithFixedDelay(() -> {
                 PointType currentLocation = locationProvider.getLocation();
                 if (!Objects.equals(currentLocation, previousLocation)) {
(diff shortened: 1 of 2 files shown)


extensions/binding/org.eclipse.smarthome.binding.astro/src/main/java/org/eclipse/smarthome/binding/astro/discovery/AstroDiscoveryService.java
```diff
@@ -62,10 +62,10 @@ protected void startScan() {
     @Override
     protected void startBackgroundDiscovery() {
         if (astroDiscoveryJob == null) {
-            astroDiscoveryJob = scheduler.scheduleAtFixedRate(() -> {
+            astroDiscoveryJob = scheduler.scheduleWithFixedDelay(() -> {
```
amitjoy: According to this scheduled work, it is expected to have a delay of 60 seconds. As far as I know, the primary difference between `scheduleAtFixedRate(...)` and `scheduleWithFixedDelay(..)` is that, in `scheduleAtFixedRate(...)` the scheduled work can finish within the period specified (60s) whereas in `scheduleWithFixedDelay(..)`, there will be 60s delay after the work gets finished.

In simple words, if the work takes `t` amount of time, the total time required to execute the work again is `t + 60s` whereas in `scheduleAtFixedRate(...)`, the work can gets finished within the 60s period. I am not sure which one would be appropriate for this scenario.

kaikreuzer: > I am not sure which one would be appropriate for this scenario.

But I am sure, just check [D.2 of our guidelines](https://www.eclipse.org/smarthome/documentation/development/guidelines.html#d-runtime-behavior) 😎

amitjoy: 🤣 
Thanks for sharing the info.

Issue #3484, opened 2017-05-18
Introduce a differentiation between handler and device configuration


kaikreuzer:
Currently, the config descriptions on Things are meant to define what configuration a handler requires in order to access and use a device.
It is a common misconception that those configurations should represent the configuration of a remote device itself.

Device configuration management is clearly a need for many (not externally managed) device bindings, though.

This issue is meant as a follow-up on the discussion at https://www.eclipse.org/forums/index.php?t=msg&th=1079343 and https://github.com/eclipse/smarthome/pull/3245.

kgoderis:
+1 and for completeness : https://github.com/eclipse/smarthome/issues/1977

Now, @kaikreuzer, how would you treat those pieces of information that are part of the configuration of the remote device, *and* that are required by the handle in order to access and use a device? If duplication is a no-no in these circumstances, that would mean you need a system where you store everything together, but maybe tagged for specific purposes. (Well, we have Configuration and Properties, of course)

So, in essence we have different categories to consider, right?:
 - configuration of the handler
 - configuration of the remote device
 - properties of the remote device

htreu:
Hi all, I will pick this up and move the discussion forward. Since my head works best visually my approach is *specification by example*. Please try to come up with concrete issues for your requirements.

From the forum threads I learned the following structural requirements:

#### Handler Configuration
Handler configuration is merely about configuration of communication parameters, like ip address/host & port, timeouts, access tokens, thing ids, etc. While some of the parameters are directly found on the thing/bridge itself (like ip address, thing ids) their only purpose on the thing handler configuration is to help it reach the specific device. Changes to the handler configuration will not be propagated to the thing itself. It would be the other way round: If a communication parameter of a thing/bridge changes, the corresponding parameter in the handler configuration has to be adopted.
The current config description model with its single level hierarchy, non-list-support is sufficient for this purpose.

#### Thing Configuration
Thing configuration should describe and model the configuration of the physical device. It should not only provide a view (read-only) but provide an interface to alter the thing configuration. It is the bindings responsibility to implement the read/write access to the corresponding thing.
The most simple idea is to make an addition to the existing config description: a boolean flag ´thing-config=true` might tag a parameter to be used as a thing config parameter in contrast to the handler configuration. This would make the config description completely reusable with minimal effort in the model and the UI.
My examples for thing configuration entries:
* HomeMatic: secure communication
* radio power for wireless devices
* reboot, reset, power-cycle (more like actions, but device specific. better on items?)
[...]

PR #3740, opened 2017-06-25
Sonos binding: add new channel clearqueue

### LINKED ISSUE #2894: [sonos] Add support for clear player queue
The current sonos binding does not provide a channel to clear the current queue of a player. The playlist channel only allows to add tracks from a playlist to a player's queue. Thus, when using the playlist channel multiple times on the same player with different playlists, the queue will represent a mix of the playlists added.

In many use cases, a user would prefer to clear the queue so to only play tracks from the selected playlist.

My proposal is to implement a new channel to enable the user to clear the queue of a player. Keeping the playlist and the proposed clearqueu channel separated, will allow the user to choose the behaviour when adding tracks to the queue.

lolodomo:
Fix #2894

### DIFF
--- a/extensions/binding/org.eclipse.smarthome.binding.sonos/ESH-INF/thing/CONNECT.xml
+++ b/extensions/binding/org.eclipse.smarthome.binding.sonos/ESH-INF/thing/CONNECT.xml
@@ -29,6 +29,7 @@
 			<channel id="notificationsound" typeId="notificationsound" />
 			<channel id="notificationvolume" typeId="notificationvolume" />
 			<channel id="playlist" typeId="playlist" />
+			<channel id="clearqueue" typeId="clearqueue" />
 			<channel id="playqueue" typeId="playqueue" />
 			<channel id="playtrack" typeId="playtrack" />
 			<channel id="playuri" typeId="playuri" />
@@ -48,12 +49,12 @@
 			<channel id="zonegroupid" typeId="zonegroupid" />
 			<channel id="zonename" typeId="zonename" />
 			<channel id="coordinator" typeId="coordinator" />
-			<!-- Extended SONOS channels (PLAY5, CONNECT & CONNECT:AMP only) -->
-			<channel id="linein" typeId="linein" />
-			<channel id="playlinein" typeId="playlinein" />
 			<channel id="sleeptimer" typeId="sleeptimer" />
 			<channel id="currenttransporturi" typeId="currenttransporturi" />
 			<channel id="currenttrackuri" typeId="currenttrackuri" />
+			<!-- Extended SONOS channels -->
+			<channel id="linein" typeId="linein" />
+			<channel id="playlinein" typeId="playlinein" />
 		</channels>
 
 		<properties>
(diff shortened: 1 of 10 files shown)


lolodomo:
I kept the same philosophy as for the other switch channels in the binding like for example the stop channel, we don't care the switch state.
But I agree that could be something to change for all channels of this kind.

sjka:
At least it is consistent then 😄  Makes sense imho. 

Still I would like to understand whether we are missing something conceptually in the framework, or this binding is modeled in a creative way.

Do channels like this indicate a need for "write-only" channels which only accept commands but do not have a state (similar as trigger channels, just that the "event" is going the other way)? How and if to introduce such a channel kind is of course by far beyond this PR and should not be discussed here - I only want to understand this here and I would be fine with it how it is implemented currently.

Or is it rather that such a channel actually should be the representation of the device being in "cleanup" state for even a very short-lived moment? Then I'd agree it's weird that it remains in the ON state and should switch off again (consistently, i.e. for all these channels, best in a separate PR again).

maggu2810:
> Do channels like this indicate a need for "write-only" channels which only accept commands but do not have a state (similar as trigger channels, just that the "event" is going the other way)? 

Isn't this already supported by the framework.
Configure the autoupdate to not be used for that channel, don't set a state in the binding itself and handle the command only. Sure, the state will be NULL all the time, but isn't that an indication that the state is not available at the moment?

lolodomo:
I confirm it is a write only command channel without state representing device state.

lolodomo:
In the Sonos binding you have at least 10 channels of this kind. Sometimes you have another channel that reflects the state. For example, the control channel (player) is indirectly updated when using the stop channel as command.
There is no channel in the binding representing the state of the device playing queue.

lolodomo:
So what is the status for this PR ?

Maybe a new bug should be declared if you think the Sonos binding have unexpected handling of channels ?

maggu2810:
@sjka What do you think about.
Isn't a channel that does never receive a state update already similar to a "write-only" channel?
Should we create a separate PR to use that approach (we need to stop the auto update for that channels) for some Sonos channels?
Can we continue with that PR as it uses the same mechanism already used by other channels of the binding (I don't know, but IIRC @lolodomo stated that above)

sjka:
As I understand: 

> I only want to understand this here and I would be fine with it how it is implemented currently.

And this holds true. No need to change anything here, I really did not intend to block this PR with it. And I'm not even sure whether we need such a dedicated write-only channel kind. Need to make up my mind. And of course, we can discuss it elsewhere. 

This PR LGTM!

lolodomo:
Can this PR be merged please ? I am waiting for it to implement #3741 that will require changes in the channels file too.

sjka:
Oh sorry, yes, sure.
Thanks!

kaikreuzer:
> And I'm not even sure whether we need such a dedicated write-only channel kind. Need to make up my mind. And of course, we can discuss it elsewhere.

Just as some random thoughts, before I forget to write them down:
- I agree that this PR is in line with how it is done in the rest of the Sonos binding, so all fine.
- In general, I think it is the wrong approach: We should rather have a single channel that accepts such kinds of commands (as the split into separate channels should usually be done by checking, which "states" are handled independently of each other - so as we do not have any states here, there is no need for a channel for each of them).
- Currently, many developers model it as switches, because it is the easiest way to make the functionality available through UIs - this is a sign that we are lacking features regarding UI integration of other type than switched.
- We should have something like "commandOptions", similar to the already existing "stateOptions". This meta-data could be provided to UIs so that they could render buttons for such features (similar as the "mapping" functionality of Switch widgets in sitemaps).
- Administrative features (so nothing that the real end user wants to use on a daily basis) might also rather be modelled as a [device management](https://github.com/eclipse/smarthome/issues/3484) feature, once this is available.

Just my 2 cents for the discussion that might or might not take place "elsewhere" ;-)

Issue #3728, opened 2017-06-23
astro binding missing events


dvehof:
I am using the following openhab2 SNAPSHOT:
```
Package: openhab2
Versions: 
2.1.0~20170623003852-1 (/var/lib/apt/lists/openhab.jfrog.io_openhab_openhab-linuxpkg_dists_unstable_main_binary-armhf_Packages)
 Description Language: 
                 File: /var/lib/apt/lists/openhab.jfrog.io_openhab_openhab-linuxpkg_dists_unstable_main_binary-armhf_Packages
                  MD5: 326ece3868ef747762229ee66dc0da02

openhab> bundle:list | grep -i astro
177 | Active   |  80 | 0.9.0.201706211517     | Astro Binding
openhab> feature:list | grep -i astro
openhab-binding-astro1                      | 1.10.0.SNAPSHOT  |          | Uninstalled | openhab-addons-legacy-2.1.0-SNAPSHOT | Astro Binding (1.x)
esh-binding-astro                           | 0.9.0.SNAPSHOT   |          | Started     | addons-2.1.0-SNAPSHOT                |
openhab-binding-astro                       | 2.1.0.SNAPSHOT   | x        | Started     | addons-2.1.0-SNAPSHOT                | Astro Binding
```

Sometimes my rules, based on astro events, are not fired. As I look into my events.log I am missing some events of the astro binding. Examples (These are ALL events of my astro:sun:rolladen Thing in the events log in that time:

```
2017-06-21 22:21:00.036 [ChannelTriggeredEvent     ] - astro:sun:rolladen:set#event triggered START
2017-06-21 22:25:00.023 [ChannelTriggeredEvent     ] - astro:sun:rolladen:set#event triggered END
2017-06-21 22:44:00.042 [ChannelTriggeredEvent     ] - astro:sun:rolladen:civilDusk#event triggered END
2017-06-22 08:00:00.047 [ChannelTriggeredEvent     ] - astro:sun:rolladen:rise#event triggered END
2017-06-22 22:26:00.044 [ChannelTriggeredEvent     ] - astro:sun:rolladen:set#event triggered END
2017-06-22 22:44:00.054 [ChannelTriggeredEvent     ] - astro:sun:rolladen:civilDusk#event triggered END
2017-06-23 08:00:00.037 [ChannelTriggeredEvent     ] - astro:sun:rolladen:rise#event triggered START
```

As you can see, sometimes the START events are missing, sometimes the END events are missing. 
[...]

PR #3746, opened 2017-06-26
Synchronous Execution of Astro Startup Jobs and minor fixes

amitjoy:
The scheduling operations schedules few jobs and executes 2 specific jobs immediately. These two jobs are executed in a different thread to prevent blocking the thread that spawns it. 

In addition, during handler initialize, channel linking and unlinking, we reschedule the jobs. And every time, before we reschedule the jobs, we remove the jobs from the ESH Core scheduler that are associated with the current thing.

As the jobs that execute immediately are handled by different threads, it is sometime noticed that while rescheduling, removal of jobs did occur after the jobs got rescheduled. This created unexpected issues in ESH Core scheduler. In this PR, we have rendered the asynchronous operations of executing immediate jobs in different threads to synchronous execution so that we can make sure the removal always happens before rescheduling.

Another note is that previously we required `equals` and `hashcode` to be overriden by `AbstractJob` but currently it is not used anymore. These `equals` and `hashcode` affected normal executions of scheduled jobs as well. Hence, the methods have been removed.

### DIFF
--- a/extensions/binding/org.eclipse.smarthome.binding.astro/src/main/java/org/eclipse/smarthome/binding/astro/handler/AstroThingHandler.java
+++ b/extensions/binding/org.eclipse.smarthome.binding.astro/src/main/java/org/eclipse/smarthome/binding/astro/handler/AstroThingHandler.java
@@ -21,8 +21,6 @@ import java.util.Date;
 import java.util.List;
 import java.util.Objects;
 import java.util.Queue;
-import java.util.concurrent.CopyOnWriteArrayList;
-import java.util.concurrent.Future;
 import java.util.concurrent.LinkedBlockingQueue;
 import java.util.concurrent.locks.Lock;
 import java.util.concurrent.locks.ReentrantLock;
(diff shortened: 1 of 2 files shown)


kaikreuzer:
Could you add a short description to the PR about what it does and fixes exactly?

amitjoy:
@kaikreuzer Sure

sjka:
Thanks for this fix! 

I can confirm that the equals/hashcode implementation indeed was messing up the internal calculations of the next execution time in the scheduler.

Doing the initial job scheduling synchronously also sounds like a pretty good idea to me.

@amitjoy could you please fix the commit message and additionally include a sign-off-by line?

sjka:
PS: fixes #3728
(not to self: auto-close won't work...)

amitjoy:
@sjka 😢 How could I forgot it? Fixing it.

sjka:
All good, that's what we have all this auto-validation stuff for 😄

PR #1589, opened 2016-06-01
Added missing import for javax.xml.bind in manifest file

satish-nikam:
Lifx binding uses class javax.xml.bind.DatatypeConverter, but does not import javax.xml.bind in manifest xml. This causes issue when running ESH in karaf (using project https://github.com/maggu2810/shk).

Signed-off-by: Satish Nikam <satniks@gmail.com>

PR #3747, opened 2017-06-26
Added methods to find broadcastAddress(es), useful in bindings for discovery.

mherwege:
Signed-off-by: Mark Herwege <mark.herwege@telenet.be>

### DIFF
--- a/bundles/core/org.eclipse.smarthome.core/src/main/java/org/eclipse/smarthome/core/net/NetUtil.java
+++ b/bundles/core/org.eclipse.smarthome.core/src/main/java/org/eclipse/smarthome/core/net/NetUtil.java
@@ -7,11 +7,15 @@
  */
 package org.eclipse.smarthome.core.net;
 
+import java.net.Inet4Address;
 import java.net.Inet6Address;
 import java.net.InetAddress;
+import java.net.InterfaceAddress;
 import java.net.NetworkInterface;
 import java.net.SocketException;
+import java.util.ArrayList;
 import java.util.Enumeration;
+import java.util.List;
 
 import org.slf4j.Logger;
 import org.slf4j.LoggerFactory;
@@ -20,6 +24,7 @@ import org.slf4j.LoggerFactory;
  * Some utility functions related to network interfaces etc.
  *
  * @author Markus Rathgeb - Initial contribution and API
+ * @author Mark Herwege - Added methods to find broadcast address(es)
  */
 public class NetUtil {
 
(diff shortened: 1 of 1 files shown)


mherwege:
As suggested by @kaikreuzer in review of openHAB Niko Home Control binding [#1589](https://github.com/openhab/openhab2-addons/pull/1589).

bundles/core/org.eclipse.smarthome.core/src/main/java/org/eclipse/smarthome/core/net/NetUtil.java
```diff
@@ -55,9 +60,53 @@ public static String getLocalIpv4HostAddress() {
...
     }
 
+    /**
+     * Get all broadcast addresses on the current host
+     *
+     * @return list of broadcast addresses
+     */
+    public static List<String> getAllBroadcastAddresses() {
+        try {
+            List<String> broadcastAddresses = new ArrayList<String>();
```
maggu2810: Isn't a `LinkedList` better if we don't know the size (to reduce reallocations and copies)?

bundles/core/org.eclipse.smarthome.core/src/main/java/org/eclipse/smarthome/core/net/NetUtil.java
```diff
@@ -55,9 +60,53 @@ public static String getLocalIpv4HostAddress() {
             }
             return hostAddress;
         } catch (SocketException ex) {
-            LOGGER.error("Could not retrieve network interface: " + ex.getMessage(), ex);
+            LOGGER.error("Could not retrieve network interface: {}", ex.getMessage(), ex);
```
maggu2810: Good catch

bundles/core/org.eclipse.smarthome.core/src/main/java/org/eclipse/smarthome/core/net/NetUtil.java
```diff
@@ -55,9 +60,53 @@ public static String getLocalIpv4HostAddress() {
...
             return null;
         }
     }
 
+    /**
+     * Get all broadcast addresses on the current host
+     *
+     * @return list of broadcast addresses
```
maggu2810: If the `SocketException` is thrown you return `null`.
So, this should be stated in the JavaDoc.
But wouldn't it make sense to throw the exception, so the caller is notified about the error -- or if the caller cannot handle it at all (does he need to know about the error) return an empty list?

bundles/core/org.eclipse.smarthome.core/src/main/java/org/eclipse/smarthome/core/net/NetUtil.java
```diff
@@ -55,9 +60,53 @@ public static String getLocalIpv4HostAddress() {
...
+     * Get the first candidate for a broadcast address
+     *
+     * @return broadcast address
+     */
+    public static String getBroadcastAddress() {
+        String broadcastAddress = null;
+        final List<String> broadcastAddresses = getAllBroadcastAddresses();
+        for (String current : broadcastAddresses) {
+            if (broadcastAddress != null) {
+                LOGGER.warn("Found multiple broadcast addresses - ignoring {}", broadcastAddress);
```
maggu2810: The JavaDoc state "get the first candidate", so I don't see any reason to throw a warning if there are multiple candidates. The function should return the first one and not check if there is more then one, or should it?

bundles/core/org.eclipse.smarthome.core/src/main/java/org/eclipse/smarthome/core/net/NetUtil.java
mherwege: Returning an empty list, and documented.

bundles/core/org.eclipse.smarthome.core/src/main/java/org/eclipse/smarthome/core/net/NetUtil.java
mherwege: Replaced with a LinkedList.

bundles/core/org.eclipse.smarthome.core/src/main/java/org/eclipse/smarthome/core/net/NetUtil.java
mherwege: I used the same structure as in the `getLocalIpv4HostAddress()` method. It does not add value, so I removed it and just return the first one, or null if there is none. Null return is documented.

bundles/core/org.eclipse.smarthome.core/src/main/java/org/eclipse/smarthome/core/net/NetUtil.java
```diff
@@ -55,7 +60,46 @@ public static String getLocalIpv4HostAddress() {
...
+        }
+    }
+
+    /**
+     * Get all broadcast addresses on the current host
+     *
+     * @return list of broadcast addresses, empty list if no broadcast addresses found
+     */
+    public static List<String> getAllBroadcastAddresses() {
+        List<String> broadcastAddresses = new LinkedList<String>();
```
maggu2810: Hm, shouldn't the formatter replace `new LinkedList<String>()` be replaced by an `Diamond Operator`?

maggu2810: Hm, okay, seems it is not activated by the ESH formatter.

maggu2810:
core and core tests succeeded, the error is not related

PR #3748, opened 2017-06-26
Classic UI: escape HTML characters

### LINKED ISSUE #3743: Classic UI: bad display when value contains XML
This can be checked for example with the zonegroup channel of the Sonos binding.
Some character escaping is probably missing.

lolodomo:
Fix #3743

### DIFF
--- a/extensions/ui/org.eclipse.smarthome.ui.classic/snippets/layer.html
+++ b/extensions/ui/org.eclipse.smarthome.ui.classic/snippets/layer.html
@@ -1,5 +1,5 @@
 <root>
-        <title set="wa%id%" style="%labelstyle%">%label%</title>
+        <title set="wa%id%" style="%labelstyle%"><![CDATA[%label%]]></title>
         <part>
                 <destination mode="replace" zone="wa%id%" create="true" />
                 <data><![CDATA[
--- a/extensions/ui/org.eclipse.smarthome.ui.classic/src/main/java/org/eclipse/smarthome/ui/classic/internal/render/AbstractWidgetRenderer.java
+++ b/extensions/ui/org.eclipse.smarthome.ui.classic/src/main/java/org/eclipse/smarthome/ui/classic/internal/render/AbstractWidgetRenderer.java
@@ -8,13 +8,14 @@
 package org.eclipse.smarthome.ui.classic.internal.render;
 
 import java.io.IOException;
-import java.net.URI;
-import java.net.URISyntaxException;
+import java.io.UnsupportedEncodingException;
 import java.net.URL;
+import java.net.URLEncoder;
 import java.util.HashMap;
 import java.util.Map;
 
 import org.apache.commons.io.IOUtils;
+import org.apache.commons.lang.StringEscapeUtils;
 import org.apache.commons.lang.StringUtils;
 import org.eclipse.smarthome.core.items.Item;
 import org.eclipse.smarthome.core.items.ItemNotFoundException;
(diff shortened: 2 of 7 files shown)


lolodomo:
For information, I was not able to fix the particular case of the selection element: when you open the sub-page to select amongs a list, if your label is defined as "TV <TV-ok>" in the sitemap mapping, only "TV" will be displayed.

lolodomo:
As a side note, I replaced the code in escapeURLPath by the code from Basic UI. Using `new URI(null, null, path, null).toString();`was failing with the exception in cases when `URLEncoder.encode(path, "UTF-8");`is working. Of course, icons are still displayed correctly.

lolodomo:
@resetnow : could you please review this PR ? The changes are more or less the same changes you already reviewed for Basic UI.

lolodomo:
How do you enter a console command when you start OH from Eclipse ?

An alternative to test your case would be to define a rule.

resetnow:
I think console for input is already allocated in the default run configuration, just paste the command where the output is.

lolodomo:
I was not aware that it was possible ! But it works.I can confirm your test case.

lolodomo:
@resetnow : what's wrong with that HTML code: `<title set="wa0201" style="">Message test &lt;i&gt;test&lt;/i&gt;</title>`
The text betwen <i> and </i> appears in italic

lolodomo:
Same problem with an hardcoded label in sitemap like this one:
```
Text label="Group <i>Demo</i>" icon="firstfloor" {
}
```

lolodomo:
It seems to be exactly the same problem I found in the selection list and I was not able to fix !
I hope you can tell me what's wrong with this HTML code and what to do to fix it.

9037568:
I'm curious why you'd ask something to be rendered in italics and then be surprised when it's rendered in italics?

lolodomo:
We just want to display the value of an item that can contain anything without any interpretation by the HTML browser. Italic was just an example but the problem is in fact with any XML tag.
I don"t understand why escaping is working everywhere except in the title tag. Probably something obvious but I am not a confirmed HTML developer.

9037568:
[Maybe this](https://stackoverflow.com/questions/3848664/prevent-browse-to-render-html-when-using-htmlentities-function-in-php) has a solution?

lolodomo:
I will have to try how Basic UI handles that case. As the escaping code is the same in Basic UI and Classic UI, U would not understand why it us correct in Basic UI and not in Classic UI.

resetnow:
One possible case is that the data transferred in a long-lining request are escaped only once. Since long-polling response is an XML document itself and "title" element doesn't have CDATA escaping, all the HTML entities are transformed back into their respective symbols once the document tree is parsed in broswer. https://github.com/eclipse/smarthome/blob/067bd5122b4666452bda319cb4e1de13eda9ab00/extensions/ui/org.eclipse.smarthome.ui.classic/snippets/layer.html#L216:56, July 4, 2017, lolodomo <notifications@github.com>:I will have to try how Basic UI handles that case. As the escaping code is the same in Basic UI and Classic UI, U would not understand why it us correct in Basic UI and not in Classic UI. —You are receiving this because you were mentioned.Reply to this email directly, view it on GitHub, or mute the thread.     --Sent from Yandex.Mail for mobile

lolodomo:
I just checked the layer.html in Basic UI and I can find this:
`<title><![CDATA[ %label% ]]></title>`
while in Classic UI: 
`<title set="wa%id%" style="%labelstyle%">%label%</title>`
Il will try to add CDATA...

lolodomo:
Problem with title is now fixed.

sjka:
Thank you all!

PR #3749, opened 2017-06-26
Basic UI: escape HTML characters

### LINKED ISSUE #3744: Basic UI: bad refresh of value containing XML
The value containing XML seems to be correctly displayed, even if truncated, when the page is first opened but when an update is sent the dispayed value rhen looks like an empty string.
Can be checked with the zonegroup channel of the Sonos binding

lolodomo:
Fix #3744


lolodomo:
@resetnow : this one is for you ,)
All my tests are positive.

extensions/ui/org.eclipse.smarthome.ui.basic/src/main/java/org/eclipse/smarthome/ui/basic/internal/render/SwitchRenderer.java
```diff
@@ -82,7 +82,8 @@ public boolean canRender(Widget w) {
                 String button = getSnippet("button");
                 button = StringUtils.replace(button, "%item%", w.getItem());
                 button = StringUtils.replace(button, "%cmd%", mapping.getCmd());
```
resetnow: I think we need escape here as well. Test case:

```
Switch item=DemoString mappings=["value__1"="description_1", "value_\"_2"="description_2"]
```

resetnow: Also, using `StringEscapeUtils.escapeHTML` here is probably not the right choice — only quotes and backslashes need to be prepended with `\` so resulting HTML attribute has correct value when read from Javascript.

lolodomo: I thought about mappings from number or ON/OFF/OPEN/CLOSE... for example but not from a general string.
You're right, it has to be tested and fixed for general string commands too.

lolodomo:
Fixed.

sjka:
Thanks!

PR #3750, opened 2017-06-27
named astro jobs

sjka:
...so they are better distinguishable, e.g. in scheduler logs.

Also, hiding the thingUID from subclasses.

### DIFF
--- a/extensions/binding/org.eclipse.smarthome.binding.astro/src/main/java/org/eclipse/smarthome/binding/astro/internal/job/AbstractJob.java
+++ b/extensions/binding/org.eclipse.smarthome.binding.astro/src/main/java/org/eclipse/smarthome/binding/astro/internal/job/AbstractJob.java
@@ -14,7 +14,12 @@ package org.eclipse.smarthome.binding.astro.internal.job;
  */
 public abstract class AbstractJob implements Job {
 
-    protected String thingUID;
+    private final String thingUID;
+
+    public AbstractJob(String thingUID) {
+        checkArgument(thingUID != null, "The thingUID must not be null");
+        this.thingUID = thingUID;
+    }
 
     @Override
     public String getThingUID() {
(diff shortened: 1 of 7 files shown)


extensions/binding/org.eclipse.smarthome.binding.astro/src/main/java/org/eclipse/smarthome/binding/astro/internal/job/AbstractJob.java
```diff
@@ -14,7 +14,12 @@
  */
 public abstract class AbstractJob implements Job {
 
-    protected String thingUID;
+    private final String thingUID;
+
+    public AbstractJob(String thingUID) {
+        checkArgument(thingUID != null, "Thing UID cannot be null");
```
htreu: instead of `cannot be null` I would prefer `must not be null` in error messages which point the caller of this method to his failure. An instance returned by a method `cannot be null` if it will never be null. A method parameter can be null any time but `must not be null` to guarantee the correct behaviour of the method.

extensions/binding/org.eclipse.smarthome.binding.astro/src/main/java/org/eclipse/smarthome/binding/astro/internal/job/PublishPlanetJob.java
```diff
@@ -28,17 +28,21 @@
...
+        AstroThingHandler astroHandler = AstroHandlerFactory.getHandler(getThingUID());
         if (checkNull(astroHandler, "AstroThingHandler is null")) {
             return;
         }
         astroHandler.publishDailyInfo();
     }
 
+    @Override
+    public String toString() {
+        return "Publish plane job " + getThingUID();
```
htreu: typo: planet

PR #3751, opened 2017-06-27
various small improvements in the expression scheduler

sjka:
* increased some log statements to debug level
* additional trace logging
* using locks and conditions instead of interrupts
* wrapping given runnables so their hashcode/equals does no harm

### DIFF
--- a/bundles/core/org.eclipse.smarthome.core/src/main/java/org/eclipse/smarthome/core/scheduler/ExpressionThreadPoolManager.java
+++ b/bundles/core/org.eclipse.smarthome.core/src/main/java/org/eclipse/smarthome/core/scheduler/ExpressionThreadPoolManager.java
@@ -13,15 +13,20 @@ import java.util.Date;
 import java.util.HashMap;
 import java.util.List;
 import java.util.Map;
+import java.util.Map.Entry;
 import java.util.concurrent.CancellationException;
 import java.util.concurrent.ConcurrentHashMap;
 import java.util.concurrent.ExecutorService;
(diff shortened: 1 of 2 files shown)


bundles/core/org.eclipse.smarthome.core/src/main/java/org/eclipse/smarthome/core/scheduler/ExpressionThreadPoolManager.java
```diff
@@ -195,97 +213,130 @@ public void run() {
...
                             scheduled.remove(e);
+                            logger.trace("Cleaning up finished expression '{}'", e);
                         }
 
-                        if (firstExecution != null) {
-                            while (now.before(firstExecution)) {
+                        if (earliestExecution != null) {
+                            boolean newExpression = false;
+                            while (!newExpression && new Date().before(earliestExecution)) {
+                                long sleepDuration = earliestExecution.getTime() - now.getTime();
```
kaikreuzer: In the line above you check against an "updated now" - isn't the risk here that you slightly sleep too long if you use the "previous now" here?

sjka: Good point, there is - I will inline this calculation again

sjka: done

PR #3752, opened 2017-06-27
Added Composite Jobs for range events if scheduled at same instant

amitjoy:
The scheduler works on the unit of seconds. For range events, it can also schedule two events to be executed at the same instant. Due to parallel execution, there are inconsistency about the executions of the range events that are scheduled for same instants.

Therefore one second to the end event has been added to assure the synchronous execution of the range start and end events.

### DIFF
--- a/extensions/binding/org.eclipse.smarthome.binding.astro/src/main/java/org/eclipse/smarthome/binding/astro/internal/job/CompositeJob.java
+++ b/extensions/binding/org.eclipse.smarthome.binding.astro/src/main/java/org/eclipse/smarthome/binding/astro/internal/job/CompositeJob.java
@@ -1,0 +1,51 @@
+/**
+ * Copyright (c) 2014-2017 by the respective copyright holders.
+ * All rights reserved. This program and the accompanying materials
+ * are made available under the terms of the Eclipse Public License v1.0
+ * which accompanies this distribution, and is available at
+ * http://www.eclipse.org/legal/epl-v10.html
+ */
+package org.eclipse.smarthome.binding.astro.internal.job;
+
+import java.util.List;
+
+/**
+ * {@link CompositeJob} comprises multiple {@link Job}s to be executed in order
+ *
+ * @author Markus Rathgeb - Initial contribution
+ * @author Amit Kumar Mondal - Minor modifications
+ */
+public final class CompositeJob extends AbstractJob {
+
+    private final List<Job> jobs;
+
+    /**
+     * Constructor
+     * 
+     * @param thingUID thing UID
+     * @param jobs the jobs to execute
+     * @throws IllegalArgumentException
+     *             if {@code jobs} is {@code null} or empty
+     */
+    public CompositeJob(String thingUID, List<Job> jobs) {
+        super(thingUID);
+        checkArgument(jobs != null, "Jobs must not be null");
+        checkArgument(!jobs.isEmpty(), "Jobs must not be empty");
+
+        this.jobs = jobs;
+
+        boolean notMatched = jobs.stream().anyMatch(j -> !j.getThingUID().equals(thingUID));
+        checkArgument(!notMatched, "The jobs must associate the same thing UID");
+    }
+
+    @Override
+    public void run() {
+        jobs.forEach(j -> {
+            try {
+                j.run();
+            } catch (RuntimeException ex) {
+                logger.warn("Job execution failed.", ex);
+            }
+        });
+    }
+}
--- a/extensions/binding/org.eclipse.smarthome.binding.astro/src/main/java/org/eclipse/smarthome/binding/astro/internal/job/Job.java
+++ b/extensions/binding/org.eclipse.smarthome.binding.astro/src/main/java/org/eclipse/smarthome/binding/astro/internal/job/Job.java
@@ -7,12 +7,18 @@
  */
 package org.eclipse.smarthome.binding.astro.internal.job;
 
+import static java.util.Arrays.asList;
+import static java.util.Calendar.SECOND;
+import static java.util.Collections.singletonList;
 import static java.util.Objects.isNull;
+import static java.util.stream.Collectors.toList;
+import static org.apache.commons.lang.time.DateUtils.truncatedEquals;
 import static org.eclipse.smarthome.binding.astro.AstroBindingConstants.*;
 import static org.eclipse.smarthome.binding.astro.internal.util.DateTimeUtils.*;
 
 import java.lang.invoke.MethodHandles;
 import java.util.Calendar;
+import java.util.List;
 
 import org.eclipse.smarthome.binding.astro.handler.AstroThingHandler;
 import org.eclipse.smarthome.binding.astro.internal.config.AstroChannelConfig;
@@ -65,20 +71,34 @@ public interface Job extends Runnable {
      */
     public static void scheduleEvent(String thingUID, AstroThingHandler astroHandler, Calendar eventAt, String event,
             String channelId) {
+        scheduleEvent(thingUID, astroHandler, eventAt, singletonList(event), channelId);
+    }
+
+    /**
+     * Schedules an {@link EventJob} instance
+     *
+     * @param thingUID the Thing UID
+     * @param astroHandler the {@link ThingHandler} instance
+     * @param eventAt the {@link Calendar} instance denoting scheduled instant
+     * @param events the event IDs to schedule
+     * @param channelId the channel ID
+     */
+    public static void scheduleEvent(String thingUID, AstroThingHandler astroHandler, Calendar eventAt,
+            List<String> events, String channelId) {
         boolean thingNull = checkNull(thingUID, "Thing UID is null");
         boolean astroHandlerNull = checkNull(astroHandler, "AstroThingHandler is null");
         boolean eventAtNull = checkNull(eventAt, "Scheduled Instant is null");
-        boolean eventNull = checkNull(event, "Event is null");
+        boolean eventsNull = checkNull(events, "Events list is null");
         boolean channelIdNull = checkNull(channelId, "Channel ID is null");
 
-        if (thingNull || astroHandlerNull || eventAtNull || eventNull || channelIdNull) {
+        if (thingNull || astroHandlerNull || eventAtNull || eventsNull || channelIdNull || events.isEmpty()) {
             return;
         }
         AstroChannelConfig config = astroHandler.getThing().getChannel(channelId).getConfiguration()
                 .as(AstroChannelConfig.class);
         Calendar instant = applyConfig(eventAt, config);
-        Job eventJob = new EventJob(thingUID, channelId, event);
-        schedule(thingUID, astroHandler, eventJob, instant);
+        List<Job> jobs = events.stream().map(e -> new EventJob(thingUID, channelId, e)).collect(toList());
+        schedule(thingUID, astroHandler, new CompositeJob(thingUID, jobs), instant);
     }
 
     /**
@@ -98,8 +118,15 @@ public interface Job extends Runnable {
         if (thingNull || astroHandlerNull || rangeNull || channelIdNull) {
             return;
         }
-        scheduleEvent(thingUID, astroHandler, range.getStart(), EVENT_START, channelId);
-        scheduleEvent(thingUID, astroHandler, range.getEnd(), EVENT_END, channelId);
+        
+        Calendar start = range.getStart();
+        Calendar end = range.getEnd();
+        if (truncatedEquals(start, end, SECOND)) {
+            scheduleEvent(thingUID, astroHandler, start, asList(EVENT_START, EVENT_END), channelId);
+        } else {
+            scheduleEvent(thingUID, astroHandler, start, EVENT_START, channelId);
+            scheduleEvent(thingUID, astroHandler, end, EVENT_END, channelId);
+        }
     }
 
     /**


extensions/binding/org.eclipse.smarthome.binding.astro/src/main/java/org/eclipse/smarthome/binding/astro/internal/job/Job.java
```diff
@@ -111,8 +112,16 @@ public static void scheduleRange(String thingUID, AstroThingHandler astroHandler
         if (thingNull || astroHandlerNull || rangeNull || channelIdNull) {
             return;
         }
-        scheduleEvent(thingUID, astroHandler, range.getStart(), EVENT_START, channelId);
-        scheduleEvent(thingUID, astroHandler, range.getEnd(), EVENT_END, channelId);
+        
+        //add 1 second to the last scheduled event if both the events comprise same time instant
+        int secondsPadding = 1;
```
sjka: Maybe I missed something, but wouldn't it be much shorter when using DateUtils directly to compare those two dates?

```java
Calendar end = range.getEnd();
if (DateUtils.truncatedEquals(range.getStart(), end, SECOND) {
    end.add(SECOND, 1);
}
```

regarding the padding: either inline the value directly or declare a constant field for it.

extensions/binding/org.eclipse.smarthome.binding.astro/src/main/java/org/eclipse/smarthome/binding/astro/internal/job/Job.java
```diff
@@ -39,7 +40,7 @@
 
     /** Logger Instance */
     public final Logger logger = LoggerFactory.getLogger(MethodHandles.lookup().lookupClass());
-
+
```
sjka: These blanks just slipped in. 
Are you using the ESH IDE with the preset formatter? (Just asking to see whether it has a bug we need to fix...)

extensions/binding/org.eclipse.smarthome.binding.astro/src/main/java/org/eclipse/smarthome/binding/astro/internal/job/Job.java
amitjoy: I agree. I didn't know about the `DateUtils#truncatedEquals` method. It's more concise. will make the necessary modifications. For the constant, it is only used in one place, that's why I thought it's not a good idea to declare a constant at the interface.

amitjoy:
> This statement just made me think - the events are executed still asynchronously in different threads, this PR does not change that. And in a multi-threaded environment we simply cannot guarantee/assure the order in which scheduled tasks are executed. We just make it more likely to happen in the right order. So maybe we should be a little more generous with respect to the padding...? What do you think?

I agree it is still asynchronous. In case of range events, when START event and END event are scheduled at the same instant, the ordering of executions cannot be controlled explicitly. That's why, the ordering of the start and end event executions are controlled by delaying the end event by 1 more second. This is of course a hack to assure that the start and end events always occur one after another even though they are getting executed in different threads.

You are absolutely right. The previous 2 liner statement was kinda misleading. The events are still asynchronous. My opinion is that delaying the end event to 1 more second will not do any harm, I believe.

sjka:
We are on the same page here. My question was more about whether we should increase the padding to something bigger, like 5 or even 10 seconds?

amitjoy:
We could delay the end event to 5 or 10 more seconds but if the user needs the events to be scheduled at the same instant, then it is user specific requirement and to assure that START and END events occur one after another, we delay it to 1 sec which would have a minimal impact on what user needs. What do you think?

sjka:
You are right - 1 second _should_ work. And if it doesn't, let's agree we would need to fix it in the scheduler.

extensions/binding/org.eclipse.smarthome.binding.astro/src/main/java/org/eclipse/smarthome/binding/astro/internal/job/Job.java
sjka: okay, so could you just manually remove the blanks then please?

extensions/binding/org.eclipse.smarthome.binding.astro/src/main/java/org/eclipse/smarthome/binding/astro/internal/util/DateTimeUtils.java
```diff
@@ -167,7 +167,7 @@ public static boolean isTimeGreaterEquals(Calendar cal1, Calendar cal2) {
         Calendar truncCal2 = DateUtils.truncate(cal2, Calendar.MINUTE);
         return truncCal1.getTimeInMillis() >= truncCal2.getTimeInMillis();
     }
-
+
```
sjka: Here two please.

amitjoy:
@sjka Sure. If 1 doesn't work, we must fix it in scheduler. As you advised, I have amended my PR with the required changes.

maggu2810:
Wait a moment before merging, I will have a look at, too.

maggu2810:
@amitjoy Only a quick and dirty change for discussion: https://github.com/amitjoy/smarthome/compare/fix_range_events_delay...maggu2810:pr3752?expand=1

amitjoy:
@maggu2810 The changes are promising. How can we proceed? I can make these necessary changes and add you as co-commiter in the PR. What do you think?

maggu2810:
@amitjoy Merge / cherry-pick that changes into your branch or integrate it manually.
Make it looking good. :wink: 
Add another Signed-off-by or leave it. I don't care for that few lines of code.

amitjoy:
@maggu2810 I have updated the PR according to suggestions.

extensions/binding/org.eclipse.smarthome.binding.astro/src/main/java/org/eclipse/smarthome/binding/astro/internal/job/Job.java
```diff
@@ -111,8 +131,15 @@ public static void scheduleRange(String thingUID, AstroThingHandler astroHandler
...
             return;
         }
-        scheduleEvent(thingUID, astroHandler, range.getStart(), EVENT_START, channelId);
-        scheduleEvent(thingUID, astroHandler, range.getEnd(), EVENT_END, channelId);
+        
+        Calendar start = range.getStart();
+        Calendar end = range.getEnd();
+        if (truncatedEquals(start, end, SECOND)) {
+            scheduleEvent(thingUID, astroHandler, start, asList(EVENT_START, EVENT_END), channelId);
+            return;
```
sjka: please use an `if {...} else {...}` block here - the early return is of course correct, but always dangerous for the next one who needs to do some archeology here (or even change it).

amitjoy:
@sjka I have amended the PR.

extensions/binding/org.eclipse.smarthome.binding.astro/src/main/java/org/eclipse/smarthome/binding/astro/internal/job/CompositeJob.java
```diff
@@ -0,0 +1,47 @@
+/**
+ * Copyright (c) 2017 by the respective copyright holders.
```
maggu2810: see comment below

extensions/binding/org.eclipse.smarthome.binding.astro/src/main/java/org/eclipse/smarthome/binding/astro/internal/job/CompositeJob.java
```diff
@@ -0,0 +1,47 @@
...
+     * @param jobs the jobs to execute
+     * @throws IllegalArgumentException
+     *             if {@code jobs} is {@code null} or empty
+     */
+    public CompositeJob(List<Job> jobs) {
+        checkArgument(jobs != null, "Jobs must not be null");
+        checkArgument(!jobs.isEmpty(), "Jobs must not be empty");
+
+        this.jobs = jobs;
+        this.thingUID = jobs.get(0).getThingUID();
```
maggu2810: Why have you removed the check that every provided job is using the same thing UID?

extensions/binding/org.eclipse.smarthome.binding.astro/src/main/java/org/eclipse/smarthome/binding/astro/internal/job/Job.java
```diff
@@ -1,20 +1,26 @@
 /**
- * Copyright (c) 2014-2017 by the respective copyright holders.
+ * Copyright (c) 2017 by the respective copyright holders.
```
maggu2810: Why have you removed the `2014-`?
Isn't the range added by the licence maven plugin that we use to update the header and to ensure every file is using the correct one?

extensions/binding/org.eclipse.smarthome.binding.astro/src/main/java/org/eclipse/smarthome/binding/astro/internal/job/Job.java
```diff
@@ -66,7 +72,7 @@ public static void schedule(String thingUID, AstroThingHandler astroHandler, Job
             logger.error("{}", ex.getMessage(), ex);
         }
     }
-
+
```
maggu2810: This trailing whitespaces indicates that your formatter is not working.

extensions/binding/org.eclipse.smarthome.binding.astro/src/main/java/org/eclipse/smarthome/binding/astro/internal/job/Job.java
amitjoy: According to [1] - The range in license header denotes the start year of the content creation and the current year.

[1] - https://eclipse.org/legal/copyrightandlicensenotice.php

extensions/binding/org.eclipse.smarthome.binding.astro/src/main/java/org/eclipse/smarthome/binding/astro/internal/job/Job.java
amitjoy: 5-6 days back I have set the SmartHome environment using Eclipse Oxygen RC release using Oomph Profile. I am using the current formatter. I will try to manually remove these lines.

extensions/binding/org.eclipse.smarthome.binding.astro/src/main/java/org/eclipse/smarthome/binding/astro/internal/job/CompositeJob.java
amitjoy: My bad. I might have overlooked it.

extensions/binding/org.eclipse.smarthome.binding.astro/src/main/java/org/eclipse/smarthome/binding/astro/internal/job/Job.java
maggu2810: If we would like to follow that rule we need to change the project specific `Code Style` guidelines and update all other source files.
IMHO this could be discussed in a separate issue.

As long as the `Code Style` for that project uses `mvn license:format` to keep the license header up to date we should follow that rule. Shouldn't we?

https://www.eclipse.org/smarthome/documentation/development/guidelines.html#a-code-style

amitjoy: You are right. I will amend this PR.

extensions/binding/org.eclipse.smarthome.binding.astro/src/main/java/org/eclipse/smarthome/binding/astro/internal/job/Job.java
maggu2810: I could be wrong -- but I am pretty sure, the formatting rules and the save actions contain the removal of trailing white spaces.
Please check it.
I re-installed the IDE using the "latest release", so Oxygen RC yesterday, too.
I hope it will not contain some regressions about white space handling or save actions.

amitjoy:
@maggu2810 I have amended the PR according your suggestions.

extensions/binding/org.eclipse.smarthome.binding.astro/src/main/java/org/eclipse/smarthome/binding/astro/internal/job/CompositeJob.java
```diff
@@ -0,0 +1,50 @@
...
+
+        boolean notMatched = jobs.stream().noneMatch(j -> j.getThingUID().equals(thingUID));
```
maggu2810: Miss I something?

Don't we need to ensure that `allMatch == true`.
None match will only be true if no one matches.

amitjoy: Ah. definitely. It should better off with anyMatch -> if any of the elements doesn't have associated thing UID that matches the expected thing UID, it must throw IAE.

extensions/binding/org.eclipse.smarthome.binding.astro/src/main/java/org/eclipse/smarthome/binding/astro/internal/job/CompositeJob.java
```diff
@@ -0,0 +1,50 @@
...
+    }
+
+    @Override
+    public String getThingUID() {
+        return thingUID;
+    }
+
+    @Override
+    public void run() {
+        jobs.forEach(Job::run);
```
maggu2810: Hm, this differs from my suggested implementation a little bit (https://github.com/amitjoy/smarthome/compare/fix_range_events_delay...maggu2810:pr3752?expand=1#diff-2af4ee8503592a50a47a62e370e62fffR40).

forEach:
"Performs the given action for each element of the Iterable until all elements have been processed or the action throws an exception."

So, your one stopped if a job's run method throws a runtime excpetion (silently).
This could be hard to detect.

Mine logs a warning, and continue with the next job.

Do you want to use lambdas and stream whenever possible or why have modified it this way?

amitjoy: Iterating over the list using the `Iterable#forEach` is inherently thread-safe while iterating through a collection. I opted `Stream#forEach` out of this because it will not give us ordered execution. This is my primary motive. I could add logging behavior to catch the Runtime Exception as well.

amitjoy:
@maggu2810 Could you kindly have a look? I have amended this PR with your suggested improvements.

maggu2810:
```
[ERROR] Failed to execute goal org.eclipse.tycho:tycho-compiler-plugin:1.0.0:compile (default-compile) on project org.eclipse.smarthome.binding.astro: Compilation failure: Compilation failure:
[ERROR] /home/travis/build/eclipse/smarthome/extensions/binding/org.eclipse.smarthome.binding.astro/src/main/java/org/eclipse/smarthome/binding/astro/internal/job/CompositeJob.java:[30]
[ERROR] public CompositeJob(List<Job> jobs) {
[ERROR] ^^^^^^^^^^^^^^^^^^^^^^^^^^^^
[ERROR] Implicit super constructor AbstractJob() is undefined. Must explicitly invoke another constructor
[ERROR] 1 problem (1 error)
[ERROR] -> [Help 1]
```

amitjoy:
Ah. The other PR from @sjka got merged. I will rebase this PR.

extensions/binding/org.eclipse.smarthome.binding.astro/src/main/java/org/eclipse/smarthome/binding/astro/internal/job/CompositeJob.java
```diff
@@ -0,0 +1,58 @@
...
+
+/**
+ * {@link CompositeJob} comprises multiple {@link Job}s to be executed in order
+ *
+ * @author Markus Rathgeb - Initial contribution
+ * @author Amit Kumar Mondal - Minor modifications
+ */
+public final class CompositeJob extends AbstractJob {
+
+    private final String thingUID;
```
maggu2810: Drop that member, it is part of `AbstractJob` already.

extensions/binding/org.eclipse.smarthome.binding.astro/src/main/java/org/eclipse/smarthome/binding/astro/internal/job/CompositeJob.java
```diff
@@ -0,0 +1,58 @@
...
+     * @param jobs the jobs to execute
+     * @throws IllegalArgumentException
+     *             if {@code jobs} is {@code null} or empty
+     */
+    public CompositeJob(String thingUID, List<Job> jobs) {
+        super(thingUID);
+        checkArgument(jobs != null, "Jobs must not be null");
+        checkArgument(!jobs.isEmpty(), "Jobs must not be empty");
+
+        this.thingUID = thingUID;
```
maggu2810: Drop that line

extensions/binding/org.eclipse.smarthome.binding.astro/src/main/java/org/eclipse/smarthome/binding/astro/internal/job/CompositeJob.java
```diff
@@ -0,0 +1,58 @@
...
+        this.jobs = jobs;
+
+        boolean notMatched = jobs.stream().anyMatch(j -> !j.getThingUID().equals(thingUID));
+        checkArgument(!notMatched, "The jobs must associate the same thing UID");
+    }
+
+    @Override
+    public String getThingUID() {
```
maggu2810: Drop that member, it is part of `AbstractJob` already.

extensions/binding/org.eclipse.smarthome.binding.astro/src/main/java/org/eclipse/smarthome/binding/astro/internal/job/CompositeJob.java
amitjoy: Good catch. Thanks for noticing.

extensions/binding/org.eclipse.smarthome.binding.astro/src/main/java/org/eclipse/smarthome/binding/astro/internal/job/CompositeJob.java
amitjoy: Forgot to remove due to quick refactoring.

sjka:
Thanks!

PR #3754, opened 2017-06-27
[core] Added support for translating default system channels

### LINKED ISSUE #3623: add (german) translations for system channel types
System channel labels are currently only available in English. There are bindings that expect German texts, so this should be added (further languages upon request I would say).

See https://github.com/openhab/openhab2-addons/issues/1668.

cweitkamp:
* Added german translations

Closes #3623 

### DIFF
--- a/bundles/core/org.eclipse.smarthome.core.thing/ESH-INF/i18n/DefaultSystemChannels_de.properties
+++ b/bundles/core/org.eclipse.smarthome.core.thing/ESH-INF/i18n/DefaultSystemChannels_de.properties
@@ -1,6 +1,6 @@
 channel-type.system.signal-strength.label = Signalstärke
 channel-type.system.low-battery.label = Niedriger Batteriestatus
-channel-type.system.battery-level.label = Batterie Level
+channel-type.system.battery-level.label = Batterieladung
 channel-type.system.trigger.label = Auslöser
 channel-type.system.rawbutton.label = Einfache Taste
 channel-type.system.button.label = Taste
--- a/bundles/core/org.eclipse.smarthome.core.thing/ESH-INF/i18n/DefaultSystemChannels_de.properties
+++ b/bundles/core/org.eclipse.smarthome.core.thing/ESH-INF/i18n/DefaultSystemChannels_de.properties
@@ -1,0 +1,6 @@
+channel-type.system.signal-strength.label = Signalstärke
+channel-type.system.low-battery.label = Niedriger Batteriestatus
+channel-type.system.battery-level.label = Batterie Level
+channel-type.system.trigger.label = Auslöser
+channel-type.system.rawbutton.label = Einfache Taste
+channel-type.system.button.label = Taste
--- a/bundles/core/org.eclipse.smarthome.core.thing/src/main/java/org/eclipse/smarthome/core/thing/DefaultSystemChannelTypeProvider.java
+++ b/bundles/core/org.eclipse.smarthome.core.thing/src/main/java/org/eclipse/smarthome/core/thing/DefaultSystemChannelTypeProvider.java
@@ -8,11 +8,17 @@
 package org.eclipse.smarthome.core.thing;
 
 import java.math.BigDecimal;
+import java.util.ArrayList;
 import java.util.Arrays;
 import java.util.Collection;
 import java.util.Collections;
+import java.util.List;
 import java.util.Locale;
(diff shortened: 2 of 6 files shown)


cweitkamp:
The current implementation works, but I am not really happy with it. Two points:

1. I copied a lot code from the [XmlChannelTypeProvider](https://github.com/eclipse/smarthome/blob/master/bundles/core/org.eclipse.smarthome.core.thing.xml/src/main/java/org/eclipse/smarthome/core/thing/xml/internal/XmlChannelTypeProvider.java) class into the `DefaultSystemChannelTypeProvider` class. Both classes implements the `ChannelTypeProvider` interface, so maybe it makes sense to create a new abstract `LocalizedChannelTypeProvider` subclass containing the business logic for translating channels, states and more.
2. This is more a general architecture design question: Why are the default system channels hard coded into their own class? Isn't it more useful to define a xml-file for them? Like in every binding? If so, the first point is obsolete.

sjka:
wrt 1.: see #3423 (i.e. [AbstractXmlBasedProvider](https://github.com/sjka/smarthome/blob/85699470184f1649a1781c4e3a4f9ca14a3e47e8/bundles/config/org.eclipse.smarthome.config.xml/src/main/java/org/eclipse/smarthome/config/xml/AbstractXmlBasedProvider.java)) 😄

bundles/core/org.eclipse.smarthome.core.thing/ESH-INF/i18n/DefaultSystemChannels_de.properties
```diff
@@ -0,0 +1,6 @@
+channel-type.system.signal-strength.label = Signalst�rke
+channel-type.system.low-battery.label = Niedriger Batteriestatus
+channel-type.system.battery-level.label = Batterie Level
```
sjka: Hmmm, that somehow sounds "Denglisch" to me. The internet comes up with "Batteriezustand" or "Batterieladezustand". Not sure if they are better...

bundles/core/org.eclipse.smarthome.core.thing/src/main/java/org/eclipse/smarthome/core/thing/DefaultSystemChannelTypeProvider.java
```diff
@@ -89,34 +100,107 @@
                     new EventOption(CommonTriggerEvents.LONG_PRESSED, null))),
             null);
 
+    private class LocalizedChannelTypeKey {
```
sjka: Could you make this a static class? Then I think you also don't need to check for the outer type in hashcode and equals.

bundles/core/org.eclipse.smarthome.core.thing/src/main/java/org/eclipse/smarthome/core/thing/DefaultSystemChannelTypeProvider.java
```diff
@@ -89,34 +100,107 @@
...
+        public String locale;
```
sjka: final

bundles/core/org.eclipse.smarthome.core.thing/src/main/java/org/eclipse/smarthome/core/thing/DefaultSystemChannelTypeProvider.java
```diff
@@ -89,34 +100,107 @@
...
+        public UID uid;
```
sjka: final

bundles/core/org.eclipse.smarthome.core.thing/ESH-INF/i18n/DefaultSystemChannels_de.properties
cweitkamp: What about "Batterieladung"?

cweitkamp:
@sjka Thanks for your review. I addressed all your comments. I will hold my changes until #3423 is merged.

bundles/core/org.eclipse.smarthome.core.thing/ESH-INF/i18n/DefaultSystemChannels_de.properties
sjka: 👍

sjka:
Thanks!

PR #3755, opened 2017-06-27
config options provider for language & region settings

sjka:
I was just wondering why we don't have a config options provider for language/region settings. So please consider this PR as a question 😄 


bundles/config/org.eclipse.smarthome.config.core/OSGI-INF/I18nConfigOptionsProvider.xml
```diff
@@ -0,0 +1,17 @@
+<?xml version="1.0" encoding="UTF-8"?>
```
kaikreuzer: How about using the nice DS annotations that you have introduced recently? Adding further XMLs only makes the refactoring harder in future!

sjka: Just caught by my own lazyness to port them all.
I'll adapt this PR!

sjka: done

bundles/config/org.eclipse.smarthome.config.core/src/main/java/org/eclipse/smarthome/config/core/internal/i18n/I18nConfigOptionsProvider.java
```diff
@@ -0,0 +1,51 @@
...
+ *
+ * @author Simon Kaufmann - initial contribution and API
+ *
+ */
+@Component(immediate = true)
+public class I18nConfigOptionsProvider implements ConfigOptionProvider {
+
+    @Override
+    public Collection<ParameterOption> getParameterOptions(URI uri, String param, Locale locale) {
+        if (uri.toString().equals("system:i18n")) {
```
amitjoy: As the contract doesn't specify any throws clause, we cannot add any invariants here. Hence we could do this -> `"system:i18n".equals(String.valueOf(uri))`

bundles/config/org.eclipse.smarthome.config.core/src/main/java/org/eclipse/smarthome/config/core/internal/i18n/I18nConfigOptionsProvider.java
```diff
@@ -0,0 +1,51 @@
...
+            Locale translation = locale != null ? locale : Locale.getDefault();
+            if (param.equals("language")) {
```
amitjoy: `param` can also be `null`, we can do this -> `"language".equals(param)`.

bundles/config/org.eclipse.smarthome.config.core/src/main/java/org/eclipse/smarthome/config/core/internal/i18n/I18nConfigOptionsProvider.java
```diff
@@ -0,0 +1,51 @@
...
+                return getAvailable(locale, l -> new ParameterOption(l.getLanguage(), l.getDisplayLanguage(translation)));
+            } else if (param.equals("region")) {
+                return getAvailable(locale, l -> new ParameterOption(l.getCountry(), l.getDisplayCountry(translation)));
+            } else if (param.equals("variant")) {
+                return getAvailable(locale, l -> new ParameterOption(l.getVariant(), l.getDisplayVariant(translation)));
+            }
+        }
+        return null;
```
amitjoy: According to the contract, we need to return empty list.

bundles/config/org.eclipse.smarthome.config.core/src/main/java/org/eclipse/smarthome/config/core/internal/i18n/I18nConfigOptionsProvider.java
```diff
@@ -0,0 +1,51 @@
...
+    }
+
+    private Collection<ParameterOption> getAvailable(Locale locale, Function<Locale, ParameterOption> mapFunction) {
+        return Arrays.stream(Locale.getAvailableLocales()).map(l -> mapFunction.apply(l)).distinct()
+                .sorted(Comparator.comparing(a -> a.getLabel())).collect(Collectors.toList());
```
amitjoy: Just a syntactic sugar: `ParameterOption::getLabel`

bundles/config/org.eclipse.smarthome.config.core/src/main/java/org/eclipse/smarthome/config/core/internal/i18n/I18nConfigOptionsProvider.java
sjka: a param with a null name? how should that work?

bundles/config/org.eclipse.smarthome.config.core/src/main/java/org/eclipse/smarthome/config/core/internal/i18n/I18nConfigOptionsProvider.java
sjka: uri is a non-null argument (see caller)

bundles/config/org.eclipse.smarthome.config.core/src/main/java/org/eclipse/smarthome/config/core/internal/i18n/I18nConfigOptionsProvider.java
sjka: right - javadoc and caller implementation don't match

bundles/config/org.eclipse.smarthome.config.core/src/main/java/org/eclipse/smarthome/config/core/internal/i18n/I18nConfigOptionsProvider.java
sjka: I guess that's personal flavor, isn't it?

bundles/config/org.eclipse.smarthome.config.core/src/main/java/org/eclipse/smarthome/config/core/internal/i18n/I18nConfigOptionsProvider.java
amitjoy: I had a quick look on the contract and the contract doesn't specify uri to be non null. If you think, the callers will never use null, you can ignore this.

bundles/config/org.eclipse.smarthome.config.core/src/main/java/org/eclipse/smarthome/config/core/internal/i18n/I18nConfigOptionsProvider.java
amitjoy: No clue if param is never null in the callers. Just the contract was not clear about it.

bundles/config/org.eclipse.smarthome.config.core/src/main/java/org/eclipse/smarthome/config/core/internal/i18n/I18nConfigOptionsProvider.java
sjka: #3761

bundles/config/org.eclipse.smarthome.config.core/src/main/java/org/eclipse/smarthome/config/core/internal/i18n/I18nConfigOptionsProvider.java
amitjoy: definitely. It's upto you.

bundles/config/org.eclipse.smarthome.config.core/src/main/java/org/eclipse/smarthome/config/core/internal/i18n/I18nConfigOptionsProvider.java
sjka: Oh, the contract in fact is pretty clear:
[ConfigDescriptionParameter.java#L204](https://github.com/eclipse/smarthome/blob/7b9d32430384f1f0bb06381014394b090d8681cb/bundles/config/org.eclipse.smarthome.config.core/src/main/java/org/eclipse/smarthome/config/core/ConfigDescriptionParameter.java#L204), [config-description-1.0.0.xsd#L59](https://github.com/eclipse/smarthome/blob/78b6ef19d9fc8da7969f41228d3f3b56f21b8a7f/bundles/config/org.eclipse.smarthome.config.xml/config-description-1.0.0.xsd#L59)
It's just at the wrong place/not replicated everywhere (yet), depending on your perspective.

amitjoy: You are right. I only looked into `ConfigOptionProvider` interface but its callers handle the invariants separately. I believe it would also be beneficial if such invariants can also be added to the interfaces.

PR #3761, opened 2017-06-28
adapt ConfigOptionProvider javadoc to reality

sjka:
...as several implementations return null. Also, it does not make
much sense to instantiate empty collections all the time, as the
majority of call will be returned with no results.

### DIFF
--- a/bundles/config/org.eclipse.smarthome.config.core/src/main/java/org/eclipse/smarthome/config/core/ConfigOptionProvider.java
+++ b/bundles/config/org.eclipse.smarthome.config.core/src/main/java/org/eclipse/smarthome/config/core/ConfigOptionProvider.java
@@ -28,8 +28,7 @@ public interface ConfigOptionProvider {
      *            the parameter name for which the requested options shall be returned
      * @param locale
      *            locale
-     * @return the configuration options provided by this provider (not
-     *         null, could be empty)
+     * @return the configuration options provided by this provider if any or {@code null} otherwise
      */
     Collection<ParameterOption> getParameterOptions(URI uri, String param, Locale locale);
 }
--- a/bundles/core/org.eclipse.smarthome.core.persistence/src/main/java/org/eclipse/smarthome/core/persistence/internal/PersistenceServiceRegistryImpl.java
+++ b/bundles/core/org.eclipse.smarthome.core.persistence/src/main/java/org/eclipse/smarthome/core/persistence/internal/PersistenceServiceRegistryImpl.java
@@ -89,13 +89,14 @@ public class PersistenceServiceRegistryImpl implements ConfigOptionProvider, Per
 
     @Override
     public Collection<ParameterOption> getParameterOptions(URI uri, String param, Locale locale) {
-        Set<ParameterOption> options = new HashSet<>();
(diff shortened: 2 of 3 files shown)


bundles/config/org.eclipse.smarthome.config.core/src/main/java/org/eclipse/smarthome/config/core/ConfigOptionProvider.java
```diff
@@ -28,8 +28,7 @@
      *            the parameter name for which the requested options shall be returned
      * @param locale
      *            locale
-     * @return the configuration options provided by this provider (not
-     *         null, could be empty)
+     * @return the configuration options provided by this provider if any (may be null)
```
maggu2810: So, can you add some information on which situations we should return an empty collection and on which one we should return null?
What makes the difference?

Can we also change the current implementations in the ESH repo to use a consistent logic.

E.g.
* the URI is not handled => null, otherwise empty or filled collection
* URI or param not handled => null, otherwise a collection that contains the parameter

sjka: What is the value of differentiating it? 
I simply adapted to JavaDoc do explicitly allow returning null, i.e. to indicate that the caller is able to handle it.

maggu2810:
Isn't it much simpler to be able to use:

```Java
for (final ParameterOption option : provider.getParameterOptions()) {
    ...
}
```

instead of

```Java
final Collection<ParameterOption> options = provider.getParameterOptions();
if (options != null) {
    for (final ParameterOption option : provider.getParameterOptions()) {
        ...
    }
}
```

> What is the value of differentiating it?

If there is no difference between an empty collection and null, I would prefer to change the implementation.
If `Collections.empty...` is used, no new instances are created, because this methods return singeltons.


Shouldn't be the documentation of the API be as detailed as possible?
Previously the developer has been told to return never null, but an empty collection.
Now he is allowed to return both.
I don't know in which situations I should return the one and in which the other.
Now the documentation of the interface is changed to fit some implementations of the interface.

sjka:
> Now the documentation of the interface is changed to fit some implementations of the interface.

No, it is changed to fit the implementation of the caller. 

We have a similar pattern everywhere (e.g. ThingHandlerFactories, ConfigDescriptionProviders, StateDescriptionProvider,....): They all are allowed to return null in case they don't have anything to contribute.

sjka:
I made the javadoc more clear so that it does not imply anymore that there is a difference between an empty collection and null.

Also, I changed the persistence service registry to return null in case it does not have anything. It was the only implementation that _alway_ returned an empty collection.

maggu2810:
Thanks.

What about `org.eclipse.smarthome.magic.binding.internal.MagicServiceImpl`?
It also contains `return Collections.emptyList();` twice.

sjka:
Ouch, well... Had it closed because I was switching between old branches and therefore missed it. Thanks for catching this!

maggu2810:
Travis failed because of the Magic bundle test.
The test requires an empty collection.
I pressed "merge" too fast.

maggu2810:
See: #3781

PR #3762, opened 2017-06-28
add Identifiable interface - used by Registry and Provider

maggu2810:
Related to: https://github.com/eclipse/smarthome/pull/3670#discussion_r123556157

### DIFF
--- a/bundles/core/org.eclipse.smarthome.core/src/main/java/org/eclipse/smarthome/core/common/registry/Identifiable.java
+++ b/bundles/core/org.eclipse.smarthome.core/src/main/java/org/eclipse/smarthome/core/common/registry/Identifiable.java
@@ -16,7 +16,7 @@ package org.eclipse.smarthome.core.common.registry;
  *
  * @author Markus Rathgeb - Initial contribution
  */
-public interface UniqueIdProvider<T> {
+public interface Identifiable<T> {
 
     /**
      * Get the unique identifier.
--- a/bundles/automation/org.eclipse.smarthome.automation.api/src/main/java/org/eclipse/smarthome/automation/Rule.java
+++ b/bundles/automation/org.eclipse.smarthome.automation.api/src/main/java/org/eclipse/smarthome/automation/Rule.java
@@ -15,6 +15,7 @@ import java.util.Set;
 
 import org.eclipse.smarthome.config.core.ConfigDescriptionParameter;
 import org.eclipse.smarthome.config.core.Configuration;
+import org.eclipse.smarthome.core.common.registry.UniqueIdProvider;
 
 /**
  * Rule is built from {@link Module}s and consists of three sections:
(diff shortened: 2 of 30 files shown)


bundles/core/org.eclipse.smarthome.core/src/main/java/org/eclipse/smarthome/core/common/registry/Identifiable.java
```diff
@@ -0,0 +1,27 @@
...
+ * http://www.eclipse.org/legal/epl-v10.html
+ */
+package org.eclipse.smarthome.core.common.registry;
+
+/**
+ * Interface for classes that instances provide a system wide unique identifier.
+ *
+ * <p>
+ * The requirement "unique" should be seen at least unique per respective type.
+ * Two unrelated stuff that is not hold at the same place every could perhaps return the same identifier.
```
sjka: In this sentence something got scrambled

bundles/core/org.eclipse.smarthome.core/src/main/java/org/eclipse/smarthome/core/common/registry/Identifiable.java
```diff
@@ -0,0 +1,27 @@
...
+ * @author Markus Rathgeb - Initial contribution
+ */
+public interface Identifiable<T> {
+
+    /**
+     * Get the unique identifier.
+     *
+     * @return the unique identifier
+     */
+    T getUID();
```
sjka: As we _have_ a `UID` class but are not restricted to it and uniqueness is relative, wouldn't `getId()` more appropriate?

maggu2810: As written in the references issue, most of the respective classes already contain a method `getUid()`. The only one using `getId()` has been the link type.
Rules, tempalte, modules, items, things, ... (IIRC) using `getUID()`.

bundles/core/org.eclipse.smarthome.core/src/main/java/org/eclipse/smarthome/core/common/registry/Identifiable.java
maggu2810: I considered this yesterday, too -- but the PC has already been shut down :wink:

maggu2810:
> Just wondering: Do you already know if this is going to be binary compatible with existing bindings? I.e. do they need to be recompiled?

As long as no changed method is used (WRT the changes to the link) I don't think it should matter.
But I haven't tested it, but recompiled the whole tree.

bundles/core/org.eclipse.smarthome.core/src/main/java/org/eclipse/smarthome/core/common/registry/Identifiable.java
sjka: 👍  very good point, indeed 😒

maggu2810:
What about this one https://github.com/eclipse/smarthome/pull/3762#discussion_r124725109

kaikreuzer:
Oops, sorry, that slipped my attention when merging... Might I ask you for a separate PR for it? 🤕

sjka:
Added the APIBreaking label, as the Registry`s signatures are not binary compatible:

```java
o.e.sh.core.common.registry.Registry.update(java.lang.Object)
```

became

```java
o.e.sh.core.common.registry.Registry.update(o.e.sh.core.common.registry.Identifiable)
```

PR #3764, opened 2017-06-28
[audio] Added getSupportedStreams() and UnsupportedAudioStreamException

### LINKED ISSUE #3566: [audio] New exception type if audio sink doesn't support the served AudioStream.
When looking into [https://github.com/openhab/openhab2-addons/pull/2303](https://github.com/openhab/openhab2-addons/pull/2303) we came up with the question if we need an additional exception type `UnsupportedAudioStreamException` to be thrown if an audio sink doesn't support the served `AudioStream`. Currently e.g. the [Sonos](https://github.com/eclipse/smarthome/blob/master/extensions/binding/org.eclipse.smarthome.binding.sonos/src/main/java/org/eclipse/smarthome/binding/sonos/internal/SonosAudioSink.java) binding throws an `UnsupportedAudioFormatException` - which type is not exactly matching the use case - whereas other bindings (e.g. Kodi in the linked PR or [AllPlay](https://github.com/openhab/openhab2-addons/blob/master/addons/binding/org.openhab.binding.allplay/src/main/java/org/openhab/binding/allplay/internal/AllPlayAudioSink.java)) only log a warning or throw an individual exception.

Any thoughts?

cweitkamp:
Closes #3566 

### DIFF
--- a/bundles/core/org.eclipse.smarthome.core.audio/src/main/java/org/eclipse/smarthome/core/audio/UnsupportedAudioStreamException.java
+++ b/bundles/core/org.eclipse.smarthome.core.audio/src/main/java/org/eclipse/smarthome/core/audio/UnsupportedAudioStreamException.java
@@ -11,7 +11,7 @@ package org.eclipse.smarthome.core.audio;
  * Thrown when a requested {@link AudioStream} is not supported by an {@link AudioSource} or {@link AudioSink}
  * implementation
  * 
- * @author Christoph Weitkamp - Added UnsupportedAudioStreamException
+ * @author Christoph Weitkamp - Initial contribution and API
  * 
  */
 public class UnsupportedAudioStreamException extends AudioException {
--- a/bundles/core/org.eclipse.smarthome.core.audio/src/main/java/org/eclipse/smarthome/core/audio/UnsupportedAudioStreamException.java
+++ b/bundles/core/org.eclipse.smarthome.core.audio/src/main/java/org/eclipse/smarthome/core/audio/UnsupportedAudioStreamException.java
@@ -1,0 +1,57 @@
+/**
+ * Copyright (c) 2014-2017 by the respective copyright holders.
+ * All rights reserved. This program and the accompanying materials
+ * are made available under the terms of the Eclipse Public License v1.0
+ * which accompanies this distribution, and is available at
+ * http://www.eclipse.org/legal/epl-v10.html
+ */
+package org.eclipse.smarthome.core.audio;
+
+/**
+ * Thrown when a requested {@link AudioStream} is not supported by an {@link AudioSource} or {@link AudioSink}
+ * implementation
+ * 
+ * @author Christoph Weitkamp - Added getSupportedStreams() and UnsupportedAudioStreamException
+ * 
+ */
+public class UnsupportedAudioStreamException extends AudioException {
+
+    private static final long serialVersionUID = 1L;
+
+    /**
+     * Unsupported {@link AudioStream}
+     */
+    private Class<? extends AudioStream> unsupportedAudioStreamClass;
+
+    /**
+     * Constructs a new exception with the specified detail message, unsupported format, and cause.
+     *
+     * @param message The message
+     * @param unsupportedAudioStreamClass The unsupported audio stream class
+     * @param cause The cause
+     */
+    public UnsupportedAudioStreamException(String message, Class<? extends AudioStream> unsupportedAudioStreamClass,
+            Throwable cause) {
+        super(message, cause);
+        this.unsupportedAudioStreamClass = unsupportedAudioStreamClass;
+    }
+
+    /**
+     * Constructs a new exception with the specified detail message and unsupported format.
+     *
+     * @param message The message
+     * @param unsupportedAudioStreamClass The unsupported audio stream class
+     */
+    public UnsupportedAudioStreamException(String message, Class<? extends AudioStream> unsupportedAudioStreamClass) {
+        this(message, unsupportedAudioStreamClass, null);
+    }
+
+    /**
+     * Gets the unsupported audio stream class.
+     *
+     * @return The unsupported audio stream class
+     */
+    public Class<? extends AudioStream> getUnsupportedAudioStreamClass() {
+        return unsupportedAudioStreamClass;
+    }
+}
--- a/bundles/core/org.eclipse.smarthome.core.audio/src/main/java/org/eclipse/smarthome/core/audio/UnsupportedAudioStreamException.java
+++ b/bundles/core/org.eclipse.smarthome.core.audio/src/main/java/org/eclipse/smarthome/core/audio/UnsupportedAudioStreamException.java
@@ -11,7 +11,7 @@ package org.eclipse.smarthome.core.audio;
  * Thrown when a requested {@link AudioStream} is not supported by an {@link AudioSource} or {@link AudioSink}
  * implementation
  * 
- * @author Christoph Weitkamp - Added getSupportedStreams() and UnsupportedAudioStreamException
+ * @author Christoph Weitkamp - Added UnsupportedAudioStreamException
  * 
  */
 public class UnsupportedAudioStreamException extends AudioException {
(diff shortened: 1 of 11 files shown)


cweitkamp:
CAUTION: This is a change breaking PR for bindings using an `AudioSink`. It itroduces a new abstract method `getSupportedStreams()`.

I will post (or push) an example implementation later.

sjka:
Could you please also adapt the voice bundle?

```
[ERROR] Failed to execute goal org.eclipse.tycho:tycho-compiler-plugin:1.0.0:compile (default-compile) on project org.eclipse.smarthome.core.voice: Compilation failure: Compilation failure:
[ERROR] /home/travis/build/eclipse/smarthome/bundles/core/org.eclipse.smarthome.core.voice/src/main/java/org/eclipse/smarthome/core/voice/internal/DialogProcessor.java:[190]
[ERROR] sink.process(audioStream);
[ERROR] ^^^^^^^^^^^^^^^^^^^^^^^^^
[ERROR] Unhandled exception type UnsupportedAudioStreamException
[ERROR] /home/travis/build/eclipse/smarthome/bundles/core/org.eclipse.smarthome.core.voice/src/main/java/org/eclipse/smarthome/core/voice/internal/VoiceManagerImpl.java:[169]
[ERROR] sink.process(audioStream);
[ERROR] ^^^^^^^^^^^^^^^^^^^^^^^^^
[ERROR] Unhandled exception type UnsupportedAudioStreamException
[ERROR] 2 problems (2 errors)
```

cweitkamp:
@sjka Done.

I also added potential examples for `getSupportedFormats()` and `getSupportedStreams()` to the unit test class `AudioSinkFake`.

sjka:
Thanks!

Now, as you already predicted above, Travis has got another wish:

```
public class SonosAudioSink implements AudioSink {
[ERROR] ^^^^^^^^^^^^^^
[ERROR] The type SonosAudioSink must implement the inherited abstract method AudioSink.getSupportedStreams()
[ERROR] 1 problem (1 error)
```

cweitkamp:
@sjka Done.

I applied those changes to binding.sonos, io. javasound and io.webaudio to prevent good boy Travis to fail again.

bundles/core/org.eclipse.smarthome.core.audio/src/main/java/org/eclipse/smarthome/core/audio/UnsupportedAudioStreamException.java
```diff
@@ -0,0 +1,57 @@
...
+ * which accompanies this distribution, and is available at
+ * http://www.eclipse.org/legal/epl-v10.html
+ */
+package org.eclipse.smarthome.core.audio;
+
+/**
+ * Thrown when a requested {@link AudioStream} is not supported by an {@link AudioSource} or {@link AudioSink}
+ * implementation
+ * 
+ * @author Christoph Weitkamp - Added UnsupportedAudioStreamException
```
kaikreuzer: Put "Initial contribution and API" here.

bundles/core/org.eclipse.smarthome.core.voice/src/main/java/org/eclipse/smarthome/core/voice/internal/DialogProcessor.java
```diff
@@ -184,12 +186,17 @@ protected void say(String text) {
                 }
             }
             if (null == voice) {
-                throw new TTSException("Unable to find a suitable voice");
+                throw new TTSException("Unable to find a voice");
```
kaikreuzer: why did you change this? It seems unrelated to this PR and the message is not correct anymore as the system probably was able to find a voice, but not one of the right locale.

bundles/core/org.eclipse.smarthome.core.voice/src/main/java/org/eclipse/smarthome/core/voice/internal/VoiceManagerImpl.java
```diff
@@ -139,20 +141,24 @@ public void say(String text, String voiceId, String sinkId) {
             } else if (voiceId.contains(":")) {
                 // it is a fully qualified unique id
                 String[] segments = voiceId.split(":");
-                tts = ttsServices.get(segments[0]);
-                voice = getVoice(tts.getAvailableVoices(), segments[1]);
+                tts = getTTS(segments[0]);
```
kaikreuzer: Why did you change this? This potentially returns you the default tts instead of the one specified in `segments[0]`, so it unnecessarily complicates things here.

extensions/binding/org.eclipse.smarthome.binding.sonos/src/main/java/org/eclipse/smarthome/binding/sonos/internal/SonosAudioSink.java
```diff
@@ -74,41 +82,46 @@ public void process(AudioStream audioStream) throws UnsupportedAudioFormatExcept
...
+                    throw new UnsupportedAudioFormatException("Sonos only supports MP3 or WAV.", format);
                 }
+            } else {
+                logger.warn("We do not have any callback url, so Sonos cannot play the audio stream!");
             }
+        } else {
+            IOUtils.closeQuietly(audioStream);
+            throw new UnsupportedAudioStreamException(
+                    "Sonos can only handle FixedLengthAudioStreams and URLAudioStreams.", audioStream.getClass());
+            // TODO: Instead of throwing an exception, we could ourselves try to wrap it into a
```
kaikreuzer: Is this todo still relevant? If I understand @maggu2810 right, it should now be rather the framework, which decides to do the conversion to a FixedLengthAudioStream, in case it sees that the sink does not support anything else.

extensions/io/org.eclipse.smarthome.io.javasound/src/main/java/org/eclipse/smarthome/io/javasound/internal/JavaSoundAudioSink.java
```diff
@@ -71,6 +75,11 @@ public void process(AudioStream audioStream) throws UnsupportedAudioFormatExcept
     }
 
     @Override
+    public Set<Class<? extends AudioStream>> getSupportedStreams() {
+        return Collections.emptySet();
```
kaikreuzer: Why empty? It supports ALL streams!

sjka:
Thanks! 
(btw, this time I made the build fail - will be fixed in a few minutes and I will trigger it again)

bundles/core/org.eclipse.smarthome.core.voice/src/main/java/org/eclipse/smarthome/core/voice/internal/VoiceManagerImpl.java
cweitkamp: I changed it for readability and to prevent NPE (in case of `segment[0]` contains a not suitable value). Have a look at [https://github.com/eclipse/smarthome/pull/3764/files#diff-2bdbf6176fb6bd9b1af9a6cb230b6293R472](https://github.com/eclipse/smarthome/pull/3764/files#diff-2bdbf6176fb6bd9b1af9a6cb230b6293R472). The `getTTS(String)` method returns exaclty the same like before (`ttsServices.get(id)`).

cweitkamp:
> Is it by intention that the newly introduced method getSupportedStreams is actually not used anywhere? I thought the idea of it was to allow the framework to check it upfront, instead of having exceptions being thrown at runtime...

@kaikreuzer thanks for review. No, of course not. My intention was to change it in a follow-up PR. I can do that in this PR if you preferr.

kaikreuzer:
Many thanks @cweitkamp! Ok, all fine for me to make use of this feature in an upcoming PR.

kaikreuzer:
We need to remember that once the next ESH stable is pushed out, quite some bindings in openhab2-addons will break and need to be adapted.

lolodomo:
The timing looks not very good. Such a breaking API will prevent any update of OH2 bindings relying on it for any user running OH 2.1 during around 5 months. We have to pray that no urgent patch is needed for all impacted bindings.
What will happen for users using the online distribution ?

kaikreuzer:
We are talking about changes on unstable development branches, so there should be no need for praying or anything.

lolodomo:
@cweitkamp: may we ask you to provide a fix for all broken OH2 bindings ?

cweitkamp:
@lolodomo Yes of course. I am working on it.

cweitkamp:
PR [https://github.com/openhab/openhab2-addons/pull/2457](https://github.com/openhab/openhab2-addons/pull/2457)

PR #3775, opened 2017-06-30
Improved formatting of position of validation errors

### LINKED ISSUE #3770: bad formatting of validation error position in log file
I updated my oh2 to oh2.1 and now one of my sitemaps doesn't load with follwing error:
```
[WARN ] [el.core.internal.ModelRepositoryImpl] - Configuration model 'default.sitemap' has errors, therefore ignoring it: [1,219,16]: missing RULE_ID at 'label='
[1,224,16]: missing RULE_ID at 'label='
```

What does this mean? Line 219 and 224?

kaikreuzer:
Fixes #3770


sjka:
Haha, makes sense! 

Now I'm just wondering whether we might even want to go a step further and print a revolutionarily user-friendly message like this:

```...[line 1224, column 16]...```

?

kaikreuzer:
I was wondering the same, but decided against it as I prefer messages without much bloat and `[line,col]` should be a format that people understand. The main issue here was that there were all of a sudden 3 segments, which was pretty confusing.

sjka:
agreed.

Issue #3185, opened 2017-03-27
Change to DayLigth saving Time causes stop in timerelated persistence strategies


JueBag:
As reported in the OH community ( [Link](https://community.openhab.org/t/openhab-not-saving-mesaurements-after-daylight-saving-time-change/25647)) the change to DayLigth Saving Time caused persistence schedules that are only using a time based strategy (everyMinute etc.) to stop at the time-switch.

Checks done on my side:
Actual values continued to be displayed.
The check of the database (rrd4j with a "everyMinute" strategy) showed the last value to be stored at the time-switch.
Several users confirmed that after restart of the OH-instance the time-scheduled persistence was working again. 
There is also an report on a cron based rule that stopped working, can't say if it is really related to the above.

kaikreuzer:
@kgoderis This might be a bug in the cron scheduler, right?

kgoderis:
@kaikreuzer most probably. I hope I can figure this one out, not sure how I will simulate a change in DLST over here.

pdegeus:
Just found this issue after noticing my influxdb stopped collecting data at exactly 3:00 this Sunday.
If it helps, my log file shows this:
```
2017-03-26 03:00:00.018 [INFO ] [cheduler.ExpressionThreadPoolManager] - Expression '0 0 * * * ?' has no future executions anymore
2017-03-26 03:00:00.019 [INFO ] [cheduler.ExpressionThreadPoolManager] - Expression '0 0 * * * ?' has no future executions anymore
2017-03-26 03:00:00.024 [INFO ] [cheduler.ExpressionThreadPoolManager] - Expression '0 * * * * ?' has no future executions anymore
2017-03-26 03:00:00.028 [INFO ] [cheduler.ExpressionThreadPoolManager] - Expression '0 * * * * ?' has no future executions anymore
```

kaikreuzer:
Any success @kgoderis?

kgoderis:
@kaikreuzer No(t yet). As you have might have noticed I am quite absent lately due to an overload @ my businesses.

kaikreuzer:
No rush - I know that you are waiting for any other things for me and I admire how you manage to do all that stuff despite of your businesses :-)

PR #3776, opened 2017-06-30
use classic scheduler for one-time jobs

sjka:
The (cron) expression based scheduler is still used for the "daily" jobs
which will be executed every midnight. Those "daily" jobs however create
jobs for all kinds of events during the current day, which will have to
be executed only once. For those, it does not really make sense to convert
a timestamp to a cron expression, which the scheduler again with lots of
effort boils down again to a timestamp. Therefore they are now simply
executed with a calculated delay directly using the regular scheduler.

### DIFF
--- a/extensions/binding/org.eclipse.smarthome.binding.astro/src/main/java/org/eclipse/smarthome/binding/astro/handler/AstroThingHandler.java
+++ b/extensions/binding/org.eclipse.smarthome.binding.astro/src/main/java/org/eclipse/smarthome/binding/astro/handler/AstroThingHandler.java
@@ -7,27 +7,24 @@
  */
 package org.eclipse.smarthome.binding.astro.handler;
 
-import static java.util.stream.Collectors.toList;
-import static org.eclipse.smarthome.core.scheduler.CronHelper.*;
 import static org.eclipse.smarthome.core.thing.ThingStatus.*;
 import static org.eclipse.smarthome.core.thing.type.ChannelKind.TRIGGER;
 import static org.eclipse.smarthome.core.types.RefreshType.REFRESH;
 
 import java.lang.invoke.MethodHandles;
 import java.text.ParseException;
-import java.time.LocalDateTime;
-import java.time.ZoneId;
+import java.util.Calendar;
 import java.util.Date;
-import java.util.List;
-import java.util.Objects;
-import java.util.Queue;
-import java.util.concurrent.LinkedBlockingQueue;
+import java.util.HashSet;
+import java.util.Set;
+import java.util.concurrent.ScheduledFuture;
+import java.util.concurrent.TimeUnit;
 import java.util.concurrent.locks.Lock;
 import java.util.concurrent.locks.ReentrantLock;
-import java.util.function.Consumer;
 
 import org.apache.commons.lang.ArrayUtils;
 import org.apache.commons.lang.StringUtils;
+import org.apache.commons.lang.time.DateFormatUtils;
 import org.eclipse.smarthome.binding.astro.internal.config.AstroChannelConfig;
 import org.eclipse.smarthome.binding.astro.internal.config.AstroThingConfig;
 import org.eclipse.smarthome.binding.astro.internal.job.Job;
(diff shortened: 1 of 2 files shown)


extensions/binding/org.eclipse.smarthome.binding.astro/src/main/java/org/eclipse/smarthome/binding/astro/handler/AstroThingHandler.java
```diff
@@ -298,22 +270,23 @@ public void triggerEvent(String channelId, String event) {
...
      *
      * @return {@code true} if the {@code job} is added to the queue, otherwise {@code false}
      */
-    public boolean addJobToQueue(Job job) {
-        if (job != null) {
-            return scheduledJobs.add(job);
+    public void schedule(Job job, Calendar eventAt) {
+        monitor.lock();
+        try {
+            long sleepTime = eventAt.getTimeInMillis() - new Date().getTime();
```
maggu2810: Does the ESH scheduler esp. for Cron Expressions handle time changes?
If I change my system time, the delta needs to be recalculated, too.
Is this out of scope here?

extensions/binding/org.eclipse.smarthome.binding.astro/src/main/java/org/eclipse/smarthome/binding/astro/handler/AstroThingHandler.java
```diff
@@ -298,22 +270,23 @@ public void triggerEvent(String channelId, String event) {
...
+            ScheduledFuture<?> future = scheduler.schedule(job, sleepTime, TimeUnit.MILLISECONDS);
+            scheduledFutures.add(future);
+            if (logger.isDebugEnabled()) {
```
maggu2810: Shouldn't external functions called as rarely as possible while holding locks?
Should we move the logging out of the lock unlock section?

sjka: good point, yes

extensions/binding/org.eclipse.smarthome.binding.astro/src/main/java/org/eclipse/smarthome/binding/astro/handler/AstroThingHandler.java
sjka: The ESH scheduler suffers from the same problem, see #3185. Therefore I left it out of scope for _this_ PR.

extensions/binding/org.eclipse.smarthome.binding.astro/src/main/java/org/eclipse/smarthome/binding/astro/handler/AstroThingHandler.java
sjka: done

extensions/binding/org.eclipse.smarthome.binding.astro/src/main/java/org/eclipse/smarthome/binding/astro/handler/AstroThingHandler.java
maggu2810: The reason I pointed it here is, that using our own scheduler implementation we are able to handle such (corner) cases in our implementation and all usages are fixed.
Calculating the delta and use the standard scheduler will need a fix for every such usages.

But if you think it does not matter, I agree with your changes.

sjka: I see two approaches: 

* we maybe can handle this easily as the astro calculations internally use julian dates, which is not subject to such anomalies - so it would be easiest to get the amount of milliseconds to wait directly from there instead of messing around with Date/Calendar objects which for this purpose actually are not really helpful and just create chaos. So it could be that we were better off staying within JD and not taking this detour. But I have to admit that I did not yet focus in this.

* Once the calculations get fixed in the scheduler, I would be happy to re-use the same utility method to calculate the time difference here. But I still do not see that this would also require using the _expression_ based scheduling as a whole. Maybe it would make sense to create such utility methods in the ScheduledExecutorService that our ThreadPoolManager returns.

So while I agree that we need to fix it and only do it once, I don't see that it conflicts with this.

maggu2810: > we maybe can handle this easily as the astro calculations internally use julian dates,

It has been more about the drift of RTCs in embedded systems.
E.g. you realize that the system time is some minutes in the future or past and you change the system time to fix that situation (if you use OS features like adjtime or fix it by a jump is out of scope).

Okay, let's postpone that cases...

kgoderis:
@sjka I am quite late to the party ;-), but I just read the above. You know that you could have used the DateExpression class in o.e.s.c.scheduler to schedule tasks on precisely calculated dates?

sjka:
No problem, happy to have you back here 😄 
Yes, thanks for the hint! I indeed saw that class but as in the astro binding already exists a "Date" object, I didn't want to first convert it to a String that then will get parsed again. So I thought it might be easiest to go with the "classic" scheduler. Which is this special case was fully sufficient.

PR #3777, opened 2017-06-30
dropped CronHelper

sjka:
...as if you needed it, you were using the wrong scheduler.

The regular scheduler (ScheduledExecutorService) _has_ methods like
   schedule([...], long delay, [...]) and
   scheduleWithFixedDelay([...], long initialDelay, long delay, [...]);
which can be used much more efficiently for running some code at a given
point in time - either once or on a regular basis.

It's pretty unefficient and error prone to first construct an artificial cron
expression for a specific point in time and then let the expression scheduler
with lots of effort calculate the timestamp again at which it should be executed.
It itself then schedules the given job with the time difference in milliseconds
between the current time and the re-calculated execution time. So this really
did not make much sense.


sjka:
It needs #3776 to be merged first, otherwise the astro binding will break.

sjka:
PS: This is breaking the API which was just introduced a few days ago. However, I'd suggest rather doing it now when it has not been adopted than later

Issue #3590, opened 2017-06-06
scripts: ctx.ruleUID not available


maggu2810:
#3007 added the ruleUID variable to the ctx object, so we could use `ctx.ruleUID` inside of scripts.

#1783 removed ruleUID from ctx but add it to the global namespace, so a script could access the rule's UID by using `ruleUID` (without `ctx.`).

#1783 has been merged "long" time ago, but "my" instances still uses an ESH version before that changes.
Now I tried to update that instances using a more recent ESH version and I realized that all the scripts does not work anymore.

I was not aware of that change and I do not see that we discussed that move (at least not in #3007 where we discussed that it is okay to leave the ruleUID in the context of scripts).

@smerschjohann What has been the reason for that change?

kaikreuzer:
@smerschjohann Could you please comment?

smerschjohann:
hm, that was not my intention. I think it is correct that it is in the global namespace (without ctx), but I think I failed at the merge of that sourcefile. You can of course add that to the ctx object again.

PR #3780, opened 2017-06-30
ruleUID should be part of 'ctx'

maggu2810:
Fixes: https://github.com/eclipse/smarthome/issues/3590

@smerschjohann Could you have a look at?

### DIFF
--- a/bundles/automation/org.eclipse.smarthome.automation.module.script/src/main/java/org/eclipse/smarthome/automation/module/script/internal/handler/AbstractScriptModuleHandler.java
+++ b/bundles/automation/org.eclipse.smarthome.automation.module.script/src/main/java/org/eclipse/smarthome/automation/module/script/internal/handler/AbstractScriptModuleHandler.java
@@ -7,6 +7,7 @@
  */
 package org.eclipse.smarthome.automation.module.script.internal.handler;
 
+import java.util.HashMap;
 import java.util.Map;
 import java.util.Map.Entry;
 import java.util.Optional;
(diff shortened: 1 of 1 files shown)


bundles/automation/org.eclipse.smarthome.automation.module.script/src/main/java/org/eclipse/smarthome/automation/module/script/internal/handler/AbstractScriptModuleHandler.java
```diff
@@ -106,10 +107,13 @@ private void loadConfig() {
...
 
-        // make the whole context available as "ctx"
+        // Add the rule's UID to the context
+        final Map<String, Object> contextNew = new HashMap<>(context);
+        contextNew.put("ruleUID", this.ruleUID);
+
+        // make the whole (new) context available as "ctx"
         // Note: We don't use "context" here as it doesn't work on all JVM versions!
-        executionContext.setAttribute("ctx", context, ScriptContext.ENGINE_SCOPE);
-        executionContext.setAttribute("ruleUID", this.ruleUID, ScriptContext.ENGINE_SCOPE);
```
triller-telekom: is it correct that you remove the `ruleUID` from the global context again?

I understand the [comment](https://github.com/eclipse/smarthome/issues/3590#issuecomment-312091953) from @smerschjohann in a way that it should be available in **both** contexts, i.e. the local and the global one.

maggu2810: @triller-telekom You are right, I moved ruleUID from global namespace to ctx (so removed it from the global namespace).
#3007 contains the discussion to add ruleUID to "ctx".
#1783 moved ruleUID from "ctx" to the global namespace without any discussion or agreement (IMHO).
I assume (reading your comment) you are using "ruleUID" from the global now as I use it from "ctx".

You are right, the comment of @smerschjohann indicates that it should be also in the global namespace.

So, what should we do?

sjka: Looks like we are back at the same point where also  https://github.com/eclipse/smarthome/pull/3152#issuecomment-287660146 got stuck. If we ever want to have a chance to offer a versioned "API" to scripts, then the global namespace would be poisonous in general. But the ruleUID in particular is so elemental that we can safely accept the "risk" of having it globally as well, IMHO.

maggu2810:
> I understand the [comment](https://github.com/eclipse/smarthome/issues/3590#issuecomment-312091953) from @smerschjohann in a way that it should be available in **both** contexts, i.e. the local and the global one.

Changed.

PR #3225, opened 2017-04-11
JSonstorage without internal Gson packages

triller-telekom:
Work in progress for issue #2697

PR #3783, opened 2017-06-30
fix number deserialization on Json Storage

### LINKED ISSUE #3774: JSON storage BigDecimal scaling regression
After updating to openHAB 2.1.0 I encounter again BigDecimal scaling issues. I.e. longs are stored as decimals in the jsondb files. Because of this  the LIFX binding (and other bindings) revert to default values and log warnings like:

```
23:04:31.744 [WARN ] [inding.lifx.handler.LifxLightHandler] - Invalid value '300.0' for transition time, using default instead.
23:04:31.790 [WARN ] [inding.lifx.handler.LifxLightHandler] - Invalid value '300.0' for transition time, using default instead.
23:04:31.829 [WARN ] [inding.lifx.handler.LifxLightHandler] - Invalid value '0.0' for transition time, using default instead.
23:04:31.870 [WARN ] [inding.lifx.handler.LifxLightHandler] - Invalid value '300.0' for transition time, using default instead.
23:04:31.991 [WARN ] [inding.lifx.handler.LifxLightHandler] - Invalid value '300.0' for transition time, using default instead.
```

In `org.eclipse.smarthome.core.thing.Thing.json` the parameters show up like:
```
          "fadetime": 300.0
```

I'm using:
```
openhab> bundle:list|grep Json
136 | Active   |  80 | 0.9.0.b5               | Eclipse SmartHome Json Storage Service
```

@cdjackson originally fixed this with PR https://github.com/eclipse/smarthome/pull/2685.

sjka:
This basically re-introduces the same fallback mechanism
that was already implemented with #2685 but got lost with #3225.

fixes #3774


wborn:
Thanks for looking into this issue @sjka. I made a local build and the values are properly deserialized.

It does seem that with this solution the json files may contain wrong scaled values. I can reproduce this in the new unit test by comitting the JsonStorage and reloading it, e.g. by making the following changes to JSonStorageTest.java:

```
    private File tmpFile;

    @Before
    public void setUp() throws IOException {
        tmpFile = File.createTempFile("storage-debug", ".json");
        tmpFile.deleteOnExit();

        objectStorage = new JsonStorage<>(tmpFile, this.getClass().getClassLoader(), 0, 0, 0);
    }

...

    @Test
    public void testIntegerScale() throws IOException {
        objectStorage.put("DummyObject", new DummyObject());
        objectStorage.commitDatabase();

        objectStorage = new JsonStorage<>(tmpFile, this.getClass().getClassLoader(), 0, 0, 0);
        objectStorage.commitDatabase();

        String storageString = FileUtils.readFileToString(tmpFile);
        Assert.assertFalse("Number scaling issues in '" + tmpFile.getAbsolutePath() + "':\n" + storageString,
                storageString.contains("\"testLong\": 12.0"));

        DummyObject dummy = (DummyObject) objectStorage.get("DummyObject");

        Assert.assertEquals(((BigDecimal) dummy.myMap.get("testShort")).scale(), 0);
        Assert.assertEquals(((BigDecimal) dummy.myMap.get("testInt")).scale(), 0);
        Assert.assertEquals(((BigDecimal) dummy.myMap.get("testLong")).scale(), 0);
        Assert.assertEquals(((BigDecimal) dummy.myMap.get("testBigDecimal")).scale(), 0);
    }
```

---

This test fails with:

```
java.lang.AssertionError: Number scaling issues in '/tmp/storage-debug3424393783657734615.json':
{
  "DummyObject": {
    "class": "org.eclipse.smarthome.storage.json.test.JSonStorageTest$DummyObject",
    "value": {
      "myMap": {
        "testShort": 12.0,
        "testBoolean": true,
        "testBigDecimalWithTrailingZero": 12.0,
        "testString": "hello world",
        "testInt": 12.0,
        "testBigDecimal": 12.0,
        "testDouble": 12.12,
        "testFloat": 12.12,
        "testLong": 12.0
      }
    }
  }
}
```

bundles/storage/org.eclipse.smarthome.storage.json/src/main/java/org/eclipse/smarthome/storage/json/StringObjectMapDeserializer.java
```diff
@@ -48,4 +96,35 @@
...
+                } else if (JsonStorage.VALUE.equals(e.getKey())) {
+                    value = e.getValue();
+                }
+            }
+            try {
+                Class<?> loadedValueType = classLoader == null ? Class.forName(type) : classLoader.loadClass(type);
+                Object deserialize = context.deserialize(value, loadedValueType);
+                Map<String, Object> innerMap = new HashMap<>();
+                innerMap.put("class", type);
+                innerMap.put("value", deserialize);
```
wborn: Use JsonStorage.CLASS and JsonStorage.VALUE constants here too?

sjka:
As you correctly pointed out in #3774, this still is flawed, for two reasons:

* the same custom deserialization would be needed for Collections
* for "complex" structures (like e.g. a List in a Map, which is standard for config parameter with "multi"), Gson falls back to "Object" deserialization and there it takes a shortcut and does not consult any other deserializer for Numbers.

I think the way it was implemented (even before I tried fixing it) is conceptually broken and we need a better solution anyhow....

sjka:
Okay, next attempt with completely re-written deserialization code: I broke the deserialization into two parts. The first handles just the JsonStorage internal structure (i.e. the "outer" map in the json), and the entities only will be deserialized once the type information is available. With this, it became possible to register a `Configuration` specific deserializer, so we don't need to blindly handle all `Map`/`List` instances and pray that we don't hit one which is not part of the configuration object.

Due to the fact with this now the JsonStorage internally _always_ keeps `JsonElement`s (instead of some real entity objects and some generically deserialized ones), there also is no intermediate step necessary anymore. So Gson's ObjectTypeAdapter should have no chance anymore to interfere and mess up the numbers by reading them all in as `double`s.

Please note that I still refrained from including the hack from #2685. Assuming the serialization _is_ stable, then it would be better to keep the numbers as they were.

@wborn As you had a very good eye previously for discovering such flaws, may I ask you again for your feedback on this?

wborn:
Yes I can confirm that these changes now fix the issue. :-) Thanks a lot @sjka ! The content of the json files now remains correct after runtime restarts or when adding/removing/updating Things. The new unit tests will also help a great deal with preventing regressions.

Autocorrection in some form could be helpful for users (e.g. OH 2.1) who already have these wrong scaled values in their json files.

bundles/storage/org.eclipse.smarthome.storage.json/src/main/java/org/eclipse/smarthome/storage/json/OuterMapDeserializer.java
```diff
@@ -0,0 +1,86 @@
...
+     *
+     * It must be
+     * <li>a Map of Maps
+     * <li>with 2 entries each
+     * <li>with {@link JsonStorage#CLASS} and {@link JsonStorage#VALUE} being their keys
+     *
+     * @param obj the object to be analyzed
+     * @return {@code true} if it is the outer storage map
+     */
+    private boolean isOuterMap(JsonObject obj) {
```
htreu: I´m unsure about method ordering, do we want private methods grouped at the end of the class? or is there no ESH default?

bundles/storage/org.eclipse.smarthome.storage.json/src/main/java/org/eclipse/smarthome/storage/json/JsonStorage.java
```diff
@@ -186,8 +192,7 @@ private T deserialize(Map<String, Object> jsonValue) {
                 loadedValueType = (Class<T>) classLoader.loadClass((String) jsonValue.get(CLASS));
             }
 
-            String jsonString = mapper.toJson(jsonValue.get("value"));
-            value = mapper.fromJson(jsonString, loadedValueType);
+            value = innerMapper.fromJson((JsonElement) jsonValue.get(VALUE), loadedValueType);
```
htreu: 'jsonValue'  is defined in the method declaration but is actually the 'innerMap'. Its a little bit strange to have a Map called jsonValue (I comment here because its not part of your PR). Can we come up with a better name?

bundles/storage/org.eclipse.smarthome.storage.json/src/main/java/org/eclipse/smarthome/storage/json/JsonStorage.java
```diff
@@ -203,10 +208,10 @@ private T deserialize(Map<String, Object> jsonValue) {
 
             FileReader reader = new FileReader(inputFile);
             Map<String, Map<String, Object>> type = new HashMap<String, Map<String, Object>>();
```
htreu: `type` is only created to retrieve the Class from it later. Can´t we just use `HashMap.class` in line 211? Wouldnt the generics be erased from the type during runtime anyway?

bundles/storage/org.eclipse.smarthome.storage.json.test/.classpath
```diff
@@ -1,6 +1,5 @@
 <?xml version="1.0" encoding="UTF-8"?>
 <classpath>
-	<classpathentry kind="src" path="src/main/groovy"/>
```
sjka: good catch, sure!

bundles/storage/org.eclipse.smarthome.storage.json.test/src/main/java/org/eclipse/smarthome/storage/json/JsonStorageServiceOSGiTest.java
```diff
@@ -0,0 +1,125 @@
...
+
+    @Test
+    public void testSerialization() {
+        assertThat(storage.getKeys().size(), is(0));
+
+        storage.put("Key1", new PersistedItem("String", Arrays.asList("LIGHT", "GROUND_FLOOR")));
+        storage.put("Key2", new PersistedItem("Number", Arrays.asList("TEMPERATURE", "OUTSIDE")));
+        assertThat(storage.getKeys().size(), is(2));
+
+        storage.get("Key1");
```
sjka: No idea. Groovy-to-Java conversion apparently throttles down the brain 😉

bundles/storage/org.eclipse.smarthome.storage.json/src/main/java/org/eclipse/smarthome/storage/json/JsonStorage.java
sjka: No, it makes a difference. The type information actually _is_ available and gets used (see [%24Gson%24Types.java#L317](https://github.com/google/gson/blob/a300148003e3a067875b1444e8268b6e0f0e0e02/gson/src/main/java/com/google/gson/internal/%24Gson%24Types.java#L317))

bundles/storage/org.eclipse.smarthome.storage.json/src/main/java/org/eclipse/smarthome/storage/json/OuterMapDeserializer.java
sjka: I don't think we have any conventions about this (yet?)

sjka:
Thanks for the good feedback! I will update the PR.

bundles/storage/org.eclipse.smarthome.storage.json/src/main/java/org/eclipse/smarthome/storage/json/JsonStorage.java
sjka: yes. Which makes me asking myself why the internal data structure carrying class & value has to be a map at all. It would be much less confusing if it were a dedicated class. Let me quickly check what it will feel like...

bundles/storage/org.eclipse.smarthome.storage.json/src/main/java/org/eclipse/smarthome/storage/json/JsonStorage.java
```diff
@@ -223,15 +228,17 @@ private File getBackupFile(int age) {
         File[] files = folder.listFiles();
 
         // Get an array of file times from the filename
-        int count = files.length;
-        for (int i = 0; i < count; i++) {
-            if (files[i].isFile()) {
-                String[] parts = files[i].getName().split(SEPARATOR);
-                if (parts.length != 2 || !parts[1].equals(file.getName())) {
-                    continue;
+        if (files != null) {
```
sjka: done - although github didn't realize 😱

htreu:
@maggu2810 I guess this is good to go. Can you merge this?

bundles/storage/org.eclipse.smarthome.storage.json/src/main/java/org/eclipse/smarthome/storage/json/JsonStorage.java
```diff
@@ -117,11 +121,8 @@ public JsonStorage(File file, ClassLoader classLoader, int maxBackupFiles, int w
 
     @Override
     public T put(String key, T value) {
-        Map<String, Object> val = new LinkedHashMap<String, Object>();
-        val.put(CLASS, value.getClass().getName());
-        val.put(VALUE, value);
-
-        Map<String, Object> previousValue = map.get(key);
+        StorageEntry val = new StorageEntry(value.getClass().getName(), entityMapper.toJsonTree(value));
+        StorageEntry previousValue = map.get(key);
```
maggu2810: Shouldn't put already return the previous value?
So, let's use:

```Java
final StorageEntry previousValue = map.put(key, val);
```

bundles/storage/org.eclipse.smarthome.storage.json/src/main/java/org/eclipse/smarthome/storage/json/ConfigurationDeserializer.java
```diff
@@ -0,0 +1,84 @@
...
+ *
+ */
+public class ConfigurationDeserializer implements JsonDeserializer<Configuration> {
+
+    @Override
+    public Configuration deserialize(JsonElement json, Type typeOfT, JsonDeserializationContext context)
+            throws JsonParseException {
+        Configuration configuration = new Configuration();
+        JsonObject configurationObject = json.getAsJsonObject();
+        JsonObject propertiesObject = configurationObject.get("properties").getAsJsonObject();
```
maggu2810: Shouldn't we be "save" and test if `configurationObject.get("properties")` returns `null`?

sjka:
Yes and Yes. PR is updated.

Issue #3786, opened 2017-07-03
Trigger a discovery for all installed bindings at once


afuechsel:
Currently it is only possible, to trigger a discovery for a single bindig id: 

```` 
POST /rest/discovery/bindings/{bindingId}/scan
```` 

It would be useful to have a chance to trigger the discovery for all available bindings (those that are listed with `GET /discovery`). For example: 

```` 
POST /rest/discovery/scan
````

sjka:
What problem are you trying to address with this? Or rather: what is the use-case? 

Do you really think it is a good idea to start the active discovery always for _all_ existing bindings _at the same time_? Think of bindings which cause lots of network traffic/broadcasts and others which put their controllers into inclusion mode for scanning. 

I'd assume that bindings which support scanning for new devices which can do that without causing much harm already support background discovery. And others intentionally don't implement it, because it is "expensive" in some way to run the discovery. 

Of course you can argue that it's simply not wise to hit such a "search all" button - but I'm not convinced that it is even most of the time a good idea, having my neighbors joining my zigbee networks just because I wanted to add e.g. a new homematic device.

afuechsel:
I would agree to "it's simply not wise to hit such a 'search all' button". The basic idea is that sometimes one is not willing to wait for the background discovery to kick in and hopefully find all things and it would be better to actively trigger a "search all" tomake sure everything available is really in the inbox. 

The use case behind is a search for all bridges of several bindings at once.

sjka:
What about e.g. https://github.com/eclipse/smarthome/issues/3605#issuecomment-312574411? You answer simply ignored my concern about this not being a good idea even most of the time. 

> it would be better

I'm simply not convinced this actually _is_ "better".

afuechsel:
I did not ignore your concern. I simply see no other way to implement a full refresh of the inbox with all installed bindings. Currently I have installed 16 bindings supporting discovery and my PR just triggers all 16 discovery services that are all gracefully executed and refresh the inbox. I don't see any reason why we should *not* implement such a feature, only because it might be "expensive".

rodneyrehm:
> What problem are you trying to address with this? Or rather: what is the use-case?

basically a forced refresh of the inbox triggered by qivicon UI. We've been doing this since the inbox was introduced (via internal APIs) and are simply migrating that to the REST world. We also allow the user to trigger this refresh, should they not have found what they were hoping to pair.

[...]

PR #3787, opened 2017-07-03
Added a method to trigger the discovery to all installed bindings at …

afuechsel:
…once.

Fixes: #3786


afuechsel:
I will reject this PR then.

PR #3795, opened 2017-07-04
Break circle dependency in GroupItem members

### LINKED ISSUE #3794: java.lang.StackOverflowError in GroupItem.collectMembers()
The PR #3464 introduced the possibility to retrieve also sub groups from a group item.

However there was a check missing that leads to a StackOverflowError as reported in the [KNX Forum](https://knx-user-forum.de/forum/supportforen/openhab/1108731-openhabian-teilweise-absturz-openhab-2-1)

triller-telekom:
Fixes #3794

### DIFF
--- a/bundles/core/org.eclipse.smarthome.core/src/main/java/org/eclipse/smarthome/core/items/GroupItem.java
+++ b/bundles/core/org.eclipse.smarthome.core/src/main/java/org/eclipse/smarthome/core/items/GroupItem.java
@@ -116,11 +116,12 @@ public class GroupItem extends GenericItem implements StateChangeListener {
 
     private void collectMembers(Set<Item> allMembers, Set<Item> members) {
         for (Item member : members) {
+            if (allMembers.contains(member)) {
+                continue;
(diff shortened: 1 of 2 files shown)


bundles/core/org.eclipse.smarthome.core/src/main/java/org/eclipse/smarthome/core/items/GroupItem.java
```diff
@@ -116,6 +116,9 @@ public GroupFunction getFunction() {
 
     private void collectMembers(Set<Item> allMembers, Set<Item> members) {
         for (Item member : members) {
+            if (allMembers.contains(member)) {
+                return;
```
sjka: how about `continue` instead of `return`? 😉

triller-telekom: Just noticed that and changed the code but your `continue` is even better ;)

PR #3804, opened 2017-07-05
Automation: Triggers may specify an optional event

### LINKED ISSUE #3713: Second parameter on core.ChannelEventTrigger to allow for filtering on the payload
EventChannels can generate numerous different events. 
Currently it is not possible to specify which one of them should trigger an event.

In order to be able to distinguish between the different events one would need a filter on the payload. Please consider adding a second parameter on core.ChannelEventTrigger which can be used as payload filter.

triller-telekom:
The old rule engine supported specifying an event on a trigger channel
to have the rule fire only if this event occured on that trigger
channel. With this patch the new automation engine supports this feature
too.

Fixes #3713


bundles/automation/org.eclipse.smarthome.automation.module.core/src/main/java/org/eclipse/smarthome/automation/module/core/handler/GenericEventTriggerHandler.java
```diff
@@ -87,6 +91,15 @@ public void receive(Event event) {
             if (!event.getTopic().contains(source)) {
                 return;
             }
+            // if the optional parameter "eventOnChannel" is set different from the defaultValue (""), compare it to the
+            // received event on the channel
+            if (eventOnChannel != null && !eventOnChannel.equals("")) {
+                if (event instanceof ChannelTriggeredEvent) {
```
htreu: line 96 to 98 can be reduced to `if (StringUtils.isNotBlank(eventOnChannel) && event instanceof ChannelTriggeredEvent && !eventOnChannel.equals(((ChannelTriggeredEvent) event).getEvent())) {`

This may also be a private method `private boolean eventMatchesChannelConfiguration(event)` so it will only be one if clause around the early return.

danchom: I think it is not good idea to add dependencies to the  third party libs for such check.

bundles/automation/org.eclipse.smarthome.automation.module.core/src/main/java/org/eclipse/smarthome/automation/module/core/handler/GenericEventTriggerHandler.java
```diff
@@ -52,13 +54,15 @@
     private static final String CFG_EVENT_TOPIC = "eventTopic";
     private static final String CFG_EVENT_SOURCE = "eventSource";
     private static final String CFG_EVENT_TYPES = "eventTypes";
+    private static final String CFG_CHANNEL_EVENT = "eventOnChannel";
```
danchom: This properties is specific for the channel events and there is no place for it in this class. If you want more specific handler it is better to extends this one and add your property and apply it to the Then the core.ChannelEventTrigger.  
Actually GenericEventTriggerHandler implements EventFilter which at the moment just check this:
    @Override
    public boolean apply(Event event) {
        logger.trace("->FILTER: {}:{}", event.getTopic(), source);
        return event.getTopic().contains(source);
    }
May be better approach is to add filter configuration property and change the apply method to check for it.
For example:
  public boolean apply(Event event) {
        logger.trace("->FILTER: {}:{}", event.getTopic(), source);
         String topic = event.getTopic();
        return topic.contains(source) && (filter== null || topic.contains(filter));
    }
where the filter can be your eventOnChannel
Is this approach suitable is for you?

triller-telekom:
@danchom Thank you for your feedback. I was also wondering what is the best way to integrate this new feature. I have now created a new module which extends the `GenericHandler` and filters within the `apply` function. Please have a look on it again. Thanks!

bundles/automation/org.eclipse.smarthome.automation.module.core/src/main/java/org/eclipse/smarthome/automation/module/core/factory/CoreModuleHandlerFactory.java
```diff
@@ -44,10 +45,11 @@
 
     private Logger logger = LoggerFactory.getLogger(CoreModuleHandlerFactory.class);
 
-    private static final Collection<String> types = Arrays.asList(
-            new String[] { ItemCommandTriggerHandler.MODULE_TYPE_ID, ItemStateTriggerHandler.UPDATE_MODULE_TYPE_ID,
-                    ItemStateTriggerHandler.CHANGE_MODULE_TYPE_ID, ItemStateConditionHandler.ITEM_STATE_CONDITION,
-                    ItemCommandActionHandler.ITEM_COMMAND_ACTION, GenericEventTriggerHandler.MODULE_TYPE_ID,
+    private static final Collection<String> types = Arrays
```
sjka: There is a shortcut to this. Instead of writing

 ```java
Arrays.asList(new String[] {"", ...})
```

you can actually write

```java
Arrays.asList("", ...)
```

bundles/automation/org.eclipse.smarthome.automation.module.core/src/main/java/org/eclipse/smarthome/automation/module/core/handler/ChannelPayloadTriggerHandler.java
```diff
@@ -0,0 +1,55 @@
...
+import org.apache.commons.lang.StringUtils;
+import org.eclipse.smarthome.automation.Trigger;
+import org.eclipse.smarthome.core.events.Event;
+import org.eclipse.smarthome.core.thing.events.ChannelTriggeredEvent;
+import org.osgi.framework.BundleContext;
+import org.slf4j.Logger;
+import org.slf4j.LoggerFactory;
+
+/**
+ * This is an ModuleHandler implementation for Trigger Channel with specific events
```
sjka: [...] for trigger channel**s** with [...]

bundles/automation/org.eclipse.smarthome.automation.module.core/src/main/java/org/eclipse/smarthome/automation/module/core/handler/ChannelPayloadTriggerHandler.java
```diff
@@ -0,0 +1,55 @@
...
+ *
+ * @author Stefan Triller - Initial contribution
+ *
+ */
+public class ChannelPayloadTriggerHandler extends GenericEventTriggerHandler {
+
+    private final Logger logger = LoggerFactory.getLogger(ChannelPayloadTriggerHandler.class);
+
+    private String eventOnChannel;
```
sjka: private **final** String

bundles/automation/org.eclipse.smarthome.automation.module.core/META-INF/MANIFEST.MF
```diff
@@ -5,6 +5,7 @@ Bundle-SymbolicName: org.eclipse.smarthome.automation.module.core
 Bundle-Version: 0.9.0.qualifier
 Bundle-RequiredExecutionEnvironment: JavaSE-1.8
 Import-Package: com.google.common.collect,
+ org.apache.commons.lang,
```
triller-telekom: I am actually using this because its very convenient here: `StringUtils.isNotBlank(eventOnChannel)`

In case the user enter multiple blanks, i.e. "   " into paperUi in the event field, he will notice that and then might wonder why this trigger does not fire. The function makes a `null` check obsolete too.

sjka: Sure, it fully makes sense. 

As the automation parts happen to be used in other products as well, I just want to be sure this does not add any additional burden to the maintainers.

bundles/automation/org.eclipse.smarthome.automation.module.core/ESH-INF/automation/moduletypes/ChannelTrigger.json
```diff
@@ -18,16 +18,25 @@
...
                   }
                ]
+            },
+            {
+               "name":"eventOnChannel",
+               "type":"TEXT",
+               "label":"Event",
+               "description":"the event on the channel to react on",
+               "required":false,
+               "defaultValue":""
```
kaikreuzer: Why do you define a default? It is an optional parameter anyhow, so if not provided, it should be rather null.

bundles/automation/org.eclipse.smarthome.automation.module.core/ESH-INF/automation/moduletypes/ChannelTrigger.json
```diff
@@ -18,16 +18,25 @@
                      "name": "kind"
                   }
                ]
+            },
+            {
+               "name":"eventOnChannel",
```
kaikreuzer: Why not merely "event"?

bundles/automation/org.eclipse.smarthome.automation.module.core/ESH-INF/automation/moduletypes/ChannelTrigger.json
```diff
@@ -18,16 +18,25 @@
...
+               "type":"TEXT",
+               "label":"Event",
+               "description":"the event on the channel to react on",
+               "required":false,
+               "defaultValue":""
             }
          ],
          "children":[
```
kaikreuzer: Please completely remove the "children" part and instead implement a handler for `core.ChannelEventTrigger`.

bundles/automation/org.eclipse.smarthome.automation.module.core/src/main/java/org/eclipse/smarthome/automation/module/core/handler/ChannelPayloadTriggerHandler.java
```diff
@@ -0,0 +1,55 @@
+/**
+ * Copyright (c) 1997, 2015 by ProSyst Software GmbH and others.
```
kaikreuzer: no, "by Deutsche Telekom AG and others".

bundles/automation/org.eclipse.smarthome.automation.module.core/src/main/java/org/eclipse/smarthome/automation/module/core/handler/ChannelPayloadTriggerHandler.java
```diff
@@ -0,0 +1,55 @@
...
+import org.slf4j.Logger;
+import org.slf4j.LoggerFactory;
+
+/**
+ * This is an ModuleHandler implementation for Trigger Channel with specific events
+ *
+ * @author Stefan Triller - Initial contribution
+ *
+ */
+public class ChannelPayloadTriggerHandler extends GenericEventTriggerHandler {
```
kaikreuzer: Please don't extend `GenericEventTriggerHandler`, but do it similar to `ItemStateTriggerHandler`.
Rename to `ChannelEventTriggerHandler`.

kaikreuzer: Please don't extend `GenericEventTriggerHandler`, but do it similar to `ItemStateTriggerHandler`.
Rename to `ChannelEventTriggerHandler`.

bundles/automation/org.eclipse.smarthome.automation.module.core/META-INF/MANIFEST.MF
kaikreuzer: Also remember the discussion about Commons Lang3, so whereever possible, I would refrain from using the lib at places where we do not yet have any dependency on it.

danchom: I missed the mf file, sorry, I've seen that the usage of StringUtils was remove from code.

bundles/automation/org.eclipse.smarthome.automation.module.core/src/main/java/org/eclipse/smarthome/automation/module/core/handler/ChannelPayloadTriggerHandler.java
kaikreuzer: Why didn't you change it as I had suggested?

triller-telekom: This file is not part of this PR anymore. I now have renamed the module to `ChannelEventTriggerHandler.java` and this one has in the license information: "Copyright (c) 1997, 2015 by the respective copyright holders." like the other new files which are created.

kaikreuzer: Right, that is why I asked why you didn't change it to "by Deutsche Telekom AG and others". It should be the same as e.g. https://github.com/eclipse/smarthome/blob/b3149975fc65dfad6dbd74bf35478bb7f9a786a5/bundles/automation/org.eclipse.smarthome.automation.module.media/src/main/java/org/eclipse/smarthome/automation/module/media/internal/MediaScriptScopeProvider.java#L2.

bundles/automation/org.eclipse.smarthome.automation.module.core/src/main/java/org/eclipse/smarthome/automation/module/core/handler/ChannelEventTriggerHandler.java
```diff
@@ -0,0 +1,120 @@
...
+                properties);
+    }
+
+    @Override
+    public void receive(Event event) {
+        if (ruleEngineCallback != null) {
+            logger.trace("Received Event: Source: {} Topic: {} Type: {}  Payload: {}", event.getSource(),
+                    event.getTopic(), event.getType(), event.getPayload());
+
+            Map<String, Object> values = Maps.newHashMap();
```
maggu2810: Why do we import `com.google.common.collect.Maps` and use `Maps.newHashMap()` instead of simply using `new HashMap()`? What's the difference?

triller-telekom: I guess there is no difference but the import `com.google.common.collect.Maps` was available so I used it. Shall I create a follow-up PR and change it?

kaikreuzer: Yes, please!

triller-telekom: BTW: That was copy&paste from `GenericEventTriggerHandler`. I will fix is there as well...

PR #3806, opened 2017-07-05
Hue: fix #3805

### LINKED ISSUE #3805: Hue binding: bad item type for channel type "switch"
https://github.com/eclipse/smarthome/blob/master/extensions/binding/org.eclipse.smarthome.binding.hue/ESH-INF/thing/channels.xml#L35

The value for tag "item-type" is incorrectly set to `SwitchItem `rather than `Switch`.

lolodomo:

### DIFF
--- a/extensions/binding/org.eclipse.smarthome.binding.hue/ESH-INF/thing/channels.xml
+++ b/extensions/binding/org.eclipse.smarthome.binding.hue/ESH-INF/thing/channels.xml
@@ -32,7 +32,7 @@
 
 	<!-- Switch Channel -->
     <channel-type id="switch">
-        <item-type>Switch/item-type>
+        <item-type>Switch</item-type>
         <label>Switch</label>
         <description>The switch channel allows to switch the light on and off.
         </description>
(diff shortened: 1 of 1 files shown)


sjka:
As it seems, the ip validation tool checks for the "committer" in git as well (which is some garbage when creating commits via Github's UI). The rest is okay though (author, signed-off) and I manually checked that the ECA is valid. The committer will be overridden anyway when squash-merging the PR, so it's safe to overrule the failed check.

extensions/binding/org.eclipse.smarthome.binding.hue/ESH-INF/thing/channels.xml
```diff
@@ -32,7 +32,7 @@
 
 	<!-- Switch Channel -->
     <channel-type id="switch">
-        <item-type>SwitchItem</item-type>
+        <item-type>Switch/item-type>
```
sjka: the `<` of the closing tag got lost

lolodomo:
Very sorry for that, that is the risk when just editing in Git. That is now fixed.

sjka:
Haha, no problem - it took me a few moments too to figure out what exactly made travis fail 😄

sjka:
Thanks!

sjka:
Thanks!

PR #3807, opened 2017-07-05
[astro] fix invalid event creation

htreu:
In case your location does imply not having a valid Day or Night range the creation of trigger events failed. This PR fixes this.


sjka:
Thanks!

PR #3810, opened 2017-07-06
Better debug output if framework tries to set a wrong state

triller-telekom:


sjka:
```
Tried to set invalid state 100.0 (StringType) on item autoUP of type RollershutterItem, ignoring it
```

Much better, indeed. Thanks!

PR #3816, opened 2017-07-07
[LIFX] Update ThingStatus at most once a second

wborn:
With this PR the `LifxLightHandler` will only update the ThingStatus when there is a change.

Each received message currently always results in a `ThingStatus.ONLINE` update. This causes a lot of unnecessary `ThingStatusInfoEvent` s, logging etc. which generate unnecessary CPU load.

### DIFF
--- a/extensions/binding/org.eclipse.smarthome.binding.lifx/src/main/java/org/eclipse/smarthome/binding/lifx/handler/LifxLightHandler.java
+++ b/extensions/binding/org.eclipse.smarthome.binding.lifx/src/main/java/org/eclipse/smarthome/binding/lifx/handler/LifxLightHandler.java
@@ -10,6 +10,9 @@ package org.eclipse.smarthome.binding.lifx.handler;
 import static org.eclipse.smarthome.binding.lifx.LifxBindingConstants.*;
 import static org.eclipse.smarthome.binding.lifx.internal.LifxUtils.increaseDecreasePercentType;
 
+import java.math.BigDecimal;
+import java.time.Duration;
+import java.time.LocalDateTime;
 import java.util.ArrayList;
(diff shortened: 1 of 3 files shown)


kaikreuzer:
It isn't recommended for bindings to do this check. The framework already checks whether there is a change or not and actually issues different events (updated vs changed). The logging can be configured to only log change events (that's the default in openHAB), so additional updates should do no harm.
By filtering it, you now suppress the information on when the light has been seen online last (we do not show this anywhere yet, but I would like to have such a feature in future).

So I vote against this PR, sorry 😎

wborn:
When my 38 lights just remain online the light state is continuously polled. Each received message results in an online update. This PR saves with my 38 lights about ~1535 updates per minute.

When you start sending commands to your lights for some animation this number will rapidly increase because more messages are received.

The online/offline status works very well for LIFX lights. So if you want to know when the light was last online you can still use the transition from online to offline for this.

kaikreuzer:
> ~1535 updates per minute.

Ok, that's a lot indeed :-)

>  you can still use the transition from online to offline for this

Not, if it hasn't gone offline at all. But ok, if you say that an ONLINE status cannot be older than a second in any case (as otherwise it would have gone OFFLINE already), my argument probably does not matter to much here.
Nonetheless: With your change, we would see a "last ONLINE message" which has a date from when it went online instead of "now-1s". So could you find a way to at least regularly do a setStatus, even if it hasn't changed? Doing it once a second would still reduce the number of messages from 1535 to only 38 :-)

wborn:
@kaikreuzer with the new commit it will now again allow for same status updates with a 1 second interval. I've also replaced some durations to use the Java 8 Duration so the units are more clear.

PR #3814, opened 2017-07-07
Simplify construction of StringListType

holmes:
It'd be really nice to be able to pass a List<String> to construct a
StringListType. A parameterized array is great for testing but not
that useful in real life.

Signed-off-by: Jason Holmes holmes.j@gmail.com (github: holmes)

PR #3817, opened 2017-07-07
Simplify construction of StringListType(s)

holmes:
It'd be really nice to be able to pass a List<String> to construct a
StringListType. A parameterized array is great for testing but not
that useful in real life.

Signed-off-by: Jason Holmes <holmes.j@gmail.com>

### DIFF
--- a/bundles/core/org.eclipse.smarthome.core/src/main/java/org/eclipse/smarthome/core/library/types/StringListType.java
+++ b/bundles/core/org.eclipse.smarthome.core/src/main/java/org/eclipse/smarthome/core/library/types/StringListType.java
@@ -12,8 +12,9 @@ import java.util.Arrays;
 import java.util.Collections;
 import java.util.Formatter;
 import java.util.List;
+import java.util.function.Function;
+import java.util.stream.Collectors;
 
-import org.apache.commons.lang.StringUtils;
 import org.eclipse.smarthome.core.types.Command;
 import org.eclipse.smarthome.core.types.State;
 
@@ -37,11 +38,14 @@ public class StringListType implements Command, State {
         typeDetails = Collections.emptyList();
     }
 
+    public StringListType(List<String> rows) {
+        typeDetails = new ArrayList<>(rows);
+    }
+
     public StringListType(StringType... rows) {
-        typeDetails = new ArrayList<String>(rows.length);
-        for (StringType row : rows) {
-            typeDetails.add(row.toString());
-        }
+        typeDetails = Arrays.stream(rows)
+                .map(StringType::toString)
+                .collect(Collectors.toList());
     }
 
     public StringListType(String... rows) {
@@ -49,15 +53,13 @@ public class StringListType implements Command, State {
     }
 
     /**
-     * Deserialize the input string,
-     * splitting it on every delimiter not preceeded by a backslash
+     * Deserialize the input string, splitting it on every delimiter not preceded by a backslash.
      */
     public StringListType(String serialized) {
-        String[] rows = serialized.split(REGEX_SPLITTER);
-        typeDetails = new ArrayList<String>(rows.length);
-        for (String row : rows) {
-            typeDetails.add(row.replace(ESCAPED_DELIMITER, DELIMITER));
-        }
+        typeDetails = Arrays.stream(
+                serialized.split(REGEX_SPLITTER))
+                .map(s -> s.replace(ESCAPED_DELIMITER, DELIMITER))
+                .collect(Collectors.toList());
     }
 
     public String getValue(final int index) {
@@ -90,11 +92,9 @@ public class StringListType implements Command, State {
 
     @Override
     public String toFullString() {
-        List<String> parts = new ArrayList<>(typeDetails.size());
-        for (String row : typeDetails) {
-            parts.add(row.replace(DELIMITER, ESCAPED_DELIMITER));
-        }
-        return StringUtils.join(parts, DELIMITER);
+        return typeDetails.stream()
+                .map(s -> s.replace(DELIMITER, ESCAPED_DELIMITER))
+                .collect(Collectors.joining(DELIMITER));
     }
 
     public static StringListType valueOf(String value) {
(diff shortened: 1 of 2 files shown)


holmes:
Updated version of https://github.com/eclipse/smarthome/pull/3814

I ended up re-forking and doing a bunch of things to try and change the committer (not the author). Found a way.

holmes:
@htreu any chance you can approve and merge this? I'd love to use it in a PR in openhab.

Thanks!

bundles/core/org.eclipse.smarthome.core/src/main/java/org/eclipse/smarthome/core/library/types/StringListType.java
```diff
@@ -38,12 +38,16 @@ public StringListType() {
...
     public StringListType(StringType... rows) {
-        typeDetails = new ArrayList<String>(rows.length);
+        typeDetails = new ArrayList<>(rows.length);
         for (StringType row : rows) {
             typeDetails.add(row.toString());
         }
     }
 
+    public StringListType(List<String> rows) {
+        this.typeDetails = new ArrayList<>(rows);
```
htreu: Would it be possible to let the convenience constructors delegate to one constructor which is responsible for setting the `typeDetails`? My suggestion is to use your new c`tor and let all the others delegate to it.

bundles/core/org.eclipse.smarthome.core/src/main/java/org/eclipse/smarthome/core/library/types/StringListType.java
```diff
@@ -38,12 +38,16 @@ public StringListType() {
     }
 
     public StringListType(StringType... rows) {
-        typeDetails = new ArrayList<String>(rows.length);
+        typeDetails = new ArrayList<>(rows.length);
         for (StringType row : rows) {
```
htreu: not part of your PR but this could well be rewritten to use streams: `Arrays.stream(rows).map(r -> r.toString()).collect(Collectors.asList())` and feed this into your c`tor. wdyt?

bundles/core/org.eclipse.smarthome.core/src/main/java/org/eclipse/smarthome/core/library/types/StringListType.java
holmes: sure thing, delegated them all.

bundles/core/org.eclipse.smarthome.core/src/main/java/org/eclipse/smarthome/core/library/types/StringListType.java
holmes: used streams throughout.

bundles/core/org.eclipse.smarthome.core/src/main/java/org/eclipse/smarthome/core/library/types/StringListType.java
```diff
@@ -90,15 +87,13 @@ public String toString() {
...
     @Override
     public String toFullString() {
-        List<String> parts = new ArrayList<>(typeDetails.size());
-        for (String row : typeDetails) {
-            parts.add(row.replace(DELIMITER, ESCAPED_DELIMITER));
-        }
-        return StringUtils.join(parts, DELIMITER);
+        return typeDetails.stream()
+            .map(s -> s.replace(DELIMITER, ESCAPED_DELIMITER))
+            .collect(Collectors.joining(",'"));
```
sjka: looks like a little extra `' ` slipped in here.
Please also use the `DELIMITER` constant.

bundles/core/org.eclipse.smarthome.core/src/main/java/org/eclipse/smarthome/core/library/types/StringListType.java
```diff
@@ -37,27 +36,25 @@ public StringListType() {
...
-            typeDetails.add(row.toString());
-        }
+        this(Arrays.stream(rows)
+            .map(StringType::toString)
+            .collect(Collectors.toList()));
     }
 
     public StringListType(String... rows) {
-        typeDetails = Arrays.asList(rows);
+        this(Arrays.asList(rows));
```
sjka: What I don't like about this approach is that we are now creating a collection twice. Therefore I'd prefer the previous approach with directly assigning the newly created collection to `typeDetails`.

bundles/core/org.eclipse.smarthome.core/src/main/java/org/eclipse/smarthome/core/library/types/StringListType.java
```diff
@@ -37,27 +36,25 @@ public StringListType() {
...
+    public StringListType(List<String> rows) {
+        this.typeDetails = new ArrayList<>(rows);
+    }
+
     public StringListType(StringType... rows) {
-        typeDetails = new ArrayList<String>(rows.length);
-        for (StringType row : rows) {
-            typeDetails.add(row.toString());
-        }
+        this(Arrays.stream(rows)
```
sjka: Please avoid creating a collection twice, i.e. assign the result of the collector directly to `typeDetails` instead of passing it to the other constructor.

holmes:
Updated.

sjka:
Looking at the test results, something still does not seem right:

```
Failed tests: 
  StringListTypeTest.testValueOf:78 expected:<a[],b> but was:<a[\],b>
  StringListTypeTest.testEquals:35 null
```

holmes:
Fixed. I even updated the tests and ran them locally.

I forgot to replace the escaped delimiters w/ the regular delimiter in the serialized constructor.

sjka:
> I even updated the tests and ran them locally.

That just earned you some bonus points 👍 
Thanks a lot!

sjka:
PS: fyi in case you don't already know it: It will need a little while until an ESH version including your change flows over to openHAB snapshots. Here is how you can track it: https://github.com/eclipse/smarthome/wiki/ESH-used-in-openHAB

PR #3818, opened 2017-07-08
[Classic Iconset] Add figcaptions

ThomDietrich:
Add HTML5 figure and figcaption elements to the iconset preview:

grafik

You'll find that the style is inline. I promise to move it to css as soon as style.css is unified between ESH and openhab-docs. If you insist I will look into it now.

Signed-off-by: Thomas Dietrich <Thomas.Dietrich@tu-ilmenau.de>


kaikreuzer:
Although it is more than 1000 loc, we can merge it as most of the changes are only due to generated content.

Issue #3741, opened 2017-06-25
Considering a default pattern for string and number channels


lolodomo:
All the string channels provided by the Sonos binding are defined without any pattern. As the result, the channel value is not displayed in UI if you use the auto-link mode of Paper UI and you forgot to define a label with pattern "%s" in your sitemap.

I propose to add a default pattern "%s" to these string channels.
@kgoderis @kaikreuzer : are you ok with that proposal ?

lolodomo:
Any feedback ?

kgoderis:
Ok for me

sjka:
makes sense

lolodomo:
I have done the changes for the Sonos binding but in fact almost all bindings should be fixed !.

lolodomo:
Finally I think the good approach was my PR #3262 that would define a default pattern whatever the channel (all bindings). I just have to adjust the test to be sure to add a default pattern only when expected.

lolodomo:
To be clear, here is my general proposal that will apply to all bindings and all UIs.
When no pattern is defined in label from sitemap, no pattern is defined in label from tem and no pattern is defined in channel, if the channel is associated to a string or number item, the UI item registry uses a default pattern to return the item state in the label.
For string item, the default pattern would be naturally "%s". For number item, I propose "%.0f".

I ping @kaikreuzer as he was worried when I first try to fix that.

sjka:
Hmmm, I'm a little undecided....

Pros:
* some bindings define patterns, some don't. This proposal just fosters consistency from a user's perspective
* for automatically created items this certainly makes sense (as you pointed out in #3262)
* To my personal experience this is the by far more common use-case than _not_ displaying a state

Cons:
* for hand-crafted `*.items` files it feels wrong to mess with whatever the user defined
* it turns the current default behavior upside-down. So people _not_ wanting to display a state at a place will be disappointed

So I'm pretty confident this sounds like a good idea to do in the `ChannelItemProvider`. This would also mean it does not mess with user's configurations and will apply only to items which are linked to channels. And as @kaikreuzer already pointed out: "[...] that there are many valid use cases where we do not want the state to be provided/displayed; we should make sure that this is not broken".  

I'm not so sure though about the implications of doing it for all such items in the UI layer. For this, I don't have a complete enough picture on how people are using them currently, which use-cases will be broken and how common they are.

lolodomo:
[...]

PR #3822, opened 2017-07-10
UI: provide a default pattern for string and number channels if not defined

lolodomo:
Fix issue #3741


bundles/core/org.eclipse.smarthome.core/src/main/java/org/eclipse/smarthome/core/items/GenericItem.java
```diff
@@ -401,15 +403,44 @@ public StateDescription getStateDescription() {
...
+                    if (stateDescription.isReadOnly()) {
+                        readOnly = true;
+                    }
+                    if (options == null) {
+                        options = stateDescription.getOptions();
+                    }
                 }
             }
+            if (found) {
+                fullStateDescription = new StateDescription(minimum, maximum, step, pattern, readOnly, options);
```
sjka: The idea here is to mix up state descriptions of different providers, right? 
This will lead to quite unexpected results, therefore I don't think this is a good idea. Also I wouldn't want to mix this into a PR which intends to add a default state description at all (or I did not understand how it is related).

bundles/ui/org.eclipse.smarthome.ui/src/main/java/org/eclipse/smarthome/ui/internal/items/ItemUIRegistryImpl.java
```diff
@@ -307,6 +309,14 @@ public String getLabel(Widget w) {
                             }
                         } else {
                             state = item.getState();
+                            if (stateDescription != null && stateDescription.getOptions() != null) {
```
sjka: How are the changes within this file related to the PR? I would not have expected any changes needed in the UI layer for this.

bundles/core/org.eclipse.smarthome.core.thing/src/main/java/org/eclipse/smarthome/core/thing/internal/ChannelStateDescriptionProvider.java
```diff
@@ -57,7 +61,29 @@ public StateDescription getStateDescription(String itemName, Locale locale) {
...
+                    StateDescription state = channelType.getState();
+                    if ((channelType.getItemType() != null) && ((state == null) || (state.getPattern() == null))) {
+                        String pattern = null;
+                        if (channelType.getItemType().equalsIgnoreCase("String")) {
+                            pattern = "%s";
+                        } else if (channelType.getItemType().equalsIgnoreCase("Number")) {
+                            pattern = "%.0f";
+                        }
+                        if (pattern != null) {
+                            logger.debug("Provide a default pattern {} for item {}", pattern, itemName);
```
sjka: please reduce it to trace level

extensions/ui/org.eclipse.smarthome.ui.basic/src/main/java/org/eclipse/smarthome/ui/basic/internal/render/AbstractWidgetRenderer.java
```diff
@@ -83,9 +83,11 @@ protected String preprocessSnippet(String snippet, Widget w) {
         snippet = StringUtils.replace(snippet, "%widget_id%", itemUIRegistry.getWidgetId(w));
         snippet = StringUtils.replace(snippet, "%icon_type%", config.getIconType());
         snippet = StringUtils.replace(snippet, "%item%", w.getItem() != null ? w.getItem() : "");
-        snippet = StringUtils.replace(snippet, "%label%", getLabel(w));
-        snippet = StringUtils.replace(snippet, "%value%", getValue(w));
-        snippet = StringUtils.replace(snippet, "%has_value%", new Boolean(hasValue(w)).toString());
+        // Optimization: avoid calling 3 times itemUIRegistry.getLabel(w)
```
sjka: While I agree this makes sense, I also would have preferred to keep it separate.

bundles/core/org.eclipse.smarthome.core/src/main/java/org/eclipse/smarthome/core/items/GenericItem.java
lolodomo: Yes, that is exactly the idea. It gives us a chance to get options from channel for example. I could add a comment.
What are the unexpected results ?
Do you prefer I keep this method unchanged and i create another one with that code ?
My PR is clearly to fix two different things that are more or less linked to state description. Splitting in two PRs means more work, more time. But ok, I will take the time !

bundles/ui/org.eclipse.smarthome.ui/src/main/java/org/eclipse/smarthome/ui/internal/items/ItemUIRegistryImpl.java
lolodomo: It is the fix for options. If you have options defined for a string channel, geLabel(w) should translate the value to its label. By the way, I think my code is not ewactly at the right place, I have to consider the case where the user requests for a transformation.

extensions/ui/org.eclipse.smarthome.ui.basic/src/main/java/org/eclipse/smarthome/ui/basic/internal/render/AbstractWidgetRenderer.java
lolodomo: I will create a separate PR !

lolodomo:
@sjka : split done, you have here only the fix for issue #3741

lolodomo:
Regarding `ChannelStateDescriptionProviderOSGiTest`, my Eclipse IDE is showing an error at this line: https://github.com/eclipse/smarthome/blob/master/bundles/core/org.eclipse.smarthome.core.thing.test/src/test/java/org/eclipse/smarthome/core/thing/internal/ChannelStateDescriptionProviderOSGiTest.java#L76
JavaOSGiTest.bundleContext is not visible.

Same error in `TranslationProviderOSGiTest`.

Is there something broken with current tests ?

lolodomo:
`mvn clean install` on `org.eclipse.smarthome.core.thing.test` succeeeded.
Should I just ignore the error in Eclipse ?

lolodomo:
Tests have been added.

bundles/core/org.eclipse.smarthome.core.thing.test/src/test/java/org/eclipse/smarthome/core/thing/internal/ChannelStateDescriptionProviderOSGiTest.java
```diff
@@ -153,22 +193,38 @@ public void presentItemStateDescription() {
...
-        Item numberItem = null;
-        for (final Item item : items) {
-            if (item.getType().equals("Number")) {
-                numberItem = item;
-                break;
-            }
+        Item item;
+        try {
+            item = itemRegistry.getItem("TestItem");
+        } catch (ItemNotFoundException e) {
```
htreu: The test could also be declared as `throws Exception` and fail happily if this happens. that would reduce lines 218 - 223 to `Item item = itemRegistry.getItem("TestItem");`.

bundles/core/org.eclipse.smarthome.core.thing.test/src/test/java/org/eclipse/smarthome/core/thing/internal/ChannelStateDescriptionProviderOSGiTest.java
```diff
@@ -182,6 +238,77 @@ public void presentItemStateDescription() {
         Assert.assertEquals("SOUND", opt.getValue());
         Assert.assertEquals("My great sound.", opt.getLabel());
 
+        try {
+            item = itemRegistry.getItem("TestItem2");
+        } catch (ItemNotFoundException e) {
```
htreu: same here

bundles/core/org.eclipse.smarthome.core.thing.test/src/test/java/org/eclipse/smarthome/core/thing/internal/ChannelStateDescriptionProviderOSGiTest.java
```diff
@@ -153,22 +193,38 @@ public void presentItemStateDescription() {
                 null, "test thing", new Configuration());
         Assert.assertNotNull(thing);
```
htreu: It would be nice to have all `assert*` methods to be static imports like in the other java tests.

bundles/core/org.eclipse.smarthome.core.thing.test/src/test/java/org/eclipse/smarthome/core/thing/internal/ChannelStateDescriptionProviderOSGiTest.java
```diff
@@ -153,22 +193,38 @@ public void presentItemStateDescription() {
...
-                break;
-            }
+        Item item;
+        try {
+            item = itemRegistry.getItem("TestItem");
+        } catch (ItemNotFoundException e) {
+            item = null;
         }
-        Assert.assertNotNull(numberItem);
+        Assert.assertNotNull(item);
```
htreu: This is obsolete since the next line will fail if `item` is null. Please also review other `assertNotNull` checks.

bundles/core/org.eclipse.smarthome.core.thing.test/src/test/java/org/eclipse/smarthome/core/thing/internal/ChannelStateDescriptionProviderOSGiTest.java
lolodomo: If I do that, the next tests will not be run. Not a problem ?

lolodomo:
Ok, tests are now fixed according to your suggestions.

lolodomo:
The Travis error has apparently nothing to do with my changes.

lolodomo:
To be very clear, this PR changes the current behaviour.

The old behavior is to display a value for a string or number item only if a pattern is found either in the sitemap or in the item or in the thing channel, and only if there is no `[]` in the label at the first level. It could lead to no display of value, a frequent case for users using Paper UI to automatically link channels to items.

The new behavior is to always display a value for a string or number item except if there is a `[]` in the label at the first level. The pattern is taken first in the sitemap then in the item then in the thing channel and finally a default one is considered. The default one is `[%s]` for string item and `[%.0f]` for number items.

There is no change for other kinds of items.

lolodomo:
Maybe we should consider a default pattern for DateTime items too.

bundles/core/org.eclipse.smarthome.core.thing.test/src/test/java/org/eclipse/smarthome/core/thing/internal/ChannelStateDescriptionProviderOSGiTest.java
htreu: The next assertions in this test method will not run, but as this aspect fails anyway this is not a problem. The other test methods should be executed individually.

lolodomo:
Are you still wanting for me on this PR ? I think I handled all requests ?

sjka:
Thanks for the ping. No, the change looks good to me now.
Due to it's nature of changing the current behavior (to the better in my belief though), I would like to give @kaikreuzer / @maggu2810 also a change to express their consent here.

bundles/core/org.eclipse.smarthome.core.thing/src/main/java/org/eclipse/smarthome/core/thing/internal/ChannelStateDescriptionProvider.java
```diff
@@ -57,7 +61,29 @@ public StateDescription getStateDescription(String itemName, Locale locale) {
             Channel channel = thingRegistry.getChannel(channelUID);
             if (channel != null) {
                 ChannelType channelType = thingTypeRegistry.getChannelType(channel, locale);
-                return channelType != null ? channelType.getState() : null;
+                if (channelType != null) {
+                    StateDescription state = channelType.getState();
+                    if ((channelType.getItemType() != null) && ((state == null) || (state.getPattern() == null))) {
+                        String pattern = null;
+                        if (channelType.getItemType().equalsIgnoreCase("String")) {
```
kaikreuzer: Better use the constant `CoreItemFactory.STRING`.

bundles/core/org.eclipse.smarthome.core.thing/src/main/java/org/eclipse/smarthome/core/thing/internal/ChannelStateDescriptionProvider.java
```diff
@@ -57,7 +61,29 @@ public StateDescription getStateDescription(String itemName, Locale locale) {
...
+                            pattern = "%s";
+                        } else if (channelType.getItemType().equalsIgnoreCase("Number")) {
```
kaikreuzer: Better use the constant `CoreItemFactory.NUMBER`.

bundles/core/org.eclipse.smarthome.core.thing/src/main/java/org/eclipse/smarthome/core/thing/internal/ChannelStateDescriptionProvider.java
```diff
@@ -57,7 +61,29 @@ public StateDescription getStateDescription(String itemName, Locale locale) {
             Channel channel = thingRegistry.getChannel(channelUID);
             if (channel != null) {
                 ChannelType channelType = thingTypeRegistry.getChannelType(channel, locale);
-                return channelType != null ? channelType.getState() : null;
+                if (channelType != null) {
+                    StateDescription state = channelType.getState();
```
kaikreuzer: Could you call the variable `stateDescription`? (I know that the name of the getter method of ChannelType is not ideal...)

lolodomo: Done

bundles/core/org.eclipse.smarthome.core.thing/src/main/java/org/eclipse/smarthome/core/thing/internal/ChannelStateDescriptionProvider.java
lolodomo: org.eclipse.smarthome.core.library is not exported.

kaikreuzer: Interesting. This is against our own guidelines, so feel free to add this export as part of this PR :-)

lolodomo:
Done

sjka:
Great, thanks!

Issue #2729, opened 2017-01-03
BasicUI does not map option values while PaperUI does


crnjan:
I have a binding with one of the channels defined as below:

    <channel-type id="lastErrorCode">
        <item-type>String</item-type>
        <label>Last error information</label>
        <category>Error</category>
        <state readOnly="true" pattern="%s">
            <options>
                <option value="0">Sensor radiator return (GT1)</option>
                ...
                <option value="20">Three phase incorrect order</option>
                <option value="21">Power failure</option>
                <option value="22">High delta GT8/GT9</option>
            </options>
        </state>
    </channel-type>

When channel has a value, i.e. 22, [PaperUI](https://community-openhab-org.s3-eu-central-1.amazonaws.com/optimized/2X/b/b54af72d5c9e9ae57d929d39907c2a4385278aa3_1_690x343.jpg) displays the option value, as expected - "High delta GT8/GT9", but within [BasicUI](https://community-openhab-org.s3-eu-central-1.amazonaws.com/optimized/2X/7/7d8a05290855e38ebc03febea39445775258eba9_1_690x339.jpg) I see the actual value, that is 22 (same on iOS client).

Code can be found [here](https://github.com/openhab/openhab2-addons/pull/1634).

sjka:
This is not necessarily a flaw in the BasicUI, but could be solved in the core framework instead. We have two `StateDescriptionProvider`s: The `GenericItemProvider` which has the information from the items-DSL (including any custom formatting pattern) and the `ChannelStateDescriptionProvider` which has the information from the linked channel (including the state options as you outlined above). 

The GenericItem just takes the StateDescription from whatever provider is first in the list. Therefore the info from the second one gets lost. 

IMHO the fact that we can have multiple `StateDescriptionProvider` makes sense. But they should be disjunct so we don't have two different ones answering for the same item. The `ChannelStateDescriptionProvider` somehow has a special role (i.e. it looks up default from the channel definitions). It should NOT be registered as `StateDescriptionProvider` and should be used to lookup the defaults which another `StateDescriptionProvider` can override.

Of course, the `ItemUIRegistryImpl` in the end also has to deal with these options and replace the options' values with labels for rendering purposes.

In that way, not every UI needs to do the lookup for item -> linked channels -> channel types -> options on its own.

@kaikreuzer, @cdjackson:  WDYT?

kaikreuzer:
[...]

PR #3825, opened 2017-07-11
UI: display mapped option value

lolodomo:
Fix issue #2729

### DIFF
--- a/bundles/core/org.eclipse.smarthome.core/src/main/java/org/eclipse/smarthome/core/items/GenericItem.java
+++ b/bundles/core/org.eclipse.smarthome.core/src/main/java/org/eclipse/smarthome/core/items/GenericItem.java
@@ -27,6 +27,7 @@ import org.eclipse.smarthome.core.types.RefreshType;
 import org.eclipse.smarthome.core.types.State;
 import org.eclipse.smarthome.core.types.StateDescription;
 import org.eclipse.smarthome.core.types.StateDescriptionProvider;
+import org.eclipse.smarthome.core.types.StateOption;
 import org.eclipse.smarthome.core.types.UnDefType;
 import org.slf4j.Logger;
 import org.slf4j.LoggerFactory;
@@ -401,15 +402,32 @@ abstract public class GenericItem implements ActiveItem {
 
     @Override
     public StateDescription getStateDescription(Locale locale) {
+        StateDescription result = null;
+        List<StateOption> stateOptions = Collections.emptyList();
         if (stateDescriptionProviders != null) {
             for (StateDescriptionProvider stateDescriptionProvider : stateDescriptionProviders) {
                 StateDescription stateDescription = stateDescriptionProvider.getStateDescription(this.name, locale);
-                if (stateDescription != null) {
-                    return stateDescription;
+
+                // as long as no valid StateDescription is provided we reassign here:
+                if (result == null) {
+                    result = stateDescription;
+                }
+
+                // if the current StateDescription does provide options and we don't already have some, we pick them up
+                // here
+                if (stateDescription != null && !stateDescription.getOptions().isEmpty() && stateOptions.isEmpty()) {
+                    stateOptions = stateDescription.getOptions();
                 }
             }
         }
-        return null;
+
+        // we recreate the StateDescription if we found a valid one and state options are given:
+        if (result != null && !stateOptions.isEmpty()) {
+            result = new StateDescription(result.getMinimum(), result.getMaximum(), result.getStep(),
+                    result.getPattern(), result.isReadOnly(), stateOptions);
+        }
+
+        return result;
     }
 
     /**
--- a/bundles/core/org.eclipse.smarthome.core/src/main/java/org/eclipse/smarthome/core/items/Item.java
+++ b/bundles/core/org.eclipse.smarthome.core/src/main/java/org/eclipse/smarthome/core/items/Item.java
@@ -128,14 +128,16 @@ public interface Item extends Identifiable<String> {
     public String getCategory();
 
     /**
-     * Returns the state description (uses the default locale).
+     * Returns the first provided state description (uses the default locale).
+     * If options are defined on the channel, they are included in the returned state description.
      *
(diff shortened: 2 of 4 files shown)


htreu:
If I understand the code correctly this would for all DSL based items inherit the full stateDescription from the linked channel-type and then render a state even if the user set the label of the item to be w/o a pattern. According to the discussion in #3741 the idea was to somehow migrate the items files and add a dedicated `pattern` attribute to the items definition to overcome this. Will this also be part of the PR?

lolodomo:
Updating the DSL to split label and pattern is not the object of this PR.
The main purpose here is to retrieve the channel options and display the mapped option value.

lolodomo:
PR ready, I added two tests for options.

bundles/core/org.eclipse.smarthome.core/src/main/java/org/eclipse/smarthome/core/items/GenericItem.java
```diff
@@ -412,6 +414,53 @@ public StateDescription getStateDescription(Locale locale) {
         return null;
     }
 
+    @Override
+    public StateDescription getMixedStateDescription() {
+        return getMixedStateDescription(null);
```
htreu: The JavaDoc tells the default locale is used. You pass in `null` here instead. Please also review the `getStateDescription` method which behaves analogous. Alternatively the JavaDoc in `Item` should be changed.

htreu:
> The main purpose here is to retrieve the channel options and display the mapped option value.

...which should work seamlessly with this PR. Nonetheless does it also change the current behaviour on item rendering by inheriting the state pattern from the channel-type. This will change a lot of sitemap entries.

lolodomo:
`getStateDescription` was already retrieving the pattern from the channel if not defined in the item. Nothing is changed here by using `getMixedStateDescription`. If no provider returns a state description with pattern, no value is displayed.

What is changed **by the other PR** is that now `ChannelStateDescriptionProvider` will now build and provide a pattern for String and Number channels even if not defined in the channel definition. So in this case, the value will be displayed even if the user just use "label" as label. It is exactly what we try to fix (in the other PR). The user still has the ability to use "label []" as item or sitemap label to disable the display of a string or number channel.
All your concerns are in fact relative to the other PR but not this one.

kaikreuzer:
> relative to the other PR but not this one.

So what does this one exactly do/fix? I am also very surprised by changed to core classes such as GenericItem, which we should really avoid whenever possible. And a new method like `getMixedStateDescription` really does not sound obvious to me - seeing it, I have no clue what to expect from it... It rather sounds like some hack that covers some underlying architectural issue.

sjka:
To my understanding it builds a state description by mixing the properties contributed by several/all providers. For each property, the non-null value from the provider with the highest ranking will be used.

As I already pointed out in https://github.com/eclipse/smarthome/pull/3822#discussion_r126447982, I'm pretty sure this will lead to funny surprises and therefore IMHO is not a good idea to do.

lolodomo:
Hey guys, I try to address the problem that was discussed in the original issue. The current method does not return the channel options in case you have a state description that is provided by the item provider that has a higher priority (for example with a pattern in the item label).
Let me know how I can retrieve the channel options ?
I am sure there will be no surprise with my current implementation.
But if you have another idea, let me know.

lolodomo:
In case it is still not clear, this PR fixes issue #2729 , display of mapped option value.

lolodomo:
That being said, using `getMixedStateDescription` rather than `getStateDescription` will fix a bug with label too in a very particular case. The use case is you have a state description returned by the item provider that contains for example a min and a max value but without any pattern; a pattern is defined for the channel and a state description with this pattern is returned by the channel state description provider. In this case, no value will be displayed because `getStateDescription` stops at the first state description (without pattern) while `getMixedStateDescription` will provide a state description containing the min and max values + the pattern returned by the channel state description provider.

htreu:
@lolodomo I can follow your arguments and your code does indeed fix the original problem. My concerns are in the direction of @kaikreuzer and @sjka: by introducing a new `getMixed...` method we indicate that the original `getStateDescription` method might not be the right one to call. And even if the _mixing_ can be described via JavaDoc it is not clear what really happens under the hood w/o reading the code and knowing the exact ranking configuration of the runtime. On top of this there might be additional `StateDescriptionProvider` implementations and the _mixing_ will become more and more unintuitive.
As for this issue I would suggest a little less intrusive fix: retrieve the stateDescription just like it is done now but then loop through all the stateDescriptionProviders and add the first options list you can find if none was set already. Like this https://github.com/htreu/smarthome/commit/b56e0b0e7cd2b192334c718076e531cd6f28c147

wdyt?

lolodomo:
What you are doing is exactly what I was doing except that you do the mix only for the options.
Doing it inside `getStateDescription` rather than creating a new method means that you are impacting other calls to `getStateDescription`. That's probably ok but I am not 100% sure. This method is used in the REST API if I correctly remember.
Of course, with your code, the fix for mapped option value will work too.
But as I explained in my previous message, we need the mix on the pattern too (not for the issue relative to options but for the display of value in general).

htreu:
Yes, my proposal is to only fix the options issue with the same method as you did. But leave the rest as is which gives us the opportunity to come up with a good concept on how to provide consistent state descriptions (if this is still an issue then).

lolodomo:
Before I change my code, is htreu's proposal ok for everybody ?
I will add a TODO in the code of getLabel to explain the use case that is not working for getting the pattern. Hopefully, it is not a common use case and the user has a workaround.

lolodomo:
@sjka @kaikreuzer : is it ok for you if I change `getStateDescription` as @htreu's proposed, to get a mix only for options in StateDescription (leaving a little bug in pattern handling) ?

lolodomo:
Can I have a feedback so that I can finish this bug fix ?

kaikreuzer:
Sorry @lolodomo, this got lost during my vacation. Yes, the suggestion from @htreu is ok for me!

lolodomo:
Ok, done.

PR #3833, opened 2017-07-12
remove apache commons http client from target platform

kaikreuzer:
Afair, we do not have any dependency on the Apache Commons HTTP client anymore.


sjka:
I'm also not aware of any within ESH. 
Travis agrees.

Issue #3812, opened 2017-07-07
[RFC] Mqtt binding(s)


davidgraeff:
I have the following binding proposals and would like to have feedback if those are actually a good thing to have in ESH.

### Broker configuration binding
_Proposed package name: MqttBroker_ (I am open for name suggestions)

At the moment it is possible to configure brokers in a textual way. Without intensively watching the log file, there is no way to tell if the provided configuration is correct and if the connection to the broker have been established, neither via the GUI (PaperUI) nor machine readable (REST interface).

I propose a binding to make the configuration of brokers more intuitive and just let the user add them as things (and sync them to the io.transport.mqtt MqttService). Features would be:

* Rich status feedback
* Auto discovery of brokers (mqtt-sn, zeroconf, best effort subnet scan)
* Configurable reconnect time (this is not possible via textual configuration yet)
* Certificate and public key pinning for ssl connections

Personally I would also like to embedded a simple mqtt server like [Moquette](https://github.com/andsel/moquette) and list this in the discovery as an option to the user. Setting up a working Mqtt Broker would be just a button click away for ESH users of this binding.

### Mqtt Things
_Proposed package name: MqttThings_ (I am open for name suggestions)

At this very moment there exist no means to actually connect mqtt topics to ESH things. There are a lot of [commercial](https://github.com/mqtt/mqtt.github.io/wiki/servers) and non-commercial (Owntracks, EasyESP) Mqtt products out there who deserve their own binding.

But for a start I propose a generic Mqtt binding where you can bind topics to things (very much like the Openhab 1.x Mqtt binding, but modern). This should be part of the ESH extensions for 1) it to serves as a implementation examples for other Mqtt binding implementers and 2) allow the ESH consumers to easily interact with one of the core IoT protocols. The following features would be part of the binding:

* List all configured brokers of the io.transport.mqtt MqttService and let the user add them as bridges.
* Manually add a thing (Switch-, Contact-, Number-, Range-thing) and bind it to a Mqtt topic state and command.
* Auto discover topics (if retain messages==true)

### Mqtt2Eventbus
_Proposed package name: Mqtt2Eventbus_ (I am open for name suggestions)

It might be a good idea to have a way to synchronize two or more instances of ESH ([forum post](https://community.openhab.org/t/integrating-two-instances-of-openhab-via-mqtt-leading-to-infinite-loop/30454)). Mqtt could be a perfect protocol to archive this.

I am looking forward to some feedback.

Cheers, David

[...]

PR #3835, opened 2017-07-12
Mqtt: Fix NPE and other issues

davidgraeff:
Now that I start to use the io.transport.mqtt bundle to implement some of the ideas in #3812, I encountered some problems that this PR addresses.

## NPE
The ReconnectPolicy family of classes need an internal reference to the MqttBrokerConnection (to call start() again). The reference is set when `MqttBrokerConnection.setReconnectPolicy` is called. That was at least the plan, but didn't happen. The result was an NPE in the PeriodicReconnectPolicy object.

## Tests
* Move test classes from the mqtt.test package to the mqtt package to have access to package private methods.
* Add extensive tests for ReconnectPolicy (there were none before).
* Extend tests for MqttConnectionObserver.

## Necessary API changes
* Rename MqttConnectionObserver.setConnected to MqttConnectionObserver.connectionStateChanged and add a second Throwable parameter to gain access to the error that happened during a connect. There was no way to retrieve any error message for displaying it to the user before.
* Make addConnectionObserver not immediately call the callback back. This is unnecessary and breaks a simple pattern to call `MqttBrokerConnection.start()` synchronously (more details in the _java doc_ for `start()`).
* Add own MqttException instead of leaking the internal paho MqttException.
* Add own MqttPublishCallback and MqttPublishResult to not leak the internal paho callback interface.

As usual there are some added JavaDocs and comments here and there when I touched the ReconnectPolicy and related classes.


bundles/io/org.eclipse.smarthome.io.transport.mqtt/src/main/java/org/eclipse/smarthome/io/transport/mqtt/MqttConnectionObserver.java
```diff
@@ -11,14 +11,18 @@
...
+     * requires that the control is returned quickly to not stall the Mqtt thread.
      *
-     * @param connected true if the connection is established, false if disconnected.
+     * @param connected True if the connection is established, false if disconnected.
+     * @param error An exception object (might be a MqttException) with the reason why
+     *            a connection failed.
      */
-    public void setConnected(boolean connected);
-
+    public void connectionStateChanged(boolean connected, Throwable error);
```
htreu: This is an API breaking change. If we get aways with this, how about "extending" it a little further by passing an enum `ConnectionState` with `CONNECTED, DISCONNECTED` instead of the boolean. This will allow us to extend the API w/o breaking it. wdyt?

bundles/io/org.eclipse.smarthome.io.transport.mqtt/src/main/java/org/eclipse/smarthome/io/transport/mqtt/reconnect/AbstractReconnectPolicy.java
```diff
@@ -22,10 +22,21 @@
...
     }
 
+    /**
+     * Return the brokerConnection object that this reconnect policy is assigned to.
+     */
+    public MqttBrokerConnection getBrokerConnection() {
+        return brokerConnection;
+    }
+
+    public boolean isReconnecting() {
```
htreu: Is this also a candidate for being `abstract`? This would force the implementations to think about what to return here.

bundles/io/org.eclipse.smarthome.io.transport.mqtt/src/main/java/org/eclipse/smarthome/io/transport/mqtt/MqttConnectionObserver.java
davidgraeff: I agree and have it implemented this way. The API is only 1 day old, that's why I think if we act 'quick', it is worth it. Before actually using the API in a project I haven't noticed that there is no way of retrieving any information on what went wrong for a connection attempt.

davidgraeff:
@htreu: I just noticed, that the manifest file was incomplete and some classes are in the wrong package. Could you please remove your approval and do a review of the second commit as well. This is what happened:
* Export org.eclipse.smarthome.io.transport.mqtt.reconnect
* Export org.eclipse.smarthome.io.transport.mqtt.sslcontext
* Move classes from org.eclipse.smarthome.io.transport.sslcontext to org.eclipse.smarthome.io.transport.**mqtt**.sslcontext
* Because nobody was able to use the org.eclipse.smarthome.io.transport.mqtt.reconnect namespace anyway so far, this should not be an API breakage: Rename Reconnect**Policy** to Reconnect**Strategy**. Strategy is the more correct term here.
* Add constructors to `PeriodicReconnectStrategy`.

davidgraeff:
And I've got this question: Some methods in `MqttBrokerConnection` can throw a `MqttException`. This is a `org.eclipse.paho.client.mqttv3` class and forces users of the `MqttBrokerConnection` class to import that package. Is this all right or should we encapsulate this exception in an own `MqttBrokerConnection` to hide the implementation details?

sjka:
> Is this all right or should we encapsulate this exception in an own `MqttException` to hide the implementation details?

Please do the latter to keep the API clean and not leak the library classes.

davidgraeff:
In a third commit, I have added a `MqttException` class to no longer leak the internal library one.

bundles/io/org.eclipse.smarthome.io.transport.mqtt/src/main/java/org/eclipse/smarthome/io/transport/mqtt/MqttWillAndTestament.java
```diff
@@ -66,6 +66,28 @@ public static MqttWillAndTestament fromString(String string) {
...
+     * @param payload The optional payload. Can be null.
+     * @param qos Valid values are 0 (Deliver at most once),1 (Deliver at least once) or 2</li>
+     * @param retain true if messages shall be retained
+     */
+    public MqttWillAndTestament(String topic, byte[] payload, int qos, boolean retain) {
+        this.topic = topic;
+        this.payload = payload;
+        this.qos = qos;
+        this.retain = retain;
+        if (StringUtils.isBlank(this.topic)) {
```
htreu: this should be the first call/block in the constructor.

bundles/io/org.eclipse.smarthome.io.transport.mqtt/src/main/java/org/eclipse/smarthome/io/transport/mqtt/MqttWillAndTestament.java
```diff
@@ -66,6 +66,28 @@ public static MqttWillAndTestament fromString(String string) {
         return result.isValid() ? result : null;
     }
 
+    private MqttWillAndTestament() {
+
```
htreu: please comment the empty block to indicate why its empty.

bundles/io/org.eclipse.smarthome.io.transport.mqtt/src/main/java/org/eclipse/smarthome/io/transport/mqtt/MqttException.java
```diff
@@ -0,0 +1,110 @@
...
+    private Throwable cause;
+
+    /**
+     * Constructs a new <code>MqttException</code> with the specified code
+     * as the underlying reason.
+     *
+     * @param reasonCode the reason code for the exception.
+     */
+    public MqttException(int reasonCode) {
+        super();
```
davidgraeff: Are u sure? Because the base classes (Exception, Throwable) do this as well and in throwable there is actually work done. I thought it is common to do this in exceptions. I will remove this if you are absolutely sure though.

bundles/io/org.eclipse.smarthome.io.transport.mqtt/src/main/java/org/eclipse/smarthome/io/transport/mqtt/MqttWillAndTestament.java
davidgraeff: done

bundles/io/org.eclipse.smarthome.io.transport.mqtt/src/main/java/org/eclipse/smarthome/io/transport/mqtt/MqttWillAndTestament.java
davidgraeff: done

davidgraeff:
I discovered, that the publish() method leaked another paho class (IMqttActionListener) and therefore added MqttPublishCallback and MqttPublishResult. The MqttBrokerConnection.publish() method now also returns the messageID.

bundles/io/org.eclipse.smarthome.io.transport.mqtt/src/main/java/org/eclipse/smarthome/io/transport/mqtt/MqttException.java
htreu: Please see the last `Note` in the official Java documentation: https://docs.oracle.com/javase/tutorial/java/IandI/super.html

The no-arg super constructor is always called.

davidgraeff: done

davidgraeff:
@htreu I have incorporated all suggested changes

kaikreuzer:
Thanks @davidgraeff!

PR #3837, opened 2017-07-13
[audio] Refactored use of filename, filename extension and separator

cweitkamp:
* Refactored use of filename, filename extension and separator
* Added util class for handling filenames and filename extensions
* Applied changes to Sonos binding


cweitkamp:
Pointing to this [discussion](https://github.com/eclipse/smarthome/pull/3598#discussion_r120611406).

I think we schould discuss the naming of the util class and if we should consider to move the declarations of the `*_EXTENSION` contants into the util class, too.

bundles/core/org.eclipse.smarthome.core.audio/src/main/java/org/eclipse/smarthome/core/audio/utils/AudioStreamUtils.java
```diff
@@ -0,0 +1,75 @@
+/**
```
htreu: Whats is the reason for this class? We have `org.apache.commons.io` already in the classpath which provides `FilenameUtils`. The methods are identical to these implemented here.

bundles/core/org.eclipse.smarthome.core.audio/src/main/java/org/eclipse/smarthome/core/audio/URLAudioStream.java
```diff
@@ -52,27 +57,34 @@ public URLAudioStream(String url) throws AudioException {
...
-                        Matcher matcher = plsStreamPattern.matcher(line);
-                        if (matcher.find()) {
-                            url = matcher.group(1);
+            final String filename = url.toLowerCase();
+            final String extension = AudioStreamUtils.getExtension(filename);
+            switch (extension) {
+                case M3U_EXTENSION:
+                    InputStream isM3U = new URL(url).openStream();
+                    String urlsM3U = IOUtils.toString(isM3U);
+                    for (String line : urlsM3U.split("\n")) {
```
htreu: Line 65 & 66 might also be a call to `IOUtils.readLines(is)` which does provide a List\<String\> ready to use.

bundles/core/org.eclipse.smarthome.core.audio/src/main/java/org/eclipse/smarthome/core/audio/URLAudioStream.java
```diff
@@ -52,27 +57,34 @@ public URLAudioStream(String url) throws AudioException {
...
+                            url = line;
                             break;
                         }
                     }
-                }
+                    break;
+                case PLS_EXTENSION:
+                    InputStream isPLS = new URL(url).openStream();
+                    String urlsPLS = IOUtils.toString(isPLS);
+                    for (String line : urlsPLS.split("\n")) {
```
htreu: same here to read lines by `IOUtils.readLines(is)`.

bundles/core/org.eclipse.smarthome.core.audio/src/main/java/org/eclipse/smarthome/core/audio/utils/AudioStreamUtils.java
cweitkamp: @htreu thanks for your review. Have a look at this [discussion](https://github.com/eclipse/smarthome/pull/3598#discussion_r120611406). I suggested to use that util class, but we decided to be less dependant. That is the reason for my own implementation.

bundles/core/org.eclipse.smarthome.core.audio/src/main/java/org/eclipse/smarthome/core/audio/URLAudioStream.java
cweitkamp: Fair point. The use of `IOUtils.toString(InputStream input)` is deprecated since `org.apache.commons.io v2.5`, so we should adopt that anyway.

cweitkamp:
@htreu I applied the changes.

bundles/core/org.eclipse.smarthome.core.audio/src/main/java/org/eclipse/smarthome/core/audio/URLAudioStream.java
```diff
@@ -52,27 +57,32 @@ public URLAudioStream(String url) throws AudioException {
...
-                for (String line : urls.split("\n")) {
-                    if (!line.isEmpty() && line.startsWith("File")) {
-                        Matcher matcher = plsStreamPattern.matcher(line);
-                        if (matcher.find()) {
-                            url = matcher.group(1);
+            final String filename = url.toLowerCase();
+            final String extension = AudioStreamUtils.getExtension(filename);
+            switch (extension) {
+                case M3U_EXTENSION:
+                    final InputStream isM3U = new URL(url).openStream();
```
sjka: what about closing the stream?

bundles/core/org.eclipse.smarthome.core.audio/src/main/java/org/eclipse/smarthome/core/audio/URLAudioStream.java
```diff
@@ -52,27 +57,32 @@ public URLAudioStream(String url) throws AudioException {
...
+                    for (final String line : IOUtils.readLines(isM3U)) {
+                        if (!line.isEmpty() && !line.startsWith("#")) {
+                            url = line;
                             break;
                         }
                     }
-                }
+                    break;
+                case PLS_EXTENSION:
+                    final InputStream isPLS = new URL(url).openStream();
```
sjka: what about closing the stream?

bundles/core/org.eclipse.smarthome.core.audio/src/main/java/org/eclipse/smarthome/core/audio/URLAudioStream.java
cweitkamp: Good point. I changed it.

bundles/core/org.eclipse.smarthome.core.audio/src/main/java/org/eclipse/smarthome/core/audio/URLAudioStream.java
cweitkamp: Good point. I changed it. Maybe the `try {} catch {}` block is placed incorrect now and should be refactored / split up as well.

bundles/core/org.eclipse.smarthome.core.audio/src/main/java/org/eclipse/smarthome/core/audio/URLAudioStream.java
maggu2810: `java.io.InputStream` implements `Closeable` and so `AutoCloseable`.
Why not using try-with-resources?

cweitkamp: This design pattern is new to me, thanks for pointing to it. It looks like the best solution for this.

cweitkamp:
@maggu2810 thanks for review, I refactored the code to use the try-with-resources statement.

bundles/core/org.eclipse.smarthome.core.audio/src/main/java/org/eclipse/smarthome/core/audio/utils/AudioStreamUtils.java
```diff
@@ -0,0 +1,75 @@
...
+     * 
+     * @param filename the filename to check the extension of
+     * @param extension the extension to check for
+     * @return {@code true} if the filename has the specified extension
+     */
+    public static boolean isExtension(String filename, String extension) {
+        if (filename == null) {
+            return false;
+        }
+        if ((extension == null) || extension.isEmpty()) {
```
maggu2810: You could remove the braces around the first check.

cweitkamp:
@maggu2810 done. If you want me to squash my commits, please let me know.

maggu2810:
> If you want me to squash my commits, please let me know.

We can do it on merging.

Thank you

PR #3844, opened 2017-07-16
Respond with status code 404 for missing icons

### LINKED ISSUE #3842: IconServlet: Requesting non-existing icon automatically delivers "none" icon
Steps to reproduce:
1. Start the eclipse smarthome runtime
2. Open the browser and request a non-existing image, e.g. using this url:
 http://localhost:8080/icon/big_bulb1?state=null&format=PNG

Expected result:
As the image does not exist, the request should get the reponse code 404 and an empty response, indicating, that the requested image does not exist and the client should handle this case accordingly.

Observed result:
The request is automatically redirectred (internally, without sending a 302 or 301 to the client to indicate, that the request should be retried with the URL in the Location header, see the code starting from [here](https://github.com/eclipse/smarthome/blob/9910b321f87576eeb0e0de2fe5552837ddf5a95b/bundles/ui/org.eclipse.smarthome.ui.icon/src/main/java/org/eclipse/smarthome/ui/icon/internal/IconServlet.java#L120)), as if it would request the "none" image category and the empty "none" image is delivered to the client with a 200 OK status code.

This makes it difficult for clients to identify if the requested image was returned and, e.g. should be cached or something else, or if the image could not be found. If the requested image could not be found, it should be up to the client how it handles such cases.

FlorianSW:
If an Icon does not exist, the request requesting this icon should respond
with a 404 status code and let the client decide, how it wants to handle
such cases.

In this commit, the IconServlet is changed to do not automatically deliver
the none.png image when an icon does not exist. The basicui and classicui
are updated (as they're clients for this servlet) to automatically load
none.png whenever the img tag of an icon fails to load. The behaviour
visible for a user is the same.

Fixes #3842


kaikreuzer:
Great that you also directly address the Basic UI! From the linked issues, I see that you came across this topic through the Android app, so I assume this will even work better with the 404 than with having an empty icon delivered, right? Do you happen to know if the same is true for the iOS app?

FlorianSW:
I would need to check, if the Android app needs to be changed in order to show the same result as it does now (the none.png), however, it will at least skip the cache for these kinds of icons, which will make it easier to fix, that the image is not there.

I haven't took a look at the openHAB iOS app, as I don't have an iOS device and therefore can't test or use the app and normally doesn't take a look at the code, too, sorry :(

digitaldan:
>Do you happen to know if the same is true for the iOS app?

My guess is the IOS will just not load any icon, is the desire that it should attempt to load /icon/none.png on 404?

FlorianSW:
> My guess is the IOS will just not load any icon, is the desire that it should attempt to load /icon/none.png on 404?

Depends on what the user would expect there :) From my point of view, it could even be an empty white box, which is rendered on the device, so you do not require the user to spend data from their plan to download an image, which in fact give any additional information.

sjka:
Well, the current approach is to serve an empty (or rather: fully transparent, I believe) icon instead. If I get it right from the referenced issue, it was introduced to suppress the "broken image link" icon rendered by most browsers automatically. 

With this historic reasoning, I don't believe that is is intended to have a "no icon"-icon in every icon set to be displayed instead. So if Android & iOS can handle that themselves, that it indeed would be much better to simply don't display anything instead of loading this nothing over the air.

The question remains: What are those to apps doing once we merge this and return 404? I.e. will we need some follow-up issues in the mobile apps or will they simple continue to work nicely?

FlorianSW:
I can't say anything about the iOS app (like I already said). However, the Android app at the moment simply renders a fallback ressource, which is a transparent "nothing":
https://github.com/openhab/openhab.android/blob/master/mobile/src/main/res/drawable-hdpi/blank_icon.png

lolodomo:
I will try this change with Basic UI, Classic UI and HABdroid.

kaikreuzer:
Reading the history, it seems that this "none" icon request was actually introduced for https://github.com/eclipse/smarthome/issues/2784, i.e. for a situation where there was actually NO ICON to be displayed (and not an icon that does not exist). If an icon is defined, it is anyhow expected to exist, so I would also say that a 404 should be fine in any case and that the clients should not even do a request to the icon servlet in case that no icon is expected.

maggu2810:
<pre>
Conflicting files
extensions/ui/org.eclipse.smarthome.ui.basic/web/smarthome.js
</pre>

sjka:
Before merging this, IMHO it would be good to at least _know_ if and how this is going to break the known downstream projects, therefore I would prefer to wait for @lolodomo's feedback (thanks for offering your help, by the way!)

lolodomo:
I tested with only a text elements linked to number items (so the icon category "number" is searched and not found), by replacing in the demo:
```
Number Weather_Temp_Max         "Todays Maximum [%.1f °C]" <temperature>      (Weather, Weather_Chart)
Number Weather_Temp_Min         "Todays Minimum [%.1f °C]" <temperature>      (Weather, Weather_Chart)
```
by
```
Number Weather_Temp_Max         "Todays Maximum [%.1f °C]"      (Weather, Weather_Chart)
Number Weather_Temp_Min         "Todays Minimum [%.1f °C]"      (Weather, Weather_Chart)
```
It is OK with Basic UI and HABdroid but **not OK with Classic UI, neither with Firefox nor Chrome. With Classic UI, the bad icon is displayed.**

lolodomo:
Note that I see the log "Requested icon category number provided by no icon provider" so 404 error is returned but probably not well handled by Classic UI.

lolodomo:
For information, why are they several files appearing as updated for Classic UI while I see in fact no real change, like for example extensions/ui/org.eclipse.smarthome.ui.classic/snippets/image.html ?

lolodomo:
By the way, the code is not working in Basic UI when the icon is updated after a new state.

lolodomo:
For the bug with Basic UI, my feeling is that it could be solved by not removing the event listener.

lolodomo:
Was this PR tested ?

FlorianSW:
Sorry for not responding in the last time :(

> For information, why are they several files appearing as updated for Classic UI while I see in fact no real change, like for example extensions/ui/org.eclipse.smarthome.ui.classic/snippets/image.html ?

It seems, that I changed something in the format or some "invisible" things, as I saved this file and reverted them back (manually) and it seems git recognized some changes. The changes can be removed, of course.

> Not working for Classic UI.
Not working for Basic UI when the icon is updated after a state change.

What exactly do you mean with "not working for Classic UI", it's working fine for me.

> For the bug with Basic UI, my feeling is that it could be solved by not removing the event listener.

That would only work, if we're 100% sure, that the image none.png will always result in 200 OK, otherwise the brwoser would be in an indefinite loop replacin the source, executing the error handler, which replaces the source and so on. While I would be ok, when someone says, that the none.png will always return 200, I'm not sure, that we can be that sure for client-side problems, like no internet connection anymore, or whatever might result in an error.

I probably take a look to the code, which changes the icon, to implement adding the error handler again, which should also fix that problem :)

> Was this PR tested ?

Uhm, yes, of course? :)

lolodomo:
By not working for Classic UI, I mean Classic UI does not display the none icon when 404 is returned.
For Basic UI, take a look what happens when the state of an item is changed in an opened page and you will see that it does not work.

kaikreuzer:
@FlorianSW & @lolodomo How shall we proceed here?

FlorianSW:
Sorry, I totally missed this PR! :( @lolodomo If I'm not mistaken, I should've implemented all your points in the latest version of this PR. Would you please be so kind and recheck it?

kaikreuzer:
@lolodomo Could you please comment?

lolodomo:
I just tested with only a number item and it is apparently now working in Classic UI and in Android app.
Strangely, I can't use Basic UI with this branch, I only got a page with very big icons !!! and not the demo sitemap. I don't understand what's wrong and I am curious to know what is the origin of this page.

lolodomo:
The very big icons I see are the ones from the demo page. I don't know but the layout seems to be broken and Basic UI seems to waiting for something.
Maybe it is only a local package I should select but I don't find which one. Or Basic UI is really broken with this PR.

lolodomo:
A maven build of this branch leads to this error in my dev environment:
[ERROR] Failed to resolve target definition D:\dev\openhab2\git\smarthome\targetplatform\smarthome.target: Could not find "org.eclipse.equinox.sdk.feature.group/3.13.0.v20170531-1133" in the repositories of the current location

sjka:
That was fixed in master a while ago, so rebasing this branch would solve the target resolution problems.

Issue #3836, opened 2017-07-13
IP Cam does not work in sitemap


Mr-iX:
Since the update to OpenHAB 2.1 my connection to the IP Cam is broken.

My sitemap code:

Text item=IP_Cam_1 {
     Image url="http://192.168.56.3/snapshot.cgi?user=test&pwd=password" label="IP Cam 1"
}

I have also tried this:
Text label="IP Cam 1" {
     Image label="IP Cam 1" url="http://192.168.56.3/snapshot.cgi?user=test&pwd=password"
}

The openhab.log file:

2017-07-08 19:59:44.610 [ERROR] [.core.internal.folder.FolderObserver] - Error handling update of file '/etc/openhab2/sitemaps/Test.sitemap': index=2, size=2.
org.eclipse.emf.common.util.BasicEList$BasicIndexOutOfBoundsException: index=2, size=2
        at org.eclipse.emf.common.util.BasicEList.get(BasicEList.java:346)[63:org.eclipse.emf.common:2.11.0.v20150805-0538]
        at org.eclipse.smarthome.ui.internal.items.ItemUIRegistryImpl.getWidget(ItemUIRegistryImpl.java:516)[136:org.eclipse.smarthome.ui:0.9.0.b5]
        at org.eclipse.smarthome.io.rest.sitemap.SitemapSubscriptionService.collectWidgets(SitemapSubscriptionService.java:225)[117:org.eclipse.smarthome.io.rest.sitemap:0.9.0.b5]
        at org.eclipse.smarthome.io.rest.sitemap.SitemapSubscriptionService.modelChanged(SitemapSubscriptionService.java:276)[117:org.eclipse.smarthome.io.rest.sitemap:0.9.0.b5]
        at org.eclipse.smarthome.model.core.internal.ModelRepositoryImpl.notifyListeners(ModelRepositoryImpl.java:286)[122:org.eclipse.smarthome.model.core:0.9.0.b5]
        at org.eclipse.smarthome.model.core.internal.ModelRepositoryImpl.addOrRefreshModel(ModelRepositoryImpl.java:136)[122:org.eclipse.smarthome.model.core:0.9.0.b5]
        at org.eclipse.smarthome.model.core.internal.folder.FolderObserver.checkFile(FolderObserver.java:234)[122:org.eclipse.smarthome.model.core:0.9.0.b5]
[...]

PR #3846, opened 2017-07-17
Reload BasicUI if sitemap has been changed

triller-telekom:
This catches the IndexOutofBoundException from #3836


extensions/ui/org.eclipse.smarthome.ui.basic/web-src/smarthome.js
```diff
@@ -1709,6 +1709,16 @@
 				value,
 				title;
 
+			if(data.TYPE === "SITEMAP_CHANGED") {
```
sjka: isn't it also necessary to check if it is _this_ sitemap which changed? I.e. have you tried having to browser windows open with different sitemaps and check that only the affected one gets redirected to its main page?

bundles/io/org.eclipse.smarthome.io.rest.sitemap/src/main/java/org/eclipse/smarthome/io/rest/sitemap/internal/PageChangeListener.java
```diff
@@ -213,8 +213,13 @@ private boolean definesVisibility(Widget w, String name) {
         return false;
     }
 
-    public void sitemapContentChanged(EList<Widget> widgets) {
-        updateItemsAndWidgets(widgets);
+    public void sitemapContentChanged() {
+        for (SitemapSubscriptionCallback callback : distinctCallbacks) {
+            SitemapChangedEvent changeEvent = new SitemapChangedEvent();
```
sjka: Does it really make sense to create a new event for each callback and not create it once and send it to all of them?

extensions/ui/org.eclipse.smarthome.ui.basic/web-src/smarthome.js
triller-telekom: Nothing :) Grunt expects this format or it complains and does not generate the minified version of the file

triller-telekom: While fixing this bug I noticed that its almost impossible to detect in which page the user currently is if the the order of the pages has been changed. Thats why there is a reload for every sitemap change, also for browsers which are on the first page. Just replacing the widgets would have let to inconsistent views in the browser. This solution is now much cleaner.

bundles/ui/org.eclipse.smarthome.ui/src/main/java/org/eclipse/smarthome/ui/internal/items/ItemUIRegistryImpl.java
```diff
@@ -515,7 +515,7 @@ public Widget getWidget(Sitemap sitemap, String id) {
                     for (int i = 2; i < id.length(); i += 2) {
                         w = ((LinkableWidget) w).getChildren().get(Integer.valueOf(id.substring(i, i + 2)));
                     }
-                } catch (NumberFormatException e) {
+                } catch (NumberFormatException | IndexOutOfBoundsException e) {
```
triller-telekom: Alright, I will change it. Though the `NumberFormatException` was caught here I thought it doesn't hurt to catch the other one too...

bundles/io/org.eclipse.smarthome.io.rest.sitemap/src/main/java/org/eclipse/smarthome/io/rest/sitemap/internal/PageChangeListener.java
triller-telekom: Agreed, creating one event is sufficient.

sjka:
okay from my side.
@resetnow your call.

extensions/ui/org.eclipse.smarthome.ui.basic/web-src/smarthome.js
```diff
@@ -1709,6 +1709,16 @@
...
 				title;
 
+			if (data.TYPE === "SITEMAP_CHANGED") {
+				var oldLocation = window.location.href;
+				var parts = oldLocation.split("?");
+				if (parts.length > 1) {
+					window.location.href = parts[0] + "?sitemap="+data.sitemapName;
+				} else {
+					window.location.href = oldLocation;
+				}
```
resetnow:
```

_t.pause();
return;
```

extensions/ui/org.eclipse.smarthome.ui.basic/web-src/smarthome.js
```diff
@@ -1709,6 +1709,16 @@
 				value,
 				title;
 
+			if (data.TYPE === "SITEMAP_CHANGED") {
+				var oldLocation = window.location.href;
+				var parts = oldLocation.split("?");
+				if (parts.length > 1) {
+					window.location.href = parts[0] + "?sitemap="+data.sitemapName;
```
resetnow: Spaces:

```
"?sitemap=" + data
```

extensions/ui/org.eclipse.smarthome.ui.basic/web-src/smarthome.js
```diff
@@ -1709,6 +1709,16 @@
...
+				} else {
+					window.location.href = oldLocation;
```
resetnow:
```
window.location.reload(true);
```

kaikreuzer:
> This catches the IndexOutofBoundException from #3836 

Does it mean that #3836 is hence fixed now and can be closed?

triller-telekom:
Yes it can be closes because the sitemap problem is solved and the other problem regarding the non working D-Link camera has been extracted into #3870

PR #3850, opened 2017-07-18
use new HashMap<>() instead of Maps.newHashMap()

triller-telekom:
As discussed [here](https://github.com/eclipse/smarthome/pull/3804#pullrequestreview-50601832)

### DIFF
--- a/bundles/automation/org.eclipse.smarthome.automation.module.core/META-INF/MANIFEST.MF
+++ b/bundles/automation/org.eclipse.smarthome.automation.module.core/META-INF/MANIFEST.MF
@@ -4,8 +4,7 @@ Bundle-Name: Eclipse SmartHome Automation Module Core
 Bundle-SymbolicName: org.eclipse.smarthome.automation.module.core
 Bundle-Version: 0.9.0.qualifier
 Bundle-RequiredExecutionEnvironment: JavaSE-1.8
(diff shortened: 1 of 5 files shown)


sjka:
Please also fix the other occurrences and remove the package import.

triller-telekom:
I can't. The ´GenericEventTriggerHandler` uses `com.google.common.collect.ImmutableSet` which is not available in standard JDK afaik.

sjka:
> I can't.

You can!

```java
this.types = Collections.unmodifiableSet(new HashSet<>(Arrays.asList(((String) module.getConfiguration().get(CFG_EVENT_TYPES)).split(","))));
```

triller-telekom:
Thanks! Done...

maggu2810:
Just for your interest for the ones that would like to operate on streams:

```Java
final String[] strs = new String[] { "1", "2" };

final Set<String> set1 = Arrays.stream(strs).collect(Collectors.collectingAndThen(Collectors.toSet(), Collections::unmodifiableSet));

final Set<String> set2 = Collections.unmodifiableSet(new HashSet<>(Arrays.asList(strs)));
```

triller-telekom:
Thanks for the hint with streams. I'm still not used to them but the more often I see them I might get more familiar with them :)

PR #3855, opened 2017-07-19
Allow JSR223 scripts with same name in different directories

### LINKED ISSUE #3801: JSR223: Script names not scoped by subdirectory for script loading at system start.
Although the JSR223 scripting code loads files from subdirectories of `automation/jsr223`, it does not scope the file names by directory when loading scripts. For example if there is a file called 'script1.py' in two subdirectories, only one will be loaded at start time. If the scripts are added or modified after openHAB starts, then both will be loaded. @smerschjohann

triller-telekom:
Fixes #3801


triller-telekom:
@smerschjohann: Was there a specific reason for just comparing the filename and not the whole path? Because in the whole `ScriptFileWatcher` you are working on `URL`s which have the full path...

triller-telekom:
closed/reopened for travis rebuild

kaikreuzer:
I would assume that the reason was that he derived the UIDs from the file name and thus wanted to rule out any duplicates. As we are by now generating a UUID, I would hope that this is no issue anymore, so your change should be fine (@smerschjohann, please correct me, if I am wrong).

smerschjohann:
yes, you are right. It was a bad idea to rely only one the filename. I did not think about the "same name in different directories" case.

Thanks for this

Issue #3856, opened 2017-07-19
Missing import in MQTT bundle


maggu2810:
#3715 breaks the MQTT bundle in OSGi runtimes that does not use the bootdelegation or other approaches to provide non-declared packages.

The class "javax.naming.ConfigurationException" is imported (and used) several times in the code base:

* https://github.com/eclipse/smarthome/pull/3715/files#diff-a1df847a3da62dd495fe6468565c5b4aR15
* https://github.com/eclipse/smarthome/pull/3715/files#diff-7692fad7bdeefeb5ce45f59c59da6daaR18
* https://github.com/eclipse/smarthome/pull/3715/files#diff-e9972fc470d359e55670874a25547dc1R15
* https://github.com/eclipse/smarthome/pull/3715/files#diff-e603ec7941a6ef01820ea1a96823349bR14
* https://github.com/eclipse/smarthome/pull/3715/files#diff-280c5006c868df9f32e04321df8a6a8eR10

the only change that has applied to the manifest is about adding lazy bundle activation:

* https://github.com/eclipse/smarthome/pull/3715/files#diff-14f7b6770342b2fe60be508b7ee9cbc5R22

```text
2017-07-19 14:35:27,514 | ERROR | lixDispatchQueue | metatype                         | 2 - org.apache.felix.metatype - 1.1.2 | FrameworkEvent ERROR - org.apache.felix.metatype
org.osgi.framework.ServiceException: Service factory exception: javax/naming/ConfigurationException
	at org.apache.felix.framework.ServiceRegistrationImpl.getFactoryUnchecked(ServiceRegistrationImpl.java:352) [?:?]
	at org.apache.felix.framework.ServiceRegistrationImpl.getService(ServiceRegistrationImpl.java:247) [?:?]
	at org.apache.felix.framework.ServiceRegistry.getService(ServiceRegistry.java:350) [?:?]
	at org.apache.felix.framework.Felix.getService(Felix.java:3720) [?:?]
	at org.apache.felix.framework.BundleContextImpl.getService(BundleContextImpl.java:470) [?:?]
	at org.apache.felix.metatype.internal.ManagedServiceTracker.addingService(ManagedServiceTracker.java:52) [2:org.apache.felix.metatype:1.1.2]
	at org.osgi.util.tracker.ServiceTracker$Tracked.customizerAdding(ServiceTracker.java:941) [?:?]
[...]

PR #3857, opened 2017-07-19
add missing import to MQTT transport manifest

maggu2810:
Fixes: https://github.com/eclipse/smarthome/issues/3856


maggu2810:
I will do some further tests if other imports are missing, too.
So please wait before merging.

Issue #3756, opened 2017-06-27
[core] add more default system channel types


cweitkamp:
Based on these [comments](https://github.com/eclipse/smarthome/pull/3653#discussion_r121926314) I like to start a discussion about adding new default system channels. I like to create a list of frequently used channel types for each [channel category](https://www.eclipse.org/smarthome/documentation/development/bindings/thing-definition.html#channel-categories).

Some ideas:

| Channel Type      | Category      | Accessible Mode | Item Type | Other |
|-------------------|---------------|-----------|-|-|
| system.color             | ColorLight    | RW | Color     | |
| system.color-temperature | ColorLight    | RW | Dimmer    | pattern="%d", min="0", max="100" step="1" |
| system.brightness        | DimmableLight | RW | Dimmer    | pattern="%d %%", min="0", max="100" step="1" |
| system.volume            | SoundVolume   | RW | Dimmer    | pattern="%d %%", min="0", max="100", step="1" |
| system.mute              | SoundVolume   | RW | Switch    | |
| system.control           | Player        | RW | Player    | |
| system.stop              | Player        | RW | Switch    | |
| system.temperature | Temperature | R | Number | pattern="%.1f" |
| system.humidity | Humidity | R | Number | pattern="%d %%", min="0", max="100" |
| system.pressure | Pressure |  R | Number |pattern="%.1f" |
| system.wind-speed | Wind | R | Number | pattern="%.1f" |
| system.wind-direction | Wind | R | Number | pattern="%.0f °", min="0", max="360" |
| system.rain | Rain | R | Number | pattern="%.1f" |
| system.snow | Rain | R | Number | pattern="%.1f" |

triller-telekom:
Speaking of channel types, there is also an [ongoing discussion](https://github.com/eclipse/smarthome/pull/3629) about "group channel types" which at first glance should be the same as "channel types".

cweitkamp:
I updated my list and added some weather channels suggested by @lolodomo in PR #2649.

triller-telekom:
Since we will get quantities like "km/h", "m", "°C", etc. soon, I think you can remove them from the channels, see #1857

cweitkamp:
Yes, I agree. But shouldn't we define a default unit for each channel?

lolodomo:
[...]

PR #3858, opened 2017-07-19
[WIP] [core] Adapted DefaultSystemChannelTypeProvider to changed ESH API

cweitkamp:
Based on PR #3423 
Solution for 1. question of this [discussion](https://github.com/eclipse/smarthome/pull/3754#issuecomment-311349822) 
Preperation for ferature #3756 

I repeat my 2. question of this [discussion](https://github.com/eclipse/smarthome/pull/3754#issuecomment-311349822) - This is more a general architecture design question: Why are the default system channels hard coded into their own class? Isn't it more useful to define a xml-file for them? Like in every binding?

IMHO the `DefaultSystemChannelTypeProvider.java` class is located wrong in package `org.eclipse.smarthome.core.thing`. Should be moved to package `org.eclipse.smarthome.core.thing.xml` now.


sjka:
> Isn't it more useful to define a xml-file for them?

The whole framework is intended to work without the XML stuff. So it should also work when not deploying the XML related bundles (and obviously also bindings with don't rely in XMLs...).

cweitkamp:
Thanks. Now it is clear to me.

bundles/core/org.eclipse.smarthome.core.thing/src/main/java/org/eclipse/smarthome/core/thing/DefaultSystemChannelTypeProvider.java
```diff
@@ -100,110 +101,38 @@
...
     public DefaultSystemChannelTypeProvider() {
-        channelGroupTypes = Collections.emptyList();
-        channelTypes = Collections.unmodifiableCollection(
-                Arrays.asList(new ChannelType[] { SYSTEM_CHANNEL_SIGNAL_STRENGTH, SYSTEM_CHANNEL_LOW_BATTERY,
-                        SYSTEM_CHANNEL_BATTERY_LEVEL, SYSTEM_TRIGGER, SYSTEM_RAWBUTTON, SYSTEM_BUTTON }));
+        final Bundle bundle = FrameworkUtil.getBundle(DefaultSystemChannelTypeProvider.class);
 
+        add(bundle, SYSTEM_CHANNEL_SIGNAL_STRENGTH);
+        add(bundle, SYSTEM_CHANNEL_LOW_BATTERY);
+        add(bundle, SYSTEM_CHANNEL_LOW_BATTERY);
```
cweitkamp: double

bundles/core/org.eclipse.smarthome.core.thing/META-INF/MANIFEST.MF
```diff
@@ -15,6 +15,7 @@ Import-Package: com.google.common.base,
  org.eclipse.smarthome.config.core.dto,
  org.eclipse.smarthome.config.core.status,
  org.eclipse.smarthome.config.core.validation,
+ org.eclipse.smarthome.config.xml,
```
kaikreuzer: Sorry, that won't work. As @sjka mentioned, the framework should also work without xml support and thus core.thing must not have a dependency on config.xml.

cweitkamp:
Any suggestions how we can solve this in a better way? I don't like the idea to have duplicate lines of code in different places.

kaikreuzer:
I didn't look into the details of what you exactly need there, but it might be possible to extract a common abstract base class that could be put in config.core, so that this would be the only dependency you need then.

kaikreuzer:
@cweitkamp Are you waiting for some further and better suggestions or did you just put this aside for the moment?

cweitkamp:
@kaikreuzer I put it on-hold. The current solution is working and this is a nice to have refactoring. If anyone else is interested to work on this he should feel free to do so.

PR #3865, opened 2017-07-20
Updated Californium to 1.0.6

triller-telekom:
Signed-off-by: Stefan Triller <stefan.triller@telekom.de>

PR #3873, opened 2017-07-21
increase coap version in karaf features

sjka:
relates to #3865


maggu2810:
I will do a test build on my machine. It seems the Eclipse infra is not reachable from Travis CI all the time (or it receives something but an HTTP error code).

PR #3876, opened 2017-07-24
Mqtt things

davidgraeff:
Implementation of the second idea mentioned in #3812.

This has a strong dependency on #4146 and a soft dependency on #4142 for graphical configuration.

## Abstract
At this very moment there exist no means to actually connect mqtt topics to ESH things. There are a lot of commercial and non-commercial (Owntracks, EasyESP, Homie) Mqtt products out there who deserve their own binding.

But for a start I propose a simple generic Mqtt binding where you can bind topics to things (very much like the Openhab 1.x Mqtt binding, but modern) with the following features:

* List all configured brokers of the io.transport.mqtt.MqttService and let the user add them as bridges.
* Manually add a thing (Text, Number, OnOff-thing) and bind it to a Mqtt state and/or command topic.
* Allow to use a transformation pattern to extract a state from a formatted response like JSON. An example would be the pattern `JSONPATH:$.device.status.temperature` for an incoming message of `{device: {status: { temperature: 23.2 }}}`.
* The number thing can have a configured min, max and step so that its channel is suitable for Dimmer/Rollershutter item types.
* The onoff thing allows to configure what should be interpreted as the states and send for the states "ON" and "OFF".

## Discussion
I have added four things with one channel each (the value channel). The following item types can be used with the channel:

* TextThing: String
* NumberThing: Number
* PercentThing: Rollershutter, Dimmer
* OnOffThing: Contact, Switch

I'm not sure if we need more things to cover the basic needs, but I might have overlooked use cases.

Cheers,
David


davidgraeff:
@htreu Can you have a look at this PR? It is a simpler and smaller PR than the Mqtt Broker configuration PR and there are less (or none) subjects that need to be discussed architecture-wise.

ThomDietrich:
@davidgraeff
your arguments against a auto-discovery make perfect sense. Do you think a manual discovery process in which topics of all retained messages and of new messages as they come in would help users to add things and items? Yes I'm talking about a # subscription 😄 Sounds like an easy task and a huge time safer... 

Also: Don't you think the binding should support the other noteworthy MQTT message features retained flag and qos? Who knows what the user might need them for...

davidgraeff:
I want to keep this PR simple for the start. Additional features can be added later. But it is more likely to get this merged faster if the PR is smaller.

ThomDietrich:
Reasonable and I'd also root for a faster merge. 

QoS and Retained are both properties on per message basis actually, even though they are often meaningful on a per topic basis. 

As you see fit! Thanks for working on this crucial feature.

kaikreuzer:
As mentioned in https://github.com/eclipse/smarthome/issues/3812#issuecomment-325296730, there is no direct dependency on Paper UI work for merging this PR.

> The implementation is done, but changes are requested, see https://github.com/eclipse/smarthome/issues/3812#issuecomment-320660894.

This is the outstanding part that should be addressed before the PR can be reviewed & merged.

davidgraeff:
Adapted to dynamic channels. Tests are not adapted yet. Is the current implementation how it should look like? (Especially the ThingHandler)

weswitt:
Is this PR going to be accepted? Would like to use this.

htreu:
Hi @davidgraeff, this one also seems to be nearly finished. Here Paper UI support is also ready for extending Channels on Things. Would be nice to see your engagement again :-)

extensions/binding/org.eclipse.smarthome.binding.mqttgeneric/src/main/java/org/eclipse/smarthome/binding/mqttgeneric/handler/MqttThingHandler.java
```diff
@@ -0,0 +1,200 @@
...
+                int index = config.transformationPattern.indexOf(':');
+                if (index == -1) {
+                    throw new IllegalArgumentException(
+                            "The transformation pattern must consist of the type and the pattern separated by a colon");
+                }
+                String type = config.transformationPattern.substring(0, index).toUpperCase();
+                config.transformationPattern = config.transformationPattern.substring(index + 1);
+                config.transformationService = getTransformationService(type);
+                if (config.transformationService == null) {
+                    throw new IllegalArgumentException("The transformation type is unknown: " + type);
```
sjka: This is not so much of a good idea... 

1. Don't keep references on OSGi services (which is essentially what you are doing here, although `TransformationHelper` unfortunately hides this fact pretty well)
2. The transformation service implementation which is configured for this channel _might_ not/currently be available but will be later on. If you let initialization fail because of this, the handler won't get a chance to recover from this.

Therefore, please only remember the transformation *name* and rather grab a fresh reference every time.

extensions/binding/org.eclipse.smarthome.binding.mqttgeneric/src/main/java/org/eclipse/smarthome/binding/mqttgeneric/internal/NumberValue.java
```diff
@@ -0,0 +1,72 @@
...
+        if (isFloat) {
+            return numberValue.toString();
+        } else {
+            return String.valueOf(numberValue.intValue());
+        }
+    }
+
+    @Override
+    public State update(String updatedValue) throws IllegalArgumentException {
+        numberValue = PercentType.valueOf(updatedValue);
```
sjka: Should be `DecimalType` here I assume?

extensions/binding/org.eclipse.smarthome.binding.mqttgeneric/src/main/java/org/eclipse/smarthome/binding/mqttgeneric/internal/NumberValue.java
```diff
@@ -0,0 +1,72 @@
...
+            numberValue = ((PercentType) command);
+        } else if (command instanceof IncreaseDecreaseType) {
+            double v;
+            if (((IncreaseDecreaseType) command) == IncreaseDecreaseType.INCREASE) {
+                v = numberValue.doubleValue() - step;
+            } else {
+                v = numberValue.doubleValue() + step;
+            }
+            numberValue = new PercentType(new BigDecimal(v));
+        } else {
```
sjka: `DecimalType` is missing here

extensions/binding/org.eclipse.smarthome.binding.mqttgeneric/src/main/java/org/eclipse/smarthome/binding/mqttgeneric/internal/NumberValue.java
```diff
@@ -0,0 +1,72 @@
...
+
+    @Override
+    public State getValue() {
+        return numberValue;
+    }
+
+    @Override
+    public String update(Command command) throws IllegalArgumentException {
+        if (command instanceof StringType) {
+            numberValue = PercentType.valueOf(((StringType) command).toString());
```
sjka: `DecimalType`

sjka:
I took the freedom to rebase this PR and add some commits on-top, filed as #5101. This is _not_ meant to be a hostile take-over, just meant as an offer to push this binding forward. You are more than welcome to continue on this PR here and either implement the review feedback and other amendments yourself or take them from #5101. In the meantime I'm going to close this PR in order to prevent confusion.

davidgraeff:
Haha all good, no offense taken. Good to see progress :)

PR #3879, opened 2017-07-24
[docu] installer screenshot with oxygen

sjka:


maggu2810:
Should we use the mailing list for announcements like "The Eclipse IDE setup has been migrated from Neon to Oxygen. Please update your IDE."? So, the mailing list smarthome-dev could be used to inform every subscribed smarthome developer (that does not read every issue / PR).

sjka:
Thanks! I just pushed an update to the website.

PR #3808, opened 2017-07-06
Improve Sonos notification timeout handling

ivivanov-bg:
1. Extract the notification timeout as a thing configuration parameter
2. Increase the default notification timeout to 40 sec. (as 20 s is too short time)

PR #889, opened 2016-01-20
added unit to configuration description parameters according to issue…

tomhoefer:
… #741

Signed-off-by: Thomas Höfer <t.hoefer@telekom.de>

PR #3883, opened 2017-07-24
Sonos binding: fix bug with notification timeout handling

lolodomo:
Fix bug introduced by PR #3808

### DIFF
--- a/extensions/binding/org.eclipse.smarthome.binding.sonos/ESH-INF/config/config.xml
+++ b/extensions/binding/org.eclipse.smarthome.binding.sonos/ESH-INF/config/config.xml
@@ -9,11 +9,11 @@
 			<description>The UDN identifies the Zone Player.</description>
 			<required>true</required>
 		</parameter>
-		<parameter name="notificationTimeout" type="integer">
+		<parameter name="notificationTimeout" type="integer" unit="s">
 			<label>Notification Timeout</label>
-			<description>Specifies the amount of time for which the notification sound will be played</description>
-			<default>40</default>
-			<required>true</required>
+			<description>Specifies the amount of time in seconds for which the notification sound will be played</description>
+			<unitLabel>second</unitLabel>
+			<default>20</default>
 		</parameter>
 		<parameter name="refresh" type="integer">
 			<label>Refresh interval</label>
--- a/extensions/binding/org.eclipse.smarthome.binding.sonos/src/main/java/org/eclipse/smarthome/binding/sonos/handler/ZonePlayerHandler.java
+++ b/extensions/binding/org.eclipse.smarthome.binding.sonos/src/main/java/org/eclipse/smarthome/binding/sonos/handler/ZonePlayerHandler.java
@@ -37,7 +37,6 @@ import org.eclipse.smarthome.binding.sonos.internal.SonosMetaData;
 import org.eclipse.smarthome.binding.sonos.internal.SonosXMLParser;
 import org.eclipse.smarthome.binding.sonos.internal.SonosZoneGroup;
 import org.eclipse.smarthome.binding.sonos.internal.SonosZonePlayerState;
-import org.eclipse.smarthome.config.core.Configuration;
 import org.eclipse.smarthome.config.discovery.DiscoveryServiceRegistry;
 import org.eclipse.smarthome.core.library.types.DecimalType;
 import org.eclipse.smarthome.core.library.types.IncreaseDecreaseType;
@@ -101,12 +100,12 @@ public class ZonePlayerHandler extends BaseThingHandler implements UpnpIOPartici
     private static final int SOCKET_TIMEOUT = 5000;
 
     /**
-     * Default notification timeout
+     * Default notification timeout (in seconds)
      */
-    private static final Integer DEFAULT_NOTIFICATION_TIMEOUT = 40000;
-    
+    private static final Integer DEFAULT_NOTIFICATION_TIMEOUT = 20;
+
     /**
-     * configurable notification timeout
+     * configurable notification timeout (in seconds)
      */
     private Integer notificationTimeout = null;
 
@@ -212,15 +211,12 @@ public class ZonePlayerHandler extends BaseThingHandler implements UpnpIOPartici
 
         if (getUDN() != null) {
             onUpdate();
-            
-            if (getConfigAs(ZonePlayerConfiguration.class).notificationTimeout == null) {
-                Configuration c = editConfiguration();
-                c.put(ZonePlayerConfiguration.NOTIFICATION_TIMEOUT, DEFAULT_NOTIFICATION_TIMEOUT);
-                updateConfiguration(c);
-            }
-            
+
             this.notificationTimeout = getConfigAs(ZonePlayerConfiguration.class).notificationTimeout;
-            
+            if (this.notificationTimeout == null) {
+                this.notificationTimeout = DEFAULT_NOTIFICATION_TIMEOUT;
+            }
+
             super.initialize();
         } else {
             updateStatus(ThingStatus.OFFLINE, ThingStatusDetail.CONFIGURATION_ERROR);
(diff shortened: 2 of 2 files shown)


extensions/binding/org.eclipse.smarthome.binding.sonos/ESH-INF/config/config.xml
```diff
@@ -11,9 +11,8 @@
 		</parameter>
 		<parameter name="notificationTimeout" type="integer">
 			<label>Notification Timeout</label>
-			<description>Specifies the amount of time for which the notification sound will be played</description>
+			<description>Specifies the amount of time in seconds for which the notification sound will be played</description>
```
maggu2810: Why not using the `unit` attribute and the `unitLabel` element?

extensions/binding/org.eclipse.smarthome.binding.sonos/src/main/java/org/eclipse/smarthome/binding/sonos/handler/ZonePlayerHandler.java
```diff
@@ -101,12 +101,12 @@
     private static final int SOCKET_TIMEOUT = 5000;
 
     /**
-     * Default notification timeout
+     * Default notification timeout (in seconds)
      */
-    private static final Integer DEFAULT_NOTIFICATION_TIMEOUT = 40000;
-    
+    private static final Integer DEFAULT_NOTIFICATION_TIMEOUT = 20;
```
maggu2810: The config-description XML is using a default of 40 seconds, you change it here to 20 seconds.
I don't know what is better 20 or 40.

extensions/binding/org.eclipse.smarthome.binding.sonos/src/main/java/org/eclipse/smarthome/binding/sonos/handler/ZonePlayerHandler.java
```diff
@@ -2388,7 +2388,7 @@ private void waitForFinishedNotification() {
         // check Sonos state events to determine the end of the notification sound
         String notificationTitle = stateMap.get("CurrentTitle");
         long playstart = System.currentTimeMillis();
-        while (System.currentTimeMillis() - playstart < this.notificationTimeout.longValue()) {
+        while (System.currentTimeMillis() - playstart < (this.notificationTimeout.longValue() * 1000)) {
```
maggu2810: Why adding unneeded braces?

extensions/binding/org.eclipse.smarthome.binding.sonos/src/main/java/org/eclipse/smarthome/binding/sonos/handler/ZonePlayerHandler.java
```diff
@@ -2407,7 +2407,7 @@ private void waitForTransportState(String state) {
             while (!stateMap.get("TransportState").equals(state)) {
                 try {
                     Thread.sleep(50);
-                    if (System.currentTimeMillis() - start > this.notificationTimeout.longValue()) {
+                    if (System.currentTimeMillis() - start > (this.notificationTimeout.longValue() * 1000)) {
```
maggu2810: Why adding unneeded braces?

extensions/binding/org.eclipse.smarthome.binding.sonos/src/main/java/org/eclipse/smarthome/binding/sonos/handler/ZonePlayerHandler.java
```diff
@@ -2423,7 +2423,7 @@ private void waitForNotTransportState(String state) {
             while (stateMap.get("TransportState").equals(state)) {
                 try {
                     Thread.sleep(50);
-                    if (System.currentTimeMillis() - start > this.notificationTimeout.longValue()) {
+                    if (System.currentTimeMillis() - start > (this.notificationTimeout.longValue() * 1000)) {
```
maggu2810: Why adding unneeded braces?

extensions/binding/org.eclipse.smarthome.binding.sonos/ESH-INF/config/config.xml
```diff
@@ -11,9 +11,8 @@
 		</parameter>
 		<parameter name="notificationTimeout" type="integer">
 			<label>Notification Timeout</label>
-			<description>Specifies the amount of time for which the notification sound will be played</description>
-			<default>40</default>
-			<required>true</required>
+			<description>Specifies the amount of time in seconds for which the notification sound will be played</description>
```
maggu2810: Why not using the `unit` atrribute and the `unitLabel` element?

extensions/binding/org.eclipse.smarthome.binding.sonos/src/main/java/org/eclipse/smarthome/binding/sonos/handler/ZonePlayerHandler.java
lolodomo: I changed the XML value after your comment.

extensions/binding/org.eclipse.smarthome.binding.sonos/ESH-INF/config/config.xml
lolodomo: Because I don't know this ability ;) If it is described somewhere, I will try to change that.

lolodomo:
@maggu2810 : I changed the config setting to include `unit` and `unitLabel` but I have not tested again after this change.

lolodomo:
Please wait, there is another thing weird with PR 3808, that is the thing update. If the thing setting is not set, the binding should just use its default value without the need to update the thing setting.

lolodomo:
Last change done.

lolodomo:
Regarding the timeout value, 20s looks like a reasonable duration for notifications. Users needing very long notification will oncrease the setting.

extensions/binding/org.eclipse.smarthome.binding.sonos/ESH-INF/config/config.xml
```diff
@@ -9,11 +9,11 @@
...
 			<required>true</required>
 		</parameter>
-		<parameter name="notificationTimeout" type="integer">
+		<parameter name="notificationTimeout" type="integer" unit="s">
 			<label>Notification Timeout</label>
-			<description>Specifies the amount of time for which the notification sound will be played</description>
-			<default>40</default>
-			<required>true</required>
+			<description>Specifies the amount of time in seconds for which the notification sound will be played</description>
+			<unitLabel>second</unitLabel>
```
kaikreuzer: Isn't "s" the normal label? "5 s" looks better to me than "5 second"...

lolodomo: Be free to change it if you think it is better.

extensions/binding/org.eclipse.smarthome.binding.sonos/ESH-INF/config/config.xml
cweitkamp: Nice to know. I found PR [#889](https://github.com/eclipse/smarthome/commit/091a79696626da2298968cf4ffbe82bdadab6a66) but no directly hint in the docs.

extensions/binding/org.eclipse.smarthome.binding.sonos/ESH-INF/config/config.xml
cweitkamp: According to the code the tag `unitLabel` is ignored if the attribute `unit` is set (see [https://github.com/eclipse/smarthome/blob/master/bundles/config/org.eclipse.smarthome.config.xml/src/main/java/org/eclipse/smarthome/config/xml/ConfigDescriptionParameterConverter.java#L120](https://github.com/eclipse/smarthome/blob/master/bundles/config/org.eclipse.smarthome.config.xml/src/main/java/org/eclipse/smarthome/config/xml/ConfigDescriptionParameterConverter.java#L120). Isn't it?

kaikreuzer: Good catch, so this line should be removed completely!

PR #3888, opened 2017-07-25
Changed logger message instantiating NPE

svilenvul:
This was marked by PMD as error by the AvoidThrowingNullPointerException rule. Obviously the NPE is not thrown :), but anyway the instance here seems to be unnecessary.


svilenvul:
OK, I will add it as an exception then.

PR #3890, opened 2017-07-25
export more packages of rest.core

maggu2810:
The rest.core bundle already exports:
* org.eclipse.smarthome.io.rest.core.item
* org.eclipse.smarthome.io.rest.core.thing

I assume the packages are exported so other implementations can make use
of the DTO classes without the need to implement and maintain that
classes themselves.

The same should be true for e.g. ConfigurationService and
ConfigurableServiceDTO.

### DIFF
--- a/bundles/io/org.eclipse.smarthome.io.rest.core/META-INF/MANIFEST.MF
+++ b/bundles/io/org.eclipse.smarthome.io.rest.core/META-INF/MANIFEST.MF
@@ -54,7 +54,9 @@ Import-Package: com.google.common.base,
  org.osgi.service.cm,
  org.osgi.service.component,
  org.slf4j
-Export-Package: org.eclipse.smarthome.io.rest.core.item,
+Export-Package: org.eclipse.smarthome.io.rest.core.config,
+ org.eclipse.smarthome.io.rest.core.service,
+ org.eclipse.smarthome.io.rest.core.item,
  org.eclipse.smarthome.io.rest.core.thing
 Service-Component: OSGI-INF/*.xml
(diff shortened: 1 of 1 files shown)


kaikreuzer:
The only reason for those exports that I am aware of is the split of the SitemapResource in a separate bundle, which needs access to the EnrichedItemDTOs (but it does not extend them). In general, I would tend to say that especially the *Resource classes should be internal and the same should apply for most EnrichedDTOs.

Do you have a concrete use case where you need to access ConfigurationService and
ConfigurableServiceDTO? Extending them might be difficult as they are registered as services, so you would end up with multiple similar services... Maybe refactoring to an abstract class or interface is more applicable for your requirements?

maggu2810:
> In general, I would tend to say that especially the *Resource classes should be internal

Okay for me. There is no need that an internal packages contain "impl" or "internal" but IMHO it is a nice convention. But such a change is not part of this PR. "items" and "things" already contains also Resource and are exported. Moving all Resources to non-exported packages could break other bundles and is not part of this PR by intention.

> Do you have a concrete use case where you need to access ConfigurationService and
ConfigurableServiceDTO?

I would like to reuse that classes (the same implementation) in my REST interface implementation.

kaikreuzer:
My point was rather that nothing of that bundle is meant to be exposed - so instead of exporting more, I would prefer to export less of it...

But ok, I do not have any issue with ConfigurationService and
ConfigurableServiceDTO and I agree that moving the *Resource and Enriched*DTOs to internal packages is out of scope of this PR.

PR #3896, opened 2017-07-26
Bug fixing: 1) set default value and normalize config values of modul…

danchom:
…es of rules when they are added into rule registry

2) create a copy of modules config values when a runtime modules are created.

### DIFF
--- a/bundles/automation/org.eclipse.smarthome.automation.core/src/main/java/org/eclipse/smarthome/automation/core/internal/RuleRegistryImpl.java
+++ b/bundles/automation/org.eclipse.smarthome.automation.core/src/main/java/org/eclipse/smarthome/automation/core/internal/RuleRegistryImpl.java
@@ -234,10 +234,10 @@ public class RuleRegistryImpl extends AbstractRegistry<Rule, String, RuleProvide
         String rUID = rule.getUID();
         if (rUID == null) {
             rUID = ruleEngine.getUniqueId();
-            super.add(initRuleId(rUID, rule));
-        } else {
-            super.add(rule);
+            rule = initRuleId(rUID, rule);
         }
+        ruleEngine.normalizeRuleConfigurations(rule);
+        super.add(rule);
         Rule ruleCopy = get(rUID);
         if (ruleCopy != null) {
             return ruleCopy;
@@ -523,6 +523,7 @@ public class RuleRegistryImpl extends AbstractRegistry<Rule, String, RuleProvide
             String rUID = ruleEngine.getUniqueId();
             ruleWithUID = initRuleId(rUID, element);
         }
+        ruleEngine.normalizeRuleConfigurations(element);
         super.added(provider, ruleWithUID);
         updateRuleByTemplate(provider, ruleWithUID);
     }
--- a/bundles/automation/org.eclipse.smarthome.automation.core/src/main/java/org/eclipse/smarthome/automation/core/internal/RuntimeCondition.java
+++ b/bundles/automation/org.eclipse.smarthome.automation.core/src/main/java/org/eclipse/smarthome/automation/core/internal/RuntimeCondition.java
@@ -13,6 +13,7 @@ import java.util.Set;
 
 import org.eclipse.smarthome.automation.Condition;
 import org.eclipse.smarthome.automation.handler.ConditionHandler;
+import org.eclipse.smarthome.config.core.Configuration;
 import org.slf4j.LoggerFactory;
 
 /**
@@ -26,7 +27,8 @@ public class RuntimeCondition extends Condition {
     private Set<Connection> connections;
 
     public RuntimeCondition(Condition condition) {
-        super(condition.getId(), condition.getTypeUID(), condition.getConfiguration(), condition.getInputs());
+        super(condition.getId(), condition.getTypeUID(),
+                new Configuration(condition.getConfiguration().getProperties()), condition.getInputs());
         setConnections(Connection.getConnections(condition.getInputs(), LoggerFactory.getLogger(getClass())));
         setLabel(condition.getLabel());
         setDescription(condition.getDescription());
--- a/bundles/automation/org.eclipse.smarthome.automation.core/src/main/java/org/eclipse/smarthome/automation/core/internal/RuleEngine.java
+++ b/bundles/automation/org.eclipse.smarthome.automation.core/src/main/java/org/eclipse/smarthome/automation/core/internal/RuleEngine.java
@@ -1429,7 +1429,7 @@ public class RuleEngine implements RegistryChangeListener<ModuleType> {
         }
     }
 
-    private void normalizeRuleConfigurations(Rule rule) {
+    protected void normalizeRuleConfigurations(Rule rule) {
         List<ConfigDescriptionParameter> configDescriptions = rule.getConfigurationDescriptions();
         Map<String, ConfigDescriptionParameter> mapConfigDescriptions;
         if (configDescriptions != null) {
--- a/bundles/automation/org.eclipse.smarthome.automation.core/src/main/java/org/eclipse/smarthome/automation/core/internal/RuntimeAction.java
+++ b/bundles/automation/org.eclipse.smarthome.automation.core/src/main/java/org/eclipse/smarthome/automation/core/internal/RuntimeAction.java
@@ -11,6 +11,7 @@ import java.util.Set;
 
 import org.eclipse.smarthome.automation.Action;
 import org.eclipse.smarthome.automation.handler.ActionHandler;
+import org.eclipse.smarthome.config.core.Configuration;
 import org.slf4j.LoggerFactory;
 
 /**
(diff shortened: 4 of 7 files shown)


danchom:
ping

bundles/automation/org.eclipse.smarthome.automation.core/src/main/java/org/eclipse/smarthome/automation/core/internal/RuleRegistryImpl.java
```diff
@@ -234,10 +234,10 @@ public Rule add(Rule rule) {
         String rUID = rule.getUID();
         if (rUID == null) {
             rUID = ruleEngine.getUniqueId();
-            super.add(initRuleId(rUID, rule));
-        } else {
-            super.add(rule);
+            rule = initRuleId(rUID, rule);
         }
+        ruleEngine.normalizeRuleConfigurations(rule);
```
kaikreuzer: @sjka How do we do config normalisation on the Thing API? Don't we already expect it to be normalised when being added to the registry? We should imho make sure that we are using the same principals in both situations.

bundles/automation/org.eclipse.smarthome.automation.core/src/main/java/org/eclipse/smarthome/automation/core/internal/RuntimeCondition.java
```diff
@@ -26,7 +27,8 @@
     private Set<Connection> connections;
 
     public RuntimeCondition(Condition condition) {
-        super(condition.getId(), condition.getTypeUID(), condition.getConfiguration(), condition.getInputs());
+        super(condition.getId(), condition.getTypeUID(),
+                new Configuration(condition.getConfiguration().getProperties()), condition.getInputs());
```
kaikreuzer: This looks pretty ugly to have in all the constructors.
Two questions:
1. Why do we need a copy of the passed configuration at all?
2. If we need it, why isn't this simply done once in `Module.setConfiguration()`?

bundles/automation/org.eclipse.smarthome.automation.core/src/main/java/org/eclipse/smarthome/automation/core/internal/RuleRegistryImpl.java
danchom: >Don't we already expect it to be normalised when being added to the registry

I think not, otherwise we don't need the normalization at all. 

Actually the problem is related with the default value of configuration property. It's value  is defined by its ModuleType as string, but this string should be normalize to the type defined by the ConfigDescriptionParameter of the property. The property can be  can be Decimal or  multiple and etc.

kaikreuzer: I am not saying that this here is wrong. I am saying that we need to make sure that we follow the same principle across the different places.

bundles/automation/org.eclipse.smarthome.automation.core/src/main/java/org/eclipse/smarthome/automation/core/internal/RuntimeCondition.java
danchom: 2) The runtime objects (RuntimeCondition, RuntimeRule and etc) are real object used by the RuleEngine. We don't have a setters because these are public immutable objects and must be verified before passing them to the RuleEngine. Otherwise we can end up with partially configured object which can brake the engine. Of course we can use a builder to avoid big constructor.
1) We have to copy configuration values otherwise the user can change configuration of immutable runtime object during the rule execution with unexpected results moreover it will skip entire rule update logic.

kaikreuzer: > We don't have a setters because these are public immutable objects

Err - they all extend `Module`, which has public setters (incl. the mentioned `setConfiguration`. But anyhow, `setConfiguration` could even be private and could be called in the constructor of `Module` - that was my question (2) above.

bundles/automation/org.eclipse.smarthome.automation.core/src/main/java/org/eclipse/smarthome/automation/core/internal/RuleRegistryImpl.java
sjka: With Things, we indeed try to normalize them as early as possible, i.e. once we get hold of the configuration for the first time (e.g. in the `@PUT` annotated methods in REST resources or right after reading the DSLs files from disk. The goal really is to have only normalized configurations in the ThingRegistry.

danchom: My point of view is that ThingRegistry is a public service which has a public method updateConfiguration(). There is no guarantee that every one will call the method with normalize parameters.

bundles/automation/org.eclipse.smarthome.automation.core/src/main/java/org/eclipse/smarthome/automation/core/internal/RuntimeCondition.java
danchom: Yes you are right. We can use this method to update the configuration. The runtime object are not public and there for the method can't be call outside of rule engine.

danchom: Changing setConfiguration  visibility to private is backward incompatible change of java api. Also I think in that way the user has opportunity to change the configuration of the module several times  without creating a new module instance, but this can be happen only before adding the rule into rule engine. After that the user can't change module's configuration properties until the rule's update operation is called.

bundles/automation/org.eclipse.smarthome.automation.core/src/main/java/org/eclipse/smarthome/automation/core/internal/RuleRegistryImpl.java
kaikreuzer: The thing that worries me here is that I would expect the rule engine to treat the passed Rule object as immutable and not to do any changes to it. If we are concerned that we might see configurations that are not normalized (and we cannot deal with them), we could throw IAEs to make sure that only expected stuff is reached in. Nonetheless, I also see the point that the framework could do this work if it anyhow knows what needs to be done. 
@sjka & @maggu2810  What do you think how we should deal with that?

bundles/automation/org.eclipse.smarthome.automation.core/src/main/java/org/eclipse/smarthome/automation/core/internal/RuntimeCondition.java
kaikreuzer: Not sure why this then speaks against doing the copy within `Module.setConfiguration()`...?

bundles/automation/org.eclipse.smarthome.automation.core/src/main/java/org/eclipse/smarthome/automation/core/internal/RuleRegistryImpl.java
maggu2810: > My point of view is that ThingRegistry is a public service which has a public method updateConfiguration(). There is no guarantee that every one will call the method with normalize parameters.

￼Sure, no guarantee, but the public method could state it requires a normalized configuration.

> we could throw IAEs to make sure that only expected stuff is reached in

Okay for me.

bundles/automation/org.eclipse.smarthome.automation.core/src/main/java/org/eclipse/smarthome/automation/core/internal/RuntimeCondition.java
kaikreuzer: @adimova Could you comment on this?

adimova: I agree to copy the passed configuration and this to be done once in `Module.setConfiguration()`.
This is already done in #4468.

bundles/automation/org.eclipse.smarthome.automation.core/src/main/java/org/eclipse/smarthome/automation/core/internal/RuleRegistryImpl.java
adimova: > With Things, we indeed try to normalize them as early as possible

It is the same for Rules (there is a bug in `ConfigurationDeserializer`, but I'll make a separate PR for it). Additionally, in Rule Registry we also need normalization, because the rule configuration is resolved with default values and references. After the resolving, normalization is needed because the default values are always strings. This however will be fixed with #4468, so this PR can be closed without merge.

maggu2810:
> This however will be fixed with #4468, so this PR can be closed without merge.

Issue #3863, opened 2017-07-20
Device specific IDs to automatically ignore things in the inbox


afuechsel:
**Current situation:** If creating a thing using the inbox, the UID of this thing normally contains some device-specific identification (serial number, MAC address, whatsoever), e.g. `hue:bridge:001788153957`. If creating this thing _manually_, either from a DSL or the Paper UI, the thing UID is manually defined and not necessarily the same as in the inbox. As a consequence, the thing, although already added to the framework will appear in the inbox and could be added again. A user may click on "Ignore" for that inbox entry to avoid this situation. 

**Proposal:** This behavior is odd and could be automated by introducing a configurable service, that _auto-ignores_ existing things in the inbox. A binding must implement some identification service to create a _device specific ID_, that clearly identifies this device (e.g. serial number, MAC address, ...). The _auto-ignore service_ will be called whenever a thing changes into ONLINE state and checks whether the inbox already contains an entry with the same _device-specific ID_ and calls IGNORE for this entry. 

The _auto-ignore service_ uses only standard features of ESH and could be switched off if not required.

sjka:
> A binding must implement some identification service 

Why not adding a method to the ThingHandler, with a default implementation that simply returns `null`? We strive to hide OSGi as much as possible from binding developers, so we shouldn't stick it into their faces for this feature.

> that clearly identifies this device

Just to be clear: this does not necessarily have to be a serial number, nor does it need to be _globally_ unique or anything. The distinguishing characteristic that is unique within the system and is used by the thing handler anyway in order to communicate with the correct device would suffice.

> The auto-ignore service will be called whenever a thing changes into ONLINE state

By means of registering itself as an event listener, right?

afuechsel:
> Why not adding a method to the ThingHandler, with a default implementation that simply returns null? 

Yes, isn't an empty string or Optional.empty better? Anyway, I agree. 

> By means of registering itself as an event listener, right?

Yes

sjka:
> Yes, isn't an empty string or Optional.empty better? Anyway, I agree.

No.

afuechsel:
Just discovered a flaw in the concept: auto-ignoring when a thing comes ONLINE works fine, we need also to perform the same check in the discovery, so when a result is going to be added to the inbox it needs to be checked not only if a thing with the uid is already existing, but also a thing with the same deviceId. 

[...]

PR #3899, opened 2017-07-27
Added representation property to thing type and used this in the inbox auto ignore service

afuechsel:
implemented device id in the hue binding
added a representation property for a hue light

Fixes: #3863


extensions/binding/org.eclipse.smarthome.binding.hue/src/main/java/org/eclipse/smarthome/binding/hue/handler/HueBridgeHandler.java
```diff
@@ -519,4 +519,13 @@ private boolean isEqual(State state1, State state2) {
 
         return configStatusMessages;
     }
+
+    @Override
+    public String getDeviceId() {
+        try {
+            return bridge.getFullConfig().getConfig().getMACAddress().replaceAll(":", "").toLowerCase();
```
sjka: The [HueThingHandlerFactory](https://github.com/eclipse/smarthome/blob/0d2b3b8c88cdc5cfa4efe3b44584f67712e93e6b/extensions/binding/org.eclipse.smarthome.binding.hue/src/main/java/org/eclipse/smarthome/binding/hue/internal/HueThingHandlerFactory.java#L68) reads it from the configuration, the [HueBridgeDiscoveryParticipant](https://github.com/eclipse/smarthome/blob/0d2b3b8c88cdc5cfa4efe3b44584f67712e93e6b/extensions/binding/org.eclipse.smarthome.binding.hue/src/main/java/org/eclipse/smarthome/binding/hue/internal/discovery/HueBridgeDiscoveryParticipant.java#L46) gets if from the device directly, the [HueBridgeNupnpDiscovery](https://github.com/eclipse/smarthome/blob/7a32b35d08035ce81aa5831796740ff02c035c56/extensions/binding/org.eclipse.smarthome.binding.hue/src/main/java/org/eclipse/smarthome/binding/hue/internal/discovery/HueBridgeNupnpDiscovery.java#L83) applies some other magic - why yet a fourth way of creatively calculating the serial number? Can't you also read it from the handler configuration?

extensions/binding/org.eclipse.smarthome.binding.hue/src/main/java/org/eclipse/smarthome/binding/hue/handler/HueBridgeHandler.java
```diff
@@ -519,4 +519,13 @@ private boolean isEqual(State state1, State state2) {
...
+        } catch (Exception e) {
```
sjka: catching Exception seems a little too broad to me here. What do you attempt to protect yourself from here?

extensions/binding/org.eclipse.smarthome.binding.hue/src/main/java/org/eclipse/smarthome/binding/hue/handler/HueLightHandler.java
```diff
@@ -482,4 +483,13 @@ private int getAlertDuration(Command command) {
 
         return delay;
     }
+
+    @Override
+    public String getDeviceId() {
+        try {
+            return getLight().getUniqueID();
+        } catch (Exception e) {
```
sjka: see above

bundles/config/org.eclipse.smarthome.config.discovery/OSGI-INF/InboxAutoIgnore.xml
```diff
@@ -0,0 +1,13 @@
+<?xml version="1.0" encoding="UTF-8"?>
```
sjka: Any reasons why you didn't use DS annotations?

bundles/config/org.eclipse.smarthome.config.discovery/src/main/java/org/eclipse/smarthome/config/discovery/DiscoveryResult.java
```diff
@@ -80,12 +80,20 @@
...
      */
     public String getRepresentationProperty();
 
     /**
+     * Returns the actual value of the representation property of the result object.
+     *
+     * @see #getRepresentationProperty()
+     * @return the value of the representation property of this result object (could be null)
+     */
+    public Object getRepresentationPropertyValue();
```
sjka: Having this method in the API somehow looks weird to me.

bundles/config/org.eclipse.smarthome.config.discovery/src/main/java/org/eclipse/smarthome/config/discovery/internal/PersistentInbox.java
```diff
@@ -211,6 +212,28 @@ public synchronized boolean add(DiscoveryResult result) throws IllegalStateExcep
...
+        Thing thing = this.thingRegistry.get(result.getThingUID());
+        if (thing != null) {
+            // thing with same UID is already in registry
+            return thing;
+        }
+
+        // check, if there is a thing with the same representation property
+        Object value = result.getRepresentationPropertyValue();
+        if (value != null) {
+            Collection<Thing> things = this.thingRegistry.getAll();
```
sjka: This could be expressed much nicer with `thingRegistry.stream().map(...).filter(...)`!

bundles/config/org.eclipse.smarthome.config.discovery/src/main/java/org/eclipse/smarthome/config/discovery/internal/InboxAutoIgnore.java
```diff
@@ -0,0 +1,133 @@
...
+ * are removed from the {@link Inbox} so they could be discovered again afterwards.
+ * </p>
+ * <p>
+ * This service can be enabled or disabled by setting the {@code autoIgnore} property to either
+ * {@code true} or {@code false} via ConfigAdmin.
+ * </p>
+ *
+ * @author Andre Fuechsel - Initial Contribution
+ */
+public class InboxAutoIgnore extends AbstractTypedEventSubscriber<ThingStatusInfoChangedEvent> {
```
afuechsel: Argh, you got me. Forgot to write tests for this class. Will do so.

bundles/config/org.eclipse.smarthome.config.discovery/src/main/java/org/eclipse/smarthome/config/discovery/internal/InboxAutoIgnore.java
```diff
@@ -0,0 +1,133 @@
...
+        if (thing != null) {
+            String deviceUid = thing.getHandler().getDeviceId();
+            if (deviceUid != null) {
+                removeFromInbox(deviceUid);
+            }
+        }
+    }
+
+    private void removeFromInbox(String deviceUid) {
+        List<DiscoveryResult> results = inbox.get(new InboxFilterCriteria(DiscoveryResultFlag.IGNORED));
```
afuechsel: Sounds like a good idea. Will think about it.

bundles/config/org.eclipse.smarthome.config.discovery/src/main/java/org/eclipse/smarthome/config/discovery/internal/InboxAutoIgnore.java
```diff
@@ -0,0 +1,133 @@
...
+        if (autoIgnore) {
+            if (ThingStatus.ONLINE.equals(event.getStatusInfo().getStatus())) {
+                checkIfThingShouldBeIgnored(event);
+            } else if (ThingStatus.REMOVING.equals(event.getStatusInfo().getStatus())) {
+                removePossibleIgnoredResultInInbox(event);
+            }
+        }
+    }
+
+    private void checkIfThingShouldBeIgnored(ThingStatusInfoChangedEvent event) {
```
afuechsel: checkAndIgnoreThingIfNecessary doesn't sound good too.. maybe: checkAndIgnoreInInbox...

bundles/core/org.eclipse.smarthome.core.thing/src/main/java/org/eclipse/smarthome/core/thing/binding/ThingHandler.java
```diff
@@ -174,4 +175,18 @@
      */
     void handleRemoval();
 
+    /**
```
afuechsel: Adding this hint in the JavaDoc would require an additional import also in the MANIFEST (at least, Eclipse wanted to have that. I don't think, that's worth mentioning it. Maybe if I use the full qualified name ... will try

bundles/core/org.eclipse.smarthome.core.thing/src/main/java/org/eclipse/smarthome/core/thing/internal/ThingManager.java
```diff
@@ -544,7 +544,7 @@ public Void call() throws Exception {
                 registerAndInitializeHandler(thing, getThingHandlerFactory(thing));
             }
 
-            if (oldThing != thing) {
+            if (oldThing != thing && oldThing != null) {
```
afuechsel: Yes, but I stumpled upon it....

extensions/binding/org.eclipse.smarthome.binding.hue.test/src/test/groovy/org/eclipse/smarthome/binding/hue/test/HueBridgeHandlerOSGiTest.groovy
```diff
@@ -53,6 +56,30 @@ class HueBridgeHandlerOSGiTest extends AbstractHueOSGiTest {
     }
 
     @Test
+    void 'assert that deviceId is set correctly'() {
```
afuechsel: As it is just an addition to already existing tests....

bundles/config/org.eclipse.smarthome.config.discovery/src/main/java/org/eclipse/smarthome/config/discovery/internal/console/InboxConsoleCommandExtension.java
```diff
@@ -80,7 +80,7 @@ public void execute(String[] args, Console console) {
                     printInboxEntries(console, inbox.get(new InboxFilterCriteria(DiscoveryResultFlag.IGNORED)));
                     break;
                 case SUBCMD_CLEAR:
-                    clearInboxEntries(console, inbox.get(new InboxFilterCriteria(DiscoveryResultFlag.NEW)));
+                    clearInboxEntries(console, inbox.getAll());
```
afuechsel: Otherwise you would never be able to remove all inbox entries.

extensions/binding/org.eclipse.smarthome.binding.hue/src/main/java/org/eclipse/smarthome/binding/hue/handler/HueBridgeHandler.java
afuechsel: to avoid null checks - but you are right, will change it and just check for NullPointerExceptions too

bundles/config/org.eclipse.smarthome.config.discovery/OSGI-INF/InboxAutoIgnore.xml
afuechsel: I was not sure if I would then need to convert all other component descriptors as well. Would I?

bundles/config/org.eclipse.smarthome.config.discovery/src/main/java/org/eclipse/smarthome/config/discovery/DiscoveryResult.java
afuechsel: Why? It's a shortcut to result.getProperties().get((result.getRepresentationProperty()) and the usual null checks. I find it useful.

bundles/core/org.eclipse.smarthome.core.thing/src/main/java/org/eclipse/smarthome/core/thing/binding/ThingHandler.java
sjka: you don't need to really include a javadoc link. Just creating a connection between these two properties by prosa text should suffice.

extensions/binding/org.eclipse.smarthome.binding.hue/src/main/java/org/eclipse/smarthome/binding/hue/handler/HueBridgeHandler.java
sjka: I feared you are just covering a `catch (NPE)` here. Veto too!
Use the value from the handler configuration and your problem will vanish.

bundles/config/org.eclipse.smarthome.config.discovery/OSGI-INF/InboxAutoIgnore.xml
sjka: No, you don't - if you add a .gitinore for this specific generated XML into the OSGI-INF folder. Which is okay for that purpose until all are migrated. 

Migrating the others too just gives extra-kudos, but is not mandatory.

bundles/config/org.eclipse.smarthome.config.discovery/src/main/java/org/eclipse/smarthome/config/discovery/DiscoveryResult.java
sjka: Because
1. the name is awkward
2. we now have extra-methods for both, the key and the value
3. we provide access to the map already
4. we will never be able to get rid of this method again from the API

afuechsel: Hmm, if you insist I could remove this method - but to be honest: At the first sight I expected the method getRepresentationProperty() to return the value until I realized, that it only returns the property's name. I think, I will never ever need the representation property's name but its value, so I would rather think about getting rid of the other method. Or am I wrong?

sjka: I agree, there will be `discoveryResult.getProperties().get(discoveryResult.getRepresentationProperty()))` everywhere. The property name by itself is boring. Then let's rather change it and give it a crisp name. It will be API breaking, but I doubt anybody is using it 🤞

afuechsel: Then I would remove both methods and add only one: getRepresentationValue(). Would you agree?

afuechsel: On a second look I must disagree. Removing the existing method getRepresentationProperty would break not only the JSON of the inbox events (at least change it), but also make the method DiscoveryResultImpl#synchronize impossible as I would have no way to access the representationProperty name to copy it into the new class. 

What about leaving everything as it is and only renaming my method getRepresentationPropertyValue() into getRepresentationValue()?

bundles/config/org.eclipse.smarthome.config.discovery.test/src/test/groovy/org/eclipse/smarthome/config/setup/test/inbox/DynamicThingUpdateOSGITest.groovy
```diff
@@ -179,4 +184,40 @@ class DynamicThingUpdateOSGITest extends OSGiTest {
 
         unregisterService(thingHandlerFactory)
     }
+
+    @Test
```
afuechsel: But the persistentInbox does not mock the thing handler nor the thing handler factory. I just wanted to reuse the existing test infrastructure. Why not just renaming the test class?

extensions/binding/org.eclipse.smarthome.binding.hue/src/main/java/org/eclipse/smarthome/binding/hue/handler/HueBridgeHandler.java
afuechsel: I cannot use 
```` 
getThing().getConfiguration().get(SERIAL_NUMBER)
```` 
because when manual configuring, the serial number does not reflect the MAC address of the bridge, this could be any value. I have to get it from the device directly. But I cannot do it like the HueBridgeDiscoveryParticipant, as it relies on the UPnP result what we do not have here. Therefore I need to use my fourth way.

File: extensions/binding/org.eclipse.smarthome.binding.hue/src/main/java/org/eclipse/smarthome/binding/hue/handler/HueBridgeHandler.java (hunk 2 above)
Comment (afuechsel) [inline extensions/binding/org.eclipse.smarthome.binding.hue/src/main/java/org/eclipse/smarthome/binding/hue/handler/HueBridgeHandler.java]: see above, I cannot use the value from the handler config. I agree, that it is not the best idea to catch a possible NPW, but in this particular case I would otherwise need to guard against any possible NPE and doing the same like in the exception, so why bother? Just catch die NPE looks cleaner to me.

File: bundles/config/org.eclipse.smarthome.config.discovery/src/main/java/org/eclipse/smarthome/config/discovery/inbox/InboxFilterCriteria.java [hunk 15]
```diff
@@ -26,61 +27,137 @@
     private final String bindingId;
     private final ThingTypeUID thingTypeUID;
     private final ThingUID thingUID;
+    private final String representationValue;
     private final DiscoveryResultFlag flag;
 
+    private InboxFilterCriteria(String bindingId, ThingTypeUID thingTypeUID, ThingUID thingUID,
+            String representationValue, DiscoveryResultFlag flag) {
+        super();
```
maggu2810: Let's remove this explicit call that is done already implicit.

bundles/config/org.eclipse.smarthome.config.discovery/src/main/java/org/eclipse/smarthome/config/discovery/inbox/InboxFilterCriteria.java
```diff
@@ -26,61 +27,137 @@
     private final String bindingId;
     private final ThingTypeUID thingTypeUID;
     private final ThingUID thingUID;
+    private final String representationValue;
     private final DiscoveryResultFlag flag;
 
+    private InboxFilterCriteria(String bindingId, ThingTypeUID thingTypeUID, ThingUID thingUID,
```
maggu2810: It seems there are several constructors in that class. Couldn't we change all the constructors that them call one implementation (e.g. `this(bindingId, thingTypeUID, thingUID, null, null)`) instead of initialize members in every constructor implementation itself?

bundles/config/org.eclipse.smarthome.config.discovery/src/main/java/org/eclipse/smarthome/config/discovery/internal/InboxAutoIgnore.java
```diff
@@ -0,0 +1,142 @@
...
+        List<DiscoveryResult> results = inbox
+                .get(InboxFilterCriteria.representationFilter(deviceUid, DiscoveryResultFlag.IGNORED));
+        if (results.size() == 1) {
+            logger.debug("Removing the ignored result from the inbox for the device uid {}", deviceUid);
+            inbox.remove(results.get(0).getThingUID());
+        }
+    }
+
+    @Activate
+    protected void activate(Map<String, String> properties) {
```
maggu2810: Shouldn't it be `String, Object`?

sjka:
As my comments have been addresses, I'll leave it to @maggu2810 / @kaikreuzer to have a second look and merge it.

bundles/config/org.eclipse.smarthome.config.discovery/src/main/java/org/eclipse/smarthome/config/discovery/internal/InboxAutoIgnore.java
afuechsel: Yes, you are right.

bundles/config/org.eclipse.smarthome.config.discovery/src/main/java/org/eclipse/smarthome/config/discovery/inbox/InboxFilterCriteria.java
afuechsel: I could, but as @sjka suggested to make them all deprecated and to be removed before creating release 1.0 I didn't touch them. they are there for ages. If you insist, I will change them.

bundles/config/org.eclipse.smarthome.config.discovery/src/main/java/org/eclipse/smarthome/config/discovery/inbox/InboxFilterCriteria.java
afuechsel: Eclipse automatically created it and I didn't realize.

bundles/config/org.eclipse.smarthome.config.discovery/src/main/java/org/eclipse/smarthome/config/discovery/DiscoveryResult.java
```diff
@@ -80,12 +80,20 @@
...
      */
     public String getRepresentationProperty();
 
     /**
+     * Returns the actual value of the representation property of the result object.
+     *
+     * @see #getRepresentationProperty()
+     * @return the value of the representation property of this result object (could be null)
+     */
+    public Object getRepresentationValue();
```
kaikreuzer: I am not too fond in having such "convenience" methods as part of the API.
To me, the API is complete as it offers `getRepresentationProperty()` and `getProperties()` - a client can nicely get hold of all information that he needs.
Having such an additional method on the interface rather brings up the risk that this new method returns something different than `getProperties().get(getRepresentationProperty())`.
I'd therefore suggest to move this convenience method to the class where you actually need it (PersistentInbox).

bundles/config/org.eclipse.smarthome.config.discovery/src/main/java/org/eclipse/smarthome/config/discovery/inbox/InboxFilterCriteria.java
```diff
@@ -26,61 +27,137 @@
     private final String bindingId;
     private final ThingTypeUID thingTypeUID;
     private final ThingUID thingUID;
+    private final String representationValue;
```
kaikreuzer: I find this a bit limiting. What if a consumer wants to filter for some other property (e.g. MANUFACTURER) or so?

@ViToni recently introduced a nice new stream based filtering for registries, which has been adopted in the RuleRegistry by defining [RulePredicates](https://github.com/eclipse/smarthome/blob/master/bundles/automation/org.eclipse.smarthome.automation.api/src/main/java/org/eclipse/smarthome/automation/RulePredicates.java). This mechanism was meant to be used from now on throughout ESH, so imho it would make a lot of sense to go for it here as well, instead of introducing all those `xxxFilter()` methods, wdyt?
This would allow us to simply have a "property predicate", where the consumer can choose whether he wants to use the representation property or any other. There would thus be no need for a dedicated field for just this purpose here.

bundles/config/org.eclipse.smarthome.config.discovery/src/main/java/org/eclipse/smarthome/config/discovery/internal/InboxAutoIgnore.java
```diff
@@ -0,0 +1,142 @@
...
+ * </p>
+ * <p>
+ * This service can be enabled or disabled by setting the {@code autoIgnore} property to either
+ * {@code true} or {@code false} via ConfigAdmin.
+ * </p>
+ *
+ * @author Andre Fuechsel - Initial Contribution
+ */
+@Component(immediate = true, service = EventSubscriber.class, property = {
+        "service.config.description.uri=system:inbox", "service.config.label=Inbox Auto Ignore",
```
kaikreuzer: Service meta-data for the UI should not be on a single feature level - imho it would be better to simply have a "Inbox" entry (instead of "Inbox Auto Ignore"), which could offer multiple things that might be possible to do in the context of the inbox.
There is e.g. already an ["Auto Approve" service](https://github.com/openhab/openhab-core/blob/master/bundles/org.openhab.core/src/main/java/org/openhab/core/internal/inbox/AutoApproveService.java) in openHAB - it might make sense to move that (very small) feature into your service here.

bundles/config/org.eclipse.smarthome.config.discovery/src/main/java/org/eclipse/smarthome/config/discovery/internal/InboxAutoIgnore.java
```diff
@@ -0,0 +1,142 @@
...
+            autoIgnore = value == null || !value.toString().equals("false");
+        }
+    }
+
+    @Reference
+    protected void setThingRegistry(ThingRegistry thingRegistry) {
+        this.thingRegistry = thingRegistry;
+    }
+
+    protected void unsetThingRegistry(ThingRegistry thingRegistry) {
```
kaikreuzer: Sorry, I haven't yet used the annotations myself: Is the "unset" automatically derived from the name of the `@Reference`-annotated method?

bundles/core/org.eclipse.smarthome.core.thing/src/main/java/org/eclipse/smarthome/core/thing/binding/ThingHandler.java
```diff
@@ -174,4 +175,24 @@
...
+     * This method should return a unique identifier of the device handled by this {@link ThingHandler}.
```
kaikreuzer: So far we have avoided the word "device" everywhere - it isn't part of the ESH vocabulary. From the JavaDoc here, the right word in the ESH context would imho be "thing".

I think you used "device" here, because you felt that a new method "getThingId()" would only cause irritation - I agree.
But wouldn't the best solution to this be to go with the vocabulary, which we already introduced for the DiscoveryResult? I.e. could we simply add a new method `getRepresentationProperty()`, which gives the handler the chance to point to the property to use for unique identification?

bundles/config/org.eclipse.smarthome.config.discovery/src/main/java/org/eclipse/smarthome/config/discovery/DiscoveryResult.java
afuechsel: Ok, I agree.

bundles/config/org.eclipse.smarthome.config.discovery/src/main/java/org/eclipse/smarthome/config/discovery/inbox/InboxFilterCriteria.java
afuechsel: The problem with InboxFilterCritearia is, that they are rather old. As I didn't want to break backward compatibility, I did not remove the constructors, but made them deprecated. Not sure, if changing thic class into something more Java 8 like is not breaking the compatibility. But I agree, it's a bit ugly. Will think about it.

bundles/config/org.eclipse.smarthome.config.discovery/src/main/java/org/eclipse/smarthome/config/discovery/internal/InboxAutoIgnore.java
afuechsel: Would it be ok, to set sevice.config.label to "Inbox" then?

bundles/config/org.eclipse.smarthome.config.discovery/src/main/java/org/eclipse/smarthome/config/discovery/internal/InboxAutoIgnore.java
afuechsel: Yes, it is. No need to annotate this method as well, but looks strange, I agree.

bundles/core/org.eclipse.smarthome.core.thing/src/main/java/org/eclipse/smarthome/core/thing/binding/ThingHandler.java
afuechsel: Not sure, if I understood you right, introducing getRepresentationProperty() would not help as the method to retrieve the unique device id is different in each handler and not necessarily only getting some device property, because in the manual case the properties might also been set manually. I agree we should not use device here, so what about `getUniqueIdentifier()`or `getRepresentationValue()`?

bundles/config/org.eclipse.smarthome.config.discovery/src/main/java/org/eclipse/smarthome/config/discovery/inbox/InboxFilterCriteria.java
kaikreuzer: Well, [below](https://github.com/eclipse/smarthome/pull/3899/files#diff-165ec77b918d80c4f727ab179ac5f93eR50) you add MANY new public methods that become API - which will make it difficult to remove them again. I am ok with keeping backward compatibility by not touching the InboxFilterCriteria here at all. But instead of adding new features to it (filtering for properties), my suggestion is to introduce directly the new stream mechanism for these purposes (especially as I do not think that it will mean much more effort).

bundles/config/org.eclipse.smarthome.config.discovery/src/main/java/org/eclipse/smarthome/config/discovery/internal/InboxAutoIgnore.java
kaikreuzer: Yes, but you should also rename the class to something more general.
Btw, did I overlook the referenced config description for this service or did you forget to commit it?

bundles/core/org.eclipse.smarthome.core.thing/src/main/java/org/eclipse/smarthome/core/thing/binding/ThingHandler.java
kaikreuzer: > introducing getRepresentationProperty() would not help 

Why not? `handler.getThing().getProperties().get(hander.getRepresentationProperty())` would always return an individual result, wouldn't it?
But looking at that statement, I wonder if the RepresentationProperty shouldn't rather be a feature of the Thing and not of the Handler. Wdyt?

bundles/config/org.eclipse.smarthome.config.discovery/src/main/java/org/eclipse/smarthome/config/discovery/internal/InboxAutoIgnore.java
afuechsel: Oops - forget to add the config description. 

Renaming the class into something more general: The purpose of this service is atuomatically ignoring inbox entries - cannot find a better and more general name,. e.g. ÌnboxService`? But this is more confusing....

bundles/config/org.eclipse.smarthome.config.discovery/src/main/java/org/eclipse/smarthome/config/discovery/inbox/InboxFilterCriteria.java
afuechsel: Leaving InboxFilterCriterla completely unchanged would mean to have this odd constructors also in the 1.0 release eventually. And how to introduce the predicates then? May be I didn't get the idea of predicates yet...

bundles/config/org.eclipse.smarthome.config.discovery/src/main/java/org/eclipse/smarthome/config/discovery/internal/InboxAutoIgnore.java
kaikreuzer: Well, as mentioned above, it should not only auto-ignore entries, it should also (in future) support auto-approve entries, thus we would need a more unspecific name. I don't have a good suggestion either. Maybe something like "AutomaticInboxProcessor"?

bundles/core/org.eclipse.smarthome.core.thing/src/main/java/org/eclipse/smarthome/core/thing/binding/ThingHandler.java
afuechsel: Did you check the example implementation of `getDeviceId()`in the HueLightHandler or HueBridgeHandler for example?

For the light: 
```` 
return light.getUniqueID();
```` 
or for the bridge: 
```` 
return bridge.getFullConfig().getConfig().getMACAddress().replaceAll(":", "").toLowerCase();
````  
The value is not taken out of a property of the thing, because these properties could be set manually during creation of that thing. I was also thinking to move this method from the handler to the thing itself, but I refrained. It almost alway requires physical interaction with the device to get some unique identifier.

File: bundles/config/org.eclipse.smarthome.config.discovery/src/main/java/org/eclipse/smarthome/config/discovery/internal/InboxAutoIgnore.java (hunk 20 above)
Comment (afuechsel) [inline bundles/config/org.eclipse.smarthome.config.discovery/src/main/java/org/eclipse/smarthome/config/discovery/internal/InboxAutoIgnore.java]: This sounds like a good proposal.

File: bundles/config/org.eclipse.smarthome.config.discovery/src/main/java/org/eclipse/smarthome/config/discovery/inbox/InboxFilterCriteria.java (hunk 19 above)
Comment (kaikreuzer) [inline bundles/config/org.eclipse.smarthome.config.discovery/src/main/java/org/eclipse/smarthome/config/discovery/inbox/InboxFilterCriteria.java]: If we leave it unchanged, we can deprecate the whole InboxFilterCriterla class right now, so nothing would be left in an 1.0 release.

>  And how to introduce the predicates then?

Similar to the Registry: The PersistentInbox should get a new `stream()` method besides the current `get` and `getAll`. But you can also ask @ViToni for help as he designed this all :-)

File: bundles/core/org.eclipse.smarthome.core.thing/src/main/java/org/eclipse/smarthome/core/thing/binding/ThingHandler.java (hunk 22 above)
Comment (kaikreuzer) [inline bundles/core/org.eclipse.smarthome.core.thing/src/main/java/org/eclipse/smarthome/core/thing/binding/ThingHandler.java]: > Did you check the example implementation of getDeviceId()in the HueLightHandler or HueBridgeHandler for example?

Of course.

> For the light

... you use the very same content is already present as a [representation property](https://github.com/eclipse/smarthome/pull/3899/files#diff-8fa06d6d05ccc8c9a1724a0b84c61e5fR119).

> for the bridge

... you use the MAC address, which we specifically mention as a candidate for a representation property.

> because these properties could be set manually during creation of that thing

Afaik, it is not possible to set properties manually, only config parameter values.
But in any case: What you might not know is that ThingHandlers can update the properties at any time (irrespectively whether the Thing is discovered, defined manually (in database or DSL) and this is what the handlers are doing already all the time. So once your handler establishes the connection to the device, it should update the Thing properties. The nice side effect is that in case your thing is stored in the managed provider the updated properties are even persisted. So you will have it at hand after a restart even if your Thing hasn't come online yet.

Comment (afuechsel) [inline bundles/core/org.eclipse.smarthome.core.thing/src/main/java/org/eclipse/smarthome/core/thing/binding/ThingHandler.java]: Ok, understood. Would you then opt for implementing the getRepresentationProperty in the thing itself, instead of the handler? And like in DiscoveryResult I assume you do not want me to add a getRepresentaionPropertyValue() convenience method here?

Comment (kaikreuzer) [inline bundles/core/org.eclipse.smarthome.core.thing/src/main/java/org/eclipse/smarthome/core/thing/binding/ThingHandler.java]: > Would you then opt for implementing the getRepresentationProperty in the thing itself

Yes. But you are right, we need a mechanism to fill this field of the Thing. If discovered, we could actually set it through the DiscoveryResult. In general, I would say that the setting should be taken from the thing type as I would claim that the property (key) must always be the same for things of the same type. - so ideally, we would allow to specify it in the type. Does that make sense?

Comment (afuechsel) [inline bundles/core/org.eclipse.smarthome.core.thing/src/main/java/org/eclipse/smarthome/core/thing/binding/ThingHandler.java]: Makes sense - but we would need to extend the thing type xml then.

Comment (kaikreuzer) [inline bundles/core/org.eclipse.smarthome.core.thing/src/main/java/org/eclipse/smarthome/core/thing/binding/ThingHandler.java]: Yes, but ideally, we would then not need any change in the handlers at all.

Comment (afuechsel) [inline bundles/core/org.eclipse.smarthome.core.thing/src/main/java/org/eclipse/smarthome/core/thing/binding/ThingHandler.java]: Could you please make a suggestion, how to extend the XML?

Comment (kaikreuzer) [inline bundles/core/org.eclipse.smarthome.core.thing/src/main/java/org/eclipse/smarthome/core/thing/binding/ThingHandler.java]: It would have been nice to be able to do something like
```
        <properties>
        	<property name="vendor">Philips</property>
        	<property name="serialNumber" representation="true"></property>
        </properties>
```
but as we have a `Map<String,String>` in the Java API, there is no way to add the info to the properties themselves.
So yes, we would need to do it as you suggested (maybe more clearly call it `representationProperty`).
The two problems that you list are imho no "real" problems - it is a convention that must be followed and we actually have the same issue for channel ids and other information that needs to be kept in sync between XML and Java.

File: bundles/config/org.eclipse.smarthome.config.discovery/src/main/java/org/eclipse/smarthome/config/discovery/internal/PersistentInbox.java [hunk 23]
```diff
@@ -211,6 +206,40 @@ public synchronized boolean add(DiscoveryResult result) throws IllegalStateExcep
         return false;
     }
 
+    /**
+     * Returns an already existing thing in the registry, that represents
+     * the {@link DiscoveryResult}. This thing either has the same UID as the
+     * {@link DiscoveryResult} or the device id of the thing's handler is equal
```
kaikreuzer: not "device id"

bundles/config/org.eclipse.smarthome.config.discovery/src/main/java/org/eclipse/smarthome/config/discovery/internal/PersistentInbox.java
```diff
@@ -171,7 +166,7 @@ public Thing approve(ThingUID thingUID, String label) {
     public synchronized boolean add(DiscoveryResult result) throws IllegalStateException {
         if (result != null) {
             ThingUID thingUID = result.getThingUID();
-            Thing thing = this.thingRegistry.get(thingUID);
+            Thing thing = getExistingThingForDiscoveryResult(result);
```
kaikreuzer: Hm. This change looks against the discussed approach in a few ways:
- It changing the existing behavior by not at all adding results, although no Thing with the same UID exists. I thought that the idea was to have the "auto-ignore" as a configurable feature that does not interfere with the existing logic.
- If it is considered to be "auto-ignore", then it is wrong here, because it does not at all add the result. This make the whole feature unsymmetric: A result is present, but ignored, if the Discovery Result comes first and then the Thing is added. It is NOT present, if the Thing is added manually and then it is discovered.

Wouldn't it be much cleaner if the AutomaticInboxProcessor would also listen to Inbox events and just do the ignore functionality in either case?

bundles/core/org.eclipse.smarthome.core.thing.xml.test/src/test/resources/test-bundle-pool/ThingTypesTest.bundle/ESH-INF/thing/thing-types.xml
```diff
@@ -13,6 +13,7 @@
 		<properties>
 			<property name="vendor">Philips</property>
 		</properties>
+        <representation-property>serialNumber</representation-property>
```
kaikreuzer: replace spaces by tabs

bundles/core/org.eclipse.smarthome.core.thing.xml.test/src/test/resources/test-bundle-pool/ThingTypesTest.bundle/ESH-INF/thing/thing-types.xml
```diff
@@ -55,6 +56,7 @@
 			<property name="key1">value1</property>
 			<property name="key2">value2</property>
 		</properties>
+        <representation-property>uniqueId</representation-property>
```
kaikreuzer: replace spaces by tabs

bundles/core/org.eclipse.smarthome.core.thing/src/main/java/org/eclipse/smarthome/core/thing/binding/ThingHandler.java
```diff
@@ -31,6 +31,7 @@
  * @author Michael Grammling - Added dynamic configuration update
  * @author Thomas Höfer - Added config description validation exception to handleConfigurationUpdate operation
  * @author Stefan Bußweiler - API changes due to bridge/thing life cycle refactoring
+ * @author Andre Fuechsel - added API to get a device id
```
kaikreuzer: No, you didn't.

docs/documentation/concepts/discovery.md
```diff
@@ -15,3 +15,5 @@ In Eclipse SmartHome bindings therefore implement _Discovery Services_ for thing
...
 
 Discovery results can either be ignored or approved, where in the latter case a thing is created for them and they become available in the application. If an entry is ignored, it will be hidden in the inbox without creating a thing for it. 
+
+Eclipse SmartHome offers a service to automatically ignore discovery results in the inbox, whenever a thing is created manually, that represents the same thing, as the respective discovery result would create. This thing would either have the same thing UID or the value of its representation property is equal to the representation property's value in the discovery result. This service is enabled by default but could be disabled by setting `autoIgnore` to false.
```
kaikreuzer: Please add a new line after every sentence - makes diffing much easier!

extensions/binding/org.eclipse.smarthome.binding.hue.test/src/test/groovy/org/eclipse/smarthome/binding/hue/test/HueBridgeHandlerOSGiTest.groovy
```diff
@@ -35,6 +35,7 @@ import org.junit.Test
  * @author Oliver Libutzki - Initial contribution
  * @author Michael Grammling - Initial contribution
  * @author Denis Dudnik - switched to internally integrated source of Jue library
+ * @author Andre Fuechsel - added test for device id
```
kaikreuzer: No, you didn't.

extensions/binding/org.eclipse.smarthome.binding.hue.test/src/test/groovy/org/eclipse/smarthome/binding/hue/test/HueLightHandlerOSGiTest.groovy
```diff
@@ -48,7 +48,7 @@ import org.junit.Test
  * @author Oliver Libutzki - Initial contribution
  * @author Michael Grammling - Initial contribution
  * @author Markus Mazurczak - Added test for OSRAM Par16 50 TW bulbs
- * @author Andre Fuechsel - modified tests after introducing the generic thing types
+ * @author Andre Fuechsel - modified tests after introducing the generic thing types, added test for device id
```
kaikreuzer: No, you didn't.

extensions/binding/org.eclipse.smarthome.binding.hue.test/src/test/groovy/org/eclipse/smarthome/binding/hue/test/HueLightHandlerOSGiTest.groovy
```diff
@@ -64,6 +64,9 @@ class HueLightHandlerOSGiTest extends AbstractHueOSGiTest {
     final String OSRAM_MODEL_TYPE = "PAR16 50 TW"
     final String OSRAM_MODEL_TYPE_ID = "PAR16_50_TW"
 
+    def SW_VERSION = "5.23.1.13452"
```
kaikreuzer: Still needed?

extensions/binding/org.eclipse.smarthome.binding.hue/src/main/java/org/eclipse/smarthome/binding/hue/handler/HueBridgeHandler.java
```diff
@@ -56,7 +56,7 @@
  * @author Dennis Nobel - Initial contribution of hue binding
  * @author Oliver Libutzki
  * @author Kai Kreuzer - improved state handling
- * @author Andre Fuechsel - implemented getFullLights(), startSearch()
+ * @author Andre Fuechsel - implemented getFullLights(), startSearch(), getDeviceId()
```
kaikreuzer: No.

extensions/binding/org.eclipse.smarthome.binding.hue/src/main/java/org/eclipse/smarthome/binding/hue/handler/HueBridgeHandler.java
```diff
@@ -273,18 +265,32 @@ public void onConnectionLost(HueBridge bridge) {
      * This method is called whenever the connection to the given {@link HueBridge} is resumed.
      *
      * @param bridge the hue bridge the connection is resumed to
+     * @throws ApiException
```
kaikreuzer: In which situation?

extensions/binding/org.eclipse.smarthome.binding.hue/src/main/java/org/eclipse/smarthome/binding/hue/handler/HueBridgeHandler.java
```diff
@@ -273,18 +265,32 @@ public void onConnectionLost(HueBridge bridge) {
...
+     * @throws IOException
      */
-    public void onConnectionResumed(HueBridge bridge) {
+    private void onConnectionResumed(HueBridge bridge) throws IOException, ApiException {
         logger.debug("Bridge connection resumed. Updating thing status to ONLINE.");
+
+        FullConfig fullConfig = bridge.getFullConfig();
```
kaikreuzer: Is it safe to do this here? It has some smell to me... Note that this is an expensive, potentially long-running and blocking call.

extensions/binding/org.eclipse.smarthome.binding.hue/src/main/java/org/eclipse/smarthome/binding/hue/handler/HueBridgeHandler.java
```diff
@@ -273,18 +265,32 @@ public void onConnectionLost(HueBridge bridge) {
...
+        }
+
         updateStatus(ThingStatus.ONLINE);
     }
 
     /**
      * Check USER_NAME config for null. Call onConnectionResumed() otherwise.
      *
      * @return True if USER_NAME was not null.
+     * @throws ApiException
```
kaikreuzer: In which situations?

extensions/binding/org.eclipse.smarthome.binding.hue/src/main/java/org/eclipse/smarthome/binding/hue/handler/HueLightHandler.java
```diff
@@ -47,7 +47,8 @@
  * @author Dennis Nobel - Initial contribution of hue binding
  * @author Oliver Libutzki
  * @author Kai Kreuzer - stabilized code
- * @author Andre Fuechsel - implemented switch off when brightness == 0, changed to support generic thing types
+ * @author Andre Fuechsel - implemented switch off when brightness == 0, changed to support generic thing types, added
```
kaikreuzer: No.

extensions/binding/org.eclipse.smarthome.binding.hue/src/main/java/org/eclipse/smarthome/binding/hue/handler/HueBridgeHandler.java
afuechsel: Yes, it was previously done in the Runnable on each connection resumed. What I could do is to guard against double initialization, something like `if (!propertiesInitialized) ...`.

bundles/config/org.eclipse.smarthome.config.discovery/src/main/java/org/eclipse/smarthome/config/discovery/internal/PersistentInbox.java
afuechsel: Yes, this sounds like a reasonable approach. Will change it. But thi feature was already there, you might have overlooked it in the first place.

bundles/config/org.eclipse.smarthome.config.discovery/src/main/java/org/eclipse/smarthome/config/discovery/internal/AutomaticInboxProcessor.java
```diff
@@ -0,0 +1,155 @@
...
+        if (results.size() == 1) {
+            logger.debug("Removing the ignored result from the inbox for the device uid {}", deviceUid);
+            inbox.remove(results.get(0).getThingUID());
+        }
+    }
+
+    @Activate
+    protected void activate(Map<String, Object> properties) {
+        if (properties != null) {
+            String value = (String) properties.get("autoIgnore");
```
afuechsel: I will add some instanceof checks.

extensions/binding/org.eclipse.smarthome.binding.hue/src/main/java/org/eclipse/smarthome/binding/hue/handler/HueBridgeHandler.java
kaikreuzer: Definitely makes sense!

bundles/config/org.eclipse.smarthome.config.discovery/src/main/java/org/eclipse/smarthome/config/discovery/internal/AutomaticInboxProcessor.java
kaikreuzer: Or simply do a `toString()`, which should solve the problem nicely :-)

kaikreuzer:
> Need to have a detailed look later, just minor comments.

@maggu2810 I'll wait for your further feedback, but I can do the merge upon your approval (as I do not have to respect the 1000 loc rule since @afuechsel is from the same company).

maggu2810:
I just had a very very very short look at that changes because of the lack of time.
If it is just an addition and not an API breakage (hopefully no other one, too :wink: ), feel free to merge.

kaikreuzer:
I did many many many deep looks and as you can see we had many iterations - I am quite confident that you will like all details about it now ;-)

lolodomo:
Please note a typo in documentation (xml-refrence.md) where tag is sometimes `representationProperty` and sometimes `representation-property`.

PR #3900, opened 2017-07-27
Upgraded to more recent patch level of Jetty and jax-rs-connector 

kaikreuzer:
(to be in sync with Karaf features as well as the openHAB IDE)


sjka:
Out of curiosity: is there any particular reason why we stay with jetty 9.2 and not upgrade to 9.4 when touching it anyway?

kaikreuzer:
Yes, there are two of them:
1. Our minimal requirement is Jetty 9.2 (e.g. as an API e.g. for the HTTP Client), so having a higher version in the target platform is not a good thing.
2. Jetty 9.3.something stopped providing all dependencies on a p2 site, so it isn't easily possible to use it in the target platform, unless we try to gather all that is needed by ourselves and make sure we have proper CQs for them.

sjka:
okok, fully convinced, thanks.
As the karaf tp feature already is on jax-rs 5.3.1 and does not include jetty anyway, this look okay to me.

PR #3904, opened 2017-07-27
[config] Added unit 'ms' (milliseconds) for config parameter

cweitkamp:
* Added unit 'ms' (milliseconds) for config parameter
* Changed 'kgv' to 'kg' assuming it should be kilogram
* Reordered 'unitType' in xsd file (SI base units, derived units with special name, others)

@tomhoefer I want to complete the `unit.properties` and `unit_de.properties` files. Is there any logic for this? Sometimes we use the written name of the unit, sometimes we use the symbol of the unit?


tomhoefer:
The original PR was #889 ... there you find the "logic"

kaikreuzer:
I think this should be discussed in the context of #601 - so far, the list of units only (with a single exception) comprises base units and not "shifted" ones. After all, the unit is "s" here and "m" only stands for 10^-3. Adding it brings up the question, why we shouldn't also add µ, n, p as well as K, M G, T prefixes.

Don't get me wrong, I am fully with you that "ms" is widely used in our parameters so it makes sense to add it - I just want to make sure it is aligned with the approach that @htreu and @clinique are working on in #601.

cweitkamp:
@tomhoefer Thanks.

@kaikreuzer Absolutely. I am with you. I am strongly looking forward for the new unit feature. I may revert the support of milliseconds from this PR and only provide the kilometers bugfix.

htreu:
Right now the approach in #1857 (which is the successor of #601) is to provide quantitative values for sensor data from bindings and a system wide conversion into user defined units. There is not yet a concept of extending this to config descriptions.
Since config descriptions are directly addressing the user I´m not sure how we would benefit from bringing the two concepts together. The unit in a configuration description gives a hint to the user which quantity the underlying service/binding/thing/channel expects.
Or to put it in another way: How would the framework benefit from knowing the real unit of a configuration parameter? Can we derive new features (with real value) from this knowledge?

PR #3906, opened 2017-07-28
Enable Null annotations warnings/errors in IDE and Maven

### LINKED ISSUE #3624: Introduce possibility to use @NonNull and @Nullable
Introduce possibility to use @NonNull and @Nullable annotations in IDE and Maven build.

triller-telekom:
Enables a warning on a potential Nullpointer access and an Error if a null value is passed to a @Nonnull argument

Fixes #3624


bundles/model/org.eclipse.smarthome.model.script/META-INF/MANIFEST.MF
```diff
@@ -19,6 +19,7 @@ Require-Bundle: org.eclipse.xtext;visibility:=reexport,
  org.eclipse.smarthome.model.lazygen;resolution:=optional
 Import-Package: com.google.common.base,
  com.google.common.collect,
+ javax.annotation,
```
kaikreuzer: You should mark it optional, so that it is only required at compile time, but not at runtime.

bundles/model/org.eclipse.smarthome.model.thing/META-INF/MANIFEST.MF
```diff
@@ -19,7 +19,8 @@ Require-Bundle: org.eclipse.xtext;visibility:=reexport,
  org.objectweb.asm;bundle-version="[5.0.1,6.0.0)";resolution:=optional,
  org.eclipse.smarthome.model.lazygen;resolution:=optional,
  org.eclipse.xtext.xbase.lib
-Import-Package: org.apache.log4j,
+Import-Package: javax.annotation,
```
kaikreuzer: You should mark it optional, so that it is only required at compile time, but not at runtime.

pom.xml
```diff
@@ -156,7 +156,16 @@
...
               </extraClasspathElement>
+                <extraClasspathElement>
+                  <groupId>com.google.code.findbugs</groupId>
+                  <artifactId>jsr305</artifactId>
+                  <version>3.0.2</version>
+              </extraClasspathElement>
             </extraClasspathElements>
+            <compilerArgs>
+              <arg>-err:+nullAnnot(javax.annotation.Nullable|javax.annotation.Nonnull|org.eclipse.jdt.annotation.NonNullByDefault),+inheritNullAnnot</arg>
+			  <arg>-warn:+null</arg>
```
kaikreuzer: Please replace tabs by spaces on this line

maggu2810:
I would like to give it a try, add some annotations on the API, remove the annotation stubs in my projects and check if it play well together...

One question about the optional import...
I have done it some time in the Eclipse IDE. Add a `Nullable` or `NonNull` annotation on a method used by other bundles (e.g. in the base module handler). The Eclipse IDE complains about missing stuff if the bindings does not import (optional or not) the annotation bundle itself. It is not a problem for the runtime, but for the Eclipse IDE only. Does it work with that solution?
I can test it myself if I test the other stuff as well, but perhaps you already done it...

triller-telekom:
At first I should mention that we are using `@Nonnull` with a **lower-case n** in the middle, otherwise you might wonder why it is not working ;)

I just did the following test but I am not sure if it is what you meant:

Inside `bundles/core/org.eclipse.smarthome.core.thing/src/main/java/org/eclipse/smarthome/core/thing/binding/BaseThingHandler.java` I annotated the `State` parameter of the `updateState(ChannelUid, State)` with `@Nonnull` and had to add the optional `javax.annotation` dependency to the bundle `org.eclipse.smarthome.core.thing`.

In a binding like the astro binding for example, I can call that method and get a warning if I pass `null` as the `State` parameter to that method. But I do NOT need to add a `javax.annotation` dependency to the astro binding bundle.

Is that what you wanted to know?

maggu2810:
@triller-telekom I would like to test this branch.
After I started a build on command line it fails with the following error:

```
[ERROR] Failed to execute goal org.eclipse.tycho:tycho-compiler-plugin:1.0.0:compile (default-compile) on project org.eclipse.smarthome.config.xml: Compilation failure: Compilation failure:
[ERROR] /home/maggu2810/bin/pkgs/eclipse/esh/git/smarthome/bundles/config/org.eclipse.smarthome.config.xml/src/main/java/org/eclipse/smarthome/config/xml/test/Test.java:[3]
[ERROR] import org.eclipse.jdt.annotation.NonNull;
[ERROR] ^^^^^^^^^^^^^^^
[ERROR] The import org.eclipse.jdt cannot be resolved
[ERROR] /home/maggu2810/bin/pkgs/eclipse/esh/git/smarthome/bundles/config/org.eclipse.smarthome.config.xml/src/main/java/org/eclipse/smarthome/config/xml/test/Test.java:[4]
[ERROR] import org.eclipse.jdt.annotation.Nullable;
[ERROR] ^^^^^^^^^^^^^^^
[ERROR] The import org.eclipse.jdt cannot be resolved
[ERROR] /home/maggu2810/bin/pkgs/eclipse/esh/git/smarthome/bundles/config/org.eclipse.smarthome.config.xml/src/main/java/org/eclipse/smarthome/config/xml/test/Test.java:[27]
[ERROR] public static @Nullable Object testReturnNull() {
[ERROR] ^^^^^^^^
[ERROR] Nullable cannot be resolved to a type
[ERROR] /home/maggu2810/bin/pkgs/eclipse/esh/git/smarthome/bundles/config/org.eclipse.smarthome.config.xml/src/main/java/org/eclipse/smarthome/config/xml/test/Test.java:[31]
[ERROR] public static void testArgumentDereference(@Nullable Object obj) {
[ERROR] ^^^^^^^^
[ERROR] Nullable cannot be resolved to a type
[ERROR] /home/maggu2810/bin/pkgs/eclipse/esh/git/smarthome/bundles/config/org.eclipse.smarthome.config.xml/src/main/java/org/eclipse/smarthome/config/xml/test/Test.java:[37]
[ERROR] public static void testCheckArgForNull(@Nullable Object obj) {
[ERROR] ^^^^^^^^
[ERROR] Nullable cannot be resolved to a type
[ERROR] /home/maggu2810/bin/pkgs/eclipse/esh/git/smarthome/bundles/config/org.eclipse.smarthome.config.xml/src/main/java/org/eclipse/smarthome/config/xml/test/Test.java:[45]
[ERROR] public static @NonNull Object testReturnNull() {
[ERROR] ^^^^^^^
[ERROR] NonNull cannot be resolved to a type
[ERROR] /home/maggu2810/bin/pkgs/eclipse/esh/git/smarthome/bundles/config/org.eclipse.smarthome.config.xml/src/main/java/org/eclipse/smarthome/config/xml/test/Test.java:[49]
[ERROR] public static void testArgumentDereference(@NonNull Object obj) {
[ERROR] ^^^^^^^
[ERROR] NonNull cannot be resolved to a type
[ERROR] /home/maggu2810/bin/pkgs/eclipse/esh/git/smarthome/bundles/config/org.eclipse.smarthome.config.xml/src/main/java/org/eclipse/smarthome/config/xml/test/Test.java:[55]
[ERROR] public static void testCheckArgForNull(@NonNull Object obj) {
[ERROR] ^^^^^^^
[ERROR] NonNull cannot be resolved to a type
[ERROR] 8 problems (8 errors)
[ERROR] -> [Help 1]
```

Is `org.eclipse.jdt.annotation.NonNull` used by intention?

maggu2810:
@triller-telekom Ignore my previous message ... dirty working tree ...

maggu2810:
@triller-telekom Thanks

sjka:
How about adding a word to the documentation here:
https://github.com/eclipse/smarthome/blob/master/docs/documentation/development/guidelines.md?

triller-telekom:
I guess first we have to agree on what to use where in #4080 before we write a documentation.

maggu2810:
> How about adding a word to the documentation here:
https://github.com/eclipse/smarthome/blob/master/docs/documentation/development/guidelines.md?

I started with a summary in the wiki (as @triller-telekom also referenced in his linked  comment).

PR #3915, opened 2017-07-31
All rules are returned if the tags parameter is null or empty list #3903

### LINKED ISSUE #3903: Unable to get all rules using over Http Rest Request
I am using the `RuleResource.get(@QueryParam("prefix") final String prefix, @QueryParam("tags") final List<String> tags)` metod.
In my request `prefix` and `tags` parameters are not specified so I am expecting all rules to be returned.
However after some debugging i fond that the `tags` parameter is not null, despite the fact that it is not specified in the request, so in the current implementation only the rules that are with no tags are returned.

plamen-peev:
Fix #3903
Signed-off-by: Plamen Veselinov Peev <p.peev@prosyst.bg>

### DIFF
--- a/bundles/automation/org.eclipse.smarthome.automation.rest/src/main/java/org/eclipse/smarthome/automation/rest/internal/RuleResource.java
+++ b/bundles/automation/org.eclipse.smarthome.automation.rest/src/main/java/org/eclipse/smarthome/automation/rest/internal/RuleResource.java
@@ -1,10 +1,10 @@
 /**
- * Copyright (c) 1997, 2015 by ProSyst Software GmbH and others.
- * All rights reserved. This program and the accompanying materials
- * are made available under the terms of the Eclipse Public License v1.0
- * which accompanies this distribution, and is available at
- * http://www.eclipse.org/legal/epl-v10.html
- */
+* Copyright (c) 2015, 2017 by Bosch Software Innovations and others.
+* All rights reserved. This program and the accompanying materials
+* are made available under the terms of the Eclipse Public License v1.0
+* which accompanies this distribution, and is available at
+* http://www.eclipse.org/legal/epl-v10.html
+*/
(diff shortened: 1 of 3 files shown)


bundles/automation/org.eclipse.smarthome.automation.rest/src/main/java/org/eclipse/smarthome/automation/rest/internal/RuleResource.java
```diff
@@ -1,4 +1,8 @@
 /**
+ * Copyright (c) 2017 by Bosch Software Innovations GmbH. All rights reserved,
```
maggu2810: Has this been added by intention?

danchom: Unfortunately it is same problem as here: https://github.com/eclipse/smarthome/pull/3671/files/025f8b55b62675de1c46753f0c0fccc32e7697b8#r129799279

triller-telekom: Any news regarding the "license issue"? Because otherwise I think the PR is alright and could be merged...

danchom: Unfortunately no, but I think it should be resolve till the end of the week.

kaikreuzer: @danchom If I can help in any way with the clarification, please let me know. And remember that we soon will have different headers anyhow...

PR #3917, opened 2017-07-31
BaseThingHandler initialize() now abstract, does not set Thing to ONLINE anymore

triller-telekom:
The background for this PR is that a lot of new binding developers do the super call and "accidentally" set their thing to online.

### DIFF
--- a/bundles/core/org.eclipse.smarthome.core.thing.test/src/test/groovy/org/eclipse/smarthome/core/thing/binding/BindingBaseClassesOSGiTest.groovy
+++ b/bundles/core/org.eclipse.smarthome.core.thing.test/src/test/groovy/org/eclipse/smarthome/core/thing/binding/BindingBaseClassesOSGiTest.groovy
@@ -116,6 +116,8 @@ class BindingBaseClassesOSGiTest extends OSGiTest {
             // check getBridge works
             assertThat getBridge().getUID().toString(), is("bindingId:type1:bridgeId")
         }
+
+        void initialize() {}
     }
 
     class SimpleBridgeHandler extends BaseBridgeHandler {
@@ -131,6 +133,8 @@ class BindingBaseClassesOSGiTest extends OSGiTest {
         public void updateBridgetatus(ThingStatus status) {
             updateStatus(status)
         }
+
+        void initialize() {}
     }
 
 
@@ -313,6 +317,8 @@ class BindingBaseClassesOSGiTest extends OSGiTest {
         public void handleCommand(ChannelUID channelUID, Command command) {
             // not implemented
         }
+
+        void initialize() {}
 
         @Override
         public Collection<ConfigStatusMessage> getConfigStatus() {
--- a/bundles/core/org.eclipse.smarthome.core.thing.test/src/test/groovy/org/eclipse/smarthome/core/thing/binding/BindingBaseClassesOSGiTest.groovy
+++ b/bundles/core/org.eclipse.smarthome.core.thing.test/src/test/groovy/org/eclipse/smarthome/core/thing/binding/BindingBaseClassesOSGiTest.groovy
@@ -117,7 +117,10 @@ class BindingBaseClassesOSGiTest extends OSGiTest {
             assertThat getBridge().getUID().toString(), is("bindingId:type1:bridgeId")
         }
 
-        void initialize() {}
+        @Override
+        void initialize() {
+            updateStatus(ThingStatus.ONLINE);
+        }
     }
 
     class SimpleBridgeHandler extends BaseBridgeHandler {
@@ -134,7 +137,10 @@ class BindingBaseClassesOSGiTest extends OSGiTest {
             updateStatus(status)
         }
 
-        void initialize() {}
+        @Override
+        void initialize() {
+            updateStatus(ThingStatus.ONLINE);
+        }
     }
 
 
@@ -318,7 +324,10 @@ class BindingBaseClassesOSGiTest extends OSGiTest {
             // not implemented
         }
 
-        void initialize() {}
+        @Override
+        void initialize() {
+            updateStatus(ThingStatus.ONLINE);
+        }
 
         @Override
         public Collection<ConfigStatusMessage> getConfigStatus() {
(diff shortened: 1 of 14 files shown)


lolodomo:
Please don't merge this PR, **it reverts one of my recent bug fix in the Sonos binding** (bug making TTS not working). Please first rebase your code on current code source.

And does this change not break existing bindings that call `super.initialize()` ?

lolodomo:
Replacing `super.initialize()` by `updateStatus(ThingStatus.ONLINE)` is I think not what is expected. It is at least not for the Sonos binding.

lolodomo:
For the Sonos binding, just suppressing the call to `super.initialize()` should be correct.

triller-telekom:
> Please don't merge this PR, it reverts one of my recent bug fix in the Sonos binding (bug making TTS not working). Please first rebase your code on current code source.

Sorry I did a rebase but messed up the merge :( I have created a new commit fixing this merge.

Regarding the change from `super.initialize()` to `updateStatus(ThingStatus.ONLINE)` is in ALL cases right because that is what `super` was doing ;) The only question is if the `ONLINE` state makes sense, i.e. some bindings set it to `OFFLINE` immediately afterwards. I am adjusting the bindings as well.

The background for this PR is that a lot of new binding developers do the `super` call and "accidentally" set their thing to online.

lolodomo:
For Sonos, don't set to ONLINE. It is set either to ONLINE or OFFLINE.just after when trying to contact the device.

lolodomo:
In fact, just before and not just after, by the thread created by the call to onUpdate.

I was one of the developer who has no idea that super.initialize() was setting the state to ONLINE !

Don't forget to check that this call is not inside the binding archetype.

lolodomo:
By the way, setting status to ONLINE in replacement to super.initialize() is a secure option that will avoid regression with binding remaining in INITIALIZING state.
It is just unnecessary for the Sonos binding because I am sure about that ;)

triller-telekom:
Thank you for your input, I have changed the Sonos Binding as you said.

Regarding the archtype, this is already OK, see [here](https://github.com/eclipse/smarthome/blob/master/tools/archetype/binding/src/main/resources/archetype-resources/src/main/java/handler/__bindingIdCamelCase__Handler.java#L53)

maggu2810:
> The background for this PR is that a lot of new binding developers do the super call and "accidentally" set their thing to online.

Hm, this should be added to the initial comment.

The base class implementation is already abstract, so not every interface method must be implemented. Why not removing the implementation at all? It should not be necessary to mark the method abstract as a non-abstract class needs to implement all the not-yet-implemented interface methods.

I assume this will break a lot of the openHAB bindings as well. @kaikreuzer WDYT about that change?

martinvw:
@maggu2810 a PR to fix the bindings already merged has been created and looks fine, https://github.com/openhab/openhab2-addons/pull/2506#issuecomment-319116951 . However I share your concerns even more for bindings developed outside of openHAB or pending as PR at our side, we have 72 new bindings pending.

mjagdis:
Is the only justification, "a lot of new binding developers do the super call and "accidentally" set their thing to online."? Seriously?

The archetype's comments are ok. The original comments in the base class are ok. Your "fixes" to bindings simply replace the super call with a direct updateStatus so it's just inlining and fixes nothing that you aren't breaking yourself. Abstracting or removing the base implementation is just an API change that breaks in development and out-of-tree bindings.

If bindings *do* get it wrong raise PRs to fix *them* properly.

And be thankful that Linus doesn't do ESH/openHAB...

triller-telekom:
I did some tests regarding the binary compatibility of this PR, turns out that unfortunately it isn't. :( Even bindings which override `initialize()` and do NOT call `super.initialize()` would be broken. I was not aware of this problem. So thank you for pointing this out.
```
10:29:24.011 DEBUG o.e.s.b.w.h.WeatherUndergroundHandler[:79] - Initializing WeatherUnderground handler.
10:29:24.011 INFO  s.e.ThingStatusInfoChangedEvent[:43] - 'weatherunderground:weather:stefansWeather2' changed from UNINITIALIZED to INITIALIZING
10:29:24.012 INFO  s.event.ThingStatusInfoEvent[:43] - 'weatherunderground:weather:stefansWeather2' updated: UNINITIALIZED (HANDLER_INITIALIZING_ERROR): org.eclipse.smarthome.core.thing.binding.BaseThingHandler.initialize()V
10:29:24.013 INFO  s.e.ThingStatusInfoChangedEvent[:43] - 'weatherunderground:weather:stefansWeather2' changed from INITIALIZING to UNINITIALIZED (HANDLER_INITIALIZING_ERROR): org.eclipse.smarthome.core.thing.binding.BaseThingHandler.initialize()V
10:29:24.038 ERROR o.e.s.c.t.i.ThingManager[:743] - Exception occurred while initializing handler of thing 'weatherunderground:weather:stefansWeather2': java.lang.AbstractMethodError: org.eclipse.smarthome.core.thing.binding.BaseThingHandler.initialize()V
java.util.concurrent.ExecutionException: java.lang.AbstractMethodError: org.eclipse.smarthome.core.thing.binding.BaseThingHandler.initialize()V
	at java.util.concurrent.FutureTask.report(FutureTask.java:122)
	at java.util.concurrent.FutureTask.get(FutureTask.java:206)
	at org.eclipse.smarthome.core.common.SafeMethodCaller.callAsynchronous(SafeMethodCaller.java:194)
	at org.eclipse.smarthome.core.common.SafeMethodCaller.call(SafeMethodCaller.java:83)
	at org.eclipse.smarthome.core.common.SafeMethodCaller.call(SafeMethodCaller.java:67)
	at org.eclipse.smarthome.core.thing.internal.ThingManager.doInitializeHandler(ThingManager.java:728)
```

I will now change the PR in a way that I fix the bindings and change the `BaseThinghandler.initialize()` to print a warning that it should not be called. If we get reports/issues that people see this warning we can then fix the binding and eventually in the future, if we have another binary incompatible change, chnage it to be an abstract method.

triller-telekom:
I have made this change non API breaking by marking the method as `@Deprecated` and logging a warning. Are you fine with this change now?

@maggu2810 @lolodomo @mjagdis

Issue #3919, opened 2017-07-31
Resource leak: Hue binding


maggu2810:
* Description: Resource leak: '<unassigned Closeable value>' is never closed
* Resource: HttpClient.java	
* Path: /org.eclipse.smarthome.binding.hue/src/main/java/org/eclipse/smarthome/binding/hue/internal
* Location: line 72
* Type: Java Problem

PR #3920, opened 2017-07-31
fix resource leak

maggu2810:
Fixes: https://github.com/eclipse/smarthome/issues/3919

### DIFF
--- a/extensions/binding/org.eclipse.smarthome.binding.hue/src/main/java/org/eclipse/smarthome/binding/hue/internal/HttpClient.java
+++ b/extensions/binding/org.eclipse.smarthome.binding.hue/src/main/java/org/eclipse/smarthome/binding/hue/internal/HttpClient.java
@@ -68,10 +68,12 @@ class HttpClient {
     }
 
     private static String convertStreamToString(InputStream is) {
-        try {
-            return new Scanner(is).useDelimiter("\\A").next();
-        } catch (NoSuchElementException e) {
-            return "";
+        try (final Scanner scanner = new Scanner(is)) {
+            try {
+                return scanner.useDelimiter("\\A").next();
+            } catch (NoSuchElementException ex) {
+                return "";
+            }
         }
     }
 


maggu2810:
Hm, I don't see any reason to use an third party (`IOUtils`) function for something that could be solved our self without adding lot's of code. As we could use `try-with-resources` that solves the resource leak with standard language features, I prefer using Java features only.

htreu:
To be clear on my first post: I was not talking about your refactoring, which makes perfect sense here. My suggestion is to drop the private method `convertStreamToString ` entirely and replace it by the IOUtils call.

We touched this topic quite a few times now. And I guess its matter of taste somehow: How far do you want to take protection from the 'not invented here'  syndrome vs. implementing small features and not depend on too many external libraries.
From my pov I would rather depend on wide spread external libraries then doing stuff on my own. And when it comes to dependencies which are already there its even more of an argument for me.

maggu2810:
@htreu To be clear, too: It is okay for me to use wide spread external libraries that are present in the most runtimes at all (so IMHO the most common ones).

I am not familiar with the Hue code, but IMHO introducing a separate function or using `IOUtils` directly depends how often the method `convertStreamToString` is used and how the implementation using `IOUtils` looks like.
I assume `IOUtils.toString(is)` does not do the job. We need to split on "\\A" and take the first entry. Correct?
Isn't a separate function with a represent name better if we need to use such a split and selection multiple times?

As you know the code base better, can you give me an example how the change would look like?

htreu:
I guess the main thing here is the line `scanner.useDelimiter("\\A").next()` which was technically not clear to me either but I must admit I did take the exact same line from stackoverflow in the past to convert InputStreams to Strings. I did a little research and found this article: https://community.oracle.com/blogs/pat/2004/10/23/stupid-scanner-tricks

The line effectively takes the whole content of the InputStream and reads it into the String. The \A refers to the beginning of the Stream. So the Scanner will tokenize the whole Stream until it reaches the next beginning (which is not present here) resulting in the whole string being taken as the first token.

With this in mind I guess its fine to replace the method. Its only called once in the HttpClient in the hue binding.

On another note: We should replace the internal HttpClient from the hue binding by the jetty client (or even better by the `org.eclipse.smarthome.io.net.http.HttpUtil`) but thats another PR with some more analysis :-)

kaikreuzer:
The PR as it is is imho fine, so I merge it.

Nonetheless, I agree with @htreu that it makes sense to refactor the code and remove this method as its implementation is not easily understandable and we allow extensions the use of Commons IO, which offers a stable feature for getting a String out of an InputStream.

maggu2810:
> I agree with @htreu that it makes sense to refactor the code and remove this method

It's fine for me, too.
Perhaps @htreu would like a PR that address it. :wink:

htreu:
here you go :-)

PR #3925, opened 2017-08-01
Add exported packages that are not imported

svilenvul:


maggu2810:
... because every exported package should be imported again?

svilenvul:
According to the [guidelines](https://www.eclipse.org/smarthome/documentation/development/guidelines.html#b-osgi-bundles), yes: 
> Every exported package of a bundle must be imported by the bundle itself again.

PR #3930, opened 2017-08-01
Implemented Option to choose default network interface

### LINKED ISSUE #3884: Allow central configuration of the primary network interface
Currently, `NetUtil.getLocalIpv4HostAddress()` makes a best effort guess about which is the IP address that is used to be reachable for external devices/systems. If users have multiple network interfaces, this has a high risk to fail.

I think we therefore need to provide a central configuration where the desired network address can be chosen.

triller-telekom:
Fixes #3884 


triller-telekom:
Close/reopen due to eclipse repository not available...

bundles/config/org.eclipse.smarthome.config.core/src/main/java/org/eclipse/smarthome/config/core/net/NetworkConfigOptionProvider.java
```diff
@@ -0,0 +1,81 @@
+/**
+ * Copyright (c) 2014-2017 by the respective copyright holders.
+ * All rights reserved. This program and the accompanying materials
+ * are made available under the terms of the Eclipse Public License v1.0
+ * which accompanies this distribution, and is available at
+ * http://www.eclipse.org/legal/epl-v10.html
+ */
+package org.eclipse.smarthome.config.core.net;
```
kaikreuzer: Is there any need to have this in an exported package?

bundles/core/org.eclipse.smarthome.core/ESH-INF/config/networkinterface.xml
```diff
@@ -0,0 +1,17 @@
+<?xml version="1.0" encoding="UTF-8"?>
+<config-description:config-descriptions xmlns:xsi="http://www.w3.org/2001/XMLSchema-instance"
+	xmlns:config-description="http://eclipse.org/smarthome/schemas/config-description/v1.0.0"
+	xsi:schemaLocation="http://eclipse.org/smarthome/schemas/config-description/v1.0.0
+        http://eclipse.org/smarthome/schemas/config-description-1.0.0.xsd">
+
+	<config-description uri="system:network">
+		<parameter name="defaultInterface" type="text">
+			<label>Default Interface</label>
+			<description><![CDATA[
```
kaikreuzer: No need to put this in a CDATA section.

bundles/core/org.eclipse.smarthome.core/ESH-INF/config/networkinterface.xml
```diff
@@ -0,0 +1,17 @@
+<?xml version="1.0" encoding="UTF-8"?>
+<config-description:config-descriptions xmlns:xsi="http://www.w3.org/2001/XMLSchema-instance"
+	xmlns:config-description="http://eclipse.org/smarthome/schemas/config-description/v1.0.0"
+	xsi:schemaLocation="http://eclipse.org/smarthome/schemas/config-description/v1.0.0
+        http://eclipse.org/smarthome/schemas/config-description-1.0.0.xsd">
+
+	<config-description uri="system:network">
+		<parameter name="defaultInterface" type="text">
```
kaikreuzer: Does this parameter really hold a network interface identifier or rather an IPv4 address?
Specifying the interface probably makes sense as it does not have to be updated, if the runtime is assigned a new ip address (on the same interface and subnet) - but this means that your NetworkInterfaceService should also be adapted.

triller-telekom:
rebased against current master to resolve conflicts

bundles/core/org.eclipse.smarthome.core/ESH-INF/config/networkinterface.xml
```diff
@@ -0,0 +1,16 @@
...
+<config-description:config-descriptions xmlns:xsi="http://www.w3.org/2001/XMLSchema-instance"
+	xmlns:config-description="http://eclipse.org/smarthome/schemas/config-description/v1.0.0"
+	xsi:schemaLocation="http://eclipse.org/smarthome/schemas/config-description/v1.0.0
+        http://eclipse.org/smarthome/schemas/config-description-1.0.0.xsd">
+
+	<config-description uri="system:network">
+		<parameter name="primaryAddress" type="text">
+			<label>Primary Address</label>
+			<description><![CDATA[<p>The primary network to be used</p>
+			<p>Alternative: Enter an IP address (XXX.XXX.XXX.XXX) manually</p>]]></description>
```
kaikreuzer: You do not mention the alternative to the alternative :-)

Better say:
```
A subnet (e.g. 192.168.1.0/24) or an IP address (e.g. 192.168.1.5)
```

bundles/config/org.eclipse.smarthome.config.core/META-INF/MANIFEST.MF
```diff
@@ -4,12 +4,13 @@ Bundle-Name: Eclipse SmartHome Config Core
...
 Bundle-Version: 0.9.0.qualifier
 Bundle-Vendor: Eclipse.org/SmartHome
-Bundle-RequiredExecutionEnvironment: JavaSE-1.8
+Bundle-RequiredExecutionEnvironment: JavaSE-1.8
 Import-Package: com.google.common.base,
  com.google.common.collect,
  com.google.gson,
  org.apache.commons.lang.reflect,
  org.eclipse.jdt.annotation;resolution:=optional,
+ org.apache.commons.net.util,
```
kaikreuzer: We imho must not introduce this dependency to the ESH core bundles.

bundles/core/org.eclipse.smarthome.core/META-INF/MANIFEST.MF
```diff
@@ -39,6 +39,7 @@ Import-Package: com.google.common.base,
  org.apache.commons.io,
  org.apache.commons.lang,
  org.eclipse.jdt.annotation;resolution:=optional,
+ org.apache.commons.net.util,
```
kaikreuzer: We imho must not introduce this dependency to the ESH core bundles.

bundles/core/org.eclipse.smarthome.core/src/main/java/org/eclipse/smarthome/core/net/NetUtil.java
```diff
@@ -25,12 +31,51 @@
...
+    @SuppressWarnings("unchecked")
+    protected void activate(ComponentContext componentContext) {
+        Dictionary<String, Object> props = componentContext.getProperties();
+        modified((Map<String, Object>) props);
+    }
+
+    @Modified
+    public synchronized void modified(Map<String, Object> config) {
+        String defaultInterfaceConfig = (String) config.get(PRIMARY_ADDRESS);
+        if (defaultInterfaceConfig == null || defaultInterfaceConfig.equals("")) {
```
kaikreuzer: better use `.isEmpty()`

bundles/core/org.eclipse.smarthome.core/src/main/java/org/eclipse/smarthome/core/net/NetworkAddressProvider.java
```diff
@@ -0,0 +1,19 @@
...
+
+/**
+ * Interface that provides access to configured network addresses
+ *
+ * @author Stefan Triller - initial contribution
+ *
+ */
+public interface NetworkAddressProvider {
+
+    public String getPrimaryIpv4HostAddress();
```
kaikreuzer: How about adding a NonNull annotation? :-)
On an interface, you can remove the "public" keyword.

docs/documentation/development/bindings/thing-handler.md
```diff
@@ -180,6 +180,27 @@ Furthermore bindings can specify a localized description of the thing status by
 rate_limit=Device is blocked by remote service for {0} minutes. Maximum limit of {1} configuration changes per {2} has been exceeded. For further info please refer to device vendor.
 ```
 
+## Offering a callback URL
+
+Some things might require a `callback` URL which should be bound to a certain network interface. A user can configure his default network address via Paper UI under `Configuration -> System -> Network Settings`. To obtain this configured address the `ThingHandlerFactory` needs a `service reference` to the `NetworkInterfaceService` in its `OSGI-INF/MyHandlerFactory.xml`:
```
kaikreuzer: Please add a new line for every sentence - this makes diffing and reviews easier.

The correct name seems to be `NetworkAddressProvider`.
The example is not good, because we by now ask people to use DS annotations, not XML files.

docs/documentation/development/bindings/thing-handler.md
```diff
@@ -180,6 +180,27 @@ Furthermore bindings can specify a localized description of the thing status by
 rate_limit=Device is blocked by remote service for {0} minutes. Maximum limit of {1} configuration changes per {2} has been exceeded. For further info please refer to device vendor.
 ```
 
+## Offering a callback URL
```
kaikreuzer: I am not sure whether this is the correct place for the documentation of this feature as the feature is completely independent of bindings. It is often used for the implementation of AudioSinks and potentially other extensions. We should rather have a place for such general system services.

extensions/binding/org.eclipse.smarthome.binding.sonos/src/main/java/org/eclipse/smarthome/binding/sonos/internal/SonosHandlerFactory.java
```diff
@@ -171,4 +172,12 @@ protected void unsetAudioHTTPServer(AudioHTTPServer audioHTTPServer) {
         this.audioHTTPServer = null;
     }
 
+    protected void setNetworkAddressProvider(NetworkAddressProvider networkUtil) {
```
kaikreuzer: change networkUtil to networkAddressProvider

extensions/binding/org.eclipse.smarthome.binding.sonos/src/main/java/org/eclipse/smarthome/binding/sonos/internal/SonosHandlerFactory.java
```diff
@@ -171,4 +172,12 @@ protected void unsetAudioHTTPServer(AudioHTTPServer audioHTTPServer) {
...
+        this.networkAddressprovider = networkUtil;
```
kaikreuzer: change networkAddressprovider to networkAddressProvider

extensions/binding/org.eclipse.smarthome.binding.sonos/src/main/java/org/eclipse/smarthome/binding/sonos/internal/SonosHandlerFactory.java
```diff
@@ -171,4 +172,12 @@ protected void unsetAudioHTTPServer(AudioHTTPServer audioHTTPServer) {
...
+    }
+
+    protected void unsetNetworkAddressProvider(NetworkAddressProvider networkUtil) {
```
kaikreuzer: dito

bundles/config/org.eclipse.smarthome.config.core/src/main/java/org/eclipse/smarthome/config/core/net/internal/NetworkConfigOptionProvider.java
```diff
@@ -0,0 +1,96 @@
...
+        if (param.equals(PARAM_PRIMARY_ADDRESS)) {
+            return getIPv4Addresses();
+        }
+        return null;
+    }
+
+    private List<ParameterOption> getIPv4Addresses() {
+        ArrayList<ParameterOption> interfaceOptions = new ArrayList<>();
+
+        HashSet<String> subnets = new HashSet<>();
```
kaikreuzer: Change first occurrence to `Set`

bundles/config/org.eclipse.smarthome.config.core/src/main/java/org/eclipse/smarthome/config/core/net/internal/NetworkConfigOptionProvider.java
```diff
@@ -0,0 +1,96 @@
...
+
+                    if (addr.isLoopbackAddress() || (addr instanceof Inet6Address)) {
+                        continue;
+                    }
+
+                    String ipv4Address = addr.getHostAddress();
+                    String subNetString = NetUtil.getIpv4NetAddress(ipv4Address, ifAddr.getNetworkPrefixLength()) + "/"
+                            + String.valueOf(ifAddr.getNetworkPrefixLength());
+
+                    subnets.add(subNetString);
```
kaikreuzer: Although you didn't add any annotation to `NetUtil.getIpv4NetAddress`, I assume that `subNetString` can be null here.

bundles/core/org.eclipse.smarthome.core/src/main/java/org/eclipse/smarthome/core/net/NetUtil.java
```diff
@@ -25,17 +31,63 @@
...
+                LOGGER.warn("Error in IP configuration, will continue to use first interface");
+                primaryIP = NetUtil.getLocalIpv4HostAddress();
+            } else {
+                primaryIP = ip;
+            }
+        } else {
+            primaryIP = addrString[0];
+        }
+
+        return primaryIP;
```
kaikreuzer: How did you convince your IDE that primaryIP is definitely not null here?

bundles/core/org.eclipse.smarthome.core/src/main/java/org/eclipse/smarthome/core/net/NetUtil.java
```diff
@@ -25,17 +31,63 @@
...
     }
 
     /**
      * Get the first candidate for a local IPv4 host address (non loopback, non localhost).
      */
+    @Deprecated
```
kaikreuzer: Please add a message about what people should now use instead.

bundles/core/org.eclipse.smarthome.core/src/main/java/org/eclipse/smarthome/core/net/NetUtil.java
```diff
@@ -104,4 +156,108 @@ public static String getBroadcastAddress() {
...
+    }
+
+    /**
+     * Get the network address a specific ip address is in
+     *
+     * @param ipAddressString ipv4 address of the device (i.e. 192.168.5.1)
+     * @param netMask netmask in bits (i.e. 24)
+     * @return network a device is in (i.e. 192.168.5.0)
+     */
+    public static String getIpv4NetAddress(String ipAddressString, short netMask) {
```
kaikreuzer: where are the null annotation?

bundles/core/org.eclipse.smarthome.core/src/main/java/org/eclipse/smarthome/core/net/NetUtil.java
```diff
@@ -104,4 +156,108 @@ public static String getBroadcastAddress() {
...
     }
 
+    /**
+     * Converts a netmask in bits into a string representation
+     * i.e. 24 bits -> 255.255.255.0
+     *
+     * @param prefixLength bits of the netmask
+     * @return string representation of netmask (i.e. 255.255.255.0)
+     */
+    public static String networkPrefixLengthToNetmask(int prefixLength) {
```
kaikreuzer: where are the null annotation?

docs/_includes/documentation-menu.html
```diff
@@ -16,6 +16,7 @@
 			<li><a href="{{docu}}/features/rest.html">REST API</a></li>
 			<li><a href="{{docu}}/features/dsl.html">Textual Configuration</a></li>
 			<li><a href="{{docu}}/features/internationalization.html">Internationalization</a></li>
+			<li><a href="{{docu}}/features/providedServices.html">Provided Services</a></li>
```
kaikreuzer: "Provided Services" sounds weird.
How about something like "Utilities", "System Services" or "Framework Services/Utilities"?
Note that not everything here might be real "services" (e.g. our caching classes, etc.).
@maggu2810 Any good suggestion or preference from your end?

docs/documentation/features/providedServices.md
```diff
@@ -0,0 +1,23 @@
...
+layout: documentation
+---
+
+# Provided Services
+
+In this chapter useful services of the Eclipse SmartHome project are described. 
+
+## Obtaining the default IP address
+
+Some things might require a `callback` URL which should be bound to a certain network interface. A user can configure his default network address via Paper UI under `Configuration -> System -> Network Settings`. To obtain this configured address the `ThingHandlerFactory` needs a `service reference` to the `NetworkInterfaceService` in its `MyHandlerFactory.java`:
```
kaikreuzer: Every sentence in a new line please.

kaikreuzer: Change `NetworkInterfaceService`.

kaikreuzer: Do not talk about "things" and "handlers", we are not in the binding docs anymore. Rather describe your service in general and merely mention in a side note that it might be helpful for callback functionality in ThingHandlers and AudioSinks.

bundles/core/org.eclipse.smarthome.core/src/main/java/org/eclipse/smarthome/core/net/NetUtil.java
```diff
@@ -25,17 +31,63 @@
  *
  * @author Markus Rathgeb - Initial contribution and API
  * @author Mark Herwege - Added methods to find broadcast address(es)
+ * @author Stefan Triller - Converted to OSGi service with primary ipv4 conf
  */
-public class NetUtil {
+@Component(name = "org.eclipse.smarthome.network", property = { "service.config.description.uri=system:network",
+        "service.config.label=Network Settings", "service.config.category=system" })
+public class NetUtil implements NetworkAddressProvider {
```
kaikreuzer: Sorry, just realised your name choice now: "Provider" has a very special meaning in many contexts in ESH. `NetworkAddressService` would probably a more neutral choice.

bundles/core/org.eclipse.smarthome.core/src/main/java/org/eclipse/smarthome/core/net/NetUtil.java
```diff
@@ -25,17 +31,63 @@
...
+    @Override
+    public String getPrimaryIpv4HostAddress() {
+        String primaryIP;
+
+        String[] addrString = primaryAddress.split("/");
+        if (addrString.length > 1) {
+            String ip = getIPv4inSubnet(primaryAddress);
+            if (ip == null) {
+                // an error has occurred, using first interface like nothing has been configured
+                LOGGER.warn("Error in IP configuration, will continue to use first interface");
```
kaikreuzer: Change to:
```
`"Invalid address '{}', will use first interface instead.", primaryAddress);`

File: bundles/core/org.eclipse.smarthome.core/src/main/java/org/eclipse/smarthome/core/net/NetUtil.java [hunk 24]
```diff
@@ -104,4 +156,108 @@ public static String getBroadcastAddress() {
...
+
+                    // use first IP within this subnet
+                    if (subNetString.equals(subnet)) {
+                        return ipv4Address;
+                    }
+                }
+            }
+        } catch (SocketException ex) {
+            LOGGER.error("Could not retrieve network interface: {}", ex.getMessage(), ex);
+            return null;
```
kaikreuzer: this line is superfluous.

docs/documentation/features/providedServices.md
```diff
@@ -0,0 +1,23 @@
+---
+layout: documentation
+---
+
+# Provided Services
+
+In this chapter useful services of the Eclipse SmartHome project are described. 
+
+## Obtaining the default IP address
```
kaikreuzer: Use the service name for the header: "Network Address Service" or something like that.

extensions/binding/org.eclipse.smarthome.binding.sonos/src/main/java/org/eclipse/smarthome/binding/sonos/discovery/ZonePlayerDiscoveryParticipant.java
```diff
@@ -29,6 +30,7 @@
  *
  * @author Karel Goderis - Initial contribution
  */
+@Component(name = "org.eclipse.smarthome.binding.sonos.discovery.zoneplayer", immediate = true)
```
kaikreuzer: Afaik, we do not need a specific service name for that one, so I'd suggest to go for the default and remove that name property here.

bundles/core/org.eclipse.smarthome.core/src/main/java/org/eclipse/smarthome/core/net/NetUtil.java
```diff
@@ -104,4 +159,107 @@ public static String getBroadcastAddress() {
...
+    }
+
+    /**
+     * Get the network address a specific ip address is in
+     *
+     * @param ipAddressString ipv4 address of the device (i.e. 192.168.5.1)
+     * @param netMask netmask in bits (i.e. 24)
+     * @return network a device is in (i.e. 192.168.5.0)
+     */
+    public static @NonNull String getIpv4NetAddress(@NonNull String ipAddressString, short netMask) {
```
kaikreuzer: If you want to assure to always return valid non-null results, you should at least also throw IllegalArgumentExceptions. Or what would you return if I pass you "ABC" as my ipAddressString?

bundles/core/org.eclipse.smarthome.core/src/main/java/org/eclipse/smarthome/core/net/NetUtil.java
```diff
@@ -104,4 +159,107 @@ public static String getBroadcastAddress() {
...
+        String subnetMaskString = networkPrefixLengthToNetmask(netMask);
+
+        String[] netMaskOctets = subnetMaskString.split("\\.");
+        String[] ipv4AddressOctets = ipAddressString.split("\\.");
+        String netAddress = "";
+        for (int i = 0; i < 4; i++) {
+            netAddress += Integer.parseInt(ipv4AddressOctets[i]) & Integer.parseInt(netMaskOctets[i]);
```
kaikreuzer: You'll need to catch NumberFormatException - this should not bubble up in the call stack.

bundles/core/org.eclipse.smarthome.core/src/main/java/org/eclipse/smarthome/core/net/NetUtil.java
kaikreuzer: This method is a wonderful candidate for a couple of unit tests!

bundles/core/org.eclipse.smarthome.core/src/main/java/org/eclipse/smarthome/core/net/NetUtil.java
```diff
@@ -25,17 +32,65 @@
...
+    private static final String PRIMARY_ADDRESS = "primaryAddress";
     private static final Logger LOGGER = LoggerFactory.getLogger(NetUtil.class);
 
-    private NetUtil() {
+    private static final Pattern IPV4_PATTERN = Pattern
+            .compile("^(([01]?\\d\\d?|2[0-4]\\d|25[0-5])\\.){3}([01]?\\d\\d?|2[0-4]\\d|25[0-5])$");
+
+    private String primaryAddress;
+
+    @SuppressWarnings("unchecked")
```
kaikreuzer: Why that?

bundles/core/org.eclipse.smarthome.core/src/main/java/org/eclipse/smarthome/core/net/NetUtil.java
```diff
@@ -25,17 +32,65 @@
...
+    protected void activate(ComponentContext componentContext) {
```
kaikreuzer: Instead of pulling the properties out of the context, why don't you simply implement `activate(Map<String, Object> props)`?

bundles/core/org.eclipse.smarthome.core/src/main/java/org/eclipse/smarthome/core/net/NetUtil.java
```diff
@@ -25,17 +32,65 @@
...
+        modified((Map<String, Object>) props);
+    }
+
+    @Modified
+    public synchronized void modified(Map<String, Object> config) {
+        String primaryAddressConf = (String) config.get(PRIMARY_ADDRESS);
+        if (primaryAddressConf == null || primaryAddressConf.isEmpty() || !isValidIPConfig(primaryAddressConf)) {
+            // if none is specified we return the default one for backward compatibility
+            LOGGER.warn("Non valid IP configuration found, will continue to use first interface");
+            primaryAddress = NetUtil.getLocalIpv4HostAddress();
```
kaikreuzer: You are calling a deprecated method here. Should be avoided :-)

bundles/core/org.eclipse.smarthome.core/src/main/java/org/eclipse/smarthome/core/net/NetworkAddressService.java
```diff
@@ -0,0 +1,22 @@
...
+/**
+ * Interface that provides access to configured network addresses
+ *
+ * @author Stefan Triller - initial contribution
+ *
+ */
+public interface NetworkAddressService {
+
+    @Nullable
+    String getPrimaryIpv4HostAddress();
```
kaikreuzer: please add JavaDoc

docs/documentation/features/frameworkUtilities.md
```diff
@@ -0,0 +1,25 @@
+---
+layout: documentation
+---
+
+# Provided Services
```
kaikreuzer: Shouldn't it now say "Framework Utilities"?

docs/documentation/features/frameworkUtilities.md
```diff
@@ -0,0 +1,25 @@
...
+
+In this chapter useful services of the Eclipse SmartHome project are described. 
+
+## Network Address Service
+
+A user can configure his default network address via Paper UI under `Configuration -> System -> Network Settings`.
+To obtain this configured address the `ThingHandlerFactory` needs a `service reference` to the `NetworkAddressService` in its `MyHandlerFactory.java`:
```
kaikreuzer: This is still binding specific doc

docs/documentation/features/frameworkUtilities.md
```diff
@@ -0,0 +1,25 @@
...
+
+```java
+@Reference
```
kaikreuzer: Do you really think we need to explain how OSGi DS works for every single service we offer?
Wouldn't it be more important to state that the NetworkAddressService IS an OSGi service that can be used?

docs/documentation/features/frameworkUtilities.md
```diff
@@ -0,0 +1,25 @@
...
+protected void setNetworkAddressService(NetworkAddressService networkAddressService) {
+	this.networkAddressService = networkAddressService;
+	}
+protected void unsetNetworkAddressService(NetworkAddressService networkAddressService) {
+	this.networkAddressService = null;
+}
+```
+
+Now the `MyHandlerFactory` can obtain the configured IP address via `networkAddressService.getPrimaryIpv4HostAddress()`.
+This IP address can for example be used for things (i.e. AudioSinks) that require a callback URL which they can offer to a device.
```
kaikreuzer: No, examples are ThingHandlerFactory and AudioSink implementations. (note: thing!=AudioSink)

bundles/core/org.eclipse.smarthome.core/src/main/java/org/eclipse/smarthome/core/net/NetUtil.java
```diff
@@ -25,17 +31,64 @@
...
+    protected void activate(Map<String, Object> props) {
+        modified(props);
+    }
+
+    @Modified
+    public synchronized void modified(Map<String, Object> config) {
+        String primaryAddressConf = (String) config.get(PRIMARY_ADDRESS);
+        if (primaryAddressConf == null || primaryAddressConf.isEmpty() || !isValidIPConfig(primaryAddressConf)) {
+            // if none is specified we return the default one for backward compatibility
+            LOGGER.warn("Non valid IP configuration found, will continue to use first interface");
```
kaikreuzer: As >90% of users will have exactly one interface, there is no need to set this configuration parameter.
Why should you bother them with a warning?
You should only show a warning if the parameter is invalid.

bundles/core/org.eclipse.smarthome.core/src/main/java/org/eclipse/smarthome/core/net/NetUtil.java
```diff
@@ -104,4 +186,127 @@ public static String getBroadcastAddress() {
...
+     * @param netMask netmask in bits (i.e. 24)
+     * @return network a device is in (i.e. 192.168.5.0)
+     */
+    public static @NonNull String getIpv4NetAddress(@NonNull String ipAddressString, short netMask) {
+
+        String errorString = "IP '" + ipAddressString + "' not a valid IPv4 address or netmask '" + netMask
+                + "' out of bounds (1-31)";
+
+        if (!isValidIPConfig(ipAddressString) || netMask < 1 || netMask > 31) {
+            throw new IllegalArgumentException(errorString);
```
kaikreuzer: You should mention this exception in the JavaDoc.

bundles/core/org.eclipse.smarthome.core/src/main/java/org/eclipse/smarthome/core/net/NetworkAddressService.java
```diff
@@ -0,0 +1,27 @@
...
+ *
+ * @author Stefan Triller - initial contribution
+ *
+ */
+public interface NetworkAddressService {
+
+    /**
+     * Returns the user configured primary IPv4 address of the system
+     * 
+     * @return IPv4 address as a String in format xxx.xxx.xxx.xxx
```
kaikreuzer: .. or null if bla bla bla...?

docs/documentation/features/frameworkUtilities.md
```diff
@@ -0,0 +1,14 @@
...
+---
+
+# Framework Utilities
+
+In this chapter useful services/utilities of the Eclipse SmartHome project are described. 
+
+## Network Address Service
+
+A user can configure his default network address via Paper UI under `Configuration -> System -> Network Settings`.
+The `NetworkAddressService` is an OSGi service that can be used like any other OSGi service by adding a service reference to it.
```
kaikreuzer: This should be the first sentence of the paragraph. Maybe also mention that it is a configurable service, mentioning its service id. The Paper UI example above is then just one way - in general anything that uses ConfigurationAdmin can be used to configure the service.

bundles/config/org.eclipse.smarthome.config.core/src/main/java/org/eclipse/smarthome/config/core/net/internal/NetworkConfigOptionProvider.java
```diff
@@ -0,0 +1,97 @@
...
+
+                for (InterfaceAddress ifAddr : current.getInterfaceAddresses()) {
+                    InetAddress addr = ifAddr.getAddress();
+
+                    if (addr.isLoopbackAddress() || (addr instanceof Inet6Address)) {
+                        continue;
+                    }
+
+                    String ipv4Address = addr.getHostAddress();
+                    String subNetString = NetUtil.getIpv4NetAddress(ipv4Address, ifAddr.getNetworkPrefixLength()) + "/"
```
kaikreuzer: What do you do if this now throws the newly introduced `IllegalArgumentException`?

triller-telekom:
close/reopen because of unreachable maven repo

bundles/core/org.eclipse.smarthome.core.test/src/test/java/org/eclipse/smarthome/core/net/NetUtilTest.java
```diff
@@ -0,0 +1,68 @@
...
+
+        network = NetUtil.getIpv4NetAddress("192.168.5.8", (short) 27);
+        assertThat(network, is("192.168.5.0"));
+
+        network = NetUtil.getIpv4NetAddress("192.168.5.8", (short) 29);
+        assertThat(network, is("192.168.5.8"));
+
+        try {
+            network = NetUtil.getIpv4NetAddress("192.168.5.8", (short) 32);
+            ;
```
kaikreuzer: Is this line required?

bundles/core/org.eclipse.smarthome.core.test/src/test/java/org/eclipse/smarthome/core/net/NetUtilTest.java
```diff
@@ -0,0 +1,68 @@
...
+        } catch (IllegalArgumentException iae) {
+            assertThat(iae.getMessage(),
+                    is("IP '192.168.5.8' not a valid IPv4 address or netmask '32' out of bounds (1-31)"));
```
kaikreuzer: I would actually except a clear message about the error and not "this or that is wrong".
```
"Netmask '32' is out of bounds (1-31)"
```
Would be the expected result, do you agree?

bundles/core/org.eclipse.smarthome.core.test/src/test/java/org/eclipse/smarthome/core/net/NetUtilTest.java
```diff
@@ -0,0 +1,68 @@
...
+        }
+        try {
+            network = NetUtil.getIpv4NetAddress("192.168.58", (short) 24);
+        } catch (IllegalArgumentException iae) {
+            assertThat(iae.getMessage(),
+                    is("IP '192.168.58' not a valid IPv4 address or netmask '24' out of bounds (1-31)"));
```
kaikreuzer: "IP '192.168.58' is not a valid IPv4 address."

bundles/core/org.eclipse.smarthome.core.test/src/test/java/org/eclipse/smarthome/core/net/NetUtilTest.java
```diff
@@ -0,0 +1,68 @@
...
+        }
+
+        try {
+            network = NetUtil.getIpv4NetAddress("SOME_TEXT", (short) 24);
+        } catch (IllegalArgumentException iae) {
+            assertThat(iae.getMessage(),
+                    is("IP 'SOME_TEXT' not a valid IPv4 address or netmask '24' out of bounds (1-31)"));
```
kaikreuzer: "IP 'SOME_TEXT' is not a valid IPv4 address"

bundles/config/org.eclipse.smarthome.config.core/src/main/java/org/eclipse/smarthome/config/core/net/internal/NetworkConfigOptionProvider.java
```diff
@@ -0,0 +1,103 @@
...
+                for (InterfaceAddress ifAddr : current.getInterfaceAddresses()) {
+                    InetAddress addr = ifAddr.getAddress();
+
+                    if (addr.isLoopbackAddress() || (addr instanceof Inet6Address)) {
+                        continue;
+                    }
+
+                    String ipv4Address = addr.getHostAddress();
+                    try {
+                        // InetAddress.getHostAddress() is missing a @NonNull annotation
```
kaikreuzer: You can make it clear by code, which conversion you are suppressing - no need for this comment. Just do above the following:
```
        @SuppressWarnings("null")
        @NonNull
        String ipv4Address = addr.getHostAddress();
```
I think that is the better style we should follow.

docs/_includes/documentation-menu.html
maggu2810: The choosen "Framework Utilities" is okay for me

PR #3932, opened 2017-08-02
Add ExpiringCacheAsync cache and test

davidgraeff:
This PR contains an asynchronous cache implementation, modeled after the already existing ExpiringCache. A unit test suite is included.

## The status quo
It is a good practice to return as fast as possible from the `handleCommand(ChannelUID channelUID, Command command)` method of a **ThingHandler** to not block callers especially UIs. Usually commands modify an internal state and push that state via a separate (network) thread to the target device, no delay here. 

The problem is the "RefreshType". In my opinion this should have been handled in another function, but it is how it is now. If we refreshed the internal state just recently, we can return it immediately via the usual `updateState(channel, state)` method. It the state is too old, we need to fetch it first and then return it. Until now either bindings have their own non-blocking solution or use a blocking way of retrieve the new state.

## ExpiringCacheAsync
The new cache class solves this issue, by returning a state immediately if not expired yet or issues a fetch request and returns the state asynchronously.

## Usage pattern

A **handleCommand** implementation with the interesting *RefreshType* could look like this:
```
public void handleCommand(ChannelUID channelUID, Command command) {
    if (command instanceof RefreshType) {
        switch (channelUID.getId()) {
            case CHANNEL_1:
                cache1.getValue((ValueType value) -> updateState(CHANNEL_1, value);
                break;
            ...
        }
    }
}
```
The handler would create a cache like this:
```
cache1 = new ExpiringCacheAsync<ValueType>(cacheDeviceStateTimeInMS, () -> {
            performRequestValue();
        });
```

The method that is responsible for updating the cache, in this case `performRequestValue` would perform whatever is necessary to update the state. ExpiringCacheAsync will not help in spawning other threads, so the method would need to delegate work to other threads on its own. When the other thread return with a result, it just do this:
```
cache1.setValue(v);
```

### DIFF
--- a/bundles/core/org.eclipse.smarthome.core/src/main/java/org/eclipse/smarthome/core/cache/ExpiringCacheAsync.java
+++ b/bundles/core/org.eclipse.smarthome.core/src/main/java/org/eclipse/smarthome/core/cache/ExpiringCacheAsync.java
@@ -1,6 +1,5 @@
 /**
  * Copyright (c) 2010-2017 by the respective copyright holders.
- *
  * All rights reserved. This program and the accompanying materials
  * are made available under the terms of the Eclipse Public License v1.0
  * which accompanies this distribution, and is available at
@@ -24,7 +23,7 @@ import org.eclipse.jdt.annotation.Nullable;
  *
  * @param <V> the type of the cached value
  */
-public class ExpiringCacheAsync<@Nullable V> {
+public class ExpiringCacheAsync<V> {
     protected final long expiry;
     protected long expiresAt = 0;
     protected CompletableFuture<V> currentNewValueRequest = null;
@@ -53,6 +52,7 @@ public class ExpiringCacheAsync<@Nullable V> {
      *         `getValue().thenAccept(value->useYourValueHere(value));`. If you need the value synchronously you can use
      *         `getValue().get()`.
      */
+    @SuppressWarnings("null")
     public CompletableFuture<V> getValue(@NonNull Supplier<CompletableFuture<V>> requestNewValueFuture) {
         if (isExpired()) {
             return refreshValue(requestNewValueFuture);
@@ -97,6 +97,7 @@ public class ExpiringCacheAsync<@Nullable V> {
         if (currentNewValueRequest == null) {
             throw new IllegalArgumentException("We expect a CompletableFuture for refreshValue() to work!");
         }
+        @SuppressWarnings("null")
         CompletableFuture<V> t = currentNewValueRequest.thenApply(newValue -> {
             // No request is ongoing anymore, update the value and expire time
             currentNewValueRequest = null;
--- a/bundles/core/org.eclipse.smarthome.core/src/main/java/org/eclipse/smarthome/core/cache/ExpiringCacheAsync.java
+++ b/bundles/core/org.eclipse.smarthome.core/src/main/java/org/eclipse/smarthome/core/cache/ExpiringCacheAsync.java
@@ -1,0 +1,131 @@
+/**
+ * Copyright (c) 2010-2017 by the respective copyright holders.
+ *
+ * All rights reserved. This program and the accompanying materials
+ * are made available under the terms of the Eclipse Public License v1.0
+ * which accompanies this distribution, and is available at
+ * http://www.eclipse.org/legal/epl-v10.html
+ */
+package org.eclipse.smarthome.core.cache;
+
+import java.util.concurrent.CompletableFuture;
+import java.util.concurrent.TimeUnit;
+import java.util.function.Supplier;
+
+import org.eclipse.jdt.annotation.NonNull;
+import org.eclipse.jdt.annotation.Nullable;
+
+/**
+ * Complementary class to {@link org.eclipse.smarthome.core.cache.ExpiringCache}, implementing an asynchronous variant
+ * of an expiring cache. An instance returns the cached value immediately to the callback if not expired yet, otherwise
+ * issue a fetch in another thread and notify callback implementors asynchronously.
+ *
+ * @author David Graeff - Initial contribution
+ *
+ * @param <V> the type of the cached value
+ */
+public class ExpiringCacheAsync<@Nullable V> {
+    protected final long expiry;
+    protected long expiresAt = 0;
+    protected CompletableFuture<V> currentNewValueRequest = null;
+    @Nullable
+    protected V value;
+
+    /**
+     * Create a new instance.
+     *
+     * @param expiry the duration in milliseconds for how long the value stays valid. Must be greater than 0.
+     * @throws IllegalArgumentException For an expire value <=0.
+     */
+    public ExpiringCacheAsync(long expiry) throws IllegalArgumentException {
+        if (expiry <= 0) {
+            throw new IllegalArgumentException("Cache expire time must be greater than 0");
+        }
+        this.expiry = TimeUnit.MILLISECONDS.toNanos(expiry);
+    }
+
+    /**
+     * Returns the value - possibly from the cache, if it is still valid.
+     *
+     * @param requestNewValueFuture If the value is expired, this supplier is called to supply the cache with a future
+     *            that on completion will update the cached value
+     * @return the value in form of a CompletableFuture. You can for instance use it this way:
+     *         `getValue().thenAccept(value->useYourValueHere(value));`. If you need the value synchronously you can use
+     *         `getValue().get()`.
+     */
+    public CompletableFuture<V> getValue(@NonNull Supplier<CompletableFuture<V>> requestNewValueFuture) {
+        if (isExpired()) {
+            return refreshValue(requestNewValueFuture);
+        } else {
+            return CompletableFuture.completedFuture(value);
+        }
+    }
+
+    /**
+     * Invalidates the value in the cache.
+     */
+    public void invalidateValue() {
+        expiresAt = 0;
+    }
+
+    /**
+     * Returns an arbitrary time reference in nanoseconds.
+     * This is used for the cache to determine if a value has expired.
+     */
+    protected long getCurrentNanoTime() {
+        return System.nanoTime();
+    }
+
+    /**
+     * Refreshes and returns the value asynchronously. Use the return value like with getValue() to get the refreshed
+     * value.
+     *
+     * @param requestNewValueFuture This supplier is called to supply the cache with a future
+     *            that on completion will update the cached value. The supplier will not be used,
+     *            if there is already an ongoing refresh.
+     * @return the new value in form of a CompletableFuture.
+     */
+    public synchronized @NonNull CompletableFuture<V> refreshValue(
+            @NonNull Supplier<CompletableFuture<V>> requestNewValueFuture) {
+        expiresAt = 0;
+        // There is already an ongoing refresh, just return that future
+        if (currentNewValueRequest != null) {
+            return currentNewValueRequest;
+        }
+        // We request a value update from the supplier now
+        currentNewValueRequest = requestNewValueFuture.get();
+        if (currentNewValueRequest == null) {
+            throw new IllegalArgumentException("We expect a CompletableFuture for refreshValue() to work!");
+        }
+        CompletableFuture<V> t = currentNewValueRequest.thenApply(newValue -> {
+            // No request is ongoing anymore, update the value and expire time
+            currentNewValueRequest = null;
+            value = newValue;
+            expiresAt = getCurrentNanoTime() + expiry;
+            return value;
+        });
+        // The @NonNull annotation forces us to check the return value of thenApply.
+        if (t == null) {
+            throw new IllegalArgumentException("We expect a CompletableFuture for refreshValue() to work!");
+        }
+        return t;
+    }
+
+    /**
+     * Checks if the value is expired.
+     *
+     * @return true if the value is expired
+     */
+    public boolean isExpired() {
+        return expiresAt < getCurrentNanoTime();
+    }
+
+    /**
+     * Return the raw value, no matter if it is already
+     * expired or still valid.
+     */
+    @Nullable
+    public V getLastKnownValue() {
+        return value;
+    }
+}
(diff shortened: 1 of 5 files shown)


bundles/core/org.eclipse.smarthome.core/src/main/java/org/eclipse/smarthome/core/cache/ExpiringCacheAsync.java
```diff
@@ -0,0 +1,137 @@
...
+     */
+    public boolean isExpired() {
+        return expiresAt < getCurrentNanoTime();
+    }
+
+    /**
+     * Return the raw value, no matter if it is already
+     * expired or still valid.
+     */
+    public V getExpiredValue() {
```
maggu2810:
```
getExpiredValue: no matter if it is already expired or still valid
```

Why not using a name like `getLastKnownValue`

bundles/core/org.eclipse.smarthome.core/src/main/java/org/eclipse/smarthome/core/cache/ExpiringCacheAsync.java
```diff
@@ -0,0 +1,137 @@
...
+            callback.accept(value);
+        }
+        waitingCacheCallbacks.clear();
+    }
+
+    /**
+     * Returns an arbitrary time reference in nanoseconds.
+     * This is used for the cache to determine if a value has expired.
+     */
+    public long getCurrentNanoTime() {
```
maggu2810: I don't see any reason to encapsulate the one function call with another one.

davidgraeff: Because of tests :) This way I can intercept the call and inject a time and don't need to "wait" for expire tests.

bundles/core/org.eclipse.smarthome.core/src/main/java/org/eclipse/smarthome/core/cache/ExpiringCacheAsync.java
davidgraeff: Yup, I can do that. It's a matter of taste I guess. I wanted to make it very obvious that the value might be expired.

kaikreuzer:
Sounds like a nice feature. Two questions:
1. Is there any binding in ESH where you think this would be immediately useful? If so, include the refactoring in this PR.
2. To make all the cache features better know, it would be cool if they were mentioned in the developer documentation. A good place might be [this new page](https://github.com/eclipse/smarthome/pull/3930/files#diff-5aaeecad42648308dcaa38da862c2c47), once that referenced PR is merged.

davidgraeff:
Not in ESH but in Openhab, the network binding.

davidgraeff:
@kaikreuzer I have added the documentation part. I'm a non native English speaker, a review of a local would be highly appreciated.

bundles/core/org.eclipse.smarthome.core/src/main/java/org/eclipse/smarthome/core/cache/ExpiringCacheAsync.java
maggu2810: I still thing that this name should be changed :wink:

davidgraeff:
Should an async cache provide a CompletableFuture interface next to the callback interface?

bundles/core/org.eclipse.smarthome.core/src/main/java/org/eclipse/smarthome/core/cache/ExpiringCacheAsync.java
kaikreuzer: @davidgraeff Would you do @maggu2810 this favor?

docs/documentation/features/frameworkUtilities.md
```diff
@@ -11,4 +11,54 @@ In this chapter useful services/utilities of the Eclipse SmartHome project are d
...
+If we refreshed a value of the internal state in a `ThingHandler` just recently, we can return it immediately via the usual `updateState(channel, state)` method in response to a "RefreshType" command.
+If the state is too old, we need to fetch it first and this may involve network calls, interprocess operations or anything else that will would block for a considerable amout of time.
+
+A common usage case of the `ExpiringCacheAsync` cache type is in a `ThingHandler` to encapsulate one value of an internal state and attach an expire time on that value.
+Consumers of the cache request the value via a callback.
+A cache action will be called if the value is expired, but does not expect the called operation to return a value immediately.
+
+
+A **handleCommand** implementation with the interesting *RefreshType* could look like this:
+```
```
kaikreuzer: please add an empty line before and after any code block

kaikreuzer:
> Should an async cache provide a CompletableFuture interface next to the callback interface?

Could probably make sense as this is the normal Java mechanism for such async operations.
Do we need it next to the callback or could it even replace the callback interface completely?

davidgraeff:
Thanks to Java8 Lambdas, callbacks are easily usable and their usage results in readable, short code. For most use cases this is a good option, whereas CompletableFuture supports chaining and more complex scenarios. I'm not sure, I don't see many or any places in ESH that use CompletableFuture, that's why I have choosen callbacks, I guess.

kaikreuzer:
>  I don't see many or any places in ESH that use CompletableFuture

This is mainly due to the fact that until recently we had to be Java7 compliant with the code.
So to be honest, I don't have that much experience with CompletableFutures myself so far, so I am not a good person to judge where it makes sense and where it doesn't. So I fully trust you that your suggestions are sensible. :-)

davidgraeff:
@kaikreuzer @maggu2810 I have changed the interface to use CompletableFuture in a second commit. If we agree on this instead of the callback one, I squash commits. The documentation is updated as well, which shows how to use the class.

docs/documentation/features/frameworkUtilities.md
```diff
@@ -11,4 +11,66 @@ In this chapter useful services/utilities of the Eclipse SmartHome project are d
...
+            ...
+        }
+    }
+}
+```
+
+The interesting part is the `updater`. If the value is not yet expired, the returned CompletableFuture will complete immediately and the given code is executed.
+If the value is expired, the updater will be used to request a refreshed value.
+
+An updater can be any class that implements `Supplier<CompletableFuture<VALUE_TYPE>>`. The method `CompletableFuture<VALUE_TYPE> get()` needs to be implemented, for example like this:
```
kaikreuzer: Supplier is a functional interface, so wouldn't it be much easier to use a lambda here instead of having to implement a class?

docs/documentation/features/frameworkUtilities.md
```diff
@@ -11,4 +11,66 @@ In this chapter useful services/utilities of the Eclipse SmartHome project are d
...
+    @Override
+    CompletableFuture<double> get() {
+       if (c != null) {
+          c = new CompletableFuture<double>();
+          doSuperImportantAsyncStuffHereToGetRefreshedValue();
+       }
+       return c;
+    }
+    
+    @Override
```
kaikreuzer: What is this overriding?

davidgraeff: This is an example, that maps to a real case where someone would implement an interface to get notified of the async refreshed value. The example is not meant to be copy pasted and to work out of the box, that would require a lot of boiler code.

_Sent from my Lge LG-F700K using [FastHub](https://play.google.com/store/apps/details?id=com.fastaccess.github)_

docs/documentation/features/frameworkUtilities.md
davidgraeff: I can't follow you. How would such an example look like? Here the class is responsible for fetching a value, get is async, and notify the future.

_Sent from my Lge LG-F700K using [FastHub](https://play.google.com/store/apps/details?id=com.fastaccess.github)_

kaikreuzer: Ok, for encapsulating the logic in a class, it makes sense.
I was merely looking at the unit tests as an example on how to use it and that could be written as
```
        CompletableFuture<Double> v = new CompletableFuture<Double>();

        // Define a supplier which returns a future that is not yet completed
        Supplier<CompletableFuture<Double>> s = spy(() -> {
            return v;
        });
```

docs/documentation/features/frameworkUtilities.md
kaikreuzer: Right, but then the class should also implement some further custom interface to make clear that this method here is defined in some specific interface (currently, the code suggests that it is a method defined in `Supplier`).

davidgraeff: True, I'll add it

davidgraeff: done

davidgraeff:
Travis needs to rerun

kaikreuzer:
I reran it, but I guess that won't help fixing the two test failures:

```
Tests in error: 
  ExpiringCacheAsyncTest.testFetchValue:39 NullPointer
  ExpiringCacheAsyncTest.testMutipleGetsWhileRefetching:82 Mockito 
```

davidgraeff:
Tests fixed

bundles/core/org.eclipse.smarthome.core/src/main/java/org/eclipse/smarthome/core/cache/ExpiringCacheAsync.java
```diff
@@ -0,0 +1,126 @@
...
+ *
+ * @param <V> the type of the cached value
+ */
+public class ExpiringCacheAsync<V> {
+    // The fields are package private on purpose, to allow easier unit testing.
+    final long expiry;
+    long expiresAt = 0;
+    CompletableFuture<V> currentNewValueRequest = null;
+    @Nullable
+    V value;
```
maggu2810: IIRC this line break is added by the Eclipse IDE because of the missing visibility keyword.
IMHO a type annotation should be in front of the type and not a line above (same for interfaces where we omit the redundant public keyword on methods).

Perhaps the IDE settings will support this some time... So, not your fault.

bundles/core/org.eclipse.smarthome.core/src/main/java/org/eclipse/smarthome/core/cache/ExpiringCacheAsync.java
```diff
@@ -0,0 +1,126 @@
...
+    /**
+     * Create a new instance.
+     *
+     * @param expiry the duration in milliseconds for how long the value stays valid. Must be greater than 0.
+     * @throws IllegalArgumentException For an expire value <=0.
+     */
+    public ExpiringCacheAsync(long expiry) throws IllegalArgumentException {
+        this.expiry = TimeUnit.MILLISECONDS.toNanos(expiry);
+        if (expiry <= 0) {
+            throw new IllegalArgumentException("Cache expire time must be greater than 0");
```
maggu2810: Wouldn't it make sense to check the argument in front of using it?

docs/documentation/features/frameworkUtilities.md
```diff
@@ -11,4 +11,67 @@ In this chapter useful services/utilities of the Eclipse SmartHome project are d
...
+## Caching
+
+The framework provides some caching solutions for common scenarios.
+
+### Simple expiring and reloading cache
+
+A common usage case is in a `ThingHandler` to encapsulate one value of an internal state and attach an expire time on that value. A cache action will be called to refresh the value if it is expired. This is what `ExpiringCache` implements. If `handleCommand(ChannelUID channelUID, Command command)` is called with the "RefreshType" command, you just return `cache.getValue()`. 
+
+It is a good practice to return as fast as possible from the `handleCommand(ChannelUID channelUID, Command command)` method to not block callers especially UIs.
+Use this type of cache only, if your refresh action is a quick to compute, blocking operation. If you deal with network calls, consider the next cache implementation instead.
```
maggu2810: the "next" cache implementation

Perhaps you could state "the asynchronously cache implementation". Just to not depend on the order.

docs/documentation/features/frameworkUtilities.md
```diff
@@ -11,4 +11,67 @@ In this chapter useful services/utilities of the Eclipse SmartHome project are d
...
+            ...
+        }
+    }
+}
+```
+
+The interesting part is the `updater`. If the value is not yet expired, the returned CompletableFuture will complete immediately and the given code is executed.
+If the value is expired, the updater will be used to request a refreshed value.
+
+An updater can be any class that implements `Supplier<CompletableFuture<VALUE_TYPE>>`. The method `CompletableFuture<VALUE_TYPE> get()` needs to be implemented, for example like this:
```
maggu2810: I don't think we should introduce such long names for generic type parameters and follow the common guidelines.

https://docs.oracle.com/javase/tutorial/java/generics/types.html

By convention, type parameter names are single, uppercase letters. This stands in sharp contrast to the variable naming conventions that you already know about, and with good reason: Without this convention, it would be difficult to tell the difference between a type variable and an ordinary class or interface name.

davidgraeff: I disagree. For documentation i prefer long parameter names, so that the reader immediately get what I'm talking about. A single "V" would need additional explanation. The code uses "V" by the way.

bundles/core/org.eclipse.smarthome.core/src/main/java/org/eclipse/smarthome/core/cache/ExpiringCacheAsync.java
davidgraeff: I was surprised as well about this format setting, but I didn't fight with Eclipse here and just accepted it. I might just declare the fields protected to correct this formatting.

davidgraeff:
I have amended some text to the documentation to explain the example in more detail

davidgraeff:
rebased

maggu2810:
I retriggered the build.

davidgraeff:
Hm, fails with:
> [ERROR] org.eclipse.smarthome.core.cache.ExpiringCacheAsync.java:[3]
Line does not match expected header line of '^ \* All rights reserved\. This program and the accompanying materials$'.

davidgraeff:
Why is there no newline in the template. That's not pleasing to read and doesn't look very common. But I'll change it to match the template.

PR #3938, opened 2017-08-02
[Astro] Apply channel config before scheduling range events

### LINKED ISSUE #3882: [astro] Event order wrong when using ealiest/latest
In case the earliest/latest limits enforce events to be triggered at the same point in time, the order is not guaranteed.

See https://github.com/eclipse/smarthome/issues/3728#issuecomment-317259064

triller-telekom:
Fixes #3882

### DIFF
--- a/extensions/binding/org.eclipse.smarthome.binding.astro/src/main/java/org/eclipse/smarthome/binding/astro/internal/job/Job.java
+++ b/extensions/binding/org.eclipse.smarthome.binding.astro/src/main/java/org/eclipse/smarthome/binding/astro/internal/job/Job.java
@@ -70,8 +70,8 @@ public interface Job extends Runnable {
      * @param channelId the channel ID
      */
     public static void scheduleEvent(String thingUID, AstroThingHandler astroHandler, Calendar eventAt, String event,
-            String channelId) {
-        scheduleEvent(thingUID, astroHandler, eventAt, singletonList(event), channelId);
+            String channelId, boolean configAlreadyApplied) {
+        scheduleEvent(thingUID, astroHandler, eventAt, singletonList(event), channelId, configAlreadyApplied);
     }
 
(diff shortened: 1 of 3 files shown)


extensions/binding/org.eclipse.smarthome.binding.astro/src/main/java/org/eclipse/smarthome/binding/astro/internal/job/Job.java
```diff
@@ -94,9 +94,12 @@ public static void scheduleEvent(String thingUID, AstroThingHandler astroHandler
...
         }
-        AstroChannelConfig config = astroHandler.getThing().getChannel(channelId).getConfiguration()
-                .as(AstroChannelConfig.class);
-        Calendar instant = applyConfig(eventAt, config);
+        Calendar instant = eventAt;
+        if (!configAlreadyApplied) {
+            AstroChannelConfig config = astroHandler.getThing().getChannel(channelId).getConfiguration()
+                    .as(AstroChannelConfig.class);
+            instant = applyConfig(eventAt, config);
+        }
```
maggu2810: Any reason to initialize instant twice if `configAlreadyApplied == false`?

What about:
```Java
final Calendar instant;
if (!configAlreadyApplied) {
    AstroChannelConfig config = astroHandler.getThing().getChannel(channelId).getConfiguration()
            .as(AstroChannelConfig.class);
    instant = applyConfig(eventAt, config);
} else {
    instant = eventAt;
}
```

triller-telekom: I was saving the "else" branch. But you are right, maybe its better to have the variable `final`. I have changed it now.

PR #3939, opened 2017-08-02
removed getKey() method from AbstractManagedProvider

kaikreuzer:
This is superfluous since we now have identifiable entities.


kaikreuzer:
Yes, I guess so - although hardly anyone should have implemented his own registry and the getKey() method itself wasn't public, so the impact should be very limited.

PR #3950, opened 2017-08-03
Local server for NTP binding tests

### LINKED ISSUE #3231: [Test Failure] NtpOSGiTest fails randomly on my local machine during mvn clean install
Sometimes the test runs fine and sometimes it fails, I cannot really specify when. But it just happend twice after running "mvn clean install" -> reboot laptop -> "mvn clean install".

I am running Arch Linux on my machine.

testDateTimeChannelWithUnknownHost(org.eclipse.smarthome.binding.ntp.test.NtpOSGiTest)  Time elapsed: 10.182 sec  <<< FAILURE!
java.lang.AssertionError: The thing status was not communication error
Expected: is <COMMUNICATION_ERROR>
     but: was <NONE>
        at org.eclipse.smarthome.binding.ntp.test.NtpOSGiTest.lambda$4(NtpOSGiTest.java:489)

doandzhi:
A simple local NTP server which removes the dependence on external systems for the tests of NTP binding. There is an issue regarding this [here](https://github.com/eclipse/smarthome/issues/3231#issuecomment-295646370). Fixes #3231.


extensions/binding/org.eclipse.smarthome.binding.ntp.test/NtpOSGiTest.launch
```diff
@@ -34,9 +34,9 @@
...
...t:default,org.eclipse.equinox.region@default:false,org.eclipse.equinox.registry@default:default,org.eclipse.equinox.transforms.hook@default:false,org.eclipse.equinox.util@default:default,org.eclipse.equinox.weaving.hook@default:false,org.eclipse.help@default:default,org.eclipse.jface.databinding@default:default,org.eclipse.jface@default:default,org.eclipse.osgi.compatibility.state@default:false,org.eclipse.osgi.services@default:default,org.eclipse.osgi@-1:true,org.eclipse.swt.win32.win32.x86_64@default:false,org.eclipse.swt@default:default,org.eclipse.team.core@default:default,org.eclipse.ui.trace@default:default,org.eclipse.ui.workbench@default:default,org.eclipse.ui@default:default,org.hamcrest.core@default:default,org.hamcrest.integration@default:default,org.hamcrest.library@default:default,org.hamcrest.text@default:default,org.hamcrest@default:default,org.jsr-305@default:default,org.junit@default:default,org.jupnp@default:default,org.mockito@default:default,org.objenesis@default:default,org.slf4j.api@default:default,org.slf4j.jcl@default:default,org.w3c.css.sac@default:default,org.w3c.dom.events@default:default,org.w3c.dom.smil@default:default,org.w3c.dom.svg@default:default"/>
```
triller-telekom: This almost doubles the amount of plugins which are started for the test. Are you 100% you need all of them? What about starting from the old config and adding only those which you need additionally?

extensions/binding/org.eclipse.smarthome.binding.ntp.test/NtpOSGiTest.launch
```diff
@@ -34,9 +34,9 @@
...
+<stringAttribute key="selected_workspace_plugins" value="org.eclipse.smarthome.binding.ntp.test@default:false,org.eclipse.smarthome.binding.ntp@default:default,org.eclipse.smarthome.config.core.test@default:false,org.eclipse.smarthome.config.core@default:default,org.eclipse.smarthome.config.discovery.test@default:false,org.eclipse.smarthome.config.discovery@default:default,org.eclipse.smarthome.config.xml.test@default:false,org.eclipse.smarthome.config.xml@default:default,org.eclipse.smarthome.core.autoupdate@default:default,org.eclipse.smarthome.core.test@default:false,org.eclipse.smarthome.core.thing.test@default:false,org.eclipse.smarthome.core.thing.xml.test@default:default,org.eclipse.smarthome.core.thing.xml@default:default,org.eclipse.smarthome.core.thing@default:default,org.eclipse.smarthome.core@default:default,org.eclipse.smarthome.io.console@default:default,org.eclipse.smarthome.test@default:default"/>
```
triller-telekom: Same as my comment above. I hardly doubt that you need all of them here.

extensions/binding/org.eclipse.smarthome.binding.ntp.test/src/test/java/org/eclipse/smarthome/binding/ntp/server/SimpleNTPServer.java
```diff
@@ -0,0 +1,100 @@
...
+                try {
+                    socket = new DatagramSocket(port);
+                } catch (SocketException e) {
+                    logger.error("Occured an error {}. Couldn't open a socket on this port:" + port, e);
+                }
+                while (isRunning) {
+                    try {
+                        socket.receive(request);
+                        handleRequest(request);
+                    } catch (IOException e) {
```
triller-telekom: What if your array with 48 bytes is too small for the request? Don't you get an `ArrayIndexOutOfBoundException` here?

extensions/binding/org.eclipse.smarthome.binding.ntp.test/src/test/java/org/eclipse/smarthome/binding/ntp/server/SimpleNTPServer.java
```diff
@@ -0,0 +1,100 @@
...
+    private int port;
+    private volatile boolean isRunning;
+    private byte[] array = new byte[48];
+    private final DatagramPacket request = new DatagramPacket(array, array.length);
+    private Logger logger = LoggerFactory.getLogger(SimpleNTPServer.class);
+
+    /**
+     * The server must use an available port to be able to start.
+     * According to RFC 793, the port is a 16 bit unsigned int.
+     * This means the range is 0 - 65535.
```
triller-telekom: range is 1 - 65535

extensions/binding/org.eclipse.smarthome.binding.ntp.test/src/test/java/org/eclipse/smarthome/binding/ntp/test/NtpOSGiTest.java
```diff
@@ -8,7 +8,8 @@
 package org.eclipse.smarthome.binding.ntp.test;
 
 import static org.hamcrest.CoreMatchers.*;
-import static org.junit.Assert.*;
+import static org.junit.Assert.assertThat;
+import static org.junit.Assert.fail;
```
triller-telekom: why did you split the wildcard `org.junit.Assert.*` into 2 specific ones?

extensions/binding/org.eclipse.smarthome.binding.ntp.test/src/test/java/org/eclipse/smarthome/binding/ntp/test/NtpOSGiTest.java
```diff
@@ -159,7 +171,8 @@ public void testStringChannelTimeZoneUpdate() {
 
         Configuration configuration = new Configuration();
         configuration.put(NtpBindingConstants.PROPERTY_TIMEZONE, TEST_TIME_ZONE_ID);
-
+        configuration.put(NtpBindingConstants.PROPERTY_NTP_SERVER_HOST, TEST_HOSTNAME);
+        configuration.put(NtpBindingConstants.PROPERTY_NTP_SERVER_PORT, TEST_PORT);
```
triller-telekom: If you have to pass these 2 properties to the `initialize` method in every test method, how about setting them **inside** the `initialize` method and remove the duplicate code from the test methods? Because the server and port won't change for any testmethod, right?

extensions/binding/org.eclipse.smarthome.binding.ntp/ESH-INF/thing/thing-types.xml
```diff
@@ -31,6 +32,11 @@
 				</description>
 				<default>30</default>
 			</parameter>
+			<parameter name="serverPort" type="integer" required="true">
```
triller-telekom: If you have a default value you do not need to specify `required=true` here.

extensions/binding/org.eclipse.smarthome.binding.ntp.test/src/test/java/org/eclipse/smarthome/binding/ntp/server/SimpleNTPServer.java
doandzhi: Actually I receive an empty buffer and later I only use the port and address to send the response. I choose to make it 48 just in case. (the size of an IPv6 header is at least 40 bytes, the size of a UDP header is 8 bytes.)

doandzhi:
Hello, I addressed your comments but the Travis build fails randomly. I suppose that there is an internal problem. :)

triller-telekom:
Travis fails are fixed now and indeed it was not your fault. :)

Thank you for addressing my comments, but what about the amount of plugins that you added in this run configuration `NtpOSGiTest.launch`? I am very sure you do not need that much.

doandzhi:
Ok, I agree that this is not very elegant but [these](https://user-images.githubusercontent.com/17512690/29069787-11fc0bbe-7c45-11e7-9e2b-2a0c144336e2.PNG) were required in order to work properly.

maggu2810:
@triller-telekom Have all your requests been fixed?

triller-telekom:
@maggu2810 If you ignore the launch configuration since it's not that critical, then yes :)

PR #3960, opened 2017-08-05
WU binding: automatic discovery of local weather

### LINKED ISSUE #3953: WU binding: auto discovery with locale location
I will implement this feature.

lolodomo:
Use DS annotation for service definition

Fix #3953 


extensions/binding/org.eclipse.smarthome.binding.weatherunderground/src/main/java/org/eclipse/smarthome/binding/weatherunderground/discovery/WeatherUndergroundDiscoveryService.java
```diff
@@ -0,0 +1,110 @@
...
+
+    @Override
+    protected void startBackgroundDiscovery() {
+        if (discoveryJob == null) {
+            discoveryJob = scheduler.scheduleWithFixedDelay(() -> {
+                PointType currentLocation = locationProvider.getLocation();
+                if (!Objects.equals(currentLocation, previousLocation)) {
+                    logger.debug("Location has been changed from {} to {}: Creating new discovery results",
+                            previousLocation, currentLocation);
+                    createResults(currentLocation);
```
htreu: In case the location changes this will also reset the API key to `ToBeDefined`.

extensions/binding/org.eclipse.smarthome.binding.weatherunderground/src/main/java/org/eclipse/smarthome/binding/weatherunderground/discovery/WeatherUndergroundDiscoveryService.java
```diff
@@ -0,0 +1,110 @@
...
+                discoveryJob = null;
+                logger.debug("Stopped Weather Underground device background discovery");
+            }
+        }
+    }
+
+    public void createResults(PointType location) {
+        ThingUID localWeatherThing = new ThingUID(THING_TYPE_WEATHER, LOCAL);
+        Map<String, Object> properties = new HashMap<>(3);
+        properties.put(APIKEY, "ToBeDefined");
```
htreu: This should only be done initially if the thing does not yet exist or if the API key is not entered yet.

extensions/binding/org.eclipse.smarthome.binding.weatherunderground/src/main/java/org/eclipse/smarthome/binding/weatherunderground/discovery/WeatherUndergroundDiscoveryService.java
lolodomo: Only the discovery result is updated, not the thing irself. So I don't see any problem here.

extensions/binding/org.eclipse.smarthome.binding.weatherunderground/src/main/java/org/eclipse/smarthome/binding/weatherunderground/discovery/WeatherUndergroundDiscoveryService.java
lolodomo: If the rhing is already created, this discovery will be ignored by the system.

htreu: I did start my ESH with your changes and could see this behaviour. After changing my system location and waiting for the autodiscovery to pick up the changes the thing goes offline with the error message `API key has to be fixed`.
For the location it should definitely update the thing (which it does) but leave the API key untouched.

lolodomo: I am very surprised of this behaviour but I believe you :)

@kaikreuzer : is it expected or a bug that a new discovery will update a created thing ?

If the behaviour is this one and is expected, I don't jnow whar to do. I know it exists an extended feature of the discovery service that allows knowing if a thing is already existing but how to get this thing and get its configuration settings ? Should I link the thing registry to my discovery service to retrieve my thing ??

lolodomo: Maybe it can work if I provide only the location property when the thing is already existing.

lolodomo: @htreu @kaikreuzer I don't find in the code where would be done the thing update when a thing is discovered ?

lolodomo: @htreu : I was able to reproduce your case. I found the code that is doing the thing update. I created the issue #3975 because I think it is not the expected behaviour.
Until we have a discussion and conclusion about discovery and thing update, I put this PR in stand-by.

htreu:
Hi @lolodomo with this (https://github.com/eclipse/smarthome/issues/3975#event-1196793335) out of the way here is my suggestion for this PR:
- for the WeatherUnderground things it would be nice to have a bridge which holds the API token
- all weather information things will be configured within this bridge and will only hold their specific location. 
- the autodiscovery can then start as soon as the bridge is present.
- the auto update feature will then exactly do what it was designed for and update the location for the "local weather" thing.

wdyt?

extensions/binding/org.eclipse.smarthome.binding.weatherunderground/src/main/java/org/eclipse/smarthome/binding/weatherunderground/discovery/WeatherUndergroundDiscoveryService.java
lolodomo: In fact, I thing the discovery service should not set the API key property.
I previously added this property because without it, the thing handler was not started.

lolodomo:
@htreu : this option was abandonned but I don't remember why now ;)

I posted a message just before. In fact, I think the discovery service should just not update de API key property. I will first try this option.

lolodomo:
It is now fixed, my solution is working.
In fact, there was no valid reason for the discovery service to set the API key property.

triller-telekom:
Seeing that every discovery result is "useless" for the user until he adds the correct API key, I think we should change the binding as @htreu mentioned. A user could then add a bridge with a correct API key and a discovery would produce the correct result.

Also if the user adds an additional location he does not need to add the API key there again. Isn't this the better option? WDYT?

kaikreuzer:
I agree, it should only be necessary at a single place to provide the API key - this should be a "bridge", which provides the access to WU. New things for different locations can then be added very easily without bothering about authentication. Note that we do the same in other bindings that require a web account (e.g. the Tankerkönig binding).

lolodomo:
Ok guys, your enhancement request is acceptable but should be the subject of another issue.
In my initial thoughts that were probably shared with everybody in the "weather binding" issue, I think that I considered not usefull to have systematically 2 different things to create when 95% of users will use only one location.

Until this change is implemented by someone, my current PR (being useful and working well) could be merged ?

lolodomo:
And by the way, the future introduction of a bridge thing should probably change nothing to the code of my current feature as the discovery service is not dealing wih any API key.

htreu:
From a coding pov I agree to your arguments. The change in thing structure is not related to the auto discovery or auto update you implemented. If we agree on having a bridge thing is a better way of modelling the WU binding we only need a volunteer to implement the new structure ;-)
I guess its better to have it sooner then later.
From my pov this can be merged and the structure change is a different story.

extensions/binding/org.eclipse.smarthome.binding.weatherunderground/OSGI-INF/WeatherUndergroundDiscoveryService.xml
```diff
@@ -0,0 +1,17 @@
+<?xml version="1.0" encoding="UTF-8"?>
```
kaikreuzer: Shouldn't we go for annotated DS now?
@htreu You had updated the archetype accordingly, but we haven't mentioned anything in the coding guidelines yet. Time for it, do you agree?

extensions/binding/org.eclipse.smarthome.binding.weatherunderground/src/main/java/org/eclipse/smarthome/binding/weatherunderground/discovery/WeatherUndergroundDiscoveryService.java
```diff
@@ -0,0 +1,109 @@
+/**
+ * Copyright (c) 2014-2017 by the respective copyright holders.
+ * All rights reserved. This program and the accompanying materials
+ * are made available under the terms of the Eclipse Public License v1.0
+ * which accompanies this distribution, and is available at
+ * http://www.eclipse.org/legal/epl-v10.html
+ */
+package org.eclipse.smarthome.binding.weatherunderground.discovery;
```
kaikreuzer: Please change to `org.eclipse.smarthome.binding.weatherunderground.internal.discovery` as you are not exporting this package.

extensions/binding/org.eclipse.smarthome.binding.weatherunderground/src/main/java/org/eclipse/smarthome/binding/weatherunderground/handler/WeatherUndergroundHandler.java
```diff
@@ -298,6 +298,9 @@ private WeatherUndergroundJsonData getWeatherData(Set<String> features, String l
             } else if (result.getResponse() == null) {
                 errorDetail = "missing response sub-object";
             } else if (result.getResponse().getErrorDescription() != null) {
+                if ("keynotfound".equals(result.getResponse().getErrorType())) {
```
kaikreuzer: How is this related to this feature?
Note that if the api key is a mandatory config parameter, the framework will make sure that the handler isn't even started, if the key is not set. So I would claim that this is not related, right?

extensions/binding/org.eclipse.smarthome.binding.weatherunderground/OSGI-INF/WeatherUndergroundDiscoveryService.xml
htreu: I agree but would prefer to delay this until I´m back from vacation.

extensions/binding/org.eclipse.smarthome.binding.weatherunderground/src/main/java/org/eclipse/smarthome/binding/weatherunderground/internal/config/WeatherUndergroundConfiguration.java
```diff
@@ -15,6 +15,9 @@
  */
 public class WeatherUndergroundConfiguration {
 
+    public static final String APIKEY = "apikey";
```
kaikreuzer: This constant does not seem to be used (anymore).

extensions/binding/org.eclipse.smarthome.binding.weatherunderground/OSGI-INF/WeatherUndergroundDiscoveryService.xml
kaikreuzer: Delay changing this to annotations or delay adding the info to the documentation?

htreu: I was referring to delaying the docs. anyway, here you go :-) https://github.com/eclipse/smarthome/pull/4002

htreu: changing to annotations is up to @lolodomo I guess.

extensions/binding/org.eclipse.smarthome.binding.weatherunderground/src/main/java/org/eclipse/smarthome/binding/weatherunderground/handler/WeatherUndergroundHandler.java
lolodomo: It is to have a clear message about wrong API key displayed in Paper UI (and something that can easily be translated). Consider that as an additional improvement of the binding.

extensions/binding/org.eclipse.smarthome.binding.weatherunderground/src/main/java/org/eclipse/smarthome/binding/weatherunderground/internal/config/WeatherUndergroundConfiguration.java
lolodomo: It was at a certain of this PR. I kept it thinking it could be used one day later. But I can suppress it.

extensions/binding/org.eclipse.smarthome.binding.weatherunderground/src/main/java/org/eclipse/smarthome/binding/weatherunderground/handler/WeatherUndergroundHandler.java
kaikreuzer: ok :-)

extensions/binding/org.eclipse.smarthome.binding.weatherunderground/src/main/java/org/eclipse/smarthome/binding/weatherunderground/discovery/WeatherUndergroundDiscoveryService.java
lolodomo: Will do it.

extensions/binding/org.eclipse.smarthome.binding.weatherunderground/OSGI-INF/WeatherUndergroundDiscoveryService.xml
kaikreuzer: > I am not refactoring the binding lol just providing a new simple feature.

Yes, and new features should use the current guidelines. No worries, you will like it as everything becomes much easier and less error-prone! See https://github.com/eclipse/smarthome/pull/4002.

lolodomo:
All review comments taken into account.

I can confirm that DS annotation is cool to declare service ;)

kaikreuzer:
Great, thanks, looks all good!
Just one small last details (which hasn't yet communicated): The XMLs are generated by the IDE as well as by the Maven build, so there is no need to have them in the repo - we therefore add a .gitignore entry for them. Could you do that as well here? Thanks!

lolodomo:
Done.

Issue #3093, opened 2017-03-01
Display localized description for a thing status info


tomhoefer:
## Desired impact in front of user

In our project we have to display a localized description for a thing status, e.g »Device is blocked by remote service for 60 minutes. Maximum limit of 10 configuration changes per hour has been exceeded. For further info please refer to the device vendor. « in case of device status ```ONLINE:CONFIGURATION_PENDING```.

## Problem

The ESH device status infrastructure  is not able to handle localized thing status description strings.

### Solution (Binding)

The corresponding binding does not provide a localized string in the ```ThingStatus.description``` property, but instead provides the reference of the localization string, e.g. ```@text/rate_limit```.

The binding is able to provide placeholder values as a JSON-serialized array of strings:

```
@text/rate_limit ["60","10"]
```

```
rate_limit=Device is blocked by remote service for {0} minutes. Maximum limit of {1} configuration changes per hour has been exceeded. For further info please refer to device vendor
```
The corresponding thing should provide a new property ```localizedStatus```, so the UI can avoid requesting the REST services for things that don't support translated status messages

### Solution (ESH)

The existing REST endpoints ```/rest/things``` and ```/rest/things/<ThingUID>``` as well as a new endpoint ```/rest/things/<ThingUID>/status``` resolve the ```ThingStatus.description``` property using the I18N infrastructure already in place in ESH.

If the string in ```ThingStatus.description``` is not a localization-message-id (begins with ```@text/```) the string is to be returned verbatim.

For events and REST requests that don't contain a preferred language the localization-message must be translated to the ESH default locale (provided by locale provider).

kaikreuzer:
Sounds feasible to me in general with one exception:

> The corresponding thing should provide a new property localizedStatus

The properties describe the Thing (device), not the features of a binding. I would therefore abstain from misusing the properties for such an information.

As a result, I would suggest that the information in the event simply has to be considered to be in default locale in any case and if a UI expects a localized version, it has to do the REST request in any case. (alternatively, your solution can keep this info somewhere else and decide based on the Thing type / binding ID whether it wants to display a localized value or not).

P.S.: `ONLINE:CONFIGURATION_PENDING` - really? If your access is blocked for 60 minutes and thus you cannot access the device, I would very much expect an OFFLINE status.

tomhoefer:
> The properties describe the Thing (device), not the features of a binding. I would therefore abstain from misusing the properties for such an information.
I agree. I will think about to a different solution

[...]

PR #3961, opened 2017-08-05
NTP binding: fix issue #3962 and add French translation

lolodomo:
Also add the unit to a configuration setting

### DIFF
--- a/extensions/binding/org.eclipse.smarthome.binding.ntp/ESH-INF/i18n/ntp_fr.properties
+++ b/extensions/binding/org.eclipse.smarthome.binding.ntp/ESH-INF/i18n/ntp_fr.properties
@@ -1,0 +1,31 @@
+# binding
+binding.ntp.name = Extension NTP
+binding.ntp.description = L'extension NTP interroge le serveur de temps configuré et fournit la date et l'heure actuelles.
+
+# thing types
+thing-type.ntp.ntp.label = Serveur NTP
+thing-type.ntp.ntp.description = Un serveur de temps fournissant la date et l'heure actuelles.
+
+# thing type configuration
+thing-type.config.ntp.ntp.hostname.label = Nom d'hôte du serveur
+thing-type.config.ntp.ntp.hostname.description = Le nom d'hôte du serveur de temps.
+thing-type.config.ntp.ntp.refreshInterval.label = Fréquence de rafraîchissement
+thing-type.config.ntp.ntp.refreshInterval.description = La fréquence en secondes de mise à jour de l'heure.
+thing-type.config.ntp.ntp.refreshNtp.label = Fréquence de rafraîchissement NTP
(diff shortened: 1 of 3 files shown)


lolodomo:
@kaikreuzer @cweitkamp : for good practice in binding internationalization, is `updateStatus` the only method in which you can use `"@text/xxx"` ?
Example in the Yahoo Weather binding: https://github.com/eclipse/smarthome/blob/master/extensions/binding/org.eclipse.smarthome.binding.yahooweather/src/main/java/org/eclipse/smarthome/binding/yahooweather/handler/YahooWeatherHandler.java#L188

If not, which are the other methods supporting that feature ?

cweitkamp:
This looks like an easy peasy way to translate `updateStatus` messages. I will update the internationalization documentation shortly and add this one to the [using custom keys](http://www.eclipse.org/smarthome/documentation/features/internationalization.html#using-custom-keys) section.

The other way I am aware of is the `LocaleProvider` / `TranslationProvider` approach you are using in this PR. Later one looks much more complex than the custom key solution.

lolodomo:
@kaikreuzer @maggu2810 @sjka : I would appreciate a review of this PR to be sure my approach with binding internationalization is good, before applying them to other bindings.

kaikreuzer:
Sorry to be late to comment, I feel slightly flooded by the number of PRs this week...

I think this idea of yours in this PR leads a bit too far. The i18n support within ESH is made in a way that the same system can answer stuff in different languages (mainly for the REST API). It is NOT meant for fully translating all aspects of the system to another language (such as log files etc.).
So the default language for all stuff that is a "technical" artifact (such as log entries, events, etc.) is english.
Regarding the localisation of thing status messages, you might be interested in https://github.com/eclipse/smarthome/issues/3093.

lolodomo:
Ok, I will reduce the scope of this PR and forgive all tranlations of log messages.
I will consider the translation of the thing status message (for NTP binding and other bindings I started to translate in French).
I think the tranlation of the automatic discovered thing name is important too as it is directly seen by the user in Paper UI.

kaikreuzer:
> I think the tranlation of the automatic discovered thing name is important too as it is directly seen by the user in Paper UI.

Our approach here was so far to use official product names, which are in 90% of the cases international and do not require any translation...

lolodomo:
For the discovered thing name, I think we should think of something similar to what is doable with ConfigStatus/ThigStatus so that it can be used easily by any binding. That means the thing name has to be localized. I will study this enhancement.

lolodomo:
@kaikreuzer : I am talking about the label of the thing. For NTP, it is hardcoded as "Local Time". It is this label that I would like to make localized.
I think we have already something very similar for config status so it should not be very difficult to add this feature to the creation of thing result. I will take a look.

lolodomo:
Ready for new review, I finally only include the "basics".
I will include other stuff in another PR later.

Issue #3965, opened 2017-08-05
Hue binding: lightId configuration setting AND property


lolodomo:
`lightId` is the main configuration setting of a hue thing but an additional `lightId` property is created with the same value.
Is there a reason to have this in double ?
Should we suppress the property ?

kaikreuzer:
I would agree that it is not needed as a property as this is always identical to the config param.
@afuechsel, do you agree?

Btw, I also noticed that we have "serialNumber" as a configuration parameter on the hue bridge, which does not seem to make much sense to me - I would think that it is enough to have it as a property. Does anyone see a reason against removing this param?

lolodomo:
I see that too this weekend and did not understand why the serial mumber is a configuration setting.

afuechsel:
Yes, I was wondering, why I could actually SET the serial number when creating a thing manually. +1 for removing it.

lolodomo:
I will propose a fix for light id and serial number.

afuechsel:
#3985 for the other clean up tasks found in the hue binding

lolodomo:
If you want to deal with the serial number configuration setting, you are welcome.

kaikreuzer:
https://github.com/eclipse/smarthome/pull/3985 includes the serial number cleanup already.

lolodomo:
Now fixed

PR #3976, opened 2017-08-07
Hue binding: remove lightId property

lolodomo:
Fix issue #3965

Also add unit to pollingInterval configuration setting

### DIFF
--- a/extensions/binding/org.eclipse.smarthome.binding.hue/src/main/java/org/eclipse/smarthome/binding/hue/handler/HueLightHandler.java
+++ b/extensions/binding/org.eclipse.smarthome.binding.hue/src/main/java/org/eclipse/smarthome/binding/hue/handler/HueLightHandler.java
@@ -133,7 +133,6 @@ public class HueLightHandler extends BaseThingHandler implements LightStatusList
                 if (vendor != null) {
                     updateProperty(Thing.PROPERTY_VENDOR, vendor);
                 }
-                updateProperty(LIGHT_ID, fullLight.getId());
                 updateProperty(LIGHT_UNIQUE_ID, fullLight.getUniqueID());
                 isOsramPar16 = OSRAM_PAR16_50_TW_MODEL_ID.equals(modelId);
                 propertiesInitializedSuccessfully = true;
--- a/extensions/binding/org.eclipse.smarthome.binding.hue/ESH-INF/thing/bridge.xml
+++ b/extensions/binding/org.eclipse.smarthome.binding.hue/ESH-INF/thing/bridge.xml
@@ -29,7 +29,7 @@
                 </description>
                 <required>false</required>
             </parameter>
-            <parameter name="pollingInterval" type="integer" min="1" step="1">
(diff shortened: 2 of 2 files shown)


extensions/binding/org.eclipse.smarthome.binding.hue/src/main/java/org/eclipse/smarthome/binding/hue/handler/HueLightHandler.java
```diff
@@ -133,7 +133,6 @@ private synchronized void initializeProperties() {
                 if (vendor != null) {
                     updateProperty(Thing.PROPERTY_VENDOR, vendor);
                 }
-                updateProperty(LIGHT_ID, fullLight.getId());
```
kaikreuzer: Please note that this change will break https://github.com/eclipse/smarthome/issues/3899 as long as https://github.com/eclipse/smarthome/issues/3863#issuecomment-320099215 isn't yet implemented (correct, @afuechsel?).

afuechsel: No, don't think so, as UNIQUE_ID is used as representation property, not the light id

kaikreuzer: Ah, ok, and the UNIQUE_ID is something like a Mac address of the bulb?

lolodomo: I kept unchanged the set of the LIGHT_UNIQUE_ID property.

lolodomo:
HueBindingConstants in fact defines at the same place the configuration settings and properties.
By the way, I should suppress SERIAL_NUMBER and the binding should use the constant from Thing.

lolodomo:
For the SERIAL_NUMBER, you're right, it seems to be necessary by the bridge handler factory. So I finally propose to keep it !

Regarding the backward compatibily, I have not tested it, I only tested with new things.

So to avoid problems with backward compatibily, I will reduce my PR to only the suppress of the lightId property, which was in fact my initial intention.

kaikreuzer:
>  it seems to be necessary by the bridge handler factory. So I finally propose to keep it !

You are giving up to quickly! I think it should still be removed, we just have to figure out, how to refactor this code.

lolodomo:
I let someone else doing the change with serial number and testing the backward compatibility. I will myself focus on other things.
But I will at least update my PR to cover the lightId property.

kaikreuzer:
ok, fine!

lolodomo:
Updated PR.

kaikreuzer:
Merge conflicts...

lolodomo:
What amount of time just for 2 updated lines ! ;)

afuechsel:
Lgtm

PR #3979, opened 2017-08-07
[Hue] Updated and added german translation

cweitkamp:


lolodomo:
Reading your comments, maybe all channel type options like channel-type.hue.alert.option.NONE should be replaced by channel-type.hue.alert.state.option.NONE ?
If you confirm, can you please update the French file too ?

cweitkamp:
I am going to test it later. If it works I will update this PR and update the French translation file too.

lolodomo:
You can find one important part of the syntax in this class: https://github.com/eclipse/smarthome/blob/246c2298eba0ee54d739f363b8a88af8effe86a5/bundles/core/org.eclipse.smarthome.core.thing/src/main/java/org/eclipse/smarthome/core/thing/i18n/ThingTypeI18nUtil.java#L26
My feeling is that there is nothing for channel group label and channel group description.

lolodomo:
Looking at the code, I can confirm that `state.option.NONE` is required rather than `option.NONE`.

cweitkamp:
@lolodomo I changed it. French file too.

extensions/binding/org.eclipse.smarthome.binding.hue/ESH-INF/i18n/hue_de.properties
```diff
@@ -1 +1,62 @@
...
+channel-type.hue.color_temperature.description = Erm�glicht die Steuerung der Farbtemperatur. Von Tageslichtwei� (0) bis Warmwei� (100).
+channel-type.hue.alert.label = Alarm
+channel-type.hue.alert.description = Erm�glicht ein tempor�res Blinken.
+channel-type.hue.alert.state.option.NONE = Kein Blinken
+channel-type.hue.alert.state.option.SELECT = Einmaliges Blinken
+channel-type.hue.alert.state.option.LSELECT = Mehrfaches Blinken
+channel-type.hue.effect.label = Farbeffekt
+channel-type.hue.effect.description = Erm�glicht einen automatischen Farbwechsels.
+
+config-status.error.missing-ip-address-configuration=Es wurde keine IP Adresse f\u00FCr die Hue Bridge angegeben.
```
triller-telekom: Why don't you use the german Umlaut "ü" here instead of the `\u00FC`? Is there any reason for it?

Except for this minor issue this PR looks fine for me.

kaikreuzer: @cweitkamp If you could change this single character, we are good to merge :-)

cweitkamp: @triller-telekom It wasn't me. Was there before my PR.

@kaikreuzer Done.

PR #3980, opened 2017-08-08
Mqtt: Fix reloading configuration

davidgraeff:
## Need to be done
* At the moment the reconnect strategy closes forever. This is not compatible with the restart-able MqttBrokerConnection object.
* Rename MqttBrokerConnection.close() to stop()

## MqttService:
* Use OSGi DS annotations instead of handwiring the XML file.
* Use modified() so that the service is not restarted after the configuration changed.
* Don't use ManagedService, the Dictonary based update() signature isn't very modern anyway.


kaikreuzer:
@davidgraeff Regarding the current behavior of the ConfigDispatcher, it can be argued whether this is a bug or a feature. It is on intention that it does not delete anything from the ConfigAdmin.
I would suggest to decouple this discussion from your MQTT work and create a separate PR for it - but [let us first agree on what we actually need](https://github.com/openhab/openhab-distro/issues/396#issuecomment-320898405).

davidgraeff:
@htreu @kaikreuzer : Done

bundles/io/org.eclipse.smarthome.io.transport.mqtt/src/main/java/org/eclipse/smarthome/io/transport/mqtt/MqttService.java
```diff
@@ -34,6 +41,8 @@
  * @author Davy Vanherbergen
  * @author Markus Rathgeb - Synchronize access to broker connections
  */
+@Component(immediate = true, service = { MqttService.class, ManagedService.class }, configurationPid = {
+        "org.eclipse.smarthome.mqtt" }, name = "MQTTConnectionService", property = "service.pid=org.eclipse.smarthome.mqtt")
 public class MqttService implements ManagedService {
```
maggu2810: I assume there is no need to implement managed service at all anymore. Is it?

davidgraeff: True, but I wasn't sure if anything relied on MqttService being a ManagedService. Do we need ManagedServices at all in ESH?

kaikreuzer: No, they are outdated. We only use configurations for DS components throughout ESH.

davidgraeff: I have removed ManagedServices

bundles/io/org.eclipse.smarthome.io.transport.mqtt/.gitignore
```diff
@@ -0,0 +1 @@
+OSGI-INF/*.xml
```
htreu: PLease create the OSGI-INF folder and place the .gitignore inside with content *.xml.

davidgraeff:
@htreu Suggested changes incorporated

bundles/io/org.eclipse.smarthome.io.transport.mqtt/src/main/java/org/eclipse/smarthome/io/transport/mqtt/MqttService.java
```diff
@@ -34,7 +39,9 @@
  * @author Davy Vanherbergen
  * @author Markus Rathgeb - Synchronize access to broker connections
  */
-public class MqttService implements ManagedService {
+@Component(immediate = true, service = { MqttService.class }, configurationPid = {
+        "org.eclipse.smarthome.mqtt" }, name = "MQTTConnectionService", property = "service.pid=org.eclipse.smarthome.mqtt")
```
maggu2810: WRT to #4113, should we set the name ever?

bundles/io/org.eclipse.smarthome.io.transport.mqtt/src/main/java/org/eclipse/smarthome/io/transport/mqtt/MqttService.java
```diff
@@ -387,6 +410,7 @@ public void unregisterMessageProducer(String brokerName, MqttMessageProducer com
      * @param eventPublisher EventPublisher
      */
     @Deprecated
+    @Reference(cardinality = ReferenceCardinality.MANDATORY, policy = ReferencePolicy.DYNAMIC)
```
maggu2810: I assume we don't try to solve the "dynamic so volatile" topic here. Correct? It has been this way before and so we leave it for now...

bundles/io/org.eclipse.smarthome.io.transport.mqtt/src/main/java/org/eclipse/smarthome/io/transport/mqtt/MqttBrokerConnection.java
```diff
@@ -577,6 +580,15 @@ public synchronized void start() throws MqttException, ConfigurationException {
     public synchronized void close() {
         logger.trace("Closing connection to broker '{}'", getName());
         try {
+            // Abort a connection attempt
+            isConnecting = false;
+
+            // Stop reconnect strategy from trying to reconnect
+            if (reconnectStrategy != null) {
+                reconnectStrategy.connectionEstablished();
```
maggu2810: Why do you consider that a reconnect strategy implementation handles "close" and "connectionEstablished" the same way?
IMHO the reconnect strategy should be closed and leave it to the implementation if they act same on both methods.

bundles/io/org.eclipse.smarthome.io.transport.mqtt/src/main/java/org/eclipse/smarthome/io/transport/mqtt/reconnect/PeriodicReconnectStrategy.java
```diff
@@ -53,8 +53,12 @@ public PeriodicReconnectStrategy(int reconnectFrequency, int firstReconnectAfter
 
     @Override
     public void lostConnection() {
-        logger.info("Starting connection helper to periodically try restore connection to broker '{}'",
-                getBrokerConnection().getName());
+        if (scheduledTask != null && !scheduledTask.isDone()) {
```
maggu2810: Isn't connection established and lost connection called as a callback and potential from different thread. Can we really omit read / write locks? Are the functions called by the same thread all the time (also isReconnecting)?

bundles/io/org.eclipse.smarthome.io.transport.mqtt/src/main/java/org/eclipse/smarthome/io/transport/mqtt/MqttService.java
maggu2810: @kaikreuzer WDYT?

bundles/io/org.eclipse.smarthome.io.transport.mqtt/src/main/java/org/eclipse/smarthome/io/transport/mqtt/reconnect/PeriodicReconnectStrategy.java
davidgraeff: We can go with your PR instead, but it was work to create all those changes. Would be nice if you do all the additional things that this PR does.

maggu2810: I don't care which one is used. I don't want to replace your one if just forget that you already fixed the reconnect issue.
So, I am fine with using yours, but we should still consider to find an accurate fix.
I didn't check e.g. "isDone()"

bundles/io/org.eclipse.smarthome.io.transport.mqtt/src/main/java/org/eclipse/smarthome/io/transport/mqtt/MqttBrokerConnection.java
davidgraeff: It should be the same for the strategy in regard that it should stop to reconnect. But if you see reasoning to split it, go for it.

bundles/io/org.eclipse.smarthome.io.transport.mqtt/src/main/java/org/eclipse/smarthome/io/transport/mqtt/reconnect/PeriodicReconnectStrategy.java
davidgraeff: I have no time to change this PR in the next days. But I can give you push access to my repo. I have seen that synchronized the access. Might be a good idea to make this thread safe, yes.

maggu2810: @davidgraeff I added the reconnect relevant part of your changes to #4160. Can you review it and give your approval or comments what should be changed?

If we get me access rights to this one, I will remove the changes of the reconnect part from this PR, so it contains the reloading configuration fix only.

bundles/io/org.eclipse.smarthome.io.transport.mqtt/src/main/java/org/eclipse/smarthome/io/transport/mqtt/MqttService.java
kaikreuzer: No, we shouldn't.

davidgraeff:
@maggu2810 I am on 3G network and had an internet outage so couldn't ask this question in your PR:
Does your solution allows the ReconnectStrategy to be used again? It didn't look like it. Because the MqttBrokerConnection allows to close and start a connection, there is no final close.

maggu2810:
Hm, no if the strategy has been closed it is closed.

IIRC in Java is it common practice that a closed object (I could be wrong, will check it) is not usable anymore.
If there is a start there should IMHO been a stop, so you could call start, stop, start, stop, ... but a close is normally used if resources should be freed and the reference is "thrown away".

davidgraeff:
I didn't name the method close, it was already there. But I recommend to rename it to stop(), even if it is an API break, as it seems to cause confusion. For the MqttBrokerConnection it makes sense that it is restart-able. Some settings like the clientID need a restarted connection.

maggu2810:
Okay, if a broker connection and its reconnect strategy needs to be stoppable and startable again, the reconnect strategy change needs to be changed again.
IIRC there has already been a renaming of a MQTT broker connection function (connection state changed?), so I assume it would be also okay to change `close` to `stop` now.

We could introduce a `start` and `stop` for the reconnection strategy, too to handle some init / cleanup if necessary correctly.

davidgraeff:
Yes, a start() can be introduced as well. For the PeriodicStrategy I recommend to release resources in the "connectionEstablished()" though. There is no need to keep any timers or executors around.

maggu2810:
I will have a look at.

maggu2810:
@davidgraeff I removed the reconnect related changes from this PR. I also removed the modified component name from MQTT service as this will result in different XML names between Maven and Eclipse IDE.

maggu2810:
As I also edited this PR now, I assume I cannot approve or review it anymore...

PR #3983, opened 2017-08-08
Change for the automatic inbox processor 

afuechsel:
Compare the value of the representation property not only to a value of a property but also of a configuration parameter

### DIFF
--- a/bundles/config/org.eclipse.smarthome.config.discovery/src/main/java/org/eclipse/smarthome/config/discovery/internal/AutomaticInboxProcessor.java
+++ b/bundles/config/org.eclipse.smarthome.config.discovery/src/main/java/org/eclipse/smarthome/config/discovery/internal/AutomaticInboxProcessor.java
@@ -92,7 +92,8 @@ public class AutomaticInboxProcessor extends AbstractTypedEventSubscriber<ThingS
 
     private String getRepresentationValue(DiscoveryResult result) {
         return result.getRepresentationProperty() != null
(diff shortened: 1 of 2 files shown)


bundles/config/org.eclipse.smarthome.config.discovery/src/main/java/org/eclipse/smarthome/config/discovery/internal/AutomaticInboxProcessor.java
```diff
@@ -92,7 +92,8 @@ public void thingAdded(Inbox inbox, DiscoveryResult result) {
 
     private String getRepresentationValue(DiscoveryResult result) {
         return result.getRepresentationProperty() != null
-                ? (String) result.getProperties().get(result.getRepresentationProperty()) : null;
+                ? (String) result.getProperties().get(result.getRepresentationProperty())
```
kaikreuzer: I would be interested what has changed on your IDE setup that causes this line break all of a sudden. Do you happen to know?

afuechsel: I would be interested as well. Just double checked my settings of a brand new freshly installed Eclipse Oxygen using the Eclipse SmartHome setup - it says line braks at column 120 - still this formatting got changed somehow. Not sure why. May it be, that Oxygen has NEW formatting settings, appling them for ? : constructs?

PR #3985, opened 2017-08-08
Clean up Hue binding

afuechsel:
- removed parameter bridge on public API, which would expose the internal hue bridge
- removed the config parameter serial number from the hue bridge
- added the representation property to toString of DiscoveryResultImpl and to the console output of the inbox


kaikreuzer:
FTR: Added APIBreaking label because `onNotAuthenticated` is public and in an exported package (but I doubt that anyone uses it) and since serialNumber has been removed as a config parameter.

PR #3986, opened 2017-08-08
Added Nullannotations on Binding API

triller-telekom:


triller-telekom:
If I do not add the optional import to all the bindings the following "problem" occurs: In the binding to have a handler that derives from the `BaseThingHandler`. Inside my handler I can call `updateProperty(@NonNull String name, String value)` from the `BaseThingHandler` with `updateProperty(null, "test")` and there won't be any warning or error.

However, if I do add the optional import, the null analysis starts and I see the error. So to me it seems like every bundle has to be aware of the annotations, otherwise the null analysis ignores it.

Regarding the position of the annotation on return values of methods I made it consistent with the documentation of Eclipse: https://help.eclipse.org/oxygen/index.jsp?topic=%2Forg.eclipse.jdt.doc.user%2Ftasks%2Ftask-using_null_annotations.htm

But If you find the other position more intuitive I can change that.

maggu2810:
> Regarding the position of the annotation on return values of methods I made it consistent with the documentation of Eclipse: https://help.eclipse.org/oxygen/index.jsp?topic=%2Forg.eclipse.jdt.doc.user%2Ftasks%2Ftask-using_null_annotations.htm

Can you point me to the location you refer on that page?
The return type should be annotated.
So, `public @Nullable Object foo()` is more intuitive then `@Nullable public Object foo()`.
There is an example using a static method and a generic, the annotation is used after the static keyword and not before.
public / protected / private are modifiers for the function and so in front of the return type.

triller-telekom:
I was referring to the very first example:
```
@NonNull String getString(String maybeString) {
        if (maybeString != null)
            return maybeString;                         // the above null check is required
        else
            return "<n/a>";
    }
```

But you are right the public / protected / private modifiers are not in these examples.

Also I just noticed that Kai has done it like you suggested in his PR too.

So I will change it in my PR.

triller-telekom:
Done

maggu2810:
Thanks for your work

PR #3988, opened 2017-08-08
[Discussion] ConfigDispatcher: Notify services of altered/removed configuration

davidgraeff:
## Config.Dispatch:
Until now the config dispatcher didn't notify services if a configuration tuple was removed or the entire configuration files was removed.
The MqttService needs this information though, to remove old configured broker connections.

This works only if the configuration file is dedicated to the service (no multiple service pids
in one file), but this is the recommended way anyway.

Discussion can be found [here](https://github.com/openhab/openhab-distro/issues/396#issuecomment-321021872)

### DIFF
--- a/bundles/config/org.eclipse.smarthome.config.dispatch/META-INF/MANIFEST.MF
+++ b/bundles/config/org.eclipse.smarthome.config.dispatch/META-INF/MANIFEST.MF
@@ -10,5 +10,6 @@ Import-Package: org.apache.commons.io;version="2.2.0",
  org.eclipse.smarthome.config.core,
  org.eclipse.smarthome.core.service,
  org.osgi.service.cm;version="1.5.0",
+ org.osgi.framework;version="1.5.0",
  org.slf4j;version="1.7.2"
 Service-Component: OSGI-INF/*.xml
--- a/bundles/config/org.eclipse.smarthome.config.dispatch/src/main/java/org/eclipse/smarthome/config/dispatch/internal/ConfigDispatcher.java
+++ b/bundles/config/org.eclipse.smarthome.config.dispatch/src/main/java/org/eclipse/smarthome/config/dispatch/internal/ConfigDispatcher.java
@@ -29,6 +29,7 @@ import org.apache.commons.io.IOUtils;
 import org.apache.commons.lang.StringUtils;
 import org.eclipse.smarthome.config.core.ConfigConstants;
 import org.eclipse.smarthome.core.service.AbstractWatchService;
+import org.osgi.framework.InvalidSyntaxException;
 import org.osgi.service.cm.Configuration;
 import org.osgi.service.cm.ConfigurationAdmin;
 import org.osgi.service.cm.ManagedService;
@@ -47,18 +48,18 @@ import org.slf4j.LoggerFactory;
  * <p>
  * The format of the configuration file is similar to a standard property file, with the exception that the property
  * name can be prefixed by the service pid of the {@link ManagedService}:
- * 
+ *
  * <p>
  * &lt;service-pid&gt;:&lt;property&gt;=&lt;value&gt;
- * 
+ *
  * <p>
  * In case the pid does not contain any ".", the default service pid namespace is prefixed, which can be defined by the
  * program argument "smarthome.servicepid" (default is "org.eclipse.smarthome").
- * 
+ *
  * <p>
  * If no pid is defined in the property line, the default pid namespace will be used together with the filename. E.g. if
  * you have a file "security.cfg", the pid that will be used is "org.eclipse.smarthome.security".
- * 
+ *
  * <p>
  * Last but not least, a pid can be defined in the first line of a cfg file by prefixing it with "pid:", e.g.
  * "pid: com.acme.smarthome.security".
@@ -68,6 +69,20 @@ import org.slf4j.LoggerFactory;
  * @author Ana Dimova - reduce to a single watch thread for all class instances
  */
 public class ConfigDispatcher extends AbstractWatchService {
+    /**
+     * Represents a result of parseLine().
+     */
+    private class ParseLineResult {
+        public String pid;
+        public String property;
+        public String value;
+
+        public ParseLineResult(String pid, String property, String value) {
+            this.pid = pid;
+            this.property = property;
+            this.value = value;
+        }
+    }
 
     private static String getPathToWatch() {
         String progArg = System.getProperty(SERVICEDIR_PROG_ARGUMENT);
@@ -151,6 +166,22 @@ public class ConfigDispatcher extends AbstractWatchService {
             } catch (IOException e) {
                 logger.warn("Could not process config file '{}': {}", path, e);
             }
+        } else if (kind == ENTRY_DELETE) {
+            // Detect if a service specific configuration file was removed. We want to
+            // notify the service in this case with an updated empty configuration.
+            File configFile = path.toFile();
+            if (configFile.isHidden() || configFile.isDirectory() || !configFile.getName().endsWith(".cfg")) {
+                return;
+            }
+            String pid = pidFromFilename(configFile);
+            try {
+                Configuration configs[] = configAdmin.listConfigurations("(service.pid=" + pid + ")");
+                if (configs.length > 0) {
+                    configs[0].update((Dictionary) new Properties());
+                }
+            } catch (IOException | InvalidSyntaxException ignored) {
+                return;
+            }
         }
     }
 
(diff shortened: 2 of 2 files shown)


maggu2810:
I like to use separate files for different PIDs.

But I don't know if I understand your "remove configuration" logic.

Let's assume there is a service (PID: a.b.c) that is using the following default configuration:
* key1=kv1_dfl1
* key2=kv2_dfl2

Scenario:
* key1 is set by a configuration file `a.b.c.cfg` with an entry key1=val_a
* key2 is set by another mechanism that uses configuration admin (e.g. the WebConsole of Karaf https://karaf.apache.org/manual/latest/webconsole)
* the file `a.b.c.cfg` is removed
* key1 and key2 are removed from config admin

Is this what you are doing now?

davidgraeff:
Yes, that's what I'm doing here and what you mentioned is the main argument again this approach, because GUI or console configuration would be wiped as well. But the current way of handling configuration is not usable for the MqttService and causes all kind of confusion for users as seen in the OH forum.
If you follow the link to the discussion, you will see my proposed solution. To sum it up: key-value pairs would need to be marked with FROM_TEXT_FILE and only those would be altered/removed by configuration file changes.

kaikreuzer:
> This works only if the configuration file is dedicated to the service, but this is the recommended way anyway.

No, this is not true. It is one of the main features of this service to allow multiple settings in the same file - see e.g. [smarthome.cfg](https://github.com/eclipse/smarthome/blob/master/distribution/smarthome/conf/smarthome.cfg) in ESH or  [runtime.cfg](https://github.com/openhab/openhab-distro/blob/master/distributions/openhab/src/main/resources/conf/services/runtime.cfg) in openHAB.
I think this is an extremely valuable feature since otherwise you end up confronting your users with a lot of technical details (service pids do not mean anything to them). This is one of my main issues about Karaf: You have 30 different config files with long technical names (the pids) and the average (non-technical) user has no clue where to find a certain setting.

I think for the suggestions done here, there are already perfect solutions available: The one is Felix FileInstall (that is used in Karaf), which makes it possible to have one file per service and which perfectly syncs with the content in ConfigurationAdmin. For being notified about removals, it is easy to subscribe to ConfigurationAdmin.

Note that the [original discussion](https://github.com/openhab/openhab-distro/issues/396#issuecomment-321021872) is about the fact that outdated config entries remain in ConfigurationAdmin if entries from a ConfigDispatcher cfg file are removed. I therefore still think that the most straight-forward solution is the [introduction of a marker](https://github.com/openhab/openhab-distro/issues/396#issuecomment-320899961), which signals whether content in ConfigurationAdmin should be fully wiped or not. The main question for me is what would be the best default (on or off) for it.

davidgraeff:
If it is a service specific file, then exclusive=yes, if it is a mixed service file, then exclusive=no as default.

davidgraeff:
@complicated service file names for users:
the file name can be influenced by setting it to something else than default of course, even easier with the annotation now.

davidgraeff:
With the needs of MqttService and maybe other services: should those talk to ConfigurationAdmin directly for remove notifications, next to implementing modified? At the moment I can't really think of what I would need to change.

maggu2810:
> I think this is an extremely valuable feature since otherwise you end up confronting your users with a lot of technical details (service pids do not mean anything to them).

I don't know if I get the point here.
We confront a user with technical details if we should use a file `a.b.c.cfg` to set `e=f` and `m.n.o.cfg` to set `g=h` and he doesn't need to know such details if a file `sample.cfg` uses entries `a.b.c.e=f` and `m.n.o.g=h`?

If the ESH mechanism that sync to ConfigAdmin should be extended to consider the removal of entries, I think the default should be false. As we could break existing setups that rely on the old behavior.

davidgraeff:
The design is flawed @maggu2810, as seen by the many confused users in the forum (Openhab forum) and ESH has not reached 1.0. The default should be the correct way and not the backwards compatible way. Other implementors of ESH will probably not expose config files to users, so there is no impact to them.
In my opinion, two different defaults make sense depending on the file being a misc or a pure service configuration file.

maggu2810:
> The default should be the correct way and not the backwards compatible way.

I agree.
But I have to admit that I has been told often that we should keep backward compatibility and I does not know how much third party products get broken... :wink:
So, I assume it is not about my personal preference but about keeping products running.

davidgraeff:
But honestly, who is relying on the behaviour that you can only add to the config dispatcher and not remove anything ^^.

_Sent from my Lge LG-F700K using [FastHub](https://play.google.com/store/apps/details?id=com.fastaccess.github)_

kaikreuzer:
> But I have to admit that I has been told often that we should keep backward compatibility and I does not know how much third party products get broken... 

Nobody knows, that's why our natural impulse should be to keep backward compatibility where possible and only break it if there are good arguments for it - so it is imho always a case-by-case decision.

> two different defaults make sense depending on the file being a misc or a pure service configuration file.

I am not so sure if this is the best solution - is it a rather hidden (=surprising) feature. How would we tell apart the one from the other? By adding a line that uses a different config-pid, the behavior for the already existing entries will change? Sounds dangerous to me.

Could we maybe make it slightly more explicit? I just found that we already support a "pid marker" in the files, i.e. we can specify:
```
pid:org.eclipse.smarthome.mqtt
host.1=localhost
host.2=192.168.1.123
```
How about using such a pid entry as a marker that we want have exclusive rights on that pid and hence will remove any other configuration that might already exist?

davidgraeff:
@kaikreuzer What happens if such a marked file gets removed? Do we need to cache which file is marked?

kaikreuzer:
No - if it is removed, it means you are not not caring about the configuration anymore and you are happy with what is in ConfigAdmin. If this is not what you want, you need to keep the file and remove its entries (apart from the pid line).

davidgraeff:
The marker idea is better than the current situation I guess.

I'm not satisfied with how deleted exclusive files are handled, because as a user I would probably expect a different behaviour and I can already feel complains coming in like "I deleted all files to start fresh but ESH/OH still seems to have MQTT broker / Persistent services / SSL keys configured".

But again it's better than before so can we please implement this.

davidgraeff:
Nope. Busy with Mqtt, SSL and lwm2m.

kaikreuzer:
> What happens if such a marked file gets removed? Do we need to cache which file is marked?

Contemplating longer about that, maybe this isn't such a bad idea after all. I agree that from a user perspective, it would be the more expected behavior.
What we need to ensure is that it works the same way whether the system is running or not when a file is deleted. If a file is deleted and the OSGi framework is started "clean", i.e. with removed OSGi cache, it currently already works as expected. So we only have to address the situation of subsequent starts without a cleaned cache. If we would put a list of "exclusive pids" that we have fed into ConfigAdmin within the ConfigDispatchers bundle data file, we could check this list upon startup and remove all entries for which we do not find a file anymore.
Entries in non-exclusive files should imho not be considered, though, this would make the mechanism overly complex.

htreu:
I started work on this.

davidgraeff:
@htreu @kaikreuzer I had this awesome idea: Can the marker, that we agreed on, be part of the filename? Example:
* The service BlaBlub is very undemanding and only want a few configuration values and don't care about removals: The configurationPid would be "blablub" for a file "blablub.cfg".
* The service NetSecurity needs its cfg file content to be in sync with its internal view of the configuration. It declares a configurationPid of "netsecurity.exclusive" for a file "netsecurity.exclusive.cfg".

This way we each service can affect how its configuration is handled. And we don't need to keep track which file was exclusive if it gets deleted, because the filename tells us.

kaikreuzer:
> a file "netsecurity.exclusive.cfg"

I don't really think that this is nicely readable - it would imho mess up the services folder.

> This way each service can affect how its configuration is handled.

No, because it would not be defined what should happen if the file is marked as exclusive and contains entries for 10 different pids. I prefer to leave it to the content to be consistent.

> And we don't need to keep track which file was exclusive if it gets deleted, because the filename tells us.

A file that has been deleted cannot tell you much anymore, can it? So I don't see why we would not have to track the files.

davidgraeff:
The file watcher tells us the deleted filename, and if the exclusive marker is encoded into the filename, we don't need a separate list somewhere in memory.

If a file is marked as exclusive the parser should not even accept the pid:key=value syntax.

kaikreuzer:
How should the file watcher tell you, if you delete a file while your runtime is down?

davidgraeff:
Hm, ok, we need at least the file list in the OSGI cache and compare it on next start.

PR #3978, opened 2017-08-07
[Tradfri] Updated and added german translation

cweitkamp:
Signed-off-by: Christoph Weitkamp <github@christophweitkamp.de>

PR #3993, opened 2017-08-09
Adjust description of Tradfri channel types

htreu:
In conjunction with #3978 the channel-type description should not refer to the channel-type but to the actual functionality of the device.

### DIFF
--- a/extensions/binding/org.eclipse.smarthome.binding.tradfri/ESH-INF/thing/thing-types.xml
+++ b/extensions/binding/org.eclipse.smarthome.binding.tradfri/ESH-INF/thing/thing-types.xml
@@ -5,7 +5,8 @@
         xsi:schemaLocation="http://eclipse.org/smarthome/schemas/thing-description/v1.0.0 http://eclipse.org/smarthome/schemas/thing-description-1.0.0.xsd">
 
     <bridge-type id="gateway">
-        <label>Trådfri Gateway</label>
+        <label>TRÅDFRI Gateway</label>
+        <description>IKEA TRÅDFRI IP Gateway</description>
 
         <config-description>
             <parameter name="host" type="text" required="true">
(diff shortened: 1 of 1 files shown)


extensions/binding/org.eclipse.smarthome.binding.tradfri/ESH-INF/thing/thing-types.xml
```diff
@@ -6,6 +6,7 @@
 
     <bridge-type id="gateway">
         <label>Trådfri Gateway</label>
+        <description>IKEA TRÅDFRI IP Gateway</description>
```
kaikreuzer: We should decide whether we use capitals for TRåDFRI or not - label & desc should be in sync.

extensions/binding/org.eclipse.smarthome.binding.tradfri/ESH-INF/thing/thing-types.xml
```diff
@@ -93,10 +94,7 @@
     <channel-type id="brightness">
         <item-type>Dimmer</item-type>
         <label>Brightness</label>
-        <description>
-            The brightness channel allows to control the brightness of a light.
-            It is also possible to switch the light on and off.
-        </description>
+        <description>Control the brightness and switch the light on and off.</description>
```
kaikreuzer: Did you check the formulations on our other bindings? I wonder if we should somewhere document some guidelines on how labels & descriptions should be written, wdyt?

extensions/binding/org.eclipse.smarthome.binding.tradfri/ESH-INF/thing/thing-types.xml
cweitkamp: +1 for capitals: IKEA TRÅDFRI. It's the official Product name.

htreu: agreed.

extensions/binding/org.eclipse.smarthome.binding.tradfri/ESH-INF/thing/thing-types.xml
htreu: no, I did not check other bindings. right now the formulations are in different forms. Lifx has something like `Sets the temperature of the light` while Hue has `The color temperature channel allows to set...`. I would tend not to refer to `channel` in the description but describe the capability/property of the thing it controls. wdyt?

kaikreuzer: Yes, I'd agree.

Issue #3963, opened 2017-08-05
Hue binding: XML thing types in error on tag representation-property 


lolodomo:
When I open one of the XML thing types of the hue binding, I see an error on the tag `representation-property`.
@afuechsel : was the DTD updated ?

lolodomo:
@kaikreuzer : that is probably more something for you. The updated file has probably not be copied on the WEB.

kaikreuzer:
You are right, I didn't yet find the time to upload the latest version to eclipse.org/smarthome.
And trying it now, the git server isn't responding... I'll try again tomorrow.

lolodomo:
Without that, it looks like you are not able to create any hue thing.

kaikreuzer:
You mean it has an impact on the runtime side, really? Imho the xsd should only be checked in the IDE.

lolodomo:
When I run the server from Eclipse and try to create a hue bridge thing from its discovery, it fails with the current code. I am forced to update the XML to suppress the new tag if I want to have something working.

I have not tested outside Eclipse.

kaikreuzer:
Do you have a log of that failure?

lolodomo:
I can provide it later if you want. The error si something relative to XML validation.

afuechsel:
This would be interesting. There should be no XML validation in the runtime.

lolodomo:
Here are the log at startup:
```
2017-08-08 19:36:20.582 [DEBUG] [c.x.o.XmlDocumentBundleTracker:329  ] - Reading the XML document '/ESH-INF/binding/binding.xml' in module 'org.eclipse.smarthome.binding.ntp'...
2017-08-08 19:36:20.629 [INFO ] [rthome.event.InboxUpdatedEvent:43   ] - Discovery Result with UID 'hue:bridge:0017881108fc' has been updated.
2017-08-08 19:36:20.645 [DEBUG] [s.c.d.internal.PersistentInbox:190  ] - Updated discovery result for 'hue:bridge:0017881108fc'.
2017-08-08 19:36:20.660 [DEBUG] [c.x.o.XmlDocumentBundleTracker:329  ] - Reading the XML document '/ESH-INF/thing/bridge.xml' in module 'org.eclipse.smarthome.binding.hue'...
[...]

PR #3995, opened 2017-08-09
Add Static code analysis tool to ESH build

### LINKED ISSUE #2437: Static Code Analysis Checks

svilenvul:
Closes #2437 

Before merging:
 - ~~CQ should be approved~~;
 - ~~the version of the tool has to be updated with the latest that is released~~; 
 - ~~the high priority issues should be fixed~~;
 - ~~https://github.com/eclipse/smarthome/issues/3963 is resolved. See https://github.com/openhab/static-code-analysis/issues/136~~


tools/static-code-analysis/checkstyle/suppressions.xml
```diff
@@ -16,4 +16,6 @@
     <!-- Eclipse SmartHome specific suppressions-->
     <!-- These bundles are generated trough XText -->
     <suppress files=".+org.eclipse.smarthome.model.+" checks="RequireBundleCheck|ExportInternalPackageCheck|ManifestPackageVersionCheck|ImportExportedPackagesCheck"/>
+    <!-- Some source files have different headers -->
+    <suppress files=".+org.eclipse.smarthome.automation.+" checks="HeaderCheck"/>
```
htreu: Do we really want to suppress checks or rather update the automation bundles to adhere to our guidelines?

kaikreuzer: Yes, because we currently have a different arrangement for the license header of the automation component.

htreu: okay, tnx.

extensions/binding/org.eclipse.smarthome.binding.hue.test/src/test/groovy/org/eclipse/smarthome/binding/hue/test/HueBridgeHandlerOSGiTest.groovy
```diff
@@ -218,7 +218,7 @@ class HueBridgeHandlerOSGiTest extends AbstractHueOSGiTest {
     }
 
     @Test
-    void 'assert that a status configuration message for missing bridge IP is properly returned (IP is an empty string)'() {
+    void 'assert that a status configuration message for missing bridge IP is properly returned - IP is an empty string'() {
```
maggu2810: Can you create a separate PR for that change?
I expect that it will be okay to change the long groovy names.

extensions/binding/org.eclipse.smarthome.binding.hue.test/src/test/groovy/org/eclipse/smarthome/binding/hue/test/HueBridgeHandlerOSGiTest.groovy
```diff
@@ -197,7 +197,7 @@ class HueBridgeHandlerOSGiTest extends AbstractHueOSGiTest {
     }
 
     @Test
-    void 'assert that a status configuration message for missing bridge IP is properly returned (IP is null)'() {
+    void 'assert that a status configuration message for missing bridge IP is properly returned - IP is null'() {
```
maggu2810: Can you create a separate PR for that change?
I expect that it will be okay to change the long groovy names.

bundles/core/org.eclipse.smarthome.core/src/main/java/org/eclipse/smarthome/core/items/ItemUtil.java
```diff
@@ -78,7 +78,7 @@ public static void assertValidItemName(String itemName) throws IllegalArgumentEx
 
     public static State convertToAcceptedState(State state, Item item) {
         if (state == null) {
-            LOGGER.error("A conversion of null was requested", new NullPointerException("state should not be null"));
+            LOGGER.error("A conversion of null was requested", new NullPointerException("state should not be null")); // NOPMD
```
maggu2810: What is the exact PMD warning that is suppressed here?

extensions/binding/org.eclipse.smarthome.binding.hue.test/src/test/groovy/org/eclipse/smarthome/binding/hue/test/HueBridgeHandlerOSGiTest.groovy
svilenvul: Yes, of course.

kaikreuzer:
Created CQ 13980 for the tool dependency.

kaikreuzer:
Woah, that was quick: The CQ is already approved!

bundles/core/org.eclipse.smarthome.core/src/main/java/org/eclipse/smarthome/core/items/ItemUtil.java
svilenvul: See https://github.com/eclipse/smarthome/pull/3888

svilenvul:
The latest build fails with the hue binding. After the update of the XML things schema, the build should be green :-).

svilenvul:
@kaikreuzer 

Friday the only remaining errors were caused from the .xsd file that is not up to date - #3963.

Should I ignore these errors or wait for the update of the schema file ?

kaikreuzer:
Wait for the schema file - it should be up to date now! I retrigger Travis...

svilenvul:
Retrigger Travis.

svilenvul:
The build failed after 49 mins :
```
...
Still running (49 of 60): ./.build.sh
Still running (50 of 60): ./.build.sh

The job exceeded the maximum time limit for jobs, and has been terminated.
```
@kaikreuzer I guess the build is too slow with the static-analysis-tool added.

sjka:
I restarted it once again, just to make sure it was not caused by a slow repository, but the result was the same. That's indeed not so good news...
The global timeout for travis builds indeed is 50 minutes (see https://docs.travis-ci.com/user/customizing-the-build#Build-Timeouts), so there is nothing much we can do about it. 
The build until now took something just above 20 minutes. Is it at least plausible that the code analysis adds _that_ much to it? Sounds like a lot, so I'm wondering whether something went seriously wrong or not...

svilenvul:
When we talk about speed/performance, FindBugs is always the main suspected :-) .

```
mvn static-code-analysis:findbugs
....
[INFO] Total time: 18:46 min
[INFO] Finished at: 2017-08-15T11:45:17+03:00
[INFO] Final Memory: 238M/1127M
```
Checkstyle checks are taking 06:32 min and PMD 02:41 min.

I guess we can speed it up a bit if we tweak the configuration of FindBugs a little bit, but I am not sure if it will be enough.

I would appreciate help here.

svilenvul:
I have tried several scenarios to reduce the time that is needed from FindBugs without much success:
- applied some improvements to the configuration - reduced the time with 2 minutes;
- reduced the number of detectors from 60 to 3 (the one that log the most errors so far) :-) and FindBugs still needs 12 minutes to check all bundles (that is only 6 minutes less build time).

I have carefully looked at the maven log (with debug activated for FindBugs) and I see no unexpected behavior. I will not attach here the logs here as they are 100 MB.

When the timeout is fixed to 50 minutes, I would propose to disable the `findbugs` goal at all. All the custom checks that we have are implemented for Checkstyle. The FindBugs detectors are not following the coding guidelines (with the exception of the logger detectors), they are extra checks in this case. We will keep the main functionality of the tool without FindBugs.

With FindBugs disabled my local build has finished for 23 minutes: 
```
[INFO] BUILD SUCCESS
[INFO] ------------------------------------------------------------------------
[INFO] Total time: 23:28 min
[INFO] Finished at: 2017-08-15T17:15:02+03:00
[INFO] Final Memory: 626M/1894M
[INFO] ------------------------------------------------------------------------
```

Please let me know what do you think about disabling the FindBugs checks for the ESH repository.

maggu2810:
Instead of disabling checks we should consider to start the split of core and extensions into different repositories. This will reduce the build time for the respective parts and perhaps allow to run all tests.

kaikreuzer:
But it also means more overhead in maintaining the builds, so I am not too keen on that.
Just to prove the point: When was the last time that we have build a Designer, since it has been moved to its own repo...?

J-N-K:
Maybe that has something to do with the general state of the designer (i.e. the fact that the current version is quite useless). However, this should be decided my those that have to maintain the repos.

kaikreuzer:
@J-N-K Its days are numbered anyhow, see https://dev.eclipse.org/mhonarc/lists/smarthome-dev/msg00146.html.

J-N-K:
@kaikreuzer, that is what I wanted to say. The designer is in bad shape and not much development is being done there (because better alternatives are on their way, which is good). Therefore I think the designer is not a good comparison. IMHO splitting extensions and core makes more modular solutions possible (e.g. core+specialized extension, leaving out the "standard" extensions).

maggu2810:
> But it also means more overhead in maintaining the builds, so I am not too keen on that.

Okay, then I missed some news. AFAIK you prefer a split of core and extensions.

> Just to prove the point: When was the last time that we have build a Designer, since it has been moved to its own repo...?

IMHO it is more about that the designer is not really maintained and not because it has been moved to another repo.

maggu2810:
@svilenvul I stumbled over a file using non-common modifier order. Shouldn't we force the usage of the common order?

svilenvul:
We have a check for modifier order - https://github.com/openhab/static-code-analysis/commit/fca697a729df9dcd253ea9983ef07954c933d98b.

svilenvul:
I can suggest more more "workarounds":
- additional Jenkins profile (fast profile) with FindBugs disabled ?

svilenvul:
@kaikreuzer, I guess it is up to you. 

> Please let me know what do you think about disabling the FindBugs checks for the ESH repository.

What we are going to do in this situation ?

kaikreuzer:
> Okay, then I missed some news. AFAIK you prefer a split of core and extensions -- some time ago.

No, you are only always underestimating my pragmatism. While I agree that it would be nice to have it in a separate repo and also to version it separately from the core, I have learned from having openHAB split into several repos what overhead this means in terms of build & release management. After all, we duplicate the number of build jobs, the repos to update for new tools like the static code analysis or other upgrades, the efforts on doing releases and the number of issue trackers (while people today already never know where to post something).
Seeing that we didn't manage to do an official release for a long time let's me doubt that we are in a position to think about separate versioning of core and extensions (which would imho be the main reason for a repo split).

> Please let me know what do you think about disabling the FindBugs checks for the ESH repository.

I don't think that genereally disabling valuable checks is an option. We should rather think about how to reduce the situations, where we execute them. I would claim that the build time of the nightly job on our hudson is not an issue, so I would prefer to have it executing the full test/check suite.

Wrt Travis, I am not sure what our options are. If you could add a parameter which deactivates the FindBugs checks and we set that in the Travis configuration only, it might be an acceptable compromise.

svilenvul:
Great, I have deactivated the FindBugs checks only in Travis.

This has caused [some errors](https://github.com/openhab/static-code-analysis/commit/6a63fc3e60163734a62def6385988bbd0ae93e01#commitcomment-23718985) in the report generation goal, but they are already fixed with this [commit](https://github.com/openhab/static-code-analysis/commit/6a63fc3e60163734a62def6385988bbd0ae93e01).

kaikreuzer:
> so maybe it is better to release new version first and then add it to the ESH build.

New version 0.3.0 has been released - do you want to update the PR accordingly? I think we would be good to merge then!

svilenvul:
@kaikreuzer, I have rebased and updated the PR.

Do you think we should add a few sentences for the tool in the ESH documentation like we have done this in the openHAB docs - http://docs.openhab.org/developers/development/bindings.html#static-code-analysis ? Where would be the best place for that - maybe in the Coding Guidelines ?

kaikreuzer:
Yes, I guess the coding guidelines are the best place for it.

svilenvul:
@kaikreuzer, I am ready with the updates on this PR.

docs/documentation/development/guidelines.md
```diff
@@ -68,3 +68,11 @@ Note that this list also serves as a checklist for code reviews on pull requests
...
 1. Likewise, bundles that accept external requests (such as servlets) must not log errors or warnings if incoming requests are incorrect. Instead, appropriate error responses should be returned to the caller.
+
+## Static code analysis
+
+The Eclipse SmartHome Maven build includes [tooling for static code analysis](https://github.com/openhab/static-code-analysis) that will validate your code against the Coding Guidelines and some additional best practices. Information about the checks can be found [here](https://github.com/openhab/static-code-analysis/blob/master/docs/included-checks.md).
+
+The tool will generate an idividual report for each binding that you can find in `../your_binding_directory/target/code-analysis/report.html` file and a report for the whole build that contains links to the individual reports in the `../smarthome/target/summary_report.html`. The tool categorizes the found issues by priority: 1(error),2(warning) or 3(info). If any error is found within your code the Maven build will end with failure. You will receive detailed information (path to the file, line and message) listing all problems with Prioriry 1 on the console.
```
maggu2810: It is not only about bindings but every bundle that is build (for each binding => for each bundle). And to be fair it would be not only be used for bundles, but for every artifact in the reactor, so "for each artifact" would be correct.

maggu2810: Why do you add a `../` in front of the path.
Normally you call `mvn ...` in the root of the reactor and IMHO the relative paths should be relative to the root.

pom.xml
```diff
@@ -469,6 +470,120 @@
...
+      <id>tycho-snapshots</id>
+      <url>https://oss.sonatype.org/content/groups/public/</url>
+    </pluginRepository>
+    <pluginRepository>
+      <id>cbi-releases</id>
+      <url>https://repo.eclipse.org/content/repositories/cbi-releases/</url>
+    </pluginRepository>
+    <pluginRepository>
+      <id>buchen-maven-repo</id>
+      <url>http://buchen.github.io/maven-repo</url>
```
maggu2810: Hasen't this one been removed previously by intention?
Why do you add it again?

pom.xml
```diff
@@ -469,6 +470,120 @@
...
+      <url>https://openhab.jfrog.io/openhab/libs-release</url>
+    </pluginRepository>
+  </pluginRepositories>
+
+  <repositories>
+    <repository>
+      <id>eclipse</id>
+      <name>Eclipse Snapshot Repository</name>
+      <layout>default</layout>
+      <url>https://repo.eclipse.org/content/repositories/snapshots/</url>
```
maggu2810: This has been removed some days ago, why do you need the Eclipse Snapshot Repository?

docs/documentation/development/guidelines.md
svilenvul: Thanks, I was wondering how to be more precise here, but without much success. I will change it to relative to the root of the reactor.

pom.xml
svilenvul: As I see I only need both JFrog repositories (one for the plugin and one for its dependencies) , I will remove the others.

docs/documentation/development/guidelines.md
svilenvul: At the moment it will be used only on bundles - there are two profiles in this PR that activate the tool for bundles and bindings (the one is obsolete now, I will remove it). Initially the idea was to have separate rulesets included in each profile, but this changed over time. 

Activating the checks for each artifact is not tested well enough, some of the OSGi specific checks might have problems with that.

svilenvul:
I hope I have answered all questions. Do you have any other concerns ?

pom.xml
```diff
@@ -469,6 +470,81 @@
...
+              <version>${sat.version}</version>
+          </plugin>
+        </plugins>
+      </build>
+    </profile>
   </profiles>
 
+  <pluginRepositories>
+    <pluginRepository>
+      <id>openhab-artifactory-release</id>
```
kaikreuzer: Actually, the official repo for the releases is "jcenter" at https://jcenter.bintray.com/.

It probably makes sense that I also request to have it published to Maven central, then we would not at all need a specific repo here.

pom.xml
```diff
@@ -469,6 +470,81 @@
...
+      <url>https://openhab.jfrog.io/openhab/libs-release</url>
+    </pluginRepository>
+  </pluginRepositories>
+
+  <repositories>
+    <repository>
```
kaikreuzer: Isn't the pluginRepository enough? Why do we need it as a dependency repository as well?

svilenvul: Good question, I haven't taught about that much until you asked. 
The tool(as a Maven *plugin*) internally starts the Maven plugins for Checkstyle, PMD and FindBugs and adds itself as a *dependency* to these plugins, so the included custom checks that are contained there are available in the classpath.

kaikreuzer: Did you test if it also works without it?

pom.xml
```diff
@@ -469,6 +470,81 @@
...
+              <version>${sat.version}</version>
+          </plugin>
+        </plugins>
+      </build>
+    </profile>
   </profiles>
 
+  <pluginRepositories>
+    <pluginRepository>
+      <id>JCenter Repository</id>
```
kaikreuzer: id should imho rather be "jcenter"

pom.xml
```diff
@@ -469,6 +470,81 @@
...
+      <url>https://jcenter.bintray.com</url>
+    </pluginRepository>
+  </pluginRepositories>
+
+  <repositories>
+    <repository>
+      <id>openhab-artifactory-release</id>
```
kaikreuzer: id should be "jcenter"

pom.xml
svilenvul: Yes, I tested it, it is not working without this repository.

kaikreuzer:
Thanks!

PR #3998, opened 2017-08-09
Magic Binding: Added channel with options and translation

triller-telekom:


bundles/test/org.eclipse.smarthome.magic/ESH-INF/i18n/magic_de.properties
```diff
@@ -0,0 +1,5 @@
+channel-type.hue.alert.label = Alarm
```
htreu: the binding id should be `magic`.

bundles/test/org.eclipse.smarthome.magic/ESH-INF/i18n/magic_de.properties
```diff
@@ -0,0 +1,5 @@
...
+channel-type.hue.alert.description = Erm�glicht ein tempor�res Blinken.
+channel-type.hue.alert.state.option.NONE = Kein Blinken
+channel-type.hue.alert.state.option.SELECT = Einmaliges Blinken
+channel-type.hue.alert.state.option.LSELECT = Mehrfaches Blinken
```
hakan42: Please add a line feed at the end of the file.

triller-telekom:
Thanks for noticing!

Stupid copy&paste... I had the correct one and then "force switched" branches so it got lost and I had to copy it again from the hue binding. Anyway, I have fixed it now.

PR #4013, opened 2017-08-10
Fix wrong service name in bindings archtype

triller-telekom:

### DIFF
--- a/tools/archetype/binding/src/main/resources/archetype-resources/src/main/java/internal/__bindingIdCamelCase__HandlerFactory.java
+++ b/tools/archetype/binding/src/main/resources/archetype-resources/src/main/java/internal/__bindingIdCamelCase__HandlerFactory.java
@@ -29,7 +29,7 @@ import org.osgi.service.component.annotations.Component;
  *
  * @author ${author} - Initial contribution
  */
-@Component(service = ThingHandlerFactory.class, immediate = true)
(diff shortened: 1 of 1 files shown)


tools/archetype/binding/src/main/resources/archetype-resources/src/main/java/internal/__bindingIdCamelCase__HandlerFactory.java
```diff
@@ -29,7 +29,7 @@
  *
  * @author ${author} - Initial contribution
  */
-@Component(service = ThingHandlerFactory.class, immediate = true)
+@Component(service = ThingHandlerFactory.class, immediate = true, configurationPolicy = ConfigurationPolicy.OPTIONAL, name = "binding.${bindingId}")
```
kaikreuzer: @triller-telekom Why did you actually add `configurationPolicy = ConfigurationPolicy.OPTIONAL` in this PR? Afaik, this is the default setting, so I wonder if we should really specifically add it.

triller-telekom: Now as you say it, I overlooked that it is default. I guess I copied it from another service that had it specified. I fixed the archetype, see PR #4125

PR #4016, opened 2017-08-10
Add exceptions to analysis tool

svilenvul:
See #3995

### DIFF
--- a/bundles/model/org.eclipse.smarthome.model.script/src/org/eclipse/smarthome/model/script/actions/LogAction.java
+++ b/bundles/model/org.eclipse.smarthome.model.script/src/org/eclipse/smarthome/model/script/actions/LogAction.java
@@ -7,6 +7,7 @@
  */
 package org.eclipse.smarthome.model.script.actions;
 
+import org.quartz.utils.FindbugsSuppressWarnings;
 import org.slf4j.Logger;
 import org.slf4j.LoggerFactory;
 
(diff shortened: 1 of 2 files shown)


bundles/model/org.eclipse.smarthome.model.script/src/org/eclipse/smarthome/model/script/actions/LogAction.java
```diff
@@ -7,6 +7,7 @@
  */
 package org.eclipse.smarthome.model.script.actions;
 
+import org.quartz.utils.FindbugsSuppressWarnings;
```
kaikreuzer: `org.quartz.utils`? Really?

svilenvul: [edu.umd.cs.findbugs.annotations.SuppressFBWarnings](http://findbugs.sourceforge.net/api/edu/umd/cs/findbugs/annotations/SuppressFBWarnings.html) would be the best choice, but it is not part of the Target Platform at the moment.

I was also surprised that the code above is working, but it is.

svilenvul: OK, I found them in the `com.eclipsesource.jaxrs.swagger-all.source_1.5.7.jar`, I will update the PR

kaikreuzer: `org.quartz` soon won't be part of the target platform anymore, that's why I am asking :-)

svilenvul: Nope, they are not exported.

svilenvul: Can we add the spotbugs eclipse plugin to the target platform? It exports the annotations package.

svilenvul:
Will be moved to xml filter file that will be part of #3995

PR #4023, opened 2017-08-11
Add spotbugs to target platform

svilenvul:
The plugin exports annotations for FindBugs(SpotBugs).

See https://github.com/eclipse/smarthome/pull/4016#discussion_r132476289
See #3995

### DIFF
--- a/targetplatform/smarthome.target
+++ b/targetplatform/smarthome.target
@@ -1,5 +1,5 @@
 <?xml version="1.0" encoding="UTF-8" standalone="no"?>
-<?pde version="3.8"?><target name="SmartHome" sequenceNumber="115">
+<?pde version="3.8"?><target name="SmartHome" sequenceNumber="116">
 <locations>
 <location includeAllPlatforms="false" includeConfigurePhase="false" includeMode="planner" includeSource="true" type="InstallableUnit">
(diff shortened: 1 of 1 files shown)


kaikreuzer:
Sorry, we cannot add this without a CQ approval.

So before creating a CQ, we need an agreement that we definitely need that. Do I get you right that those annotations are our only chance to define finegrained exceptions to our rules?

svilenvul:
No, finegrained exceptions can be added in a filter file as well (it can filter methods in a class), but it will be a lot more difficult to maintain.

Should we discuss if we need annotations for Checkstyle and PMD as well ? Should we bring that back to https://github.com/eclipse/smarthome/issues/2437?

svilenvul:
Actually Checkstyle and PMD support adding exceptions with comments, so annotations there should not be needed.

kaikreuzer:
Sounds like my preferred option then 👍

svilenvul:
OK, If  I got it right than I am going back to the filter file.

kaikreuzer:
I was referring to "support adding exceptions with comments" as my preferred option :-)
But anything that does not introduce a new dependency for the target platform is a good solution...

PR #4024, opened 2017-08-11
Added Null check on Representation Property

amitjoy:
`org.eclipse.smarthome.core.thing.type.ThingType.getRepresentationProperty()` javadoc says that representation property of this thing type can be null. Without any representation property results in NPE.

### DIFF
--- a/bundles/config/org.eclipse.smarthome.config.discovery/src/main/java/org/eclipse/smarthome/config/discovery/internal/AutomaticInboxProcessor.java
+++ b/bundles/config/org.eclipse.smarthome.config.discovery/src/main/java/org/eclipse/smarthome/config/discovery/internal/AutomaticInboxProcessor.java
@@ -11,8 +11,10 @@ import static org.eclipse.smarthome.config.discovery.inbox.InboxPredicates.*;
 
 import java.util.List;
 import java.util.Map;
+import java.util.Objects;
 import java.util.stream.Collectors;
 
+import org.eclipse.smarthome.config.core.Configuration;
 import org.eclipse.smarthome.config.discovery.DiscoveryResult;
 import org.eclipse.smarthome.config.discovery.DiscoveryResultFlag;
 import org.eclipse.smarthome.config.discovery.inbox.Inbox;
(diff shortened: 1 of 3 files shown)


bundles/config/org.eclipse.smarthome.config.discovery/src/main/java/org/eclipse/smarthome/config/discovery/internal/AutomaticInboxProcessor.java
```diff
@@ -79,13 +81,13 @@ public void receiveTypedEvent(ThingStatusInfoChangedEvent event) {
...
         if (autoIgnore) {
             String value = getRepresentationValue(result);
-            if (value != null) {
-                Thing thing = thingRegistry.stream()
-                        .filter(t -> value.equals(getRepresentationPropertyValueForThing(t))).findFirst().orElse(null);
-                if (thing != null) {
-                    logger.debug("Auto-ignoring the inbox entry for the representation value {}", value);
-                    inbox.setFlag(result.getThingUID(), DiscoveryResultFlag.IGNORED);
-                }
+            Thing thing = thingRegistry.stream()
```
tomhoefer: I think it is worth to write a unit test for this purpose ....

amitjoy: Done

triller-telekom:
Thanks for the fix, but while you are at it anyway, how about raising a warning in the IDE for the next one who uses `ThingType.getRepresentationProperty()`?

Just add a `@Nullable` in front of the return type `String` here: https://github.com/eclipse/smarthome/blob/master/bundles/core/org.eclipse.smarthome.core.thing/src/main/java/org/eclipse/smarthome/core/thing/type/ThingType.java#L312

amitjoy:
@tomhoefer I have added a new unit test to check thing when it does not contain any representation property.

@triller-telekom I have added a new commit to introduce JDT null-check annotation in `ThingType#getRepresentationProperty()`

Issue #4028, opened 2017-08-11
Sonos binding: adding a representation property


lolodomo:
A representation propoerty should be added to each Sonos thing type that will reference the UDN thing configuration setting.
Doing that, it will avoid to have Sonos in the inbox when they already exist as thing even with the non default thing UID.

I will implement this change.

lolodomo:
Now implemented and PR merged

PR #4035, opened 2017-08-11
Sonos binding: add representation property

lolodomo:
Also add the room name to the label of a discovered Sonos device to help the user distinguishing discovered Sonos of the same kind

Fix issue #4028 


extensions/binding/org.eclipse.smarthome.binding.sonos/src/main/java/org/eclipse/smarthome/binding/sonos/discovery/ZonePlayerDiscoveryParticipant.java
```diff
@@ -51,11 +51,12 @@ public DiscoveryResult createResult(RemoteDevice device) {
                 } catch (Exception e) {
                     // ignore and use default label
                 }
+                label += " (" + roomName + ")";
```
kaikreuzer: This one is interesting.
The room name used to be the "representation property" as it is the information which is useful to the user to identify the device. The new UDN is really just something very technical, which won't be helpful to the user at all :-(
But I guess this is the side effect of the new auto-ignore feature - not much we can do about it.

PR #4038, opened 2017-08-12
Hue binding: localized thing status description

lolodomo:


lolodomo:
@cweitkamp : once this PR is merged, I encourage you to update the German translation. I have added the new entries in comments in the German file.

sjka:
Adding the API-Breaking label because solutions with additional languages will have to amend their translation files.

sjka:
On a second though, I think this would alter the meaning of the APIBreaking label and would pollute the list of code-breaking changes. Therefore I added a new "Translation" label for this purpose.

PR #4040, opened 2017-08-12
BasicUI: build frontend files as a part of Maven build process

resetnow:


resetnow:
It's WIP because I want to see how it goes on Travis.

resetnow:
Okay, travis fails because download.itemis.com is down and some repo is hosted there. Anyway, the build is working for me locally.

kaikreuzer:
I can retrigger Travis.

resetnow:
But the mentioned repo is still down, and other PRs fail too with the same error: https://travis-ci.org/eclipse/smarthome/builds/263839542

maggu2810:
Would it make sense to move antlr-generator-3.2.0-patch.jar to our github repo similar to the third party repo?

kaikreuzer:
I would have to dig for details, but I am sure that there was some reason, why Xtext/MWE didn't simply include this bundle in their project to host it at Eclipse, but rather prompt the user if he agrees to download it and then fetches it from externally...

resetnow:
Forgive my ignorance but how is antrl related to BasicUI?

kaikreuzer:
Not at all, it was just the artifact that caused the build to be red because its repo was down - but it is only required for other parts of the build, not for the Basic UI.

resetnow:
Both of these packages are somewhere deep into the dependency chain. Updating packages listed in `package.json` to newer versions may solve the problem, but it will require new opening new CQs.

triller-telekom:
I feared that this is the case but wanted to ask anyway ;)

One more thing I just noticed while switching branches: You should define a ".gitignore" for the auto-generated files:

```
node/
npm_cache/
web/smarthome.css
web/smarthome.js
```

resetnow:
Updated.

PR #4067, opened 2017-08-16
factored communication out of ThingManager into own component

sjka:
...in order to separate this concern from the rest of the diverse duties of the ThingManager.

Please note: the logic is simply moved from the ThingManager into the CommunicationManager as it is. There is no point in nitpicking it as I'm currently reworking it completely (will be in another PR).

### DIFF
--- a/bundles/core/org.eclipse.smarthome.core.thing/src/main/java/org/eclipse/smarthome/core/thing/internal/CommunicationManager.java
+++ b/bundles/core/org.eclipse.smarthome.core.thing/src/main/java/org/eclipse/smarthome/core/thing/internal/CommunicationManager.java
@@ -1,0 +1,208 @@
+/**
+ * Copyright (c) 2014-2017 by the respective copyright holders.
+ * All rights reserved. This program and the accompanying materials
+ * are made available under the terms of the Eclipse Public License v1.0
+ * which accompanies this distribution, and is available at
+ * http://www.eclipse.org/legal/epl-v10.html
+ */
+package org.eclipse.smarthome.core.thing.internal;
+
+import java.util.Set;
+import java.util.concurrent.TimeoutException;
+
+import org.eclipse.smarthome.core.common.SafeMethodCaller;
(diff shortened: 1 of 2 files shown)


bundles/core/org.eclipse.smarthome.core.thing/src/main/java/org/eclipse/smarthome/core/thing/internal/InboundCommunication.java
```diff
@@ -0,0 +1,34 @@
...
+ * It is tailored to handle the inbound communication, i.e. state updates, commands and triggers coming from the
+ * binding. It allows the {@link ThingManager} to delegate such aspects of the {@link ThingHandlerCallback} internally
+ * to the {@link CommunicationManager}.
+ *
+ * @author Simon Kaufmann - initial contribution and API
+ *
+ */
+public interface InboundCommunication {
+
+    void stateUpdated(ChannelUID channelUID, State state);
```
kaikreuzer: As this is an interface, could you please add a short JavaDoc to the methods as well?

bundles/core/org.eclipse.smarthome.core.thing/src/main/java/org/eclipse/smarthome/core/thing/internal/InboundCommunication.java
```diff
@@ -0,0 +1,34 @@
...
+package org.eclipse.smarthome.core.thing.internal;
+
+import org.eclipse.smarthome.core.thing.ChannelUID;
+import org.eclipse.smarthome.core.thing.Thing;
+import org.eclipse.smarthome.core.thing.binding.ThingHandlerCallback;
+import org.eclipse.smarthome.core.types.Command;
+import org.eclipse.smarthome.core.types.State;
+
+/**
+ * This interface provides (internal) access to the {@link CommunicationManager}.
```
maggu2810: Should an interface knows it implementation class?
If it is only about injecting the communication manager to the thing manager, why using an interface at all?

PR #4075, opened 2017-08-17
added autoApprove feature for Inbox (moved from openhab-core)

kaikreuzer:
In a second commit, I migrated all other DS components of the same bundle to annotations, so I would suggest to merge through rebase.


maggu2810:
Build failed...

maggu2810:
> suggest to merge through rebase

Can you merge
* added autoApprove feature for Inbox (moved from openhab-core)
* added documentation
and
* changed services to DS annotations
* fixed bug that cause unit tests to fail

then I can merge using rebase

kaikreuzer:
Let's say the separate commits were helpful for the review, but I guess it can be squashed on merge after all - since the changes are in different classes, it won't do much harm and resorting the commits to only partially squash them in a different order is a bit of a hassle that I'd prefer to avoid :-)

PR #4079, opened 2017-08-17
Fixed support for configurable background discoveries

kaikreuzer:
Since its inception, the AbstractDiscoveryService was designed in a way that background discovery should be configurable for each service. It turns out that hardly any binding correctly supports this - with this PR, I am trying to fix this for ESH bindings.

A few remarks:
- Similar to the binding configurations, which require the service pid to be `binding.<id>`, I have defined the service pids for (configurable) discovery services to be `discovery.<id>`.
- I changed the configuration parameter name from `backgroundDiscovery.enabled` to simply `background`, because it is a parameter for "discovery" and having dots in the parameter name is pretty unusual and rather confusing. Note that this is a breaking change - but as I believe that nobody uses it yet, I would say we should do it.
- I have added activate/modified methods where they didn't exist - according to DS specs, they must exist on the implementing class itself and not only on parent classes.
- I didn't migrate to DS annotations as this would have meant too much work in this context. I'd hope we find some other volunteer to work on that for the existing bindings.


maggu2810:
Hm, wouldn't it be possible to use a certain prefix for the PIDs (e.g. `esh`)?
It is like a "global namespace pollution" if we use ESH and other stuff in the same OSGi container.

kaikreuzer:
As I said, it is defined for bindings since long this way, so I am only following this.

You actually should consider "binding" and "discovery" as namespaces, which gather many different similar services. The main goal was here to NOT use ESH or solution specific prefixes, so that it is easily possible to move bindings to different namespaces (as has been done with LIFX, Astro, etc.) without breaking any compatibility.

kaikreuzer:
fixed failing test.

triller-telekom:
The change looks alright to me. But lets wait for travis.

The only annoying thing is that you now have to override the `modified` method and implement it with `super.modified()` which looks superfluous to me. But if the specs really require it...

maggu2810:
> according to DS specs, they must exist on the implementing class itself and not only on parent classes.

Really, I looked ad 4.2.0 and 6.0.0 compendium spec and don't agree.
The method needs to be accessible by the component implementation, so protected methods of a super class are a match.

kaikreuzer:
How do you add a `@Modified' annotation to a method that does not exist in your class?

maggu2810:
Sure, the annotation needs to be present in the implementation class and so also the annotated method.

> I have added activate/modified methods where they didn't exist - according to DS specs, they must exist on the implementing class itself and not only on parent classes.

added ... methods where "they" ... "they" must exist...

If the second "they" is about the DS annotations I agree.

kaikreuzer:
> If the second "they" is about the DS annotations I agree.

If it helps that you agree, then yes, that's what it was about ;-)
And yes, at closer inspection, you are right that the spec only expects them to be present in the class hierarchy.

mhilbush:
@kaikreuzer I had relied on the following in `runtime.cfg` to allow the user to disable background discovery for the globalcache binding.  

```
org.openhab.binding.globalcache.discovery.GlobalCacheDiscoveryService:backgroundDiscovery.enabled=false
```

It is now broken as a result of this PR.  The problem is that I'm not sure what I need to change in the globalcache binding.  As part of this PR, you had made changes to the *Discovery.xml files; however, since https://github.com/openhab/openhab2-addons/commit/52f768a3f8f80ac30d98dfb286bac134116baf99, those files are no longer there? 

What would be the correct way for me to restore this functionality to the globalcache binding?

triller-telekom:
This [line](https://github.com/openhab/openhab2-addons/blob/master/addons/binding/org.openhab.binding.globalcache/src/main/java/org/openhab/binding/globalcache/internal/discovery/GlobalCacheDiscoveryService.java#L42) needs these 2 additional parameters:

```
modified="modified" name="discovery.globalcache"
```

and you should override `modified`

```
@Override
protected void modified(Map<String, Object> configProperties) {
       super.modified(configProperties);
 }
```

The `.xml` files are gone because we we are using DS-annotations now, like the `@Component` in the line referenced above. Out of these annotations the `*.xml` files will be auto-generated.

mhilbush:
Thanks!

mhilbush:
@triller-telekom I'm getting around to fixing this in the globalcache binding, and the above doesn't seem to work.

`modified="modified"` is flagged by the IDE as an undefined attribute.

`name="discovery.globalcache"` causes an exception to be thrown. org.osgi.service.component.ComponentException: The component name 'discovery.globalcache' has already been registered by Bundle 242 (org.openhab.binding.globalcache) as Component of Class org.openhab.      binding.globalcache.internal.discovery.GlobalCacheDiscoveryService

Here's what I've found to work.  I'd like to get some feedback whether this is correct.
```
@Component(service = DiscoveryService.class, immediate = true, configurationPid = "discovery.globalcache")
```
It also appears that the modified method is not needed.  It works with and without that method.  Upon changing the configuration parameter, rather than modified being called, the discovery service is deactivated then activated.

And, so I can get the documentation correct, in runtime.cfg, I would add the following line to disable background discovery.
```
discovery.globalcache:background=false
```

triller-telekom:
What you did is correct, i.e.
```
@Component(service = DiscoveryService.class, immediate = true, configurationPid = "discovery.globalcache")
```
See the [Astro Binding](https://github.com/eclipse/smarthome/blob/master/extensions/binding/org.eclipse.smarthome.binding.astro/src/main/java/org/eclipse/smarthome/binding/astro/internal/discovery/AstroDiscoveryService.java#L44) as an example which does it in the same way.

I do not know why there is an exception thrown if you add the "name" parameter, however we are using "configurationPid" now, so what you did is correct.

I was wring with adding `modified="modified"`, this is only the case for writing the xml file, in our case with DS annotations you have to put the `@Modified` annotation on your `modify` method.

sjka:
> modified="modified" is flagged by the IDE as an undefined attribute.

There indeed is no `modified="modified"` property - it should be an annotation like this:

```java
@Method
protected void modified(...) {
    ...
}
```

kaikreuzer:
>  I was too slow

And wrong as well, as the annotation is `@Modified` (as mentioned by @triller-telekom) and not `@Method` 😛

mhilbush:
Thanks!  `@Modified` seems to have done the trick.  :-)

A couple additional findings:

- When you add the `@Modified` annotation, the IDE also adds `@Override`.  So, you actually end up with:
```
@Override
@Modified
protected void modified(Map<String, Object> configProperties) {
    super.modified(configProperties);
}
```

- It appears the Astro binding is incorrect, as it does not include the `@Modified` annotation.
https://github.com/eclipse/smarthome/blob/master/extensions/binding/org.eclipse.smarthome.binding.astro/src/main/java/org/eclipse/smarthome/binding/astro/internal/discovery/AstroDiscoveryService.java#L68
```
@Override
protected void modified(Map<String, Object> configProperties) {
    super.modified(configProperties);
}
```

triller-telekom:
@mhilbush So you needed the `@Modified` annotation because without it the method was not called?

From the [OSGi specification](https://osgi.org/download/r4v42/r4.cmpn.pdf) under point 112.5.12. it says

> The name of the modified method is specified by the modified attribute. See
Component Element on page 311. The prototype and selection priority of the
modified method is identical to that of the activate method. See Activate
Method on page 320.

whereas under 112.5.8 for the `activate` method this is written:

> A component instance can have an activate method. The name of the activate
method can be specified by the activate attribute. See Component Element
on page 311. If the activate attribute is not specified, the default
method name of activate is used. 

Note the "can" in the last quote, i.e. "The name of the activate
method **can** be specified", and for the `modified` method there is no such "can", i.e. I am not sure what the default now is.

So what was the outcome of your experiments, did you need the annotation for `modified` to make it work?

mhilbush:
> So you needed the @Modified annotation because without it the method was not called?

Correct. Without the `@Modified` annotation, when I changed the parameter to enable/disable background discovery, the binding was deactivated then activated.  Once I added `@Modified` the `modified` method was called.

triller-telekom:
@mhilbush Thanks, I have created #4967 to fix it for the astro discovery service.

PR #4082, opened 2017-08-17
fix resource leak

maggu2810:
The datagram socket must be closed.


kaikreuzer:
Err, this class is actually dead code and should be removed. The hue binding uses an UPnPDiscoveryParticipant for discovery and not this class (which came in from the formerly external library).

Are you ok to remove it or do you actually use it?

maggu2810:
I don't use it. I just realized that there is a resource leak. If we could remove the whole class without loosing functionality this is okay for me to.

maggu2810:
done

kaikreuzer:
Much better - this change is definitely free of bugs ;-)

PR #4086, opened 2017-08-17
fixes #4074

### LINKED ISSUE #4074: Changing BaseThingHandler bridgeStatusChanged behaviour
Currently this method works as follows:

bridgeStatus changes from ONLINE to OFFLINE: thingStatus is set to OFFLINE, thingStatusDetail to BRIDGE_OFFLINE
bridgeStatus changes from OFFLINE to ONLINE: checks whether the thingStatusDetail is BRIDGE_OFFLINE and if so, sets the thing Status to ONLINE

On my opinion this is wrong. If the thing was OFFLINE (e.g. because of mis-configuration) when the bridge goes offline, the thingStatusDetail (here: CONFIGURATION_ERROR) is lost, because it is set to BRIDGE_OFFLINE. When the bridge goes back online, the thingStatus changes to ONLINE, even if the reason for the thing to be OFFLINE in the first place is still there.

My suggestion is to change it so that the methods checks whether the thing is already OFFLINE and only set thingStatusDetail to BRIDGE_OFFLINE if the reason for the thing to go OFFLINE is indeed the bridge changing its status.

J-N-K:


bundles/core/org.eclipse.smarthome.core.thing/src/main/java/org/eclipse/smarthome/core/thing/binding/BaseThingHandler.java
```diff
@@ -582,10 +582,10 @@ protected boolean isInitialized() {
 
     @Override
     public void bridgeStatusChanged(ThingStatusInfo bridgeStatusInfo) {
-        if (bridgeStatusInfo.getStatus() == ThingStatus.ONLINE
-                && getThing().getStatusInfo().getStatusDetail() == ThingStatusDetail.BRIDGE_OFFLINE) {
-            updateStatus(ThingStatus.ONLINE, ThingStatusDetail.NONE);
-        } else if (bridgeStatusInfo.getStatus() == ThingStatus.OFFLINE) {
+        if (ThingStatus.ONLINE.equals(bridgeStatusInfo.getStatus())
```
maggu2810: Hm, my understanding has been that the base implementation should look like:

```Java
if (bridgeStatusInfo.getStatus() == ThingStatus.OFFLINE) {
    updateStatus(ThingStatus.OFFLINE, ThingStatusDetail.BRIDGE_OFFLINE);
}
```

J-N-K: Correct. This needs to be updated (and of course the test, too). Will do tomorrow.

kaikreuzer:
I would say that it is ok to call super, if we make sure (and document) that the implementation in the BaseThingHandler only deals with the OFFLINE status.

The rest of the implementation should hence deal only with other situations (and if it wants to do any additional action upon a change to OFFLINE, it is still allowed to do so).

extensions/binding/org.eclipse.smarthome.binding.hue/src/main/java/org/eclipse/smarthome/binding/hue/handler/HueLightHandler.java
```diff
@@ -486,4 +487,14 @@ private int getAlertDuration(Command command) {
 
         return delay;
     }
+
+    @Override
+    public void bridgeStatusChanged(ThingStatusInfo bridgeStatusInfo) {
+        super.bridgeStatusChanged(bridgeStatusInfo);
+        if (ThingStatus.ONLINE.equals(bridgeStatusInfo.getStatus())
+                && ThingStatus.OFFLINE.equals(getThing().getStatusInfo().getStatus())
```
kaikreuzer: Do we really have to check all of this here or wouldn't it be enough to (always) act if the bridge switched to ONLINE?

extensions/binding/org.eclipse.smarthome.binding.hue/src/main/java/org/eclipse/smarthome/binding/hue/handler/HueLightHandler.java
```diff
@@ -486,4 +487,14 @@ private int getAlertDuration(Command command) {
...
+                && ThingStatusDetail.BRIDGE_OFFLINE.equals(getThing().getStatusInfo().getStatusDetail())) {
+            updateStatus(ThingStatus.ONLINE, ThingStatusDetail.NONE);
```
kaikreuzer: As mentioned in https://github.com/eclipse/smarthome/issues/4074#issuecomment-323281682, I would claim that setting the status ONLINE is many cases a bug. In the example of the hue binding, the handler determines the correct status in `onLightStateChanged`. So either we do not set a state here at all and wait for the next update (triggered through the regular polling) or we set it to UNKNOWN for the time being.

Same should be checked for the other bindings.

extensions/binding/org.eclipse.smarthome.binding.hue/src/main/java/org/eclipse/smarthome/binding/hue/handler/HueLightHandler.java
J-N-K: What if the thing state changed after it was set to offline because of the bridge going offline?

extensions/binding/org.eclipse.smarthome.binding.hue/src/main/java/org/eclipse/smarthome/binding/hue/handler/HueLightHandler.java
J-N-K: I checked that and changed the PR.

J-N-K:
REMOVING should possibly result in OFFLINE/BRIDGE_OFFLINE as well. It's difficult to say what happens for UNKNOWN, I would leave it as it is.

J-N-K:
I think it is necessary to adjust the openhab2-addons, too, otherwise they won't work anymore. I'm preparing the changes to restore the current situation (i.e. turning the thing ONLINE if the bridge goes ONLINE), but I'm not able to investigate each binding if this is the perfectly correct way. (see [OH2 2654](https://github.com/openhab/openhab2-addons/pull/2654).

kaikreuzer:
>  I'm not able to investigate each binding if this is the perfectly correct way.

Absolutely clear. Thanks for your help!

extensions/binding/org.eclipse.smarthome.binding.hue/src/main/java/org/eclipse/smarthome/binding/hue/handler/HueLightHandler.java
kaikreuzer: Can this happen? We said that the only valid status for the Thing is OFFLINE.BRIDGE_OFFLINE while the bridge is offline. So this here should be the first moment when this changes.

J-N-K: That is true. What happens if the bridge is OFFLINE for some reason and the thing goes OFFLINE.BRIDGE_OFFLINE, too. Then the user changes the configuration of the thing and sets an invalid configuration. The thing status might change to OFFLINE.CONFIGURATION_ERROR.

J-N-K:
retriggering build

extensions/binding/org.eclipse.smarthome.binding.hue/src/main/java/org/eclipse/smarthome/binding/hue/handler/HueLightHandler.java
kaikreuzer: Valid remark, this is all nasty :-/
But if we say that it is ok to have the Thing being OFFLINE for another reason than BRIDGE_OFFLINE when a bridge is OFFLINE, this also means that the BaseThingHandler implementation should only switch to OFFLINE.BRIDGE_OFFLINE, if it isn't OFFLINE for another reason already - BUT if this other reason is "COMMUNICATION_ERROR" changing it to BRIDGE_OFFLINE would imho be the correct thing, while OFFLINE.CONFIGURATION_ERROR should stay. I really don't know how to solve this best...

@maggu2810 & @sjka Any good ideas from you?

maggu2810: * Bridge is offline
* Thing is offline
* Thing Status detail is ... (e.g. bridge offline or communication error or whatever)
* Bridge status changes to offline

Wouldn't it make sense (because it should succeed in the most cases) to handle the status transition similar as on initialization?
So, e.g. check configuration and if it does not fit, set offline:communication error, ...
I assume initialize and handling the bridge status change from offline to online could share some code.

I don't think we could rely on one status detail is more prioritized then the other one in offline state.

kaikreuzer: Ok, so in @J-N-K's example above, it would mean that even if the status might have changed due to an invalid configuration, we would call the very same logic here, which simply should in any case re-check the validity of the configuration and do a best-effort approach to determine the current online status.
The result would be a more complex implementation than the current one (here in the HueLightHandler), but a more consistent one across bindings, correct?

maggu2810: I think so.
Does it need more complex in the hue binding? Do we break anything if we simply call `void initializeThing(ThingStatus bridgeStatus)`?

J-N-K: The more I think about it, probably BaseThingHandler should be abstract and each binding should find the correct way how to deal with bridgeStatusChanged.

martinvw: just curious is there a reason to go for equals instead of == for enums I would personally always prefer ==

maggu2810: I don't think there is a reason but personal preference.
Enums are singletons so there shouldn't be any difference.

kaikreuzer: > Does it need more complex in the hue binding?

Yes, because at the moment, it simply [switches to ONLINE](https://github.com/eclipse/smarthome/pull/4086/files/feb025795bec72ddec485c3c999ff1a60ac69d9e#diff-2ba4d1bdb2c89e2c8b66b342a88ed23eR497) and does not check whether it has a valid configuration, nor if it is (still) known/reachable on the hue bridge.

> Do we break anything if we simply call void initializeThing(ThingStatus bridgeStatus)?

For the Hue binding, this would probably be exactly the right thing to do. But this method only exists in that handler, so we cannot generalise it - we can only point to it as a good way to deal with it. But it means we have not formalized it and as such we are prone to implementation errors by binding developers.

> The more I think about it, probably BaseThingHandler should be abstract and each binding should find the correct way how to deal with bridgeStatusChanged.

Maybe that is the solution in the end.
@sjka Would appreciate your opinion here as well - maybe you have a good idea that nobody else thought of, who knows...

maggu2810: > For the Hue binding, this would probably be exactly the right thing to do. But this method only exists in that handler, so we cannot generalise it - we can only point to it as a good way to deal with it. But it means we have not formalized it and as such we are prone to implementation errors by binding developers.

My comment has been only about the Hue binding.

J-N-K: @kaikreuzer, `abstract` might not be a good idea, that way even binding developers that don't use brigdes at all would need to implement it. 

So, how do we go on from here?

J-N-K:
do we agree that there should not be any implementation at all in the BaseThingHandler?

J-N-K:
Where do we stand here? I'm willing to update the PR, but there should be a definite decision.

triller-telekom:
I just read through all comments of this PR and issue #4074. I guess that the summary of it is in @kaikreuzer's [post](https://github.com/eclipse/smarthome/issues/4074#issuecomment-324594265)

> When the bridge goes offline, the reason for the Thing to be offline is clearly the bridge - so setting it to OFFLINE.BRIDGE_OFFLINE is the only correct thing to do. Now if the bridge goes online again, we have to do a status transition to a correct status for the thing - we should not do guesswork here. So [...]  we leave it entirely to the binding to find the right current status.

This means that we effectively remove from the current implementation the part that sets the thing to `ONLINE` after the bridge goes online.

@J-N-K Would you like to have a look on how many ESH and OH bindings would be affected? I guess that would only be those that have a bridge and simply call `super()` in this method, right?

J-N-K:
The build failed in PaperUI, I'll retrigger it if @kaikreuzer confirms that this is the correct implementation. I think all ESH bindings are covered by this PR, I'll check the OH2 bindings after this is confirmed. Please do not merge before OH2 bindings are ready to be updated.

triller-telekom:
@J-N-K  Thanks for your effort on continuing this PR!

To me the code now looks like what we discussed above, let's wait for @kaikreuzer to confirm what I quoted from him and then you can adjust the OH2 bindings.

J-N-K:
@kaikreuzer, if you find some time, please comment. Thank you.

PR #4097, opened 2017-08-20
ClassicUI: add favicon and webapp manifest

resetnow:


resetnow:
You probably have seen this but just in case there is an accompanying PR in core: https://github.com/openhab/openhab-core/pull/198

kaikreuzer:
Yes, thanks - just waiting to do an ESH stable build before merging the change in openhab-core.

Issue #3898, opened 2017-07-27
Replace ESH Designer with LSP


sjka:
As discussed in https://dev.eclipse.org/mhonarc/lists/smarthome-dev/msg00144.html, we should get rid of the Eclipse SmartHome Designer and replace it with an LSP interface which can be used by arbitrary editors.

This includes:

* An update to Xtext 2.12
* Generating the required LSP artifacts within the build
* Exposing the generated language server(s) from the runtime
* Dropping all editor related bundles from eclipse/smarthome repository and the entire eclipse/smarthome-designer repository

...with all the nitty-gritty details that will be faced when implementing it.

Issue #1402, opened 2016-04-19
Use Xtext web as a potential replacement for the Designer


kaikreuzer:
Since Xtext 2.9.2, there is an initial support for web-based DSL editors, see https://www.youtube.com/watch?v=QtW6WBJztPw.

This would be a huge gain over the current Designer, since
- it does not require to build and maintain an RCP application for different platform
- there is no problem with transferring the file from a desktop to the runtime instance (see https://github.com/eclipse/smarthome/issues/619)
- the editor has direct access to all installed add-ons (e.g. rule actions), so it is easy to solve https://github.com/eclipse/smarthome/issues/997

kaikreuzer:
@chriskn told me that he successfully prototyped this feature and it seems as if there should be no blockers regarding the implementation of such a feature for the runtime.

A remaining question is, where this web editor should be built in. There are imho several options:

1. Build a new simple UI specifically for the textual configuration support.
Pros: easy to provide as an option, UI can be designed specifically for that purpose (with a file explorer, etc.)
Cons: Yet another UI, which might duplicate some features from other UIs and which makes it more difficult for the users to know what to do where

2. Add it to the Paper UI
Pros: It is the existing "admin" UI in ESH and thus the most natural place
Cons: Paper UI is designed for "touch"/mobile support, while text editing really only makes sense for PC use. Paper UI is meant to be "graphical/easy" administration, the textual files are so far the "non UI option" for the system configuration, so from a functional view, it does not really fit to the Paper UI. Furthermore, it isn't easily possible to add it as an optional feature, hence it has to be "baked in" and the code would be within the Paper UI even for solutions that do not support any textual configuration.

3. Only offer the runtime side and leave the UI to the solution
For openHAB, I could e.g. imagine that support is added in HABmin, but this is nothing that can be moved to ESH directly.
Pros: No decision needed at ESH
Cons: No UI available in ESH to demo this feature (possibly (1) could be done in a simplistic way to mitigate this issue)

What do you think? @cdjackson, what is your opinion regarding a possible HABmin integration?

maggu2810:
I agree with your cons for point 2.

For point 1 and 3 this means, offer the runtime side for both, but differ in the UI (no one for 1, add one for 3).
I would prefer to add a **simply** one (or if someone have time, it could be as good as possible) but at least one, that could be used as a showcase what is possible.
Other solutions are free to not use the simply one and add a more extended version (e.g. as part of HABmin, etc.).

[...]

PR #4100, opened 2017-08-21
Xtext upgrade

sjka:
This contributes towards #3898 and #1402: It upgrades Xtext to the latest release version. This however entails some additional upgrades due to Xtext's minimal requirements:

* Xtext 2.12
* Eclipse 4.7 (Oxygen)
* Guava 15 (requires PB CQ)

Build, runtime and designer still seem to work well for me, so the upgrade was pretty smooth. However, as the equinox upgrade it already uncovered #4091 and #4092, I'm happy for feedback if there are any more surprises related to this...


kaikreuzer:
Could you check what needs to be done in https://github.com/eclipse/smarthome/blob/master/features/karaf/esh-tp/src/main/feature/feature.xml#L167-L205 (maybe with some help of @maggu2810, who had some libs repackaged for the current version)?

kaikreuzer:
I successfully tested this and it all looks good to me.
I have added https://github.com/eclipse/smarthome/pull/4100/commits/79a486fbca45b308b41e3dfdb853e447848b0c37, which also updates the Karaf feature. I even built an openHAB distro with this and all looked good.

@maggu2810 Please also have a look, especially at my changes to the Karaf feature and let me know in case I did anything wrong there.

sjka:
As smooth as it goes until now, I created CQ 14019 for Guava 15

sjka:
FTR: The CQ got approved.

kaikreuzer:
Cool, let's go for it and see what breaks ;-)

Issue #2226, opened 2016-09-29
Trigger Channels Part 2


kaikreuzer:
With https://github.com/eclipse/smarthome/issues/1043, we have introduced trigger channels to the framework, but as already mentioned early in https://github.com/eclipse/smarthome/issues/1043#issuecomment-184598078 that was only a first step for the basic functionality.

Imho, we are left with a dangerous situation now, if people start using it although what they would really need for their purposes are the profile, which I'd like to call part 2.

In order to discuss about how to design this second part, I would like to take [the LeapMotion controller as an example](https://www.youtube.com/watch?v=czq3S13tGVk) for a triggering input device. As can be seen in the video, it can be e.g. used to control Color items, i.e. it sends ON/OFF, percentage values for dimming and HSB values for the colors. The raw events (and thus what we are currently left with) of this device are not really suitable. There are hundreds of these events per second and they are very generic like "4 fingers at position x,y,z" or "circle to right with radius x cm". Of course one could say that such raw events are emitted and you would have to implement rules, which then try to convert them to some meaningful commands. But this obviously does not make much sense.

Instead, the binding itself should have a chance to define configuration options and depending on those configuration options it could then directly provide item commands such as ON/OFF, so that such a trigger channel could be directly linked to an existing item (which e.g. represents a light and is linked to a hue bulb).

markusjw:
If I understand your last paragraph correctly, it matches my opinion. There are two more or less separate things:

1. To link trigger channels to items, we might need trigger channel types that match the item types. A thing (let's say a rocker switch) can now implement two trigger channels (predefined type system.button) for the two buttons. But at the same time, it could also implement a "switch trigger channel" with on/off functionality, combining both buttons/positions of the rocker switch.

2. The mapping of hand gestures to trigger channels is specific to the Leap Motion binding. Or are there generic concepts needed for other bindings as well?

kaikreuzer:
Regarding 1:
- For a rocker, I would rather prefer a single channel with events like up_pressed, down_pressed, released - since they are not really independent (you cannot press both sides at the same time).
[...]

Issue #1043, opened 2016-02-15
Introduce trigger channels for things, which allow modelling of buttons and remote controls


kaikreuzer:
Currently, all channels on a Thing are supposed to be linked to an item and thus they inherently carry some state and are a potential entity to be displayed in the UIs.

For wall-buttons, remote controls, etc. this usually does not make sense. Their buttons rather represent the very same as switch widget on a UI, i.e. they are just a mean to operate some item (e.g. a light bulb).
To properly model this, we should introduce a new kind of channel. The idea is to be able to use such new channels directly as triggers for rules (without the need to have any item/link) for them. Alternatively (this is how openHAB 1 works), links could be created to existing items (which would be the equivalent of wiring the button to a lightbulb - the trigger channel must send commands to the item in this case).

maggu2810:
What differs such channels from other ones?
So, if I can use a channel (a wall-button) as a trigger, why do I need to create an item so use a light channel as an output?
I do not know, if we open using channels directly where we stop?
Using channels as a source but not sink, make that sense?

Could you give more details how you think this could be modelled?

Will a channel emit commands or state updates?
Will a channel consume commands or state updates?

kaikreuzer:
> What differs such channels from other ones?

The main difference is that they do not have a state, but only issue events.

> why do I need to create an item so use a light channel as an output?

You do not, that's the idea of using it as a trigger for rules only. The option to link them to an existing "light" item would actually be something like a "simple, hard-wired rule", which does not have to go through the rule engine itself.

> Using channels as a source but not sink, make that sense?

Yes, this is one way you could think of them.

> Will a channel emit commands or state updates?
> Will a channel consume commands or state updates?

I think in the first place, they will only send events like PRESSED, LONG_PRESSED, DOUBLE_PRESSED, RELEASED or something like this (to be specified). These would be the "output" values of the rule trigger module.
For the option to directly wire such a channel to an existing item, such a link would need some configuration (I like speaking of certain "profiles"), which convert the events like PRESSED to according commands like ON or UP.

> Could you give more details how you think this could be modelled?

We will have to discuss this together - I only wanted to capture the requirement in a first step.
I think it would need something like an "eventDescription", which is similar to the "stateDescription" of the regular channels.

[...]

PR #4108, opened 2017-08-22
Introduction of Profiles

sjka:
This PR now reworks the factored CommunicationManager (#4067) and makes its behavior modifiable by "profiles". Such profiles are basically exchangeable parts which the communication manager delegates its actual duties to. I also outlined their purpose in the profiles.md file below to become part of the documentation. 

The idea is to make the used profile configurable on ItemChannelLinks. In this first step the profile selection is more or less static though. Although I already anticipated a "ProfileFactory" interface (mainly for testing purposes here), this PR is still missing the configurability by purpose. As this will involve touching many files and the Thing-DSL, I would like to handle this in a separate follow-up PR. 

Using profiles also opens up another interesting use-case: I noticed that so far no binding implements the `ThingHandler.handleUpdate(...)` method. Technically spoken some very few do, but they don't do anything there. The main idea of this method apparently was to allow devices to run as "slaves" of other devices, i.e. react on item updates only. A typical use-case for this would be a light switch with an indicator light which should follow the state of the the physical lamp without the risk of running out-of-sync. 
Therefore I already added a "slave" profile which does exactly that: it listens to item updates, transforms them to commands (if possible) and forwards them to the `ThingHandler.handleCommand(...)` directly. At the same time it cuts off all other communication coming out of the thing handle. With this it would be easily possible to run _any_  device now as a slave without the need to fix/implement this in all of the existing bindings. 

At the same time I deprecated the `ThingHandler.handleUpdate(...)` method to prevent any new binding to implement it. I think it is fair to allow for a grace period before we drop this method from the interface.

The profiles here (especially the TriggerProfiles) are required for the second part of #1043 a.k.a. #2226. I only created a RawButtonTriggerProfile here as an example and proof-of-concept. But there are more required, of course. Especially for the system channels they make much sense! Nevertheless, I would like to keep the work of implementing these out this PR as well.


davidgraeff:
Hm, at some points I find it very hard to follow the ESH documentation and the newly added profiles.md is one of those sections that would highly confuse me. I think this is mainly because there are no diagrams that provide an overview with whom and of how profiles are interacting with other components and there are no example usages.

sjka:
I agree, the docu is not yet complete. Examples can be added only once it actually can be configured. As this part is yet missing, there is not much to put there right now. I will add some pictures for the concrete profiles once all this here somehow got agreed. I hope it will get clearer then. For the time being I just thought it's important to write it down anywhere at all. 

For all the other confusing parts: Fully agreed. Any help is highly appreciated!

bundles/core/org.eclipse.smarthome.core.thing/src/main/java/org/eclipse/smarthome/core/thing/internal/profiles/DefaultSlaveProfile.java
```diff
@@ -0,0 +1,64 @@
...
+ * turn {@link State} updates into {@link Command}s (if possible) and then forwards those to the {@link ThingHandler}.
+ * <p>
+ * This allows devices to be operated as "slaves" of another one directly, without the need to write any rules.
+ * <p>
+ * No events are send for state updates or commands initiated by the ThingHandler. They are assumed to be passive only.
+ *
+ * @author Simon Kaufmann - initial contribution and API.
+ *
+ */
+public class DefaultSlaveProfile extends DefaultMasterProfile {
```
kaikreuzer: Why does it extend the master profile? Shouldn't it rather `implements StateProfile` as well?

bundles/core/org.eclipse.smarthome.core.thing/src/main/java/org/eclipse/smarthome/core/thing/internal/profiles/DefaultMasterProfile.java
```diff
@@ -0,0 +1,99 @@
...
+import org.eclipse.smarthome.core.types.Command;
+import org.eclipse.smarthome.core.types.State;
+import org.slf4j.Logger;
+import org.slf4j.LoggerFactory;
+
+/**
+ * This is the default profile for stateful channels.
+ *
+ * It forwards commands to the {@link ThingHandler}. In the other direction it posts events to the event bus
+ * for state updates and commands which are initiated by the handler.
```
kaikreuzer: "...and commands which are initiated by the handler" - was that really a use case for a "master" device? Do you have an example?

bundles/core/org.eclipse.smarthome.core.thing/src/main/java/org/eclipse/smarthome/core/thing/internal/profiles/RawButtonTriggerProfile.java
```diff
@@ -0,0 +1,35 @@
...
+
+import org.eclipse.smarthome.core.events.EventPublisher;
+import org.eclipse.smarthome.core.items.Item;
+import org.eclipse.smarthome.core.items.events.ItemEventFactory;
+import org.eclipse.smarthome.core.library.types.OnOffType;
+import org.eclipse.smarthome.core.thing.CommonTriggerEvents;
+import org.eclipse.smarthome.core.thing.link.ItemChannelLink;
+import org.eclipse.smarthome.core.thing.profiles.TriggerProfile;
+
+public class RawButtonTriggerProfile implements TriggerProfile {
```
kaikreuzer: JavaDoc missing

bundles/core/org.eclipse.smarthome.core.thing/src/main/java/org/eclipse/smarthome/core/thing/internal/profiles/DefaultSlaveProfile.java
```diff
@@ -0,0 +1,64 @@
...
+
+/**
+ * This is the default implementation for a slave profile.
+ *
+ * In contrast to the {@link DefaultMasterProfile} it does not forward any commands to the ThingHandler. Instead, it
+ * turn {@link State} updates into {@link Command}s (if possible) and then forwards those to the {@link ThingHandler}.
+ * <p>
+ * This allows devices to be operated as "slaves" of another one directly, without the need to write any rules.
+ * <p>
+ * No events are send for state updates or commands initiated by the ThingHandler. They are assumed to be passive only.
```
kaikreuzer: typo: are sent

kaikreuzer: Sorry if I forgot the details again: Taking the KNX wall switch as an example. Would this have to be modelled as two channels? One with master-profile only sending commands and one with slave-profile only receiving updates? What spoke against allowing slave-channels to send commands (to their "master" within ESH)?

bundles/core/org.eclipse.smarthome.core.thing/src/main/java/org/eclipse/smarthome/core/thing/internal/profiles/DefaultMasterProfile.java
sjka: Really good point - I just mindlessly took that from the old API. Now thinking about it, I don't have a good example. 

No binding in ESH is using it. 

In openhab2-addons there are only two bindings using it: [lutron](https://github.com/openhab/openhab2-addons/blob/50f7f662b4e5cf9f99b0c9a3e64f7097357d17e4/addons/binding/org.openhab.binding.lutron/src/main/java/org/openhab/binding/lutron/handler/SwitchHandler.java#L75) and [squeezebox](https://github.com/openhab/openhab2-addons/blob/52f768a3f8f80ac30d98dfb286bac134116baf99/addons/binding/org.openhab.binding.squeezebox/src/main/java/org/openhab/binding/squeezebox/handler/SqueezeBoxPlayerHandler.java#L395). I'm not familiar with both yet, so it's hard to judge whether this in fact is correct or not. On a first glance it looks like it might not be needed.

bundles/core/org.eclipse.smarthome.core.thing/src/main/java/org/eclipse/smarthome/core/thing/internal/profiles/DefaultSlaveProfile.java
sjka: I just was trying to prevent code duplication by delegating slave's `onUpdate(...)` to master's `onCommand(...)`. But I see this is confusing. I will change that.

bundles/core/org.eclipse.smarthome.core.thing/src/main/java/org/eclipse/smarthome/core/thing/internal/profiles/DefaultSlaveProfile.java
sjka: > What spoke against allowing slave-channels to send commands (to their "master" within ESH)?

Nothing really, IIRC. 

If a slave issues a command towards ESH, it will have to live with the fact that the state bounces back as an update via the master (which then gets turned into a command again by the slave profile, this time flowing from ESH towards the binding). But that shouldn't matter much - it rather makes sense because it gets the _real_ value that the master reported back as an update, which is not necessarily identical to the command that the slave initially sent towards ESH.

bundles/core/org.eclipse.smarthome.core.thing/src/main/java/org/eclipse/smarthome/core/thing/internal/profiles/DefaultMasterProfile.java
sjka: PS: Just do double-check I read you correctly: This is not only about the JavaDoc, but you are suggesting to make master's "postCommand" a no-op. Correct?

kaikreuzer: Correct.

kaikreuzer:
There are 3 test failures:
```
  ThingManagerOSGiTest.ThingManager handles post command correctly:804->OSGiTest.waitForAssert:-1->OSGiTest.waitForAssert:191->OSGiTest.waitForAssert:220 
Expected: not null
     but: was null
  DefaultSlaveProfileTest.testPostCommand:117 
No interactions wanted here:
-> at org.eclipse.smarthome.core.thing.internal.profiles.DefaultSlaveProfileTest.testPostCommand(DefaultSlaveProfileTest.java:117)
But found this interaction:
-> at org.eclipse.smarthome.core.thing.internal.profiles.DefaultSlaveProfile.postCommand(DefaultSlaveProfile.java:92)
Actually, above is the only interaction with this mock.
  DefaultMasterProfileTest.testPostCommand:121 
Wanted but not invoked:
eventPublisher.post(<Capturing argument>);
-> at org.eclipse.smarthome.core.thing.internal.profiles.DefaultMasterProfileTest.testPostCommand(DefaultMasterProfileTest.java:121)
Actually, there were zero interactions with this mock.
```

sjka:
My bad, I neglected to update the tests...
Done.

kaikreuzer:
Still one failing test :-(

bundles/core/org.eclipse.smarthome.core.thing/src/main/java/org/eclipse/smarthome/core/thing/internal/CommunicationManager.java
```diff
@@ -43,144 +58,223 @@
...
-                    logger.warn(
-                            "Cannot delegate command '{}' for item '{}' to handler for channel '{}', "
-                                    + "because no thing with the UID '{}' could be found.",
-                            command, itemName, channelUID, channelUID.getThingUID());
+    private Profile getProfile(ItemChannelLink link, Item item, Thing thing) {
+        if (thing == null) {
+            return new NoOpProfile();
+        }
+
+        Channel channel = thing.getChannel(link.getLinkedUID().getId());
```
maggu2810: So, the parameter `link` must not be null.
`ItemChannelLink link` => `@NonNull ItemChannelLink link`

bundles/core/org.eclipse.smarthome.core.thing/src/main/java/org/eclipse/smarthome/core/thing/internal/CommunicationManager.java
```diff
@@ -43,144 +58,223 @@
...
-                        logger.warn("Cannot delegate command '{}' for item '{}' to handler for channel '{}', "
-                                + "because no handler is assigned. Maybe the binding is not installed or not "
-                                + "propertly initialized.", command, itemName, channelUID);
-                    }
-                } else {
-                    logger.warn(
-                            "Cannot delegate command '{}' for item '{}' to handler for channel '{}', "
-                                    + "because no thing with the UID '{}' could be found.",
-                            command, itemName, channelUID, channelUID.getThingUID());
+    private Profile getProfile(ItemChannelLink link, Item item, Thing thing) {
```
maggu2810: We could mark the return type with `@NonNull`

bundles/core/org.eclipse.smarthome.core.thing/src/main/java/org/eclipse/smarthome/core/thing/internal/CommunicationManager.java
```diff
@@ -43,144 +58,223 @@
...
     }
 
-    @Override
-    protected void receiveUpdate(ItemStateEvent updateEvent) {
-        String itemName = updateEvent.getItemName();
+    private Profile getProfileFromFactories(ItemChannelLink link, Item item, Channel channel) {
+        for (ProfileFactory profileFactory : profileFactories.keySet()) {
+            Profile profile = profileFactory.createProfile(link, item, channel);
+            if (profile != null) {
+                profileFactories.get(profileFactory).add(link.getLinkedUID());
```
maggu2810: If you need access to the profileFactory, wouldn't it make sense to use `entrySet()` instead of `keySet()`?

bundles/core/org.eclipse.smarthome.core.thing/src/main/java/org/eclipse/smarthome/core/thing/internal/CommunicationManager.java
```diff
@@ -43,144 +58,223 @@
...
+                logger.trace("Going to use profile {} for link {}", profile, link);
+                return profile;
+            }
+        }
+        return null;
+    }
+
+    private Profile createDefaultProfile(Channel channel) {
```
maggu2810: `@NonNull Channel`

bundles/core/org.eclipse.smarthome.core.thing/src/main/java/org/eclipse/smarthome/core/thing/internal/CommunicationManager.java
```diff
@@ -43,144 +58,223 @@
...
+                    return new RawButtonTriggerProfile();
+                }
+                break;
+            default:
+                throw new NotImplementedException();
+        }
+        return null;
+    }
+
+    private void receiveCommand(ItemCommandEvent commandEvent) {
```
maggu2810: `@NonNull ItemCommandEvent`

bundles/core/org.eclipse.smarthome.core.thing/src/main/java/org/eclipse/smarthome/core/thing/internal/CommunicationManager.java
```diff
@@ -43,144 +58,223 @@
...
-                            "Cannot delegate update '{}' for item '{}' to handler for channel '{}', "
-                                    + "because no thing with the UID '{}' could be found.",
-                            newState, itemName, channelUID, channelUID.getThingUID());
+        final Item item = itemRegistry.get(itemName);
+
+        itemChannelLinkRegistry.stream().filter(link -> {
+            // all links for the item
+            return link.getItemName().equals(itemName);
+        }).filter(link -> {
+            // make sure a command event is not sent back to its source
```
maggu2810: "command" or "update" event?
Or is it about the default slave profile? If so, could you add it in the comment?

bundles/core/org.eclipse.smarthome.core.thing/src/main/java/org/eclipse/smarthome/core/thing/internal/CommunicationManager.java
```diff
@@ -43,144 +58,223 @@
...
+            ChannelUID channelUID = link.getLinkedUID();
+            Thing thing = getThing(channelUID.getThingUID());
+            Profile profile = getProfile(link, item, thing);
+            if (profile instanceof StateProfile) {
+                ((StateProfile) profile).onUpdate(link, thing, newState);
+            }
+        });
+    }
+
+    private void receiveTrigger(ChannelTriggeredEvent channelTriggeredEvent) {
```
maggu2810: `@NonNull ChannelTriggeredEvent`

bundles/core/org.eclipse.smarthome.core.thing/src/main/java/org/eclipse/smarthome/core/thing/internal/CommunicationManager.java
```diff
@@ -43,144 +58,223 @@
...
     }
 
     public void stateUpdated(ChannelUID channelUID, State state) {
-        Set<Item> items = itemChannelLinkRegistry.getLinkedItems(channelUID);
-        for (Item item : items) {
-            State acceptedState = ItemUtil.convertToAcceptedState(state, item);
-            eventPublisher
-                    .post(ItemEventFactory.createStateEvent(item.getName(), acceptedState, channelUID.toString()));
-        }
+        final Thing thing = getThing(channelUID.getThingUID());
```
maggu2810: `@NonNull ChannelUID channelUID`

bundles/core/org.eclipse.smarthome.core.thing/src/main/java/org/eclipse/smarthome/core/thing/internal/CommunicationManager.java
```diff
@@ -43,144 +58,223 @@
...
+            }
+        });
     }
 
     public void postCommand(ChannelUID channelUID, Command command) {
-        Set<String> items = itemChannelLinkRegistry.getLinkedItemNames(channelUID);
-        for (String item : items) {
-            eventPublisher.post(ItemEventFactory.createCommandEvent(item, command, channelUID.toString()));
-        }
+        final Thing thing = getThing(channelUID.getThingUID());
```
maggu2810: @NonNull ChannelUID channelUID

bundles/core/org.eclipse.smarthome.core.thing/src/main/java/org/eclipse/smarthome/core/thing/internal/profiles/RawButtonTriggerProfile.java
```diff
@@ -0,0 +1,44 @@
...
+import org.eclipse.smarthome.core.library.types.OnOffType;
+import org.eclipse.smarthome.core.thing.CommonTriggerEvents;
+import org.eclipse.smarthome.core.thing.link.ItemChannelLink;
+import org.eclipse.smarthome.core.thing.profiles.TriggerProfile;
+
+/**
+ * This profile allows a channel of the "system:rawbutton" type to be bound to an item.
+ *
+ * It reads the triggered events and uses the item's current state to calculate the new state once it detects that the
+ * button was pressed.
```
maggu2810: So, it is a toggle profile. Correct?

docs/documentation/concepts/profiles.md
```diff
@@ -0,0 +1,30 @@
...
+Trigger channels (`ChannelKind.TRIGGER`) by themselves do not maintain a state (as by their nature they only fire events). With the help of trigger profiles they can be linked to items anyway. Hence the main purpose of a trigger profile is to calculate a state based on the fired events. This state then is forwarded to the linked item by sending `ItemStateEvents`. 
+
+Trigger profiles are powerful means to implement some immediate, straight-forward logic without the need to write any rules. 
+
+Apart from that, they do not pass any commands or state updates to and from the thing handler as by their nature trigger channels are not capable of handling these.
+
+
+
+
+
```
maggu2810: Why do you add all that empty lines?

maggu2810:
Damn, merged while reviewing :wink:

kaikreuzer:
Sorry, I had put my approval 18 hours ago and the PR is 15 days old - so I was hoping you didn't have any further comments.

But never mind - I am sure @sjka will happily address everything in a follow-up PR :-)

maggu2810:
> I had put my approval 18 hours ago and the PR is 15 days old 

I am slow :wink:

sjka:
Sure, no problem

PR #4109, opened 2017-08-22
core/ConfigUtil: Do not normalize objectClass, component.name, component.id configuration values

davidgraeff:
The class `ConfigUtil` is used by `Configuration` to normalize values of a given Map<String,Object> configuration map.

I started to use `Configuration` recently also in OSGi services (in the activate and modified method), not only ThingHandlers, which works wonderfully. Except that I get a lot of warnings of the aforementioned class because it tries to normalize "objectClass" which is a complex object.

To be able to use the `Configuration` class like in the example below, the "objectClass" and two other OSGi default configuration parameters should be excluded.

```
    @Modified
    protected void modified(Map<String, Object> config) {
        configuration= new Configuration(config).as(NetworkBindingConfiguration.class);
    }
```


sjka:
That clearly makes sense!

I'm just wondering whether this is the right/only place for this check... On the first glance there also is the `normalizeTypes(Map<String, Object>, List<ConfigDescription>)` which could suffer from the same bug, couldn't it? 

In any case, I think putting this assertion into ConfigUtilTest to make sure it doesn't get broken by anybody accidentally would make sense.

davidgraeff:
I cannot open projects with groovy code unfortunately, something is misconfigured in my eclipse. I converted one of the test suites to java but converting everything is a little tedious to just have this merged.

bundles/config/org.eclipse.smarthome.config.core/src/main/java/org/eclipse/smarthome/config/core/ConfigUtil.java
```diff
@@ -138,8 +135,10 @@ public static Object normalizeType(Object value, ConfigDescriptionParameter conf
         for (Entry<String, ?> parameter : configuration.entrySet()) {
             String name = parameter.getKey();
             Object value = parameter.getValue();
-            ConfigDescriptionParameter configDescriptionParameter = configParams.get(name);
-            convertedConfiguration.put(name, normalizeType(value, configDescriptionParameter));
+            if (!name.equals("objectClass") && !name.equals("component.name") && !name.equals("component.id")) {
```
sjka: Now that I'm seeing this, I think it would be good to have a static set of strings defined once, then checking here (and above) only if the name is contained in there - just to make sure they don't run out of sync.

davidgraeff: What about a private helper method instead?

sjka: sure, even better

davidgraeff: done

bundles/config/org.eclipse.smarthome.config.core/src/main/java/org/eclipse/smarthome/config/core/ConfigUtil.java
```diff
@@ -35,11 +43,13 @@
...
      */
     public static Map<String, Object> normalizeTypes(Map<String, Object> configuration) {
-        Map<String, Object> convertedConfiguration = new HashMap<String, Object>(configuration.size());
+        Map<String, Object> convertedConfiguration = new HashMap<>(configuration.size());
         for (Entry<String, Object> parameter : configuration.entrySet()) {
             String name = parameter.getKey();
             Object value = parameter.getValue();
-            convertedConfiguration.put(name, normalizeType(value));
+            if (!isOSGiConfigParameter(name)) {
+                convertedConfiguration.put(name, normalizeType(value));
```
triller-telekom: You only add the `value` to your `convertedConfiguration` if it is not an OSGI parameter, is that by intention or is there an
```
else {
    convertedConfiguration.put(name, value);
}
```
missing?

bundles/config/org.eclipse.smarthome.config.core/src/main/java/org/eclipse/smarthome/config/core/ConfigUtil.java
```diff
@@ -138,8 +148,10 @@ public static Object normalizeType(Object value, ConfigDescriptionParameter conf
         for (Entry<String, ?> parameter : configuration.entrySet()) {
             String name = parameter.getKey();
             Object value = parameter.getValue();
-            ConfigDescriptionParameter configDescriptionParameter = configParams.get(name);
-            convertedConfiguration.put(name, normalizeType(value, configDescriptionParameter));
+            if (!isOSGiConfigParameter(name)) {
+                ConfigDescriptionParameter configDescriptionParameter = configParams.get(name);
+                convertedConfiguration.put(name, normalizeType(value, configDescriptionParameter));
```
triller-telekom: Same as above, don't you want to add an (un-normalized) OSGI parameter to your return map?

bundles/config/org.eclipse.smarthome.config.core/src/main/java/org/eclipse/smarthome/config/core/ConfigUtil.java
davidgraeff: Yes that's the entire purpose of this PR, to ignore the OSGi parameters completely. They contain either dots so cannot be matched with a variable name or are complex objects, so cannot be marshaled to values.

sjka: +1 - these values don't make any sense for the service itself. It knows its name and pid.

sjka:
Thanks!

bundles/config/org.eclipse.smarthome.config.core/src/main/java/org/eclipse/smarthome/config/core/ConfigUtil.java
triller-telekom: Ah Ok, then the name of this PR was a bit misleading for me, because it soudned like you only want to omit the "normalization" and not the entire parameters. thanks for the clarification!

PR #4113, opened 2017-08-23
Set configurationPid instead of service name on archetype

triller-telekom:


hakan42:
stupid question, is the configurationPolicy important to set?

triller-telekom:
@hakan42 if the binding has a configuration which will be stored in configadmin and that should appear as a section in paperui, then yes :)

kaikreuzer:
@hakan42 The `configurationPid` is now used as a replacement for `name`, see [this discussion](https://github.com/openhab/openhab2-addons/pull/2559#issuecomment-324055141).

hakan42:
o.k., I will adapt this in the senseBox binding immediately.

PR #4116, opened 2017-08-23
Use configurationPid instead of name in OSGI services

triller-telekom:


maggu2810:
Isn't the configuration PID equal to the name of the component if not explicitly set?
Has it been different on 4.2?

Could we please add the reason (e.g. link to a bug report, forum entry, etc.) why a PR / change is done?
For me it is not clear.

maggu2810:
Ah okay, will continue reading https://github.com/eclipse/smarthome/pull/4113

kaikreuzer:
> Could we please add the reason (e.g. link to a bug report, forum entry, etc.) why a PR / change is done?

We definitely should - @triller-telekom, please always add some comments about the "why" on such PRs, thanks!

triller-telekom:
Sorry @maggu2810 next time I will include a link to the discussion

maggu2810:
@triller-telekom All fine, thank you :wink:

PR #4117, opened 2017-08-23
Removing unnecessary repos

kaikreuzer:
Our build is continuously failing with timeouts on weird remote repos and it is also generally terribly slow.
I just checked the additional repos that we include in our pom.xml and it seems that we do not require any of them anymore. We e.g. do not use snapshots from Tycho and the buchen-maven-repo is only required by the Designer RCP build, which is not part of this repo anymore.

### DIFF
--- a/pom.xml
+++ b/pom.xml
@@ -471,32 +471,4 @@
     </profile>
   </profiles>
 
-  <pluginRepositories>
-    <pluginRepository>
-      <id>tycho-snapshots</id>
-      <url>https://oss.sonatype.org/content/groups/public/</url>
-    </pluginRepository>
-    <pluginRepository>
-      <id>cbi-releases</id>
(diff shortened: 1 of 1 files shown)


davidgraeff:
Travis supports a cache mechanism. Can all of this p2 stuff be cached, so that no downloads at all happen unless there is a new dependency?

maggu2810:
Travis job has still not be started...
Caching for Travis would make sense to reduce the build time.
Broken dependencies could be identified by the nightly build.

kaikreuzer:
I didn't yet look at what needs to be done to Travis for caching p2 artifacts. @davidgraeff If you know the details, feel free to create a PR.

@maggu2810 I restarted the Travis job last night, now it has run and is green.

PR #4118, opened 2017-08-23
WU binding: example updated in the documentation

lolodomo:
It is not allowed to use "Location" as an item name.

### DIFF
--- a/extensions/binding/org.eclipse.smarthome.binding.weatherunderground/README.md
+++ b/extensions/binding/org.eclipse.smarthome.binding.weatherunderground/README.md
@@ -131,7 +131,7 @@ demo.items:
 String Conditions "Conditions [%s]" {channel="weatherunderground:weather:CDG:current#conditions"}
 Image Icon "Icon" {channel="weatherunderground:weather:CDG:current#icon"}
 DateTime ObservationTime "Observation time [%1$tH:%1$tM]" <clock>  {channel="weatherunderground:weather:CDG:current#observationTime"}
-String Location "Location [%s]" {channel="weatherunderground:weather:CDG:current#location"}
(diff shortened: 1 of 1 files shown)


extensions/binding/org.eclipse.smarthome.binding.weatherunderground/README.md
```diff
@@ -131,7 +131,7 @@ demo.items:
 String Conditions "Conditions [%s]" {channel="weatherunderground:weather:CDG:current#conditions"}
 Image Icon "Icon" {channel="weatherunderground:weather:CDG:current#icon"}
 DateTime ObservationTime "Observation time [%1$tH:%1$tM]" <clock>  {channel="weatherunderground:weather:CDG:current#observationTime"}
-String Location "Location [%s]" {channel="weatherunderground:weather:CDG:current#location"}
+String ObservationLocation "Location [%s]" {channel="weatherunderground:weather:CDG:current#location"}
```
htreu: I dont want to make too much of a discussion here, but what does `ObservationLocation` actually mean?

lolodomo: Location of xeather observation.
It is the terminology of WU: https://www.wunderground.com/weather/api/d/docs?d=data/conditions

htreu: thanks for clarification.

sjka:
Thanks!

kaikreuzer:
> It is not allowed to use "Location" as an item name.

@lolodomo Just as a small correction: It is allowed, but when using keywords as names, they have to be escaped as `^Location`.

PR #4122, opened 2017-08-24
Migrate to the new Xtext generator

sjka:
This PR again is another step towards #3898. It migrates the DSLs to the new Xtext generator according to https://typefox.io/xtexts-new-generator-migration.

With the new generator, by default the commons parts for all supported IDEs (Eclipse, IntelliJ, Web, LSP...) are split out into `*.ide` bundles. I have been sticking to this default, as with #3898 we will be dropping the `*.ui` bundles. The only other option would have been to put this stuff into the main bundles of each DSL, but then again the LSP stuff cannot be optional. So all in all the default looks sensible and keeps modularization clean. Therefore I kept it.

As the generator changed, the configuration in the `*.mwe2` files completely changed. They even look cleaner now.


kaikreuzer:
You tend to forget about our feature definitions :-)
This time it is not only the Karaf features that need to be adapted, but also the p2 features.

sjka:
I would tend to add the `*.ide` to a future "LSP" karaf-feature once we have the language server (future PR). The same applies to the p2 runtime features: they are currently not needed there.

sjka:
I have added them to the `.runtime.model` p2 feature.

bundles/model/org.eclipse.smarthome.model.item/META-INF/MANIFEST.MF
```diff
@@ -1,48 +1,50 @@
...
+ org.eclipse.emf.mwe2.launch;resolution:=optional,
+ com.ibm.icu;resolution:=optional,
+ org.eclipse.xtext.util,
+ org.eclipse.emf.ecore;visibility:=reexport,
+ org.eclipse.emf.common,
+ org.antlr.runtime,
+ org.eclipse.xtext.common.types,
+ org.eclipse.smarthome.model.lazygen;resolution:=optional,
+ org.eclipse.xtext.xbase.lib,
+ org.eclipse.xtend.lib
```
kaikreuzer: This is a new dependency. Do we really need it (mandatorily) or only optional for the IDE to be able to resolve annotations?

bundles/model/org.eclipse.smarthome.model.item/META-INF/MANIFEST.MF
```diff
@@ -1,48 +1,50 @@
...
+ org.apache.log4j,
+ org.eclipse.smarthome.core.autoupdate,
+ org.eclipse.smarthome.core.common.registry,
+ org.eclipse.smarthome.core.items,
+ org.eclipse.smarthome.core.items.dto,
+ org.eclipse.smarthome.core.library.items,
+ org.eclipse.smarthome.core.library.types,
+ org.eclipse.smarthome.core.types,
+ org.eclipse.smarthome.model.core,
+ org.eclipse.xtend.lib.annotations,
```
kaikreuzer: This is a new dependency. Do we really need it (mandatorily) or only optional for the IDE to be able to resolve annotations?

kaikreuzer:
@sjka Would be cool if you could make this dependency optional on ALL model projects, not just on items.

sjka:
There is a require bundle dependency chain of `o.e.xtext` -> `o.e.xtext.util` -> `o.e.xtend.lib`, hence they are in fact not as optional as some might wish. And xbase also requires it. Therefore I don't believe a runtime using DSLs would work without the `o.e.xtend.lib` bundle...
Although apparently there is no direct dependency of the source code to it, having it as an "optional" dependency would indicate that you can leave it out. Which obviously is not the case. Therefore I would rather prefer to keep them as mandatory.

kaikreuzer:
> having it as an "optional" dependency would indicate that you can leave it out.

Hm, it would only indicate is as much as you can leave out any other transitive dependency. A resolver would anyhow notice that xtext requires it and include it. Nonetheless, I would think that the clean way is to only mention our direct dependencies - an optional means that we need it for compilation/IDE, but we do not ourselves depend on it at runtime. Maybe xtext changes in future and does NOT have this dependency anymore at runtime - then our manifest would become incorrect, just because some dependency in external bundles have changed.

sjka:
Okay, I made them optional now in all languages. 

But please be aware that xtend lib includes stuff that the classes generated from xtend files potentially can reference. By chance we don't need them today (if I didn't miss anything) - but in case any of the xtend files changes (or the xtend generator) this dependency could suddenly become mandatory again. However, nothing is gonna break because as mentioned above at runtime this dependency is required anyway.

bundles/model/org.eclipse.smarthome.model.rule.ide/src/org/eclipse/smarthome/model/rule/ide/RulesIdeModule.xtend
```diff
@@ -0,0 +1,11 @@
+/*
+ * generated by Xtext 2.12.0
+ */
+package org.eclipse.smarthome.model.rule.ide
+
+
+/**
+ * Use this class to register ide components.
+ */
+class RulesIdeModule extends AbstractRulesIdeModule {
```
martinvw: @sjka or anyone else, I cannot find this parent class when compiling locally (and GitHub can also not find it), I assume the smarthome build still works, so any hints?

maggu2810: Keep the ctrl (or strg) key pressed and click on the `AbstractRulesIdeModule`.
It will direct you to that implementation.

Also a find for the respective file should give you the result:
```
$ find -name "AbstractRulesIdeModule.*"
./bundles/model/org.eclipse.smarthome.model.rule.ide/target/classes/org/eclipse/smarthome/model/rule/ide/AbstractRulesIdeModule.class
./bundles/model/org.eclipse.smarthome.model.rule.ide/src-gen/org/eclipse/smarthome/model/rule/ide/AbstractRulesIdeModule.java
```

martinvw: @maggu2810 thanks!

Then I fear that the generation failed somehow, because I had this error while compiling with everything with maven. It was while tracing a bug which disappeared anyway after a reboot...

maggu2810: Hm, never (AFAIK) seen such an error while compiling using Maven. Are you sure that the Eclipse IDE has been closed while compiling with Maven?

martinvw: I'm sure it was not closed while compiling, I hear a possible work around / solution :-)

maggu2810: You need to close the Eclipse IDE before you start a Maven build. :wink:

PR #4124, opened 2017-08-24
Fix Weatherunderground thing is stuck in Initializing state

### LINKED ISSUE #4123: Weather Underground Binding gets stuck in Initializing state if location parameter is wrong
If the location parameter contains a space between latitude and longitude the thing never comes online and is stuck in initializing state.

I will provide a PR for it.

triller-telekom:
Fixes #4123

### DIFF
--- a/extensions/binding/org.eclipse.smarthome.binding.weatherunderground/ESH-INF/i18n/weatherunderground.properties
+++ b/extensions/binding/org.eclipse.smarthome.binding.weatherunderground/ESH-INF/i18n/weatherunderground.properties
@@ -7,4 +7,4 @@ offline.conf-error-missing-apikey = The "API key" parameter must be configured.
 offline.conf-error-missing-location = The "location" parameter must be configured.
 offline.conf-error-syntax-language = The value of the "language" parameter must contain 2 letters.
 offline.conf-error-min-refresh = The minimum value of the "refresh interval" parameter must be 5 minutes.
-offline.uri-error = Could not create URI to fetch data, please check if parameters are correct and otherwise report a bug.
+offline.uri-error = Could not create URI to fetch data, please check if the location parameter is correct.
--- a/extensions/binding/org.eclipse.smarthome.binding.weatherunderground/ESH-INF/i18n/weatherunderground.properties
+++ b/extensions/binding/org.eclipse.smarthome.binding.weatherunderground/ESH-INF/i18n/weatherunderground.properties
@@ -7,3 +7,4 @@ offline.conf-error-missing-apikey = The "API key" parameter must be configured.
 offline.conf-error-missing-location = The "location" parameter must be configured.
 offline.conf-error-syntax-language = The value of the "language" parameter must contain 2 letters.
 offline.conf-error-min-refresh = The minimum value of the "refresh interval" parameter must be 5 minutes.
+offline.uri-error = Could not create URI to fetch data, please check if parameters are correct and otherwise report a bug.
--- a/extensions/binding/org.eclipse.smarthome.binding.weatherunderground/ESH-INF/thing/thing-types.xml
+++ b/extensions/binding/org.eclipse.smarthome.binding.weatherunderground/ESH-INF/thing/thing-types.xml
@@ -58,7 +58,7 @@
 				<label>API Key</label>
 				<description>API key to access the Weather Underground service</description>
 			</parameter>
-			<parameter name="location" type="text" required="true" pattern="[0-9]{1,2}(\.[0-9]*)*,[0-9]{1,2}(\.[0-9]*)*">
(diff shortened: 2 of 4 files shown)


extensions/binding/org.eclipse.smarthome.binding.weatherunderground/ESH-INF/i18n/weatherunderground.properties
```diff
@@ -7,3 +7,4 @@ offline.conf-error-missing-apikey = The "API key" parameter must be configured.
 offline.conf-error-missing-location = The "location" parameter must be configured.
 offline.conf-error-syntax-language = The value of the "language" parameter must contain 2 letters.
 offline.conf-error-min-refresh = The minimum value of the "refresh interval" parameter must be 5 minutes.
+offline.uri-error = Could not create URI to fetch data, please check if parameters are correct and otherwise report a bug.
```
kaikreuzer: - "and otherwise report a bug" is a pretty unusual (and rather inappropriate) message for a Thing status.
- "please check if parameters are correct" - can't we figure out ourselves if a parameter is invalid and if so, which and provide more details here?

extensions/binding/org.eclipse.smarthome.binding.weatherunderground/ESH-INF/thing/thing-types.xml
```diff
@@ -58,7 +58,7 @@
 				<label>API Key</label>
 				<description>API key to access the Weather Underground service</description>
 			</parameter>
-			<parameter name="location" type="text" required="true">
+			<parameter name="location" type="text" required="true" pattern="[0-9]{1,2}(\.[0-9]*)*,[0-9]{1,2}(\.[0-9]*)*">
```
kaikreuzer: Is this pattern valid for all "Multiple syntaxes are supported"?

mhilbush:
Will this have any affect on the other location syntaxes that are allowed in the location parameter?

extensions/binding/org.eclipse.smarthome.binding.weatherunderground/src/main/java/org/eclipse/smarthome/binding/weatherunderground/handler/WeatherUndergroundHandler.java
```diff
@@ -337,6 +337,11 @@ private WeatherUndergroundJsonData getWeatherData(Set<String> features, String l
             error = "Error parsing Weather Underground response";
             errorDetail = e.getMessage();
             statusDescr = "@text/offline.comm-error-parsing-response";
+        } catch (IllegalArgumentException e) {
+            // catch Illegal character in path at index XX: http://api.wunderground.com/...
```
kaikreuzer: It is pretty hard to say which line of the code block actually might throw this exception. Could that be moved closer to the code that it handles?

extensions/binding/org.eclipse.smarthome.binding.weatherunderground/src/main/java/org/eclipse/smarthome/binding/weatherunderground/handler/WeatherUndergroundHandler.java
```diff
@@ -337,6 +337,11 @@ private WeatherUndergroundJsonData getWeatherData(Set<String> features, String l
...
+            error = "Error creating URI to fetch data";
```
kaikreuzer: pretty unspecific message...

triller-telekom:
> Will this have any affect on the other location syntaxes that are allowed in the location parameter?

Thanks for reminding me that there are more possibilities, I have adjusted the regex and created more detailed error messages

lolodomo:
I think a good pratice could be to add a new entry in all translation propoerty files when a new entry is added in the default (english) property file. The entry would be the same as in the default file but put in comments. Then we will know that all lines in comments in a translation file are lines that need to be translated.
If we don't do that, translation files will quickly become outdated and any update will require a binding review to know what is missing.
WDYT ?

triller-telekom:
> I think a good pratice could be to add a new entry in all translation propoerty files when a new entry is added in the default (english) property file. The entry would be the same as in the default file but put in comments. 

That' actually a good idea. I have added my new entry as a comment in the french translation file.

extensions/binding/org.eclipse.smarthome.binding.weatherunderground/ESH-INF/thing/thing-types.xml
```diff
@@ -58,7 +58,7 @@
 				<label>API Key</label>
 				<description>API key to access the Weather Underground service</description>
 			</parameter>
-			<parameter name="location" type="text" required="true">
+			<parameter name="location" type="text" required="true" pattern="(^-?[0-9]{1,2}\.[0-9]+,-?[0-9]{1,2}\.[0-9]+$)|(^([A-Z]|[a-z])+\/([A-Z]|[a-z])+$)|(^([A-Z]|[a-z])+$)|(^pws:([A-Z]|[a-z]|[0-9])+$)">
```
sjka: The regex doesn't even match all of the examples from the readme, so maybe it's better to drop it again?

triller-telekom:
Dropped the pattern from the location parameter because it was a guess only anyway and we do not exactly know their API.

sjka:
Thanks!

PR #4125, opened 2017-08-24
Remove ConfigurationPolicy.OPTIONAL from bindings archetype as its de…

triller-telekom:
…fault anyway


kaikreuzer:
Thanks!

PR #4126, opened 2017-08-24
Expose the command line delimiter

garytse:
The javadoc in method executeCommandLineAndWaitResponse mentions @@ as the command line delimiter.  It is easier to expose it for external reference.


bundles/io/org.eclipse.smarthome.io.net/src/main/java/org/eclipse/smarthome/io/net/exec/ExecUtil.java
```diff
@@ -30,7 +30,10 @@
  */
 public class ExecUtil {
 
-    private static final String CMD_LINE_DELIMITER = "@@";
+	/**
```
kaikreuzer: Please use the auto-formatter to avoid adding tabs in the source code.

kaikreuzer:
Thanks - and now please squash the two commits into one :-)

PR #4128, opened 2017-08-24
Change retrieving the firmware version to be async

ivivanov-bg:
The current timeout of 2 seconds is way to short on very first initialization of the CoapClient

According to https://tools.ietf.org/html/rfc7252#section-4.8.2 the initial timeout should be 93 seconds.

The proposal is to increase the timeout from 2 to 20 secs. This way it will be relatively small but still long enough for first initialization

### DIFF
--- a/extensions/binding/org.eclipse.smarthome.binding.tradfri/src/main/java/org/eclipse/smarthome/binding/tradfri/handler/TradfriGatewayHandler.java
+++ b/extensions/binding/org.eclipse.smarthome.binding.tradfri/src/main/java/org/eclipse/smarthome/binding/tradfri/handler/TradfriGatewayHandler.java
@@ -137,7 +137,7 @@ public class TradfriGatewayHandler extends BaseBridgeHandler implements CoapCall
      */
     public void startScan() {
         if (endPoint != null) {
-            updateGatewayInfoSynchronious();
+            updateGatewayInfo();
             deviceClient.get(new TradfriCoapHandler(this));
         }
     }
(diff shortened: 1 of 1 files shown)


kaikreuzer:
That's an interesting finding. The [JavaDoc](https://github.com/eclipse/californium/blob/245fac1e9dc1547cc6f42ffa984ee6cc6b660144/californium-core/src/main/java/org/eclipse/californium/core/CoapClient.java#L135-L137) indeed says:
> Under normal circumstances this property should be set to at least the <em>EXCHANGE_LIFETIME</em>

And `EXCHANGE_LIFETIME` is [defined as 247s](https://github.com/eclipse/californium/blob/76d023b97d7ada9b259e4d87432a9383b2574b3b/californium-core/src/main/java/org/eclipse/californium/core/network/config/NetworkConfigDefaults.java#L81). (note that this is NOT the same as the `MAX_TRANSMIT_WAIT`, which is 93s).

So I would actually claim that the correct fix is to completely remove the call to `setTimeout(TIMEOUT)` and rather go with the default.

BUT: This setting only specifies how long synchronous calls will block. Do we actually use any synchronous call? I am only aware of async communication, so I am not sure whether this fix will actually fix anything. Did you come across a bug where you identified this setting to be the root cause?

ivivanov-bg:
Hi,

Yes - we do have a synchronous call - in updateGatewayInfoSynchronious() for retrieving the firmware version of the gateway

kaikreuzer:
> Yes - we do have a synchronous call - in updateGatewayInfoSynchronious()

Hm, ok, this has only been recently introduced by yourself. I didn't notice when reviewing it that it is a blocking call. Any reason why this cannot be done asynchronous? It looks to me like a perfect candidate for it...

ivivanov-bg:
Should it be done similar to: requestDeviceDetails ?

I thought it's dangerous because this runs in different thread and I don't see any synchronization on setURI

And also - the other option (to have another client) is way to heavy

kaikreuzer:
>  I don't see any synchronization on setURI

Well, if this is a problem, we will have the same problem with requestDeviceDetails, so I do not see this as an argument :-)

ivivanov-bg:
> we will have the same problem with requestDeviceDetails

We might have it - just not yet been hit :)

Ok - I'll change it in similar fashion

kaikreuzer:
When implementing it that way I had a look at the Californium sources and am confident that there is no synchronization issue, since the URI is set in a newly created instance for every async call.

ivivanov-bg:
>since the URI is set in a newly created instance for every async call

I also checked the code and I think there might be an issue as the changing of the URI is not synchronized.
So - when you do: deviceClient.setURI() you change the uri of the instance and later in the get there is: asynchronous(newGet().setURI(this.uri), handler);

But at this point someone else might have already changed the URI.


Anyway - I'll do the change and if there is an issue at some time in the future - then will think about a fix

kaikreuzer:
> But at this point someone else might have already changed the URI.

Well, as the `deviceClient` is private to our handler, it can be only ourself. But yes, as a safeguard we could actually synchronize the `requestDeviceDetails` method.

ivivanov-bg:
Did I missed something else ??

PR #4131, opened 2017-08-25
Astro: Schedule Jobs 30sec after midnight to ensure to be on the next…

### LINKED ISSUE #4018: Astro binding - Channel 'astro:sun:local:set#event' does not trigger
Dear,
When I was on build #986 (OH 2.2). I freshly installed the Astro binding and made the required settings. 

I have the following rule:
`rule "Zonsondergang - nachtverlichting aan"

when
    Channel 'astro:sun:local:set#event' triggered START 
then
	logInfo("Astro","De zon gaat onder. Slaapwel.")
        sendCommand(KNX_GV_Nachtverlichting_4_2_0, ON)
end`

It does not get triggered. At midnight I do see this:

`2017-08-10 00:00:00.371 [INFO ] [thome.binding.astro.internal.job.Job] - Scheduled Astro event-jobs for thing astro:moon:local
2017-08-10 00:00:00.876 [INFO ] [thome.binding.astro.internal.job.Job] - Scheduled Astro event-jobs for thing astro:sun:local
`
If I check out the value of my Astro-items, the sun should have risen on 05:50 and will set at 21h37. So the values of all Astro-items are calculated correctly.

My current build is #1003.

It seems I am not the only one with this issue. See [this thread](https://community.openhab.org/t/astro-binding-problem-events-do-not-trigger-multiple-triggers-solved/30320/12) on the OpenHAB Community.

Thank you for looking in to this.

triller-telekom:
… day

Fixes #4018

### DIFF
--- a/extensions/binding/org.eclipse.smarthome.binding.astro/src/main/java/org/eclipse/smarthome/binding/astro/handler/AstroThingHandler.java
+++ b/extensions/binding/org.eclipse.smarthome.binding.astro/src/main/java/org/eclipse/smarthome/binding/astro/handler/AstroThingHandler.java
@@ -52,7 +52,7 @@ import org.slf4j.LoggerFactory;
  */
 public abstract class AstroThingHandler extends BaseThingHandler {
 
-    private static final String DAILY_MIDNIGHT = "0 0 0 * * ? *";
+    private static final String DAILY_MIDNIGHT = "30 0 0 * * ? *";
 
     /** Logger Instance */
     protected final Logger logger = LoggerFactory.getLogger(MethodHandles.lookup().lookupClass());


maggu2810:
Can you please give me some more details why 0h 0m 30s? is on the next day and 0h 0m 0s isn't?

triller-telekom:
The problem is that our scheduler calculates the milliseconds until the next execution and in the cases of the bug that I am fixing the scheduler and the system time are a few (milli)seconds off and thus the expression scheduled to be executed at 0h 0m 0s is executed before midnight which is the day before. And in case of the Astro binding we are only interested in the day and not not the actual time of the cron execution.

maggu2810:
> the expression scheduled to be executed at 0h 0m 0s is executed before midnight which is the day before

Hm, it seems to me that the scheduler is then buggy and not the astro binding.
Isn't the normal behavior that the execution is at the given time or a little bit later.
If a cron expression that is using "00:00:00" as execution trigger is executed on the old day, I would say this is an unexpected behavior.

Changing the trigger cron expression will mask the error but not really solve the bug.

triller-telekom:
Indeed you are right. But there a quite a few bug reports regarding the random skipping of Astro events [here](https://github.com/eclipse/smarthome/issues/4018) and in the openHAB [forum](https://community.openhab.org/t/astro-binding-problem-events-do-not-trigger-multiple-triggers-solved/30320/12) so the idea was to have a quick fix for the next ESH stable and then fix the scheduler. Because fixing the scheduler will take longer.

kaikreuzer:
As a temporary workaround fix, this looks ok for me - the "real" fix will have to address #4145 then.
@maggu2810 If you are fine with it as well, feel free to merge.

maggu2810:
Can we create an issue to revert that change after #4145 has been fixed?
It seems to be wrong that a cron expression using the name "DAILY_MIDNIGHT" has an offset of 30 seconds.

triller-telekom:
Done: #4167

PR #4136, opened 2017-08-25
Export the package org.eclipse.smarthome.core.thing.link.events

hsudbrock:
Currently, the package org.eclipse.smarthome.core.thing.link.events is not exported.
In consequence, one cannot use the ItemChannelLinkXXXEvent classes in that package
to listen to those events. Moreover, this is unlike for other events packages, like, e.g.,
the package org.eclipse.smarthome.core.items.events which is exported.

This is why I suggest to export org.eclipse.smarthome.core.thing.link.events.
Or is there a reason why it should not be exported?

Signed-off-by: Henning Sudbrock <henning.sudbrock@telekom.de>


hsudbrock:
Just had a another look at the coding guidelines - in addition, the package should then also be imported by its bundle. I will add an additional pull request for fixing that issue introduced in this pull request.

Issue #3859, opened 2017-07-19
[BasicUI] Allow switching between sitemaps


demsot:
With a click on the home button in the basic ui, the ui should open the default sitemap, which is entered in the paper ui:
![image](https://user-images.githubusercontent.com/21989078/28369113-e67d7710-6c96-11e7-9bc6-c576c730d614.png)

ThomDietrich:
Reference: https://community.openhab.org/t/switch-between-sitemaps-in-basic-ui/31898

sjka:
Isn't that completely confusing? I'd expect a "home" button to lead to the "home page", not to the configuration page where is set what "home" is.

kaikreuzer:
I agree with @sjka - after all, IF a default sitemap is set, its "home" is the expected result. If no default is set, the selection list makes sense and this is imho already the case.

triller-telekom:
I just tried out what the "home button" is doing: Independent of which sitemap is configured as the default one, this button brings you back to the main page of the **current** sitemap. And I think that is the expected behavior.

Imagine a user being on a secondary sitemap deep in a nested frame and wants to go back to the main page, so intuitively he clicks on that button. He will expect to land on the first page of that secondary sitemap, otherwise he will probably be confused that the sitemap has been changed.

I would consider this not as a bug but working as expected.

The only addition I could imagine is a second button, maybe containing a "1" which the user can click to land not on the main page of the current sitemap but on the one which he has configured to be his default one.

ThomDietrich:
I disagree to close this issue. We should discuss the details though.

**Problem 1:** When inside the main view of one sitemap, the home button doesn't have a functionality -> not intuitive
**Problem 2:** It should be possible to navigate to the list of (other) sitemaps via a UI button -> not available

The home button should open up the list of available sitemaps.

sjka:
1. The Home Button should be a "Home" button. Nothing else. 
2. The Basic UI must not link to the PaperUI

> When inside the main view of a sitemap, the home button doesn't have a functionality -> not intuitive

Oh, it has. It provides a visual clue that you _are_ on the main view.

> It should be possible to navigate to the list of (other) sitemaps via a UI button -> not available

That's something completely different. So far this issue was about adding some functionality to the Home button. Not about selecting another sitemap.

kaikreuzer:
> The home button should open up the list of available sitemaps.

I fully disagree. The home button, as on any webpage, should go to the homepage. If you are already there, so be it.
[...]

Issue #3523, opened 2017-05-26
[BasicUI] Icons & font MaterialIcons-Regular are missing


htreu:
In BasicUI some of the icons do not show up, i.e. the group navigation (arrow left/right) or the color picker items. In addition the requested font MaterialIcons-Regular is not available from the web/fonts folder. I wonder why this ever worked because the files for this font never were part of the repository according to git history.

![image](https://cloud.githubusercontent.com/assets/1729268/26488119/75ff7cea-4202-11e7-835b-b98f0c479302.png)

I can see that @resetnow did changes to material-icons.css which indicate that this should have been working some day with 072e76c7be6efadd93821fba17b82516b3265c83 what am I missing here?

When I revert 072e76c7be6efadd93821fba17b82516b3265c83 and put current version of MaterialIcons-Regular.woff2 in the web/fonts folder everything works well.

Tested in Chrome/Safari/Firefox on macOS.

kaikreuzer:
@htreu I didn't look into it, but just want to mention that all looks good in the very latest openHAB distro build nonetheless:

<img width="1058" alt="screen shot 2017-05-26 at 11 25 37" src="https://cloud.githubusercontent.com/assets/3244965/26489046/112d9b36-4206-11e7-81e0-06e06c16352c.png">

htreu:
In openHAB the fonts are provided by a fragment, see https://github.com/openhab/openhab-core/tree/master/bundles/org.openhab.ui.basicui/patch/web/fonts

htreu:
It seems the font provided by the above mentioned fragment in openHAB is different from the MaterialIcons-Regular font which is already part of ESH in PaperIU.

I suggest to add the PaperUI version of the font to BasicUI (we already have a CQ for this version) and remove the patch here.

@resetnow is there anything to be said against this?

resetnow:
IIRC the mentioned commit contains changes​ caused by using the same Material Icons font as PaperUI. I'll check if it's correct. I have to note that for me the fonts provided by openHAB overlay work well.

htreu:
@resetnow yes the overlay works well. But in ESH the font is completely missing and the version from PaperUI (which would be nice to use since it already has CQ) uses the "old" codes.

kaikreuzer:
@resetnow & @htreu Any suggestion how to proceed here? Could we simply switch to the version that is also used by the Paper UI?

resetnow:
[...]

PR #4137, opened 2017-08-25
Basic UI styling changed to SASS + initial theming support

kubawolanin:
Inspired by [this thread](https://community.openhab.org/t/basic-ui-dark-theme/33603) I started to play around the Basic UI front-end. Treat this PR as a proposal and a discussion point, rather than a full-blown feature :-)

So far I've managed to:
* clean up the stylesheets, 
* add some CSS custom properties (variables) in `_theming.scss` file 
* and move from LESS to SCSS.

The reason for the last bullet point was to potentially utilize Material Design Lite in [scss version only](https://github.com/tomatau/mdl-scss). That would give us some more flexibility wrt styling (e.g. extend css classes with internal mdl components or create your own).

CSS Custom Properties, however, are a great way to support theming in the runtime. 
I've [applied them in HABPanel](https://github.com/openhab/org.openhab.ui.habpanel/pull/187) as well and amount of possibilities they gave to the UI theming is huge 😎 

~~**I couldn't find a way to test my work** within a openhab2 dev environment. I hope I will find some help here. I've run `gulp` task to build the front-end. Then I cleanend and run the Eclipse workspace.
None of my changes appeared on the running UI though.
I didn't find any information regarding building Basic UI specifically in [the documentation](https://eclipse.org/smarthome/documentation/features/ui/basic/readme.html).~~

Edit: It all works now :-) forgot to enable the plugin in Run configuration 🙄 Thx @resetnow 

Cheers!

Signed-off-by: Kuba Wolanin <hi@kubawolanin.com> (github: kubawolanin)

### DIFF
--- a/extensions/ui/org.eclipse.smarthome.ui.basic/ESH-INF/config/config.xml
+++ b/extensions/ui/org.eclipse.smarthome.ui.basic/ESH-INF/config/config.xml
@@ -5,6 +5,15 @@
         http://eclipse.org/smarthome/schemas/config-description-1.0.0.xsd">
 
 	<config-description uri="ui:basic">
+		<parameter name="theme" type="text" required="true">
+			<label>Theme</label>
+			<description>Defines the UI theme.</description>
+			<options>
+				<option value="default">Default</option>
+				<option value="dark">Dark</option>
+			</options>
+			<default>default</default>
+		</parameter>
 		<parameter name="iconType" type="text" required="true">
 			<label>Icon Format</label>
 			<description>Defines whether bitmap or vector icons are used by the UI.</description>
--- a/extensions/ui/org.eclipse.smarthome.ui.basic/gulpfile.js
+++ b/extensions/ui/org.eclipse.smarthome.ui.basic/gulpfile.js
@@ -3,19 +3,19 @@
 
 	var
 		gulp = require("gulp"),
-		less = require("gulp-less"),
+		sass = require("gulp-sass"),
 		uglify = require("gulp-uglify"),
 		eslint = require("gulp-eslint");
 
 	var
 		sources = {
 			js: "web-src/smarthome.js",
-			less: "web-src/smarthome.less"
+			sass: "web-src/smarthome.scss"
 		};
 
 	gulp.task("css", function() {
-		return gulp.src(sources.less)
-			.pipe(less({ compress: true }))
+		return gulp.src(sources.sass)
+			.pipe(sass({outputStyle: 'compressed'}).on('error', sass.logError))
 			.pipe(gulp.dest("web"));
 	});
 
--- a/extensions/ui/org.eclipse.smarthome.ui.basic/snippets/main.html
+++ b/extensions/ui/org.eclipse.smarthome.ui.basic/snippets/main.html
@@ -61,7 +61,7 @@
 			</div>
 		</script>
 	</head>
-	<body class="mdl-color--grey-100 mdl-color-text--grey-700" data-sitemap="%sitemap%" data-page-id="%id%" data-icon-type="%icon_type%">
+	<body class="mdl-color-text--grey-700" data-sitemap="%sitemap%" data-page-id="%id%" data-theme="%theme%" data-icon-type="%icon_type%">
 	<div class="mdl-layout mdl-js-layout mdl-layout--fixed-header">
 		<header class="mdl-layout__header mdl-layout__header--scroll navigation navigation-home">
 			<div class="mdl-layout__header-row">
@@ -78,7 +78,7 @@
 			</div>
 			<div class="ui__loading mdl-progress mdl-js-progress mdl-progress__indeterminate"></div>
 		</header>
-		<main class="mdl-layout__content mdl-color--grey-100">
+		<main class="mdl-layout__content">
 			<div class="page-content mdl-grid">
 				%children%
 			</div>
--- a/extensions/ui/org.eclipse.smarthome.ui.basic/CONTRIBUTING.md
+++ b/extensions/ui/org.eclipse.smarthome.ui.basic/CONTRIBUTING.md
@@ -1,6 +1,6 @@
 ## Contributing
 
-Javascript linting, compressing and LESS compilation is handled through Grunt. Therefore, please run `grunt` instead of changing compressed CSS/JS manually. Default Grunt task will also check the code using ESLint.
+Javascript linting, compressing and LESS compilation is handled through Gulp. Therefore, please run `gulp` instead of changing compressed CSS/JS manually. Default Gulp task will also check the code using ESLint.
 
 Installing build dependencies:
 
--- a/extensions/ui/org.eclipse.smarthome.ui.basic/package.json
+++ b/extensions/ui/org.eclipse.smarthome.ui.basic/package.json
@@ -8,7 +8,7 @@
   "dependencies": {
     "gulp": "^3.9.0",
     "gulp-eslint": "^4.0.0",
-    "gulp-less": "^3.3.2",
+    "gulp-sass": "^3.1.0",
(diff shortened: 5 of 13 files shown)


resetnow:
>I couldn't find a way to test my work 

Assuming you have imported orc.eclipse.smarthome.ui.basic bundle and enabled it in the run configuration, it should work the way you described. Try changing any file in `web/` directory and see if you changes appear in browser.

extensions/ui/org.eclipse.smarthome.ui.basic/snippets/main.html
```diff
@@ -61,7 +61,7 @@
 			</div>
 		</script>
 	</head>
-	<body class="mdl-color--grey-100 mdl-color-text--grey-700" data-sitemap="%sitemap%" data-page-id="%id%" data-icon-type="%icon_type%">
+	<body class="mdl-color-text--grey-700" data-sitemap="%sitemap%" data-page-id="%id%" data-icon-type="%icon_type%">
```
resetnow: These classes may need to be left as is for browsers without CSS variables support — BasicUI still works with IE 11 and not-so-recent Android versions. This will introduce a graceful fallback.

https://caniuse.com/#feat=css-variables

kubawolanin: CSS Custom Properties will fallback gracefully as well if you pass a second argument to them, e.g
```css
background: var(--body-bg, #f5f5f5);
```

And this is what I did in my follow-up commit.

kubawolanin:
Thanks @resetnow for the quick review!
I've added a fallback values to the CSS custom properties, see [here](https://github.com/kubawolanin/smarthome/blob/7493737811d4b20328d7baff364e59060a962fd2/extensions/ui/org.eclipse.smarthome.ui.basic/web-src/_theming.scss#L57).

I've also added a few themes already 😺 
So when you go to http://<yourESH>/paperui/index.html#/configuration/services and click on Basic UI configuration, you'll be able to change `theme` property to one of the following values:

```
default
dark
blue-grey
deep-orange
green
```

And here's a little sneak peek of the themes:

basic-ui-themes

At this point I think I can remove "WIP" flag from the title and we can proceed with the proper review :-)

Cheers!

resetnow:
>I've added a fallback values to the CSS custom properties, see here.

I don't think that's how fallback works. I have tested <a href="https://codepen.io/vladislav/pen/vJbwrv">this page</a> in a couple of browsers and it doesn't work for me, am I doing this right?





I think a better solution would be specifying the same property twice, only the second time using the variable syntax.

```css
color: black;
color: var(--color);
```

kubawolanin:
@resetnow well spotted! I've added better fallback now. Thanks

ThomDietrich:
@kubawolanin may I ask about the new home icon in the top left corner (what/why) and if that maybe means that you've touched https://github.com/eclipse/smarthome/issues/3859 (careful, the issue is a bit confusing)

resetnow:
@ThomDietrich Judging by the lack of arrows this is a result of openHAB bundles not being picked up by runtime.

kubawolanin:
Hey @ThomDietrich 👋 
No, I haven't touched it :-) My browser complained about MaterialIcons file being missing from the source so I've quickly added it temporarily to my dev environment to see if it helps.
As you can see, the group frame's arrows are not rendering too.
Edit: just found out that the issue is known -> https://github.com/eclipse/smarthome/issues/3523

extensions/ui/org.eclipse.smarthome.ui.basic/ESH-INF/config/config.xml
```diff
@@ -5,6 +5,18 @@
         http://eclipse.org/smarthome/schemas/config-description-1.0.0.xsd">
 
 	<config-description uri="ui:basic">
+		<parameter name="theme" type="text" required="true">
```
kaikreuzer: This is now a pretty different approach than we have in the Paper UI.
With the Paper UI, we allow ESH solutions to add a custom theme on top by adding a fragment. It is NOT a feature exposed to the user, though. I think this makes sense in terms of ensuring a "corporate identity", i.e. allowing solutions to make sure that their UIs carry their logo and match there general CI.

While I can understand that it can be useful to allow users to choose a light or dark theme, I am not so convinced, if we really need orange/blue/green color options (resp. rather allow solutions to add additional themes besides light/dark through a fragment).

extensions/ui/org.eclipse.smarthome.ui.basic/gulpfile.js
```diff
@@ -3,19 +3,19 @@
 
 	var
 		gulp = require("gulp"),
-		less = require("gulp-less"),
+		sass = require("gulp-sass"),
```
kaikreuzer: This is a new tool that we didn't use before, correct? If so, we will require a CQ and approval by Eclipse.

resetnow: We could also stick with LESS unless it doesn't support some of the required features.

extensions/ui/org.eclipse.smarthome.ui.basic/ESH-INF/config/config.xml
kubawolanin: I'll leave just Default and Dark then. We can work on the fragment solution later on if that's preferred.

extensions/ui/org.eclipse.smarthome.ui.basic/gulpfile.js
kubawolanin: I'd suggest to go with SASS because of the fact that MDL uses it natively by default. In the future we might extend/modify MDL stylesheets and possibly load them directly from `package.json` instead of hosting them as a static asset. 

See my comment from the first post:
> The reason for the last bullet point was to potentially utilize Material Design Lite in scss version only. That would give us some more flexibility wrt styling (e.g. extend css classes with internal mdl components or create your own).

I believe this approach will give us more flexibility overall and now is a good time to change it (since there's not much stylesheet to be maintained and conversion from LESS was pretty straightforward).

What's a process of submitting CQs? Is that something a PR creator should do?

Cheers! :-)

extensions/ui/org.eclipse.smarthome.ui.basic/ESH-INF/config/config.xml
kubawolanin: Done

extensions/ui/org.eclipse.smarthome.ui.basic/gulpfile.js
kubawolanin: @resetnow @kaikreuzer 🏓

resetnow: I think eventually having MDL building from sources is a good point.
 
>Is that something a PR creator should do?

No, as far as I understand, some people from Eclipse org can do it, including @kaikreuzer

kaikreuzer: Correct, this has to be done by project committers.
Just to clarify: This is a build-time only dependency, i.e. we are not distributing anything from that project, right?

kaikreuzer:
I have created CQ 14162 for gulp-sass approval.

extensions/ui/org.eclipse.smarthome.ui.basic/gulpfile.js
kubawolanin: @kaikreuzer that's right.

kubawolanin:
@kaikreuzer is there any update on the IP check of gulp-sass?
Thanks!

kaikreuzer:
Not yet, waiting for it...

kaikreuzer:
CQ has just been approved!

PR #4147, opened 2017-08-28
automation: add missing optional o.e.jdt.annotation dependency

htreu:


kaikreuzer:
Did the same locally already, so I see the need ;-)

PR #4149, opened 2017-08-28
updated IDE and Maven build to latest Xtext, MWE2 and Tycho plugins

kaikreuzer:


sjka:
Just tried it out: works for me too 👍

PR #4148, opened 2017-08-28
Exposing a Language Server

sjka:
This is a first shot on the Language Server support so that we can jointly work on it. 

The new bundle `o.e.sh.model.lsp` exposes a Language Server on (currently fixed) port 5007.

relates to openhab/openhab-vscode#31
fixes #3898
Signed-off-by: Simon Kaufmann <simon.kfm@googlemail.com>

PR #4150, opened 2017-08-28
Updated README with skipTests flag

kubawolanin:
Thanks @kaikreuzer for [the info](https://github.com/eclipse/smarthome/pull/4148#issuecomment-325430137)! :-)

Signed-off-by: Kuba Wolanin <hi@kubawolanin.com> (github: kubawolanin)


maggu2810:
Doesn't we require a Signed-off-line that fits to the guidelines? (correct braces and without github name)?

kaikreuzer:
I'll correct the syntax when squashing - I think the formal sign-off requirement is fulfilled.

PR #4153, opened 2017-08-29
Optimize the JAX-RS implementation to grab only RESTResources

### LINKED ISSUE #4151: Circular Reference in Automation Component
Similar to #4143, the now used DS implementation from Felix provoked yet another circular dependency:

```
!MESSAGE Circular reference detected trying to get service {org.eclipse.smarthome.automation.module.script.ScriptExtensionProvider}={service.id=136, service.bundleid=88, service.scope=bundle, component.name=org.eclipse.smarthome.automation.module.script.rulesupport.internal.LoaderScriptExtension, component.id=129}
 stack of references: ServiceReference: {org.eclipse.smarthome.automation.module.script.ScriptExtensionProvider}={service.id=136, service.bundleid=88, service.scope=bundle, component.name=org.eclipse.smarthome.automation.module.script.rulesupport.internal.LoaderScriptExtension, component.id=129}
ServiceReference: {org.eclipse.smarthome.automation.module.script.internal.ScriptExtensionManager}={service.id=139, service.bundleid=104, service.scope=bundle, component.name=org.eclipse.smarthome.automation.module.script.internal.ScriptExtensionManager, component.id=137}
```

Checking the code, I do not find any real issue. What looks suspicious is that in the exception, there are two places with:
```
	at com.eclipsesource.jaxrs.publisher.internal.ResourceTracker.addingService(ResourceTracker.java:39)
...
	at com.eclipsesource.jaxrs.publisher.internal.ResourceTracker.addingService(ResourceTracker.java:39)
```

This might be a hint that in fact the OSGi JAX-RS Connector is causing the issue as it activates all services (in order to check whether they have JAX-RS annotations or not).

sjka:
This PR adds another bundle `o.e.sh.io.rest.optimize` which has a dependency to the eclipsesource JAX-RS implementation and registers a service which teaches it some manners. 

In particular, it maintains a whitelist of services/service-interfaces which are relevant in the context of JAX-RS, so that the adaptor does not need anymore to consume each and every service just to check if it has some JAX-RS annotations. 

fixes #4151


maggu2810:
I am not sure if I understand this filter correctly.
Does this filter only allow the object classes in the whitelist to be respected by JAX-RS?
As you add this bundle to the esh-base Karaf feature, I assume that the all non-ESH REST interfaces using JAX-RS in our product will stop working.
Correct?

maggu2810:
Would it be possible to use an expression that allows all object classes but filters the org.eclipse.smarthome.* ones to the one that are part of your whitelist?

sjka:
> As you add this bundle to the esh-base Karaf feature, I assume that the all non-ESH REST interfaces using JAX-RS in our product will stop working.

Yes, fully correct 😄 

The JAX-RS implementation registers a ServiceTracker for _every_ service and whenever a new service is added, it immediately gets an instance of it - just to check if it has any JAX-RS annotations. That basically forces all services to be activated immediately and doesn't leave the SCR any chance to sort the activation order in the way it had in mind. 

The idea of applying the filter only to the ESH namespace would relief at least the pain we have seen in several issues here with our services. Nevertheless, you might run into the same with troubles with others. But that's a use-at-your-own risk-situation then. I'm fine with it and will update this PR accordingly.

maggu2810:
@sjka Thanks.
Another question: Do you know how multiple resource filters are handled by JAX-RS?
Blacklist vs. Whitelist, intersection vs. union

sjka:
I just had a look: it's a the-winner-takes-it-all-situation 😒 :

https://github.com/hstaudacher/osgi-jax-rs-connector/blob/745ebabced77e50170fbc117825caf2fc8276ebd/bundles/com.eclipsesource.jaxrs.publisher/src/com/eclipsesource/jaxrs/publisher/internal/Activator.java#L94-L100

Thinking a little more about the o.e.sh.*-restriction.... Would you consider it an option to set a service property (like e.g. `jax.rs=true`) as well? I fear we might still have quite a bunch of non-ESH service that it will start and scan otherwise.

maggu2810:
> Would you consider it an option to set a service property (like e.g. `jax.rs=true`) as well?

Yes, I can. No problem.

But it is a very "confusion" situation:
* Adding the jaxrs.publisher will force every service to get activated immediately
* Adding a resource filter for jaxrs.publisher is already a little bit in-deterministic as the "winner-takes-it-all"
* If we mask all services not using a special package or the service property we could break third party stuff that package we don't know and that does not set the property. It feels wrong to me that as soon as we add a special ESH bundle to the OSGi runtime other stuff does not work anymore because we require that the property is set. What is wrong with the o.e.sh.* filter? Do you see any problem?

kaikreuzer:
I agree that using a service property is rather ugly and would not solve the issue for third party extensions that are beyond one's own control.

How about this:
- Have a whitelist of ESH services to allow, filter out all other ESH services, but allow all services in other namespaces.
- Let a fragment add an additional property file to the bundle, which provides another whitelist. If this is present, only allow non-ESH services from that whitelist and filter out the rest.

This should give us the flexibility to cover most cases nicely. If only the ESH bundle is present, the behavior is at least better than before and it does not break any other services that might exist in solutions today. If a solution wants to further optimize the filtering, it is able to do so via a fragment.

maggu2810:
I wounder why you are using an activator and register the service programmatically.
Do you need this working without an OSGi Declarative Services implementation?

If we could DS, we could also inject filter strings by consuming other services (sure, also possible with service trackers but not such nicely).
We could also rely on configuration admin to get additional settings injected.

But a fragment should to the job, too.

sjka:
> I wounder why you are using an activator and register the service programmatically.

Because of this limitation:

>  <b>Please Note:</b> Right now the {@link ResourceFilter} service must be registered before the publisher bundle is started. Dynamic behavior may follow in future versions.

Therefore using start-levels looked like a suitable way to guarantee this this requirement is met.

The proposal with the fragments sounds reasonable to me: we can achieve the same result with less surprises and still have some degree of configurability.

sjka:
I updated this PR with the suggested fragment approach. 

> If we could DS, we could also inject filter strings by consuming other services (sure, also possible > with service trackers but not such nicely).
> We could also rely on configuration admin to get additional settings injected.

While using Services and/or Config Admin for the extension would have been nice, my fear is that this will be error prone or a least very unintuitive as the JAX-RS implementation does _not_ support any dynamic (re)configuration currently. Both of these mechanisms would imply this was possible.

sjka:
PS: 

> As you add this bundle to the esh-base Karaf feature, I assume [...]

I mindlessly put it there. Technically it's optional, so having it as a separate feature would also make sense.

As I don't have to maintain a Karaf based solution it's your call.

bundles/io/org.eclipse.smarthome.io.rest.optimize/META-INF/MANIFEST.MF
```diff
@@ -0,0 +1,11 @@
+Manifest-Version: 1.0
+Bundle-ManifestVersion: 2
+Bundle-Name: Eclipse SmartHome REST Interface JAX-RS optimization Bundle
+Bundle-SymbolicName: org.eclipse.smarthome.io.rest.optimize
```
maggu2810: Shouldn't we set singleton to true?

maggu2810:
Hi @sjka, as written above the changes looks good to me.
IMHO the singleton flag for the bundle makes sense because different versions of that bundle on the same OSGi doesn't make sense.

> As I don't have to maintain a Karaf based solution it's your call.

Okay, no prob. As we do  non blacklist any non ESH bundle (package) anymore, it could stay in the base feature.

sjka:
> For the other readers (if there is any), the current filter expression (without a whitelist.txt) looks like: [...]

Thanks for the summary! 

Just for the record: I have added all non-ESH services that ESH relies on. This is useless as long as there is no whitelist.txt, but as soon as there is a whitelist.txt the framework keeps working even if somebody did not guess what it needs internally (which would have been pretty unfair, I think).

> What do you think, should we add the singleton flag to the bundle manifest?

Yes, good point. Done.

maggu2810:
I agree that it make sense to have the standard services in the hard coded whitelist.

PR #4154, opened 2017-08-29
Scheduler: Detect time jumps and reschedule jobs

### LINKED ISSUE #3185: Change to DayLigth saving Time causes stop in timerelated persistence strategies

triller-telekom:
The ExpressionThreadPoolExecutor now calculates the time when he plans to wake up next. If this time is exceeded by 3 seconds, all jobs will be rescheduled.

Fixes #3185

### DIFF
--- a/bundles/core/org.eclipse.smarthome.core.test/META-INF/MANIFEST.MF
+++ b/bundles/core/org.eclipse.smarthome.core.test/META-INF/MANIFEST.MF
@@ -7,6 +7,7 @@ Bundle-Vendor: Eclipse.org/SmartHome
 Fragment-Host: org.eclipse.smarthome.core
 Bundle-RequiredExecutionEnvironment: JavaSE-1.8
 Import-Package: groovy.lang,
+ org.apache.commons.lang.time,
  org.codehaus.groovy.reflection,
  org.codehaus.groovy.runtime,
  org.codehaus.groovy.runtime.callsite,
--- a/bundles/core/org.eclipse.smarthome.core.test/src/test/java/org/eclipse/smarthome/core/scheduler/ExpressionThreadPoolExecutorTest.java
+++ b/bundles/core/org.eclipse.smarthome.core.test/src/test/java/org/eclipse/smarthome/core/scheduler/ExpressionThreadPoolExecutorTest.java
@@ -8,16 +8,46 @@
 package org.eclipse.smarthome.core.scheduler;
 
 import static org.junit.Assert.*;
+import static org.mockito.Mockito.when;
+import static org.mockito.MockitoAnnotations.initMocks;
 
 import java.text.ParseException;
+import java.util.Calendar;
+import java.util.Date;
 
+import org.apache.commons.lang.time.DateUtils;
 import org.eclipse.smarthome.core.scheduler.ExpressionThreadPoolManager.ExpressionThreadPoolExecutor;
+import org.junit.After;
+import org.junit.Before;
 import org.junit.Test;
+import org.mockito.Mock;
+import org.mockito.invocation.InvocationOnMock;
+import org.mockito.stubbing.Answer;
+import org.slf4j.Logger;
+import org.slf4j.LoggerFactory;
 
 public class ExpressionThreadPoolExecutorTest {
 
-    private ExpressionThreadPoolExecutor scheduler = ExpressionThreadPoolManager.getExpressionScheduledPool("test");
+    private ExpressionThreadPoolExecutor scheduler;
     protected boolean success;
+
+    private final Logger logger = LoggerFactory.getLogger(ExpressionThreadPoolExecutorTest.class);
+
+    protected boolean fakeOn = false;
+
+    @Mock
+    DateWrapper dw;
+
+    @Before
+    public void setup() {
+        initMocks(this);
+        scheduler = ExpressionThreadPoolManager.getExpressionScheduledPool("test");
+    }
+
+    @After
+    public void after() {
+        scheduler = null;
+    }
 
     @Test
     public void testScheduleAndRemove() throws ParseException, InterruptedException {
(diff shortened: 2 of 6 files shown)


sjka:
@kgoderis can you please have a look?

kgoderis:
@triller-telekom @sja at first sight - I did not test the code in the IDE - it looks ok

sjka:
Okay, let me ask a very stupid question then...
As this PR claims to fix #3185 which was about the change from ST to DST: Has it been proven that this appeared to the scheduler as a "time jump"? I'm missing this proof in the test-case - and also that it would survive such a ST-DST change now.
From looking at the code (which is hard by itself and is even harder with this huge diff), I'm really not sure whether this PR fixes the right problem - and even if it fixes another problem (i.e. time drift), then I don't feel confident to judge whether it accidentally also fixes the time change or not.

triller-telekom:
@kgoderis Yes you are right, the changes in the `AbstractExpression.java` where just fixes that occurred as "bugs" to me while debugging, but I am fine with reverting my changes in there for now if you want to fix them later. Shall I?

@sjka Indeed this PR fixes a time drift, but a the change from ST to DST and back is one hour. And this drift will be detected once the scheduler will wake up (maximum 60 seconds after the change), which will then reschedule the jobs. @kgoderis Is that also what you understand while reading the code?

kgoderis:
@triller-telekom Indeed, better change back in order not to break things. And I do read/understand the code as you point out. e.g. a change ST->DST is a 1-hour drift that goes beyond the 3 second limit. Now, that being said, I never observed small drifts in the scheduler, so, I guess the 3 second value is somewhat arbitrary ?

triller-telekom:
@kgoderis I have reverted the `AbstractExpression.java`. The 3 seconds are arbitrary, yes. There have to be a few milliseconds anyway, otherwise there will always be a rescheduling since we cannot guarantee that there is not a small drift due to the CPU being busy on another thread. I choose 3 seconds to be on the safe side :)

maggu2810:
It has been years ago as I implemented a scheduler in C but IIRC a DST change does not need to be 1 hour on all locations. There is / has been also changes of e.g. 30 minutes.

triller-telekom:
Yes I also read that there are 30min changes from ST to DST. There are even countries which have a :45 offset to the UTC timezone, crazy...

So indeed we have to catch at least the 30min jump.

sjka:
> Has it been proven that this appeared to the scheduler as a "time jump"?

Guys, why are do you keep ignoring my question so persistently? 😩 

All I've been asking for were these few lines of code:

```java

        CronExpression c = new CronExpression("0 0 6 * * ? *");
        Date now = new Date(2017 - 1900, 02, 25, 18, 0, 0);
        Date next = c.getTimeAfter(now);
        long diff = next.getTime() - now.getTime();
        System.out.println(now);
        System.out.println(next);
        System.out.println(diff + "ms");
        System.out.println(TimeUnit.MILLISECONDS.toHours(diff) + "h");
```

It outputs

```
Sat Mar 25 18:00:00 CET 2017
Sun Mar 26 06:00:00 CEST 2017
39600000ms
11h
```

As you can see, it _correctly_ (!!!) outputs 39600000ms, which equals to 11h. So the CronExpression in fact _considers_ the time-shift!!!

So, my concerns remain valid: What caused the failure of the scheduler? And how is this PR fixing it?

triller-telekom:
You are calling the `getTimeAfter(now)` method on the saturday at 6pm. But the real code just calls `new Date()` every time the scheduler wakes up, see: [here](https://github.com/eclipse/smarthome/blob/master/bundles/core/org.eclipse.smarthome.core/src/main/java/org/eclipse/smarthome/core/scheduler/ExpressionThreadPoolManager.java#L171)

So we might run into [this problem](https://stackoverflow.com/questions/42773913/using-new-java-util-date-returns-an-hour-behind-with-dst) where `new Date()` is off by one hour after the `SYSTEM TIME` has switched.

But if this is the case, I guess my fix here has no chance to detect such a `SYSTEM TIME` change :( So we really would be better off changing to the new Java 8 `ZonedDateTime`. WDYT @kgoderis ?

kgoderis:
> Has it been proven that this appeared to the scheduler as a "time jump"?
> Guys, why are do you keep ignoring my question so persistently? 😩

In fact, you do have valid point here ;-) 

@triller-telekom I think the problem is not coming from a misconfiguration of the host (cfr the stackoverflow.com post - I do like this part of the post : "The old date time classes including Date and Calendar are a bloody mess. Avoid them. They are now legacy, supplanted by the java.time classes." - maybe that is something we have to correct in the whole ESH code base), but from a glitch in the running algorithm

That is, referring back to the orignal post in the forum and so forth, the problem seems to stem from cron expressions that execute at a frequency higher that 1 hour, e.g. their next execution time is less that an hour away, with 1 hour being the assumed shift in CET to CEST. 

So, what I *think* that happens is that while the algorithm is running, the system time suddenly shifts one hour forward, e.g. 2AM -> 3AM. The candidate dates that are/were being calculated in the meantime, are suddenly one hour behind the "new" system time (which we fetch through Date()), and therefore invalidated (we do not return candidates that are in the past) , resulting in those "has no future executions anymore" messages.

I need to plunge into the code. What would be the best way to simulate a DST change on a system?

kgoderis:
*sigh*

For cron expressions we reset the set of candidate dates with every call to getTimeAfter(), so my initial thinking is wrong. This is a real nasty problem then. Maybe it is due to system not switching over to DST as suggested by @triller-telekom but I would be quite a coincidence that so many OH2 users do not have their system switch over correctly.

kaikreuzer:
@triller-telekom What shall we do with this PR?

triller-telekom:
It looks like there were no problems with the last change from DST to ST on the last weekend, so I will close this PR.

PR #4157, opened 2017-08-30
remove some potential NP access

maggu2810:


bundles/automation/org.eclipse.smarthome.automation.core/src/main/java/org/eclipse/smarthome/automation/core/internal/RuleEngine.java
```diff
@@ -1442,7 +1443,7 @@ private void normalizeRuleConfigurations(Rule rule) {
 
     }
 
-    private <T extends Module> void normalizeModuleConfigurations(List<T> modules) {
+    private <T extends Module> void normalizeModuleConfigurations(List<@NonNull T> modules) {
```
maggu2810: If we state that the list must not contain an entry that is equal to null, the Eclipse nullness checker knows that the elements on iterating are non-null, too.
And IMHO it is also good to inform the caller that the list must not contain a null entry.

bundles/automation/org.eclipse.smarthome.automation.module.script.defaultscope/src/main/java/org/eclipse/smarthome/automation/module/script/defaultscope/internal/ItemRegistryDelegate.java
```diff
@@ -62,7 +62,11 @@ public boolean containsValue(Object value) {
 
     @Override
     public State get(Object key) {
-        return itemRegistry.get((String) key).getState();
+        final Item item = itemRegistry.get((String) key);
+        if (item == null) {
+            return null;
```
maggu2810: What should we do if the item is not known (null or undef state, null, illegal argument exception)?

bundles/core/org.eclipse.smarthome.core.thing/src/main/java/org/eclipse/smarthome/core/thing/dto/ThingDTOMapper.java
```diff
@@ -43,11 +43,10 @@ public static ThingDTO map(Thing thing) {
 
         String thingTypeUID = thing.getThingTypeUID().getAsString();
         String thingUID = thing.getUID().toString();
-        @SuppressWarnings("null") // thing.getBridgeUID() is checked against null before use
-        String bridgeUID = thing.getBridgeUID() != null ? thing.getBridgeUID().toString() : null;
+        final ThingUID bridgeUID = thing.getBridgeUID();
 
-        return new ThingDTO(thingTypeUID, thingUID, thing.getLabel(), bridgeUID, channelDTOs,
-                toMap(thing.getConfiguration()), thing.getProperties(), thing.getLocation());
+        return new ThingDTO(thingTypeUID, thingUID, thing.getLabel(), bridgeUID != null ? bridgeUID.toString() : null,
```
maggu2810: Handle it this way we don't need to suppress a warning at all.

extensions/binding/org.eclipse.smarthome.binding.hue/src/main/java/org/eclipse/smarthome/binding/hue/internal/HueThingHandlerFactory.java
```diff
@@ -100,7 +100,9 @@ protected synchronized void removeHandler(ThingHandler thingHandler) {
                 // remove discovery service, if bridge handler is removed
                 HueLightDiscoveryService service = (HueLightDiscoveryService) bundleContext
                         .getService(serviceReg.getReference());
-                service.deactivate();
+                if (service != null) {
```
maggu2810: getService can AFAIK return null

extensions/binding/org.eclipse.smarthome.binding.ntp/src/main/java/org/eclipse/smarthome/binding/ntp/handler/NtpHandler.java
```diff
@@ -134,15 +134,19 @@ public void initialize() {
...
+                        logger.debug("No format set in channel config for {}. Using default format.", stringChannelUID);
+                        dateTimeFormat = new SimpleDateFormat(DATE_PATTERN_WITH_TZ);
+                    }
                 } else {
-                    logger.debug("No format set in channel config for {}. Using default format.", stringChannelUID);
-                    dateTimeFormat = new SimpleDateFormat(DATE_PATTERN_WITH_TZ);
+                    logger.debug("Missing channel: '{}'", stringChannelUID.getId());
                 }
-            } catch (Exception ex) {
+            } catch (RuntimeException ex) {
```
maggu2810: Changed it on the go: We should not catch non runtime exception if there aren't any.

extensions/binding/org.eclipse.smarthome.binding.sonos/src/main/java/org/eclipse/smarthome/binding/sonos/handler/ZonePlayerHandler.java
```diff
@@ -1743,7 +1747,8 @@ public void setRepeat(Command command) {
 
     public Boolean isShuffleActive() {
         return ((stateMap.get("CurrentPlayMode") != null) && stateMap.get("CurrentPlayMode").startsWith("SHUFFLE"))
-                ? true : false;
+                ? true
+                : false;
```
maggu2810: Changed by the formatter

bundles/automation/org.eclipse.smarthome.automation.module.script.defaultscope/src/main/java/org/eclipse/smarthome/automation/module/script/defaultscope/internal/ItemRegistryDelegate.java
kaikreuzer: The Javadoc says `@return element or null if no element was found`, so null should be just right.

PR #4158, opened 2017-08-30
Astro: Sun phase calculation fix

### LINKED ISSUE #4138: Astro binding :  Peculiarities of the Sun phase channel
I am using the Astro binding in my openHAB system, using the Sun / Phase channel as a way to check the "state of daylight" in my rules. 

While testing this I was looking at the log files for a "full cycle" (i.e. a 24 hour period) and I noticed a couple of strange things. 

Here is a snip of my log file:

2017-08-14 03:55:00.003 [ItemStateChangedEvent     ] - AstroSunHome_sPhase_Name changed from UNDEF to NAUTIC_DAWN
2017-08-14 05:02:00.003 [ItemStateChangedEvent     ] - AstroSunHome_sPhase_Name changed from NAUTIC_DAWN to CIVIL_DAWN
2017-08-14 05:50:00.004 [ItemStateChangedEvent     ] - AstroSunHome_sPhase_Name changed from CIVIL_DAWN to SUN_RISE
2017-08-14 05:54:00.002 [ItemStateChangedEvent     ] - AstroSunHome_sPhase_Name changed from SUN_RISE to DAYLIGHT
2017-08-14 13:34:00.002 [ItemStateChangedEvent     ] - AstroSunHome_sPhase_Name changed from DAYLIGHT to NOON
2017-08-14 21:14:00.003 [ItemStateChangedEvent     ] - AstroSunHome_sPhase_Name changed from NOON to SUN_SET
2017-08-14 21:18:00.003 [ItemStateChangedEvent     ] - AstroSunHome_sPhase_Name changed from SUN_SET to CIVIL_DUSK
2017-08-14 22:06:00.007 [ItemStateChangedEvent     ] - AstroSunHome_sPhase_Name changed from CIVIL_DUSK to NAUTIC_DUSK
2017-08-14 23:13:00.003 [ItemStateChangedEvent     ] - AstroSunHome_sPhase_Name changed from NAUTIC_DUSK to ASTRO_DUSK
2017-08-15 00:00:00.023 [ItemStateChangedEvent     ] - AstroSunHome_sPhase_Name changed from ASTRO_DUSK to UNDEF

The first thing I notice is that apparently the Sun / Phase channel will never report NIGHT as a state (contrary to what the documentation states) but instead goes back to UNDEF. 

The next thing is that the Sun / Phase channel can report a stated called NOON (and this is as stated in the documentation). As can been seen in the log this is reported as the phase from 13:34 until SUN_SET. 

As far as I know, however, NOON is a specific point in time (12:00 / midday) and not a period of time, thus (IMO) NOON should not really be reported on the Sun / Phase channel.

gerrieg:
fixes #4138

### DIFF
--- a/extensions/binding/org.eclipse.smarthome.binding.astro/src/main/java/org/eclipse/smarthome/binding/astro/internal/model/Range.java
+++ b/extensions/binding/org.eclipse.smarthome.binding.astro/src/main/java/org/eclipse/smarthome/binding/astro/internal/model/Range.java
@@ -62,10 +62,10 @@ public class Range {
      * Returns true, if the given calendar matches into the range.
      */
     public boolean matches(Calendar cal) {
-        if (start != null && end != null) {
-            return cal.getTimeInMillis() >= start.getTimeInMillis() && cal.getTimeInMillis() < end.getTimeInMillis();
-        }
-        return false;
+        long matchStart = start != null ? start.getTimeInMillis()
+                : DateTimeUtils.truncateToMidnight(cal).getTimeInMillis();
+        long matchEnd = end != null ? end.getTimeInMillis() : DateTimeUtils.truncateToMidnight(cal).getTimeInMillis();
+        return cal.getTimeInMillis() >= matchStart && cal.getTimeInMillis() < matchEnd;
     }
 
     @Override
--- a/extensions/binding/org.eclipse.smarthome.binding.astro/src/main/java/org/eclipse/smarthome/binding/astro/internal/model/Range.java
+++ b/extensions/binding/org.eclipse.smarthome.binding.astro/src/main/java/org/eclipse/smarthome/binding/astro/internal/model/Range.java
@@ -62,9 +62,12 @@ public class Range {
      * Returns true, if the given calendar matches into the range.
      */
     public boolean matches(Calendar cal) {
+        if (start == null && end == null) {
+            return false;
+        }
         long matchStart = start != null ? start.getTimeInMillis()
                 : DateTimeUtils.truncateToMidnight(cal).getTimeInMillis();
-        long matchEnd = end != null ? end.getTimeInMillis() : DateTimeUtils.truncateToMidnight(cal).getTimeInMillis();
+        long matchEnd = end != null ? end.getTimeInMillis() : DateTimeUtils.endOfDayDate(cal).getTimeInMillis();
         return cal.getTimeInMillis() >= matchStart && cal.getTimeInMillis() < matchEnd;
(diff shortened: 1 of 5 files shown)


extensions/binding/org.eclipse.smarthome.binding.astro/src/main/java/org/eclipse/smarthome/binding/astro/internal/model/Range.java
```diff
@@ -62,10 +62,10 @@ public long getDuration() {
...
      */
     public boolean matches(Calendar cal) {
-        if (start != null && end != null) {
-            return cal.getTimeInMillis() >= start.getTimeInMillis() && cal.getTimeInMillis() < end.getTimeInMillis();
-        }
-        return false;
+        long matchStart = start != null ? start.getTimeInMillis()
+                : DateTimeUtils.truncateToMidnight(cal).getTimeInMillis();
+        long matchEnd = end != null ? end.getTimeInMillis() : DateTimeUtils.truncateToMidnight(cal).getTimeInMillis();
+        return cal.getTimeInMillis() >= matchStart && cal.getTimeInMillis() < matchEnd;
```
sjka: If no start/end are given, then both conditions are comparing against `DateTimeUtils.truncateToMidnight(cal)`. How can this ever be true? Doesn't `matchEnd` needs to add 24h?

gerrieg: Range start/end are always for the current day. If no start/end are given (e.g night in northern hemisphere), there is nothing to match and should return false. So it's OK.

sjka: oh, I see. Thanks for the clarification!

sjka: No, wait, I still don't fully get it - can it happen that only one is given? I.e. if there is only a start, then it would also return false (because the second part is using the previous midnight), even if cal is after matchStart.

gerrieg: you are right! end must be truncated to endOfDay and not to midnight. Thanks!!!

sjka:
Thank you!

Issue #3867, opened 2017-07-20
[Tradfri] Exception occurred while informing handler:Task 


cweitkamp:
After several restarts or changes in related *.items files I get a lot of errors like this:
```
2017-07-20 18:24:13.137 [ERROR] [ome.core.thing.link.ThingLinkManager] - Exception occurred while informing handler:Task java.util.concurrent.ScheduledThreadPoolExecutor$ScheduledFutureTask@1ef875f rejected from java.util.concurrent.ScheduledThreadPoolExecutor@47d5df[Terminated, pool size = 0, active threads = 0, queued tasks = 0, completed tasks = 3]
java.util.concurrent.RejectedExecutionException: Task java.util.concurrent.ScheduledThreadPoolExecutor$ScheduledFutureTask@1ef875f rejected from java.util.concurrent.ScheduledThreadPoolExecutor@47d5df[Terminated, pool size = 0, active threads = 0, queued tasks = 0, completed tasks = 3]
	at java.util.concurrent.ThreadPoolExecutor$AbortPolicy.rejectedExecution(ThreadPoolExecutor.java:2047)[:1.8.0_65]
	at java.util.concurrent.ThreadPoolExecutor.reject(ThreadPoolExecutor.java:823)[:1.8.0_65]
	at java.util.concurrent.ScheduledThreadPoolExecutor.delayedExecute(ScheduledThreadPoolExecutor.java:326)[:1.8.0_65]
	at java.util.concurrent.ScheduledThreadPoolExecutor.schedule(ScheduledThreadPoolExecutor.java:533)[:1.8.0_65]
	at java.util.concurrent.ScheduledThreadPoolExecutor.execute(ScheduledThreadPoolExecutor.java:622)[:1.8.0_65]
	at java.util.concurrent.Executors$DelegatedExecutorService.execute(Executors.java:668)[:1.8.0_65]
	at org.eclipse.californium.core.network.CoapEndpoint.runInProtocolStage(CoapEndpoint.java:692)[175:californium-osgi:1.0.5]
	at org.eclipse.californium.core.network.CoapEndpoint.sendRequest(CoapEndpoint.java:390)[175:californium-osgi:1.0.5]
	at org.eclipse.californium.core.CoapClient.send(CoapClient.java:897)[175:californium-osgi:1.0.5]
	at org.eclipse.californium.core.CoapClient.send(CoapClient.java:879)[175:californium-osgi:1.0.5]
[...]

Issue #3685, opened 2017-06-16
[Tradfri] Handler takes more than 5000ms


kaikreuzer:
From my log:
```
21:36:15.095 [INFO ] [me.event.ThingStatusInfoChangedEvent] - 'tradfri:gateway:gwb072bf27b71d' changed from UNINITIALIZED (HANDLER_CONFIGURATION_PENDING) to INITIALIZING
21:36:20.095 [WARN ] [ome.core.thing.internal.ThingManager] - Initializing handler for thing 'tradfri:gateway:gwb072bf27b71d' takes more than 5000ms.
21:36:20.097 [INFO ] [smarthome.event.ThingUpdatedEvent   ] - Thing 'tradfri:gateway:gwb072bf27b71d' has been updated.
21:36:20.101 [INFO ] [me.event.ThingStatusInfoChangedEvent] - 'tradfri:gateway:gwb072bf27b71d' changed from INITIALIZING to UNKNOWN
```
Note that my gateway was in a weird state - it never responded and I had to reboot it.
Nonetheless, we should make sure that the handler initialization does not block in such a case.

kaikreuzer:
What I am also not quite clear about is why we see a ThingUpdatedEvent here...

PR #4159, opened 2017-08-30
Implemented two fixes/improvements for the Tradfri binding

### LINKED ISSUE #3869: [Tradfri] Binding bug
**Expected Behavior**

I expect the switch item to always reflect the status of the light.

**Current Behavior**

A while after configuring the binding and linking the dimmer item with the bulb the status of the light is no longer reflected. The light can be controlled through the dimmer item but its status is no longer update if the light is switched on/off or the brightness changes using the Tradfri Android app or the physical remote/dimmer.

**Possible Solution**

**Steps to Reproduce (for Bugs)**

Configure binding
Assign dimmer item to "tradfri channel"
Use as normal. Check using the physical remote and app and the status updates correctly on OpenHAB.
A while later....
Binding works to control the light but is no longer receiving status update.
Context

I cannot get the status of the light as it stops changing to reflect the actual status if turned on using the IKEA Tradfri remote. The IKEA Tradfri app updates immediately as soon as I turn the light on/off from within OpenHAB so I know the gateway is working fine.

**Your Environment**

Using OpenHAB 2.1 from the official repo on a Raspberry Pi 2

**Environment name and version (e.g. Chrome 39, node.js 5.4, Java 8, ...):**
Using Google Chrome, Firefox and Opera

**Operating System and version (desktop or mobile, Windows 10, Raspbian Jessie, ...):**
Raspbian lite (latest release as of 19/07/2017

AlexanderKostadinov28:
1. **`TradfriDiscoveryService.java`** - Removal of older discovery results implemented for `TradfriDiscoveryService#startScan()`, so that devices are instantaneously retrieved in inbox after deletion (e.g. if pairing process is invoked directly afterwards).
2. **`TradfriLightHandler.java`** - Fixed bulb channels not receiving updates applied to the physical device from remote control (fixes #3869, but the issue should be also reproduced with bridge remaining `ONLINE`). 
The problem was that, when at some point the bridge goes `OFFLINE` for some reason, after it and the bulb become `ONLINE` again, `coapClient.startObserve(this)` was not called for the bulb while the observe relation, on the other hand, has timed out. This is why this logic is implemented in `public void bridgeStatusChanged(ThingStatusInfo bridgeStatusInfo)` for the light handler.

### DIFF
--- a/extensions/binding/org.eclipse.smarthome.binding.tradfri/src/main/java/org/eclipse/smarthome/binding/tradfri/handler/TradfriLightHandler.java
+++ b/extensions/binding/org.eclipse.smarthome.binding.tradfri/src/main/java/org/eclipse/smarthome/binding/tradfri/handler/TradfriLightHandler.java
@@ -23,6 +23,7 @@ import org.eclipse.smarthome.core.thing.ChannelUID;
 import org.eclipse.smarthome.core.thing.Thing;
 import org.eclipse.smarthome.core.thing.ThingStatus;
 import org.eclipse.smarthome.core.thing.ThingStatusDetail;
+import org.eclipse.smarthome.core.thing.ThingStatusInfo;
 import org.eclipse.smarthome.core.thing.binding.BaseThingHandler;
 import org.eclipse.smarthome.core.types.Command;
 import org.eclipse.smarthome.core.types.RefreshType;
@@ -145,6 +146,17 @@ public class TradfriLightHandler extends BaseThingHandler implements CoapCallbac
             
             logger.debug("Updating thing for lightId {} to state {dimmer: {}, colorTemp: {}, devicefirmware: {}, modelId: {}, vendor: {}}"
                     , state.getLightId(), dimmer, colorTemp, devicefirmware, modelId, vendor);
+        }
+    }
+    
+    @Override
+    public void bridgeStatusChanged(ThingStatusInfo bridgeStatusInfo) {
+        super.bridgeStatusChanged(bridgeStatusInfo);
+
+        if (bridgeStatusInfo.getStatus() == ThingStatus.ONLINE) {
+            scheduler.schedule(() -> {
+                coapClient.startObserve(this);
+            }, 0, TimeUnit.SECONDS);
         }
     }
 
--- a/extensions/binding/org.eclipse.smarthome.binding.tradfri/src/main/java/org/eclipse/smarthome/binding/tradfri/internal/discovery/TradfriDiscoveryService.java
+++ b/extensions/binding/org.eclipse.smarthome.binding.tradfri/src/main/java/org/eclipse/smarthome/binding/tradfri/internal/discovery/TradfriDiscoveryService.java
@@ -48,6 +48,7 @@ public class TradfriDiscoveryService extends AbstractDiscoveryService implements
 
     @Override
     protected void startScan() {
+        removeOlderResults(getTimestampOfLastScan());
         handler.startScan();
     }
 


extensions/binding/org.eclipse.smarthome.binding.tradfri/src/main/java/org/eclipse/smarthome/binding/tradfri/handler/TradfriGatewayHandler.java
```diff
@@ -79,17 +79,40 @@ public void handleCommand(ChannelUID channelUID, Command command) {
     @Override
     public void initialize() {
         GatewayConfig configuration = getConfigAs(GatewayConfig.class);
+
+        if (isConfigurationValid(configuration)) {
+            scheduler.schedule(new Runnable() {
```
kaikreuzer: Could you use Java8 syntax here? (`schedule(() -> {... }, 0, TimeUnit.SECONDS);`)

extensions/binding/org.eclipse.smarthome.binding.tradfri/src/main/java/org/eclipse/smarthome/binding/tradfri/handler/TradfriGatewayHandler.java
```diff
@@ -79,17 +79,40 @@ public void handleCommand(ChannelUID channelUID, Command command) {
...
+
+                @Override
+                public void run() {
+                    try {
+                        setupGateway(configuration);
+                    } catch (final Throwable t) {
```
kaikreuzer: Do we really need this catch?

extensions/binding/org.eclipse.smarthome.binding.tradfri/src/main/java/org/eclipse/smarthome/binding/tradfri/handler/TradfriGatewayHandler.java
```diff
@@ -79,17 +79,40 @@ public void handleCommand(ChannelUID channelUID, Command command) {
...
     public void initialize() {
         GatewayConfig configuration = getConfigAs(GatewayConfig.class);
+
+        if (isConfigurationValid(configuration)) {
+            scheduler.schedule(new Runnable() {
+
+                @Override
+                public void run() {
+                    try {
+                        setupGateway(configuration);
```
kaikreuzer: Did you analyze WHY this can take longer than 5 seconds?
After all, the code only initialises some Java instances where I do not expect any long running operations. The remote calls are already done in a separately scheduled job (`startScan`).

extensions/binding/org.eclipse.smarthome.binding.tradfri/src/main/java/org/eclipse/smarthome/binding/tradfri/handler/TradfriLightHandler.java
```diff
@@ -147,6 +148,17 @@ public void onUpdate(JsonElement data) {
...
     }
+    
+    @Override
+    public void bridgeStatusChanged(ThingStatusInfo bridgeStatusInfo) {
+        super.bridgeStatusChanged(bridgeStatusInfo);
+
+        if (bridgeStatusInfo.getStatus() == ThingStatus.ONLINE) {
+            scheduler.schedule(() -> {
+                coapClient.startObserve(this);
+            }, 3, TimeUnit.SECONDS);
```
kaikreuzer: Why should we wait 3 seconds here?

extensions/binding/org.eclipse.smarthome.binding.tradfri/src/main/java/org/eclipse/smarthome/binding/tradfri/handler/TradfriGatewayHandler.java
AlexanderKostadinov28: Will do it.

extensions/binding/org.eclipse.smarthome.binding.tradfri/src/main/java/org/eclipse/smarthome/binding/tradfri/handler/TradfriGatewayHandler.java
AlexanderKostadinov28: I did not perform deeper analysis, since I could not reproduce initialization taking more than 5 seconds (all my trials resulted in the process taking no more than a second). However, since you said in the ticket that you tested on a HB in a weird state, I implemented a mechanism to prevent possible stall and make sure initialization always passes within the 5-second interval.

extensions/binding/org.eclipse.smarthome.binding.tradfri/src/main/java/org/eclipse/smarthome/binding/tradfri/handler/TradfriGatewayHandler.java
AlexanderKostadinov28: No, I will remove this.

extensions/binding/org.eclipse.smarthome.binding.tradfri/src/main/java/org/eclipse/smarthome/binding/tradfri/handler/TradfriLightHandler.java
AlexanderKostadinov28: I re-used the logic from `initialize()` which to me makes sense.

AlexanderKostadinov28:
Considering your change request comment, observing the logs proved that the bridge always went OFFLINE and then ONLINE before remote actions on the bulb become not reflected on the channels' states.

extensions/binding/org.eclipse.smarthome.binding.tradfri/src/main/java/org/eclipse/smarthome/binding/tradfri/handler/TradfriLightHandler.java
kaikreuzer: Imho in initialize this is done because the coap client is instantiated right before, so it gives it some time before doing the first request.
At this code here, we do not do any changes to the coap client instance, so I do not see any reason to postpone the observe request.

extensions/binding/org.eclipse.smarthome.binding.tradfri/src/main/java/org/eclipse/smarthome/binding/tradfri/handler/TradfriGatewayHandler.java
kaikreuzer: I would avoid introducing async scheduled jobs just because of some guess. Looking at the log of #3685, it might even have hung in the configuration checks, so I am not sure if this code change here would at all address the issue. I would therefore rather refrain from this change.

kaikreuzer:
What about my other question?

> Also, could you explain how this fix is related to #3867? I cannot see a link to a missing observe subscription in it.

AlexanderKostadinov28:
Considering issue #3867, I had the impression it is a mirror of #3869, but in a deeper look it is different - so it is not covered and I updated the PR description.

extensions/binding/org.eclipse.smarthome.binding.tradfri/src/main/java/org/eclipse/smarthome/binding/tradfri/handler/TradfriGatewayHandler.java
AlexanderKostadinov28: I understand your opinion, but what do you think about moving the configuration check plus gateway setup in the scheduled job altogether - in such case it is going to make sense and prevent from initialization slow-down. Furthermore, I am not that sure we can reproduce this whatsoever. :)

kaikreuzer: >  moving the configuration check plus gateway setup in the scheduled job altogether

No, this does not make sense, because the framework already calls `initialize()` in a separate thread. Note that the 5s warning is really just a warning and the system fully recovers from it and continues - so instead of hiding the message, we should rather keep it as a mechanism to hint us at potential problems.

AlexanderKostadinov28: Okay, in such a case, I also believe this is to be reverted. I will do that shortly.

Issue #4161, opened 2017-08-30
Visual Studio and "configuration model x.rules is either empty or cannot be parsed correctly" error


Dan1001:
Hi,

I'm running OH2.2 snapshot on Ubuntu 14.04. I usually use SublimeText as an editor, editing locally. When I use Visual Studio and the OH extension (also locally) I get the above error when editing rules.

I appreciate a number of people have experienced similar issues with a variety of editors when editing across a Samba networked drive, but a local problem seems unusual, and it was suggested I raise as an issue (see [here](https://community.openhab.org/t/visual-studio-code-openhab-extension/30205/72)).

thanks!

Dan

kaikreuzer:
Could you post a log file with exceptions here?

Dan1001:
Sure - do you mean just:

2017-08-30 14:04:43.325 [INFO ] [el.core.internal.ModelRepositoryImpl] - Refreshing model 'sonos.rules'
2017-08-30 14:04:43.326 [WARN ] [el.core.internal.ModelRepositoryImpl] - Configuration model 'sonos.rules' is either empty or cannot be parsed correctly!
2017-08-30 14:04:43.969 [INFO ] [el.core.internal.ModelRepositoryImpl] - Loading model 'sonos.rules'

or something else?

Dan

kaikreuzer:
That is a well known issue when using Samba shares - there simply is a moment in which the file is empty and thus cannot be parsed. So the warning is technically correct.
But your log shows that it is correctly loaded as soon as the content is available. So I expect that everything works, right?

Dan1001:
The oddity is that this isn't a Samba share - I'm locally editing files on a local drive, and those files aren't shared over the network. 

As this is a live system I stopped using Visual Studio when I saw the error - but I'll play with it a little and let you know if I see any real problems.

Dan1001:
On second thoughts, get a series of errors when I edit an items file:

2017-08-30 14:58:22.347 [INFO ] [el.core.internal.ModelRepositoryImpl] - Refreshing model 'sensors.items'
2017-08-30 14:58:22.353 [WARN ] [ore.common.registry.AbstractRegistry] - org.eclipse.smarthome.core.library.items.NumberItem with key 'CoffeePower' already exists! Failed to add a second with the same UID!
[and so on for each item]

then:

2017-08-30 14:58:22.571 [INFO ] [el.core.internal.ModelRepositoryImpl] - Refreshing model 'sensors.items'

[...]

PR #4163, opened 2017-08-30
Fix GenericItemProvider missing added event on startup

### LINKED ISSUE #617: Having multiple *.items configuration files results in duplicate item binding processing.
migrated from Bugzilla [#423465](https://bugs.eclipse.org/bugs/show_bug.cgi?id=423465)
status NEW severity _normal_ in component _DSL_ for _---_
Reported in version _unspecified_ on platform _All_
Assigned to: Project Inbox

On 2013-12-06 12:58:15 -0500, Kai Kreuzer wrote:

> Moved from https://code.google.com/p/openhab/issues/detail?id=173

On 2015-04-27 10:54:12 -0400, Dennis Nobel wrote:

> Not sure if this might be solved already because of some GenericItemProvider refactorings?

### LINKED ISSUE #4162: NPE in GenericItemProvider
I got this NPE in the logs:

```
15:52:09.793 INFO  o.e.s.m.c.i.ModelRepositoryImpl[:130] - Refreshing model 'demo.items'
15:52:09.805 ERROR o.e.s.m.c.i.f.FolderObserver[:246] - Error handling update of file '/Users/henning/work/smarthome-master2/git/smarthome/distribution/smarthome/conf/items/demo.items': null.
	at org.eclipse.smarthome.model.item.internal.GenericItemProvider.toItemMap(GenericItemProvider.java:388)
	at org.eclipse.smarthome.model.item.internal.GenericItemProvider.modelChanged(GenericItemProvider.java:354)
	at org.eclipse.smarthome.model.core.internal.ModelRepositoryImpl.notifyListeners(ModelRepositoryImpl.java:287)
	at org.eclipse.smarthome.model.core.internal.ModelRepositoryImpl.addOrRefreshModel(ModelRepositoryImpl.java:136)
	at org.eclipse.smarthome.model.core.internal.folder.FolderObserver.checkFile(FolderObserver.java:234)
	at org.eclipse.smarthome.model.core.internal.folder.FolderObserver.processWatchEvent(FolderObserver.java:297)
	at org.eclipse.smarthome.core.service.WatchQueueReader.run(WatchQueueReader.java:204)
```

Steps to reproduce:

- Start ESH in the IDE using `SmartHome-Runtime.launch`
- Edit `distribution/smarthome/conf/items/demo.items`
- add a label to one of the items like `String DemoString "Demo String"`
- save file

--> NPE

htreu:
This patch makes the GenericItemProvider independent from the ADDED event the ModelRepositoryImpl emits on startup. When the GenericItemProvider reads its models it will add the items to its internal map and prevent NPEs on further modified events.

Fixes #4162.

Also:
- fixes duplicate added events for items see https://github.com/eclipse/smarthome/issues/4161#issuecomment-326000211 and closes #617
- minor cosmetics

### DIFF
--- a/bundles/model/org.eclipse.smarthome.model.item/src/org/eclipse/smarthome/model/item/internal/GenericItemProvider.java
+++ b/bundles/model/org.eclipse.smarthome.model.item/src/org/eclipse/smarthome/model/item/internal/GenericItemProvider.java
@@ -9,11 +9,14 @@ package org.eclipse.smarthome.model.item.internal;
 
 import java.util.ArrayList;
 import java.util.Collection;
+import java.util.Collections;
 import java.util.HashMap;
 import java.util.List;
 import java.util.Locale;
 import java.util.Map;
 import java.util.concurrent.ConcurrentHashMap;
+import java.util.function.Function;
+import java.util.stream.Collectors;
 
 import org.apache.commons.lang.StringUtils;
 import org.eclipse.emf.common.util.EList;
@@ -67,9 +70,6 @@ public class GenericItemProvider extends AbstractProvider<Item>
 
     private Integer rank;
 
-    public GenericItemProvider() {
-    }
-
     protected void activate(Map<String, Object> properties) {
         Object serviceRanking = properties.get(Constants.SERVICE_RANKING);
         if (serviceRanking instanceof Integer) {
@@ -86,6 +86,12 @@ public class GenericItemProvider extends AbstractProvider<Item>
 
     public void setModelRepository(ModelRepository modelRepository) {
         this.modelRepository = modelRepository;
+
+        // process models which are already parsed by modelRepository:
+        for (String modelName : modelRepository.getAllModelNamesOfType("items")) {
+            modelChanged(modelName, EventType.ADDED);
+        }
+
         modelRepository.addModelRepositoryChangeListener(this);
     }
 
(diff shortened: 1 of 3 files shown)


bundles/model/org.eclipse.smarthome.model.item/src/org/eclipse/smarthome/model/item/internal/GenericItemProvider.java
```diff
@@ -157,6 +157,8 @@ public void removeBindingConfigReader(BindingConfigReader reader) {
                 }
             }
         }
+        itemsMap.put(modelName, items);
```
kaikreuzer: This looks like a brute force fix to me. Note that the itemsMap is also filled in `modelChanged`, so you are now doing it several times.
Wouldn't the better place the `getAll()` method as this is the one that gets called (once) before added events are received? And if this would additionally check, whether an entry in the map already exists, it would be even nicer.

htreu: You are right and in addition the line introduces just another bug when multiple *.items models are present. Only one model would be written to itemsMap. Update is on the way.

htreu: Oh noes, wrote that before I had the first coffee... Its a map idiot (<- note to myself).

PR #4165, opened 2017-08-30
Rest API: possibility to add multiple items at once

### LINKED ISSUE #4060: [REST API] Ability to add/edit multiple items within one request
## Background

I'm working on a [home structure generator](https://github.com/openhab/openhab-core/pull/172) for openHAB. It allows you to create an Items collection with simple point-and-click. 

There are two forms of generated Items:
* Text file - Xtext based `*.items` file
* REST API call - Items are structured as objects that [API](demo.openhab.org/doc/index.html) understands

In order to accomplish the feature from the second bulletpoint I need to be able to pass an array of Items in one PUT request.
Otherwise hundreds of requests would be needed for each Item separately.

## Problem statement

According to [the REST docs](demo.openhab.org/doc/index.html) I can do a `PUT /items/{itemname}` to add a new Item.
There's no option to add an array of Items within one request.

See this comment: https://github.com/openhab/openhab-core/pull/172#issuecomment-322549324

## Possible solution

REST API should allow to receive `PUT /items` with the following parameter:

parameter: `body`
value:
```json
[{
  "type": "string",
  "name": "string",
  "label": "string",
  "category": "string",
  "tags": [
    "string"
  ],
  "groupNames": [
    "string"
  ],
  "groupType": "string",
  "function": {
    "name": "string",
    "params": [
      "string"
    ]
  }
}, {
  "type": "string",
  "name": "string",
  "label": "string",
  "category": "string",
  "tags": [
    "string"
  ],
  "groupNames": [
    "string"
  ],
  "groupType": "string",
  "function": {
    "name": "string",
    "params": [
      "string"
    ]
  }
},
...
]
```

triller-telekom:
Fixes #4060


bundles/io/org.eclipse.smarthome.io.rest.core/src/main/java/org/eclipse/smarthome/io/rest/core/item/ItemResource.java
```diff
@@ -385,7 +390,7 @@ public Response removeMember(@PathParam("itemName") @ApiParam(value = "item name
             @ApiResponse(code = 404, message = "Item not found or item is not editable.") })
     public Response removeItem(@PathParam("itemname") @ApiParam(value = "item name", required = true) String itemname) {
 
-        if (managedItemProvider.remove(itemname) == null) {
+        if (itemname == null || managedItemProvider.remove(itemname) == null) {
```
maggu2810: Hm, shouldn't we respond with a `bad request` if the item name is not given?
The parameter is marked "required" and if it is null, it is missing on the request.

```Java
if (itemname == null) {
    return Response.status(Status.BAD_REQUEST).build();
}
```

bundles/io/org.eclipse.smarthome.io.rest.core/src/main/java/org/eclipse/smarthome/io/rest/core/item/ItemResource.java
```diff
@@ -513,6 +506,103 @@ public Response createOrUpdateItem(
...
+
+        // build response
+        List<JsonObject> responseList = new ArrayList<>();
+
+        for (GroupItemDTO item : wrongTypes) {
+            responseList.add(buildStatusObject(item.name, "error", "Received HTTP PUT request at '" + uriInfo.getPath()
+                    + "' with an invalid item type '" + item.type + "'."));
+        }
+        for (ActiveItem item : failedItems) {
+            responseList.add(buildStatusObject(item.getName(), "error", "Cannot update non-managed Item"));
```
maggu2810: Shouldn't we use lower case ("item")?

bundles/io/org.eclipse.smarthome.io.rest.core/src/main/java/org/eclipse/smarthome/io/rest/core/item/ItemResource.java
triller-telekom: Sure, I did not think about this, I just wanted to fix the possible NPE which the IDE pointed me to :)

bundles/io/org.eclipse.smarthome.io.rest.core/src/main/java/org/eclipse/smarthome/io/rest/core/item/ItemResource.java
```diff
@@ -385,6 +390,10 @@ public Response removeMember(@PathParam("itemName") @ApiParam(value = "item name
             @ApiResponse(code = 404, message = "Item not found or item is not editable.") })
     public Response removeItem(@PathParam("itemname") @ApiParam(value = "item name", required = true) String itemname) {
 
+        if (itemname == null) {
+            return Response.status(Status.BAD_REQUEST).build();
```
kaikreuzer: Can this situation at all happen? After all, it is a part of the URI path, so how could it be null?
If it can happen, then please also add 400 as a potential response to the swagger annotations of this method.

bundles/io/org.eclipse.smarthome.io.rest.core/src/main/java/org/eclipse/smarthome/io/rest/core/item/ItemResource.java
```diff
@@ -513,6 +510,103 @@ public Response createOrUpdateItem(
...
+                // item does not yet exist, create it
+                managedItemProvider.add(activeItem);
+                createdItems.add(activeItem);
+            } else if (managedItemProvider.get(itemName) != null) {
+                // item already exists as a managed item, update it
+                managedItemProvider.update(activeItem);
+                updatedItems.add(activeItem);
+            } else {
+                // Item exists but cannot be updated
+                logger.warn("Cannot update existing item '{}', because is not managed.", itemName);
```
kaikreuzer: typo: because _it_ is not managed.

bundles/io/org.eclipse.smarthome.io.rest.core/src/main/java/org/eclipse/smarthome/io/rest/core/item/ItemResource.java
maggu2810: Ah, I didn't realized that it has been marked as a `@PathParam`. Don't know if it is then possible to be null.

triller-telekom: Looking at this problem again I think you are right. Since `itemname` is part of the path it cannot be `null` here. So I added `@NonNull` to satisfy the IDE :) (which was my original intention by touching this line at all)

PR #4166, opened 2017-08-31
the reconnect timing should be constant after object creation

maggu2810:
The reconnect timing should not be changed after the periodic reconnect strategy has been connected.
Using unmodifiable members will make that values thread safe.
The caller also does the internal state the reconnect strategy currently is using, so it make IMHO no sense to change some timing without knowing at which time it will affect the
current behavior.
If the MQTT broker connection should use another timed strategy you could create a new reconnect strategy and set it for the existing connection.


kaikreuzer:
Could you please rebase the branch?

maggu2810:
done

maggu2810:
Would it be possible to merge this one?

I will implement https://github.com/eclipse/smarthome/pull/3980#issuecomment-326222638 (allow to start-stop-start-stop a reconnect strategy) and this will result in a conflict again if we merge one of the both.

maggu2810:
Thanks

PR #4168, opened 2017-08-31
allows to start-stop-start-stop-... a reconnect strategy

maggu2810:
A MQTT Broker Connection must be able to handle start and stop iterations.

Related to: https://github.com/eclipse/smarthome/pull/3980#issuecomment-326222638


maggu2810:
@davidgraeff WDYT about?

davidgraeff:
I would like to adapt the unit test in a later pr, but I need a way to query the strategy if it is started. Should we introduce a isStarted() function?

maggu2810:
I don't care. But IMHO this should be added to the specific implementation only because it should not be necessary to have such a method in general.

davidgraeff:
Can you add it to the PeriodicStrategy?

maggu2810:
> Can you add it to the PeriodicStrategy?

done

PR #4169, opened 2017-08-31
POC: Encrypted credentials

paulianttila:
I quickly tested idea to have encrypted credentials on property files. Idea was to store passphrase in the ENV variable. Or even make it more secure and ask passphrase during startup (currently from karaf console), then passphrase is not stored on target system disk at all but only in RAM. Karaf console command extension can be used to encrypt credentials which can be then stored in property files.

I also implemented simple crypto action in openHAB side, which could be used to decrypt credentials from rules. In fact, rules was my motivation for this as I use some rules to fetch iPhone location info, but I don't like to have my apple id credentials in plain text format in rule files.

This PR is just a POC and starting point for the discussion around the topic.

Signed-off-by: Pauli Anttila <pauli.anttila@gmail.com>

### DIFF
--- a/bundles/config/org.eclipse.smarthome.config.dispatch/src/main/java/org/eclipse/smarthome/config/dispatch/internal/ConfigDispatcher.java
+++ b/bundles/config/org.eclipse.smarthome.config.dispatch/src/main/java/org/eclipse/smarthome/config/dispatch/internal/ConfigDispatcher.java
@@ -29,6 +29,7 @@ import org.apache.commons.io.IOUtils;
 import org.apache.commons.lang.StringUtils;
 import org.eclipse.smarthome.config.core.ConfigConstants;
 import org.eclipse.smarthome.core.service.AbstractWatchService;
+import org.eclipse.smarthome.io.crypto.Crypto;
 import org.osgi.service.cm.Configuration;
 import org.osgi.service.cm.ConfigurationAdmin;
 import org.osgi.service.cm.ManagedService;
@@ -47,18 +48,18 @@ import org.slf4j.LoggerFactory;
  * <p>
  * The format of the configuration file is similar to a standard property file, with the exception that the property
  * name can be prefixed by the service pid of the {@link ManagedService}:
- * 
+ *
  * <p>
  * &lt;service-pid&gt;:&lt;property&gt;=&lt;value&gt;
- * 
+ *
  * <p>
  * In case the pid does not contain any ".", the default service pid namespace is prefixed, which can be defined by the
  * program argument "smarthome.servicepid" (default is "org.eclipse.smarthome").
- * 
+ *
  * <p>
  * If no pid is defined in the property line, the default pid namespace will be used together with the filename. E.g. if
  * you have a file "security.cfg", the pid that will be used is "org.eclipse.smarthome.security".
- * 
+ *
  * <p>
  * Last but not least, a pid can be defined in the first line of a cfg file by prefixing it with "pid:", e.g.
  * "pid: com.acme.smarthome.security".
@@ -108,6 +109,7 @@ public class ConfigDispatcher extends AbstractWatchService {
     private final Logger logger = LoggerFactory.getLogger(ConfigDispatcher.class);
 
     private ConfigurationAdmin configAdmin;
+    private Crypto crypto;
 
     @Override
     public void activate() {
@@ -128,6 +130,14 @@ public class ConfigDispatcher extends AbstractWatchService {
 
     protected void unsetConfigurationAdmin(ConfigurationAdmin configAdmin) {
         this.configAdmin = null;
+    }
+
+    protected void setCryptoModule(Crypto crypto) {
+        this.crypto = crypto;
+    }
+
+    protected void unsetCryptoModule(Crypto crypto) {
+        this.crypto = null;
     }
 
     @Override
@@ -252,6 +262,22 @@ public class ConfigDispatcher extends AbstractWatchService {
 
             String property = contents[1];
             String value = contents[2];
+            logger.debug("'{}'='{}'", property, value);
+
+            if (value != null) {
+                if (value.startsWith("ENC(") && value.endsWith(")")) {
+                    String val = value.substring(value.indexOf("(") + 1, value.indexOf(")"));
+                    try {
+                        String decryptedText = crypto.decrypt(val);
+                        logger.debug("decryptedText: {}", decryptedText); // TODO: remove
+                        value = decryptedText;
+                    } catch (Exception e) {
+                        logger.warn("Error occured during property '" + property + "' value decryption. ", e);
+                        continue;
+                    }
+                }
+            }
+
             Configuration configuration = configAdmin.getConfiguration(pid, null);
             if (configuration != null) {
                 Dictionary configProperties = configMap.get(configuration);
--- a/bundles/io/org.eclipse.smarthome.io.console/src/main/java/org/eclipse/smarthome/io/console/internal/extension/CryptoConsoleCommandExtension.java
+++ b/bundles/io/org.eclipse.smarthome.io.console/src/main/java/org/eclipse/smarthome/io/console/internal/extension/CryptoConsoleCommandExtension.java
@@ -1,0 +1,87 @@
+/**
+ * Copyright (c) 2014-2017 by the respective copyright holders.
+ * All rights reserved. This program and the accompanying materials
+ * are made available under the terms of the Eclipse Public License v1.0
+ * which accompanies this distribution, and is available at
+ * http://www.eclipse.org/legal/epl-v10.html
+ */
+package org.eclipse.smarthome.io.console.internal.extension;
+
+import java.util.Arrays;
+import java.util.List;
+
+import org.eclipse.smarthome.io.console.Console;
+import org.eclipse.smarthome.io.console.extensions.AbstractConsoleCommandExtension;
+import org.eclipse.smarthome.io.crypto.Crypto;
+
+/**
+ * Console command extension for crypto module functions.
+ *
+ * @author Pauli Anttila - Initial contribution
+ *
+ */
+public class CryptoConsoleCommandExtension extends AbstractConsoleCommandExtension {
+
+    private Crypto crypto;
+
+    private static final String SUBCMD_ENCRYPT = "encrypt";
+    private static final String SUBCMD_SET_PASSPHRASE = "setpassphrase";
+
+    public CryptoConsoleCommandExtension() {
+        super("crypto", "Crypto utils");
+    }
+
+    protected void setCryptoModule(Crypto crypto) {
+        this.crypto = crypto;
+    }
+
+    protected void unsetCryptoModule(Crypto crypto) {
+        this.crypto = null;
+    }
+
+    @Override
+    public List<String> getUsages() {
+        return Arrays
+                .asList(new String[] { buildCommandUsage(SUBCMD_ENCRYPT + " <plaintext>", "Encrypt plain text value"),
+                        buildCommandUsage(SUBCMD_SET_PASSPHRASE + " <passphrase>", "Set encryption passphrase") });
+    }
+
+    @Override
+    public void execute(String[] args, Console console) {
+        if (args.length > 0) {
+            String subCommand = args[0];
+            switch (subCommand) {
+                case SUBCMD_ENCRYPT:
+                    if (args.length > 1) {
+                        String value = args[1];
+                        try {
+                            String encryptedText = crypto.encrypt(value);
+                            console.println("Encrypted value: " + encryptedText);
+                        } catch (Exception e) {
+                            console.println("Error occured during value encryption: " + e.getMessage());
+                        }
+                    } else {
+                        console.println("Specify the plaintext to encrypt: " + this.getCommand() + " " + SUBCMD_ENCRYPT
+                                + " <plaintext>");
+                    }
+                    break;
+                case SUBCMD_SET_PASSPHRASE:
+                    if (args.length > 1) {
+                        String key = args[1];
+                        crypto.setEncPassphrase(key.getBytes());
+                    } else {
+                        console.println("Specify the passphrase: " + this.getCommand() + " " + SUBCMD_SET_PASSPHRASE
+                                + " <passphrase>");
+                    }
+                    break;
+                default:
+                    console.println("Unknown command '" + subCommand + "'");
+                    printUsage(console);
+                    break;
+            }
+        } else {
+            printUsage(console);
+        }
+    }
+
+}
--- a/bundles/io/org.eclipse.smarthome.io.encryption/.settings/org.eclipse.jdt.core.prefs
+++ b/bundles/io/org.eclipse.smarthome.io.encryption/.settings/org.eclipse.jdt.core.prefs
@@ -1,0 +1,8 @@
+eclipse.preferences.version=1
+org.eclipse.jdt.core.compiler.codegen.inlineJsrBytecode=enabled
+org.eclipse.jdt.core.compiler.codegen.targetPlatform=1.8
+org.eclipse.jdt.core.compiler.compliance=1.8
+org.eclipse.jdt.core.compiler.problem.assertIdentifier=error
+org.eclipse.jdt.core.compiler.problem.enumIdentifier=error
+org.eclipse.jdt.core.compiler.problem.forbiddenReference=warning
+org.eclipse.jdt.core.compiler.source=1.8
--- a/bundles/io/org.eclipse.smarthome.io.encryption/.settings/org.maven.ide.eclipse.prefs
+++ b/bundles/io/org.eclipse.smarthome.io.encryption/.settings/org.maven.ide.eclipse.prefs
@@ -1,0 +1,9 @@
+#Fri Feb 19 22:30:18 CET 2010
+activeProfiles=
+eclipse.preferences.version=1
+fullBuildGoals=process-test-resources
+includeModules=false
+resolveWorkspaceProjects=true
+resourceFilterGoals=process-resources resources\:testResources
+skipCompilerPlugin=true
+version=1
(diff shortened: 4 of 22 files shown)


bundles/io/org.eclipse.smarthome.io.encryption/src/main/java/org/eclipse/smarthome/io/crypto/internal/CryptoActivator.java
```diff
@@ -0,0 +1,65 @@
...
+ */
+public final class CryptoActivator implements BundleActivator {
+
+    private static byte[] encKey;
+
+    /**
+     * Called whenever the OSGi framework starts our bundle
+     */
+    @Override
+    public void start(BundleContext bc) throws Exception {
```
triller-telekom: How do you ensure that this bundle is started before all the other bundles that want to read their (potentially encrypted) configuration? There are some discussions in issue #1896 already, but it is not implemented (yet).

bundles/io/org.eclipse.smarthome.io.encryption/src/main/java/org/eclipse/smarthome/io/crypto/internal/CryptoActivator.java
```diff
@@ -0,0 +1,65 @@
...
+
+        String key = System.getenv(CryptoModule.ENV_VARIABLE_ENC_PASSPHRASE);
+        if ("ASK".equalsIgnoreCase(key)) {
```
triller-telekom: How about also adding a third option to pass the password as a command line argument? Of course the OS has to be configured in a way to that it hides this argument in the "ps" command but that's up to OS since we would just offer this option.

bundles/io/org.eclipse.smarthome.io.encryption/src/main/java/org/eclipse/smarthome/io/crypto/internal/CryptoActivator.java
```diff
@@ -0,0 +1,65 @@
...
+            while (true) {
+                System.out.println("Give secret key to continue: ");
+                BufferedReader br = new BufferedReader(new InputStreamReader(System.in));
+                String line = br.readLine();
+                if (!line.isEmpty()) {
+                    // System.out.println("secret key: " + line);
```
triller-telekom: Please remove this line

bundles/io/org.eclipse.smarthome.io.encryption/src/main/java/org/eclipse/smarthome/io/crypto/internal/CryptoActivator.java
```diff
@@ -0,0 +1,65 @@
...
+        } else {
+            encKey = key.getBytes();
+        }
+    }
+
+    /**
+     * Called whenever the OSGi framework stops our bundle
+     */
+    @Override
+    public void stop(BundleContext bc) throws Exception {
```
triller-telekom: Don't you want to call `clearKey()` here?

bundles/io/org.eclipse.smarthome.io.encryption/src/main/java/org/eclipse/smarthome/io/crypto/internal/CryptoModule.java
```diff
@@ -0,0 +1,137 @@
...
+ */
+public class CryptoModule implements Crypto {
+    public final static String ENV_VARIABLE_ENC_PASSPHRASE = "CFG_ENCRYPTION_PASSPHRASE";
+
+    private final Logger logger = LoggerFactory.getLogger(CryptoModule.class);
+
+    private byte[] salt = { (byte) 0xB8, (byte) 0x62, (byte) 0xF2, (byte) 0x29, (byte) 0x85, (byte) 0xE8, (byte) 0x11,
+            (byte) 0x96 };
+
+    private int iterationCount = 2000;
```
triller-telekom: Please make this a contant.

bundles/io/org.eclipse.smarthome.io.encryption/src/main/java/org/eclipse/smarthome/io/crypto/internal/CryptoModule.java
```diff
@@ -0,0 +1,137 @@
...
+ *
+ * @author Pauli Anttila - Initial contribution
+ *
+ */
+public class CryptoModule implements Crypto {
+    public final static String ENV_VARIABLE_ENC_PASSPHRASE = "CFG_ENCRYPTION_PASSPHRASE";
+
+    private final Logger logger = LoggerFactory.getLogger(CryptoModule.class);
+
+    private byte[] salt = { (byte) 0xB8, (byte) 0x62, (byte) 0xF2, (byte) 0x29, (byte) 0x85, (byte) 0xE8, (byte) 0x11,
```
triller-telekom: Please make this a contant.

bundles/io/org.eclipse.smarthome.io.encryption/src/main/java/org/eclipse/smarthome/io/crypto/internal/CryptoModule.java
```diff
@@ -0,0 +1,137 @@
...
+            throw new IllegalStateException("Secret key not initialized");
+        }
+    }
+
+    @Override
+    public void setEncPassphrase(byte[] secret) {
+        try {
+            String text = new String(secret, "UTF-8");
+            KeySpec keySpec = new PBEKeySpec(text.toCharArray(), salt, iterationCount);
+            secretKey = SecretKeyFactory.getInstance("PBEWithMD5AndDES").generateSecret(keySpec);
```
triller-telekom: We definately do not want to use MD5 for security related hashes, nor DES to encrypt something. How about using PBEWithHmacSHA256AndAES_256 instead?

If PBEWithHmacSHA256AndAES_256 is not available and also with regards to US import/export restrictions we might want to have this configurable in a config file anyway so people with the JCE installed can use better algorithms.

bundles/io/org.eclipse.smarthome.io.encryption/src/main/java/org/eclipse/smarthome/io/crypto/internal/CryptoModule.java
```diff
@@ -0,0 +1,137 @@
...
+    public final static String ENV_VARIABLE_ENC_PASSPHRASE = "CFG_ENCRYPTION_PASSPHRASE";
+
+    private final Logger logger = LoggerFactory.getLogger(CryptoModule.class);
+
+    private byte[] salt = { (byte) 0xB8, (byte) 0x62, (byte) 0xF2, (byte) 0x29, (byte) 0x85, (byte) 0xE8, (byte) 0x11,
+            (byte) 0x96 };
+
+    private int iterationCount = 2000;
+    private Cipher dcipher;
+    private Cipher ecipher;
```
triller-telekom: Why do you have 2 separate ciphers for encryption and decryption?

bundles/io/org.eclipse.smarthome.io.encryption/src/main/java/org/eclipse/smarthome/io/crypto/internal/CryptoModule.java
davidgraeff: I recon the salt should not be a constant (visible for everyone that can read source codes, including the bad guys), but be an initially computed value (hashing the mac address, serial no of the cpu or anything). That way the salt is at least different for every user and not public available.

bundles/io/org.eclipse.smarthome.io.encryption/src/main/java/org/eclipse/smarthome/io/crypto/internal/CryptoModule.java
davidgraeff: The java documentation explicitly mentions necessary algorithms to be shipped with every jre, so you should select the safest one as a fallback, if the configured one cannot be found. You might as well first try a stronger one, and catch the exception, before trying the jre guaranteed one.

bundles/io/org.eclipse.smarthome.io.encryption/src/main/java/org/eclipse/smarthome/io/crypto/Crypto.java
```diff
@@ -0,0 +1,29 @@
...
+ * Crypto module API.
+ *
+ * @author Pauli Anttila - Initial contribution
+ *
+ */
+public interface Crypto {
+
+    public String encrypt(String plainText) throws InvalidKeyException, IllegalBlockSizeException, BadPaddingException;
+
+    public String decrypt(String encryptedText)
```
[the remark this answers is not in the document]
davidgraeff: I had those two methods in my suggestion for a SecretsFault as well. But what if the user changes the master password? All the old encrypted secrets cannot be decrypted anymore.

bundles/io/org.eclipse.smarthome.io.encryption/src/main/java/org/eclipse/smarthome/io/crypto/Crypto.java
```diff
@@ -0,0 +1,29 @@
...
+
+/**
+ * Crypto module API.
+ *
+ * @author Pauli Anttila - Initial contribution
+ *
+ */
+public interface Crypto {
+
+    public String encrypt(String plainText) throws InvalidKeyException, IllegalBlockSizeException, BadPaddingException;
```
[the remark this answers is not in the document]
davidgraeff: Replace all the exceptions by GeneralSecurityException

bundles/io/org.eclipse.smarthome.io.console/src/main/java/org/eclipse/smarthome/io/console/internal/extension/CryptoConsoleCommandExtension.java
```diff
@@ -0,0 +1,87 @@
...
+ * Console command extension for crypto module functions.
+ *
+ * @author Pauli Anttila - Initial contribution
+ *
+ */
+public class CryptoConsoleCommandExtension extends AbstractConsoleCommandExtension {
+
+    private Crypto crypto;
+
+    private static final String SUBCMD_ENCRYPT = "encrypt";
```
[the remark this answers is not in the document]
davidgraeff: I would say, per default, all "password" context fields are encrypted

bundles/config/org.eclipse.smarthome.config.dispatch/src/main/java/org/eclipse/smarthome/config/dispatch/internal/ConfigDispatcher.java
```diff
@@ -252,6 +262,22 @@ private void processConfigFile(File configFile) throws IOException, FileNotFound
 
             String property = contents[1];
             String value = contents[2];
+            logger.debug("'{}'='{}'", property, value);
+
+            if (value != null) {
+                if (value.startsWith("ENC(") && value.endsWith(")")) {
```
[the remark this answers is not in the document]
davidgraeff: Maybe we should extend the idea of configuration value modifiers. I need base64 for example for config values. Do we have OSGi capabilities already? A service (like this) could provide the capability to decrypt ENC wrapped text, another service provide the capability to decode base64 wrapped text.

davidgraeff:
> how do we ensure that this bundle is started before all other bundles which read configuration files?

This service describes itself as dependency for ConfigDispatcher, therefore it will be started before reading configurations, I thing?

> what about an additional option to pass password as command line argument?

This is not secure unfortunately. You can see the full command line of all running applications without having administrative access rights.

> with regards to US import/export restrictions we might want to have this configurable in a config file anyway so people with the JCE installed can use better algorithms.
cipher algorithms should be store in a config file for easier change

I agree, but the default, before any config file exists, should be the safest way possible, by testing what cypher is available.

davidgraeff:
To increase the speed of review process so that I can continue with my SSL service, can we discuss the API of a SecretsVault/Crypto service first and submit that as a first PR?

bundles/io/org.eclipse.smarthome.io.encryption/src/main/java/org/eclipse/smarthome/io/crypto/internal/CryptoModule.java
triller-telekom: A salt is not a secret, it just protects the user against pre-computed rainbow tables for the used hash algorithm. But I agree that having an individual salt per installation is even better than a global one for all ESH installations.

paulianttila:
Thank you both @triller-telekom and @davidgraeff about the review. In fact, my intention was not that you use your time for code review but start discussion. Code itself definitely is not in that shape that it could be merged. I tried to indicate that by proof of concept word, but maybe I picked a wrong word or at least I have should mentation that better in the PR text.

paulianttila:
>> how do we ensure that this bundle is started before all other bundles which read configuration files?

> This service describes itself as dependency for ConfigDispatcher, therefore it will be started before reading configurations, I thing?

Yes, that's right.

>> what about an additional option to pass password as command line argument?

> This is not secure unfortunately. You can see the full command line of all running applications without having administrative access rights.

I agree as well. Environment variables has been seen pretty save in security point of view. I actually liked most the idea, that master password is given during startup, then password is not stored in the file system at all. I tried also implement simple http servlet, where user could give the master password, but that broach didn't work because on bundle dependencies. Yubikey support would be awesome.

paulianttila:
> Entering the password at startup, the more secure it is, also the more inconvenient it is. Imagine your SmartHome installation suffering a power outage while you are on vacation. The system might restart automatically but then hang while waiting for the password to be entered. So I guess the environment variable would be a good compromise.

I fully agree, even my own environment is backed by the ups and I could give password remotely pretty much anywhere. Environment variable for sure should be the default broach, but maybe we could also provide the more secure way. Yubikey support would be pretty elegant solution, as it's cheep and works like a HSM, but there is no native java support, so it's need e.g. OpenSC to provide PKCS#11 API.

davidgraeff:
> Yubikey support would be pretty elegant solution, as it's cheep and works like a HSM, but there is no native java support, so it's need e.g. OpenSC to provide PKCS#11 API.

You can write a wrapper around the start up script, that extract the secret from your Yubikey and pass it to the environment for ESH. A master password should do it for now.

> There is a chicken egg problem as configDispatcher need cryptoModule, therefore cryptoModule can't use generic configuration files.

I haven't thought about that. There is only one cipher used anyway for encryption/decryption. That could be chosen by selecting the safest option available and printing a warning to the user if the safest option is not really safe actually. The selection is stored in the OSGi cache. No configuration file should be necessary here. I doubt it anyway, that if we choose a reasonable cipher, that any user will ever change that default (except for the worse).

davidgraeff:
> Unfortunately I don't know if we can use OSGi capabilities as @davidgraeff mentioned in his comment , because they were introduced in OSGi 4.3, see here and AFAIK ESH is based on OSGi 4.2, am I right @kaikreuzer ?

Framework | Version | OSGi spec support | Release date
-- | -- | -- | --
Apache Felix | 5.2.0 | R6 | September 25, 2015
Eclipse Equinox | 4.7 | R6 (parts R5) | June 28, 2017



According to this page https://projects.eclipse.org/projects/rt.equinox/releases/4.7.0-oxygen.
This is confusing, are we not using a recent (2 years old) version of the OSGi specification?

kaikreuzer:
> AFAIK ESH is based on OSGi 4.2, am I right @kaikreuzer ?

As [documented here as C.2](https://www.eclipse.org/smarthome/documentation/development/guidelines.html#c-language-levels-and-libraries)

triller-telekom:
My concerns regarding the security are maybe a bit too comprehensive. In the end ESH is a framework and not a solution ready to be used.

So, how about defining the whole "encryption module" only on the basis of interfaces rather than a full fledged implementation? It should be possible to replace all security relevant parameters such as keys, path to key stores, hash algorithms, encryption algorithms, passwords etc. with a different implementation. This way we would leave out the whole security discussion from ESH, but still provide the possibility to implement security in a solution based on ESH.

One such base implementation (bundle) that we ship with ESH could use the default algorithms of Java and store the passwords in a config file (readable only by the user that starts ESH) or in an environment variable. A different solution could replace our bundle and use more secure ciphers and maybe also a Yubikey.

BTW:

> The algorithm could be chosen by selecting the safest option available and printing a warning to the user if the safest option is not really safe actually. The selection is stored in the OSGi cache. No configuration file should be necessary here. I doubt it anyway, that if we choose a reasonable algorithm, that any user will ever want to change that default (except for the worse).

This is not true, because ESH can be used in commercial solutions which might suffer from compliance rules that enforce them to use certain ciphers, key length etc. So a configuration should definitely be possible. 

WDYT?

bundles/io/org.eclipse.smarthome.io.encryption/pom.xml
```diff
@@ -0,0 +1,22 @@
...
+  <parent>
+    <groupId>org.eclipse.smarthome.bundles</groupId>
+    <artifactId>io</artifactId>
+    <version>0.9.0-SNAPSHOT</version>
+  </parent>
+
+  <properties>
+    <bundle.symbolicName>org.eclipse.smarthome.io.crypto</bundle.symbolicName>
+    <bundle.namespace>org.eclipse.smarthome.io.crypto</bundle.namespace>
+  </properties>
```
wborn: Please remove these properties. See #6060.

bundles/io/org.eclipse.smarthome.io.encryption/.settings/org.eclipse.jdt.core.prefs
```diff
@@ -0,0 +1,8 @@
+eclipse.preferences.version=1
+org.eclipse.jdt.core.compiler.codegen.inlineJsrBytecode=enabled
+org.eclipse.jdt.core.compiler.codegen.targetPlatform=1.8
+org.eclipse.jdt.core.compiler.compliance=1.8
+org.eclipse.jdt.core.compiler.problem.assertIdentifier=error
+org.eclipse.jdt.core.compiler.problem.enumIdentifier=error
+org.eclipse.jdt.core.compiler.problem.forbiddenReference=warning
+org.eclipse.jdt.core.compiler.source=1.8
```
wborn: Please remove this file because it causes the default Eclipse project settings to be overridden. See also #6106.

bundles/io/org.eclipse.smarthome.io.encryption/.settings/org.maven.ide.eclipse.prefs
```diff
@@ -0,0 +1,9 @@
+#Fri Feb 19 22:30:18 CET 2010
+activeProfiles=
+eclipse.preferences.version=1
+fullBuildGoals=process-test-resources
+includeModules=false
+resolveWorkspaceProjects=true
+resourceFilterGoals=process-resources resources\:testResources
+skipCompilerPlugin=true
+version=1
```
wborn: Please remove this file because it causes the default Eclipse project settings to be overridden. See also #6106.

davidgraeff:
Can be closed from my side. The API need to be discussed first.

PR #3839, opened 2017-07-15
Mqtt Broker configuration binding

davidgraeff:
## Abstract
At the moment it is possible to configure brokers in a textual way. Without intensively watching the log file, there is no way to tell if the provided configuration is correct and if the connection to the broker have been established, neither via the GUI (PaperUI) nor machine readable (REST interface).

This binding makes the configuration of brokers more intuitive and let the user add broker connections as things. The connections are synced to the `io.transport.mqtt.MqttService` and are therefore available for all other bindings to use.

Features include: 
* **2 different Bridges**: SystemBrokerConnection, BrokerConnection 
* **Rich status feedback** for system broker connections. No longer skimming through log files if a connection doesn't work. 
* **Auto discovery of brokers**: Enumeration of MqttService shared broker connections, best effort subnet and localhost scan on standard Mqtt ports
* **Configurable reconnect time**: This is not possible via textual configuration yet
* **Security** Certificate and public key pinning for ssl connections to brokers 
* **Extensive tests** for all handlers, the discovery services and the SSLContextProvider.

## Discussion

Points that need to be discussed might be:
* **No channels** There are no channels on the broker Things. Everything is published to Thing properties right now. The connection status could be made available via a "status channel" additionally.
* **Security** The public key and certificate hashes for key and certificate pinning are stored in the thing configuration. I'm not sure if this is the safest places to store them as the configuration can be changed via the rest interface. It is at least the only way to let the user configure them easily at the moment.

I'm happy to hear some feedback.

Cheers,
David

PR #4172, opened 2017-09-01
Core/net: Add getInterfaceAddresses.

davidgraeff:
This completes getAllBroadcastAddresses() and is useful for discovery services.
A usecase is for example to compute the list of all assignable addresses for an interface to scan for a service/device.

This is used by the OH network binding and the Mqtt broker binding. See https://github.com/eclipse/smarthome/pull/3839#discussion_r135989687

The actual list of addresses is computed via:
```
NetUtil.getInterfaceAddresses(true).stream().map(interfaceIP -> Arrays
                .asList(new SubnetUtils(interfaceIP).getInfo().getAllAddresses()))
                .flatMap(List::stream).collect(Collectors.toList());
```

### DIFF
--- a/bundles/core/org.eclipse.smarthome.core/src/main/java/org/eclipse/smarthome/core/net/NetUtil.java
+++ b/bundles/core/org.eclipse.smarthome.core/src/main/java/org/eclipse/smarthome/core/net/NetUtil.java
@@ -14,9 +14,12 @@ import java.net.InterfaceAddress;
 import java.net.NetworkInterface;
 import java.net.SocketException;
 import java.util.Enumeration;
+import java.util.HashSet;
+import java.util.Iterator;
 import java.util.LinkedList;
 import java.util.List;
 import java.util.Map;
+import java.util.Set;
 import java.util.regex.Pattern;
 
 import org.eclipse.jdt.annotation.NonNull;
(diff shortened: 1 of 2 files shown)


docs/documentation/features/frameworkUtilities.md
```diff
@@ -11,4 +11,7 @@ In this chapter useful services/utilities of the Eclipse SmartHome project are d
 The `NetworkAddressService` is an OSGi service that can be used like any other OSGi service by adding a service reference to it. Its OSGi service name is `org.eclipse.smarthome.network`.
 A user can configure his default network address via Paper UI under `Configuration -> System -> Network Settings`.
 One can obtain the configured address via the `getPrimaryIpv4HostAddress()` method on the service.
-This service is useful for example in the `ThingHandlerFactory` or an `AudioSink` where one needs a specific IP address of the host system to provide something like a `callback` URL.
\ No newline at end of file
+This service is useful for example in the `ThingHandlerFactory` or an `AudioSink` where one needs a specific IP address of the host system to provide something like a `callback` URL.
+
+Some static methods like `getAllBroadcastAddresses()` for retrieving all interface broadcast addresses or `getInterfaceAddresses()` for retrieving all assigned interface addresses might be
+usefull as well for discovery services.
```
kaikreuzer: I assume this line break was by accident?

bundles/core/org.eclipse.smarthome.core/src/main/java/org/eclipse/smarthome/core/net/NetUtil.java
```diff
@@ -187,6 +190,50 @@ public static String getBroadcastAddress() {
...
+        try {
+            en = NetworkInterface.getNetworkInterfaces();
+        } catch (SocketException ex) {
+            LOGGER.error("Could not find interface IP addresses: {}", ex.getMessage(), ex);
+            return interfaceIPs;
+        }
+
+        while (en.hasMoreElements()) {
+            NetworkInterface networkInterface = en.nextElement();
+            boolean isLoopback = true;
```
maggu2810: It does not care, I just wonder if you need "isLoopback" on your initial implementation...

```Java
boolean isLoopback = true;
try {
    isLoopback = networkInterface.isLoopback();
} catch (SocketException ignored) {
}
if (isLoopback) {
    continue;
}
```

Why not shrink it to this one:

```Java
try {
    if (networkInterface.isLoopback()) {
        continue;
    }
} catch (SocketException ignored) {
}
```

davidgraeff: `NetworkConfigOptionProvider` in core.net.internal does it like this:

```
try {
            final Enumeration<NetworkInterface> interfaces = NetworkInterface.getNetworkInterfaces();
            while (interfaces.hasMoreElements()) {
                final NetworkInterface current = interfaces.nextElement();
                if (!current.isUp() || current.isLoopback() || current.isVirtual()) {
                    continue;
                }

                for (InterfaceAddress ifAddr : current.getInterfaceAddresses()) {
                    InetAddress addr = ifAddr.getAddress();

                    if (addr.isLoopbackAddress() || (addr instanceof Inet6Address)) {
                        continue;
                    }

                    @SuppressWarnings("null")
                    @NonNull
                    String ipv4Address = addr.getHostAddress();
                    try {
                        String subNetString = NetUtil.getIpv4NetAddress(ipv4Address, ifAddr.getNetworkPrefixLength())
                                + "/" + String.valueOf(ifAddr.getNetworkPrefixLength());
                        subnets.add(subNetString);
                    } catch (IllegalArgumentException ex) {
                        logger.error("Could not calculate network address: {} Ignoring IP {}", ex.getMessage(),
                                ipv4Address, ex);
                    }
                }
            }
            logger.error("Could not retrieve network interface: {}", ex.getMessage(), ex);
            return null;
        }
```

They even check for "up" and "not virtual"

maggu2810: You are right. Is there is one SocketException the whole loop is stopped. Locally exception handling would be better.

PR #4173, opened 2017-09-01
MqttService: Remove all deprecated methods, rename close to stop.

### LINKED ISSUE #4170: MqttService API break
As we don't have any consumers of the MQTT API right now (they would have been affected by the reconnect bug that just recently got fixed), we should discuss some little API breaks that I strongly recommend:

In MqttBrokerConnection:
* Rename close() to stop(). The semantic of close is usually, that the object cannot be used anymore. Stop would be matching with start(). Alternatively call start() open() instead.
* Remove all deprecated methods.

Cheers, David

### LINKED ISSUE #5162: Unable to use MQTT Last Will and Testament with dynamic data
I am using `ESH version 0.9.0 io.transport.mqtt` bundle and trying to implement a custom plugin for MQTT. Having a look into the code I saw that it is pretty straightforward and well structured, thanks for this!

One concern of mine is the MQTT Last Will and Testament functionality. Creating an instance of the given class `MqttWillAndTestament`  and passing it to the `.setLastWill()` method of the `MqttBrokerConnection` didn't work at all. 

Digging into the code of this `MqttBrokerConnection` class I see that this field is used for building the `MqttConnectOptions` object in the  `createMqttOptions()`. This object is passed to the `.connect()` method of `org.eclipse.paho.client.mqttv3.MqttAsyncClient` later in the `MqttBrokerConnection.start()`.

My headaches start here: When I configure the MqttService for a given brokername and later use its `.getBrokerConnection()` method passing the corresponding broker name the mentioned things has already happened. Although I get a proper connection and I am able to publish or subscribe using this connection I have no LWT. The problem comes when I have to set this Last Will which by the time of creation of this connection is null. With apologies for my ignorance I ask you for advice/ help. 

@davidgraeff Since you are deeply involved in the io.transport.mqtt bundle's implementation I would be very glad if you share your opinion on this one.

davidgraeff:
This PR prepares the io.transport.mqtt bundle for the discussed MQTT architecture.

* Remove `name` from MqttBrokerConnection. This is only necessary for the MqttService.
  Consumers of MqttBrokerConnection should not be bothered with this detail.
* The MqttService requires a brokerID for addConnection() and removeConnection() and the observer interface adds the brokerID next to the connection as well, to compensate for the removed `name`.
* Remove `textualConfigured` flag. Again this is not interesting for anything outside MqttService.
* Don't parse configuration parameters ourselves, but use a configuration values class instead.
* Use Nullable where it makes sense.
* The code was modernized
* ESH compatible implementation of the service factory pattern including a config description for GUIs
* Add MqttService.setLastWill method. Fixes #5162

API break:
* All deprecated methods are removed.
* MqttBrokerConnection.close -> stop
* MqttBrokerConnection.getURL() -> getHost(), getPort(), isSecure().
* MqttBrokerConnection constructor changed

Config interface break:
* There is no "url" anymore, it is now "host"+"port"+"secure"
* "retain" is now "retainMessages"
* "user" is now "username"
* "pass" is now "password"
* "lw" is now "lwtTopic"+"lwtMessage"+"lwtQos"+"lwtRetain"
* "clientid" is now "clientID"
* Uses OSGI service factories. 

Documentation
* Add more public class documentation especially for `MqttService` and its purpose.

Fixes #4170


kaikreuzer:
Oops, merge conflicts...

davidgraeff:
I would agree. The changes would be:

* Make host, port, secure, clientID final
* The clientID must be set in the constructor. Null is allowed, an ID will be generated in that case.

Everything else (qos, reconnect time etc) aren't connection parameters but take effect immediately for all following published messages.

davidgraeff:
I changed the PR accordingly. I'm just not sure about username and password. Those are technically speaking not connection parameters and can change between stop/start cycles and are optional.

bundles/io/org.eclipse.smarthome.io.transport.mqtt/src/main/java/org/eclipse/smarthome/io/transport/mqtt/MqttService.java
```diff
@@ -217,212 +209,57 @@ public boolean addBrokerConnection(MqttBrokerConnection connection) {
...
      */
-    public MqttBrokerConnection addBrokerConnection(Map<String, String> brokerConnectionConfig)
-            throws ConfigurationException, MqttException {
-        // Extract mandatory fields
-        String brokerID = brokerConnectionConfig.get(NAME_PROPERTY);
-        if (StringUtils.isBlank(brokerID)) {
-            throw new ConfigurationException("MQTT Broker property 'name' is not provided");
-        }
-        brokerID = brokerID.toLowerCase();
+    public synchronized MqttBrokerConnection addBrokerConnection(@NonNull String brokerID,
```
maggu2810: @kaikreuzer Do I remember correctly that you allowed `@Nullable` on return types to enable checks for the consumer code?
If it is allowed I would prefer to add it here.

bundles/io/org.eclipse.smarthome.io.transport.mqtt/src/main/java/org/eclipse/smarthome/io/transport/mqtt/MqttService.java
```diff
@@ -182,33 +178,29 @@ public boolean hasBrokerObservers() {
      * @param brokerName to look for.
      * @return existing connection or null
      */
-    public MqttBrokerConnection getBrokerConnection(String brokerName) {
-        synchronized (brokerConnections) {
-            return brokerConnections.get(brokerName.toLowerCase());
-        }
+    public synchronized MqttBrokerConnection getBrokerConnection(String brokerName) {
```
maggu2810: See other comment about `Nullable` on return type.

bundles/io/org.eclipse.smarthome.io.transport.mqtt/src/main/java/org/eclipse/smarthome/io/transport/mqtt/MqttService.java
davidgraeff: Hm, NonNull is the java standard. I agree that it sometimes makes sense to add it, as a way to document parameters, for instance Integer: A consumer might feel like he always needs to provide a value, even if `null` would be absolutely acceptable.

But for return types I would say Nullable is not reasonable.

maggu2810: Please read https://github.com/eclipse/smarthome/pull/4080

davidgraeff: Nah a twenty page long discussion doesn't help :) A link or quote of the developer guidelines would convince me of course :D

maggu2810: There isn't a developer guidelines about the usage of nullness annotation. At least we didn't fully agree.

davidgraeff: At least everything that has been agreed on, should be put in the developer guidelines. Should I open an issue?

maggu2810: I don't think so. I have added a summary here:
https://github.com/eclipse/smarthome/wiki/Type-annotations

kaikreuzer:
@maggu2810 Was this PR now ok for you?
If so, what might be left is to adapt the use of null annotations to our new guideline (i.e. using `@NonNullByDefault` on the class level.

bundles/io/org.eclipse.smarthome.io.transport.mqtt/src/main/java/org/eclipse/smarthome/io/transport/mqtt/MqttBrokerConnection.java
```diff
@@ -165,31 +154,31 @@ public AbstractReconnectStrategy getReconnectStrategy() {
...
-     * tcp://localhost:1883 or ssl://localhost:8883
-     *
-     * @return url for the MQTT broker.
+     * Get the MQTT broker port
      */
-    public String getUrl() {
-        return url;
+    @Nullable
+    public Integer getPort() {
+        return port;
```
maggu2810: On the constructor you state.
```text
* @param port A port or null to select the default port for a secure or insecure connection
```

So I would assume that a port is choosen now and we could return the explicit set one, the insecure or secure default.
I am pretty sure that a connection uses a port ;)

maggu2810:
> @maggu2810 Was this PR now ok for you?

If the MQTT transport and reconnect is still working (using the non-deprecated methods), all is okay for me. :wink: 
Didn't test it myself.

maggu2810:
Parts of the nullness fixes could be taken from here:
https://github.com/eclipse/smarthome/pull/4290/commits/ae1a8ac4de6180e7c3bc46d06b7d1a05875d7507

davidgraeff:
* I adapted the PR including the Nullness fixes.
* I removed the custom nested configuration parser, as we agreed in https://github.com/eclipse/smarthome/issues/4178, that a central place to parse nested configuration is a good idea, and I expect the "Configuration" class to handle the details for me.

At the moment that means, this service will not be configurable without changes to the "Configuration" class.

davidgraeff:
Because all my other Mqtt PRs are basing on the API changes, introduced here, I will resurrect this PR and adapt the service multi-config concept.

htreu:
@davidgraeff ping me when ready for review.

davidgraeff:
@htreu Code-wise it is complete. But travis failed because of another issue, therefore I do not know if if will pass the static code check yet.

Could you have a look at the MqttBrokerConnectionServiceInstance file, if I'm using the service-factory in a right way?
 https://github.com/eclipse/smarthome/pull/4173/files#diff-9365221333b769c30777cfc2914e2001R39

htreu:
Travis seems to be unrelated.

For the MqttBrokerConnectionServiceInstance please have a look at the `MagicMultiInstanceService` and the `MagicMultiInstanceServiceMarker` classes.
* `MagicMultiInstanceService` is the implementation of the actual service which can be configured multiple times.
* `MagicMultiInstanceServiceMarker` is a marker class which mainly carries the relevant information about the service to Paper UI.

The `ConfigurableService.SERVICE_PROPERTY_DESCRIPTION_URI + "=test:multipleMagic"` refers to the ESH-INF/config/multipleConfig.xml file where the configuration for the service is described.

Once you have everything setup correctly Paper UI should allow you to add/edit/remove instances of your service.

davidgraeff:
@htreu If have adjusted the PR. Could you have a look?

bundles/io/org.eclipse.smarthome.io.transport.mqtt/src/main/java/org/eclipse/smarthome/io/transport/mqtt/MqttService.java
```diff
@@ -1,244 +1,58 @@
...
-        public final @Nullable String pwd;
-        public final @Nullable String clientId;
-        public final @Nullable Integer keepAlive;
-        public final @Nullable Integer qos;
-        public final @Nullable Boolean retain;
-        public final @Nullable String lwt;
-        public final @Nullable String lwtTopic;
-        public final byte @Nullable [] lwtMessage;
-        public final @Nullable Integer lwtQos;
-        public final @Nullable Boolean lwtRetain;
```
htreu: Please review the latest changes @maggu2810 did on the LWT configuration. It is now possible to either pass the encoded String with `topic:message:qos:retained` or provide all values in separate fields (see https://github.com/eclipse/smarthome/pull/5231).

bundles/io/org.eclipse.smarthome.io.transport.mqtt/ESH-INF/config/brokerConnectionInstance.xml
```diff
@@ -0,0 +1,73 @@
...
+			<description>A connection name helps to identify a broker connection. If nothing is set, an automatic name based on the host and port will be generated.</description>
+			<default></default>
+		</parameter>
+		
+        <parameter name="host" type="text" required="true">
+            <label>Broker host</label>
+            <description>Hostname or IP of the broker</description>
+            <default></default>
+        </parameter>
+        <parameter name="host" type="integer" required="false">
```
htreu: name should be `host`

bundles/io/org.eclipse.smarthome.io.transport.mqtt/src/main/java/org/eclipse/smarthome/io/transport/mqtt/MqttBrokerConnectionConfig.java
```diff
@@ -0,0 +1,86 @@
...
+    // MQTT parameters
+    public BigDecimal qos = BigDecimal.valueOf(MqttBrokerConnection.DEFAULT_QOS);
+    public Boolean retainMessages = false;
+    public @Nullable String lastWill;
+    /** Keepalive in seconds */
+    public @Nullable BigDecimal keepAlive;
+
+    /**
+     * Return the brokerID of this connection. This is either the name or host:port(:s).
+     * This method might return null, if none of the parameters is set.
```
htreu: The method does actually _not_ return null but the empty String is nothing is configured.

bundles/io/org.eclipse.smarthome.io.transport.mqtt/src/main/java/org/eclipse/smarthome/io/transport/mqtt/MqttService.java
```diff
@@ -1,244 +1,58 @@
...
- *
- * See the NOTICE file(s) distributed with this work for additional
- * information regarding copyright ownership.
- *
- * This program and the accompanying materials are made available under the
- * terms of the Eclipse Public License 2.0 which is available at
- * http://www.eclipse.org/legal/epl-2.0
- *
- * SPDX-License-Identifier: EPL-2.0
+ * Copyright (c) 2014-2017 by the respective copyright holders.
```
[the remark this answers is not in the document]
davidgraeff: Unfortunately I do have to correct license headers myself. `mvn license:format` only corrects xml headers for me after 10min of downloading various files (it is time for gradle!)

bundles/io/org.eclipse.smarthome.io.transport.mqtt/src/main/java/org/eclipse/smarthome/io/transport/mqtt/internal/MqttBrokerConnectionServiceInstance.java
```diff
@@ -0,0 +1,118 @@
...
+        MqttBrokerConnectionConfig config = new Configuration(configMap).as(MqttBrokerConnectionConfig.class);
+
+        try {
+            // Compute brokerID and make sure it is not empty
+            String brokerID = config.getBrokerID();
+            if (StringUtils.isBlank(brokerID)) {
+                logger.warn("Ignore invalid broker connection configuration: {}", config);
+                return;
+            }
+            assert brokerID != null;
```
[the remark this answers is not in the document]
davidgraeff: Correct, but at least my eclipse will issue a warning. I'll investigate.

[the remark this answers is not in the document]
davidgraeff: The static analyser, used by eclipse, is not recognizing the result of `StringUtils.isBlank`. The assert is unfortunately necessary here.

bundles/io/org.eclipse.smarthome.io.transport.mqtt/src/main/java/org/eclipse/smarthome/io/transport/mqtt/internal/MqttBrokerConnectionServiceInstance.java
```diff
@@ -0,0 +1,118 @@
...
+
+            // Add connection and make sure it succeeded
+            assert mqttService != null;
```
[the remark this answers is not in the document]
davidgraeff: True, but eclipse is issuing a warning. I guess that's because `mqttService` is a field variable and the static analyser doesn't guarantee that no concurrent thread or side effect of a method call changes the field. I would leave it for now.

davidgraeff:
@htreu Issues addressed

bundles/io/org.eclipse.smarthome.io.transport.mqtt/ESH-INF/config/brokerConnectionInstance.xml
```diff
@@ -0,0 +1,115 @@
...
+            <label>Message parameters</label>
+            <advanced>true</advanced>
+        </parameter-group>
+        <parameter-group name="group_lastwill_params">
+            <label>Last will</label>
+            <description>All last-will parameters</description>
+            <advanced>true</advanced>
+        </parameter-group>
+
+		<parameter name="name" type="text" required="false" groupName="group_connection">
```
htreu: please run auto formatter once.

bundles/io/org.eclipse.smarthome.io.transport.mqtt/ESH-INF/config/brokerConnectionInstance.xml
```diff
@@ -0,0 +1,115 @@
...
+            </options>
+            <default>0</default>
+        </parameter>
+        <parameter name="retainMessages" type="boolean" required="false" groupName="group_message_params">
+            <label>Retain messages</label>
+            <description>Messages send by this connection are retained</description>
+            <default>false</default>
+        </parameter>
+
+
```
htreu: empty line

bundles/io/org.eclipse.smarthome.io.transport.mqtt/src/main/java/org/eclipse/smarthome/io/transport/mqtt/MqttBrokerConnection.java
```diff
@@ -86,44 +84,44 @@
...
     private class ClientCallbacks implements MqttCallback {
         @Override
-        public synchronized void connectionLost(Throwable t) {
-            if (t instanceof MqttException) {
-                MqttException e = (MqttException) t;
-                logger.info("MQTT connection to '{}' was lost: {} : ReasonCode {} : Cause : {}", getName(),
-                        e.getMessage(), e.getReasonCode(),
-                        (e.getCause() == null ? "Unknown" : e.getCause().getMessage()));
+        public synchronized void connectionLost(@Nullable Throwable exception) {
+            assert exception != null;
```
htreu: `exception` being null seems to be a legit case. Its way too hard to throw errors here.
We do not use `assert` in ESH at all: Assertions need to be switched on during compilation what we don't do. Thus they will be removed from the resulting class files. And in this particular case it is hiding a NPE when `exception` actually _is_ null.
Please review other occurrences and replace them by throwing IllegalArgumentException or IllegalStateException.

bundles/io/org.eclipse.smarthome.io.transport.mqtt/src/main/java/org/eclipse/smarthome/io/transport/mqtt/MqttBrokerConnection.java
```diff
@@ -133,24 +131,28 @@ public void messageArrived(String topic, MqttMessage message) {
...
+     *            If none is specified, a default is generated. The client id cannot
+     *            be longer than 23 characters.
+     * @throws IllegalArgumentException
+     */
+    public MqttBrokerConnection(String host, @Nullable Integer port, boolean secure, @Nullable String clientId) {
+        this.host = host;
+        this.secure = secure;
+        if (clientId == null) {
+            clientId = MqttClient.generateClientId();
+            assert clientId != null;
```
htreu: please remove or refactor

bundles/io/org.eclipse.smarthome.io.transport.mqtt/src/main/java/org/eclipse/smarthome/io/transport/mqtt/internal/MqttBrokerConnectionServiceInstance.java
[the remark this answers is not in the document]
htreu: Please remove it and ignore the warning. Or add an extra null check to the StringUtils check.

bundles/io/org.eclipse.smarthome.io.transport.mqtt/src/main/java/org/eclipse/smarthome/io/transport/mqtt/internal/MqttBrokerConnectionServiceInstance.java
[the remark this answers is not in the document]
htreu: Please remove and ignore/refactor

bundles/io/org.eclipse.smarthome.io.transport.mqtt/src/main/java/org/eclipse/smarthome/io/transport/mqtt/internal/MqttBrokerConnectionServiceInstance.java
```diff
@@ -0,0 +1,118 @@
...
+            connection = mqttService.addBrokerConnection(brokerID, config);
+            if (connection == null) {
+                logger.warn("Ignore existing broker connection configuration for: {}", brokerID);
+                return;
+            }
+            // Start connection
+            assert connection != null;
```
htreu: Please remove and ignore/refactor

bundles/io/org.eclipse.smarthome.io.transport.mqtt/src/main/java/org/eclipse/smarthome/io/transport/mqtt/MqttBrokerConnection.java
davidgraeff: I checked with the documentation. All parameters on this interface are meant to be non null. The annotations are just missing and assert is basically used to compensate. I have added some comments

bundles/io/org.eclipse.smarthome.io.transport.mqtt/src/main/java/org/eclipse/smarthome/io/transport/mqtt/MqttBrokerConnection.java
davidgraeff: Some asserts can not be removed with my eclipse, which was setup by oomph and the official installer. I get errors like "Null type mismatch (type annotations): required '@NonNull String' but this expression has type '@Nullable String'. The problem severity defaults to "error" for those null checks.

davidgraeff:
I have no idea, why you get "java.lang.IllegalArgumentException: Can not set java.lang.Integer field org.eclipse.smarthome.io.transport.mqtt.MqttBrokerConnectionConfig.lwtQos to java.lang.String" to be honest. The conf desc xml says `<parameter name="lwtQos" type="integer" ...>` and it is defined as `public Integer lwtQos` in **MqttBrokerConnectionConfig**.

bundles/io/org.eclipse.smarthome.io.transport.mqtt/src/main/java/org/eclipse/smarthome/io/transport/mqtt/MqttBrokerConnection.java
maggu2810: > Assertions need to be switched on during compilation what we don't do.

That's not correct.
assertions could be enabled at runtime to use additional checks (e.g. in the development phase) that are not enabled default

davidgraeff: Ah I see (http://etutorials.org/Programming/Java+performance+tuning/Chapter+6.+Exceptions+Assertions+Casts+and+Variables/6.2+Assertions/). So it actually imposes a runtime overhead in contrast to C/C++. I'll use local variables then to get rid of the warnings.

htreu:
I don't understand the exception either. Does it work for you?
Also: This time Travis really hangs at MQTT transport with 
```
[ERROR] Failed to execute goal org.eclipse.tycho:tycho-compiler-plugin:1.1.0:compile (default-compile) on project org.eclipse.smarthome.io.transport.mqtt: Compilation failure: Compilation failure: 
[ERROR] /home/travis/build/eclipse/smarthome/bundles/io/org.eclipse.smarthome.io.transport.mqtt/src/main/java/org/eclipse/smarthome/io/transport/mqtt/MqttService.java:[47] 
[ERROR] 	@Component(immediate = true, service = { MqttService.class }, configurationPid = {
[ERROR]         "org.eclipse.smarthome.mqtt" }, property = "service.pid=org.eclipse.smarthome.mqtt")
[ERROR] 	                                                                                 ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
[ERROR] Type mismatch: cannot convert from String[] to String
```

maggu2810:
See: https://github.com/eclipse/smarthome/pull/4625

davidgraeff:
The reason for the IllegalArgumentException is this:
Some time ago numbers in service configuration maps were dispatched as BigDecimal objects. Now they are Strings instead. We use config/core/Configuration for parsing the map and it fails on line 89ff, because it still expects BigDecimals:
```
if (value != null && value instanceof BigDecimal && !typeName.equals("BigDecimal")) {
       BigDecimal bdValue = (BigDecimal) value;
     if (typeName.equalsIgnoreCase("Float")) {
         value = bdValue.floatValue();
     } else if (typeName.equalsIgnoreCase("Double")) {
         value = bdValue.doubleValue();
     } else if (typeName.equalsIgnoreCase("Long")) {
         value = bdValue.longValue();
     } else if (typeName.equalsIgnoreCase("Integer") || typeName.equalsIgnoreCase("int")) {
         value = bdValue.intValue();
     }
}
```

Configuration need to be changed to also accept Strings

davidgraeff:
I have addressed the Configuration class shortcoming here: #5262

bundles/io/org.eclipse.smarthome.io.transport.mqtt/ESH-INF/config/brokerConnectionInstance.xml
```diff
@@ -0,0 +1,115 @@
...
+			<description>A broker connection is either a non-secure tcp connection or a tls secure connection</description>
+			<default>false</default>
+		</parameter>
+
+		<parameter name="username" type="text" required="false" groupName="group_connection">
+			<label>Broker username</label>
+			<description>Broker username</description>
+			<default />
+		</parameter>
+		<parameter name="password" type="text" required="false" groupName="group_connection">
```
htreu: Please give this parameter the context `password`.

bundles/io/org.eclipse.smarthome.io.transport.mqtt/src/main/java/org/eclipse/smarthome/io/transport/mqtt/MqttBrokerConnection.java
```diff
@@ -49,81 +51,76 @@
...
+                reconnectStrategy.lostConnection();
             }
-
-            reconnectStrategy.lostConnection();
         }
 
         @Override
-        public void deliveryComplete(IMqttDeliveryToken token) {
+        public void deliveryComplete(@Nullable IMqttDeliveryToken token_) {
+            IMqttDeliveryToken token = (@NonNull IMqttDeliveryToken) token_;
```
htreu: this looks like an accident waiting to happen :-) Please simply null-check here.

bundles/io/org.eclipse.smarthome.io.transport.mqtt/src/main/java/org/eclipse/smarthome/io/transport/mqtt/MqttBrokerConnection.java
```diff
@@ -49,81 +51,76 @@
...
 
         @Override
-        public void messageArrived(String topic, MqttMessage message) {
-            logger.trace("Received message on topic '{}' : {}", topic, new String(message.getPayload()));
-            for (Map.Entry<String, List<MqttMessageSubscriber>> entry : consumers.entrySet()) {
-                final String target = entry.getKey();
-                final List<MqttMessageSubscriber> consumerList = entry.getValue();
-
+        public void messageArrived(@Nullable String topic_, @Nullable MqttMessage message_) {
+            byte[] payload = ((@NonNull MqttMessage) message_).getPayload();
```
htreu: Okay, I get what you are doing here. Instead of defining the methods args `@Nullable` and do the cast here you may add `@NonNullByDefault({})` at `ClientCallbacks` class level. This will "switch off" null annotation checks in this class. It hurts the eyes a little less then casting @Nullabel to @NonNull.

bundles/io/org.eclipse.smarthome.io.transport.mqtt/src/main/java/org/eclipse/smarthome/io/transport/mqtt/MqttBrokerConnectionConfig.java
```diff
@@ -0,0 +1,90 @@
...
+    public @Nullable String username;
+    public @Nullable String password;
+    public @Nullable String clientID;
+    // MQTT parameters
+    public Integer qos = MqttBrokerConnection.DEFAULT_QOS;
+    public Boolean retainMessages = false;
+    /** Keepalive in seconds */
+    public @Nullable Integer keepAlive;
+    // Last will parameters
+    public @Nullable String lwtTopic;
```
htreu: Now the "old" configuration of lwt is not supported anymore. With https://github.com/eclipse/smarthome/pull/5231 @maggu2810 made sure both options could be used.
I thinks its fine that the UI (config description) only provides the single fields. But text based configuration may depend on the single line config. wdyt?

davidgraeff: Usually yes, it would be a good idea to preserve old fields. But: This PR is a API brake, code-wise as well as configuration-wise. Remember that we used the dot syntax before to express multiple mqtt instances and we use osgi service factories with a config file each now. Any user of this service (which I doubt there are any, because we provide no means to use Mqtt really) need to migrate anyway ^^.

davidgraeff:
@htreu  Done

bundles/io/org.eclipse.smarthome.io.transport.mqtt/src/main/java/org/eclipse/smarthome/io/transport/mqtt/MqttBrokerConnectionConfig.java
htreu: @maggu2810 fine for you?

maggu2810: It is okay for me to drop the old way to configure the LWT.

I am pretty sure there are guys the use this service. There has been one that inspired me to implement that change (to support a more generic topic and payload) and I am using that service, too. Already two of them. :wink: But as @davidgraeff also pointed out, changes in its usage needs to be done regardless if we keep this field or not, so I would prefer to remove old stuff and stop maintaining it on this change and all the other ones that will come some time.

htreu:
Strange: Travis complains about

```
[ERROR] /home/travis/build/eclipse/smarthome/bundles/io/org.eclipse.smarthome.io.transport.mqtt/src/main/java/org/eclipse/smarthome/io/transport/mqtt/MqttService.java:[48] 
[ERROR] 	@Component(immediate = true, service = { MqttService.class }, configurationPid = {
[ERROR]         "org.eclipse.smarthome.mqtt" }, property = { Constants.SERVICE_PID + "=org.eclipse.smarthome.mqtt" })
[ERROR] 	                                                                                 ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
[ERROR] Type mismatch: cannot convert from String[] to String
```

Locally it does compile fine.

davidgraeff:
Hm, but why. A fast lookup on google let me find this example: `@Component(property={EventConstants.EVENT_TOPIC+"=osgi/enroute/*"})`. This is what we do here.

htreu:
My guess is, that `configurationPid` is the parameter in question (although it is in fact defined as String[] in `@Component`). I have no clue why we see this error.
Please give it a try.

maggu2810:
That artifact is used for the annotations on a Maven build (see `pom.xml`):

```
    <dependency>
      <groupId>org.apache.felix</groupId>
      <artifactId>org.apache.felix.scr.ds-annotations</artifactId>
      <version>1.2.10</version>
    </dependency>
```

This artifact itself uses the `org.osgi.service.component.annotations` package of

```
    <dependency>
      <groupId>org.osgi</groupId>
      <artifactId>org.osgi.compendium</artifactId>
      <version>4.3.0</version>
    </dependency>
```

Have a look at [Release 4 Version 4.3 - Annotation Type Component]: `configurationPid` needs to be a `String`, not an array.
That differs from [Release 6 - Annotation Type Component]: `configurationPid` is a `String[]` (and settings one value for an array with one element is supported in annotations AFAIK)

[Release 4 Version 4.3 - Annotation Type Component]: https://osgi.org/javadoc/r4v43/residential/org/osgi/service/component/annotations/Component.html
[Release 6 - Annotation Type Component]: https://osgi.org/javadoc/r6/residential/org/osgi/service/component/annotations/Component.html

htreu:
Thanks @maggu2810. I always searched our org.eclipse.osgi.services-3.6.0 dependency where String[] is defined. Where do we define the OSGi DS Annotation version to be used? Or: who is right? the maven dependency or the IDE dependency?

> and settings one value for an array with one element is supported in annotations AFAIK

Yes, I just observed that in the annotation processor sources.

maggu2810:
The development guidelines states: "The minimum OSGi framework version supported is OSGi R4.2, no newer features must be used."

The OSGi stuff that is available in the Eclipse IDE provides much more recent versions, so in this case the Maven one is "more correct" as it uses an older version of the annotations.

Another example: In the Eclipse IDE you can also annotate a field in a component using `@Reference` but this is supported with starting with OSGi R6 only. Also in this case you will succeed in the IDE but fail with Maven.

htreu:
Thanks. And to complete the picture (since you referenced R4V4.**3**) we can use these because no runtime support is needed (I see no *.annotation package in the 4.2 API).
Would it be an option to exchange the org.eclipse annotations by the org.osgi annotations in the IDE (sorry if this had been discussed before...)?

maggu2810:
> exchange the org.eclipse annotations by the org.osgi annotations in the IDE

Which one ("org.eclipse annotations") do you mean?
Isn't the problem that the IDE is using the org.osgi annotations in a more recent version that the Maven build which is caused by the stuff that is present in the Target Platform?

davidgraeff:
Guys :) I changed the configurationpid to a simple string and would like to go on with this PR.

htreu:
@davidgraeff agreed :-)
Once again Travis is taking the lead:
```
[ERROR] Failed to execute goal org.eclipse.tycho:tycho-compiler-plugin:1.1.0:compile (default-compile) on project org.eclipse.smarthome.io.transport.mqtt: Compilation failure: Compilation failure: 
[ERROR] /home/travis/build/eclipse/smarthome/bundles/io/org.eclipse.smarthome.io.transport.mqtt/src/main/java/org/eclipse/smarthome/io/transport/mqtt/MqttService.java:[135] 
[ERROR] 	connection = new MqttBrokerConnection(config.host, config.port, config.secure, config.clientID);
[ERROR] 	                                      ^^^^^^^^^^^
[ERROR] Null type mismatch (type annotations): required '@NonNull String' but this expression has type '@Nullable String'
```

davidgraeff:
Because the IDE has more background knowledge. Look at the code in question, especially
**config.host != null**:
```
if (StringUtils.isNotBlank(config.host) && config.host != null) {
        connection = new MqttBrokerConnection(config.host, config.port, config.secure, config.clientID);
        brokerConnections.put(brokerID, connection);
} 
```

htreu:
come on, don't let travis win!
Now tests fail due to unresolved dependencies. Following imports are missing:
```
org.mockito.invocation,
org.mockito.stubbing
```

After correcting this there is also one test left failing:

```
java.lang.AssertionError: Name getter expected:<1883> but was:<null>
	at org.junit.Assert.fail(Assert.java:88)
	at org.junit.Assert.failNotEquals(Assert.java:834)
	at org.junit.Assert.assertEquals(Assert.java:118)
	at org.eclipse.smarthome.io.transport.mqtt.MqttBrokerConnectionTest.setterGetterTests(MqttBrokerConnectionTest.java:169)
```

You now always initialise the port. Now the test assumption is false.

davidgraeff:
@htreu Tests corrected. And squashed.

PR #4177, opened 2017-09-01
Add the ability to localize a thing discovery

### LINKED ISSUE #4025: Localizing the thing discovery result
It would be a good idea to localize the label of a thing discovery result to give the opportunity to each binding implementing a discovery service to provide a label that can be translated in different languages.

Several ESH bindings are currently providing a label that is hardcoded in English like for exemple "Local Time" in the NTP binding, "Local Sun" and "Local Moon" in the Astro binding or "Local Weather" in the WU binding. This label is displayed in Paper UI in the inbox. Hopefully, the user has the ability in Paper UI to update this label before the thing is finally created.

This enhancement would require the creation of a new translation service, something similar to what was implemented for the thing status description where you can use `"@text/xxx"` as description. It should be possible to use `"@text/xxx"` in the `withLabel` method of the discovey result builder.

Of course, the translation service should be requested in each place where the discovery result is used, for example in the console command displaying the discovery entries of the inbox.

If noone is voluntary, I could try to impement that.

lolodomo:
Use this feature in the astro, the NTP and the WU bindings.
Update the astro and NTP bindings to use DS annotations to declare the services.

Fixes #4025


extensions/binding/org.eclipse.smarthome.binding.ntp/ESH-INF/i18n/ntp_de.properties
```diff
@@ -34,3 +34,6 @@ channel-type.config.ntp.string-channel.DateTimeFormat.description = Format f
 offline.conf-error-init-handler = Fehler bei der Initialisierung des NTP handler.
 offline.comm-error-unknown-host = Zeitserver {0} ist unbekannt. Systemzeit wird zur�ckgegeben.
 offline.comm-error-connection = Verbindung zum Zeitserver {0} kann nicht aufgebaut werden. Systemzeit wird zur�ckgegeben.
+
+# Discovery result
+#discovery.ntp.ntp.local.label = Local Time
```
cweitkamp: You can use 'Lokale Zeit' for geman translation. ;-)

lolodomo: Done

lolodomo:
Nobody to review this PR ?

extensions/binding/org.eclipse.smarthome.binding.astro/OSGI-INF/.gitignore
```diff
@@ -0,0 +1 @@
+*.xml
```
J-N-K: It is quite unexpected from the PR title that you changed to automatic XML generation. Wouldn't it be better to move that to a seperate PR and leave this one for the i18n only?

extensions/binding/org.eclipse.smarthome.binding.astro/src/main/java/org/eclipse/smarthome/binding/astro/discovery/AstroDiscoveryService.java
```diff
@@ -31,11 +38,14 @@
  * @author Gerhard Riegler - Initial Contribution
  * @author Stefan Triller - Use configured location
  */
+@Component(service = DiscoveryService.class, immediate = true, configurationPid = "discovery.astro")
```
J-N-K: see above

extensions/binding/org.eclipse.smarthome.binding.astro/src/main/java/org/eclipse/smarthome/binding/astro/discovery/AstroDiscoveryService.java
```diff
@@ -119,4 +132,22 @@ protected void unsetLocationProvider(LocationProvider locationProvider) {
         this.locationProvider = null;
     }
 
+    @Reference
+    protected void setLocaleProvider(final LocaleProvider localeProvider) {
```
J-N-K: These methods seem to be common for all i18n discovery services. Couldn't you move them to a new class "AbstractI18nDiscoveryService" or something like that from which the actual implementation inherits? This would reduce code duplication.

extensions/binding/org.eclipse.smarthome.binding.astro/src/main/java/org/eclipse/smarthome/binding/astro/internal/AstroHandlerFactory.java
```diff
@@ -28,6 +30,7 @@
  *
  * @author Gerhard Riegler - Initial contribution
  */
+@Component(service = ThingHandlerFactory.class, immediate = true)
```
J-N-K: see above

extensions/binding/org.eclipse.smarthome.binding.astro/src/main/java/org/eclipse/smarthome/binding/astro/discovery/AstroDiscoveryService.java
```diff
@@ -31,11 +38,14 @@
...
  * @author Stefan Triller - Use configured location
  */
+@Component(service = DiscoveryService.class, immediate = true, configurationPid = "discovery.astro")
 public class AstroDiscoveryService extends AbstractDiscoveryService {
     private final Logger logger = LoggerFactory.getLogger(AstroDiscoveryService.class);
     private static final int DISCOVER_TIMEOUT_SECONDS = 30;
     private static final int LOCATION_CHANGED_CHECK_INTERVAL = 60;
     private LocationProvider locationProvider;
+    private Locale locale;
+    private DiscoveryResultI18nUtil discoveryResultI18nUtil;
```
J-N-K: These members seem to be common for all i18n discovery services. Couldn't you move them to a new class "AbstractI18nDiscoveryService" or something like that from which the actual implementation inherits? This would reduce code duplication.

bundles/config/org.eclipse.smarthome.config.discovery/src/main/java/org/eclipse/smarthome/config/discovery/i18n/DiscoveryResultI18nUtil.java
```diff
@@ -0,0 +1,56 @@
...
+
+        String label = i18nProvider.getText(bundle, key, defaultLabel, locale);
+
+        return new DiscoveryResultImpl(discoveryResult.getThingTypeUID(), discoveryResult.getThingUID(),
+                discoveryResult.getBridgeUID(), discoveryResult.getProperties(),
+                discoveryResult.getRepresentationProperty(), label, discoveryResult.getTimeToLive());
+    }
+
+    private String inferKey(DiscoveryResult discoveryResult, String lastSegment) {
+        return "discovery." + discoveryResult.getThingUID().getAsString().replaceAll(":", ".") + "." + lastSegment;
```
J-N-K: Using the ThingUID works for the bindings you selected. For the WeMo binding this will not work, because the ThingUID will be different for each thing, but the label would need translation independently of that ([see here](https://github.com/eclipse/smarthome/blob/7fe6f9081a0c5800ee8bbe968d22f7fab5db96a3/extensions/binding/org.eclipse.smarthome.binding.wemo/src/main/java/org/eclipse/smarthome/binding/wemo/discovery/WemoDiscoveryParticipant.java#L44)).

extensions/binding/org.eclipse.smarthome.binding.weatherunderground/ESH-INF/i18n/weatherunderground.properties
```diff
@@ -8,3 +8,6 @@ offline.conf-error-missing-location = The "location" parameter must be configure
 offline.conf-error-syntax-language = The value of the "language" parameter must contain 2 letters.
 offline.conf-error-min-refresh = The minimum value of the "refresh interval" parameter must be 5 minutes.
 offline.uri-error = Could not create URI to fetch data, please check if the location parameter is correct.
+
+# Discovery result
+discovery.weatherunderground.weather.local.label = Local Weather
```
J-N-K: German translation is "Lokales Wetter"

extensions/binding/org.eclipse.smarthome.binding.ntp/src/main/java/org/eclipse/smarthome/binding/ntp/discovery/NtpDiscovery.java
```diff
@@ -26,8 +33,12 @@
  *
  * @author Marcel Verpaalen - Initial contribution
  */
+@Component(service = DiscoveryService.class, immediate = true, configurationPid = "discovery.ntp")
 public class NtpDiscovery extends AbstractDiscoveryService {
 
+    private Locale locale;
```
J-N-K: see above

extensions/binding/org.eclipse.smarthome.binding.ntp/src/main/java/org/eclipse/smarthome/binding/ntp/discovery/NtpDiscovery.java
```diff
@@ -63,11 +74,30 @@ private void discoverNtp() {
...
-                    .withLabel("Local Time").build();
+            DiscoveryResult result = this.discoveryResultI18nUtil.createLocalizedDiscoveryResult(this,
+                    DiscoveryResultBuilder.create(uid).withProperties(properties).withLabel("Local Time").build(),
+                    this.locale);
             thingDiscovered(result);
         }
 
     }
 
+    @Reference
```
J-N-K: see above

extensions/binding/org.eclipse.smarthome.binding.ntp/src/main/java/org/eclipse/smarthome/binding/ntp/internal/NtpHandlerFactory.java
```diff
@@ -23,10 +26,12 @@
  * @author Marcel Verpaalen - Initial contribution
  * @author Markus Rathgeb - Add locale provider support
  */
+@Component(service = ThingHandlerFactory.class, immediate = true)
```
J-N-K: see above

extensions/binding/org.eclipse.smarthome.binding.weatherunderground/src/main/java/org/eclipse/smarthome/binding/weatherunderground/internal/discovery/WeatherUndergroundDiscoveryService.java
```diff
@@ -46,6 +50,8 @@
     private static final int LOCATION_CHANGED_CHECK_INTERVAL = 60;
 
     private LocationProvider locationProvider;
+    private Locale locale;
```
J-N-K: see above

extensions/binding/org.eclipse.smarthome.binding.weatherunderground/src/main/java/org/eclipse/smarthome/binding/weatherunderground/internal/discovery/WeatherUndergroundDiscoveryService.java
```diff
@@ -123,4 +129,22 @@ protected void unsetLocationProvider(LocationProvider locationProvider) {
         this.locationProvider = null;
     }
 
+    @Reference
```
J-N-K: see above

extensions/binding/org.eclipse.smarthome.binding.astro/OSGI-INF/.gitignore
lolodomo: Because I had to reference new OSGi services. It was a perfect opportunity to move to DS annotations.

extensions/binding/org.eclipse.smarthome.binding.astro/src/main/java/org/eclipse/smarthome/binding/astro/discovery/AstroDiscoveryService.java
lolodomo: That is probably doable. Will study that.

extensions/binding/org.eclipse.smarthome.binding.astro/src/main/java/org/eclipse/smarthome/binding/astro/discovery/AstroDiscoveryService.java
lolodomo: It will require to declare a new specific service that will reference itself the required services. Then each binding discovery service that needs i18n will use this service.
@kai suggested to not create a new specific service for this feature, so I tried to use/enhance the existing binding discovery services.

bundles/config/org.eclipse.smarthome.config.discovery/src/main/java/org/eclipse/smarthome/config/discovery/i18n/DiscoveryResultI18nUtil.java
lolodomo: For WeMo like for Sonos for example, the used label is built from the product name returned by UPnP. I don't see a real need for translation. How could you anticipate all possible discovered things ?
I am clearly not covering this.

extensions/binding/org.eclipse.smarthome.binding.weatherunderground/ESH-INF/i18n/weatherunderground.properties
lolodomo: This was provided in an additional commit.

bundles/config/org.eclipse.smarthome.config.discovery/src/main/java/org/eclipse/smarthome/config/discovery/i18n/DiscoveryResultI18nUtil.java
J-N-K: WeMo uses a default label "WeMo device" if nothing specific is obtained from the device. This could be translated. I'm working on a OneWire binding, discovered temnperature sensors get the label "temperature sensor" or something like that, this oculd be translated, too.

lolodomo: Ok with you. It should be possible to translate the default label. I think it could be done without any change to my PR by using @text/xxx as label. I will check that.

lolodomo: ... by using 
```
@text/xxx
```

extensions/binding/org.eclipse.smarthome.binding.astro/src/main/java/org/eclipse/smarthome/binding/astro/discovery/AstroDiscoveryService.java
lolodomo: Unfortunately, it seems not possible to inherit annotations from a super class.
If it was possible, I could have moved this to the new abstract class.

extensions/binding/org.eclipse.smarthome.binding.astro/src/main/java/org/eclipse/smarthome/binding/astro/discovery/AstroDiscoveryService.java
lolodomo: I created a new abstract class that replaced the util class.

bundles/config/org.eclipse.smarthome.config.discovery/src/main/java/org/eclipse/smarthome/config/discovery/i18n/DiscoveryResultI18nUtil.java
lolodomo: Ok, I can confirm that it is possible to use
```
"@text/discovery.ntp.defaultLabel"
```
as label in the binding and define for example in the property file:
```
discovery.ntp.defaultLabel = here is a test
```

lolodomo:
Ok, as suggested, I have now a new version with a new abstract class `AbstractI18nDiscoveryService`. It allows to factorize few things and it replaces the previous util class I have created.
Unfortunately, it is not possible to factorize in the abstract class the service reference declaration because DS declarative annotation is not inherited from a super class. @sjka @kai : could you please confirm that annoying point ? As a consequence the set/unset methods have to be defined in all discovery class that extends AbstractI18nDiscoveryService rather than only in class AbstractI18nDiscoveryService.

As explained in one of my comments, this PR allows any binding to set a reference to a property as discovery label by using `"@text/propertyname"`. This could be used for example by few bindings to define a localized default label for all discovered things.

lolodomo:
Git is sometimes stupid with conflict detection !
Once I have a positive review, I will rebase everything (I don't want to do it many times).

lolodomo:
I am going to simplify my code by overwritting thingDiscovered.

lolodomo:
@J-N-K. What is still mussing ?

lolodomo:
Code simplified and rebased. Now ready for final review and merge.

bundles/config/org.eclipse.smarthome.config.discovery/src/main/java/org/eclipse/smarthome/config/discovery/i18n/AbstractI18nDiscoveryService.java
```diff
@@ -0,0 +1,107 @@
...
+ */
+public abstract class AbstractI18nDiscoveryService extends AbstractDiscoveryService {
+
+    protected TranslationProvider i18nProvider;
+    protected Locale locale;
+
+    /**
+     * Creates a new instance of this class with the specified parameters.
+     *
+     * @param supportedThingTypes
```
maggu2810: @kaikreuzer Can we add a guideline for the JavaDoc format in the source code?
I don't see any reason why we should add a line break between the parameter name and the description.
I don't see any reason to add empty lines between `@param`, `@return`, `@throws` etc.

maggu2810: http://www.oracle.com/technetwork/java/javase/documentation/index-137868.html

`@param`

```
Additional spaces can be inserted between the name and description so that the descriptions line up in a block.
```

The only about new lines in general is between description and tags.

```
Insert a blank comment line between the description and the list of tags, as shown.
```

I know that there is no "forbidden" stated in that document but at least we in our project should use the same format (at least for every newly created one). So can we agree to drop that line breaks after the parameter name (exception name) and also the empty lines between the tags?

lolodomo: For information, I just copied/pasted this from the parent (existing) class without any change.

maggu2810: @lolodomo Thanks for that information.

I know that there are some usage of that style. I would not create a comment that you need to change an existing one that is not related to your changes.
I just want to ensure that new code / documentation is using the style we use.

As long as there is no guideline -- so "the style we have to use" -- I just could ask you to change it but I would like to know if @kaikreuzer and @sjka agree with me.

lolodomo: But maybe the correct approach in my case would be to just inherit the java doc rather than duplicating it ?

kaikreuzer: > but I would like to know if @kaikreuzer and @sjka agree with me.

Yes, I agree. But as this is code formatting, I would refrain from specifically adding it to the guidelines (because we have a million other such things like where to put the brackets...) - we simply should make sure that the existing code base follows that scheme.

kaikreuzer:
Sorry for being late with my review comments.
I have to admit that I overlooked the aspect that the "I18NUtil" classes also require the TranslationService to be passed in - [my suggestion](https://github.com/eclipse/smarthome/issues/4025#issuecomment-326211952) was assuming that it is enough to statically use the Util class, which would have tremendously simplified the implementation...

My wish would be to have as little implementation overhead as possible for binding developers. We should not end up with a feature that is only used in exceptional cases, but it should imho be the normal thing to do when providing binding internationalisation.

So without having thought it through in depth, I would see different options to improve the implementation of this PR:
- We could add the i18n specific things directly to AbstractDiscoveryService, instead of adding a new class AbstractI18nDiscoveryService. We would still have the issue that people need to declare the setter/unsetter methods correctly for using it.
- We could have discovery services set `@text/...` strings in the DiscoveryResult labels only; the replacement could be done centrally within the `PersistentInbox´, if the keys are uniquely defined - if this is possible, it would mean no additional efforts at all for the binding developers.

Wdyt?

lolodomo:
I was thinking about your first option this morning. It will avoid having two different classes. I think it is a good idea. i can provide this change very quickly.

With your second option, we loose the key naming convention and all binding will be able to use their own convention for key naming. Not ideal I think. But this solution should be doable if we have already a particular service for PersistentInbox.

lolodomo:
PR updated to suppress the new abstract class.

J-N-K:
I was thing of weatherunderground_de.properties. But that file is missing at all in the repo (only fr exists), so forget about that comment.

bundles/config/org.eclipse.smarthome.config.discovery/src/main/java/org/eclipse/smarthome/config/discovery/AbstractDiscoveryService.java
```diff
@@ -249,6 +258,20 @@ protected synchronized void stopScan() {
      *            Holds the information needed to identify the discovered device.
      */
     protected void thingDiscovered(DiscoveryResult discoveryResult) {
+        if (this.i18nProvider != null && this.locale != null) {
```
maggu2810: Just an idea: We could change this method signature to

```Java
protected void thingDiscovered(DiscoveryResult discoveryResult, final @Nullable TranslationProvider translationProvider, final @Nullable Locale locale)
```

and create an obsolete method

```Java
protected final void thingDiscovered(DiscoveryResult discoveryResult) {
    thingDiscovered(discoveryResult, null, null);
}
```

Why at all? To ensure that the "I read not every updated documentation" developer knows that he should provide the two additional references to the abstract base class implementation.

kaikreuzer: The good thing is that the developers do not need to provide these two additional references at all - only the ones who want to provide translated labels would have to make sure that their DiscoveryService gets LocaleProvider and TranslationProvider injected.

lolodomo:
Is there still something preventing the merge of my PR ?
I see new PR updating discoveryservice  for NTP and astro bindings. I am afraid they could introduce new conflcts with my PR. It alreadly takes me a lot of time to rebase everything a first time, I am not sure I will do it a second time.

lolodomo:
Excellent, thank you.

afuechsel:
I am missing the actual usage of the new feature. The PR created only the chance to set localized labels using the usual constant definition. But the Astro Binding still calls thingDiscovery with a label "Local Moon" instead of providing "@text/discovery.astro.moon.local.label"
Did I overlook something?

afuechsel:
@lolodomo am I right?

lolodomo:
Yes, I think you"re wrong. There will be a translation as soon as the expected label is declared in the relevant property file.

afuechsel:
@lolodomo You are right, I did not understand the inferKey functionality correctly in the first place.

ivivanov-bg:
And how would you suggest to do the translation for discovery services registered with service reference (eg. ```HueLightDiscoveryService``` )

afuechsel:
Like this for example - will create a PR for it: 

````java 
    @NonNullByDefault({})
    private ServiceReference<?> translationProviderSrvRef;
    @NonNullByDefault({})
    private ServiceReference<?> localeProviderSrvRef;

    public void activate(BundleContext bundleContext) {
        hueBridgeHandler.registerLightStatusListener(this);
        translationProviderSrvRef = bundleContext.getServiceReference(TranslationProvider.class.getName());
        if (translationProviderSrvRef != null) {
            this.i18nProvider = (TranslationProvider) bundleContext.getService(translationProviderSrvRef);
        }
        localeProviderSrvRef = bundleContext.getServiceReference(LocaleProvider.class.getName());
        if (localeProviderSrvRef != null) {
            this.localeProvider = (LocaleProvider) bundleContext.getService(localeProviderSrvRef);
        }
    }

    public void deactivate(BundleContext bundleContext) {
        removeOlderResults(new Date().getTime(), hueBridgeHandler.getThing().getUID());
        hueBridgeHandler.unregisterLightStatusListener(this);
        if (translationProviderSrvRef != null) {
            bundleContext.ungetService(translationProviderSrvRef);
            this.i18nProvider = null;
        }
        if (localeProviderSrvRef != null) {
            bundleContext.ungetService(localeProviderSrvRef);
            this.localeProvider = null;
        }
    }
```` 

BundleContext can be injected by the caller. It's a little bit awkward, but would work I guess. Creating PR to prove it.

ivivanov-bg:
I have some second thoughts about this approach but let's discuss it in the PR you will create

afuechsel:
@ivivanov-bg Just discovered that it does *not* make any sense for the Hue binding. I will *not* create any PR. 

Explanation: The key for localization is retrieved using the thing UID in `AbstractDiscoveryServce#inferKey`, so it be for example `discovery.hue.0210.001788141f1a.1.label` which cannot be declared in a property file. Furthermore, we have several lamps of the same type connected to a bridge, the label should clearly distinguish between Lamp 1, 2 or 3. Therfore the HueLightDiscoveryService retrieves the label of the result from the hue bridge directly. 

Said that it proves for me that having a label on a discovery result is more or less mandatory, so why does the DiscoveryResultBuilder not make it mandatory? @kaikreuzer any thoughts about that?

afuechsel:
@lolodomo There is another thing I don't understand: the inferKey method used the Thing UID to create a key. So this approach would not work for all bridge based bindings, that use some bridge id in their thing UIDs. Wouldn't it be better to use the thing type UID to retrieve the key?

lolodomo:
The inferkey is only for the simple case and simple binding. More complex bindings with bridge and different thing types should implement the usual and classic  i18n translation mechanism.

Issue #4095, opened 2017-08-19
[Sonos] Error when using setpoint for volume in sitemap


metbril:
I use a setpoint for my volume and found that (since openHAB 2.1?) this no longer is working.
As a workaround I can use a slider, that has less finegrained control for me. I am sure that the setpoint has worked with OH2.0, build don't know when it broke.

Running OH2.1 Stable on a RPi3B/openhabian

sonos.items
```
Dimmer Sonos_Living_Volume  "Volume [%.0f %%]" <soundvolume>      (Sonos_Living) { channel="sonos:PLAY1:living:volume" }
```

default.sitemap
```
Slider item=Sonos_Living_Volume
Setpoint item=Sonos_Living_Volume minValue=0 maxValue=100 step=5
```
openhab.log
```
17:14:31.601 [INFO ] [smarthome.event.ItemCommandEvent    ] - Item 'Sonos_Living_Volume' received command 10.0
17:14:31.618 [ERROR] [ome.core.thing.internal.ThingManager] - Exception occurred while calling handler: org.jupnp.model.types.InvalidValueException: Can't convert string to number or not in range: 10.0
java.util.concurrent.ExecutionException: org.jupnp.model.types.InvalidValueException: Can't convert string to number or not in range: 10.0
	at org.eclipse.smarthome.core.common.SafeMethodCaller.executeDirectly(SafeMethodCaller.java:220)[99:org.eclipse.smarthome.core:0.9.0.b5]
	at org.eclipse.smarthome.core.common.SafeMethodCaller.callAsynchronous(SafeMethodCaller.java:189)[99:org.eclipse.smarthome.core:0.9.0.b5]
	at org.eclipse.smarthome.core.common.SafeMethodCaller.call(SafeMethodCaller.java:83)[99:org.eclipse.smarthome.core:0.9.0.b5]
	at org.eclipse.smarthome.core.common.SafeMethodCaller.call(SafeMethodCaller.java:67)[99:org.eclipse.smarthome.core:0.9.0.b5]
	at org.eclipse.smarthome.core.thing.internal.ThingManager.receiveCommand(ThingManager.java:374)[106:org.eclipse.smarthome.core.thing:0.9.0.b5]
	at org.eclipse.smarthome.core.items.events.AbstractItemEventSubscriber.receive(AbstractItemEventSubscriber.java:47)[99:org.eclipse.smarthome.core:0.9.0.b5]
[...]

PR #4180, opened 2017-09-03
Sonos: convert command with decimal value to expected integer value

lolodomo:
Should fix issue #4095


kaikreuzer:
@lolodomo Better write "fixes #xxxx" instead of "fixes issue #xxxx" as this will trigger the Github auto-close feature once a PR is merged.

Issue #4178, opened 2017-09-02
Support hierarchical configuration & exclusive cfg files


davidgraeff:
I'm facing the problem in two services (MqttService and #4105 SSLService) that I'm in a need of a simple 2-level hierarchical configuration (parameters grouped in instance-able sections). I'm calling this "sections" in the followup text. I propose to add support to the framework in a way that GUIs can interact with sections (create/modify/delete) and that they can be stored in configuration files a human readable way.

## Comparison to ManagedService+ManagedServiceFactory

OSGi provides `ManagedServiceFactory` for similar use cases like this. Disadvantages:
* Ancient interface (Updates are published through `Dictionary`s)
* Also not supported by ESH at the moment:
   * REST/GUI can't delete instances nor create them
   * ConfigDispatcher can't handle the filename format "pid-subname.cfg" for a factory managed service.
   * The current solution doesn't support writing back to cfg files either
* We still need nested configurations (for example to configure a list of user credentials for a service)

## User interactions

The user would interact with sections as follows:
* A GUI (PaperUI) should be able to create a new section instance, edit an existing one and to remove an instance. Optional limits in the xml description prevent to many or to less instances.
* A cfg compatible file format can be used to express sections (INI sections, yaml indents).

### Example use cases (MqttService)
```
normal=config
other=value

[broker:connection1]
host=123.123.123.123
port=1234
secure=true

[broker:connection2]
....
```

The section type here is "broker" and two instances are configured. The instance name can alternatively be omited:
```
[broker]
host=123.123.123.123
port=1234
secure=true
```

A third alternative is to keep the key=value structure. Dots are used to separate key and instance name. Multiple levels are separated by hashes:
```
broker.connection1#host=123.123.123.123
broker.connection1#port=1234
broker.connection1#secure=true
```

This would be the flat representation of nested key,value structures for ConfigAdmin.

### Example use cases (SSLService)
```
algorithmConfig=secureAlgorithm

[context:default]
javaKeystore=default.keystore
keystorePassword=bfa7823ha7
privateKeyPassword=asdg23jh5a
automaticRefresh=true

[context:specialContext]
domain=www.mysmarthome.com
javaKeystore=external.keystore
keystorePassword=bfa7823ha7
privateKeyPassword=asdg23jh5a
useACME=true
```

### Other use cases
There are probably more use cases than only the two services that needs this. For instance I can think of user management related configuration where a section instance would consist of a username, a password, ssh key hash, etc.

## Implementation

[...]

Issue #3692, opened 2017-06-17
Allow add-ons to provide custom meta-data on items


kaikreuzer:
This is a continuation of the discussion at https://github.com/eclipse/smarthome/issues/1093#issuecomment-303422314.

We have multiple add-ons (HomeKit, Alexa, ImperiHome, ...) that want to expose items to external systems and which require some special meta-information like a mapping to the type system of the remote system.

So far the approach was to use semantic tagging, but this has the drawback that it isn't possible to have different meta-data for different add-ons and to decide which ones to include or exclude on a per add-on basis.

I think we therefore need some other mechanism to provide that information. So far my recommendation usually was to come up with an add-on specific configuration file and create a dedicated REST endpoint for exposing the data.
@digitaldan [challenged that](https://github.com/eclipse/smarthome/issues/1093#issuecomment-303735814) and fears that we will end up with too many endpoints that would fragment the REST API and be a potential security risk.

Also, we should have a common way how to configure such meta-data - we should not force every add-on to come up with its own UI to do this job.

martinvw:
A very generic idea. So what if add-ons could implement a certain class to provide scoped (prefixed like Alexa:) meta-tags. The ui could then provide a drop down based on all those provided tags

anthonygillet:
Continuing the discussion from 1093...

> Note that a boolean flag (expose/don't expose) won't be enough in many situations - you might want to add a specific homekit-type mapping to one item and a custom label to use with Alexa on another item - so I am afraid, the UI will have to be a bit more complex...

Could you elaborate on what you mean here?  I'm not sure exactly why a boolean flag would not work.  I figured that the complexity should be pushed down to the individual addons.  Once an item is exposed to an addon, the addons themselves could provide additional metadata if necessary.  For example, HomeKit internally will need to map the OH semantic tags to HK accessory types.  Shouldn't it be up to the HomeKit addon to provide a way to configure that if desired, rather than OH?

You mention custom labels - I presume you mean the ability to specify alternative names for items to be controlled by voice.  That seems like a pretty complex issue.  For now, users can achieve multiple labels / alternative voice names by simply having multiple items that work together through groups, rules, or sharing channels. (not ideal, I know)

I agree that you don't want to get into a situation where you have to live with a temporary solution forever.  However, a simple boolean UI would unblock development of this (seemingly highly-desired) feature, and keeping it simple would allow iteration later if more control is desired at the OH level.

digitaldan:
[...]

PR #4181, opened 2017-09-03
Allow Configuration to handle nested structures and lists

davidgraeff:
One step towards #4178

The `Configuration` class provides unmarshalling to a POJO class. The `as` method is extended to support nested structures and list of structures.

An example configuration class structure that benefits from nested unmarshalling and from list unmarshalling:
```java
class ServiceConfiguration {
   BigDecimal importantNumber;
   String niceString;
   CoordinateConfiguration theCoordinates;
   List<UserConfiguration> users;
}

class CoordinateConfiguration {
   BigDecimal x;
   BigDecimal y;
}
class UserConfiguration {
   String username;
}
```

## Port of tests to java
I ported some relevant tests to java and added a test case for nested and list fields

## Marshal class back to Map<String, Object>
Add the method `Map<String,Object> readFromConfigurationClassInstance(Object configurationClassInstance) `

## Discussion
* I marked `ConfigUtil:normalizeType(Object value)` deprecated in favour of `ConfigUtil:normalizeType(Object value, @Nullable ConfigDescriptionParameter)`. In my opinion with the deprecated method we are hiding the fact, that we always create a Normalizer in that specific method in a `null` case. The consumer should specifically confirm with null that he will not provide a config description and happily accepts a created Normalizer. An alternative is to create a default Normalizer in the constructor. In that case `ConfigUtil:normalizeType(Object value)` is a valid method signature again.
* I removed the exception that is thrown in `ConfigUtil:normalizeType` (https://github.com/eclipse/smarthome/pull/4181/files#diff-810126e50ad4f2c2ad5d985ffd577ed4L87) if the object is not a String or a Number. It was a non checked exception, not documented and not in the unit tests and not part of the API, so I guess nothing depends on this exception to be thrown. Instead the value object is just returned as-is. This change is necessary to allow Map and List objects.


sjka:
Just a (very) minor remark: `Configuration` is still using the newly deprecated method signature:

https://github.com/davidgraeff/smarthome/blob/4faa4b1b5d1a77d72a080292f04dc8f284dd469b/bundles/config/org.eclipse.smarthome.config.core/src/main/java/org/eclipse/smarthome/config/core/Configuration.java#L234

Otherwise this looks great!

> I marked ConfigUtil:normalizeType(Object value) deprecated in favour of [...]

👍 

> I removed the exception that is thrown in ConfigUtil:normalizeType  [...] if the object is not a String or a Number. [...] Instead the value object is just returned as-is. 

I'm also not aware of any code "relying" on it. The only aspect I can think of is that it pointed out beautifully that there something was seriously wrong. On the other hand: there is still something going to explode nicely downstream if the types don't match, so I'd be okay to leave it as it is now and see how it goes.

sjka:
Could you please rebase this PR?

bundles/config/org.eclipse.smarthome.config.core/src/main/java/org/eclipse/smarthome/config/core/Configuration.java
```diff
@@ -53,63 +56,143 @@ public Configuration(Map<String, Object> properties) {
...
+    /**
+     * Creates an instance of the given configuration class (a default constructor is required) and
+     * tries to find a matching property value for each non-final field.
+     * Nested classes are supported.
+     * List fields are supported as long as the property value is of type Collection.
+     *
+     * @param configurationClass
+     * @return
+     */
+    @Nullable
```
maggu2810: Type annotations should be placed at the type and not on the visible modifier

bundles/config/org.eclipse.smarthome.config.core/src/main/java/org/eclipse/smarthome/config/core/Configuration.java
```diff
@@ -53,63 +56,143 @@ public Configuration(Map<String, Object> properties) {
...
+    }
+
+    /**
+     * Creates an instance of the given configuration class (a default constructor is required) and
+     * tries to find a matching property value for each non-final field.
+     * Nested classes are supported.
+     * List fields are supported as long as the property value is of type Collection.
+     *
+     * @param configurationClass
+     * @return
```
maggu2810: an instance of the given class on success, null on error

kaikreuzer:
> Otherwise this looks great!

@sjka This pretty much contradicts [your comment here](https://github.com/eclipse/smarthome/issues/3692#issuecomment-317672269) - although I was specifically talking about Maps, the general theme was the same: Having more complex structures as configurations.

The big issue with the suggestions (mine as well as David's) is that we do not stand a chance to provide proper config descriptions for them, nor to have it in any way supported in the UIs - and that was the ultimate goal of all the configuration infrastructure.

So although I'd personally like to be able to have more complex configurations, I do not see a way to move away from flat lists of primitive types. Do I miss anything?

kaikreuzer:
Hm, ok, I just read over https://github.com/eclipse/smarthome/issues/4178 and I see that I completely lost track of all activities over the last week - my suggestion would be to first discuss the general vision on #4178 and only process the individual PRs afterwards.

Ok for everyone?

davidgraeff:
That was the plan. The PRs are proof of concept. It is possible to allow complex structures in the core. I haven't check the GUI situation yet. 

Another point to analyse is, how this would influence thing configuration. Because the dtd is shared between every way of configuration, things would be capable of having complex configs as well. This would need to be supported by storage, or the config needs to be flattened at some point.

davidgraeff:
I will alter this PR, and add annotations for configuration fields. Annotations that I have in mind are:
* FieldRequired
* SecretVaultField
* Base64Chars
* Base64Bytes

htreu:
@davidgraeff is there any activity planed towards your latest proposal for this PR? Please give a short update.

davidgraeff:
Can be closed.

PR #4186, opened 2017-09-05
add nullness annotations for persistence service

maggu2810:


sjka:
```
[ERROR] Failed to execute goal org.eclipse.tycho:tycho-compiler-plugin:1.0.0:compile (default-compile) on project org.eclipse.smarthome.core.test: Compilation failure: Compilation failure:
[ERROR] /home/travis/build/eclipse/smarthome/bundles/core/org.eclipse.smarthome.core.test/src/test/groovy/org/eclipse/smarthome/core/items/TestItem.java:[28]
[ERROR] return null;
[ERROR] ^^^^
[ERROR] Null type mismatch: required '@NonNull List<Class<? extends State>>' but the provided value is null
[ERROR] /home/travis/build/eclipse/smarthome/bundles/core/org.eclipse.smarthome.core.test/src/test/groovy/org/eclipse/smarthome/core/items/TestItem.java:[33]
[ERROR] return null;
[ERROR] ^^^^
[ERROR] Null type mismatch: required '@NonNull List<Class<? extends Command>>' but the provided value is null
[ERROR] /home/travis/build/eclipse/smarthome/bundles/core/org.eclipse.smarthome.core.test/src/test/java/org/eclipse/smarthome/core/library/types/ArithmeticGroupFunctionTest.java:[213]
[ERROR] return null;
[ERROR] ^^^^
[ERROR] Null type mismatch: required '@NonNull List<Class<? extends State>>' but the provided value is null
[ERROR] /home/travis/build/eclipse/smarthome/bundles/core/org.eclipse.smarthome.core.test/src/test/java/org/eclipse/smarthome/core/library/types/ArithmeticGroupFunctionTest.java:[218]
[ERROR] return null;
[ERROR] ^^^^
[ERROR] Null type mismatch: required '@NonNull List<Class<? extends Command>>' but the provided value is null
[ERROR] -> [Help 1]
```

\*haha*

bundles/core/org.eclipse.smarthome.core/src/main/java/org/eclipse/smarthome/core/items/Item.java
```diff
@@ -77,7 +77,7 @@
      *
      * @return a list of data types that can be used to update the item state
      */
-    public List<Class<? extends State>> getAcceptedDataTypes();
+    public @NonNull List<@NonNull Class<@NonNull ? extends State>> getAcceptedDataTypes();
```
sjka: Maybe stupid question: What does `Class<@NonNull ? extends State>`tell me?

maggu2810: Okay, perhaps for Class this is really not necessary.
It would state that if the generic template type T (`? extends State`) is used e.g. on some methods return value, you already state that in this case the type is non null.
So, `T cast(Object obj)` would return a non-null type / same for newInstance.
But it is not only not necessary, but also wrong in this case.

> *haha*

Is a failing test such funny? :wink:

sjka: >> haha

> Is a failing test such funny? 😉

No. Just a (bad) attempt of a Nelson (from The Simpsons) cite.

sjka: PS ...and thanks for the explanation!

maggu2810:
Shouldn't the configuration for the test launch configuration in the Eclipse IDE and the one used by Maven (at least "what" is tested) be equal?
If I start the tests in the Eclipse IDE (after fixing the launch configuration to be runnable again), all tests succeeded.

maggu2810:
> Just a (bad) attempt of a Nelson (from The Simpsons) quote.

All fine :wink: 

Tests has been fixed. IMHO non-null empty collections are okay. At least the JavaDoc already states that  e.g. a list is returned...

sjka:
I think so too.

PR #4189, opened 2017-09-05
NetUtils: Accomodate for different needs of getInterfaceAddresses()

davidgraeff:
I'm sorry for the API break. The method is only in core since a few days but I propose these changes:

* Adapt the name, instead of `getAllInterfaceAddresses` it is `getAllInterfaceAddresses` to be in line with other methods of the class.
* Return a collection instead of a set.
* Introduce a CIDR notation class. The boolean flag is therefore removed from the signture of the method.
* Adapt NetworkConfigOptionProvider to use the API. I thing the advantages of a CIDR class (in this case filter for IPv4 addresses) can be seen.

### DIFF
--- a/bundles/core/org.eclipse.smarthome.core/src/main/java/org/eclipse/smarthome/core/net/NetUtil.java
+++ b/bundles/core/org.eclipse.smarthome.core/src/main/java/org/eclipse/smarthome/core/net/NetUtil.java
@@ -198,7 +198,6 @@ public class NetUtil implements NetworkAddressService {
      *
      * @return The collected IPv4 and IPv6 Addresses
      */
-    @SuppressWarnings("null")
     public static Collection<CidrAddress> getAllInterfaceAddresses() {
         Collection<CidrAddress> interfaceIPs = new ArrayList<>();
         Enumeration<NetworkInterface> en;
@@ -221,7 +220,10 @@ public class NetUtil implements NetworkAddressService {
             }
 
             for (InterfaceAddress cidr : networkInterface.getInterfaceAddresses()) {
-                interfaceIPs.add(new CidrAddress(cidr.getAddress(), cidr.getNetworkPrefixLength()));
+                final InetAddress address = cidr.getAddress();
+                assert address != null; // NetworkInterface.getInterfaceAddresses() should return only non-null
+                                        // addresses
+                interfaceIPs.add(new CidrAddress(address, cidr.getNetworkPrefixLength()));
             }
         }
 
--- a/bundles/core/org.eclipse.smarthome.core/src/main/java/org/eclipse/smarthome/core/net/NetUtil.java
+++ b/bundles/core/org.eclipse.smarthome.core/src/main/java/org/eclipse/smarthome/core/net/NetUtil.java
@@ -13,13 +13,12 @@ import java.net.InetAddress;
 import java.net.InterfaceAddress;
 import java.net.NetworkInterface;
 import java.net.SocketException;
+import java.util.ArrayList;
+import java.util.Collection;
 import java.util.Enumeration;
-import java.util.HashSet;
-import java.util.Iterator;
 import java.util.LinkedList;
 import java.util.List;
 import java.util.Map;
-import java.util.Set;
 import java.util.regex.Pattern;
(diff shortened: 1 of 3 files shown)


maggu2810:
If mapping and filtering of the streamed addresses are the common use case, have you considered about returning a stream of `CidrAddress`? If the returned collection is mostly streamed and filtered, we don't need to create the collection, create a stream but could return the stream and the minor use cases that needs a collection could call the `collect` method of the stream.

davidgraeff:
I need to collect the results of the `NetworkInterface.getNetworkInterfaces()` (which returns an Enumeration) and therefore create a collection anyway. A stream is very powerful and you can iterate the stream as well (I consider that the main use case), but in this case we would throw away the computed collection.

maggu2810:
@davidgraeff  My suggestion has been to use streams as early as possible.

e.g.
```Java
    public static Stream<CidrAddress> getAllInterfaceAddresses() {
        Enumeration<NetworkInterface> en;
        try {
            en = NetworkInterface.getNetworkInterfaces();
            LOGGER.error("Could not find interface IP addresses: {}", ex.getMessage(), ex);
            return Stream.empty();
        }

        return enumerationAsStream(en).filter((NetworkInterface networkInterface) -> {
            try {
                return networkInterface.isUp() && !networkInterface.isLoopback();
            } catch (SocketException ignored) {
                return false;
            }
        }).flatMap(networkInterface -> networkInterface.getInterfaceAddresses().stream()
                .filter(cidr -> cidr.getAddress() != null).map(cidr -> {
                    final InetAddress address = cidr.getAddress();
                    assert address != null; // already filtered null address
                    return new CidrAddress(address, cidr.getNetworkPrefixLength());
                }));
    }
```

Create a stream from an enumeration could be done using e.g.

```Java
    private static <T> Stream<T> enumerationAsStream(Enumeration<T> e) {
        return StreamSupport.stream(Spliterators.spliteratorUnknownSize(new Iterator<T>() {
            @Override
            public T next() {
                return e.nextElement();
            }

            @Override
            public boolean hasNext() {
                return e.hasMoreElements();
            }
        }, Spliterator.ORDERED), false);
    }
```

But no need to change it, just an idea if the main use case is to use the result as a stream.

davidgraeff:
@maggu2810 Personally I like the stream API as well, but I'm not so sure if this is efficient anymore especially due to flatMap and if we try to force usage of streams here. It doesn't look more readable and has a helper method. The origin of the problem is of course the dated Enumeration that is returned, but I would go with the collection if you agree.

maggu2810:
Sure, I agree.
Readibility is a must ;)

bundles/core/org.eclipse.smarthome.core/src/main/java/org/eclipse/smarthome/core/net/NetUtil.java
```diff
@@ -210,23 +211,17 @@ public static String getBroadcastAddress() {
...
-                InterfaceAddress interfaceAddress = it.next();
-                if (addPrefix) {
-                    interfaceIPs.add(interfaceAddress.getAddress().getHostAddress() + "/"
-                            + interfaceAddress.getNetworkPrefixLength());
-                } else {
-                    interfaceIPs.add(interfaceAddress.getAddress().getHostAddress());
-                }
 
+            for (InterfaceAddress cidr : networkInterface.getInterfaceAddresses()) {
+                interfaceIPs.add(new CidrAddress(cidr.getAddress(), cidr.getNetworkPrefixLength()));
```
maggu2810: You could remote `@SuppressWarnings("null")` for the whole method implementation using e.g.

```Java
final InetAddress address = cidr.getAddress();
assert address != null; // NetworkInterface.getInterfaceAddresses() should return only non-null addresses
interfaceIPs.add(new CidrAddress(address, cidr.getNetworkPrefixLength()));
```

WDYT? I am not a friend of supress warnings for the whole function. An assert should not change anything on runtime as they are default deactivated.

davidgraeff: I agree

Issue #4187, opened 2017-09-05
Inbox auto ignored results are never removed


afuechsel:
When adding e.g. a hue bridge manually, its result is automatically set to IGNORED in the inbox as expected. When removing the bridge, the IGNORED entry in the inbox must be automatically deleted so a new discovery can find it again. It turns out, that this is not happening most of the time.

The reason is, that we do not have such a thing as a Pre-Delete-Hook, where we could easily check and remove possible ignored entries.Listening to ThingStatusChangedEvents as currently implemented does not help, because a REMOVED event indicates that a thing is already removed and REMOVING is not always fired and sometimes directly followed by a REMOVED. In most cases we are accessing the ThingRegistry too late to get informations about the thing being removed (it just returns null).

dominicdesu:
This fix might have caused a small issue with the hue binding, please see #4375

PR #1682, opened 2016-06-16
Thing properties are never persisted 

afuechsel:
Removed callback.thingUpdated(…) from BaseThingHandler#updateProperties and #updateProperty

fixes: #588

Signed-off-by: Andre Fuechsel <andre.fuechsel@telekom.de>

Issue #588, opened 2015-12-03
Early calls to updateProperties() throws warnings in AbstractManagerProvider.update()


kaikreuzer:
migrated from Bugzilla [#466778](https://bugs.eclipse.org/bugs/show_bug.cgi?id=466778)
status NEW severity *normal* in component *Core* for *---*
Reported in version *unspecified* on platform *PC*
Assigned to: Project Inbox

On 2015-05-07 16:13:16 -0400, karel goderis wrote:

> When creating a new Thing, and ThingHandler, code like
> 
> Map<String, String> properties = editProperties();
> properties.put(selector.getChannelID(),
> entry.getValue().getAsString());
> updateProperties(properties);
> 
> throws a lot of these in the logs
> 
> 2015-05-07 22:08:43 [WARN ] [.c.c.r.AbstractManagedProvider:138  ] - Could not update element with key tesla:models:v12killer in ManagedThingProvider, because it does not exists.
> 
> The call originates from the this.callback.thingUpdated(thing) here below
> 
>   protected void updateProperties(Map<String, String> properties) {
>         boolean propertiesUpdated = false;
>         for (Entry<String, String> property : properties.entrySet()) {
>             String propertyName = property.getKey();
>             String propertyValue = property.getValue();
>             String existingPropertyValue = thing.getProperties().get(propertyName);
>             if (existingPropertyValue == null || !existingPropertyValue.equals(propertyValue)) {
>                 this.thing.setProperty(propertyName, propertyValue);
>                 propertiesUpdated = true;
>             }
>         }
>         if (propertiesUpdated) {
>             synchronized (this) {
>                 if (this.callback != null) {
>                     this.callback.thingUpdated(thing);
>                 } else {
>                     throw new IllegalStateException("Could not update properties, because callback is missing");
>                 }
>             }
>         }
>     }
> 
> which contains
> 
>        public void thingUpdated(Thing thing) {
>             thingUpdatedLock.add(thing.getUID());
>             thingRegistry.update(thing);
>             thingUpdatedLock.remove(thing.getUID());
>         }
> 
> Is there a timing problem as to the Thing being created but not yet being added to the thingRegistry?

On 2015-05-07 18:20:58 -0400, Dennis Nobel wrote:

> How do you add the thing? From the DSL or UI?

On 2015-05-08 03:18:44 -0400, karel goderis wrote:

> The thing is created from the DSL.
[...]

PR #4192, opened 2017-09-05
AutomaticInboxProcessor registered as an RegistryListener 

afuechsel:
for the thing registry to handle REMOVED events correctly

Fixes: #4187

### DIFF
--- a/bundles/config/org.eclipse.smarthome.config.discovery/src/main/java/org/eclipse/smarthome/config/discovery/internal/AutomaticInboxProcessor.java
+++ b/bundles/config/org.eclipse.smarthome.config.discovery/src/main/java/org/eclipse/smarthome/config/discovery/internal/AutomaticInboxProcessor.java
@@ -79,8 +79,6 @@ public class AutomaticInboxProcessor extends AbstractTypedEventSubscriber<ThingS
         }
     }
 
-    // InboxListemer methods
-
     @Override
     public void thingAdded(Inbox inbox, DiscoveryResult result) {
         if (autoIgnore) {
@@ -107,8 +105,6 @@ public class AutomaticInboxProcessor extends AbstractTypedEventSubscriber<ThingS
     @Override
     public void thingRemoved(Inbox inbox, DiscoveryResult result) {
     }
-
-    // RegistryListener methods
 
     @Override
     public void added(Thing element) {
--- a/bundles/config/org.eclipse.smarthome.config.discovery/src/main/java/org/eclipse/smarthome/config/discovery/internal/AutomaticInboxProcessor.java
+++ b/bundles/config/org.eclipse.smarthome.config.discovery/src/main/java/org/eclipse/smarthome/config/discovery/internal/AutomaticInboxProcessor.java
@@ -19,6 +19,7 @@ import org.eclipse.smarthome.config.discovery.DiscoveryResult;
 import org.eclipse.smarthome.config.discovery.DiscoveryResultFlag;
 import org.eclipse.smarthome.config.discovery.inbox.Inbox;
 import org.eclipse.smarthome.config.discovery.inbox.InboxListener;
+import org.eclipse.smarthome.core.common.registry.RegistryChangeListener;
 import org.eclipse.smarthome.core.events.AbstractTypedEventSubscriber;
 import org.eclipse.smarthome.core.events.EventSubscriber;
 import org.eclipse.smarthome.core.thing.Thing;
@@ -55,7 +56,7 @@ import org.slf4j.LoggerFactory;
         "service.config.description.uri=system:inbox", "service.config.label=Inbox", "service.config.category=system",
         "service.pid=org.eclipse.smarthome.inbox" })
 public class AutomaticInboxProcessor extends AbstractTypedEventSubscriber<ThingStatusInfoChangedEvent>
-        implements InboxListener {
+        implements InboxListener, RegistryChangeListener<Thing> {
 
     private final Logger logger = LoggerFactory.getLogger(this.getClass());
 
@@ -78,6 +79,8 @@ public class AutomaticInboxProcessor extends AbstractTypedEventSubscriber<ThingS
         }
     }
 
+    // InboxListemer methods
+
     @Override
     public void thingAdded(Inbox inbox, DiscoveryResult result) {
         if (autoIgnore) {
@@ -97,12 +100,6 @@ public class AutomaticInboxProcessor extends AbstractTypedEventSubscriber<ThingS
         }
     }
 
-    private String getRepresentationValue(DiscoveryResult result) {
-        return result.getRepresentationProperty() != null
-                ? Objects.toString(result.getProperties().get(result.getRepresentationProperty()), null)
-                : null;
-    }
-
     @Override
     public void thingUpdated(Inbox inbox, DiscoveryResult result) {
     }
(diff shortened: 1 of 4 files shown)


bundles/config/org.eclipse.smarthome.config.discovery/src/main/java/org/eclipse/smarthome/config/discovery/internal/AutomaticInboxProcessor.java
```diff
@@ -78,6 +79,8 @@ public void receiveTypedEvent(ThingStatusInfoChangedEvent event) {
         }
     }
 
+    // InboxListemer methods
```
maggu2810: Listener (m => n)

maggu2810: Do we need such marker? I don't think we used it on all the other places.

extensions/binding/org.eclipse.smarthome.binding.hue/src/main/java/org/eclipse/smarthome/binding/hue/handler/HueBridgeHandler.java
```diff
@@ -278,6 +278,7 @@ private void onConnectionResumed() throws IOException, ApiException {
                 properties.put(Thing.PROPERTY_SERIAL_NUMBER, config.getMACAddress().replaceAll(":", "").toLowerCase());
                 properties.put(Thing.PROPERTY_FIRMWARE_VERSION, config.getSoftwareVersion());
                 updateProperties(properties);
+                updateThing(thing);
```
maggu2810: How is this change related to the `AutomaticInboxProcessor registered as an RegistryListener`?

extensions/binding/org.eclipse.smarthome.binding.hue/src/main/java/org/eclipse/smarthome/binding/hue/handler/HueLightHandler.java
```diff
@@ -136,6 +136,7 @@ private synchronized void initializeProperties() {
                 }
                 updateProperty(LIGHT_UNIQUE_ID, fullLight.getUniqueID());
                 isOsramPar16 = OSRAM_PAR16_50_TW_MODEL_ID.equals(modelId);
+                updateThing(thing);
```
maggu2810: How is this change related to the `AutomaticInboxProcessor registered as an RegistryListener`?

bundles/config/org.eclipse.smarthome.config.discovery/src/main/java/org/eclipse/smarthome/config/discovery/internal/AutomaticInboxProcessor.java
htreu: typo

afuechsel: No, will remove it. :)

extensions/binding/org.eclipse.smarthome.binding.hue/src/main/java/org/eclipse/smarthome/binding/hue/handler/HueBridgeHandler.java
afuechsel: If the properties would not be persisted, the automatic inbox listener would be unable to retrieve the representation property during deletion. Please note, that this was on purpose as long as the hue binding was updating the properties in each polling cycle. Now that it is guarded, that problem won't exist anymore and the properties should be persisted.

extensions/binding/org.eclipse.smarthome.binding.hue/src/main/java/org/eclipse/smarthome/binding/hue/handler/HueLightHandler.java
afuechsel: see above

bundles/config/org.eclipse.smarthome.config.discovery/src/main/java/org/eclipse/smarthome/config/discovery/internal/AutomaticInboxProcessor.java
afuechsel: already removed as @maggu2810 suggested

extensions/binding/org.eclipse.smarthome.binding.hue/src/main/java/org/eclipse/smarthome/binding/hue/handler/HueBridgeHandler.java
sjka: Just came across this here. How does it relate to https://github.com/eclipse/smarthome/pull/1682 and esp. https://github.com/eclipse/smarthome/issues/588#issuecomment-226462285?

afuechsel: @sjka I see your point but I wonder how to solve this issue without persisting the properties. As the thing is already gone when the removed callback is called I would not be able to get the property value. Hmm.

kaikreuzer: @afuechsel I is always a bad smell to do some specific code changes in one binding for a concept that should be applicable in the whole framework (and thus be the same for all bindings). Note that after #1682 had been implemented, we meanwhile had https://github.com/eclipse/smarthome/pull/2629 in place. This allows to do Thing updates (structure AND configuration) and persists those updates in case it is a managed Thing. Imho, what is missing in that PR is that it also tries to persist properties in the very same way. If this was done, it would be in line with how configuration updates are handled and it would be exactly what you need here (i.e. this line can be removed as the framework already takes care of everything). Do you agree?

afuechsel: I think, if properties would be persisted in some other way, this line could be removed. But did I understand correctly, that there is still some tiny piece missing (persisting the properties)? Would that mean to revert #1682?

sjka: Yes. See #4688.

afuechsel: Well, then this PR needs to be changed too.

sjka: Yes. See #4688.

afuechsel: Ah, didn't see it in the first place :)

PR #4196, opened 2017-09-05
storage: add containsKey method

maggu2810:
The storage interface returns null if you call get for a key that is not
present in the storage.
If you want to check if a key is present (and differ between non-present
or present but an assigned null value) you currently need to call
`storage.getKeys().contains(key)`.
Normally there exist better implementation that does not require to
create a collection of all keys first and then check that collection.

### DIFF
--- a/bundles/core/org.eclipse.smarthome.core/src/main/java/org/eclipse/smarthome/core/storage/Storage.java
+++ b/bundles/core/org.eclipse.smarthome.core/src/main/java/org/eclipse/smarthome/core/storage/Storage.java
@@ -41,6 +41,16 @@ public interface Storage<T> {
     T remove(@NonNull String key);
 
     /**
+     * Check if the storage contains a key.
+     *
+     * @param key the key
+     * @return true if the storage contains the key, otherwise false
+     */
+    default boolean containsKey(@NonNull String key) {
+        return getKeys().contains(key);
+    }
+
+    /**
      * Gets the value mapped to the key specified.
      *
      * @param key the key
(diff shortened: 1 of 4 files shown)


bundles/core/org.eclipse.smarthome.core/src/main/java/org/eclipse/smarthome/core/storage/Storage.java
```diff
@@ -41,6 +41,16 @@
     T remove(@NonNull String key);
 
     /**
+     * Check if the storage contains a key.
+     *
+     * @param key the key
+     * @return true if the storage contains the key, otherwise false
+     */
+    default boolean containsKey(@NonNull String key) {
+        return getKeys().contains(key);
```
maggu2810: I have added a default implementation to not break any API -- if you use it in your custom products.
Perhaps it would be better to drop that default implementation to force the implementer to use a better / custom implementation. 
WDYT?

sjka: We have our own implementation, but that would be easily manageable. So from my side it would be okay to drop it and break the API in that place.

maggu2810: @kaikreuzer Okay for you to drop the default implementation and risk an API break?

kaikreuzer: Yes, ok for me!

maggu2810:
Removed the default implementation.

PR #4203, opened 2017-09-06
Maintaining ReadyMarkers in a dedicated service

sjka:
...instead of using the SCR for it.

The reason behind this is that on at least one commercial OSGi server when running with activated security manager, registering services takes quite long (up to several hundreds of milliseconds per service). When running on limited hardware resources, this significantly impacts the startup times. Therefore it does not look like such a nice idea to register hundreds of services just as "Ready Markers". 

As an improvement to this, I suggest to have a single service which holds these ready markers and other components can register as listeners. 

This does not mean we shouldn't use services for the "startup level"solution (see #1896). However, it simply turned out that it is not necessarily a good idea to flood the SCR with "marker" services.


bundles/core/org.eclipse.smarthome.core/OSGI-INF/org.eclipse.smarthome.core.internal.service.ReadyServiceImpl.xml
```diff
@@ -0,0 +1,7 @@
+<?xml version="1.0" encoding="UTF-8"?>
```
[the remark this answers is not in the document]
sjka: oh yes, thanks!

maggu2810:
What is the reason to open the trackers async and create multiple schedulers that only use case is to schedule one Job on activation? Does the open of a tracker block the activation such long?

bundles/core/org.eclipse.smarthome.core/src/main/java/org/eclipse/smarthome/core/service/ReadyMarkerFilter.java
```diff
@@ -0,0 +1,65 @@
...
+ * @author Simon Kaufmann - initial contribution and API.
+ *
+ */
+public final class ReadyMarkerFilter {
+
+    private final String identifier;
+    private final String type;
+
+    public ReadyMarkerFilter() {
+        this.type = null;
```
maggu2810: `this(null, null);`

bundles/core/org.eclipse.smarthome.core/src/main/java/org/eclipse/smarthome/core/service/ReadyMarker.java
```diff
@@ -7,12 +7,80 @@
...
+
+    private final String type;
+    private final String identifier;
+
+    public ReadyMarker(@NonNull String type, @NonNull String identifier) {
+        this.type = type;
+        this.identifier = identifier;
+    }
+
+    public String getType() {
```
maggu2810: You could mark that the return type is non null.

bundles/core/org.eclipse.smarthome.core/src/main/java/org/eclipse/smarthome/core/service/ReadyMarker.java
```diff
@@ -7,12 +7,80 @@
...
+        return type;
+    }
+
+    public String getIdentifier() {
```
maggu2810: You could mark that the return type is non null.

bundles/core/org.eclipse.smarthome.core/src/main/java/org/eclipse/smarthome/core/internal/service/ReadyServiceImpl.java
```diff
@@ -0,0 +1,119 @@
...
+/**
+ * Implementation of the {@link ReadyService} interface.
+ *
+ * @author Simon Kaufmann - initial contribution and API.
+ *
+ */
+@Component
+public class ReadyServiceImpl implements ReadyService {
+
+    private static final Logger logger = LoggerFactory.getLogger(ReadyServiceImpl.class);
```
maggu2810: I didn't see the need to use a logger in a static context, so could we keep the logger instance non static?

bundles/core/org.eclipse.smarthome.core/src/main/java/org/eclipse/smarthome/core/internal/service/ReadyServiceImpl.java
```diff
@@ -0,0 +1,119 @@
...
+    private static final ReadyMarkerFilter ANY = new ReadyMarkerFilter();
```
maggu2810: Perhaps we could make the ANY constructor (that uses null twice) private to and use a static function that returns an instance (no need for a singleton). Perhaps the meaning of `ReadyMarkerFilter.any()` is more intuitive then `new ReadyMarkerFilter()`.

sjka: If i make that private, there is no public constructor left. My intention was to keep the one with arguments private in order to prevent the constructor-mess we had at other places and rather have the builder-like pattern using `with...()`-Methods. 

So do I understand you right that using `ReadyMarkerFilter.any().withType("...");` is the usage you are suggesting?

maggu2810: No :wink: 

If I understand the current situation correctly you can thee different type of ready markers. 

* filter type, created by `ReadyMarkerFilter.withType(String type)`
* filter identifier, created by `ReadyMarkerFilter.withIdentifier(String type)`
* filter nothing (the any one), created by `new ReadyMarkerFilter()`

Correct me if I am wrong.

So, I just think about making all constructors private and use for the last one a factory method, too. All three functions should return immutable objects.

* filter type, created by `ReadyMarkerFilter.withType(String type)`
* filter identifier, created by `ReadyMarkerFilter.withIdentifier(String type)`
* filter nothing (the any one), created by `ReadyMarkerFilter.any()`

sjka:
> Does the open of a tracker block the activation such long?

Yes, during tracker-opening, all matching bundles will be "added" to that tracker synchronously by the framework. 
I couldn't believe it myself, but running it on this specific runtime with enabled security manager on a rather-not-so-powerful hardware took several seconds to scan all bundles.

bundles/core/org.eclipse.smarthome.core/src/main/java/org/eclipse/smarthome/core/internal/service/ReadyServiceImpl.java
sjka: Ah, that's the misunderstanding... the `with...(...)`-Methods are not static. Therefore you can create:

* filter nothing, created by `new ReadyMarkerFilter()`
* filter type, created by `new ReadyMarkerFilter().withType(String type)`
* filter identifier, created by `new ReadyMarkerFilter().withIdentifier(String identifier)`
* filter for both, created by  `new ReadyMarkerFilter().withType(String type).withIdentifier(String identifier)`

The reasoning was to have a design is that it can be evolved/extended with further parameters without creating a mess with the string arguments in constructors. I'm really not sure it these two parameters will be the only ones needed in future.

maggu2810: > the `with...(...)`-Methods are not static

It seems I have forgotten how to read...

bundles/core/org.eclipse.smarthome.core.binding.xml/src/main/java/org/eclipse/smarthome/core/binding/xml/internal/XmlBindingInfoProvider.java
```diff
@@ -49,18 +54,27 @@
...
 
     @Activate
     public void activate(ComponentContext componentContext) {
         XmlDocumentReader<BindingInfoXmlResult> bindingInfoReader = new BindingInfoReader();
 
         bindingInfoTracker = new XmlDocumentBundleTracker<>(componentContext.getBundleContext(), XML_DIRECTORY,
-                bindingInfoReader, this, READY_MARKER);
-        bindingInfoTracker.open();
+                bindingInfoReader, this, READY_MARKER, readyService);
+        trackerJob = scheduler.schedule(() -> {
```
kaikreuzer: submit?

bundles/core/org.eclipse.smarthome.core.thing.xml/src/main/java/org/eclipse/smarthome/core/thing/xml/internal/XmlThingTypeProvider.java
```diff
@@ -56,18 +61,28 @@
...
     @Activate
     protected void activate(BundleContext bundleContext) {
         XmlDocumentReader<List<?>> thingTypeReader = new ThingDescriptionReader();
-
         thingTypeTracker = new XmlDocumentBundleTracker<List<?>>(bundleContext, XML_DIRECTORY, thingTypeReader, this,
-                READY_MARKER);
-        thingTypeTracker.open();
+                READY_MARKER, readyService);
+
+        trackerJob = scheduler.schedule(() -> {
```
kaikreuzer: submit?

bundles/core/org.eclipse.smarthome.core/src/main/java/org/eclipse/smarthome/core/service/ReadyService.java
```diff
@@ -0,0 +1,103 @@
...
+ * <p>
+ * Interested parties may register as a tracker for {@link ReadyMarker}s. Optionally they can provide a
+ * {@link ReadyMarkerFilter} in order to restrict the {@link ReadyMarker}s they get notified for.
+ * <p>
+ * Alternatively, {@link #isReady(ReadyMarker)} can be used to check for any given {@link ReadyMarker}.
+ *
+ * @author Simon Kaufmann - initial contribution and API.
+ *
+ */
+@NonNullByDefault
```
kaikreuzer: Did we already agree on using that annotation?

bundles/core/org.eclipse.smarthome.core/src/main/java/org/eclipse/smarthome/core/internal/service/ReadyServiceImpl.java
```diff
@@ -0,0 +1,119 @@
...
+                notifyTrackers(readyMarker, tracker -> tracker.onReadyMarkerRemoved(readyMarker));
+                logger.debug("Removed ready marker {}", readyMarker);
+            }
+        } finally {
+            rwlTrackers.readLock().unlock();
+        }
+    }
+
+    private void notifyTrackers(ReadyMarker readyMarker, Consumer<ReadyTracker> action) {
+        trackers.entrySet().stream().filter(entry -> {
```
amitjoy: Not a change request but a suggestion: 

```java
trackers.entrySet().stream().filter(entry -> entry.getValue().apply(readyMarker)).map(entry -> entry.getKey())
.forEach(action);
````

Comment (amitjoy) [reply bundles/core/org.eclipse.smarthome.core/src/main/java/org/eclipse/smarthome/core/internal/service/ReadyServiceImpl.java]: or

```java
trackers.entrySet().stream().filter(entry -> entry.getValue().apply(readyMarker)).map(Entry::getKey)
.forEach(action);
```

bundles/core/org.eclipse.smarthome.core/src/main/java/org/eclipse/smarthome/core/service/ReadyService.java
sjka: No idea, I lost the overview...

bundles/core/org.eclipse.smarthome.core/src/main/java/org/eclipse/smarthome/core/internal/service/ReadyServiceImpl.java
sjka: I've been using the block-notation on purpose, to force the auto-formatter to insert line-breaks in places where the human brain can follow. I couldn't figure out how to make JDT do do something similar automatically. Do you by any change have an idea?

amitjoy: Ah now I understand. What I usually do to achieve this is that I manually make changes to the stream notations and turn of autoformatter for certain section.

```java
// @formatter:off
trackers.entrySet().stream()
                    .filter(entry -> entry.getValue().apply(readyMarker))
                    .map(Entry::getKey)
                    .forEach(action);
// @formatter:on
```

bundles/core/org.eclipse.smarthome.core/src/main/java/org/eclipse/smarthome/core/service/ReadyService.java
kaikreuzer: Check [this](https://github.com/eclipse/smarthome/pull/4080#issuecomment-328115241) to get an overview again :-)

Issue #4179, opened 2017-09-03
Component XML generation broken for services with a configuration-pid


sihui62:
https://community.openhab.org/t/oh-2-2-0-snapshots-1025-1028-zwave-npe/33711/65

kaikreuzer:
The issue is that Felix Fileinstall writes this entry to the config file:
```
autoLinks = B"false"
```
I assume that the link service does not correctly handle boolean (instead of string) values.

maggu2810:
```Java
Object value = context.getProperties().get("autoLinks");
autoLinks = value == null || !value.toString().equals("false");
```

I assume there is a missing "to lower case" between `toString()` and `equals`

maggu2810:
I has been wrong. The code should be correct:

```Java
    public static void test(final Object value) {
        final boolean flag = value == null || !value.toString().equals("false");
        System.out.println(String.format("%s => %b", value, flag));
    }

    public static void main(final String args[]) {
        test("true");
        test("false");
        test(true);
        test(false);
        test(Boolean.TRUE);
        test(Boolean.FALSE);
    }
```

```text
true => true
false => false
true => true
false => false
true => true
false => false
```

kaikreuzer:
My suspicion was apparently wrong. It seems that the configuration is just fine.
If I call GET http://localhost:8080/rest/services/org.eclipse.smarthome.links/config I get
```
{
  "felix.fileinstall.filename": "file:/Users/kai/Downloads/oh/userdata/etc/org.eclipse.smarthome.links.cfg",
  "autoLinks": false
}
```
Also, the toggle in the Paper UI correctly is in "off" state:

<img width="851" alt="screen shot 2017-09-05 at 16 46 07" src="https://user-images.githubusercontent.com/3244965/30067145-48138c4c-925a-11e7-8394-a7567d53496a.png">

Nonetheless, the "Items" menu is not visible.
I therefore assume that it must be an issue in the Paper UI - @htreu Could you have a look?

kaikreuzer:
We have tracked down the issue to be a bug in the XML generation during the Maven build:
https://issues.apache.org/jira/browse/FELIX-5506

This effectively means that all our services that rely on a specific configuration-pid are currently broken. 😢

maggu2810:
:open_mouth:

maggu2810:
Should we add the `Critical` label?

kaikreuzer:
[...]

PR #4207, opened 2017-09-06
[LIFX] Add representation property and use DS annotations

wborn:
This PR improves the LIFX Binding by:

* Using the macAddress as representation-property
* Declaring components using DS annotations instead of XML files


maggu2810:
Perhaps we should wait until [this] one has been resolved.

[this]: https://github.com/eclipse/smarthome/issues/4179#issuecomment-327473607

Issue #1477, opened 2016-05-08
"Integers" sent in configuration now include scale


cdjackson:
If I send a configuration value of -2, it is now received as a BigDecimal of -2.0, which is not equal to -2.  This seems to have changed recently (in the last week or two) - prior to this it was received without the scale setting and the change broke a load of comparisons in ZWave.

Given that comparisons of BigDecimal checks the scale, it's annoying that integer config types are not marshalled without the scale.

Maybe this is unavoidable, in which case feel free to close, but I thought I'd raise it in case others also experience problems with this change.

kaikreuzer:
Can you confirm that the config value is declared as an integer and not as a decimal?
I would have thought that [this test](https://github.com/eclipse/smarthome/pull/1423/files#diff-5e38bc0bc2283f77cf75e3a4d2bb168aR52) covers your use case - I fully agree that if you compare it against -2, it should be true.
@sjka Could you have a look, please?

cdjackson:
This is the definition. The actual number is -232323 as you can see below. This has changed in the past week or so.

```
            <parameter name="controller_exclude" type="integer" groupName="actions">
                <label>Exclude Devices</label>
                <description>Puts the controller into Exclusion mode.</description>
                <advanced>true</advanced>
                <default>0</default>
                <options>
                    <option value="-232323">send</option>
                </options>
            </parameter>
```

maggu2810:
Don't know which commit and why the scale has been added, just about the comparison:
If the scale does not matter for the one that consumes the BigDecimal, I would prefer to use compareTo instead of equal.

maggu2810:
Don't know which commit and why the scale has been added, just about the comparison:
If the scale does not matter for the one that consumes the BigDecimal, I would prefer to use compareTo instead of equal. Just if this helps you in some way.
Last but not least, we should not add a scale.

kaikreuzer:
>  I would prefer to use compareTo instead of equal

That's a very valid point!

> Last but not least, we should not add a scale.

What exactly are the reasons against it? It seem to be a valuable piece of information.

maggu2810:
>> Last but not least, we should not add a scale.

> What exactly are the reasons against it? It seem to be a valuable piece of information.

I have meant "we should not increase the scale". If the user sets "2" we should not set the BigDecimal to "2.0". There is no reason to increase the scale.

kaikreuzer:
> There is no reason to increase the scale.

Yes, fully agree.

[...]

PR #4213, opened 2017-09-07
Magic: Add configurable thing + use DS annotations

triller-telekom:


bundles/test/org.eclipse.smarthome.magic/src/main/java/org/eclipse/smarthome/magic/binding/handler/MagicConfigurableThingHandler.java
```diff
@@ -0,0 +1,32 @@
...
+    public void handleCommand(@NonNull ChannelUID channelUID, Command command) {
+    }
+
+    @Override
+    public void initialize() {
+        updateStatus(ThingStatus.ONLINE);
+    }
+
+    @Override
+    public void handleConfigurationUpdate(Map<String, Object> configurationParameters) {
```
maggu2810: What's the reason to implement that function if only the super method is called?

triller-telekom: I used it to set a break point there while debugging #1477 :) Since the magic bundle is our "testing bundle", maybe its useful for others too.

maggu2810: Ah, okay for me

PR #4214, opened 2017-09-07
drop google collect from bindings

maggu2810:
The binding "digitalstrom" has been skipped because there is already a big fat PR in queue and we want to prevent merge conflicts.


maggu2810:
I also fixed the Hue binding test launch configuration so it can be started in the IDE again (and dropped a lot of bundles that should not be part of that launch configuration at all).

PR #4215, opened 2017-09-07
Fix https://github.com/eclipse/smarthome/issues/3912 in CronExpression

kgoderis:

### DIFF
--- a/bundles/core/org.eclipse.smarthome.core/src/main/java/org/eclipse/smarthome/core/scheduler/CronExpression.java
+++ b/bundles/core/org.eclipse.smarthome.core/src/main/java/org/eclipse/smarthome/core/scheduler/CronExpression.java
@@ -11,11 +11,13 @@ import java.text.ParseException;
 import java.util.ArrayList;
 import java.util.Calendar;
 import java.util.Date;
+import java.util.LinkedList;
 import java.util.List;
 import java.util.StringTokenizer;
 import java.util.TimeZone;
 
 import org.apache.commons.lang.StringUtils;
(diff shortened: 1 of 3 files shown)


bundles/core/org.eclipse.smarthome.core/src/main/java/org/eclipse/smarthome/core/scheduler/CronExpression.java
```diff
@@ -223,16 +225,29 @@ public static boolean isValidExpression(String cronExpression) {
...
+
+        if (yearsPart == null) {
+            List<@NonNull CronExpressionPart> ep = getExpressionParts();
+            List<@NonNull CronExpressionPart> parts = new LinkedList<@NonNull CronExpressionPart>();
+            parts.addAll(ep);
+
+            try {
+                parts.add(new YearsExpressionPart("*"));
+            } catch (ParseException e) {
+                throw new IllegalArgumentException("Year rule part can contain * as a token");
```
sjka: ~~can~~ -> cannot

kgoderis:
> How about a little regression test in CronExpressionTest?

I am not familiar with how to craft these kind of tests....

kaikreuzer:
@kgoderis We already have a test for this and you have removed the "Ignore" from it - so maybe it would suffice to also [remove the FIXME comment](https://github.com/eclipse/smarthome/pull/4215/files#diff-90461fad9ab4c7bd9de458243afdb5bcR81)?

kgoderis:
done.

PR #4216, opened 2017-09-08
Delete orphan configuration from ConfigAdmin

htreu:
This PR supersedes #3988.

Configuration files which are marked with `pid:<service.pid>` in the first line are considered _exclusive_ config files for the given PID. Modifications (insert line, remove line, delete file) will be reflected by configAdmin. Once an exclusive configuration is altered configAdmin will update the specific service with the new configuration.

In a second commit the JavaTest was extended to also provide an _afterLastCall_  callback for the _waitForAssert_ method. The o.e.sh.test bundle is now also a test bundle with a basic test of the new callback.


htreu:
https://github.com/eclipse/smarthome/pull/4216/commits/580e54be301048b3c964891f288b3ac3341f4b22 is not used anymore since the last commit introduced JUnits TemporaryFolder to the test. The test automatically cleans up all files after each test run. So the commit might as well be dropped.

kaikreuzer:

bundles/config/org.eclipse.smarthome.config.dispatch/META-INF/MANIFEST.MF
```diff
@@ -5,12 +5,13 @@ Bundle-RequiredExecutionEnvironment: JavaSE-1.8
...
- org.eclipse.smarthome.config.core,
- org.eclipse.smarthome.core.service,
- org.osgi.service.cm;version="1.5.0",
+Import-Package: com.google.gson,
+ com.google.gson.annotations,
+ org.apache.commons.io,
+ org.apache.commons.lang,
+ org.eclipse.smarthome.config.core,
+ org.eclipse.smarthome.core.service,
+ org.osgi.framework;version="1.5.0",
```
kaikreuzer: Please remove the version constraints (unless we have them?)
Also remove it from slf4j while you are at it :-)

bundles/config/org.eclipse.smarthome.config.dispatch/src/main/java/org/eclipse/smarthome/config/dispatch/internal/ConfigDispatcher.java
```diff
@@ -65,22 +74,14 @@
...
-        } else {
-            return ConfigConstants.getConfigFolder() + File.separator + SERVICES_FOLDER;
-        }
-    }
+    private final Logger logger = LoggerFactory.getLogger(ConfigDispatcher.class);
 
-    public ConfigDispatcher() {
-        super(getPathToWatch());
-    }
+    private final Gson GSON = new Gson();
```
kaikreuzer: Looks like a static field, if the name is in capitals...

bundles/config/org.eclipse.smarthome.config.dispatch/src/main/java/org/eclipse/smarthome/config/dispatch/internal/ConfigDispatcher.java
```diff
@@ -151,6 +163,62 @@ protected void processWatchEvent(WatchEvent<?> event, Kind<?> kind, Path path) {
...
+    }
+
+    private void loadExclusivePIDList() {
+        try (FileReader reader = new FileReader(exclusivePIDStore)) {
+            exclusivcePIDMap = GSON.fromJson(reader, ExclusivePIDMap.class);
+            if (exclusivcePIDMap != null) {
+                exclusivcePIDMap.initializeProcessPIDMapping();
+            }
+        } catch (JsonSyntaxException | JsonIOException | IOException e) {
+            logger.error("Error loading exclusive pids from {}.", exclusivePIDStore.getAbsolutePath(), e);
```
kaikreuzer: Don't we also end up here if the file does not exist (which is a perfectly valid case)?

bundles/config/org.eclipse.smarthome.config.dispatch/src/main/java/org/eclipse/smarthome/config/dispatch/internal/ConfigDispatcher.java
```diff
@@ -105,15 +106,26 @@ public ConfigDispatcher() {
 
     private static final String PID_MARKER = "pid:";
 
-    private final Logger logger = LoggerFactory.getLogger(ConfigDispatcher.class);
+    private static final String EXCLUSIVE_PID_STORE_FILE = "configdispatcher_pid_list.json";
+
+    private ExclusivePIDMap exclusivcePIDMap;
```
kaikreuzer: typo: exclusivce

bundles/config/org.eclipse.smarthome.config.dispatch/src/main/java/org/eclipse/smarthome/config/dispatch/internal/ConfigDispatcher.java
```diff
@@ -151,6 +163,62 @@ protected void processWatchEvent(WatchEvent<?> event, Kind<?> kind, Path path) {
...
+        try (FileWriter writer = new FileWriter(exclusivePIDStore)) {
+            exclusivcePIDMap.setCurrentExclusivePIDList();
+            GSON.toJson(exclusivcePIDMap, writer);
+        } catch (JsonIOException | IOException e) {
+            logger.error("Error storing exclusive PID list in bundle data file.", e);
+        }
+    }
+
+    private void processOrphanExclusivePIDs() {
+        for (String orpahnPID : exclusivcePIDMap.getOrpahnPIDs()) {
```
kaikreuzer: typo: orpahnPID

bundles/config/org.eclipse.smarthome.config.dispatch/src/main/java/org/eclipse/smarthome/config/dispatch/internal/ConfigDispatcher.java
```diff
@@ -273,26 +359,126 @@ private void processConfigFile(File configFile) throws IOException, FileNotFound
...
+                processedPIDMapping.put(pid, null);
+            }
+        }
+
+        /**
+         * Collect PIDs which where not processed (mapped path is null).
+         *
+         * @return the list of PIDs which where not processed either during #activate or on file deleted event.
+         */
+        public List<String> getOrpahnPIDs() {
```
kaikreuzer: typo: Orpahn

bundles/config/org.eclipse.smarthome.config.dispatch/src/main/java/org/eclipse/smarthome/config/dispatch/internal/ConfigDispatcher.java
```diff
@@ -151,6 +163,62 @@ protected void processWatchEvent(WatchEvent<?> event, Kind<?> kind, Path path) {
...
+        } catch (JsonIOException | IOException e) {
+            logger.error("Error storing exclusive PID list in bundle data file.", e);
+        }
+    }
+
+    private void processOrphanExclusivePIDs() {
+        for (String orpahnPID : exclusivcePIDMap.getOrpahnPIDs()) {
+            try {
+                Configuration configuration = configAdmin.getConfiguration(orpahnPID, null);
+                configuration.delete();
```
kaikreuzer: Is this a "silent" operation, if the configuration didn't exist at all?

bundles/config/org.eclipse.smarthome.config.dispatch/src/main/java/org/eclipse/smarthome/config/dispatch/internal/ConfigDispatcher.java
```diff
@@ -151,6 +163,62 @@ protected void processWatchEvent(WatchEvent<?> event, Kind<?> kind, Path path) {
...
+        } else if (kind == ENTRY_DELETE) {
+            // Detect if a service specific configuration file was removed. We want to
+            // notify the service in this case with an updated empty configuration.
+            File configFile = path.toFile();
+            if (configFile.isHidden() || configFile.isDirectory() || !configFile.getName().endsWith(".cfg")) {
+                return;
+            }
+
+            exclusivcePIDMap.setFileRemoved(configFile.getAbsolutePath());
+            processOrphanExclusivePIDs();
```
kaikreuzer: Shouldn't you also update the list on CREATE and MODIFIED events?

bundles/config/org.eclipse.smarthome.config.dispatch/META-INF/MANIFEST.MF
htreu: to be honest, I dont know about this constraint. I was fooled by the constraint on `org.osgi.service.cm;version="1.5.0"` and left it on `org.osgi.framework`, too. Will remove it.

bundles/config/org.eclipse.smarthome.config.dispatch/src/main/java/org/eclipse/smarthome/config/dispatch/internal/ConfigDispatcher.java
htreu: My impression was (just as for the logger) that the instance(s) from ConfigDispatcher are controlled by the OSGi framework and we dont give it static references to help unload/load the bundle during runtime. Is this not the case here?

htreu: In case I´m right I could change the name to gson just as for the logger ;-)

kaikreuzer: It is. But then you should comply to the Java naming conventions and not use capitals for that field.

bundles/config/org.eclipse.smarthome.config.dispatch/src/main/java/org/eclipse/smarthome/config/dispatch/internal/ConfigDispatcher.java
htreu: yes, I will separate the IOException and only log to debug.

bundles/config/org.eclipse.smarthome.config.dispatch/src/main/java/org/eclipse/smarthome/config/dispatch/internal/ConfigDispatcher.java
htreu: In line 207 the configuration object is created if none existed before. So with regards to object references this will always proceed. I´m not sure what "didn't exist at all" could otherwise mean in OSGi context.

bundles/config/org.eclipse.smarthome.config.dispatch/src/main/java/org/eclipse/smarthome/config/dispatch/internal/ConfigDispatcher.java
htreu: absolutely. when the file is created during runtime and then removed when the system is shutdown the configuration will also remain in the configAdmin store.

bundles/config/org.eclipse.smarthome.config.dispatch/src/main/java/org/eclipse/smarthome/config/dispatch/internal/ConfigDispatcher.java
kaikreuzer: I meant if we are sure that we are not causing an "config CREATED" and "config DELETED" event on ConfigAdmin for no reason.

bundles/config/org.eclipse.smarthome.config.dispatch/src/main/java/org/eclipse/smarthome/config/dispatch/internal/ConfigDispatcher.java
kaikreuzer: So why didn't you adapt the code then?

bundles/config/org.eclipse.smarthome.config.dispatch/src/main/java/org/eclipse/smarthome/config/dispatch/internal/ConfigDispatcher.java
```diff
@@ -171,44 +181,47 @@ protected void processWatchEvent(WatchEvent<?> event, Kind<?> kind, Path path) {
...
-            if (exclusivcePIDMap != null) {
-                exclusivcePIDMap.initializeProcessPIDMapping();
+            exclusivePIDMap = gson.fromJson(reader, ExclusivePIDMap.class);
+            if (exclusivePIDMap != null) {
+                exclusivePIDMap.initializeProcessPIDMapping();
             }
-        } catch (JsonSyntaxException | JsonIOException | IOException e) {
-            logger.error("Error loading exclusive pids from {}.", exclusivePIDStore.getAbsolutePath(), e);
+        } catch (JsonSyntaxException | JsonIOException e) {
+            logger.error("Error parsing exclusive pids from {}.", exclusivePIDStore.getAbsolutePath(), e.getMessage());
```
kaikreuzer: 2 params, but only one placeholder...

bundles/config/org.eclipse.smarthome.config.dispatch/src/main/java/org/eclipse/smarthome/config/dispatch/internal/ConfigDispatcher.java
```diff
@@ -171,44 +181,47 @@ protected void processWatchEvent(WatchEvent<?> event, Kind<?> kind, Path path) {
...
+        } catch (IOException e) {
+            logger.debug("Error loading exclusive pids from {}.", exclusivePIDStore.getAbsolutePath(), e.getMessage());
```
kaikreuzer: 2 params, but only one placeholder...

PR #4219, opened 2017-09-08
Extended the zoom channel category to also allow Number and Dimmer

sjka:
I just stumbled upon the ZOOM channel category, which is listed as a String type. Maybe I missed something, but wouldn't it rather make sense to be of Numeric type? A quick search on Github didn't bring up an existing binding using it. Is there any particular reason why it needs to be a String? 

### DIFF
--- a/docs/_data/categories_channel.csv
+++ b/docs/_data/categories_channel.csv
@@ -32,4 +32,4 @@ Temperature,"R, RW",Number,temperature.png
 Water,R,"Switch, Number",water.png
 Wind,R,Number,wind.png
 Window,"R, RW","String, Switch",window.png
-Zoom,RW,String,zoom.png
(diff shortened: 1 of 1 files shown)


kaikreuzer:
I think this is the general question of whether we want to cover functionality that is not part of the core item types through numbers (enumerations) or strings. The idea for zoom imho was to allow commands like "zoomin", "zoomout", maybe even "pan", "tilt" or whatever. Using such commands in textual rules is more user friendly than having to use 1,2,3.

Apart from that, a "normal" zoom function could also be modelled as a Dimmer item, where increase/decrease are good fits and you can adjust it between 0-100%.

So maybe the best change would be to allow all those options (string, number, dimmer)? After all, it is only meant to get a rough idea about the channel.

sjka:
okay, fine by me too.

PR #4225, opened 2017-09-08
Renamed a few packages to internal in the org.eclipse.smarthome.io.re…

VelinYordanov:
…st.core bundle.

### DIFF
--- a/bundles/io/org.eclipse.smarthome.io.rest.core.test/src/test/groovy/org/eclipse/smarthome/io/rest/core/internal/discovery/InboxResourceOSGITest.groovy
+++ b/bundles/io/org.eclipse.smarthome.io.rest.core.test/src/test/groovy/org/eclipse/smarthome/io/rest/core/internal/discovery/InboxResourceOSGITest.groovy
@@ -5,7 +5,7 @@
  * which accompanies this distribution, and is available at
  * http://www.eclipse.org/legal/epl-v10.html
  */
-package org.eclipse.smarthome.io.rest.core.discovery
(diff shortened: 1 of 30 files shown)


maggu2810:
I myself imported stuff of the REST bundle some time ago because I would like to reuse existing code. But this was not about the resources but the DTO and the mappers (and one service).
But also about this ones, I stop doing it.

About the REST resources:
I don't know if it make sense to reuse the REST resources at all.
I agree we should use the same approach for all RESTResources.
IMHO we could make them internal

kaikreuzer:
I'd agree - let's keep the RESTResources private, while exporting the DTOs and mappers.

@VelinYordanov Would you be able to refactor the code accordingly?

VelinYordanov:
@kaikreuzer Would that suffice?

kaikreuzer:
Yes, thanks, I think this looks good as it is (@maggu2810, do you agree?).
So all there is left to do is to adapt the unit tests accordingly.

kaikreuzer:
We already have this: The SitemapResource uses the ItemDTOs (and yes, potential side effects are expected and wanted ;-)).

maggu2810:
My comment has been about non-ESH bundles that reuse the DTOs -- so ones that are not under our control.

VelinYordanov:
@kaikreuzer The tests should be ok now.

sjka:
Would have been nice if _all_ `<implementation...` tags were adapted in the XMLs. Now many resources are broken 😒

PR #4229, opened 2017-09-09
Sonos: play TuneIn by a station id

lolodomo:
Channel table fully updated in the documentation with the access mode added for each channel


lolodomo:
@kgoderis : could you please review this change ?

sjka:
Just tried it out - works like a charm for me 👍

extensions/binding/org.eclipse.smarthome.binding.sonos/src/main/java/org/eclipse/smarthome/binding/sonos/handler/ZonePlayerHandler.java
```diff
@@ -132,6 +134,8 @@
 
     private Map<String, String> stateMap = Collections.synchronizedMap(new HashMap<String, String>());
 
+    private List<SonosMusicService> msuicServices;
```
wborn: I don't own or use any Sonos devices. But "msuicServices"  looks like a typo. :-)

kaikreuzer: Good catch :-)
Ah, sorry, don't know where my head is these days... I somehow mixed up Sonos with LIFX... So forget about my testing request unless you can make your bulbs play some music 🤣

lolodomo: lol

lolodomo:
Typo fixed + everything squashed.

Issue #3913, opened 2017-07-30
[Lifx] direct IP address support


paulianttila:
Currently lifx addon implementation is based on lifx discovery protocol, which works fine when bulbs and server is running on the same network. For security reason, my lifx bulbs are in dedicated VLAN and server (OH) is running on different VLAN. Broadcast communication is not allowed between different VLANs, so lifx addon can't find bulbs and therefore I can't control them. 

Currently only deviceId (mac address) can be manually configured to addon, and if I have understand correctly, addon then finds the IP address of the bulb by using UDP broadcast communication. 

My request is to enhance the addon and make possible to configure IP address to bulb configuration, which addon could then use directly.

htreu:
Hey @wborn, is this something you could have a look at? Thanks in advance :-)

wborn:
@paulianttila With #4231 you can now optionally configure the host of the light. I've succesfully tested controlling lights from another machine on another location via a VPN and Docker in bridge mode when UDP broadcasts don't work. :-)

paulianttila:
@wborn, just tested your changes and host parameter seems to works great. Now I can control bulbs in my vlan environment without python script. Big thanks to you 👍

PR #4231, opened 2017-09-09
[LIFX] Add optional host configuration parameter, support new products and improvements

wborn:
* Add host configuration parameter for lights that do not respond to UDP broadcasts (fixes https://github.com/eclipse/smarthome/issues/3913)
* Add support for new [products](https://lan.developer.lifx.com/docs/lifx-products) (most notably the new [Downlights](https://community.lifx.com/t/you-guys-working-on-a-new-gu10/2519/9?u=wborn))
* Add deviceId configuration parameter pattern to prevent mistakes/exceptions
* Fix static code analysis findings
* Makes the LIFX Binding periodically update network interface information. This for instance resolves the issue that network interfaces are not (or still) being used for broadcasts after they go up (or down).
* The code for this information was duplicated across the LifxLightDiscovery and LifxLightCommunicationHandler classes and has now been extracted into LifxNetworkUtil.
* All utility classes have now been grouped in the .util package.
* Update Thing status to configuration error when Device ID and Host are null
* Introduce logId property so logging identifies light depending on MAC/IP configuration
* Add logId where missing to simplify debugging
* Use lambdas for listener registration and iteratating over listeners
* Resolve @NonNull warnings
* Fix properties not always being updated
* Use a LifxLightContext for sharing common variables between handler helper objects
* Improve exception handling by catching specific exceptions and using more specific error messages


extensions/binding/org.eclipse.smarthome.binding.lifx/ESH-INF/config/config.xml
```diff
@@ -4,8 +4,14 @@
     xsi:schemaLocation="http://eclipse.org/smarthome/schemas/config-description/v1.0.0 http://eclipse.org/smarthome/schemas/config-description-1.0.0.xsd">
 
     <config-description uri="thing-type:lifx:light">
-        <parameter name="deviceId" type="text" required="true">
+        <parameter name="deviceId" type="text" pattern="([0-9A-Fa-f]{12})" required="true">
             <label>LIFX device ID</label>
+            <description>Identifies the light e.g. "D073D5A1A1A1"</description>
+        </parameter>
+        <parameter name="host" type="text" required="false">
```
kaikreuzer: I am not clear on how the deviceId and the host parameter relate to each other. Which one is required for the handler in order to identify and communicate with a certain bulb? One of both? If so, what happens if both values are set (as in your example in the README), but do not correspond to the same light?

wborn: The binding always uses the deviceId to identify the light. The host parameter is only required for communications when the binding can not automatically discover the light using UDP broadcasts.

> If so, what happens if both values are set (as in your example in the README), but do not correspond to the same light?

The binding [ignores](https://github.com/wborn/smarthome/blob/37a425bc68fa3abb9a3ac0b6423d1dedd4258e94/extensions/binding/org.eclipse.smarthome.binding.lifx/src/main/java/org/eclipse/smarthome/binding/lifx/internal/LifxLightCommunicationHandler.java#L332) incoming packets that do not match the light MACAddress (deviceId). So in that scenario the light will remain offline. The binding will still send packets resulting from commands to the light.

Maybe we should add a helpful status description for this scenario? It should be possible to detect this.

kaikreuzer: > The binding ignores incoming packets that do not match the light MACAddress (deviceId).

Do we know from which IP the packets originate from? If so, are there any real cases where the packet is NOT from the bulb at that IP and thus must be ignored? I know that in early times, LIFX bulbs were meant to act as a router to 6LoWPAN networks, but afaik this isn't true anymore and thus every bulb should have a unique IP - so if this is the case, the IP would indeed be the better configuration parameter (while the mac could be fully omitted).

wborn: > Do we know from which IP the packets originate from? If so, are there any real cases where the packet is NOT from the bulb at that IP and thus must be ignored?

The IP of the light from which the packet originates is always known. Using just the IP as a configuration parameter only makes sense when people have a static IP configured for their lights. Otherwise the broadcast is needed to find the light and match it with the MAC.

> LIFX bulbs were meant to act as a router to 6LoWPAN networks, but afaik this isn't true anymore and thus every bulb should have a unique IP

Lights no longer act as routers for other lights. It was not stable and thus removed. At the same time the V2 LAN protocol was introduced.

> while the mac could be fully omitted

In theory the MAC could be removed as a configuration parameter for those people who run their lights with static IPs. The only issue with that is that most of the current code assumes the MAC is readily available and uses it for instance with logging and locking.

Another issue could be that the user should then configure either the MAC or the IP. Is it possible to define such a constraint within the `<config-description>`?

kaikreuzer: > Another issue could be that the user should then configure either the MAC or the IP. Is it possible to define such a constraint within the <config-description>?

I think giving the choice (and not always requiring the MAC) would be the best option. Unfortunately, there isn't such a constraint for config descriptions. Probably the easiest way would be to declare both as optional and set the Thing status to CONFIGURATION_ERROR, if none of the two parameters is provided. Wdyt?

wborn: > set the Thing status to CONFIGURATION_ERROR

Yes that would solve this issue.

After some more thought I think there will be other issues when there will be 2 properties (MAC or IP) that identify a Thing.

E.g. the `<representation-property>` can no longer be set to the MAC. The DiscoveryService will also not be able to tell if a Thing has a static or dynamic IP and thus which property should identify the Thing.

kaikreuzer: > E.g. the <representation-property> can no longer be set to the MAC.

Why not? The MAC is the invariable identifier, so even if you find the bulb at 5 IP addresses, you can notice that it is the very same. So the MAC should stay as the representation-property.

> DiscoveryService will also not be able to tell if a Thing has a static or dynamic IP and thus which property should identify the Thing.

The discovery can decide, which config parameter to use - and if it easily gets hold of the MAC, it should use it. Where is the problem?

wborn: > Where is the problem?

Wouldn't discovery then create new discovery results (with MAC configuration parameters) when you have added all your lights using only IP configuration parameters? 

A workaround could be to automatically update an empty MAC configuration parameter when a packet is received from the light.

kaikreuzer: Hm, good point. I was actually thinking that the MAC would be added as a property and thus the "representation-property" info would do the trick, but you are right that this isn't possible as the MAC is already a config param. So yes, setting this param by the handler might solve the issue.

wborn:
With the latest commit the deviceId (MAC) is now an optional property for lights @kaikreuzer. Since a lot of code (logging, throttling) was depending on the MAC, there are quite some code changes. I also made some additional improvements while I was at it.

PR #4233, opened 2017-09-09
Fix NPE when EnrichedItemDTOMapper considers transforming null Strings

wborn:
This NPE occurred when showing the Thing Control page in Paper UI while testing the new Nest Binding. The Nest Binding had [updated](https://github.com/openhab/openhab2-addons/blob/1987aecc480cb48700ca0fb6584a0ba3327a1e6e/addons/binding/org.openhab.binding.nest/src/main/java/org/openhab/binding/nest/handler/NestCameraHandler.java#L75) the public camera URL with a `new StringType(null)` because public sharing was not enabled for the respective camera.

As a result `/rest/items?recursive=false` returned a 500 response code and the whole Paper UI Control page stopped working.

Perhaps the Nest Binding should not post such updates. But ESH should be able to handle such item states when it does not immediately reject such updates.

With this fix such items do not have a "state" property in the resulting JSON and Paper UI shows "undefined" for such states.


kaikreuzer:
> Perhaps the Nest Binding should not post such updates. But ESH should be able to handle such item states when it does not immediately reject such updates.

I agree with both statements - thanks for the fix!

PR #4234, opened 2017-09-10
Core/config: Add NestedConfigConverter

davidgraeff:
One step towards #4178.

The added class converts between a flat map of <key,value> tuples to a nested map and back.
`key`s in the flat map express a hierarchy by the hash sign and dots. The schema is:

key.instanceName#key.instanceName#...#key=value

The `instanceName` can be anything but need to be unique among all instances at the same
hierarchy level for a given key.
The `key` part identifies the field name for the nested structure.

The pairs of `key`.`instanceName`, separated by a dot, that form a hierarchy are separated by dashes.

## Example
A flat map looks like this:

```
autostart=true
broker.broker1#host=123.123.123.123
broker.broker1#secure=true
broker.broker2#host=124.124.124.124
broker.broker2#secure=true
```

and will be converted to:

```
autostart=true
broker
   broker1
     host=123.123.123.123
     secure=true
   broker2
     host=124.124.124.124
     secure=true
```

This class is useful in conjunction with {@link Configuration} and its
(un-)marshaling functionality.


triller-telekom:
Since we now have #4685 and #4847 I think we can close this PR because the aforementioned PRs supersede this one by implementing the OSGi service factory pattern.

@kaikreuzer @sjka

davidgraeff:
OSGi service factory support should do it

Issue #4140, opened 2017-08-27
Dynamic changes to the channel type parameters


ppieczul:
## Introduction

The goal for this topic is to agree on the changes required in the framework in order to support dynamic changes to some of the channel type characteristics. An examples are: value to text mappings (options) and format string, which are provided as part of the channel type definition, either in xml or through a channel type provider programatically. Also, we can think of extending it to other channel type parameters, i.e. min/max/step values and read-only attribute, although I don't have a real life use case for them today.

Current design seems to assume that channel types are to be available when thing becomes online and they should not change during its lifetime. For example, Paper UI is designed around this assumption and is able to refresh channel changes dynamically, but not channel types.

In some circumstances it is needed to at least update the definition of channels types, based on the information received in the runtime from a remote device under binding supervision. One example is Loxone binding, which can receive from the Miniserver an update with a new definition of labels for lighting controllers, while the device remains connected and operating all the time. It is a change in the device configuration that we need to reflect in the openHAB.

Before we jump to the design of what needs to be done, I'd like to agree on the assumptions for the changes. Here is how I see this situation.

## End user use cases

I think that there is basically one requirement that affects what end user will see:

  * User should see the dynamic changes to the available channel type options and format immediately as they happen on the remote device (thing). They should become visible in the UIs and rest API, providing that the user has not provided custom configuration of these parameters through configuration files. There should be no interruption to the thing operation, for example thing should not go offline and online again for these changes to be visible.

@Kai, for now I would like to review and agree on the assumptions, the use cases and set of parameters that we should be allowing for dynamic modifications, so that we are on the same page wrt requirements for the change.

Thank you
Pawel

kaikreuzer:
FTR: This discussion was started during the Loxone bindin review here: https://github.com/openhab/openhab2-addons/pull/2423#discussion_r132446174

My suggestion there was not to allow channel types to be dynamic, but rather to provide a way to dynamically contribute to a StateDescription from within a binding (currently this is only static through the `ChannelStateDescriptionProvider`, which derives its input from the channel type.

davidgraeff:
I thing, I had the same problem (assuming that I understood the problem) in the OpenHAB Yamaha binding. The device returned a list of supported input modes, and I needed to create a channel type dynamically with the given modes as state options:

[...]

PR #4238, opened 2017-09-10
Add dynamic state description provider (#4140)

ppieczul:
Please see #4140 for the context.
Work in progress.

Signed-off-by: Pawel Pieczul <pieczul@gmail.com> (github: ppieczul)

### DIFF
--- a/bundles/core/org.eclipse.smarthome.core.thing/src/main/java/org/eclipse/smarthome/core/thing/internal/ChannelStateDescriptionProvider.java
+++ b/bundles/core/org.eclipse.smarthome.core.thing/src/main/java/org/eclipse/smarthome/core/thing/internal/ChannelStateDescriptionProvider.java
@@ -72,10 +72,7 @@ public class ChannelStateDescriptionProvider implements StateDescriptionProvider
             ChannelUID channelUID = boundChannels.iterator().next();
             Channel channel = thingRegistry.getChannel(channelUID);
             if (channel != null) {
-                StateDescription stateDescription = getDynamicStateDescription(channel, locale);
-                if (stateDescription != null) {
-                    return stateDescription;
-                }
+                StateDescription stateDescription = null;
                 ChannelType channelType = thingTypeRegistry.getChannelType(channel, locale);
                 if (channelType != null) {
                     stateDescription = channelType.getState();
@@ -98,16 +95,22 @@ public class ChannelStateDescriptionProvider implements StateDescriptionProvider
                             }
                         }
                     }
-                    return stateDescription;
                 }
+                StateDescription dynamicStateDescription = getDynamicStateDescription(channel, stateDescription,
+                        locale);
+                if (dynamicStateDescription != null) {
+                    return dynamicStateDescription;
+                }
+                return stateDescription;
             }
         }
         return null;
     }
 
-    private StateDescription getDynamicStateDescription(Channel channel, Locale locale) {
+    private StateDescription getDynamicStateDescription(Channel channel, StateDescription originalStateDescription,
+            Locale locale) {
         for (DynamicStateDescriptionProvider provider : dynamicStateDescriptionProviders) {
-            StateDescription stateDescription = provider.getStateDescription(channel, locale);
+            StateDescription stateDescription = provider.getStateDescription(channel, originalStateDescription, locale);
             if (stateDescription != null) {
                 return stateDescription;
             }
--- a/bundles/core/org.eclipse.smarthome.core.thing/src/main/java/org/eclipse/smarthome/core/thing/internal/ChannelStateDescriptionProvider.java
+++ b/bundles/core/org.eclipse.smarthome.core.thing/src/main/java/org/eclipse/smarthome/core/thing/internal/ChannelStateDescriptionProvider.java
@@ -72,12 +72,13 @@ public class ChannelStateDescriptionProvider implements StateDescriptionProvider
             ChannelUID channelUID = boundChannels.iterator().next();
             Channel channel = thingRegistry.getChannel(channelUID);
             if (channel != null) {
+                StateDescription stateDescription = getDynamicStateDescription(channel, locale);
+                if (stateDescription != null) {
+                    return stateDescription;
+                }
                 ChannelType channelType = thingTypeRegistry.getChannelType(channel, locale);
                 if (channelType != null) {
-                    StateDescription stateDescription = getDynamicStateDescription(channel, locale);
-                    if (stateDescription == null) {
-                        stateDescription = channelType.getState();
-                    }
+                    stateDescription = channelType.getState();
                     if ((channelType.getItemType() != null)
                             && ((stateDescription == null) || (stateDescription.getPattern() == null))) {
                         String pattern = null;
@@ -98,8 +99,6 @@ public class ChannelStateDescriptionProvider implements StateDescriptionProvider
                         }
                     }
                     return stateDescription;
-                } else {
-                    return null;
                 }
             }
         }
--- a/bundles/core/org.eclipse.smarthome.core.thing/src/main/java/org/eclipse/smarthome/core/thing/internal/ChannelStateDescriptionProvider.java
+++ b/bundles/core/org.eclipse.smarthome.core.thing/src/main/java/org/eclipse/smarthome/core/thing/internal/ChannelStateDescriptionProvider.java
@@ -7,9 +7,11 @@
  */
 package org.eclipse.smarthome.core.thing.internal;
 
+import java.util.List;
 import java.util.Locale;
 import java.util.Map;
 import java.util.Set;
+import java.util.concurrent.CopyOnWriteArrayList;
 
 import org.eclipse.smarthome.core.items.Item;
 import org.eclipse.smarthome.core.library.CoreItemFactory;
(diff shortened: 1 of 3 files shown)


bundles/core/org.eclipse.smarthome.core.thing/src/main/java/org/eclipse/smarthome/core/thing/type/DynamicStateDescriptionRegistry.java
```diff
@@ -0,0 +1,70 @@
...
+import org.osgi.service.component.annotations.ReferencePolicy;
+
+/**
+ * The {@link DynamicStateDescriptionRegistry} tracks all {@link StateDescription}s provided by registered
+ * {@link DynamicStateDescriptionProvider}s.
+ *
+ * @author Pawel Pieczul - Initial contribution
+ *
+ */
+@Component(immediate = true, service = DynamicStateDescriptionRegistry.class)
```
triller-telekom: if you implement your interface you can omit the `service=` part.

bundles/core/org.eclipse.smarthome.core.thing/src/main/java/org/eclipse/smarthome/core/thing/type/DynamicStateDescriptionRegistry.java
```diff
@@ -0,0 +1,70 @@
...
+     *
+     * @param channelUID channel UID
+     * @param locale (can be null)
+     * @return state description or null if no state description provided
+     */
+    public StateDescription getStateDescription(ChannelUID channelUID, Locale locale) {
+        for (DynamicStateDescriptionProvider provider : dynamicStateDescriptionProviders) {
+            StateDescription stateDescription = provider.getStateDescription(channelUID, locale);
+            if (stateDescription != null) {
+                return stateDescription;
```
triller-telekom: This will use the first matching one, is that what you intend to do? What if there are multiple providers that return a description?

ppieczul: I can think of a practical use case with single provide for a particular channel. What could be a case for more than one provider? We could try to merge the descriptions, but I just don't see a reason to do that.

ppieczul:
@triller-telekom thanks for your comments.
I'd like to get this provider done and I have the following questions:
  * Is the concept generally ok?
  * How can I compile a single smarthome project using mvn - compilation from a sub-directory fails as well as running top level mvn with --projects option. It it known to fail or should I nail it down?
  * What is the best way to run the tests, so I can add my new ones? I could not get it working under Eclipse (junit4 missing) and also from mvn (as I am getting the error to select a single project)
  * Are there any other parts of the framework I am missing but I should modify adding this provider?

Thanks
Pawel

bundles/core/org.eclipse.smarthome.core.thing/src/main/java/org/eclipse/smarthome/core/thing/internal/Activator.java
```diff
@@ -11,6 +11,7 @@
 import org.eclipse.smarthome.core.common.osgi.ServiceBinder.Bind;
 import org.eclipse.smarthome.core.common.osgi.ServiceBinder.Unbind;
 import org.eclipse.smarthome.core.thing.type.ChannelTypeRegistry;
+import org.eclipse.smarthome.core.thing.type.DynamicStateDescriptionRegistry;
```
triller-telekom: Do you really need the code in this `Activator` class?

Instead the `ChannelStateDescriptionProvider` could get a service reference
```
@Reference
protected void setDynamicStateDescriptionRegistry(DynamicStateDescriptionRegistry dynamicStateDescriptionRegistry) {
        this.dynamicStateDescriptionRegistry = dynamicStateDescriptionRegistry;
}
```

bundles/core/org.eclipse.smarthome.core.thing/src/main/java/org/eclipse/smarthome/core/thing/internal/ChannelStateDescriptionProvider.java
```diff
@@ -68,7 +68,10 @@ public StateDescription getStateDescription(String itemName, Locale locale) {
             if (channel != null) {
                 ChannelType channelType = thingTypeRegistry.getChannelType(channel, locale);
                 if (channelType != null) {
-                    StateDescription stateDescription = channelType.getState();
+                    StateDescription stateDescription = thingTypeRegistry.getStateDescription(channel, locale);
```
triller-telekom: if you use the service reference mentioned above you should be able to query your `DynamicStateDescriptionRegistry` directly and not through the `ThingTypeRegistry`.

bundles/core/org.eclipse.smarthome.core.thing/src/main/java/org/eclipse/smarthome/core/thing/type/DynamicStateDescriptionRegistry.java
triller-telekom: I see the problem that merging might be difficult, but when I was reading the code I just thought that in theory you could have multiple providers for the same channel and if you return the result from the first one, if this is what you really intend to do. If it is, then all is good.

bundles/core/org.eclipse.smarthome.core.thing/src/main/java/org/eclipse/smarthome/core/thing/type/ThingTypeRegistry.java
```diff
@@ -161,6 +162,21 @@ public ChannelType getChannelType(Channel channel, Locale locale) {
...
     }
 
+    /**
+     * Returns the state description for a given channel, if provided by a {@link DynamicStateDescriptionProvider}
+     *
+     * @param channel channel
+     * @param locale locale (can be null)
+     * @return state description or null if none provided for this channel
+     */
+    public StateDescription getStateDescription(Channel channel, Locale locale) {
```
triller-telekom: Please remove this method because you do not need to go through the `ThingTypeRegistry` but you can call your `DynamicStateDescriptionRegistry` directly.

kaikreuzer:
@triller-telekom Instead of looking at specific code places here, I think it would be worthwhile to first discuss the [big picture of @ppieczul](https://github.com/eclipse/smarthome/issues/4140#issuecomment-328304811) on the issue itself.

ppieczul:
@triller-telekom The code looks much simpler now. Is this looking better?

ppieczul:
@triller-telekom I would like to extend the tests and followed your advice above.
On any junit plugin test run attempt, including the launch configuration, I am getting:

`Required plug-in 'org.eclipse.jdt.junit4.runtime' could not be found.`

Any obvious to me attempts to fix it are not successful and Internet does not advise much.
 
Does it sound familiar to you and any quick hint what am I missing?

ppieczul:
I managed to progress using maven only and my understanding of how this works improved a bit. 
Thanks for your help.
I suppose the current version is what may be final. Looking forward to some feedback.

ppieczul:
@kaikreuzer @triller-telekom could you kindly take a look at this small code now, it may be all there needs to be for this feature. Thanks

bundles/core/org.eclipse.smarthome.core.thing/src/main/java/org/eclipse/smarthome/core/thing/internal/ChannelStateDescriptionProvider.java
```diff
@@ -68,7 +74,10 @@ public StateDescription getStateDescription(String itemName, Locale locale) {
             if (channel != null) {
                 ChannelType channelType = thingTypeRegistry.getChannelType(channel, locale);
                 if (channelType != null) {
-                    StateDescription stateDescription = channelType.getState();
+                    StateDescription stateDescription = getDynamicStateDescription(channel, locale);
```
sjka: In the way you implemented it here it will only work if there is a channelType defined. Keep in mind: there are bindings which work perfectly well without channel types. With this PR they actually should get the chance to also provide state descriptions for channels now, so this will be a big improvement. You will just have to move some of the logic out of the `if (channelType != null) {` block.

ppieczul: @sjka: I reshuffled the logic. An idea came to my mind in the meantime, that maybe we could pass the original state description as an argument to this provider, so it can return it with modifications of only some parameters. I don't have a use case, but it would be more broad and also work for original case. What do you think?

sjka: Excellent idea! 
I already had some doubts whether we maybe should try merging both results, but that all felt a bit awkward. Your suggestion of handing over the "static" description (if it exists) makes much more sense, because then the dynamic provide can decide to accept all the attributes it is not interested in.

ppieczul:
@sjka I added the original state description as an argument to the provider and added a test. Can you please take a look now? Thanks

sjka:
Awesome, thanks!

PR #4239, opened 2017-09-10
Configuration write back service 

davidgraeff:
One step closer to #4178.

Configuration write back service for services that use exclusive pid files and declare support via property.

## Abstract
Some services would need to have changed or extended configuration written back to the according cfg file (or a new cfg file being generated if not existing). Those services should declare a special property in their DS, which also requires the cfg file to be an exclusive cfg file.

Use-case: The SSL Service generates a Private/Public-KeyPair on start and writes the passwords back to the configuration. Those passwords are necessary for the user  to obtain to 1) backup the generated SSL configuration, 2) for the service to open the certificate file on next start.

A ConfigWriteBack service observes the ConfigAdmin via ConfigurationEvent events and write changes of affected configurations back to the cfg file, in the same file format that ConfigDispatcher can read (might be a nested configuration in future).

This is similar to Apache FileInstall felix.fileinstall.enableConfigSave=true, and provides the missing parts next to ConfigDispatcher for true exclusive cfg file support.


bundles/config/org.eclipse.smarthome.config.writeback.test/about.html
```diff
@@ -0,0 +1,28 @@
+<!DOCTYPE html PUBLIC "-//W3C//DTD XHTML 1.0 Strict//EN"
```
htreu: about.html is obsolete since EPL-2.0 was introduced.

bundles/config/org.eclipse.smarthome.config.writeback.test/build.properties
```diff
@@ -0,0 +1,6 @@
+bin.includes = .,\
+               META-INF/,\
+               OSGI-INF/,\
+               about.html
```
htreu: remove about.html

bundles/config/org.eclipse.smarthome.config.writeback/about.html
```diff
@@ -0,0 +1,28 @@
+<!DOCTYPE html PUBLIC "-//W3C//DTD XHTML 1.0 Strict//EN"
```
htreu: about.html is obsolete now.

bundles/config/org.eclipse.smarthome.config.writeback/build.properties
```diff
@@ -0,0 +1,6 @@
+source.. = src/main/java/
+output.. = target/classes/
+bin.includes = META-INF/,\
+               .,\
+               OSGI-INF/,\
+               about.html
```
htreu: remove about.html

bundles/config/org.eclipse.smarthome.config.writeback.test/META-INF/MANIFEST.MF
```diff
@@ -0,0 +1,13 @@
+Manifest-Version: 1.0
+Bundle-ManifestVersion: 2
+Bundle-Name: Eclipse SmartHome Config Writeback Tests
+Bundle-SymbolicName: org.eclipse.smarthome.config.writeback.test
+Bundle-Version: 0.9.0.qualifier
```
htreu: we are at 0.10.0 now

bundles/config/org.eclipse.smarthome.config.writeback.test/pom.xml
```diff
@@ -0,0 +1,89 @@
+<project xmlns="http://maven.apache.org/POM/4.0.0" xmlns:xsi="http://www.w3.org/2001/XMLSchema-instance" xsi:schemaLocation="http://maven.apache.org/POM/4.0.0 http://maven.apache.org/xsd/maven-4.0.0.xsd">
+  
+  <parent>
+    <groupId>org.eclipse.smarthome.bundles</groupId>
+    <artifactId>config</artifactId>
+    <version>0.9.0-SNAPSHOT</version>
```
htreu: 0.10.0-SNAPSHOT

bundles/config/org.eclipse.smarthome.config.writeback/META-INF/MANIFEST.MF
```diff
@@ -0,0 +1,15 @@
+Manifest-Version: 1.0
+Bundle-ManifestVersion: 2
+Bundle-Name: Eclipse SmartHome Config Writeback
+Bundle-SymbolicName: org.eclipse.smarthome.config.writeback
+Bundle-Version: 0.9.0.qualifier
```
htreu: 0.10.0

bundles/config/org.eclipse.smarthome.config.writeback/pom.xml
```diff
@@ -0,0 +1,25 @@
+<?xml version="1.0" encoding="UTF-8"?>
+<project xmlns:xsi="http://www.w3.org/2001/XMLSchema-instance"
+  xmlns="http://maven.apache.org/POM/4.0.0"
+  xsi:schemaLocation="http://maven.apache.org/POM/4.0.0 http://maven.apache.org/maven-v4_0_0.xsd">
+
+  <parent>
+    <groupId>org.eclipse.smarthome.bundles</groupId>
+    <artifactId>config</artifactId>
+    <version>0.9.0-SNAPSHOT</version>
```
htreu: 0.10.0-SNAPSHOT

htreu:
@kaikreuzer, @sjka & @maggu2810 please comment on if/how we plan to support configuration write back.

kaikreuzer:
> This is similar to Apache FileInstall felix.fileinstall.enableConfigSave=true

This is correct, so whoever wants to use such a mechanism in his solution, he is better off to directly use FileInstall.
The ConfigDispatcher is meant as a one-way mechanism: It is one source for configuration settings that are fed into ConfigAdmin.

Note that specifically openHAB regards the conf folder to be read-only, hence this service wouldn't be applicable to openHAB anyhow (and I assume that this is the primary intention).

I'd therefore suggest to close this PR.

Issue #4230, opened 2017-09-09
Maven build failing for any binding due to checkstyle code analysis


lolodomo:
If I run a "mvn clean install" at the root of the repo, everything is fine.

But if I run it inside "extensions/binding" or "extensions/binding/org.eclipse.smarthome.binding.sonos", I got this kind of error:
```
[ERROR] Failed to execute goal org.openhab.tools:static-code-analysis:0.3.0:checkstyle (default) on project org.eclipse.smarthome.binding.sonos: Unable to execute mojo: An error has occurred in Checkstyle report generation. Failed during checkstyle execution: Unable to find suppressions file at location: D:\dev\openhab2\git\smarthome\extensions\binding\org.eclipse.smarthome.binding.sonos\tools\static-code-analysis\checkstyle\suppressions.xml: Could not find resource 'D:\dev\openhab2\git\smarthome\extensions\binding\org.eclipse.smarthome.binding.sonos\tools\static-code-analysis\checkstyle\suppressions.xml'.
```
I can confirm that I don't have any tools directory inside the binding structure; this directory seems to be present only at the root of the repo.

J-N-K:
The same issue occurs in IDE. I didn't report it because I thought it was a local issue.

maggu2810:
I have a fix in mind. Will create a PR next week.

PR #4240, opened 2017-09-11
static-code-analyzer use directory realtive to root pom

maggu2810:
```sh
pushd extensions/binding/org.eclipse.smarthome.binding.astro/; mvn clean install; popd
```

before:
```text
[INFO] BUILD FAILURE
...
[ERROR] Failed to execute goal org.openhab.tools:static-code-analysis:0.3.0:checkstyle (default) on project org.eclipse.smarthome.binding.astro: Unable to execute mojo: An error has occurred in Checkstyle report generation. Failed during checkstyle execution: Unable to find suppressions file at location: /home/rathgeb/bin/pkgs/eclipse/esh/git/smarthome/extensions/binding/org.eclipse.smarthome.binding.astro/tools/static-code-analysis/checkstyle/suppressions.xml: Could not find resource '/home/rathgeb/bin/pkgs/eclipse/esh/git/smarthome/extensions/binding/org.eclipse.smarthome.binding.astro/tools/static-code-analysis/checkstyle/suppressions.xml'. -> [Help 1]
```

after:
```text
[INFO] BUILD SUCCESS
```

Fixes: https://github.com/eclipse/smarthome/issues/4230

### DIFF
--- a/pom.xml
+++ b/pom.xml
@@ -71,6 +71,23 @@
   <build>
     <plugins>
       <plugin>
+        <groupId>org.commonjava.maven.plugins</groupId>
+        <artifactId>directory-maven-plugin</artifactId>
+        <version>0.2</version>
+        <executions>
+          <execution>
+            <id>directories</id>
+            <goals>
+              <goal>highest-basedir</goal>
+            </goals>
+            <phase>initialize</phase>
+            <configuration>
+              <property>basedirRoot</property>
+            </configuration>
+          </execution>
+        </executions>
+      </plugin>
+      <plugin>
         <groupId>${tycho-groupid}</groupId>
         <artifactId>tycho-maven-plugin</artifactId>
         <version>${tycho-version}</version>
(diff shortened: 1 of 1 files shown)


kaikreuzer:
Interesting plugin. I guess we will need to request a build-time dependency CQ for it, before we can merge.

pom.xml
```diff
@@ -71,6 +71,23 @@
   <build>
     <plugins>
       <plugin>
+        <groupId>org.commonjava.maven.plugins</groupId>
+        <artifactId>directory-maven-plugin</artifactId>
+        <version>0.1</version>
```
kaikreuzer: There is a version 0.2 available (see https://mvnrepository.com/artifact/org.commonjava.maven.plugins/directory-maven-plugin) - any special reason why you nonetheless chose 0.1?

kaikreuzer:
I have created CQ 14162 for it.

pom.xml
maggu2810: No special reason, I use the version 0.1 in several projects and it is doing its job.
It seems the version 0.2 has been released on July 31th, 2017 and I was not aware of.
If 0.2 is working as expected, let's use the newer one.

maggu2810:
CQ has been approved but I don't see any version constrained in the CQ.
Doesn't the version matter?
Should I update this PR to use 0.2 and we are ready for merge?

kaikreuzer:
It has only been PMC-approved, not yet closed by the IP team, so please wait for that.
Yes, for build-time dependencies, the version does not matter, so please update to 0.2 if that works as well.

maggu2810:
> so please update to 0.2 if that works as well

Seems to do its job.

kaikreuzer:
CQ has been approved!

PR #4245, opened 2017-09-11
fix shuffling of items at load time

### LINKED ISSUE #4232: Random order of group members
I moved from OH snapshot 1003 to snapshot 1034 and unfortunately discovered that group members are now presented in a random order and no more like since the origin of OH in the order they are declared in the item file.
Why having changed that ? This was practical to have a kind of control on the group members order.
It has been changed after snapshot 1003.

sjka:
...by using a LinkedHashMap which retains the insertion order.

fixes #4232


lolodomo:
@sjka : thank you

PR #4249, opened 2017-09-11
Renamed a package in the org.eclipse.smarthome.io.rest.log bundle to internal.

VelinYordanov:


sjka:
As far as I can see, these the package that this PR now exports only contains constants which are used for the internal package. Therefore I'm wondering whether it is the right way to export this package, or rather move the constants to the internal one?

maggu2810:
Has this PR already been changed?
It seems that the constant are now internal (which is correct, IMHO).
Shouldn't the PR title be changed, too?

PR #4256, opened 2017-09-13
Added time zone in the PaperUI and in the I18ProviderImpl 

### LINKED ISSUE #3936: Timezone Service Proposal
Hello, I started the [refactoring of the DateTimeType](https://github.com/eclipse/smarthome/issues/3886) using a Java 8 class called **ZonedDateTime** which stores the timezone unlike the Calendar class (the current implementation).

   While adapting the tests of the **Astro** binding I faced difficulties working with the timezone. The given timestamps are with a certain **timezone** and to achieve the same timestamps in the test results they have to use a timezone _property_ in order to convert the values properly. Otherwise _local timezone_ is used which will cause test failures when they are not in correct timezone. The solution was to use the timezone offset as a configuration property.

   I would like to propose  implementing a  service which provides the **_timezone_** of a given _latitude_ and _longtitude_. This can be **reusable** and I believe it will come in handy later.

doandzhi:
Querying the underlying OS for a timezone and setting this as a default in the LocationProvider. A user could then choose a different one via a drop down list in **Paper UI**.
timezonesnimka1
yee

Fixes #3936 

### DIFF
--- a/bundles/config/org.eclipse.smarthome.config.core/src/main/java/org/eclipse/smarthome/config/core/internal/i18n/I18nConfigOptionsProvider.java
+++ b/bundles/config/org.eclipse.smarthome.config.core/src/main/java/org/eclipse/smarthome/config/core/internal/i18n/I18nConfigOptionsProvider.java
@@ -8,10 +8,16 @@
 package org.eclipse.smarthome.config.core.internal.i18n;
 
 import java.net.URI;
+import java.time.ZoneId;
 import java.util.Arrays;
 import java.util.Collection;
+import java.util.Collections;
 import java.util.Comparator;
+import java.util.LinkedList;
+import java.util.List;
 import java.util.Locale;
+import java.util.TimeZone;
+import java.util.concurrent.TimeUnit;
 import java.util.function.Function;
 import java.util.stream.Collectors;
 
@@ -23,29 +29,64 @@ import org.osgi.service.component.annotations.Component;
  * {@link ConfigOptionProvider} that provides a list of
  *
  * @author Simon Kaufmann - initial contribution and API
+ * @author Erdoan Hadzhiyusein - Added time zone
  *
  */
 @Component(immediate = true)
 public class I18nConfigOptionsProvider implements ConfigOptionProvider {
 
+    private static final String NO_OFFSET_FORMAT = "(GMT) %s";
+    private static final String NEGATIVE_OFFSET_FORMAT = "%s (GMT%d:%02d)";
+    private static final String POSITIVE_OFFSET_FORMAT = "%s (GMT+%d:%02d)";
+
     @Override
     public Collection<ParameterOption> getParameterOptions(URI uri, String param, Locale locale) {
         if (uri.toString().equals("system:i18n")) {
             Locale translation = locale != null ? locale : Locale.getDefault();
-            if (param.equals("language")) {
-                return getAvailable(locale, l -> new ParameterOption(l.getLanguage(), l.getDisplayLanguage(translation)));
-            } else if (param.equals("region")) {
-                return getAvailable(locale, l -> new ParameterOption(l.getCountry(), l.getDisplayCountry(translation)));
-            } else if (param.equals("variant")) {
-                return getAvailable(locale, l -> new ParameterOption(l.getVariant(), l.getDisplayVariant(translation)));
-            }
+            return processParamType(param, locale, translation);
         }
         return null;
+    }
+
+    private Collection<ParameterOption> processParamType(String param, Locale locale, Locale translation) {
+        switch (param) {
+            case "language":
+                return getAvailable(locale,
+                        l -> new ParameterOption(l.getLanguage(), l.getDisplayLanguage(translation)));
+            case "region":
+                return getAvailable(locale,
+                        l -> new ParameterOption(l.getCountry(), l.getDisplayCountry(translation)));
+            case "variant":
+                return getAvailable(locale,
+                        l -> new ParameterOption(l.getVariant(), l.getDisplayVariant(translation)));
+            case "timezone":
+                return ZoneId.getAvailableZoneIds().stream().map(TimeZone::getTimeZone).sorted((tz1, tz2) -> {
+                    return tz2.getRawOffset() - tz2.getRawOffset();
+                }).map(tz -> {
+                    return new ParameterOption(tz.getID(), getTimeZoneRepresentation(tz));
+                }).collect(Collectors.toList());
+            default:
+                return null;
+        }
+    }
+
+    private static String getTimeZoneRepresentation(TimeZone tz) {
+        long hours = TimeUnit.MILLISECONDS.toHours(tz.getRawOffset());
+        long minutes = TimeUnit.MILLISECONDS.toMinutes(tz.getRawOffset()) - TimeUnit.HOURS.toMinutes(hours);
+        minutes = Math.abs(minutes);
+        final String result;
+        if (hours > 0) {
+            result = String.format(POSITIVE_OFFSET_FORMAT, tz.getID(), hours, minutes);
+        } else if (hours < 0) {
+            result = String.format(NEGATIVE_OFFSET_FORMAT, tz.getID(), hours, minutes);
+        } else {
+            result = String.format(NO_OFFSET_FORMAT, tz.getDisplayName(Locale.getDefault()));
+        }
+        return result;
     }
 
     private Collection<ParameterOption> getAvailable(Locale locale, Function<Locale, ParameterOption> mapFunction) {
         return Arrays.stream(Locale.getAvailableLocales()).map(l -> mapFunction.apply(l)).distinct()
                 .sorted(Comparator.comparing(a -> a.getLabel())).collect(Collectors.toList());
     }
-
 }
--- a/bundles/core/org.eclipse.smarthome.core/src/main/java/org/eclipse/smarthome/core/i18n/TimeZoneProvider.java
+++ b/bundles/core/org.eclipse.smarthome.core/src/main/java/org/eclipse/smarthome/core/i18n/TimeZoneProvider.java
@@ -1,0 +1,25 @@
+/**
+ * Copyright (c) 2014-2017 by the respective copyright holders.
+ * All rights reserved. This program and the accompanying materials
+ * are made available under the terms of the Eclipse Public License v1.0
+ * which accompanies this distribution, and is available at
+ * http://www.eclipse.org/legal/epl-v10.html
+ */
+package org.eclipse.smarthome.core.i18n;
+
+import java.time.ZoneId;
+
+/**
+ * This interface describes a provider for time zone.
+ *
+ * @author Erdoan Hadzhiyusein - Initial contribution and API
+ */
+public interface TimeZoneProvider {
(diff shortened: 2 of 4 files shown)


bundles/config/org.eclipse.smarthome.config.core/src/main/java/org/eclipse/smarthome/config/core/internal/i18n/I18nConfigOptionsProvider.java
```diff
@@ -23,29 +28,65 @@
...
+                case "region":
+                    return getAvailable(locale,
+                            l -> new ParameterOption(l.getCountry(), l.getDisplayCountry(translation)));
+                case "variant":
+                    return getAvailable(locale,
+                            l -> new ParameterOption(l.getVariant(), l.getDisplayVariant(translation)));
+                case "timezone":
+                    Collection<ParameterOption> collection = new LinkedList<ParameterOption>();
+                    String[] ids = TimeZone.getAvailableIDs();
+                    List<TimeZone> timeZones = Arrays.asList(ids).stream().map(TimeZone::getTimeZone)
```
maggu2810: Instead of `Arrays.asList(array).stream()` you could use `Arrays.stream(array)`

bundles/config/org.eclipse.smarthome.config.core/src/main/java/org/eclipse/smarthome/config/core/internal/i18n/I18nConfigOptionsProvider.java
```diff
@@ -23,29 +28,65 @@
...
             }
         }
         return null;
     }
 
+    private static String getTimeZoneRepresentation(TimeZone tz) {
+        long hours = TimeUnit.MILLISECONDS.toHours(tz.getRawOffset());
+        long minutes = TimeUnit.MILLISECONDS.toMinutes(tz.getRawOffset()) - TimeUnit.HOURS.toMinutes(hours);
+        minutes = Math.abs(minutes);
+        String result = "";
```
maggu2810: Use `final String result;` so you ensure that result needs to be assigned in one of the branches and you don't need to assign it twice.

bundles/core/org.eclipse.smarthome.core/src/main/java/org/eclipse/smarthome/core/i18n/LocationProvider.java
```diff
@@ -23,4 +26,10 @@
      */
     PointType getLocation();
 
+    /**
+     * Provides access to the time zone property
+     *
+     * @return the time zone set in the Paper UI and if there isn't one, the default time zone of the underlying system
+     */
+    TimeZone getTimeZone();
```
maggu2810: Shouldn't there be a `TimeZoneProvider` interface that is implemented by `I18nProviderImpl` similar to `LocaleProvider` and `LocationProvider`?

bundles/core/org.eclipse.smarthome.core/src/main/java/org/eclipse/smarthome/core/internal/i18n/I18nProviderImpl.java
```diff
@@ -125,14 +124,40 @@ protected synchronized void modified(Map<String, Object> config) {
...
+            }
+        }
     }
 
     @Override
     public PointType getLocation() {
         return location;
     }
 
+    public TimeZone getTimeZone() {
```
maggu2810: Your IDE should add `@Override`.

bundles/config/org.eclipse.smarthome.config.core/src/main/java/org/eclipse/smarthome/config/core/internal/i18n/I18nConfigOptionsProvider.java
```diff
@@ -23,29 +28,65 @@
...
+                            l -> new ParameterOption(l.getLanguage(), l.getDisplayLanguage(translation)));
+                case "region":
+                    return getAvailable(locale,
+                            l -> new ParameterOption(l.getCountry(), l.getDisplayCountry(translation)));
+                case "variant":
+                    return getAvailable(locale,
+                            l -> new ParameterOption(l.getVariant(), l.getDisplayVariant(translation)));
+                case "timezone":
+                    Collection<ParameterOption> collection = new LinkedList<ParameterOption>();
+                    String[] ids = TimeZone.getAvailableIDs();
```
maggu2810: I assume you are more familar with the topic of Java 8's new time API because you implemented https://github.com/eclipse/smarthome/pull/4259

But shouldn't we use `ZoneId` instead of `TimeZone`?

```
There is an existing time zone class in Java—java.util.TimeZone—but it isn’t used by Java SE 8 be-cause all JSR 310 classes are immutable and time zone is mutable.
```

There is also `ZoneId.getAvailableZoneIds()`, ...

doandzhi:
Thanks for the quick reaction. I agree with the corrections you proposed.

maggu2810:
@doandzhi Hi, does it make more sense to provide a time zone or a zone id in the java code?

bundles/config/org.eclipse.smarthome.config.core/src/main/java/org/eclipse/smarthome/config/core/internal/i18n/I18nConfigOptionsProvider.java
```diff
@@ -23,29 +29,65 @@
...
+                    return getAvailable(locale,
+                            l -> new ParameterOption(l.getLanguage(), l.getDisplayLanguage(translation)));
+                case "region":
+                    return getAvailable(locale,
+                            l -> new ParameterOption(l.getCountry(), l.getDisplayCountry(translation)));
+                case "variant":
+                    return getAvailable(locale,
+                            l -> new ParameterOption(l.getVariant(), l.getDisplayVariant(translation)));
+                case "timezone":
+                    Collection<ParameterOption> collection = new LinkedList<ParameterOption>();
```
sjka: please factor out this logic into its own method so it stays easy to keep an overview of this switch statement (good idea, by the way!). 

And while you are at it: why didn't you keep using the stream? e.g. like 

```java
        return ZoneId.getAvailableZoneIds().stream().map(TimeZone::getTimeZone).sorted((tz1, tz2) -> {
            return tz2.getRawOffset() - tz2.getRawOffset();
        }).map(tz -> {
            return new ParameterOption(tz.getID(), getTimeZoneRepresentation(tz));
        }).collect(Collectors.toList());

```

bundles/core/org.eclipse.smarthome.core/src/main/java/org/eclipse/smarthome/core/internal/i18n/I18nProviderImpl.java
```diff
@@ -125,7 +125,29 @@ protected synchronized void modified(Map<String, Object> config) {
...
+        }
+    }
+
+    private void setTimeZone(final String timeZone) {
+        if (timeZone != null) {
+            try {
+                this.timeZone = TimeZone.getTimeZone(timeZone);
+            } catch (IllegalArgumentException e) {
+                this.timeZone = TimeZone.getDefault();
+                logger.warn("Could not set a new timezone, the system's default time zone will be used ", e);
```
sjka: ~~timezone~~ -> time zone

...to keep it consistent

bundles/core/org.eclipse.smarthome.core/src/main/java/org/eclipse/smarthome/core/internal/i18n/I18nProviderImpl.java
```diff
@@ -77,15 +82,10 @@ protected synchronized void modified(Map<String, Object> config) {
...
-        if (location != null) {
-            try {
-                this.location = PointType.valueOf(location);
-            } catch (IllegalArgumentException e) {
-                // preserve old location or null if none was set before
-                logger.warn("Could not set new location, keeping old one: ", e.getMessage());
-            }
-        }
+        setTimeZone(timeZone);
+        setLocation(location);
```
sjka: why did you factor out these, but not the language?

sjka:
PS: What I _initially_ wanted to ask you but forgot: Could you please rebase your branch onto the latest master? Thanks!

sjka:
Cool, thanks. No, that wasn't a requirement - it's easier to read right now, all good!

I just realized that you didn't answer @maggu2810's [question above](https://github.com/eclipse/smarthome/pull/4256#issuecomment-331172878). Considering that in #4259 you always need ZoneIDs and that it makes sense to stay within the new java.time scope, this is a very valid question to consider.

doandzhi:
Hey, @maggu2810! Regarding the changes you proposed in _I18nProviderImpl.java_ (using java.time object instead of String to store the time zone) I made the needed corrections. I would be very glad if you can take a look at some time so I can integrate the changes directly in #4259 soon.

bundles/core/org.eclipse.smarthome.core/src/main/java/org/eclipse/smarthome/core/i18n/TimeZoneProvider.java
```diff
@@ -0,0 +1,25 @@
...
+ * @author Erdoan Hadzhiyusein - Initial contribution and API
+ */
+public interface TimeZoneProvider {
+
+    /**
+     * Provides access to the time zone property.
+     *
+     * @return the time zone set in the Paper UI and if there isn't one, the default time zone of the underlying system
+     */
+    TimeZone getTimeZone();
```
sjka: If I understood \@maggu2810's question right, he was asking wether it might make more sense to return a `java.time.ZoneId` here instead of a `java.util.TimeZone`. Considering that with the `java.time.*` classes that you are using in #4259 you always have to call `timeZone.toZoneId()` on it anyway, IMHO this would make sense. Or did I miss anything?

bundles/core/org.eclipse.smarthome.core/src/main/java/org/eclipse/smarthome/core/i18n/TimeZoneProvider.java
```diff
@@ -0,0 +1,25 @@
...
+ * are made available under the terms of the Eclipse Public License v1.0
+ * which accompanies this distribution, and is available at
+ * http://www.eclipse.org/legal/epl-v10.html
+ */
+package org.eclipse.smarthome.core.i18n;
+
+import java.util.TimeZone;
+
+/**
+ * This interface describes a provider for a location and time zone.
```
sjka: This interface describes a provider for a ~~location and~~ time zone.

doandzhi:
Thanks for the appropriate comments. I have to be sorry for not considering these actually. Had a closer look and I think it looks better now. :)

sjka:
👍 
Thanks!

maggu2810:
> Hello, @maggu2810! Regarding the changes you proposed in I18nProviderImpl.java (using java.time object instead of String to store the time zone) I made the needed corrections. I would be very glad if you can take a look at some time so I can integrate the changes directly in #4259 soon.

Sorry, has been not available the last week.
Thanks @sjka for having a look at (you are right with your assumption what has been my meaning).
And thanks @doandzhi for your work.

PR #4258, opened 2017-09-13
save-action: final modifiers on private fields

### LINKED ISSUE #3651: Save Actions: final modifiers
As just pointed out in https://github.com/eclipse/smarthome/pull/3650#pullrequestreview-43702946, there are several mistakes which could not happen if we were using `final` modifiers for method arguments and private fields (where applicable), ideally supported by the IDE:

image

So far, we treat method arguments as "virtually final" in reviews, so it would just formalize what we are doing anyway. On the other side, visually it adds some bloat. As will #3624. 

As I could not find any public discussion related to this, I think it makes sense to have it here.

sjka:
fixes #3651

### DIFF
--- a/targetplatform/EclipseSmartHome.setup
+++ b/targetplatform/EclipseSmartHome.setup
@@ -568,7 +568,7 @@
       <setupTask
           xsi:type="setup:PreferenceTask"
           key="/instance/org.eclipse.jdt.ui/sp_cleanup.make_private_fields_final"
-          value="false"/>
+          value="true"/>
       <setupTask
           xsi:type="setup:PreferenceTask"
           key="/instance/org.eclipse.jdt.ui/sp_cleanup.make_type_abstract_if_missing_method"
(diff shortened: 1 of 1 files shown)


targetplatform/EclipseSmartHome.setup
```diff
@@ -568,15 +568,15 @@
...
           key="/instance/org.eclipse.jdt.ui/sp_cleanup.make_private_fields_final"
-          value="false"/>
+          value="true"/>
       <setupTask
           xsi:type="setup:PreferenceTask"
           key="/instance/org.eclipse.jdt.ui/sp_cleanup.make_type_abstract_if_missing_method"
           value="false"/>
       <setupTask
           xsi:type="setup:PreferenceTask"
           key="/instance/org.eclipse.jdt.ui/sp_cleanup.make_variable_declarations_final"
```
martinvw: What does this one do?

sjka: If you look at the screenshot in https://github.com/eclipse/smarthome/issues/3651#issue-235538193, this corresponds to the "outer" checkbox, i.e. it switches on the feature in general.

martinvw: Great!

PR #4262, opened 2017-09-13
Added @NonNull to property labels and getThingTypeUID

### LINKED ISSUE #4099: Question: Missing @NonNull in Thing.getThingTypeUID()
As discussed with @triller-telekom https://github.com/openhab/openhab2-addons/pull/2556#issuecomment-323672397

>>Because of the call to thing.getThingTypeUID() we do require a null check but I would suppose a thing always has an ThingTypeUID what do you think

>I think there is an annotation missing in the Thing class because a thing should always have a UID.

@triller-telekom and me think `@NonNull` annotation is missing on [getThingTypeUID](https://github.com/eclipse/smarthome/blob/master/bundles/core/org.eclipse.smarthome.core.thing/src/main/java/org/eclipse/smarthome/core/thing/Thing.java#L163) do you agree?

martinvw:
Added @NonNull to property labels and getThingTypeUID

Minor changes:

* Interface fields are always public static final, leave out the modifiers
* Corrected @return to @param in setLabel docblock
* Corrected field name in setBridgeUID docblock

Closes #4099

### DIFF
--- a/bundles/core/org.eclipse.smarthome.core.thing/src/main/java/org/eclipse/smarthome/core/thing/Thing.java
+++ b/bundles/core/org.eclipse.smarthome.core.thing/src/main/java/org/eclipse/smarthome/core/thing/Thing.java
@@ -32,19 +32,19 @@ import org.eclipse.smarthome.core.thing.binding.ThingHandler;
 public interface Thing extends Identifiable<ThingUID> {
 
     /** the key for the vendor property */
-    public static final String PROPERTY_VENDOR = "vendor";
+    @NonNull String PROPERTY_VENDOR = "vendor";
 
     /** the key for the model ID property */
-    public static final String PROPERTY_MODEL_ID = "modelId";
+    @NonNull String PROPERTY_MODEL_ID = "modelId";
 
     /** the key for the serial number property */
-    public static final String PROPERTY_SERIAL_NUMBER = "serialNumber";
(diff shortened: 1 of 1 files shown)


maggu2810:
Shouldn't we wait with nullable annotations until we finalize the coding guidelines?

martinvw:
Yes maybe we should, I did not follow the complete discussion my view was limited to linked issue, where a PR was 'requested' 🙂

kaikreuzer:
@maggu2810 I guess [I encouraged him](https://github.com/eclipse/smarthome/issues/4099#issuecomment-324259800) at a time when we thought that the discussions were already through - but in the light of the recent discussions on #4080, I would agree that we might want with that right now.

kaikreuzer:
With [these new guidelines](https://github.com/eclipse/smarthome/pull/4275/files#diff-24ba7e763168d1631a43ebdef19bcd5d), I would claim that this PR is obsolete.

kaikreuzer:
You might want to create a new PR with your other "minor changes" though - as they do not fit to the title of this PR, I preferred to close it and ask you for a new one :-)

martinvw:
> You might want to create a new PR with your other "minor changes" though - as they do not fit to the title of this PR, I preferred to close it and ask you for a new one :-)

Done, see #4293

PR #4267, opened 2017-09-14
Security API

davidgraeff:
A first API proposal for a security package. #4169 (POC: Encrypted credentials ) and #4105 (TLS Service) are targeted. The idea is to first agree on this API and to provide implementations for both parts in follow up PRs. 

## Secure Vault API 
* It is recommended to use char[] instead of String for secrets (https://stackoverflow.com/a/8889285)
* A vault service, one per requesting bundle. See http://blog.vogella.com/2017/02/13/control-osgi-ds-component-instances/. Proposed API:

```java
/**
 * The {@link SecretsVaultService} can be used to store and retrieve secrets.
 *
 * Secrets can only be retrieved by the bundle that has stored them and by default
 * the backend is required to store secrets encrypted only.
 *
 * Because of the way how the String class is handled internally, you are strongly
 * encouraged to use char or byte arrays or the ByteBuffer class
 * for keeping encrypted secrets in memory.
 *
 * The encryption algorithm, the storage location and format depend on the implementation.
 */
public interface SecretsVaultService {
   @Nullable ByteBuffer retrieve(String key) throws IOException, GeneralSecurityException;
   void set(String key, @Nullable ByteBuffer secret) throws IOException, GeneralSecurityException;
}
```
* A management interface to retrieve all keys, for GUIs like PaperUI. Including a way to change the master password.
```java
/**
 * The {@link SecretsVaultManagement} allows to retrieve all keys for a given
 * bundle or all bundles. The actual secrets cannot be retrieved via this
 * management interface and are only accessible by the {@link SecretsVaultService}
 * for the bundle that stored the secrets.
 */
public interface SecretsVaultService {
   Enumeration<SimpleEntry<String, String>> getAllKeys();
   Enumeration<String> getAllKeys(String bundlename);

    /**
     * Set/Delete a secret, identified by the given key, bundlename and old secret value.
     *
     * @param bundlename The fully qualified bundle name, e.g. org.eclipse.smarthome.binding.abc
     * @param key An identifier
     * @param oldSecret The old secret value. It need to match the currently stored value.
     * @param secret A secret as a ByteBuffer. If the ByteBuffer is empty, an empty secret will be stored.
     *            If null is given, the secret will be erased.
     * @return Return true if the modification was successful
     * @throws IOException If the underlying file can't be opened or is not writable.
     * @throws GeneralSecurityException If the encryption algorithm fails, this exception is thrown
     */
     boolean set(String bundlename, String key, ByteBuffer oldSecret, @Nullable ByteBuffer secret)
            throws IOException, GeneralSecurityException;
    /**
     * All bundle secrets are stored in a protected format. An implementation could be a java {@link KeyStore} for
     * example. A master password is used for encryption. This method allows to change the master password by
     * re-encrypting all bundle secrets with the new password.
     *
     * The implementation need to guarantee that the master password change happens in a transactional way and any
     * changes will be rolled back if an error happens.
     *
     * @param oldSecret The old secret value. It need to match the current master password
     * @param secret
     * @return Returns false if the new master password could not be applied.
     */
     boolean setMasterSecret(ByteBuffer oldSecret, ByteBuffer secret);
}
```

* Annotation interface for fields in a configuration holder object. `Secret` annotated fields will automatically be retrieved from {@link SecretsVaultService} and are instantly available for consumption. Usage example:

```java
class YourConfig {
   public String username;
   public @Secret("pwd") Bytebuffer password;
   public @Secret Bytebuffer bitcoinWalletKey;
};
```

Configuration holder objects can be used in services like this:
```java
void activate(@Nullable Map<String, Object> configMap) {
   YourConfig config = new Configuration(configMap).as(YourConfig.class);
}
```

## Network secure sockets
A central service interface to provide SSL/DTLS configuration for server-like services within ESH and derived products. Server like services are for example:

* The MQTT embedded broker
* The :8080 web-page (The REST endpoint, the UIs)

```java
public interface SecureSocketServers {
    /**
     * To retrieve a working secure socket implementation , you would call this
     * method and get an asynchronous callback as soon as the context is ready. It can take a while (some seconds)
     * to resolve a new SSLContext if there is none so far.
     *
     * Describe what features your secure socket implementation
     * should fulfil (mandatory) and what features are nice to have (optional).
     *
     * For example if you need a DTLS, you would set `mandatory` to
     * {SecureSocketCap.DTLS}.
     */
    CompletableFuture<SSLContext> requestSecureSocket(Set<SecureSocketCap> mandatory,
       Set<SecureSocketCap> optional, @Nullable String contextTag);
}
```


davidgraeff:
@triller-telekom Can you have a look at this API proposal?

bundles/io/org.eclipse.smarthome.io.security/src/main/java/org/eclipse/smarthome/io/security/api/Cryptographic.java
```diff
@@ -0,0 +1,65 @@
...
+
+    /**
+     * Encrypts the given bytes by the given password.
+     *
+     * @param password The password as byte array.
+     * @param plain The plain bytes to encrypt
+     * @return Returns the ciphered text.
+     * @throws GeneralSecurityException
+     */
+    byte[] encrypt(byte[] password, byte[] plain) throws GeneralSecurityException;
```
triller-telekom: In the `SecretsVaultService` you are working on `ByteBuffer`s and here you are using `byte[]` if the `ByteBuffer` is really more convenient maybe we should use it here too, or change both to use `byte[]` to have it consistent.

bundles/io/org.eclipse.smarthome.io.security/src/main/java/org/eclipse/smarthome/io/security/api/Cryptographic.java
```diff
@@ -0,0 +1,65 @@
...
+public interface Cryptographic {
+    /**
+     * Encrypts the given text by the given password.
+     *
+     * @param password The password as byte array.
+     * @param plain The plain text to encrypt
+     * @return Returns the ciphered text.
+     * @throws GeneralSecurityException
+     */
+    char[] encrypt(byte[] password, char[] plain) throws GeneralSecurityException;
```
triller-telekom: I am thinking about whether it makes sense to support `String plain` instead of the `char[]` which would be more convenient to use. Also it makes sense to provide the password as a `String`. The same applies to the `decrypt` method.

davidgraeff: I would agree convenient-wise but please read the linked article in the security vault issue. The problem is: strings are allocated on the VM heap, and their content cannot be overwritten or cleared (immutable). The passwords would remain in memory until the next Garbage collection. 
We shouldn't risk that for a security API. The user can still use strings and call toCharArray if he doesn't care.

bundles/io/org.eclipse.smarthome.io.security/src/main/java/org/eclipse/smarthome/io/security/api/Cryptographic.java
davidgraeff: I'm using ByteBuffer if the argument can be null and byte arrays otherwise.

bundles/io/org.eclipse.smarthome.io.security/src/main/java/org/eclipse/smarthome/io/security/api/TlsNetworkServer.java
```diff
@@ -0,0 +1,87 @@
...
+ */
+package org.eclipse.smarthome.io.security.api;
+
+import java.io.IOException;
+import java.net.InetSocketAddress;
+import java.security.GeneralSecurityException;
+
+import javax.net.ssl.SSLContext;
+
+import org.eclipse.californium.elements.Connector;
```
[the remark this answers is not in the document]
davidgraeff: Java9 hast native Support for dtls, but right now a SSLContext is not enough unfortunately. If you look at the TLS service pr, I create those two with different default algorithms. To my knowledge it is also not possible to actually create a dtlsConnector with just an sslcontext/sslengine. And the purpose is to make it easy for the API consumer to create a secure tcp or udp socket. It doesn't need to be californium but should be something.

bundles/io/org.eclipse.smarthome.io.security/src/main/java/org/eclipse/smarthome/io/security/api/Cryptographic.java
```diff
@@ -0,0 +1,65 @@
...
+     * @throws GeneralSecurityException
+     */
+    byte[] decrypt(byte[] password, byte[] cipher) throws GeneralSecurityException;
+
+    /**
+     * Returns the master password hash. Can be used to encrypt/decrypt, but
+     * remember that it will change, if the user changes the master password
+     * and old cipher will not be able to be decrypted anymore.
+     */
+    byte[] masterPasswordHash();
```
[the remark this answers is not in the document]
davidgraeff: I wanted to avoid such a listener. Everybody can register for this listener and an evil binding just logs the new plain text password and has full access. 

Instead I used a different pattern, see the security vault issue in the section Use-case.

bundles/io/org.eclipse.smarthome.io.security/src/main/java/org/eclipse/smarthome/io/security/api/Cryptographic.java
davidgraeff: The password argument can be char or byte. The problem I see with chars is that they are 16bit wide. What if your password is {1,2,3,4,5} (read as bytes). How do you express this in chars? I think only bytes allow us the full range of possible password combinations.

paulianttila: Is password the best term? Would secret, secretKey, encyptionKey or passphrase be better? It depends on algorithm what kind of parameters is need for encryption (e.g. IV), so is "password" enough?

bundles/io/org.eclipse.smarthome.io.security/src/main/java/org/eclipse/smarthome/io/security/api/Cryptographic.java
```diff
@@ -0,0 +1,65 @@
...
+
+    /**
+     * Decrypts the given text with the given password.
+     *
+     * @param password The password as byte array.
+     * @param cipher The cipher to decrypt
+     * @return Returns the plain text
+     * @throws GeneralSecurityException
+     */
+    char[] decrypt(byte[] password, char[] cipher) throws GeneralSecurityException;
```
paulianttila: Cipher? Does cipher means here encrypted password? Cipher term is normally used as a term for used algorithm.

triller-telekom: He probably means `cipherText`. But I agree, this name choice might be confusing.

bundles/io/org.eclipse.smarthome.io.security/src/main/java/org/eclipse/smarthome/io/security/api/Cryptographic.java
triller-telekom: You mean that if you use `char[]` we would need to add a padding, because `byte` is 8bit and `char` is 16bit, so the `bytes` have to be a multiple of 2?

bundles/io/org.eclipse.smarthome.io.security/src/main/java/org/eclipse/smarthome/io/security/api/Cryptographic.java
triller-telekom: If there is a param defined as `byte[] param` one can also pass `null`, so I am confused about your explanation.

bundles/io/org.eclipse.smarthome.io.security/src/main/java/org/eclipse/smarthome/io/security/api/Cryptographic.java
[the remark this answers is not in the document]
triller-telekom: True, we definitely do not want to have a listener where everyone can obtain the password from :) But then we would need a "migration service". If you change the password I would not want to restart the system and wait for a re-encryption of all secrets. This should happen on the fly. If you offer your proposed `boolean migrateMasterPassword(String oldPassword, String newPassword)` method it should re-encrypt all secrets immediately and lock the configurations until this migration has finished and then store newly added data encrypted with the new password.

bundles/io/org.eclipse.smarthome.io.security/src/main/java/org/eclipse/smarthome/io/security/api/Cryptographic.java
triller-telekom: > It depends on algorithm what kind of parameters is need for encryption (e.g. IV), so is "password" enough?

I think this should be left to the implementation and not stated on the interface. I would like to have the interface "as simple as possible", i.e. the less parameters which are "optional", the better.

bundles/io/org.eclipse.smarthome.io.security/src/main/java/org/eclipse/smarthome/io/security/api/Cryptographic.java
davidgraeff: I can't use annotations on a byte array, at least it didn't work for me. I have no problems of using ByteBuffer all the time. It just wraps a byte array and has almost no additional runtime costs.

bundles/io/org.eclipse.smarthome.io.security/src/main/java/org/eclipse/smarthome/io/security/api/Cryptographic.java
davidgraeff: I see this problem:
Char a = 1; -> bytes {0,1}. And that we read bytes from files anyway. We do we want to convert to utf8, 16bit-wide and then back to a byte representation (where we can fail miserably if done wrong).

bundles/io/org.eclipse.smarthome.io.security/src/main/java/org/eclipse/smarthome/io/security/api/Cryptographic.java
davidgraeff: encryptedDate would be an option

bundles/io/org.eclipse.smarthome.io.security/src/main/java/org/eclipse/smarthome/io/security/api/Cryptographic.java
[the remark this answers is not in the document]
davidgraeff: Can we lock configadmin? A restart sounds more failsafe proof. It's a bad comparison, I know, but when I change the password of my router, I need to restart it. If I change a Windows login password, I need to logout and login again. I think people are used to the restart pattern if something central like the master password is changed. But open to suggestions.

bundles/io/org.eclipse.smarthome.io.security/src/main/java/org/eclipse/smarthome/io/security/api/Cryptographic.java
triller-telekom: If it doesn't add "much" (I don't think "no") additional runtime cost we could use it. But I wouldn't make this decision dependent on whether you can add a null annotation or not. Otherwise you also have to replace all "int", "boolean", etc. parameters in the code as well because you cannot add the null annotations on primitives.

bundles/io/org.eclipse.smarthome.io.security/src/main/java/org/eclipse/smarthome/io/security/api/Cryptographic.java
paulianttila: > I think this should be left to the implementation and not stated on the interface. I would like to have the interface "as simple as possible", i.e. the less parameters which are "optional", the better.

I agree, but why password is then in the interface as it's not enough for encryption anyway. This is the reason why I didn't introduce it in my PoC test, but keeped interface even simpler.

triller-telekom: Ah now I got your problem. The problem is the conversion with a charset... In this case I think it makes sense to use `byte[] passPhrase`.

bundles/io/org.eclipse.smarthome.io.security/src/main/java/org/eclipse/smarthome/io/security/api/Cryptographic.java
triller-telekom: encryptedDat**e** implies that it is a date from a calendar, so I would rather use encrytedDat**a**

bundles/io/org.eclipse.smarthome.io.security/src/main/java/org/eclipse/smarthome/io/security/api/Cryptographic.java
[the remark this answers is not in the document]
triller-telekom: Another thought: What if you supply a new password but do **not** restart immediately but leave the system running for days/weeks... Which password will be used for new secrets that will be entered?

Also where do you store the new password, that you provided via rest api, in memory? But you do not have that memory if you restart the system...

bundles/io/org.eclipse.smarthome.io.security/src/main/java/org/eclipse/smarthome/io/security/api/Cryptographic.java
triller-telekom: > I agree, but why password is then in the interface as it's not enough for encryption anyway. This is the reason why I didn't introduce it in my PoC test, but keeped interface even simpler.

Hmm good point actually, because then the password would also stay within the security service and will not end up elsewhere in the system (possibly even in Strings). @davidgraeff Is there any reason against this proposal?

bundles/io/org.eclipse.smarthome.io.security/src/main/java/org/eclipse/smarthome/io/security/api/Cryptographic.java
davidgraeff: Actually I do this. For the Mqtt connection object for example. The port parameter is an Integer instead of an int so that the user can set the parameter explicitly to null if he doesn't care. For booleans it makes sense too sometimes if you need a tristate (on/off/disabled).

bundles/io/org.eclipse.smarthome.io.security/src/main/java/org/eclipse/smarthome/io/security/api/Cryptographic.java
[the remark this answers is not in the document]
davidgraeff: To stick to the router example: I get a hint message that the settings will only be applied if I restart the system.

The basic implementation does not have a rest service interface. The password can only be changed via environment variable on a restart.

An advanced (but not via ESH provided) implementation would have a PasswordChangeService or similar, which writes a new password to a file and forces a restart for example and offers a rest service interface. what do u think?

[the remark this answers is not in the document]
triller-telekom: I still do not quite understand how exactly this "environment variable" thing will look like. Do you have one variable "ESH_PASSWORD" which will be used by default and another one "ESH_PASSWORD_NEW" which the user has to set if he wants to change the current one? Who resets "ESH_PASSWORD_NEW" to "" once "ESH_PASSWORD" has been updated? And afterwards the user has to restart the whole ESH instance?

> PasswordChangeService or similar, which writes a new password to a file and forces a restart

How do you want to enforce a restart of the complete system from within a service? Does that work?

Also if the restart work, does the PasswordChangeService then start before all the other bindings, read the new password, migrate all secrets and then remove the file containing the new password, or what is your proposed work-flow?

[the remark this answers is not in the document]
davidgraeff: If a service has the access rights of an OSGi console, it can end and restart services. But that's not the point. 

We either have observers for a password change with old and new password and we agreed that's a bad idea, or we restart the framework. There's no other option that I can see here. 

My idea was a environment variable with the current password (/ new password) and if it is a change then there's a variable with the old password as well. We check the current password first and then the old password if existing to decrypt. If the old matches we re encrypt everything.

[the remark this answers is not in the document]
davidgraeff: Reencrypt. Github mobile doesn't allow edits. Sigh.

bundles/io/org.eclipse.smarthome.io.security/src/main/java/org/eclipse/smarthome/io/security/api/Cryptographic.java
davidgraeff: Yes. I again refer to the Use-case of the other issue. With an API where the password is not given by the user but master password derived, everything is lost on a master password change. Because we don't have password changed listeners to re encrpyt all kind of different consumers. That wouldn't work anyway, what if a bundle is temporarily not enabled and later reenabled. It would miss the change and again encrypted data is lost.

bundles/io/org.eclipse.smarthome.io.security/src/main/java/org/eclipse/smarthome/io/security/api/TlsNetworkServer.java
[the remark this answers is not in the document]
davidgraeff: I can thing of a custom return class where everything necessary is included to create a dtlsConnector. This way we are independent of californium.

bundles/io/org.eclipse.smarthome.io.security/src/main/java/org/eclipse/smarthome/io/security/api/SecretsVaultService.java
```diff
@@ -0,0 +1,48 @@
...
+     * @throws GeneralSecurityException If the decryption algorithm fails, this exception is thrown
+     */
+    @Nullable
+    ByteBuffer getSecret(@NonNull String key) throws IOException, GeneralSecurityException;
+
+    /**
+     * Set or delete a secret, identified by the given key.
+     *
+     * @param key An identifier
+     * @param secret A secret as a ByteBuffer. Can be null to delete the secret.
```
paulianttila: Addition: If the vault previously contained a value for the key, the old value is replaced.

bundles/io/org.eclipse.smarthome.io.security/src/main/java/org/eclipse/smarthome/io/security/api/SecretsVaultService.java
```diff
@@ -0,0 +1,48 @@
...
+     * @param key An identifier
+     * @return Either null if no secret could be found under the given key, or the secret.
+     * @throws IOException If the underlying file can't be opened or read.
+     * @throws GeneralSecurityException If the decryption algorithm fails, this exception is thrown
+     */
+    @Nullable
+    ByteBuffer getSecret(@NonNull String key) throws IOException, GeneralSecurityException;
+
+    /**
+     * Set or delete a secret, identified by the given key.
```
paulianttila: Should delete have own method rather than use null as special behaviour?

bundles/io/org.eclipse.smarthome.io.security/src/main/java/org/eclipse/smarthome/io/security/api/Cryptographic.java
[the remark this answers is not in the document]
triller-telekom: Sorry I still don't get what you mean. :(

> My idea was a environment variable with the current password (/ new password)

Is that now one or two variables?

> and if it is a change

I am assuming "currentPW" and "newPW" are two variables.

> then there's a variable with the old password as well

So you copy the content of "currentPW" to "oldPW" and the content of "newPW" to "currentPW"?

> We check the current password first and then the old password

Why do you even introduce a third variable? Two are enough, aren't they? One for the old and one for the new password.

bundles/io/org.eclipse.smarthome.io.security/src/main/java/org/eclipse/smarthome/io/security/api/Cryptographic.java
triller-telekom: I think there is a slight misunderstanding. What @paulianttila  and me mean is that the security service should take care of handling the password, i.e. reading it from and environment variable, password file, etc. If the user changes the password this service can also take care of the re-encryption. Other services that are using the security service then only pass the data to it.

Regarding the problem that the password has been changed and that services cannot decrypt their data anymore: We have this problem also if the API for `encrypt()` and `decrypt()` contain the password parameter, because then it is up to them to detect that password change in the environment variables or whereever it has been changed and then call these methods with the new function.

So I think the better option is to keep everything inside the security module to handle this password change in one central point.

bundles/io/org.eclipse.smarthome.io.security/src/main/java/org/eclipse/smarthome/io/security/api/SecretsVaultService.java
triller-telekom: For me both is fine, but maybe an extra method would be cleaner since this method does not need to accept `null`.

bundles/io/org.eclipse.smarthome.io.security/src/main/java/org/eclipse/smarthome/io/security/api/Cryptographic.java
davidgraeff: I'm on a mobile only these days so it's quite hard to express myself with touch keyboards. But: no the problem does not exist if password is a parameter:
* The binding chooses a random key B.
* B is stored in the secret vault: sec("aKey",B)
* B is restored on binding start instead of generating it new from now on.
* so existing as long as the user doesn't purge everything.
* The binding encrypts: C = enc(B,toStore).
* The master password can be changed. Because everything in the security vault is reencrypted and still retrievable even after change.

bundles/io/org.eclipse.smarthome.io.security/src/main/java/org/eclipse/smarthome/io/security/api/Cryptographic.java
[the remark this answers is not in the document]
davidgraeff: Two variables only.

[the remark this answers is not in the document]
davidgraeff: It's current and old. Old is optional and should it be set for a change but it's not relevant if it still exists on later starts because we always first try current. And if that works we ignore old. But again: if there's another better approach I'm open to that. I only want the API here. And even the API can be extended, just not changed later on.

bundles/io/org.eclipse.smarthome.io.security/src/main/java/org/eclipse/smarthome/io/security/api/SecretsVaultService.java
davidgraeff: Less is better sometimes :D if there's delete I expect add. Set/get is easy and simple

bundles/io/org.eclipse.smarthome.io.security/src/main/java/org/eclipse/smarthome/io/security/api/Cryptographic.java
triller-telekom: I think we are talking about different things here...

1. You are saying that the binding (lets call it "user", since it uses the security service) creates its own encyption key(s) and uses it for further actions.

2. What I am saying is that we just create ONE encryption key (derived from the master password given in the environment variable) for the whole ESH installation. All secrets will be encrypted with this single key.

So I guess we should distinguish these 2 use-cases. Case 2 can take care of encrypting all config values that should not be stored as plain text.

So why do you need case 1? This would be a general mechanism in ESH which you can pass a key + value to and obtain the encrypted result. I do not see the need for different keys in ESH, IMHO one main key is enough and all secrets can be encrypted with that one.

davidgraeff: I would agree, that one secret would be awesome in ESH. But please tell me how the following scenario works and if we just assume that the master password just can't be changed ever:

* We have a binding WEB that needs an api token T.
* We use `encrypt(T)` and get T_ENC. T_ENC is encrypted just by the master password.
* T is given via a text cfg file: apiToken=enc(T_ENC). 

* The user changes the master password.
* Binding WEB starts at some point, gets the config and uses `decrypt(T_ENC)`.
* But it doesn't get T back. It gets T_WRONGLY_DECRYPTED back, because the underlying password changed.

How is that supposed to work?

paulianttila: All credentials should be saved to secret vault not in bindings configuration. Binding configuration parameter is just a reference to secret vault. So e.g. WEB bindings apiToken configuration parameter could be apiToken=SECRET(WEB_apiToken). BaseThingHandler recognise that WEB_apiToken is reference to secret vault credential and query WEB_apiToken from secret vault. Secret vault decrypt the credential and return decrypted credential, which is then passed to binding. 

Cryptographic API is then not needed at all or it's internal API for secret vault cryptographic providers. Secret vault fully handles the master key change and it's transparent from the bindings or other services.

For backward compatible reason, user could add secret vault reference to configuration by special SECRET(key) name, but we could also introduce new credential type, which is then automatically handled by the configuration UI's. So e.g. Paper UI could show all possible secret vault key names in the drop down menu, where user can select correct one. Then secret vault could have dedicated interface to manage secrets (add, remove, change, query all key names, etc).

davidgraeff: Yes, that's a perfect suggestion. So we would have: add(key,secret)/delete(key)/get(key)/modify(key,secret).
Add throws if it would overwrite or secret is null.
Modify throws if key is not existing.

And we have getAllKeys() but without the actual decrypted secrets for the rest interface.

Get()/modify() should only work for the bundle that stored the secret. So the rest interface etc will not be able to access the secrets at all.

Configadmin should not decrypt therefore, but the configuration class, that can be used in the correct bundle context then.

paulianttila: > So the rest interface etc will not be able to access the secrets at all.

Well, there could be (maybe later) dedicated REST interface also for secret vault where user could manage the secrets. But yes, for security reason, existing credentials should not ever exposed as decrypted to the REST interface. Only add and modify could be possible including secret, and then remove and getAllKeys. That REST interface should only be supported via TLS connections and hopefully someday restricted by the user roles as well.

triller-telekom: That sounds like a plan. So the Secret Vault service keeps all secrets to himself and handles the encryption/decryption of configuration values transparently for all other services.

The implementation of this Secret vault service should be configurable, i.e. algorithms and keys sizes should be changeable with secure default (as far as possible with the US import/export restrictions).

For the REST interface I think at the moment we do not have any other choice than what @paulianttila  suggests since we do not have roles to restrict it. I still have an aching stomach with the modification through the REST interface because potentially everyone could change the secrets then. And changing a secret in this context is almost the same as reading the current one in plaintext since for example as an attacker if I replace the secret key of an ssl connection service, I am able to read the traffic. Any good ideas on restricting the modification even more?

The only idea I have now is to maybe restrict the `modify` command on the REST interface to only be called from external hosts, i.e. the user in his browser and not a malicious bundle executing the call. WDYT?

Also we need a way for users using text files to configure their ESH instance to somehow store a given plain text token in the secrets vault service and refer to that in the config file.

davidgraeff: Modify API could look like this:
`void modify(key, old_secret, new_secret)`
So the rest interface would ask for the old secret as well. That way it cannot be changed that easily by malicious code on top of the rest interface.

davidgraeff:
@paulianttila @triller-telekom: I have adapted the API proposal

bundles/io/org.eclipse.smarthome.io.security/src/main/java/org/eclipse/smarthome/io/security/api/Cryptographic.java
triller-telekom: Sounds good.

bundles/io/org.eclipse.smarthome.io.security/src/main/java/org/eclipse/smarthome/io/security/api/SecretsVaultService.java
```diff
@@ -0,0 +1,61 @@
...
+import java.security.GeneralSecurityException;
+
+import org.eclipse.jdt.annotation.NonNullByDefault;
+import org.eclipse.jdt.annotation.Nullable;
+
+/**
+ * The {@link SecretsVaultService} can be used to store and retrieve secrets.
+ *
+ * Secrets can only be retrieved by the bundle that stored them and by default
+ * will be encrypted and and stored on disk immediately.
```
triller-telekom: Typo "and and"

bundles/io/org.eclipse.smarthome.io.security/src/main/java/org/eclipse/smarthome/io/security/api/Cryptographic.java
```diff
@@ -0,0 +1,74 @@
...
+/**
+ * {@link Cryptographic} allows you to encrypt and decrypt. The algorithm and
+ * algorithm parameterisation is configured centralised for the framework.
+ *
+ * You can use the master password hash or a derived password to encrypt secrets, please be
+ * aware that a change of the master password can occur at any time though.
+ *
+ * @author David Graeff - Initial contribution
+ */
+public interface Cryptographic {
```
triller-telekom: As @paulianttila [suggested](https://github.com/eclipse/smarthome/pull/4267/files#r139303413) this interface can be removed, because the `SecretsVaultService` is doing the encryption internally, right?

bundles/io/org.eclipse.smarthome.io.security/src/main/java/org/eclipse/smarthome/io/security/api/SecretsVaultService.java
```diff
@@ -0,0 +1,61 @@
...
+
+    /**
+     * Store a secret, identified by the given key.
+     *
+     * @param key An identifier
+     * @param secret A secret as a ByteBuffer. If the ByteBuffer is empty, an empty secret will be stored.
+     * @throws IOException If the underlying file can't be opened or is not writable.
+     * @throws GeneralSecurityException If the encryption algorithm fails, this exception is thrown
+     */
+    void add(String key, ByteBuffer secret) throws IOException, GeneralSecurityException;
```
triller-telekom: Should this method not throw an exception if the given key already exists and not silently overwrite it? Because as far as I understand the current API a modification has to go through `SecretsVaultManagement.modify()`?

bundles/io/org.eclipse.smarthome.io.security/src/main/java/org/eclipse/smarthome/io/security/api/SecretsVaultService.java
```diff
@@ -0,0 +1,61 @@
...
+     * Retrieve a secret, identified by the given key.
+     *
+     * @param key An identifier
+     * @return Either the secret as ByteBuffer that was stored by this bundle under the given key
+     *         or null if no secret could be found.
+     * @throws IOException If the underlying storage file can't be opened or read.
+     * @throws GeneralSecurityException If the decryption algorithm fails, this exception is thrown
+     */
+    @Nullable
+    ByteBuffer retrieve(String key) throws IOException, GeneralSecurityException;
```
triller-telekom: I still do not understand how this retrieving of  a secret works. As a user of this new security API if I want to retrieve a secret for my bundle what do I have to do?

1. List<String> mySecrets = SecretsVaultManagement.getAllKeys("myBundle")
2. for(String secret : mySecrets) { SecretsVaultService.retrieve(secret) }

If this is correct for `retrieve`, how does `add` work? Do I use `SecretsVaultManagement.modify` also to add new keys? If so we should rename it to something like `addOrModify` to make it more transparent.

bundles/io/org.eclipse.smarthome.io.security/src/main/java/org/eclipse/smarthome/io/security/api/SecretsVaultService.java
```diff
@@ -0,0 +1,61 @@
...
+     * @throws GeneralSecurityException If the encryption algorithm fails, this exception is thrown
+     */
+    void add(String key, ByteBuffer secret) throws IOException, GeneralSecurityException;
+
+    /**
+     * Clears the secret at the given key. Does nothing if the secret does not exist.
+     *
+     * @param key An identifier
+     */
+    void delete(String key);
```
triller-telekom: Since with a `delete` and `add` we could simulate a `modify`, should we also require the `oldSecret` as a parameter if one wants to delete its key?

bundles/io/org.eclipse.smarthome.io.security/src/main/java/org/eclipse/smarthome/io/security/api/TlsNetworkServer.java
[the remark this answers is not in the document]
triller-telekom: Can you please change it to your proposed custom class?

bundles/io/org.eclipse.smarthome.io.security/src/main/java/org/eclipse/smarthome/io/security/api/SecretsVaultService.java
davidgraeff: As a consumer of the api, you know your keys. You probably would not iterate over all your stored secrets.

Add only adds and should either return with a Boolean or throw if the key is already existing.

If the consumer needs to modify, it would delete and add again.

---

The rest API can't add with the current proposal, I forgot about that. We probably rename modify to "set" and allow to add a secret as well with an empty old secret if there's nothing saved so far under the current key.

bundles/io/org.eclipse.smarthome.io.security/src/main/java/org/eclipse/smarthome/io/security/api/SecretsVaultService.java
davidgraeff: Yes. It should return a Boolean or throw.

bundles/io/org.eclipse.smarthome.io.security/src/main/java/org/eclipse/smarthome/io/security/api/SecretsVaultService.java
triller-telekom: > If the consumer needs to modify, it would delete and add again.

Why do we offer `SecretsVaultManagement.modify` (or `set` after the change below) then?

> The rest API can't add with the current proposal, I forgot about that. We probably rename modify to "set" and allow to add a secret as well with an empty old secret if there's nothing saved so far under the current key.

So you suggest to rename `SecretsVaultManagement.modify` to `SecretsVaultManagement.set` and allow for adding new secrets and overwriting existing ones (with matching oldSecret)? That's what I meant, so fine with me.

davidgraeff: To answer your question: The service interface is to be used within a bundle for that bundles secrets. The old secret is not necessary to modify a secret here. 

The management interface can be used by any bundle but might be security restricted internally and the bundle name needs to be supplied as well as old secret to change a secret.

bundles/io/org.eclipse.smarthome.io.security/src/main/java/org/eclipse/smarthome/io/security/api/SecretsVaultService.java
davidgraeff: See other reply. This interface is used within a bundle and accesses only that bundles secrets. The old secret does not need to be provided here. That was my plan at least, I don't know if this is realizable.

bundles/io/org.eclipse.smarthome.io.security/src/main/java/org/eclipse/smarthome/io/security/api/SecretsVaultService.java
triller-telekom: I understand your separation but how do we ensure that the implementation of the service interface is not called by someone else? I mean this implementation has a `delete` and `add` method that are public, right? But maybe this is a stupid question because we would have to expose these methods from the implementation somehow first before others can use it?

davidgraeff: I thought about the OSGi service factory. If a bundle requests the secret vault service, an specific instance for that bundle is created. I haven't tested this with code yet.

bundles/io/org.eclipse.smarthome.io.security/src/main/java/org/eclipse/smarthome/io/security/api/Cryptographic.java
davidgraeff: Should we provide this anyway? We need something like this to implement the secret vault service and could export it for other use cases

triller-telekom: You mean for the secret vault service you need some methods that actually do the encryption and decryption of the configuration values? I am not sure if we should define an interface for that because this strongly depends on the implementation how this is done. As we might also need to provide an IV if the underlying cryptographic function needs it. So my suggestion is that we leave this entirely to the implementation of the secret vault service. @paulianttila  WDYT?

davidgraeff:
After some research I came to the conclusion that it is actually not that easy to move away from the californium elements connector. The facts are:
* SSLContext and DTLSConnector are working similar in hiding the actual secrets away from the API consumer
* Both are build with a `KeyStore` that holds all private/public keys, certificates and PSKs.

To be able to create a DTLSConnector just with the API we provide here, we would need access to the `KeyStore`. We cannot expose `KeyStore` though:
  * It could be manipulated and even stored again.
  * We would need to expose secrets like the `KeyStore` password and private key passwords as well.

However: It is possible to create a DTLS connection in Java9 with just a SSLEngine (derived from a SSLContext). I created an issue for californium: https://github.com/eclipse/californium/issues/421.

So we either keep the californium elements connector (maybe by declaring the API as non-stable) or switch to Java9 soon.

bundles/io/org.eclipse.smarthome.io.security/src/main/java/org/eclipse/smarthome/io/security/api/SecretsVaultService.java
```diff
@@ -0,0 +1,64 @@
...
+import java.nio.ByteBuffer;
+import java.security.GeneralSecurityException;
+
+import org.eclipse.jdt.annotation.NonNullByDefault;
+import org.eclipse.jdt.annotation.Nullable;
+
+/**
+ * The {@link SecretsVaultService} can be used to store and retrieve secrets.
+ *
+ * Secrets can only be retrieved by the bundle that stored them and by default
```
paulianttila: > Secrets can only be retrieved by the **bundle that stored them**

How this is handled? Should secret storing and modifying be possible e.g. from karaf console (not bundle which will use the secrets)?

davidgraeff: This is possible via the management interface of this proposal, that is also used by the rest interface. Retrieving is not possible. Modifying requires to supply the old secret.

bundles/io/org.eclipse.smarthome.io.security/META-INF/MANIFEST.MF
```diff
@@ -0,0 +1,14 @@
+Manifest-Version: 1.0
+Bundle-ManifestVersion: 2
+Bundle-Name: Eclipse SmartHome security API
+Bundle-SymbolicName: org.eclipse.smarthome.io.security;singleton:=true
+Bundle-Version: 0.9.0.qualifier
+Bundle-Vendor: Eclipse.org/SmartHome
+Bundle-RequiredExecutionEnvironment: JavaSE-1.8
+Export-Package: org.eclipse.smarthome.io.security.api
+Import-Package: org.eclipse.californium.elements,
```
triller-telekom: I think this is a left over and should be removed too, right?

bundles/io/org.eclipse.smarthome.io.security/src/main/java/org/eclipse/smarthome/io/security/api/SecretsVaultService.java
triller-telekom: Could you please give it a try and give some details on how that could work? If it doesn't I would clearly vote for an additional argument to the `delete` method which is the `oldValue` and only delete it if it matches.

triller-telekom:
> So we either keep the californium elements connector for now (maybe by declaring the API as non-stable)

I saw that you have removed the californium dependency, so did you decide for the option below already?

> or switch to Java9 soon for just the SSLContext createSSLContext(...) API to be usable for DTLS connections.

This is clearly not an option, a switch in the major revision of java is a huge step for ESH and cannot be done within days/weeks.

> To be able to create a DTLSConnector just with the API we provide here, we would need access to the KeyStore

Instead of providing our "holy keystore" to them, how about handing over a "dummy keystore" with just the keys that are necessary? Unfortunately in this case we need to offer a method to extract the public and private key from our "holy keystore" and put it into the dummy one which isn't nice either. WDYT?

davidgraeff:
Exactly, we expose the private key. And the API would allow to enumerate all private keys. The entire purpose of this API would be defeated.

davidgraeff:
I removed the part, because we should first agree on this part of the API, and when I implement lwm2m we can discuss, on what additional API (or temporary API) is necessary to support DTLS as well.

bundles/io/org.eclipse.smarthome.io.security/META-INF/MANIFEST.MF
davidgraeff: yes

davidgraeff:
> Could you please give it a try and give some details on how that could work? If it doesn't I would clearly vote for an additional argument to the delete method which is the oldValue and only delete it if it matches.

I'm not sure if I understand. What prevents me from doing this: `delete(key, retrieve(key))`?

triller-telekom:
> I removed the part, because we should first agree on this part of the API. When I implement lwm2m we can discuss, what additional API (or temporary API) is necessary to support DTLS as well.

Great, let's discuss this outside of this PR.

> I'm not sure if I understand. What prevents me from doing this: `delete(key, retrieve(key))`?

You are right, but this brings us back to [your proposed OSGI service factory](https://github.com/eclipse/smarthome/pull/4267#discussion_r140295159) and if it is able to successfully hide this service from other bundles than the one which is allowed to use it. any news on that one?

davidgraeff:
@triller-telekom According to this page (http://blog.vogella.com/2017/02/13/control-osgi-ds-component-instances/) it is possible for a factory to create an instance per bundle. I thought that @paulianttila will adapt his solution. If I would need to implement the API, I would just add it to the SSLService package and would not touch anything in core (I would not adapt the Configuration class like @paulianttila did).

triller-telekom:
> @triller-telekom According to this page (http://blog.vogella.com/2017/02/13/control-osgi-ds-component-instances/) it is possible for a factory to create an instance per bundle.

Sounds exactly like what we need!

> I thought that @paulianttila will adapt his solution.

Can you give me some context what exactly you mean by that?

> If I would need to implement the API, I would just add it to the SSLService package and would not touch anything in core (I would not adapt the Configuration class like @paulianttila did).

I think the security vault is a core service and I would appreciate if @paulianttila adopts his configuration PR #4169 so the current `Configuration` service can use the new security vault service.

maggu2810:
I assume as long as we have to support OSGi 4.2 we need to use features of DS 1.1 only.

sjka:
At least service factories should be available in DS 1.1 already (see OSGi R4.2 CMPN 112.4.6 "Service Element").

davidgraeff:
I thought we are on 4.3? It is much more of a hassle with 4.2, as pointed out by the linked article. If we are on 4.2 indeed, can we discuss to move to 4.3? Considering that the standard is super old in IT years by now.

maggu2810:
See this read https://dev.eclipse.org/mhonarc/lists/smarthome-dev/msg00056.html

davidgraeff:
New versions usually come with improved developer comfort or in this case necessary functionality. It usually makes sense to not fall behind too much. In this case we would update, according to the semantic versioning, to a compatible minor version.

Why does ESH not has a yearly dependency version update vote or something similar? Everybody that is using ESH and interested in its development would subscribe to this poll and is able to comment on why a newer version might not fit his or other use-cases and point out release blockers. At the moment I don't see how ESH can stay up-to-date with libraries and dependencies, if everybody is afraid of touching what is already there with no clear update path.

davidgraeff:
To sum my last post up: I have no idea how a security API can be realized with OSGi 4.2 only.

triller-telekom:
> I have no idea how a security vault API can be realized with OSGi 4.2 only.

What about @sjka's comment about service factories, those are available, do we need anything else?

> At least service factories should be available in DS 1.1 already (see OSGi R4.2 CMPN 112.4.6 "Service Element").

davidgraeff:
@triller-telekom If you know how to do this, sure. The only material that I found on the web for bundle instances uses DS 1.2.

triller-telekom:
In OSGi R4.2 CMPN 112.4.6 "Service Element" I read:

> Service Element
The service element is optional. It describes the service information to be
used when a component configuration is to be registered as a service.
A service element has the following attribute:
• servicefactory – Controls whether the service uses the ServiceFactory
concept of the OSGi Framework. The default value is false. If
servicefactory is set to true, a different component configuration is
created, activated and its component instance returned as the service
object for each distinct bundle that requests the service. Each of these
component configurations has the same component properties. Otherwise,
the same component instance from the single component configuration
is returned as the service object for all bundles that request the
service.
The servicefactory attribute must not be true if the component is a factory
component or an immediate component. This is because SCR is not free to
create component configurations as necessary to support servicefactory. A
component description is ill-formed if it specifies that the component is a
factory component or an immediate component and servicefactory is set to
true.

If @sjka is right that DS 1.1 is OSGI 4.2 then this should work. I also found this on the openHAB page: http://docs.openhab.org/developers/prerequisites/osgids.html

maggu2810:
> If @sjka is right that DS 1.1 is OSGI 4.2

Have a look at the top page header of the OSGi R4.2 document in the capter DS :wink:

triller-telekom:
> Have a look at the top page header of the OSGi R4.2 document in the capter DS :wink:

Such a large document and I was just jumping into the mentioned paragraph. But that's good news that DS 1.1 supports service factories. @davidgraeff

davidgraeff:
So the vault service would just declare itself as service factory, which leads to a new instance for every bundle requesting it. That's nice. So we can continue on agreeing on the API :)

triller-telekom:
I think this service factory discussion was the last open point where we disagreed, right? I just scrolled through our comments  but didn't find any other issues. So if we implement the vault service as a factory and have the management service as a general service to configure all vaults, I am fine with this API now.

triller-telekom:
@davidgraeff Are you also fine with the current content (interfaces) of this PR? If so, please let us know and we will merge it soon so you can start implementing the basic implementation that uses whatever best algorithms the default JDK offers.

triller-telekom:
@davidgraeff: ping :)

bundles/io/org.eclipse.smarthome.io.security/src/main/java/org/eclipse/smarthome/io/security/api/SecretsVaultManagement.java
```diff
@@ -0,0 +1,56 @@
...
+
+    /**
+     * Modify a secret, identified by the given key, bundlename and old secret value.
+     *
+     * @param bundlename The fully qualified bundle name, e.g. org.eclipse.smarthome.binding.abc
+     * @param key An identifier
+     * @param oldSecret The old secret value
+     * @param secret A secret as a ByteBuffer. If the ByteBuffer is empty, an empty secret will be stored.
+     * @return Return true if the modification was successful
+     * @throws IOException If the underlying file can't be opened or is not writable.
```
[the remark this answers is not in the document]
davidgraeff: IOException perfectly expresses the type of errors that can occur, if you think about it. The only layer (apart from "no memory") that could cause problems for this function is the storage layer. I agree that we should not talk about files here, but IOException also works for Memory-only, Special hardware or DB storage.

bundles/io/org.eclipse.smarthome.io.security/src/main/java/org/eclipse/smarthome/io/security/api/SecretsVaultService.java
```diff
@@ -0,0 +1,64 @@
...
+ * will be encrypted and stored on disk immediately.
```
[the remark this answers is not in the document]
davidgraeff: I wanted to express requirements for the implementation in this API as well, I guess. The implementation / storage layer SHOULD encrypt immediately. How the storage is implemented is not that important of course.

bundles/io/org.eclipse.smarthome.io.security/src/main/java/org/eclipse/smarthome/io/security/api/TlsNetworkServer.java
```diff
@@ -0,0 +1,73 @@
...
+     * @param securityProvider Either {@link TlsNetworkServer#TLS} for TCP connections or
+     *            {@link TlsNetworkServer#DTLS} for UDP connections on Java 9 and newer.
+     * @return An SSLContext, that can be used to derive a SSLEngine or a SSLServerSocketFactory.
+     * @throws IOException If loading or saving the keyStore fails or reading or writing the configuration fails, this
+     *             exception is thrown.
+     * @throws GeneralSecurityException This exception is raised if creating the self
+     *             signed certificate fails (because the requested algorithms are not supported for instance) or if the
+     *             existing certificate is expired and service is configured to not automatically create a new one.
+     */
+    SSLContext createSSLContext(String context, String securityProvider) throws GeneralSecurityException, IOException;
```
davidgraeff: In retrospective I would say this API is not very OSGi like. The context parameter should be removed. A service factory should be used. The requester should be able to request a special SSLContext by providing a context string via service parameter.

bundles/io/org.eclipse.smarthome.io.security/src/main/java/org/eclipse/smarthome/io/security/api/annotation/Secret.java
```diff
@@ -0,0 +1,66 @@
...
+ * ```
+ *
+ * What happens instead is this:
+ *
+ * # On load
+ * When loading a service, the configuration dispatcher service will retrieve the services
+ * configuration and the services `activate(@Nullable Map<String, Object> configMap)` method
+ * is called. The `Secret` annotated fields will be null/not-set in that map.
+ * If you follow the configuration holder object pattern, you will do something like:
+ * `YourConfig config = new Configuration(configMap).as(YourConfig.class);`.
```
triller-telekom: I really like your idea of "marking" a configuration parameter as "security sensitive".

However, we do have a mechanism for that already: You can mark a config parameter with the context "password". Maybe we should exploit this instead of introducing this other mechanism with an annotation?

Why I have put my comment here is because your idea is to do the fetching+decryption of these sensitive config parameters within the static call to `new Configuration(configMap).as` or after #5292 `ConfigUtil.as`. This fetching and decryption will be done by an OSGi service, right? How do you access it from a `static` context?

bundles/io/org.eclipse.smarthome.io.security/src/main/java/org/eclipse/smarthome/io/security/api/SecureSocketCapabilities.java
```diff
@@ -0,0 +1,34 @@
...
+
+/**
+ * A set of default capabilities. Implement SecureSocketCapability and define your own for additional capabilities.
+ */
+public enum SecureSocketCapabilities implements SecureSocketCapability {
+    /** Capability: Request a TLS capable secure socket implementation */
+    TLS,
+    /**
+     * Capability: Request a DTLS capable secure socket implementation.
+     * On Java 8 this is usually a Scandium DtlsConnector. On Java 9 it is a SSLContext.
```
triller-telekom: If we have to make the distinction between `SSLCONTEXT` and `SCANDIUM_CONNECTOR`, why do you duplicate the information by also offering `DTLS` here?

bundles/io/org.eclipse.smarthome.io.security/about.html
```diff
@@ -0,0 +1,28 @@
+<!DOCTYPE html PUBLIC "-//W3C//DTD XHTML 1.0 Strict//EN"
```
triller-telekom: Please replace this file with a `NOTICE` file, because with the change to EPL-2 we are no longer using `about.html` files, see https://github.com/eclipse/smarthome/blob/master/bundles/core/org.eclipse.smarthome.core/NOTICE for an example.

bundles/io/org.eclipse.smarthome.io.security/src/main/java/org/eclipse/smarthome/io/security/api/SecureSocketRequest.java
```diff
@@ -0,0 +1,78 @@
...
+
+    /**
+     * @return Return optional capabilities. They are considered if possible to match the best secure socket
+     *         implementation. If you prefer a CA-signed certificate over a self-signed one for example, you would
+     *         express this here.
+     */
+    Set<SecureSocketCapability> optionalCapabilities();
+
+    /**
+     * Returns a potentially new secure socket implementation instance that fulfills the requested mandatory
```
triller-telekom: If you say that this method should `return` something, why is it `void`? :)

Since you use this `SecureSocketRequest` class as a callback object, maybe you should make this clear that this methods sets the socket implementation on this callback object but does not `return` it.

bundles/io/org.eclipse.smarthome.io.security/src/main/java/org/eclipse/smarthome/io/security/api/SecureSocketRequest.java
```diff
@@ -0,0 +1,78 @@
...
+
+    /**
+     * Because it is a common case to request the java frameworks own SSLContext, this specific method will get called
+     * instead of {@link SecureSocketRequest#secureSocketImplementationResponse(Object)} if the secure socket
+     * implementation matches a SSLContext.
+     *
+     * @param secureSocketContext May be null, if at least one capability could not be matched. May be null in a
+     *            subsequent call if the user removed an implementation deliberately.
+     */
+    void secureSocketImplementationResponse(@Nullable SSLContext secureSocketContext);
```
triller-telekom: see above

bundles/io/org.eclipse.smarthome.io.security/src/main/java/org/eclipse/smarthome/io/security/api/SecureSocketCapabilities.java
davidgraeff: I'd like to be the first user of this API in another binding and I plan to support Java9 SSLContexts as well as a DtlsConnector. Preferably it is a SSLContext of course. So I would only require the capability "DTLS" and don't care which implementation is actually returned. Other bindings might do this different and definitely want a DtlsConnector. They would request the SCANDIUM_CONNECTOR+DTLS capability.

bundles/io/org.eclipse.smarthome.io.security/src/main/java/org/eclipse/smarthome/io/security/api/annotation/Secret.java
davidgraeff: Yes we should avoid handling one case in different ways here. Maybe we should discuss #5300 first. As a result we might get a core annotation for the "password context" and this `Secure` annotation class is not necessary anymore.

I can imagine that the potential ConfigUtil.as method moves again to the core bundle into a class constructor that also takes a bundleContext. That way we would be able to resolve the secureVault service. Hypothetical code example:
```java
@Modified
public void modified(BundleContext bundleContext, Map<String, Object> config) {
    YourConfig configuration = new ConfigParser(bundleContext,config,YourConfig.class);
}
```

bundles/io/org.eclipse.smarthome.io.security/src/main/java/org/eclipse/smarthome/io/security/api/SecureSocketCapabilities.java
triller-telekom: Which of the possible options below is a user of the API required to ask for?
* `DTLS` AND `SSLContext` OR `DTLS` AND `SCANDIUM_CONNECTOR`
* `SSLContext` OR `SCANDIUM_CONNECTOR`
* just `DTLS` and depending on the java runtime version you dynamically get a `SSLContext` OR `SCANDIUM_CONNECTOR`

bundles/io/org.eclipse.smarthome.io.security/src/main/java/org/eclipse/smarthome/io/security/api/annotation/Secret.java
triller-telekom: Shall we leave this `Secret` annotation out of this PR and create a new issue/PR for that after we have discussed #5300?

I think it is better to discuss it afterwards and finalize this security API. Apart from my uncertainty on how one should use the `SecureSocketCapabilities` (https://github.com/eclipse/smarthome/pull/4267/files#r176436119) and the minor javadoc remarks in the `SecureSocketRequest` class: https://github.com/eclipse/smarthome/pull/4267/files#r176450243 and https://github.com/eclipse/smarthome/pull/4267/files#r176450487 I am fine with the API.

bundles/io/org.eclipse.smarthome.io.security/src/main/java/org/eclipse/smarthome/io/security/api/SecureSocketCapabilities.java
davidgraeff: It is unfortunately that complicated because Java 8 installations need to be considered. The API is more generic therefore. The implementation of this API that ESH provides would return a SSLContext only, to keep dependencies low. On Java 8, if you request `TLS`, you get something back, if you request `DTLS`, no provider can be returned. You can optionally require `SSLContext` to make sure, it is always a SSLContext. Because:

Another bundle can provide more options e.g. Scandium Connectors additionally. It is more or less undefined which secure socket provider will be choosen, if you just request `DTLS`.

I have adjusted the comments in code to explain the flags a little more.

bundles/io/org.eclipse.smarthome.io.security/pom.xml
```diff
@@ -0,0 +1,23 @@
...
+  <parent>
+    <groupId>org.eclipse.smarthome.bundles</groupId>
+    <artifactId>io</artifactId>
+    <version>0.10.0-SNAPSHOT</version>
+  </parent>
+
+  <properties>
+    <bundle.symbolicName>org.eclipse.smarthome.io.security</bundle.symbolicName>
+    <bundle.namespace>org.eclipse.smarthome.io.security</bundle.namespace>
+  </properties>
```
wborn: Please remove these properties. See #6060.

kaikreuzer:
Closing it here - once required by some feature and you want to continue working on it, please port this code over to https://github.com/openhab/openhab-core. Thanks!

Issue #4251, opened 2017-09-11
[Tradfri] Support for new RGB bulbs


hreichert:
IKEA sells new Tradfri products.

RGB bulbs, E27 socket, 600 lm, 8.6 W
packaged with 1 remote control
Product number: 703.389.51

I think they will be announced in October 2017.
The official IKEA Tradfri app does not support them fully right now. Currently it's only possible to change the color temperature (from white to orange), but you can not select a color yet. If you change the color through the remote control (RGB), the app shows the correct color.

**What about support for this RGB bulbs in the Eclipse Smarthome Tradfri Binding?**
I'm willing to implement the RGB support and contribute it, but I got only limited time resources right now... So I want to coordinate first, if somebody else is already implementing this. 

Here are some pictures:

![tradfri_rgb_1](https://user-images.githubusercontent.com/21249127/30291873-7ac52fea-9734-11e7-94f0-e87fb538ea2f.jpg)
![tradfri_rgb_2](https://user-images.githubusercontent.com/21249127/30291872-7ac4e83c-9734-11e7-8900-c8593a7c8acb.jpg)
![tradfri_rgb_3](https://user-images.githubusercontent.com/21249127/30291875-7acc5608-9734-11e7-9a6e-7b5e371eba7e.jpg)
![tradfri_rgb_4](https://user-images.githubusercontent.com/21249127/30291874-7aca0240-9734-11e7-950c-dac8943d236b.jpg)

normalfag:
ja, genau!

sjka:
> What about support for this RGB bulbs in the Eclipse Smarthome Tradfri Binding?

Yes, sure! The only limiting factor so far was their non-availability - nobody mentioned here yet that he/she was working on it. So it's good that you created this issue for coordination! 

Looking forward to your contribution!

hreichert:
CoAP response for RGB bulb:
```
{  
  "9001":"TRADFRI bulb E27 CWS opal 600lm",
  "9002":1505151864,
  "9020":1505253777,
  "9003":65550,
  "5750":2,
  "9054":0,
  "9019":1,
  "3":{  
    "0":"IKEA of Sweden",
    "1":"TRADFRI bulb E27 CWS opal 600lm",
    "2":"",
    "3":"1.3.002",
    "6":1
  },
  "3311":[  
    {  
      "5707":0,
      "5850":1,
      "9003":0,
      "5710":6554,
      "5851":8,
      "5708":0,
      "5709":13107,
[...]

PR #4271, opened 2017-09-15
[Tradfri] Support for new RGB bulbs #4251

hreichert:
Issue: #4251

Add support for the new TRADFRI full color lamps.
* New ThingType _0210_ / Extended Color Light with additional channel _color_
* Extended _TradfriLightHandler_
    * Handle color setting and retrieving
    * Changed color temperature calculation
* New class TradfriColor for colorspace conversions
* Extended _TradfriDiscoveryService_ for automatic discovery of color lights
* Tests for the above
* Documentation in README.md

I think it's ready for review.

### DIFF
--- a/extensions/binding/org.eclipse.smarthome.binding.tradfri/ESH-INF/thing/thing-types.xml
+++ b/extensions/binding/org.eclipse.smarthome.binding.tradfri/ESH-INF/thing/thing-types.xml
@@ -71,7 +71,7 @@
         </config-description>
     </thing-type>
 
-    <thing-type id="0200">
+    <thing-type id="0210">
         <supported-bridge-type-refs>
             <bridge-type-ref id="gateway"/>
         </supported-bridge-type-refs>
@@ -80,7 +80,6 @@
         <description>A dimmable light that supports full colors and color temperature settings.</description>
 
         <channels>
-            <channel id="brightness" typeId="brightness"/>
             <channel id="color_temperature" typeId="color_temperature"/>
             <channel id="color" typeId="color"/>
         </channels>
--- a/extensions/binding/org.eclipse.smarthome.binding.tradfri/ESH-INF/thing/thing-types.xml
+++ b/extensions/binding/org.eclipse.smarthome.binding.tradfri/ESH-INF/thing/thing-types.xml
@@ -71,6 +71,28 @@
         </config-description>
     </thing-type>
 
+    <thing-type id="0200">
+        <supported-bridge-type-refs>
+            <bridge-type-ref id="gateway"/>
+        </supported-bridge-type-refs>
+
+        <label>Color Light</label>
+        <description>A dimmable light that supports full colors and color temperature settings.</description>
+
+        <channels>
+            <channel id="brightness" typeId="brightness"/>
+            <channel id="color_temperature" typeId="color_temperature"/>
+            <channel id="color" typeId="color"/>
+        </channels>
+
+        <config-description>
+            <parameter name="id" type="integer" required="true">
+                <label>ID</label>
+                <description>The identifier of the light on the gateway</description>
+            </parameter>
+        </config-description>
+    </thing-type>
+
     <!-- note that this isn't yet supported by the code as we do not receive any data from the gateway for it -->
     <thing-type id="0820" listed="false">
         <supported-bridge-type-refs>
@@ -105,6 +127,13 @@
         <item-type>Dimmer</item-type>
         <label>Color Temperature</label>
         <description>Control the color temperature of the light.</description>
+        <category>ColorLight</category>
+    </channel-type>
+
+    <channel-type id="color">
+        <item-type>Color</item-type>
+        <label>Color</label>
+        <description>Control the color of the light.</description>
         <category>ColorLight</category>
     </channel-type>
 
--- a/extensions/binding/org.eclipse.smarthome.binding.tradfri/README.md
+++ b/extensions/binding/org.eclipse.smarthome.binding.tradfri/README.md
@@ -12,7 +12,7 @@ The thing type ids are defined according to the lighting devices defined for Zig
 |--------------------------|------------------|------------|
 | Dimmable Light           | 0x0100           | 0100       |
 | Colour Temperature Light | 0x0220           | 0220       |
-| Colour Light             | 0x0200           | 0200       |
+| Extended Colour Light    | 0x0210           | 0210       |
 
 The following matrix lists the capabilities (channels) for each of the supported lighting device types:
 
@@ -20,7 +20,7 @@ The following matrix lists the capabilities (channels) for each of the supported
 |-------------|:------:|:----------:|:-----:|:-----------------:|   
 |  0100       |    X   |     X      |       |                   |
 |  0220       |    X   |     X      |       |          X        |
-|  0200       |    X   |     X      |   X   |          X        |
+|  0210       |    X   |            |   X   |          X        |
 
 ## Thing Configuration
 
@@ -30,8 +30,12 @@ The devices require only a single (integer) parameter, which is their instance i
 
 ## Channels
 
-All devices support the `brightness` channel.
-The white spectrum bulbs additionally also support the `color_temperature` channel. Full color bulbs additionally also support the `color` channel.
+The dimmable bulbs support the `brightness` channel.
+The white spectrum bulbs additionally also support the `color_temperature` channel. 
+
+Full color bulbs support the `color_temperature` and `color` channels.
+Brightness can be changed with the `color` channel.
+
 Refer to the matrix above.
 
 | Channel Type ID   | Item Type | Description                                 |
@@ -48,7 +52,7 @@ demo.things:
 Bridge tradfri:gateway:mygateway [ host="192.168.0.177", code="EHPW5rIJKyXFgjH3" ] {
     0100 myDimmableBulb [ id=65537 ]    
     0220 myColorTempBulb [ id=65538 ]
-    0200 myColorBulb [ id=65539 ]
+    0210 myColorBulb [ id=65539 ]
 }
 ```
 
@@ -56,7 +60,7 @@ demo.items:
 
 ```
 Dimmer Light { channel="tradfri:0100:mygateway:myDimmableBulb:brightness" }
-Color ColorLight { channel="tradfri:0200:mygateway:myColorBulb:color" } 
+Color ColorLight { channel="tradfri:0210:mygateway:myColorBulb:color" } 
 ```
 
 demo.sitemap:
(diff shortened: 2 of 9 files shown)


extensions/binding/org.eclipse.smarthome.binding.tradfri/src/main/java/org/eclipse/smarthome/binding/tradfri/handler/TradfriLightHandler.java
```diff
@@ -243,17 +267,51 @@ private void handleColorTemperatureCommand(Command command) {
...
 
+    private void handleColorCommand(Command command) {
+        if (command instanceof HSBType) {
+            // tradfri can not handle color (xy) and brightness at once,
+            // so send color first and schedule the brightess update.
+            // 1500ms wait time because brightness setting would interrupt color fading
+            setColor((HSBType) command);
+            scheduler.schedule(() -> {
+                setBrightness(((HSBType) command).getBrightness());
+            }, 1500, TimeUnit.MILLISECONDS);
```
hreichert: Maybe it's better to calculate the delay based on the dimmer transition time?

extensions/binding/org.eclipse.smarthome.binding.tradfri/ESH-INF/thing/thing-types.xml
```diff
@@ -71,6 +71,28 @@
...
+    <thing-type id="0200">
+        <supported-bridge-type-refs>
+            <bridge-type-ref id="gateway"/>
+        </supported-bridge-type-refs>
+
+        <label>Color Light</label>
+        <description>A dimmable light that supports full colors and color temperature settings.</description>
+
+        <channels>
+            <channel id="brightness" typeId="brightness"/>
```
kaikreuzer: The color channel already accepts all the same (PercentType) commands as the brightness channel, so a thing that has the color channel, should not have the brightness channel as well as this is redundant.
See e.g. also the hue + LIFX bindings where it is the same.

hreichert: Ok, I can remove the channel. But I think that we loose functionality:
Imagine a group gWohnzimmer_Brightness for the living room.
In this group we have
* Something_Wohnzimmer_Brightness (something something)
* Tradfri_Wohnzimmer1_Brightness (dimmable bulb)
* Tradfri_Wohnzimmer2_Brightness (dimmable color temperature bulb)
* Tradfri_Wohnzimmer3_Brightness (full color bulb)

Further we have a `Slider item=gWohnzimmer_Brightness` in the sitemap.
Slide the slider and see how all lamps, also the full color bulb, change their brightness.
Without the brightness channel, one would have to use some rules, I think?
Or do I have a knot in my brain?

htreu: Iirc when sending only the brightness value (0 - 100) to a color item it will be interpreted as brightness only. So in your example all bulbbs should change their brightness.

hreichert: Ah well I think you're right.
Will verify this and update my PR accordingly.

sjka: You remember correctly! The [ColorItem](https://github.com/eclipse/smarthome/blob/74a486dfbf8201a2c48f021fa2ef4353a5ba2ca8/bundles/core/org.eclipse.smarthome.core/src/main/java/org/eclipse/smarthome/core/library/items/ColorItem.java#L83) takes care of this: If it gets updated with a percent type, it keeps everything else from the previous state and only adjusts the brightness.

extensions/binding/org.eclipse.smarthome.binding.tradfri/ESH-INF/thing/thing-types.xml
```diff
@@ -71,6 +71,28 @@
         </config-description>
     </thing-type>
 
+    <thing-type id="0200">
```
cweitkamp: In case of a color lamp with adjustable color temperature and brightness the corresponding ZigBee ID should be 210 instead of 200.

hreichert: The _real_ device ID of this lamps is 200 (0x0200), please see #4251 (confirmed on hardware).
Do we take the _real_ device ID, or the device ID that we imagine that is correct?
The whole thingType naming based on technical device IDs, that the IKEA gateway do not even expose, is anyway a bit.. hmm

kaikreuzer: When creating it this way I was hoping that IKEA would stay close to the specification, which would be the best way to figure out the capabilities of the device. This is what we ended up with in the hue binding after many 3rd party devices were supported by it. I'd guess that sooner or later IKEA will come to the same conclusion and our approach will prove to be the right choice ☺️

cweitkamp: Sry didn't look into #4251. You already complained your choice. And this is another break with the ZigBee protocol caused by IKEA developers.

To be compliant I would suggest to use the same thing-type definition as we use in the Hue and other bindings. In this case it should be the device ID of what we think the bulb is capable of -> 210. If not, we should remove the color_temperature channel as well (where later one is IMHO the worse way).

btw. I didn't mentioned it before: You did a great job so far.

hreichert: My personal feeling is also that 210 is the right choice, and not 200. But as the _real_ device ID was 210, I renamed it so stick with the previous naming scheme.
I'm happy to re-rename it to 210, cause this better reflects the capabilities of the Thing.

P.S: After re-reading my previous comment, it sounded a bit offensive. That was not intentional.

extensions/binding/org.eclipse.smarthome.binding.tradfri/ESH-INF/thing/thing-types.xml
hreichert: I tried this, but got unexpected results.
My steps:
* Breakpoints on `TradfriLightHandler.handleCommand` and `ColorItem.setState`
* Item: tradfri_0200_gwa0cc2b6d64bb_65551_color
* Set a color with BasicUI -> handleCommand receives HSBType -> good
* Send "100" to the same item -> handleCommand receives PercentType -> unexpected?
* AFTER this, the breakpoint on `ColorItem.setState` kicks in
My current code has already a special handling for PercentType on the color channel, so the functionality is OK. But the ColorItem.setState doesn't help here, I think.
What's your opinion about this?

extensions/binding/org.eclipse.smarthome.binding.tradfri/ESH-INF/thing/thing-types.xml
cweitkamp: > But as the real device ID was 210, ...

You mean 200, I guess. ;-) Okay, we give it a try.

> ... it sounded a bit offensive.

Oh no, everything was alright. I am fine with it.

hreichert: Yes, 200... All these numbers!
I changed it now to *210* in the binding and the documentation. PR update ongoing with the other changes.

hreichert:
Updated with following changes:
* Renamed ThingType 200 to 210
* Removed "brightness" channel from ThingType 210
* Improved color calculations
* Documentation

extensions/binding/org.eclipse.smarthome.binding.tradfri/ESH-INF/thing/thing-types.xml
kaikreuzer: Hm, my comment above actually meant that we SHOULD go for the real Zigbee type, because sooner or later other devices will appear with the 210 type - and maybe they then need a different logic to be supported (e.g. hardware instead of software color temp setting). So I was actually fine with the "200" choice.

>  If not, we should remove the color_temperature channel as well

@cweitkamp The other option would be that we think about doing "software calculation" on those (hue) as well - do you think that would be possible?

extensions/binding/org.eclipse.smarthome.binding.tradfri/README.md
```diff
@@ -12,14 +12,15 @@ The thing type ids are defined according to the lighting devices defined for Zig
...
-
+| Extended Colour Light    | 0x0210           | 0210       |
 
 The following matrix lists the capabilities (channels) for each of the supported lighting device types:
 
 | Thing type  | On/Off | Brightness | Color | Color Temperature |
 |-------------|:------:|:----------:|:-----:|:-----------------:|   
 |  0100       |    X   |     X      |       |                   |
 |  0220       |    X   |     X      |       |          X        |
+|  0210       |    X   |            |   X   |          X        |
```
kaikreuzer: I am noticing that this table was already incorrect - neither 0220 nor 0210 should have an X for On/Off...

extensions/binding/org.eclipse.smarthome.binding.tradfri/ESH-INF/thing/thing-types.xml
kaikreuzer: But I don't have a too strong opinion on 200 vs. 210 - if you both now agreed that 210 is the better choice, so be it. We can still revisit this once we see more devices that bring us more clarity on how the gateway handles them.

hreichert: Well, @cweitkamp has also good arguments.
I'm absolutely against removing the color_temperature channel from the new thing. See my discussion above; but replace "brightness" with "color temperature". And this can not be handled with the color channel, at least not without rules or something.

extensions/binding/org.eclipse.smarthome.binding.tradfri/README.md
hreichert: Very good point, as there is no "onoff" channel. Should I open a new issue for this? I want to extend the documentation anyway with more examples.

kaikreuzer: You could simply remove the column as part of this PR - or we first merge and you can come up with a new PR that enhances the documentation in general.

hreichert: I vote for option 2 - new PR when the enhancements are done.

kaikreuzer: ok!

PR #4275, opened 2017-09-15
Include Null annotation usage in Coding Guidelines

triller-telekom:


kaikreuzer:
Shouldn't you also adapt the archetype accordingly?

triller-telekom:
Done.

maggu2810:
Thanks @triller-telekom

PR #4276, opened 2017-09-15
drop obsolete link resources from ThingResource

### LINKED ISSUE #4273: [Rest] DELETE /rest/things/*/channels/*/link
As reported in https://github.com/openhab/openhab-core/issues/209 the rest api offers an endpoint `DELETE /rest/things/*/channels/*/link` which does not really function also it seems to be duplicate compared to `DELETE /rest/links/*/*` and the implementations differ.

Maybe they (there is also an add endpoint) should be removed otherwise there some issues which I quickly spotted:

* Wrong annotation for itemName, that is why it's not shown in Swagger, also its not validated in anyway before passing it to the internal method, so I see no way it could properly report if the link is not found or something
* Missing else if the link does not exist, so you would at least have some feedback, imho that could/should also be a 404
* The implementation does not match the one in https://github.com/eclipse/smarthome/blob/master/bundles/io/org.eclipse.smarthome.io.rest.core/src/main/java/org/eclipse/smarthome/io/rest/core/link/ItemChannelLinkResource.java#L101 but accepts the same parameters, can / should they share some code, should they contain the same code or as suggested should this one be removed?

Regards

sjka:
...as they were replaced by the ItemChannelLinkResource.

Removing them should not do any harm as they were broken anyway (see #4273).

fixes #4273

### DIFF
--- a/bundles/io/org.eclipse.smarthome.io.rest.core/src/main/java/org/eclipse/smarthome/io/rest/core/thing/ThingResource.java
+++ b/bundles/io/org.eclipse.smarthome.io.rest.core/src/main/java/org/eclipse/smarthome/io/rest/core/thing/ThingResource.java
@@ -49,9 +49,7 @@ import org.eclipse.smarthome.config.core.status.ConfigStatusInfo;
 import org.eclipse.smarthome.config.core.status.ConfigStatusService;
 import org.eclipse.smarthome.config.core.validation.ConfigValidationException;
 import org.eclipse.smarthome.core.auth.Role;
-import org.eclipse.smarthome.core.items.GenericItem;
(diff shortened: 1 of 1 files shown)


sjka:
PS: Is it "API-Breaking" if we break a broken (in the sense of defective) API?

martinvw:
Thanks for picking it up so quickly 👌 

> PS: Is it "API-Breaking" if we break a broken (in the sense of defective) API?

😀

sjka:
As far a I can see it suffers from the same issue: There is an "itemName" method parameter which is not part of the path spec, therefore it cannot hold any sensible value. So even if the rest would work, the item creation at the bottom of the method is broken.

PR #4277, opened 2017-09-15
PaperUI: Fix update thing when no config-description is given.

### LINKED ISSUE #4265: [Paper UI] Unable to update a Thing when its type has no configuration description
When a Thing Type does not have a `<config-description>` XML block, you can not update any Things of this type in Paper UI. An example of this are the Nest Binding [Thing types](https://github.com/openhab/openhab2-addons/tree/master/addons/binding/org.openhab.binding.nest/ESH-INF/thing) (except for the type defined in bridge.xml).

Because of this it is not possible to update following Thing properties for such Thing Types:
* Name
* Location
* Bridge

Reproduction scenario:
1. Open Paper UI
2. Add Nest Binding and add a Nest Account
3. Let it detect Nest devices and add these via "Inbox"
4. Now open "Configuration -> Things"
5. Click on a Thing (e.g. a Cam, Protect or Thermostat) to open its details
6. Click the edit configuration button on top of the screen
7. Update the "Name" and "Location"
8. Now when you click the "Checkmark" on top of the screen nothing happens and the changes are not applied.

When you monitor the network traffic via the browser debug console in step 7, it shows a 404 not found occurs for a request to `/rest/config-descriptions`.

When I recompile the code and give these thing types a `<config-description>` XML block it is possible to update these Things.

htreu:
Fixes #4265.

### DIFF
--- a/extensions/ui/org.eclipse.smarthome.ui.paper/web-src/js/services.js
+++ b/extensions/ui/org.eclipse.smarthome.ui.paper/web-src/js/services.js
@@ -453,6 +453,9 @@ angular.module('PaperUI.services', [ 'PaperUI.services.repositories', 'PaperUI.c
             }
         },
         setConfigDefaults : function(originalConfiguration, groups, sending) {
+            if (!groups) { // no config groups available for this configuration
+                return originalConfiguration;
+            }
             var configuration = {};
             angular.copy(originalConfiguration, configuration);
             for (var i = 0; i < groups.length; i++) {


extensions/ui/org.eclipse.smarthome.ui.paper/web-src/js/services.js
```diff
@@ -453,6 +453,9 @@ angular.module('PaperUI.services', [ 'PaperUI.services.repositories', 'PaperUI.c
             }
         },
         setConfigDefaults : function(originalConfiguration, groups, sending) {
+            if (!groups) { // no config groups available for this configuration
+                return {};
```
sjka: hmmm, just a wild guess from looking at the function signature (so I might be completely wrong), but shouldn't it return `originalConfiguration` in this case?

htreu: good catch. Since originalConfiguration is always {} at this point it technically does not matter. but it looks awful this way.

kaikreuzer:
Before merging this, I would actually like to understand if this is at all a valid situation, see https://github.com/eclipse/smarthome/issues/4265#issuecomment-329886798.

kaikreuzer:
Technically, I think this fix is indeed required as in theory a thing type does not require to provide meta-data for its configuration (while a configuration still is mandatory) - so ok to merge.

PR #4280, opened 2017-09-16
added nullable imports for ESH bundles

maggu2810:


maggu2810:
@kaikreuzer Can we merge this one (soon)?
We agreed to use nullness annotations. Base classes and interfaces started using `@NonNullByDefault`. The optional import should be okay for all bundles, we will not see any "missing in classpath" in the IDE and we don't get merge conflicts because one PR is merged in front of the others (that fixes some nullness stuff).
Or do you see any problem here?

maggu2810:
This one has been done manually. For the classpath file of all projects because of external annotations I have written a script... But that is another topic we first need to agree and out of scope of this PR :wink:

PR #4285, opened 2017-09-17
Add missing details from Items article

ThomDietrich:
Besides cleaning up markdown, I've added ImageItem, LocationItem and StringType

Ready to be merged.

Question: Shouldn't we add javadoc links for clarity? In my personal opinion documentation like this article is nice but can hardly replace the facts a professional can derive from code...

Signed-off-by: Thomas Dietrich <Thomas.Dietrich@tu-ilmenau.de>


kaikreuzer:
Thanks!

> Shouldn't we add javadoc links for clarity?

This is under "Concepts" menu, not under "Development"; so I would tend to say the main audience are not necessarily developers, so JavaDoc links might be a bit too much for it.

Issue #4286, opened 2017-09-17
Implement a OneWire binding


J-N-K:
OneWire devices are a perfect use case for the new thing concept in ESH. There should be a binding for that.

PR #4287, opened 2017-09-17
Feature: onewire binding

J-N-K:
Implements issue #4286. This is just to keep mergability checks in place. Still needs a lot of work before it is ready to merge, although it is working at my place.

- [x] implement basic functions
- [x] check units in channels
- [x] add tests
- [x] add code documentation
- [x] add user documentation


J-N-K:
@sjka, @maggu2810, @kaikreuzer, I would like to hear comments on the general design of this binding. I‘m aware of the fact that this is not in a mergable state at the moment and IMO it‘s too early for a complete review, but if there are severe objections against the overall design, it makes no sense to refine the code. Thanks in advance.

sjka:
Hey @J-N-K,

generally it makes much sense to have such a binding IMHO, therefore thanks a lot for putting all that effort in it!

* The modeling with the owserver as a bridge and all the sensors as individual child things certainly makes sense. Especially because some of the things can have a dynamic amount of channels themselves again. Also I really like the clean separation between the handler(s) and the client which connects to the server very much!

* I noticed a `scale` configuration parameter in several thing-types. While this so far is the only way of doing it "by hand", you actually might want to closely watch #4818 - it is really close to being merged, therefore I think it's worth waiting for it. It saves you some effort as you don't need to take care of configuring and converting the units within the binding anymore and this soon is going to be "the way" of doing it.

* I don't really get the sense of the abstract `OwBaseBridgeHandler` class - as there is only one implementation anyway, just merge it with `OwserverHandler` for now. Or did I miss anything?

* Your `OwBaseBridgeHandler` exposes lots of methods which simply delegate to the `OwServerConnection`. I think it would be cleaner to directly hand out the `OwServerConnection` as a "client" to the corresponding thing handler.

* Don't "remember" the `lastDigitalState` - whenever you've got "fresh" information, send it out as a state update. Leave optimizations to the framework. The rationale behind this is that is sometimes is important how old an information is - so when you cache the old state and only send updates when it changes it gets impossible to tell whether the current item state is stale or _really_ the most recent information.

* In the Digital I/O devices, you dynamically create and remove channels based on the thing configuration. While this is a very valid option for scenarios where device can describe itself, I think there is a nicer way of doing this now. As the user (currently by configuration) anyway tells the system what is there and what is not, why don't you simply define the 1 (with channel config for i/o) or 4 channel-types (I/O x Inverted) and declare on the corresponding thing-types that they can be "extended" by these (see [docu](https://www.eclipse.org/smarthome/documentation/development/bindings/xml-reference.html#xml-structure-for-thing-descriptions) or the [Magic Binding](https://github.com/eclipse/smarthome/blob/master/bundles/test/org.eclipse.smarthome.magic/ESH-INF/thing/thing-types.xml) as an example). This would eliminate lots of ugly initialization code for the dynamics in the handlers.

* I assume this is one of the things you would have cleaned up anyway, but please avoid logging warning like in `OwServerConnection` currently - ideally it would be on debug/trace level and the user would simply see the thing going OFFLINE/COMMUNICATION_ERROR with the corresponding error message.

Now this got longer than I wanted to and for sure my amount of feedback looks worse than it is meant to be. Your code (once again 😄 ) looks pretty well structured (also unlike what I expected from the ~5k lines of code) and I'm really looking forward to having this binding! Keep up the good work 👍

J-N-K:
@sjka, thanks for your comments. 

- I think that probably #4818 and units will be merged before this one, so I'll change to that after it is available. For the time being, I'll stick with this solution, as I use the bidning at my home.
- In principle other bridges could be added (e.g. a direct serial or USB bridge as suggested in #4801). I tried to define necessary methods to use the devices and force a bridge implementation to provide these by making them abstract in the `OwBaseBridgeHandler`. This is alos the reason I did not expose the `owServerConnection` to the devices, as it surely will not be available in other bridge types.
- I'll have a look at the extensible-option. From a quick look, I'm not sure how to make sure the user is not adding more channels than available and how to preserve the channel order if something changed from in to out. 

I'll now continue to refine and document the code and add tests.

htreu:
@J-N-K, @sjka iirc the `scale` setting is about the precision of the temperature sensor which is used during measurement.
This is an excerpt from the data sheet ([DS18B20 - PDF](https://datasheets.maximintegrated.com/en/ds/DS18B20.pdf)):
```
The resolution of the temperature
sensor is user-configurable to 9, 10, 11, or 12 bits,
corresponding to increments of 0.5°C, 0.25°C, 0.125°C,
and 0.0625°C, respectively. The default resolution at
power-up is 12-bit. 
```

I would expect the binding to provide this setting as its not quantity or unit specific.
Another thing from the data sheet:

```
The DS18B20 output temperature data is calibrated in
degrees Celsius; for Fahrenheit applications, a lookup
table or conversion routine must be used.
```

So the values are always provided in °C. When QuantityType lands the state updates can be given in °C.

J-N-K:
From my point this is ready for a first review. Quite sure this will take some time, It's 7k loc.

htreu:
Thanks for the ping guys. I will start with a review soon(TM).

extensions/binding/org.eclipse.smarthome.binding.onewire/README.md
```diff
@@ -0,0 +1,183 @@
...
+For best performance it is recommended to set the resolution only as high as needed.
+
+The second parameter is the `offset`.
+Onewire sensors are usually well calibrated but depending on the mounting position of the sensor it may be necessary to apply an offset.
+The `offset` parameter consists of the value itself and it's unit (e.g. `-1.0 °C`).
+ 
+ 
+## Full Example
+
+text
```
htreu: please provide a full example.

extensions/binding/org.eclipse.smarthome.binding.onewire/README.md
```diff
@@ -0,0 +1,183 @@
...
+
+### demo.things:
```
htreu: It would be nice to also see a more complex example, maybe with additional channel options and multiple sensor ids.

extensions/binding/org.eclipse.smarthome.binding.onewire/src/main/java/org/eclipse/smarthome/binding/onewire/internal/DigitalIoLogic.java
```diff
@@ -0,0 +1,26 @@
...
+ * http://www.eclipse.org/legal/epl-2.0
+ *
+ * SPDX-License-Identifier: EPL-2.0
+ */
+package org.eclipse.smarthome.binding.onewire.internal;
+
+import org.eclipse.jdt.annotation.NonNullByDefault;
+
+/**
+ * The {@link DigitalIoLogic} provides the direction and mode of a digital IO channel
```
htreu: JavaDoc copied from DigitalIoMode?

extensions/binding/org.eclipse.smarthome.binding.onewire/src/main/java/org/eclipse/smarthome/binding/onewire/internal/OwHandlerFactory.java
```diff
@@ -0,0 +1,109 @@
...
+import org.osgi.service.component.annotations.Reference;
+
+/**
+ * The {@link OwHandlerFactory} is responsible for creating things and thing
+ * handlers.
+ *
+ * @author Jan N. Klug - Initial contribution
+ */
+@NonNullByDefault
+@Component(service = ThingHandlerFactory.class, immediate = true, name = "binding.onewire", configurationPolicy = ConfigurationPolicy.OPTIONAL)
```
htreu: AFAIK the handler factories do not have to be active immediate. See #5511 for reference. Also please add the `configurationPid` property.

extensions/binding/org.eclipse.smarthome.binding.onewire/src/main/java/org/eclipse/smarthome/binding/onewire/internal/Util.java
```diff
@@ -0,0 +1,105 @@
...
+
+    /**
+     * convert measured value to temperature with offset
+     *
+     * @param temperature measured value
+     * @param offset offset to apply
+     * @return originalState if not DecimalType, temperature otherwise
+     * @throws OwException in case of conversion error
+     */
+    public static QuantityType<Temperature> convertTemperature(QuantityType<Temperature> temperature,
```
htreu: The offset handling should in the future be done by the new Offset-Profile. Please see #5679 which provides a system profile which can be configured for any channel. Therefor I suggest you drop the binding specific offset support here.

extensions/binding/org.eclipse.smarthome.binding.onewire/src/main/java/org/eclipse/smarthome/binding/onewire/internal/Util.java
```diff
@@ -0,0 +1,105 @@
...
+        Double aH = relativeHumidity.doubleValue() / 100 * saturationVaporPressure / (461.52 * (273.15 + theta));
+        State absoluteHumidity = new QuantityType<Density>(aH, SmartHomeUnits.KILOGRAM_PER_CUBICMETRE).toUnit("g/m³");
+        if (absoluteHumidity != null) {
+            return absoluteHumidity;
+        } else {
+            throw new IllegalArgumentException("could not change unit");
+        }
+    }
+
+    public static State calculateDewpoint(QuantityType<Temperature> temperature,
```
htreu: Also nice! See my last comment about derived values. No need to change anything right now :-)

extensions/binding/org.eclipse.smarthome.binding.onewire/src/main/java/org/eclipse/smarthome/binding/onewire/internal/Util.java
```diff
@@ -0,0 +1,105 @@
...
+    }
+
+    /**
+     * calculate absolute humidity in g/m³ from measured values
+     *
+     * @param temperature the measured temperature
+     * @param relativeHumidity the measured relative humidity
+     * @return the corresponding absolute humidity
+     */
+    public static State calculateAbsoluteHumidity(QuantityType<Temperature> temperature,
```
htreu: nice. In the light of #4748  maybe we can extract this when a solution is found.

extensions/binding/org.eclipse.smarthome.binding.onewire/src/main/java/org/eclipse/smarthome/binding/onewire/internal/device/AbstractDigitalOwDevice.java
```diff
@@ -0,0 +1,141 @@
...
+import org.eclipse.smarthome.core.library.types.DecimalType;
+import org.eclipse.smarthome.core.library.types.OnOffType;
+import org.eclipse.smarthome.core.thing.Channel;
+import org.eclipse.smarthome.core.thing.Thing;
+import org.eclipse.smarthome.core.types.Command;
+import org.eclipse.smarthome.core.types.State;
+import org.eclipse.smarthome.core.types.StateDescription;
+
+/**
+ * The {@link AbstractDigitalOwDevice} class defines an DS18x20 device
```
htreu: JavaDoc copied from the DS18x20 class?

extensions/binding/org.eclipse.smarthome.binding.onewire/src/main/java/org/eclipse/smarthome/binding/onewire/internal/device/DS18x20.java
```diff
@@ -0,0 +1,88 @@
...
+    }
+
+    @Override
+    public void configureChannels() {
+        Thing thing = callback.getThing();
+        Channel temperatureChannel = thing.getChannel(CHANNEL_TEMPERATURE);
+
+        if (temperatureChannel != null) {
+            Configuration channelConfiguration = temperatureChannel.getConfiguration();
+            if (channelConfiguration.get(CONFIG_OFFSET) != null) {
```
htreu: see my comment about offset handling earlier.

extensions/binding/org.eclipse.smarthome.binding.onewire/src/main/java/org/eclipse/smarthome/binding/onewire/internal/device/DS2438.java
```diff
@@ -0,0 +1,205 @@
...
+            } else {
+                HUMIDITY_PARAMETER.replace(THING_TYPE_OWSERVER, new OwserverDeviceParameter("/humidity"));
+            }
+        }
+
+        Channel temperatureChannel = thing.getChannel(CHANNEL_TEMPERATURE);
+
+        if (temperatureChannel != null) {
+            Configuration channelConfiguration = temperatureChannel.getConfiguration();
+            if (channelConfiguration.get(CONFIG_OFFSET) != null) {
```
htreu: again offset handling.

extensions/binding/org.eclipse.smarthome.binding.onewire/src/main/java/org/eclipse/smarthome/binding/onewire/internal/device/OwDeviceParameter.java
```diff
@@ -0,0 +1,31 @@
...
+import java.util.HashMap;
+
+import org.eclipse.smarthome.core.thing.ThingTypeUID;
+
+/**
+ * The {@link OwDeviceParameter} defines a specific device map
+ *
+ * @author Jan N. Klug - Initial contribution
+ */
+public class OwDeviceParameter extends HashMap<ThingTypeUID, Object> {
```
htreu: Its in general not a good thing to extend the Java internal collections. Are you able to refactor this so the `OwDeviceParameter` holds a map internally and delegates appropriate methods?

extensions/binding/org.eclipse.smarthome.binding.onewire/src/main/java/org/eclipse/smarthome/binding/onewire/internal/discovery/OwDiscoveryItem.java
```diff
@@ -0,0 +1,339 @@
...
+    private final String familyId;
+    private final String sensorId;
+    private OwSensorType sensorType = OwSensorType.UNKNOWN;
+    private String vendor = "generic";
+    private String sensorTypeId = "";
+
+    private String hwRevision = "";
+    private String prodDate = "";
+
+    OwPageBuffer pages = new OwPageBuffer();
```
htreu: can also be private

extensions/binding/org.eclipse.smarthome.binding.onewire/src/main/java/org/eclipse/smarthome/binding/onewire/internal/handler/DigitalIOThingHandler.java
```diff
@@ -0,0 +1,136 @@
...
+        // sensor configuration
+        try {
+            sensors.get(0).configureChannels();
+        } catch (OwException e) {
+            updateStatus(ThingStatus.OFFLINE, ThingStatusDetail.CONFIGURATION_ERROR, e.getMessage());
+            return;
+        }
+
+        validConfig = true;
+        refreshInterval = 10 * 1000;
```
htreu: this overwrites the configuration.

extensions/binding/org.eclipse.smarthome.binding.onewire/src/main/java/org/eclipse/smarthome/binding/onewire/internal/handler/DigitalIOThingHandler.java
```diff
@@ -0,0 +1,136 @@
...
+
+        updateStatus(ThingStatus.UNKNOWN, ThingStatusDetail.NONE);
```
htreu: this overwrites any status set in the `super.initialized` call.

extensions/binding/org.eclipse.smarthome.binding.onewire/src/main/java/org/eclipse/smarthome/binding/onewire/internal/handler/IButtonThingHandler.java
```diff
@@ -0,0 +1,54 @@
...
+    @Override
+    public void initialize() {
+        super.initialize();
+
+        sensors.add(new DS2401(sensorIds.get(0), this));
+
+        refreshInterval = 10 * 1000;
+        validConfig = true;
+
+        updateStatus(ThingStatus.UNKNOWN, ThingStatusDetail.NONE);
```
htreu: this overwrites the status from the `super` call.

extensions/binding/org.eclipse.smarthome.binding.onewire/src/main/java/org/eclipse/smarthome/binding/onewire/internal/handler/IButtonThingHandler.java
```diff
@@ -0,0 +1,54 @@
...
+        super(thing, dynamicStateDescriptionProvider);
+    }
+
+    @Override
+    public void initialize() {
+        super.initialize();
+
+        sensors.add(new DS2401(sensorIds.get(0), this));
+
+        refreshInterval = 10 * 1000;
```
htreu: this overwrites the user config.

extensions/binding/org.eclipse.smarthome.binding.onewire/src/main/java/org/eclipse/smarthome/binding/onewire/internal/handler/TemperatureSensorThingHandler.java
```diff
@@ -0,0 +1,62 @@
...
+    public TemperatureSensorThingHandler(Thing thing,
+            @Nullable OwDynamicStateDescriptionProvider dynamicStateDescriptionProvider) {
+        super(thing, dynamicStateDescriptionProvider);
+    }
+
+    @Override
+    public void initialize() {
+        super.initialize();
+
+        sensors.add(new DS18x20(sensorIds.get(0), this));
```
htreu: `sensorIds.get(0)` might fail if no sensor id was configured for this thing. See also the other Thing handlers where this can cause issues.

edit: I just saw that the framework will protect you in case a required parameter is missing. In case of a "bad" id (like `id="abcd"`)  I actually get the expected `java.lang.IndexOutOfBoundsException: Index: 0, Size: 0`.

extensions/binding/org.eclipse.smarthome.binding.csvlogger/.project
```diff
@@ -0,0 +1,33 @@
+<?xml version="1.0" encoding="UTF-8"?>
```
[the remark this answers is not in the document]
J-N-K: done

extensions/binding/org.eclipse.smarthome.binding.onewire.test/src/test/java/org/eclipse/smarthome/binding/onewire/device/DS2401Test.java
```diff
@@ -0,0 +1,45 @@
...
+    @Test
+    public void presenceTestOn() {
+        presenceTest(OnOffType.ON);
+    }
+
+    @Test
+    public void presenceTestOff() {
+        presenceTest(OnOffType.OFF);
+    }
+}
```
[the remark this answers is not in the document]
J-N-K: done

extensions/binding/org.eclipse.smarthome.binding.onewire/ESH-INF/thing/DigitalIO.xml
```diff
@@ -0,0 +1,129 @@
...
+		<label>Digital I/O</label>
+		<description>A single digital I/O device (DS2405)</description>
+		<channels>
+			<channel id="present" typeId="present" />
+			<channel id="digital0" typeId="dio">
+				<label>Digital I/O 0</label>
+			</channel>
+		</channels>
+		<properties>
+			<property name="modelId">Sensor Model</property>
```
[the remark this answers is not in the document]
J-N-K: I'll answer this only once and will adopt the final result to all occurences. It is not exactly true that this information is useless. E.g. in the case of the two-channel digital I/O two different sensors can be used (DS2405 and DS2413), the same holds true for the temperature sensor (DS18B20, DS18S20). 

And for the MS-TH there might be different manufacturers of the same sensor. I decided to keep it conistent for all things and added the information even where only one or none of the information is useful. Especially for the multisensors this makes it much easier to handle if all things use the same properties.

extensions/binding/org.eclipse.smarthome.binding.onewire/src/main/java/org/eclipse/smarthome/binding/onewire/internal/handler/DigitalIOThingHandler.java
J-N-K: I have made a change here. Please tell me what you think. If ok, I'll change the other occurences, too.

extensions/binding/org.eclipse.smarthome.binding.onewire/src/main/java/org/eclipse/smarthome/binding/onewire/internal/handler/IButtonThingHandler.java
J-N-K: done

extensions/binding/org.eclipse.smarthome.binding.onewire/src/main/java/org/eclipse/smarthome/binding/onewire/internal/handler/TemperatureSensorThingHandler.java
J-N-K: this will be solved by the changes i made above to the DigitalIOThingHandler intialization code (then this will never be called if the ids are invalid).

extensions/binding/org.eclipse.smarthome.binding.onewire/ESH-INF/thing/channel-types.xml
```diff
@@ -0,0 +1,130 @@
...
+	<channel-type id="supplyvoltage">
+		<item-type>Number:Voltage</item-type>
+		<label>Supply Voltage</label>
+		<description></description>
+		<state readOnly="true" pattern="%.1f %unit%" />
+	</channel-type>
+	<!-- Light Channel -->
+	<channel-type id="light">
+		<item-type>Number:Illuminance</item-type>
+		<label>Lightness</label>
```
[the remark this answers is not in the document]
J-N-K: I was thinking about that, too. However the german "Helligkeit" which is used in the documentation of the manufacturer is translated to "lightness" according to dict.leo.org. iButtonLink calls it "light level" sensor. I can change it to whatever you think fits best.

extensions/binding/org.eclipse.smarthome.binding.onewire/ESH-INF/thing/iButton.xml
```diff
@@ -0,0 +1,34 @@
...
+			<property name="thingTypeVersion">1</property>
+		</properties>
+		<config-description>
+			<parameter name="id" type="text">
+				<context>network_address</context>
+				<label>Sensor ID</label>
+				<description>Sensor ID in format: xx.xxxxxxxxxxxx)</description>
+				<required>true</required>
+			</parameter>
+			<parameter name="refresh" type="integer" min="0">
```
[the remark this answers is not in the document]
J-N-K: changed in handler code

extensions/binding/org.eclipse.smarthome.binding.onewire/ESH-INF/thing/DigitalIO.xml
```diff
@@ -0,0 +1,129 @@
+<?xml version="1.0" encoding="UTF-8"?>
+<thing:thing-descriptions bindingId="onewire" xmlns:xsi="http://www.w3.org/2001/XMLSchema-instance" xmlns:thing="http://eclipse.org/smarthome/schemas/thing-description/v1.0.0" xsi:schemaLocation="http://eclipse.org/smarthome/schemas/thing-description/v1.0.0 http://eclipse.org/smarthome/schemas/thing-description-1.0.0.xsd">
+	<thing-type id="digitalio">
+		<supported-bridge-type-refs>
+			<bridge-type-ref id="owserver" />
+		</supported-bridge-type-refs>
+		<label>Digital I/O</label>
+		<description>A single digital I/O device (DS2405)</description>
+		<channels>
+			<channel id="present" typeId="present" />
```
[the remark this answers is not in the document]
J-N-K: Sometimes 1W busses tend to be instable (depending e.g. on cabling). If this happens the sensors seem to vanish and reappear later. This can be easliy monitored by this state. If it is possible to use persistence services with thing status, I can remove it.

extensions/binding/org.eclipse.smarthome.binding.onewire/META-INF/MANIFEST.MF
```diff
@@ -0,0 +1,28 @@
...
+ org.eclipse.smarthome.core.library.types,
+ org.eclipse.smarthome.core.library.unit,
+ org.eclipse.smarthome.core.thing,
+ org.eclipse.smarthome.core.thing.binding,
+ org.eclipse.smarthome.core.thing.binding.builder,
+ org.eclipse.smarthome.core.thing.type,
+ org.eclipse.smarthome.core.types,
+ org.osgi.framework,
+ org.slf4j,
+ tec.uom.se.unit
```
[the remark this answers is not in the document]
J-N-K: removed

extensions/binding/org.eclipse.smarthome.binding.onewire/README.md
```diff
@@ -0,0 +1,183 @@
...
+Discovery is supported for things. You have to add the bridges manually.  
+
+## Thing Configuration
+
+It is strongly recommended to use discovery and PaperUI for thing configuration.
+Please note that:
+
+* All things need a bridge.
+* The sensor id parameter supports only the dotted format, including the family id (e.g. `28.7AA256050000`).
+* Refresh time is the minimum time in seconds between to checks of that thing.
```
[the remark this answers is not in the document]
J-N-K: two

extensions/binding/org.eclipse.smarthome.binding.onewire/README.md
```diff
@@ -0,0 +1,183 @@
...
+It defaults to 300s for analog channels and 10s for digital channels.
+* All things (except the bridges) provide presence detection via the `present`channel.
```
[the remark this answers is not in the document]
J-N-K: done

extensions/binding/org.eclipse.smarthome.binding.onewire/README.md
```diff
@@ -0,0 +1,183 @@
...
+
+There are no configuration options for the owserver besides the network address.
+It consists of two parts: `address` and `port`.
+
+The `address` parameter is used to denote the location of the owserver instance. 
+It supports both, a hostname or an IP address. 
+It defaults to `localhost`.
+
+The `port` parameter is used to adjust non-standard OWFS installations.
+It defaults to `4304`, which is normally correct.
```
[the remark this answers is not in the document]
J-N-K: reworded

extensions/binding/org.eclipse.smarthome.binding.onewire/README.md
J-N-K: done

extensions/binding/org.eclipse.smarthome.binding.onewire/README.md
J-N-K: I'll have a look at that. Since I don't use it, I'll have to figure out how this works in textual configuration.

extensions/binding/org.eclipse.smarthome.binding.onewire/src/main/java/org/eclipse/smarthome/binding/onewire/internal/DigitalIoLogic.java
J-N-K: done

extensions/binding/org.eclipse.smarthome.binding.onewire/src/main/java/org/eclipse/smarthome/binding/onewire/internal/OwHandlerFactory.java
J-N-K: done

extensions/binding/org.eclipse.smarthome.binding.onewire/src/main/java/org/eclipse/smarthome/binding/onewire/internal/device/AbstractDigitalOwDevice.java
J-N-K: done

extensions/binding/org.eclipse.smarthome.binding.onewire/src/main/java/org/eclipse/smarthome/binding/onewire/internal/discovery/OwDiscoveryItem.java
J-N-K: done

extensions/binding/org.eclipse.smarthome.binding.onewire/src/main/java/org/eclipse/smarthome/binding/onewire/internal/Util.java
J-N-K: Do you have an ETA for merging? I'll remove it as soon as this is merged.

J-N-K:
I left some comments and included most of yours in my latest commit. Regarding the `OwDeviceParameter` refactoring: can be done, will take until the weekend. I'm a bit busy ATM.

extensions/binding/org.eclipse.smarthome.binding.onewire/src/main/java/org/eclipse/smarthome/binding/onewire/internal/device/OwDeviceParameter.java
J-N-K: done

extensions/binding/org.eclipse.smarthome.binding.onewire/ESH-INF/thing/DigitalIO.xml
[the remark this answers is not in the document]
htreu: The thing status is always a live status. Why should it be persisted? Rules are already able to react on the thing status. On restart the thing status will again be set by each corresponding thing handler.

extensions/binding/org.eclipse.smarthome.binding.onewire/ESH-INF/thing/DigitalIO.xml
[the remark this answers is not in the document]
htreu: If thing handlers always do overwrite this placeholders its okay for me.

extensions/binding/org.eclipse.smarthome.binding.onewire/ESH-INF/thing/channel-types.xml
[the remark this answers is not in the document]
htreu: Since this seems to be a physical measurement value I suggest to use "Illuminance" or may be "Luminosity" (see https://dict.leo.org/englisch-deutsch/Lichtst%C3%A4rke). Is it measured in candela here? You may also provide a state description pattern.

extensions/binding/org.eclipse.smarthome.binding.onewire/src/main/java/org/eclipse/smarthome/binding/onewire/internal/Util.java
htreu: @J-N-K we are about to merge #5679. You can remove the binding specific offset handling.

extensions/binding/org.eclipse.smarthome.binding.onewire/ESH-INF/thing/channel-types.xml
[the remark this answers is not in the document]
J-N-K: I changed it to illuminance since the measured value is in lux. State pattern added.

htreu:
@J-N-K please have a look at the Travis output:

```
Results :
Failed tests: 
  IButtonThingHandlerTest.testInitializationEndsWithUnknown:71->JavaTest.waitForAssert:69->JavaTest.waitForAssert:116->JavaTest.waitForAssert:154->lambda$0:71 expected:<UNKNOWN> but was:<UNINITIALIZED>
  DigitalIOThingHandlerTest.testInitializationEndsWithUnknown:76->JavaTest.waitForAssert:69->JavaTest.waitForAssert:116->JavaTest.waitForAssert:154->lambda$0:76 expected:<UNKNOWN> but was:<UNINITIALIZED>
  TemperatureSensorThingHandlerTest.testInitializationEndsWithUnknown:78->JavaTest.waitForAssert:69->JavaTest.waitForAssert:116->JavaTest.waitForAssert:154->lambda$0:78 expected:<UNKNOWN> but was:<UNINITIALIZED>
Tests in error: 
  IButtonThingHandlerTest.testRefresh:77 » IndexOutOfBounds Index: 0, Size: 0
  DigitalIOThingHandlerTest.testRefresh:82 » IndexOutOfBounds Index: 0, Size: 0
  TemperatureSensorThingHandlerTest.testRefresh:84 » IndexOutOfBounds Index: 0, ...
Tests run: 44, Failures: 3, Errors: 3, Skipped: 0
```

extensions/binding/org.eclipse.smarthome.binding.onewire/src/main/java/org/eclipse/smarthome/binding/onewire/internal/handler/OwBaseThingHandler.java
```diff
@@ -0,0 +1,264 @@
...
+    protected final List<AbstractOwDevice> sensors = new ArrayList<AbstractOwDevice>();
+    protected final List<String> sensorIds = new ArrayList<String>();
+    protected long lastRefresh = 0;
+    protected long refreshInterval = 300 * 1000;
+    protected boolean validConfig = false;
+
+    protected @Nullable OwDynamicStateDescriptionProvider dynamicStateDescriptionProvider;
+
+    public OwBaseThingHandler(Thing thing,
+            @Nullable OwDynamicStateDescriptionProvider dynamicStateDescriptionProvider) {
```
triller-telekom: This is a static mandatory reference in your factory, so how can it be `null` here?

extensions/binding/org.eclipse.smarthome.binding.onewire/src/main/java/org/eclipse/smarthome/binding/onewire/internal/handler/OwBaseBridgeHandler.java
```diff
@@ -0,0 +1,162 @@
...
+                        } else {
+                            logger.trace("{} not initialized, skipping refresh", owThing.getLabel());
+                        }
+                    } else {
+                        logger.debug("{} handler missing", owThing.getLabel());
+                    }
+                    thingCount--;
+                }
+            }
+        } catch (RuntimeException e) {
```
triller-telekom: Please do not catch `RuntimeException` directly, this will also accidentally catch `NullpointerException`s.... Please catch specific ones or throw properly typed Exceptions, which is even better...

extensions/binding/org.eclipse.smarthome.binding.onewire/src/main/java/org/eclipse/smarthome/binding/onewire/internal/handler/OwBaseBridgeHandler.java
```diff
@@ -0,0 +1,162 @@
...
+            schedulerAborted(e);
+        }
+    };
+
+    protected ScheduledFuture<?> refreshTask = scheduler.scheduleWithFixedDelay(owRefresher, 1, 1000,
```
triller-telekom: Do you want to make these `1000` milliseconds configurable?

extensions/binding/org.eclipse.smarthome.binding.onewire/src/main/java/org/eclipse/smarthome/binding/onewire/internal/handler/OwBaseBridgeHandler.java
```diff
@@ -0,0 +1,162 @@
...
+    private void schedulerAborted(RuntimeException e) {
+        logger.debug("bridge {}: refresh scheduler crashed: {}", getThing().getUID(), e.getMessage());
+        refreshTask.cancel(false);
+        refreshTask = scheduler.scheduleWithFixedDelay(owRefresher, 1, 1000, TimeUnit.MILLISECONDS);
+    }
+
+    @Override
+    public void dispose() {
+        refreshable = false;
+        refreshTask.cancel(false);
```
triller-telekom: Please check if it is non null and running before you cancel it.

extensions/binding/org.eclipse.smarthome.binding.onewire/src/main/java/org/eclipse/smarthome/binding/onewire/internal/OwException.java
```diff
@@ -0,0 +1,26 @@
...
+ * SPDX-License-Identifier: EPL-2.0
+ */
+package org.eclipse.smarthome.binding.onewire.internal;
+
+/**
+ * The {@link OwException} class defines an exception for handling OneWireExceptions
+ *
+ * @author Jan N. Klug - Initial contribution
+ */
+public class OwException extends RuntimeException {
```
triller-telekom: Is there any particular reason why you made this a `RuntimeException` instead of a regular one? I would prefer this to be a regular exception.

extensions/binding/org.eclipse.smarthome.binding.onewire/ESH-INF/thing/DigitalIO.xml
```diff
@@ -0,0 +1,129 @@
...
+				<label>Digital I/O 0</label>
+			</channel>
+		</channels>
+		<properties>
+			<property name="modelId">Sensor Model</property>
+			<property name="vendor">Vendor</property>
+			<property name="sensorCount">1</property>
+			<property name="thingTypeVersion">1</property>
+		</properties>
+		<config-description>
```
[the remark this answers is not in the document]
J-N-K: `thingTypeVersion` may differ for different things in the future. The idea is that the binding can auto-update things if futre versions of the binding add properties or channels that are not present in old versions. Is it possible to override properties in a thing-type (e.g. the `sensorCount` could default to `1` in the "basething" but overridden in some things).

extensions/binding/org.eclipse.smarthome.binding.onewire/src/main/java/org/eclipse/smarthome/binding/onewire/internal/handler/OwBaseThingHandler.java
J-N-K: Don't know. Will check.

extensions/binding/org.eclipse.smarthome.binding.onewire/ESH-INF/thing/DigitalIO.xml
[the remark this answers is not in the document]
triller-telekom: Unfortunately you are right in the point that properties cannot be overridden if you reference the config-description. But I am wondering why these properties are part of the config-description and not part of the thing-type directly?

Updating thing-types due to a new binding version, such ass adding new channels should not be done within a binding itself, but we should rather aim for a framework wide implementation for all bindings, see https://github.com/eclipse/smarthome/pull/5762#pullrequestreview-132326757 (last sentence was about adding a new channel).

extensions/binding/org.eclipse.smarthome.binding.onewire/src/main/java/org/eclipse/smarthome/binding/onewire/internal/OwException.java
J-N-K: Changed. Improved Exception handling.

extensions/binding/org.eclipse.smarthome.binding.onewire/src/main/java/org/eclipse/smarthome/binding/onewire/internal/handler/OwBaseBridgeHandler.java
J-N-K: IDE complains about null-check (redundant null-check), isCancelled checked.

extensions/binding/org.eclipse.smarthome.binding.onewire/about.html
```diff
@@ -0,0 +1,27 @@
+<!DOCTYPE html PUBLIC "-//W3C//DTD XHTML 1.0 Strict//EN"
+        "http://www.w3.org/TR/xhtml1/DTD/xhtml1-strict.dtd">
+<html xmlns="http://www.w3.org/1999/xhtml">
+<head>
+    <meta http-equiv="Content-Type" content="text/html; charset=ISO-8859-1"/>
+    <title>About</title>
+</head>
+<body lang="EN-US">
+<h2>About This Content</h2>
+<p>June 5, 2006</p>
```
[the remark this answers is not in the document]
J-N-K: done

extensions/binding/org.eclipse.smarthome.binding.onewire.test/.classpath
```diff
@@ -0,0 +1,7 @@
+<?xml version="1.0" encoding="UTF-8"?>
+<classpath>
+	<classpathentry kind="con" path="org.eclipse.jdt.launching.JRE_CONTAINER/org.eclipse.jdt.internal.debug.ui.launcher.StandardVMType/JavaSE-1.7"/>
```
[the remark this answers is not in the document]
J-N-K: done

extensions/binding/org.eclipse.smarthome.binding.onewire.test/about.html
```diff
@@ -0,0 +1,27 @@
+<!DOCTYPE html PUBLIC "-//W3C//DTD XHTML 1.0 Strict//EN"
+        "http://www.w3.org/TR/xhtml1/DTD/xhtml1-strict.dtd">
+<html xmlns="http://www.w3.org/1999/xhtml">
+<head>
+    <meta http-equiv="Content-Type" content="text/html; charset=ISO-8859-1"/>
+    <title>About</title>
+</head>
+<body lang="EN-US">
+<h2>About This Content</h2>
+<p>June 5, 2006</p>
```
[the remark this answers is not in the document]
J-N-K: done

extensions/binding/org.eclipse.smarthome.binding.onewire/src/main/java/org/eclipse/smarthome/binding/onewire/internal/handler/OwBaseBridgeHandler.java
J-N-K: I removed the check. It never happened again after I changed the OwException handling.

extensions/binding/org.eclipse.smarthome.binding.onewire/ESH-INF/thing/DigitalIO.xml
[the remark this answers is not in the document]
J-N-K: I don't understand the first part. They are not part of the config-description as far as I can see? I removed the `thingTypeVersion` and the update code.

extensions/binding/org.eclipse.smarthome.binding.onewire.test/pom.xml
```diff
@@ -0,0 +1,140 @@
...
+              <level>4</level>
+              <autoStart>true</autoStart>
+            </bundle>
+          </bundleStartLevel>
+        </configuration>
+      </plugin>
+    </plugins>
+
+   </build>
+
```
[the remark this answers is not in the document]
J-N-K: done

extensions/binding/org.eclipse.smarthome.binding.onewire.test/pom.xml
```diff
@@ -0,0 +1,140 @@
...
+            <bundle>
+              <id>org.eclipse.smarthome.core.binding.xml</id>
+              <level>4</level>
+              <autoStart>true</autoStart>
+            </bundle>
+          </bundleStartLevel>
+        </configuration>
+      </plugin>
+    </plugins>
+
```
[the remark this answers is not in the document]
J-N-K: done

extensions/binding/org.eclipse.smarthome.binding.onewire.test/pom.xml
```diff
@@ -0,0 +1,140 @@
...
+  <version>0.10.0-SNAPSHOT</version>
+  <packaging>eclipse-test-plugin</packaging>
+
+  <name>Eclipse Smarthome OneWire Binding Tests</name>
+
+  <properties>
+    <bundle.symbolicName>org.eclipse.smarthome.binding.onewire.test</bundle.symbolicName>
+    <bundle.namespace>org.eclipse.smarthome.binding.onewire</bundle.namespace>
+  </properties>
+
```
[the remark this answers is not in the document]
J-N-K: done

extensions/binding/org.eclipse.smarthome.binding.onewire/src/main/java/org/eclipse/smarthome/binding/onewire/internal/handler/DigitalIOThingHandler.java
J-N-K: @htreu, could you comment on this?

triller-telekom:
Thank you for your changes!

Could you please also update your tests? They seem to be broken because your `OWException` is now an `Exception` and has to be caught :)

triller-telekom:
Thank you for fixing the tests. The code looks pretty good now.

The only point which is left for me is the one raised by @htreu about the "present" channel of all things, see https://github.com/eclipse/smarthome/pull/4287#discussion_r197129889 What your use-case needs is essentially a possibility to persist the Thing ONLINE/OFFLINE information, not sure if this should be done as a channel...

I would like to hear the opinion from @kaikreuzer about that one. Otherwise I am fine with the binding. @htreu has even tested it and reported it to work.

extensions/binding/org.eclipse.smarthome.binding.onewire/src/main/java/org/eclipse/smarthome/binding/onewire/internal/handler/OwBaseThingHandler.java
J-N-K: fixed

kaikreuzer:
@J-N-K I agree with @htreu and @triller-telekom that the `present` channel does not make sense and should be removed. If you as a user need this information, you can track online/offline changes in rules (for ANY thing, not just for 1wire things) and do whatever you think is appropriate (incrementing counters, updating other item states, sending notifications, etc.).

extensions/binding/org.eclipse.smarthome.binding.onewire/src/main/java/org/eclipse/smarthome/binding/onewire/internal/handler/OwBaseThingHandler.java
```diff
@@ -0,0 +1,267 @@
...
+    }
+
+    @Override
+    public void initialize() {
+        configure();
+    }
+
+    protected boolean configure() {
+        Configuration configuration = getConfig();
+        Map<String, String> properties = editProperties();
```
kaikreuzer: When you call `editProperties` it is expected that you also call `updateProperties`, which you don't.
To simply access current properties, you could simply do `thing.getProperties()` instead.

extensions/binding/org.eclipse.smarthome.binding.onewire/src/main/java/org/eclipse/smarthome/binding/onewire/internal/handler/OwBaseThingHandler.java
```diff
@@ -0,0 +1,267 @@
...
+    /**
+     * update properties of the sensor if missing
+     *
+     */
+    protected void updateSensorProperties() {
+        Map<String, String> properties = editProperties();
+
+        Bridge bridge = getBridge();
+        if (bridge == null) {
+            logger.info("updating thing properties failed, no bridge available");
```
kaikreuzer: reduce to debug

extensions/binding/org.eclipse.smarthome.binding.onewire/src/main/java/org/eclipse/smarthome/binding/onewire/internal/owserver/OwserverConnection.java
```diff
@@ -0,0 +1,452 @@
...
+        OwserverPacket requestPacket = new OwserverPacket(OwserverMessageType.DIR, "/");
+
+        write(requestPacket);
+
+        OwserverPacket returnPacket = null;
+        do {
+            try {
+                returnPacket = read(false);
+            } catch (OwException e) {
+                logger.info("getDirectory may have returned incomplete result: {}", e.getMessage());
```
kaikreuzer: Is that a software issue? Then better do WARN and include the exception as a last param.

extensions/binding/org.eclipse.smarthome.binding.onewire/src/main/java/org/eclipse/smarthome/binding/onewire/internal/owserver/OwserverConnection.java
```diff
@@ -0,0 +1,452 @@
...
+    /**
+     * close the connection to the owserver instance
+     */
+    private void close() {
+        if (owserverSocket != null) {
+            try {
+                owserverSocket.close();
+            } catch (IOException e) {
+                owserverConnectionState = OwserverConnectionState.FAILED;
+                logger.warn("could not close connection: {}", e.getMessage());
```
kaikreuzer: Issue to be reported? Then include the exception.

extensions/binding/org.eclipse.smarthome.binding.onewire/src/main/java/org/eclipse/smarthome/binding/onewire/internal/owserver/OwserverPacket.java
```diff
@@ -0,0 +1,330 @@
...
+
+/**
+ * The {@link OwserverPacket} class provides a single packet for communication with the owserver
+ *
+ * @author Jan N. Klug - Initial contribution
+ */
+
+@NonNullByDefault
+public class OwserverPacket {
+    protected Logger logger = LoggerFactory.getLogger(OwserverPacket.class);
```
kaikreuzer: add "final"

kaikreuzer: and are you sure that you want it protected and not private?

extensions/binding/pom.xml
```diff
@@ -34,6 +34,8 @@
     <module>org.eclipse.smarthome.binding.lirc</module>
     <module>org.eclipse.smarthome.binding.mqtt</module>
     <module>org.eclipse.smarthome.binding.mqtt.test</module>
+    <module>org.eclipse.smarthome.binding.onewire</module>
```
kaikreuzer: Please also add the binding to:
- https://github.com/eclipse/smarthome/blob/master/features/karaf/esh-ext/src/main/feature/feature.xml
- https://github.com/eclipse/smarthome/blob/master/features/org.eclipse.smarthome.feature.runtime.binding/feature.xml
- https://github.com/eclipse/smarthome/blob/master/docs/_includes/documentation-menu.html

kaikreuzer: In my personal alphabet, "o" comes after "n" ;)

extensions/binding/org.eclipse.smarthome.binding.onewire.test/about.html
```diff
@@ -0,0 +1,27 @@
+<!DOCTYPE html PUBLIC "-//W3C//DTD XHTML 1.0 Strict//EN"
```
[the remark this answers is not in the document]
J-N-K: Could you provide an example for that? I can't find any in the existing bindings.

extensions/binding/org.eclipse.smarthome.binding.onewire/ESH-INF/thing/common.xml
```diff
@@ -0,0 +1,125 @@
...
+				<label>Humidity Sensor-Type</label>
+				<limitToOptions>true</limitToOptions>
+				<options>
+					<option value="/humidity">HIH-3610</option>
+					<option value="/HIH4000/humidity">HIH-4000</option>
+					<option value="/HTM1735/humidity">HTM-1735</option>
+					<option value="/DATANAB/humidity">Datanab</option>
+				</options>
+				<default>/humidity</default>
+				<required>true</required>
```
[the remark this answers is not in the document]
J-N-K: Agreed on the "required" setting. But I think it is more transparent to the iser if he can select the sensor he has rather than leaving the selection empty (especially when configuiring via Paper UI).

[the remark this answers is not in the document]
kaikreuzer: The Paper UI should actually pro-actively fill the field with the provided default, so the selection should not be empty (otherwise I'd consider it a bug in the Paper UI, which you might report).

extensions/binding/org.eclipse.smarthome.binding.onewire/src/main/java/org/eclipse/smarthome/binding/onewire/internal/discovery/OwDiscoveryService.java
```diff
@@ -0,0 +1,181 @@
...
+        // resolve all non-DS2438
+        Iterator<Entry<String, String>> associationMapIterator = associationMap.entrySet().iterator();
+        while (associationMapIterator.hasNext()) {
+            Entry<String, String> entry = associationMapIterator.next();
+            if (!entry.getKey().substring(0, 2).equals("26")) {
+                if (owDiscoveryItems.containsKey(entry.getKey()) && owDiscoveryItems.containsKey(entry.getValue())) {
+                    owDiscoveryItems.get(entry.getValue()).addAssociatedSensor(owDiscoveryItems.get(entry.getKey()));
+                    owDiscoveryItems.remove(entry.getKey());
+                } else {
+                    logger.info("cannot resolve association {}->{}", entry.getKey(), entry.getValue());
```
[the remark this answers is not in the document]
J-N-K: this is most probably an hardware fault. I have added a note on that.

extensions/binding/org.eclipse.smarthome.binding.onewire/src/main/java/org/eclipse/smarthome/binding/onewire/internal/owserver/OwserverConnection.java
J-N-K: reduced to DEBUG, this migth happen in case of issues with the network or busg in the owserver, not in the binding.

extensions/binding/org.eclipse.smarthome.binding.onewire/src/main/java/org/eclipse/smarthome/binding/onewire/internal/owserver/OwserverPacket.java
J-N-K: removed as it is not used at all

extensions/binding/org.eclipse.smarthome.binding.onewire/ESH-INF/thing/DigitalIO.xml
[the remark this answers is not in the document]
J-N-K: does that mean it should be removed entirely or just left blank?

extensions/binding/pom.xml
J-N-K: oops

extensions/binding/org.eclipse.smarthome.binding.onewire/ESH-INF/thing/DigitalIO.xml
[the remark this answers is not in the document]
kaikreuzer: remove entirely

extensions/binding/org.eclipse.smarthome.binding.onewire/src/main/java/org/eclipse/smarthome/binding/onewire/internal/discovery/OwDiscoveryService.java
```diff
@@ -0,0 +1,181 @@
...
+                    OwDiscoveryItem owDiscoveryItem = new OwDiscoveryItem(owBridgeHandler, sensorId);
+                    owDiscoveryItems.put(owDiscoveryItem.getSensorId(), owDiscoveryItem);
+                    if (owDiscoveryItem.hasAssociatedSensorIds()) {
+                        for (String associatedSensorId : owDiscoveryItem.getAssociatedSensorIds()) {
+                            associationMap.put(associatedSensorId, owDiscoveryItem.getSensorId());
+                        }
+                    }
+                    logger.trace("found sensor {} (id: {})", owDiscoveryItem.getSensorType(), sensorId);
+                } catch (OwException e) {
+                    logger.info("sensor-id {}: {}", sensorId, e.getMessage());
```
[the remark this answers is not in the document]
kaikreuzer: Ok, thanks for the explanation, I'm ok with INFO then, but I would suggest to alter the message to something more user friendly, telling him that it isn't supported.

kaikreuzer:
Good news, the CQ has already been approved!

PR #4290, opened 2017-09-17
Nullness fixes

maggu2810:
The commits are separated, so rebase without squash should be okay.

### DIFF
--- a/bundles/core/org.eclipse.smarthome.core/src/main/java/org/eclipse/smarthome/core/common/registry/Identifiable.java
+++ b/bundles/core/org.eclipse.smarthome.core/src/main/java/org/eclipse/smarthome/core/common/registry/Identifiable.java
@@ -7,11 +7,14 @@
  */
 package org.eclipse.smarthome.core.common.registry;
 
+import org.eclipse.jdt.annotation.NonNullByDefault;
+
 /**
  * Interface for classes that instances provide an identifier.
  *
  * @author Markus Rathgeb - Initial contribution
  */
+@NonNullByDefault
 public interface Identifiable<T> {
 
     /**
--- a/bundles/io/org.eclipse.smarthome.io.transport.mqtt/src/main/java/org/eclipse/smarthome/io/transport/mqtt/MqttService.java
+++ b/bundles/io/org.eclipse.smarthome.io.transport.mqtt/src/main/java/org/eclipse/smarthome/io/transport/mqtt/MqttService.java
@@ -19,6 +19,8 @@ import java.util.concurrent.CopyOnWriteArrayList;
 import javax.naming.ConfigurationException;
 
 import org.apache.commons.lang.StringUtils;
+import org.eclipse.jdt.annotation.NonNullByDefault;
+import org.eclipse.jdt.annotation.Nullable;
 import org.eclipse.smarthome.core.events.EventPublisher;
 import org.osgi.service.component.annotations.Activate;
 import org.osgi.service.component.annotations.Component;
@@ -41,13 +43,15 @@ import org.slf4j.LoggerFactory;
  */
 @Component(immediate = true, service = { MqttService.class }, configurationPid = {
         "org.eclipse.smarthome.mqtt" }, property = "service.pid=org.eclipse.smarthome.mqtt")
+@NonNullByDefault
 public class MqttService {
     private static final String NAME_PROPERTY = "name";
     private final Logger logger = LoggerFactory.getLogger(MqttService.class);
     private final Map<String, MqttBrokerConnection> brokerConnections = new ConcurrentHashMap<String, MqttBrokerConnection>();
     private final List<MqttBrokersObserver> brokersObservers = new CopyOnWriteArrayList<>();
+
     @Deprecated
-    private EventPublisher eventPublisher;
+    private @Nullable EventPublisher eventPublisher;
 
     /**
      * The expected service configuration looks like this:
@@ -127,7 +131,12 @@ public class MqttService {
 
         for (Map<String, String> brokerConfig : brokerConfigs.values()) {
             try {
-                addBrokerConnection(brokerConfig).start();
+                final MqttBrokerConnection conn = addBrokerConnection(brokerConfig);
+                if (conn == null) {
+                    logger.warn("MqttBroker connection name already present.");
+                    continue;
+                }
+                conn.start();
             } catch (ConfigurationException e) {
                 logger.warn("MqttBroker connection configuration faulty: {}", e.getMessage());
             } catch (MqttException e) {
@@ -182,7 +191,7 @@ public class MqttService {
      * @param brokerName to look for.
      * @return existing connection or null
      */
-    public MqttBrokerConnection getBrokerConnection(String brokerName) {
+    public @Nullable MqttBrokerConnection getBrokerConnection(String brokerName) {
         synchronized (brokerConnections) {
             return brokerConnections.get(brokerName.toLowerCase());
         }
@@ -222,17 +231,17 @@ public class MqttService {
      * @throws ConfigurationException Most likely your provided name and url are invalid.
      * @throws MqttException
      */
-    public MqttBrokerConnection addBrokerConnection(Map<String, String> brokerConnectionConfig)
+    public @Nullable MqttBrokerConnection addBrokerConnection(Map<String, String> brokerConnectionConfig)
             throws ConfigurationException, MqttException {
         // Extract mandatory fields
         String brokerID = brokerConnectionConfig.get(NAME_PROPERTY);
-        if (StringUtils.isBlank(brokerID)) {
+        if (brokerID == null || brokerID.isEmpty()) {
             throw new ConfigurationException("MQTT Broker property 'name' is not provided");
         }
         brokerID = brokerID.toLowerCase();
 
         final String brokerURL = brokerConnectionConfig.get("url");
-        if (StringUtils.isBlank(brokerURL)) {
+        if (brokerURL == null || brokerURL.isEmpty()) {
             throw new ConfigurationException("MQTT Broker property 'url' is not provided");
         }
         // Add the connection
@@ -295,7 +304,7 @@ public class MqttService {
      * @param brokerName The broker name
      * @return Returns the removed broker connection, or null if there was none with the given name.
      */
-    public MqttBrokerConnection removeBrokerConnection(String brokerName) {
+    public @Nullable MqttBrokerConnection removeBrokerConnection(String brokerName) {
         synchronized (brokerConnections) {
             MqttBrokerConnection connection = brokerConnections.remove(brokerName.toLowerCase());
             if (connection != null) {


kaikreuzer:
Travis asks you to also adjust `ItemRegistryDelegate`...

bundles/core/org.eclipse.smarthome.core/src/main/java/org/eclipse/smarthome/core/common/registry/Identifiable.java
```diff
@@ -7,11 +7,14 @@
...
 package org.eclipse.smarthome.core.common.registry;
 
+import org.eclipse.jdt.annotation.NonNullByDefault;
+
 /**
  * Interface for classes that instances provide an identifier.
  *
  * @author Markus Rathgeb - Initial contribution
  */
+@NonNullByDefault
```
kaikreuzer: Don't we still have the issue that `Rule.getUID()` is still allowed to return null and thus is `@nullable`?

maggu2810: Hm, you are right, the current concept allows rules without an UID.
But does it make sense to state that we have an interface `Identifiable` that provides an UID and allow an implementing class to return null on that method (so "I am itentifiable but don't give you an UID)?
Shouldn't we throw an `IllegalStateException` on the ` Rule.getUID()` method if there is no UID assigned? As soon as someone wants to call `getUID` because it relies that the reference is an `Identifiable` it should also provide an UID. If the `Rule` has no UID assigned, the method should not be called at all (the usage of that reference should know about the fact that there is no UID yet assigned and should not call that method).

maggu2810:
> Travis does not seem to like the changes to `ItemRegistryDelegate`...

Crazy :wink: Will have a look at.

maggu2810:
>> Travis does not seem to like the changes to `ItemRegistryDelegate`...

> Crazy :wink: Will have a look at.

Ah, it is about the missing external annotations.
The key and the value of an entry set is `@Nullable` or `@NonNull` dependent on the type that is used for the key and the value (the entry set only contains entries of a map that contains only types that are allowed in the map). As long as there is not the correct external annotation also an entry set could return nullables also if it has been declared that there are no null entries (key or value).
I will remove that commit and move it to my other branch.

bundles/core/org.eclipse.smarthome.core/src/main/java/org/eclipse/smarthome/core/common/registry/Identifiable.java
kaikreuzer: Note that I brought this topic up [here](https://github.com/eclipse/smarthome/pull/3933#issue-247309524) already:

> I think we have introduced a bug when adding the Identifiable interface: I would claim that Identifiable.getUID() should return a non-null value. Adding such an annotation gets us into trouble with the Rule class, though as we allow the creation of rules without any UID (which is added only by the rule registry upon add). I think we should have a closer look at how this can best be resolved - for the moment, I didn't add the annotation on getUID().

Throwing an `IllegalStateException` might be a solution, but @danchom should be involved in that refactoring.

maggu2810: Great, I identified that I replied to your comment :wink:

danchom: >Shouldn't we throw an IllegalStateException on the Rule.getUID() method if there is no UID assigned? 

For me no, because it will be hard for users to know for already used strings which are used by the RuleEngine for ruleUIDs. At the moment the RuleEngine grantee unique ids for added rules.
Actuality the Rule is Identifiable after the object is added to RuleRegistry. Before that it is just a POJO object which carries rule's information and there is no guarantee that it is a valid rule object.

maggu2810: > it will be hard for users to know for already used strings which are used by the RuleEngine as ruleUIDs.

I don't understand.
So, if a "not added rule" does not contain an UID it is not identifiable.

> Before that it is just a POJO object which carries rule's information

So, should we create a DTO or POJO object?

danchom: >So, if a "not added rule" does not contain an UID it is not identifiable.

It depends if it is added to the registry or not. After add operation it will be identifiable. May be the the rule object must not be Identifiable at all?

adimova: We can forbid the creation of a Rule/ModuleType/RuleTemplate by constructors that accept UID=null as a parameter and to add a method to the RuleRegistry, so the user can get а valid UID

maggu2810: The `ScriptedAutomationManager` contains the following line:

```Java
        Rule rule = element.getUID() == null ? new Rule(generateUID()) : new Rule(element.getUID());
```

in the `Rule addRule(Rule element)` method.

@adimova Do I understand your suggestion correctly that we create an e.g. `RuleUtils` class with a static method `public static String generateRuleUID()` and change the `Rule()` constructor to use this method to generate a new random UID on construction time? Using that approach we could ensure that every rule contains a UID as soon as the object is created, the given one or a generated one.

kaikreuzer: @maggu2810 I am afraid that won't work as the logic to determine a UID is done within the RuleRegistry through checking which other rule UIDs are already assigned - so if I understand it right, @adimova suggests to have that method available on the RuleRegistry, not statically on a Utils class.

danchom: BTW implementing Identifiable interface grantee that the rule must have getUID() method, but this is not related in any way to return value of this method.

maggu2810: @kaikreuzer Wouldn't this mean that the RuleRegistry reference needs to be accessed by the Rule constructor? IIRC The generate UID uses the UUID class so potential duplicates are rather rare (if ever).

At least we need a solution.
* References that are stored in a `Registry` needs to implement the Identifiable interface.
* The `Identifiable` interface makes only sense if it returns a non-nullable type (I assume we all agree).
* The `Rule` is stored in the `RuleRegistry`, so it must not return null on `getUID`

Perhaps @adimova could give us more details how to fix that.

adimova: @maggu2810 i don't like the 

> an e.g. RuleUtils class with a static method public static String generateRuleUID() 

approach, because the default constructors of Automation objects ware introduced for usage from GSON parser only, and every time when we use them, that will reserve a UID which will be replaced after that in case the user wants its own UID. The reserved UID can't be used anymore. 
I actually want to remove the Rule() constructor and the default constructors of other Automation objects. GSON can use plugged deserializers instead of default constructors. For Module types and Templates we have not such logic that prevents "null" for UIDs and the plugged deserializers will solve this problem on common way.

maggu2810: > @maggu2810 i don't like the ... approach

No prob, it has been only a suggestion without knowing the automation infra. I am pretty sure you know the better solution at all.

IIRC Gson is used for the REST interface and the persistent `Storage` (e.g. mapdb or json storage).
* Rules that are persisted have already an UID assigned. Correct?
* Rules that are transfered from / to the REST interface needs to use the `RuleDTO` class (part of `org.eclipse.smarthome.automation.api`.

> BTW implementing Identifiable interface grantee that the rule must have getUID() method, but this is not related in any way to return value of this method.

@danchom The whole discussion is about that the interface should require to return a non-null type. So this is related.

adimova: @maggu2810 I'm answering to the:

> Rules that are persisted have already an UID assigned. Correct?

yes, this is correct

> Rules that are transfered from / to the REST interface needs to use the RuleDTO class (part of org.eclipse.smarthome.automation.api.

The REST interface uses RuleRegistry and can request a valid UID to use it for RuleDTO objects, so my proposition - the RuleRegistry to be extended with such method as generateRuleUID() will solve that issue

maggu2810: @adimova Can you create a PR?

adimova: ok

bundles/io/org.eclipse.smarthome.io.transport.mqtt/src/main/java/org/eclipse/smarthome/io/transport/mqtt/MqttService.java
```diff
@@ -222,17 +231,17 @@ public boolean addBrokerConnection(MqttBrokerConnection connection) {
      * @throws ConfigurationException Most likely your provided name and url are invalid.
      * @throws MqttException
      */
-    public MqttBrokerConnection addBrokerConnection(Map<String, String> brokerConnectionConfig)
+    public @Nullable MqttBrokerConnection addBrokerConnection(Map<String, String> brokerConnectionConfig)
             throws ConfigurationException, MqttException {
         // Extract mandatory fields
         String brokerID = brokerConnectionConfig.get(NAME_PROPERTY);
-        if (StringUtils.isBlank(brokerID)) {
+        if (brokerID == null || brokerID.isEmpty()) {
```
htreu: Imho it really is a pity that we have to "work around" the annotations now just because the compiler is not able to recognise `StringUtils.isBlank` as an implicit null check. I just had a similar situation in a builder pattern (see https://github.com/eclipse/smarthome/pull/4301/files#diff-8f975ae88094089585192875908ba08bR79) which I could solve by doing some API changes. In the end it turned out these changes where a turn for the better.

Just to clarify: I thought `@NonNullByDefault` should guarantee the call to `brokerConnectionConfig.get(NAME_PROPERTY)` to always return a non null value? So why is the null check (or StringUtils.isBlank) required at all?

bundles/io/org.eclipse.smarthome.io.transport.mqtt/src/main/java/org/eclipse/smarthome/io/transport/mqtt/MqttService.java
```diff
@@ -222,17 +231,17 @@ public boolean addBrokerConnection(MqttBrokerConnection connection) {
...
             throw new ConfigurationException("MQTT Broker property 'name' is not provided");
         }
         brokerID = brokerID.toLowerCase();
 
         final String brokerURL = brokerConnectionConfig.get("url");
```
htreu: another question about the non null behaviour of maps: With `@NonNullByDefault` we force the compiler to interpret a call to `brokerConnectionConfig.get("url")` to always return a non null value. How can this be guaranteed at all? I get the impression that any call to a map is interpreted as non null value, making additional null checks obsolete. Thats a dangerous behaviour:

```
Object value = myMap.get("key");
if (value == null) {
   throw ....
}
```

in this example the body of the if clause is reported as dead code. I guess for Maps the default should be `Map<@NoNull KEY_TYPE, @Nullable VALUE_TYPE>`.

wdyt?

maggu2810: > another question about the non null behaviour of maps: With `@NonNullByDefault` we force the compiler to interpret a call to `brokerConnectionConfig.get("url")` to always return a non null value.

No, you define the default for your classes or packages but not for the map itself.
A map's get method could always return null e.g. your ask for a value of a non-existing key.

> I guess for Maps the default should be `Map<@NoNull KEY_TYPE, @Nullable VALUE_TYPE>`.

No, this is not correct. If you look at the Map API null is allowed for the key if the implementation allows null keys (stated in the JavaDoc).

Also `@NonNull K` is not correct at all, because of the upper / lower bounds of generic types. In the generic type declaration you need to use `K extends @NonNull Object`.

The checker framework has a good documentation about that. I realized it a few days ago.

bundles/io/org.eclipse.smarthome.io.transport.mqtt/src/main/java/org/eclipse/smarthome/io/transport/mqtt/MqttService.java
maggu2810: > Imho it really is a pity that we have to "work around" the annotations now just because the compiler is not able to recognise `StringUtils.isBlank` as an implicit null check.

You are right, the Eclipse Nullness Checker should be improved and the EEAs.
The Checker Framework can handle such situations already.
There I could use astub files (similar to EEA) and state that if the method returns true, the reference used in the argument is non-null.

```Java
class StringUtils {
    static @EnsuresNonNull("#1") boolean isBlank(@Nullable String str);
}
```

maggu2810:
This PR contains three commits and it seems at least the `Identifiable` blocks the other ones.
Let's clean it up.

* fix nullness of MQTT service
    => could be done on the go in https://github.com/eclipse/smarthome/pull/4173
* Identifiable should make use of '@NonNullByDefault' ￼…
    => we are waiting for a PR of @adimova first https://github.com/eclipse/smarthome/pull/4290#discussion_r139631218 
* fix nullness annotations for automation commands utils ￼…
    => moved to https://github.com/eclipse/smarthome/pull/4319

bundles/core/org.eclipse.smarthome.core/src/main/java/org/eclipse/smarthome/core/common/registry/Identifiable.java
adimova: @kaikreuzer, @maggu2810 we need to agree on a decision about the `@NonNull` UID and `@NonNullByDefault` for the `getUID()` method in the Identifiable interface. The possible solutions are:

1. `Null` to be possible for UID - this is the current situation when `Null` is a valid UID. 
- Advantage: no changes :-)
- Disadvantage: all "not added rules" are equal

2. System UID that will be recognized and replaced by the Registry after adding the Rule - this is similar to the current situation, but without default constructors, and if the constructors are called with `Null`, the system UID (constant) will be used instead. 
 - Advantage: `@NonNullByDefault` can be used for the `getUID()` method.
 - Disadvantage: all "not added rules" are equal

3. Add a `generateUID()` method in Registry. It can be added in:
  a) Rule Registry, or
  b) Abstract Registry
 - Advantage: `@NonNullByDefault` can be used for the `getUID()` method.
 - Disadvantage: The IDs generated by the `generateUID()` will be unique at the time of generation. When you try to add the Rule in the Registry, the ID my have already been taken for another rule, and adding will fail. In the add method, the Registry does not know if the rule UID was generated by the `generateUID()` method, or set explicitly by the user.

4. The user must provide a correct UID for the rule
 - Advantage: the Rule registry acts as all other Registries.
 - Disadvantage: we make it more difficult for the user

All resolutions have negatives and we should discuss which is the most effective one.

maggu2810: Didn't we already agree that the `getUID()` of an `Identifiable` should return an identifiable ID and so not null already?

About your point 3: Couldn't this (the UID generation) be make clever that a collision is such a rare case that it will "never" be happen (see e.g. https://en.wikipedia.org/wiki/Universally_unique_identifier#Collisions)?

adimova: > Didn't we already agree that the getUID() of an Identifiable should return an identifiable ID and so not null already?

@maggu2810 I agree with this, but I can't get this decision alone, I should convince our architects. I am actually asking by their request.

> About your point 3: Couldn't this (the UID generation) be make clever that a collision is such a rare case that it will "never" be happen?

It is not about "when the same UUID is generated more than once", it is about that the same UUID could be generated form more than one generators, each of them with its own way to generate a UUID for its rules:

- scripting
- file provider
- bundle provider
- by the user it self
- third party app

I am actually voting for point 4 but let me say again I can't get this decision by my self

kaikreuzer: I agree that option 4 is probably the best. This would at least also make it consistent with how we deal with it for Things, where a UID is also generated (by the UI) if a thing is manually created.

To reduce the impact, we could imho simply generate a `java.util.UUID` in the Rule constructor, if no UID is passed from the client. This way we would ensure to always have a UID while all client code can stay as it is. Note that the ScriptedRuleSupport already uses with `UUID`s, so it would be in sync with this part as well. The only downside would be that the UIDs are not "nice" compared to the current "rule_x", but as they are an internal identifier not really meant for any end user, this is imho acceptable.

adimova: See #4438 to confirm is it what we agreed here

maggu2810:
Rebased

Issue #4133, opened 2017-08-25
Cron Scheduler Replacement?


kaikreuzer:
We have learned over the past months that implementing a cron scheduler is not a trivial task - our current implementation has caused quite some issues, which are hard to analyse, debug and fix.

Instead of fully inventing it on our own, we should check whether using existing implementations might be a viable option. Two potential candidates have already been mentioned in the past and I would like to at least check their pros & cons to come to an informed decision:

- [cron-utils](https://github.com/jmrozanec/cron-utils): Seems to be an actively maintained project with a nice API and many features. Not sure whether it can deal with scheduling jobs itself though, it seems that it only calculates next execution times.
- [OSGi enRoute Scheduler](https://github.com/osgi/osgi.enroute/tree/master/osgi.enroute.scheduler.simple.provider): Looks like a small and straight-forward scheduler that can also deal with cron expressions and which might even become [part of the OSGi spec](https://github.com/osgi/design/blob/master/rfcs/rfc0226/rfc-0226-Scheduling.pdf) one day.

maggu2810:
FYI: The same github person that hosts the cron-utils project also hosts an cron-utils-scheduler project.
Perhaps we could have a look at this one, too.

kgoderis:
Cons : it only supports cron expressions

I initially implemented the scheduler as part of a larger exercise to have a built in caldav server, but that was never reviewed, and to replace the quartz scheduler. I would not like to throw that effort away

This aside, imho this kind of effort is energy badly spent. What kind of issues did we have, and that we are unable to solve? Does the 2-3 bugs we had to fix really make that we have throw all this away? I don it understand your reasoning here

kaikreuzer:
> Cons : it only supports cron expressions

Right, that's a good point - caldav support can indeed become an important feature.

> What kind of issues did we have, and that we are unable to solve?

We had a couple of issue that haven't been solved: https://github.com/eclipse/smarthome/issues/3185, https://github.com/eclipse/smarthome/issues/3912 and https://github.com/eclipse/smarthome/issues/4145 and a few others that were solved.

The problem is that all of them are rather critical issues and quite some people actually already tried looking into them, but failed to come up with a fix. I have to admit that I personally struggle to understand the code and its inner working - so I am of no huge help here either.

triller-telekom:
[...]

Issue #3325, opened 2017-04-27
Handling exceptions from the schedulers


martinvw:
I think that it is a pity that at this moment (almost) every Runnable in openHAB (and most likely the also in the Eclipse Smarthome) contain a `catch Exception` and some (even) contain a catch Throwable. But this should imho not be a responsibility of the bindings, their code should be clean and straight forward. 

One of the many examples is this one (chosen because its the first binding in the list): https://github.com/openhab/openhab2-addons/blob/master/addons/binding/org.openhab.binding.airquality/src/main/java/org/openhab/binding/airquality/handler/AirQualityHandler.java#L120

However I would like to make the framework responsible for this, there are a few solutions for this:

1. Handle them in the [`ThreadPoolExecutor.afterExecute`](https://docs.oracle.com/javase/7/docs/api/java/util/concurrent/ThreadPoolExecutor.html#afterExecute(java.lang.Runnable,%20java.lang.Throwable)) [example](http://blog.kavinyao.com/2014/11/java-thread-pool/#gist14515810)
2. Implementing a `ThreadFactory` and registering the `UncaughtExceptionHandler` [example](http://farenda.com/java/java-uncaughtexceptionhandler-executorservice/)
3. Something custom with wrapping the runnables, a minor benefit as that we could really tailor the logging for example to log from the runnable class.

I was really curious how you feel about this subject, so please let me know.

Note that it is also considered on a list of secure coding habits: https://www.securecoding.cert.org/confluence/display/java/TPS03-J.+Ensure+that+tasks+executing+in+a+thread+pool+do+not+fail+silently

sjka:
Thanks for bringing up this topic. Indeed, it's a good idea!

Probably I'd use the first option, as it is straight-forward and you have the Runnable at hand, hence you can get a logger for it. So I don't see a need for the third option. However, I don't have any strong feelings about it. 

Another question would be: What do you want to do then? Of course, logging the error. But afaik most handlers then also take some action on their Things, i.e. set them to OFFLINE/... How should this be done? If you want to do it also in the generic exception handling, then you ought to make sure that this doesn't interfere with all the other usages of the scheduler as t is heavily used by the framework itself too. On the other hand, if that's not done centrally, then the bindings still need to catch exceptions and nothing is won.

martinvw:
Nope I agree that option 3 does not add too much value, I started out with the third option in head but realised that the JDK already offered us some more standard solution for it.

[...]

PR #4295, opened 2017-09-18
[WIP] Alternate scheduler implementation

sjka:
...supporting cron expressions and potentially other notations/formats in future.

This implementation is a stripped-down version of the [OSGi enRoute Scheduler](https://github.com/osgi/osgi.enroute/tree/master/osgi.enroute.base.api/src/osgi/enroute/scheduler/api) with @pkriens consent, adapted to Java 8's `CompletableFuture` instead of OSGi `Promise` in order to stay compliant with OSGi 4.2. It is implemented by far cleaner and simpler hence easier to maintain and is yet more powerful. 

It not only allows scheduling a job to be executed at a certain point in time (by an `Instant`), after a given delay (by a `Duration`) or at a given cron expression (see [Scheduler.java](https://github.com/eclipse/smarthome/compare/master...sjka:scheduler2?expand=1#diff-a46656866fefd61ccd01a439f036b799) for the API and [SchedulerTest.java](https://github.com/eclipse/smarthome/compare/master...sjka:scheduler2?expand=1#diff-c83f2d87fce54fcfd86eb2ba83a07201) for examples), but also allows declaring an OSGi service, implementing `CronJob` which then will automatically be picked up and executed according to the given cron configuration (see [CronJob.java](https://github.com/eclipse/smarthome/compare/master...sjka:scheduler2?expand=1#diff-719c42ac68f5a96a1837487151075d6e)). If used programmatically, the scheduler is accessed as an OSGi service and will take care for cleaning up all scheduled jobs in case the using bundle gets stopped. It can be extended to support additional scheduler definitions in future - the contract for this into the scheduler implementation is Java's `TemporalAdjuster` which needs to be implemented.

This PR is meant to further progress the discussion of #4133 by show-casing an alternative to our current implementation. If this reaches a consensus to continue on this way, I will take care of the yet open points:

* Migration of all current usages to this new scheduler
* Renaming it to it's final package name
* And certainly some final cleanups

Also-by: Peter Kriens <Peter.Kriens@aqute.biz>


kaikreuzer:
Looks indeed like a very clean implementation - most code in this PR are tests, JavaDoc and Interfaces ;-)

@maggu2810 @kgoderis What do you think of it?

bundles/core/org.eclipse.smarthome.core/src/main/java/org/eclipse/smarthome/core/scheduler2/CronJob.java
```diff
@@ -0,0 +1,104 @@
...
+ * &#64;daily                  Run once a day at midnight                                                   0 0 * * *
+ * &#64;hourly                 Run once an hour at the beginning of the hour                                0 * * * *
+ * &#64;reboot                 Run at startup                                                               @reboot
+ * </pre>
+ * <p>
+ * Please not that for the constants we follow the Java 8 Date & Time constants.
+ * Major difference is the day number. In Quartz this is 0-6 for SAT-SUN while
+ * here it is 1-7 for MON-SUN.
+ *
+ * @param <T>
```
maggu2810: Where is T used in this interface at all?

sjka: Oh, that was a victim of my stripping-down which I didn't clean up yet. The enRoute implementation had a pretty fancy mechanism to covert the service configuration into an arbitrary POJO - similar as our `Configuration.as(...)` just far more flexible. 
As we already have our configuration mechanism I didn't want to introduce another one, so I though it would be sufficient here to leave it as a map, which optionally can be feed into our `Configuration`.
Anway: done.

pkriens:
Is it sufficiently to state here that I fully wrote the original and own the original copyright and approve of this derivation to be licensed under EPL?

kaikreuzer:
@pkriens, yes thanks - as you have a valid ECA and @sjka includes `Also-by: Peter Kriens <Peter.Kriens@aqute.biz>` in the commit message, this is all clean from an IPR perspective.

@maggu2810 What is your impression here? Would you prefer to stick with our current implementation or might it be worth to consider this as a replacement?

maggu2810:
Didn't have a deeper look yet.

kgoderis:
I will have a deeper look as well, but I think the API has to be altered slightly in order to better facilitate some other expression formats. The actual implementation is rather Cron centric and could be factored out / generalised for any type of expression format.

bundles/core/org.eclipse.smarthome.core/src/main/java/org/eclipse/smarthome/core/internal/scheduler2/InternalSchedulerImpl.java
```diff
@@ -0,0 +1,320 @@
...
+
+            schedule(s, s.next());
+            return null;
+        }, instant);
+        if (s.canceled) {
+            s.promise.cancel(true);
+        }
+    }
+
+    private class ScheduleCron extends Schedule {
```
kgoderis: I would generalise this class in order to use any TemporalAdjuster instead of a CronAdjuster

bundles/core/org.eclipse.smarthome.core/src/main/java/org/eclipse/smarthome/core/internal/scheduler2/InternalSchedulerImpl.java
```diff
@@ -0,0 +1,320 @@
...
+        volatile CompletableFuture<?> promise;
+        volatile boolean canceled;
+        Throwable exception;
+
+        abstract Instant next();
+
+        abstract void doIt() throws Exception;
+    }
+
+    private static class PeriodSchedule extends Schedule {
```
kgoderis: If we want to introduce other types of expressions, one might consider to remove this class altogether, and rather craft a TemporalAdjuster that deals with the desired Period instead.

kgoderis:
I will refrain from a detailed review here [I am not a software engineer by profession], but I found the solution using the TemporalAdjusters and CompletableFutures a rather elegant one that will permit to use other types of expressions in a straight-forward manner. These Java 8 techniques are definitely better than the current scheduler (implemented with Java 7 in mind) However, I think a bit of generalisation has to be done to step away from a Cron-only view of the world. [I also wonder why the the actual implementation of the scheduler is kept separate in InternalSchedulerImpl - it only renders things a bit more complicated].

pkriens:
The separation is needed to do cleanup when a bundle disconnects I recall.

bundles/core/org.eclipse.smarthome.core/src/main/java/org/eclipse/smarthome/core/internal/scheduler2/InternalSchedulerImpl.java
sjka: Good point! That was one of the open things to clean up which I wanted to do only if this way gets more or less the "approval" of everybody. I quickly did it nevertheless: It indeed turned out nicely, I think.

kgoderis:
@sjka Simon, will we not break the API the day we introduce other scheduler expression types? It is rather cron-only for the time being.

bundles/automation/org.eclipse.smarthome.automation.module.timer/src/main/java/org/eclipse/smarthome/automation/module/timer/handler/TimeOfDayTriggerHandler.java
```diff
@@ -72,10 +72,13 @@ public void run() {
     @Override
     public synchronized void dispose() {
```
martinvw: This looks duplicated can this live elsewhere?

bundles/core/org.eclipse.smarthome.core/src/main/java/org/eclipse/smarthome/core/internal/scheduler2/DelegatedScheduler.java
```diff
@@ -0,0 +1,128 @@
...
+                    return;
+                }
+                Iterator<Closeable> iterator = closeables.iterator();
+                c = iterator.next();
+                iterator.remove();
+            }
+
+            try {
+                c.close();
+            } catch (Throwable t) {
```
martinvw: Do not catch `Throwable`'s you will also catch for example `OutOfMemoryError`s, this will not improve the stabilitly and predictability of the code.

bundles/core/org.eclipse.smarthome.core/src/main/java/org/eclipse/smarthome/core/internal/scheduler2/DelegatedScheduler.java
```diff
@@ -0,0 +1,128 @@
...
+public class DelegatedScheduler implements Scheduler {
+
+    private InternalSchedulerImpl delegate;
+    private final Set<Closeable> closeables = new HashSet<>();
+
+    @Deactivate
+    void close() {
+        while (true) {
+            Closeable c;
+            synchronized (closeables) {
```
martinvw: Does the added complexity of closing with the double loop really improve the safety afaik ik someone has a reference to the deactivated scheduler he might still add a `Closable` which will not be closed.

bundles/core/org.eclipse.smarthome.core/src/main/java/org/eclipse/smarthome/core/internal/scheduler2/InternalSchedulerImpl.java
```diff
@@ -0,0 +1,298 @@
...
+            } else {
+                return false;
+            }
+        }
+    }
+
+    @Override
+    public <T> CompletableFuture<T> before(CompletableFuture<T> promise, Duration timeout) {
+        CompletableFuture<T> d = new CompletableFuture<T>();
+        Unique only = new Unique();
```
martinvw: Would there really go something wrong without this `Unique`? It looks like that it should work without it.

bundles/core/org.eclipse.smarthome.core/src/main/java/org/eclipse/smarthome/core/internal/scheduler2/InternalSchedulerImpl.java
```diff
@@ -0,0 +1,298 @@
...
+                return true;
+            } else {
+                return false;
+            }
+        }
+    }
+
+    @Override
+    public <T> CompletableFuture<T> before(CompletableFuture<T> promise, Duration timeout) {
+        CompletableFuture<T> d = new CompletableFuture<T>();
```
martinvw: Can d have a real name what is where is it used for?

martinvw: Maybe take a look at some of the other one letter variables as well

bundles/core/org.eclipse.smarthome.core/src/main/java/org/eclipse/smarthome/core/internal/scheduler2/InternalSchedulerImpl.java
```diff
@@ -0,0 +1,298 @@
...
+            s.promise.cancel(true);
+        };
+    }
+
+    private void schedule(Schedule s, Instant instant) {
+        s.promise = at(() -> {
+            try {
+                s.doIt();
+            } catch (Throwable t) {
+                if (s.exception != null) {
```
martinvw: What happens in this three lines and why?

bundles/core/org.eclipse.smarthome.core/src/main/java/org/eclipse/smarthome/core/scheduler2/CronJob.java
```diff
@@ -0,0 +1,101 @@
...
+ * chronos.
+ * <p>
+ * The Unix Cron defines a syntax that is used by the Cron service. A user
+ * should register a Cron service with the {@link CronJob#CRON} property. The
+ * value is according to the {link http://en.wikipedia.org/wiki/Cron}.
+ * <p>
+ *
+ * <pre>
+ * * * * * * * *
+ * | │ │ │ │ │ |
```
martinvw: How does the docblock relate to anything in this interface? Is see no parameters to pass an cronexpression?

bundles/core/org.eclipse.smarthome.core/src/main/java/org/eclipse/smarthome/core/scheduler2/TimeoutException.java
```diff
@@ -0,0 +1,26 @@
...
+public class TimeoutException extends Throwable {
+    private static final long serialVersionUID = 1L;
+
+    private TimeoutException() {
+    }
+
+    /**
+     * The singleton timeout exception
+     */
+    public static TimeoutException SINGLETON = new TimeoutException();
```
martinvw: Why a singleton exception? In this way you do not have a valid stacktrace.

bundles/core/org.eclipse.smarthome.core/src/main/java/org/eclipse/smarthome/core/internal/scheduler2/PeriodicAdjuster.java
```diff
@@ -0,0 +1,43 @@
...
+ *
+ * @author Peter Kriens - initial contribution and API
+ *
+ */
+class PeriodicAdjuster implements TemporalAdjuster {
+
+    private final Iterator<Duration> iterator;
+    private Duration current;
+
+    PeriodicAdjuster(Duration first, Duration... delays) {
```
martinvw: Maybe just change the constructor with only var args givne it's package private already.

bundles/core/org.eclipse.smarthome.core/src/main/java/org/eclipse/smarthome/core/internal/scheduler2/InternalSchedulerImpl.java
```diff
@@ -0,0 +1,298 @@
...
+        CompletableFuture<Instant> deferred = new CompletableFuture<>();
+        Instant start = Instant.now();
+        ScheduledFuture<?> schedule = executor.schedule(() -> {
+            deferred.complete(start);
+        }, duration.toMillis(), TimeUnit.MILLISECONDS);
+        deferred.exceptionally((Throwable e) -> {
+            if (e instanceof CancellationException) {
+                schedule.cancel(true);
+            }
+            return null;
```
martinvw: Do we also log the exception somewhere and do we or do we not have any control of whether the schedule should be cancelled when certain types of exceptions occur, see also https://github.com/eclipse/smarthome/issues/3325

bundles/core/org.eclipse.smarthome.core/src/main/java/org/eclipse/smarthome/core/internal/scheduler2/CronAdjuster.java
```diff
@@ -0,0 +1,551 @@
...
+        Temporal isOk(Temporal t) {
+            if (checker.matches(t)) {
+                return null;
+            }
+
+            Temporal out = t.plus(1, type.getBaseUnit());
+
+            switch (type) {
+                case YEAR:
+                    out = out.with(ChronoField.MONTH_OF_YEAR, 1);
```
martinvw: Missing breaks or intentional, if intentional maybe document that, because fall through is not that often used on purpose

bundles/core/org.eclipse.smarthome.core/src/main/java/org/eclipse/smarthome/core/internal/scheduler2/CronAdjuster.java
```diff
@@ -0,0 +1,551 @@
...
+            year = null;
+        }
+
+        fields = new Field[] { year, month, dayOfWeek, dayOfMonth, hours, minutes, seconds };
+    }
+
+    private Map<String, String> doEnv(String[] entries) {
+        Map<String, String> map = new HashMap<String, String>();
+        if (entries.length > 1) {
+            for (int i = 0; i < entries.length - 1; i++) {
```
martinvw: Are you ignoring the last entry?

bundles/core/org.eclipse.smarthome.core/src/main/java/org/eclipse/smarthome/core/internal/scheduler2/CronAdjuster.java
```diff
@@ -0,0 +1,551 @@
...
+            year = parse(parts[6], ChronoField.YEAR);
+        } else {
+            year = null;
+        }
+
+        fields = new Field[] { year, month, dayOfWeek, dayOfMonth, hours, minutes, seconds };
+    }
+
+    private Map<String, String> doEnv(String[] entries) {
+        Map<String, String> map = new HashMap<String, String>();
```
martinvw: Please use the diamond operator, ie replace the second occurrence of the same generic by `<>`

bundles/core/org.eclipse.smarthome.core/src/main/java/org/eclipse/smarthome/core/internal/scheduler2/CronAdjuster.java
```diff
@@ -0,0 +1,551 @@
...
+
+    /*
+     * A function interface that we use to check a Temporal to see if it matches
+     * a part of the specification. We combine these checkers in and and ors.
+     */
+    interface Checker {
+        boolean matches(Temporal t);
+    }
+
+    Checker ALWAYS_FALSE = (t) -> false;
```
martinvw: Can mor things be private in this class or maybe moved out and can the order be more strict, static fields, fields, constructors, methods. Then somewhere else the inner classes, depending on where you need to use them.

bundles/core/org.eclipse.smarthome.core.test/src/test/java/org/eclipse/smarthome/core/internal/scheduler2/SchedulerTest.java
```diff
@@ -0,0 +1,175 @@
...
+        d.completeExceptionally(e);
+        assertTrue(before.isDone());
+        assertTrue(before.isCompletedExceptionally());
+    }
+
+    @Test
+    public void testTimeout() throws InterruptedException, InvocationTargetException {
+        CompletableFuture<Integer> d = new CompletableFuture<>();
+        CompletableFuture<Integer> before = si.before(d, Duration.ofMillis(100));
+        Thread.sleep(200);
```
martinvw: To me it does not make complete sense, I would expect that the scheduler runs a separate thread so the only way to test a timeout is by passing in something that will take to long complete, or am I missing some essential detail here?

kaikreuzer:
@sjka Any chance to rebase this and address @martinvw's comments?

kaikreuzer:
@martinvw Do I remember correctly that you were interested taking this work over? It's a pity to see this not progressing since more than a year by now...

Hilbrand:
@kai otherwise I'll be willing to pick this up.

martinvw:
@kaikreuzer @Hilbrand I don't recall that 😄. But if @Hilbrand can do it, that's great 👍

kaikreuzer:
@martinvw Then I might have mixed you both up. I just remember that it was a very bright Dutch guy, who mentioned this to me 😄. So @Hilbrand, go for it!

Hilbrand:
I have a number of questions:
1. The number of scheduled threads is in the code in this pr set to 10. Should that be made configurable? And is 10 enough?
1. In the current scheduler there are also `DateExpression` and `RecurrenceExpression`. Should that functionality also be present in the new implementation? One usage of `RecurrenceExpression` in `TimeOfDayTriggerHandler` was replaced by a cron scheduler.
1. Is any part current scheduler used outside the eclipse Smart Home core in any other (Smart Home or openHAB) projects?
1. Will there be a feature wanted to get the remaining time (or time till next trigger)? Or have the ability to programmatically pause/resume a scheduled job?
1. Or get an id and/or find a scheduled job by id? (All these features might not necessarily be added now, but if these are possible features I can see if that will have an impact in implementation and design in accordingly)
1. A practical. I guess the easiest why for me to continue is to open the changes in a new pr. Is that ok?

maggu2810:
> Is any part current scheduler used outside the eclipse Smart Home core in any other (Smart Home or openHAB) projects?

I need to check our code base but I don't think so. But also if we use it, I will give it a go as we could change our code easily to the alternate scheduler implementation.

bundles/core/org.eclipse.smarthome.core/src/main/java/org/eclipse/smarthome/core/internal/scheduler2/CronAdjuster.java
```diff
@@ -0,0 +1,551 @@
...
+        }
+
+        if (b == ALWAYS_TRUE) {
+            return a;
+        }
+
+        return (temporal) -> a.matches(temporal) && b.matches(temporal);
+    }
+
+    public Map<String, String> getEnv() {
```
Hilbrand: How the map returned by `getEnv()` be used? It doesn't seem to be used anywhere and CronAdjuster is a package private class. Is this something to pass to the `run` method in `CronJob`? or should it be made available somewhere else?

sjka: If you can't find any usages, then there probably aren't any. I just haven't come to cleaning up the adjusters, so potentially they are not needed. I would try to keep the API as slim as possible, therefore feel free to dump everything which appears to be not needed.

kaikreuzer:
Superseded by https://github.com/eclipse/smarthome/pull/6646

PR #4297, opened 2017-09-18
Fix Nullannotation problem

triller-telekom:
Given that GenericThingProvider.xtend is correct it creates a Channel with
a nullable ChannelTypeUID and a nullable acceptedItemType

solves Null type mismatch: required '@NonNull String' but the provided
value is inferred as @Nullable	GenericThingProvider.xtend
/org.eclipse.smarthome.model.thing/src/org/eclipse/smarthome/model/thing/
internal	line 349	Xtend Problem

and

Null type mismatch: required '@NonNull ChannelTypeUID' but the provided
value is inferred as @Nullable	GenericThingProvider.xtend
/org.eclipse.smarthome.model.thing/src/org/eclipse/smarthome/model/thing/
internal	line 352	Xtend Problem

### DIFF
--- a/bundles/core/org.eclipse.smarthome.core.thing/src/main/java/org/eclipse/smarthome/core/thing/binding/builder/ChannelBuilder.java
+++ b/bundles/core/org.eclipse.smarthome.core.thing/src/main/java/org/eclipse/smarthome/core/thing/binding/builder/ChannelBuilder.java
@@ -11,7 +11,6 @@ import java.util.HashSet;
 import java.util.Map;
 import java.util.Set;
 
-import org.eclipse.jdt.annotation.NonNull;
 import org.eclipse.jdt.annotation.NonNullByDefault;
 import org.eclipse.jdt.annotation.Nullable;
 import org.eclipse.smarthome.config.core.Configuration;
@@ -90,7 +89,7 @@ public class ChannelBuilder {
      * @param properties properties to add
      * @return channel builder
      */
-    public ChannelBuilder withProperties(Map<@NonNull String, String> properties) {
+    public ChannelBuilder withProperties(Map<String, String> properties) {
         this.properties = properties;
         return this;
     }
--- a/bundles/core/org.eclipse.smarthome.core.thing/src/main/java/org/eclipse/smarthome/core/thing/binding/builder/ChannelBuilder.java
+++ b/bundles/core/org.eclipse.smarthome.core.thing/src/main/java/org/eclipse/smarthome/core/thing/binding/builder/ChannelBuilder.java
@@ -12,6 +12,8 @@ import java.util.Map;
 import java.util.Set;
 
 import org.eclipse.jdt.annotation.NonNull;
+import org.eclipse.jdt.annotation.NonNullByDefault;
+import org.eclipse.jdt.annotation.Nullable;
 import org.eclipse.smarthome.config.core.Configuration;
 import org.eclipse.smarthome.core.thing.Channel;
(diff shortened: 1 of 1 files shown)


bundles/core/org.eclipse.smarthome.core.thing/src/main/java/org/eclipse/smarthome/core/thing/binding/builder/ChannelBuilder.java
```diff
@@ -87,7 +90,7 @@ private ChannelBuilder(ChannelUID channelUID, String acceptedItemType, Set<Strin
      * @param properties properties to add
      * @return channel builder
      */
-    public @NonNull ChannelBuilder withProperties(Map<@NonNull String, String> properties) {
+    public ChannelBuilder withProperties(Map<@NonNull String, String> properties) {
```
maggu2810: The `@NonNull` could be removed for the map's key.

triller-telekom: Oh sorry I overlooked that one, I have removed it now.

maggu2810:
AFAIK a channel without a channel type is valid,

But is a channel without an acceptedItemType a valid one? Is an accepted item type not a requirement?

triller-telekom:
If an acceptedItemtype is a requirement, then the [GenericThingProvider](https://github.com/eclipse/smarthome/blob/master/bundles/model/org.eclipse.smarthome.model.thing/src/org/eclipse/smarthome/model/thing/internal/GenericThingProvider.xtend#L350) should make sure that in all branches above this call set one or throw an exception before creating the channel.

maggu2810:
FYI this has been my change this weekend https://github.com/maggu2810/smarthome/commit/2404ab459bfe28e01e2f5205b00ad891f44fa568

But as written above the model.thing.test failed because it expects 4 channels and got only 2...

So, let's wait for @sjka or @kaikreuzer which argument is mandatory and which optional.

triller-telekom:
You had a check for `itemType === null` in there but this is now gone? See [here](https://github.com/maggu2810/smarthome/commit/2404ab459bfe28e01e2f5205b00ad891f44fa568#diff-07878baae2f7b44ae95eb70159660f59R350)

sjka:
In the DSL, you either have to specify an itemType (and leave out the channel type - so it is legal to define channels without a channel type), or you specify a channel type (which in turn has an optional item-type attribute only in its XSD). So effectively you can legally configure a channel without both! Question is whether any binding requires and/or supports this.

From looking at the code, it seems that the `ChannelItemProvider` is the only user of the itemType (a.k.a. acceptedItemType) and it is well-prepared for it being `null` (it won't create an item them). The PaperUI as far a I can see is too. So in reality it is not _really_ required.

Defining channels manually without a channel type currently is a supported scenario, IIRC. In many cases they are crucial though for bindings in order to distinguish the manually created ones, because there it cannot rely on the names anymore - unless they are all identical in what would be the channel type, then of course it could be omitted. I'm not aware of any binding where this is the case though.
I actually thought we would infer the channel type by matching its name to the ones from the thing type. But I can't seem to find this, so probably I don't remember correctly. 

Anyway, back to the null-annotations: Without changing the DSL and presumably also the REST implementation, you can't simply assume they're non-null and also can't ignore them if either channel type or accepted item type are null.

maggu2810:
@sjka Thanks for all that details. Then I assume this PR looks good.

PR #4298, opened 2017-09-18
Introduced channel category types and revisited all category lists

kaikreuzer:
This PR restructures the (icon) categories in a way that it better groups different category types (e.g. all locations, all "measurable properties", etc.) together. The goal of this is to work towards a better match to an ontology and to have a clean definition of which icons an icon set has to provide mandatorily.

The Classic Iconset page now looks like this:

classic

The different category types are separated from each other. I have also removed the different states for the same icon in order to provide a better overview. The different available states are shown as a hover text:

screen shot 2017-09-19 at 00 12 32

A future enhancement should be to replace this hover text by a pop up showing the real icons instead.

I mark this PR as API breaking as I have also removed and renamed certain categories, so that users might have to adapt their setups accordingly.


sjka:
This looks plausible to me and having the structure really helps for getting an overview!

I completely feel incapable to decide upon whether the set of categories is complete or anything important would be missing. But I guess that was the same before, just harder to notice in the big mess of icons.

maggu2810:
> I completely feel incapable to decide upon whether the set of categories is complete or anything important would be missing. But I guess that was the same before, just harder to notice in the big mess of icons.

Same here

ThomDietrich:
@kaikreuzer just a friendly request: Could you mention me in such PRs/Issues? Thanks ;)
Great improvement!!

PR #4301, opened 2017-09-19
Add extensible flag for ThingType

### LINKED ISSUE #4146: Add a property to thing types that declares extensibility
A few thing types have a fixed set of defined channels, while others are meant to have a flexible number of channels defined by the user.

We would need a way to identify the "extensibility" of a thing type, so that this can be checked by the framework and nicely be supported by UIs (see e.g. #4142).

Possibly it is also good to give a list of channel-types which are allowed to be added (not all types will be valid in general, I guess).

htreu:
This closes #4146.

- refactor ThingType constructors to builder pattern with new ThingTypeBuilder
- depcrecate all ThingType constructors, replace invocations by builder
- add extensible attribute with list of extensible channel-type ids to thing-type
- expose extensible channel-type id list to REST resource

Also: Migrate org.eclipse.smarthome.core.thing.xml.test from groovy to Java.


bundles/test/org.eclipse.smarthome.test/src/main/java/org/eclipse/smarthome/test/java/JavaOSGiTest.java
```diff
@@ -58,7 +58,7 @@ public void bindBundleContext() {
      *
      * @return bundle context
      */
-    private BundleContext getBundleContext() {
+    protected BundleContext getBundleContext() {
```
maggu2810: This method is called by `bindBundleContext()` that is annotated with `@Before`.

You should access the bundleContext member directly -- it is already `protected`.
Perhaps this method name is confusing and should be named `initBundleContext`.

bundles/core/org.eclipse.smarthome.core.thing/src/main/java/org/eclipse/smarthome/core/thing/type/ThingType.java
```diff
@@ -39,40 +37,35 @@
 
     private final List<ChannelGroupDefinition> channelGroupDefinitions;
     private final List<ChannelDefinition> channelDefinitions;
+    private final List<String> extensibleChannelTypeIds;
     private final List<String> supportedBridgeTypeUIDs;
-    private final Map<@NonNull String, String> properties;
+    private final @NonNull Map<@NonNull String, String> properties;
```
triller-telekom: As @maggu2810 already [suggested](https://github.com/eclipse/smarthome/pull/4301#pullrequestreview-63596110) we should use null annotations. According to [our guidelines](https://github.com/eclipse/smarthome/pull/4275/files#diff-24ba7e763168d1631a43ebdef19bcd5dR27) we should use `@NonNullByDefault` on the class level and remove it here.

htreu:
I will extend this PR by null annotations. Maybe I will have to stop at least at some level...

kaikreuzer:

bundles/core/org.eclipse.smarthome.core.thing.test/src/test/java/org/eclipse/smarthome/core/thing/type/ThingTypeBuilderTest.java
```diff
@@ -0,0 +1,230 @@
...
+    @Test
+    public void whithConfigDescriptionURI_shouldSetConfigDescriptionURI() throws Exception {
+        ThingType thingType = builder.withConfigDescriptionURI(new URI(CONF_URI)).build();
+
+        assertThat(thingType.getConfigDescriptionURI(), is(new URI(CONF_URI)));
+    }
+
+    @Test
+    public void withExtensibleChannelTypeIds_shouldSetUnmodifiableExtensibleChannelTypeIds() {
+        @SuppressWarnings("null")
```
sjka: Is this really needed with the global suppression on class-level?

bundles/core/org.eclipse.smarthome.core.thing.xml.test/build.properties
```diff
@@ -1,4 +1,5 @@
 source.. = src/test/groovy/,\
+		   src/test/java/,\
```
sjka: is `src/test/groovy` still valid? It looks to me like no groovy file is remaining, therefore I'd expect the builder to complain here.

Kudos btw for migrating them all!

bundles/core/org.eclipse.smarthome.core.thing.xml/src/main/java/org/eclipse/smarthome/core/thing/xml/internal/ThingTypeConverter.java
```diff
@@ -95,12 +97,22 @@ protected ThingTypeXmlResult unmarshalType(HierarchicalStreamReader reader, Unma
...
         return thingTypeXmlResult;
     }
 
+    protected List<String> getExtensibleChannelTypeIds(Map<String, String> attributes) {
+        String extensible = attributes.get("extensible");
+        if (extensible == null) {
+            return Collections.emptyList();
+        }
+
+        return Arrays.asList(extensible.split(","));
```
sjka: Should we maybe also trim the strings?
Keep in mind us poor human beings who were ~~forced~~ taught in school to waste paper with spaces like this `extensible="alarm, brightness, foo, bar"`

bundles/core/org.eclipse.smarthome.core.thing.xml/src/main/java/org/eclipse/smarthome/core/thing/xml/internal/ThingTypeXmlResult.java
```diff
@@ -129,15 +133,22 @@ public ConfigDescription getConfigDescription() {
...
 
-    public ThingType toThingType() throws ConversionException {
-
-        ThingType thingType = new ThingType(this.thingTypeUID, this.supportedBridgeTypeUIDs, this.label,
-                this.description, this.category, this.listed, this.representationProperty,
-                toChannelDefinitions(this.channelTypeReferences),
-                toChannelGroupDefinitions(this.channelGroupTypeReferences), toPropertiesMap(),
-                this.configDescriptionURI);
+    ThingTypeBuilder getBuilder() {
+        return ThingTypeBuilder.instance(thingTypeUID, label) //
```
sjka: good idea 😉

bundles/core/org.eclipse.smarthome.core.thing.xml/thing-description-1.0.0.xsd
```diff
@@ -38,7 +38,14 @@
         </xs:sequence>
         <xs:attribute name="id" type="config-description:idRestrictionPattern" use="required"/>
         <xs:attribute name="listed" type="xs:boolean" default="true" use="optional"/>
+        <xs:attribute name="extensible" type="thing-description:namespaceIdListRestrictionPattern" use="optional"/>
     </xs:complexType>
+    
+    <xs:simpleType name="namespaceIdListRestrictionPattern">
+        <xs:restriction base="xs:string">
+            <xs:pattern value="[A-Za-z0-9\-_.]+(,[A-Za-z0-9\-_.]+)*" />
```
sjka: Also here (see above): Wouldn't be allowing an optional space be nice?

bundles/core/org.eclipse.smarthome.core.thing/src/main/java/org/eclipse/smarthome/core/thing/type/ThingTypeBuilder.java
```diff
@@ -0,0 +1,206 @@
...
+import org.apache.commons.lang.StringUtils;
+import org.eclipse.jdt.annotation.NonNull;
+import org.eclipse.jdt.annotation.NonNullByDefault;
+import org.eclipse.jdt.annotation.Nullable;
+import org.eclipse.smarthome.core.thing.ThingTypeUID;
+
+/**
+ * A {@link ThingType} builder.
+ *
+ * @author Henning Treu - initial contribution
```
sjka: 👍  I like that idea! It gives us an API which is much easier to extend and evolve.

bundles/test/org.eclipse.smarthome.test/src/main/java/org/eclipse/smarthome/test/java/JavaOSGiTest.java
sjka: +1 - would you mind changing the method name accordingly while you are at it anyway?

sjka:
@maggu2810 (this is more a note to myself): this PR exceeds the 1000 LOC, it [must](https://eclipse.org/legal/EclipseLegalProcessPoster.pdf) either be merged by kai/me or needs to undergo an IP review otherwise.

bundles/core/org.eclipse.smarthome.core.thing.test/src/test/java/org/eclipse/smarthome/core/thing/type/ThingTypeBuilderTest.java
htreu: no, obsolete.

bundles/core/org.eclipse.smarthome.core.thing.xml.test/build.properties
htreu: obsolete, indeed.

bundles/core/org.eclipse.smarthome.core.thing.xml/src/main/java/org/eclipse/smarthome/core/thing/xml/internal/ThingTypeConverter.java
htreu: I did actually thought about that. In the XSD I did specify a pattern which does not allow spaces. In case the XML is validated against the XSD this will fail early. But we may also add trimming, better save then sorry :-)

bundles/core/org.eclipse.smarthome.core.thing.xml/thing-description-1.0.0.xsd
htreu: yes, we might allow that, update is on the way.

bundles/test/org.eclipse.smarthome.test/src/main/java/org/eclipse/smarthome/test/java/JavaOSGiTest.java
htreu: agreed.

maggu2810:
> @maggu2810 (this is more a note to myself): this PR exceeds the 1000 LOC, it [must](https://eclipse.org/legal/EclipseLegalProcessPoster.pdf) either be merged by kai/me or needs to undergo an IP review otherwise. 

Figure 2?

maggu2810:
Shouldn't we add a documentation about the extensible flag in this PR?

sjka:
> Figure 2?

Yes, it's funny:
* If I merge it (same employer): Figure 2 
* If you merge it: Figure 3 ending up at "Go To: Everything Else (Figure 6)"

Just to make this very clear: I didn't want to imply that you should not _review_ it. I just wanted to point out that this shortcut exists where we can benefit from as I can't remember that we ran into this "Figure 2" situation before.

htreu:
omg, we have a `"Figure 2" situation` ;-)

I will update the documentation, stay tuned.

sjka:
Thanks for the docu update. Having gone through the code, it looks good to me! And conceptually I also don't see any controversy. Therefore: Thanks!

PR #4304, opened 2017-09-19
Renamed a few packages to internal in the sample bundles.

VelinYordanov:


kaikreuzer:
Hm, that was a bad timing - @maggu2810's https://github.com/eclipse/smarthome/pull/4308 shuffled all our manifests, so I am afraid you will need to rebase all your PRs that moved classes (and thus also touch a manifest). Sorry for that!

PR #4308, opened 2017-09-20
RFC: Rework manifest

maggu2810:
Yesterday I removed my old openHAB IDE setup and started a fresh checkout using the openHAB setup and choose every repo.
After the IDE has been started and all repos has been cloned I realized some error messages that are related to the optional nullness annotation support that is not present on all bundles:
* The project was not built since its build path is incomplete. Cannot find the class file for org.eclipse.jdt.annotation.NonNullByDefault.
* The type org.eclipse.jdt.annotation.NonNullByDefault cannot be resolved. It is indirectly referenced from required .class files	...

For the ESH bundles I added the JTD annotation import manually: https://github.com/eclipse/smarthome/pull/4280

Now I have written a small tool (it is not finished yet) that read every manifest, adds the optional import if it is missing and apply some other cleanups:
* sort Import-Package, Export-Package and Require-Bundle section alphabetical
* remove duplicates from the above sections
* use "niceness" to add line breaks for some lists (e.g. the sections above and some others)

As a side effect the attributes (except the manifest version) seems to be ordered alphabetical, too.

The tool has been written just for personal interest to know how to automate such a task and to support the openHAB repositories to add that import to all that manifests.
I don't know if we should apply this for the ESH manifests, too. Perhaps some of them looks better, but I leave it up to you.

### DIFF
--- a/bundles/auth/org.eclipse.smarthome.auth.jaas/META-INF/MANIFEST.MF
+++ b/bundles/auth/org.eclipse.smarthome.auth.jaas/META-INF/MANIFEST.MF
@@ -1,14 +1,15 @@
 Manifest-Version: 1.0
+Bundle-ActivationPolicy: lazy
+Bundle-License: http://www.eclipse.org/legal/epl-v10.html
+Bundle-ManifestVersion: 2
 Bundle-Name: Eclipse SmartHome JAAS Auth
 Bundle-RequiredExecutionEnvironment: JavaSE-1.8
+Bundle-SymbolicName: org.eclipse.smarthome.auth.jaas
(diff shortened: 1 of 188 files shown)


kaikreuzer:
> to support the openHAB repositories to add that import to all that manifests.

That's excellent, many thanks for your efforts - looking forward to the according PRs.

Wrt ESH: I'd be ok to apply it here, too - it all looks nicely sorted. Not sure on the (rebase required) impact on open PRs, though; if that is an issue we could also do without that change.

maggu2810:
If you have had a short look over the files and Travis CI succeeded, I assume we could merge it as we would have a consistent layout for all manifest files then.

PR #4311, opened 2017-09-20
Renamed a few packages to internal in the sonos binding.

VelinYordanov:

### DIFF
--- a/extensions/binding/org.eclipse.smarthome.binding.sonos/META-INF/MANIFEST.MF
+++ b/extensions/binding/org.eclipse.smarthome.binding.sonos/META-INF/MANIFEST.MF
@@ -7,15 +7,10 @@ Bundle-RequiredExecutionEnvironment: JavaSE-1.8
 Bundle-SymbolicName: org.eclipse.smarthome.binding.sonos
 Bundle-Vendor: Eclipse.org/SmartHome
 Bundle-Version: 0.9.0.qualifier
-Export-Package: 
- org.eclipse.smarthome.binding.sonos,
- org.eclipse.smarthome.binding.sonos.handler
 Import-Package: 
  org.apache.commons.io,
  org.apache.commons.lang,
  org.eclipse.jdt.annotation;resolution:=optional,
- org.eclipse.smarthome.binding.sonos,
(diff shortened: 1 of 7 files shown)


maggu2810:
Without an opened IDE: Couldn't every handler be moved to internal, so also the `ZonePlayerHandler` one?

kaikreuzer:
Hm, you guys were too quick for me.

I was just about to point out that [our archetype creates the handlers in exactly this exported package](https://github.com/eclipse/smarthome/tree/master/tools/archetype/binding/src/main/resources/archetype-resources/src/main/java/handler), so unless we change the definition there, I don't think that we should move any handlers to internal packages.

As a background: The archetype had been designed this way to allow the possibility to extend bindings through other bundles, e.g. for a more specific device support - such a bundle could extend an existing (generic) handler and add some specific handling for certain devices.

maggu2810:
Hi Kai,
you are right, the archetype currently uses an exported handler package.
But should every bundle export its handler also if the handler is not been designed to be extended only because perhaps someone uses this in a binding?
So, shouldn't be the default that handlers are not exported as long as someone wants to do it?

kaikreuzer:
> if the handler is not been designed to be extended 

Well, the point is that as it is in an exported package, it is assumed that it is possible to extend it.
Every handler implements the same interface and as such it should always be possible to extend the implementation by some special processing. The "original" handler does not even have to know that it is/can be extended.

> as long as someone wants to do it?

The problem here is that if this "someone" is a solution that builds on an (already released) ESH, it would be too late to do the refactoring for the specific binding. That's why it is imho nicer to design it in a way that this option is available from the start.

I acknowledge that this feature isn't yet used by any binding, so its value hasn't been proven. Nonetheless, I like it from an architectural perspective and do not see a strong reason to change it - but I am open to discuss it; we just should make sure that all bindings are in line with our approach.

maggu2810:
I assume that the API of the most specific handlers is also using internal data types. For example the Zone Player handler of the Sonos binding introduce public methods that uses (e.g. return) `SonosMetaData` that itself is not exported and so bundle private.
As long as the handler are exported because the intention is that a solution can extend them (without using fragments) we need to ensure that no internal package is part of its API.

PR #4312, opened 2017-09-20
Renamed the discovery package to internal in the wemo binding.

VelinYordanov:


maggu2810:
The Wemo test is broken now, so as you changed Wemo binding stuff, do you think it could be related?

VelinYordanov:
@maggu2810 Should be ok now.

PR #4314, opened 2017-09-21
Dynamic say config

sjka:
* migrated DS XMLs to annotations in that the automation.module.media bundle
* converted say action to java provider in order to support selecting the audio sink based on a dynamic list.


triller-telekom:
LGTM

PR #4320, opened 2017-09-22
add nullness handling to storage and its implementations

maggu2810:
A value of a key could be potential null. It depends on the respective implementation if it differs between a existing key that value is null or a nullable value states that the key does not exist.

Do you agree?

### DIFF
--- a/bundles/storage/org.eclipse.smarthome.storage.json/src/main/java/org/eclipse/smarthome/storage/json/internal/JsonStorage.java
+++ b/bundles/storage/org.eclipse.smarthome.storage.json/src/main/java/org/eclipse/smarthome/storage/json/internal/JsonStorage.java
@@ -20,7 +20,8 @@ import java.util.Timer;
 import java.util.TimerTask;
 import java.util.concurrent.ConcurrentHashMap;
 
-import org.eclipse.jdt.annotation.NonNull;
+import org.eclipse.jdt.annotation.NonNullByDefault;
+import org.eclipse.jdt.annotation.Nullable;
 import org.eclipse.smarthome.config.core.Configuration;
 import org.eclipse.smarthome.core.storage.Storage;
 import org.slf4j.Logger;
@@ -44,6 +45,7 @@ import com.google.gson.JsonSyntaxException;
  * @author Stefan Triller - Removed dependency to internal GSon packages
  * @author Simon Kaufmann - Distinguish between inner and outer de-/serialization, keep json structures in map
  */
+@NonNullByDefault
 public class JsonStorage<T> implements Storage<T> {
 
     private final Logger logger = LoggerFactory.getLogger(JsonStorage.class);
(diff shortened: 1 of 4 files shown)


bundles/storage/org.eclipse.smarthome.storage.json/src/main/java/org/eclipse/smarthome/storage/json/internal/JsonStorage.java
```diff
@@ -21,6 +21,8 @@
 import java.util.concurrent.ConcurrentHashMap;
 
 import org.eclipse.jdt.annotation.NonNull;
```
triller-telekom: There is still @NonNull in use in this class, which should be superseded by @NonNullByDefault.

bundles/storage/org.eclipse.smarthome.storage.mapdb/src/main/java/org/eclipse/smarthome/storage/mapdb/internal/MapDbStorage.java
```diff
@@ -9,10 +9,13 @@
 
 import java.util.ArrayList;
 import java.util.Collection;
+import java.util.Collections;
 import java.util.HashSet;
 import java.util.Map;
 
 import org.eclipse.jdt.annotation.NonNull;
```
triller-telekom: There is still @NonNull in use in this class, which should be superseded by @NonNullByDefault.

bundles/test/org.eclipse.smarthome.test/src/main/groovy/org/eclipse/smarthome/test/storage/VolatileStorage.java
```diff
@@ -12,6 +12,8 @@
 import java.util.concurrent.ConcurrentHashMap;
 
 import org.eclipse.jdt.annotation.NonNull;
```
triller-telekom: There is still @NonNull in use in this class, which should be superseded by @NonNullByDefault.

maggu2810:
> I agree that a `Map` should by default be annotated as `Map<@NonNull K, @Nullable V)`. But I vaguely remember that some maps do accept `null` as a key?

Sorry, I don't understand. You agree to which comment of mine?
And do you point to a Map usage of the Map interface?

triller-telekom:
I was referring to:

> A value of a key could be potential null. It depends on the respective implementation if it differs between a existing key that value is null or a nullable value states that the key does not exist.
Do you agree?

And I was pointing to wherever implementations of the `Map` interface are used.

And for our `Storage` interface I think it also makes sense to only allow @NonNull keys but @Nullable values.

maggu2810:
> And for our `Storage` interface I think it also makes sense to only allow @NonNull keys but @Nullable values.

Agree.

But this is all about our `Storage` interface that is completely independent of a `Map` interface.
If an implementation uses a `Map` it is up to the implementation if it choose to use one that allows nullable keys or one that does not support it.
So, I was confused that you "agree that a `Map`" but I didn't refer to a map in my comment.

PR #4323, opened 2017-09-22
Configurable ItemChannelLinks

### LINKED ISSUE #583: Add support for properties on ItemChannelLinks

sjka:
This allows adding some configuration properties onto links, e.g. though

```xtext
  Switch someItem {
      channel="a:b:c:channel" [
          foo="bar",
          answer=42,
          always=true
      ]
  }
```

or the REST API.

As a consequence, ItemChannelLinks now also can be updated through the REST API.

❗️ This is API breaking, it adds a `Configuration` argument to `BindingConfigReader.processBindingConfiguration(...)`. I wouldn't expect many implementations out there - apart from the OH1 compatibility layer. Let me know any concerns, than I will think of something different - but I didn't want to clutter the API without a need.

fixes #583

### DIFF
--- a/bundles/io/org.eclipse.smarthome.io.rest.core/src/main/java/org/eclipse/smarthome/io/rest/core/internal/link/ItemChannelLinkResource.java
+++ b/bundles/io/org.eclipse.smarthome.io.rest.core/src/main/java/org/eclipse/smarthome/io/rest/core/internal/link/ItemChannelLinkResource.java
@@ -22,13 +22,13 @@ import javax.ws.rs.core.Response;
 import javax.ws.rs.core.Response.Status;
 import javax.ws.rs.core.UriInfo;
 
+import org.eclipse.smarthome.config.core.Configuration;
 import org.eclipse.smarthome.core.auth.Role;
 import org.eclipse.smarthome.core.thing.ChannelUID;
 import org.eclipse.smarthome.core.thing.link.AbstractLink;
 import org.eclipse.smarthome.core.thing.link.ItemChannelLink;
 import org.eclipse.smarthome.core.thing.link.ItemChannelLinkRegistry;
 import org.eclipse.smarthome.core.thing.link.ThingLinkManager;
-import org.eclipse.smarthome.core.thing.link.dto.AbstractLinkDTO;
 import org.eclipse.smarthome.core.thing.link.dto.ItemChannelLinkDTO;
 import org.eclipse.smarthome.io.rest.JSONResponse;
 import org.eclipse.smarthome.io.rest.RESTResource;
@@ -68,7 +68,7 @@ public class ItemChannelLinkResource implements RESTResource {
     @ApiResponses(value = {
             @ApiResponse(code = 200, message = "OK", response = ItemChannelLinkDTO.class, responseContainer = "Collection") })
     public Response getAll() {
-        Stream<AbstractLinkDTO> linkStream = itemChannelLinkRegistry.getAll().stream().map(this::toBeans);
+        Stream<ItemChannelLinkDTO> linkStream = itemChannelLinkRegistry.getAll().stream().map(this::toBeans);
         return Response.ok(new Stream2JSONInputStream(linkStream)).build();
     }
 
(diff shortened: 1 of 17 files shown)


bundles/io/org.eclipse.smarthome.io.rest.core/src/main/java/org/eclipse/smarthome/io/rest/core/internal/link/ItemChannelLinkResource.java
```diff
@@ -85,10 +85,28 @@ public Response isAutomatic() {
...
+            link = new ItemChannelLink(itemName, new ChannelUID(channelUid), new Configuration());
+        } else {
+            if (bean.channelUID != null && !bean.channelUID.equals(channelUid)) {
+                return Response.status(Status.BAD_REQUEST).build();
+            }
+            if (bean.itemName != null && !bean.itemName.equals(itemName)) {
+                return Response.status(Status.BAD_REQUEST).build();
+            }
+            link = new ItemChannelLink(itemName, new ChannelUID(channelUid), new Configuration(bean.configuration));
+
```
triller-telekom: Unnecessary newline

bundles/model/org.eclipse.smarthome.model.item/src/org/eclipse/smarthome/model/Items.xtext
```diff
@@ -39,7 +39,26 @@ ModelItemType :
...
+ValueType returns ecore::EJavaObject:
+    STRING | NUMBER | BOOLEAN
+;
+
+BOOLEAN returns ecore::EBoolean: 
+    'true' | 'false'
+;
+
+NUMBER returns ecore::EBigDecimal:
+    ID ('.' ID )?
```
triller-telekom: A number can contain any char from `ID`, so also `a-z`, `A-Z`, `_`?

sjka: Stolen from [Thing.xtext](https://github.com/eclipse/smarthome/blob/09d8095efe9933216874e488b379cd9a10aa85e2/bundles/model/org.eclipse.smarthome.model.thing/src/org/eclipse/smarthome/model/thing/Thing.xtext#L88-L90). Why should it be different?

kaikreuzer: @triller-telekom Think about numbers like `-4.52E+6` and you will understand why it is that way :-)

triller-telekom: @kaikreuzer Thank you for this explanation, but I still don't understand why you allow different characters than `e` or `E` and `_` for a number. Or is it just out of convenience reasons that you do not want to define a slightly different `ID` terminal? :)

kaikreuzer:
@sjka A question that I just stumbled over:

As you know, the `channel` property accepts a list of channels e.g.

```
channel="a:b:c:channel, d:e.f:channel, g:h:i:channel"
```
How does your chosen syntax for the configurations fit to this?

sjka:
The same configuration obviously gets applied to all the links, i.e.

```
      channel="a:b:c:channel, d:e:f:channel, g:h:i:channel" [
          foo="bar",
          answer=42,
          always=true
      ]
```

However, as far as I can see it is allowed to use the same binding configuration multiple times, i.e. if there should be different link configurations, then you'd simply split the one string up and make it three different ones like this:

```
      channel="a:b:c:channel" [
          foo="bar",
          answer=42,
          always=true
      ],
      channel="d:e:f:channel" [
          foo="bar",
          answer=43,
          always=true
      ],
      channel="g:h:i:channel" [
          foo="bar",
          answer=44,
          always=true
      ]
```

kaikreuzer:
> as far as I can see it is allowed to use the same binding configuration multiple times

Ok, thanks, that would be a solution. Do you just "believe" that it should work or did you check it?

sjka:
I'm a strong believer in things I have tried out! 
I created #4378 in order to make you become a believer too 😉

kaikreuzer:
Thanks :-)

PR #4325, opened 2017-09-22
fix the REST implementation locations

### LINKED ISSUE #4322: Several REST resources broken
#4225 missed to adapt all `<implementation...` tags, therefore DS doesn't find the classes, amongst these are:

* DiscoveryResource
* BindingResource
* ChannelTypeResource
* ExtensionResource
* ItemResource

While fixing them, we might as well port them to annotations.

sjka:
Due to lack of time I didn't port them to annotations right now, but this at least fixes all the SCR errors I saw during startup.

fixes #4322

### DIFF
--- a/bundles/io/org.eclipse.smarthome.io.rest.core/OSGI-INF/bindingresource.xml
+++ b/bundles/io/org.eclipse.smarthome.io.rest.core/OSGI-INF/bindingresource.xml
@@ -9,10 +9,10 @@
 
 -->
 <scr:component xmlns:scr="http://www.osgi.org/xmlns/scr/v1.1.0" name="org.eclipse.smarthome.io.rest.core.internal.binding.BindingResource">
-   <implementation class="org.eclipse.smarthome.io.rest.core.binding.BindingResource"/>
+   <implementation class="org.eclipse.smarthome.io.rest.core.internal.binding.BindingResource"/>
    <service>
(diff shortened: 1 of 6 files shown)


bundles/io/org.eclipse.smarthome.io.rest.core/OSGI-INF/bindingresource.xml
```diff
@@ -9,10 +9,10 @@
 
 -->
 <scr:component xmlns:scr="http://www.osgi.org/xmlns/scr/v1.1.0" name="org.eclipse.smarthome.io.rest.core.internal.binding.BindingResource">
-   <implementation class="org.eclipse.smarthome.io.rest.core.binding.BindingResource"/>
+   <implementation class="org.eclipse.smarthome.io.rest.core.internal.binding.BindingResource"/>
    <service>
-      <provide interface="org.eclipse.smarthome.io.rest.core.binding.BindingResource"/>
       <provide interface="org.eclipse.smarthome.io.rest.RESTResource"/>
+      <provide interface="org.eclipse.smarthome.io.rest.core.internal.binding.BindingResource"/>
```
kaikreuzer: Question: Does it actually make sense to keep these "provide" lines, even after moving the resources to an internal package? Imho the "provide RESTResource" should be enough, right?

sjka: Good point - nope, they're pretty much useless now.

PR #4329, opened 2017-09-25
mentioned the need for line breaks in docs

kaikreuzer:


maggu2810:
The "ip-validation" fails.

Why are there two Travis CI jobs now?
* continuous-integration/travis-ci/pr
* continuous-integration/travis-ci/push

sjka:
> The "ip-validation" fails.

IP validations fails because it was edited with then web UI, hence the committer was an anonymous account - with the squashing the commit gets my commit anyway. The sign-off line was correct.

> Why are there two Travis CI jobs now?

no idea either

kaikreuzer:
> Why are there two Travis CI jobs now?

Probably because the branch was created by Github editor here in this repo and not in my fork - so it reacted on pr & push - I'd think that the "push" hook should actually not be set for Travis...

PR #4332, opened 2017-09-25
Explain OSGi discovery service in binding documentation

triller-telekom:
Binding developers who are not familiar with OSGi do not know what
"register as an OSGi service" in our documentation means, see
https://github.com/openhab/openhab2-addons/pull/2659#discussion_r140767229

### DIFF
--- a/docs/documentation/development/bindings/discovery-services.md
+++ b/docs/documentation/development/bindings/discovery-services.md
@@ -64,6 +64,7 @@ The following example shows the implementation of the above mentioned methods in
     protected void startBackgroundDiscovery() {
         logger.debug("Start WeMo device background discovery");
         if (wemoDiscoveryJob == null || wemoDiscoveryJob.isCancelled()) {
+            wemoDiscoveryJob = scheduler.scheduleWithFixedDelay(wemoDiscoveryRunnable, 0, refreshInterval, TimeUnit.SECONDS);
             wemoDiscoveryJob = scheduler.scheduleAtFixedRate(wemoDiscoveryRunnable, 0, refreshInterval, TimeUnit.SECONDS);
         }
     }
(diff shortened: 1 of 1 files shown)


triller-telekom:
I also took the freedom to correct a `scheduleAtFixedRate` vs `scheduleWithFixedDelay` in an example in the same file in an additional commit.

docs/documentation/development/bindings/discovery-services.md
```diff
@@ -41,6 +41,16 @@ The following example is taken from the `HueLightDiscoveryService`, it calls `th
...
 
+## Registering as an OSGi service
+
+The `Discovery` class of a binding which implements `AbstractDiscoveryService` should be annotated with
+
+```java
+@Component(service = DiscoveryService.class, immediate = true, configurationPid = "discovery.<bindingID>")
+```
+
+where `<bindingID>` is the package name of the binding, i.e. `astro` for the Astro binding. Such a registered service will be picked up automatically by the framework.
```
sjka: The term "binding id" is known to every binding developer, I guess. IMHO referring to the "package name" is more confusing that what it helps because the first association would be "org.eclipse.smarthome.binding.myBinding".

docs/documentation/development/bindings/discovery-services.md
```diff
@@ -54,6 +64,7 @@ The following example shows the implementation of the above mentioned methods in
     protected void startBackgroundDiscovery() {
         logger.debug("Start WeMo device background discovery");
         if (wemoDiscoveryJob == null || wemoDiscoveryJob.isCancelled()) {
+            wemoDiscoveryJob = scheduler.scheduleWithFixedDelay(wemoDiscoveryRunnable, 0, refreshInterval, TimeUnit.SECONDS);
```
sjka: Don't you want to delete the `scheduleAtFixedRate`-line?

triller-telekom: Of course... Stupid line picking if another line gets changed by accident and you don't want to commit it...

lolodomo:
Internationalization of discovery result should be explained somewhere.

sjka:
@lolodomo happy to accept a follow-up for #4177 😉

PR #4342, opened 2017-09-26
Renamed a package in the org.eclipse.smarthome.core.thing bundle.

VelinYordanov:


VelinYordanov:
@sjka  Very well, should I add a suppression for this bundle then?

sjka:
That would be awesome. I will try to remember removing it once I export the package.

PR #4344, opened 2017-09-26
[Tradfri] Fix color light bug when brightness is less than 2%

ivivanov-bg:


sjka:
What exactly is "the bug"? And could you please also amend `TradfriColorTest` accordingly?

ivivanov-bg:
The bug is that when the brightness of the color bulb is set to 1% - the xyBrightness / 2.54 returns 0 and the brightness value of the created HSB type is 0 and the brightness channel is updated with 0

sjka:
The point is to prevent regressions and to document how it is expected to behave. Otherwise the next one who cleans up some code or refactors it otherwise might run into the same issue without noticing it and then some users/customers will have to discover it.

sjka:
I just noticed that there already is a similar workaround for the brightness in the [TradfriLightHandler](https://github.com/eclipse/smarthome/blob/c6b4126b320e89de792bf595ffc9603431387d00/extensions/binding/org.eclipse.smarthome.binding.tradfri/src/main/java/org/eclipse/smarthome/binding/tradfri/handler/TradfriLightHandler.java#L352). 

It is surprising to have the same workaround applied at different layers, for the brightness in the handler, for the color in TradfriColor. Do you see any chance to unify these?

ivivanov-bg:
Ok - I will.

One question: Any particular reason to have only one light handler and check if the light is a color one instead of inherit it to let's say `ColorLightHandler` ?

sjka:
No, not really - just the huge overlap, I guess. The only places where this distinction is made are in the `onUpdate()` method, afaics. So I also don't see a need for adding a class hierarchy right now. The experience with others (e.g. hue, lifx,...) was the same.

hreichert:
@ivivanov-bg Thanks for fixing, I did not thought of this when writing the calculation.

extensions/binding/org.eclipse.smarthome.binding.tradfri.test/src/test/java/org/eclipse/smarthome/binding/tradfri/internal/TradfriColorTest.java
```diff
@@ -67,35 +84,51 @@ public void testFromHSBTypeKnownGood1() {
...
 
     @Test
     public void testConversionReverse() {
         // convert from HSBType
-        TradfriColor color = TradfriColor.fromHSBType(HSBType.GREEN);
+        TradfriColor color = TradfriColor.fromHSBType(new HSBType("120,100,1"));
         assertNotNull(color);
         assertEquals(0, (int) color.rgbR);
-        assertEquals(254, (int) color.rgbG); // 254 instead of 255 - only approximated calculation
+        assertEquals(2, (int) color.rgbG); // 254 instead of 255 - only approximated calculation
```
htreu: The value is `2` but the comment says 254.

extensions/binding/org.eclipse.smarthome.binding.tradfri.test/src/test/java/org/eclipse/smarthome/binding/tradfri/internal/TradfriColorTest.java
```diff
@@ -67,35 +84,51 @@ public void testFromHSBTypeKnownGood1() {
...
+        assertEquals(0, color.hsbType.getHue().intValue());
+        assertEquals(100, color.hsbType.getSaturation().intValue());
+        assertEquals(1, color.hsbType.getBrightness().intValue());
+    }
 
     @Test
     public void testConversionReverse() {
         // convert from HSBType
-        TradfriColor color = TradfriColor.fromHSBType(HSBType.GREEN);
+        TradfriColor color = TradfriColor.fromHSBType(new HSBType("120,100,1"));
```
htreu: the test is just fine w/o your changes. If the HSB(120,100,1) adds another corner case please add a separate test case.

extensions/binding/org.eclipse.smarthome.binding.tradfri.test/src/test/java/org/eclipse/smarthome/binding/tradfri/internal/TradfriColorTest.java
htreu: see my comment for line 107, I guess the test case is just fine as it is.

extensions/binding/org.eclipse.smarthome.binding.tradfri/src/main/java/org/eclipse/smarthome/binding/tradfri/internal/TradfriColor.java
```diff
@@ -279,5 +279,21 @@ public static PercentType calculateColorTemperature(int xyX, int xyY) {
...
+    
+    
+    /**
+     * Converts the xyBrightness value to PercentType
+     *
+     * @param xyBrightness xy brightness level 0 to 254
+     * @return {@link PercentType} with brightness level (0 = light is off, 1 = lowest, 100 = highest)
+     */
+    public static PercentType xyBrightnessToPercentType(int xyBrightness) {
+        if (xyBrightness > 254) {
```
htreu: out of curiosity: why are values > 254 a problem? All we have to do is adopt the `2.54` in line 296 and we could also deal with 255 or greater. wdyt?

ivivanov-bg: Well - my understanding for a PercentType is to be between 0 and 100. So - if `xyBrightness > 254` then `xyBrightness / 2.54` will return value `> 100` and it will throw `IllegalArgumentException`

htreu: Yes I get that. All we have to adopt is the divisor (2.54) to a value that matches our upper bound (254). In case we move the upper bound to 255 the divisor should be adopted to 2.55 to match the PercentType. I just wanted to know where the 254 originates and why there are possible values grater then 254.

ivivanov-bg: I guess this comes from the IKEA API (not not really sure)

May be @sjka or @kaikreuzer can give more info

hreichert: Yes, the range 0 to 254 is defined for Zigbee ZLL brightness and the IKEA gateway adopts this.

Issue #4317, opened 2017-09-21
Improve unhelpful logging of rule engine


ThomDietrich:
https://github.com/eclipse/smarthome/blob/a4dea98a1c91a36d4dbe6d5d616879988f176459/bundles/model/org.eclipse.smarthome.model.rule.runtime/src/org/eclipse/smarthome/model/rule/runtime/internal/engine/RuleEngineImpl.java#L341-L345

Hello, the above logging lines are often encountered while DSL rule development. Their minimalistic form can sometimes be confusing or non-helpful. One interesting example is:

```
Rule 'myRuleTest': java.lang.Number
```

Also quite common is:

```text
Rule 'myRuleTest': The name '<XCastedExpressionImpl>.blaaa' cannot be resolved to an item or type.
```

I believe this could easily be improved for the sake of user experience. One simple approach would be to distinguish commonly known error reasons and add constructive feedback to the logging message.

@sjka I believe you are the right person to trigger in this matter? What do you think?
@martinvw I think we discussed something similar a while back.

martinvw:
@ThomDietrich yes we should work on this, maybe we can start by collecting some example which lead to those errors and work from there on improving the error handling. If needed we can even pair up around the smarthome day 👍

sjka:
I absolutely agree, it would help a lot if these error messages would be more helpful. However, while this sounds like a trivial task on the first glance, it probably will turn out to be pretty nasty in the end: these error messages are not created by _us_, but by the xtext parser, lexer and the scoping. Therefore I second @martinvw, we need to collect them (ideally a sample rule + error message) and then have a deeper look individually on what can be done in order to support xtext in that situation to generate better error messages. Sometimes there might have been taken a short-cut so that the scoping is clueless what to expect and we can change it to the better. Sometimes we might be able to add our custom validations with also our own error messages. In any case it's gonna be some effort.
Nevertheless, I'm looking forward to meeting you guys at the [Smart Home Day](https://www.eclipsecon.org/europe2017/smarthome) in Ludwigsburg!

rkoshak:
One common misleading error that comes to mind is, for example, when I pass something to a Number Item that cannot be parsed into a Number I end up getting a NullPointerException (NPE) rather than a ParseException (or whatever would be appropriate). In fact, NPE seems to be the default exception thrown for pretty much all type and parsing problems.

[...]

PR #4346, opened 2017-09-26
log a warning if a rule cannot convert an argument to a type

sjka:
...so that sendCommand or postUpdate don't fail with a weird NPE anymore
if e.g. a non-parsable string is attempted to be sent to a number item.

See https://github.com/eclipse/smarthome/issues/4317#issuecomment-331917181

### DIFF
--- a/bundles/model/org.eclipse.smarthome.model.script/src/org/eclipse/smarthome/model/script/actions/BusEvent.java
+++ b/bundles/model/org.eclipse.smarthome.model.script/src/org/eclipse/smarthome/model/script/actions/BusEvent.java
@@ -7,8 +7,10 @@
  */
 package org.eclipse.smarthome.model.script.actions;
 
+import java.util.List;
 import java.util.Map;
 import java.util.Map.Entry;
+import java.util.stream.Collectors;
 
 import org.eclipse.smarthome.core.events.EventPublisher;
 import org.eclipse.smarthome.core.items.GroupItem;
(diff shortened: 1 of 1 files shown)


bundles/model/org.eclipse.smarthome.model.script/src/org/eclipse/smarthome/model/script/actions/BusEvent.java
```diff
@@ -77,14 +79,25 @@ static public Object sendCommand(String itemName, String commandString) {
             try {
                 Item item = registry.getItem(itemName);
                 Command command = TypeParser.parseCommand(item.getAcceptedCommandTypes(), commandString);
-                publisher.post(ItemEventFactory.createCommandEvent(itemName, command));
+                if (command != null) {
+                    publisher.post(ItemEventFactory.createCommandEvent(itemName, command));
+                } else {
+                    LoggerFactory.getLogger(BusEvent.class).warn(
```
triller-telekom: Since the logger is now used at least 2 times (also in line 91), maybe its time for storing it in a local variable than fetching it twice.

bundles/model/org.eclipse.smarthome.model.script/src/org/eclipse/smarthome/model/script/actions/BusEvent.java
```diff
@@ -140,14 +153,24 @@ static public Object postUpdate(String itemName, String stateString) {
             try {
                 Item item = registry.getItem(itemName);
                 State state = TypeParser.parseState(item.getAcceptedDataTypes(), stateString);
-                publisher.post(ItemEventFactory.createStateEvent(itemName, state));
+                if (state != null) {
+                    publisher.post(ItemEventFactory.createStateEvent(itemName, state));
+                } else {
+                    LoggerFactory.getLogger(BusEvent.class).warn(
```
triller-telekom: see above

bundles/model/org.eclipse.smarthome.model.script/src/org/eclipse/smarthome/model/script/actions/BusEvent.java
```diff
@@ -140,14 +153,24 @@ static public Object postUpdate(String itemName, String stateString) {
...
+                            getAcceptedDataTypeNames(item));
+                }
             } catch (ItemNotFoundException e) {
                 LoggerFactory.getLogger(BusEvent.class).warn("Item '{}' does not exist.", itemName);
             }
         }
         return null;
     }
 
+    private static <T extends State> List<String> getAcceptedDataTypeNames(Item item) {
```
triller-telekom: It is a pity that you have to double this method via copy&paste. Isn't there a common place for it that can be accessed from both classes?

sjka: place it not the problem. Stupid java is:


"The method theCommondMethodThatWeAllWantedToWrite(List<Class<? extends Type>>) in the type BusEvent is not applicable for the arguments (List<Class<? extends Command>>)"

bundles/model/org.eclipse.smarthome.model.script/src/org/eclipse/smarthome/model/script/actions/BusEvent.java
sjka: hmmm, these are both exceptional cases which will occur rarely to never. I'm not really convinced that this justifies a static field. Remember, OSGi is a _dynamic_ module system.

PR #4347, opened 2017-09-26
Update WemoLightHandler subscription handling

### LINKED ISSUE #4282: Wemo Light Handler: remove subscription
@hmerk Hi, I have a question about the current `removeSubscription` implementation.

```Java
    private synchronized void removeSubscription() {
        logger.debug("Removing WeMo GENA subscription for '{}'", this);
        if (service.isRegistered(this)) {
            String subscription = null;

            if ((subscriptionState.get(subscription) != null) && subscriptionState.get(subscription).booleanValue()) {
                logger.debug("WeMo {}: Unsubscribing from service {}...", getUDN(), subscription);
                service.removeSubscription(this, "bridge1");
            }

            subscriptionState = new HashMap<String, Boolean>();
            service.unregisterParticipant(this);

        }
    }
```

You set `subscription` to `null` and use this as a key for the `subscriptionState` map.
A key of a map is IMHO non null by default, so what do you want to check here?
What am I missing here?

hmerk:
Closes #4282

Signed-off-by: Hans-Jörg Merk hans-joerg.merk@t-online.de

### DIFF
--- a/extensions/binding/org.eclipse.smarthome.binding.wemo/src/main/java/org/eclipse/smarthome/binding/wemo/handler/WemoLightHandler.java
+++ b/extensions/binding/org.eclipse.smarthome.binding.wemo/src/main/java/org/eclipse/smarthome/binding/wemo/handler/WemoLightHandler.java
@@ -64,6 +64,8 @@ public class WemoLightHandler extends BaseThingHandler implements UpnpIOParticip
      */
     private static final int DIM_STEPSIZE = 5;
 
+    protected final static String SUBSCRIPTION = "bridge1";
+
     protected final static int SUBSCRIPTION_DURATION = 600;
 
     /**
@@ -378,12 +380,10 @@ public class WemoLightHandler extends BaseThingHandler implements UpnpIOParticip
         if (service.isRegistered(this)) {
             logger.debug("Checking WeMo GENA subscription for '{}'", this);
 
-            String subscription = "bridge1";
-
-            if ((subscriptionState.get(subscription) == null) || !subscriptionState.get(subscription).booleanValue()) {
-                logger.debug("Setting up GENA subscription {}: Subscribing to service {}...", getUDN(), subscription);
-                service.addSubscription(this, subscription, SUBSCRIPTION_DURATION);
-                subscriptionState.put(subscription, true);
+            if ((subscriptionState.get(SUBSCRIPTION) == null) || !subscriptionState.get(SUBSCRIPTION).booleanValue()) {
+                logger.debug("Setting up GENA subscription {}: Subscribing to service {}...", getUDN(), SUBSCRIPTION);
+                service.addSubscription(this, SUBSCRIPTION, SUBSCRIPTION_DURATION);
+                subscriptionState.put(SUBSCRIPTION, true);
             }
 
         } else {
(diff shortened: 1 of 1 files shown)


extensions/binding/org.eclipse.smarthome.binding.wemo/src/main/java/org/eclipse/smarthome/binding/wemo/handler/WemoLightHandler.java
```diff
@@ -393,13 +393,13 @@ private synchronized void onSubscription() {
     }
 
     private synchronized void removeSubscription() {
-        logger.debug("Removing WeMo GENA subscription for '{}'", this);
         if (service.isRegistered(this)) {
-            String subscription = null;
+        logger.debug("Removing WeMo GENA subscription for '{}'", this);
```
triller-telekom: I think the formatter did not work correctly here, there is an indentation level missing.

extensions/binding/org.eclipse.smarthome.binding.wemo/src/main/java/org/eclipse/smarthome/binding/wemo/handler/WemoLightHandler.java
```diff
@@ -393,13 +393,13 @@ private synchronized void onSubscription() {
...
+            String subscription = "bridge1";
```
triller-telekom: While you are extracting this "magic" string, I think you should make it a constant with capital letters on top of the class, so it is easier to find.

hmerk: Sure, will change it to a constant.

extensions/binding/org.eclipse.smarthome.binding.wemo/src/main/java/org/eclipse/smarthome/binding/wemo/handler/WemoLightHandler.java
hmerk: Indeed, will fix it

triller-telekom:
@hmerk: Do you have this PR still on your list? :)

hmerk:
@triller-telekom Yes, still on my list but actually not finding time for any coding.

hmerk:
@triller-telekom 
I finally found time to get this sorted out and already squashed my changes.
Hope it is ok now.

triller-telekom:
@hmerk: Thank you very much!

extensions/binding/org.eclipse.smarthome.binding.wemo/src/main/java/org/eclipse/smarthome/binding/wemo/handler/WemoLightHandler.java
```diff
@@ -64,6 +64,8 @@
      */
     private static final int DIM_STEPSIZE = 5;
 
+    protected final static String SUBSCRIPTION = "bridge1";
```
kaikreuzer: @hmerk Just for my understanding: Can you elaborate what "bridge1" here exactly means and whether this also works, if you have multiple ESH instances OR multiple WeMo devices? I see that this constant was used before your PR already, but it simply looks fishy to me...

hmerk:
"bridge1" is the name of the eventservice we subscribe to. It's name is definde in the devices firmware, like "basicevent1" for switches or "insight1" for Insight switches.
I will create another PR next week to give the constant a more meaningfull name like EVENTSERVICE and add a short info to the code what it is about.
It has no influence to multiple devices or instances.

PR #4218, opened 2017-09-08
Core/Network: Allow IPv6 for default network interface. 

davidgraeff:
Clean up NetUtils. Fix NetworkConfigOptionProvider.

Fixes #4205

Signed-off-by: David Gräff <david.graeff@web.de>

PR #4348, opened 2017-09-26
prevent NPE if no network interface is present

kaikreuzer:
If running on an offline host without network interfaces, we currently see exceptions like
```
	at org.eclipse.smarthome.core.net.NetUtil.getPrimaryIpv4HostAddress(NetUtil.java:71) ~[?:?]
	at org.openhab.ui.dashboard.internal.DashboardService.activate(DashboardService.java:76) ~[?:?]
```

This fix should prevent this (and the Javadoc also already specifically notes that null is returned if not interface is present).

### DIFF
--- a/bundles/core/org.eclipse.smarthome.core/src/main/java/org/eclipse/smarthome/core/net/NetUtil.java
+++ b/bundles/core/org.eclipse.smarthome.core/src/main/java/org/eclipse/smarthome/core/net/NetUtil.java
@@ -66,6 +66,11 @@ public class NetUtil implements NetworkAddressService {
 
     @Override
     public String getPrimaryIpv4HostAddress() {
+        if (primaryAddress == null) {
+            // we do not seem to have any network interfaces
+            return null;
+        }
+
         String primaryIP;
 
         String[] addrString = primaryAddress.split("/");


kaikreuzer:
Being offline? That's a paramount use case!

triller-telekom:
This PR will create merge conflicts with #4218 which has just been rebased and its test cases fixed. maybe we want to merge that one first and then apply this fix?

davidgraeff:
@maggu2810 We need to decide if we want proper IPv6 support and also consider the DHCP address change, like in the other PR implemented. If we decide to have these, then `getPrimaryIpv4HostAddress()` might return null if the machine is offline.

maggu2810:
If the machine is offline it is okay for me to return `null`.
But if the machine has been offline if the system starts (service `activate` and `modified` has been called) and request later the primary host address (and the system is online now), I would expect to get the current one and not the latest checked one (so null, because it has been offline at that time).

davidgraeff:
Oh that works as you expect it in the other PR. The configuration value is stored in a field. And the actual matching with an interface is performed when an address is requested.

kaikreuzer:
So do I get it right that this PR should be closed in favor of #4218? Let's do so then, but make sure that #4218 gets merged :-)

Issue #4349, opened 2017-09-26
[Tradfri] Bug in color space conversion


hreichert:
As reported in the OpenHAB community:
https://community.openhab.org/t/ikea-tradfri-gateway/26135/126

> when I change the color in the IKEA Trådfri iOS App to the darkest blue, I get the following error:

```
==> /var/log/openhab2/openhab.log <==
2017-09-26 19:17:37.493 [ERROR] [alifornium.core.network.CoapEndpoint] - Exception in protocol stage thread: Value must be between 0 and 100
java.lang.IllegalArgumentException: Value must be between 0 and 100
	at org.eclipse.smarthome.core.library.types.PercentType.validateValue(PercentType.java:50) [98:org.eclipse.smarthome.core:0.9.0.201709260841]
	at org.eclipse.smarthome.core.library.types.PercentType.<init>(PercentType.java:35) [98:org.eclipse.smarthome.core:0.9.0.201709260841]
	at org.eclipse.smarthome.core.library.types.HSBType.fromRGB(HSBType.java:114) [98:org.eclipse.smarthome.core:0.9.0.201709260841]
	at org.eclipse.smarthome.binding.tradfri.internal.TradfriColor.constructHsbTypeFromRgbWithBrightnessPercent(TradfriColor.java:247) [186:org.eclipse.smarthome.binding.tradfri:0.9.0.201709260841]
	at org.eclipse.smarthome.binding.tradfri.internal.TradfriColor.fromCie(TradfriColor.java:133) [186:org.eclipse.smarthome.binding.tradfri:0.9.0.201709260841]
	at org.eclipse.smarthome.binding.tradfri.handler.TradfriLightHandler$LightData.getColor(TradfriLightHandler.java:423) [186:org.eclipse.smarthome.binding.tradfri:0.9.0.201709260841]
	at org.eclipse.smarthome.binding.tradfri.handler.TradfriLightHandler.onUpdate(TradfriLightHandler.java:140) [186:org.eclipse.smarthome.binding.tradfri:0.9.0.201709260841]
	at org.eclipse.smarthome.binding.tradfri.internal.TradfriCoapHandler.onLoad(TradfriCoapHandler.java:60) [186:org.eclipse.smarthome.binding.tradfri:0.9.0.201709260841]
	at org.eclipse.californium.core.CoapClient$ObserveMessageObserverImpl.deliver(CoapClient.java:1067) [175:californium-osgi:1.0.6]
[...]

PR #4351, opened 2017-09-26
[Tradfri] Fix bug in color space conversion

hreichert:
Ensure that calculated RGB values are not negative.
Added new unit test case.

Fixes: #4349

### DIFF
--- a/extensions/binding/org.eclipse.smarthome.binding.tradfri/src/main/java/org/eclipse/smarthome/binding/tradfri/internal/TradfriColor.java
+++ b/extensions/binding/org.eclipse.smarthome.binding.tradfri/src/main/java/org/eclipse/smarthome/binding/tradfri/internal/TradfriColor.java
@@ -124,6 +124,11 @@ public class TradfriColor {
         // green = green <= 0.0031308 ? 12.92 * green : (1.0 + 0.055) * Math.pow(green, (1.0 / 2.4)) - 0.055;
         // blue = blue <= 0.0031308 ? 12.92 * blue : (1.0 + 0.055) * Math.pow(blue, (1.0 / 2.4)) - 0.055;
 
+        // calculated values can be slightly negative, so cap them to minimum 0.0
+        red = Math.max(0.0, red);
+        green = Math.max(0.0, green);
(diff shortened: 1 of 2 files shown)


extensions/binding/org.eclipse.smarthome.binding.tradfri/src/main/java/org/eclipse/smarthome/binding/tradfri/internal/TradfriColor.java
```diff
@@ -124,9 +124,9 @@ public static TradfriColor fromCie(int xyX, int xyY, int xyBrightness) {
         // green = green <= 0.0031308 ? 12.92 * green : (1.0 + 0.055) * Math.pow(green, (1.0 / 2.4)) - 0.055;
         // blue = blue <= 0.0031308 ? 12.92 * blue : (1.0 + 0.055) * Math.pow(blue, (1.0 / 2.4)) - 0.055;
 
-        int redRounded = (int) Math.round(red * 255.0);
-        int greenRounded = (int) Math.round(green * 255.0);
-        int blueRounded = (int) Math.round(blue * 255.0);
+        int redRounded = ensureRgbRange((int) Math.round(red * 255.0));
```
triller-telekom: I wonder if this is the right place for the fix. Shouldn't we make sure that red/green/blue is not negative and not above 1? At least the later is already checked in lines 108-120 so you would only need to check for negative numbers.

hreichert: Thanks for your comment, that would be indeed better.
I'm going to change this and rebase on master, as there are now duplicate method names in the tests I guess.

hreichert: Done.

hreichert:
Updated PR with suggested change from @triller-telekom

sjka:
Thanks!

PR #4352, opened 2017-09-27
MQTT: Fix wildcard subscribers

### LINKED ISSUE #4333: MQTT: Fix wildcard topics
Following is the implementation of addconsumer. Problem that i am facing is the line with <<<-----. I treplaces all '+' with regex format which i think MQTT broker like mosquitto do not understand. I have also attached an output from mosquitto when a topic has '+' character in it.

> public boolean addConsumer(MqttMessageSubscriber subscriber) throws MqttException {
>         String topic = subscriber.getTopic();
>         // Prepare topic for regex pattern matching taking place in messageArrived.
>         topic = StringUtils.replace(StringUtils.replace(topic, "+", "[^/]*"), "#", ".*");  <<<------
>         synchronized (consumers) {
>             List<MqttMessageSubscriber> subscriberList = consumers.get(topic);
>             if (subscriberList == null) {
>                 subscriberList = new ArrayList<>();
>             }
>             subscriberList.add(subscriber);
>             consumers.put(topic, subscriberList);
>         }
>         if (isConnected()) {
>             try {
>                 client.subscribe(topic, qos);
>             } catch (org.eclipse.paho.client.mqttv3.MqttException e) {
>                 throw new MqttException(e);
>             }
>         }
>         return true;
>     }

**MOSQUITTO.log**
1506352001: Received SUBSCRIBE from openhab
1506352001:     cctvchannel/motion_alert/2/2/[^/]*/reply (QoS 0)
1506352001: openhab 0 cctvchannel/motion_alert/2/2/[^/]*/reply
1506352001: Sending SUBACK to openhab
1506352001: Received SUBSCRIBE from openhab
1506352001:     cctvchannel/motion_alert/2/3/[^/]*/reply (QoS 0)
1506352001: openhab 0 cctvchannel/motion_alert/2/3/[^/]*/reply
1506352001: Sending SUBACK to openhab
1506352006: Received SUBSCRIBE from openhab

davidgraeff:
Fixes #4333

### DIFF
--- a/bundles/io/org.eclipse.smarthome.io.transport.mqtt/src/main/java/org/eclipse/smarthome/io/transport/mqtt/MqttBrokerConnection.java
+++ b/bundles/io/org.eclipse.smarthome.io.transport.mqtt/src/main/java/org/eclipse/smarthome/io/transport/mqtt/MqttBrokerConnection.java
@@ -347,9 +347,8 @@ public class MqttBrokerConnection {
      * @throws MqttException If connected and the subscribe fails, this exception is thrown.
      */
     public boolean addConsumer(MqttMessageSubscriber subscriber) throws MqttException {
-        String topic = subscriber.getTopic();
         // Prepare topic for regex pattern matching taking place in messageArrived.
-        topic = StringUtils.replace(StringUtils.replace(topic, "+", "[^/]*"), "#", ".*");
+        String topic = StringUtils.replace(StringUtils.replace(subscriber.getTopic(), "+", "[^/]*"), "#", ".*");
         synchronized (consumers) {
             List<MqttMessageSubscriber> subscriberList = consumers.get(topic);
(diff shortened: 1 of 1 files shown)


maggu2810:
> I cant judge if this is MQTT technically OK.

It should :wink: 

@htreu The fix delegate the unmodified topic to the subscribe call. So, should we create a test checks that the client's subscribe method is called with the return value of `getTopic()` of the supplied `MqttMessageSubscriber`. I find tests very useful, but I don't know if it make sense in this case.
@davidgraeff WDYT?

davidgraeff:
Let's fix this bug first

htreu:
@maggu2810 yes, you are right the PR does indeed fix the call to client#subscribe. My comment was about the String replacement which was not obvious to me. But not part of this PR, so all is fine :-)

PR #4355, opened 2017-09-27
[documentation] Changed 'bridge-type'-tag to 'thing-type'-tag

cweitkamp:
- Changed 'bridge-type'-tag to 'thing-type'-tag
- Changed some spellings

### DIFF
--- a/docs/documentation/development/bindings/thing-definition.md
+++ b/docs/documentation/development/bindings/thing-definition.md
@@ -60,7 +60,7 @@ The channel type definition is specified on the same level as the thing type def
 That way channels can be reused in different things.
 
 The granularity of channel types should be on its semantic level, i.e. very fine-grained:
-If a Thing measures two temperature values, one for indoor and one for outdoor, this should be modelled as two different channel types.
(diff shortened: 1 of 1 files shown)


docs/documentation/development/bindings/thing-definition.md
```diff
@@ -310,15 +310,15 @@ Having this property identified per binding it could be used as the `representat
 
 The `representation property` will be defined in the thing type XML: 
 
-```xml 
- <bridge-type id="thingTypeId">
+```xml
```
afuechsel: Well, this was a copy/paste from the hue bridge thing type. Therefore it was bridge type. Not sure, if this should be changed.

afuechsel: Of course the closing tag should also be bridge type.

cweitkamp: The `representation property` is valid for both. It can be used to identify bridge types as well as thing types. So it doesn't matter which we use in this example. Imho the thing type fits better, because this page is about things. ;-)

PR #4359, opened 2017-09-29
PaperUI: allow extension channels on things

### LINKED ISSUE #4142: PaperUI: add new channels to things
For generic bindings like KNX (see https://github.com/openhab/openhab2-addons/pull/2323), MQTT (see #3876) and others, where things are used to structure generic things with a user defined sub-list of channels the PaperUI should provide a feature to add channels to things.
This results in the following tasks:
- identify if its allowed to add channels to a specific thing
- provide an "Add channel" button to those things
- Select one of the available channel-types for the thing
- Edit the channel configuration
- Add the channel and configuration to the thing via REST

Only the "identify if channels can be added to thing" part needs changes in the core framework. The first idea is to provide a flag on the thing-type definition to indicate this.

htreu:
Closes #4142.

Also: Refactor the configuration module to separate the things controllers and views.
Also: Refactor the things controllers and split them into separate files.

For review: To better differentiate between refactoring and new functionality try to go though this commit by commit.


sjka:
Works nicely as far as I can see.

When trying it out I was wondering whether we should define an "extensible" thing-type in the magic binding. I think it would be helpful. But maybe in a separate PR, in order to not further blow this one up. Wdyt?

htreu:
I will supply an update to the magic bundle asap.

htreu:
See #4372 for the magic PR ;-)

PR #4363, opened 2017-10-01
[hue] Fix incorrect usage of trailing slash in hue API call

### LINKED ISSUE #4339: Hue-Binding is breaking the Hue API rules
I'm not be able to connect my software based Hue Bridge (https://github.com/mariusmotea/diyHue) with the Hue binding to openHAB. The binding will create a new user, but the status is always changing from offline to online and back to offline immediately. 
The log of my software based Hue bridge shows that the Hue binding is trying to access the bridge as follows: "/api/username/". According to the Hue API it should be only "/api/username". Even this is working with real Hue bridges the last "/" is breaking the API protocol rules and is causing an error on the software based bridge because a following command for example like "lights" etc. is missing.

dominicdesu:
Fixes #4339

Signed-off-by: Dominic Lerbs <dominicdesu@hallojapan.de>

### DIFF
--- a/extensions/binding/org.eclipse.smarthome.binding.hue/src/main/java/org/eclipse/smarthome/binding/hue/internal/HueBridge.java
+++ b/extensions/binding/org.eclipse.smarthome.binding.hue/src/main/java/org/eclipse/smarthome/binding/hue/internal/HueBridge.java
@@ -903,10 +903,13 @@ public class HueBridge {
     }
 
     private String getRelativeURL(String path) {
-        if (username == null) {
-            return "http://" + ip + "/api/" + path;
-        } else {
-            return "http://" + ip + "/api/" + enc(username) + "/" + path;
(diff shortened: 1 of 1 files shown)


sjka:
When reading this, I personally doubt that this is designed right when thinking about the original ideas of RESTful applications as here it's not the whole collection that gets replaced but instead something is added to an existing collection. 

Nevertheless, I see that this actually _is_ according to the Hue API documentation and I can also confirm that it _still_ works nicely. Therefore I agree, it makes sense to stick to the documentation precisely.

sjka:
Thanks!

sjka:
Ouch, I missed to see that Travis CI did not pick up this PR. And promptly you failed to run & adapt the tests which relied on the trailing slash in the stubs. 

I quickly created #4365 to get our build running again.

dominicdesu:
@sjka : Sorry and thanks!

PR #4366, opened 2017-10-02
[Tradfri] Added FLOALT panels

cweitkamp:
- Added "FLOALT" panels and "TRADFRI bulb E27 WS clear 950lm" to `COLOR_TEMP_MODELS`
- Added representation-property to thing types
- Refactored xml files and applied formatter on xml files
- Changed spelling of TRÅDFRI throughout the whole binding

After you added a FLOALT panel to your environment it is correclty discovered as 0220 thing type. Once it becomes not reachable it is additionally discovered as 0100 thing type. This fix should prevent this behavior.

### DIFF
--- a/extensions/binding/org.eclipse.smarthome.binding.tradfri/ESH-INF/config/config.xml
+++ b/extensions/binding/org.eclipse.smarthome.binding.tradfri/ESH-INF/config/config.xml
@@ -1,0 +1,17 @@
+<?xml version="1.0" encoding="UTF-8"?>
+<config-description:config-descriptions
+	xmlns:xsi="http://www.w3.org/2001/XMLSchema-instance" xmlns:config-description="http://eclipse.org/smarthome/schemas/config-description/v1.0.0"
+	xsi:schemaLocation="http://eclipse.org/smarthome/schemas/config-description/v1.0.0 http://eclipse.org/smarthome/schemas/config-description-1.0.0.xsd">
+
+	<config-description uri="thing-type:tradfri:device">
+		<parameter-group name="device">
+			<label>Device</label>
+			<description>Device specific settings.</description>
+		</parameter-group>
+		<parameter name="id" type="integer" required="true" groupName="device">
+			<label>ID</label>
+			<description>The identifier of the device on the gateway.</description>
+		</parameter>
+	</config-description>
+
+</config-description:config-descriptions>
--- a/extensions/binding/org.eclipse.smarthome.binding.tradfri/ESH-INF/config/config.xml
+++ b/extensions/binding/org.eclipse.smarthome.binding.tradfri/ESH-INF/config/config.xml
@@ -4,11 +4,7 @@
 	xsi:schemaLocation="http://eclipse.org/smarthome/schemas/config-description/v1.0.0 http://eclipse.org/smarthome/schemas/config-description-1.0.0.xsd">
 
 	<config-description uri="thing-type:tradfri:device">
-		<parameter-group name="device">
-			<label>Device</label>
-			<description>Device specific settings.</description>
-		</parameter-group>
-		<parameter name="id" type="integer" required="true" groupName="device">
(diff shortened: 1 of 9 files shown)


kaikreuzer:
Thanks, great to see support for the FLOALT devices.

The behavior worries me a bit though: We will see the same for every new upcoming device that we do not yet specifically support. Shouldn't we make sure that we do not add additional discovery results? The recently introduced representation-property of things might probably help here. Would you want to have a look into that?

cweitkamp:
I added the representation-property to the thing type definition. But I am not sure if this is the solution for the real problem (see comment in [TradfriDiscoveryService#82](https://github.com/eclipse/smarthome/pull/4366/files#diff-7e60db33554b2d3a2f4f6f3bb82c6581R82)).

extensions/binding/org.eclipse.smarthome.binding.tradfri/ESH-INF/config/config.xml
```diff
@@ -0,0 +1,17 @@
+<?xml version="1.0" encoding="UTF-8"?>
+<config-description:config-descriptions
+	xmlns:xsi="http://www.w3.org/2001/XMLSchema-instance" xmlns:config-description="http://eclipse.org/smarthome/schemas/config-description/v1.0.0"
+	xsi:schemaLocation="http://eclipse.org/smarthome/schemas/config-description/v1.0.0 http://eclipse.org/smarthome/schemas/config-description-1.0.0.xsd">
+
+	<config-description uri="thing-type:tradfri:device">
+		<parameter-group name="device">
```
kaikreuzer: Please don't use parameter groups if there is just a single parameter to put into it...

extensions/binding/org.eclipse.smarthome.binding.tradfri/ESH-INF/thing/thing-types.xml
```diff
@@ -1,139 +1,127 @@
...
+			<channel id="color" typeId="color" />
+		</channels>
+
+		<representation-property>id</representation-property>
+
+		<config-description-ref uri="thing-type:tradfri:device" />
+	</thing-type>
+
+	<!-- note that this isn't yet supported by the code as we do not receive any data 
+		from the gateway for it -->
```
kaikreuzer: why is here a new line?

cweitkamp:
I am sorry. Maybe it got lost after a forced push: [TradfriDiscoveryService.java#L82](https://github.com/eclipse/smarthome/blob/master/extensions/binding/org.eclipse.smarthome.binding.tradfri/src/main/java/org/eclipse/smarthome/binding/tradfri/internal/discovery/TradfriDiscoveryService.java#L82).

kaikreuzer:
Ok, you are right with the comment. If we do discovery while the bulb is unreachable, we will actually discover the wrong (or let's say a too simple) type. But the representation-property should nonetheless avoid that we see a new discovery result. So if the Thing has the wrong type, the user can delete it and re-discover it with the new type.

cweitkamp:
Seems to work very well. I put a version similar to this PR - without the COLOR_TEMP_MODELS check - into my prod environment yesterday. The discovery shows an INFO logging for new devices with wrong thing type from time to time but nothing pops up in my Paper UI inbox. Should we maybe remove it completely? wdyt?

extensions/binding/org.eclipse.smarthome.binding.tradfri/ESH-INF/config/config.xml
cweitkamp: No problem. I removed it.

hreichert:
Side note: All IKEA products got a common identifier in their name. I used this in #4271 to distinguish between color lights and spectrum lights.
* " W " -> dimmable only
* " WS " -> white spectrum
* " CWS  " -> color white spectrum
Maybe this is a more future-safe approach to detect the type?

PR #4372, opened 2017-10-04
Magic bundle: add extensible thing to the magic test bundle

htreu:
Also: extend the color channel to be configurable.


bundles/test/org.eclipse.smarthome.magic/src/main/java/org/eclipse/smarthome/magic/binding/handler/MagicExtensibleThingHandler.java
```diff
@@ -0,0 +1,25 @@
+package org.eclipse.smarthome.magic.binding.handler;
```
kaikreuzer: License header?

htreu: done.

sjka:
Thanks

PR #3799, opened 2017-07-04
Changes to the Tradfri Binding to discover and use Groups

S0urceror:
I have updated the Tradfri Gateway binding so that it also correctly identifies groups (IPSO: 15004) of lamps bound to a remote. With the update I made they appear in the OpenHab Inbox as new Things. You can bind these to a Switch item.

Now it is easy to switch all lamps within a group on or off. Also it is possible to make a rule that executes when you select On/Off in the remote. Enabling to use the Ikea remotes for other purposes in OpenHAB as well.

PR #4373, opened 2017-10-04
[Tradfri] Added support for remote controller and motion sensor devices

cweitkamp:
The TRÅDFRI controller and sensor devices currently cannot be observed directly using the coap protocol. We assume that they are communicating directly with the bulbs or lamps without routing their commands through the gateway. This makes it nearly impossible to trigger events or redirect any kind of information to ESH/OH2. Except using workarounds like defining TRÅDFRI groups (see PR #3799).

My approach adds the possibility to include those devices using the binding to the users environment and retrieve read-only data like `presence status` or `battery level`. I am happy to hear any feedback.

- [x] Moved config classes to internal package
- [x] Added support for remote controller and motion sensor devices (read-only, battery level and battery low channels)
- [x] Implement an `abstract TradfriThingHandler` for `TradfriControllerHandler`, `TradfriLightHandler` and `TradfriSensorHandler`
- [x] Extract classes `ControllerData`, `LightData` and `SensorData` from handlers, move to own package (e.g. `o.e.s.b.tradfri.internal.model`) and implement a reusable `abstract TradfriData` class
- [x] Documentation
- [x] Unit Tests
- [x] Find a person who owns a TRÅDFRI dimmer

### DIFF
--- a/extensions/binding/org.eclipse.smarthome.binding.tradfri/ESH-INF/thing/thing-types.xml
+++ b/extensions/binding/org.eclipse.smarthome.binding.tradfri/ESH-INF/thing/thing-types.xml
@@ -70,7 +70,7 @@
 			<bridge-type-ref id="gateway" />
 		</supported-bridge-type-refs>
 
-		<label>Occupancy Sensor Unit</label>
+		<label>Occupancy Sensor</label>
 		<description>This represents the motion sensor capable of reporting the battery level.</description>
 
 		<channels>
@@ -88,7 +88,7 @@
 			<bridge-type-ref id="gateway" />
 		</supported-bridge-type-refs>
 
-		<label>Non-Color Control Unit</label>
+		<label>Non-Colour Controller</label>
 		<description>This represents the wireless dimmer sensor capable of reporting the battery level.</description>
 
 		<channels>
@@ -106,7 +106,7 @@
 			<bridge-type-ref id="gateway" />
 		</supported-bridge-type-refs>
 
-		<label>Non-Color Scene Control Unit</label>
+		<label>Non-Colour Scene Controller</label>
 		<description>This represents the remote control capable of reporting the battery level.</description>
 
 		<channels>
--- a/extensions/binding/org.eclipse.smarthome.binding.tradfri/ESH-INF/thing/thing-types.xml
+++ b/extensions/binding/org.eclipse.smarthome.binding.tradfri/ESH-INF/thing/thing-types.xml
@@ -7,24 +7,7 @@
 		<label>TRÅDFRI Gateway</label>
 		<description>IKEA TRÅDFRI IP Gateway</description>
 
-		<config-description>
-			<parameter name="host" type="text" required="true">
-				<context>network-address</context>
-				<label>Host</label>
-				<description>Hostname or IP address of the IKEA TRÅDFRI gateway</description>
-			</parameter>
-			<parameter name="port" type="integer" required="false">
-				<label>Port</label>
-				<description>Port for accessing the gateway</description>
-				<advanced>true</advanced>
-				<default>5684</default>
-			</parameter>
-			<parameter name="code" type="text" required="true">
-				<context>password</context>
-				<label>Security Code</label>
-				<description>Security code printed on the label underneath the gateway.</description>
-			</parameter>
-		</config-description>
+		<config-description-ref uri="bridge-type:tradfri:gateway" />
 	</bridge-type>
 
 	<!-- thing types for devices -->
@@ -82,16 +65,53 @@
 		<config-description-ref uri="thing-type:tradfri:device" />
 	</thing-type>
 
-	<!-- note that this isn't yet supported by the code as we do not receive any data from the gateway for it -->
+	<thing-type id="0107">
+		<supported-bridge-type-refs>
+			<bridge-type-ref id="gateway" />
+		</supported-bridge-type-refs>
+
+		<label>Occupancy Sensor Unit</label>
+		<description>This represents the motion sensor capable of reporting the battery level.</description>
+
+		<channels>
+			<channel id="battery_level" typeId="system.battery-level" />
+			<channel id="battery_low" typeId="system.low-battery" />
+		</channels>
+
+		<representation-property>id</representation-property>
+
+		<config-description-ref uri="thing-type:tradfri:device" />
+	</thing-type>
+
 	<thing-type id="0820" listed="false">
 		<supported-bridge-type-refs>
 			<bridge-type-ref id="gateway" />
 		</supported-bridge-type-refs>
 
 		<label>Non-Color Control Unit</label>
+		<description>This represents the wireless dimmer sensor capable of reporting the battery level.</description>
 
 		<channels>
-			<channel id="brightness" typeId="brightness" />
+			<channel id="battery_level" typeId="system.battery-level" />
+			<channel id="battery_low" typeId="system.low-battery" />
+		</channels>
+
+		<representation-property>id</representation-property>
+
+		<config-description-ref uri="thing-type:tradfri:device" />
+	</thing-type>
+
+	<thing-type id="0830">
+		<supported-bridge-type-refs>
+			<bridge-type-ref id="gateway" />
+		</supported-bridge-type-refs>
+
+		<label>Non-Color Scene Control Unit</label>
+		<description>This represents the remote control capable of reporting the battery level.</description>
+
+		<channels>
+			<channel id="battery_level" typeId="system.battery-level" />
+			<channel id="battery_low" typeId="system.low-battery" />
 		</channels>
 
 		<representation-property>id</representation-property>
--- a/extensions/binding/org.eclipse.smarthome.binding.tradfri/README.md
+++ b/extensions/binding/org.eclipse.smarthome.binding.tradfri/README.md
@@ -4,23 +4,35 @@ This binding integrates the IKEA TRÅDFRI gateway and devices connected to it (s
 
 ## Supported Things
 
-Besides the gateway (thing type "gateway"), the binding currently supports dimmable warm white bulbs as well as white spectrum bulbs.
+Beside the gateway (thing type "gateway"), the binding currently supports colored bulbs, dimmable warm white bulbs as well as white spectrum bulbs.
+The binding also supports read-only data from remote controls and motion sensors (e.g. the battery status).
+The TRÅDFRI controller and sensor devices currently cannot be observed right away.
+We assume that they are communicating directly with the bulbs or lamps without routing their commands through the gateway.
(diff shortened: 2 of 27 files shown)


hreichert:
@cweitkamp Sounds great! Would be very cool to integrate the battery levels in my existing monitoring!

S0urceror:
Hopefully I have time soon to add my Group support back into this release. Great work!

cweitkamp:
I pushed my latest changes and I think this PR has left the [WIP] status. I am looking forward for review comments and any kind of feedback.

I own a remote controller and a motion sensor. If anybody owns a TRÅDFRI wireless dimmer, please contact me for testing it.

sjka:
I have both, a wireless remote control and a wireless dimmer. They are both discovered successfully and added to the inbox. After having them approved, the dimmer stayed offline at first, but after a restart into debug mode it then also worked instantly. So I have no idea what initially went wrong. 
In any case, they both nicely report their battery levels. Is there anything else I should have a closer look at?

cweitkamp:
Sounds good. Question is which thing-type is used for each device? I assume `830` in both cases. Do you think that is correct? Imho the wireless dimmer should be discovered as thing-type `820`. Currently I have no clue how to distinguish between the response results of those devices - except comparing against the json property `TYPE = "5750"` -> `DEVICE_MODEL = "1"`. Wdyt?

sjka:
Yes, they were discovered as `830` both. 

I agree, according to the ZLL standard `820` would be the better match for the wireless dimmer.

Here is what they are discovered with:

```
payload: {"9001":"TRADFRI remote control","9002":1497419274,"9020":1507493476,"9003":65541,"9054":0,"15009":[{"9003":0}],"5750":0,"9019":1,"3":{"0":"IKEA of Sweden","1":"TRADFRI remote control","2":"","3":"1.2.214","6":3,"9":60}} 
payload: {"9001":"TRADFRI wireless dimmer","9002":1495312356,"9020":1507525938,"9003":65536,"9054":0,"15009":[{"9003":0}],"5750":0,"9019":0,"3":{"0":"IKEA of Sweden","1":"TRADFRI wireless dimmer","2":"","3":"1.2.248","6":3,"9":74}} 
```

I also don't see any significant difference apart from the `DEVICE_MODEL` text. So there is not much choice, is it?

cweitkamp:
Thanks for testing. I added a suitable check for the `DEVICE_MODEL`.

cweitkamp:
```
<<<<<<< feature-tradfri-remote-control
 * @author Christoph Weitkamp - Added support for remote controller and motion sensor devices (read-only battery level)
=======
 * @author Andre Fuechsel - fixed the results removal
>>>>>>> master
```

@afuechsel I definitely need some help with this merge conflict :sweat:. What should I do? :smile:

afuechsel:
@cweitkamp What an interesting conflict. :) Looks like the only one. I leave it up to you to solve the author's comment conflict :D

cweitkamp:
All,

I removed the NTP features in this PR, to get this approach merged soon. I want to make space for other necessary changes.

Reading and updating the NTP server configuration from the gateway is possible using the COAP interface, but currently leads to some not very nice issues. My approach was not good enough. I will keep on working on it in the future.

extensions/binding/org.eclipse.smarthome.binding.tradfri/README.md
```diff
@@ -4,23 +4,35 @@ This binding integrates the IKEA TRÅDFRI gateway and devices connected to it (s
...
+The thing type ids are defined according to the lighting devices defined for ZigBee LightLink ([see page 24, table 2](https://www.nxp.com/documents/user_manual/JN-UG-3091.pdf).
+These are:
+
+| Device type                     | ZigBee Device ID | Thing type |
+|---------------------------------|------------------|------------|
+| Dimmable Light                  | 0x0100           | 0100       |
+| Colour Temperature Light        | 0x0220           | 0220       |
+| Extended Colour Light           | 0x0210           | 0210       |
+| Occupancy Sensor Unit           | 0x0107           | 0107       |
+| Non-Color Control Unit          | 0x0820           | 0820       |
```
kaikreuzer: According to the spec referenced above, the name should be "Non-Colour Controller"

extensions/binding/org.eclipse.smarthome.binding.tradfri/README.md
```diff
@@ -4,23 +4,35 @@ This binding integrates the IKEA TRÅDFRI gateway and devices connected to it (s
...
+
+The thing type ids are defined according to the lighting devices defined for ZigBee LightLink ([see page 24, table 2](https://www.nxp.com/documents/user_manual/JN-UG-3091.pdf).
+These are:
+
+| Device type                     | ZigBee Device ID | Thing type |
+|---------------------------------|------------------|------------|
+| Dimmable Light                  | 0x0100           | 0100       |
+| Colour Temperature Light        | 0x0220           | 0220       |
+| Extended Colour Light           | 0x0210           | 0210       |
+| Occupancy Sensor Unit           | 0x0107           | 0107       |
```
kaikreuzer: Where did you get the 0x0107 from? I cannot find it in the spec referenced above.

extensions/binding/org.eclipse.smarthome.binding.tradfri/README.md
```diff
@@ -4,23 +4,35 @@ This binding integrates the IKEA TRÅDFRI gateway and devices connected to it (s
...
+| Non-Color Control Unit          | 0x0820           | 0820       |
+| Non-Color Scene Control Unit    | 0x0830           | 0830       |
```
kaikreuzer: -> "Non-Colour Scene Controller"

extensions/binding/org.eclipse.smarthome.binding.tradfri/src/main/java/org/eclipse/smarthome/binding/tradfri/handler/TradfriControllerHandler.java
```diff
@@ -0,0 +1,76 @@
...
+
+    @Override
+    public void handleCommand(ChannelUID channelUID, Command command) {
+        if (command instanceof RefreshType) {
+            logger.debug("Refreshing channel {}", channelUID);
+            coapClient.asyncGet(this);
+            return;
+        }
+
+        logger.warn("The controller is a read-only device and cannot handle commands.");
```
kaikreuzer: reduce to debug

extensions/binding/org.eclipse.smarthome.binding.tradfri/src/main/java/org/eclipse/smarthome/binding/tradfri/handler/TradfriSensorHandler.java
```diff
@@ -0,0 +1,76 @@
...
+
+    @Override
+    public void handleCommand(ChannelUID channelUID, Command command) {
+        if (command instanceof RefreshType) {
+            logger.debug("Refreshing channel {}", channelUID);
+            coapClient.asyncGet(this);
+            return;
+        }
+
+        logger.warn("The sensor is a read-only device and cannot handle commands.");
```
kaikreuzer: reduce to debug

extensions/binding/org.eclipse.smarthome.binding.tradfri/src/main/java/org/eclipse/smarthome/binding/tradfri/handler/TradfriThingHandler.java
```diff
@@ -0,0 +1,128 @@
...
+            }, 10, TimeUnit.SECONDS);
+        }
+    }
+
+    @Override
+    public void bridgeStatusChanged(ThingStatusInfo bridgeStatusInfo) {
+        super.bridgeStatusChanged(bridgeStatusInfo);
+
+        if (bridgeStatusInfo.getStatus() == ThingStatus.ONLINE) {
+            scheduler.schedule(() -> {
```
kaikreuzer: you can use "submit" instead

extensions/binding/org.eclipse.smarthome.binding.tradfri/src/main/java/org/eclipse/smarthome/binding/tradfri/internal/discovery/TradfriDiscoveryService.java
```diff
@@ -105,6 +112,19 @@ public void onUpdate(String instanceId, JsonObject data) {
                         thingType = THING_TYPE_DIMMABLE_LIGHT;
                     }
                     thingId = new ThingUID(thingType, bridge, Integer.toString(id));
+                    // } else if (TYPE_SWITCH.equals(type) && data.has(SWITCH)) {
```
kaikreuzer: This can be removed as you are handling it below, right?

extensions/binding/org.eclipse.smarthome.binding.tradfri/README.md
cweitkamp: see https://www.nxp.com/docs/en/user-guide/JN-UG-3114.pdf (page 71ff)

extensions/binding/org.eclipse.smarthome.binding.tradfri/README.md
cweitkamp: In relation to the above linked document we maybe can use "Dimmer Switch" (0x0104) for this device as well.

extensions/binding/org.eclipse.smarthome.binding.tradfri/README.md
kaikreuzer: Might be worth to update the link above then as well!

extensions/binding/org.eclipse.smarthome.binding.tradfri/ESH-INF/thing/thing-types.xml
```diff
@@ -82,16 +65,53 @@
 		<config-description-ref uri="thing-type:tradfri:device" />
 	</thing-type>
 
-	<!-- note that this isn't yet supported by the code as we do not receive any data from the gateway for it -->
+	<thing-type id="0107">
+		<supported-bridge-type-refs>
+			<bridge-type-ref id="gateway" />
+		</supported-bridge-type-refs>
+
+		<label>Occupancy Sensor Unit</label>
```
kaikreuzer: Could you please also update the Thing types to match the README? (remove "Unit" on the new types)

cweitkamp: Oh, yes. Done.

phelicksen:
Good job, been looking forward to this! Thanks!

cweitkamp:
Sounds reasonable, but we should move those public constants into `TradfriBindingConstants.java`  and leave the `TradfriGatewayConfig.java` where it is - in an internal package.

PR #4374, opened 2017-10-04
Leverage link configuration to select the profile

### LINKED ISSUE #2226: Trigger Channels Part 2

sjka:
With this, the ItemChannelLink configuration can be used to select
which profile should be used:

```
Switch someItem {
      channel="a:b:c:channel1" [
          profile="master"
      ]
      channel="a:b:c:channel2" [
          profile="system:slave"
      ]
      channel="a:b:c:channel3" [
          profile="mybinding:foo"
      ]
}
```

The profile identifiers are namespaced, so bindings can bring their own.
There is a "system" namespace which also is the default if no namespace is
given (like in the "master" example above).

There is a config description available for the links under

```
    link:<itemName> -> <channelUID>
```

which also includes all applicable profiles in its options.

fixes #2226

### DIFF
--- a/bundles/core/org.eclipse.smarthome.core.thing/src/main/java/org/eclipse/smarthome/core/thing/profiles/ProfileFactory.java
+++ b/bundles/core/org.eclipse.smarthome.core.thing/src/main/java/org/eclipse/smarthome/core/thing/profiles/ProfileFactory.java
@@ -7,29 +7,34 @@
  */
 package org.eclipse.smarthome.core.thing.profiles;
 
-import org.eclipse.jdt.annotation.NonNull;
+import java.util.Collection;
+
+import org.eclipse.jdt.annotation.NonNullByDefault;
 import org.eclipse.jdt.annotation.Nullable;
-import org.eclipse.smarthome.core.items.Item;
-import org.eclipse.smarthome.core.thing.Channel;
-import org.eclipse.smarthome.core.thing.link.ItemChannelLink;
 
 /**
- * Implementors are capable of creating a {@link Profile} for one or several links.
+ * Implementors are capable of creating a {@link Profile} instances.
  *
  * @author Simon Kaufmann - initial contribution and API.
  *
  */
+@NonNullByDefault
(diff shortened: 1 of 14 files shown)


bundles/core/org.eclipse.smarthome.core.thing/src/main/java/org/eclipse/smarthome/core/thing/profiles/ProfileFactory.java
```diff
@@ -7,29 +7,34 @@
...
  *
  * @author Simon Kaufmann - initial contribution and API.
  *
  */
+@NonNullByDefault
 public interface ProfileFactory {
 
     /**
-     * Create a {@link Profile} instance for the given link
+     * Create a {@link Profile} instance for the given profile tye ID.
```
kaikreuzer: typo: tye

PR #4379, opened 2017-10-05
JsonPath transformation service comply with the contract

### LINKED ISSUE #4376: JSONPATH returning NULL on item not found can lead to problems with openhab REST API
NOTE: Reported here as per request of @kaikreuzer in https://github.com/openhab/openhab2-addons/issues/2767

System: Ubuntu server 16.04 LTS on Xeon 1240L V3. Openhab stable from ppa

I recently included Sonoff B1 lightbulbs with Tasmota fw in my system. Here is an example item config:

```
Switch B1_1_Sw          "B1-1-Switch"           (gALL, gLIGHTS)         { mqtt="<[broker:b1-1/stat/RESULT:state:JSONPATH($.POWER)],  >[broker:b1-1/cmnd/POWER:command:*:${command}]" }
Number B1_1_Dim         "B1-1-Dimmer"                                   { mqtt="<[broker:b1-1/stat/RESULT:state:JSONPATH($.Dimmer)], >[broker:b1-1/cmnd/Dimmer:command:*:default]" }
Number B1_1_CT          "B1-1-ColorTemp"                                { mqtt="<[broker:b1-1/stat/RESULT:state:JSONPATH($.CT)],     >[broker:b1-1/cmnd/CT:command:*:default], >[broker:b1-1/cmnd/CT:command:*: ]" }
String B1_1_Col         "B1-1-Color"                                    { mqtt="<[broker:b1-1/stat/RESULT:state:JSONPATH($.Color)],  >[broker:b1-1/cmnd/Color:command:*:default]" }
```

The B1 with tasmota software answers to all commands with

> stat/RESULT {“POWER”:“ON”, “Dimmer”:100, “Color”:“FFFFFFFFFF”}

(of course with appropriate values). It also answers with this to the cmnd/CT command which sets the color temperature. If you want the color temperature in a reply, you have to send an empty cmnd/CT and you will get something like this.

> stat/RESULT {“CT”:500}

Now, JSONPATH returns NULL if it can’t find the path item you are asking for, and as both the standard as well as the CT reply use the RESULT path, there’s always something going to be NULL. If you send a regular command, CT will be NULL and if you ask for CT, POWER, Dimmer and Color will be NULL.

That lead to this messages in the log:

> 2017-10-03 20:12:50.277 [WARN ] [b.core.events.EventPublisherDelegate] - given new state is NULL, couldn’t post update for 'B1_1_CT’
> 2017-10-03 20:12:50.282 [WARN ] [b.core.events.EventPublisherDelegate] - given new state is NULL, couldn’t post update for 'B1_1_CT’
> 2017-10-03 20:12:50.325 [WARN ] [b.core.events.EventPublisherDelegate] - given new state is NULL, couldn’t post update for 'B1_1_Sw’
> 2017-10-03 20:12:50.325 [WARN ] [b.core.events.EventPublisherDelegate] - given new state is NULL, couldn’t post update for ‘B1_1_Dim’

And after a short while the openhab REST API would reply to an items request with a 500 java.lang.NullPointerException.

See also https://community.openhab.org/t/connection-lost/34744/15

sjka:
...as it was returning null in case the path pointed to an element which did not exist.
Now it throws a TransformationException. 

I'm not 100% sure though if this _really_ is _always_ the right way of fixing it. A missing element in JSON is somewhat similar to "intentionally null", i.e. `UndefType.UNDEF` might also be a consideration. Any deeper insights anybody?

fixes #4376

### DIFF
--- a/extensions/transform/org.eclipse.smarthome.transform.jsonpath.test/META-INF/MANIFEST.MF
+++ b/extensions/transform/org.eclipse.smarthome.transform.jsonpath.test/META-INF/MANIFEST.MF
@@ -8,4 +8,6 @@ Bundle-Version: 0.9.0.qualifier
 Fragment-Host: org.eclipse.smarthome.transform.jsonpath
 Import-Package: 
  org.eclipse.jdt.annotation;resolution:=optional,
+ org.hamcrest,
  org.junit;version="4.0.0"
+Require-Bundle: org.hamcrest
--- a/extensions/transform/org.eclipse.smarthome.transform.jsonpath.test/src/test/java/org/eclipse/smarthome/transform/jsonpath/internal/JSonPathTransformationServiceTest.java
+++ b/extensions/transform/org.eclipse.smarthome.transform.jsonpath.test/src/test/java/org/eclipse/smarthome/transform/jsonpath/internal/JSonPathTransformationServiceTest.java
@@ -7,8 +7,9 @@
  */
 package org.eclipse.smarthome.transform.jsonpath.internal;
 
+import static org.junit.Assert.assertEquals;
+
 import org.eclipse.smarthome.core.transform.TransformationException;
-import org.eclipse.smarthome.transform.jsonpath.internal.JSonPathTransformationService;
 import org.junit.Assert;
 import org.junit.Before;
 import org.junit.Test;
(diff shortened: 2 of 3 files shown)


kaikreuzer:
> Any deeper insights anybody?

Not from my end.
Shall we wait for insights from others or "just merge"...?

sjka:
Then suggest using merging it because it now fits to the javadoc contract. If it turns out this is not appropriate, then we still can adapt it afterwards.

martinvw:
Hmm, reading your suggestion `UndefType.UNDEF` or `UndefType.NULL` might make more sense, it is after all an element which may be only missing this time. And it's not uncommon in JSON the leave out any `null` values, see for example some NPE's I got in the Nest binding because I do not posses all Nest devices.

htreu:
@martinvw imho the transformation service is right about throwing a TransformationException. Its in the responsibility of the caller to act appropriate. And from the code I saw the exception is handled well by all callers.

Issue #2826, opened 2017-01-14
Event-based Triggers not working with constants


ThomDietrich:
Rule conditions involving constant variables or "strings" are never considered. Is this supposed to be supported?

Minimal Example:

```java
val Number MODE_FINISHED = 3 // constant


rule "State Machine Rule 1"
when
    Item OpState changed
then
    if (OpState.state == MODE_FINISHED) {
        // This is working
    }
end


rule "State Machine Rule 2"
when
    Item OpState changed to 3
then
    // This is working
end


rule "State Machine Rule 3"
when
    Item OpState changed to MODE_FINISHED
then
    // This is NOT working
end


rule "State Machine Rule 4"
when
    Item OpState_String changed to "FINISHED"
then
    // This is NOT working
end
```

kaikreuzer:
No sorry, this is not supported.

What you need to know is that only the part between "then" and "end" is the Xbase script that has access to the variables and constant declarations. I do not see any good way to change this, it would rather be a huge and ugly hack.

ThomDietrich:
Okay already expected this to be such a kind of a problem. Thanks!

MrMontesa:
Just for confirmation as to what I understand from this issue:
In rules it is not possible to react on certain string item changes. E.g:
`2017-04-16 21:48:03.074 [ItemStateChangedEvent     ] - harmonyhub_activity changed from PowerOff to Fernsehen
`

As this issue is referenced quite often in community forum it would be great to
a) use this issue as a feature request since I think there is neccissity to support this?
b) explain a possible workaround solution as long as this feature is not supported?

Thanks much for your help

kaikreuzer:
Actually, my comment was about
```
Item OpState changed to MODE_FINISHED
```
Directly providing a string should actually work - so this should be analyzed, why this isn't triggered with
```
Item OpState_String changed to "FINISHED"
```
So let's re-open the issue.

ThomDietrich:
@kaikreuzer I was okay to leave it with this but solving the forth case (comparison with a string) would be appreciated.

@MrMontesa the workaround is actually the very first example I've given.

MrMontesa:
Hey @ThomDietrich ,
You're right for the basic rule trigger. But once you need a more sophisticated trigger like:
`Item XYZ changed from "string1" to "string2"`
You're example will be quite complex depending on the changes you need to cover.

use-case:
[...]

PR #4383, opened 2017-10-06
Support for quoted strings in rule triggers

sjka:
According to #2826, the following works

```xtext
rule "Test Rule 1"
when
    Item test changed to world
then
    // This is IS working
end
```

while the following is NOT working:

```xtext
rule "Test Rule 2"
when
    Item test changed to "world"
then
    // This is NOT working
end
```

With this PR, I so far have only added a test-case for this. And indeed it fails and results as expected in 

```
testChangedEventTrigger_withoutQuotes(org.eclipse.smarthome.model.rule.runtime.internal.engine.RuleTest)  Time elapsed: 0.639 sec
testChangedEventTrigger_withQuotes(org.eclipse.smarthome.model.rule.runtime.internal.engine.RuleTest)  Time elapsed: 0.034 sec  <<< FAILURE!
org.junit.ComparisonFailure: expected:<[world]> but was:<["world"]>
	at org.junit.Assert.assertEquals(Assert.java:115)
	at org.junit.Assert.assertEquals(Assert.java:144)
	at org.eclipse.smarthome.model.rule.runtime.internal.engine.RuleTest.assertUpdateTriggerWithNewState(RuleTest.java:82)
	at org.eclipse.smarthome.model.rule.runtime.internal.engine.RuleTest.testChangedEventTrigger_withQuotes(RuleTest.java:66)
```

So the quotes somehow end up inside the string! 

It looks like this behavior gets better when defining the trigger in [Rules.xtext](https://github.com/eclipse/smarthome/blob/09d8095efe9933216874e488b379cd9a10aa85e2/bundles/model/org.eclipse.smarthome.model.rule/src/org/eclipse/smarthome/model/rule/Rules.xtext) instead as

```xtext
ChangedEventTrigger:
	'Item' item=ItemName 'changed' ('from' oldState=ValidState)? ('to' newState=ValidState)?
;		
ValidState:
	ID | Number | STRING
;
```

directly like this:

```xtext
ChangedEventTrigger:
	'Item' item=ItemName 'changed' ('from' oldState=(ID|Number|STRING)? ('to' newState=(ID|Number|STRING))?
;		
```

However, we then would lose the nice way of integrating the content proposal in [RulesProposalProvider.java](https://github.com/eclipse/smarthome/blob/3ae5d7b1b284f2bde6199bfe8dad950fd38cae39/bundles/model/org.eclipse.smarthome.model.rule.ui/src/org/eclipse/smarthome/model/rule/ui/contentassist/RulesProposalProvider.xtend). So there must be a better way to "trick" xtext into behaving well. Any suggestions? @chriskn or @cdietrich maybe?

relates to #2826

### DIFF
--- a/bundles/model/org.eclipse.smarthome.model.rule.runtime/src/org/eclipse/smarthome/model/rule/runtime/internal/engine/RuleEngineImpl.java
+++ b/bundles/model/org.eclipse.smarthome.model.rule.runtime/src/org/eclipse/smarthome/model/rule/runtime/internal/engine/RuleEngineImpl.java
@@ -96,7 +96,7 @@ public class RuleEngineImpl implements ItemRegistryChangeListener, StateChangeLi
     @Activate
     public void activate() {
         injector = RulesStandaloneSetup.getInjector();
-        triggerManager = injector.getInstance(RuleTriggerManager.class);
+        triggerManager = new RuleTriggerManager(injector);
 
         if (!isEnabled()) {
             logger.info("Rule engine is disabled.");
@@ -440,5 +440,9 @@ public class RuleEngineImpl implements ItemRegistryChangeListener, StateChangeLi
         } else if (event instanceof ThingStatusInfoChangedEvent) {
             receiveThingStatus((ThingStatusInfoChangedEvent) event);
         }
+    }
+
+    RuleTriggerManager getTriggerManager() {
+        return triggerManager;
     }
 }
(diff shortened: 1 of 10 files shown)


sjka:
I revisited this PR and I _think_ I found a way to assist Xtext in retaining the original type information and thereby getting rid of the excess quotation marks.

It looks just a _little_ bit bit uglier in the xtext file, but on the other hand that is completely kept internal and the result is fully backward compatible with respect to existing files written in the Rules DSL.

PR #4386, opened 2017-10-08
Basic/Classic UI: support for HTTP Live Stream

lolodomo:
Add support for HLS in compatible WEB browsers
Use encoding="HLS" in the video sitemap element.

### DIFF
--- a/extensions/ui/org.eclipse.smarthome.ui.basic/snippets/video.html
+++ b/extensions/ui/org.eclipse.smarthome.ui.basic/snippets/video.html
@@ -4,6 +4,6 @@
 		data-control-type="video"
 		data-widget-id="%widget_id%"
 	>
-	   <video autoplay controls src="%url%"></video>
+	   <video autoplay controls src="%url%" %media_type%></video>
 	</div>
 </div>
--- a/extensions/ui/org.eclipse.smarthome.ui.basic/src/main/java/org/eclipse/smarthome/ui/basic/internal/render/VideoRenderer.java
+++ b/extensions/ui/org.eclipse.smarthome.ui.basic/src/main/java/org/eclipse/smarthome/ui/basic/internal/render/VideoRenderer.java
@@ -9,6 +9,8 @@ package org.eclipse.smarthome.ui.basic.internal.render;
 
 import org.apache.commons.lang.StringUtils;
 import org.eclipse.emf.common.util.EList;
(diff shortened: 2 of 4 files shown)


lolodomo:
I tested it with Chrome on Android.
I can't test with Apple devices but it should work in Safari (Mac or iOs).

lolodomo:
As an example, this change allows the rendering of video from the Netatmo Welcome camera.

You can find compatible browsers with HLS in wikipedia: https://en.wikipedia.org/wiki/HTTP_Live_Streaming

lolodomo:
I tested with the 5 m3u8 links proposed on this page: https://bitmovin.com/mpeg-dash-hls-examples-sample-streams/

It is working well in Chrome (Android) and in Microsoft Edge (Windows 10).

Unfortunately Netatmo welcome links are not working in Microsoft Edge.

lolodomo:
Even if I am very confident, it would be cool if someone can test with Safari browser on iOS and Mac.

sjka:
At lest the middle one works on Safari @ macOS and Safaris @ iOS. I'm not sure whether it's just that the others are simply not available/overloaded right now or it that points to some other issues.

None of them works in Chrome @ macOS/Windows 10, which is expected though if I read [this Wikipedia page](https://en.wikipedia.org/wiki/HTTP_Live_Streaming) correctly.

This is according to your expectations, right?

lolodomo:
Are your tests really done with the updated UIs from this PR ?
The 5 links were available yesterday and I am expecting that they all works in Safari with my changes.
As a reminder, they all work in Chrome/Android anec Edge/Windows 10 with this PR.

sjka:
Ouch, my bad - got the IPs mixed up, sorry. 
They are indeed _all_ good on Safari on both OSes!

sjka:
Just one question regarding the encoding configuration - unfortunately we have no documentation about sitemaps here, so the best I could find was the one from [openHAB](http://docs.openhab.org/configuration/sitemaps.html#element-type-video). There it says that the encoding attribute "can stay left empty for auto selection". This does not hold true not anymore, right? Is it somehow possible to auto-detect HLS streams?

lolodomo:
I will update this documentation once this PR is merged to explain that "hls" has to be set for m3u8 files (HLS).

lolodomo:
Thank you for your additional tests @sjka . I think you tested wit success on all compatible WEB browsers. Note that it should work in Chrome/iOS too.

PR #4387, opened 2017-10-09
PaperUI: refactor config description editor to directive

htreu:
Also: reduce export scope of configService, onwards & upwards


sjka:
Looks nicer in the html and configuration in PaperUI still works. 
Thanks!

PR #4388, opened 2017-10-09
[Tradfri] Increase default command delay

ivivanov-bg:
After some more tests with different devices
it turned out that 500 ms is not enough for some basic operations.
So - increasing the delay to 600 ms

### DIFF
--- a/extensions/binding/org.eclipse.smarthome.binding.tradfri/src/main/java/org/eclipse/smarthome/binding/tradfri/internal/TradfriCoapClient.java
+++ b/extensions/binding/org.eclipse.smarthome.binding.tradfri/src/main/java/org/eclipse/smarthome/binding/tradfri/internal/TradfriCoapClient.java
@@ -28,7 +28,7 @@ import org.slf4j.LoggerFactory;
 public class TradfriCoapClient extends CoapClient {
 
     private static final int TIMEOUT = 2000;
-    private static final int DEFAULT_DELAY_MILLIS = 500;
+    private static final int DEFAULT_DELAY_MILLIS = 600;
     private final Logger logger;
     private final LinkedList<PayloadCallbackPair> commandsQueue;
     private final Runnable commandExecutor;


kaikreuzer:
> So - increasing the delay to 600 ms

Why does the code contain 1000 then?

ivivanov-bg:
Should be OK now

kaikreuzer:
Are you really sure that this is the right way to address the issues? It feels as if there is some reliability issue in the gateway and we do not really know why. From my testing, the gateway should "in general" be able to process many requests per second. So in what situations does it fail? Probably due to some concurrent processing. So yes, sending less requests will reduce the problem. Once we do not send any commands anymore, the problem will even be gone ;-).
But who tells us that the same issue does not happen, if the gateway is busy because some other app/system is using it at the same time? This would not be under our control.

So in an ideal world, IKEA would fix the concurrency issues on the gateway itself. Until then, wouldn't it be cleaner to implement some retry mechanism, i.e. make sure that we treat the communication bi-directional and thus check that after sending a command, we receive an acknowledgement/update?

ivivanov-bg:
From my tests - it seems that the problem is no the gateway but the bulb itself. And in the part of changing color temperature.

While the bulb is changing it's color - it ignores all other requests. I don't know if it has to do with the DEFAULT_DIMMER_TRANSITION_TIME which is sent with each command or it's a hardware limitation for the bulb.

The problem with the retry mechanism is that there is no way (or at least I was not able to find how) to know if the command was applied (or is still been processed).

I tried using synchronous mode for the communication but it is only between the gateway and the binding and the gateway acknowledges all commands immediately - so - no real difference.

hreichert:
I can confirm that the fading is the problem.
Tried it directly with the "deconz" zigbee controller:
* Send color temperature to bulb
* Immediately send brightness change
* Results in: Interrupted color temperature change

ivivanov-bg:
OK - then I think we can also implement new channel for `SCENE` command - but in anyway - this timeout is still required because without it - the bulbs are not handling all commands and the issuer of the commands is not notified about that.

hreichert:
> OK - then I think we can also implement new channel for SCENE command
Maybe sometime in the future ;) And this would not help for this issue. I just wanted to point out why it "seems" to work in the IKEA app. At first I wondered how they send color temperature and brightness at the same time, but well, they don't - just predefined scenes.

> this timeout is still required because without it - the bulbs are not handling all commands and the issuer of the commands is not notified about that.
Yes. Maybe IKEA solves this sometime and somehow in the gateway and/or in the bulbs, but for now, this seems like a good workaround.

sjka:
While I absolutely agree with @kaikreuzer that this workaround is not really the robust way of solving this, but I also see that the workaround already _is_ in the code-base. This PR hardly makes it any worse (precisely only by 20% 😉 ) but instead gives it a chance to be effective _at all_ for those situation _where_ it could be. 
Therefore I created #4410 to follow-up on this topic and would like to encourage you all to do so!

kaikreuzer:
> This PR hardly makes it any worse (precisely only by 20% 😉 )

I agree and I am fine with that. I just hope that it will reduce the problem by more than 20% because otherwise, we might see 5 more PRs adding 100ms each ;-)

Issue #1093, opened 2016-02-25
Semantic tagging


kaikreuzer:
ESH has a basic tagging implementation, which allows to add tags (as simple strings) to items.
The idea behind this is to assign items a semantic. So while the "category" refers to a taxonomy (e.g. this is a "temperature sensor", the tags are supposed to refer to an ontology (this is the "current outside temperature").

So far, a few temporary tags were introduced, like "home-group" for marking group items as a kind of room, as well as "thing" to mark it as a group of items of a physical device.
Furthermore, the new openHAB integrations for HomeKit and Amazon Echo make use of [tags like "homekit:DimmableLightbulb"](https://github.com/openhab/openhab2-addons/tree/master/addons/io/org.openhab.io.homekit#item-configuration).

With more use cases coming up, I think we need to formalise the use of tags and thus we should come up with a (or many?) ontologies to use (and hence have a fixed tag library).
Since there are many efforts for ontologies in the industry, I am not fond of the idea to define one from scratch for ESH. Nonetheless, it also does not seem as if there is anything matching our needs to 100%.
Since ontologies can reference other ontologies, my suggested approach is trying to define an ESH ontology which refers as much as possible to the work of others.

A good starting point is the [SAREF ontology](http://ontology.tno.nl/saref/). Its "saref:Device" and "saref:Service" can be replacements for "thing", while "saref:BuildingSpace" could replace "home-group".
SAREF itself already refers to other ontologies where it makes sense, such as
https://www.w3.org/TR/owl-guide/
https://www.w3.org/TR/owl-time/
https://www.w3.org/2003/01/geo/
http://www.wurvoc.org/vocabularies/om-1.6/Unit_of_measure

Note that hierarchies and references are an important part of ontologies. So we imho need to enhance the querying possibilities on the ItemRegistry to respect this. It should e.g. be possible to query for ["saref:BuildingObject"](http://ontology.tno.nl/saref/saref_BuildingObject.html) and get all items that have tags like "saref:Window", "saref:Door", "saref:Device", etc.

One of my initial ideas was also to support multiple different ontologies on ESH and provide a possibility to automatically map between them, where feasible. I fear that this might get pretty complicated though and hence I'd suggest to go for just "the one ESH ontology" for now.

beowulfe:
I'd like to think about extending the existing ontology with support for WindowCoverings, GarageDoors, Locks and Fans. (Using HomeKits terms here) Any thoughts on including these in the ontology?

[...]

PR #4390, opened 2017-10-09
[WIP] added meta data infrastructure

### LINKED ISSUE #3692: Allow add-ons to provide custom meta-data on items

kaikreuzer:
Fixes #3692.

This PR introduces a `MetaDataRegistry` which can be queried for item-related meta data. The meta data can be add-on specific and thus it is always referred to with a namespace.

There is a `ManagedMetaDataProvider`, which allows adding meta data through the REST API, but add-ons are also free to implement their own `MetaDataProvider`, if they have other sources for that information.

Wrt textual configuration, it uses the existing infrastructure for "binding configurations", i.e. the `{ ..}` part of item files. Any such definition for which there is NO reader available will end up in the  meta data registry through the `GenericMetaDataProvider`.

As an example, this allows to do definitions like:

```
Switch "My Fan" { homekit="Fan.v2", alexa="Fan"[type="oscillating", speedSteps=3] }
```
Note that also the new configurations can be used to provide key-value maps if necessary.

The item endpoint of the REST API has been extended to directly include meta data for given namespaces in an item response, so that it should be easy for external clients to get hold of the data.

Still to do:
- [ ] Handling of exceptional cases (e.g. delete requests for read-only meta data)
- [ ] Code polishing
- [ ] Unit tests
- [ ] Documentation


kaikreuzer:
@beowulfe, I would like to validate that approach with you.
My hope is that this new feature would allow to refactor the homekit add-on in a way that it does not rely on item tags (only), but on meta-data instead. A thermostat could e.g. look like:

```
Group Thermostat { homekit="Thermostat" [units="Celsius", name="Kitchen"]}
Number Temperature (Thermostat) { homekit="temperature.current" }
Number Temperature_Setpoint (Thermostat) { homekit="temperature.target" }
```

Alternatively, it could even try to derive some information when no meta-data on the group is specified:
```
Group Kitchen
Number Temperature (Kitchen) { homekit="thermostat:temperature.current" }
Number Temperature_Setpoint (Kitchen) { homekit="thermostat:temperature.target" }
```
-> auto derive a Thermostat service with the following capabilities:
```
thermostat:name=Kitchen (taken from the group's name)
thermostat:temperature.units=Celsius (once we have unit support on Number items)
thermostat:heating-cooling.current = Heat  (e.g. when target is higher than current)
thermostat:heating-cooling.target = auto
```
This would give you the full flexibility to do mappings to the HomeKit definitions. We should imho try to keep support for a common semantic tag library in the add-on, so whenever there is a good mapping from ESH tags to Homekit, that could be automatically derived without requiring from the user to specifically add homekit information.

What do you think about this approach? Would that work for you?

SpeedmaxX:
Hi, did you do any steps forward in this issue in the background?

marziman:
@kaikreuzer,

really cool stuff. Took a look and it would be great to have this officially available. I think the docu part is important next to the unit tests. Great job!

Any idea, when this can go into core? Do you need support with testing this?

BR Mehmet

bundles/core/org.eclipse.smarthome.core/src/main/java/org/eclipse/smarthome/core/items/MetaDataKey.java
```diff
@@ -0,0 +1,90 @@
...
+
+/**
+ * This class represents the key of a {@link MetaData} entity.
+ * It is a simple combination of a namespace and an item name.
+ *
+ * @author Kai Kreuzer - Initial contribution and API
+ *
+ */
+@NonNullByDefault
+public class MetaDataKey {
```
maggu2810: Hm, this would be a good candidate for the cleaned up (Generic)UID base class usage. :wink:

sjka: Nice - didn't have it on my radar either...

sjka:
As agreed with @kaikreuzer I continued this PR in order to push forward this topic.

Thereby I 
* rebased it onto the latest master (therefore - as aligned - the force-push instead of a PR against this PR)
* renamed MetaData to Metadata as it's a single word (apart from that I left the initial commit untouched)
* introduced an AbstractUID class (as already attempted in #4891 - now we have the use-case for it as @maggu2810 pointed out in https://github.com/eclipse/smarthome/pull/4390#discussion_r166456818)
* let the `MetadataRegistry` listen only on the `ManagedItemProvider` instead of the `ItemRegistry` in order to perform the clean up in its managed provider. Otherwise it would happen also when a provider disappears.
* added tests for the whole enchilada
* added some conceptual documentation
* some minor fixes here and there

Yet missing is the adaption of the Paper UI.

bundles/core/org.eclipse.smarthome.core/src/main/java/org/eclipse/smarthome/core/common/AbstractUID.java
```diff
@@ -0,0 +1,129 @@
...
+        int numberOfSegments = segments.size();
+        if (numberOfSegments < minNumberOfSegments) {
+            throw new IllegalArgumentException(
+                    String.format("UID must have at least %d segments.", minNumberOfSegments));
+        }
+        for (int i = 0; i < numberOfSegments; i++) {
+            String segment = segments.get(i);
+            validateSegment(segment, i, numberOfSegments);
+        }
+        this.segments = segments;
```
maggu2810: It seems I missed one point some time...
The list is supplied by the caller and we should ensure that if he changes the list after the UID generation the internals of the UID object are not changed anymore.
This call should be changed to something similar to: `this.segments = Collections.unmodifiableList(new ArrayList<>(segments));`

sjka: ...as did I! Fixed.

digitaldan:
This looks great,  I have the Alexa Version 3 skill about ready for beta.  The v3 api is much more advanced and will require more comprehensive meta data then was previously used.  While I am still supporting the original tags (Lighting, Switch, Thermostat), I think it would be great to have the additional functionally only available with the new metadata structure.  Right now I am thinking example Alexa items would look like:
```
  Dimmer LightSwicth   "Light"                           { alexa="PowerController.powerState,BrightnessController.brightness" }

  Group  Thermostat    "Thermostat"                             { alexa="Endpoint.Thermostat" [scale="Fahrenheit"] }	  
  Number Temperature   "Temperature [%.0f F]"    (Thermostat)   { alexa="TemperatureSensor.temperature"}
  Number HeatSetpoint  "Heat Setpoint [%.0f F]"  (Thermostat)   { alexa="ThermostatController.upperSetpoint }
  Number CoolSetpoint  "Cool Setpoint [%.0f F]"  (Thermostat)   { alexa="ThermostatController.lowerSetpoint" }
  Number Mode          "Mode [%s]"               (Thermostat)   { alexa="ThermostatController.thermostatMode" [OFF=0,HEAT=1,COOL=2,AUTO=3] }
```

Couple things, at first glance I would assume the parameters in brackets are not specific for any one meta definition, so if we have {alexa="..." homekit="..."  [units="Fahrenheit"] } , then the units parameter is part of the overall metadata, and not a child of the alexa or homekit metadata ?   

If this is the case, should we have set of standard, or at least agreed upon  parameters?  For example, I am using "scale" b/c the alexa api uses this term, but OH prefers "units", so i would rather we agree to use "units" as a standard so the user is not duplicating this per item (like [units="Fahrenheit", scale="Fahrenheit"]) . 

The same goes for the "name" parameter, i like the idea of being able to have a different spoken name then the item label if its provided, but I think we should agree that "name" (or spokenName) should be used as best practice among Alex, homekit, google, etc..  Again, I am assuming parameters are on the item and not specific to each metadata definition.

beowulfe:
Agreed on the need to standardize on some of the keys. I wonder how far we want to go with that though, and what we should do with the outliers. For HomeKit, for example, we'd want to define minimum and maximum temperatures. That's necessary for the HomeKit UI, but I'm guessing isn't something Alexa cares about.

Do we want to prefix keys like "HomeKit:maximumTemperature" to avoid collisions?

beowulfe:
Two alternatives to prefixing:

If we want to promote reuse of metadata, we may need a Registry somewhere to reserve and describe fields.

If we simply want to avoid collisions, we could reserve keys within the module and write a test that fails on duplicates.

I think I prefer the idea of a global registry for fields, because we don't want to end up with a bunch of variants on the same data.

digitaldan:
@sjka I'm a little embarrassed i can't find where the bindings parameters ( "[foo=bar]" )  are parsed in ESH,  I'm not familiar with this syntax and how it applies to the binding config, can you point me to this?

kaikreuzer:
@digitaldan & @beowulfe: I would think that this discussion should better be done on https://github.com/eclipse/smarthome/issues/3692 or even https://github.com/eclipse/smarthome/issues/1093, but not here on this PR, which "merely" introduces a technical feature for handling meta-data (but in itself not caring about the meaning of that data).

> If we want to promote reuse of metadata

No, we definitely do not want to do this. https://github.com/eclipse/smarthome/issues/3692 is all about "add-on specific metadata", while https://github.com/eclipse/smarthome/issues/1093 is about information that can be shared (and where we noticed that it isn't sufficient for many cases).

The "reuse" @digitaldan is talking about is merely to agree on names for parameters, but not actually reusing their values across add-ons.

beowulfe:
Oh, I see - I missed that the extra attributes apply to the first key outside the [.

Makes sense - I like it!

bundles/model/org.eclipse.smarthome.model.item/src/org/eclipse/smarthome/model/item/internal/GenericMetadataProvider.java
```diff
@@ -0,0 +1,95 @@
...
+ *
+ * @author Kai Kreuzer - Initial contribution and API
+ *
+ */
+@NonNullByDefault
+@Component(immediate = true, service = { MetadataProvider.class, GenericMetadataProvider.class })
+public class GenericMetadataProvider extends AbstractProvider<Metadata> implements MetadataProvider {
+
+    private final Set<Metadata> metadata = new HashSet<>();
+    private final ReadWriteLock lock = new ReentrantReadWriteLock(true);
```
maggu2810: As the lock is private and so used by this class internally only, I still think (as already comment somewhere else) a simple object and a synchronized block would be simpler and less error prone ("oh, I missed an unlock).
I see no usage of some advanced lock usage here.

bundles/model/org.eclipse.smarthome.model.item/src/org/eclipse/smarthome/model/item/internal/GenericMetadataProvider.java
```diff
@@ -0,0 +1,95 @@
...
+ * This class serves as a provider for all metadata that is found within item files.
+ * It is filled with content by the {@link GenericItemProvider}, which cannot itself implement the
+ * {@link MetadataProvider} interface as it already implements {@link ItemProvider}, which would lead to duplicate
+ * methods.
+ *
+ * @author Kai Kreuzer - Initial contribution and API
+ *
+ */
+@NonNullByDefault
+@Component(immediate = true, service = { MetadataProvider.class, GenericMetadataProvider.class })
```
maggu2810: IMHO a provider component needs be activated only if someone would like to use it.

bundles/core/org.eclipse.smarthome.core/src/main/java/org/eclipse/smarthome/core/internal/items/MetadataRegistryImpl.java
```diff
@@ -0,0 +1,108 @@
...
+
+/**
+ * This is the main implementing class of the {@link MetadataRegistry} interface. It
+ * keeps track of all declared metadata of all metadata providers.
+ *
+ * @author Kai Kreuzer - Initial contribution and API
+ *
+ */
+@Component(immediate = true, service = { MetadataRegistry.class })
+public class MetadataRegistryImpl extends AbstractRegistry<Metadata, MetadataKey, MetadataProvider>
```
maggu2810: Would it be possible to add `@NonNullByDefault`?

bundles/core/org.eclipse.smarthome.core/src/main/java/org/eclipse/smarthome/core/items/ManagedMetadataProvider.java
```diff
@@ -0,0 +1,72 @@
...
+
+/**
+ * {@link ManagedMetadataProvider} is an OSGi service, that allows to add or remove
+ * metadata for items at runtime. Persistence of added metadata is handled by
+ * a {@link StorageService}.
+ *
+ * @author Kai Kreuzer - Initial contribution
+ */
+@Component(immediate = true, service = { MetadataProvider.class, ManagedMetadataProvider.class })
+public class ManagedMetadataProvider extends AbstractManagedProvider<Metadata, MetadataKey, Metadata>
```
maggu2810: Would it be possible to add `@NonNullByDefault`?

bundles/core/org.eclipse.smarthome.core/src/main/java/org/eclipse/smarthome/core/items/Metadata.java
```diff
@@ -0,0 +1,87 @@
...
+        return key;
+    }
+
+    /**
+     * Provides the configuration meta-data.
+     *
+     * @return configuration as a map of key-value pairs
+     */
+    public Map<String, Object> getConfiguration() {
+        return Collections.unmodifiableMap(configuration);
```
maggu2810: Isn't the configuration of the `Metadata` itself already fixed / unmodifiable?
In the constructor `Collections.emptyMap()` is used if no configuration is provided, so in that case the member is already unmodifiable. We could change the constructor part if a provided configuration is assigned from `new HashMap<>(configuration)` to `Collections.unmodifiableMap(new HashMap<>(configuration))` and return `configuration` here.
As we return a view this already requires that the collection itself is not altered, because this could break (CME) the usage of the returned view.

bundles/core/org.eclipse.smarthome.core/src/main/java/org/eclipse/smarthome/core/items/MetadataPredicates.java
```diff
@@ -0,0 +1,39 @@
...
+
+import java.util.function.Predicate;
+
+/**
+ * Provides some default predicates that are helpful when working with metadata.
+ *
+ * @author Kai Kreuzer - Initial contribution and API
+ *
+ */
+public class MetadataPredicates {
```
maggu2810: Would it be possible to add `@NonNullByDefault`?

bundles/core/org.eclipse.smarthome.core/src/main/java/org/eclipse/smarthome/core/items/MetadataProvider.java
```diff
@@ -0,0 +1,21 @@
...
+import org.eclipse.smarthome.core.common.registry.Provider;
+
+/**
+ * This is a marker interface for metadata provider implementations that should be used to register those as an OSGi
+ * service.
+ *
+ * @author Kai Kreuzer - Initial contribution and API
+ *
+ */
+public interface MetadataProvider extends Provider<Metadata> {
```
maggu2810: Would it be possible to add `@NonNullByDefault`?

bundles/core/org.eclipse.smarthome.core/src/main/java/org/eclipse/smarthome/core/items/MetadataRegistry.java
```diff
@@ -0,0 +1,22 @@
...
+/**
+ * The MetadataRegistry is the central place, where additional information about items is kept.
+ *
+ * Metadata can be supplied by {@link MetadataProvider}s, which can provision them from any source
+ * they like and also dynamically remove or add data.
+ *
+ * @author Kai Kreuzer - Initial contribution and API
+ *
+ */
+public interface MetadataRegistry extends Registry<Metadata, MetadataKey> {
```
maggu2810: Would it be possible to add `@NonNullByDefault`?

bundles/model/org.eclipse.smarthome.model.item/src/org/eclipse/smarthome/model/item/internal/GenericMetadataProvider.java
sjka: Please notice: this is a ReadWriteLock! Is your suggested approach using a simple synchronized block still be as elegant for the "block read attempts only if somebody else writes to it at the same time"-use-case? Maybe I'm just not seeing it...

maggu2810: >> I see no usage of some advanced lock usage here.

> block read attempts only if somebody else writes to it at the same time

:see_no_evil: 

Time to open the eyes. Thanks

sjka:
@digitaldan your assumption is correct, the properties inside the `[...]` only apply to the meta-data which they proceed. They are parsed (as all the rest) by Xtext and end up in a [ModelItem](https://github.com/eclipse/smarthome/blob/e3831401564a2f35ead7e65f4e511f5f69f9b2ad/bundles/model/org.eclipse.smarthome.model.item/src/org/eclipse/smarthome/model/item/internal/GenericItemProvider.java#L204) instance. That all existed before, therefore it's practically impossible to find it here in this PR. This PR then newly passes these properties to the GenericMetaDataProvider [here](https://github.com/eclipse/smarthome/pull/4390/files#diff-d55eb2dcdd9211f0f1e797cd3f94d0f6R393).

Regarding the agreement of property keys etc.: I would love to reuse the config description concept which we already have for all kinds of other entities, mainly for helping UIs to render something which makes sense to the users. Unfortunately there are some architectural challenges because that would introduce cyclic dependencies. I will think about that a little, maybe we can find a way to still make that possible - and then this would also give you some means to enable re-use of such config descriptions between different namespaces.

digitaldan:
>They are parsed (as all the rest) by Xtext and end up in a ModelItem instance.
I figured this, but still could not find in ESH where this was being modeled, I assumed that the parameter format must have already existed since this PR does not change anything there, do you know where the Xtext format for this is, I'm merely curious :-)

>Regarding the agreement of property keys etc.
Since the parameters are tied to the metadata they proceed, I think this is less important, so I would not think too much about it. I'm not sure there really will be that many, so I wonder if between @beowulfe @marziman and myself that we could agree on some common names for the few that do overlap (units and spokenName ), since we represent the first big users of this (homekit, alexa, google voice).

sjka:
Sure, the Xtext part indeed was added already a while ago in a different context with [#4323](https://github.com/eclipse/smarthome/pull/4323/files#diff-d424787b9ea712f75e8d9488400839fa).

Alignment indeed makes sense, I'm sure you guys will manage that amongst yourselves. That's pragmatic.

digitaldan:
>Sure, the Xtext part indeed was added already a while ago in a different context with #4323.

Thanks, that's was what I was looking for.

docs/documentation/concepts/items.md
```diff
@@ -137,3 +137,24 @@ Here is a short table demonstrating conversions for the examples above:
...
+
+## Item Metadata
+
+Sometimes additional information is required to be attached to items for certain use-cases. 
+This could be e.g. an application which needs some hints in order to render the items in a generic way or an integration with voice controlled assistants or any other services which access the items and need to understand their "meaning".
+
+For this purpose, such meta-information can be attached to items using disjunct namespaces so they won't conflict with each other. 
+Each metadata entry has a main value and optionally additional key/value pairs. 
+There can be metadata for as many namespaces attached to an item as desired. 
+The following example show how `something` is configured for the `example` namespace, together with `bar` for the `foo` property in an `*.items`-file:
```
afuechsel: missing an "s": shows

docs/documentation/concepts/items.md
```diff
@@ -137,3 +137,24 @@ Here is a short table demonstrating conversions for the examples above:
...
+
+    Switch light { channel="...", example="something" [ foo="bar" ] }
```
afuechsel: I don't like this example. A bit more descriptive would be nice, e.g.: 

The following example shows how `lighting` is configured for the `alexa` namespace, together with `night` for the `timeOfDay` property in an `*.items`-file:
```` 
 Switch light { channel="...", alexa="lighting" [ timeOfDay="night" ] }
````

Comment (sjka) [reply docs/documentation/concepts/items.md]: > A bit more descriptive would be nice

Fully agree. I just wanted to prevent any "I have copied the alexa configuration from the documentation and it doesn't work"-situation. But you're right, the example should be improved.

File: bundles/core/org.eclipse.smarthome.core/src/main/java/org/eclipse/smarthome/core/internal/items/MetadataRegistryImpl.java (hunk 5 above)
Comment (sjka) [reply bundles/core/org.eclipse.smarthome.core/src/main/java/org/eclipse/smarthome/core/internal/items/MetadataRegistryImpl.java]: Not until `AbstractRegistry` has it. Which would suggest not to do within this PR...

File: bundles/core/org.eclipse.smarthome.core/src/main/java/org/eclipse/smarthome/core/items/ManagedMetadataProvider.java (hunk 6 above)
Comment (sjka) [reply bundles/core/org.eclipse.smarthome.core/src/main/java/org/eclipse/smarthome/core/items/ManagedMetadataProvider.java]: same here

File: bundles/config/org.eclipse.smarthome.config.core/src/main/java/org/eclipse/smarthome/config/core/items/MetadataConfigDescriptionProvider.java [hunk 13]
```diff
@@ -0,0 +1,69 @@
...
+
+import java.util.List;
+import java.util.Locale;
+
+import org.eclipse.jdt.annotation.Nullable;
+import org.eclipse.smarthome.config.core.ConfigDescriptionParameter;
+import org.eclipse.smarthome.config.core.ParameterOption;
+
+/**
+ *
```
kaikreuzer: Some general JavaDoc for that interface would be nice.

bundles/config/org.eclipse.smarthome.config.core/src/main/java/org/eclipse/smarthome/config/core/items/MetadataConfigDescriptionProvider.java
```diff
@@ -0,0 +1,69 @@
...
+     * @return the metadata namespace description
+     */
+    default String getDescription(@Nullable Locale locale) {
+        return getNamespace();
+    }
+
+    /**
+     * Get all valid options if the main metadata value should be restricted to certain values.
+     *
+     * @param locale
```
kaikreuzer: JavaDocs!

bundles/config/org.eclipse.smarthome.config.core/src/main/java/org/eclipse/smarthome/config/core/items/MetadataConfigDescriptionProvider.java
```diff
@@ -0,0 +1,69 @@
...
+     */
+    @Nullable
+    List<ParameterOption> getParameterOptions(@Nullable Locale locale);
+
+    /**
+     * Get the config descriptions for all expected parameters.
+     * <p>
+     * This list may depend on the current "main" value
+     *
+     * @param value
```
kaikreuzer: JavaDocs!

bundles/config/org.eclipse.smarthome.config.core/src/main/java/org/eclipse/smarthome/config/core/items/MetadataConfigDescriptionProvider.java
```diff
@@ -0,0 +1,69 @@
...
+import org.eclipse.jdt.annotation.Nullable;
+import org.eclipse.smarthome.config.core.ConfigDescriptionParameter;
+import org.eclipse.smarthome.config.core.ParameterOption;
+
+/**
+ *
+ * @author Simon Kaufmann - initial contribution and API
+ *
+ */
+public interface MetadataConfigDescriptionProvider {
```
kaikreuzer: NonNullByDefault?

bundles/config/org.eclipse.smarthome.config.core/src/main/java/org/eclipse/smarthome/config/core/items/MetadataConfigDescriptionProvider.java
```diff
@@ -0,0 +1,69 @@
...
+    /**
+     * Get the human-readable description of the metadata namespace
+     * <p>
+     * Overriding this method is optional - it will default to the namespace identifier.
+     *
+     * @param locale a locale, if available
+     * @return the metadata namespace description
+     */
+    default String getDescription(@Nullable Locale locale) {
+        return getNamespace();
```
kaikreuzer: I would remove the default implementation as people won't notice that they could/should overwrite it and you have already implemented the fallback [here](https://github.com/eclipse/smarthome/pull/4390/commits/5924264500e248d386d29a2802a4cb3db88e17c4#diff-152fa07bd689237e8513d6766d765f0bR122) as well - one place should be enough and this method should be allowed to return null.

bundles/config/org.eclipse.smarthome.config.core/src/main/java/org/eclipse/smarthome/config/core/internal/items/MetadataConfigDescriptionProviderImpl.java
```diff
@@ -0,0 +1,150 @@
...
+ * See the NOTICE file(s) distributed with this work for additional
+ * information regarding copyright ownership.
+ *
+ * This program and the accompanying materials are made available under the
+ * terms of the Eclipse Public License 2.0 which is available at
+ * http://www.eclipse.org/legal/epl-2.0
+ *
+ * SPDX-License-Identifier: EPL-2.0
+ */
+package org.eclipse.smarthome.config.core.internal.items;
```
kaikreuzer: I don't like a reference to "item" in the config bundle - and I do not really see a reason for it - couldn't you simply use "metadata"?

bundles/config/org.eclipse.smarthome.config.core/src/main/java/org/eclipse/smarthome/config/core/items/MetadataConfigDescriptionProvider.java
```diff
@@ -0,0 +1,69 @@
...
+ * See the NOTICE file(s) distributed with this work for additional
+ * information regarding copyright ownership.
+ *
+ * This program and the accompanying materials are made available under the
+ * terms of the Eclipse Public License 2.0 which is available at
+ * http://www.eclipse.org/legal/epl-2.0
+ *
+ * SPDX-License-Identifier: EPL-2.0
+ */
+package org.eclipse.smarthome.config.core.items;
```
kaikreuzer: -> core.metadata

kaikreuzer:
@digitaldan & @beowulfe: I have just rolled this change out to openHAB - so the metadata infrastructure should be available for Alexa & Homekit integration in the openHAB IDE (after refreshing your target platform). Please note that you are the Guinea pigs for that - if you run into any issues, please report them!

[Here](https://github.com/eclipse/smarthome/blob/master/bundles/test/org.eclipse.smarthome.magic/src/main/java/org/eclipse/smarthome/magic/internal/metadata/MagicMetadataUsingService.java) you can find an example service that uses metadata - you'd have to do something similar to get access to it.

marziman:
@kaikreuzer you forgot me. Since the Google Assistant integration should also use the metadata approach. Can you point me on some example on how this will look like in a concrete example.
So I can go and directly bring this into Google Assistant integration?

@digitaldan & @beowulfe Yes please! Lets arrange for some "common" convention to have a not-complicated way to use this in the real world integrations.
As I didnt follow the complete PR code, I would be happy with proposals. Otherwise I can give it a try and propose sth. I already have similiarities with Alexa, but some tags differ.

Thanks & BR
Mehmet

kaikreuzer:
> you forgot me

I didn't, I just thought you do not want to be a Guinea pig 🐷 !
But sure, the same is also relevant for Google Assistant integration.

> Can you point me on some example on how this will look like in a concrete example.

I gave an example DSL in the [comment above](https://github.com/eclipse/smarthome/pull/4390#issue-145476911) and a link to the code that consumes this data [just before](https://github.com/eclipse/smarthome/pull/4390#issuecomment-381139834). Some short documentation had been [added here] - that's all there is right now.

beowulfe:
Awesome! One thing that adds a decent amount of complexity to the HomeKit addon is the code that tracks items in a Group that need to be composed to form a HomeKit accessory. If I was to abstract that out from HomeKit, so we only have one implementation instead of many, where's the right place for that?

digitaldan:
@kaikreuzer fantastic
>I have just rolled this change out to openHAB 

So is this available in the nightly builds then?

digitaldan:
>So is this available in the nightly builds then?

nevermind, I see that it is :-)

kaikreuzer:
> If I was to abstract that out from HomeKit, so we only have one implementation instead of many, where's the right place for that?

@beowulfe I am not quite sure what kind of methods/features you have in mind for this. Could you elaborate on it / give an example? Is it something that could be potentially added to the `MetadataRegistry`?

beowulfe:
@kaikreuzer - that might work. Let me get something working and I'll start a PR for discussion.

marziman:
Hi @beowulfe @digitaldan 
did you already included this? Since it is a good point (and people ask me, in tagging confusion) I would be happy to follow (or help to create) the new semantics.

BR Mehmet

digitaldan:
@marziman yes we have this already implemented in the v3 version of the Alexa skill, see https://github.com/openhab/openhab-alexa/tree/v3 .  

@marziman  and @beowulfe while we have a specific metadata syntax for Alexa v3, we have been discussing keeping a more basic form that would keep things simpler for most users, and is similar to the syntax we have in tags.  We would like to have consistency with homekit and google on this.  We are proposing to support the following "labels"
```
Lighting
Switchable
Lock
Thermostat
CurrentTemperature
TargetTemperature
(homekit:)HeatingCoolingMode
LowerTemperature (new)
UpperTemperature (new)
ColorTemperature (new)
Activity (new)
Scene (new)
EntertainmentChannel
EntertainmentInput
MediaPlayer
SpeakerMute
SpeakerVolume
```

I realize this is probably the wrong thread to discuss such things, but wdyt as a starting place?  See https://github.com/openhab/openhab-alexa/tree/v3#labels-concept for more info on "labels".
 
@jsetton has been working on this as you can see at https://github.com/openhab/openhab-alexa/issues/68

marziman:
@digitaldan:

Your proposal is pretty cool on Google Assistant side. We currently use these tags:
- Lighting
- Switchable
- Thermostat
- CurrentTemperature

So this fits to the proposal. Fine from GA side, and I would like to switch to the metadata style these days.

@beowulfe: Is that fine for you, too?

Just need to setup my installation and need to try it out.

BR Mehmet

beowulfe:
Those look great to me as well. Thanks for pulling those together @digitaldan

digitaldan:
>Those look great to me as well. Thanks for pulling those together @digitaldan

As much as I would like to take the credit, @jsetton  came up with those :-)

PR #4399, opened 2017-10-11
fix to prevent double upload of p2metadata artifacts in maven 3.3.X o…

sobimibos:
…r later

### DIFF
--- a/features/pom.xml
+++ b/features/pom.xml
@@ -28,4 +28,20 @@
     <module>karaf</module>
   </modules>
 
+  <build>
+    <pluginManagement>
+      <plugins>
+        <plugin>
+          <groupId>org.eclipse.tycho</groupId>
+          <artifactId>tycho-p2-plugin</artifactId>
+          <version>${tycho-version}</version>
+          <configuration>
+            <!-- fix to prevent double upload of p2metadata artifacts  -->
+            <defaultP2Metadata>false</defaultP2Metadata>
+          </configuration>
+        </plugin>
+      </plugins>
+    </pluginManagement>
+  </build>
+
 </project>


sjka:
Could you please elaborate a little on why this change is needed (considering that the ESH build is running fine)?

sobimibos:
When using Maven 3.3.X or later and the new static code analyse (which requires at least Maven 3.3.X), the tycho-p2-plugin will deploy some artifacts (p2metadata.xml) twice to the remote repository.

This causes no harm for snapshot builds but release builds will fail because the remote repository (like nexus) does not allow multiple uploads of the same release version of an artifact.

This is a known bug, which is described here: https://bugs.eclipse.org/bugs/show_bug.cgi?id=484076. The workaround is described there as well.

 


Gesendet: Donnerstag, 12. Oktober 2017 um 11:04 Uhr
Von: "Simon Kaufmann" <notifications@github.com>
An: eclipse/smarthome <smarthome@noreply.github.com>
Cc: "Michael Bock" <mbock@gmxpro.de>, Author <author@noreply.github.com>
Betreff: Re: [eclipse/smarthome] fix to prevent double upload of p2metadata artifacts in maven 3.3.X o… (#4399)


Could you please elaborate a little on why this change is needed (considering that the ESH build is running fine)?

sjka:
Okay, thanks for the explanation. I was trying to identify the issue somehow in our build logs, but apparently we are either not deploying (nightly) or skipping the checks (stable).

kaikreuzer:
We haven't deployed any release since adding the SAT to the build, we only did snapshots so far. So I guess we should merge this to get a successful next release :-)

Issue #4321, opened 2017-09-22
Null annotations - observations and discussion


htreu:
After being confronted with the new null annotations for the first time now I would like to show some issues I came across and discuss with you how we manage to deal with them. The original PR and discussion was here: #4080.

All my examples assume the `@NonNullByDefault` annotation on class level. This implies all members, return values and method parameters are expected to be non null.

**Class members of type Map<K,V>**
The compiler treads members of type Map<K,V> as `@NonNull Map<@NonNull K, @NonNull V>`. This implies that a call to `myMap.get("invalidKey")` will return a non null value, even if the Map does not contain the key. In the process the following code will be reported as _dead code_:
```
Object value = myMap.get("invalidKey");
if (value == null) {
    throw ... // this block reported as "dead code"
}
```

**Implicit null checks**
Given a class with constructor `MyThing(String param1)` which is all non null we get the following behaviour during creation:
```
@Nullable String param1 = getConfigParam(...);
if (param1 != null) {
    new MyThing(param1); // this will work okay, null check accepted
}

if (StringUtils.isBlank(param1) {
   new MyThing(param1); // this will fail with error, implicit null check not accepted
}
```

This could very easily lead to a change in programming style which is required to adopt to the null annotations: To express the semantically _blank_ check a `StringUtils.isBlank(str)` will become a `str == null || str.isEmpty()` which is technically equal but imho feels wrong.

**Dependency injection**
Class members which are injected by OSGi DeclarativeService (DS) are not supported well: When doing no extra annotation, the often used mechanism to set these members to `null` on unbind will resolve to error. When giving the member a `@Nullable` (which semantically is correct) all invocations now require a null check since the compiler warns about potential null pointer access.
Both "positions" do not reflect the actual behaviour since we know the framework will take care of the dependency (in most cases) and we can be sure not to have a null pointer at hand.

Some of the above issues where already addressed [here](https://github.com/eclipse/smarthome/pull/4290/files#r140263834) but my feeling is we need to have another discussion on how to deal with the new situation. Right now my IDE floods my with warnings which do not seem to resolve magically.

htreu:
Another observation I want to add regards JUnit test cases:

**JUnit tests**
My Tests often follow this scheme:

```
private static final String CONST = "value";

private MyClass unitUnderTest;

public setup() {
    unitUnderTest = new MyClass(CONST)
}
```

[...]

PR #4402, opened 2017-10-11
Null annotations: alter severity of IDE warnings

htreu:
This addresses the proposal mentioned here https://github.com/eclipse/smarthome/issues/4321#issuecomment-332112083 to reduce the severity of some null annotation warnings. Also the "Null Reference" check was raised to `Error`.

Also: Make null reference check happy in `AbstractManagedProvider`.

### DIFF
--- a/bundles/core/org.eclipse.smarthome.core/src/main/java/org/eclipse/smarthome/core/common/registry/AbstractManagedProvider.java
+++ b/bundles/core/org.eclipse.smarthome.core/src/main/java/org/eclipse/smarthome/core/common/registry/AbstractManagedProvider.java
@@ -122,7 +122,7 @@ public abstract class AbstractManagedProvider<E extends Identifiable<K>, K, PE> 
         String key = getKeyAsString(element);
         if (storage.get(key) != null) {
             PE persistableElement = storage.put(key, toPersistableElement(element));
-            if (persistableElement != null) { // redundant but makes null reference check happy.
+            if (persistableElement != null) {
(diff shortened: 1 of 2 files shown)


bundles/core/org.eclipse.smarthome.core/src/main/java/org/eclipse/smarthome/core/common/registry/AbstractManagedProvider.java
```diff
@@ -122,10 +122,12 @@ public E update(E element) {
         String key = getKeyAsString(element);
         if (storage.get(key) != null) {
             PE persistableElement = storage.put(key, toPersistableElement(element));
-            E oldElement = toElement(key, persistableElement);
-            notifyListenersAboutUpdatedElement(oldElement, element);
-            logger.debug("Updated element {} in {}.", key, this.getClass().getSimpleName());
-            return oldElement;
+            if (persistableElement != null) { // redundant but makes null reference check happy.
```
maggu2810: This is not redundant.
The storage member could be changed between `storage.get(key)` and `storage.put(key, ...`.

Or am I wrong? I don't see any synchronized statements that ensure storage isn't changed.
So, I agree with the change but IMHO the comment should be removed.

htreu: Yes, you are right. The comment actually refers only to `storage.get(key)` but parallel modification does indeed require this call. I will remove the comment.

PR #4403, opened 2017-10-11
Thing REST config exception logging

cdjackson:
This probably isn't a good implementation, but I think we need to improve the exception logging in the REST resource(s). Currently, if there's an exception during a PUT, we end up with no information on how to find the issue.

Users are reporting for example -:
```
2017-Oct-06 20:08:00.760 [ERROR] [marthome.io.rest.core.internal.thing.ThingResource] - Exception during HTTP PUT request for update config at 'things/zwave:device:07cb40a2:node12/config' for thing 'zwave:device:07cb40a2:node12': null
```

I would like to log the actual exception so we have a way to debug (even if this PR isn't the best approach ;) ).

### DIFF
--- a/bundles/io/org.eclipse.smarthome.io.rest.core/src/main/java/org/eclipse/smarthome/io/rest/core/internal/thing/ThingResource.java
+++ b/bundles/io/org.eclipse.smarthome.io.rest.core/src/main/java/org/eclipse/smarthome/io/rest/core/internal/thing/ThingResource.java
@@ -419,8 +419,7 @@ public class ThingResource implements RESTResource {
                     ex.getValidationMessages());
             return Response.status(Status.BAD_REQUEST).entity(ex.getValidationMessages(locale)).build();
         } catch (Exception ex) {
-            logger.error("Exception during HTTP PUT request for update config at '{}' for thing '{}': {}",
-                    uriInfo.getPath(), thingUID, ex.getMessage());
+            logger.error("Exception during HTTP PUT request for update config at '{}'", uriInfo.getPath(), ex);
             return JSONResponse.createResponse(Status.INTERNAL_SERVER_ERROR, null, ex.getMessage());
         }
 


bundles/io/org.eclipse.smarthome.io.rest.core/src/main/java/org/eclipse/smarthome/io/rest/core/internal/thing/ThingResource.java
```diff
@@ -411,6 +411,7 @@ public Response updateConfiguration(@HeaderParam(HttpHeaders.ACCEPT_LANGUAGE) St
         } catch (Exception ex) {
             logger.error("Exception during HTTP PUT request for update config at '{}' for thing '{}': {}",
                     uriInfo.getPath(), thingUID, ex.getMessage());
+            logger.debug("Exception during HTTP PUT request ", ex.getMessage());
```
sjka: This doesn't actually add any additional value, I think you meant something like this:

```java
if (logger.isDebugEnabled()) {
    logger.debug("", ex);
}
```
right?

htreu: isn´t the `logger.isDebugEnabled()` call only required when extensive String concats or other  "expensive" operations are done before the call to `logger.debug` actually happens? Doing only `logger.debug("oops", ex)` should be okay here.

sjka: Sure, good point - yes it is!

What I actually _meant_ to write down (should have copy-and-pasted it) was:

```java
if (logger.isDebugEnabled()) {
    logger.error("", ex);
}
```

but simply a 

```java
logger.debug("", ex);
```

would totally suffice of course.

maggu2810: @htreu Do you consider to change `ex.getMessage()` to `ex`?

htreu: Yes, on a second though I also wouldn't mind to have the full exception with stack trace in the original logging. Since we already log an error we consider this as a critical bug which results typically from coding mistake.

@cdjackson wdyt?

kaikreuzer: @cdjackson ping!

cdjackson: Sorry - not sure why I didn't see any of these comments (I don't seem to have any emails). Anyway, yes, I would be perfectly happy if we just had the single log and it logged the full stack track.

Do you want me to up update this PR with such implementation? (might be early next week as I'm now in DE/FR enroute to the conference).

triller-telekom: I would also vote for logging the full exception `ex` instead of just `ex.getMessage()`.

htreu:
hi @cdjackson 

> Do you want me to up update this PR with such implementation?

that would be nice :-)
Looking forward.

kaikreuzer:
Hey @cdjackson, could you finish this small PR, so that it can be merged?

cdjackson:
Sorry for the delay - I've updated the PR. Hopefully it's an acceptable solution ;)

cdjackson:
Are you sure that works here?

I thought that the definition for the logger commands will only print the stacktrace if it's the second argument? (maybe I'm wrong though!)

cdjackson:
Ok, looks like I remembered incorrectly ;) I've updated it...

maggu2810:
The exception reference itself should be used as last element without an parameter in the string.

maggu2810:
You have written:

> I thought that the definition for the logger commands will only print the stacktrace if it's the second argument? (maybe I'm wrong though!)

I tried to explain: 

> The exception reference itself should be used as last element without an parameter in the string.

---

> Sorry - I’m not sure what you mean - isn’t that what I’ve done?

Yes, it is what you have done in your last commit.

cdjackson:
Ah - ok, sorry - I thought you were commenting on the last commit. Thanks for the explanation :)

bundles/io/org.eclipse.smarthome.io.rest.core/src/main/java/org/eclipse/smarthome/io/rest/core/internal/thing/ThingResource.java
```diff
@@ -419,8 +419,7 @@ public Response updateConfiguration(@HeaderParam(HttpHeaders.ACCEPT_LANGUAGE) St
                     ex.getValidationMessages());
             return Response.status(Status.BAD_REQUEST).entity(ex.getValidationMessages(locale)).build();
         } catch (Exception ex) {
-            logger.error("Exception during HTTP PUT request for update config at '{}' for thing '{}': {}",
-                    uriInfo.getPath(), thingUID, ex.getMessage());
+            logger.error("Exception during HTTP PUT request for update config at '{}'", uriInfo.getPath(), ex);
```
htreu: I guess `thingUID` is also part of `uriInfo.getPath()`.

cdjackson: Yes - I didn't see the point in duplicating the UID given we were increasing the logging...

htreu:
This is ready to get merged.

kaikreuzer:
Wow, what a long discussion for a single line logging statement 😇

PR #4404, opened 2017-10-12
Fix Hue test: create internal package

htreu:

### DIFF
--- a/extensions/binding/org.eclipse.smarthome.binding.hue.test/src/test/groovy/org/eclipse/smarthome/binding/hue/internal/HueLightHandlerOSGiTest.groovy
+++ b/extensions/binding/org.eclipse.smarthome.binding.hue.test/src/test/groovy/org/eclipse/smarthome/binding/hue/internal/HueLightHandlerOSGiTest.groovy
@@ -5,7 +5,7 @@
  * which accompanies this distribution, and is available at
  * http://www.eclipse.org/legal/epl-v10.html
  */
-package org.eclipse.smarthome.binding.hue.test
+package org.eclipse.smarthome.binding.hue.internal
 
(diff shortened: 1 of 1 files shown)


sjka:
I'm really struggling with figuring out what was the problem here. They worked before in the IDE, in maven locally, on Travis & Jenkins. You are outnumbered 😉 

Do you have a log of what went wrong for you? I mean, changing the package certainly doesn't hurt, especially as the test previously was in its own package, but the difference in the builds somehow makes me nervous...

sjka:
Hmmm, now one of the tests had its moment of madness:

```
startSearchIsCalled(org.eclipse.smarthome.binding.hue.internal.HueLightDiscoveryServiceOSGiTest)  Time elapsed: 10.059 sec  <<< FAILURE!
java.lang.AssertionError: 
Expected: is not null
     but: was null
	at org.junit.Assert.assertThat(Assert.java:956)
	at org.junit.Assert.assertThat(Assert.java:923)
	at org.eclipse.smarthome.binding.hue.internal.HueLightDiscoveryServiceOSGiTest.lambda$3(HueLightDiscoveryServiceOSGiTest.java:200)
```

htreu:
> what was the problem here

On master my IDE shows an error on `org.eclipse.smarthome.binding.hue.test.HueLightHandlerOSGiTest` because the outer class `o.e.s.binding.hue.internal.HttpClient` is not visible in the import statement of the test. 
Besides from that I can see that the test compiles and runs fine locally using maven and also in the IDE.

This infers that the public static inner class `Result` is visible and accessible, although its outer class is not (from a different package).

This PR fixes the error marker by moving the test into the same package as the outer class.

htreu:
Closing in favour of #4413.

PR #4405, opened 2017-10-12
Fix oxygen repo URL in targetplatform

htreu:


sjka:
IMHO the constellation is a little weird, but due to the lack of a better proposal from my side let's use it for now in order to get the builds going again.

PR #1888, opened 2016-07-21
Support state options in control UI

aounhaider1:
fixes https://github.com/eclipse/smarthome/issues/1886
Signed-off-by: Aoun Bukhari <bukhari@itemis.de>

PR #4408, opened 2017-10-12
[PaperUI] fix UNDEF/NULL display

### LINKED ISSUE #4407: [PaperUI] UndefType.NULL is shown as -NaN
On a NumberItem I have the value of UndefType.NULL. In PaperUI Control it is shown as "-NaN", which may be misleading. I would suggest to print something like "UNDEFINED" instead.

sjka:
there used to be a `-` displayed only, but apparently this got broken as a side-effect of #1888 and became `-NaN`. This is my attempt to bring back our lovely `-`.

fixes #4407


mdicke2s:
That was fast :-) thank you very much

PR #4039, opened 2017-08-12
Fix label/description localization for channel definition

lolodomo:
Signed-off-by: Laurent Garnier <lg.hc@free.fr>

PR #4411, opened 2017-10-12
[i18n] Fixed translation of channel definitions (label and description)

cweitkamp:
- Added test case for translating channel definitions and channel group definitions
- Fixed translation of channel definitions (suggested by @ppieczul )
- Added 'channel' identifier in translation component
- Updated documentation

See https://github.com/openhab/openhab2-addons/pull/2768#issuecomment-335045841

An additional step I want to test if it's working for channel definitions with default system channel-types, too. That didn't work before. See https://github.com/eclipse/smarthome/issues/3623#issuecomment-308498668

Also-by: Pawel Pieczul <pieczul@gmail.com>

### DIFF
--- a/bundles/core/org.eclipse.smarthome.core.thing.xml.test/src/test/java/org/eclipse/smarthome/core/thing/xml/test/SystemWideChannelTypesTest.java
+++ b/bundles/core/org.eclipse.smarthome.core.thing.xml.test/src/test/java/org/eclipse/smarthome/core/thing/xml/test/SystemWideChannelTypesTest.java
@@ -103,31 +103,27 @@ public class SystemWideChannelTypesTest extends JavaOSGiTest {
         // install test bundle
         Bundle sysBundle = SyntheticBundleInstaller.install(bundleContext, SYSTEM_CHANNELS_BUNDLE_NAME);
         assertThat(sysBundle, is(notNullValue()));
-        List<ThingType> thingTypes = thingTypeProvider.getThingTypes(null).stream()
-                .filter(it -> it.getUID().getId().equals("wireless-router")).collect(Collectors.toList());
 
-        assertThat(thingTypes.size(), is(1));
+        ThingType wirelessRouterType = thingTypeProvider.getThingTypes(null).stream()
+                .filter(it -> it.getUID().getAsString().equals("SystemChannels:wireless-router")).findFirst().get();
+        assertThat(wirelessRouterType, is(notNullValue()));
 
-        Collection<ChannelDefinition> channelDefs = thingTypes.get(0).getChannelDefinitions();
-
+        Collection<ChannelDefinition> channelDefs = wirelessRouterType.getChannelDefinitions();
         assertThat(channelDefs.size(), is(3));
 
-        List<ChannelDefinition> myChannel = channelDefs.stream().filter(
+        ChannelDefinition myChannel = channelDefs.stream().filter(
                 it -> it.getId().equals("test") && it.getChannelTypeUID().getAsString().equals("system:my-channel"))
-                .collect(Collectors.toList());
+                .findFirst().get();
+        assertThat(myChannel, is(notNullValue()));
 
-        List<ChannelDefinition> sigStr = channelDefs.stream()
-                .filter(it -> it.getId().equals("sigstr")
-                        && it.getChannelTypeUID().getAsString().equals("system:signal-strength"))
-                .collect(Collectors.toList());
+        ChannelDefinition sigStr = channelDefs.stream().filter(it -> it.getId().equals("sigstr")
+                && it.getChannelTypeUID().getAsString().equals("system:signal-strength")).findFirst().get();
+        assertThat(sigStr, is(notNullValue()));
 
-        List<ChannelDefinition> lowBat = channelDefs.stream().filter(
+        ChannelDefinition lowBat = channelDefs.stream().filter(
                 it -> it.getId().equals("lowbat") && it.getChannelTypeUID().getAsString().equals("system:low-battery"))
-                .collect(Collectors.toList());
-
-        assertThat(myChannel.size(), is(1));
-        assertThat(sigStr.size(), is(1));
-        assertThat(lowBat.size(), is(1));
+                .findFirst().get();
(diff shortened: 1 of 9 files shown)


lolodomo:
Be careful, this PR would break all existing translations due to the change of the key !
If you want to go in this direction, I would appreciate (request) an update of all existing translation files for bindings in ESH and OH2.

lolodomo:
Note that the ability to translate system channels is something I added and is in use in the Netatmo binding : https://github.com/openhab/openhab2-addons/blob/master/addons/binding/org.openhab.binding.netatmo/ESH-INF/i18n/netatmo_fr.properties#L231
Unfortunately, I discovered that it is not always working. It works well when I run from Eclipse but not in my production environment.

ppieczul:
@lolodomo Looking at the code, the way it worked for channels before this pr is:
  * if there is a localized channel TYPE, use that label and description
  * otherwise use the label and description from the CHANNEL (not type)
Would it be possible that you saw such behavior? You could see the translation for some channels and others not, depending whether they have localized types.

cweitkamp:
We have to look at this PR from two different point of views.

First we want to fix the translations for channel definitions. @lolodomo introduced this feature in PR #4039 . He contributed the same lines of code in the `ThingTypeI18nLocalizationService` class like we do in this PR. The code somehow got changed in PR #3510 . Maybe the cause of a rebase or merge conflict. @triller-telekom wdyt? What were the purposes of your changes? Were there any?

Second we should discuss my proposal to insert the `channel` identifier. Let me explain my thoughts. We have a special syntax for each basic entity - may I call it like that? - e.g. `channel-type`, `thing-type`, and so on. On top of that we have a different syntax for `config` or `state` definitions. PR #4039 introduced a way to overrule the initial definitions for labels or descriptions of a parent entity. We are able to define a `channel-type` and define different labels or descriptions for the channel definitions. Same for `channel-group-types`. Later ones are translated with this syntax `thing-type.<binding-id>.<thing-type-id>.group.<channel-group-id>.*` whereas first ones currently should be translated with this syntax `thing-type.<binding-id>.<thing-type-id>.<channel-id>.*`. IMHO there is something missing. My suggestion changes this syntax to `thing-type.<binding-id>.<thing-type-id>.channel.<channel-id>.*` which looks a little more compliant because you have to use a special keyword `channel` similar to the channel group definitions.

I agree with @lolodomo that this is an APIBreaking change and we have to adopt some translation files to keep the old functionality. But this way of translation ain't documented until this PR gets merged and the work to do will be small. I volunteer to do that.

lolodomo:
I was not aware that my code has been erased !

I am going to think about your key change proposal.

triller-telekom:
#3510 was just about adding the `category` field to the thing type. If I accidentally broke any translation feature due to a wrongly done rebasing because of a merge conflict, I am sorry. That was not my intention.

@cweitkamp While you are at it now, I would appreciate if you fix whatever I broke from #4039. So this leaves only the second point open for discussion.

lolodomo:
@cweitkamp : why not rather `channel-group.<binding-id>.<thing-type-id>.<channel-group-id>.*` and `channel.<binding-id>.<thing-type-id>.<channel-id>.*` for channel ?

lolodomo:
it looks like I used thing-type.<binding-id>.<thing-type-id>.group.<channel-group-id>.* in the WU binding and using the snapshot 1064 the translations are apparently correctly applied when I use Paper UI.

it looks like I used thing-type.<binding-id>.<thing-type-id>.<channel-id>.* in the Netatmo binding to translate system channels and using the snapshot 1064 the translations are apparently not applied when I use Paper UI. Is it what was broken and you are fixing here ?

Finally I think there is not too much bindings to update if the syntax for channel and channel group are changed. Did you use the current syntax in few bindings on your side ?

cweitkamp:
> Is it what was broken and you are fixing here ?

Yes, that is exactly the issue my fix addresses.

> Finally I think there is not too much bindings to update if the syntax for channel and channel group are changed. Did you use the current syntax in few bindings on your side ?

Agreed. I cannot remember to have used it in one of the bindings I translated. But there maybe others. If so, the number shouldn't be that high.

May I ask you to provide the French translations for the default system channel-types in the core if the ESH framework itself? (https://github.com/eclipse/smarthome/tree/master/bundles/core/org.eclipse.smarthome.core.thing/ESH-INF/i18n)

lolodomo:
WDYT about my alternative proposal for keys ?

lolodomo:
I will add the french translation for default system channels. You did not provide a translation for the description ?

cweitkamp:
> You did not provide a translation for the description ?

Because the don't have one. ;-)

> WDYT about my alternative proposal for keys ?

I already tried to explain my proposal. To be honest, imho my version looks like the right choice, because you are translating a thing type specific channel group or channel. I think it is easier to understand and fits better in a sorted properties file. And the third point would be the number of already contributed translations. The way to translate channel group types was introduced and documented some time ago, so we could expect that it is used in several bindings.

cweitkamp:
> An additional step I want to test if it's working for channel definitions with default system channel-types, too.

I added a test case for this, too. It works.

triller-telekom:
@cweitkamp Did you have a look at the bindings in ESH which might be affected by your change in the translation key? Are there any? And if so, can you please adjust them with this PR too?

cweitkamp:
No, I didn't do that, yet. Thanks for the reminder.

cweitkamp:
@triller-telekom I checked the translations of all ESH bindings. I didn't find matching translations which has to be changed.

triller-telekom:
@cweitkamp Thanks for checking. So for ESH everything seems to be fine then, but what about OH? Are there any bindings that have to be changed so a follow up PR on the openhab2-addons repo would be necessary?

cweitkamp:
In OH2 we have to change at least something in the Netatmo binding. I will provide an PR over there after a new ESH snapshot is available.

kaikreuzer:
Thanks!
FYI @cweitkamp, I'd plan a new ESH stable by the end of this week (probably Friday).

cweitkamp:
PR in OH2 https://github.com/openhab/openhab2-addons/pull/2836

PR #4412, opened 2017-10-12
Sitemap REST API: bypass proxy for HLS video playlist

lolodomo:

### DIFF
--- a/bundles/io/org.eclipse.smarthome.io.rest.sitemap/src/main/java/org/eclipse/smarthome/io/rest/sitemap/internal/SitemapResource.java
+++ b/bundles/io/org.eclipse.smarthome.io.rest.sitemap/src/main/java/org/eclipse/smarthome/io/rest/sitemap/internal/SitemapResource.java
@@ -300,9 +300,10 @@ public class SitemapResource implements RESTResource, SitemapSubscriptionCallbac
                             logger.debug("Received HTTP GET request at '{}' for the unknown page id '{}'.", uri,
                                     pageId);
                         } else {
-                            logger.debug("Received HTTP GET request at '{}' for the page id '{}'. "
-                                    + "This id refers to a non-linkable widget and is therefore no valid page id.", uri,
-                                    pageId);
+                            logger.debug(
+                                    "Received HTTP GET request at '{}' for the page id '{}'. "
+                                            + "This id refers to a non-linkable widget and is therefore no valid page id.",
+                                    uri, pageId);
                         }
                     }
                     throw new WebApplicationException(404);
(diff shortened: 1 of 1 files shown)


bundles/io/org.eclipse.smarthome.io.rest.sitemap/src/main/java/org/eclipse/smarthome/io/rest/sitemap/internal/SitemapResource.java
```diff
@@ -471,16 +472,20 @@ private WidgetDTO createWidgetBean(String sitemapName, Widget widget, boolean dr
...
-            if (uri.getPort() < 0 || uri.getPort() == 80) {
-                bean.url = uri.getScheme() + "://" + uri.getHost() + "/proxy?sitemap=" + sitemapName
-                        + ".sitemap&widgetId=" + wId;
+            if (videoWidget.getEncoding() != null && videoWidget.getEncoding().toLowerCase().contains("hls")) {
+                bean.url = videoWidget.getUrl();
             } else {
-                bean.url = uri.getScheme() + "://" + uri.getHost() + ":" + uri.getPort() + "/proxy?sitemap="
-                        + sitemapName + ".sitemap&widgetId=" + wId;
+                String wId = itemUIRegistry.getWidgetId(widget);
+                if (uri.getPort() < 0 || uri.getPort() == 80) {
```
htreu: matter of taste: can this be refactored to use a `StringBuilder`? like:

```
StringBuilder sb = new StringBuilder();
sb.append(uri.getScheme()).append("://").append(uri.getHost());
if (uri.getPort() < 0 || uri.getPort() == 80) {
    sb.append(":").append(uri.getPort());
}
sb.append("/proxy?sitemap=").append(sitemapName).append(".sitemap&widgetId=").append(wId);
bean.url = sb.toString();
```

lolodomo: Yes and a method could ne created as the same code is used twice, once for image and once for video.

lolodomo:
Changed as requested.

bundles/io/org.eclipse.smarthome.io.rest.sitemap/src/main/java/org/eclipse/smarthome/io/rest/sitemap/internal/SitemapResource.java
```diff
@@ -510,6 +501,17 @@ private WidgetDTO createWidgetBean(String sitemapName, Widget widget, boolean dr
         return bean;
     }
 
+    private String buildProxyUrl(String sitemapName, Widget widget, URI uri) {
+        String wId = itemUIRegistry.getWidgetId(widget);
+        StringBuilder sb = new StringBuilder();
+        sb.append(uri.getScheme()).append("://").append(uri.getHost());
+        if (uri.getPort() >= 0 && uri.getPort() != 80) {
```
maggu2810: You moved this method only to another method, but is it really correct?
If the scheme is http the default port is 80, so all is fine.
If the scheme is https the default port should be 443 and so `uri.getPort()` should be checked to be 443.
Okay, this only affect the strange situation if someone runs https on port 80, so we should perhaps never run into.
But shouldn't it be fixed?

lolodomo: We could set the port in URL everytime ?

maggu2810: I agree. If it is the default one, it does not matter..

lolodomo:
Change done.

PR #4413, opened 2017-10-12
The HueLightHandlerOSGiTest uses this class, which results in an erro…

kaikreuzer:
…r marker in the IDE as it is package private.

As we are using OSGi and not exporting this package, it is imho fully ok to declare the class as public instead.


htreu:
Duplicate of https://github.com/eclipse/smarthome/pull/4404. In #4404 I moved the test into the same package as the package private outer class.

kaikreuzer:
Isn't this here the simpler fix? As the test is not testing HttpClient, there is no strong reason to have it in the same package...

htreu:
Agreed.

PR #4278, opened 2017-09-15
Refactored OSGi/Guice integration of DSL based models

sjka:
I just had a "quick look" into why refreshing model bundles causes such a trouble (i.e. #3157) and this is what I ended up with:

A new ServiceModule is introduced which takes care to inject ESH services via Guice in order to prevent occasional "ScriptServiceUtil is not initialized" errors.

Also, the EMF registries are cleaned up if model parsers are stopped and their corresponding resources are removed from the ModelRepository. This is to prevent class cast exceptions when old instances hit a newer version of the rule engine.

fixes #3157
Signed-off-by: Simon Kaufmann <simon.kfm@googlemail.com>

PR #4422, opened 2017-10-16
fix OSGi service injection into Language Server modules

sjka:
...doing it in a similar way as in #4278

relates to openhab/openhab-vscode#31


kubawolanin:
Thanks @sjka for the quick fix. I've tested it with openHAB distro 1063 but it doesn't seem to help. See my comment here: https://github.com/openhab/openhab-vscode/pull/31#issuecomment-337003152

Issue #2267, opened 2016-10-06
An item should not be editable if it is bound to read only channel 


emiovska:
When an item is provided from the .items file and has a custom pattern, the state description readOnly property in GenericItemProvider  is set to false by default.

Currently there are two state description providers: **GenericItemProvider** and **ChannelStateDescriptionProvider** which are sorted descending by service ranking.
Moreover the ChannelStateDescriptionProvider has service ranking -1 set in the **ChannelStateDescriptionProvider.xml** file which means that
GenericItemProvider is before the ChannelStateDescriptionProvider.

When getStateDescription method is called to get the state description of an item, the method iterates over the registered  StateDescriptionProviders and returns the first found not null value. That means that if the item is provided from .items file will always has the readOnly property in the stateDescription set to false.

According to description above there is a situation when an item provided from .items file and bound to read only channel can be edited from the UI. 
According to me, this is not the desired behavior because if the item is bound to read only channel 
(like sensor or web service) does not make sense that the item can be edited from the UI.

So, should this be considered as a bug?

kaikreuzer:
Sorry, my reply is slightly late 😲 
Yes, you are right - this is currently a limitation; you cannot define stuff like min/max/read-only in item files while the whole StateDescription information is taken from it.
What would be needed is to somehow merge information from different StateDescriptionProviders - this is something @lolodomo [suggested here](https://github.com/eclipse/smarthome/pull/3825#issuecomment-314504609) in a similar way.

lolodomo:
And my proposal was rejected lol

I was not axare of this issue but it is anither example showing why we need a kind of merge.

kaikreuzer:
> And my proposal was rejected lol

... in the context of the addressed issue. That doesn't mean that the idea itself has been rejected, that's why I pinged you here ;-)

lpapazow:
This bug can still be reproduced after [this PR](https://github.com/eclipse/smarthome/pull/3825)  has been merged. 

I could create a PR which fixes this issue in a [similar  way](https://github.com/eclipse/smarthome/pull/3825/files#diff-12183f2bd5de40fa6c6bcf35817f8aacR425) by assigning the result variable to a new StateDescription which takes the isReadOnly value from the current StateDescription and all other parameters from the previous result variable.

Would that be an appropriate way to fix the issue?

kaikreuzer:
Should be fixed by https://github.com/eclipse/smarthome/pull/4882.

PR #4428, opened 2017-10-17
the returned StateDescription combines values from GenericItemProvide…

lpapazow:
Fix for issue [2267](https://github.com/eclipse/smarthome/issues/2267).

This fix relies on the fact that  **ChannelStateDescriptionProvider** has service ranking -1 and therefore when iterating through the stateDescriptionProviders in the **getStateDescription** method,  **ChannelStateDescriptionProvider** will come after **GenericItemProvider**.

If an item that is provided from the .items file and has a custom pattern, then the **GenericItemProvider** will provide a non null **StateDescription** and on the next iteration, the result will take the **isReadOnly** value from the **ChannelStateDescriptionProvider**.

### DIFF
--- a/bundles/core/org.eclipse.smarthome.core/src/main/java/org/eclipse/smarthome/core/items/GenericItem.java
+++ b/bundles/core/org.eclipse.smarthome.core/src/main/java/org/eclipse/smarthome/core/items/GenericItem.java
@@ -408,6 +408,11 @@ abstract public class GenericItem implements ActiveItem {
                     result = stateDescription;
                 }
 
+                if (result != null && stateDescription != null) {
+                    result = new StateDescription(result.getMinimum(), result.getMaximum(), result.getStep(),
+                            result.getPattern(), stateDescription.isReadOnly(), result.getOptions());
+                }
+
                 // if the current StateDescription does provide options and we don't already have some, we pick them up
                 // here
                 if (stateDescription != null && !stateDescription.getOptions().isEmpty() && stateOptions.isEmpty()) {
--- a/bundles/io/org.eclipse.smarthome.io.rest.core.test/src/test/java/org/eclipse/smarthome/io/rest/core/internal/item/EnrichedItemDTOMapperWithTransformOSGiTest.java
+++ b/bundles/io/org.eclipse.smarthome.io.rest.core.test/src/test/java/org/eclipse/smarthome/io/rest/core/internal/item/EnrichedItemDTOMapperWithTransformOSGiTest.java
@@ -11,8 +11,10 @@ import static org.hamcrest.CoreMatchers.*;
 import static org.junit.Assert.assertThat;
 import static org.mockito.Mockito.when;
 import static org.mockito.MockitoAnnotations.initMocks;
+import static org.mockito.Mockito.mock;
 
 import java.math.BigDecimal;
+import java.util.Arrays;
 import java.util.List;
 
 import org.eclipse.smarthome.core.library.items.NumberItem;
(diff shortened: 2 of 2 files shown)


bundles/core/org.eclipse.smarthome.core/src/main/java/org/eclipse/smarthome/core/items/GenericItem.java
```diff
@@ -408,6 +408,11 @@ public StateDescription getStateDescription(Locale locale) {
                     result = stateDescription;
                 }
 
+                if (result != null && stateDescription != null) {
```
kaikreuzer: You are here in any case overwriting a previous readOnly information, I don't think this is a valid approach.

htreu:
We had a similar discussion once in #2729. The outcome was that only the state description options where taken from the `ChannelStateDescriptionProvider` in case the `GenericItemProvider` had its own state description to offer.

One solution would be to take all properties from the `ChannelStateDescriptionProvider` the "other"  StateDescriptionProviders can not offer (e.g. `readOnly`, `options`, `min,max,step`). But that would be a convention (which I dont like btw.).

kaikreuzer:
> which I dont like btw.

Could you elaborate on why you do not like it? Taking the info from the first (ordered by prio) provider that serves a value (and not null), sounds to me like a valid approach (but as stated above, I have the feeling that I am missing something, so your answer to this question might help on this...)

htreu:
My reasons for dislike for a conventional "merging" of two or more StateDescriptionProviders:

- Its conventional and thus must be explained to everyone. Its not an easy algorithm but has exceptions which treat some properties different then others.
- It does not scale. While the "fix" might help us right now with 2 StateDescriptionProviders it will become more cumbersome when adding more of them: with the exceptions made (e.g. the `readOnly` property) we have a hard time telling for the n-th StateDescriptionProvider if `readOnly=false` is its default or set by intention. Although we might get around this by using `Boolean` instead of `boolean` here (but API breaking...).

ppieczul:
What if getStateDescription was modified to also pass the current (from the previous provider iterated) state description, so the queried provider can modify just he parameters it cares about and keep the others intact? This would be kind of distributed merging, would not break the API used by the bindings and need modifications of the providers only.

htreu:
@ppieczul what I like about the 'distributed merging' idea is the fact that we don´t have to come up with one algorithm now which will then be hard to modify in the future but leave the decision to every single implementation. It will although also be API breaking since StateDescriptionProviders would have to incorporate the new behaviour.

htreu:
Maybe we can gather more brain power, ideas and comments:
@lolodomo you did the last implementation of merging the state options in #3825 , wdyt?

From a framework perspective we could also factor out some StateDescriptionMergingService and do a default implementation which does a most "simple" in place merging (see https://github.com/eclipse/smarthome/pull/4428#issuecomment-338174587) but leave flexibility for solutions to define their own behaviour.

@kaikreuzer wdyt? /cc @sjka & @maggu2810

lolodomo:
On my side, I was in favor of a merge because it is required to fix a current bug we have to get the right label. This was my reason to initially implement the merge solution.

maggu2810:
If the `getStateDescription` method get the current one return a new / modified one, we need to call this method for every state description provider starting from the lowest service ranking to the highest service ranking. Correct?

We could also keep the method as it is and create a `StateDescriptionMergingService` that calls every provider for a state description starting from the highest service ranking to the lowest as long as there is a value that is null for all previously called providers. All values from the highest ones are used, if there is a null value, the next (priority sorted) provider is called and that value is used (if non-null, otherwise call the next)...
If we use a migration service that handled the merging logic a solution could provide its own implementation with a higher service ranking and another merging strategy could be used.

@htreu Does this fit to your suggestion? Keeping the `getStateDescription` as it is and implement a service that merges the description using a custom (default the above one) merge strategy?

htreu:
@maggu2810 yes that was also my intention. The current `getStateDescription` does only merge state options but the default implementation of a `StateDescriptionMergingService` should merge all properties and replace the implementation from `getStateDescription`.

lpapazow:
The idea about a StateDescriptionMergingService seems great. I will try to implement one and create another PR within the next week.

htreu:
@lpapazow I guess this is superseded by #4882 now? Can we close this one then?

lpapazow:
Yes, I forgot to mention it in #4882 but this one should be closed.

PR #4429, opened 2017-10-17
Add support for Sonos One player

ivivanov-bg:
Add support for Sonos One player

### DIFF
--- a/extensions/binding/org.eclipse.smarthome.binding.sonos/ESH-INF/i18n/sonos_fr.properties
+++ b/extensions/binding/org.eclipse.smarthome.binding.sonos/ESH-INF/i18n/sonos_fr.properties
@@ -21,6 +21,8 @@ thing-type.sonos.PLAYBAR.label = PLAYBAR
 thing-type.sonos.PLAYBAR.description = Représente une barre de son Sonos
 thing-type.sonos.zoneplayer.label = Autre Sonos
 thing-type.sonos.zoneplayer.description = Représente un équipement Sonos inconnu de l'extension
+thing-type.sonos.One.label = One
+thing-type.sonos.One.description = Represente une enceinte Sonos One
 
 thing-type.config.sonos.zoneplayer.udn.label = Nom unique (UDN)
 thing-type.config.sonos.zoneplayer.udn.description = Identifie de manière unique un Sonos
--- a/extensions/binding/org.eclipse.smarthome.binding.sonos/ESH-INF/thing/One.xml
+++ b/extensions/binding/org.eclipse.smarthome.binding.sonos/ESH-INF/thing/One.xml
@@ -1,0 +1,69 @@
+<?xml version="1.0" encoding="UTF-8"?>
+<thing:thing-descriptions bindingId="sonos"
+	xmlns:xsi="http://www.w3.org/2001/XMLSchema-instance"
+	xmlns:thing="http://eclipse.org/smarthome/schemas/thing-description/v1.0.0"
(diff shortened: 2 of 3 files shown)


extensions/binding/org.eclipse.smarthome.binding.sonos/ESH-INF/thing/ONE.xml
```diff
@@ -0,0 +1,69 @@
+<?xml version="1.0" encoding="UTF-8"?>
+<thing:thing-descriptions bindingId="sonos"
+	xmlns:xsi="http://www.w3.org/2001/XMLSchema-instance"
+	xmlns:thing="http://eclipse.org/smarthome/schemas/thing-description/v1.0.0"
+	xsi:schemaLocation="http://eclipse.org/smarthome/schemas/thing-description/v1.0.0 http://eclipse.org/smarthome/schemas/thing-description-1.0.0.xsd">
+
+	<!-- Player ONE Thing Type -->
+	<thing-type id="One" listed="false">
```
kaikreuzer: Shouldn't this be better "ONE" in capitals (to be in sync with the PLAYXs?

ivivanov-bg: Well - it could be - but the the discovery has to be tuned too because the player announces itself as 'One' and not 'ONE'

lolodomo: The discovery code could ignore the case.

kaikreuzer: I'd agree that it would be better to harden the discovery code.

FrankZimmer: The real Name of Play 1 is "PLAY:1" and the real name of the one is "One" and NOT "ONE". So i recommend to use the real name for the ID. Sonos changed their pattern: http://www.sonos.com/de-de/shop/one.html

extensions/binding/org.eclipse.smarthome.binding.sonos/ESH-INF/thing/ONE.xml
```diff
@@ -0,0 +1,69 @@
...
+			<channel id="tuneinstationid" typeId="tuneinstationid" />
+			<channel id="volume" typeId="volume" />
+			<channel id="zonegroup" typeId="zonegroup" />
+			<channel id="zonegroupid" typeId="zonegroupid" />
+			<channel id="zonename" typeId="zonename" />
+			<channel id="coordinator" typeId="coordinator" />
+			<channel id="sleeptimer" typeId="sleeptimer" />
+			<channel id="currenttransporturi" typeId="currenttransporturi" />
+			<channel id="currenttrackuri" typeId="currenttrackuri" />
+		</channels>
```
lolodomo: Is there no line-in in the One ? If line-in is present, take example on the PLAY:5.

lolodomo:
If "One" is kept as thing type id, there is nothing to change in the discovery code.
Please add the following lines in the french translation:
thing-type.sonos.One.label = One
thing-type.sonos.One.description = Représente une enceinte Sonos One

extensions/binding/org.eclipse.smarthome.binding.sonos/ESH-INF/thing/ONE.xml
kaikreuzer: Ok, but then these lines are clearly wrong in the first place: https://github.com/eclipse/smarthome/pull/4429/files#diff-148f59c83d8c2a479dea48c22dd67b70R10

PR #4432, opened 2017-10-17
Migrate "Genarate All Models" launch config to sequential launch group

htreu:
Instead of calling an mwe2 workflow with all model generating workflows inside (which is broken) this provides a launch group and calls all model generating launch configs in sequence.

for reference, see https://www.eclipse.org/forums/index.php?t=msg&th=1089453&goto=1774478


kaikreuzer:
I am a bit confused now - what is the difference to https://github.com/eclipse/smarthome/pull/4421?

htreu:
TooManyOpenTabsException :-D

PR #4435, opened 2017-10-18
Introduce HttpContextFactoryService

### LINKED ISSUE #4433: HTTP resources not resolved from fragment bundles
The registered HTTP services and resources (ClassicUI, BasicUI, PaperUI) do not resolve resource requests using the bundles ClassLoader any more. This looks like a bug in the current Equinox implementation used by ESH, see https://bugs.eclipse.org/bugs/show_bug.cgi?id=526150.
This has the effect that resources from fragment bundles are not loaded any more. When deploying the org.openhab.ui.paperui fragment bundle with the IDE runtime setup the fragment resources do not get served.

Workaround: as described in the bug report we will provide our own implementation of the HttpContext which behaves correct according to the OSGi spec.

htreu:
This fixes #4433.

The new HttpContextFactoryService is able to create `org.osgi.service.http.HttpContext` for a given bundle. The default implementation guarantees a OSGi specification conform resource resolving.


bundles/io/org.eclipse.smarthome.io.net/src/main/java/org/eclipse/smarthome/io/net/http/internal/HttpContextFactoryServiceImpl.java
```diff
@@ -0,0 +1,70 @@
...
+            return true;
+        }
+
+        @Override
+        public URL getResource(String name) {
+            if (name != null) {
+                String resourceName = name;
+                if (name.startsWith("/")) {
+                    resourceName = name.substring(1);
+                }
```
maggu2810: As you need to use the check in every case, why not using one assignment all the time instead of assign it twice sometimes:
```java
String resourceName;
if (name.startsWith("/")) {
    resourceName = name.substring(1);
} else {
    resourceName = name;
}
```

kaikreuzer:
To follow up on @maggu2810's questions on #4433:

> I assume it should be able for solutions to use its own factory implementation or do we need to handle all the different possible context providers in ESH itself and make it configurable?

I am not sure. So far, no solution asked for being able to replace the http context by something custom. And for authorisation enablement, I think it makes sense to leave the context creation to the framework (as it can then best integrate with other services that might exist for the permission checks). But if we want to provide all due flexibility, it would be also fine for me to allow custom factory implementations (possibly realised through a higher service rank?).

> Do you think using a shared context for all servlets could be a use case, too?

I am not really sure whether this is possible as currently the HttpContexts always also hold the bundle for which they were created. If we do not need this, it should be ok to share the same instance, but if those are not heavy, it might not be worth to go for shared ones.

bundles/io/org.eclipse.smarthome.io.net/src/main/java/org/eclipse/smarthome/io/net/http/HttpContextFactoryService.java
```diff
@@ -0,0 +1,33 @@
...
+ *
+ * @author Henning Treu - initial contribution and API
+ *
+ */
+public interface HttpContextFactoryService {
+
+    /**
+     * Creates an {@link HttpContext} according to the OSGi specification of
+     * {@link HttpService#createDefaultHttpContext()} as described here:
+     * https://www.knopflerfish.org/releases/6.0.0/docs/javadoc/org/osgi/service/http/HttpService.html#createDefaultHttpContext
```
kaikreuzer: We imho should not reference knopflerfish.org here - anyhow, the reference to the method of the interface should be good enough.

bundles/io/org.eclipse.smarthome.io.net/src/main/java/org/eclipse/smarthome/io/net/http/HttpContextFactoryService.java
```diff
@@ -0,0 +1,33 @@
...
+     *
+     * @param bundle the bundle which will be used by this {@link HttpContext} to resolve resources.
+     * @return the {@link HttpContext} for the given bundle.
+     */
+    HttpContext createHttpContext(Bundle bundle);
```
kaikreuzer: We might also want to call this `createDefaultHttpContext`, so that it is easier to add other methods in future, such as `getUnauthenticatedHttpContext` or similar.

bundles/io/org.eclipse.smarthome.io.net/src/main/java/org/eclipse/smarthome/io/net/http/internal/HttpContextFactoryServiceImpl.java
```diff
@@ -0,0 +1,72 @@
...
+import org.osgi.framework.Bundle;
+import org.osgi.service.component.annotations.Component;
+import org.osgi.service.http.HttpContext;
+import org.osgi.service.http.HttpService;
+
+/**
+ * Create {@link HttpContext} instances when registering servlets, resources or filters using the
+ * {@link HttpService#registerServlet} and corresponding methods.
+ * The resulting {@link HttpContext} complies with the OSGi specification when it comes to resource resolving, see
+ * https://www.knopflerfish.org/releases/6.0.0/docs/javadoc/org/osgi/service/http/HttpService.html#createDefaultHttpContext
```
kaikreuzer: remove link

extensions/ui/org.eclipse.smarthome.ui.paper/src/main/java/org/eclipse/smarthome/ui/paper/internal/PaperUIApp.java
```diff
@@ -7,7 +7,14 @@
  */
 package org.eclipse.smarthome.ui.paper.internal;
 
+import org.eclipse.smarthome.io.net.http.HttpContextFactoryService;
+import org.osgi.framework.Bundle;
+import org.osgi.framework.BundleContext;
 import org.osgi.service.component.ComponentContext;
+import org.osgi.service.component.annotations.Component;
```
kaikreuzer: don't you need those as optional import in the manifest?

htreu: yes. The import is also missing in the o.e.s.io.net MANIFEST. I dont know how that even worked.

maggu2810: As long as they exist in the classpath on compile time no one will complain.
Missing annotation class definitions should not matter on runtime.

htreu: anyhow: added and marked as optional.

maggu2810: https://stackoverflow.com/a/3568041

bundles/io/org.eclipse.smarthome.io.net/META-INF/MANIFEST.MF
```diff
@@ -10,6 +10,7 @@ Export-Package:
  org.eclipse.smarthome.io.net.exec,
  org.eclipse.smarthome.io.net.http
 Import-Package: 
+ javax.servlet.http;version="3.1.0",
```
maggu2810: Which feature of the Servlet are we using that has been introduced in 3.1.0 and is not available before?

htreu: none, the version was added by the IDE quick fix action, removed.

extensions/ui/org.eclipse.smarthome.ui.paper/src/main/java/org/eclipse/smarthome/ui/paper/internal/PaperUIApp.java
htreu: @maggu2810 thanks for the link. I was wondering about why the IDE did not complain. Although o.e.s.io.net was missing the import completely the service XML was generated and no compile error present.

maggu2810:
> @sjka & @maggu2810: Any further comments about the questions above? I would be fine with this PR for the start - I think the other requirements can be addressed later without having to break anything.

Wouldn't it be "better" to wrap the HTTP context created by the used HTTP service implementation and workaround the get resource bug? So overriding only this function but delegate the other ones?
If a specific HTTP service inject a HTTP context with an improved logic this one get's lost.
I don't know if there could be much logic involved but the factory could create the context by using the service (add as an argument) and wrap that context by delegate every method except the get resource one.

htreu:
I also thought about the delegation pattern at first but the HttpContext does only define three methods with the following specs:
- `getMimeType` to always return `null`
- `handleSecurity` to be always implementation specific
- `getResource` to behave like we did in this PR

So there is actually little room for delegation to the "original" (read library) implementation.

maggu2810:
> I also thought about the delegation pattern at first but the HttpContext does only define three methods with the following specs:
> - `getMimeType` to always return `null`
> - `handleSecurity` to be always implementation specific
> - `getResource` to behave like we did in this PR

Which spec are you referring that states that `getMimeType` have to return `null`?

See: https://osgi.org/javadoc/r4v42/org/osgi/service/http/HttpContext.html
> MIME type (e.g. text/html) of the name or null to indicate that the Http Service should determine the MIME type itself.

So, `null` is possible, but not a "must".

If the used HTTP service provides a HttpContext that `handleSecurity` method handles all the logic described in the spec, shouldn't we delegate it to the implementation the HTTP services provides? Now we have our own implementation that does perhaps not fit to the HTTP service expected one.

I would prefer to only override the method we want a specific behavior and keep / delegate the other ones. I assume we all don't know about all the HTTP service implementation and so context implementations in the wild. If a product uses and configure a specific one, this should still do its expected job / behavior for the ESH framework.

htreu:
I refer to the HttpService#getDefaultHttpContext method spec here: https://www.knopflerfish.org/releases/6.0.0/docs/javadoc/org/osgi/service/http/HttpService.html#createDefaultHttpContext

But you are right, even for `handleSecurity` it might be worth using the HttpContext created by the HttpService using a wrapper.

@kaikreuzer wdyt?

maggu2810:
> kaikreuzer requested changes

@kaikreuzer Is there still something missing or can you revoke your change request?

PR #4438, opened 2017-10-18
Some nullness fixes for Automation

adimova:
extends #4290

### DIFF
--- a/bundles/automation/org.eclipse.smarthome.automation.api/src/main/java/org/eclipse/smarthome/automation/Rule.java
+++ b/bundles/automation/org.eclipse.smarthome.automation.api/src/main/java/org/eclipse/smarthome/automation/Rule.java
@@ -8,32 +8,41 @@
 package org.eclipse.smarthome.automation;
 
 import java.util.ArrayList;
-import java.util.Collection;
 import java.util.Collections;
+import java.util.HashSet;
 import java.util.List;
 import java.util.Set;
+import java.util.UUID;
 
+import org.eclipse.jdt.annotation.NonNullByDefault;
+import org.eclipse.jdt.annotation.Nullable;
+import org.eclipse.smarthome.automation.template.RuleTemplate;
 import org.eclipse.smarthome.config.core.ConfigDescriptionParameter;
 import org.eclipse.smarthome.config.core.Configuration;
 import org.eclipse.smarthome.core.common.registry.Identifiable;
 
 /**
- * Rule is built from {@link Module}s and consists of three sections:
- * ON/IF/THEN.
+ * An automation Rule is built from {@link Module}s and consists of three parts:
  * <ul>
- * <li>ON - contains {@link Trigger} modules. The triggers defines what fires the the Rule.
- * <li>IF - contains {@link Condition} modules which determinate if the Rule is satisfied or not. When all conditions
- * are satisfied then the Rule can proceed with execution of THEN part.
- * <li>THEN - contains actions which have to be executed by the {@link Rule}
+ * <li><b>Triggers:</b> a list of {@link Trigger} modules. Each {@link Trigger} from this list
+ * can start the evaluation of the Rule. A Rule with an empty list of {@link Trigger}s can
+ * only be triggered through the {@link RuleRegistry#runNow(String, boolean, java.util.Map)} method,
+ * or directly executed with the {@link RuleRegistry#runNow(String)} method.
+ * <li><b>Conditions:</b> a list of {@link Condition} modules. When a Rule is triggered, the
+ * evaluation of the Rule's {@link Condition}s will determine if the Rule will be executed.
+ * A Rule will be executed only when all it's {@link Condition}s are satisfied. If the {@link Condition}s
+ * list is empty, the Rule is considered satisfied.
+ * <li><b>Actions:</b> a list of {@link Action} modules. These modules determine the actions that
+ * will be performed when a Rule is executed.
  * </ul>
- * Rules can have <code>tags</code> - non-hierarchical keywords or terms for
- * describing them. They help for classifying the items and allow them to be
- * found.
+ * Additionally, Rules can have <code><b>tags</b></code> - non-hierarchical keywords or terms for describing them.
+ * They can help the user to classify or label the Rules, and to filter and search them.
  *
  * @author Yordan Mihaylov - Initial Contribution
  * @author Ana Dimova - Initial Contribution
  * @author Vasil Ilchev - Initial Contribution
  */
+@NonNullByDefault
 public class Rule implements Identifiable<String> {
 
     protected List<Trigger> triggers;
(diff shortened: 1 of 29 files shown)


bundles/automation/org.eclipse.smarthome.automation.api/src/main/java/org/eclipse/smarthome/automation/Rule.java
```diff
@@ -76,19 +85,20 @@ public Rule(String uid) {
...
+        this.triggers = triggers == null ? Collections.<Trigger> emptyList() : triggers;
+        this.conditions = conditions == null ? Collections.<Condition> emptyList() : conditions;
+        this.actions = actions == null ? Collections.<Action> emptyList() : actions;
+        this.configDescriptions = configDescriptions == null ? new ArrayList<ConfigDescriptionParameter>()
+                : configDescriptions;
+        this.configuration = configurations == null ? new Configuration() : configurations;
         setTemplateUID(templateUID);
-        setVisibility(visibility);
+        this.visibility = visibility == null ? Visibility.VISIBLE : visibility;
+        tags = Collections.<String> emptySet();
```
maggu2810: Shouldn't `Collections.emptySet()` already do its job? The type could be identified implicit.

bundles/automation/org.eclipse.smarthome.automation.api/src/main/java/org/eclipse/smarthome/automation/Rule.java
```diff
@@ -146,7 +156,7 @@ public void setName(String ruleName) throws IllegalStateException {
      * @return a {@link Set} of tags
      */
     public Set<String> getTags() {
-        return tags = tags != null ? tags : Collections.<String> emptySet();
+        return tags;
```
maggu2810: Do you return the internal set by intention, so it could be modified by the caller?
Shouldn't we return a copy of the current set, so a modification does not result in a changed tag set without calling setTag.
Another point is if two consumers call getTags() and one iterate over the returned reference while the other one modifies the returned reference the first one could run into a CME.

Also the set function should create a new set with the content of the supplied argument instead of using the given reference that could be changed by the caller and so the tags could be modified without calling setTags.

bundles/automation/org.eclipse.smarthome.automation.api/src/main/java/org/eclipse/smarthome/automation/Rule.java
```diff
@@ -321,10 +312,10 @@ public Module getModule(String moduleId) {
      * @return list of modules of defined type or all modules when the type is not
      *         specified.
      */
-    @SuppressWarnings("unchecked")
+    @SuppressWarnings({ "unchecked", "null" })
```
maggu2810: Don't initialize result with null, but use `final List<T> result;` first.
Then add an else branch
```Java
else {
    result = Collections.emptyList();
}
```

you could then use a simple `return result;` at the end.

Or miss I something?
Why do you need to suppress the "null" warning?

bundles/automation/org.eclipse.smarthome.automation.parser.gson/src/main/java/org/eclipse/smarthome/automation/parser/gson/internal/RuleGSONParser.java
```diff
@@ -33,19 +33,20 @@
...
                 } else {
                     Rule rule = gson.fromJson(jr, Rule.class);
-                    Set<Rule> rules = new HashSet<Rule>();
                     rules.add(rule);
-                    return rules;
                 }
+                return rules;
             }
         } catch (Exception e) {
+            e.printStackTrace();
```
maggu2810: This line should be removed.

bundles/automation/org.eclipse.smarthome.automation.api/src/main/java/org/eclipse/smarthome/automation/Rule.java
adimova: > Shouldn't Collections.emptySet() already do its job? The type could be identified implicit.

Yes, you are correct. This was copy-pasted from old, java5 compatible code. I'll fix it

> I also didn't check if every null warning suppression is really necessary. Do we need all of them?

All null warning suppressions were added because of Eclipse warnings. I'll review them again

bundles/automation/org.eclipse.smarthome.automation.api/src/main/java/org/eclipse/smarthome/automation/Rule.java
adimova: OK

bundles/automation/org.eclipse.smarthome.automation.parser.gson/src/main/java/org/eclipse/smarthome/automation/parser/gson/internal/RuleGSONParser.java
adimova: OK

bundles/automation/org.eclipse.smarthome.automation.core/src/main/java/org/eclipse/smarthome/automation/core/internal/RuleRegistryImpl.java
```diff
@@ -280,20 +255,25 @@ public Rule add(Rule rule) {
...
+                    if (ruleUIDs == null) {
+                        ruleUIDs = new HashSet<String>(11);
+                        mapTemplateToRules.put(templateUID, ruleUIDs);
+                    }
+                    ruleUIDs.add(uid);
                 }
-                ruleUIDs.add(uid);
             }
+        } catch (RuntimeException re) {
+            re.printStackTrace();
```
kaikreuzer: printStackTrace?

bundles/automation/org.eclipse.smarthome.automation.api/src/main/java/org/eclipse/smarthome/automation/Rule.java
```diff
@@ -108,11 +117,11 @@ public String getUID() {
      *
      * @return unique id of this {@link Rule}
```
kaikreuzer: this javadoc does not fit to the method

bundles/automation/org.eclipse.smarthome.automation.api/src/main/java/org/eclipse/smarthome/automation/Rule.java
```diff
@@ -76,19 +86,18 @@ public Rule(String uid) {
...
-        this.uid = uid;
-        setTriggers(triggers);
-        setConditions(conditions);
-        setActions(actions);
-        setConfigurationDescriptions(configDescriptions);
-        setConfiguration(configurations);
+    public Rule(@Nullable String uid, @Nullable List<Trigger> triggers, @Nullable List<Condition> conditions,
+            @Nullable List<Action> actions, @Nullable List<ConfigDescriptionParameter> configDescriptions,
+            @Nullable Configuration configurations, @Nullable String templateUID, @Nullable Visibility visibility) {
+        this.uid = uid == null ? ID_PREFIX + UUID.randomUUID().toString() : uid;
```
kaikreuzer: I would actually suggest to leave away the ID_PREFIX as it only makes the UID even longer without providing much benefit.

bundles/automation/org.eclipse.smarthome.automation.api/src/main/java/org/eclipse/smarthome/automation/Rule.java
adimova: for the user convenience, this will be fixed in RuleUtils class

bundles/automation/org.eclipse.smarthome.automation.api/src/main/java/org/eclipse/smarthome/automation/Rule.java
adimova: ok

bundles/automation/org.eclipse.smarthome.automation.api/src/main/java/org/eclipse/smarthome/automation/Rule.java
adimova: ok

bundles/automation/org.eclipse.smarthome.automation.core/src/main/java/org/eclipse/smarthome/automation/core/internal/RuleRegistryImpl.java
adimova: ok

maggu2810:
Travis CI failed in the Thing XML tests. I assume this is not related to that changes, but I restarted a Travis CI build to hopefully get a successful build.

maggu2810:
Build succeeded.
@kaikreuzer Have all your requested changes been comitted? For me it seems like they have been fixed.

kaikreuzer:
Looks good. Thanks @adimova!

steve-bate:
There's a report of recently broken behavior related to the SimpleRule implementation.

https://community.openhab.org/t/jsr223-jython-script-rule-loading-error-in-build-1071/35601

It seems that something in the SmartHome code is trying to use the default constructor for SimpleRule and that no longer exists after the explicit constructor was added. Can somebody take a look?

PR #4442, opened 2017-10-19
Consider group assignments

### LINKED ISSUE #4434: Item group assignment does not take effect without service restart
System: openhab stable from PPA on Ubuntu server 16.04 LTS

I noticed that when I change the group assignment of an item in the .items-file, the change does not take effect without a service restart. openhab.log indicates that the file change was detected and the model was updated. Other item changes do take effect without service restart.

sjka:
...when comparing items.

fixes #4434


sjka:
@kaikreuzer I'm not sure if there is any reason why the group names were not considered. If there was, then we could easily fix it also here:
 
https://github.com/eclipse/smarthome/blob/f4247ba748e5271beccde8fa63d63b2be4df6c12/bundles/model/org.eclipse.smarthome.model.item/src/org/eclipse/smarthome/model/item/internal/GenericItemProvider.java#L364

PR #4439, opened 2017-10-18
Improving scale transformation service

clinique:
to ensure that they match in the same order than the scale definition file as discussed
in PR openhab/openhab-docs#486

Signed-off-by: Gaël L'hopital <glhopital@gmail.com>

PR #4443, opened 2017-10-19
Improved the Scale Transformation article.

Confectrician:
This PR is moved from openHAB documentation.

Main discussion took part in:
https://github.com/openhab/openhab-docs/pull/486

_We already discussed there about a more handy example in the future,
but there are no details yet, so i wanted to finish this reworking on the current state._

For the record:

Related: #4439

Closes openhab/openhab-docs#486
Closes openhab/openhab-docs#484

_Since this is my first PR on a eclipse repo, feel free to comment if there is something missing.
I have done all the steps from the contribution guidelines.
(Eclipse Account, GitHub Settings in there, Contributor Agreement)
Most things should be ok this way._

Signed-off-by:  Jerome Luckenbach <github@luckenba.ch> (github: @Confectrician)

### DIFF
--- a/extensions/transform/org.eclipse.smarthome.transform.scale/README.md
+++ b/extensions/transform/org.eclipse.smarthome.transform.scale/README.md
@@ -10,7 +10,7 @@ Range expressions always contain two parts.
 The range to scale on, which is located left from the equality sign and the corresponding output string on the right of it.
 A range consists of two bounds. Both are optional, the range is then open. Both bounds can be inclusive or exclusive.
 
-| Scale Expression | Returns XYZ when the given Value is                        |
+| Scale Expression | Returns XYZ when the given value is                        |
 |------------------|------------------------------------------------------------|
 | `[12..23]=XYZ`   | `between (or equal to) 12 and 23`                          |
 | `]12..23[=XYZ`   | `between 12 and 23 (12 and 23 are excluded in this case.)` |
@@ -31,7 +31,7 @@ Example item:
 Number Uv_Sensor_Level "UV Level [SCALE(uvindex.scale):%s]"
 ```
 
-Referenced scale file `uvindex.scale` in the `transform` folder::
+Referenced scale file `uvindex.scale` in the `transform` folder:
 
 ```python
 [..3]=1
@@ -42,7 +42,7 @@ Referenced scale file `uvindex.scale` in the `transform` folder::
 ```
 
 Each value the item receives, will be categorized to one of the five given ranges.
-Values **lower than or equal to 3** are catched with `[..3]=1`.
+Values **lower than or equal to 3** are caught with `[..3]=1`.
 Greater values are catched in ranges with 2 values as criteria.
 The only condition here is that the received value has to be lower than or equal to `100` in our example, since we haven't defined other cases yet.
 If **none** of the configured conditions matches the given value, the response will be empty.
--- a/extensions/transform/org.eclipse.smarthome.transform.scale/README.md
+++ b/extensions/transform/org.eclipse.smarthome.transform.scale/README.md
@@ -1,6 +1,6 @@
 # Scale Transformation Service
 
-Transform a given input by matching it to ranges.
+Transform a given input by matching it to specified ranges.
 The input string must be in numerical format.
 
 The file is expected to exist in the `transform` configuration directory and its ending has to be `.scale`.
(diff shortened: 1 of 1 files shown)


extensions/transform/org.eclipse.smarthome.transform.scale/README.md
```diff
@@ -1,16 +1,51 @@
 # Scale Transformation Service
 
-Transform the input by matching it between limits of ranges in a scale file.  The input string must be in numerical format.
+Transform a given input by matching it between limits of ranges.
```
sjka: I know, you just broke the lines here, but would you agree that "Transform a given input by matching it to ranges" is less confusing?

extensions/transform/org.eclipse.smarthome.transform.scale/README.md
```diff
@@ -1,16 +1,51 @@
...
+[..3]=1
+]3..6]=2
+]6..8]=3
+]8..10]=4
+[10..100]=5
+```
+
+Each value the item receives, will be categorized to one of the five given ranges.
+Values **lower than or equal to 3** are caught with `[..3]=1`.
+Greater values are catched in ranges with 2 values as criteria.
```
sjka: ~~are catched~~ -> "are **caught**" (or maybe rather "are **matched**")

extensions/transform/org.eclipse.smarthome.transform.scale/README.md
```diff
@@ -1,16 +1,51 @@
...
+The only condition here is that the received value has to be lower than or equal to `100` in our example, since we haven't defined other cases yet.
```
sjka: lower or equal ~~to~~ than

extensions/transform/org.eclipse.smarthome.transform.scale/README.md
Confectrician: Totally agree.

Wdyt about:
"Transform a given input by matching it to __specific__ ranges."
or
"Transform a given input by matching it to __specified__ ranges."

Just to make directly clear, that there are no fixed ranges somewhere in esh that are used for that.

extensions/transform/org.eclipse.smarthome.transform.scale/README.md
Confectrician: I think "matched" it will be, when I compare the options.
That sounds/reads the best. (My opinion)

extensions/transform/org.eclipse.smarthome.transform.scale/README.md
sjka: 👍  "specified"

Confectrician:
>I'm even thinking if it would make sense to mention the term "discretization" somewhere, although I'm not exactly sure whether it mathematically is fully correct or not.

That would be something for an opener sentence.
I'll look if i can create a useful sentence for that.

>One additional thought: Do you think it would make sense to explain the "inclusive" and "exclusive" bracket notations a little (for the less savvy users)? I'm not sure if that becomes clear if somebody never heard of that...

Good point. I have thought about that too already.
We discussed in the openHAB PR already about a better understandable example.
I think, this should be combined then.
The explanation on its own may be confusing too.
I would prefer an updated example with a matching explanation on a later point.

Wdyt?

sjka:
> I would prefer an updated example with a matching explanation on a later point.

Sure, there is no pressure to do it here in this PR. This is a huge improvement as it is now!

[For the record: The automatic IP validation fails because of the edit directly within Github. I manually verified that your ECA is valid, so that's fine.]

Confectrician:
Hm not sure, but a first try:

"The Scale Transformation Service is a an easy to handle tool, that can help you with the discretization of number inputs."

second sentence would then be changed to something like this:
"It transforms a given input by matching it to specified ranges."

sjka:
Yes, for me this totally makes sense. I like it!
From my point of view we you may update the PR accordingly (just [drop the comma](https://www.usingenglish.com/forum/threads/72534-comma-before-quot-that-quot)) and then it's ready to be merged.

Confectrician:
Here we go. Update is done.

One off topic thing for my understanding:
> [For the record: The automatic IP validation fails because of the edit directly within Github. I manually verified that your ECA is valid, so that's fine.]

Did this fail due to the missing sign off within the PR message?
Just as a reminder for me to do it better next time. :)

sjka:
Thanks!

> Did this fail due to the missing sign off within the PR message?

No, they were fine (but yes, a missing sign-off line also makes it fail). The problem is that github's web editor created commits with itself as the committer and not you. So the only way around it would be not using the web editor. However, doing the lookup manually really is nothing more than two clicks and a c'n'p of your email address. I'm very happy to do that manually if that helps setting the barrier as low as possible for improving our documentation.

Issue #4400, opened 2017-10-11
New Rule Engine causes deadlock in concierge


sjka:
see https://github.com/eclipse/smarthome-packaging-sample/issues/13#issuecomment-335803723 for details.

To me this looks like a bug in concierge (and therefore I opened an issue there). Nevertheless, this brings me back to the question whether it is really a good idea to mess with bundle resolution within ESH, or if that rather should be left to the solution built on top... I personally consider this a design flaw!

kaikreuzer:
I agree - bundles should not interfere with the lifecycle of other bundles; this should be changed in the automation module.

PR #4451, opened 2017-10-23
avoid interfering with bundle lifecycle

kaikreuzer:
We must not interfere with the bundle lifecycle, but rather only inspect resolved bundles.

The only exception might be to trigger a re-resolution of host bundles, if a fragment is added later; but if we need this, this should only be done if we really identified this situation.
I am not too sure if we really require it as fragments are rather ugly here and consumers should be encouraged to rather put automation resources within bundles. In any case, I would prefer to have this fix merged quickly and have fragment support dealt with as an option later on, if required.

fixes https://github.com/eclipse/smarthome/issues/4400


maggu2810:
@kaikreuzer Automation integration tests failed

```text
Failed tests:
 HostFragmentSupportTest.asserting that the installation of the host-fragment provides the resources correctly:135->OSGiTest.waitForAssert:191->OSGiTest.waitForAssert:220
Expected: is <4>
 but: was <2>
```

kaikreuzer:
updated PR.

aluedt:
Happy to hear this fixes #4400. As we avoid using snapshot versions for my current project I'd be interested in when this will make it into a release or the next beta version? Is there a roadmap for the 0.9.0 release or beta 6 including this fix?

adimova:
OK for me too

kaikreuzer:
> The only exception might be to trigger a re-resolution of host bundles, if a fragment is added later

I have discussed this with @sjka and we came to the conclusion (or rather: he convinced me) that even this is wrong.
If fragments are installed after startup, it is up to the container management of the solution to decide whether this should trigger a re-resolution of host bundles or not - this decision is a general one and not at all specific to the automation module here, so it must not be decided here in the application logic. It can be rather very counter-productive and error-prone if this is done here AND the solution's container management does it as well. I have therefore completely removed the unit test about fragment resource resolution and think that this is the only correct solution.

> Is there a roadmap for the 0.9.0 release or beta 6 including this fix?

@aluedt There is not yet a fixed date for the 0.9.0 release, but I want to start the planning/discussion around this soon. Beta releases we usually only do upon request - as I would also like to update the sample packaging to the latest code base soon, I'd be ok to do a beta 6 within the next day as soon as this PR is merged (and no other critical issues are open).

kaikreuzer:
Thanks @adimova!

@maggu2810 & @sjka It seems @adimova is not yet commissioned as a committer on this repo by @eclipsewebmaster, so if one of you could do the merge, it would be awesome :-)

Issue #4446, opened 2017-10-20
Firmware Update: Model-specific firmwares


hsudbrock:
### The issue

Currently, in Eclipse SmartHome, a firmware is always related to exactly one thing type. Moreover, the `o.e.sh.core.thing.firmware.FirmwareUpdateService` determines the latest firmware available for a thing based solely on the thing's type.

Consider two devices that, from the perspective of Eclipse SmartHome, provide the same functionality and are controlled using the same protocols, but that are actually different device models and need different firmware versions. (Example: Two different models of a roller shutter from the same company.) For the binding, it makes sense to use one thing type and thing handler for both devices (to avoid the additional complexity of having two thing types and handlers that provide the same function).

In this case, the current way of handling firmware update in Eclipse SmartHome is not sufficient, since firmwares are related to exactly one thing type. To obtain the correct firmware in the described case, one needs to identify the latest firmware based on thing type and the thing's model ID.

### Solution suggestions

I suggest to extend the Eclipse SmartHome firmware code such that firmwares can be for a specific model (in addition to being for a specific thing type).  I see two ways how to consider a thing's model in the Eclipse SmartHome firmware code:

1. Adding a boolean property `isModelSpecific` to `o.e.sh.core.thing.binding.firmware.Firmware` such that a firmware can indicate that it must only be used for things with a corresponding model ID.
2. Extending the `o.e.sh.core.thing.firmware.FirmwareProvider` interface such that firmware providers can (optionally) provide firmwares based on thing type and model ID.  (Could be done in a backwards compatible way by providing default implementations for the new interface methods.)

For the first approach, the `o.e.sh.core.thing.firmware.FirmwareUpdateService` has to be changed such that it only considers firmwares that are suitable for a given thing (based on the new `isModelSpecific` flag, the firmware's `model`, and the thing's `modelId` property).

For the second approach, the `o.e.sh.core.thing.firmware.FirmwareRegistry`has to be extended to provide firmwares based on thing type and model ID as well. Moreover, the `o.e.sh.core.thing.firmware.FirmwareUpdateService` has to be adapted to use the new methods of the `FirmwareRegistry` when finding suitable firmwares for a thing.

My preference is the first approach, since it does not require to adapt existing interfaces.

I would be happy about feedback on these suggestions; I am also willing to implement this improvement :smile:

### Credit
Credit for suggesting such an improvement in Eclipse SmartHome as well as initial ideas for implementation goes to @tomhoefer.

[...]

PR #4454, opened 2017-10-24
Support model-restricted firmwares in firmware update

hsudbrock:
In https://github.com/eclipse/smarthome/issues/4446, I suggest to extend the Eclipse SmartHome firmware code with the possibility to have firmwares that are not suitable for all things of a given thing type, but only for things with a specific model ID. This PR is a proposal for an implementation of this feature as follows:

* Extend the class `o.e.sh.core.thing.binding.firmware.Firmware` with the boolean flag `modelRestricted`, so that firmware providers can specify that a firmware is restricted to the model specified in the Firmware's `model` field. Add a utility method (`isSuitableFor(Thing)`) to `Firmware` to check whether the firmware is suitable for a given thing (using the new flag, the firmware's model, and the thing's modelId).
* Adapt the class `o.e.sh.core.thing.firmware.FirmwareUpdateService` so that it (a) only considers suitable firmwares when computing the firmware status of a thing, and (b) does not allow firmware updates on a thing for model-restriced firmwares that do not match the thing's modelID.

Moreover, this PR contains:
* Extend the `FirmwareDTO` with the new `modelRestricted` field.
* Migration of the `FirmwareTest` from Groovy to Java.

Related Issue: https://github.com/eclipse/smarthome/issues/4446
Signed-off-by: Henning Sudbrock <henning.sudbrock@telekom.de>


eclipsewebmaster:
Issue tracker reference:
https://bugs.eclipse.org/bugs/show_bug.cgi?id=4446

bundles/core/org.eclipse.smarthome.core.thing/src/main/java/org/eclipse/smarthome/core/thing/binding/firmware/Firmware.java
```diff
@@ -285,11 +299,27 @@ public boolean isPrerequisiteVersion(String firmwareVersion) {
...
+        return hasSameThingType(thing) && hasRequiredModel(thing);
+    }
+
     @Override
     public int compareTo(Firmware firmware) {
         return -internalVersion.compare(new Version(firmware.getVersion()));
     }
 
+    private boolean hasSameThingType(Thing thing) {
+        return Objects.equal(this.getUID().getThingTypeUID(), thing.getThingTypeUID());
```
afuechsel: Note for Java 7 and later: This method should be treated as deprecated; use Objects.equals(java.lang.Object, java.lang.Object) instead. It's now plain Java. Side note: can any of these arguments be null? I doubt, so why not just using: 
```` 
this.getUID().getThingTypeUID().equals(thing.getThingTypeUID())
````

File: bundles/core/org.eclipse.smarthome.core.thing/src/main/java/org/eclipse/smarthome/core/thing/binding/firmware/Firmware.java [hunk 2]
```diff
@@ -285,11 +299,27 @@ public boolean isPrerequisiteVersion(String firmwareVersion) {
...
+    }
+
+    private boolean hasRequiredModel(Thing thing) {
+        if (isModelRestricted()) {
+            return Objects.equal(this.getModel(), thing.getProperties().get(PROPERTY_MODEL_ID));
```
afuechsel: see above

bundles/core/org.eclipse.smarthome.core.thing/src/main/java/org/eclipse/smarthome/core/thing/firmware/FirmwareUpdateService.java
```diff
@@ -327,6 +325,11 @@ private ProgressCallbackImpl getProgressCallback(ThingUID thingUID) {
         return progressCallbackMap.get(thingUID);
     }
 
+    private Firmware getLatestSuitableFirmware(Thing thing) {
+        return firmwareRegistry.getFirmwares(thing.getThingTypeUID()).stream()
```
afuechsel: Question: does getFirmwares return a sorted list, so that findFirst is taking the correct one (latest one)?

bundles/core/org.eclipse.smarthome.core.thing/src/main/java/org/eclipse/smarthome/core/thing/binding/firmware/Firmware.java
hsudbrock: Thanks for the hint to the `java.util.Objects`, I will change this.

You are right, the UID of the firmware and the thing type UID in this UID cannot be null. In my opinion, the code using `Objects.equals(...)` is a little easier to read than the `this.getUID().getThingTypeUID().equals(...)` (since then a reader can see right at the beginning that equality is checked, instead of seeing the equals call in the middle of the line). Do you have a reason why this should be changed anyway?

bundles/core/org.eclipse.smarthome.core.thing/src/main/java/org/eclipse/smarthome/core/thing/binding/firmware/Firmware.java
hsudbrock: Here, in contrast to above, the two compared objects could both be null. Hence, `Objects.equals()` is, I think, the right choice. (Of course using the `java.util.Objects` class as suggested above.)

bundles/core/org.eclipse.smarthome.core.thing/src/main/java/org/eclipse/smarthome/core/thing/firmware/FirmwareUpdateService.java
hsudbrock: Yes, it does (not a list, but a collection), cf. the Javadoc of the method: _The collection is sorted in descending order, i.e. the latest firmware will be the first element in the collection._

sjka:
Thanks!

PR #4455, opened 2017-10-24
Improve DMX binding

J-N-K:
refactoring: rename class Channel to DmxChannel to avoid problems with  ESH-class Channel
add tests for FadeAction
make Chaser turn to off after completed with infinite holdTime (-1)

### DIFF
--- a/extensions/binding/org.eclipse.smarthome.binding.dmx/src/main/java/org/eclipse/smarthome/binding/dmx/internal/action/ActionState.java
+++ b/extensions/binding/org.eclipse.smarthome.binding.dmx/src/main/java/org/eclipse/smarthome/binding/dmx/internal/action/ActionState.java
@@ -18,8 +18,8 @@ package org.eclipse.smarthome.binding.dmx.internal.action;
  * @author Jan N. Klug
  */
 public enum ActionState {
-    waiting,
-    running,
-    completed,
-    completedfinal
+    WAITING,
+    RUNNING,
+    COMPLETED,
+    COMPLETEDFINAL
 }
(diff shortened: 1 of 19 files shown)


extensions/binding/org.eclipse.smarthome.binding.dmx/src/main/java/org/eclipse/smarthome/binding/dmx/internal/action/ActionState.java
```diff
@@ -0,0 +1,25 @@
...
+ *
+ * waiting : not started yet
+ * running : action is running
+ * completed : action has completed, proceed to next action
+ * completedfinal : action has completed, hold here
+ *
+ * @author Jan N. Klug
+ */
+public enum ActionState {
+    waiting,
```
htreu: enum values should be all upper case.

extensions/binding/org.eclipse.smarthome.binding.dmx/src/main/java/org/eclipse/smarthome/binding/dmx/internal/multiverse/DmxChannel.java
```diff
@@ -253,8 +254,10 @@ public synchronized Integer getNewHiResValue(long calculationTime) {
         if (hasRunningActions()) {
             BaseAction action = actions.get(0);
             value = action.getNewValue(this, calculationTime);
-            if (action.isCompleted()) {
+            if (ActionState.completed.equals(action.getState())) {
```
htreu: since `ActionState` are enum values this should be `if (action.getState() == ActionState.COMPLETED) {`. please check other comparison, too.

J-N-K:
Oops. Forgot to commit & push my changes.

sjka:
Now it seems to be complete, TravisCI likes it. 
Thanks!

PR #4469, opened 2017-10-30
[Tradfri] Use first IP address which is reachable

ivivanov-bg:
Get first IP address which is reachable in the network

### DIFF
--- a/extensions/binding/org.eclipse.smarthome.binding.tradfri/src/main/java/org/eclipse/smarthome/binding/tradfri/internal/discovery/TradfriDiscoveryParticipant.java
+++ b/extensions/binding/org.eclipse.smarthome.binding.tradfri/src/main/java/org/eclipse/smarthome/binding/tradfri/internal/discovery/TradfriDiscoveryParticipant.java
@@ -7,6 +7,8 @@
  */
 package org.eclipse.smarthome.binding.tradfri.internal.discovery;
 
+import java.io.IOException;
+import java.net.InetAddress;
 import java.util.Collections;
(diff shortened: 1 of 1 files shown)


ivivanov-bg:
The simplest way to reproduce the issue is:
1. Reboot the bridge with the LAN cable un-plugged.
   -> Bridge assigns to itself a local link address (in the rangle169.254.0.1 to 169.254.255.254)
2. Plug the LAN cable 
  ->DHCP server assigns an IP

3. mDNS result contains the following addresses:
  -> /<local-link address>, /<LAN ip address>, /<IPv6 address>

And with the current version of the discovery participant the local link address is used (which is usually not accessible by other devices)


I suppose this is bug in the Tradfri Gateway firmware but I also think that the binding is the place where those bugs have to be workarounded (if possible) until they are fixed.
(The other option is to have binding which makes the device unusable)

kaikreuzer:
I tried really hard, but I cannot reproduce this behavior.
Can you confirm that you are still seeing it with the latest firmware?

ivivanov-bg:
I can't check now because I don't have gateway currently.
But it was very easy to reproduce: 
1. Power on the gateway with no LAN cable plugged.
2. After it boots (may be in 10 to 20 secs) plug the lan cable
3. Start discovery

kaikreuzer:
I did exactly these steps and it worked smoothly 3 times in a row.
I'd suggest to close this for now, unless anyone is still able to reproduce this issue.

ivivanov-bg:
OK - close it and will keep an eye on this

PR #4471, opened 2017-10-30
Move UPnP & mDNS Discovery Services to separate bundles

### LINKED ISSUE #4470: UpnpDiscoveryService will fail when org.jupnp.registry package is not available
We saw a lifecycle issue in bundle `org.eclipse.smarthome.config.discovery`.

The bundle declares an `import-package: org.jupnp.registry;resolution:=optional`. So bundle is fine to live without any jUPnP classes in runtime.

On the other side there is a DS component defined like that:
```
@Component(immediate = true, service = DiscoveryService.class, configurationPid = "discovery.upnp")
public class UpnpDiscoveryService extends AbstractDiscoveryService implements RegistryListener {
```

This means that this component will be immediately activated. But when as this point in time the optional package is missing, the DS comp can not be activated as the interface to implement is missing. Which will result in a `ClassNotFoundException` or `NoClassDefFoundError`. Which will result in an Unsatisfied DS component, and furthermore other DS comps unsatisfied.

Thats what we observed when updating bundles in runtime.

We propose to move this UPnP specific logic out of the generic `o.e.sh.config.discovery` bundle to the `org.eclipse.smarthome.io.transport.upnp` bundle where there can be easily mandatory references to jUPnP classes/interfaces.

What do you think?

sjka:
fixes #4470


sjka:
considering how many bindings this breaks, I'm not sure sure if it really is a good idea. 

The alternative would be making o.e.sh.config.discovery depend on jupnp mandatorily, which also sucks. Any better proposals?

JochenHiller:
Indeed: was a bad decision to put the class `UpnpDiscoveryParticipant ` into package `org.eclipse.smarthome.config.discovery` which is somehow unrelated to UPnP.
Even extracting the whole stuff into its own bundle does not relieve this API breaking change.
And split-packages is no nice solution.

No better idea at the moment. Interesting question about API evolvement...

kaikreuzer:
I also do not really see a way around breaking the API.
But if we do so (and have to adapt so many bindings), we should be sure that we do it the right way.
Imho io.transport is not the correct place either - it would draw in a dependency to the Thing API in those bundles, although they are generally considered to be "lower-level" bundles that are used by bindings (and thus Things). The best place would seem to me to `config.discovery.upnp`and `config.discovery.mdns` - introducing new bundles just for an interface and an internal service is not too nice, but probably the cleanest approach.

sjka:
Having dedicated bundles for them would in fact allow us to keep the old interfaces for a short while and make the corresponding discovery service implementations listen to both (without running into cyclic dependencies). With this we could allow a "grace period" and therefore avoid the hard breaking of the APIs. I will update this PR accordingly.

kaikreuzer:
Thanks, looks good!
I am just missing the new bundles being added in the feature definitions.

maggu2810:
Moving the discovery for mDNS and JUPnP to separate bundles make sense.
Do we really need a "grace period" and to keep the old one for a while?

JochenHiller:
> Do we really need a "grace period" and to keep the old one for a while?
@maggu2810 At least it is important for commercial use of that APIs, including us. Otherwise we would need a synchronized delivery of artifacts from ESH, and all bindings using these APIs. Which makes life very hard. So we appreciate the idea of having a grace period. We try to follow the API change ASAP, so we can get rid of the interim solution in near future.

PR #4472, opened 2017-10-31
Extracted JUPnP optional dependencies to a failsafe dedicated component

amitjoy:
Proposal for #4470 

Signed-off-by: Amit Kumar Mondal <admin@amitinside.com>


sjka:
This indeed would solve the aspect nicely that the service fails miserably if no UPnP classes are available!

However, this does not address the fact they they will never become available and therefore your users won't be able to find anything via UPnP. The background is that per spec in OSGi bundles don't need to get re-resolved once an optional dependency can be satisfied later on. For that reason I'd rather like to follow the approach Kai suggested [here](https://github.com/eclipse/smarthome/pull/4471#issuecomment-340609675) in order to get rid of the optional dependencies completely.

PR #4475, opened 2017-11-01
Sonos: no update of linein channel when unexpected

lolodomo:
The linein channel must not be updated for thing types that don't own this channel.


triller-telekom:
LGTM

PR #4480, opened 2017-11-01
Sonos: cover art download in a separate thread

### LINKED ISSUE #4447: Sonos: player state undefined due to state TRANSITIONING
It looks like Sonos stopped to send the state PLAYING after the state TRANSITIONING in certain conditions.
I can reproduce that when switching to the next song of my Spotify playlist with the Sonos application. The result is that the player controls are unset due to the state value TRANSITIONING.
By the way, I think we should consider TRANSITIONING as PLAYING for the player and enable the play button.

lolodomo:
Fixes #4447

### DIFF
--- a/extensions/binding/org.eclipse.smarthome.binding.sonos/src/main/java/org/eclipse/smarthome/binding/sonos/internal/handler/ZonePlayerHandler.java
+++ b/extensions/binding/org.eclipse.smarthome.binding.sonos/src/main/java/org/eclipse/smarthome/binding/sonos/internal/handler/ZonePlayerHandler.java
@@ -683,10 +683,15 @@ public class ZonePlayerHandler extends BaseThingHandler implements UpnpIOPartici
             case CURRENTALBUMART:
                 url = getAlbumArtUrl();
                 if (url != null) {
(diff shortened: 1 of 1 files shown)


extensions/binding/org.eclipse.smarthome.binding.sonos/src/main/java/org/eclipse/smarthome/binding/sonos/internal/handler/ZonePlayerHandler.java
```diff
@@ -683,10 +683,18 @@ protected void updateChannel(String channeldD) {
...
                 url = getAlbumArtUrl();
                 if (url != null) {
-                    RawType image = HttpUtil.downloadImage(url, true, 500000);
-                    if (image != null) {
-                        newState = image;
-                    }
+                    // We run the update of the covert art channel in a different thread
+                    // because it can take time
+                    newState = null;
+                    scheduler.schedule(new Runnable() {
```
kaikreuzer: 1. please use "submit" instead of schedule
2. please use Java8 syntax and not anonymous inner classes

lolodomo:
Change done.

lolodomo:
In this case, yes ;)

PR #4483, opened 2017-11-02
[core] Adjustments to ExpiringCacheMap

martinvw:
I wanted the `ExpiringCacheMap` to be more like a Java 8 `Map`, ideally it would have implemented the `Map` interface but our `get` and `put` are not symmetrical, for very good reasons. 

Given that checking whether an entry exists and adding it, if it doesn't is very common I added it, I also changed the constructor to accept a duration to make more readable code possible.

Example:

```java
ExpiringCacheMap<String, RawData> cache = new ExpiringCacheMap<>(Duration.ofMinutes(15));

...

RawType image = cache.putIfAbsentAndGet(url, () -> {
    logger.debug("Try to download the content of URL {}", url);
    return HttpUtil.downloadImage(url);
});

``` 

@cweitkamp wdyt?

PS: I did not name if `putIfAbsent` because it behaves differently
PS2: could we already use proper java 8 here?

### DIFF
--- a/bundles/core/org.eclipse.smarthome.core.test/src/test/java/org/eclipse/smarthome/core/cache/ExpiringCacheMapTest.java
+++ b/bundles/core/org.eclipse.smarthome.core.test/src/test/java/org/eclipse/smarthome/core/cache/ExpiringCacheMapTest.java
@@ -13,188 +13,197 @@ import java.util.Collection;
 import java.util.LinkedHashSet;
 import java.util.LinkedList;
 import java.util.Set;
+import java.util.concurrent.TimeUnit;
 import java.util.function.Supplier;
 
 import org.apache.commons.lang.RandomStringUtils;
 import org.junit.Test;
-import org.slf4j.Logger;
-import org.slf4j.LoggerFactory;
+import org.junit.Before;
 
 /**
  * Test class for the {@link ExpiringCacheMap} class.
  *
  * @author Christoph Weitkamp - Initial contribution.
+ * @author Martin van Wingerden - Added tests for putIfAbsentAndGet
  */
 public class ExpiringCacheMapTest {
+    private static final long CACHE_EXPIRY = TimeUnit.SECONDS.toMillis(2);
 
-    private final Logger logger = LoggerFactory.getLogger(ExpiringCacheMapTest.class);
+    private static final String RESPONSE_1 = "ACTION 1";
+    private static final String RESPONSE_2 = "ACTION 2";
 
-    public static final long CACHE_EXPIRY = 2 * 1000; // 2s
+    private static final Supplier<String> CACHE_ACTION = () -> RandomStringUtils.random(8);
+    private static final Supplier<String> PREDICTABLE_CACHE_ACTION_1 = () -> RESPONSE_1;
+    private static final Supplier<String> PREDICTABLE_CACHE_ACTION_2 = () -> RESPONSE_2;
 
-    public static final Supplier<String> CACHE_ACTION = () -> RandomStringUtils.random(8);
+    private static final String FIRST_TEST_KEY = "FIRST_TEST_KEY";
+    private static final String SECOND_TEST_KEY = "SECOND_TEST_KEY";
 
-    public static final String FIRST_TEST_KEY = "FIRST_TEST_KEY";
-    public static final String SECOND_TEST_KEY = "SECOND_TEST_KEY";
-
+    private ExpiringCacheMap<String, String> subject;
+    
+    @Before
+    public void setUp() {
+        subject = new ExpiringCacheMap<>(CACHE_EXPIRY);
+    }
+    
+    
     @Test(expected = IllegalArgumentException.class)
-    public void testAddIllegalArgumentException1() throws IllegalArgumentException {
-        final ExpiringCacheMap<String, String> cache = new ExpiringCacheMap<>(CACHE_EXPIRY);
-
-        final Supplier<String> action = null;
-        cache.put(FIRST_TEST_KEY, action);
+    public void testPutIllegalArgumentException1() throws IllegalArgumentException {
+        subject.put(FIRST_TEST_KEY, (Supplier<String>) null);
     }
 
     @Test(expected = IllegalArgumentException.class)
-    public void testAddIllegalArgumentException2() throws IllegalArgumentException {
-        final ExpiringCacheMap<String, String> cache = new ExpiringCacheMap<>(CACHE_EXPIRY);
-
-        final ExpiringCache<String> item = null;
-        cache.put(FIRST_TEST_KEY, item);
+    public void testPutIllegalArgumentException2() throws IllegalArgumentException {
+        subject.put(FIRST_TEST_KEY, (ExpiringCache<String>) null);
     }
 
     @Test(expected = IllegalArgumentException.class)
-    public void testAddIllegalArgumentException3() throws IllegalArgumentException {
-        final ExpiringCacheMap<String, String> cache = new ExpiringCacheMap<>(CACHE_EXPIRY);
+    public void testPutIllegalArgumentException3() throws IllegalArgumentException {
+        subject.put(null, CACHE_ACTION);
+    }
+    
+    @Test
+    public void testPut() {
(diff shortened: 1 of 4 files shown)


cweitkamp:
As I already mentioned, I like your idea. Extending the `ExpiringCacheMap` to match the `Map` interface would be a very nice improvement.

The naming of the method `putIfAbsentAndGet` seems to be reasonable. In the initial contribution we agreed that the normal `put` methods shouldn't return values because of possible performance issues. This differs from other cache implementations (e.g. Googles Guava). By using this new method the developer has to consider on his own if he wants to accept a possible delay or not. We can think about a notice in the JavaDoc.

cweitkamp:
Just to share another idea with you - while implementing https://github.com/openhab/openhab2-addons/pull/2304 the following situation come to my mind. In my approach I use the `ExpiringCacheMap` to store downloaded image data (BLOBs). The idea behind it is obvious: We don't want to download the data every time to prevent to much overhead. This could leads to the situation that we store a lot of (possible) big data in our cache. Do you think that could be a problem? Should we think about introducing a - maybe optional, parameterized - garbage collector for the cache?

martinvw:
> Just to share another idea with you - while implementing openhab/openhab2-addons#2304 the following situation come to my mind. In my approach I use the ExpiringCacheMap to store downloaded image data (BLOBs). The idea behind it is obvious: We don't want to download the data every time to prevent to much overhead. This could lead to the situation that the memory usage of the cache could grow up to an large size. Do you think that could be a problem? Should we think about introducing a - maybe optional, parameterized - garbage collector for the cache?

I added a `SoftReference` wdyt?

I will adjust the docblock, good idea

martinvw:
Can you do a full review?

cweitkamp:
@martinvw Could you please have a look what Travis doesn't like?

martinvw:
> @martinvw Could you please have a look what Travis doesn't like?

@cweitkamp it really likes it when the tests also compile *face-slap*

bundles/core/org.eclipse.smarthome.core/src/main/java/org/eclipse/smarthome/core/cache/ExpiringCacheMap.java
```diff
@@ -72,6 +83,51 @@ public void put(K key, ExpiringCache<V> item) {
...
+     * @param key the key with which the specified value is to be associated
+     * @param action the action for the item to be associated with the specified key to retrieve/calculate the value
+     * @return the (cached) value for the specified key
+     */
+    public V putIfAbsentAndGet(K key, Supplier<V> action) {
+        return putIfAbsentAndGet(key, new ExpiringCache<>(expiry, action));
+    }
+
+    /**
+     * If the specified key is not already associated with a value, associate it with the given expiringCache.
```
cweitkamp: Could you change `expiringCache` to `{@link ExpiringCache}`?

bundles/core/org.eclipse.smarthome.core/src/main/java/org/eclipse/smarthome/core/cache/ExpiringCacheMap.java
```diff
@@ -72,6 +83,51 @@ public void put(K key, ExpiringCache<V> item) {
     }
 
     /**
+     * If the specified key is not already associated with a value, associate it with the given expiringCache.
```
cweitkamp: Could you change `expiringCache` to `{@link ExpiringCache}`?

bundles/core/org.eclipse.smarthome.core.test/src/test/java/org/eclipse/smarthome/core/cache/ExpiringCacheMapTest.java
```diff
@@ -13,188 +13,197 @@
...
+    public void testPutIfAbsentAndGet() {
+        String response = subject.putIfAbsentAndGet(FIRST_TEST_KEY, PREDICTABLE_CACHE_ACTION_1);
+
+        assertEquals(RESPONSE_1, response);
+        assertEquals(RESPONSE_1, subject.get(FIRST_TEST_KEY));
 
-        cache.put(null, CACHE_ACTION);
+        response = subject.putIfAbsentAndGet(FIRST_TEST_KEY, PREDICTABLE_CACHE_ACTION_2);
+
+        assertEquals(RESPONSE_1, response);
```
cweitkamp: I know that we as developer are responsible for the values of `RESPONSE_1` and `RESPONSE_2` because we are free to assign to them whatever we like. But this test case premises,  that those values are different. So let's add an `assertNotEquals(RESPONSE_1,RESPONSE_2)` - with an inline comment why we added it - to make sure the test will always work like expected. Wdyt?

martinvw: Yes, it crossed my mind for a while too that the test can succeed with less code ;-)

I will come up with something.

martinvw:
@sjka do you have time for a quick review of this?

bundles/core/org.eclipse.smarthome.core/src/main/java/org/eclipse/smarthome/core/cache/ExpiringCache.java
```diff
@@ -51,17 +51,17 @@ public ExpiringCache(long expiry, Supplier<V> action) {
      */
     @Nullable
     public synchronized V getValue() {
-        if (value == null || isExpired()) {
+        if (value.get() == null || isExpired()) {
```
maggu2810: Couldn't the soft reference be freed between the twice calls to `get()`?
Shouldn't we use a safe programming style and safe the current reference?

```Java
final V valueCur = value.get();
if (valueCur == null || isExpired()) {
    return refreshValue();
} else {
    retturn valueCur;
}
```

bundles/core/org.eclipse.smarthome.core/src/main/java/org/eclipse/smarthome/core/cache/ExpiringCache.java
```diff
@@ -72,9 +72,9 @@ public synchronized void invalidateValue() {
      */
     @Nullable
     public synchronized V refreshValue() {
-        value = action.get();
+        value = new SoftReference<>(action.get());
```
maggu2810: Same topic as above:

```Java
final V valueNew = action.get();
value = new SoftReference<>(valueNew);
expiresAt = System.nanoTime() + expiry;
return valueNew;
```

bundles/core/org.eclipse.smarthome.core/src/main/java/org/eclipse/smarthome/core/cache/ExpiringCache.java
martinvw: Yes we should indeed

bundles/core/org.eclipse.smarthome.core/src/main/java/org/eclipse/smarthome/core/cache/ExpiringCache.java
martinvw: same as above ;-)

martinvw:
@maggu2810 thanks for good comments, I updated the PR

maggu2810:
> the initial contribution we agreed that the normal `put` methods shouldn't return values because of possible performance issues.

Can you point me to that discussion?
Do you really inspect any performance impact in Java if a reference is returned?
The Java Map `putIfAbsent` returns the value and IMHO we should do the same and remove the call to get after `putIfAbsent` is called internally.

martinvw:
We store an action, and returning a reference that action is not useful. But call the actual action (while you might not be interested in the result at that moment) will be or at least can be a painful operation.

So we do not actually really implement the `Map` interface but we have a put of an `Action<T>` and a get which returns `T`, the actual result of that action.

cweitkamp:
> Can you point me to that discussion?

@maggu2810 Not a "real" discussion. More a proposal. See https://github.com/eclipse/smarthome/pull/3572#discussion_r120820691.

maggu2810:
Okay, I missed the difference between the item and the value of the item -- the last one could trigger a refresh...

PR #4484, opened 2017-11-02
Update README.md

MyNameIsThomas:
New device (radio) added to list


kaikreuzer:
Thanks! (merging with small patch exception)

PR #4487, opened 2017-11-02
switched to latest JUPnP 2.3.0 release

kaikreuzer:

### DIFF
--- a/features/karaf/esh-tp/src/main/feature/feature.xml
+++ b/features/karaf/esh-tp/src/main/feature/feature.xml
@@ -50,9 +50,16 @@
   </feature>
 
   <feature name="esh-tp-httpclient" version="${project.version}">
-    <capability>esh.tp;feature=httpclient;version=4.2.3</capability>
-    <bundle dependency="true">mvn:org.apache.httpcomponents/httpcore-osgi/4.2.3</bundle>
-    <bundle>mvn:org.apache.httpcomponents/httpclient-osgi/4.2.3</bundle>
+    <capability>esh.tp;feature=httpclient;version=9.3.14.v20161028</capability>
+    <bundle dependency="true">mvn:javax.servlet/javax.servlet-api/3.1.0</bundle>
+    <bundle dependency="true">mvn:org.eclipse.jetty/jetty-client/9.3.14.v20161028</bundle>
+    <bundle dependency="true">mvn:org.eclipse.jetty/jetty-http/9.3.14.v20161028</bundle>
+    <bundle dependency="true">mvn:org.eclipse.jetty/jetty-util/9.3.14.v20161028</bundle>
+    <bundle dependency="true">mvn:org.eclipse.jetty/jetty-io/9.3.14.v20161028</bundle>
+    <bundle dependency="true">mvn:org.eclipse.jetty/jetty-proxy/9.3.14.v20161028</bundle>
+    <bundle dependency="true">mvn:org.eclipse.jetty.websocket/websocket-api/9.3.14.v20161028</bundle>
+    <bundle dependency="true">mvn:org.eclipse.jetty.websocket/websocket-common/9.3.14.v20161028</bundle>
+    <bundle dependency="true">mvn:org.eclipse.jetty.websocket/websocket-client/9.3.14.v20161028</bundle>
   </feature>
 
   <feature name="esh-tp-jax-rs" version="${project.version}">
@@ -141,11 +148,11 @@
   </feature>
 
   <feature name="esh-tp-jupnp" description="UPnP/DLNA library for Java" version="${project.version}">
-    <capability>esh.tp;feature=jupnp;version=2.2.0</capability>
-
+    <capability>esh.tp;feature=jupnp;version=2.3.0</capability>
     <feature dependency="true">http</feature>
+    <feature dependency="true">scr</feature>
     <feature dependency="true">esh-tp-httpclient</feature>
-    <bundle>mvn:org.jupnp/org.jupnp/2.2.0</bundle>
+    <bundle>mvn:org.jupnp/org.jupnp/2.3.0</bundle>
   </feature>
 
   <feature name="esh-tp-jmdns" description="An implementation of multi-cast DNS in Java." version="${project.version}">
--- a/distribution/smarthome/logback_debug.xml
+++ b/distribution/smarthome/logback_debug.xml
@@ -21,7 +21,7 @@
 
 	<logger name="OSGi" level="WARN" />
 	<logger name="org.eclipse.jetty" level="WARN" />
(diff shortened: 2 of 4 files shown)


features/karaf/esh-tp/src/main/feature/feature.xml
```diff
@@ -141,11 +148,11 @@
   </feature>
 
   <feature name="esh-tp-jupnp" description="UPnP/DLNA library for Java" version="${project.version}">
-    <capability>esh.tp;feature=jupnp;version=2.2.0</capability>
-
+    <capability>esh.tp;feature=jupnp;version=2.3.0</capability>
     <feature dependency="true">http</feature>
     <feature dependency="true">esh-tp-httpclient</feature>
-    <bundle>mvn:org.jupnp/org.jupnp/2.2.0</bundle>
+    <bundle dependency="true">mvn:org.osgi/org.osgi.compendium/4.2.0</bundle>
```
kaikreuzer: @maggu2810 Do you have any advice how to correctly handle this? org.jupnp has a dependency on `org.osgi.component`. How do I express this in a neutral way? I somehow couldn't figure out how all our other OSGi dependencies are resolved, we seem to refer to the compendium spec nowhere in the feature definitions...

maggu2810: You should never depend on an API bundle.
As we don't want to force a specific implementation we already use `dependency="true"`, so the referenced dependency it is only installed if no other one satisfy the requirements.
For a the Component / DS part, you should use the SCR feature.

Does this one work for you:
```text
<feature dependency="true">scr</feature>
```

kaikreuzer: Thank you, seems to work! I have updated the PR accordingly.

features/karaf/esh-tp/src/main/feature/feature.xml
```diff
@@ -50,9 +50,16 @@
   </feature>
 
   <feature name="esh-tp-httpclient" version="${project.version}">
-    <capability>esh.tp;feature=httpclient;version=4.2.3</capability>
-    <bundle dependency="true">mvn:org.apache.httpcomponents/httpcore-osgi/4.2.3</bundle>
-    <bundle>mvn:org.apache.httpcomponents/httpclient-osgi/4.2.3</bundle>
+    <capability>esh.tp;feature=httpclient;version=9.3.14.v20161028</capability>
```
maggu2810: Is there any usage of `esh-tp-httpclient` except `esh-tp-jupnp` (e.g. in some openHAB feature)?
If there is no other usage of that feature, couldn't we move the bundle dependencies (still using `dependency="true"`) to `esh-tp-jupnp` itself and get rid of that feature?

kaikreuzer: Not yet, but as you ask: I think we should maybe add an "esh-io-transport-http" feature that extensions should then reference, if they require the HTTP client libs. Similar to how they currently express a dependency to coap, mqtt, etc. So in theory, a small solution that neither uses upnp nor any binding that requires esh-io-transport-http would not need the Jetty client bundles being present. Wdyt?

maggu2810: > So in theory, a small solution that neither uses upnp nor any binding that requires esh-io-transport-http would not need the Jetty client bundles being present. Wdyt?

Hm, isn't this already the case regardless of the `esh-io-transport-http`? If no one has a requirement that is fullfilled by that dependencies no feature / bundle of that one is installed.
Or do I miss something.

The point is, if we create a http client feature that is using the Jetty implementation, which bundles should be part of that feature? Also the Websocket bundles? Do we really need such a http client feature?

kaikreuzer: > If no one has a requirement

My question is: How do bindings express that requirement? By defining a dependency to `esh-tp-httpclient`? So far, the "tp" features are only used to define the general tp, but not used for expressing dependencies for extensions.

>  if we create a http client feature that is using the Jetty implementation, which bundles should be part of that feature?

None - it would be exactly [the same as for CoAP](https://github.com/eclipse/smarthome/blob/master/features/karaf/esh-core/src/main/feature/feature.xml#L254-L257).

>  Also the Websocket bundles?

I would say yes; unless we think they are too huge and want to split them into a separate feature that would need to be referenced independently.

maggu2810: > My question is: How do bindings express that requirement? By defining a dependency to esh-tp-httpclient? So far, the "tp" features are only used to define the general tp, but not used for expressing dependencies for extensions.

They do. The feature states a requirement on a capability that is provided by the respective feature. Requirement and capabilities are provided by the package imports and export and there are our manual defined esh.tp ones in the feature definition.

> None - it would be exactly the same as for CoAP.

The `esh-io-transport-coap` feature adds a requirement for `esh.tp;filter:="(feature=coap)"`.
That capability is provided by e.g. `esh-tp-coap` but could also provided by other ones.

> I think we should maybe add an "esh-io-transport-http" feature that extensions should then reference

Now I get your point. Currently there is a tp feature (which no non-tp feature should directly depend on) and you would like to create a non-tp io-transport-http one that extensions could depend on.

But this make the Jetty API part of the ESH transport HTTP API.
What about "The ESH API (used by bindings and extensions) should not use externals API"? If Jetty breaks its API ESH would be affected.
For mDNS and JUPnP this is different, because we don't directly use their API in bindings but rely on the discovery wrapper of the ESH framework. So Jetty's (and perhaps CoAP's -- I didn't check) API will get an API ESH "provides" / delegates.
Is my understanding correct?

kaikreuzer: No. All JUPnP, jmDNS, Californium (CoAP), Paho (MQTT), XStream (XML), GSON (JSON) and Jetty (HTTP) libs are used as the official API for ESH, see https://github.com/eclipse/smarthome/blob/master/docs/documentation/development/bindings/dependencies.md#optional-bundles.
They are considered to be widespread and thus the effort to define a proprietary API in ESH has not been done.
But anyhow, this discussion is rather out of scope of this issue. The decision to use the Jetty client in bindings was already discussed [more than 2 years ago]() and all bindings have been adapted accordingly.
This PR only makes this more explicit now - afaics, the dependencies had so far been resolved through the "http" dependency of Karaf (which in turn depends on pax-http), which simply brings a [full Jetty, including client libs](https://github.com/ops4j/org.ops4j.pax.web/blob/master/pax-web-features/src/main/resources/features.xml#L30-L47) into the framework.

> Currently there is a tp feature (which no non-tp feature should directly depend on) and you would like to create a non-tp io-transport-http one that extensions could depend on.

That's exactly right, so I have just added such a non-tp io-transport-http to this PR.

PR #4489, opened 2017-11-03
Added logging for endpoint destruction

kaikreuzer:
As discussed in https://github.com/eclipse/californium/pull/452#issuecomment-341703735.

### DIFF
--- a/extensions/binding/org.eclipse.smarthome.binding.tradfri/src/main/java/org/eclipse/smarthome/binding/tradfri/internal/TradfriCoapEndpoint.java
+++ b/extensions/binding/org.eclipse.smarthome.binding.tradfri/src/main/java/org/eclipse/smarthome/binding/tradfri/internal/TradfriCoapEndpoint.java
@@ -1,0 +1,38 @@
+/**
+ * Copyright (c) 2014-2017 by the respective copyright holders.
+ * All rights reserved. This program and the accompanying materials
+ * are made available under the terms of the Eclipse Public License v1.0
+ * which accompanies this distribution, and is available at
+ * http://www.eclipse.org/legal/epl-v10.html
+ */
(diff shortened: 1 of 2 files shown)


extensions/binding/org.eclipse.smarthome.binding.tradfri/src/main/java/org/eclipse/smarthome/binding/tradfri/internal/TradfriCoapEndpoint.java
```diff
@@ -0,0 +1,36 @@
...
+
+    private final Logger logger = LoggerFactory.getLogger(TradfriCoapEndpoint.class);
+
+    public TradfriCoapEndpoint(DTLSConnector dtlsConnector, NetworkConfig standard) {
+        super(dtlsConnector, standard);
+    }
+
+    @Override
+    public synchronized void destroy() {
+        logger.debug("Destroying CoAP endpoint.", new Exception("Endpoint destroyed"));
```
maggu2810: Shouldn't we add an enclosing "if is debug log is enabled"?
The exception instance is created regardless of the log level and so the class instance needs already to catch the whole stack trace etc. Isn't the creation of an exception and its stack trace array building a very expensive operation?
The exception is only for logging so it does not really care which class is used, but I still prefer a runtime one :wink: (if you don't care).

kaikreuzer: Yes, it is potentially expensive, but it is a piece of code that ideally should hardly ever be executed.
But your comment is correct in general, so I have updated the PR as suggested.

maggu2810:
So, hopefully you detect the bug now. Good luck :wink:

PR #4496, opened 2017-11-04
Chart servlet: new optional setting to set a maximum chart width

### LINKED ISSUE #4493: Chart servlet: performance problem on RPI
Using android app and opehHAB server on my RPI 2, I see that it takes several seconds to display a chart.
As a remember the android app is requesting a big chart with my tablet (2560x1280).

After tracing the chart servlet on my RPI2, I discovered that the server takes around 7 seconds to handle the HTTPS request and build the PNG data. What takes most of the time (5 seconds) is the writing of the PNG data done at this line: https://github.com/eclipse/smarthome/blob/master/bundles/ui/org.eclipse.smarthome.ui/src/main/java/org/eclipse/smarthome/ui/internal/chart/ChartServlet.java#L281

Any idea whether we can make `ImageIO.write` run faster ?

What would be the solution to avoid that case ? Should the server build a smaller chart in case the CPU is slow ?

The discussion started here: https://github.com/openhab/openhab.android/issues/440

lolodomo:
Fixes #4493


lolodomo:
This new setting is optional, it allows mainly on slow hardware like RPI to choose between speed and size for the chart.
It concerns UIs that request very hi-res charts like the Android app.

lolodomo:
Maybe in another PR.

Issue #4257, opened 2017-09-13
[HttpUtil] Suggestion for using a builder like pattern for Http requests


martinvw:
Hello,

I was using the `HttpUtil` and found it a little bit inconvenient to use with all those `null` parameters in between, so I prepared a small suggestion which makes it more readable in my opinion. Wdyt?

Is:

```java
String httpMethod = "POST";
String url = "http://" + host + "/cgi-bin/MUH44TP_getsetparams.cgi";
String content = "{tag:ptn}";
InputStream stream = new ByteArrayInputStream(content.getBytes(StandardCharsets.UTF_8));
int timeout = 5000;
String response = HttpUtil.executeUrl(httpMethod, url, null, stream, null, timeout);
```
Suggestion:

```java
HttpUtil.postTo(url)
    .withContent("{tag:ptn}")
    .withTimeout(5, TimeUnit.SECONDS)
    .getContentAsString();
```

Current:

```java
String METHOD = "GET";
int TIMEOUT = 30 * 1000; // 30 seconds
Properties HEADERS = new Properties();
HEADERS.put("User-Agent", "openHAB / senseBox binding " + version.toString());
String query = SENSEMAP_API_URL_BASE + "/boxes/" + senseBoxId;
String body = HttpUtil.executeUrl(METHOD, query, HEADERS, null, null, TIMEOUT);
``` 
Suggestion:

```java
HttpUtil.getFrom(SENSEMAP_API_URL_BASE + "/boxes/" + senseBoxId)
    .header("User-Agent", "openHAB / senseBox binding " + version)
    .withTimeout(30, TimeUnit.SECONDS)
    .getContentAsString();
```

Any thoughts on this?

igilham:
Looks way better and should make it easier to avoid putting params in the wrong positions, NPEs etc.

martinvw:
@sjka wdyt? If enough people like it I might be able to build a first sketch of it in the train to home tonight :-)

kaikreuzer:
Would be ok for me.
But please note that the HttpUtil was meant to be a helper class to make simple HTTP interaction possible for rules, i.e. the main "client" of that class should be `org.eclipse.smarthome.model.script.actions.HTTP` for which the syntax is pretty irrelevant.

HttpUtil is NOT meant to be the HTTP client of choice for bindings or other Java code - we actually suggest to use the Jetty client directly. So in this context, I am not sure how much effort we should spend on enhancing the HttpUtil class.

[...]

PR #4497, opened 2017-11-04
Net utils: optional timeout parameter for the downloadImage methods

lolodomo:

### DIFF
--- a/bundles/io/org.eclipse.smarthome.io.net/src/main/java/org/eclipse/smarthome/io/net/http/HttpUtil.java
+++ b/bundles/io/org.eclipse.smarthome.io.net/src/main/java/org/eclipse/smarthome/io/net/http/HttpUtil.java
@@ -52,6 +52,8 @@ public class HttpUtil {
 
     private static Logger logger = LoggerFactory.getLogger(HttpUtil.class);
 
+    private static final int DEFAULT_TIMEOUT_MS = 5000;
+
     private static final HttpClient CLIENT = new HttpClient(new SslContextFactory());
 
     private static class ProxyParams {
@@ -68,8 +70,8 @@ public class HttpUtil {
      * set into the {@link HttpClient}.
      *
      * @param httpMethod the HTTP method to use
-     * @param url the url to execute (in milliseconds)
-     * @param timeout the socket timeout to wait for data
+     * @param url the url to execute
+     * @param timeout the socket timeout in milliseconds to wait for data
      *
      * @return the response body or <code>NULL</code> when the request went wrong
      * @throws IOException when the request execution failed, timed out or it was interrupted
@@ -84,11 +86,11 @@ public class HttpUtil {
      * set into the {@link HttpClient}.
      *
      * @param httpMethod the HTTP method to use
-     * @param url the url to execute (in milliseconds)
+     * @param url the url to execute
      * @param content the content to be send to the given <code>url</code> or <code>null</code> if no content should be
      *            send.
      * @param contentType the content type of the given <code>content</code>
-     * @param timeout the socket timeout to wait for data
+     * @param timeout the socket timeout in milliseconds to wait for data
      *
      * @return the response body or <code>NULL</code> when the request went wrong
      * @throws IOException when the request execution failed, timed out or it was interrupted
@@ -105,12 +107,12 @@ public class HttpUtil {
      * set into the {@link HttpClient}.
      *
      * @param httpMethod the HTTP method to use
-     * @param url the url to execute (in milliseconds)
+     * @param url the url to execute
      * @param httpHeaders optional http request headers which has to be sent within request
      * @param content the content to be send to the given <code>url</code> or <code>null</code> if no content should be
      *            send.
      * @param contentType the content type of the given <code>content</code>
-     * @param timeout the socket timeout to wait for data
+     * @param timeout the socket timeout in milliseconds to wait for data
      *
      * @return the response body or <code>NULL</code> when the request went wrong
      * @throws IOException when the request execution failed, timed out or it was interrupted
@@ -128,12 +130,12 @@ public class HttpUtil {
      * Executes the given <code>url</code> with the given <code>httpMethod</code>
      *
      * @param httpMethod the HTTP method to use
-     * @param url the url to execute (in milliseconds)
+     * @param url the url to execute
      * @param httpHeaders optional HTTP headers which has to be set on request
      * @param content the content to be send to the given <code>url</code> or <code>null</code> if no content should be
      *            send.
      * @param contentType the content type of the given <code>content</code>
-     * @param timeout the socket timeout to wait for data
+     * @param timeout the socket timeout in milliseconds to wait for data
      * @param proxyHost the hostname of the proxy
      * @param proxyPort the port of the proxy
      * @param proxyUser the username to authenticate with the proxy
@@ -163,12 +165,12 @@ public class HttpUtil {
      * Executes the given <code>url</code> with the given <code>httpMethod</code>
      *
      * @param httpMethod the HTTP method to use
-     * @param url the url to execute (in milliseconds)
+     * @param url the url to execute
      * @param httpHeaders optional HTTP headers which has to be set on request
      * @param content the content to be send to the given <code>url</code> or <code>null</code> if no content should be
      *            send.
      * @param contentType the content type of the given <code>content</code>
-     * @param timeout the socket timeout to wait for data
+     * @param timeout the socket timeout in milliseconds to wait for data
      * @param proxyHost the hostname of the proxy
      * @param proxyPort the port of the proxy
      * @param proxyUser the username to authenticate with the proxy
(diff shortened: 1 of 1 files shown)


bundles/io/org.eclipse.smarthome.io.net/src/main/java/org/eclipse/smarthome/io/net/http/HttpUtil.java
```diff
@@ -74,7 +76,7 @@
      * @return the response body or <code>NULL</code> when the request went wrong
      * @throws IOException when the request execution failed, timed out or it was interrupted
      */
-    public static String executeUrl(String httpMethod, String url, int timeout) throws IOException {
+    public static String executeUrl(String httpMethod, String url, long timeout) throws IOException {
```
martinvw: Does this break compatibility with existing jar's on the marketplace, is it really needed? A timeout of more then 2.000.000 looks too high to me anyway.

bundles/io/org.eclipse.smarthome.io.net/src/main/java/org/eclipse/smarthome/io/net/http/HttpUtil.java
```diff
@@ -377,7 +393,22 @@ public static RawType downloadImage(String url) {
...
+    }
+
+    /**
+     * Download the image data from an URL.
+     *
+     * @param url the URL of the image to be downloaded
+     * @param scanTypeInContent true to allow the scan of data to determine the content type if not found in the headers
+     * @param maxContentLength the maximum data size in bytes to trigger the download; any negative value to ignore the
+     *            data size
+     * @param timeout the socket timeout to wait for data
```
martinvw: Can you add the unit here

bundles/io/org.eclipse.smarthome.io.net/src/main/java/org/eclipse/smarthome/io/net/http/HttpUtil.java
```diff
@@ -388,17 +419,18 @@ public static RawType downloadImage(String url, boolean scanTypeInContent, long
      * @param scanTypeInContent true to allow the scan of data to determine the content type if not found in the headers
      * @param maxContentLength the maximum data size in bytes to trigger the download; any negative value to ignore the
      *            data size
+     * @param timeout the socket timeout to wait for data
```
martinvw: Please add the unit here

bundles/io/org.eclipse.smarthome.io.net/src/main/java/org/eclipse/smarthome/io/net/http/HttpUtil.java
```diff
@@ -52,6 +52,8 @@
 
     private static Logger logger = LoggerFactory.getLogger(HttpUtil.class);
 
+    private static final long DEFAULT_TIMEOUT = 5000;
```
martinvw: Given the value I would hope it's seconds, can you make it clear by changing the name for example `DEFAULT_TIMEOUT _MS`

Alternative what would you think about accepting a Duration, see for example the following fragment:

```java
downloadImage(url, Duration.ofSeconds(15));

// or with a static import
downloadImage(url, ofSeconds(15));
```

bundles/io/org.eclipse.smarthome.io.net/src/main/java/org/eclipse/smarthome/io/net/http/HttpUtil.java
```diff
@@ -363,7 +365,21 @@ private static void startHttpClient(HttpClient client) {
...
+     * Download the image data from an URL.
+     *
+     * If content type is not found in the headers, the data is scanned to determine the content type.
+     *
+     * @param url the URL of the image to be downloaded
+     * @param timeout the socket timeout to wait for data
+     * @return a RawType object containing the image, null if the content type could not be found or the content type is
+     *         not an image
+     */
+    public static RawType downloadImage(String url, long timeout) {
```
martinvw: It might make the API a little bit confusing if the second parameter is sometimes a boolean `scanTypeInContent ` and sometime the timeout.

bundles/io/org.eclipse.smarthome.io.net/src/main/java/org/eclipse/smarthome/io/net/http/HttpUtil.java
lolodomo: I will rename the constant.

bundles/io/org.eclipse.smarthome.io.net/src/main/java/org/eclipse/smarthome/io/net/http/HttpUtil.java
lolodomo: I made the mistake of thinking that max int value was 32767 !
I will return to int type.

bundles/io/org.eclipse.smarthome.io.net/src/main/java/org/eclipse/smarthome/io/net/http/HttpUtil.java
lolodomo: Will do it.

bundles/io/org.eclipse.smarthome.io.net/src/main/java/org/eclipse/smarthome/io/net/http/HttpUtil.java
lolodomo: What do you propose considering that I don't want to break the current API ? To add only the new method with four arguments, the last one being the timeout ?

martinvw: this one is not breaking but might make it less easy to use, but I still intend to work on https://github.com/eclipse/smarthome/issues/4257 so maybe only introduce it in indeed in the 4 parameter version. It is up to you and the maintainers here :-)

lolodomo:
Changes done.

bundles/io/org.eclipse.smarthome.io.net/src/main/java/org/eclipse/smarthome/io/net/http/HttpUtil.java
```diff
@@ -69,7 +71,7 @@
      *
      * @param httpMethod the HTTP method to use
      * @param url the url to execute (in milliseconds)
```
martinvw: Noe you adder the unit maybe also remove the unit from the url....

bundles/io/org.eclipse.smarthome.io.net/src/main/java/org/eclipse/smarthome/io/net/http/HttpUtil.java
```diff
@@ -377,7 +379,22 @@ public static RawType downloadImage(String url) {
...
+     *
+     * @param url the URL of the image to be downloaded
+     * @param scanTypeInContent true to allow the scan of data to determine the content type if not found in the headers
+     * @param maxContentLength the maximum data size in bytes to trigger the download; any negative value to ignore the
+     *            data size
+     * @param timeout the socket timeout in milliseconds to wait for data
+     * @return a RawType object containing the image, null if the content type could not be found or the content type is
+     *         not an image or the data size is too big
+     */
+    public static RawType downloadImage(String url, boolean scanTypeInContent, long maxContentLength, int timeout) {
```
martinvw: Given the current executeurl methods putting the timeout in the end is already so maybe just do it, it makes the code In your binding better.

lolodomo: Yes, that is what is done, the timeout parameter is the last parameter.
Sorry I don't understand your comment.

lolodomo:
I fixed the comments.
But I don't understand your last request regarding the position of the timeout parameter.

bundles/io/org.eclipse.smarthome.io.net/src/main/java/org/eclipse/smarthome/io/net/http/HttpUtil.java
martinvw: I referred to my own earlier comment:

> It might make the API a little bit confusing if the second parameter is sometimes a boolean scanTypeInContent and sometime the timeout.

I yesterday noticed (because it was shown in the diff now) that the existing methods have the same pattern applied that every method has the timeout as the last parameter, meaning that it is sometimes the second, third or fourth.

So I was actually down voting my previous comment and suggesting that if it was needed for better code in the calling side, that it is proper to append it also to the single argument method.

lolodomo:
Ok I restored the version with the timeout as second parameter.

martinvw:
Perfect, just to be clear still looks good to me LGTM

bundles/io/org.eclipse.smarthome.io.net/src/main/java/org/eclipse/smarthome/io/net/http/HttpUtil.java
```diff
@@ -388,17 +419,18 @@ public static RawType downloadImage(String url, boolean scanTypeInContent, long
      * @param scanTypeInContent true to allow the scan of data to determine the content type if not found in the headers
      * @param maxContentLength the maximum data size in bytes to trigger the download; any negative value to ignore the
      *            data size
+     * @param timeout the socket timeout in milliseconds to wait for data
      * @return a RawType object containing the downloaded data, null if the content type does not match the expected
      *         type or the data size is too big
      */
     public static RawType downloadData(String url, String contentTypeRegex, boolean scanTypeInContent,
-            long maxContentLength) {
+            long maxContentLength, int timeout) {
```
sjka: Maybe I miss something here, but isn't this breaking the API? Keep in mind - the purpose of this class is (as I learned) primarily to be exposed in scripts and rules. Therefore this would breaking endusers' code here.

martinvw: You are right this method does not seem have a backwards compatible variant, missed that :-)

lolodomo: I will fix that.

lolodomo:
Change done, hoping that it is the last one.

sjka:
Yes, I think so 😆 
Thanks!!

PR #4514, opened 2017-11-07
aligned equals implementation of GenericItem

sjka:
...with the rest of the framework: The other entities (e.g.
Rule, ThingImpl,...) only check for the identity but not for
all properties.


sjka:
@hsudbrock, @kaikreuzer after having talked about #4442 and having looked around a little, I really think the previous PR was going into the wrong direction. Therefore I'd suggest cutting down the equals implementation of GenericItem so that it matches the same scheme as the other entities. 

Unless anybody is aware of anything that lets this look like maybe not such a good idea...

sjka:
Adding the APIBreaking label because Items might behave different, especially when herding them inside collections.

kaikreuzer:
Uah, this somehow looks risky 😟 . But at least Travis thinks it is all fine and safe and I don't have any good arguments what might break...
Nonetheless, I don't dare pressing the merge button yet, maybe tomorrow.

Issue #4464, opened 2017-10-29
CommunicationManager & DefaultMasterProfile : linking items through channels not supported by default


kgoderis:
@sjka @kaikreuzer DefaultMasterProfile does not support items configuration like for example:

`DateTime Time "Time [%1$tT]" { channel="ntp:ntp:myhometown:dateTime", channel="knx:generic:ip1:1_1_12:9_1_1"}`

Well, it does, but in comparison to for example openHAB 1, it is no longer possible to have a binding, in casu NTP, update the Time Item, and automatically have the runtime post an update for that Item so that the same State can be handled by the other binding, in casu KNX

e.g. I presume it was by design that DefaultMasterProfile contains

```
    @Override
    public void onUpdate(ItemChannelLink link, Thing thing, State state) {
        // no-op
    }
```

Maybe I overlooked something, but shouldn't a simple Profile, similar to RawButtonToggleProfile, be provided in the core to enable these kinds of item configurations? Or, from a user perspective, should these kind of Item configuration be handled through Rules?

kgoderis:
Ok, The above example/usage can be solved by assigning a DefaultSlaveProfile to the given knx channel, but on a more general note, referring to https://github.com/eclipse/smarthome/pull/2881, which is superseded by Profiles concept, there might be a benefit to do away with the Autoupdate binding, and rather re-implement it through Profiles, possibly extending it with the inter-binding and intra-binding bridging concepts of https://github.com/eclipse/smarthome/pull/2881

That being said, I am wondering if we could/should enable access to the ItemChannelLinkRegistry for these and other more advanced use-cases?

sjka:
> The above example/usage can be solved by assigning a DefaultSlaveProfile to the given knx channel

That's exactly its purpose!

> there might be a benefit to do away with the Autoupdate binding, and rather re-implement it through Profiles

Yes, that's the plan to make it a configuration of the default master profile. The tricky part will be to keep backward compatibility for openHAB's compatibility layer.

> possibly extending it with the inter-binding and intra-binding bridging concepts

I don't think this should be in scope here. You brought this up in the context of the KNX binding. This however is something that _could_ be "implemented" in KNX side by assigning common GAs and therefore rather _should_ be done there. I'd really like to avoid KNX concepts sneaking into ESH.

kgoderis:
[...]

PR #4516, opened 2017-11-08
Profile API improvements

sjka:
Having had some first [feedback](https://community.openhab.org/t/enocean-bindings-opencean-or-aleoncean/12493/55) and some chats with @kaikreuzer, it obviously made sense to rework and simplify the profile API at some points. This is the current result.

* Simplified the API for implementing profiles
  (at the price of a little more complexity on CommunicationManager side)
* Allow implementing profiles without exposing Items and ItemChannelLinks
  to bindings
* Restore backward-compatibility by making the default master profile
  forward handleUpdate() calls to ThingHandlers again (see also https://github.com/eclipse/smarthome/issues/4464#issuecomment-341425684)
* Exposing constants in SystemProfiles
* Allow for easier evolvement of the API
* Having the ProfileCallbacks requires less code within profiles and therefore
  eliminates the need to export the default profiles for inheritance.
* Renamed system profiles from master/slave to default/follow

Although this technically *is* APIBreaking (and therefore adding the corresponding label), there cannot yet exist a productive usage of this API, therefore I don't see any risks involved here.


bundles/core/org.eclipse.smarthome.core.thing/src/main/java/org/eclipse/smarthome/core/thing/profiles/SystemProfiles.java
```diff
@@ -0,0 +1,86 @@
...
+        @Override
+        public String getLabel() {
+            return "Slave";
+        }
+    };
+
+    public static final TriggerProfileType RAWBUTTON_TOGGLE_SWITCH_TYPE = new TriggerProfileType() {
+        @Override
+        public ProfileTypeUID getUID() {
+            return SLAVE;
```
kungfoolfighting: Wrong UID

bundles/core/org.eclipse.smarthome.core.thing/src/main/java/org/eclipse/smarthome/core/thing/internal/link/ItemChannelLinkConfigDescriptionProvider.java
```diff
@@ -90,21 +86,41 @@ public ConfigDescription getConfigDescription(URI uri, Locale locale) {
...
-    @Reference(cardinality = ReferenceCardinality.MULTIPLE, policy = ReferencePolicy.DYNAMIC)
-    public void addProfileAdvisor(ProfileAdvisor profileAdvisor) {
-        profileAdvisors.add(profileAdvisor);
+    private boolean isSupportedItemType(ProfileType profileType, Item item) {
+        return profileType.getSupportedItemTypes() == ProfileType.ANY_ITEM_TYPE
+                || profileType.getSupportedItemTypes().contains(item.getType());
+    }
+
+    private boolean isSupportedChannelType(ProfileType profileType, Channel channel) {
+        return profileType instanceof StateProfileType
```
maggu2810: `profileType instanceof TriggerProfileType`?

bundles/core/org.eclipse.smarthome.core.thing/src/main/java/org/eclipse/smarthome/core/thing/profiles/ProfileType.java
```diff
@@ -0,0 +1,43 @@
...
+ * @author Simon Kaufmann - initial contribution and API.
+ *
+ */
+@NonNullByDefault
+public interface ProfileType extends Identifiable<ProfileTypeUID> {
+
+    /**
+     * Constant for mathing *ANY* item type.
+     */
+    public static final Collection<String> ANY_ITEM_TYPE = new ArrayList<>(0);
```
maggu2810: Just a minor notice:
We can declare variables in Java interfaces. By default, these are public, final and static. 
So, no need to add the default modifiers.

bundles/core/org.eclipse.smarthome.core.thing/src/main/java/org/eclipse/smarthome/core/thing/profiles/SystemProfiles.java
```diff
@@ -0,0 +1,86 @@
...
+import org.eclipse.smarthome.core.thing.DefaultSystemChannelTypeProvider;
+import org.eclipse.smarthome.core.thing.type.ChannelTypeUID;
+
+/**
+ * System profile constants.
+ *
+ * @author Simon Kaufmann - initial contribution and API.
+ *
+ */
+public interface SystemProfiles {
```
maggu2810: Same comments about default modifiers as above.

bundles/core/org.eclipse.smarthome.core.thing/src/main/java/org/eclipse/smarthome/core/thing/profiles/TriggerProfileType.java
```diff
@@ -0,0 +1,36 @@
...
+ * @author Simon Kaufmann - initial contribution and API.
+ *
+ */
+@NonNullByDefault
+public interface TriggerProfileType extends ProfileType {
+
+    /**
+     * Constant for mathing *ANY* channel type.
+     */
+    public static final Collection<ChannelTypeUID> ANY_CHANNEL_TYPE = new ArrayList<>(0);
```
maggu2810: see above (default modifiers)

bundles/core/org.eclipse.smarthome.core.thing/src/main/java/org/eclipse/smarthome/core/thing/internal/link/ItemChannelLinkConfigDescriptionProvider.java
kungfoolfighting: No I think that StateProfileType is what is meant here, but I don't quite understand why a StateProfileType would always support all ChannelTypes. Could there not be profiles that only support a specific channel type?

bundles/core/org.eclipse.smarthome.core.thing/src/main/java/org/eclipse/smarthome/core/thing/internal/link/ItemChannelLinkConfigDescriptionProvider.java
```diff
@@ -90,21 +86,41 @@ public ConfigDescription getConfigDescription(URI uri, Locale locale) {
...
+                || ((TriggerProfileType) profileType)
```
kungfoolfighting: While it is currently the case that usually profiles fall within the StateProfileType or the TriggerProfileType group, it would make sense to check if the profile type is an instance of TriggerProfileType here anyways in order to make sure we dont get a problem here later, in case other types are added.

kungfoolfighting:
I might have done something wrong, but a working item definition that used the rawbutton channel and a toggle profile stopped working after I merged this patch.
I get a log entry that says It was not able to find a profile factory for this profile type. This message is from line 211 in CommunicationManager.

sjka:
Thanks, I will have a look.

sjka:
What does your item definition look like?

bundles/core/org.eclipse.smarthome.core.thing/src/main/java/org/eclipse/smarthome/core/thing/internal/link/ItemChannelLinkConfigDescriptionProvider.java
```diff
@@ -107,7 +107,7 @@ private boolean isSupportedItemType(ProfileType profileType, Item item) {
     }
 
     private boolean isSupportedChannelType(ProfileType profileType, Channel channel) {
-        return profileType instanceof StateProfileType
+        return profileType instanceof TriggerProfileType
```
kungfoolfighting: This must be wrong now. It would now always return true if profileType is of type TriggerProfileType. That would mean that the rest of the condition is superfluous.

philomatic:
Thanks for being compliant with me! 😄 

If I get this right there is no need for another ProfileCallback implementation right now (Profiles alone  "should" be enough for information flow)?
It's irritating in the first moment, but reduces complexity for new profiles and code duplication **a lot**!

LGTM!

philomatic:
How do you want to deal with

```
@Deprecated
void handleUpdate(ChannelUID channelUID, State newState);
```

in ThingManager?

I think we should either stay backward-compatible by forwarding handleUpdate() calls (and therefore cancel deprecation) or drop support for handleUpdate() completely before 1.0.
Just mentioning this so we won't have to deal with it twice.

Best regards

kungfoolfighting:
@sjka I don't have the code at work atm. But it looks something like this:
`Switch ToggleSwitch {channel="bindingname:thingID:triggerChannelName" [profile="system:"RawButtonToggleSwitchProfile"] }`
Thinking about it, maybe I forgot to rename the profile from RawButtonToggleProfile to RawButtonToggleSwitchProfile. It was very late yesterday. So I will check tonight and keep you posted.

sjka:
> maybe I forgot to rename the profile

That's why I've been asking... It should read `rawbutton-toggle` or `system:rawbutton-toggle` now.

kungfoolfighting:
Right My mistake. So It if the ProfileTypeUID didnt change, then it is still the correct string, being "system:rawbutton-toggle". It did work yesterday before the merge and I was happily toggling my lights. I just made a mistake right now. So I had a working item, and now it no longer works. 
I will post the exact logger message tonight. but it was something like
"no ProfileFactory found which supports 'itemname_system_rawbutton-toggle_channelname'" I am not exactly certain anymore, but it concatenated parts with underscore and the system_rawbutton_toggle was part of it. 
Maybe that helps a bit. But don't spend too much time investigating on this flimsy information before I have verified it.

sjka:
No, my initial post was wrong (and I corrected it immediately, but that doesn't send out notification mails). The profile ID indeed has changed and got a `-switch` appended. Sure, just try it out once you find the time and let me know...

kaikreuzer:
> How do you want to deal with `@Deprecated void handleUpdate`?

-> to be discussed in https://github.com/eclipse/smarthome/issues/4525.

bundles/core/org.eclipse.smarthome.core.thing/src/main/java/org/eclipse/smarthome/core/thing/profiles/ProfileCallback.java
```diff
@@ -0,0 +1,51 @@
...
+     * @param state
+     */
+    void handleUpdate(State state);
+
+    /**
+     * Send a command to the framework.
+     *
+     * @param command
+     */
+    void sendCommandEvent(Command command);
```
kaikreuzer: I would remove the word "Event" here - after all, you are sending a command, not an event.

bundles/core/org.eclipse.smarthome.core.thing/src/main/java/org/eclipse/smarthome/core/thing/profiles/ProfileCallback.java
```diff
@@ -0,0 +1,51 @@
...
+
+    /**
+     * Send a state update to the framework.
+     *
+     * @param state
+     */
+    void sendStateEvent(State state);
```
kaikreuzer: I would remove the word "Event" here - after all, you are sending a state, not an event.

bundles/core/org.eclipse.smarthome.core.thing/src/main/java/org/eclipse/smarthome/core/thing/profiles/StateProfile.java
```diff
@@ -21,46 +17,35 @@
  * @author Simon Kaufmann - initial contribution and API.
  *
  */
+@NonNullByDefault
 public interface StateProfile extends Profile {
```
kaikreuzer: I wonder if it would make sense to make the method names more self-explanatory here.
After all, ALL methods are called from "the outside", so ideally they could be all "on...". We would then only need to add some information about the sender like
`onCommandFromItem`, `onStateUpdateFromItem`, `onCommandFromHandler` and `onStateUpdateFromHandler` or something like that?

kungfoolfighting:
After I appended the -switch to the profile name, the toggle profile worked again.
I am still not able to make my own profile work. 
Do I need to define some XML or something?
What sort of information do I need to provide so you can tell me what I am doing wrong?
None of the break points in any ofthe functions in the profile factory are ever hit.
I have basically just taken the toggle profile as a template and done everything the same way.
When looking into the code I simply cannot find how the registries are being filled with the factories though. It is all obfuscated through the osgi services.

bundles/core/org.eclipse.smarthome.core.thing/src/main/java/org/eclipse/smarthome/core/thing/profiles/TriggerProfile.java
```diff
@@ -24,9 +22,21 @@
...
+     * @param event the event payload
+     */
+    void onTrigger(String event);
+
+    /**
+     * Will be called if the linked item's state is updated.
+     *
+     * @param state the new {@link State}
+     */
+    void onUpdate(State state);
```
kaikreuzer: Maybe better `onStateUpdate`?

bundles/core/org.eclipse.smarthome.core.thing/src/main/java/org/eclipse/smarthome/core/thing/profiles/ProfileTypeProvider.java
```diff
@@ -0,0 +1,22 @@
...
+import org.eclipse.smarthome.core.common.registry.Provider;
+
+/**
+ * A {@link ProfileTypeProvider} is responsible for providing {@link ProfileType}s.
+ *
+ * @author Simon Kaufmann - initial contribution and API.
+ *
+ */
+@NonNullByDefault
+public interface ProfileTypeProvider extends Provider<ProfileType> {
```
kaikreuzer: Shouldn't this provider rather be similar to `ThingTypeProvider` etc., i.e. specifically able to provide a profile type for a given locale (such that its label is localisable)?

bundles/core/org.eclipse.smarthome.core.thing/src/main/java/org/eclipse/smarthome/core/thing/profiles/Profile.java
```diff
@@ -19,6 +21,7 @@
  * @author Simon Kaufmann - initial contribution and API.
  *
  */
+@NonNullByDefault
 public interface Profile {
```
kaikreuzer: I would expect to have a `ProfileTypeUID getProfileTypeUID();` method here.

bundles/core/org.eclipse.smarthome.core.thing/src/main/java/org/eclipse/smarthome/core/thing/internal/profiles/SystemProfileFactory.java
```diff
@@ -0,0 +1,106 @@
...
+    }
+
+    @Override
+    public @NonNull Collection<@NonNull ProfileTypeUID> getSupportedProfileTypeUIDs() {
+        return SUPPORTED_PROFILE_TYPE_UIDS;
+    }
+
+    @Override
+    public void addProviderChangeListener(ProviderChangeListener<@NonNull ProfileType> listener) {
+        // TODO Auto-generated method stub
```
kaikreuzer: replace or delete comment

bundles/core/org.eclipse.smarthome.core.thing/src/main/java/org/eclipse/smarthome/core/thing/internal/profiles/SystemProfileFactory.java
```diff
@@ -0,0 +1,106 @@
...
+    }
+
+    @Override
+    public void removeProviderChangeListener(ProviderChangeListener<@NonNull ProfileType> listener) {
+        // TODO Auto-generated method stub
```
kaikreuzer: replace or delete comment

bundles/core/org.eclipse.smarthome.core.thing/src/main/java/org/eclipse/smarthome/core/thing/profiles/StateProfile.java
philomatic: +1

Profiles (as well as channels) are between Items and ThingHandlers, they don't belong to one of them.

Thereby stateUpdated / onUpdate was confusing to me too.

bundles/core/org.eclipse.smarthome.core.thing/src/main/java/org/eclipse/smarthome/core/thing/internal/profiles/ProfileTypeRegistry.java
```diff
@@ -19,6 +20,10 @@
  * @author Simon Kaufmann - initial contribution and API.
  *
  */
-public interface ProfileTypeRegistry extends Registry<ProfileType, ProfileTypeUID> {
+public interface ProfileTypeRegistry {
+
+    public List<ProfileType> getProfileTypes();
```
kaikreuzer: This method should take a locale as well.

bundles/core/org.eclipse.smarthome.core.thing/src/main/java/org/eclipse/smarthome/core/thing/profiles/Profile.java
```diff
@@ -24,4 +25,13 @@
 @NonNullByDefault
 public interface Profile {
 
+    ProfileTypeUID getProfileTypeUID();
```
kaikreuzer: Would you also spend it a single line of JavaDoc? :-)

bundles/core/org.eclipse.smarthome.core.thing/src/main/java/org/eclipse/smarthome/core/thing/profiles/Profile.java
```diff
@@ -19,6 +22,16 @@
  * @author Simon Kaufmann - initial contribution and API.
  *
  */
+@NonNullByDefault
 public interface Profile {
 
+    ProfileTypeUID getProfileTypeUID();
```
kungfoolfighting: This should probably be static. It was annoying to me that it wasnt static when deriving from it.

kungfoolfighting:
I checked the message I get when my profile doesn't work, it is:
`2017-11-10 18:13:34.127 [WARN ] [e.s.c.t.i.CommunicationManager:212  ] - no ProfileFactory found which supports 'jenocean:rocker-channel-to-ONOFF-profile'`

The item definition looks like this:

`Switch testswitch1 "Test Switch" {channel="jenocean:RockerSwitch:714f6f57:channel_a_rocker" [ profile="jenocean:rocker-channel-to-ONOFF-profile" ]}`

Does that give you a hint?

maggu2810:
I see several non-null violations

* `Identifiable<T>` is using `@NonNullByDefault`
* `ProfileType` extends `Identifiable<ProfileTypeUID>` and is using `@NonNullByDefault`
* `StateProfileType` extends `ProfileType` and is using `@NonNullByDefault`
* `SystemProfiles` contains a new annonymous `StateProfileType` implementations and does not use `@NonNullByDefault`

Issue #4302, opened 2017-09-19
REST: get things of type


maggu2810:
There is currently no method to get all things of a special type only (or I didn't find them).
We are using setups that "get all things" call results into a > 2 MiB response.
Sometimes we need to know only things of a given type first.
I can create that interface in our custom product or you could extend the thing resource to support this request officially.
WDYT?

sjka:
I don't see a reason _not_ to put it here. "go" from my side.

kaikreuzer:
We didn't go into detailed filtering mechanisms on any rest resource from the start, but the it was clear that this would be required sooner or later.

When adding those now for certain cases, we should make sure to find a good general scheme that we can apply to all different kinds of resources. Ideally, we should have flexible query parameters, maybe also support substring matching and somehow cleverly feed that into a stream-based get method on the registries.

maggu2810:
The `GET` method does not support data in the content, so we could use the URL for the filter information. Correct?

htreu:
Correct. We could provide something like `/rest/things?thing-type=hue:lamp` which will make `thing-type` a pre defined query parameter for the things resource. The alternative is to encode the whole filter into a single parameter like `rest/things?filter=thing-type%3Dhue%3Alamp%3Blabel%3DMy%20Hue%20Lamp` which would be more flexible in a way that we create filters dynamically from the filter string. This way nearly every attribute or property of a resource may be used as a filter.

maggu2810:
We would need a implementation that could be reused e.g.

```Java
boolean testLdapExpression(final String expression, final Map<String, String> map)
```

The `map` is filled with key value pairs and is under control of the caller.
The thing resource could handle it this way (pseudo code):

```Java
for (final ThingDTO thing : getAllThings()) {
    final Map<String,String> props = new HashMap<>();
    props.put("thing-type", thing.thingType);
    props.put("label", thing.label);
    if (testLdapExpression(filter, props) {
        result.add(thing);
    }
}
```

another implementation will put other properties...

kaikreuzer:
I am personally not a huge fan of LDAP filters and imho this is also rather unusual in the JS/REST world - but maybe we should ask some web experts on that :-)

htreu:
What I like better on the explicit query parameter name (i.e. `/rest/things?thing-type=hue:lamp`) is the fact that the swagger documentation will give you direct access to all available parameters. In addition the URI is more human readable and I suspect we will not have that many query params.

[...]

PR #4517, opened 2017-11-08
optionally limit fields in /rest/item result

sjka:
While we have #4302 for _filtering_ REST APIs, [this forum post](https://www.eclipse.org/forums/index.php?t=rview&goto=1775927) made me experiment a little how we could (generically) introduce a mechanism to _map_ the resultset to something smaller. IMHO it would work nicely like this:

I'd suggest introducing an optional query parameter where it is possible to specify the field names which should only be included in the result, i.e.

```
/rest/items?fields=name,type
```

would give only 

```json
[
  {
    "type": "Dimmer",
    "name": "sonos_PLAY1_RINCON_949F3E140B2E01400_volume"
  },
  {
    "type": "String",
    "name": "sonos_PLAY1_RINCON_949F3E140B2E01400_zonename"
  },
  {
    "type": "Player",
    "name": "sonos_PLAY1_RINCON_949F3E85A92801400_control"
  },
  {
    "type": "String",
    "name": "sonos_PLAY1_RINCON_949F3E140B2E01400_tuneinstationid"
  },
```

What do you think? The very same mechanism obviously could be used for other resources/entities too, of course, if viable.


htreu:
Looks good! How about extracting the function into a DTOMapper and passing this into the Resource via OSGi DS? It will be plain simple to incorporate it into other Resources then.

sjka:
Yepp, that's pretty much what I had in mind. I was just waiting for some word of encouragement before touching one - and thereby migrating all the services in the o.e.sh.io.rest.core bundle to DS Annotations.

htreu:
ha, I knew it. A W E S O M E!!eins11

maggu2810:
If the filter contains `id` and `properties` a result could look like this one:

```text
[
  {
    "id": "1",
    "properties": {
      "id": "2"
    }
  }
]
```

Correct?

Do we need to be able to filter also one of the both IDs? Or is it okay to keep every ID field if `id` is used?

htreu:
From what I understand form the code it will only filter top level fields. So in your example the full "properties" object is passed but not filtered at all.

sjka:
Yes, it's only filtering attributes on top level. They will be either in or out completely. 
For the sake of simplicity I though this might be sufficient. Otherwise the mapping instructions might end up much bigger than the actual content and we would need to invent yet another mapping language which can be squeezed into URLs. Certainly possible - but really needed?

htreu:
For nested level filtering it could be something like `fields=id,properties.id,properties.value`. But I see this as an improvement when really needed and not in this PR.
wdyt?

PR #4518, opened 2017-11-09
Expose the item triggering the execution of a rule

jmsjones:
This PR adds the implicit variable "triggeringItem" to text file based rules. The variable is useful to identify which item triggered a rule when the rule can be triggered by more than one item. The implementation is consistent with the existing implicit variables "receivedCommand" and "previousState".

This change is only useful for rules with multiple conditions.
`Item Switch01 changed or`
`Item Switch02 changed`
The plan is to extend this further to support rules bound to groups.


jmsjones:
Some discussions about the kind of problem I am trying to solve:
[https://community.openhab.org/t/which-item-updated-my-group-item/25273/2](https://community.openhab.org/t/which-item-updated-my-group-item/25273/2)
[https://community.openhab.org/t/more-efficient-rule/4730](https://community.openhab.org/t/more-efficient-rule/4730)

jmsjones:
Closing PR to move commits from master branch to feature branch.

PR #4523, opened 2017-11-09
Fix group function requires StringType with regex as parameter

### LINKED ISSUE #4460: Group Operator Count does not work as expected
I saw that group operator count was not documented.
Per comments it is expected to work this way : 
```
* Group:Number:COUNT(".") will count all items having a string state of one character
* Group:Number:COUNT("[5-9]") will count all items having a string state between 5 and 9
```

So I was about to give him some unit tests but I face an issue : 
- Per global logic (in ItemDTOMapper.mapFunction) group operators that accepts arguments are expected to return the same type than the operator but this does not work for Count as the operator is StringType while output is DecimalType.
I don't know how this happened to work on OH1 but ESH complains : 

Definition : 
```
Group:String:COUNT(".*") gSensorAirq	"Qualité Air [%d]"	<airquality>	(gSensors)


2017-10-27 08:00:24.128 [ERROR] [pse.smarthome.core.items.GenericItem] - Tried to set invalid state 5 (DecimalType) on item gSensorAirq of type StringItem, ignoring it
```
but value of the group seems good.

If I switch definition to Number : 
```
Group:Number:COUNT(".*") gSensorAirq	"Qualité Air [%d]"	<airquality>	(gSensors)

2017-10-27 09:45:05.068 [WARN ] [arthome.core.items.dto.ItemDTOMapper] - State '.*' is not valid for a group item with base type 'Number'
```

while changing label value format to %s :
```
Group:String:COUNT(".*") gSensorAirq	"Qualité Air [%s]"	<airquality>	(gSensors)
```
does not log error and does not work at all.

Has somebody an idea on how to address this ?

triller-telekom:
Fixes #4460

### DIFF
--- a/bundles/core/org.eclipse.smarthome.core.test/src/test/java/org/eclipse/smarthome/core/items/dto/ItemDTOMapperTest.java
+++ b/bundles/core/org.eclipse.smarthome.core.test/src/test/java/org/eclipse/smarthome/core/items/dto/ItemDTOMapperTest.java
@@ -11,12 +11,13 @@
  */
 package org.eclipse.smarthome.core.items.dto;
 
-import static org.hamcrest.CoreMatchers.instanceOf;
+import static org.hamcrest.CoreMatchers.*;
 import static org.junit.Assert.assertThat;
 
 import org.eclipse.smarthome.core.items.GroupFunction;
 import org.eclipse.smarthome.core.library.items.NumberItem;
 import org.eclipse.smarthome.core.library.types.ArithmeticGroupFunction;
+import org.eclipse.smarthome.core.library.types.StringType;
 import org.junit.Test;
 
 public class ItemDTOMapperTest {
@@ -34,6 +35,9 @@ public class ItemDTOMapperTest {
         GroupFunction gFunc = ItemDTOMapper.mapFunction(item1, gFuncDTO);
 
         assertThat(gFunc, instanceOf(ArithmeticGroupFunction.Count.class));
+        assertThat(gFunc.getParameters().length, is(1));
+        assertThat(gFunc.getParameters()[0], instanceOf(StringType.class));
+
     }
 
 }
--- a/bundles/core/org.eclipse.smarthome.core.test/src/test/java/org/eclipse/smarthome/core/items/dto/ItemDTOMapperTest.java
+++ b/bundles/core/org.eclipse.smarthome.core.test/src/test/java/org/eclipse/smarthome/core/items/dto/ItemDTOMapperTest.java
@@ -1,0 +1,39 @@
+/**
+ * Copyright (c) 2014-2017 by the respective copyright holders.
+ * All rights reserved. This program and the accompanying materials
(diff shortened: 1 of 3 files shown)


sjka:
Looks reasonable to me. 

However, this is yet another "edge case" which had to be found by somebody manually (thanks @clinique!) and now got fixed. But nothing is going to raise an alert once the next fix-for-some-other-edge-case would break this again. After #3314, #3332, #3493 (and I remember some more) I really hope you know what I'm missing here, right? 😢

clinique:
Maybe adding some unit tests and some lines in the documentation would be worse. @triller-telekom : thanks for this correction !

bundles/core/org.eclipse.smarthome.core/src/main/java/org/eclipse/smarthome/core/items/dto/ItemDTOMapper.java
```diff
@@ -132,8 +123,12 @@ public static GroupFunction mapFunction(Item baseItem, GroupFunctionDTO function
                 }
                 break;
             case "COUNT":
-                if (args.size() == 1) {
-                    groupFunction = new ArithmeticGroupFunction.Count(args.get(0));
+                State countParam = null;
+                if (function.params != null && function.params.length == 1) {
+                    countParam = new StringType(function.params[0]);
+                }
+                if (countParam != null) {
```
htreu: you might combine this into one `if` clause like

```
if (function.params != null && function.params.length == 1) {
    groupFunction = new ArithmeticGroupFunction.Count(new StringType(function.params[0]));
} else {
    LoggerFactory.getLogger....
}
```

htreu:
Since the `ArithmeticGroupFunctionTest` class is already there a test for the `COUNT` function is a now brainer ;-)

triller-telekom:
Thanks for the comments, I have created 2 test cases, one just for the count function and one for the ItemDTOMapper that tests "Group:Number:Count(".*hello.*")"

bundles/core/org.eclipse.smarthome.core.test/src/test/java/org/eclipse/smarthome/core/items/dto/ItemDTOMapperTest.java
```diff
@@ -0,0 +1,39 @@
...
+        // testing Group:Number:Count(".*hello.*")
+        NumberItem item1 = new NumberItem("item1");
+
+        GroupFunctionDTO gFuncDTO = new GroupFunctionDTO();
+        gFuncDTO.name = "COUNT";
+        gFuncDTO.params = new String[] { ".*hello.*" };
+
+        GroupFunction gFunc = ItemDTOMapper.mapFunction(item1, gFuncDTO);
+
+        assertThat(gFunc, instanceOf(ArithmeticGroupFunction.Count.class));
```
htreu: since the key change in this PR is to create a `StringType` parameter for the count function you might also check this here with `gFunc#getParameters`.

sjka:
Thank you!

PR #4526, opened 2017-11-09
Expose the item triggering the execution of a rule

jmsjones:
This PR adds the implicit variable "triggeringItem" to text file based rules. The variable is useful to identify which item triggered a rule when the rule can be triggered by more than one item. The implementation is consistent with the existing implicit variables "receivedCommand" and "previousState".

This change is only useful for rules with multiple conditions.
```
Item Switch01 changed or
Item Switch02 changed
```
The plan is to extend this further to support rules bound to groups (future PR).

Some discussions about the kind of problem I am trying to solve:
[https://community.openhab.org/t/which-item-updated-my-group-item/25273](https://community.openhab.org/t/which-item-updated-my-group-item/25273)
[https://community.openhab.org/t/more-efficient-rule/4730](https://community.openhab.org/t/more-efficient-rule/4730)

### DIFF
--- a/bundles/model/org.eclipse.smarthome.model.rule.runtime/src/org/eclipse/smarthome/model/rule/runtime/internal/engine/RuleEngineImpl.java
+++ b/bundles/model/org.eclipse.smarthome.model.rule.runtime/src/org/eclipse/smarthome/model/rule/runtime/internal/engine/RuleEngineImpl.java
@@ -200,7 +200,7 @@ public class RuleEngineImpl implements ItemRegistryChangeListener, StateChangeLi
         if (!starting && triggerManager != null) {
             Iterable<Rule> rules = triggerManager.getRules(CHANGE, item, oldState, newState);
 
-            executeRules(rules, oldState);
+            executeRules(rules, item, oldState);
         }
     }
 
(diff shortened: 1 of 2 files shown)


jmsjones:
The CI build failure occurred building the Paper UI. The Paper UI code is not affected by this PR. How can the build be restarted?

maggu2810:
I have started the CI build again.

sjka:
This addition clearly makes a lot of sense!

However, as the "old" rule engine implementation needs to be replaced by the "new" automation engine under the hood sooner or later. Hence I'm just not 100% sure if it is a good idea to extend the Rule-DSL-"API" at this point in time, before the migration. This _might_ make things more complicated, especially as it's something potentially many users will rely upon...

jmsjones:
I understand wanting to think ahead to the migration, however, unless this is taking place very soon I don't think we should hinder enhancements intended to address problems that users are having today. If it is feature that many users will rely upon, then we need to make sure the new engine supports it as well. I have started digging into the code for the automation engine and might be able to contribute there as well.

sjka:
Maybe I was not clear enough about the engine migration: Ideally existing rules written in todays `*.rules` file continue to work seamlessly on top of the new automation engine. And of course, _both_  engines must support that feature of course. The only question is how such information is going to be exposed, i.e. what implicit variables are there and how they are accessed. 
The more users are going to use it _now_ the more grumpy users you will create _in future_ if this cannot be done in a compatible way and therefore breaking their rule scripts silently. For this reason I would refrain from adding more stuff to the old rule scripts _right now_.

PR #4529, opened 2017-11-10
only reinitialize handler if anything in the configuration is changed

kaikreuzer:
This tremendously speeds up the reaction time of the Paper UI if you leave the Thing configuration page without doing any changes on it.


sjka:
The `org.eclipse.smarthome.core.thing.binding.BindingBaseClassesOSGiTest.assert configuration parameters are validated()` test now fails. It expects a `ConfigValidationException` which now doesn't get thrown anymore: The config was empty before and the test inserts `parameter: null`, i.e. the value _still_ is null -> no change -> no validation.

PR #4530, opened 2017-11-10
various lifecycle improvements and fixes on Tradfri

kaikreuzer:


sjka:
It indeed gets more robust - couldn't provoke the NPE anymore which I got easily before by restarting the bundle.

PR #4534, opened 2017-11-12
Basic UI: fix refresh of hidden/visible image/chart

### LINKED ISSUE #4492: Basic UI requesting chart data even for hidden charts
After tracing the chart servlet, I discovered that if in your sitemap, you have several charts defined with different periods and a visibility attribute set to only display the chart for the selected period, in fact Basic UI is requesting the server for all charts, meaning even for charts it does not render.
This leads to useless activity on server side.
Basic UI should consider the visibility property.

### LINKED ISSUE #4507: Basic UI: Refresh disabled forever for image/chart when they are first hidden
The JS code is disabling the refresh of image/chart that are hidden. It is done at this line:
https://github.com/eclipse/smarthome/blob/master/extensions/ui/org.eclipse.smarthome.ui.basic/web-src/smarthome.js#L445
It happens because `_t.image.clientWidth === 0` when the image/chart is hidden.
The problem is that the refresh is never re-enabled when the image/chart is then made visible.

lolodomo:
Fixes #4492
Fixes #4507
Ignore refresh when inappropriate

### DIFF
--- a/extensions/ui/org.eclipse.smarthome.ui.basic/src/main/java/org/eclipse/smarthome/ui/basic/internal/render/ChartRenderer.java
+++ b/extensions/ui/org.eclipse.smarthome.ui.basic/src/main/java/org/eclipse/smarthome/ui/basic/internal/render/ChartRenderer.java
@@ -34,6 +34,8 @@ public class ChartRenderer extends AbstractWidgetRenderer {
 
     private final Logger logger = LoggerFactory.getLogger(ChartRenderer.class);
 
+    private final static String URL_NONE_ICON = "images/none.png";
+
     @Override
     public boolean canRender(Widget w) {
         return w instanceof Chart;
@@ -77,8 +79,16 @@ public class ChartRenderer extends AbstractWidgetRenderer {
             if (chartTheme != null) {
                 chartUrl += "&theme=" + chartTheme;
             }
-            // add timestamp to circumvent browser cache
-            String url = chartUrl + "&t=" + (new Date()).getTime();
+            String url;
+            boolean ignoreRefresh;
+            if (!itemUIRegistry.getVisiblity(w)) {
+                url = URL_NONE_ICON;
+                ignoreRefresh = true;
+            } else {
+                // add timestamp to circumvent browser cache
+                url = chartUrl + "&t=" + (new Date()).getTime();
+                ignoreRefresh = false;
+            }
 
             String snippet = getSnippet("chart");
             snippet = preprocessSnippet(snippet, w);
@@ -92,6 +102,7 @@ public class ChartRenderer extends AbstractWidgetRenderer {
             snippet = StringUtils.replace(snippet, "%id%", itemUIRegistry.getWidgetId(w));
             snippet = StringUtils.replace(snippet, "%proxied_url%", chartUrl);
             snippet = StringUtils.replace(snippet, "%valid_url%", "true");
+            snippet = StringUtils.replace(snippet, "%ignore_refresh%", ignoreRefresh ? "true" : "false");
             snippet = StringUtils.replace(snippet, "%url%", url);
 
             sb.append(snippet);
--- a/extensions/ui/org.eclipse.smarthome.ui.basic/web-src/smarthome.js
+++ b/extensions/ui/org.eclipse.smarthome.ui.basic/web-src/smarthome.js
@@ -343,12 +343,12 @@
 			_t.visible = state;
 		};
 
-		_t.setValue = function(value, itemState) {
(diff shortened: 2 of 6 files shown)


lolodomo:
@resetnow : please confirm that I don't need anymore to provide the file smarthome.js in the web directory but only the one in the web-src directory ?

If you can review, that would be cool.

resetnow:
>I don't need anymore to provide the file smarthome.js in the web directory

That's correct.

>If you can review, that would be cool.

Later today.

extensions/ui/org.eclipse.smarthome.ui.basic/web-src/smarthome.js
```diff
@@ -431,22 +434,50 @@
...
+				_t.activateRefresh();
+			} else {
+				_t.formRow.classList.add(o.formRowHidden);
+				_t.disactivateRefresh();
+			}
+
+			_t.visible = state;
+		};
+
+		_t.disactivateRefresh = function() {
```
resetnow: You probably meant `deactivateRefresh`

extensions/ui/org.eclipse.smarthome.ui.basic/web-src/smarthome.js
```diff
@@ -431,22 +434,50 @@
...
+			if (interval !== undefined) {
+				clearInterval(interval);
+				interval = undefined;
```
resetnow: JS type are weird but I think here `null` is more suitable as it's used in the rest of the code

extensions/ui/org.eclipse.smarthome.ui.basic/src/main/java/org/eclipse/smarthome/ui/basic/internal/render/ChartRenderer.java
```diff
@@ -77,8 +77,13 @@ public boolean canRender(Widget w) {
             if (chartTheme != null) {
                 chartUrl += "&theme=" + chartTheme;
             }
-            // add timestamp to circumvent browser cache
-            String url = chartUrl + "&t=" + (new Date()).getTime();
+            String url;
+            if (!itemUIRegistry.getVisiblity(w)) {
+                url = "images/none.png";
```
resetnow: Feels a bit weird here, should be a static const maybe?

extensions/ui/org.eclipse.smarthome.ui.basic/web-src/smarthome.js
```diff
@@ -431,22 +434,50 @@
...
+			}
 
-		var
 			interval = setInterval(function() {
 				if (_t.image.clientWidth === 0) {
 					clearInterval(interval);
 					return;
 				}
-				_t.image.setAttribute("src", _t.url + "&t=" + Date.now());
+				if (_t.image.getAttribute("src").startsWith(_t.url)) {
```
resetnow: Couldn't figure out what this is for.

lolodomo: It is to avoid a refresh of the image using the proxied URL when the image was set with a raw data for example.

extensions/ui/org.eclipse.smarthome.ui.basic/web-src/smarthome.js
lolodomo: Ok I will change that.

extensions/ui/org.eclipse.smarthome.ui.basic/web-src/smarthome.js
lolodomo: Yes

extensions/ui/org.eclipse.smarthome.ui.basic/src/main/java/org/eclipse/smarthome/ui/basic/internal/render/ChartRenderer.java
lolodomo: Ok, why not.

extensions/ui/org.eclipse.smarthome.ui.basic/web-src/smarthome.js
lolodomo: Maybe I should rather set a variable to store the information whether we should refresh or not.

lolodomo:
@resetnow : I have taken into account all your comments except one that remains to be considered.

lolodomo:
@resetnow : everything is now ok and working well.

kaikreuzer:
@resetnow If you can approve the changes, we should be good to merge!

Issue #3327, opened 2017-04-28
Remove "bundleContext" (and similar) from BaseThingHandler


kaikreuzer:
When the Thing infrastructure was refactored to NOT have thing handlers registered as OSGi services anymore, a few parts where "forgotten" in the BaseThingHandler:

```
    protected ThingRegistry thingRegistry;
    protected ItemChannelLinkRegistry linkRegistry;
    protected BundleContext bundleContext;

    @SuppressWarnings("rawtypes")
    private ServiceTracker thingRegistryServiceTracker;
    @SuppressWarnings("rawtypes")
    private ServiceTracker linkRegistryServiceTracker;
```
I would claim that none of those should be in there - especially not as `protected` so that it becomes available to individual handler implementations.

Removing them will be an API breaking change, but therefore it should imho be done rather sooner than later.

sjka:
I agree we should do it rather sooner than later! 

I just remember a discussion with @sbussweiler about the bundleContext. There were some bindings using it, but nevertheless we should clean that up!

maggu2810:
> kaikreuzer:
> but therefore it should imho be done rather sooner than later.

> sjka
> I agree we should do it rather sooner than later!

Okay for me as well

htreu:
This is now duplicated by https://github.com/eclipse/smarthome/issues/5182.

PR #4539, opened 2017-11-13
the bundle context is a nullable member of the base thing handler

maggu2810:
The bundle context member of the base thing handler itself is nullable.
If the thing handler is active, the bundle context should not be null.
Add a method to get the bundle context (non-null) or a illegal state exception if this method is called if the member variable is null / the handler is not active.

As we rely on the most usage that the handler has been activated, we should rely on a non-null member. But as we cannot really ensure that the handler has been activated an IllegalStateException with a understandable message is much better then a NPE.

### DIFF
--- a/bundles/core/org.eclipse.smarthome.core.thing/src/main/java/org/eclipse/smarthome/core/thing/binding/BaseThingHandlerFactory.java
+++ b/bundles/core/org.eclipse.smarthome.core.thing/src/main/java/org/eclipse/smarthome/core/thing/binding/BaseThingHandlerFactory.java
@@ -45,8 +45,8 @@ public abstract class BaseThingHandlerFactory implements ThingHandlerFactory {
 
     protected @Nullable BundleContext bundleContext;
 
-    private Map<String, ServiceRegistration<ConfigStatusProvider>> configStatusProviders = new ConcurrentHashMap<>();
-    private Map<String, ServiceRegistration<FirmwareUpdateHandler>> firmwareUpdateHandlers = new ConcurrentHashMap<>();
+    private final Map<String, ServiceRegistration<ConfigStatusProvider>> configStatusProviders = new ConcurrentHashMap<>();
+    private final Map<String, ServiceRegistration<FirmwareUpdateHandler>> firmwareUpdateHandlers = new ConcurrentHashMap<>();
 
     private @Nullable ServiceTracker<ThingTypeRegistry, ThingTypeRegistry> thingTypeRegistryServiceTracker;
     private @Nullable ServiceTracker<ConfigDescriptionRegistry, ConfigDescriptionRegistry> configDescriptionRegistryServiceTracker;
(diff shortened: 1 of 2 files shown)


maggu2810:
Perhaps we could make the bundle context member of the base thing handler private some time...

sjka:
I actually would prefer to not expose the bundle context to bindings _at all_ (see https://github.com/eclipse/smarthome/issues/3327#issuecomment-298162492). The handler should not mess around with it. Therefore adding this protected method goes into the wrong direction.

maggu2810:
Hm, do you mix the thing handler and the thing handler factory?
The base thing handler is no service anymore, but the factory is still a component.

sjka:
Errr...., yes. Sorry!

bundles/core/org.eclipse.smarthome.core.thing/src/main/java/org/eclipse/smarthome/core/thing/binding/BaseThingHandlerFactory.java
```diff
@@ -92,6 +92,22 @@ protected void deactivate(ComponentContext componentContext) {
...
+     * @return the bundle context
+     * @throws IllegalArgumentException if the bundle thing handler is not active
+     */
+    protected BundleContext getBundleContext() {
+        final BundleContext bundleContext = this.bundleContext;
+        if (bundleContext != null) {
+            return bundleContext;
+        } else {
+            throw new IllegalStateException(
+                    "The bundle context is missing (it seems your thing handler is used but not active).");
```
sjka: "your _thing handler_" doesn't make sense then, right?

maggu2810: :wink:

extensions/binding/org.eclipse.smarthome.binding.tradfri/src/main/java/org/eclipse/smarthome/binding/tradfri/internal/TradfriHandlerFactory.java
```diff
@@ -79,14 +79,14 @@ protected synchronized void removeHandler(ThingHandler thingHandler) {
     private void registerDiscoveryService(TradfriGatewayHandler bridgeHandler) {
         TradfriDiscoveryService discoveryService = new TradfriDiscoveryService(bridgeHandler);
         discoveryService.activate();
-        this.discoveryServiceRegs.put(bridgeHandler.getThing().getUID(), bundleContext
+        this.discoveryServiceRegs.put(bridgeHandler.getThing().getUID(), getBundleContext()
```
kaikreuzer: afair, this has been copied from hue binding and we have similar code in other bindings as well - so it might make sense to adapt them likewise.

maggu2810: > so it might make sense to adapt them likewise

As already written above the bundle context member variable of the base class should be made private and the usage should be changed to use the get method. WDYT? What about a transition phase because of that API breakage? Do we need one?

PR #4546, opened 2017-11-14
Update coding guidelines

triller-telekom:


triller-telekom:
General stuff is because of things that came up during reviews, i.e. which happened more often so I decided it should be in our guidelines.

Regarding the exception for the null analysis please see our discussion here: https://github.com/eclipse/smarthome/issues/4321#issuecomment-342218053

docs/documentation/development/guidelines.md
```diff
@@ -26,13 +36,22 @@ Note that this list also serves as a checklist for code reviews on pull requests
 1. Packages that contain classes that are not meant to be used by other bundles should have "internal" in their package name.
 1. We are using [null annotations](https://wiki.eclipse.org/JDT_Core/Null_Analysis) from the Eclipse JDT project. Therefore every bundle should have an **optional** `Import-Package` dependency to `org.eclipse.jdt.annotation`.
 Classes should be annotated by `@NonNullByDefault` and return types, parameter types, generic types etc. are annotated with `@Nullable` only.
+For fields that store injected services, i.e. they will never be `null` in an OSGI framework, we annotate them with
```
kaikreuzer: This is only valid for mandatory and static services - you might also directly want to mention OSGi Declarative Services here (which is named in the points above as well).

docs/documentation/development/guidelines.md
```diff
@@ -26,13 +36,22 @@ Note that this list also serves as a checklist for code reviews on pull requests
...
+```
+
+to skip the `null` evaluation for these fields.
+Fields within `ThingHandler` classes that are initialized within the `initialize()` method may also be annotated like this, because the framework ensures that they will never be `null`.
 There is **no need** for a `@NonNull` annotation because it is set as default.
 The transition of existing classes could be a longer process but if you want to use nullness annotation in a class / interface you need to set the default for the whole class and annotate all types that differ from the default.
 Test classes do not have to be annotated.
 
-## B. OSGi Bundles
+## C. OSGi Bundles
```
kaikreuzer: Please avoid changing existing guideline identifiers as they are referenced in many places.

docs/documentation/development/guidelines.md
```diff
@@ -26,13 +36,22 @@ Note that this list also serves as a checklist for code reviews on pull requests
...
 Classes should be annotated by `@NonNullByDefault` and return types, parameter types, generic types etc. are annotated with `@Nullable` only.
+For fields that store injected services, i.e. they will never be `null` in an OSGI framework, we annotate them with
+
+```java
+@NonNullByDefault({})
+private MyService injectedService;
+```
+
+to skip the `null` evaluation for these fields.
+Fields within `ThingHandler` classes that are initialized within the `initialize()` method may also be annotated like this, because the framework ensures that they will never be `null`.
```
kaikreuzer: > because the framework ensures that they will never be `null`

This is not correct. It only (tries to) ensure that no methods are called before initialize() has been called. It's the duty of the handler implementation to ensure that such fields are set in initialize(). Note that I have seen many situations where fields are NOT set if the initialization fails and the Thing is set to OFFLINE - so this can actually be a tricky part.

docs/documentation/development/guidelines.md
```diff
@@ -26,13 +26,22 @@ Note that this list also serves as a checklist for code reviews on pull requests
 1. Packages that contain classes that are not meant to be used by other bundles should have "internal" in their package name.
 1. We are using [null annotations](https://wiki.eclipse.org/JDT_Core/Null_Analysis) from the Eclipse JDT project. Therefore every bundle should have an **optional** `Import-Package` dependency to `org.eclipse.jdt.annotation`.
 Classes should be annotated by `@NonNullByDefault` and return types, parameter types, generic types etc. are annotated with `@Nullable` only.
+For fields that store injected mandatory and static OSGi Declarative Services, i.e. they will never be `null` in an OSGI framework, we annotate them with
```
kaikreuzer: This formulation is not correct. I'd suggest to change it to:
> Fields that get a static and mandatory reference injected through OSGi Declarative Services can be annotated with

docs/documentation/development/guidelines.md
```diff
@@ -26,13 +26,22 @@ Note that this list also serves as a checklist for code reviews on pull requests
...
+
+```java
+@NonNullByDefault({})
+private MyService injectedService;
+```
+
+to skip the `null` evaluation for these fields.
```
kaikreuzer: -> in order to skip the `null`evaluation for them.

kaikreuzer:
Thanks!

PR #4549, opened 2017-11-14
Allow profiles to access a scheduler

### LINKED ISSUE #4544: Scheduler in Profiles
For some profile implementations it would make sense to have access to a scheduler (as pointed out in the [forum](https://www.eclipse.org/forums/index.php/t/1089956/)). Therefore it must be exposed somehow to the binding without allowing access to the ThreadPoolManager directly. 

My proposal would be using the `ProfileCallback` class and include something like `ProfileCallback.getScheduler()`.

sjka:
...by providing a ProfileContext.

It is designed in a way that it can be extended in future without
breaking the ProfileFactory API (again).

fixes #4544

### DIFF
--- a/bundles/core/org.eclipse.smarthome.core.thing/src/main/java/org/eclipse/smarthome/core/thing/internal/ProfileContextImpl.java
+++ b/bundles/core/org.eclipse.smarthome.core.thing/src/main/java/org/eclipse/smarthome/core/thing/internal/ProfileContextImpl.java
@@ -1,0 +1,43 @@
+/**
+ * Copyright (c) 2014-2017 by the respective copyright holders.
+ * All rights reserved. This program and the accompanying materials
+ * are made available under the terms of the Eclipse Public License v1.0
+ * which accompanies this distribution, and is available at
+ * http://www.eclipse.org/legal/epl-v10.html
+ */
+package org.eclipse.smarthome.core.thing.internal;
+
+import java.util.concurrent.ExecutorService;
+
+import org.eclipse.jdt.annotation.NonNullByDefault;
+import org.eclipse.smarthome.config.core.Configuration;
+import org.eclipse.smarthome.core.common.ThreadPoolManager;
+import org.eclipse.smarthome.core.thing.profiles.ProfileContext;
+
+/**
+ * {@link ProfileContext} implementation.
+ *
+ * @author Simon Kaufmann - initial contribution and API.
+ *
+ */
+@NonNullByDefault
+public class ProfileContextImpl implements ProfileContext {
+
+    private static final String THREAD_POOL_NAME = "profiles";
+    private final Configuration configuration;
+
+    public ProfileContextImpl(Configuration configuration) {
+        this.configuration = configuration;
+    }
+
+    @Override
+    public Configuration getConfiguration() {
+        return configuration;
+    }
+
+    @Override
+    public ExecutorService getExecutorService() {
+        return ThreadPoolManager.getPool(THREAD_POOL_NAME);
+    }
+
+}
--- a/bundles/core/org.eclipse.smarthome.core.thing.test/src/test/java/org/eclipse/smarthome/core/thing/internal/CommunicationManagerTest.java
+++ b/bundles/core/org.eclipse.smarthome.core.thing.test/src/test/java/org/eclipse/smarthome/core/thing/internal/CommunicationManagerTest.java
@@ -15,7 +15,6 @@ import java.util.Arrays;
 import java.util.stream.Collectors;
 import java.util.stream.Stream;
 
-import org.eclipse.smarthome.config.core.Configuration;
 import org.eclipse.smarthome.core.events.EventPublisher;
 import org.eclipse.smarthome.core.items.ItemRegistry;
 import org.eclipse.smarthome.core.items.events.ItemEventFactory;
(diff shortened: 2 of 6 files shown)


maggu2810:
```text
[ERROR] Failed to execute goal org.eclipse.tycho:tycho-compiler-plugin:1.0.0:compile (default-compile) on project org.eclipse.smarthome.core.thing: Compilation failure: Compilation failure: 

[ERROR] /home/travis/build/eclipse/smarthome/bundles/core/org.eclipse.smarthome.core.thing/src/main/java/org/eclipse/smarthome/core/thing/internal/profiles/SystemProfileFactory.java:[46] 

[ERROR] 	public class SystemProfileFactory implements ProfileFactory, ProfileAdvisor, ProfileTypeProvider {

[ERROR] 	             ^^^^^^^^^^^^^^^^^^^^

[ERROR] The type SystemProfileFactory must implement the inherited abstract method ProfileFactory.createProfile(ProfileTypeUID, ProfileCallback, ProfileContext)

[ERROR] /home/travis/build/eclipse/smarthome/bundles/core/org.eclipse.smarthome.core.thing/src/main/java/org/eclipse/smarthome/core/thing/internal/profiles/SystemProfileFactory.java:[58] 

[ERROR] 	public Profile createProfile(ProfileTypeUID profileTypeUID, ProfileCallback callback, Configuration configuration) {

[ERROR] 	               ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^

[ERROR] The method createProfile(ProfileTypeUID, ProfileCallback, Configuration) of type SystemProfileFactory must override or implement a supertype method
```

sjka:
It _did_ - happily sitting in my local workspace 🤦‍♂️ 
Oh well...

maggu2810:
I am fine with merging this PR.
Should we wait for some comments as it is API breaking?

sjka:
Afaik currently it's only @kungfoolfighting who would suffer but at the same time he requested it, so that shouldn't be troubling. 
Maybe leave it a couple of hours so others have a chance to comment.

kungfoolfighting:
Love it, thanks!

bundles/core/org.eclipse.smarthome.core.thing/src/main/java/org/eclipse/smarthome/core/thing/internal/ProfileContextImpl.java
```diff
@@ -0,0 +1,43 @@
...
+    }
+
+    @Override
+    public Configuration getConfiguration() {
+        return configuration;
+    }
+
+    @Override
+    public ExecutorService getExecutorService() {
+        return ThreadPoolManager.getPool(THREAD_POOL_NAME);
```
kungfoolfighting: When working with this change, i stumbled over this here. 
I am using a scheduledExcecutor in my version. 
Could you call "getScheduledPool" here?
It allows me to do easier scheduling of recurring tasks after initial delay, which is exactly what I need for dimming.

sjka: Yes, of course (-> #4555) - that was the idea actually...

maggu2810: I thought about it yesterday while reviewing your changes, too.
My assumption has been that you would like to prevent the usage of
* `scheduleAtFixedRate(Runnable command, long initialDelay, long period, TimeUnit unit)`
* `scheduleWithFixedDelay(Runnable command, long initialDelay, long delay, TimeUnit unit)`
in profiles. As it does not make sense to start repetitive stuff.

kungfoolfighting: I use the repetitive scheduling to transform long pressed into dimmer increase and decrease commands inside a profile. Should I not be doing that?

sjka: Yes, I think so too!
Using the scheduler for e.g. repetitively sending commands while a buttons is pressend perfectly makes sense. 
There just shouldn't be any fancy long-running stuff in there, eg. implementing a wake-up light that increases the brightness over 30min is better put into a rule.

maggu2810: @kungfoolfighting perhaps your profile should stop increasing if max is reached or decreasing if min is reached. Let's assume an wireless rocker (e.g. EnOcean) and a lost release button telegram. Your scheduled repetitive task should stop some time. :wink:

kungfoolfighting: @maggu2810 
There is no way to detect minimum or maximum, since the profile and channel are stateless.
But I have already handled the case of lost release telegrams. I schedule both a repeating task as well as a task that interrupts the repeating task after timeout.

sjka: > since the profile [...] are stateless

That's not correct - the same profile instance is used for all communications per link. You therefore can keep as much state as you need, if that helps.

kungfoolfighting: Sorry bad choice of words. I meant to say that trigger profiles are trigger profiles because the channels they support do not mirror the state of the items to which the are linked. As such I would not want to depend on the item state to make them work. 

But as I said: there isn't a real problem anyways. It all works with the change.

PR #4550, opened 2017-11-14
Corrected null annotations in thing package

triller-telekom:

### DIFF
--- a/bundles/core/org.eclipse.smarthome.core.thing/src/main/java/org/eclipse/smarthome/core/thing/binding/BaseBridgeHandler.java
+++ b/bundles/core/org.eclipse.smarthome.core.thing/src/main/java/org/eclipse/smarthome/core/thing/binding/BaseBridgeHandler.java
@@ -9,7 +9,9 @@ package org.eclipse.smarthome.core.thing.binding;
 
 import java.util.List;
 
-import org.eclipse.jdt.annotation.NonNull;
+import org.eclipse.jdt.annotation.DefaultLocation;
+import org.eclipse.jdt.annotation.NonNullByDefault;
+import org.eclipse.jdt.annotation.Nullable;
 import org.eclipse.smarthome.core.thing.Bridge;
 import org.eclipse.smarthome.core.thing.Thing;
 import org.eclipse.smarthome.core.thing.ThingUID;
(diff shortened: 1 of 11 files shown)


bundles/core/org.eclipse.smarthome.core.thing/src/main/java/org/eclipse/smarthome/core/thing/binding/BaseBridgeHandler.java
```diff
@@ -25,12 +27,14 @@
  * @author Dennis Nobel - Initial contribution
  * @author Stefan Bußweiler - Added implementation of BridgeHandler interface
  */
+@NonNullByDefault({ DefaultLocation.PARAMETER, DefaultLocation.RETURN_TYPE, DefaultLocation.ARRAY_CONTENTS,
```
martinvw: Just my two cents, do we really need / want to define all these things every time...

triller-telekom: Not everywhere, I have documented this exception in an update PR to our coding guidelines: https://github.com/eclipse/smarthome/pull/4546/files#diff-24ba7e763168d1631a43ebdef19bcd5dR39

The problem are the OSGI services which are injected and stored as a field. Thus we have to exclude fields from the analysis of classes that retrieve services via injection.

martinvw: 😢 imho that is pretty ugly

maggu2810: Can't we ask for another JDT nullness annotation to mark that a variable be ignored by the nullness check at all (so, not allowed to be used for arguments and return types, but for fields only)? Example `@NullnessCheckIgnore`?
We only need compiler and IDE support if the annotation itself could be configured (so no need for a new annotation bundle).

htreu: > not allowed to be used for arguments and return types, but for fields only

FTR: there are unfortunately rare occasions where injected services are used as return values. In this situations a null check would of cause circumvent an compiler error.

maggu2810:
See https://github.com/eclipse/smarthome/issues/4321#issuecomment-344996256

Issue #4164, opened 2017-08-30
Switch to EPLv2 and introduce new license headers


kaikreuzer:
The EPLv2 has been released, see https://www.eclipse.org/org/press-release/20170829eplv2.php.

According to the press release, Eclipse projects are expected to migrate to that new version.

We have a long standing discussion about needing new license headers, which move the list of copyright holders away from the header into NOTICE files.

kaikreuzer:
Here is the handbook on the new license header: https://www.eclipse.org/projects/handbook/#ip-copyright-headers

My suggestion is to go for the "more generic copyright header" (as this was specifically introduced upon our request).

We will have to address this before doing a 0.9.0 release.

maggu2810:
> The {date} is either a year or a range of years with the first and last years of the range separated by a comma (e.g. “2004” or “2004, 2017”). The first year is when the contents of the file were first created and the last year is the last year is when the contents were last modified.

"range separated by a comma", so we also need to change "2014-2017" to "2014, 2017", too.

kaikreuzer:
FTR, next after #4552 is to replace the about.thml files with a NOTICE file.
We will have to decide on the exact content of those files, which will be done in a subsequent PR before the 0.9.0 release.

PR #4552, opened 2017-11-14
Update license to EPL-2.0

htreu:
This addresses #4164.

All file headers are updated to the "generic" copyright header, the LICENSE file copies the EPL-2.0 text and all SUA occurrences have been replaced by a copy of the EPL-2.0 [as stated here](https://www.eclipse.org/projects/handbook/#legaldoc-plugins).

All about.html files will be replaced in a separate PR by NOTICE files which will include the legal entity copyright holders along with the third party content licenses.


htreu:
Of cause the static code analysis is not yet able to handle the new license headers, hence the travis failure.
Should be update the tool first? it would then have to be able to check both versions. I suggest to first get this straight, then update the static analysis, then merge this PR.
wdyt?

kaikreuzer:
How about simply adapting the [check configuration](https://github.com/eclipse/smarthome/blob/master/tools/static-code-analysis/checkstyle/ruleset.properties#L2)?

htreu:
My comment was about synchronisation of the merge of this PR and the static-code-analysis. I will do the PR for static-code-analysis and use the next version of it (0.4.0) here.

kaikreuzer:
Thanks - reviewing this PR, I just came across one small issue: 

The license headers of the (archetype-generated) files for the binding skeleton now also refer to the "Contributors to the Eclipse Foundation" instead of the "respective copyright holders".

This is now only correct for code that is meant to end up at ESH, but not for all others (e.g. bindings in openhab2-addons).

I guess we should add a placeholder in those headers and use [the `vendorName` property](https://github.com/openhab/openhab2-addons/blob/master/addons/binding/create_openhab_binding_skeleton.sh#L18) to fill in the correct string.

tools/archetype/binding.test/src/main/resources/archetype-resources/src/test/java/__bindingIdCamelCase__HandlerTest.java
```diff
@@ -1,7 +1,12 @@
 #set( $dt = $package.getClass().forName("java.util.Date").newInstance() )
 #set( $year = $dt.getYear() + 1900 )
+#if( $vendorName == "Eclipse.org/SmartHome" )
```
kaikreuzer: The vendorName is currently not set when generating a skeleton in ESH, see https://github.com/eclipse/smarthome/blob/master/extensions/binding/create_binding_skeleton.sh, so this won't actually work. Maybe you should simply check whether it is set or not?

tools/archetype/binding.test/src/main/resources/archetype-resources/src/test/java/__bindingIdCamelCase__HandlerTest.java
```diff
@@ -1,7 +1,12 @@
...
+    #set( $copyright = "Contributors to the Eclipse Foundation" )
+#else
+    #set( $copyright = "by the respective copyright holders." )
+#end
 /**
- * Copyright (c) 2014,${year} Contributors to the Eclipse Foundation
+ * Copyright (c) 2014,${year} ${copyright}
```
kaikreuzer: As we are anyhow on it: Would it be possible to add an optional property startYear and insert this here? If not set, it could default to 2014.

tools/archetype/binding.test/src/main/resources/archetype-resources/src/test/java/__bindingIdCamelCase__HandlerTest.java
htreu: I beg to differ ;-) See https://github.com/eclipse/smarthome/blob/master/tools/archetype/binding/src/main/resources/META-INF/maven/archetype-metadata.xml#L8 and https://github.com/eclipse/smarthome/blob/master/tools/archetype/binding.test/src/main/resources/META-INF/maven/archetype-metadata.xml#L8

tools/archetype/binding.test/src/main/resources/archetype-resources/src/test/java/__bindingIdCamelCase__HandlerTest.java
htreu: sure, hold on.

tools/archetype/binding.test/src/main/resources/archetype-resources/src/test/java/__bindingIdCamelCase__HandlerTest.java
kaikreuzer: Ok, you won 🥇

PR #4553, opened 2017-11-14
Profile type builder

sjka:
I wonder whether a builder pattern would be more convenient (in terms of implementation as well as future extensibility) for creating profile types (see [SystemProfiles.java](https://github.com/eclipse/smarthome/compare/master...sjka:profileTypeBuilder?expand=1#diff-2eea43cbaf238cd83f8550390ec9098b) for the effect).

cc: @kungfoolfighting, @maggu2810

### DIFF
--- a/bundles/core/org.eclipse.smarthome.core.thing/src/main/java/org/eclipse/smarthome/core/thing/internal/profiles/StateProfileTypeImpl.java
+++ b/bundles/core/org.eclipse.smarthome.core.thing/src/main/java/org/eclipse/smarthome/core/thing/internal/profiles/StateProfileTypeImpl.java
@@ -1,0 +1,51 @@
+/**
+ * Copyright (c) 2014-2017 by the respective copyright holders.
+ * All rights reserved. This program and the accompanying materials
+ * are made available under the terms of the Eclipse Public License v1.0
+ * which accompanies this distribution, and is available at
+ * http://www.eclipse.org/legal/epl-v10.html
+ */
+package org.eclipse.smarthome.core.thing.internal.profiles;
+
+import java.util.Collection;
+
+import org.eclipse.jdt.annotation.NonNull;
+import org.eclipse.jdt.annotation.NonNullByDefault;
+import org.eclipse.smarthome.core.thing.profiles.ProfileTypeUID;
+import org.eclipse.smarthome.core.thing.profiles.StateProfileType;
+
+/**
+ * Default implementation of a {@link StateProfileType}.
+ *
+ * @author Simon Kaufmann - initial contribution and API.
+ *
+ */
+@NonNullByDefault
+public class StateProfileTypeImpl implements StateProfileType {
+
+    private final ProfileTypeUID profileTypeUID;
+    private final String label;
+    private final Collection<String> supportedItemTypes;
+
+    public StateProfileTypeImpl(ProfileTypeUID profileTypeUID, String label, Collection<String> supportedItemTypes) {
+        this.profileTypeUID = profileTypeUID;
+        this.label = label;
+        this.supportedItemTypes = supportedItemTypes;
+    }
+
+    @Override
+    public @NonNull ProfileTypeUID getUID() {
+        return profileTypeUID;
+    }
+
+    @Override
+    public @NonNull Collection<@NonNull String> getSupportedItemTypes() {
+        return supportedItemTypes;
+    }
+
+    @Override
+    public @NonNull String getLabel() {
+        return label;
+    }
+
+}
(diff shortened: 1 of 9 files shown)


kaikreuzer:
You'll need to rebase...

maggu2810:
Please wait, I will test something in front of merging.

bundles/core/org.eclipse.smarthome.core.thing/src/main/java/org/eclipse/smarthome/core/thing/profiles/ProfileTypeBuilder.java
```diff
@@ -0,0 +1,145 @@
...
+    }
+
+    /**
+     * Create a profile type instance with the previously given parameters.
+     *
+     * @return the according subtype of
+     */
+    @SuppressWarnings("unchecked")
+    public T build() {
+        final String lbl = label;
```
kungfoolfighting: Why is the label not a constructor parameter to the builder, if it apparently is mandatory for the construction of a profiletype?

sjka: The whole intention of the builder is to not end up in a constructor-overload-mess when adding additional parameters later on. Therefore I put only the identifier there. 
That was the reason but I don't have any strong feelings about it. In the end it's a matter of taste, I guess...

bundles/core/org.eclipse.smarthome.core.thing/src/main/java/org/eclipse/smarthome/core/thing/internal/profiles/StateProfileTypeImpl.java
```diff
@@ -0,0 +1,51 @@
...
+    private final Collection<String> supportedItemTypes;
+
+    public StateProfileTypeImpl(ProfileTypeUID profileTypeUID, String label, Collection<String> supportedItemTypes) {
+        this.profileTypeUID = profileTypeUID;
+        this.label = label;
+        this.supportedItemTypes = supportedItemTypes;
+    }
+
+    @Override
+    public @NonNull ProfileTypeUID getUID() {
```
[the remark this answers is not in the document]
sjka: Yes, sure, I forgot to clean them up... Thanks!

maggu2810:
@sjka I created a PR for your branch yesterday.

bundles/core/org.eclipse.smarthome.core.thing/src/main/java/org/eclipse/smarthome/core/thing/profiles/ProfileTypeBuilder.java
maggu2810: @sjka I agree that we should prevent a "constructor-overload-mess". IMHO there should be one private constructor that consumes all mandatory fields. All mandatory fields should be supplied already by the static factory methods.

sjka:
Just seen it - I was fighting my mail pile in reverse order ;-)

bundles/core/org.eclipse.smarthome.core.thing/src/main/java/org/eclipse/smarthome/core/thing/profiles/TriggerProfileType.java
```diff
@@ -23,13 +22,8 @@
...
 
     /**
-     * Constant for matching *ANY* channel type.
-     */
-    Collection<ChannelTypeUID> ANY_CHANNEL_TYPE = new ArrayList<>(0);
-
-    /**
      *
-     * @return a collection of ChannelTypeUIDs or {@link #ANY_CHANNEL_TYPE}.
+     * @return a collection of ChannelTypeUIDs (may be empty all are supported).
```
maggu2810: I am not a native speaker but isn't there a "if" missing (may be empty if all are supported)?

bundles/core/org.eclipse.smarthome.core.thing/src/main/java/org/eclipse/smarthome/core/thing/profiles/ProfileTypeBuilder.java
sjka: Yes, that what I meant (regardless of whether they are static factory methods or the constructor itself). Nevertheless, I count it as being overruled here. Will amend it accordingly.

maggu2810: I like to understand the different approaches.

@sjka if we introduce a new mandatory field some time and do not change the "construction" (regardless if new or a static method) methods, the usage will still compile and seems to be valid. So, no error on compilation. But as soon as the build method checks the mandatory fields at runtime all unchanged (still successful compiled) usage will run into errors at runtime.

Wouldn't it better to catch such errors (they will occur) already on compile time?

To prevent big fat construction methods there are also a step builder pattern (at least I found it here "https://stackoverflow.com/a/35847487").

sjka: I obviously didn't think it through completely - of course it would be better to fail loudly at compile time! It's an "incompatible" change, therefore implementors must notice it rather sooner than later.

You convinced me. Case closed. 😄

maggu2810:
I assume I am not allowed to merge anymore as it contains a commit of mine.

PR #4563, opened 2017-11-16
Added system rocker channels and associated system events.

### LINKED ISSUE #4541: System channels for 2-button trigger channels (a.k.a. rockers)
We currently have the following two channels:

| Channel | Events |
|:---|:---|
| system.button | SHORT_PRESSED & LONG_PRESSED & DOUBLE_PRESSED |
| system.rawbutton | PRESSED & RELEASED |

As discussed in the context of the [enocean binding](https://community.openhab.org/t/enocean-bindings-opencean-or-aleoncean/12493/58) but certainly also relevant for others, it would make sense to also have something like `system.rocker` and `system.rawrocker` for 2-button controls with their corresponding events.

kungfoolfighting:
Closes #4541

Signed-off-by: Jan Kemmler <kickass_kemmler@gmx.de>

### DIFF
--- a/bundles/core/org.eclipse.smarthome.core.thing/ESH-INF/i18n/DefaultSystemChannels_de.properties
+++ b/bundles/core/org.eclipse.smarthome.core.thing/ESH-INF/i18n/DefaultSystemChannels_de.properties
@@ -4,3 +4,4 @@ channel-type.system.battery-level.label = Batterieladung
 channel-type.system.trigger.label = Auslöser
 channel-type.system.rawbutton.label = Einfache Taste
 channel-type.system.button.label = Taste
+channel-type.system.rawrocker.label = Wipptaster
--- a/bundles/core/org.eclipse.smarthome.core.thing/src/main/java/org/eclipse/smarthome/core/thing/DefaultSystemChannelTypeProvider.java
+++ b/bundles/core/org.eclipse.smarthome.core.thing/src/main/java/org/eclipse/smarthome/core/thing/DefaultSystemChannelTypeProvider.java
@@ -102,6 +102,18 @@ public class DefaultSystemChannelTypeProvider implements ChannelTypeProvider {
                     new EventOption(CommonTriggerEvents.LONG_PRESSED, null))),
             null);
 
+    /**
+     * System wide trigger {@link ChannelType} which triggers "DIR1_PRESSED", "DIR1_RELEASED", "DIR2_PRESSED" and
+     * "DIR2_RELEASED" events.
+     */
+    public static final ChannelType SYSTEM_RAWROCKER = new ChannelType(new ChannelTypeUID("system:rawrocker"), false,
+            null, ChannelKind.TRIGGER, "Raw rocker switch", null, null, null, null,
+            new EventDescription(Arrays.asList(new EventOption(CommonTriggerEvents.DIR1_PRESSED, null),
+                    new EventOption(CommonTriggerEvents.DIR1_RELEASED, null),
+                    new EventOption(CommonTriggerEvents.DIR2_PRESSED, null),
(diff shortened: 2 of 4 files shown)


maggu2810:
We still need to know if we would like to have a rawrocker and a rocker or a rawrocker only.

kungfoolfighting:
Yeah, but that could easily be another PR, right? 
there is no need to create lots of channels that are not needed yet. I would always do these things incrementally and then provide what is needed when it is needed. That was my plan here at least.

kungfoolfighting:
What is the procedure for this PR?
I need this PR to be merged for the next PR that is based on this one.
Is there anything I still need to do?
The ip-validation failed initially, but I have since registered this ECA for my email address.

bundles/core/org.eclipse.smarthome.core.thing/src/main/java/org/eclipse/smarthome/core/thing/DefaultSystemChannelTypeProvider.java
```diff
@@ -102,6 +102,18 @@
                     new EventOption(CommonTriggerEvents.LONG_PRESSED, null))),
             null);
 
+    /**
+     * System wide trigger {@link ChannelType} which triggers "DIR1_PRESSED", "DIR1_RELEASED", "DIR2_PRESSED" and
+     * "DIR2_RELEASED" events.
+     */
+    public static final ChannelType SYSTEM_RAWROCKER = new ChannelType(new ChannelTypeUID("system:rawrocker"), false,
+            null, ChannelKind.TRIGGER, "Raw rocker switch", null, null, null, null,
```
sjka: "Raw rocker **button**"

bundles/core/org.eclipse.smarthome.core.thing/ESH-INF/i18n/DefaultSystemChannels_de.properties
```diff
@@ -4,3 +4,4 @@ channel-type.system.battery-level.label = Batterieladung
 channel-type.system.trigger.label = Ausl�ser
 channel-type.system.rawbutton.label = Einfache Taste
 channel-type.system.button.label = Taste
+channel-type.system.rawrocker.label = Wipptaster
```
sjka: It would be cool to also immediately amend the french translation...
@lolodomo heeeeelp!?!

sjka:
> The ip-validation failed initially, but I have since registered this ECA for my email address.

Yep, thanks for doing so. Don't worry about the IP validation flag now - it double-checked manually that your ECA ended up correctly in the database, so we can simply overrule the automatic check.

bundles/core/org.eclipse.smarthome.core.thing/ESH-INF/i18n/DefaultSystemChannels_de.properties
lolodomo: For that I should understand what means raw rocker lol

kungfoolfighting: A rocker switch is a switch that you can press on two sides, for on the top and the bottom.
It then has a rocking motion, that is why it is called that.
It could also be called see-saw switch :)

Something like this: https://images.homedepot-static.com/productImages/8f485b21-6d89-4d7d-a32f-3e1c14177c06/svn/white-leviton-switches-wss0s-d0w-64_1000.jpg

lolodomo: Ok I see now. You can call it in French: Bouton à bascule

Issue #4465, opened 2017-10-29
Introduce an API for serial communication


kaikreuzer:
We so far do not have any way to allow ESH bindings access to serial ports. Within openHAB, RXTX is used, which is not an option for ESH due to its LGPL license, so any binding that uses serial 
and which is developed on the openHAB side is automatically incompatible with ESH. Furthermore, RXTX is not maintained anymore and seems to have several issues, so a different implementation might be helpful.

Some discussion can be found [here](https://github.com/openhab/openhab2-addons/pull/2246#issuecomment-340220633) - let's follow up on that.

kaikreuzer:
>> And as pointed out, most of the other libs use exactly the same API and merely moved it to their own namespace in order to avoid any license/IP issues.

> I'm pretty sure there is no license/IP issue here but it's more about them not being compliant to the API. Since afaik sun / oracle created such api/interfaces for others to implement not to restrict it.

Seeing is believing. The issue is that I cannot find any license information about javax.comm - and only a written license will give you any permission, otherwise the strict copyright laws are in place.
But as soda.dk includes the javax.comm APIs and is licensed under the EPL, this might be the safest way to get hold of it for ESH.

maggu2810:
As soda.dk serial's implementation that is using java.comm API is part of Kura already, shouldn't the Eclipse IP team already have had a look on this one?

sjka:
Sure, CQ 8156 (by Kura) even was PBing the original CQ 1330 (by OHF).

kaikreuzer:
> RXTX is used, which is not an option for ESH due to its LGPL license

FTR, as mentioned [here](https://github.com/eclipse/smarthome/pull/3633#issuecomment-343774064): RXTX actually IS an option, since we have an approved CQ 9215 for nrjavaserial as a works-with dependency. Not wanting to judge whether the API itself is a good option or not, but it is the one which has been used extensively so far.

Another option that should be mentioned is OpenJDKs [Device I/O](http://openjdk.java.net/projects/dio/). While it was specifically meant as a modern API for hardware access on SBCs, the project is almost dead and hardly anything has happened in the past 2 years, so it might not be a good pick either :-/

aploese:
You may have a look at https://github.com/aploese/spsw
Ready build jars are in the snapshot repo (https://oss.sonatype.org/content/repositories/snapshots)
Its OSGi ready...(Service interface and provider implementation)

aploese:
[...]

PR #4568, opened 2017-11-16
introduced config option provider for serial ports

kaikreuzer:
This PR adds an optional bundle, which can provide a list of available serial ports as configuration parameter options.

FTR, I am aware that https://github.com/eclipse/smarthome/issues/4465 has not been decided yet, but I think this feature is pretty useful and the implementation so small that it can be easily adapted once we reach a consensus on #4465.


htreu:
@kaikreuzer please update the documentation at `docs/documentation/development/bindings/xml-reference.md` and append the new context.

htreu:
And karaf feature validation is not happy.

kaikreuzer:
> please update the documentation at docs/documentation/development/bindings/xml-reference.md

You mean [this is not good enough](https://github.com/eclipse/smarthome/pull/4568/files#diff-8e67ab95e48e5025e43751c8d5255b9e)?

Issue #4369, opened 2017-10-03
Test ESH with Java 9


kaikreuzer:
Just as a reminder to ourselves: Now that Java 9 is out, we should make sure that ESH runs smoothly on it as well. (Note that this does not mean that we will move the code base from 8 to 9 any time soon - we should simply try to increase the number of JVMs that ESH can run on).

lewie:
A nightly Build openHAB 2.2, started in Debian 7 jessie with oracle 9 SDK:

```
root@core ~# sh /data/OH/start.sh
Launching the openHAB runtime...
/data/OH/runtime/bin/karaf: 198: [: Illegal number:
/data/OH/runtime/bin/karaf: 198: [: Illegal number:

                          __  _____    ____
  ____  ____  ___  ____  / / / /   |  / __ )
 / __ \/ __ \/ _ \/ __ \/ /_/ / /| | / __  |
/ /_/ / /_/ /  __/ / / / __  / ___ |/ /_/ /
\____/ .___/\___/_/ /_/_/ /_/_/  |_/_____/
    /_/                        2.2.0-SNAPSHOT
                               Build #1054

Hit '<tab>' for a list of available commands
and '[cmd] --help' for help on a specific command.
Hit '<ctrl-d>' or type 'system:shutdown' or 'logout' to shutdown openHAB.

openhab> 
WARNING: An illegal reflective access operation has occurred
WARNING: Illegal reflective access by org.eclipse.jetty.util.BufferUtil (file:/data/OH22/userdata/cache/org.eclipse.osgi/87/0/bundleFile) to field java.nio.MappedByteBuffer.fd
WARNING: Please consider reporting this to the maintainers of org.eclipse.jetty.util.BufferUtil
WARNING: Use --illegal-access=warn to enable warnings of further illegal reflective access operations
WARNING: All illegal access operations will be denied in a future release
log:tail
18:44:57.923 [INFO ] [i.dashboard.internal.DashboardService] - Started dashboard at http://192.168.0.174:8080
18:44:57.936 [INFO ] [i.dashboard.internal.DashboardService] - Started dashboard at https://192.168.0.174:8443
18:44:58.059 [INFO ] [.basic.internal.servlet.WebAppServlet] - Started Basic UI at /basicui/app
18:44:58.072 [INFO ] [marthome.ui.paper.internal.PaperUIApp] - Started Paper UI at /paperui
[...]

PR #4569, opened 2017-11-16
Fixes on UPnP infrastucture

### LINKED ISSUE #4531: [Sonos] UPnP action invocation results in an empty result, which sets the Thing to OFFLINE
UPnP is broken !
The following line is failing (returning an empty result):
https://github.com/eclipse/smarthome/blob/fee1c67521d3d2b84bdf8ec82552a50bfad6faf5/extensions/binding/org.eclipse.smarthome.binding.sonos/src/main/java/org/eclipse/smarthome/binding/sonos/internal/handler/ZonePlayerHandler.java#L837
The result with the Sonos binding is that the thing is set to OFFLINE.
I have 2 Sonos and strangely it happens only for one of my 2 Sonos.

It looks like a critical bug in the new library version we are relying on (jUPnP).

I switch back to snapshot 1065 immediately.
@kaikreuzer : I think you should announce in the forum to absolutely avoid the snapshot 1078 for all users using UPnP, until a fix is found.

kaikreuzer:
- Improved UPnP presence handling in UpnpIOService
- Switched from DS XML to annotations
- Removed unneeded dependency in WeMo tests
- updated Jetty library to version 9.3.15

Fixes #4531


kaikreuzer:
I'll have a look at the failing WeMo tests tomorrow - no energy for that left today 😲

kaikreuzer:
Fixed all tests and added some other improvements - let's touch wood that Travis succeeds :-)

sjka:
> updated Jetty library to version 9.3.15

I confirm: sonos works again!

sjka:
While talking about jetty upgrades... Did you still have https://github.com/eclipse/smarthome/issues/4369#issuecomment-333996110 in mind when picking the version? Maybe it would make sense to skip directly ahead to 9.3.16?

kaikreuzer:
I would leave Java9 out of the picture for now - we can upgrade once we want to support this.
For the moment, I prefer sticking to older patch levels as e.g. Pax-Web and Karaf both are still on those as well.

sjka:
Fine by me. 

Code looks good as far as I can tell, and I haven't managed yet to break the UPnP stuff anymore.

Dan1001:
Fantastic! Will this then go into the next snapshot version?

extensions/binding/org.eclipse.smarthome.binding.sonos/src/main/java/org/eclipse/smarthome/binding/sonos/internal/handler/ZonePlayerHandler.java
```diff
@@ -2889,7 +2878,10 @@ public String getCurrentZoneName() {
     @Override
     public void onStatusChanged(boolean status) {
         if (status) {
-            updateStatus(ThingStatus.ONLINE);
+            if (getThing().getStatus() != ThingStatus.ONLINE) {
+                updateStatus(ThingStatus.ONLINE);
```
lolodomo: I think you should let the polling job set the thing ONLINE through the call to updatePlayerState. That way the thing would be set to ONLINE in one unique place.
By the way, that is a good idea to run the job immediately after the device is found through UPnP, it can make the thing going ONLINE a little faster.

lolodomo:
@kaikreuzer : I added a comment relative to your code in method onStatusChanged.

extensions/binding/org.eclipse.smarthome.binding.sonos/src/main/java/org/eclipse/smarthome/binding/sonos/internal/handler/ZonePlayerHandler.java
lolodomo: At the same time, that means you could generate a conflict between this one time job and the regular scheduled job, in particular each one requesting GENA subscriptions.

lolodomo: As there is a lock implemented for GENA subscriptions, it should be finally probably ok.

PR #4571, opened 2017-11-16
Document source field in AbstractEvent class

### LINKED ISSUE #727: AbstractEvent.source usage not consistent
All,
It is not really a bug as such, but the usage of the AbstractEvent.source throughout the core is not 100% consistent. Sometimes this is a hardcoded string, sometimes this is an UID, and often, it is simply 'null'

What should/would be the convention to use for this source field? getClass().getSimpleName() of the calling class?

triller-telekom:
Fixes #727


kaikreuzer:
No, sorry, this does not address #727 as expected. The issue is about using a consistent naming/content scheme for the source field. Saying it has to be "pre-negotiated" does not help us here.

PR #4574, opened 2017-11-16
annotate some provider interfaces

maggu2810:
Sometimes I would like to add one annotation, I end up with changing millions (okay, that's not really the correct number, but a lot of) files. :wink: 
So, let's do small steps.


sjka:
They're virulent 😉

PR #4576, opened 2017-11-16
remove mandatory reference from nullness analysis

maggu2810:
Related to https://github.com/eclipse/smarthome/issues/4321#issuecomment-344996256

### DIFF
--- a/bundles/core/org.eclipse.smarthome.core.thing/src/main/java/org/eclipse/smarthome/core/thing/i18n/ThingTypeI18nLocalizationService.java
+++ b/bundles/core/org.eclipse.smarthome.core.thing/src/main/java/org/eclipse/smarthome/core/thing/i18n/ThingTypeI18nLocalizationService.java
@@ -13,7 +13,6 @@ import java.util.Locale;
 
 import org.eclipse.jdt.annotation.NonNull;
 import org.eclipse.jdt.annotation.NonNullByDefault;
-import org.eclipse.jdt.annotation.Nullable;
 import org.eclipse.smarthome.core.i18n.TranslationProvider;
 import org.eclipse.smarthome.core.thing.type.BridgeType;
 import org.eclipse.smarthome.core.thing.type.ChannelDefinition;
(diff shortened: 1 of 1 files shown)


sjka:
Looks nice - the compiler though has a different opinion on that:

```
[ERROR] Failed to execute goal org.eclipse.tycho:tycho-compiler-plugin:1.0.0:compile (default-compile) on project org.eclipse.smarthome.core.thing: Compilation failure: Compilation failure: 
[ERROR] /home/travis/build/eclipse/smarthome/bundles/core/org.eclipse.smarthome.core.thing/src/main/java/org/eclipse/smarthome/core/thing/i18n/ThingTypeI18nLocalizationService.java:[41] 
[ERROR] 	private ThingTypeI18nUtil thingTypeI18nUtil;
[ERROR] 	                          ^^^^^^^^^^^^^^^^^
[ERROR] The @NonNull field thingTypeI18nUtil may not have been initialized
[ERROR] /home/travis/build/eclipse/smarthome/bundles/core/org.eclipse.smarthome.core.thing/src/main/java/org/eclipse/smarthome/core/thing/i18n/ThingTypeI18nLocalizationService.java:[49] 
[ERROR] 	this.thingTypeI18nUtil = null;
[ERROR] 	                         ^^^^
[ERROR] Null type mismatch: required '@NonNull ThingTypeI18nUtil' but the provided value is null
[ERR
```

maggu2810:
@triller-telekom Shouldn't be the IDE and the Maven nullness check settings be equal?

maggu2810:
@sjka Sorry about that, I tested with the IDE only.

sjka:
Nah, all good! If we can get it to work like that, I think this would be a really sweet solution for these I-know-they'll-never-be-null-under-normal-circumstances fields!

triller-telekom:
@maggu2810 the settings are "somehow" equal :P Unfortunately the settings int he Eclipse IDE UI do not directly map 1:1 to command line settings for the JDT compiler. But I will have a look at it...

maggu2810:
@triller-telekom You don't need to check that.
@triller-telekom @sjka We get the error because of the JDT compiler version that is used by the Tycho compiler plugin version 1.0.0. It's not such up-to-date as the one in the IDE.
After the next tycho release it should work -- at least it do with 1.1.0-SNAPSHOT.

Anyone knows their release dates?

triller-telekom:
Seems like they are working to get java 9 support into tycho and have some trouble to get it wot work: https://dev.eclipse.org/mhonarc/lists/tycho-user/msg07782.html

kaikreuzer:
With #4604 being merged, I expect Travis to succeed now - relaunching it.

PR #4585, opened 2017-11-18
added dependency="true" where missing

kaikreuzer:

### DIFF
--- a/features/karaf/esh-tp/src/main/feature/feature.xml
+++ b/features/karaf/esh-tp/src/main/feature/feature.xml
@@ -79,7 +79,7 @@
   </feature>
 
   <feature name="esh-kat.cpy-jersey-min-2.22.2" version="${project.version}">
-    <feature>http</feature>
+    <feature dependency="true">http</feature>
     <bundle start-level="36">mvn:org.glassfish.jersey.containers/jersey-container-servlet/2.22.2</bundle>
     <bundle start-level="36">mvn:org.glassfish.jersey.media/jersey-media-sse/2.22.2</bundle>
     <bundle start-level="36">mvn:org.glassfish.jersey.media/jersey-media-multipart/2.22.2</bundle>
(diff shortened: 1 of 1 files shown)


kaikreuzer:
Please allow me to self-merge this very small fix - thanks!

features/karaf/esh-tp/src/main/feature/feature.xml
```diff
@@ -177,7 +177,7 @@
   <feature name="esh-tp-serial" version="${project.version}">
     <capability>esh.tp;feature=serial;version=3.12.0</capability>
 
-    <bundle>mvn:org.openhab/nrjavaserial/3.12.0.OH</bundle>
+    <bundle dependency="true">mvn:org.openhab/nrjavaserial/3.12.0.OH</bundle>
```
maggu2810: I don't think this is correct. If you mark this bundle as `dependency="true"` you have to add a special requirement.
The feature `esh-tp-serial` is already marked with `dependency="true"` and so installed if necessary only.
But if I install `esh-tp-serial` manually because I want the serial TP feature to be installed, also the bundle should be installed. So, if you don't want to be to strict about the bundle itself, you need to add at least the requirement for the specific package that should be available.

kaikreuzer: Hm, I tried to align it with the rest and we e.g. have it the same way for esh-tp-apache-commons, esh-tp-httpclient and some others. But I have to admit that I didn't really yet get the meaning&effect of dependency=true on bundles. Are the dependent bundles only transitive dependencies that will be there anyhow? Or what makes the difference and how should it be used?

maggu2810: The TP feature file is not as clean as it should be at the moment. So okay for me to keep it this way for the moment.
I will try to add a summary about that feature file itself, my initial intention and what I think we should do. Will ping you as soon as I will find a time slot.

PR #4587, opened 2017-11-19
[Tradfri] Removed setting the gateway status to 'UNKNOWN' to prevent the thing handlers to be initialized before the gateway

### LINKED ISSUE #4631: Tradfri Handler Lifecycle
https://github.com/eclipse/smarthome/blob/7e1e42c32ee7189c9c68064611aac4a276fddec7/extensions/binding/org.eclipse.smarthome.binding.tradfri/src/main/java/org/eclipse/smarthome/binding/tradfri/handler/TradfriGatewayHandler.java#L124

Any particular reason for setting the status to UNKNOWN (beside for the LightHandlerTest) 
This status basically breaks the proper lifecycle handling because it allows the Light Handler to start initializing BEFORE the gateway is initialized which may cause the following two calls to return null:
https://github.com/eclipse/smarthome/blob/7e1e42c32ee7189c9c68064611aac4a276fddec7/extensions/binding/org.eclipse.smarthome.binding.tradfri/src/main/java/org/eclipse/smarthome/binding/tradfri/handler/TradfriThingHandler.java#L63
https://github.com/eclipse/smarthome/blob/7e1e42c32ee7189c9c68064611aac4a276fddec7/extensions/binding/org.eclipse.smarthome.binding.tradfri/src/main/java/org/eclipse/smarthome/binding/tradfri/handler/TradfriThingHandler.java#L67

cweitkamp:
Closes #4631

See https://community.openhab.org/t/new-features-in-ikea-tradfri-binding-are-not-visible/36031/8

```
2017-11-15 14:53:51.515 [INFO ] [ome.event.ThingStatusInfoEvent:47   ] - 'tradfri:gateway:XXXXXXX' updated: UNKNOWN
2017-11-15 14:53:51.518 [INFO ] [nt.ThingStatusInfoChangedEvent:47   ] - 'tradfri:gateway:XXXXXXX' changed from INITIALIZING to UNKNOWN
2017-11-15 14:53:51.526 [INFO ] [ome.event.ThingStatusInfoEvent:47   ] - 'tradfri:0220:XXXXXXX:65540' updated: INITIALIZING
2017-11-15 14:53:51.535 [INFO ] [nt.ThingStatusInfoChangedEvent:47   ] - 'tradfri:0220:XXXXXXX:65540' changed from UNINITIALIZED (BRIDGE_UNINITIALIZED) to INITIALIZING
2017-11-15 14:53:51.539 [INFO ] [ome.event.ThingStatusInfoEvent:47   ] - 'tradfri:0220:XXXXXXX:65540' updated: OFFLINE (BRIDGE_OFFLINE): Tradfri gateway offline 'UNKNOWN'
2017-11-15 14:53:51.542 [INFO ] [ome.event.ThingStatusInfoEvent:47   ] - 'tradfri:0100:XXXXXXX:65538' updated: INITIALIZING
2017-11-15 14:53:51.548 [INFO ] [nt.ThingStatusInfoChangedEvent:47   ] - 'tradfri:0100:XXXXXXX:65538' changed from UNINITIALIZED (BRIDGE_UNINITIALIZED) to INITIALIZING
2017-11-15 14:53:51.565 [ERROR] [.c.thing.link.ThingLinkManager:283  ] - Exception occurred while informing handler: nulljava.lang.NullPointerException: null
	at org.eclipse.smarthome.binding.tradfri.handler.TradfriLightHandler.handleCommand(TradfriLightHandler.java:119)
	at org.eclipse.smarthome.core.thing.binding.BaseThingHandler.channelLinked(BaseThingHandler.java:205)
	at org.eclipse.smarthome.core.thing.link.ThingLinkManager.informHandlerAboutLinkedChannel(ThingLinkManager.java:281)
	at org.eclipse.smarthome.core.thing.link.ThingLinkManager.receiveTypedEvent(ThingLinkManager.java:321)
	at org.eclipse.smarthome.core.thing.link.ThingLinkManager.receiveTypedEvent(ThingLinkManager.java:1)
	at org.eclipse.smarthome.core.events.AbstractTypedEventSubscriber.receive(AbstractTypedEventSubscriber.java:50)
	at org.eclipse.smarthome.core.internal.events.OSGiEventManager$1.call(OSGiEventManager.java:207)
	at org.eclipse.smarthome.core.internal.events.OSGiEventManager$1.call(OSGiEventManager.java:1)
	at org.eclipse.smarthome.core.common.SafeMethodCaller$CallableWrapper.call(SafeMethodCaller.java:181)
	at java.util.concurrent.ThreadPoolExecutor.runWorker(ThreadPoolExecutor.java:1142)
.....
2017-11-15 14:53:51.611 [ERROR] [.c.thing.link.ThingLinkManager:283  ] - Exception occurred while informing handler: nulljava.lang.NullPointerException: null
	at org.eclipse.smarthome.binding.tradfri.handler.TradfriControllerHandler.handleCommand(TradfriControllerHandler.java:70)
	at org.eclipse.smarthome.core.thing.binding.BaseThingHandler.channelLinked(BaseThingHandler.java:205)
	at org.eclipse.smarthome.core.thing.link.ThingLinkManager.informHandlerAboutLinkedChannel(ThingLinkManager.java:281)
	at org.eclipse.smarthome.core.thing.link.ThingLinkManager.receiveTypedEvent(ThingLinkManager.java:321)
	at org.eclipse.smarthome.core.thing.link.ThingLinkManager.receiveTypedEvent(ThingLinkManager.java:1)
	at org.eclipse.smarthome.core.events.AbstractTypedEventSubscriber.receive(AbstractTypedEventSubscriber.java:50)
	at org.eclipse.smarthome.core.internal.events.OSGiEventManager$1.call(OSGiEventManager.java:207)
	at org.eclipse.smarthome.core.internal.events.OSGiEventManager$1.call(OSGiEventManager.java:1)
	at org.eclipse.smarthome.core.common.SafeMethodCaller$CallableWrapper.call(SafeMethodCaller.java:181)
	at java.util.concurrent.ThreadPoolExecutor.runWorker(ThreadPoolExecutor.java:1142)
.....
2017-11-15 14:53:54.370 [INFO ] [ome.event.ThingStatusInfoEvent:47   ] - 'tradfri:gateway:XXXXXXX' updated: ONLINE
2017-11-15 14:53:54.372 [INFO ] [nt.ThingStatusInfoChangedEvent:47   ] - 'tradfri:gateway:XXXXXXX' changed from UNKNOWN to ONLINE
2017-11-15 14:53:54.373 [INFO ] [ome.event.ThingStatusInfoEvent:47   ] - 'tradfri:0100:XXXXXXX:65538' updated: ONLINE
2017-11-15 14:53:54.380 [INFO ] [nt.ThingStatusInfoChangedEvent:47   ] - 'tradfri:0100:XXXXXXX:65538' changed from OFFLINE (BRIDGE_OFFLINE): Tradfri gateway offline 'UNKNOWN' to ONLINE
2017-11-15 14:53:54.389 [INFO ] [ome.event.ThingStatusInfoEvent:47   ] - 'tradfri:0220:XXXXXXX:65540' updated: ONLINE
2017-11-15 14:53:54.414 [INFO ] [nt.ThingStatusInfoChangedEvent:47   ] - 'tradfri:0220:XXXXXXX:65540' changed from OFFLINE (BRIDGE_OFFLINE): Tradfri gateway offline 'UNKNOWN' to ONLINE
.....
```

### DIFF
--- a/extensions/binding/org.eclipse.smarthome.binding.tradfri/src/main/java/org/eclipse/smarthome/binding/tradfri/handler/TradfriControllerHandler.java
+++ b/extensions/binding/org.eclipse.smarthome.binding.tradfri/src/main/java/org/eclipse/smarthome/binding/tradfri/handler/TradfriControllerHandler.java
@@ -74,6 +74,8 @@ public class TradfriControllerHandler extends TradfriThingHandler {
             logger.debug("Refreshing channel {}", channelUID);
             if (coapClient != null) {
                 coapClient.asyncGet(this);
+            } else {
+                logger.debug("Rejecting command '{}' because 'coapClient' is null.", command);
             }
             return;
         }
--- a/extensions/binding/org.eclipse.smarthome.binding.tradfri/src/main/java/org/eclipse/smarthome/binding/tradfri/handler/TradfriControllerHandler.java
+++ b/extensions/binding/org.eclipse.smarthome.binding.tradfri/src/main/java/org/eclipse/smarthome/binding/tradfri/handler/TradfriControllerHandler.java
@@ -72,9 +72,7 @@ public class TradfriControllerHandler extends TradfriThingHandler {
     public void handleCommand(ChannelUID channelUID, Command command) {
         if (command instanceof RefreshType) {
             logger.debug("Refreshing channel {}", channelUID);
-            if (coapClient != null) {
-                coapClient.asyncGet(this);
-            }
+            coapClient.asyncGet(this);
             return;
         }
 
--- a/extensions/binding/org.eclipse.smarthome.binding.tradfri/src/main/java/org/eclipse/smarthome/binding/tradfri/handler/TradfriControllerHandler.java
+++ b/extensions/binding/org.eclipse.smarthome.binding.tradfri/src/main/java/org/eclipse/smarthome/binding/tradfri/handler/TradfriControllerHandler.java
@@ -74,8 +74,6 @@ public class TradfriControllerHandler extends TradfriThingHandler {
             logger.debug("Refreshing channel {}", channelUID);
             if (coapClient != null) {
                 coapClient.asyncGet(this);
-            } else {
-                logger.debug("Rejecting command '{}' because 'coapClient' is null.", command);
             }
             return;
         }
--- a/extensions/binding/org.eclipse.smarthome.binding.tradfri/src/main/java/org/eclipse/smarthome/binding/tradfri/handler/TradfriControllerHandler.java
+++ b/extensions/binding/org.eclipse.smarthome.binding.tradfri/src/main/java/org/eclipse/smarthome/binding/tradfri/handler/TradfriControllerHandler.java
@@ -72,7 +72,9 @@ public class TradfriControllerHandler extends TradfriThingHandler {
     public void handleCommand(ChannelUID channelUID, Command command) {
         if (command instanceof RefreshType) {
             logger.debug("Refreshing channel {}", channelUID);
-            coapClient.asyncGet(this);
+            if (coapClient != null) {
+                coapClient.asyncGet(this);
+            }
             return;
         }
 
--- a/extensions/binding/org.eclipse.smarthome.binding.tradfri/src/main/java/org/eclipse/smarthome/binding/tradfri/handler/TradfriGatewayHandler.java
+++ b/extensions/binding/org.eclipse.smarthome.binding.tradfri/src/main/java/org/eclipse/smarthome/binding/tradfri/handler/TradfriGatewayHandler.java
@@ -121,7 +121,6 @@ public class TradfriGatewayHandler extends BaseBridgeHandler implements CoapCall
 
                 establishConnection();
             } else {
-                updateStatus(ThingStatus.UNKNOWN);
                 // Running async operation to retrieve new <'identity','key'> pair
                 scheduler.execute(() -> {
                     boolean success = obtainIdentityAndPreSharedKey();
--- a/extensions/binding/org.eclipse.smarthome.binding.tradfri/src/main/java/org/eclipse/smarthome/binding/tradfri/handler/TradfriLightHandler.java
+++ b/extensions/binding/org.eclipse.smarthome.binding.tradfri/src/main/java/org/eclipse/smarthome/binding/tradfri/handler/TradfriLightHandler.java
@@ -122,9 +122,7 @@ public class TradfriLightHandler extends TradfriThingHandler {
         if (active) {
             if (command instanceof RefreshType) {
                 logger.debug("Refreshing channel {}", channelUID);
-                if (coapClient != null) {
-                    coapClient.asyncGet(this);
-                }
+                coapClient.asyncGet(this);
                 return;
             }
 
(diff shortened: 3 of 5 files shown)


maggu2810:
I don't understand this fix.

The `coapClient` variable should be non null if the handler is initialized. Correct?
`handleCommand` should be called only for initialized handlers. Correct?

If the above is correct, then the framework should be fixed.
Sure, we could add code to every binding to check if a handler has been initialized if we know it should be. But I don't think so.

WDYT?

kaikreuzer:
I agree that this should be fixed in the framework - we should ensure that `channelLinked` is not called if the handler is not yet initialised.

cweitkamp:
> The `coapClient` variable should be non null if the handler is initialized. Correct?

Yes, that's correct.

I agree to fix this in the framework.

After a little research I found this line of code:
https://github.com/eclipse/smarthome/blob/master/bundles/core/org.eclipse.smarthome.core.thing/src/main/java/org/eclipse/smarthome/core/thing/link/ThingLinkManager.java#L273

Isn't that exactly the check we want to do?

kaikreuzer:
Looks like it indeed. So the question is: Why isn't it working...?

maggu2810:
@cweitkamp Are you able to reproduce the problem? So you could add some debug message at the time the coap member variable is changed (the new call in initialize and the null assignment in dispose). Also you could add this check temporarily in your working copy and if the coap is null you could log the thing status with all its details.

Perhaps there is a concurrency issue.

maggu2810:
Is the thing status synchronized between the different threads?

kaikreuzer:
We can get in such a state in Tradfri, if the gateway id is invalid and thus the uri for the coapClient is not correct and the coapClient stays null and the Thing is set to OFFLINE (but initialized). This should be fixed, but it is not the situation from the log above, I assume.

kaikreuzer:
> Is the thing status synchronized between the different threads?

This should not really matter as it would only mean that we identify the "initialized" situation too late, but not too early.

vk45:
log file from openHAB_Runtime.lauch executed from Eclipse on aWindows system powerd by I7-Processsor.
I checked for NPE within the tradfri code and report if coapClient is NULL to have a better readable log
[openhab_debug_smarthome_tradfri_009.log](https://github.com/eclipse/smarthome/files/1492781/openhab_debug_smarthome_tradfri_009.log)

cweitkamp:
I am really not sure how to proceed and set up a solution for this? Do you have any advice?

> Are you able to reproduce the problem?

@maggu2810 The problem is reproducable. But I am afraid we have to ask @vk45 for help. He observed this issue in his environment. I am only the intermediary. ;-)

There is only one line of code where the `coapClient` really changed:
https://github.com/eclipse/smarthome/blob/7e1e42c32ee7189c9c68064611aac4a276fddec7/extensions/binding/org.eclipse.smarthome.binding.tradfri/src/main/java/org/eclipse/smarthome/binding/tradfri/handler/TradfriThingHandler.java#L66
It is never set to `null` again. We use `coapClient.shutdown()` in the `dispose()` method of the handler.

ivivanov-bg:
>I am really not sure how to proceed and set up a solution for this? Do you have any advice?

I guess you can try to remove setting the gateway status to UNKNOWN. This will prevent the light handler been initialized before the gateway handler goes to online state (which I think should be the expected behavior)

You should also ignore the ```creationOfTradfriLightHandler``` test because it will fail (as there is no real gateway and the handler will always be in INITIALIZING state)

vk45:
I made a short dirty hack in TradfriThingHandler.initialize and added a  case UNKNOWN with the content of the ONLINE case , that works

but yours is a better idea:
I removed updateStatus(ThingStatus.UNKNOWN) from TradfriGatewayHandler.setupCoapClient().
now it works.

cweitkamp:
Thanks for your help. I removed setting the gateway status to 'UNKNOWN' and reverted my previous changes.

cweitkamp:
Just want to ask if we can make any progress in this case. Is the solution sufficient or shall we cancel the PR?

kaikreuzer:
Sorry @cweitkamp I lost the updates of that discussion.
It all seems to make sense and I think the solution is fine!

cweitkamp:
@kaikreuzer Never mind. Thank you.

PR #4592, opened 2017-11-20
Revert "added dependency="true" where missing"

kaikreuzer:
Reverts eclipse/smarthome#4585

Should fix https://github.com/openhab/openhab-distro/issues/571


maggu2810:
I assume you would like to revert only one not four commits.

kaikreuzer:
I used the "revert" button of Github. No idea why it created 4 commits, but if you check the resulting change, it is just the two lines concerned. So if you squash+merge, all should be fine.

PR #4593, opened 2017-11-20
Maven tycho plugin warning

maggu2810:
Get rid of 

```text
[WARNING] 
[WARNING] Some problems were encountered while building the effective model for org.eclipse.smarthome.archetype:org.eclipse.smarthome.magic.test:eclipse-test-plugin:0.9.0-SNAPSHOT
[WARNING] 'build.plugins.plugin.version' for org.eclipse.tycho:target-platform-configuration is missing. @ line 27, column 15
[WARNING] 'build.plugins.plugin.version' for org.eclipse.tycho:tycho-surefire-plugin is missing. @ line 63, column 15
[WARNING] 
[WARNING] It is highly recommended to fix these problems because they threaten the stability of your build.
[WARNING] 
[WARNING] For this reason, future Maven versions might no longer support building such malformed projects.
[WARNING] 
```

on every Maven build.

### DIFF
--- a/bundles/automation/org.eclipse.smarthome.automation.core.test/pom.xml
+++ b/bundles/automation/org.eclipse.smarthome.automation.core.test/pom.xml
@@ -23,7 +23,7 @@
   <build>
     <plugins>
       <plugin>
-        <groupId>org.eclipse.tycho</groupId>
+        <groupId>${tycho-groupid}</groupId>
         <artifactId>tycho-surefire-plugin</artifactId>
         <version>${tycho-version}</version>
       </plugin>
--- a/bundles/automation/org.eclipse.smarthome.automation.core.test/pom.xml
+++ b/bundles/automation/org.eclipse.smarthome.automation.core.test/pom.xml
@@ -25,7 +25,6 @@
       <plugin>
         <groupId>${tycho-groupid}</groupId>
         <artifactId>tycho-surefire-plugin</artifactId>
-        <version>${tycho-version}</version>
       </plugin>
     </plugins>
   </build>
--- a/bundles/automation/org.eclipse.smarthome.automation.event.test/pom.xml
+++ b/bundles/automation/org.eclipse.smarthome.automation.event.test/pom.xml
@@ -22,7 +22,6 @@
       <plugin>
         <groupId>${tycho-groupid}</groupId>
         <artifactId>tycho-surefire-plugin</artifactId>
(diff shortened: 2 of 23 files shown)


martinvw:
@maggu2810 how does it resolve now and not before is the group-id not the same? Ie another plugin?

maggu2810:
A property could be changed e.g. by command line etc.
So, if the plugin management is using the property key, the (non management) plugin sections should use the property, too.

martinvw:
So in our case it changed and thus has a different value otherwise this change would not resolve the warming. Or do i miss a crucial detail here?

maggu2810:
You are right. There is something strange (and interesting).
Need to analyze the whole tree, to see why it is working in the one case but not in the other...

Another question would be, why do we use a property for the groupId at all? Do we consider that it will be changed?

maggu2810:
@martinvw Can you filter your results to drop that places that defines a version itself? If you have a section that defines the version for that plugin, why should you see a warning?

maggu2810:
I have added another one, to get consistent property usage:
https://github.com/eclipse/smarthome/pull/4593/commits/ff66289935fc18c2cb96cc6a95ca6ce3b129e69e

The next step will be that I remove the redundant version declaration as it should be derived from the parent / root POM.
I set this PR to WIP until it has been fixed.

maggu2810:
> The next step will be that I remove the redundant version declaration as it should be derived from the parent / root POM.

Done

martinvw:
This one can maybe be answered by @kaikreuzer / @sjka 

> Another question would be, why do we use a property for the groupId at all? Do we consider that it will be changed?

@maggu2810 so this makes the warning disappear so can we conclude that `does maven do a literal string matching before substituting properties.`

Do you still need refinement of the filter? I was having diner...

Maybe we could consider whether we should do a cleanup for all `org.eclipse.tycho` plug-ins such as: 'tycho-p2-plugin', 'tycho-source-feature-plugin' and maybe others.

PS: sorry for making it explode :-)

maggu2810:
> Maybe we could consider whether we should do a cleanup for all `org.eclipse.tycho` plug-ins such as: 'tycho-p2-plugin', 'tycho-source-feature-plugin' and maybe others.

What do you miss?

martinvw:
Ah, I see you already did that complete job :-D

martinvw:
Thanks!

Issue #4570, opened 2017-11-16
State of zonegroup item contains XML


aounhaider1:
In sonos PLAY:1 the state of zonegroup item contains a large XML. I am just wondering if this is a feature or a bug?
Here is the item:
```xml
{
  "link": "https://192.168.2.154/rest/items/sonos_PLAY3_RINCON_B8E9373719EE01400_zonegroup",
  "state": "<ZoneGroups><ZoneGroup Coordinator=\"RINCON_B8E9373719EE01400\" ID=\"RINCON_B8E9373719EE01400:1928870658\"><ZoneGroupMember UUID=\"RINCON_B8E9373719EE01400\" Location=\"http://192.168.2.152:1400/xml/device_description.xml\" ZoneName=\"Lounge\" Icon=\"x-rincon-roomicon:living\" Configuration=\"1\" SoftwareVersion=\"39.2-47040\" MinCompatibleVersion=\"38.0-00000\" LegacyCompatibleVersion=\"25.0-00000\" BootSeq=\"59\" WirelessMode=\"0\" WirelessLeafOnly=\"0\" HasConfiguredSSID=\"1\" ChannelFreq=\"2437\" BehindWifiExtender=\"0\" WifiEnabled=\"1\" Orientation=\"0\" RoomCalibrationState=\"4\" SecureRegState=\"2\" VoiceState=\"0\"/></ZoneGroup></ZoneGroups>",
  "stateDescription": {
    "pattern": "%s",
    "readOnly": true,
    "options": []
  },
  "type": "String",
  "name": "sonos_PLAY3_RINCON_B8E9373719EE01400_zonegroup",
  "label": "Zone Group",
  "tags": [],
  "groupNames": [
    "sonos_PLAY3_RINCON_B8E9373719EE01400"
  ]
}
```
Screenshot of PaperUI control page:
<img width="804" alt="screen shot 2017-11-16 at 15 52 02" src="https://user-images.githubusercontent.com/5161937/32897509-2446099c-cae6-11e7-86be-ca56095aa317.png">

aounhaider1:
@kgoderis maybe you know about this?

kaikreuzer:
[RTFM](https://github.com/eclipse/smarthome/blob/master/extensions/binding/org.eclipse.smarthome.binding.sonos/README.md#channels): "XML formatted string with the current zonegroup configuration"

The question remains whether this channel is helpful to anyone. I would vote to simply remove it.

[...]

PR #4595, opened 2017-11-20
PaperUI: Collapse long string stateText

aounhaider1:
Before:

screen shot 2017-11-16 at 15 52 02

After:

screen shot 2017-11-20 at 16 14 17

### DIFF
--- a/extensions/ui/org.eclipse.smarthome.ui.paper/web-src/css/parameterDescription.css
+++ b/extensions/ui/org.eclipse.smarthome.ui.paper/web-src/css/parameterDescription.css
@@ -31,6 +31,10 @@
 	margin-bottom: 0;
 }
 
+.state > .ellipsis p {
+    font-size: 30px;
+}
+
 .ellipsis:after {
 	content: "\02026 more";
 	cursor: pointer; box-sizing : content-box;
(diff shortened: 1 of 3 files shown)


htreu:
Hi @aounhaider1 what is the reason for an item to have such a long state text? What device/channel does provide this text? Is it meant to be bound to an item?

aounhaider1:
The zonegroup channel of Sonos have this stateText. This PR is just a boundry check in case an item has a long state text.

kaikreuzer:
@htreu FYI: https://github.com/eclipse/smarthome/issues/4570

htreu:
Although the implementation of the ellipsis feature here is okay from a coding pov, I guess we should close this PR since it fixes the "bug" in the wrong place.
Even when the sonos channel in question stays like it is it does not deliver end-user information when bound to an item. It should at least be filtered/transformed in on way or the other.

aounhaider1:
Ok. But I think that at least the text should be wrapped inside the card. I'll close this PR.

htreu:
@aounhaider1 yes, wrapping the text would actually make sense. I guess with XML there is the problem of not having whitespace separators to allow word wrap.

PR #4596, opened 2017-11-20
Add serial number to discovery result 

ivivanov-bg:
Show mac address in gateway discovery result


ivivanov-bg:
And what happens if you have more than one Gateway in the network ?
How can you distinguish them- (and provide the proper security code printed on the gateway (right below the Serial Number, which is practically the MAC address)

maggu2810:
Hm, a discovery result contains properties. The UI (didn't know the current Paper UI implementation) could show the properties the same way as we could look at the properties on things.

kaikreuzer:
> And what happens if you have more than one Gateway in the network ?

Isn't the IP address used as a representation property?
As the name suggests, this property should be used by UIs to give users a hint on how to identify a certain device.

ivivanov-bg:
>Isn't the IP address used as a representation property?

Yes - it is. But:
1. I don't think the IP address is a good candidate for representation property (as it's not hardly bound to the device like the serial number for eg. is and it's also not printed on the device itself - again like the serial number is)
2. Even if we agree about the proper representation property - the UI currently doesn't show it.

I know this can also be seen as an issue of the UI but as I already said - this is done for Sonos and Hue long before the concept for representation property existed.

So I would suggest to have similar behavior for tradfri (as a workaround) and once the UI is updated - to just remove the workarounds from all bindings.

PS: May be I labeled the PR wrong - it should be "Show serial number in gateway discovery result". (And the fact that the serial number is the same as the mac address should not be taken into account.)

ivivanov-bg:
Updated the discovery to not change the label but add the serial number as a property instead.

maggu2810:
Looks good to me

PR #4597, opened 2017-11-20
PaperUI: Fix config description binding error

htreu:
This fixes massive JavaScript console error logging when AngularJS recalculates the digest loop.


extensions/ui/org.eclipse.smarthome.ui.paper/web-src/js/directives/directive.configDescription.js
```diff
@@ -3,6 +3,9 @@ angular.module('PaperUI.directive.configDescription', []) //
 
     var controller = function($scope) {
         $scope.getName = function(parameter, option) {
+            if (!option) {
+                return undefined;
```
kubawolanin: not sure if your ESLint configuration allows it but simple `return;` should suffice as it returns `undefined` under the hood

PR #4598, opened 2017-11-20
PaperUI: Fix switch state for dimmer item with category 'DimmableLight'

htreu:
Fixes a bug where the UI switch state for a dimmer item with category "DimmableLight" does not get updated. When the state change is not initiated by PaperUI but from the backend the switch will now also display the correct On/Off state.

### DIFF
--- a/extensions/ui/org.eclipse.smarthome.ui.paper/web-src/js/controllers.control.js
+++ b/extensions/ui/org.eclipse.smarthome.ui.paper/web-src/js/controllers.control.js
@@ -350,6 +350,10 @@ angular.module('PaperUI.controllers.control', []) //
         sendCommand(brightness);
     }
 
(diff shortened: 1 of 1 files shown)


htreu:
The DimmerItemController is responsible for updating two UI controls in sync: there is the slider which adjusts the percentage value and there is the switch which is on for brightness > 0. The slider is directly bound to the item state by `ng-model="item.state"` in the html. The switch is bound to a synthetic property in the controller $scope (`$scope.state.switchState`). The $scope did not get updated when the item state did change from the outside. The new `$scope.$watch(...)` fixes this.

kaikreuzer:
@maggu2810 Question answered? If so, feel free to merge.

maggu2810:
@htreu Thanks :wink:

Issue #4591, opened 2017-11-20
OSGiEventManager should do safe method calls and identify handlers that are blocking


kaikreuzer:
[There are reports](https://community.openhab.org/t/jetty-update-karaf-4-1-3-upgrade-and-full-lsp-support/36354/7?u=kai) where the OSGiEventManager gets blacklisted by the OSGi EventAdmin, which results in an unusable system state. Instead of being blamed, the OSGiEventManager should instead make sure to identify blocking handlers itself and return quickly.

sjka:
Even if it would take such buggers out of the game, I still see two remaining problems:

The default timeout of e.g. [felix event admin](http://felix.apache.org/documentation/subprojects/apache-felix-event-admin.html) is 5s, which is exactly the timeout of the safe method caller in this case. So at the point where it identifies a slow event subscriber, it is already to late and the event manager got banned itself. 

If the internal timeouts are fixed, then the same would arise if there are multiple slow subscribers because they are informed serially. 

The "proper" solution I guess would be to inform the subscribers asynchronously completely (and thereby taking care to ban slow subscribers itself). Wdyt?

sjka:
no major objections so far, so I'll look into making an according PR proposal.

PR #4599, opened 2017-11-20
notify handlers asynchronously so that longer running operations do not block us

kaikreuzer:
Currently, the ThingLinkManager might process events very slowly, which can be a cause for the issue described in #4591. Notifying the handlers asynchronously should solve this.


maggu2810:
Doesn't this increase the potential change about a "not initialized anymore" call to the channel (un)linked method?

kaikreuzer:
True. I've now moved the check into the runnable, so that it is checked immediately before the call - does that look better?

maggu2810:
Yes, the change is so the same as before, but the other issue is resolved.
Looks good to me. Let's wait for Travis and then merge the PR.

sjka:
> Yes, the change is so the same as before [...]

At least there is a pretty high probability that is is 😉  

Let's see how the chances are in reality of running into a race condition at this place - the consequences would not be of real harm though, so I'm fine with it!

PR #4600, opened 2017-11-21
added deprecation info on bundleContext as bindings must not use it

kaikreuzer:


maggu2810:
Thank you

PR #4602, opened 2017-11-21
Unblock event dispatcher

### LINKED ISSUE #4591: OSGiEventManager should do safe method calls and identify handlers that are blocking

sjka:
* remove "hanging" event subscribers from the game
* inform _all_ event subscribers asynchronously

fixes #4591


bundles/core/org.eclipse.smarthome.core/src/main/java/org/eclipse/smarthome/core/internal/events/OSGiEventManager.java
```diff
@@ -197,24 +204,24 @@ private Event createESHEvent(final EventFactory eventFactory, final String type,
...
-                if (filter == null || filter.apply(event)) {
-                    SafeMethodCaller.call(new ActionWithException<Void>() {
-
-                        @Override
-                        public Void call() throws Exception {
+            EventFilter filter = eventSubscriber.getEventFilter();
+            if (filter == null || filter.apply(event)) {
+                scheduler.submit(() -> {
+                    try {
+                        SafeMethodCaller.call((ActionWithException<Void>) () -> {
```
kaikreuzer: This looks not efficient to me. You are submitting a job to be executed in a thread of the eventManager thread pool and then you do a (blocking) safe method call, which again spawns another thread.

Imho what is missing is a method in the SafeMethodCaller to do async (non-blocking) calls - if that existed, a single thread would be enough here.

wborn: Will the order of events still be maintained when an eventSubscriber is receiving events asynchronously?

maggu2810:
> Will the order of events still be maintained when an eventSubscriber is receiving events asynchronously?

That is a good point. We should ensure that this changes and also #4599 does not result in an un-ordered event handling. Is this (e.g. with #4599) the case now? Should we force a single threaded pool?

sjka:
My suggestion rather would be to enhance the safe method caller mechanism to only execute one thread in parallel per "target" (whatever this is - depends on the use-case) and thereby solving two aspects:

1. guarantee the execution order within such a "target" context
2. don't punish other parts of the system if one actor (e.g. a handler) hangs

I wouldn't use single-executor pools though in order to keep the overall number of threads low.

vkolotov:
Hi all, I'm having the #4591 issue as well. This happens very often and you have to restart OH to make it go away. Looking forward to seeing it fixed/merged.

sjka:
superseeded by #4655

bundles/core/org.eclipse.smarthome.core/src/main/java/org/eclipse/smarthome/core/internal/events/OSGiEventManager.java
sjka: Very good point - no, this was not guaranteed anymore. 
#4655 will take care of this by executing them _asynchronously_ but _sequentially_ (i.e. not in parallel).

PR #4605, opened 2017-11-21
add profile to use tycho snapshot

maggu2810:
Let's use the SNAPSHOT temporary as long as there is no more recent
release.
We need to use a more recent Tycho version (>1.0.0) to use a more recent
JDT compiler.
A more recent JDT compiler is necessary because of better nullness
handling.

### DIFF
--- a/pom.xml
+++ b/pom.xml
@@ -540,6 +540,23 @@
         </plugins>
       </build>
     </profile>
+    <profile>
+      <id>tycho-snapshot</id>
+      <activation>
+        <property>
+          <name>tycho-snapshot</name>
+        </property>
+      </activation>
+      <properties>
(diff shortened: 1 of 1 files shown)


kaikreuzer:
What does this bring us with #4604 in place?

maggu2810:
Nothing. Because of #4606 we cannot test if newly added annotations would work with a more recent JDT compiler using the Maven build (without changing the compiler arguments, too) anymore.

kaikreuzer:
So shall we close this PR and use the Tycho snapshot after the 0.9.0 release?

maggu2810:
I add a further commit to the branch, so I could still use that profile to check the nullness annotations using Maven. But no need to integrate it here.

PR #4606, opened 2017-11-21
Annotate discovery package

triller-telekom:
This PR needs #4550 as a prerequisite.

Also-by: Markus Rathgeb <maggu2810@gmail.com>

### DIFF
--- a/bundles/config/org.eclipse.smarthome.config.discovery/src/main/java/org/eclipse/smarthome/config/discovery/DiscoveryListener.java
+++ b/bundles/config/org.eclipse.smarthome.config.discovery/src/main/java/org/eclipse/smarthome/config/discovery/DiscoveryListener.java
@@ -9,6 +9,8 @@ package org.eclipse.smarthome.config.discovery;
 
 import java.util.Collection;
 
+import org.eclipse.jdt.annotation.NonNullByDefault;
+import org.eclipse.jdt.annotation.Nullable;
 import org.eclipse.smarthome.core.thing.ThingTypeUID;
 import org.eclipse.smarthome.core.thing.ThingUID;
 
@@ -23,6 +25,7 @@ import org.eclipse.smarthome.core.thing.ThingUID;
  *
  * @see DiscoveryService
  */
+@NonNullByDefault
 public interface DiscoveryListener {
 
     /**
@@ -52,7 +55,7 @@ public interface DiscoveryListener {
     /**
      * Removes all results belonging to one of the given types that are older
      * than the given timestamp.
-     * 
+     *
      * @param source
      *            the discovery service which is the source of this event (not
      *            null)
@@ -65,5 +68,7 @@ public interface DiscoveryListener {
      *            instead
      * @return collection of thing UIDs of all removed things
      */
-    Collection<ThingUID> removeOlderResults(DiscoveryService source, long timestamp, Collection<ThingTypeUID> thingTypeUIDs);
+    @Nullable
+    Collection<ThingUID> removeOlderResults(DiscoveryService source, long timestamp,
+            @Nullable Collection<ThingTypeUID> thingTypeUIDs);
 }
--- a/bundles/config/org.eclipse.smarthome.config.discovery/src/main/java/org/eclipse/smarthome/config/discovery/DiscoveryListener.java
+++ b/bundles/config/org.eclipse.smarthome.config.discovery/src/main/java/org/eclipse/smarthome/config/discovery/DiscoveryListener.java
@@ -68,7 +68,6 @@ public interface DiscoveryListener {
      *            instead
      * @return collection of thing UIDs of all removed things
      */
-    @Nullable
     Collection<ThingUID> removeOlderResults(DiscoveryService source, long timestamp,
             @Nullable Collection<ThingTypeUID> thingTypeUIDs);
 }
--- a/bundles/config/org.eclipse.smarthome.config.discovery/src/main/java/org/eclipse/smarthome/config/discovery/AbstractDiscoveryService.java
+++ b/bundles/config/org.eclipse.smarthome.config.discovery/src/main/java/org/eclipse/smarthome/config/discovery/AbstractDiscoveryService.java
@@ -233,10 +233,11 @@ public abstract class AbstractDiscoveryService implements DiscoveryService {
                 scheduledStop.cancel(false);
                 scheduledStop = null;
             }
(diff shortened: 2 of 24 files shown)


maggu2810:
Using the more recent JDT compiler:

```text
[ERROR] /home/maggu2810/bin/pkgs/eclipse/esh/git/smarthome/bundles/config/org.eclipse.smarthome.config.discovery/src/main/java/org/eclipse/smarthome/config/discovery/dto/DiscoveryResultDTOMapper.java:[50]
[ERROR] ? new ThingTypeUID(discoveryResultDTO.thingTypeUID)
[ERROR] ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
[ERROR] Null type mismatch (type annotations): required '@NonNull String' but this expression has type '@Nullable String'
```

bundles/config/org.eclipse.smarthome.config.discovery/src/main/java/org/eclipse/smarthome/config/discovery/DiscoveryListener.java
```diff
@@ -65,5 +68,7 @@
      *            instead
      * @return collection of thing UIDs of all removed things
      */
-    Collection<ThingUID> removeOlderResults(DiscoveryService source, long timestamp, Collection<ThingTypeUID> thingTypeUIDs);
+    @Nullable
```
maggu2810: Why do we support a nullable return type? The JavaDoc states that a collection is returned and wouldn't it be simpler if the return value could be used in a for each loop without check for null first?

maggu2810:
I have also created a PR against your repo.

bundles/config/org.eclipse.smarthome.config.discovery/src/main/java/org/eclipse/smarthome/config/discovery/DiscoveryListener.java
triller-telekom: Sure, but then I have to fix at least the wemo binding, hue binding and a tradfri test:

https://github.com/eclipse/smarthome/blob/master/extensions/binding/org.eclipse.smarthome.binding.wemo/src/main/java/org/eclipse/smarthome/binding/wemo/handler/WemoMakerHandler.java#L325

https://github.com/eclipse/smarthome/blob/master/extensions/binding/org.eclipse.smarthome.binding.wemo/src/main/java/org/eclipse/smarthome/binding/wemo/handler/WemoHandler.java#L490

https://github.com/eclipse/smarthome/blob/master/extensions/binding/org.eclipse.smarthome.binding.wemo/src/main/java/org/eclipse/smarthome/binding/wemo/handler/WemoCoffeeHandler.java#L505

https://github.com/eclipse/smarthome/blob/master/extensions/binding/org.eclipse.smarthome.binding.hue/src/main/java/org/eclipse/smarthome/binding/hue/handler/HueBridgeHandler.java#L531

https://github.com/eclipse/smarthome/blob/master/extensions/binding/org.eclipse.smarthome.binding.tradfri.test/src/test/java/org/eclipse/smarthome/binding/tradfri/discovery/TradfriDiscoveryServiceTest.java#L73

Since in ESH these bindings alone violate the @NonNull I did not put it there to also not break any openHAB bindings.

So my question now is: Shall I fix those ESH bindings within this PR? Create a separate one because its independent from this discovery package? How shall we deal with such situations?

triller-telekom:
@maggu2810 Thank your for your PR, I have included it. Can you still reproduce https://github.com/eclipse/smarthome/pull/4606#issuecomment-346172428 ? Because in my IDE everything is alright... Or maybe you have some annotation on the `ThingUID` constructor to require @NonNull? because the current master does not have that.

How do we proceed regarding the binding fixing from comment https://github.com/eclipse/smarthome/pull/4606#discussion_r152487647 ?

maggu2810:
For me a comment like "collection of thing UIDs of all removed things" states that a collection is returned, that can be empty if there is nothing removed.

I would also expect that I could write something like this one:

```Java
for (final ThingUID uid : listener.removeOlderResults(...)) {
    ...
}
```

So, I would prefer to return an empty list and so non-null.
Fixing the bindings is IMHO the better solutions as we can some time rely that null should not returned and we don't need to insert a check for the return value on every usage.
I also don't see a difference between null and an empty list, so no reason to use null (an empty collection is already present as a singleton).

maggu2810:
IIRC I don't see any further errors after inclusion of my PR.

triller-telekom:
I have changed `removeOlderResults` to return @NonNull and fixed the bindings/tests what returned `null` before to now return `Collections.emptyList()`.

Please have a look at it again.

PS: My goal is to annotate the hue binding to have an example for others that we can reference in a blog post on the ESH website soon. That's why I annotated the binding API first (maybe there are other packages that I come across while annotating the hue binding though).

maggu2810:
> My goal is to annotate the hue binding

:+1: 

> I annotated the binding API

:+1: 

> maybe there are other packages that I come across while annotating the hue binding

:+1: 

:wink:

kaikreuzer:
I just had to notice that this is actually source (not binary) API breaking as it changed the method signature of `DiscoveryListener.removeOlderResults()`. As can be seen on this PR here as well, quite some code in ESH required adaption as well.

I wonder whether this change is worth it. For methods on interfaces that are often not implemented, it is convenient to simply keep the auto-generated (empty) code. In the case here, this results in a compilation error that the user has to fix although he isn't interested in this method.
Another argument against it might be that we say that framework code should be defensive and always potentially expect null values - so on interfaces that we expect 3rd parties to implement, the return values might often be better marked nullable. Wdyt?

maggu2810:
If interfaces defines methods that are "often not implemented" with "implemented" interpret as " keep the auto-generated (empty) code" (that returns null) why doesn't we introduce a default method in the interface.
If the method is required but not ofter used, there should be one.

If we need to harden the framework to work with every third party code that developers does not follow the signature, we are back to the situation before the annotations. We have to check every method argument of the API (public in sense of used by 3rd party developers) if there is a null argument given and every method that is called by the framework if null is returned (as long as it is not a primitve type).

This PR did not change the handling of the return value. So, if the framework handles the null return value correctly before it is still doing.
So currently we will still handle 3rd party code that returns null, but we mark that it should not return null.
For me it is the same as we sometime keep the null checks for non-null marked arguments. If it will break the framework at some future state we should keep it as the error is not direct visible. But this is not an arguemnt to not mark something that is should not be null.

PR #4607, opened 2017-11-21
removed zonegroup channel

kaikreuzer:
See https://github.com/eclipse/smarthome/issues/4570#issuecomment-345862380

### DIFF
--- a/extensions/binding/org.eclipse.smarthome.binding.sonos/ESH-INF/thing/One.xml
+++ b/extensions/binding/org.eclipse.smarthome.binding.sonos/ESH-INF/thing/One.xml
@@ -48,7 +48,7 @@
 			<channel id="stop" typeId="stop" />
 			<channel id="tuneinstationid" typeId="tuneinstationid" />
 			<channel id="volume" typeId="volume" />
-			<channel id="zonegroup" typeId="zonegroup" />
+			<channel id="zonegroupid" typeId="zonegroupid" />
 			<channel id="zonename" typeId="zonename" />
 			<channel id="coordinator" typeId="coordinator" />
 			<channel id="sleeptimer" typeId="sleeptimer" />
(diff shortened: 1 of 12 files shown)


extensions/binding/org.eclipse.smarthome.binding.sonos/ESH-INF/thing/One.xml
```diff
@@ -49,7 +49,6 @@
 			<channel id="tuneinstationid" typeId="tuneinstationid" />
 			<channel id="volume" typeId="volume" />
 			<channel id="zonegroup" typeId="zonegroup" />
-			<channel id="zonegroupid" typeId="zonegroupid" />
```
AlexanderKostadinov28: Is there a specific reason that for the One the `zonegroupid` channel is removed instead of the `zonegroup` (like for the other players)? To me this seems to be a mistake. :)

AlexanderKostadinov28:
@kaikreuzer Can you please give further information why this commit is API-breaking and in terms of what?

extensions/binding/org.eclipse.smarthome.binding.sonos/ESH-INF/thing/One.xml
sjka: @kaikreuzer have you seen this comment?

kaikreuzer: Thanks @AlexanderKostadinov28, good catch! I certainly only removed it to check if anyone is doing a thorough review ;-)

kaikreuzer:
@AlexanderKostadinov28 I added API breaking because we remove a "feature". So for the unlikely situation that anybody uses that XML channel, it will obviously break his setup. No other features are broken by this PR, though.

PR #4613, opened 2017-11-23
Chart periode: add H as alias to h

mueller-ma:
D, W, M, Y are uppercase, but h is lowercase. This adds H as uppercase since uppercase H is not bound to anything else.
Also see `date` (from manpage):
````
%H     hour (00..23)
````
Signed-off-by: mueller-ma <mueller-ma@users.noreply.github.com>


mueller-ma:
Dont want to create an eclipse account...

PR #4614, opened 2017-11-23
Fix GroupItem state calculation

htreu:
GroupItem calculates its state over all direct members. For sub-groups the state of the sub-group is taken into the calculation.
The `GroupItem#getStateAs` method however did calculate the state over all non-GroupItem members and sub-members.

Also: Unit test added.

### DIFF
--- a/docs/documentation/concepts/items.md
+++ b/docs/documentation/concepts/items.md
@@ -35,32 +35,32 @@ Group items can themselves be members of other group items.
 Cyclic membership is not forbidden but strongly not recommended.
 User interfaces might display group items as single entries and provide navigation to its members.
 
-Example for a Group item simply collection other items:
+Example for a Group item as a simple collection of other items:
 ```
     Group groundFloor
     Switch kitchenLight (groundFloor)
     Switch livingroomLight (groundFloor)
 ``` 
 
-In addition Group items can calculate their own state derived from their members using a specific group function.
+### Derive Group State from Member Items
 
-### Derive group state from member items
-
-Group items can derive their own state depending on their member items.
+Group items can derive their own state from their member items.
 To derive a state the group item must be constructed using a base item and a group function.
-When calculating state, group functions will recursively traverse the group's members and also take members of subgroups into account.
+When calculating the state, group functions recursively traverse the group's members and also take members of subgroups into account.
 If a subgroup however defines a state on its own (having base item & group function set) traversal stops and the state of the subgroup member is taken. 
 
 Available group functions:
 
-| Function           | Parameters  | Base Item | Description |
-|--------------------|-------------|-----------|-------------|
-| EQUALITY           | - | \<all\> | Sets the state of the members if all have equal state. Otherwise UNDEF is set |
-| AND, OR, NAND, NOR | \<activeState\>, \<passiveState\> | \<all\> (must match active & passive state) | Sets the <activeState> if the member state <activeState> evaluates to `true` under the boolean term. Otherwise the <passiveState> is set. |
-| SUM, AVG, MIN, MAX | - | Number | Sets the state according to the arithmetic function over all member states. |
-| COUNT              | \<regular expression\> | Number | Sets the state to the number of members matching the given regular expression with their states. |
+| Function           | Parameters                    | Base Item                                   | Description                                                                                                                                     |
+|--------------------|-------------------------------|---------------------------------------------|-------------------------------------------------------------------------------------------------------------------------------------------------|
+| EQUALITY           | -                             | \<all\>                                     | Sets the state of the members if all have equal state. Otherwise UNDEF is set.                                                                  |
+| AND, OR, NAND, NOR | <activeState>, <passiveState> | \<all\> (must match active & passive state) | Sets the \<activeState\>, if the member state \<activeState\> evaluates to `true` under the boolean term. Otherwise the \<passiveState\> is set.|
+| SUM, AVG, MIN, MAX | -                             | Number                                      | Sets the state according to the arithmetic function over all member states.                                                                     |
+| COUNT              | <regular expression>          | Number                                      | Sets the state to the number of members matching the given regular expression with their states.                                                |
+
 
 Examples for derived states on group items when declared in the item DSL:
+
 - `Group:Number:COUNT(".*")` counts all members of the group matching the given regular expression, here any character or state (simply count all members).
 - `Group:Number:AVG` calculates the average value over all member states which can be interpreted as `DecimalTypes`.
 - `Group:Switch:OR(ON,OFF)` sets the group state to `ON` if any of its members has the state `ON`, `OFF` if all are off.    
--- a/docs/documentation/concepts/items.md
+++ b/docs/documentation/concepts/items.md
@@ -28,11 +28,43 @@ The following item types are currently available (alphabetical order):
 | String         | Stores texts | String |
 | Switch         | Typically used for lights (on/off) | OnOff |
 
-Group Items can derive their own state depending on their member items.
+## Group Items
 
-- AVG displays the average of the item states in the group.
-- OR displays an OR of the group, typically used to display whether any item in a group has been set.
-- other aggregations: AND, SUM, MIN, MAX, NAND, NOR
+Group items collect other items into groups.
+Group items can themselves be members of other group items.
+Cyclic membership is not forbidden but strongly not recommended.
+User interfaces might display group items as single entries and provide navigation to its members.
+
+Example for a Group item simply collection other items:
+```
+    Group groundFloor
+    Switch kitchenLight (groundFloor)
+    Switch livingroomLight (groundFloor)
+``` 
+
+In addition Group items can calculate their own state derived from their members using a specific group function.
+
+### Derive group state from member items
+
+Group items can derive their own state depending on their member items.
+To derive a state the group item must be constructed using a base item and a group function.
+When calculating state, group functions will recursively traverse the group's members and also take members of subgroups into account.
+If a subgroup however defines a state on its own (having base item & group function set) traversal stops and the state of the subgroup member is taken. 
+
+Available group functions:
+
+| Function           | Parameters  | Base Item | Description |
+|--------------------|-------------|-----------|-------------|
+| EQUALITY           | - | \<all\> | Sets the state of the members if all have equal state. Otherwise UNDEF is set |
+| AND, OR, NAND, NOR | \<activeState\>, \<passiveState\> | \<all\> (must match active & passive state) | Sets the <activeState> if the member state <activeState> evaluates to `true` under the boolean term. Otherwise the <passiveState> is set. |
+| SUM, AVG, MIN, MAX | - | Number | Sets the state according to the arithmetic function over all member states. |
+| COUNT              | \<regular expression\> | Number | Sets the state to the number of members matching the given regular expression with their states. |
+
+Examples for derived states on group items when declared in the item DSL:
+- `Group:Number:COUNT(".*")` counts all members of the group matching the given regular expression, here any character or state (simply count all members).
+- `Group:Number:AVG` calculates the average value over all member states which can be interpreted as `DecimalTypes`.
+- `Group:Switch:OR(ON,OFF)` sets the group state to `ON` if any of its members has the state `ON`, `OFF` if all are off.    
+- `Group:Switch:AND(ON,OFF)` sets the group state to `ON` if all of its members have the state `ON`, `OFF` if any of the group members has a different state than `ON`.
 
 ## State and Command Type Formatting
 
--- a/bundles/core/org.eclipse.smarthome.core.test/src/test/java/org/eclipse/smarthome/core/items/GroupItemTest.java
+++ b/bundles/core/org.eclipse.smarthome.core.test/src/test/java/org/eclipse/smarthome/core/items/GroupItemTest.java
@@ -197,6 +197,41 @@ public class GroupItemTest extends JavaOSGiTest {
     }
 
     @Test
+    public void testGetStateAs_shouldEqualStateUpdate() {
+        // Main group uses AND function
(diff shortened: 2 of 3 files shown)


kaikreuzer:
Contemplating about this, I think neither is correct - the best would be to differentiate between group items that have a state themselves and those that don't.

Take this setup as an example:
```
Group:Number:AVG MyAverageHouseTemperature

Group Kitchen (MyAverageHouseTemperature)
Group Office (MyAverageHouseTemperature)
Group:Number:AVG Cellar (MyAverageHouseTemperature)

Number Temp_Kitchen (Kitchen)
Switch Light_Kitchen (Kitchen)
Number Temp_Office (Office)
Switch Light_Office (Office)
Number Temp1_Cellar (Cellar)
Number Temp2_Cellar (Cellar)
```
`MyAverageHouseTemperature` should take `Temp_Kitchen`, `Temp_Office` and `AverageCellarTemperature` directly into account, but NOT `Temp1_Cellar` and `Temp2_Cellar`. 

Does that make sense?

htreu:
Imho it does _not_ make sense to have a GroupItem without state (`Kitchen` & `Office` in your example) and expect its members to somehow be taken into account when another group calculates its state. It would break the Principle of Least Surprise. When you expect `MyAverageHouseTemperature` to calculate the average temp from other groups those should themselves provide a suitable state for the calculation.

kaikreuzer:
Well, that's a matter of perspective.
To me it beakrs the Principle of Least Surprise if it does NOT regard any items that are within this group (by being part of a sub-group).

Imagine e.g. groups where you want to count how many items have a specific state. You simply cannot define your subgroups to have a state here, their only purpose is the grouping. You would have to define multiple such groups, which makes it all much more complicated.

pfink:
I fully agree @htreu, the behavior implemented by this PR would be the one that I'd expect. The approach suggested by @kaikreuzer maybe fits better for special use cases, but it's VERY unlikely that a user would expect that exact behavior.

> Imagine e.g. groups where you want to count how many items have a specific state. You simply cannot define your subgroups to have a state here, their only purpose is the grouping.

Why can't I define the subgroup to have a state here? Even if their "initial" purpose (counting states) was something else, I see no problem here to provide the state needed for them.

htreu:
My main concern and the reason for bringing up Principle of Least Surprise _was_ the inconsistent behaviour of groups with or without a function.
Thinking about this for a while now I guess we should really go with the suggestion @kaikreuzer did: it adds more flexibility by providing both: groups with functions and own state as terminal items vs. groups with transient behaviour in regard to their members.
It certainly makes a difference if a sub-group provides its own average over its members or if the sub-members are taken into the overall average calculation: Its `(((a + b) / 2) + ((c + d) / 2)) / 2` vs. `(a + b + c + d) / 4` which is not equivalent.

It would be nice to have some more opinions and comments on this one: @maggu2810, @sjka wdyt?

sjka:
If a group takes into account only its direct members, then as a user I have the choice: I can either make aggregated state (i.e. sub-groups) be a member of that specific group or the individual items directly, depending on what I want. If the group would always take into account the entire tree, then I we lose the opportunity to have such aggregated states. Therefore I'd like to second @htreu and @pfink here.

Considering that group memberships are defined on child-level anyway, this fits nicely in the way that at the item itself it is defined what should happen with its values, i.e. it's not a "calculate the state of subgroups x, y, z", but rather a "subgroup x or item y should contribute to group a" approach.

maggu2810:
> Its `(((a + b) / 2) + ((c + d) / 2)) / 2` vs. `(a + b + c + d) / 4` which is not equivalent.

@htreu Perhaps I am wrong but you have chosen a "bad" example. In your case it is equivalent. :wink: 
But I agree. If I would like to get the average of x groups, I expect the addition of the group states / values and a division by x. I would not expect that the members of the groups are added and divided by the number of elements.

htreu:
omg, you are right. I wanted to make a generic version of it and then came up with this...
Nonetheless calculating average differs using averages of subgroups vs. subelements directly.

The question here is: do we want simple groups w/o own state to be transient when it comes to calculating state for parent groups? It would be more flexible but comes with the cost of complexity.

With Kai´s example https://github.com/eclipse/smarthome/pull/4614#issuecomment-346617138 what would one expect for `MyAverageHouseTemperature` to do with the groups `Kitchen` and `Office`? Right now they would not add any value to the calculation. Is this bad configuration?

When it comes to the `COUNT` function, what should be the result here:

```
Group:Number:COUNT(".*") parent
Group subGroup1 (parent)

Switch switch1 (parent)
Switch sub-switch1 (subGroup)
Switch sub-switch2 (subGroup)
```

with all Switch items having a valid state?

kaikreuzer:
> this fits nicely in the way that at the item itself it is defined what should happen with its values

This is the main issue for me here. A group (without function) serves as a kind of "alias" for a set of items. I simply want to use this alias and not being forced to re-iterate everything on all the single items.

```
Group:Number:COUNT("ON") OnItems
Group:Number:COUNT("OFF") OnItems

Group MyLights (OnItems, OffItems)
Switch Light1 (MyLights)
Switch Light2 (MyLights)
Switch Light3 (MyLights)
...
Switch Light1000 (MyLights)
```
1. How would you solve this if groups are not considered?
2. What purpose would a Group-without-function at all serve as a member of a Group-with-function? If it has no purpose, it should not be a member in the first place. So when allowing to add it as a member, it should result in some changed behavior, shouldn't it?

pfink:
> How would you solve this if groups are not considered?

I don't think it can be solved unless you add both groups to all elements.

> What purpose would a Group-without-function at all serve as a member of a Group-with-function? If it has no purpose, it should not be a member in the first place. So when allowing to add it as a member, it should result in some changed behavior, shouldn't it?

You can still use it for UI or for rules to iterate its members. Additionally, commands to the parent group will be passed through to all of it's members.

@kaikreuzer: If I understood you correctly, this is what you suggest:
```
Group:Number:COUNT(".*") parent
Group subGroup1 (parent)
Group:Number:COUNT("ON") subGroup2 (parent)

Switch switch1 (parent)
Switch sub-switch1 (subGroup1) // is ON
Switch sub-switch2 (subGroup1) // is OFF
Switch sub-switch3 (subGroup2) // is ON
Switch sub-switch4 (subGroup2) // is OFF
Switch sub-switch5 (subGroup2) // is OFF
```
State `subGroup1`: None
State `subGroup2`: 1
State `parent`: 3 (sub-switch1 + sub-switch2 + sub-switch3)

I think it would be fine to do it like it. As I said, it's definitely **not** _Principle of Least Surprise_, it's more _Principle of most possibilities for power users_. But as I am a power user, I like the latter one ;-)

maggu2810:
@pfink What about `switch1`?

Just to be sure that I understand the comments correctly...

If a group A contains another group B...
* ... and group B has a function, then the state of the group B is ...
* ... and group B has not a function, then the members of group B are ...
taken for the calculation.

Just added a split between function and state.
If a group has a function it should also have a state.
If we decide if we use the members or the group state itself, we should decide if a function has been assigned.

pfink:
@maggu2810 Yes, switch1 should of course also be counted, I corrected the example.

htreu:
Why would in the above example the state of `parent` be 5? The group function is defined to `COUNT(".*")` which simply counts all members. By accident the group function for `subGroup2` is also `COUNT` but how should this be represented in the state which `parent` calculates?
Imagine another definition of subGroup2: `Group:Number:AND(ON, OFF) subGroup2 (parent)` which would change the state of subGroup2 to `OFF`. Why should this change the result for `parent`? Should the calculation on a parent group also depend on the type of function on sub groups? That would add another level of complexity which is hard to explain and also hard to maintain.

Imho we have three options on how to proceed here:
1. Define all groups transitive, regardless of their function/state to forward their members recursively. This will allow for groups used as aliases.
1. Define all groups as terminal members of parent groups. When state is available and parent group function calculates, the sub-group state is taken into account.
1. Define groups w/o base item/function as transitive, passing their members recursively to parent group calculations. Define groups with base item/function as terminal and taking their state into account during parent group calculation as of any other item.

maggu2810:
@htreu Sorry, I missed to look at the function at all.
It does not make sense, that the state "2" is added if `COUNT(".*")` is used, as it makes no difference if the state if 1, 2, 3, 4 (it would be different if `SUM` is used).

As I am not very familiar with groups:
What is the state of a group without a function but only non-group members? UnDefType.NULL?

Am I right that (3) is what I described as my understanding above?

htreu:
> What is the state of a group without a function but only non-group members? UnDefType.NULL?

Correct, from what I see in the code all items are initialised with `state = UnDefType.NULL`.

> Am I right that (3) is what I described as my understanding above?

Correct.

pfink:
Oh man, sorry. I completely mixed up the number examples with the count examples. I revised my post again and I think it finally makes sense and is appropriately showing what Kai meant...

htreu:
So if we have consensus I will author the changes and update this PR.

pfink:
Would be fine for me. But an open question would be how to handle group states that do not fit into the group function (e.g. the subgroup has the state `ON` whilst the aggregation function is `AVG`). Should the group and its members skipped in this case or should it be treated the same way as a group with no state (so group the function is applied to its members)?

maggu2810:
If we agree on option 3 (that should be equal to https://github.com/eclipse/smarthome/pull/4614#issuecomment-347780449) I don't see any difference between the subgroup state `ON` of an non-group item member that state is `ON`.

docs/documentation/concepts/items.md
```diff
@@ -28,11 +28,43 @@ The following item types are currently available (alphabetical order):
...
-Group Items can derive their own state depending on their member items.
-
-- AVG displays the average of the item states in the group.
-- OR displays an OR of the group, typically used to display whether any item in a group has been set.
-- other aggregations: AND, SUM, MIN, MAX, NAND, NOR
+## Group Items
+
+Group items collect other items into groups.
+Group items can themselves be members of other group items.
+Cyclic membership is not forbidden but strongly not recommended.
```
kaikreuzer: Really? I thought to remember that we still had issues in such situations somewhere.

docs/documentation/concepts/items.md
```diff
@@ -28,11 +28,43 @@ The following item types are currently available (alphabetical order):
...
+User interfaces might display group items as single entries and provide navigation to its members.
+
+Example for a Group item simply collection other items:
```
kaikreuzer: simply collection -> as a simple collection of

docs/documentation/concepts/items.md
```diff
@@ -28,11 +28,43 @@ The following item types are currently available (alphabetical order):
...
+```
+    Group groundFloor
+    Switch kitchenLight (groundFloor)
+    Switch livingroomLight (groundFloor)
+``` 
+
+In addition Group items can calculate their own state derived from their members using a specific group function.
```
kaikreuzer: -> "In addition, "

docs/documentation/concepts/items.md
```diff
@@ -28,11 +28,43 @@ The following item types are currently available (alphabetical order):
...
+
+### Derive group state from member items
```
kaikreuzer: Other headlines here use capitalised words

docs/documentation/concepts/items.md
kaikreuzer: I would actually remove the whole sentence - this content comes in the next section again.

docs/documentation/concepts/items.md
```diff
@@ -28,11 +28,43 @@ The following item types are currently available (alphabetical order):
...
+
+Group items can derive their own state depending on their member items.
```
kaikreuzer: depending on -> from

docs/documentation/concepts/items.md
```diff
@@ -28,11 +28,43 @@ The following item types are currently available (alphabetical order):
...
+To derive a state the group item must be constructed using a base item and a group function.
+When calculating state, group functions will recursively traverse the group's members and also take members of subgroups into account.
```
kaikreuzer: calculating state -> calculating the state
group functions will recursively traverse -> group functions recursively traverse

docs/documentation/concepts/items.md
```diff
@@ -28,11 +28,43 @@ The following item types are currently available (alphabetical order):
...
+If a subgroup however defines a state on its own (having base item & group function set) traversal stops and the state of the subgroup member is taken. 
+
+Available group functions:
+
+| Function           | Parameters  | Base Item | Description |
```
kaikreuzer: Could you transform that to a pretty table? https://www.tablesgenerator.com/markdown_tables

docs/documentation/concepts/items.md
```diff
@@ -28,11 +28,43 @@ The following item types are currently available (alphabetical order):
...
+|--------------------|-------------|-----------|-------------|
+| EQUALITY           | - | \<all\> | Sets the state of the members if all have equal state. Otherwise UNDEF is set |
+| AND, OR, NAND, NOR | \<activeState\>, \<passiveState\> | \<all\> (must match active & passive state) | Sets the <activeState> if the member state <activeState> evaluates to `true` under the boolean term. Otherwise the <passiveState> is set. |
+| SUM, AVG, MIN, MAX | - | Number | Sets the state according to the arithmetic function over all member states. |
+| COUNT              | \<regular expression\> | Number | Sets the state to the number of members matching the given regular expression with their states. |
+
+Examples for derived states on group items when declared in the item DSL:
+- `Group:Number:COUNT(".*")` counts all members of the group matching the given regular expression, here any character or state (simply count all members).
```
kaikreuzer: add empty line

docs/documentation/concepts/items.md
htreu: At least the GroupItem is capable of collecting its members and not to loop indefinitely ;-) What du you suggest to state in the docs?

kaikreuzer: How about "Cyclic memberships must be avoided."?

PR #4618, opened 2017-11-23
Inbox: Representation properties are unique per binding

### LINKED ISSUE #4191: Inbox auto-ignores discovery results when representation property value matches, but thing type does not
The inbox auto-ignore feature ignores discovery results if the representation property value of the discovery result is the same as the representation property value of an existing thing. In consequence, discovery results are ignored even if there is an existing thing of a different type with the same representation property value.

Context: The representation property value is something that is in some sense unique for all things of that type of thing (e.g., serial number, ID assigned by bridge, MAC address, ...).

 I believe that one should not compare representation property values of different thing types, as this could clash (e.g., if the serial numbers for some type of light bulb look, by chance, the same as the serial numbers of some type of switch).

I added a unit test illustrating this behavior here: https://github.com/hsudbrock/smarthome/tree/thing-type-not-considered-when-autoignoring-in-inbox

In a discussion with @kaikreuzer, it came up that it might be necessary to also not compare representation property values if things/discovery results are attached to different bridges, to not ignore a discovery result for one bridge just because there is a thing attached to another bridge with the same ID assigned to by the bridge.

triller-telekom:
Do not ignore a discovery result if it has the same
representation property as a thing that already exists, if
their bindingId differs.

Fixes #4191

Also-by: Henning Sudbrock <henning.sudbrock@telekom.de>

### DIFF
--- a/bundles/config/org.eclipse.smarthome.config.discovery.test/src/test/java/org/eclipse/smarthome/config/discovery/internal/AutomaticInboxProcessorTest.java
+++ b/bundles/config/org.eclipse.smarthome.config.discovery.test/src/test/java/org/eclipse/smarthome/config/discovery/internal/AutomaticInboxProcessorTest.java
@@ -306,6 +306,27 @@ public class AutomaticInboxProcessorTest {
     }
 
     @Test
+    public void testOneThingOutOfTwoWithSameRepresentationPropertyButDifferentBindingIdIsBeingRemoved() {
+        inbox.add(DiscoveryResultBuilder.create(THING_UID).withProperty(DEVICE_ID_KEY, DEVICE_ID)
+                .withRepresentationProperty(DEVICE_ID_KEY).build());
+        inbox.setFlag(THING_UID, DiscoveryResultFlag.IGNORED);
+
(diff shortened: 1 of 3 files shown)


triller-telekom:
@afuechsel Since you created the `AutomaticInboxProcessor`, did I miss anything?

afuechsel:
Lgtm

triller-telekom:
@afuechsel yesterday @hsudbrock mentioned that there might be a second point in the code where, i.e. i am not considering the scenario where a thing exists and a new discovery result is created. But @hsudbrock thought there might also be a scenario where a discovery result exists and you create a thing.

Since you wrote that code I just want to double check that I am not missing anything, thanks :)

sjka:
What about `AutomaticInboxProcessor.removeFromInbox(..)`?

triller-telekom:
@sjka Thanks for spotting this. I have added it in an additional commit.

afuechsel:
Thanks, @sjka, I didn't think about this too.

hsudbrock:
@triller-telekom What about `AutomaticInboxProcessor#thingAdded`?. Do you think it makes sense to extend the unit test suite further, since these additional changes are not yet covered by unit tests?

triller-telekom:
Thank you @hsudbrock it sound logical that there is another point in the code, thanks for pointing me to it. I have also created additional test cases.

PR #4621, opened 2017-11-23
Added rawrocker profiles closes #4542

### LINKED ISSUE #4542: System trigger profiles
PRs #4108 & #4516 so far only introduced the [RawButtonToggleSwitchProfile](https://github.com/eclipse/smarthome/blob/9cddf010bae0cc7fcc181aeb95716f4bf0bf01c3/bundles/core/org.eclipse.smarthome.core.thing/src/main/java/org/eclipse/smarthome/core/thing/internal/profiles/RawButtonToggleSwitchProfile.java) as an example for a trigger profile. 

As discussed in the ESH forum ([here](https://www.eclipse.org/forums/index.php/t/1089956/) and [here](https://www.eclipse.org/forums/index.php/t/1089653/)) and in the [OH community](https://community.openhab.org/t/enocean-bindings-opencean-or-aleoncean/12493/58), it would make sense to have system profiles for all the system trigger channels filling the following matrix:

| | Switch | Dimmer | Color | Rollershutter |
|:---|:---|:---|:---|:---|
| button | | | | |
| rawbutton | rawbutton-toggle-switch | | | |
| rocker (see #4541) | | | | |
| rawrocker (see #4541) | | | | |

Anything else missing?

kungfoolfighting:
This PR creates system profiles for rawrocker channels.
There is still need for a profile to convert to roller shutters after this PR.
Signed-off-by: Jan Kemmler <kickass_kemmler@gmx.de>

### DIFF
--- a/bundles/core/org.eclipse.smarthome.core.thing/src/main/java/org/eclipse/smarthome/core/thing/internal/profiles/RawRockerToDimmerProfile.java
+++ b/bundles/core/org.eclipse.smarthome.core.thing/src/main/java/org/eclipse/smarthome/core/thing/internal/profiles/RawRockerToDimmerProfile.java
@@ -1,10 +1,14 @@
 /**
- * Copyright (c) 2010-2017 by the respective copyright holders.
+ * Copyright (c) 2014,2017 Contributors to the Eclipse Foundation
  *
- * All rights reserved. This program and the accompanying materials
- * are made available under the terms of the Eclipse Public License v1.0
- * which accompanies this distribution, and is available at
- * http://www.eclipse.org/legal/epl-v10.html
+ * See the NOTICE file(s) distributed with this work for additional
+ * information regarding copyright ownership.
+ *
+ * This program and the accompanying materials are made available under the
+ * terms of the Eclipse Public License 2.0 which is available at
+ * http://www.eclipse.org/legal/epl-2.0
+ *
+ * SPDX-License-Identifier: EPL-2.0
  */
 
 package org.eclipse.smarthome.core.thing.internal.profiles;
--- a/bundles/core/org.eclipse.smarthome.core.thing/src/main/java/org/eclipse/smarthome/core/thing/internal/profiles/RawRockerToDimmerProfile.java
+++ b/bundles/core/org.eclipse.smarthome.core.thing/src/main/java/org/eclipse/smarthome/core/thing/internal/profiles/RawRockerToDimmerProfile.java
@@ -1,0 +1,125 @@
+/**
+ * Copyright (c) 2010-2017 by the respective copyright holders.
+ *
+ * All rights reserved. This program and the accompanying materials
+ * are made available under the terms of the Eclipse Public License v1.0
+ * which accompanies this distribution, and is available at
+ * http://www.eclipse.org/legal/epl-v10.html
+ */
+
+package org.eclipse.smarthome.core.thing.internal.profiles;
+
+import java.util.concurrent.ScheduledFuture;
+import java.util.concurrent.TimeUnit;
+
+import org.eclipse.jdt.annotation.NonNullByDefault;
+import org.eclipse.jdt.annotation.Nullable;
+import org.eclipse.smarthome.core.library.types.IncreaseDecreaseType;
+import org.eclipse.smarthome.core.library.types.OnOffType;
+import org.eclipse.smarthome.core.thing.CommonTriggerEvents;
+import org.eclipse.smarthome.core.thing.profiles.ProfileCallback;
+import org.eclipse.smarthome.core.thing.profiles.ProfileContext;
+import org.eclipse.smarthome.core.thing.profiles.ProfileTypeUID;
+import org.eclipse.smarthome.core.thing.profiles.SystemProfiles;
+import org.eclipse.smarthome.core.thing.profiles.TriggerProfile;
+import org.eclipse.smarthome.core.types.Command;
+import org.eclipse.smarthome.core.types.State;
+
+/**
+ * The {@link RawRockerToDimmerProfile} transforms rocker switch channel events into dimmer commands.
+ *
+ * @author Jan Kemmler - Initial contribution
+ */
+@NonNullByDefault
+public class RawRockerToDimmerProfile implements TriggerProfile {
+
+    private final ProfileCallback callback;
+
+    ProfileContext context;
+
+    @Nullable
+    private ScheduledFuture<?> dimmFuture;
+    @Nullable
+    private ScheduledFuture<?> timeoutFuture;
+
+    private long pressedTime = 0;
+
+    final class DimmerIncreaseDecreaseTask implements Runnable {
+        private ProfileCallback callback;
+        private Command command;
+
+        DimmerIncreaseDecreaseTask(ProfileCallback callback, Command command) {
+            this.callback = callback;
+            this.command = command;
+        }
+
+        @Override
+        public void run() {
+            callback.sendCommand(command);
+        }
+    };
+
+    @SuppressWarnings("null")
+    RawRockerToDimmerProfile(ProfileCallback callback, ProfileContext context) {
+        this.callback = callback;
+        this.context = context;
+    }
+
+    @Override
+    public ProfileTypeUID getProfileTypeUID() {
+        return SystemProfiles.RAWROCKER_TO_DIMMER;
+    }
+
+    @Override
+    public void onStateUpdateFromItem(State state) {
+
+    }
+
+    @Override
+    public void onTriggerFromHandler(String event) {
+        if (CommonTriggerEvents.DIR1_PRESSED.equals(event)) {
+            buttonPressed(IncreaseDecreaseType.INCREASE);
+        } else if (CommonTriggerEvents.DIR1_RELEASED.equals(event)) {
+            buttonReleased(OnOffType.ON);
+        } else if (CommonTriggerEvents.DIR2_PRESSED.equals(event)) {
+            buttonPressed(IncreaseDecreaseType.DECREASE);
+        } else if (CommonTriggerEvents.DIR2_RELEASED.equals(event)) {
+            buttonReleased(OnOffType.OFF);
+        }
+    }
+
+    private void buttonPressed(Command commandToSend) {
+        if (null != timeoutFuture) {
+            timeoutFuture.cancel(false);
+        }
+        if (null != dimmFuture) {
+            dimmFuture.cancel(false);
+        }
+
+        dimmFuture = context.getExecutorService().scheduleWithFixedDelay(
+                new DimmerIncreaseDecreaseTask(callback, commandToSend), 550, 200, TimeUnit.MILLISECONDS);
+        timeoutFuture = context.getExecutorService().schedule(new Runnable() {
+            @Override
+            public void run() {
+                if (null != dimmFuture) {
+                    dimmFuture.cancel(false);
+                }
+            }
+        }, 10000, TimeUnit.MILLISECONDS);
+        pressedTime = System.currentTimeMillis();
+    }
+
+    private void buttonReleased(Command commandToSend) {
+        if (null != timeoutFuture) {
+            timeoutFuture.cancel(false);
+        }
+        if (null != dimmFuture) {
+            dimmFuture.cancel(false);
+        }
+
+        if (System.currentTimeMillis() - pressedTime <= 500) {
+            callback.sendCommand(commandToSend);
+        }
+    }
+
+}
--- a/bundles/core/org.eclipse.smarthome.core.thing/src/main/java/org/eclipse/smarthome/core/thing/internal/profiles/RawRockerToOnOffProfile.java
+++ b/bundles/core/org.eclipse.smarthome.core.thing/src/main/java/org/eclipse/smarthome/core/thing/internal/profiles/RawRockerToOnOffProfile.java
@@ -1,10 +1,14 @@
 /**
- * Copyright (c) 2010-2017 by the respective copyright holders.
+ * Copyright (c) 2014,2017 Contributors to the Eclipse Foundation
  *
- * All rights reserved. This program and the accompanying materials
- * are made available under the terms of the Eclipse Public License v1.0
- * which accompanies this distribution, and is available at
- * http://www.eclipse.org/legal/epl-v10.html
+ * See the NOTICE file(s) distributed with this work for additional
+ * information regarding copyright ownership.
+ *
+ * This program and the accompanying materials are made available under the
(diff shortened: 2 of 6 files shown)


bundles/core/org.eclipse.smarthome.core.thing/src/main/java/org/eclipse/smarthome/core/thing/internal/profiles/RawRockerToDimmerProfile.java
```diff
@@ -0,0 +1,129 @@
...
+
+    private void buttonPressed(Command commandToSend) {
+        if (null != timeoutFuture) {
+            timeoutFuture.cancel(false);
+        }
+        if (null != dimmFuture) {
+            dimmFuture.cancel(false);
+        }
+
+        dimmFuture = context.getExecutorService().scheduleWithFixedDelay(
```
sjka: When you write 

```java 
dimmFuture = context.getExecutorService().scheduleWithFixedDelay(() -> callback.sendCommand(commandToSend), 550, 200, TimeUnit.MILLISECONDS);
```

instead, you will save 15 lines of superfluous code above.

bundles/core/org.eclipse.smarthome.core.thing/src/main/java/org/eclipse/smarthome/core/thing/internal/profiles/RawRockerToDimmerProfile.java
```diff
@@ -0,0 +1,129 @@
...
+                new DimmerIncreaseDecreaseTask(callback, commandToSend), 550, 200, TimeUnit.MILLISECONDS);
+        timeoutFuture = context.getExecutorService().schedule(new Runnable() {
```
sjka: also here: please "CMD/CTRL+1" -> "Convert to lambda expression"

bundles/core/org.eclipse.smarthome.core.thing/src/main/java/org/eclipse/smarthome/core/thing/internal/profiles/RawRockerToDimmerProfile.java
```diff
@@ -0,0 +1,129 @@
...
+            @Override
+            public void run() {
+                if (null != dimmFuture) {
+                    dimmFuture.cancel(false);
+                }
+            }
+        }, 10000, TimeUnit.MILLISECONDS);
```
sjka:
```java
}, 10, TimeUnit.SECONDS);
```

would be much easier to read.

bundles/core/org.eclipse.smarthome.core.thing/src/main/java/org/eclipse/smarthome/core/thing/internal/profiles/RawRockerToDimmerProfile.java
```diff
@@ -0,0 +1,129 @@
...
+ * The {@link RawRockerToDimmerProfile} transforms rocker switch channel events into dimmer commands.
+ *
+ * @author Jan Kemmler - Initial contribution
+ */
+@NonNullByDefault
+public class RawRockerToDimmerProfile implements TriggerProfile {
+
+    private final ProfileCallback callback;
+
+    ProfileContext context;
```
sjka: `private final`

bundles/core/org.eclipse.smarthome.core.thing/src/main/java/org/eclipse/smarthome/core/thing/internal/profiles/RawRockerToDimmerProfile.java
```diff
@@ -0,0 +1,129 @@
...
+            this.command = command;
+        }
+
+        @Override
+        public void run() {
+            callback.sendCommand(command);
+        }
+    };
+
+    @SuppressWarnings("null")
```
sjka: That _shouldn't_ be necessary

bundles/core/org.eclipse.smarthome.core.thing/src/main/java/org/eclipse/smarthome/core/thing/internal/profiles/RawRockerToOnOffProfile.java
```diff
@@ -0,0 +1,61 @@
...
+    RawRockerToOnOffProfile(ProfileCallback callback) {
+        this.callback = callback;
+    }
+
+    @Override
+    public ProfileTypeUID getProfileTypeUID() {
+        return SystemProfiles.RAWROCKER_TO_ON_OFF;
+    }
+
+    /**
```
sjka: please don't repeat the javadoc of an overridden method (unless the meaning somehow changed, of course). It's already hard to keep it up-to-date at all, and I fear it's gonna be a complete mess with copies like this.

bundles/core/org.eclipse.smarthome.core.thing/src/main/java/org/eclipse/smarthome/core/thing/internal/profiles/RawRockerToDimmerProfile.java
```diff
@@ -0,0 +1,129 @@
...
+ * See the NOTICE file(s) distributed with this work for additional
+ * information regarding copyright ownership.
+ *
+ * This program and the accompanying materials are made available under the
+ * terms of the Eclipse Public License 2.0 which is available at
+ * http://www.eclipse.org/legal/epl-2.0
+ *
+ * SPDX-License-Identifier: EPL-2.0
+ */
+
```
kaikreuzer: Just out of curiosity: Was this empty line automatically generated by the license plugin? If so, we seem to have a bug there...

bundles/core/org.eclipse.smarthome.core.thing/src/main/java/org/eclipse/smarthome/core/thing/profiles/SystemProfiles.java
```diff
@@ -32,4 +34,12 @@
             .newTrigger(RAWBUTTON_TOGGLE_SWITCH, "Raw Button Toggle").withSupportedItemTypes(CoreItemFactory.SWITCH)
             .withSupportedChannelTypeUIDs(DefaultSystemChannelTypeProvider.SYSTEM_RAWBUTTON.getUID()).build();
 
-}
+    TriggerProfileType RAWBROCKER_TO_ON_OFF_TYPE = ProfileTypeBuilder
```
kaikreuzer: I assume it should be `RAWROCKER_TO_ON_OFF_TYPE` without the "B" :-)

maggu2810: There is no a `RAWBUTTON_TOGGLE_SWITCH` without a "to" between "rawbutton" and "toggle switch".
There is a rawrocker to dimmer that states the type and not the values the conversion is done two (same as toggle switch).
There is a rawrocker to on off that is using the values the conversion is applied to.

Couldn't we make it more consistent and use a "to" for all or for none?
Shouldn't we use "switch" instead of "on off"?

bundles/core/org.eclipse.smarthome.core.thing/src/main/java/org/eclipse/smarthome/core/thing/internal/profiles/RawRockerToOnOffProfile.java
```diff
@@ -0,0 +1,61 @@
...
+import org.eclipse.smarthome.core.thing.profiles.SystemProfiles;
+import org.eclipse.smarthome.core.thing.profiles.TriggerProfile;
+import org.eclipse.smarthome.core.types.State;
+
+/**
+ * The {@link RawRockerToOnOffProfile} transforms rocker switch channel events into ON and OFF commands.
+ *
+ * @author Jan Kemmler - Initial contribution
+ */
+public class RawRockerToOnOffProfile implements TriggerProfile {
```
maggu2810: I assume this profile is a good candidate for a configuration (up = on or up = off).

bundles/core/org.eclipse.smarthome.core.thing/src/main/java/org/eclipse/smarthome/core/thing/internal/profiles/RawRockerToDimmerProfile.java
```diff
@@ -0,0 +1,129 @@
...
+import org.eclipse.smarthome.core.types.Command;
+import org.eclipse.smarthome.core.types.State;
+
+/**
+ * The {@link RawRockerToDimmerProfile} transforms rocker switch channel events into dimmer commands.
+ *
+ * @author Jan Kemmler - Initial contribution
+ */
+@NonNullByDefault
+public class RawRockerToDimmerProfile implements TriggerProfile {
```
maggu2810: Perhaps I am wrong, but I see some concurrency in this class.
e.g. dimmFeature is read and written from different threads without synchronizations, button pressed is not synchronized, ...

bundles/core/org.eclipse.smarthome.core.thing/src/main/java/org/eclipse/smarthome/core/thing/profiles/SystemProfiles.java
kungfoolfighting: Oh, this is a stupid mistake.
@maggu2810 I didn't use switch for the specific reason that this profile can be used for dimmers as well, if you only want to have on and off commands sent to the dimmer. That way it doesn't confuse anyone into thinking it can only be used with switches.
What do you think?

bundles/core/org.eclipse.smarthome.core.thing/src/main/java/org/eclipse/smarthome/core/thing/internal/profiles/RawRockerToOnOffProfile.java
kungfoolfighting: This is a complicated topic. I have been thinking about profiles and their configuration.
This whole thing quickly leads to philosophical issues about where to solve the complexity issue with many different combinations of items and channels and possible translations between events and commands.
As Kai said in another post, there is also the question of having a toggle profile work on a rawrocker channel, which is another example where you now start doing tons of configuration for profiles, so that you can use them for more channel/item combinations.

The more configuration options and fewer individual profiles we use, the more complex the definition of items in *.items files will become.

The configuration for a RawRockerToDimmerProfile for example would then contain parameters for initial delay time, interval time for increasing and decreasing and definition of which event triggers up and which event triggers down. This also means that the user that wants to use this profile also has to know exactly which events each channel emits.

On the other hand, if we split these functionalities into more different profiles, the user has a longer list of different profiles from which to choose the correct one. This is also some added complexity, but it seems a bit less to me.
What do you think?
Do we want to prefer more configuration options over more profiles?

In this specific example with mapping the events upside-down, I don't even know that we really even want to have this functionality. It seems relatively specific and can easily be done via rules, if someone wants to solve the problem of having up and down exchanged.

Also: I have no idea how configurations for profiles work, is there some documentation or an example on how to do it?

bundles/core/org.eclipse.smarthome.core.thing/src/main/java/org/eclipse/smarthome/core/thing/internal/profiles/RawRockerToDimmerProfile.java
kungfoolfighting: Ah okay, that's what lambdas look like in Java. Thanks, this is much nicer!

bundles/core/org.eclipse.smarthome.core.thing/src/main/java/org/eclipse/smarthome/core/thing/internal/profiles/RawRockerToDimmerProfile.java
kungfoolfighting: No, it wasn't, I copy/pasted the header from another file and might have added the extra line manually.
I prefer doing it manually, since the automated script also adds headers to files that are unrelated to what I am working on and then I have to manually exclude those things from commits, as has happened in a previous PR of mine.

bundles/core/org.eclipse.smarthome.core.thing/src/main/java/org/eclipse/smarthome/core/thing/internal/profiles/RawRockerToOnOffProfile.java
maggu2810: There is no need that you add configuration support in this PR.
It has been just a comment WRT the forum entry about profile configuration.

For such simple stuff as swap up / down, I would like to see a configuration some time.
If I want to combine a rocker with a light, I don't want to write a rule, just because I want to switch the on / off use case. Sure, rules are an option, but isn't the intention of profiles, that we don't need to create rules for such normal use cases.

kungfoolfighting: Okay, I will add this at some point. I am still curious, however, how profile configurations work. Got a link?

sjka: They are configured at the links, not at the items (there is no support for it yet in the the Paper UI though). And indeed, I got a link (literally 😄 ): 
https://github.com/eclipse/smarthome/blob/36c9e2717e4f89cf60a0d885e7ecc6069b338f48/bundles/model/org.eclipse.smarthome.model.thing.tests/src/org/eclipse/smarthome/model/thing/tests/GenericItemChannelLinkProviderJavaTest.java#L164-L226

Within the profiles you then can get the configuration object via the `ProfileContext`.

kungfoolfighting:
@maggu can you please check if the concurrency issues are fixed now? Thank you!

kungfoolfighting:
@maggu I wondered the same thing and googled it. Apparently it should be fine.

bundles/core/org.eclipse.smarthome.core.thing/src/main/java/org/eclipse/smarthome/core/thing/internal/profiles/RawRockerToDimmerProfile.java
kaikreuzer: Surely it is ok if you do it manually. You just should not forget to do so - the files still contain the empty line...

kaikreuzer:
Many thanks!

PR #4623, opened 2017-11-24
use fresh thing instances within the profile callback

### LINKED ISSUE #4622: ProfileCallbackImpl may contain reference to an old Thing
While debugging https://github.com/openhab/openhab2-addons/issues/2854 I've encountered references to Things not properly being updated in the ProfileCallbackImpl.

As you can read in the referenced issue, the initialization of the HarmonyHubHandler can take more than 5 seconds. The warnings ofcourse does not cause any issues. What does seem to be an issue is that during initialization of the handler a new Thing is created and the handler of the old Thing is set to `null` in the ThingManager at [line 453](https://github.com/eclipse/smarthome/blob/92b4f7d1dcd262dfe9c33ca458853140cdd145af/bundles/core/org.eclipse.smarthome.core.thing/src/main/java/org/eclipse/smarthome/core/thing/internal/ThingManager.java#L453
):

        public void thingUpdated(final Thing thing, ThingTrackerEvent thingTrackerEvent) {
            ...
                if (oldThing != thing && oldThing != null) {
                    oldThing.setHandler(null);
                }
            ...
        }

Apparently `CommunicationManager.getProfile(..)` does not update the oldThing in the ProfileCallBackImpl to the newThing. So when a command is sent to the channel a ProfileCallBackImpl is retreived containing the oldThing with a null handler. I.e. resulting in the logging:

> 23:06:30.764 WARN  o.e.s.c.t.i.p.ProfileCallbackImpl[:122] - Cannot delegate update 'PowerOff' for item 'harmonyhub_hub_Test_currentActivity' to handler for channel 'harmonyhub:hub:Test:currentActivity', because no handler is assigned. Maybe the binding is not installed or not propertly initialized.

A temporary working solution for me is to comment parts of the code in CommunicationManager.getProfile(..) so it always creates a new (working) profile:

        private Profile getProfile(ItemChannelLink link, Item item, Thing thing) {
            synchronized (profiles) {
                // Profile profile = profiles.get(link.getUID());
                // if (profile != null) {
                //     return profile;
                // }
                ProfileTypeUID profileTypeUID = determineProfileTypeUID(link, item, thing);
                if (profileTypeUID != null) {
                    Profile profile = getProfileFromFactories(profileTypeUID, link, createCallback(link));
                    if (profile != null) {
                        profiles.put(link.getUID(), profile);
                        return profile;
                    }
                }
                return new NoOpProfile();
            }
        }

sjka:
fixes #4622

### DIFF
--- a/bundles/core/org.eclipse.smarthome.core.thing/src/main/java/org/eclipse/smarthome/core/thing/internal/CommunicationManager.java
+++ b/bundles/core/org.eclipse.smarthome.core.thing/src/main/java/org/eclipse/smarthome/core/thing/internal/CommunicationManager.java
@@ -134,9 +134,8 @@ public class CommunicationManager implements EventSubscriber, RegistryChangeList
     }
 
     private ProfileCallback createCallback(ItemChannelLink link) {
-        Item item = itemRegistry.get(link.getItemName());
-        Thing thing = getThing(link.getLinkedUID().getThingUID());
-        return new ProfileCallbackImpl(eventPublisher, link, thing, item);
+        return new ProfileCallbackImpl(eventPublisher, link, thingUID -> getThing(thingUID),
+                itemName -> getItem(itemName));
     }
 
     private ProfileTypeUID determineProfileTypeUID(ItemChannelLink link, Item item, Thing thing) {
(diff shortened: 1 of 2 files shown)


sjka:
This is _not_ API breaking for a change - I just clicked a few pixels too high 😄

PR #4624, opened 2017-11-24
Remove org.osgi.service.component.annotations import from manifests

wborn:
See https://github.com/openhab/openhab2-addons/pull/2880

### DIFF
--- a/bundles/config/org.eclipse.smarthome.config.core/META-INF/MANIFEST.MF
+++ b/bundles/config/org.eclipse.smarthome.config.core/META-INF/MANIFEST.MF
@@ -33,6 +33,5 @@ Import-Package:
  org.eclipse.smarthome.core.net,
  org.osgi.framework,
  org.osgi.service.component,
- org.osgi.service.component.annotations;resolution:=optional,
  org.slf4j
 Service-Component: OSGI-INF/*.xml
--- a/bundles/config/org.eclipse.smarthome.config.discovery.mdns/META-INF/MANIFEST.MF
+++ b/bundles/config/org.eclipse.smarthome.config.discovery.mdns/META-INF/MANIFEST.MF
@@ -17,6 +17,5 @@ Import-Package:
  org.eclipse.smarthome.io.transport.mdns.discovery,
  org.osgi.framework,
  org.osgi.service.component,
- org.osgi.service.component.annotations;resolution:=optional,
  org.slf4j
 Service-Component: OSGI-INF/*.xml
--- a/bundles/config/org.eclipse.smarthome.config.discovery.upnp/META-INF/MANIFEST.MF
+++ b/bundles/config/org.eclipse.smarthome.config.discovery.upnp/META-INF/MANIFEST.MF
@@ -18,6 +18,5 @@ Import-Package:
  org.jupnp.registry,
  org.osgi.framework,
  org.osgi.service.component,
- org.osgi.service.component.annotations;resolution:=optional,
  org.slf4j
 Service-Component: OSGI-INF/*.xml
--- a/bundles/config/org.eclipse.smarthome.config.discovery/META-INF/MANIFEST.MF
+++ b/bundles/config/org.eclipse.smarthome.config.discovery/META-INF/MANIFEST.MF
@@ -34,6 +34,5 @@ Import-Package:
  org.osgi.framework,
  org.osgi.service.cm,
  org.osgi.service.component,
- org.osgi.service.component.annotations;resolution:=optional,
  org.slf4j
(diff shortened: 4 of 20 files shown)


kaikreuzer:
Hm, Travis fails with
```
[ERROR] /home/travis/build/eclipse/smarthome/bundles/io/org.eclipse.smarthome.io.transport.mqtt/src/main/java/org/eclipse/smarthome/io/transport/mqtt/MqttService.java:[49] 
[ERROR] 	@Component(immediate = true, service = { MqttService.class }, configurationPid = {
[ERROR]         "org.eclipse.smarthome.mqtt" }, property = "service.pid=org.eclipse.smarthome.mqtt")
[ERROR] 	                                                                                 ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
[ERROR] Type mismatch: cannot convert from String[] to String
```
Seems we have a different annotation definition now?

From the error, it seems the MqttService should be defined as 
```
@Component(immediate = true, service = { MqttService.class }, configurationPid = {
        "org.eclipse.smarthome.mqtt" }, property = { "service.pid=org.eclipse.smarthome.mqtt" })
```
i.e. adding curly brackets around the property. Do you want to give that fix a try?

wborn:
Thanks for the suggestion. I'll give it a try later this weekend!

kaikreuzer:
Argh. Now I am completely puzzled:

```
[ERROR] /home/travis/build/eclipse/smarthome/bundles/io/org.eclipse.smarthome.io.transport.mqtt/src/main/java/org/eclipse/smarthome/io/transport/mqtt/MqttService.java:[49] 
[ERROR] 	@Component(immediate = true, service = { MqttService.class }, configurationPid = {
[ERROR]         "org.eclipse.smarthome.mqtt" }, property = { "service.pid=org.eclipse.smarthome.mqtt" })
[ERROR] 	                                                                                 ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
[ERROR] Type mismatch: cannot convert from String[] to String
```

sjka:
The problem here is the `configurationPid` - that one is expected to be a plain string and _not_ an array (i.e. drop the curly brackets there).

kaikreuzer:
Before merging, we should make sure that it does not break the XML generation, see https://github.com/openhab/openhab2-addons/pull/2880#issuecomment-346931292.

wborn:
This PR has been rebased on the most recent master. The compilation errors got fixed using @sjka proposed fix. 

I've had a look inside the JARs and see that XML is properly generated using the annotations. However after also having a look inside the JARs resulting from https://github.com/openhab/openhab2-addons/pull/2880 these also have properly generated XML. So most likely something else will still cause issues when the imports are removed.

wborn:
Apparently the OSGI-INF directories inside the JARs produced by Jenkins for the online-repo of the [openHAB build # 1096](https://openhab.jfrog.io/openhab/webapp/#/builds/openHAB-Distribution/1096/1511563757684/published/org.openhab.distro:online-repo:2.2.0-SNAPSHOT) do not contain generated XML.

There is however generated XML inside my local build of an online-repo after building openhab2-addons with [PR #2280](https://github.com/openhab/openhab2-addons/pull/2880) applied.

So the missing generated XML files may be due to an issue with the openHAB Jenkins setup.

kaikreuzer:
> I've had a look inside the JARs and see that XML is properly generated using the annotations. 

Which jars did you look at exactly? The ones produced by a local Maven build? If so, did you make sure that you closed your IDE before running the build and that you had deleted the IDE-generated XMLs? Otherwise the local build simply packages those and all seems fine (even if the maven-scr plugin fails to generated them).

wborn:
> Which jars did you look at exactly? The ones produced by a local Maven build? If so, did you make sure that you closed your IDE before running the build and that you had deleted the IDE-generated XMLs? 

I've looked into the JARs produced by my local Maven build after closing Eclipse and deleting the IDE-generated XML files.

kaikreuzer:
Ok, thanks. So all we have to find out is why the XML generation isn't working on openhab2-addons then - it all looks pretty much the same to me 😲

wborn:
It would be nice to see if the [ESH Nightly](https://ci.eclipse.org/smarthome/job/SmartHomeDistribution-Nightly/buildTimeTrend) Jenkins build correctly adds the generated XML. However it seems to have been stalled for hours.

sjka:
I gave it the needed kick. Let's see...

wborn:
I've checked several Nighlty builds and haven't found any artifacts without generated XML. :smiley:

PR #4633, opened 2017-11-26
Add WAITING step for firmware update

cdjackson:
RF devices may wait for a significant amount of time before the transfer actually begins so this adds the possibility to indicate this to the user.


tomhoefer:
lgtm

PR #4637, opened 2017-11-27
Allow null instead of empty list for removed discovery results

triller-telekom:

### DIFF
--- a/bundles/config/org.eclipse.smarthome.config.discovery/src/main/java/org/eclipse/smarthome/config/discovery/DiscoveryListener.java
+++ b/bundles/config/org.eclipse.smarthome.config.discovery/src/main/java/org/eclipse/smarthome/config/discovery/DiscoveryListener.java
@@ -98,6 +98,7 @@ public interface DiscoveryListener {
      *            if not {@code null} only results of that bridge are being removed
      * @return collection of thing UIDs of all removed things
      */
+    @Nullable
     Collection<ThingUID> removeOlderResults(DiscoveryService source, long timestamp,
             @Nullable Collection<ThingTypeUID> thingTypeUIDs, @Nullable ThingUID bridgeUID);
 
--- a/bundles/config/org.eclipse.smarthome.config.discovery/src/main/java/org/eclipse/smarthome/config/discovery/DiscoveryListener.java
+++ b/bundles/config/org.eclipse.smarthome.config.discovery/src/main/java/org/eclipse/smarthome/config/discovery/DiscoveryListener.java
@@ -75,6 +75,7 @@ public interface DiscoveryListener {
      * @deprecated use {@link #removeOlderResults(DiscoveryService, long, Collection, ThingUID)} instead
      */
     @Deprecated
+    @Nullable
     default Collection<ThingUID> removeOlderResults(DiscoveryService source, long timestamp,
             @Nullable Collection<ThingTypeUID> thingTypeUIDs) {
         return removeOlderResults(source, timestamp, thingTypeUIDs, null);
--- a/bundles/config/org.eclipse.smarthome.config.discovery/src/main/java/org/eclipse/smarthome/config/discovery/internal/DiscoveryServiceRegistryImpl.java
+++ b/bundles/config/org.eclipse.smarthome.config.discovery/src/main/java/org/eclipse/smarthome/config/discovery/internal/DiscoveryServiceRegistryImpl.java
@@ -348,7 +348,7 @@ public final class DiscoveryServiceRegistryImpl implements DiscoveryServiceRegis
     }
 
     @Override
-    public Collection<ThingUID> removeOlderResults(final DiscoveryService source, final long timestamp,
+    public @Nullable Collection<ThingUID> removeOlderResults(final DiscoveryService source, final long timestamp,
             final @Nullable Collection<ThingTypeUID> thingTypeUIDs, @Nullable ThingUID bridgeUID) {
         HashSet<ThingUID> removedResults = new HashSet<>();
         for (final DiscoveryListener listener : this.listeners) {
--- a/extensions/binding/org.eclipse.smarthome.binding.hue.test/src/test/java/org/eclipse/smarthome/binding/hue/internal/HueLightDiscoveryServiceOSGiTest.java
+++ b/extensions/binding/org.eclipse.smarthome.binding.hue.test/src/test/java/org/eclipse/smarthome/binding/hue/internal/HueLightDiscoveryServiceOSGiTest.java
@@ -19,7 +19,6 @@ import static org.junit.Assert.*;
 import java.io.IOException;
 import java.lang.reflect.Field;
 import java.util.Collection;
-import java.util.Collections;
 import java.util.concurrent.atomic.AtomicBoolean;
 
 import org.eclipse.smarthome.binding.hue.handler.HueBridgeHandler;
@@ -132,7 +131,7 @@ public class HueLightDiscoveryServiceOSGiTest extends AbstractHueOSGiTest {
             @Override
             public Collection<ThingUID> removeOlderResults(DiscoveryService source, long timestamp,
                     Collection<ThingTypeUID> thingTypeUIDs, ThingUID bridgeUID) {
-                return Collections.emptyList();
+                return null;
             }
         });
 
--- a/extensions/binding/org.eclipse.smarthome.binding.hue.test/src/test/java/org/eclipse/smarthome/binding/hue/internal/discovery/HueBridgeNupnpDiscoveryOSGITest.java
+++ b/extensions/binding/org.eclipse.smarthome.binding.hue.test/src/test/java/org/eclipse/smarthome/binding/hue/internal/discovery/HueBridgeNupnpDiscoveryOSGITest.java
@@ -18,7 +18,6 @@ import static org.junit.Assert.*;
 
 import java.io.IOException;
 import java.util.Collection;
-import java.util.Collections;
 import java.util.HashMap;
 import java.util.List;
 import java.util.Map;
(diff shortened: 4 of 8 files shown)


triller-telekom:
Based on the discussion in https://github.com/eclipse/smarthome/pull/4606#issuecomment-346897913

I would like to continue here with this proposal. Having a `default` method on the interface that returns an empty list could be an option. But this would hide the method for people using the IDE feature "Add unimplemented method" because `default` methods would not be generated and thus a developer might overlook this method and not implement it at all.

As a suggestion for our annotations I would say that we guarantee the users of our framework that the framework commits itself to the annotations, i.e. a user can rely on them that a value in NonNull or can be `null`. And for the other direction, i.e. everything that the framework is calling and what is implemented by its users we are "defensive" in the way that we do not trust the user and implement checks in case the user provided wrong data.

@maggu2810: WDYT?

maggu2810:
> For methods on interfaces that are often not implemented, it is convenient to simply keep the auto-generated (empty) code

> But this would hide the method for people using the IDE feature "Add unimplemented method"

It is really a pain that the IDE cannot create the correct binding code and a developer needs to read documentation and write code itself. :wink: 

> As a suggestion for our annotations I would say that we guarantee the users of our framework that the framework commits itself to the annotations, i.e. a user can rely on them that a value in NonNull or can be `null`. And for the other direction, i.e. everything that the framework is calling and what is implemented by its users we are "defensive" in the way that we do not trust the user and implement checks in case the user provided wrong data.

Isn't that similar to what I have written yesterday night?
We say to the binding developer, the framework would like to have a non null return value but the framework keeps defense and don't trust the user.
Same as for function arguments. The framework states that an argument must not be null but if it will crash later and not directly, we don't trust the user and check the argument regardless of the notification.

The method "remove older results" is not implemented by the framework but by the user. The framework states the user should not return null (as it does for function arguments of the framework that a user should not call with a null value). But the consumer of "remove older results" (the framework) will still be defense and handle the null return value.

sjka:
I don't really understand the discussion here. From what I can see, the framework *code* at this place does _not_ expect an empty collection, but is perfectly happy with a `null` value. The addition of the `@NonNullByDefault` annotation simply changed this, but I'd consider this a bug which this PR simply fixes. What did I miss?

maggu2810:
The repetitive discussion shows that we have different opinions about the whole nullness stuff.
If you feel more happy to keep null allowed on places that could happily life without null and add additional checks for null, ...
I use null annotations so I don't need to check for null references all the time. But I also trust developers that they could write code and not use auto-generated stuff.

I am completely fine to state that the return value here should not be null.
There is no semantic difference between null and an empty collection both states that nothing has been removed. Allowing null only involves that the handler of the return value needs to add an additional check. No problem to add this check, but not necessary.

My intention of the nullness annotations has been to remove null checks and state we don't want null at places where null doesn't bring any win.
But as long as we state "it has been this way before, so it must not be changed" or "the code generator adds `return null`" as arguments, I don't think it is a "discussion" at all.

Nothing critical, I assume I have a different point of view about this topic. That is a normal situation. For the moment I will cancel to complain about nullness stuff as it seems that you guys agree.

triller-telekom:
> Allowing null only involves that the handler of the return value needs to add an additional check. No problem to add this check, but not necessary.

So far the handler of the return value was perfectly fine with a returned `null` value, i.e. it expects `null` and implements a `null` check. 

Trusting the developers of the framework is a good thing, I agree! But trusting third party developers that implement our interface is IMHO a crucial thing when it comes to security. So internally within our framework code we can get rid of `null` checks, but I think we should keep them for places where we call external code. So how about this separation of internally without checks and externally with checks?

The problem with setting this return value to `@NonNull` is that external developers can simply ignore it and if you add an additional `null` check this check will be marked as "dead code".

I appreciate your opinion on the whole null annotation "discussion" since you are the one with the most experience in that field. So although it seems that "we guys" agree, this must not necessarily mean we are right :)

maggu2810:
> The problem with setting this return value to `@NonNull` is that external developers can simply ignore it and if you add an additional `null` check this check will be marked as "dead code".

That is only tooling.
If we don't trust external code we just need to ignore / disable such messages.
Isn't is the same as we have a function that adds an object to an internally collection and we would like to ensure that the added element is null?
Wouldn't we add a `@NonNull` annotation to our "add" method argument but still keep the check that the argument is really not null (throw an IllegalArgumentException) because it could crash some time later if someone accesses the collection members? Isn't it the same redundant code? Or should we annotate the argument as `@Nullable` just to remove the dead code warning?

> Trusting the developers of the framework is a good thing, I agree! But trusting third party developers that implement our interface is IMHO a crucial thing when it comes to security. So internally within our framework code we can get rid of `null` checks, but I think we should keep them for places where we call external code. So how about this separation of internally without checks and externally with checks?

Okay for me, too. But doesn't this fit to my comment yesterday?
We didn't remove a null check and still handle the situation a third party developer returns null.
We state that he shouldn't use null but an empty collection.

We should try to eliminate null on all places where possible and use as much "non-null" references as possible. Keeping null checks for untrusted third party code to have a stable framework should not be affected by that change.

triller-telekom:
> Or should we annotate the argument as @Nullable just to remove the dead code warning?

No we certainly should not do that. This PR was meant as a place for discussing this on a real life example.

> If we don't trust external code we just need to ignore / disable such messages.

So shall we set this to ignore? I think it would make sense. I haven't seen this warning (before null annotations) often anyway...

If we disable the warning I can close this PR and open one that changes the setting in the IDE (and maven).

image

triller-telekom:
I just found this comment from Stephan Herrmann:

https://bugs.eclipse.org/bugs/show_bug.cgi?id=459256#c1

> But to really show what you mean, simply write:

	public void method(@NonNull String param) {
	    assert param != null : "param cannot be null";
	}

How about using this instead of switching the dead code warning off? Of course this would mean that we have to change all null checks into asserts once we turn on null annotations for a class.

maggu2810:
> How about using this instead of switching the dead code warning off?

I don't understand this suggestion.
If we don't trust third party code we need to add a check regardless of the annotation.
If you use an assertion instead and assertion are disabled at runtime then you don't check the untrusted third party code anymore.

triller-telekom:
I forgot the fact that you can enable/disable assumptions at runtime...

Guess that leaves us only with the option to disable the dead code warning.

Do you think it would have a noticeable (negative) effect  if we would disable the warning?

maggu2810:
> Do you think it would have a noticeable (negative) effect if we would disable the warning?

Disable the complain about redundant checks could lead to "not the best performance" as there are e.g. null checks that does not need to be done. But as we don't want to trust 3rd party code, we needs that redundant checks.
If there are some other redundant checks I don't think it does matter a lot.

triller-telekom:
I am closing this PR in favor to #4650

kaikreuzer:
I am not really happy with the outcome here.
My paramount objective is to make life for extension developers easy - meaning having to care as little as possible about framework requirements and still write good, correct and safe code.

To reach this goal, it would be perfectly sensible to me to have the return value as `@Nullable` here - if the developer does not want to provide ANY feedback here, why force him to return an empty list instead of simply null? Why bother him with an compilation error in the tooling that he has to figure out what to do about it? Why remove a dead code warning, while it would help him identify bugs in his code?

> My intention of the nullness annotations has been to remove null checks and state we don't want null at places where null doesn't bring any win.

Right, but this doesn't apply here: We still have to do the null check anyhow and the win is that the developer can simply ignore methods that he does not want to implement.

triller-telekom:
@kaikreuzer What about the example from @maggu2810?

> Isn't is the same as we have a function that adds an object to an internally collection and we would like to ensure that the added element is null?
Wouldn't we add a @NonNull annotation to our "add" method argument but still keep the check that the argument is really not null (throw an IllegalArgumentException) because it could crash some time later if someone accesses the collection members? Isn't it the same redundant code? Or should we annotate the argument as @Nullable just to remove the dead code warning?

Would we also state @Nullable for the developer to have it "more convenient" for him?

I think that the framework should clearly define how its code should be used by developers and if we took the decision in the past that we do not like `null` for lists, but want empty lists, why not state that here?

sjka:
> if we took the decision in the past that we do not like null for lists, but want empty lists, why not state that here?

IIRC we have been talking about _returning_ empty collections - so API consumers don't need to bloat their code with null checks.

However, this particular case here is an interface which _consumers_ implement and therefore we are talking about values wich the framework _accepts_, i.e. incoming values. 

I absolutely agree that life would be much easier without `null` values everywhere. However, I think @kaikreuzer has a point in suggesting to be a little more defensive at the framework boarders - let aside the convenience for implementors.

triller-telekom:
Reopened as per https://github.com/eclipse/smarthome/pull/4650#issuecomment-351654333

kaikreuzer:
I actually expected that you also revert the changes you did [here and in the following lines](https://github.com/eclipse/smarthome/pull/4606/files#diff-3879289e932e5a84f684fadf011ece4aL126) as well - because they were due to the previous change of the API.

triller-telekom:
Since these classes are within our codebase and we said that we want to return empty collections instead of `null` wherever possible, does it hurt to have these lines in this PR?

@maggu2810 WDYT think?

kaikreuzer:
Our code is the reference on what we expect from consumers to implement. On consumer interfaces, we do not want to require anyone to bother about creating empty collections. Isn't that what we discussed when reopening this issue?

triller-telekom:
Yes I think we discussed that we do not enforce our consumers to return an empty collection and allow them to return `null`. But if we also "consume" our interfaces like you quoted here: https://github.com/eclipse/smarthome/pull/4606/files#diff-3879289e932e5a84f684fadf011ece4aL126 I do not see a problem if we return an empty collection there.

kaikreuzer:
Well, my main concern were the 5 occurrences underneath the line you have linked to.

PR #4650, opened 2017-11-29
Ignore dead code warning in IDE and maven

triller-telekom:

### DIFF
--- a/pom.xml
+++ b/pom.xml
@@ -202,7 +202,7 @@
             </extraClasspathElements>
             <compilerArgs>
               <!--<arg>-err:+nullAnnot(org.eclipse.jdt.annotation.Nullable|org.eclipse.jdt.annotation.NonNull|org.eclipse.jdt.annotation.NonNullByDefault),+inheritNullAnnot</arg>-->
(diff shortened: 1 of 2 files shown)


triller-telekom:
How often have you seen this warning in your IDE? To be honest I have only seen it a couple of times in all the years I am writing Java code with Eclipse...

However, for the framework ESH to act defensive against 3rd party solution code, we somehow need a way to enforce our null annotations. And if we skip null checks in cases where a 3rd party developer may return null because his IDE ignores our @NonNull, we are in trouble.

Please see our discussion in #4637 before I opened this PR.

maggu2810:
Some code that shows the warning because we try to harden the framework...

```Java
public class FrameworkProvidedService {
    public void foo(@NonNull Object ele) {
        // The argument "ele" is not dereferenced in this function but safed for use at another time.
        // Because we do not trust third party code that it fullfills the nullable requirements, we add an additional
        // test here to harden the framework.
        if (ele == null) { // Mutliple markers at this line: 1. Dead code, 2. Redundant null check: comparing '@NonNull
                           // Object' against null.
            throw new IllegalArgumentException("The argument must not be null");
        }
        // ...
    }
}
```

triller-telekom:
Thinking about this again it might be not be such a good idea to disable that warning.

We agreed that we have to distinguish two types of interfaces, those which are implemented by others and those that we implement ourselves. The later ones can be `@NonNull` because we are the implementers and follow our rules. For the former we want to be defensive and add a `null` check. Thus it makes sense to also allow the `@Nullable` in these cases.

I am finally closing this PR and reopen #4637 to be merged as our final decision.

PR #4653, opened 2017-11-29
Update LSP started string

ThomDietrich:
...following the wording of dashboards and including the buzzword LSP. I know it's pretty minor :)

Signed-off-by: Thomas Dietrich <thomas.dietrich@tu-ilmenau.de>


sjka:
...but it makes sense. Thanks!

(FTR: I manually checked the ECA _is_ valid)

Issue #982, opened 2016-02-04
Could not find an Event Factory for the event type 'ItemThingLinkRemovedEvent'


lolodomo:
Here is a warning message I get when I logout from OH 2 beta 2:

```
2016-02-04 23:44:38.011 [INFO ] [ternal.scheduler.WeatherJobScheduler] - Deleting weatherJob-home
2016-02-04 23:44:38.614 [INFO ] [ding.mios.internal.MiosUnitConnector] - run: Shutting down MiOS Polling thread.  Unit=house
2016-02-04 23:44:39.348 [INFO ] [ding.astro.handler.AstroThingHandler] - Deleting astro DailyJob for thing 'astro:sun:088b495d'
2016-02-04 23:44:39.355 [INFO ] [ding.astro.handler.AstroThingHandler] - Scheduled astro DailyJob at midnight for thing astro:sun:088b495d
2016-02-04 23:44:45.345 [INFO ] [me.event.ThingStatusInfoChangedEvent] - 'ntp:ntp:local' changed from ONLINE to UNINITIALIZED (HANDLER_MISSING_ERROR)
2016-02-04 23:44:45.434 [INFO ] [ding.astro.handler.AstroThingHandler] - Deleting astro DailyJob for thing 'astro:sun:088b495d'
2016-02-04 23:44:45.441 [INFO ] [me.event.ThingStatusInfoChangedEvent] - 'astro:sun:088b495d' changed from ONLINE to UNINITIALIZED (HANDLER_MISSING_ERROR)
2016-02-04 23:44:45.520 [WARN ] [ore.internal.events.OSGiEventManager] - Could not find an Event Factory for the event type 'ItemThingLinkRemovedEvent'.
2016-02-04 23:44:45.521 [WARN ] [ore.internal.events.OSGiEventManager] - Could not find an Event Factory for the event type 'ItemThingLinkRemovedEvent'.
```

I don't know if this is really important or not.

kaikreuzer:
> I don't know if this is really important or not.

It does not impact any functionality (as it is only happening during shutdown), but it is obviously not nice to have such things in the log.
I think we should in general check, if we can have a mechanism to determine whether we are during a startup or shutdown phase and suppress events in these situations.

mohsh86:
I have a similar issue, but for some reason, the ntp:ntp:local thing won' go back to initialized, and stuck at "uninitialized"

[...]

PR #4655, opened 2017-11-30
SafeCaller as a replacement for the SafeMethodCaller

### LINKED ISSUE #4591: OSGiEventManager should do safe method calls and identify handlers that are blocking

sjka:
This is an attempt to enhance the SafeMethodCaller by asynchronous execution and allowing to limit certain calls to a single thread in parallel. 

* As with the added options the SafeMethodCaller API would have exploded, it comes with a nice builder-pattern to configure how the call should be protected and executed
* It returns a dynamic proxy which allows not only to execute `Runnable`s but rather provides "safe" access to a given object's methods
* It can run calls asynchronously (i.e. return immediately)
* It comes as an OSGi services (as it internally has to maintain some state and static fields don't match the OSGi lifecycle expectations)
* It supports an "identifier" which determines a scope which can be (optionally) limited to a single thread in parallel (the target object itself is the identifier by default)

enhances #4602 (and thereby replacing it)
fixes #4591


bundles/core/org.eclipse.smarthome.core.test/src/test/java/org/eclipse/smarthome/core/internal/common/SafeCallerImplTest.java
```diff
@@ -0,0 +1,532 @@
...
+            });
+        });
+    }
+
+    @Test
+    public void testSingleThread_sync_parallel() throws Exception {
+        Runnable mock = mock(Runnable.class);
+        doAnswer(a -> sleep(500)).when(mock).run();
+
+        safeCaller.modified(new HashMap<String, Object>() {
```
htreu: could you also use `configureSingleThread();` here?

bundles/core/org.eclipse.smarthome.core.test/src/test/java/org/eclipse/smarthome/core/internal/common/SafeCallerImplTest.java
```diff
@@ -0,0 +1,532 @@
...
+            verify(mock, times(2)).run();
+        });
+    }
+
+    @Test
+    public void testSingleThread_async() throws Exception {
+        Runnable mock = mock(Runnable.class);
+        doAnswer(a -> sleep(500)).when(mock).run();
+
+        safeCaller.modified(new HashMap<String, Object>() {
```
htreu: also here: `configureSingleThread();`

bundles/core/org.eclipse.smarthome.core/src/main/java/org/eclipse/smarthome/core/internal/common/DuplicateExecutionException.java
```diff
@@ -0,0 +1,34 @@
...
+ * This program and the accompanying materials are made available under the
+ * terms of the Eclipse Public License 2.0 which is available at
+ * http://www.eclipse.org/legal/epl-2.0
+ *
+ * SPDX-License-Identifier: EPL-2.0
+ */
+package org.eclipse.smarthome.core.internal.common;
+
+/**
+ * Denotes that under there already is a thread occupied by the same context.
```
htreu: `under` may be removed.

bundles/core/org.eclipse.smarthome.core/src/main/java/org/eclipse/smarthome/core/internal/common/SafeCallManager.java
```diff
@@ -0,0 +1,64 @@
...
+/**
+ *
+ * @author Simon Kaufmann - initial contribution and API.
+ *
+ */
+@NonNullByDefault
+public interface SafeCallManager {
+
+    /**
+     * Track that a the call to the target method starts.
```
htreu: `a` is obsolete.

bundles/core/org.eclipse.smarthome.core/src/main/java/org/eclipse/smarthome/core/internal/common/SafeCallManagerImpl.java
```diff
@@ -0,0 +1,165 @@
...
+        this.threadPoolName = threadPoolName;
+        this.enforceSingleThreadPerIdentifier = enforceSingleThreadPerIdentifier;
+    }
+
+    @Override
+    public void recordCallStart(Invocation invocation, TrackingCallable wrapper) {
+        synchronized (activeIdentifiers) {
+            TrackingCallable otherWrapper = activeIdentifiers.get(invocation.getIdentifier());
+            if (enforceSingleThreadPerIdentifier && otherWrapper != null) {
+                enqueue(invocation);
```
htreu: This may have a small comment on why the invocation is queued nonetheless.

bundles/core/org.eclipse.smarthome.core/src/main/java/org/eclipse/smarthome/core/internal/common/TrackingCallable.java
```diff
@@ -0,0 +1,55 @@
...
+ */
+package org.eclipse.smarthome.core.internal.common;
+
+import java.lang.reflect.Method;
+import java.util.concurrent.Callable;
+
+import org.eclipse.jdt.annotation.Nullable;
+
+/**
+ * Wraps a {@link Callable} and tracks the start timestamp as well as the executing thread.
```
htreu: `timestamp` is not tracked anymore ;-)

bundles/core/org.eclipse.smarthome.core/src/main/java/org/eclipse/smarthome/core/internal/common/TrackingCallable.java
```diff
@@ -0,0 +1,55 @@
...
+
+import org.eclipse.jdt.annotation.Nullable;
+
+/**
+ * Wraps a {@link Callable} and tracks the executing thread.
+ *
+ * @author Simon Kaufmann - initial contribution and API.
+ *
+ */
+class TrackingCallable implements Callable<Object> {
```
maggu2810: Any reason not to mark this class as `@NonNullByDefault`?

bundles/core/org.eclipse.smarthome.core/src/main/java/org/eclipse/smarthome/core/internal/common/SafeCallManagerImpl.java
```diff
@@ -0,0 +1,168 @@
...
+                getScheduler().submit(wrapper);
+                logger.trace("Submitted {} for asynchronous execution", next);
+            }
+        }
+    }
+
+    private void handlePotentialTimeout(Invocation invocation) {
+        Object identifier = invocation.getIdentifier();
+        Invocation currentInvocation = activeAsyncInvocations.get(identifier);
+        if (currentInvocation != null && currentInvocation == invocation) {
```
maggu2810: `invocation` is not null as it has been already dereferenced above.
So, if you test `currentInvocation == invocation`, isn't the test `currentInvocation != null` redundant?

bundles/core/org.eclipse.smarthome.core/src/main/java/org/eclipse/smarthome/core/internal/common/SafeCallManager.java
```diff
@@ -0,0 +1,64 @@
...
+
+import org.eclipse.jdt.annotation.NonNullByDefault;
+
+/**
+ *
+ * @author Simon Kaufmann - initial contribution and API.
+ *
+ */
+@NonNullByDefault
+public interface SafeCallManager {
```
maggu2810: Do you plan to add the missing JavaDoc for param and return after the first reviews?

bundles/core/org.eclipse.smarthome.core/src/main/java/org/eclipse/smarthome/core/internal/common/SafeCallManager.java
```diff
@@ -0,0 +1,64 @@
...
+ * SPDX-License-Identifier: EPL-2.0
+ */
+package org.eclipse.smarthome.core.internal.common;
+
+import java.util.concurrent.ExecutorService;
+
+import org.eclipse.jdt.annotation.NonNullByDefault;
+
+/**
+ *
```
maggu2810: Mostly you add a short description about the class / interface. Do you consider to add it here, too?

bundles/core/org.eclipse.smarthome.core/src/main/java/org/eclipse/smarthome/core/internal/common/InvocationHandlerSync.java
```diff
@@ -0,0 +1,63 @@
...
+    public InvocationHandlerSync(SafeCallManager manager, T target, Object identifier, long timeout,
+            @Nullable Consumer<Throwable> exceptionHandler, @Nullable Runnable timeoutHandler) {
+        super(manager, target, identifier, timeout, exceptionHandler, timeoutHandler);
+    }
+
+    @Override
+    public Object invoke(Object proxy, Method method, Object[] args) throws Throwable {
+        TrackingCallable wrapper = new TrackingCallable(new Invocation(this, method, args));
+        if (getManager().isSafeContext()) {
+            logger.debug(MSG_CONTEXT, toString(method), getTarget());
```
maggu2810: Does it make sense to test if debugging is enabled here, so the methods `toString(method)` and `getTarget()` are only called if necessary?

bundles/core/org.eclipse.smarthome.core/src/main/java/org/eclipse/smarthome/core/internal/common/InvocationHandlerSync.java
```diff
@@ -0,0 +1,63 @@
...
+import org.slf4j.LoggerFactory;
+
+/**
+ * Synchronous invocation handler implementation.
+ *
+ * @author Simon Kaufmann - initial contribution and API.
+ *
+ * @param <T>
+ */
+public class InvocationHandlerSync<T> extends AbstractInvocationHandler<T> implements InvocationHandler {
```
maggu2810: `@NonNullByDefault`?

bundles/core/org.eclipse.smarthome.core/src/main/java/org/eclipse/smarthome/core/internal/common/InvocationHandlerAsync.java
```diff
@@ -0,0 +1,48 @@
...
+ * Asynchronuous {@link InvocationHandler} implementation.
+ *
+ * Instead of directly invoking the called method it rather queues it in the {@link SafeCallManager} for asynchronous
+ * execution.
+ *
+ * @author Simon Kaufmann - initial contribution and API.
+ *
+ * @param <T>
+ */
+class InvocationHandlerAsync<T> extends AbstractInvocationHandler<T> implements InvocationHandler {
```
maggu2810: `@NonNullByDefault`?

bundles/core/org.eclipse.smarthome.core/src/main/java/org/eclipse/smarthome/core/internal/common/AbstractInvocationHandler.java
```diff
@@ -0,0 +1,167 @@
...
+    private final Object identifier;
+    private final long timeout;
+    @Nullable
+    private final Consumer<Throwable> exceptionHandler;
+    @Nullable
+    private final Runnable timeoutHandler;
+
+    AbstractInvocationHandler(SafeCallManager manager, T target, Object identifier, long timeout,
+            @Nullable Consumer<Throwable> exceptionHandler, @Nullable Runnable timeoutHandler) {
+        super();
```
maggu2810: Okay, but already done implicit.

bundles/core/org.eclipse.smarthome.core/src/main/java/org/eclipse/smarthome/core/common/SafeCallerBuilder.java
```diff
@@ -0,0 +1,74 @@
...
+ * Builder to create a safe-call wrapper for another object.
+ *
+ * @author Simon Kaufmann - initial contribution and API.
+ *
+ * @param <T>
+ */
+@NonNullByDefault
+public interface SafeCallerBuilder<T> {
+
+    public T build();
```
maggu2810: Interface methods are `public` by definition, so no need to add the keyword here.
Same for the other methods of this interface.

maggu2810: One method of this interface miss the JavaDoc... 😉

bundles/core/org.eclipse.smarthome.core/src/main/java/org/eclipse/smarthome/core/common/SafeCaller.java
```diff
@@ -0,0 +1,37 @@
...
+ *
+ * @author Simon Kaufmann - initial contribution and API.
+ */
+@NonNullByDefault
+public interface SafeCaller {
+
+    /**
+     * Default timeout for actions in milliseconds.
+     */
+    public final int DEFAULT_TIMEOUT = 5000 /* milliseconds */;
```
maggu2810: Members / member function of interfaces are public default.

bundles/core/org.eclipse.smarthome.core/src/main/java/org/eclipse/smarthome/core/common/SafeCaller.java
```diff
@@ -0,0 +1,37 @@
...
+
+    public <T> SafeCallerBuilder<T> create(T target, Class<T> interfaceType);
```
maggu2810: Please add JavaDoc to the two `create` methods.

bundles/core/org.eclipse.smarthome.core/src/main/java/org/eclipse/smarthome/core/internal/common/TrackingCallable.java
sjka: Yes, with `@NonNullByDefault` and `@Override @Nullable public Object call() {` I get the following error:

```
The return type is incompatible with '@NonNull Object' returned from Callable<Object>.call() (mismatching null constraints)```

bundles/core/org.eclipse.smarthome.core/src/main/java/org/eclipse/smarthome/core/internal/common/InvocationHandlerAsync.java
sjka: Due to the yet missing external annotations this leads to some awkward pseudo-inherited annotations and a (in fact) redundant null-check on `method`. I have added it now as you were asking for it, but actually don't think this improved anything really.

bundles/core/org.eclipse.smarthome.core/src/main/java/org/eclipse/smarthome/core/internal/common/InvocationHandlerSync.java
sjka: dito

lolodomo:
@sjka: do you think your PR could have fixed the issue #982 ? After stopping OH with snapshot 1104, I did not find these errors in my logs.

PR #4658, opened 2017-11-30
use http for target platform resolution consistently

sjka:


maggu2810:
What has been the reason to switch from https to http?
Is there something that does not work if https is used?
Can you give us some information?

sjka:
> What has been the reason to switch from https to http?

Not a very compelling one, I'm afraid. And I somehow expected that question...

> Is there something that does not work if https is used?

Yes, apparently at least one one of our commercial adopters whose name I won't dare to disclose publicly did not yet manage to configure the proxy settings on their highly secured CI servers properly. And while they make plans to drum up a task-force to think about doing so I simply thought it can't hurt to be at least consistent on ESH side in the meantime. None of the other sites is configured using https right now - although of course they actually should.

Issue #3605, opened 2017-06-07
Discovery should be able to activate per bridge, not per binding


cdjackson:
If a binding has multiple bridges, it should be possible to activate discovery for each bridge separately - not just for the binding. At the moment it's not possible to select which bridge is activated for discovery, and therefore which bridge a new device will attach to, and this is a problem where this needs to be controlled.

kaikreuzer:
This is only relevant for bindings, which have an active pairing mechanism for having new devices join the network, right? 
We still do not have any clean support for such cases in general - e.g. it isn't possible to easily remove undesired devices from the network again, once they have joined and are listed in the inbox.

Is it a realistic use case for end users to have multiple such bridges? Or is it merely something helpful for development purposes?

cdjackson:
Yes, I guess it’s just for bindings with active join mechanisms… And yes, it’s a realistic case - we have this problem now on ZWave as people are using multiple dongles and they can’t control which dongle a device will connect to. Another use case I have is BLE devices have limited connections available - maybe as few as 3 simultaneous connections. If people want / need multiple connections, it may require multiple dongles and in order to avoid them all just joining the first bridge they find, we’d need some control over this.

Re leaving the network - I have a mechanism in ZWave, but it’s not especially nice. It uses a config parameter as an action to put the stick into exclusion mode for 30 seconds. A more standard mechanism would be nice, but that’s another issue...

rodneyrehm:
> We still do not have any clean support for such cases in general

Is that maybe because you're "abusing" the discovery/inbox - which was intended for *announcing* (IP) devices available for pairing - for actually *pairing* devices?

> Is it a realistic use case for end users to have multiple such bridges? 

I'm not sure about current HMIP support, but Homematic used to have network extenders to support larger homes. Our suolution hasn't supported this in the past, but it'd be great if we could support this in the future.

> It uses a config parameter as an action to put the stick into exclusion mode for 30 seconds. A more standard mechanism would be nice, but that’s another issue...

What happens if the device is put into the inbox (i.e. connected with the controller) and ESH goes down before the DiscoveryResult's TTL is reached (subsequently triggering the exclusion)? unless DiscoveryResults are persisted, this would lead to ESH not knowing the Thing yet, but the device thinking it's connected to the controller - no?

PR #4663, opened 2017-12-01
Added getBridge() to DiscoveryService 

afuechsel:
Added getBridge() to DiscoveryService to indicate the bridge UID that actually created this discovery service.

Before this PR one could approve a hue bridge from the inbox which in turn added all hue lights of this bridge to the inbox. When pairing another hue bridge and removing it later all hue lights have been removed from the inbox, regardless of whether they are from the first or the second bridge. With this PR the results from the first and still existing bridge are no longer removed if another bridge is removed. 


bundles/config/org.eclipse.smarthome.config.discovery.test/src/test/groovy/org/eclipse/smarthome/config/setup/test/discovery/DiscoveryServiceMockOfBridge.groovy
```diff
@@ -0,0 +1,22 @@
+package org.eclipse.smarthome.config.setup.test.discovery
```
sjka: Oh noooo, pleeease, no NEW groovy tests!

bundles/config/org.eclipse.smarthome.config.discovery.test/src/test/groovy/org/eclipse/smarthome/config/setup/test/discovery/DiscoveryServiceRegistryOSGITest.groovy
```diff
@@ -323,6 +339,55 @@ class DiscoveryServiceRegistryOSGITest extends OSGiTest {
         anotherDiscoveryServiceMockForBinding1.abortScan()
     }
 
+    @Test
+    void 'assert that removeOlderResults removes only the entries of the specified bridge' () {
```
sjka: Groovy. See above.

bundles/config/org.eclipse.smarthome.config.discovery/src/main/java/org/eclipse/smarthome/config/discovery/DiscoveryService.java
```diff
@@ -50,22 +51,32 @@
...
-    public Collection<ThingTypeUID> getSupportedThingTypes();
+    Collection<ThingTypeUID> getSupportedThingTypes();
+
+    /**
+     * If the discovery service discovers things for a specific bridge only, this method
+     * would return the UID of the bridge.
+     *
+     * @return UID of the bridge that has started this discovery service, may be {@null}
+     */
+    default @Nullable ThingUID getBridge() {
```
sjka: What if a discovery service _can_ discover things for several bridge instances - and you still would like to get rid of results for a single bridge?

bundles/config/org.eclipse.smarthome.config.discovery.test/src/test/groovy/org/eclipse/smarthome/config/setup/test/discovery/DiscoveryServiceMockOfBridge.groovy
afuechsel: It's just a copy of an existing test modified to check the new behavior. As the whole infrastructure for the DiscoveryService tests is in Groovy I see no benefit in duplicating it to Java.

bundles/config/org.eclipse.smarthome.config.discovery/src/main/java/org/eclipse/smarthome/config/discovery/DiscoveryService.java
afuechsel: Isn't it a problem that a discovery service is able to discover things for more than one bridge instance? I mean, having a bridge that creates a discovery service, I only want results of this service for this bridge

sjka:
Why don't you simply extends the `DiscoveryListener` interface with a method signature that also accepts a bridge's UID?

bundles/config/org.eclipse.smarthome.config.discovery/src/main/java/org/eclipse/smarthome/config/discovery/DiscoveryService.java
sjka: https://github.com/openhab/openhab2-addons/blob/e2befee912df9c1f65f9423a166d96a62ac8e7cc/addons/binding/org.openhab.binding.folding/src/main/java/org/openhab/binding/folding/internal/discovery/FoldingSlotDiscoveryService.java

sjka: https://github.com/openhab/openhab2-addons/blob/739a8bb2126c22d36830b21908b051ffe3399d1e/addons/binding/org.openhab.binding.yamahareceiver/src/main/java/org/openhab/binding/yamahareceiver/internal/discovery/ZoneDiscoveryService.java

cdjackson:
> Isn't it a problem that a discovery service is able to discover things for more than one bridge instance? I mean, having a bridge that creates a discovery service, I only want results of this service for this bridge

Maybe not directly linked, but since you talk about multiple bridges and only getting discovery results from one service, I'll just link this other issue...

Discovery should be able to activate per bridge, not per binding #3605

triller-telekom:
LGTM, except for the new groovy tests :(

afuechsel:
@triller-telekom well, as I said above I just copied an existing test and tailored it to suit my needs. Would it really be a good idea, to re-create the whole test infrastructure of these OSGi tests just to add another, nearly similar test? Or should I convert everything into Java? That's far beyond the scope of this PR. I know, that ESH is going away from Groovy back to Java although this makes things more difficult IMHO and I would create plain new tests in Java as required, but not when I can reuse existing test infrastructure.

sjka:
FTR: I created #4704 where I ported the complete test to Java for you.

kaikreuzer:
FTR: #4704 has been merged and this PR here needs to be rebased and adapted.

afuechsel:
Rebased, ported my own test to Java

kaikreuzer:
I'll restart Travis.

kaikreuzer:
This time it's the DMX tests :-(

afuechsel:
@kaikreuzer how shall we proceed here? I am unable to reproduce these test failures.

sjka:
PS: The IP validation seems to fail because of the merge commit which will get squashed away anyway.

PR #4673, opened 2017-12-03
Sitemap REST API: signal long polling timeout

### LINKED ISSUE #4668: Sitemap REST API: 30 seconds timeout
I discovered that the sitemap REST API has a feature to block the request until a change in the page occurs but there is also a timeout of 30 seconds (in the method `waitForChanges`). When the timeout is reached, the page data is returned even without any data change.

https://github.com/eclipse/smarthome/blob/master/bundles/io/org.eclipse.smarthome.io.rest.sitemap/src/main/java/org/eclipse/smarthome/io/rest/sitemap/internal/SitemapResource.java#L217

This is a problem in the Android app for example because the page is reloaded every 30 seconds even when nothing is changed: https://github.com/openhab/openhab.android/issues/460

I see 2 ways to fix that:

1. disable the timeout in the REST API
2. keep the timeout but add a flag in the response to notify the caller when the timeout was reached

WDYT ?

lolodomo:
Fixes #4668

### DIFF
--- a/bundles/io/org.eclipse.smarthome.io.rest.sitemap/src/main/java/org/eclipse/smarthome/io/rest/sitemap/internal/PageDTO.java
+++ b/bundles/io/org.eclipse.smarthome.io.rest.sitemap/src/main/java/org/eclipse/smarthome/io/rest/sitemap/internal/PageDTO.java
@@ -17,7 +17,7 @@ import java.util.List;
 
 /**
  * This is a data transfer object that is used to serialize page content.
- * 
+ *
  * @author Kai Kreuzer - Initial contribution and API
  *
  */
@@ -30,6 +30,7 @@ public class PageDTO {
     public String link;
     public PageDTO parent;
     public boolean leaf;
+    public boolean timeout;
 
     public List<WidgetDTO> widgets = new ArrayList<WidgetDTO>();
 
--- a/bundles/io/org.eclipse.smarthome.io.rest.sitemap/src/main/java/org/eclipse/smarthome/io/rest/sitemap/internal/SitemapResource.java
+++ b/bundles/io/org.eclipse.smarthome.io.rest.sitemap/src/main/java/org/eclipse/smarthome/io/rest/sitemap/internal/SitemapResource.java
@@ -210,13 +210,16 @@ public class SitemapResource implements RESTResource, SitemapSubscriptionCallbac
             }
         }
 
+        boolean timeout = false;
         if (headers.getRequestHeader("X-Atmosphere-Transport") != null) {
             // Make the REST-API pseudo-compatible with openHAB 1.x
             // The client asks Atmosphere for server push functionality,
             // so we do a simply listening for changes on the appropriate items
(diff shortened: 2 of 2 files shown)


lolodomo:
Note that this enhancement does not break anything as it only adds a new field that will be ignore by unmodified callers.

lolodomo:
@kaikreuzer : can you please merge this PR so that all users of the upcoming 2.2 release could benefit of a better Android app.

bundles/io/org.eclipse.smarthome.io.rest.sitemap/src/main/java/org/eclipse/smarthome/io/rest/sitemap/internal/SitemapResource.java
```diff
@@ -210,13 +210,17 @@ public Response getPageData(@Context HttpHeaders headers,
...
             // The client asks Atmosphere for server push functionality,
             // so we do a simply listening for changes on the appropriate items
-            blockUnlessChangeOccurs(sitemapname, pageId);
+            // The blocking has a timeout of 30 seconds. If this timeout is reached,
+            // we notice this information in the response object.
+            timeout = blockUnlessChangeOccurs(sitemapname, pageId);
         }
-        Object responseObject = getPageBean(sitemapname, pageId, uriInfo.getBaseUriBuilder().build(), locale);
+        PageDTO responseObject = getPageBean(sitemapname, pageId, uriInfo.getBaseUriBuilder().build(), locale);
+        responseObject.timeout = timeout;
```
kaikreuzer: Imho it would be cleaner to add `timeout` as a new parameter to `getPageBean` - this should be safe as it is a only a private method.

bundles/io/org.eclipse.smarthome.io.rest.sitemap/src/main/java/org/eclipse/smarthome/io/rest/sitemap/internal/PageDTO.java
```diff
@@ -33,6 +33,8 @@
 
     public List<WidgetDTO> widgets = new ArrayList<WidgetDTO>();
 
+    public boolean timeout;
```
kaikreuzer: Could you move this up to the same block as the other fields?

kaikreuzer:
> could benefit of a better Android app.

So time for you to create an according PR on the Android app!

bundles/io/org.eclipse.smarthome.io.rest.sitemap/src/main/java/org/eclipse/smarthome/io/rest/sitemap/internal/SitemapResource.java
lolodomo: Ok I will change that.

lolodomo:
Now that I see your real intention to merge it , yes I can start working on the Android app fix.

lolodomo:
Changes done.

Issue #4669, opened 2017-12-03
Default System Profile behaviour


kgoderis:
With reference to https://github.com/eclipse/smarthome/pull/4516, I wonder if it would not more logical to revert back the change with respect to handleUpdate(). That is, the Profiles introduced up to now do actually enable a behaviour. Therefore, it would be more logical to have "no behaviour" by default, and enable the current default behaviour with an explicitly defined profile. 

This would also be more logical from a user point of view, whereby a new wanted behaviour has to be added explicitly to the Channel definition.

maggu2810:
Hi @kgoderis, I am not sure if I understand your suggestion?
Isn't the current default profile that handle command and handle update are delegated to the handler?
What is your suggestion a default profile should look like?

kaikreuzer:
@kgoderis Isn't this a duplicate of the discussion that is supposed to be done in https://github.com/eclipse/smarthome/issues/4525?

kgoderis:
@kaikreuzer partly, in the sense that we discuss the handleUpdate() function itself in #4525

@maggu2810 Before #4516 there was no flow of Command and States between Channels connected to the same Item, e.g. default behaviour was "do nothing". All the profiles introduced up to now "add" a behaviour, or enable a behaviour. They are "additive" to the link.  With #4516 the default is behaviour is "do something", with all the defined profiles interpreted as "replace default behaviour by another behaviour". I find it more logical to have no default behaviour, and all the Profiles to be "additive". Just thinking forward to the day that there will be a use case to have more than one profile on an ItemChannelLink..... 

The changed default behaviour actually breaks things very badly in my installation, and more specifically with the KNX binding, which in itself is under discussion / not progressing at the same speed, but which is a primary future user of Profiles just in order to prevent loops of Telegrams on the KNX bus. I would hope that this change of default behaviour can be reverted at least until there is a clear view how things work out for the KNX binding. KNX is maybe not a primary concern for ESH, but it is the core technology in a few installations, notably of contributors like @kaikreuzer ;-)

kaikreuzer:
> With #4516 the default is behaviour is "do something"

No, this is not correct. The default profile is completely agnostic and simply passes commands and state updates between the framework and the binding as it used to do before.

Note that it specifically states "Restore backward-compatibility by making the default master profile forward handleUpdate() calls to ThingHandlers again". So it didn't introduce any changed behavior, it simply fixed a regression that was introduced shortly before.

[...]

Issue #4525, opened 2017-11-09
The future of handleUpdate(...)


sjka:
As mentioned [here](https://github.com/eclipse/smarthome/issues/4464#issuecomment-343155841), this is to discuss the future of the `ThingHandler.handleUpdate(...)` method. 

As I mentioned in PR #4108, I didn't really see a use-case for it nor an (open-source!) binding that actually made proper use of it at that time hence deprecating it. If we have a weird method that doesn't really make sense for any binding, then we indeed should remove it before a potential "1.0" release. On the other hand if we see it has good use, then of course it must stay!

The major use-case I identified was to "follow" another device by linking one channel to the same item together with another channel. However, then the thingHandler would need to know that it _is_ in this "following" mode and not take the lead anymore, i.e. should only react on updates but not on commands anymore. This means a thing handler implementation would need to be _either_ in default mode _or_ in follow mode but cannot be used in both scenarios. This use-case is covered by the "follow" profile now, allowing pretty much _every_ binding to be turned into this position without the need for it to even know that.

Nevertheless, there certainly might be other use-cases which I did not take into consideration. Thanks @philomatic for pointing that out! Therefore this issue here to collect and discuss them.

kungfoolfighting:
Ok, I'll give this a try:
I am probably way out of my depth so excuse me I'd the following is nonsense.
As far as I know the handleUpdate function is was gets called whenever a postUpdate is called on an item. I also assume that the reason we might get rid of the handleUpdate function is, that to the binding it usually makes no difference if an item got a command or an update. It only needs to transmit die payload to a physical device.
So the question is: could there theoretically be a binding that does care about the distinction? 

Under these conditions I could imagine a binding that links different home automation systems together so that an item in a remote system is represented by a local item. Sending a command to the item could then potentially trigger a rule on the remote system that only triggers on received commands, where as posting an update might not trigger that rule.
If we assume that the remote system might have the same concepts of things, commands and updates as OH or might even be OH then it would seems necessary to preserve an update as an update and not transform it into a command.

Most other scenarios I have come up with can be solved by using two channels and items and using one for updates and one for commands. This does have the downside of being a bit more complicated but the upside of allowing for more expressivity in the purpose of the items and channels.

kgoderis:
@sjka Shall I make a first throw at this?

[...]

PR #4674, opened 2017-12-03
Enhancement of Profiles with two new profile types

kgoderis:
“PassiveProfile" is a profile with only communication from the ThingHandler to the Item enabled, and “CommandProfile” is a profile whereby only Commands can flow between ThingHandlers and Items (but States are filtered out)

Signed-Off-By: Karel Goderis <karel.goderis@me.com>

### DIFF
--- a/bundles/core/org.eclipse.smarthome.core.thing/src/main/java/org/eclipse/smarthome/core/thing/internal/profiles/SystemCommandProfile.java
+++ b/bundles/core/org.eclipse.smarthome.core.thing/src/main/java/org/eclipse/smarthome/core/thing/internal/profiles/SystemCommandProfile.java
@@ -1,0 +1,73 @@
+/**
+ * Copyright (c) 2014,2017 Contributors to the Eclipse Foundation
+ *
+ * See the NOTICE file(s) distributed with this work for additional
+ * information regarding copyright ownership.
+ *
+ * This program and the accompanying materials are made available under the
+ * terms of the Eclipse Public License 2.0 which is available at
+ * http://www.eclipse.org/legal/epl-2.0
+ *
+ * SPDX-License-Identifier: EPL-2.0
+ */
+package org.eclipse.smarthome.core.thing.internal.profiles;
+
+import org.eclipse.jdt.annotation.NonNullByDefault;
+import org.eclipse.smarthome.core.thing.binding.ThingHandler;
+import org.eclipse.smarthome.core.thing.profiles.ProfileCallback;
+import org.eclipse.smarthome.core.thing.profiles.ProfileTypeUID;
+import org.eclipse.smarthome.core.thing.profiles.StateProfile;
+import org.eclipse.smarthome.core.thing.profiles.SystemProfiles;
+import org.eclipse.smarthome.core.types.Command;
+import org.eclipse.smarthome.core.types.State;
+
+/**
+ * This is the default implementation for a follow profile.
+ *
+ * In contrast to the {@link SystemDefaultProfile} it does not forward any state updates from the framework to the
+ * ThingHandler. Instead, it only accepts {@link Command}s and then forwards those to the {@link ThingHandler}.
+ * <p>
+ * This allows devices to be operated in an "execution-only" mode, whereby they only take commands that are explicitly
+ * given by the framework and ignore state updates that might come from the Item that is for example linked to other
+ * Channels that are used to feed back a State
+ * <p>
+ * The ThingHandler may send commands to the framework, but no state updates are forwarded.
+ *
+ * @author Karel Goderis - initial contribution and API.
+ *
+ */
+@NonNullByDefault
+public class SystemCommandProfile implements StateProfile {
+
+    private final ProfileCallback callback;
+
+    public SystemCommandProfile(ProfileCallback callback) {
+        this.callback = callback;
+    }
+
+    @Override
+    public ProfileTypeUID getProfileTypeUID() {
+        return SystemProfiles.COMMAND;
+    }
+
+    @Override
+    public void onStateUpdateFromItem(State state) {
+        // no-op
+    }
+
+    @Override
+    public void onCommandFromHandler(Command command) {
+        callback.sendCommand(command);
+    }
+
+    @Override
+    public void onCommandFromItem(Command command) {
+        callback.handleCommand(command);
+    }
+
+    @Override
+    public void onStateUpdateFromHandler(State state) {
+        // no-op
+    }
+
+}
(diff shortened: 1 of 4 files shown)


kgoderis:
@sjka huh? what happened with the Eclipse Validation? my Contributor Agreement is not expiring before 2020.

maggu2810:
For me the profile itself are okay. If there is need for, why not adding it.
I am not sure about the chossen name, let's see about a comment of @sjka as I assume he knows most about the whole profile stuff.

maggu2810:
> @sjka @kaikreuzer huh? what happened with the Eclipse Validation? my Contributor Agreement is not expiring before 2020.

A manual check state that there is a valid ECA for your address. So all fine.

kgoderis:
@maggu2810 Tx. I was not sur myself about the preferred nomenklatura. The current Profiles do not exhibit a verb for example.

That being said, they are necessary for the KNX binding (in its current state), but general enough to make them System Profiles I think.

sjka:
Generally I agree with you - the suggested profiles are "generic" enough to be put as system profiles. Still, I'm lacking a little the other use-cases apart from the generic/advanced KNX binding right now. In any case, they won't hurt having and sometimes the use-cases pop up with the availability of stuff.

Nevertheless, I don't like the name of the "passive" profile in particular, as it suggests it wouldn't be doing anything - which clearly is not the case here. Obviously, this depends on how you are looking at it. But is is confusing as it is right now. 

Problem is: I don't have a better suggestion immediately. Something like "read-only" crossed my mind (an the other one would be a "write-only" counterpart, but I'm not fully convinced either (because it still entails a point of view). Do you (or anybody else) maybe have a better idea?

kgoderis:
@sjka what about system:inform instead of passive and system:execute instead of command ?

kgoderis:
@sjka @maggu2810 can we push a bit ahead with this small mod? I need it rather urgently because my operational environment is in limbo because of the change in default Profile behaviour

sjka:
Sure...

In a very rare moment of pure ingenuity (thanks for pushing this 😉 ) `system:inbound` just crossed my mind. IMHO this would match best what we want to express: It allows the communication flow only from the device/binding-side to flow into the system (and we could complement it with a `system:outbound` profile if ever needed at a later point in time). How about that?

The `system:command` profile is fine as it is, as it only allows commands and no updates in both directions. I'd leave it as it is.

kgoderis:
@sjka ok let's do that. I will update the PR accordingly

kaikreuzer:
Sorry to come late to the party, but I am actually a bit puzzled about the discussion here.
We mainly introduced profiles for trigger channels and having the default ones for state channels was mainly a side effect. I never thought that there should be a need to add additional ones, especially as this could lead to unexpected behavior for the bindings.

So IF there is really a need for them, I would love to learn about the use case behind it as we should make sure that we do not add features (and thus complexity) to the core framework, which simply duplicates something that can already be done with it.

So @kgoderis, with reference to the discussion in https://github.com/eclipse/smarthome/issues/4669 and #4525, why do you think those profiles are required?

kgoderis:
@kaikreuzer Since the direct link between Bindings and Items evaporated with the introduction of Channels by the ESH framework the KNX binding can not be ported to ESH. Period.

As raised on various occasions, and discussed, for example https://github.com/eclipse/smarthome/pull/2881, https://github.com/openhab/openhab2-addons/pull/1438 and https://github.com/openhab/openhab2-addons/pull/2323, the KNX binding  due to the nature of its architecture needs to "guess" how the GA's are linked  together  in order to prevent "circular" Telegrams on the KNX bus. Since the GA's are defined at the level of Channels, and Channels are linked to Items, you *have* to know and control how Commands and States "flow" from one Channel to another (over the Items), in order to avoid the problem. As you can read from the discussions, the problem can not be solved at the level of the Binding, and even reversing the regression (https://github.com/eclipse/smarthome/issues/4669) will make  https://github.com/openhab/openhab2-addons/pull/2323 fail. Moreover, it was  https://github.com/openhab/openhab2-addons/pull/2323 that led to the creation of the system:follow (then called system:slave) Profile in the first place.  

You know, I have now developed and refactored the KNX 2.0 binding at least 3-4 times over the last 2 years. I have experimented with different solutions to overcome this problem. None of them work well. 

I seem to be ONLY person running a native ESH KNX binding in an operational environment, and I fear that unless other try to do the same, I will remain the only person shouting in the desert that the KNX binding needs a solution at the level of the framework. Either by supplying it with more information (e.g, source of Commands and States) so that I can figure out what to do by itself, or either go the way of Profiles or alike. *sigh* this has been the reason why I clearly indicated that I would cease to develop the KNX binding all together. No problem I guess, I will add some Binding specific Profiles to my edition of the KNX binding and run that until someone finally acknowledges there is an issue to solve and comes up with a solution......

lewie:
@kgoderis 

No, you are not alone in the desert!!

Without knowing the origin, it is definitely not possible to recognize self-initiated events coming from the openHAB event bus. Only if the origin is known, the KNX binding can prevent (filter) such openHAB events from repeatedly being sent back to the KNX bus as a KNX telegram.

This realization cost me many weeks of work and testing when I wanted to make the KNX1 binding fit for openHAB2. Gave up KNX1 - It is not possible!

We all know, it is a hard stony way to a native KNX2 binding.

tkuehne71:
Let me add a chip to the pile as well. While I am not familiar with the internal workings of the KNX binding and the Framework, let me reiterate from a User's Point of view, that the KNX binding today can run into a state where it makes the KNX bus unusable, since it Spams the bus with circular updates. Not only does this create issues with the Level of traffic, it also means that the UI will not handle setpoints for the affected items correctly (they simply get switched back to their previous state by the circular updates). As you can see from the discussion in https://github.com/openhab/openhab1-addons/issues/4547 a lot of users are running into similar Problems and are voicing their Frustration with OH2 over this.

Best,

Thomas

sjka:
As you can see in openhab/openhab2-addons#2323, the KNX binding actually doesn't even need these profiles but can live with the system profiles perfectly well. 

The question whether events should have an origin or not is unrelated to this PR (don't get me wrong, I absolutely agree that something needs to be done in order to get the KNX 1 binding working as well). But I don't see how _this_ PR is able to solve that issue.

kaikreuzer:
Thanks @sjka.
@kgoderis I would suggest that we have a closer look at the suggested implementation in https://github.com/openhab/openhab2-addons/pull/2323. If we all agree that this is a suitable way to model the channels and that this does not require any additional profiles, I think we could close this PR here, do you agree?

kgoderis:
@kaikreuzer Yes, let's do that.

kaikreuzer:
Imho, openhab/openhab2-addons#2323 works pretty well, so I would not see a need for those new profile types here right now - I'd therefore close this PR. If a new requirement comes up, we can at any time re-open it again.

PR #4679, opened 2017-12-04
Increased timeout for item commands and updates to 30 seconds

kaikreuzer:
We so far always allowed handlers up to 30 s to process a handleCommand call - this change brings backward compatibility and avoids too many warnings in the logs on existing bindings.

I also changed the timeout from int to long, so that it is easier to use it with TimeUnits.

### DIFF
--- a/bundles/core/org.eclipse.smarthome.core.thing/src/main/java/org/eclipse/smarthome/core/thing/internal/profiles/ProfileCallbackImpl.java
+++ b/bundles/core/org.eclipse.smarthome.core.thing/src/main/java/org/eclipse/smarthome/core/thing/internal/profiles/ProfileCallbackImpl.java
@@ -12,6 +12,7 @@
  */
 package org.eclipse.smarthome.core.thing.internal.profiles;
 
+import java.util.concurrent.TimeUnit;
 import java.util.function.Function;
 
 import org.eclipse.smarthome.core.common.SafeCaller;
@@ -64,7 +65,7 @@ public class ProfileCallbackImpl implements ProfileCallback {
                 if (ThingHandlerHelper.isHandlerInitialized(thing)) {
                     logger.debug("Delegating command '{}' for item '{}' to handler for channel '{}'", command,
                             link.getItemName(), link.getLinkedUID());
-                    safeCaller.create(handler).onTimeout(() -> {
+                    safeCaller.create(handler).withTimeout(TimeUnit.SECONDS.toMillis(30)).onTimeout(() -> {
                         logger.warn("Handler for thing '{}' takes more than {}ms for handling a command",
                                 handler.getThing().getUID(), SafeCaller.DEFAULT_TIMEOUT);
                     }).build().handleCommand(link.getLinkedUID(), command);
@@ -95,7 +96,7 @@ public class ProfileCallbackImpl implements ProfileCallback {
                 if (ThingHandlerHelper.isHandlerInitialized(thing)) {
                     logger.debug("Delegating update '{}' for item '{}' to handler for channel '{}'", state,
                             link.getItemName(), link.getLinkedUID());
-                    safeCaller.create(handler).onTimeout(() -> {
+                    safeCaller.create(handler).withTimeout(TimeUnit.SECONDS.toMillis(30)).onTimeout(() -> {
                         logger.warn("Handler for thing '{}' takes more than {}ms for handling an update",
                                 handler.getThing().getUID(), SafeCaller.DEFAULT_TIMEOUT);
                     }).build().handleUpdate(link.getLinkedUID(), state);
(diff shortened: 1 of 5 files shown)


maggu2810:
> We so far always allowed handlers up to 30 s to process a handleCommand call

Really? Hasn't been the intention that handleCommand has to be return fast and long running stuff should be handled in another thread?

Shouldn't we use a constant for the 30 seconds instead of adding this magic timeout number at four places?

kaikreuzer:
> Hasn't been the intention that handleCommand has to be return fast and long running stuff should be handled in another thread?

No - I actually searched for any definition, but neither our JavaDoc nor [our tutorial](https://www.eclipse.org/smarthome/documentation/development/bindings/thing-handler.html#handling-commands) says anything like this. Rather the contrary: Developers are asked there to reach out to their devices in those methods. With the nature of remote communication and potential timeouts, the current 5 second timeout is definitely not in line with that. Since I rolled out this 5 second timeout, I have seen many reports like [this one](https://community.openhab.org/t/jetty-update-karaf-4-1-3-upgrade-and-full-lsp-support/36354/90) and we imho shouldn't blame the implementations now all of a sudden for taking more than 5 seconds.

> long running stuff should be handled in another thread?

I thought about whether we should make the API more strict and ask developers to exactly do this. But thinking about it, I noticed that most likely almost all handlers will then simply wrap the whole `handleCommand` implementation in a job that they schedule right away, since almost everything that deals with real hardware might have potentially long-running operations.
Ending up with such implementations cannot be our goal; so it is probably better to deal with the threading in the framework and allow handlers more time for execution.

> Shouldn't we use a constant for the 30 seconds instead of adding this magic timeout number at four places?

Most certainly. Thanks for pointing it out, I have corrected that 😇 .

bundles/core/org.eclipse.smarthome.core.thing/src/main/java/org/eclipse/smarthome/core/thing/internal/profiles/ProfileCallbackImpl.java
```diff
@@ -64,10 +65,11 @@ public void handleCommand(Command command) {
...
                     logger.debug("Delegating command '{}' for item '{}' to handler for channel '{}'", command,
                             link.getItemName(), link.getLinkedUID());
-                    safeCaller.create(handler).onTimeout(() -> {
-                        logger.warn("Handler for thing '{}' takes more than {}ms for handling a command",
-                                handler.getThing().getUID(), SafeCaller.DEFAULT_TIMEOUT);
-                    }).build().handleCommand(link.getLinkedUID(), command);
+                    safeCaller.create(handler).withTimeout(CommunicationManager.THINGHANDLER_EVENT_TIMEOUT)
+                            .onTimeout(() -> {
+                                logger.warn("Handler for thing '{}' takes more than {}ms for handling a command",
+                                        handler.getThing().getUID(), SafeCaller.DEFAULT_TIMEOUT);
```
sjka: `SafeCaller.DEFAULT_TIMEOUT` -> `CommunicationManager.THINGHANDLER_EVENT_TIMEOUT` so that the logging won't lie

bundles/core/org.eclipse.smarthome.core.thing/src/main/java/org/eclipse/smarthome/core/thing/internal/profiles/ProfileCallbackImpl.java
```diff
@@ -95,10 +97,11 @@ public void handleUpdate(State state) {
...
                     logger.debug("Delegating update '{}' for item '{}' to handler for channel '{}'", state,
                             link.getItemName(), link.getLinkedUID());
-                    safeCaller.create(handler).onTimeout(() -> {
-                        logger.warn("Handler for thing '{}' takes more than {}ms for handling an update",
-                                handler.getThing().getUID(), SafeCaller.DEFAULT_TIMEOUT);
-                    }).build().handleUpdate(link.getLinkedUID(), state);
+                    safeCaller.create(handler).withTimeout(CommunicationManager.THINGHANDLER_EVENT_TIMEOUT)
+                            .onTimeout(() -> {
+                                logger.warn("Handler for thing '{}' takes more than {}ms for handling an update",
+                                        handler.getThing().getUID(), SafeCaller.DEFAULT_TIMEOUT);
```
sjka: dito

kaikreuzer:
@sjka Thx, fixed.

maggu2810:
If we are fine with a non fast returning `handleCommand` implementation (at least if we don't have this requirement), I am fine with that change.
@sjka if it is okay for you, feel free to merge

sjka:
Yes, I'm fine with it right now. I'm open to discussing this for the future, but I agree with @kaikreuzer that restricting this silently a couple of days before a (minor) release is not a good point time.

PR #4680, opened 2017-12-04
PaperUI: enable trigger channel linking

htreu:
This enables trigger channel linking through PaperUI.

* Trigger channels can only link to existing items
* On-the-fly item creation is not possible for Trigger channels
* Also _dummy_  items will not be created by ChannelItemProvider
* In _simple mode_ the same item selection pops up as in advanced mode


bundles/core/org.eclipse.smarthome.core.thing/src/main/java/org/eclipse/smarthome/core/thing/internal/ChannelItemProvider.java
```diff
@@ -265,6 +301,29 @@ private void createItemForLink(ItemChannelLink link) {
         }
     }
 
+    private Collection<String> getItemType(Channel channel) {
```
maggu2810: Can we change the method name to something that state what the implementation is doing (and perhaps we should use plural)?
For example `getProfileItemTypes` or something similar.
Sure, it is a private method, but it should be clear that the item types depend on the profiles and not on the accepted item type of the channel (without reading the implementation).

bundles/core/org.eclipse.smarthome.core.thing/src/main/java/org/eclipse/smarthome/core/thing/profiles/ProfileAdvisor.java
```diff
@@ -41,4 +42,20 @@
...
+     *
+     * Please note:
+     * <ul>
+     * <li>This will override any default behavior
+     * <li>A "profile" configuration on the link will override this suggestion
+     * </ul>
+     *
+     * @param channelType the {@link ChannelType} of the linked channel
+     * @param itemType the linked itemType (not applicable for trigger channels)
+     * @return the profile identifier or {@code null} if this advisor
```
triller-telekom: "if this advisor" ...? something is missing here

bundles/core/org.eclipse.smarthome.core.thing/src/main/java/org/eclipse/smarthome/core/thing/profiles/ProfileAdvisor.java
```diff
@@ -41,4 +42,20 @@
...
+     * Suggest a custom profile for a given {@link ChannelType} (and potentially also the itemType).
+     *
+     * Please note:
+     * <ul>
+     * <li>This will override any default behavior
+     * <li>A "profile" configuration on the link will override this suggestion
+     * </ul>
+     *
+     * @param channelType the {@link ChannelType} of the linked channel
+     * @param itemType the linked itemType (not applicable for trigger channels)
```
triller-telekom: I thought trigger channels can now be linked to items, can't they?

extensions/ui/org.eclipse.smarthome.ui.paper/web-src/js/controllers/controller.things.spec.js
```diff
@@ -175,51 +203,51 @@ describe('module PaperUI.controllers.things', function() {
...
-        });
-        it('should create item and link', function() {
-            spyOn(itemService, 'create').and.returnValue(deferred.promise).and.callThrough();
-            spyOn(itemService.create({}).$promise, 'then').and.returnValue(deferred.promise);
-            scope.newItemName = "N";
-            scope.createAndLink();
-            expect(itemService.create).toHaveBeenCalled();
-        });
-    });
+    // describe('tests for LinkChannelDialogController', function() {
```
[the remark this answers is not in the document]
htreu: ...or fixing the test actually.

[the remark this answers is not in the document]
htreu: Sadly the code now has `new Set` which is an ES6 feature and not supported by our PhantomJS which runs the tests. I have to investigate on how to update PhantomSJ or switch to another browser, like Chrome headless.

bundles/core/org.eclipse.smarthome.core.thing/src/main/java/org/eclipse/smarthome/core/thing/profiles/ProfileAdvisor.java
htreu: This is just a copy of the original method javaDoc. We have to ask @sjka aboiut the missing text here.

bundles/core/org.eclipse.smarthome.core.thing/src/main/java/org/eclipse/smarthome/core/thing/profiles/ProfileAdvisor.java
htreu: Yes, they do. The suggested ProfileTypeUID however will not depend on the given itemType when it comes to trigger channels.

htreu:
Test added again, fell free to comment or merge :-)

triller-telekom:
LGTM now.

bundles/core/org.eclipse.smarthome.core.thing/src/main/java/org/eclipse/smarthome/core/thing/internal/profiles/SystemProfileFactory.java
```diff
@@ -114,4 +128,13 @@ public ProfileTypeUID getSuggestedProfileTypeUID(Channel channel, @Nullable Stri
         return SUPPORTED_PROFILE_TYPE_UIDS;
     }
 
+    @Reference(cardinality = ReferenceCardinality.MANDATORY, policy = ReferencePolicy.DYNAMIC)
```
kaikreuzer: mandatory+dynamic does not really bring much value. Why not merely do `@Reference` here?

bundles/core/org.eclipse.smarthome.core.thing/src/main/java/org/eclipse/smarthome/core/thing/profiles/ProfileAdvisor.java
kaikreuzer: "... does not have any advice"

PR #4691, opened 2017-12-07
Adapt ref-tag script to allow releases without qualifier

kaikreuzer:
Also
- changed sed syntax to work correctly on Mac
- included version replacement inside package.json files
- added a small usage at the top, to be extended by experts...

### DIFF
--- a/tools/ref-tag/create.sh
+++ b/tools/ref-tag/create.sh
@@ -1,5 +1,13 @@
 #!/bin/bash
 
+# USAGE:
+# ./tools/ref-tag/create.sh <commitid>
+#
+# Environment variables to set before execution:
+# VERSION_NEW_MMR - the new version (e.g. 1.2.0) to use. Optional, if not set, the current version is kept
+# VERSION_NEW_QUALI - the new version qualifier. Possible values are "RELEASE", "SNAPSHOT", or any other string like "b1" or "alpha2". Optional, defaults to "SNAPSHOT"
+
+
 #
 # Define some functions first
 #
@@ -125,7 +133,7 @@ esac
 # Prepare some version old variables
 #
 VERSION_OLD_MMR=${VERSION_OLD%-SNAPSHOT}
-VERSION_OLD_QUALI="${VERSION_OLD_MMR}.qualifier"
+VERSION_OLD_MMRQ="${VERSION_OLD_MMR}.qualifier"
 
 #
 # Generate new version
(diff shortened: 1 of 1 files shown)


tools/ref-tag/create.sh
```diff
@@ -176,7 +189,7 @@ for FILE in \
   docs/documentation/development/testing.md \
   products/org.eclipse.smarthome.repo/category.xml
 do
-  sed 's:'"${VERSION_OLD_QUALI}"':'"${VERSION_NEW_QUALI}"':g' -i "${FILE}"
+  sed -e "s:${VERSION_OLD_MMRQ}:${VERSION_NEW_MMRQ}:g" -i "" ${FILE}
```
maggu2810: I don't understand the change from `-i "${FILE}"` to `-i "" ${FILE}`.
Can you explain please?

kaikreuzer: This is from the man page on Mac:
```
     sed [-Ealn] [-e command] [-f command_file] [-i extension] [file ...]
...
     -i extension
             Edit files in-place, saving backups with the specified extension.  If a zero-length extension is given, no backup will be saved.  It is not recommended to give a zero-length extension when in-place editing files, as you risk
             corruption or partial content in situations where disk space is exhausted, etc.

```
As you can see, the "-i" option expects a parameter and I provide a "zero-length extension" to it.

maggu2810: This change will break the ref-tool on Linux system.

man page on Linux:
```
       -i[SUFFIX], --in-place[=SUFFIX]

              edit files in place (makes backup if SUFFIX supplied)
```

Let's we want to change every `a` to `b` in the file `test123`.

```text
$ sed s:a:b:g -i "" test123 
sed: can't read : No such file or directory

$ sed s:a:b:g -i  test123 
(all fine)
```

So, on Linux it would be:
* `-i"" filename`
* `-i="" filename`
* or simply `-i filename`

kaikreuzer: Hrmpf. The easiest solution would be if you'd buy yourself a Mac ;-)
Not sure how to resolve this. Probably would need to google a while to figure out some OS independent way of doing that operation.

maggu2810: We could write a wrapper function that checks the OS and execute different command or we could use something similar to this one: https://stackoverflow.com/a/22084103

I would prefer an OS detection like this one: https://stackoverflow.com/a/3466183

tools/ref-tag/create.sh
```diff
@@ -136,22 +144,27 @@ fi
...
     *)
-      VERSION_NEW="${VERSION_NEW_MMR}.${VERSION_NEW_QUALI}"
+      VERSION_NEW="${VERSION_NEW_MMR}-${VERSION_NEW_QUALI}"
+      VERSION_NEW_MMRQ="${VERSION_NEW_MMR}.${VERSION_NEW_QUALI}"
       ;;
   esac
 else
-  VERSION_NEW_QUALI=".qualifier"
-  VERSION_NEW="${VERSION_NEW_MMR}".{VERSION_NEW_QUALI}
+  VERSION_NEW="${VERSION_NEW_MMR}-SNAPSHOT"
```
maggu2810: Didn't we need a 
```
VERSION_NEW_MMRQ="${VERSION_NEW_MMR}.qualifier"
```
here?

maggu2810:
WRT to the changes about the variable names, I could try to change the script.

maggu2810:
I could also write the whole script as a Python 3.5 one

tools/ref-tag/create.sh
kaikreuzer: looks like it

kaikreuzer:
I am pretty bad in all sort of scripting, so if you want to take over and enhance the script, please do so! I only had to do the changes in this PR in order to be able to do the release tag, but I am perfectly fine if you come up with a different/better solution.

kaikreuzer:
Closing this as @maggu2810 is about to come up with a much better script 👍

PR #4698, opened 2017-12-08
Print details on missing service references in tests

sjka:
When tidying up my repository I found this little piece of code which came in handy a couple of times already. It helps immediately spotting the reason why a service was not available (i.e. whenever `getService(...)` returns `null`) during test execution, e.g.:

```
Components implementing org.eclipse.smarthome.core.thing.ManagedThingProvider:
23 [UNSATISFIED_REFERENCE] org.eclipse.smarthome.core.thing.ManagedThingProvider in org.eclipse.smarthome.core.thing
	StorageService (org.eclipse.smarthome.core.storage.StorageService)
		No component implementing org.eclipse.smarthome.core.storage.StorageService is currently registered.
```

Before I spend any more time cleaning it up and making it nicer to read - do you agree that this would be helpful? And if so, where should it print it to? Is sysout okay?

### DIFF
--- a/bundles/test/org.eclipse.smarthome.test/META-INF/MANIFEST.MF
+++ b/bundles/test/org.eclipse.smarthome.test/META-INF/MANIFEST.MF
@@ -31,6 +31,9 @@ Import-Package: com.google.common.collect,
  org.junit.matchers;version="4.0.0",
  org.mockito,
  org.osgi.framework,
+ org.osgi.framework.dto,
  org.osgi.service.cm;resolution:=optional,
+ org.osgi.service.component.runtime,
+ org.osgi.service.component.runtime.dto,
  org.slf4j
 Require-Bundle: org.junit, org.mockito, org.hamcrest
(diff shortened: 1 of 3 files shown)


maggu2810:
I didn't do a full review but I have a short look at the code and also like the idea.
sysout for tests is IMHO okay your implementation already uses a reference so easy to change.

I agree it would be helpful and like it.

sjka:
Okay, thanks for the feedback!
I'm gonna clean it up a little then and let you know once it's worth having a closer look 😉

htreu:
Since I already gained benefit from a similar implementation this is for sure really helpful. Go on and let me know once your clean up is done.

sjka:
I've refactored it a little to make it more readable. 
Happy reviewing.

maggu2810:
@sjka Please don't squash yourself. I already reviewed the PR six days ago. Can you explain me what has been changed?

sjka:
Argh, sorry...
I just factored the new code out of `JavaOSGiTest` into its own class. Nothing changed logic-wise.

PR #4699, opened 2017-12-08
Make LSP server port configurable

### LINKED ISSUE #4589: Make LSP port configurable
Currently, the [LSP port is hard-coded](https://github.com/eclipse/smarthome/blob/d570593172b729de6e205665af94971dec3d1e7f/bundles/model/org.eclipse.smarthome.model.lsp/src/main/java/org/eclipse/smarthome/model/lsp/internal/ModelServer.java#L42). It should be possible to change it through configuration (using Configuration admin).

sjka:
Make LSP server port configurable

fixes #4589

### DIFF
--- a/bundles/model/org.eclipse.smarthome.model.lsp/src/main/java/org/eclipse/smarthome/model/lsp/internal/ModelServer.java
+++ b/bundles/model/org.eclipse.smarthome.model.lsp/src/main/java/org/eclipse/smarthome/model/lsp/internal/ModelServer.java
@@ -15,17 +15,21 @@ package org.eclipse.smarthome.model.lsp.internal;
 import java.io.IOException;
 import java.net.ServerSocket;
 import java.net.Socket;
+import java.util.Map;
 import java.util.concurrent.ExecutionException;
 import java.util.concurrent.Future;
 
 import org.eclipse.lsp4j.jsonrpc.Launcher;
 import org.eclipse.lsp4j.launch.LSPLauncher;
 import org.eclipse.lsp4j.services.LanguageClient;
+import org.eclipse.smarthome.config.core.ConfigurableService;
 import org.eclipse.smarthome.model.script.ScriptServiceUtil;
 import org.eclipse.smarthome.model.script.engine.ScriptEngine;
 import org.eclipse.xtext.ide.server.LanguageServerImpl;
+import org.osgi.framework.Constants;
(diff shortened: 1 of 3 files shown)


bundles/model/org.eclipse.smarthome.model.lsp/src/main/java/org/eclipse/smarthome/model/lsp/internal/ModelServer.java
```diff
@@ -40,22 +45,31 @@
...
     private ServerSocket socket;
 
     private ScriptServiceUtil scriptServiceUtil;
     private ScriptEngine scriptEngine;
     private Injector injector;
 
     @Activate
-    public void activate() {
+    public void activate(Map<String, Object> config) {
+        final int port = config.containsKey(KEY_PORT) ? Integer.parseInt(config.get(KEY_PORT).toString())
```
maggu2810: Shouldn't we catch a `NumberFormatException` here?

bundles/model/org.eclipse.smarthome.model.lsp/src/main/java/org/eclipse/smarthome/model/lsp/internal/ModelServer.java
```diff
@@ -66,23 +80,25 @@ public void deactivate() {
                 socket.close();
             }
         } catch (IOException e) {
-            logger.error("Error shutting down the Language Server", e);
+            logger.debug("Error shutting down the Language Server", e);
         }
     }
 
-    private void listen() {
+    private void listen(int port) {
```
maggu2810: WRT to the changed logging, shouldn't this method return a success indication (`boolean`), so the caller could at least log one error message that the server is not running / listen?

sjka: As #4722 fixed the superfluous logging in a much better way, I guess my change to the log severities here became obsolete. I'll simply remove that commit.

maggu2810:
WRT to the imminent ref-tag, I keep it up to @kaikreuzer to merge it now or after that tag.

kaikreuzer:
I should have mentioned: I created ref-tag 0.10.0.b1 yesterday, so feel free to merge whatever you are up to!

Issue #4697, opened 2017-12-08
OSGiEventManager: prevent timeouts


maggu2810:
WRT: openhab/openhab-distro#600

What do you thin about chaning the implementation of the `OSGiEventManager` a little bit?

Proposal of the changes:
* add a blocking queue of `org.osgi.service.event.Event` to the member variables
* the activate method starts a thread that reads the elements from the queue
* the deactivate method stops the thread started by the activate method
* the reader thread reads elements from the blocking queue and handles it as it is done at the moment in the `handleEvent(org.osgi.service.event.Event osgiEvent)` method
* the `handleEvent(org.osgi.service.event.Event osgiEvent)` method that is called be the `EventAdmin` adds the `Event` to the blocking queue

With this approach we can consume the events as fast as possible and return the control to the event admin.
So we don't need to change the global "blocking event handler" handling and keep the blacklisting as set default.

sjka:
Might be worth a try - especially to see if that _still_ would cause timeouts 😉

kaikreuzer:
I think the most important part on that would be that we should refrain from doing any logging inbetween 😎

maggu2810:
So, I assign this issue to me and will try to implement that change.

PR #4700, opened 2017-12-08
don't block event admin on OSGi event handling

maggu2810:
We do not want to block the event admin while handling the received OSGi
events.
So, we consume the OSGi events and add them to a internal queue.
A separate handler thread will then proceed the elements of that queue
in order.

Fixes: https://github.com/eclipse/smarthome/issues/4697

### DIFF
--- a/bundles/core/org.eclipse.smarthome.core/src/main/java/org/eclipse/smarthome/core/internal/events/EventHandler.java
+++ b/bundles/core/org.eclipse.smarthome.core/src/main/java/org/eclipse/smarthome/core/internal/events/EventHandler.java
@@ -16,7 +16,11 @@ import java.util.Arrays;
 import java.util.HashSet;
 import java.util.Map;
 import java.util.Set;
+import java.util.concurrent.ConcurrentHashMap;
+import java.util.concurrent.CopyOnWriteArraySet;
 
+import org.eclipse.jdt.annotation.NonNullByDefault;
+import org.eclipse.jdt.annotation.Nullable;
 import org.eclipse.smarthome.core.common.SafeCaller;
 import org.eclipse.smarthome.core.events.Event;
 import org.eclipse.smarthome.core.events.EventFactory;
@@ -25,18 +29,17 @@ import org.eclipse.smarthome.core.events.EventSubscriber;
 import org.slf4j.Logger;
 import org.slf4j.LoggerFactory;
 
-import com.google.common.collect.SetMultimap;
-
 /**
  * Handle Eclipse SmartHome events encapsulated by OSGi events.
  *
  * @author Markus Rathgeb - Initial contribution
  */
+@NonNullByDefault
 public class EventHandler {
 
     private final Logger logger = LoggerFactory.getLogger(EventHandler.class);
 
-    private final SetMultimap<String, EventSubscriber> typedEventSubscribers;
+    private final ConcurrentHashMap<String, CopyOnWriteArraySet<EventSubscriber>> typedEventSubscribers;
     private final Map<String, EventFactory> typedEventFactories;
     private final SafeCaller safeCaller;
 
@@ -47,7 +50,7 @@ public class EventHandler {
      * @param typedEventFactories the event factories indexed by the event type
      * @param safeCaller the safe caller to use
      */
-    public EventHandler(final SetMultimap<String, EventSubscriber> typedEventSubscribers,
+    public EventHandler(final ConcurrentHashMap<String, CopyOnWriteArraySet<EventSubscriber>> typedEventSubscribers,
             final Map<String, EventFactory> typedEventFactories, final SafeCaller safeCaller) {
         this.typedEventSubscribers = typedEventSubscribers;
         this.typedEventFactories = typedEventFactories;
@@ -78,7 +81,8 @@ public class EventHandler {
         }
     }
 
-    private void handleEvent(final String type, final String payload, final String topic, final String source) {
+    private void handleEvent(final String type, final String payload, final String topic,
+            final @Nullable String source) {
         final EventFactory eventFactory = typedEventFactories.get(type);
         if (eventFactory == null) {
             logger.warn("Could not find an Event Factory for the event type '{}'.", type);
@@ -112,8 +116,8 @@ public class EventHandler {
         return subscribers;
     }
 
-    private Event createESHEvent(final EventFactory eventFactory, final String type, final String payload,
-            final String topic, final String source) {
+    private @Nullable Event createESHEvent(final EventFactory eventFactory, final String type, final String payload,
+            final String topic, final @Nullable String source) {
         Event eshEvent = null;
         try {
             eshEvent = eventFactory.createEvent(type, topic, payload, source);
--- a/bundles/core/org.eclipse.smarthome.core/src/main/java/org/eclipse/smarthome/core/internal/events/EventHandler.java
+++ b/bundles/core/org.eclipse.smarthome.core/src/main/java/org/eclipse/smarthome/core/internal/events/EventHandler.java
@@ -1,0 +1,144 @@
+/**
+ * Copyright (c) 2014,2017 Contributors to the Eclipse Foundation
+ *
+ * See the NOTICE file(s) distributed with this work for additional
+ * information regarding copyright ownership.
+ *
+ * This program and the accompanying materials are made available under the
+ * terms of the Eclipse Public License 2.0 which is available at
+ * http://www.eclipse.org/legal/epl-2.0
+ *
+ * SPDX-License-Identifier: EPL-2.0
+ */
+package org.eclipse.smarthome.core.internal.events;
+
+import java.util.Arrays;
+import java.util.HashSet;
+import java.util.Map;
+import java.util.Set;
+
+import org.eclipse.smarthome.core.common.SafeCaller;
+import org.eclipse.smarthome.core.events.Event;
+import org.eclipse.smarthome.core.events.EventFactory;
+import org.eclipse.smarthome.core.events.EventFilter;
+import org.eclipse.smarthome.core.events.EventSubscriber;
+import org.slf4j.Logger;
+import org.slf4j.LoggerFactory;
+
+import com.google.common.collect.SetMultimap;
+
+/**
+ * Handle Eclipse SmartHome events encapsulated by OSGi events.
+ *
+ * @author Markus Rathgeb - Initial contribution
+ */
+public class EventHandler {
+
+    private final Logger logger = LoggerFactory.getLogger(EventHandler.class);
+
+    private final SetMultimap<String, EventSubscriber> typedEventSubscribers;
+    private final Map<String, EventFactory> typedEventFactories;
+    private final SafeCaller safeCaller;
+
+    /**
+     * Create a new event handler.
+     *
+     * @param typedEventSubscribers the event subscribers indexed by the event type
+     * @param typedEventFactories the event factories indexed by the event type
+     * @param safeCaller the safe caller to use
+     */
+    public EventHandler(final SetMultimap<String, EventSubscriber> typedEventSubscribers,
+            final Map<String, EventFactory> typedEventFactories, final SafeCaller safeCaller) {
+        this.typedEventSubscribers = typedEventSubscribers;
+        this.typedEventFactories = typedEventFactories;
+        this.safeCaller = safeCaller;
+    }
+
+    public void handleEvent(org.osgi.service.event.Event osgiEvent) {
+        logger.trace("Handle OSGi event (event: {})", osgiEvent);
+
+        Object typeObj = osgiEvent.getProperty("type");
+        Object payloadObj = osgiEvent.getProperty("payload");
+        Object topicObj = osgiEvent.getProperty("topic");
+        Object sourceObj = osgiEvent.getProperty("source");
+
+        if (typeObj instanceof String && payloadObj instanceof String && topicObj instanceof String) {
+            String typeStr = (String) typeObj;
+            String payloadStr = (String) payloadObj;
+            String topicStr = (String) topicObj;
+            String sourceStr = (sourceObj instanceof String) ? (String) sourceObj : null;
+            if (!typeStr.isEmpty() && !payloadStr.isEmpty() && !topicStr.isEmpty()) {
+                handleEvent(typeStr, payloadStr, topicStr, sourceStr);
+            }
+        } else {
+            logger.error(
+                    "The handled OSGi event is invalid. Expect properties as string named 'type', 'payload' and 'topic'. "
+                            + "Received event properties are: {}",
+                    Arrays.toString(osgiEvent.getPropertyNames()));
+        }
+    }
+
+    private void handleEvent(final String type, final String payload, final String topic, final String source) {
+        final EventFactory eventFactory = typedEventFactories.get(type);
+        if (eventFactory == null) {
+            logger.warn("Could not find an Event Factory for the event type '{}'.", type);
+            return;
+        }
+
+        final Set<EventSubscriber> eventSubscribers = getEventSubscribers(type);
+        if (eventSubscribers.isEmpty()) {
+            return;
+        }
+
+        final Event eshEvent = createESHEvent(eventFactory, type, payload, topic, source);
+        if (eshEvent == null) {
+            return;
+        }
+
+        dispatchESHEvent(eventSubscribers, eshEvent);
+    }
+
+    private Set<EventSubscriber> getEventSubscribers(String eventType) {
+        Set<EventSubscriber> eventTypeSubscribers = typedEventSubscribers.get(eventType);
+        Set<EventSubscriber> allEventTypeSubscribers = typedEventSubscribers.get(EventSubscriber.ALL_EVENT_TYPES);
+
+        Set<EventSubscriber> subscribers = new HashSet<EventSubscriber>();
+        if (eventTypeSubscribers != null) {
+            subscribers.addAll(eventTypeSubscribers);
+        }
+        if (allEventTypeSubscribers != null) {
+            subscribers.addAll(allEventTypeSubscribers);
+        }
+        return subscribers;
+    }
+
+    private Event createESHEvent(final EventFactory eventFactory, final String type, final String payload,
+            final String topic, final String source) {
+        Event eshEvent = null;
+        try {
+            eshEvent = eventFactory.createEvent(type, topic, payload, source);
+        } catch (Exception e) {
+            logger.error(
+                    "Creation of ESH-Event failed, "
+                            + "because one of the registered event factories has thrown an exception: {}",
+                    e.getMessage(), e);
+        }
+        return eshEvent;
+    }
+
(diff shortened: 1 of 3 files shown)


maggu2810:
Will add nullness annotations and try do drop the usage of google collections.

maggu2810:
Will add a custom thread name tomorrow.

maggu2810:
Any objections?

kaikreuzer:
I will need some quiet minutes to properly review this, because http://bholley.net/images/posts/thistall.jpg - I therefore also want @sjka to review as well.

Also would want to do testing on an openHAB distro. So please allow me a few days.

bundles/core/org.eclipse.smarthome.core/src/main/java/org/eclipse/smarthome/core/internal/events/OSGiEventManager.java
```diff
@@ -91,21 +84,25 @@ public void removedService(ServiceReference reference, Object service) {
...
     }
 
-    private final Logger logger = LoggerFactory.getLogger(OSGiEventManager.class);
+    private final Map<String, EventFactory> typedEventFactories = new ConcurrentHashMap<String, EventFactory>();
 
-    private EventAdmin osgiEventAdmin;
+    private final ConcurrentHashMap<String, CopyOnWriteArraySet<EventSubscriber>> typedEventSubscribers = new ConcurrentHashMap<>();
 
-    private final Map<String, EventFactory> typedEventFactories = new ConcurrentHashMap<String, EventFactory>();
+    // private final SetMultimap<String, EventSubscriber> typedEventSubscribers = Multimaps
```
kaikreuzer: can this be removed?

bundles/core/org.eclipse.smarthome.core/src/main/java/org/eclipse/smarthome/core/internal/events/OSGiEventManager.java
```diff
@@ -271,36 +197,52 @@ public Void run() throws Exception {
...
+        }
+        if ((value = event.getTopic()) == null || value.isEmpty()) {
+            throw new IllegalArgumentException(String.format(errorMsg, "topic"));
+        }
     }
 
     private void assertValidState(EventAdmin eventAdmin) throws IllegalStateException {
-        Preconditions.checkArgument(eventAdmin != null, "The event bus module is not available!");
+        if (eventAdmin == null) {
+            throw new IllegalArgumentException("The event bus module is not available!");
```
kaikreuzer: This should be an `IllegalStateException`.

bundles/core/org.eclipse.smarthome.core/src/main/java/org/eclipse/smarthome/core/internal/events/EventHandler.java
```diff
@@ -0,0 +1,148 @@
...
+    private final SafeCaller safeCaller;
+
+    /**
+     * Create a new event handler.
+     *
+     * @param typedEventSubscribers the event subscribers indexed by the event type
+     * @param typedEventFactories the event factories indexed by the event type
+     * @param safeCaller the safe caller to use
+     */
+    public EventHandler(final ConcurrentHashMap<String, CopyOnWriteArraySet<EventSubscriber>> typedEventSubscribers,
```
kaikreuzer: Shouldn't this take a `Map<String, Set<EventSubscriber>>`?

bundles/core/org.eclipse.smarthome.core/src/main/java/org/eclipse/smarthome/core/internal/events/ThreadedEventHandler.java
```diff
@@ -0,0 +1,94 @@
...
+
+    /**
+     * Create a new threaded event handler.
+     *
+     * @param typedEventSubscribers the event subscribers indexed by the event type
+     * @param typedEventFactories the event factories indexed by the event type
+     * @param safeCaller the safe caller to use
+     */
+    public ThreadedEventHandler(
+            final ConcurrentHashMap<String, CopyOnWriteArraySet<EventSubscriber>> typedEventSubscribers,
```
kaikreuzer: Shouldn't this take a `Map<String, Set<EventSubscriber>>`?

bundles/core/org.eclipse.smarthome.core/src/main/java/org/eclipse/smarthome/core/internal/events/EventHandler.java
```diff
@@ -0,0 +1,148 @@
...
+                            + "Received event properties are: {}",
+                    Arrays.toString(osgiEvent.getPropertyNames()));
+        }
+    }
+
+    private void handleEvent(final String type, final String payload, final String topic,
+            final @Nullable String source) {
+        final EventFactory eventFactory = typedEventFactories.get(type);
+        if (eventFactory == null) {
+            logger.warn("Could not find an Event Factory for the event type '{}'.", type);
```
kaikreuzer: Could we reduce this to debug again (as in the current master)?

bundles/core/org.eclipse.smarthome.core/src/main/java/org/eclipse/smarthome/core/internal/events/ThreadedEventHandler.java
```diff
@@ -0,0 +1,94 @@
...
+            final Map<String, EventFactory> typedEventFactories, final SafeCaller safeCaller) {
+        thread = new Thread(() -> {
+            final EventHandler worker = new EventHandler(typedEventSubscribers, typedEventFactories, safeCaller);
+            while (running.get()) {
+                try {
```
kaikreuzer: Do we run any risk to come across a RuntimeException here, which will make our thread terminate (and thus kill all further event processing from that moment on)?

bundles/core/org.eclipse.smarthome.core/src/main/java/org/eclipse/smarthome/core/internal/events/OSGiEventManager.java
```diff
@@ -91,21 +84,25 @@ public void removedService(ServiceReference reference, Object service) {
...
 
+    private EventAdmin osgiEventAdmin;
+
     private SafeCaller safeCaller;
 
     @Activate
     protected void activate(ComponentContext componentContext) {
+        eventHandler = new ThreadedEventHandler(typedEventSubscribers, typedEventFactories, safeCaller);
+
         eventSubscriberServiceTracker = new EventSubscriberServiceTracker(componentContext.getBundleContext());
```
kaikreuzer: Slightly unrelated, but as you are refactoring the code anyhow in this PR: Do you happen to know, why we work with a service tracker here instead of using a 0..n dynamic dependency? I especially wonder in the light of #4775 whether we are doing anything wrong here...

bundles/core/org.eclipse.smarthome.core/src/main/java/org/eclipse/smarthome/core/internal/events/EventHandler.java
```diff
@@ -0,0 +1,148 @@
...
+ * Handle Eclipse SmartHome events encapsulated by OSGi events.
+ *
+ * @author Markus Rathgeb - Initial contribution
+ */
+@NonNullByDefault
+public class EventHandler {
+
+    private final Logger logger = LoggerFactory.getLogger(EventHandler.class);
+
+    private final ConcurrentHashMap<String, CopyOnWriteArraySet<EventSubscriber>> typedEventSubscribers;
```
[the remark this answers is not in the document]
maggu2810: It depends.

The ThreadedEventHandler currently forces a `ConcurrentHashMap` (`ConcurrentMap` would be enough) with a value type of `CopyOnWriteArraySet`.
The ThreadedEventHandler uses the `CopyOnWriteArraySet` (and also the `ConcurrentMap`) to signal that the set (and map) will be accessed by multiple threads and so prefers that type.

The EventHandler itself could be used in a multi-threaded and single-threaded environment, so he could (and I agree, he should) use `Map<String, Set<EventSubscriber>)`.

The annoying thing is that e.g.
`Collection<SubType>` is not a replacement for `Collection<BaseType>`.
So we need to:
* introduce an unchecked cast
* remove the constrained from the `ThreadedEventHandler` and loose this information

What do you prefer?

bundles/core/org.eclipse.smarthome.core/src/main/java/org/eclipse/smarthome/core/internal/events/EventHandler.java
maggu2810: See above

bundles/core/org.eclipse.smarthome.core/src/main/java/org/eclipse/smarthome/core/internal/events/EventHandler.java
maggu2810: If you want :wink:

bundles/core/org.eclipse.smarthome.core/src/main/java/org/eclipse/smarthome/core/internal/events/EventHandler.java
```diff
@@ -0,0 +1,148 @@
...
+        }
+
+        dispatchESHEvent(eventSubscribers, eshEvent);
+    }
+
+    private Set<EventSubscriber> getEventSubscribers(String eventType) {
+        Set<EventSubscriber> eventTypeSubscribers = typedEventSubscribers.get(eventType);
+        Set<EventSubscriber> allEventTypeSubscribers = typedEventSubscribers.get(EventSubscriber.ALL_EVENT_TYPES);
+
+        Set<EventSubscriber> subscribers = new HashSet<EventSubscriber>();
```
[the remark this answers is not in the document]
maggu2810: Done

bundles/core/org.eclipse.smarthome.core/src/main/java/org/eclipse/smarthome/core/internal/events/OSGiEventManager.java
maggu2810: Sure, shouldn't be present at all.

bundles/core/org.eclipse.smarthome.core/src/main/java/org/eclipse/smarthome/core/internal/events/EventHandler.java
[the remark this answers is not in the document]
maggu2810: Done (has been present in the previous code base already).

bundles/core/org.eclipse.smarthome.core/src/main/java/org/eclipse/smarthome/core/internal/events/OSGiEventManager.java
maggu2810: Fixed.

bundles/core/org.eclipse.smarthome.core/src/main/java/org/eclipse/smarthome/core/internal/events/ThreadedEventHandler.java
maggu2810: See above

bundles/core/org.eclipse.smarthome.core/src/main/java/org/eclipse/smarthome/core/internal/events/ThreadedEventHandler.java
maggu2810: Good catch, added a catch branch.

bundles/core/org.eclipse.smarthome.core/src/main/java/org/eclipse/smarthome/core/internal/events/OSGiEventManager.java
maggu2810: I need to guess, but AFAIK ESH used service trackers a lot "in the early days".
And IIRC the core should not use DS "in the early days".

We could migrate that from ServiceTracker to DS, but this would be another change and is not part of that "topic".

maggu2810:
> There are a few null warnings when getting elements from collections - you might want to add specific annotations to avoid those.

Isn't this normal as long as we don't use EEAs?

bundles/core/org.eclipse.smarthome.core/src/main/java/org/eclipse/smarthome/core/internal/events/EventHandler.java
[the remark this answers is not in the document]
kaikreuzer: > The ThreadedEventHandler uses the CopyOnWriteArraySet to signal that the set (and map) will be accessed by multiple threads

But it only does read accesses (and you internally treat it as a [plain set](https://github.com/eclipse/smarthome/pull/4700/files#diff-761ec20041a2951418930542adc99ef0R106), so I do not really understand why we need a CopyOnWriteArraySet? Isn't it a general issue that caller should be careful with changes to a set that it has already passed to someone else?

bundles/core/org.eclipse.smarthome.core/src/main/java/org/eclipse/smarthome/core/internal/events/OSGiEventManager.java
kaikreuzer: ok.

kaikreuzer:
> Isn't this normal as long as we don't use EEAs?

Well, those are introduced by the NonNullByDefault annotation, which is also applied to the types in collections (and thus a `myCollection.get()` is always assumed to be nonNull). I don't remember the details on that discussion, but it does not seem to be related to EEAs. @triller-telekom Can you clarify how that should be best handled?

bundles/core/org.eclipse.smarthome.core/src/main/java/org/eclipse/smarthome/core/internal/events/EventHandler.java
```diff
@@ -0,0 +1,148 @@
...
+ *
+ * @author Markus Rathgeb - Initial contribution
+ */
+@NonNullByDefault
+public class EventHandler {
+
+    private final Logger logger = LoggerFactory.getLogger(EventHandler.class);
+
+    private final ConcurrentHashMap<String, CopyOnWriteArraySet<EventSubscriber>> typedEventSubscribers;
+    private final Map<String, EventFactory> typedEventFactories;
```
triller-telekom: If an `EventFactory` inside `typedEventFactories` can be `null` as checked here:
https://github.com/eclipse/smarthome/pull/4700/files#diff-761ec20041a2951418930542adc99ef0R87

you might want to define `typedEventFactories` like this to avoid the warning at the `null` check:
`private final Map<String, @Nullable EventFactory> typedEventFactories;`

bundles/core/org.eclipse.smarthome.core/src/main/java/org/eclipse/smarthome/core/internal/events/EventHandler.java
```diff
@@ -0,0 +1,148 @@
...
+
+        dispatchESHEvent(eventSubscribers, eshEvent);
+    }
+
+    private Set<EventSubscriber> getEventSubscribers(String eventType) {
+        Set<EventSubscriber> eventTypeSubscribers = typedEventSubscribers.get(eventType);
+        Set<EventSubscriber> allEventTypeSubscribers = typedEventSubscribers.get(EventSubscriber.ALL_EVENT_TYPES);
+
+        Set<EventSubscriber> subscribers = new HashSet<>();
+        if (eventTypeSubscribers != null) {
```
triller-telekom: To get rid of the `null` warning here and in line 113, you can define `typedEventSubscribers` like this:

`private final ConcurrentHashMap<String, @Nullable CopyOnWriteArraySet<EventSubscriber>> typedEventSubscribers;`

bundles/core/org.eclipse.smarthome.core/src/main/java/org/eclipse/smarthome/core/internal/events/ThreadedEventHandler.java
```diff
@@ -0,0 +1,96 @@
...
+                    final Event event = queue.poll(1, TimeUnit.HOURS);
+                    if (event == null) {
```
triller-telekom: This is the only `null` check warning that can only be removed using external annotations because `queue` is of type `java.util.concurrent.BlockingQueue` which is not annotated.

maggu2810:
It is about EEAs.

A call `map.get(key)` can always return null if the key is not available in the map, regardless if the map is using a nullable or a non-null value. The type hint does not matter in this case.

A non-null marked value should (with correct EEAs) add a warning if you put a null value to a map.

So, state that values could be null (and allow to put null values into the map) just to get rid of the warning caused by the missing EEA does not make sense.

bundles/core/org.eclipse.smarthome.core/src/main/java/org/eclipse/smarthome/core/internal/events/EventHandler.java
maggu2810: But it is not correct, the map should not contain a null value. We should not add wrong annotations just to workaround the missing EEAs.
You set the severity from errors to warnings some time ago, just because of this...

bundles/core/org.eclipse.smarthome.core/src/main/java/org/eclipse/smarthome/core/internal/events/EventHandler.java
maggu2810: See other comments

triller-telekom:
@maggu2810 Thank you fort he clarification, so you meant the second paragraph of https://github.com/maggu2810/smarthome/pull/6#issuecomment-353037388

Then I am fine with keeping the warning until we have EEAs.

bundles/core/org.eclipse.smarthome.core/src/main/java/org/eclipse/smarthome/core/internal/events/EventHandler.java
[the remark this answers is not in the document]
maggu2810: The osgi event handler implementation writes to the set.
The event handler reads from that sets.
Both itself does not need to use concurrency protected

The threaded event manager bundles the two event handler as the one writes in one thread and the other reads from another thread. The threaded event manager is the one that would like to state, "I introduce a potential concurrency issue here, so you need to use a copy on write array set to solve this issue".

So, I still think the `ThreadedEventHandler` is the one that would like to protect the caller from doing something wrong.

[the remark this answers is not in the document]
maggu2810: I will come up with a change that you can review...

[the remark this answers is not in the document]
maggu2810: Please have a look again. Now only the `ThreadedEventHandler` knows about the specific chosen collection implementations.

kaikreuzer:
@maggu2810 Thanks for the clarification, I indeed misinterpreted the annotation on the type wrongly (i.e. as being applicable for the get and not for the put).

PR #4701, opened 2017-12-08
fix license header(s)

maggu2810:


maggu2810:
I don't dare to say it, but I just started the Maven plugin without looking at the changes in detail.

Issue #3657, opened 2017-06-14
[Paper UI] Show configuration page of newly added things, if required params are not set


kaikreuzer:
If a new Thing is added from the Inbox, which does not have all required parameters set in the config (e.g. because some user password is required), no handler is instantiated and the Thing status goes to "configuration pending".

It is difficult for the user to figure out that he needs to go into the Thing configuration to fill in some more details, before he can use the device.

I'd therefore suggest that the Paper UI should directly jump to the Thing configuration page in such a situation.

kaikreuzer:
@htreu As discussed :-)

Pavion:
See also some community ideas here:
https://community.openhab.org/t/paperui-prompt-for-credentials-config-after-discovery/36892/6

htreu:
Hi @Pavion, I read [your suggestions ](https://community.openhab.org/t/paperui-prompt-for-credentials-config-after-discovery/36892/6)and like `#2` quite a lot:
Upon discovery approval you could check the configuration description of the thing and apply the check you suggested (are there empty && required properties in the config) and then redirect to the thing edit page instead of thing view. The check of the config description should be done here too, as the thing might have empty properties which are not required but set to default values by the binding.

`#1` is, as you mentioned a too generic solution, `#3` would only be the start of a way more complex wizard-like approach which I would like to start only if we have the requirements to do so. 

I would be happy to see your PR here!

Pavion:
Hello Henning and thank you for your answer. As you can see, I've submitted a PR, but can't get past the automatic checks although I've signed the ECA as required. Also I can't understand why Travis CI build fails. Do I miss something..?
As it's a rather small change regarding but one function, I'd have no problem with you or someone else taking over and implementing this solution. For me it's not the matter of putting my name somewhere but to contribute by offering a solution, which is feasible both for a developer and an end user. 

_P.S. Schönen zweiten Advent! :)_

htreu:
Hi @Pavion, I just saw your PR and joined the discussion. Thanks for taking the time and contributing! Especially for PaperUI and JavaScript there are not so many contributors.

_Ich bin etwas spät dran, Danke Dir :-)_

triller-telekom:
This was implemented in PR #4702 so this issue can be closed. @sjka @kaikreuzer

PR #4702, opened 2017-12-08
On adding Thing from Inbox check for missing settings (#3657)

Pavion:
Implementation of the issue #3657:

After adding a Thing from Inbox an additional check is performed:
- is the Thing **OFFLINE** and has **CONFIGURATION_ERROR** _or_
- is the Thing _not_ **ONLINE** and has some empty configuration entries
If so, user will be redirected to the Thing Edit page automatically.

This implementation present a seamless experience for a user, for
example allowing to discover a bridge, be prompted to enter its
credentials and to discover connected Things in a same workflow.

Small drawback: if a Thing takes long to become **ONLINE** _and_ has some
empty non-mandatory configuration entries, user still will be
redirected.

To reduce this risk, this method could be applied to bridges only,
feedback is welcome.

### DIFF
--- a/extensions/ui/org.eclipse.smarthome.ui.paper/web-src/js/controllers.setup.js
+++ b/extensions/ui/org.eclipse.smarthome.ui.paper/web-src/js/controllers.setup.js
@@ -2,6 +2,9 @@ angular.module('PaperUI.controllers.setup', []).controller('SetupPageController'
     $scope.navigateTo = function(path) {
         $location.path('inbox/' + path);
     }
+    $scope.navigateToConfig = function(path) {
+        $location.path('configuration/' + path);
+    }
     $scope.thingTypes = [];
     function getThingTypes() {
         thingTypeRepository.getAll(function(thingTypes) {
@@ -54,15 +57,37 @@ angular.module('PaperUI.controllers.setup', []).controller('SetupPageController'
                 'enableChannels' : !$scope.advancedMode
             }, result.label).$promise.then(function() {
                 thingRepository.setDirty(true);
-                toastService.showDefaultToast('Thing added.', 'Show Thing', 'configuration/things/view/' + thingUID);
                 var thingType = thingTypeRepository.find(function(thingType) {
                     return thingTypeUID === thingType.UID;
                 });
 
-                if (thingType && thingType.bridge) {
-                    $scope.navigateTo('setup/search/' + thingUID.split(':')[0]);
+                var configRequired = false;
+                var thing = thingRepository.find(function(thing) {
+                    return thing.UID === thingUID;
+                });
+                if (thing && thing.statusInfo.status === 'OFFLINE' && thing.statusInfo.statusDetail === 'CONFIGURATION_ERROR') {
+                    configRequired = true;
+                } else if (thing && thing.statusInfo.status !== 'ONLINE') {
+                    for ( var key in thing.configuration) {
+                        if (thing.configuration.hasOwnProperty(key)) {
+                            if (thing.configuration[key] === '') {
+                                configRequired = true;
+                                break;
+                            }
+                        }
+                    }
+                }
+
+                if (configRequired) {
+                    $scope.navigateToConfig('things/edit/' + thingUID);
+                    toastService.showDefaultToast('Thing added and must be configured');
                 } else {
-                    discoveryResultRepository.getAll(true);
+                    toastService.showDefaultToast('Thing added.', 'Show Thing', 'configuration/things/view/' + thingUID);
+                    if (thingType && thingType.bridge) {
+                        $scope.navigateTo('setup/search/' + thingUID.split(':')[0]);
+                    } else {
+                        discoveryResultRepository.getAll(true);
+                    }
                 }
             });
         });
--- a/extensions/ui/org.eclipse.smarthome.ui.paper/web-src/js/controllers.setup.js
+++ b/extensions/ui/org.eclipse.smarthome.ui.paper/web-src/js/controllers.setup.js
@@ -40,7 +40,7 @@ angular.module('PaperUI.controllers.setup', []).controller('SetupPageController'
         discoveryResultRepository.getAll(true);
     };
 
-}).controller('InboxEntryController', function($scope, $mdDialog, $q, inboxService, discoveryResultRepository, thingTypeRepository, thingService, toastService, thingRepository) {
+}).controller('InboxEntryController', function($scope, $mdDialog, $q, inboxService, discoveryResultRepository, thingTypeRepository, thingService, toastService, thingRepository, configDescriptionService) {
     $scope.approve = function(thingUID, thingTypeUID, event) {
         $mdDialog.show({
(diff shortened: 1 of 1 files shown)


Pavion:
There is a valid ISA for my E-Mail now. Do I need to resubmit my PR or is there an another way to rerun the validation check?

extensions/ui/org.eclipse.smarthome.ui.paper/web-src/js/controllers.setup.js
```diff
@@ -54,15 +57,37 @@ angular.module('PaperUI.controllers.setup', []).controller('SetupPageController'
...
+                var configRequired = false;
+                var thing = thingRepository.find(function(thing) {
+                    return thing.UID === thingUID;
+                });
+                if (thing && thing.statusInfo.status === 'OFFLINE' && thing.statusInfo.statusDetail === 'CONFIGURATION_ERROR') {
+                    configRequired = true;
+                } else if (thing && thing.statusInfo.status !== 'ONLINE') {
+                    for ( var key in thing.configuration) {
+                        if (thing.configuration.hasOwnProperty(key)) {
+                            if (thing.configuration[key] === '') {
```
triller-telekom: Don't you have the information regarding the required properties in your `thing.configuration` object? I mean the paperUI shows these fields in red, so the information must be there somehow :)

As @htreu suggested in https://github.com/eclipse/smarthome/issues/3657#issuecomment-349871679 you could then only redirect the user if a required field is missing instead of always redirecting the user if a field is empty.

triller-telekom:
One further remark. The travis failure is not your fault. It could not download a file and has to be restarted to try again. @kaikreuzer Can you please restart travis here?

```
[ERROR] npm ERR! shasum check failed for /home/travis/build/eclipse/smarthome/extensions/ui/org.eclipse.smarthome.ui.paper/npm_tmp/npm-7587-2b7bb585/registry.npmjs.org/material-design-icons/-/material-design-icons-2.0.0.tgz
[ERROR] npm ERR! Expected: a9c79ddf0b84ef6b2fe48a0d1157b42568a81f5a
[ERROR] npm ERR! Actual:   9bf6d1d81f6fa2f9315e4fa83ee504a81cad667f
[ERROR] npm ERR! From:     https://registry.npmjs.org/material-design-icons/-/material-design-icons-2.0.0.tgz
```

extensions/ui/org.eclipse.smarthome.ui.paper/web-src/js/controllers.setup.js
Pavion: Hello and thanks for your answer. Sure, it was also my intention. Sadly this information is not available in JS. Configuration Object in this layer is a simple Map. In my case:
`{host: "127.0.0.1", protocol: "http", password: "", poll: 5, port: "5000", username: ""}`
here password is mandatory and username not (for this test only).

maggu2810:
> @kaikreuzer Can you please restart travis here?

I was so free and triggered a new build.

extensions/ui/org.eclipse.smarthome.ui.paper/web-src/js/controllers.setup.js
htreu: There is a service to retrieve config descriptions for a given config description URI. Check out the `configDescriptionService` and review its usages for an example.
With the the result from this service you get the meta data to check for required fields.

Pavion: Thanks for this tip!! 
I've removed my old approach almost entirely and replaced it with looking into `configDescriptionService` of this `thing-type` and checking for being empty and `required`.  
If the call to `configDescriptionService` fails, it falls back to old behavior.

extensions/ui/org.eclipse.smarthome.ui.paper/web-src/js/controllers.setup.js
```diff
@@ -54,15 +57,40 @@ angular.module('PaperUI.controllers.setup', []).controller('SetupPageController'
...
+                });
+
+                if (thing) {
+                    var configRequired = false;
+                    configDescriptionService.getByUri({
+                        uri : 'thing-type:' + thing.thingTypeUID
+                    }, function(configDescription) {
+                        if (configDescription.parameters) {
+                            for (var i = 0; i < configDescription.parameters.length; i++) {
+                                if (configDescription.parameters[i]['defaultValue'] === '' && configDescription.parameters[i]['required']) {
```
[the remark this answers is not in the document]
htreu: In addition: You may also access the properties like `parameter.required`. Or is there any reason to use the associative array syntax?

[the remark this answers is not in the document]
Pavion: Thanks for pointing that out, it'll be done! But as a global question first: 
`required` and no `defaultValue` is a simple case, which should be handled here. 
But what is your opinion on `required` and empty `defaultValue===''`? 
Should an empty (but specified) defaultValue be handled here too or not? 
In `thing-type.xml` it would be: 
`            
<default></default>
<required>true</required>
`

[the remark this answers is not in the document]
htreu: You are right, the empty default value can be ignored here. The empty `default` tag should not be accepted during review anyway.

Pavion:
I've found an other case to be handled, connected with the last comment from @htreu 
if I have a `required` config entry, whose value I don't know at the time of discovery, I basically have three options:

1. I would set `<default></default>` in `things-type.xml` and don't bother me with supplying properties while handling discovery in my binding's handler (why should I, if I can't). This behavior is doubtful and should be avoided according to @htreu  
2. I would not set `<default></default>` but supply an empty parameter: `properties.put(PASSWORD, "");` in my binding. This will init a Thing, but will fail on missing password (which is not that bad but unneeded, as I've already knew it would). This way also isn't obvious for a developer.  
3. I would not set `<default></default>` **and** supply no `property`. This will fail on Thing init instantly (`Thing not initializable`). 

All cases should now be handled properly but I see the third case as the preferred one, as it causes minimal overhead and is the most intuitive.  

As for the implementation, first the new Thing with its `configuration` is retrieved, then the  `configDescription` of its `Thing Type`. All `required` parameter of the Type are now compared with Thing configuration. Is there no such parameter (3rd case) or if there is a parameter, which is empty (1st and 2nd cases) user will be redirected. In all other cases old behavior stays.

htreu:
@Pavion I'm very happy to see you working on Paper UI as it really needs some more attention (review the [list of paper UI specific issues](https://github.com/eclipse/smarthome/labels/UI-PaperUI)). It would be awesome to see more of your contributions in the future. Feel free to create PRs (and of cause issues) and ping me for review and discussion.

Cheers _und Danke_  :-)
Henning

Pavion:
Thank you Henning.
Sure I'll take a look at the other issues as soon as I find some time. 
Have a nice Christmas!

PR #4703, opened 2017-12-08
Sitemap REST API: fix for the new timeout field

lolodomo:

### DIFF
--- a/bundles/io/org.eclipse.smarthome.io.rest.sitemap/src/main/java/org/eclipse/smarthome/io/rest/sitemap/internal/SitemapResource.java
+++ b/bundles/io/org.eclipse.smarthome.io.rest.sitemap/src/main/java/org/eclipse/smarthome/io/rest/sitemap/internal/SitemapResource.java
@@ -571,6 +571,7 @@ public class SitemapResource implements RESTResource, SitemapSubscriptionCallbac
      *
      * @param widgets
      *            the widgets of the page to observe
+     * @return true if the timeout is reached
      */
     private boolean waitForChanges(EList<Widget> widgets) {
         long startTime = (new Date()).getTime();
@@ -593,7 +594,7 @@ public class SitemapResource implements RESTResource, SitemapSubscriptionCallbac
         for (GenericItem item : items) {
             item.removeStateChangeListener(listener);
         }
-        return !timeout;
+        return timeout;
     }
 
     /**


lolodomo:
@kaikreuzer : please merge quickly this fix, I made a mistake in my PR #4673, I think `waitForChanges` was returning true in case the timeout was reached but in fact it was the reverse !
We need this fix in the coming version 2.2.
PS: PR for Android app is submitted.

bundles/io/org.eclipse.smarthome.io.rest.sitemap/src/main/java/org/eclipse/smarthome/io/rest/sitemap/internal/SitemapResource.java
```diff
@@ -548,21 +548,21 @@ private Sitemap getSitemap(String sitemapname) {
     }
 
     private boolean blockUnlessChangeOccurs(String sitemapname, String pageId) {
-        boolean timeout = false;
+        boolean changeOccurred = true;
```
kaikreuzer: Is it correct to claim that changeOccurred if there is no sitemap with the given name?

lolodomo: If I set it to false, the caller will consider that the timeout was reached which is wrong in the case you mention.
Maybe I should reverse the logic so that this method returns true when the timeout is reached.

kaikreuzer: Might be easier to read then - up to you.

lolodomo:
Change done.
Finally, the code change is reduced like that.

bundles/io/org.eclipse.smarthome.io.rest.sitemap/src/main/java/org/eclipse/smarthome/io/rest/sitemap/internal/SitemapResource.java
```diff
@@ -215,7 +215,7 @@ public Response getPageData(@Context HttpHeaders headers,
             // Make the REST-API pseudo-compatible with openHAB 1.x
             // The client asks Atmosphere for server push functionality,
             // so we do a simply listening for changes on the appropriate items
-            // The blocking has a timeout of 30 seconds. If this timeout is reached,
+            // The blocking has a timeout of 30 seconds. If this timeout expired,
```
kaikreuzer: I do not actually understand what you mean by "timeout expired". I would have assumed that you mean by this that we have hit the timeout - but then it is the same as "is reached" and I wonder why you have changed it? Imho mixing "expiring" and "timout" is not helpful to explain what has happened...

lolodomo:
Ok, the PR is now fixed with an unchanged comment.

lolodomo:
@kaikreuzer : any chance that you create a new ESH build today ? ;) If you can do that, I will ask @digitaldan to merge the corresponding fix in the Android app so that the fix can be included in the coming today Android app version 2.0. If not, the fix will be included later in a version 2.0.1.

kaikreuzer:
The track record of our ESH nightly build does not really look as if we'd have a successful stable build: https://ci.eclipse.org/smarthome/job/SmartHomeDistribution-Nightly/
Not sure why this is all red for days now...

kaikreuzer:
Green again. Let's give the stable a try: https://ci.eclipse.org/smarthome/job/SmartHomeDistribution-Stable/246/
But afterwards, I have to get the openHAB Jenkins up again, which just broke down this morning :-(

lolodomo:
Oh yes, it looks like there is a problem with https://openhab.ci.cloudbees.com/

lolodomo:
The cloudbees link leads to a page saying "We're sorry - this DEV@cloud account is currently unavailable". Any idea what's going on @kaikreuzer ?

kaikreuzer:
As I said: It is down. I contacted support, but they will only get back to me during business hours.

PR #4704, opened 2017-12-08
ported DiscoveryServiceRegistryOSGITest to Java

sjka:


bundles/config/org.eclipse.smarthome.config.discovery.test/src/test/groovy/org/eclipse/smarthome/config/discovery/DiscoveryServiceRegistryOSGiTest.java
```diff
@@ -0,0 +1,383 @@
...
+                new ThingTypeUID(ANY_BINDING_ID_2, ANY_THING_TYPE_2), 3);
+
+        discoveryServiceFaultyMock = new DiscoveryServiceMock(new ThingTypeUID(FAULTY_BINDING_ID, FAULTY_THING_TYPE),
+                1, true);
+
+        extendedDiscoveryServiceMock = new ExtendedDiscoveryServiceMock(
+                new ThingTypeUID(EXTENDED_BINDING_ID, EXTENDED_THING_TYPE), 1, true);
+
+        serviceRegs.add(
+                bundleContext.registerService(DiscoveryService.class.getName(), discoveryServiceMockForBinding1, null));
```
sjka: @afuechsel this was the nitty gritty detail: the JavaOSGiTest only remembers one service registration per interface hence didn't cleanup properly (I might as well fix _that_, but not tonight anymore...). So no matter if groovy or java runs first, the second one is always gonna fail because of the junk services.

afuechsel:
@sjka Thanks for your effort

PR #4715, opened 2017-12-11
fixed JavaOSGiTest

afuechsel:
fixed JavaOSGiTest to save multiple service registrations per interface name

### DIFF
--- a/bundles/test/org.eclipse.smarthome.test/src/main/java/org/eclipse/smarthome/test/java/JavaOSGiTest.java
+++ b/bundles/test/org.eclipse.smarthome.test/src/main/java/org/eclipse/smarthome/test/java/JavaOSGiTest.java
@@ -18,7 +18,6 @@ import static org.junit.Assert.assertThat;
 import java.util.ArrayList;
 import java.util.Dictionary;
 import java.util.HashMap;
-import java.util.Iterator;
 import java.util.List;
 import java.util.Map;
 import java.util.function.Predicate;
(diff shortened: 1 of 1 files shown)


bundles/test/org.eclipse.smarthome.test/src/main/java/org/eclipse/smarthome/test/java/JavaOSGiTest.java
```diff
@@ -245,19 +253,15 @@ private BundleContext initBundleContext() {
...
-                final ServiceRegistration<?> otherReg = regs.next();
-                if (otherReg == reg) {
-                    regs.remove();
-                }
-            }
+        ServiceRegistration<?> reg = null;
+        List<ServiceRegistration<?>> regList = registeredServices.remove(interfaceName);
+        if (regList != null) {
+            reg = regList.get(0);
+            registeredServices.get(interfaceName).forEach(r -> r.unregister());
```
htreu: You just removed the map entry for `interfaceName` so this call will result in NPE.

afuechsel: ah, damn. Copy/Paste error.

bundles/test/org.eclipse.smarthome.test/src/main/java/org/eclipse/smarthome/test/java/JavaOSGiTest.java
```diff
@@ -200,10 +199,19 @@ private BundleContext initBundleContext() {
...
         return srvReg;
     }
 
+    private void saveServiceRegistration(final String interfaceName, final ServiceRegistration<?> srvReg) {
+        List<ServiceRegistration<?>> regs = registeredServices.get(interfaceName);
+        if (regs == null) {
+            regs = new ArrayList<>();
+        }
+        regs.add(srvReg);
+        registeredServices.put(interfaceName, regs);
```
maggu2810: This should be moved to the if branch. A new entry needs only to be pushed if regs is null.

PR #4717, opened 2017-12-11
WIP: Provide information about multiple instance services for UIs

triller-telekom:
The newly introduced MultipleInstanceServiceInfo interface serves as a
marker service for the service pid returned by its getServicePID()
method.

Every service that should be configured multiple times with different
configurations should have the configurationPolicy
ConfigurationPolicy.REQUIRE and thus it will not be started without a
proper configuration.

In addition one service implementing MultipleInstanceServiceInfo is
neccessary that is started immediately which announces the availability
of the forementioned-service by returning the configurationPid of it in
its getServicePID() method.


triller-telekom:
The name `MultipleInstanceServiceInfo` of the interface is still not yet final. I am open for better suggestions.

triller-telekom:
Superseded by #4847

PR #4718, opened 2017-12-11
Wemo: Set thing status to offline if polling fails or bridge config i…

### LINKED ISSUE #4712: [WeMo] [Bug] Thing not set to offline if polling fails
I missed to implement this, should be a two-liner only in every handler.

triller-telekom:
…s wrong

Fixes #4712


hmerk:
@triller-telekom 
Thanks Stefan, LGTM

extensions/binding/org.eclipse.smarthome.binding.wemo/src/main/java/org/eclipse/smarthome/binding/wemo/handler/WemoCoffeeHandler.java
```diff
@@ -106,6 +106,7 @@ public void run() {
 
             } catch (Exception e) {
                 logger.debug("Exception during poll : {}", e);
+                updateStatus(ThingStatus.OFFLINE, ThingStatusDetail.COMMUNICATION_ERROR);
```
kaikreuzer: For communication errors, it is always helpful to add `e.getMessage()` as a third parameter.

triller-telekom: Alright, done.

PR #4719, opened 2017-12-11
Use dynamic free port for FSInternetRadio test

htreu:
Take a free TCP port for the embedded jetty when running the FSInternetRadioHanderJavaTest to make it more robust against parallel execution (e.g. when running on the eclipse jenkins).


extensions/binding/org.eclipse.smarthome.binding.fsinternetradio.test/src/test/java/org/eclipse/smarthome/binding/fsinternetradio/test/TestPortUtil.java
```diff
@@ -0,0 +1,59 @@
...
+     * @throws IllegalStateException if unable to find a free port
+     */
+    public static int findFreePort() {
+        ServerSocket socket = null;
+        try {
+            socket = new ServerSocket(0);
+            socket.setReuseAddress(true);
+            int port = socket.getLocalPort();
+            try {
+                socket.close();
```
maggu2810: Why do you close the socket here? It should already be done in the finally method.
It should be enough to write `return socket.getLocalPort();` here.

extensions/binding/org.eclipse.smarthome.binding.fsinternetradio.test/src/test/java/org/eclipse/smarthome/binding/fsinternetradio/test/TestPortUtil.java
```diff
@@ -0,0 +1,59 @@
...
+                // Ignore IOException on close()
+            }
+            return port;
+        } catch (IOException e) {
+        } finally {
+            if (socket != null) {
+                try {
+                    socket.close();
+                } catch (IOException e) {
+                }
```
maggu2810: Add your comment `// Ignore IOException on close()` here.

extensions/binding/org.eclipse.smarthome.binding.fsinternetradio.test/src/test/java/org/eclipse/smarthome/binding/fsinternetradio/test/TestPortUtil.java
```diff
@@ -0,0 +1,59 @@
...
+            }
+        }
+        throw new IllegalStateException("Could not find a free TCP/IP port to start embedded Jetty HTTP Server on");
```
maggu2810: The only reason to not return the port is if above (opening socket, not the finally branch) the `IOException` is thrown.
So move this method to the catch branch and embed the `IOException` into the ISE.

PR #4722, opened 2017-12-11
reduce debug level if no client connection is established

### LINKED ISSUE #4667: LSP error when stopping OH
Running OH snapshot 1104, I got this error when stopping OH:
```
2017-12-02 09:21:24.729 [ERROR] [thome.model.lsp.internal.ModelServer] - Error accepting the client connection
java.net.SocketException: Socket closed
        at java.net.PlainSocketImpl.socketAccept(Native Method) ~[?:?]
        at java.net.AbstractPlainSocketImpl.accept(AbstractPlainSocketImpl.java:409) ~[?:?]
        at java.net.ServerSocket.implAccept(ServerSocket.java:545) ~[?:?]
        at java.net.ServerSocket.accept(ServerSocket.java:513) ~[?:?]
        at org.eclipse.smarthome.model.lsp.internal.ModelServer.listen(ModelServer.java:80) ~[?:?]
        at org.eclipse.smarthome.model.lsp.internal.ModelServer.lambda$0(ModelServer.java:58) ~[?:?]
        at java.lang.Thread.run(Thread.java:745) [?:?]
```

kaikreuzer:
It is completely normal and expected that an IOException occurs, if the server is shutdown, so there is no reason to log in error level here.

fixes #4667

### DIFF
--- a/bundles/model/org.eclipse.smarthome.model.lsp/src/main/java/org/eclipse/smarthome/model/lsp/internal/ModelServer.java
+++ b/bundles/model/org.eclipse.smarthome.model.lsp/src/main/java/org/eclipse/smarthome/model/lsp/internal/ModelServer.java
@@ -51,11 +51,8 @@ public class ModelServer {
     private ScriptEngine scriptEngine;
     private Injector injector;
 
-    private boolean isShuttingDown = false;
-
     @Activate
     public void activate() {
-        isShuttingDown = false;
         injector = Guice.createInjector(new RuntimeServerModule(scriptServiceUtil, scriptEngine));
         new Thread(() -> {
             listen();
@@ -64,7 +61,6 @@ public class ModelServer {
 
     @Deactivate
     public void deactivate() {
-        isShuttingDown = true;
         try {
             if (socket != null && !socket.isClosed()) {
                 socket.close();
@@ -86,8 +82,8 @@ public class ModelServer {
                         handleConnection(client);
                     }, "Client " + client.getRemoteSocketAddress()).start();
                 } catch (IOException e) {
-                    if (!isShuttingDown) {
-                        logger.error("Error opening socket for client connections: {}", e.getMessage());
+                    if (!socket.isClosed()) {
+                        logger.error("Error accepting client connection: {}", e.getMessage());
                     }
                 }
             }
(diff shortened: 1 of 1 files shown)


bundles/model/org.eclipse.smarthome.model.lsp/src/main/java/org/eclipse/smarthome/model/lsp/internal/ModelServer.java
```diff
@@ -82,7 +82,7 @@ private void listen() {
                         handleConnection(client);
                     }, "Client " + client.getRemoteSocketAddress()).start();
                 } catch (IOException e) {
-                    logger.error("Error accepting the client connection", e);
+                    logger.debug("Error accepting the client connection: {}", e.getMessage());
```
maggu2810: IMHO it depends if a a closed socket is the only reason for an `IOException`.
Do you know if it is the only one?

If it is not the only possible reason, we should check if the exception is an instance of `java.net.SocketException` and if the socket is closed. In this case we can use debug level or not logging anything.

If there are other cases and it is not a closed socket, we should keep level error.

WDYT?

kaikreuzer: Imho all other reasons are potentially due to the client not doing a proper connection/handshake. And this would be similar like we have HTTP clients accessing our HTTP server - we do not log such requests as an ERROR as the bug is on the client side, not on our side. So I would think that debug is the right choice here.

maggu2810: And what about
```
java.net.SocketException: Too many open files
```
this is an error on the server side that could identify a resource leak.

And it is rather easy to run out of free file handles using some third party code :wink:

kaikreuzer: Ok, you won 🥇 
But how shall we proceed here then? It seems as if the only option is to parse the exception message, because otherwise I cannot differentiate beween "Too many open files" and "Socket closed".

kaikreuzer:
@maggu2810 How about the new implementation?

maggu2810:
An idea has been to request `socket.isClosed()` again a respective exception has been called.

kaikreuzer:
I had this before, but then thought that the socket is probably also closed in many other situations (on "too many files" etc.). Or can we rule this out?

maggu2810:
The server socket is already bound to an port lines above.
And so, it already has an file descriptor (in the operation system) assigned.

On accept a new socket for the incoming connection is created and so the OS would like to fetch another file descriptor.

The failed accept of an incoming connection is okay, if the server socket has been closed.
The server socket will not be closed if there are too many files open, because it stays at it is.

PR #4728, opened 2017-12-13
reduced log level

kaikreuzer:
As long as we do not have an orderly system startup and shutdown process in place, it is a perfectly normal situation that ItemFactories are unregistered on a shutdown before the very last event is sent. This currently results in ugly logs like this on a regular shutdown:

```
2017-12-13 07:55:58.783 [WARN ] [ore.internal.events.OSGiEventManager] - Could not find an Event Factory for the event type 'ThingRemovedEvent'.
2017-12-13 07:55:58.783 [WARN ] [ore.internal.events.OSGiEventManager] - Could not find an Event Factory for the event type 'ThingRemovedEvent'.
2017-12-13 07:55:58.784 [WARN ] [ore.internal.events.OSGiEventManager] - Could not find an Event Factory for the event type 'ThingRemovedEvent'.
2017-12-13 07:55:58.784 [WARN ] [ore.internal.events.OSGiEventManager] - Could not find an Event Factory for the event type 'ThingRemovedEvent'.
2017-12-13 07:55:58.786 [WARN ] [ore.internal.events.OSGiEventManager] - Could not find an Event Factory for the event type 'ItemChannelLinkRemovedEvent'.
2017-12-13 07:55:58.786 [WARN ] [ore.internal.events.OSGiEventManager] - Could not find an Event Factory for the event type 'ItemChannelLinkRemovedEvent'.
2017-12-13 07:55:58.787 [WARN ] [ore.internal.events.OSGiEventManager] - Could not find an Event Factory for the event type 'ItemChannelLinkRemovedEvent'.
2017-12-13 07:55:58.787 [WARN ] [ore.internal.events.OSGiEventManager] - Could not find an Event Factory for the event type 'ItemChannelLinkRemovedEvent'.
2017-12-13 07:55:58.787 [WARN ] [ore.internal.events.OSGiEventManager] - Could not find an Event Factory for the event type 'ItemChannelLinkRemovedEvent'.
2017-12-13 07:55:58.787 [WARN ] [ore.internal.events.OSGiEventManager] - Could not find an Event Factory for the event type 'ItemChannelLinkRemovedEvent'.
2017-12-13 07:55:58.787 [WARN ] [ore.internal.events.OSGiEventManager] - Could not find an Event Factory for the event type 'ItemChannelLinkRemovedEvent'.
2017-12-13 07:55:58.788 [WARN ] [ore.internal.events.OSGiEventManager] - Could not find an Event Factory for the event type 'ItemChannelLinkRemovedEvent'.
2017-12-13 07:55:58.788 [WARN ] [ore.internal.events.OSGiEventManager] - Could not find an Event Factory for the event type 'ItemChannelLinkRemovedEvent'.
2017-12-13 07:55:58.788 [WARN ] [ore.internal.events.OSGiEventManager] - Could not find an Event Factory for the event type 'ItemChannelLinkRemovedEvent'.
```

As long as this is our expected behavior, I'd suggest to reduce the log level of that message.

### DIFF
--- a/bundles/core/org.eclipse.smarthome.core/src/main/java/org/eclipse/smarthome/core/internal/events/OSGiEventManager.java
+++ b/bundles/core/org.eclipse.smarthome.core/src/main/java/org/eclipse/smarthome/core/internal/events/OSGiEventManager.java
@@ -191,7 +191,7 @@ public class OSGiEventManager implements EventHandler, EventPublisher {
                 }
             }
         } else {
-            logger.warn("Could not find an Event Factory for the event type '{}'.", type);
+            logger.debug("Could not find an Event Factory for the event type '{}'.", type);
         }
     }
 


maggu2810:
How do you know that this log message is only produced on shutdown?
Is it because we never have seen it before on a normal running system?

kaikreuzer:
Well, on any properly assembled system, event factories will be there for the events that are sent.
So yes: The theory says that should never happen and in practice we have never seen it either.

But as mentioned above: Imho it is fine to increase it to WARN again, once #1896 is in place, because then we should only send events if we are started and if some factory is missing then, this is a bug that we can flag.

Issue #594, opened 2015-12-03
Date / Time handling


kaikreuzer:
migrated from Bugzilla [#479362](https://bugs.eclipse.org/bugs/show_bug.cgi?id=479362)
status NEW severity _enhancement_ in component _Core_ for _---_
Reported in version _0.8.0_ on platform _All_
Assigned to: Project Inbox

On 2015-10-08 13:25:03 -0400, Markus Rathgeb wrote:

> I would like to remove some warnings and realized that the FilterCriteria (and so some tests) of core.persistence is using the Date object.
> 
> If there is no reason we have to use that, I would like (if there is some time) to move form Date to another class.
> 
> Using Date without a Time does not allow set a begin and end time.
> So, moving to a calendar object should allow an better filtering, too.
> 
> How does the persistence / the whole ESH framework handle time references?
> Are all time references handled in UTC or in local time (e.g. using a begin / end mark, an expression in the rules, ...)?
> 
> If the local time is used do you see any problems if I use a UI on a system that using a different timezone as the system running ESH?

On 2015-10-25 01:42:01 -0400, Kai Kreuzer wrote:

> Yes, currently the assumption is always "local time of runtime".
> 
> > If the local time is used do you see any problems if I use a UI on a system that using a different timezone as the system running ESH?
> 
> Well, the UI would show the time of the runtime (e.g. my charts look the same from whereever I am). They won't tell me the sun is shining at midnight, just because I (and the UI) am on the other side of the earth.
> 
> But in general, I am also fine with moving to Calendar.

On 2015-10-25 05:47:24 -0400, Markus Rathgeb wrote:

> - I would expect that the time (stored in the persistence) is an always increasing value
> - We have to ensure, that it is stored with all informations.
> - DST flag (handle transitions from 2:59 DST to 2:00 non-DST)
> - Timezone (handle changes of time zone correctly)
> - How to handle a decreasing time?
> 
> For the last point I used a monotonic timestamp and a offset in the past.
> But I do not know, what we need to solve / grant at all?
> 
> If there is a method to get data points between a range, IMHO we must ensure, that the time is increasing only (could be solved with the offset, but perhaps we could state, that this case is out of scope of the framework and must be granted by the system).

On 2015-10-26 20:46:31 -0400, Kai Kreuzer wrote:

> Doesn't "milliseconds since 1.1.1970" (i.e. the Date class) fulfill all your requirements, assuming that the runtime does not move and stays during its lifetime in the same time zone?

[...]

PR #4738, opened 2017-12-14
Change FilterCriteria to use Java8's ZonedDateTime

lpapazow:
As suggested in this [discussion](https://github.com/eclipse/smarthome/issues/594), changed FilterCriteria  class to use Java8's ZonedDateTime class, instead of java.util.Date. 
The methods that use java.util.Date are marked as deprecated. 
When converting java.util.Date to ZonedDateTime, the default timezone is used.

Signed-off-by: Lyubomir V. Papazov <lyubomir.papazov@MUSALA.COM>


maggu2810:
Shouldn't we use the timezone provided by the `TimeZoneProvider` OSGi service instead?

lpapazow:
Sure, using this service seems way more appropriate.

However, I'm not sure what would be the best way to access the `TimeZoneProvider `service in the `FilterCriteria` class. I am currently using the default timezone in two of `FilterCriteria`'s methods:

- `setBeginDate(Date beginDate)`
- `setEndDate(Date beginDate)`
 which I've marked as deprecated in this PR.

Do you think it would be appropriate to use the default timezone in these two methods? Alternatively, I could get the `TimeZoneProvider` service from the BundleContext, use that service to get the timezone and keep the timezone into a static variable in the `FilterCriteria` class. Or maybe there is some better way to do this that I can't think of?

maggu2810:
As you marked it deprecated and there is another function that uses a better function argument, it is okay for me to use the system default one at this place.

Doesn't the default timezone / zone id be used [here](https://github.com/eclipse/smarthome/pull/4738/files#diff-e3fb993b083190530555d6efd632e55cR246) too? As this resource is a component, you could depend on the resp. service.

bundles/core/org.eclipse.smarthome.core.persistence/src/main/java/org/eclipse/smarthome/core/persistence/FilterCriteria.java
```diff
@@ -1,155 +1,227 @@
...
-        return this;
-    }
-
-    public FilterCriteria setState(State state) {
-        this.state = state;
-        return this;
-    }
-
-}
+/**
```
triller-telekom: I just noticed that the diff of this file looks rather large and also lines that are not changed are marked as changed. I think this might be a an issue with line endings ore file encoding because the indentation with spaces looks alright.

Could you please run the Eclipse formater again on this file?

bundles/model/org.eclipse.smarthome.model.persistence/OSGI-INF/persistenceextension.xml
```diff
@@ -16,4 +16,5 @@
 <scr:component xmlns:scr="http://www.osgi.org/xmlns/scr/v1.1.0" immediate="true" name="org.eclipse.smarthome.model.persistence.extensions">
    <implementation class="org.eclipse.smarthome.model.persistence.extensions.PersistenceExtensions"/>
    <reference bind="setPersistenceServiceRegistry" cardinality="1..1" interface="org.eclipse.smarthome.core.persistence.PersistenceServiceRegistry" name="PersistenceServiceRegistry" policy="static" unbind="unsetPersistenceServiceRegistry"/>
+   <reference bind="setTimeZoneProvider" cardinality="1..1" interface="org.eclipse.smarthome.core.i18n.TimeZoneProvider" name="TimeZoneProvider" policy="static" unbind="unsetTimeZoneProvider"/>
```
triller-telekom: Since we have switched to DS annotations since a while, could you please convert this xml file to annotations since you are touching it anyway. And while you are at it, would you please be so kind and also convert the `persistencemanager.xml` next to it too? Because then you can place a `.gitignore` in the `OSGI-INF` folder with `/*.xml`.

bundles/model/org.eclipse.smarthome.model.persistence/src/org/eclipse/smarthome/model/persistence/extensions/PersistenceExtensions.java
```diff
@@ -55,11 +59,19 @@ public PersistenceExtensions() {
     protected void setPersistenceServiceRegistry(PersistenceServiceRegistry registry) {
         PersistenceExtensions.registry = registry;
     }
-
+
```
triller-telekom: There are some spaces in this line, please remove them.

bundles/model/org.eclipse.smarthome.model.persistence/src/org/eclipse/smarthome/model/persistence/extensions/PersistenceExtensions.java
```diff
@@ -144,7 +156,7 @@ public static HistoricItem historicState(Item item, AbstractInstant timestamp, S
         if (service instanceof QueryablePersistenceService) {
             QueryablePersistenceService qService = (QueryablePersistenceService) service;
             FilterCriteria filter = new FilterCriteria();
-            filter.setEndDate(timestamp.toDate());
+            filter.setEndDate(ZonedDateTime.ofInstant(timestamp.toDate().toInstant(), TimeZone.getDefault().toZoneId()));
```
triller-telekom: Should this not be `timeZoneProvider.getTimeZone()`?

bundles/ui/org.eclipse.smarthome.ui/OSGI-INF/chartprovider.xml
```diff
@@ -20,4 +20,5 @@
    <service>
       <provide interface="org.eclipse.smarthome.ui.chart.ChartProvider"/>
    </service>
+   <reference bind="setTimeZoneProvider" cardinality="1..1" interface="org.eclipse.smarthome.core.i18n.TimeZoneProvider" name="TimeZoneProvider" policy="static" unbind="unsetTimeZoneProvider"/>
```
triller-telekom: I would like to kindly ask you if you can also convert this to DS-annotations but the `OSGI-INF` folder contains 4 files, so I leave it to you if you want to convert all of them or at least just this one and ignore this single created file from the Eclipse IDE in the `.gitignore` file.

lpapazow:
As requested, I converted all of the `.xml` files in `org.eclipse.smarthome.ui/OSGI-INF` and `org.eclipse.smarthome.model.persistence/OSGI-INF` (6 files totally) to annotations,  deleted all `.xml` files and added `.gitignore` files in these two folders.

Also, in the previous version of this PR the `PersistenceExtensionsTest` were breaking with NPE because they were dependent on a `timeZoneProvider` object which wasn't provided by the previous test configuration, so I added `org.eclipse.equinox.ds` to the configuration of the `tycho-surefire-plugin` in `org.eclipse.smarthome.model.persistence.tests/pom.xml`.

maggu2810:
@triller-telekom Do you want to have a final look?

bundles/ui/org.eclipse.smarthome.ui/src/main/java/org/eclipse/smarthome/ui/internal/chart/defaultchartprovider/DefaultChartProvider.java
```diff
@@ -106,6 +119,16 @@ protected void activate() {
         logger.debug("Available themes for default chart provider: {}", themeNames);
     }
 
+    @Reference(cardinality=ReferenceCardinality.MANDATORY, policy=ReferencePolicy.STATIC)
```
kaikreuzer: Mandatory-Static is the default and thus does not require to be explicitly stated here, right?

maggu2810: Yes, its already implicit set, so no need to do it explicit.

kaikreuzer:
Oops, that comment was too late... As this appears at multiple places in this PR, maybe you could create a new PR for the cleanup - thanks!

PR #4742, opened 2017-12-14
MQTT: Use original topic for subscription and removal

### LINKED ISSUE #4739: MqttBrokerConnection.removeConsumer() is not removing the consumer
I'm facing a problem while using org.eclipse.smarthome.io.transport.mqtt.

If I use addConsumer with a subscriber using the topic "+/+/+/+/+/+" and later on try to remove the same subscriber, the subscriber is not really removed from the list of consumers in MqttBrokerConnection.

In addConsumer() there is a conversion of the topic:
`String topic = StringUtils.replace(StringUtils.replace(subscriber.getTopic(), "+", "[^/]*"), "#", ".*");`

And later on `topic` is used, while in removeConsumer `subscriber.getTopic()` is used.

If the above conversion is added to removeConsumer() it seems to fix the problem.

triller-telekom:
Fixes #4739

### DIFF
--- a/bundles/io/org.eclipse.smarthome.io.transport.mqtt/src/main/java/org/eclipse/smarthome/io/transport/mqtt/MqttBrokerConnection.java
+++ b/bundles/io/org.eclipse.smarthome.io.transport.mqtt/src/main/java/org/eclipse/smarthome/io/transport/mqtt/MqttBrokerConnection.java
@@ -353,14 +353,14 @@ public class MqttBrokerConnection {
      */
     public boolean addConsumer(MqttMessageSubscriber subscriber) throws MqttException {
         // Prepare topic for regex pattern matching taking place in messageArrived.
-        String topic = StringUtils.replace(StringUtils.replace(subscriber.getTopic(), "+", "[^/]*"), "#", ".*");
+        String topic = prepareTopic(subscriber.getTopic());
         synchronized (consumers) {
             List<MqttMessageSubscriber> subscriberList = consumers.get(topic);
             if (subscriberList == null) {
                 subscriberList = new ArrayList<>();
+                consumers.put(topic, subscriberList);
             }
             subscriberList.add(subscriber);
-            consumers.put(subscriber.getTopic(), subscriberList);
         }
         if (isConnected()) {
             try {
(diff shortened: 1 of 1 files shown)


bundles/io/org.eclipse.smarthome.io.transport.mqtt/src/main/java/org/eclipse/smarthome/io/transport/mqtt/MqttBrokerConnection.java
```diff
@@ -360,7 +360,7 @@ public boolean addConsumer(MqttMessageSubscriber subscriber) throws MqttExceptio
                 subscriberList = new ArrayList<>();
             }
             subscriberList.add(subscriber);
-            consumers.put(topic, subscriberList);
+            consumers.put(subscriber.getTopic(), subscriberList);
```
maggu2810: The current state of the patch does not make sense to me.
Five lines above a look up is done to get the `subscriberList` and a new one is created if there is no one.
After that the subscriber is added to that list and put to the consumers using the key `subscriber.getTopic()`.
The lookup above it done using still the key `topic`.

So, if the original topic should be used, it needs to be fixed on all places.

I would also prefer to not use put if we already receive the list. We can add the element to the list without a further put operation.

maggu2810: Another question is why we need to replace parts of the topic at all in this method if the key consists of the original topic...

triller-telekom: > So, if the original topic should be used, it needs to be fixed on all places.

I have done that now

> I would also prefer to not use put if we already receive the list. We can add the element to the list without a further put operation.

Done.

> Another question is why we need to replace parts of the topic at all in this method if the key consists of the original topic...

That's a very good question, the comment in code says something about that it is necessary for comparing it with a regex. But I think this PR is just for making the current behavior consistent. The topic replacement should be discussed in a new issue.

maggu2810:
>> Another question is why we need to replace parts of the topic at all in this method if the key consists of the original topic...

> That's a very good question, the comment in code says something about that it is necessary for comparing it with a regex. But I think this PR is just for making the current behavior consistent. The topic replacement should be discussed in a new issue.

Okay for me. Will you create a new issue and add `@...` for the one that introduced the substring replacement?

triller-telekom:
I have opened #4769 regarding the substring replacement.

maggu2810:
@triller-telekom Thank you

Issue #2801, opened 2017-01-12
Update Guava to a more recent version


sjka:
In ESH, we are currently using Guava 10.1. 

As discussed in https://www.eclipse.org/forums/index.php/t/1065860/, on of the main benefits of this version is that is is slightly smaller than newer versions (1.5M vs. 2.4M). 

However, all Solutions I'm aware of are packaging a Guava 15 or higher anyway. So do we really need to stick to that old version? We ran into troubles a couple of times now, with #2783 being just the latest accident.

Therefore I propose to upgrade to a more recent version. Guava 20 currently is the newest, but I haven't checked it yet for any potential incompatibilities.

kaikreuzer:
> Guava 20 currently is the newest, but I haven't checked it yet for any potential incompatibilities.

If they follow semver, it would mean that 20 is as incompatible to 15 as 10 to 15 ;-)

For the few things that we are actually using, I still think that >2MB is simply terribly fat.
Maybe we should define (in a good OSGi manner) only a few packages from Guava that we allow to use - including their valid version range, which we should then also ensure in the MANIFEST.MF.

sjka:
> the few things that we are actually using

That's a pretty valid point, indeed. I have created #2805 and #2809 in order to eliminate to outliers. Then we are left with only two packages when looking at our sources (i.e. not including imports from 3rd party libraries):

```
  52 com.google.common.base
 107 com.google.common.collect
```

So it would be pretty simple to restrict usage to only these two packages, i.e. not allow usage of additional ones.

Looking at what we are using there, I see a huge overlap with apache commons. When I have some time I will have a look at how much we actually "need" and what could be replaced instantly.

maggu2810:
> Looking at what we are using there, I see a huge overlap with apache commons. When I have some time I will have a look at how much we actually "need" and what could be replaced instantly.

:+1:

kaikreuzer:
> When I have some time I will have a look at how much we actually "need" and what could be replaced instantly.

Sorry to be doubtful about that as well. Imho, using apache commons might not be the best solution either. Actually, Guava seems to be more modern and closer to the specification, at least this is what [Google claimes](https://code.google.com/archive/p/google-collections/wikis/Faq.wiki). The fact that apache commons 2.6 (which we use) does not support generics is indeed not nice.
[...]

PR #4743, opened 2017-12-14
Reduce guava usage

sjka:
see #2801 et al.

This PR doesn't intend to change any behavior - the goal was to eliminate some easy-to-replace guava usages.

### DIFF
--- a/bundles/automation/org.eclipse.smarthome.automation.module.media/src/main/java/org/eclipse/smarthome/automation/module/media/internal/MediaModuleHandlerFactory.java
+++ b/bundles/automation/org.eclipse.smarthome.automation.module.media/src/main/java/org/eclipse/smarthome/automation/module/media/internal/MediaModuleHandlerFactory.java
@@ -12,8 +12,9 @@
  */
 package org.eclipse.smarthome.automation.module.media.internal;
 
-import java.util.ArrayList;
-import java.util.Arrays;
+import static java.util.Arrays.asList;
+import static java.util.Collections.unmodifiableList;
+
 import java.util.Collection;
 
 import org.eclipse.smarthome.automation.Action;
@@ -35,13 +36,14 @@ import org.osgi.service.component.annotations.Reference;
 @Component(service = ModuleHandlerFactory.class)
 public class MediaModuleHandlerFactory extends BaseModuleHandlerFactory {
 
-    private static final Collection<String> types = Arrays.asList(SayActionHandler.TYPE_ID, PlayActionHandler.TYPE_ID);
+    private static final Collection<String> TYPES = unmodifiableList(
+            asList(SayActionHandler.TYPE_ID, PlayActionHandler.TYPE_ID));
     private VoiceManager voiceManager;
     private AudioManager audioManager;
 
     @Override
     public Collection<String> getTypes() {
-        return new ArrayList<>(types);
+        return TYPES;
     }
 
     @Override
--- a/bundles/automation/org.eclipse.smarthome.automation.module.media/src/main/java/org/eclipse/smarthome/automation/module/media/internal/MediaModuleHandlerFactory.java
+++ b/bundles/automation/org.eclipse.smarthome.automation.module.media/src/main/java/org/eclipse/smarthome/automation/module/media/internal/MediaModuleHandlerFactory.java
@@ -13,6 +13,7 @@
 package org.eclipse.smarthome.automation.module.media.internal;
 
 import java.util.ArrayList;
+import java.util.Arrays;
 import java.util.Collection;
 
 import org.eclipse.smarthome.automation.Action;
@@ -27,8 +28,6 @@ import org.eclipse.smarthome.core.voice.VoiceManager;
 import org.osgi.service.component.annotations.Component;
 import org.osgi.service.component.annotations.Reference;
 
-import com.google.common.collect.ImmutableList;
-
 /**
  *
  * @author Kai Kreuzer - Initial contribution
@@ -36,8 +35,7 @@ import com.google.common.collect.ImmutableList;
 @Component(service = ModuleHandlerFactory.class)
 public class MediaModuleHandlerFactory extends BaseModuleHandlerFactory {
 
-    private static final Collection<String> types = ImmutableList.of(SayActionHandler.TYPE_ID,
-            PlayActionHandler.TYPE_ID);
+    private static final Collection<String> types = Arrays.asList(SayActionHandler.TYPE_ID, PlayActionHandler.TYPE_ID);
     private VoiceManager voiceManager;
     private AudioManager audioManager;
 
--- a/bundles/config/org.eclipse.smarthome.config.core/src/main/java/org/eclipse/smarthome/config/core/validation/ConfigValidationException.java
+++ b/bundles/config/org.eclipse.smarthome.config.core/src/main/java/org/eclipse/smarthome/config/core/validation/ConfigValidationException.java
@@ -82,8 +82,7 @@ public final class ConfigValidationException extends RuntimeException {
      * internationalized then the default message is delivered.
      * <p>
      * If there is no TranslationProvider available then this operation will return the default validation messages by
-     * using
-     * {@link ConfigValidationException#getValidationMessages()}.
+     * using {@link ConfigValidationException#getValidationMessages()}.
      *
      * @param locale the locale to be used; if null then the default locale will be used
      *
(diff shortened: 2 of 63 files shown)


maggu2810:
I like that changes :wink:

sjka:
Then there is only Travis who needs to get convinced 😉 

Btw, with this the number of imports is now down to

```bash
➜  find . -name "*.java" -exec grep "import com.google.common" {} \; | sed -e "s#import \(.*\)\.[^\.]*;#\1#g" | sort | uniq -c | sort
   4 com.google.common.base
  28 com.google.common.collect
```

as opposed to 

```bash
  52 com.google.common.base
 107 com.google.common.collect
```

which we had in [January 2017](https://github.com/eclipse/smarthome/issues/2801#issuecomment-272159861).

maggu2810:
> Then there is only Travis who needs to get convinced

As soon as Travis agree with that changes, I will give them a go. :wink:

bundles/config/org.eclipse.smarthome.config.core/src/main/java/org/eclipse/smarthome/config/core/validation/ConfigValidationException.java
```diff
@@ -65,19 +66,20 @@ public ConfigValidationException(Bundle bundle, Collection<ConfigValidationMessa
...
+        return Collections.unmodifiableMap(ret);
     }
 
     /**
      * Retrieves the internationalized validation messages for this exception. If there is no text found to be
      * internationalized then the default message is delivered.
      * <p>
-     * If there is no TranslationProvider available then this operation will return the default validation messages by using
+     * If there is no TranslationProvider available then this operation will return the default validation messages by
+     * using
```
htreu: concat with the next line

bundles/config/org.eclipse.smarthome.config.discovery/src/main/java/org/eclipse/smarthome/config/discovery/inbox/events/InboxEventFactory.java
```diff
@@ -20,7 +20,6 @@
 import org.eclipse.smarthome.core.events.EventFactory;
 import org.osgi.service.component.annotations.Component;
 
-import com.google.common.base.Preconditions;
 import com.google.common.collect.Sets;
```
htreu: why not replace Sets usage in the constructor?

```
public InboxEventFactory() {
    super(new HashSet<>(Arrays.asList(InboxAddedEvent.TYPE, InboxUpdatedEvent.TYPE, InboxRemovedEvent.TYPE)));
}
```

bundles/core/org.eclipse.smarthome.core.test/src/test/java/org/eclipse/smarthome/core/net/HttpServiceUtilTest.java
```diff
@@ -103,10 +100,9 @@ public void shouldReturnUndefinedForException() throws InvalidSyntaxException {
         when(ref2.getProperty(HTTP_PORT)).thenReturn("8080");
         when(ref2.getProperty(Constants.SERVICE_RANKING)).thenReturn("100");
 
-        return Lists.newArrayList(ref1, ref2).toArray(new ServiceReference[0]);
+        return new ServiceReference[] { ref1, ref2 };
```
htreu: come on, what was wrong with the old implementation? btw, blame says it was me ;-)

bundles/core/org.eclipse.smarthome.core.thing.test/src/test/java/org/eclipse/smarthome/core/thing/firmware/ModelRestrictedFirmwareUpdateServiceOSGiTest.java
```diff
@@ -219,9 +219,9 @@ private FirmwareProvider createFirmwareProvider(Firmware... firmwares) {
             @Override
             public Set<Firmware> getFirmwares(ThingTypeUID thingTypeUID, Locale locale) {
                 if (thingTypeUID.equals(thingType.getUID())) {
-                    return Sets.newHashSet(firmwares);
+                    return new HashSet<>(Arrays.asList(firmwares));
                 } else {
-                    return Sets.newHashSet();
+                    return new HashSet<>();
```
htreu: this might also be `Collections.emptySet()`

bundles/automation/org.eclipse.smarthome.automation.module.media/src/main/java/org/eclipse/smarthome/automation/module/media/internal/MediaModuleHandlerFactory.java
```diff
@@ -27,17 +28,14 @@
...
 /**
  *
  * @author Kai Kreuzer - Initial contribution
  */
 @Component(service = ModuleHandlerFactory.class)
 public class MediaModuleHandlerFactory extends BaseModuleHandlerFactory {
 
-    private static final Collection<String> types = ImmutableList.of(SayActionHandler.TYPE_ID,
-            PlayActionHandler.TYPE_ID);
+    private static final Collection<String> types = Arrays.asList(SayActionHandler.TYPE_ID, PlayActionHandler.TYPE_ID);
```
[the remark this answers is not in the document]
sjka: yes, because it is meant to _be_ a constant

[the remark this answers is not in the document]
kaikreuzer: Shouldn't it be TYPES then?

bundles/config/org.eclipse.smarthome.config.discovery/src/main/java/org/eclipse/smarthome/config/discovery/inbox/events/InboxEventFactory.java
sjka: because it technically would even have to be 

```java
super(Collections.unmodifiableSet(new HashSet<>(Arrays.asList(InboxAddedEvent.TYPE, InboxUpdatedEvent.TYPE, InboxRemovedEvent.TYPE))))
```

I'd rather want to have a convenience method which does exactly this (maybe a little optimized), but don't neither clutters the code as much nor drags in a couple of MBs footprint like guava. Nevertheless, I left it out of the scope if _this_ PR to keep it manageable.

bundles/core/org.eclipse.smarthome.core.test/src/test/java/org/eclipse/smarthome/core/net/HttpServiceUtilTest.java
sjka: It worked 😄

bundles/core/org.eclipse.smarthome.core.thing.test/src/test/java/org/eclipse/smarthome/core/thing/firmware/ModelRestrictedFirmwareUpdateServiceOSGiTest.java
sjka: Right - I had a similar case where groovy was involved (and it didn't like it). But you're right in this case.

bundles/config/org.eclipse.smarthome.config.discovery/src/main/java/org/eclipse/smarthome/config/discovery/inbox/events/InboxEventFactory.java
htreu: okay, so this be part of the second or third run then.

sjka:
image

...what a fool I was to merge the other one 😢

bundles/config/org.eclipse.smarthome.config.discovery/src/main/java/org/eclipse/smarthome/config/discovery/inbox/events/InboxEventFactory.java
maggu2810: Why not using something like `Stream.of(InboxAddedEvent.TYPE, InboxUpdatedEvent.TYPE, InboxRemovedEvent.TYPE).collect(toSet())`?

kaikreuzer:
@sjka Looks as if there are still/again some files in conflict. Better ping @maggu2810 or myself quickly after resolving those, so that we do not miss the window when it can be merged...

sjka:
> Better ping [...] quickly after resolving those

@maggu2810 / @kaikreuzer: quick-ping 😄

PR #4744, opened 2017-12-14
drop designer bundles

sjka:


sjka:
Doesn't it feel great? 😄

PR #4752, opened 2017-12-15
[Basic UI] added missing material fonts

sjka:
As per CQ 10013 I understand that there is an "binary, checkin, thirdparty, unmodified" approval for "Material design icons Version: 2.0.0". Therefore I don't see why we shouldn't have them in the Basic UI, along with the others. Or what am I missing?


kaikreuzer:
Is this supposed to address https://github.com/eclipse/smarthome/issues/3523?

How does it relate to the [suggested revert to old codes](https://github.com/eclipse/smarthome/issues/3523#issuecomment-327782694)?

sjka:
No, #3523 somehow dropped out of my mind, so this PR naively tries to fix the broken symbols by adding the material fonts version which we have a valid CQ for, ignoring everything that was discussed in the issue 😩

PR #4753, opened 2017-12-15
[Basic UI] favicon

sjka:
Why shall have the Paper UI a favicon but poor little Basic UI doesn't?


kaikreuzer:
... because it is supposed to be customized by a fragment anyhow?
But sure, we can have one here as well, why not. Btw, does the Classic UI have one?

PR #4755, opened 2017-12-15
extend HSBType

J-N-K:
The HSBType is useful for dimming with the same color, because it is easy to change the brightness channel. However, the HSBType currently only has get, not set methods for hue, saturation and brightness. So changing the brightness requires the creation of a new object. This PR adds the missing set methods.

### DIFF
--- a/bundles/core/org.eclipse.smarthome.core/src/main/java/org/eclipse/smarthome/core/library/types/HSBType.java
+++ b/bundles/core/org.eclipse.smarthome.core/src/main/java/org/eclipse/smarthome/core/library/types/HSBType.java
@@ -166,15 +166,15 @@ public class HSBType extends PercentType implements ComplexType, State, Command 
                 | ((convertPercentToByte(rgb[1]) & 0xFF) << 8) | ((convertPercentToByte(rgb[2]) & 0xFF) << 0);
     }
 
-    public HSBType withHue(DecimalType h) {
+    public HSBType setHue(DecimalType h) {
         return new HSBType(h, getSaturation(), getBrightness());
     }
 
-    public HSBType withSaturation(PercentType s) {
+    public HSBType setSaturation(PercentType s) {
         return new HSBType(getHue(), s, getBrightness());
     }
 
-    public HSBType withBrightness(PercentType b) {
+    public HSBType setBrightness(PercentType b) {
         return new HSBType(getHue(), getSaturation(), b);
     }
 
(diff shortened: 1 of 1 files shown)


kaikreuzer:
Our types are supposed to be immutable objects, so I don't think that setters should be introduced anywhere. There is no such thing as "changing a value" - you are also not changing an Integer object by adding another one - you are simply taking another Integer object.

J-N-K:
Ok, what about .withBrightness (or similar) Methods that return new objects?

maggu2810:
How is it handled by all the others immutable objects of the JavaSE?
Just looked at the `BigDecimal` methods and there are `setScale` methods that return a new `BigDecimal` that the new scale has been assigned.

kaikreuzer:
@J-N-K "with" sounds like the builder pattern, this should not be a method of the class itself.

@maggu2810's suggestion is probably hard to argue against as he has the JDK on his side 🙄 .

Nonetheless, let me ask where you actually need that feature, because I would have expected that you can work with PercentTypes where ever you would deal with brightness values only.

J-N-K:
The DMX binding allows to control RGB devices (e.g. LED stripes). The output is RGB, but internally HSBType is used (because that is what is provided by ESH). The dimmer thing has a color channel, but a brightness channel, too.
Currently the brightness channel will totally screw up the color. It would be much easier to use the current color, adjust the brightness and then get the R,G,B components with adjusted brightness. Of course you can do something like

`currentColor = new HSBType(currentColor.getHue(), currentColor.getSaturation(), newBrightness);`

however something like

`currentColor = currentColor.setBrightness(newBrightness);`

is much more readable. However, if you think that it is better not to introduce that, I'll stick with the longer code.

kaikreuzer:
> The dimmer thing has a color channel, but a brightness channel, too.

This is actually an incorrect modelling. If a Thing has a color channel, it must not have a brightness channel at the same time (because color already provides everything that brightness would provide). Could you remove the brightness channel, please?

> However, if you think that it is better not to introduce that

I agree that it helps on your code and your implementation is now the similar to what there is in `BigDecimal`. I personally do not like it much as it seems to completely contradicts the Java bean model, where setters have a special semantic, which is clearly not applied here, which can easily lead to confusion (people might expect that the instance at hand now has the new value).

I don't veto a merge, but I would like to leave the decision to @maggu2810 & @sjka.

J-N-K:
@kaikreuzer, Is that true for color/switch and brightness/switch Channel combinations, too? Then it would be better to refactor the DMX dimmer to be three separate things (normal, color, tunable white) than to have one thing which is dynamically reconfigured.

kaikreuzer:
Yes, it is true for all those combinations. I agree, if it is technically feasible for you, having dedicated Thing types might be the nicer solution.

J-N-K:
Ok.

Issue #4751, opened 2017-12-15
NPE in addDiscoveryService at startup


mhilbush:
Build 1140

I see this consistently at the very beginning of startup.

```
2017-12-15 10:19:43.443 [INFO ] [voicerss.internal.VoiceRSSTTSService] - Using VoiceRSS cache folder /opt/openhab2/userdata/voicerss/cache
2017-12-15 10:19:44.836 [INFO ] [.core.internal.i18n.I18nProviderImpl] - Locale set to en_US, Location set to XX.263122,-YY.742618, Time zone set to America/New_York
2017-12-15 10:19:45.706 [ERROR] [g.eclipse.smarthome.config.discovery] - [org.eclipse.smarthome.config.discovery.internal.DiscoveryServiceRegistryImpl(42)] The addDiscoveryService method has thrown an exception
java.lang.NullPointerException: null
	at org.eclipse.smarthome.config.discovery.internal.DiscoveryServiceRegistryImpl.addDiscoveryService(DiscoveryServiceRegistryImpl.java:453) [104:org.eclipse.smarthome.config.discovery:0.10.0.b1]
	at sun.reflect.NativeMethodAccessorImpl.invoke0(Native Method) ~[?:?]
	at sun.reflect.NativeMethodAccessorImpl.invoke(NativeMethodAccessorImpl.java:62) ~[?:?]
	at sun.reflect.DelegatingMethodAccessorImpl.invoke(DelegatingMethodAccessorImpl.java:43) ~[?:?]
	at java.lang.reflect.Method.invoke(Method.java:498) ~[?:?]
	at org.apache.felix.scr.impl.inject.BaseMethod.invokeMethod(BaseMethod.java:229) [42:org.apache.felix.scr:2.0.12]
	at org.apache.felix.scr.impl.inject.BaseMethod.access$500(BaseMethod.java:39) [42:org.apache.felix.scr:2.0.12]
	at org.apache.felix.scr.impl.inject.BaseMethod$Resolved.invoke(BaseMethod.java:650) [42:org.apache.felix.scr:2.0.12]
	at org.apache.felix.scr.impl.inject.BaseMethod.invoke(BaseMethod.java:506) [42:org.apache.felix.scr:2.0.12]
	at org.apache.felix.scr.impl.inject.BindMethod.invoke(BindMethod.java:658) [42:org.apache.felix.scr:2.0.12]
[...]

PR #4766, opened 2017-12-17
handle discovery services if activated only

maggu2810:
This should be "theoretical" handle injected discovery service as soon as the component is activated.

@mhilbush Can you give it a try?

The previous implementation is really buggy. E.g. it clears the discovery service list on deactivate without handling the remove method. 

Fixes: https://github.com/eclipse/smarthome/issues/4751


mhilbush:
Thank you for the quick turnaround on this.  I would try it, but this error occurs only on my production box, and I'm not sure how to update my prod box with the changes from this PR.

PR #4747, opened 2017-12-15
Did little improvements to the NTP binding

AlexanderKostadinov28:
Changes include:
- `MANIFEST.MF` - added lazy activation policy (should be used with `Service-Component`).
- `NtpBindingConstants.java` - reordered the modifiers to comply with the Java Language Specification. Added a private constructor to hide the implicit public one.
- `NtpHandlerFactory.java` - little logical improvement.
- `NtpDiscovery.java` - Removed the unnecessary declaration of thrown exception 'java.lang.IllegalArgumentException' since it is a runtime exception. Removed unnecessary overrided methods which simply called the `super` method.

PR #4768, opened 2017-12-18
Did little improvements to the NTP binding (correct)

AlexanderKostadinov28:
@maggu2810 I opened this new PR since I had reverted and force-pushed the changes considering #4747, so the old PR can be deleted.

Changes include:
- `NtpBindingConstants.java` - Added a private constructor to hide the implicit public one.
- `NtpHandlerFactory.java` - little logical improvement.
- `NtpHandler.java` - got rid of simple `(String)` casting.

### DIFF
--- a/extensions/binding/org.eclipse.smarthome.binding.ntp/src/main/java/org/eclipse/smarthome/binding/ntp/handler/NtpHandler.java
+++ b/extensions/binding/org.eclipse.smarthome.binding.ntp/src/main/java/org/eclipse/smarthome/binding/ntp/handler/NtpHandler.java
@@ -117,25 +117,38 @@ public class NtpHandler extends BaseThingHandler {
             logger.debug("Initializing NTP handler for '{}'.", getThing().getUID());
 
             Configuration config = getThing().getConfiguration();
-            hostname = (String) config.get(PROPERTY_NTP_SERVER_HOST);
+            hostname = config.get(PROPERTY_NTP_SERVER_HOST).toString();
             port = (BigDecimal) config.get(PROPERTY_NTP_SERVER_PORT);
             refreshInterval = (BigDecimal) config.get(PROPERTY_REFRESH_INTERVAL);
             refreshNtp = (BigDecimal) config.get(PROPERTY_REFRESH_NTP);
             refreshNtpCount = 0;
 
             try {
-                timeZone = TimeZone.getTimeZone((String) config.get(PROPERTY_TIMEZONE));
+                String timeZoneConfigValue = config.get(PROPERTY_TIMEZONE).toString();
+                if (timeZoneConfigValue != null) {
+                    timeZone = TimeZone.getTimeZone(timeZoneConfigValue);
+                } else {
+                    timeZone = TimeZone.getDefault();
+                    logger.debug("{} using default TZ '{}', because configuration property '{}' is null.",
+                            getThing().getUID(), timeZone, PROPERTY_TIMEZONE);
+                }
             } catch (Exception e) {
                 timeZone = TimeZone.getDefault();
-                logger.debug("{} using default TZ: {}", getThing().getUID(), timeZone);
+                logger.debug("{} using default TZ '{}' due to an occurred exception: ", getThing().getUID(), timeZone, e);
             }
 
             try {
-                String localeString = (String) config.get(PROPERTY_LOCALE);
-                locale = new Locale(localeString);
+                String localeStringConfigValue = config.get(PROPERTY_LOCALE).toString();
+                if (localeStringConfigValue != null) {
+                    locale = new Locale(localeStringConfigValue);
+                } else {
+                    locale = localeProvider.getLocale();
+                    logger.debug("{} using default locale '{}', because configuration property '{}' is null.",
+                            getThing().getUID(), locale, PROPERTY_LOCALE);
+                }
             } catch (Exception e) {
                 locale = localeProvider.getLocale();
-                logger.debug("{} using default locale: {}", getThing().getUID(), locale);
+                logger.debug("{} using default locale '{}' due to an occurred exception: ", getThing().getUID(), locale, e);
             }
             dateTimeChannelUID = new ChannelUID(getThing().getUID(), CHANNEL_DATE_TIME);
             stringChannelUID = new ChannelUID(getThing().getUID(), CHANNEL_STRING);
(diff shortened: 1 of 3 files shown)


extensions/binding/org.eclipse.smarthome.binding.ntp/src/main/java/org/eclipse/smarthome/binding/ntp/internal/discovery/NtpDiscovery.java
```diff
@@ -32,14 +32,14 @@
...
- * *
  *
  * @author Marcel Verpaalen - Initial contribution
+ * @author Alexander Kostadinov - Little binding improvements
  */
 @Component(service = DiscoveryService.class, immediate = true, configurationPid = "discovery.ntp")
 public class NtpDiscovery extends AbstractDiscoveryService {
 
-    public NtpDiscovery() throws IllegalArgumentException {
+    public NtpDiscovery() {
```
maggu2810: The `super` constructor throws an IAE, so why not mark this one, that the IAE could be thrown?
Wouldn't it be helpful for the caller of the constructor to know that this runtime exception could be thrown?

extensions/binding/org.eclipse.smarthome.binding.ntp/src/main/java/org/eclipse/smarthome/binding/ntp/internal/NtpHandlerFactory.java
```diff
@@ -55,7 +56,7 @@ protected ThingHandler createHandler(Thing thing) {
 
         ThingTypeUID thingTypeUID = thing.getThingTypeUID();
 
-        if (thingTypeUID.equals(THING_TYPE_NTP)) {
+        if (THING_TYPE_NTP.equals(thingTypeUID)) {
```
maggu2810: @kaikreuzer How do we like to handle the `@author` section of our files? Should we really add a new author line for such changes that are only "style" and does not bring any new features or real bug fixes?

@AlexanderKostadinov28 Don't get me wrong here, it is not about "adding your name to the author section", but if every line change will be added to that section the section will be bigger then the remaining source code.

extensions/binding/org.eclipse.smarthome.binding.ntp/src/main/java/org/eclipse/smarthome/binding/ntp/handler/NtpHandler.java
```diff
@@ -12,7 +12,15 @@
  */
 package org.eclipse.smarthome.binding.ntp.handler;
 
-import static org.eclipse.smarthome.binding.ntp.NtpBindingConstants.*;
+import static org.eclipse.smarthome.binding.ntp.NtpBindingConstants.CHANNEL_DATE_TIME;
```
maggu2810: I don't know if we every had added a "rule" for static imports.
Other projects states e.g. as soon as the static imports is greater or equal then 3 (or 4 or 5) a wildcard should be used.
@kaikreuzer What is our guideline?

extensions/binding/org.eclipse.smarthome.binding.ntp/META-INF/MANIFEST.MF
```diff
@@ -25,3 +25,4 @@ Import-Package:
  org.eclipse.smarthome.core.types,
  org.slf4j
 Service-Component: OSGI-INF/*.xml
+Bundle-ActivationPolicy: lazy
```
maggu2810: Has this only be done, because the Eclipse IDE likes the lazy bundle activation or because there is any real use case?

Has this change been tested that the binding's life cycle is still working?
It should, but in some cases a binding could be activated (AFAIK) earlier then before.

extensions/binding/org.eclipse.smarthome.binding.ntp/src/main/java/org/eclipse/smarthome/binding/ntp/internal/discovery/NtpDiscovery.java
AlexanderKostadinov28: I removed it since I thought there is no sense explicitly marking it, but on a second thought you have a point - I will revert this change and push shortly.

extensions/binding/org.eclipse.smarthome.binding.ntp/src/main/java/org/eclipse/smarthome/binding/ntp/internal/NtpHandlerFactory.java
AlexanderKostadinov28: @maggu2810 I actually wondered and was unsure about the same thing, so I am with you on this as well - I saw contributors being marked for every change and decided to follow this. In my opinion, we should not add contributor lines for every little change in the code. :)

extensions/binding/org.eclipse.smarthome.binding.ntp/META-INF/MANIFEST.MF
AlexanderKostadinov28: Truth to be told, I followed the IDE advices and, in general, am not truly sure if this would have an effect at all. I also saw that it is defined for several of our bindings (e.g. SONOS, Tradfri, Astro, Lifx and DMX), so to me this was a good reason to follow the IDE advice. :) All in all, if we are unsure there is any sense in making this change, I can revert it straightforward.

extensions/binding/org.eclipse.smarthome.binding.ntp/src/main/java/org/eclipse/smarthome/binding/ntp/handler/NtpHandler.java
sjka: We have an implicit rule: our code-formatter 😄 

image

extensions/binding/org.eclipse.smarthome.binding.ntp/META-INF/MANIFEST.MF
sjka: Afaik the IDE's warning could be safely ignored in this case (some say they've even got it all wrong... 😉 ). See also Neil's and Peter's responses [here](https://stackoverflow.com/a/16989358).

AlexanderKostadinov28: In such case I will revert this for sure. ;)

extensions/binding/org.eclipse.smarthome.binding.ntp/src/main/java/org/eclipse/smarthome/binding/ntp/handler/NtpHandler.java
AlexanderKostadinov28: I will revert this then.

kaikreuzer:
@AlexanderKostadinov28 Did you see @sjka's comments?

AlexanderKostadinov28:
@kaikreuzer Sorry, I had missed that, I will find a minute to resolve @sjka's comments.

AlexanderKostadinov28:
@sjka I applied the requested changes, so you can have a look.

sjka:
Thanks!

PR #4770, opened 2017-12-18
Remove version tag from template binding in generated poms.

Hilbrand:
The version tag is very likely the same as the parent class and therefore can be omitted. Having the version here leads to lots of review comments in openhab2 bindings to remove it. So by not generating it in the first place this would reduce review and committer load.

P.s The test pom also contains some properties. I don't know if these are useful. I got a request to remove them in my binding test project. So can they be removed also or do they have added value? If they can be removed I can create another pull request and remove them.


sjka:
> I don't know if these are useful. 

No, they are not anymore and can (actually _should_) be removed.

> If they can be removed I can create another pull request and remove them.

Thanks, that's great!

Hilbrand:
I've removed the properties in this pr.

PR #4771, opened 2017-12-18
Added generating thing parameter sample configuration class.

Hilbrand:
To get access to thing configuration parameters a class can be used. Very often new bindings don't know this feature and use the properties class directly and implement their own parsing of the parameters. By using a configuration class this is all not needed. This change adds a sample configuration file so people directly see what is possible.

### DIFF
--- a/tools/archetype/binding/src/main/resources/archetype-resources/src/main/java/internal/__bindingIdCamelCase__Configuration.java
+++ b/tools/archetype/binding/src/main/resources/archetype-resources/src/main/java/internal/__bindingIdCamelCase__Configuration.java
@@ -1,0 +1,33 @@
+#set( $dt = $package.getClass().forName("java.util.Date").newInstance() )
+#set( $year = $dt.getYear() + 1900 )
+#if( $vendorName == "Eclipse.org/SmartHome" )
+    #set( $copyright = "Contributors to the Eclipse Foundation" )
+#else
+    #set( $copyright = "by the respective copyright holders." )
+#end
+/**
+ * Copyright (c) ${startYear},${year} ${copyright}
+ *
+ * See the NOTICE file(s) distributed with this work for additional
+ * information regarding copyright ownership.
+ *
+ * This program and the accompanying materials are made available under the
+ * terms of the Eclipse Public License 2.0 which is available at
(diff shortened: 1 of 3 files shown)


Hilbrand:
I agree it should be documented as well. I added it also to the template since I've reviewed numerous bindings and most of them were parsing parameters themselves. This suggest people don't read documentation (which is unfortunate of course and also leads to (lots of) other review issues as well). Most of the binding developers seem to be limited experienced developers, and could use all the help to get them develop bindings. In my ideal world I would think the best solution would be to have better support in the framework itself like that the initialize method simple passes the configuration as method parameter and that the configuration class is automatically generated, but that something that needs a little more thought.

maggu2810:
Would it also make sense to assign a default parameter on one example, so a missing configuration has already a fallback (does it work)?

Hilbrand:
You mean a default in the thing configuration xml or as initialization of the field? in the xml would be best I think and be sufficient. Although in this case I added `required=true`, which might imply it can never be null as that would trigger an warning before initialization. I could add a second example without required and a default included to make the distinction clear. (I might even just add all 3 possible configuration types: String, BigDecimal and Boolean)

maggu2810:
Need to have a look at the code tomorrow...
If the class member is instantiated with the default if missing, there is nothing to add (but to answer your question, I pointed to the configuration class member initialization).

kaikreuzer:
@maggu2810 Short friendly ping.

tools/archetype/binding/src/main/resources/archetype-resources/src/main/java/internal/__bindingIdCamelCase__Configuration.java
```diff
@@ -0,0 +1,33 @@
...
+
+/**
+ * The {@link ${bindingIdCamelCase}Configuration} class contains fields mapping thing configuration paramters.
+ *
+ * @author ${author} - Initial contribution
+ */
+public class ${bindingIdCamelCase}Configuration {
+
+    /**
+     * Sample configuration parameter. Replace with you own.
```
hakan42: missing "r", "with you*r* own" :)

Issue #4543, opened 2017-11-13
Handling RuntimeExceptions in scheduled tasks (in bindings)


Hilbrand:
While reviewing some bindings I found that almost everyone catches `Exception`. It's advised to not do this. However to do it correctly one should still catch `RuntimeException`. At least if the code is run by a scheduled task using the `scheduler` from the `BaseThingHandler`. Because if a `RuntimeException` is thrown inside a scheduled job that is not catched it will silently stop the scheduling and not log the error if nothing extra is done (which I haven't seen). A solution to this can be call get on the ScheduledFuture. Something like this:
```
scheduler.submit(() -> {
            try {
                refreshJob.get();
            } catch (InterruptedException e) {
                Thread.currentThread().interrupt();
            } catch (ExecutionException e) {
                logger.error("ExecutionException", e);
            }
        });
```
But then every binding has to add code like this, which is basically the same extra work to add the catch for RuntimeException in the scheduled task.

I asked a question about this on the eclipse forum: https://www.eclipse.org/forums/index.php?t=msg&th=1089703&goto=1775340&#msg_1775340 Based on the answers I've looking into this and came up with the following possible solution. This adds a wrapper around the `ScheduledThreadPoolExecutor` in `ThreadPoolManager`:
```
private static class WrappedScheduledExecutorService extends ScheduledThreadPoolExecutor {

        public WrappedScheduledExecutorService(int corePoolSize, ThreadFactory threadFactory) {
            super(corePoolSize, threadFactory);
        }

        @Override
        public ScheduledFuture<?> scheduleWithFixedDelay(Runnable command, long initialDelay, long delay,
                TimeUnit unit) {
            return super.scheduleWithFixedDelay(() -> wrap(command), initialDelay, delay, unit);
        }
       
       <some other methods to be wrapped not included here>

        private void wrap(Runnable command) {
            try {
                command.run();
            } catch (RuntimeException e) {
                logger.error("Unexpected Runtime Exception, scheduler stops: ", e);
                throw e;
            }
        }
    }
```
This wrapping is possibly needed on other places as well, but I left that out here.
My question is: Is the wrapping something that is a desired solution? Or are there other solutions/ideas that are preferred? Or would something like the `get()` call be preferred? The latter could be something done in the `BaseThingHandler` were it could keep track of scheduled tasks. That would also make it possible to handle the cancel in `dispose` automatically (because every binding seems to need to implement this at the moment).

triller-telekom:
[...]

PR #4772, opened 2017-12-18
Added wrapping of threads in ThreadPoolManager to log errors.

Hilbrand:
This change will log errors in the scheduled threads. This removed the burden of the user of the scheduler to log just for logging, but also catches possible bugs that are caused by unchecked exceptions that are not handled and will silently cause a failure.

This is related to the discussion in #4543. It only handles wrapping in ThreadPoolManager, so it might be on other places in wrapping could be done.  But the ThreadPoolManager is the most relevant for binding developers were the problems this solves are very relevant and help simplifying binding development.

### DIFF
--- a/bundles/core/org.eclipse.smarthome.core/src/main/java/org/eclipse/smarthome/core/common/ThreadPoolManager.java
+++ b/bundles/core/org.eclipse.smarthome.core/src/main/java/org/eclipse/smarthome/core/common/ThreadPoolManager.java
@@ -17,7 +17,6 @@ import java.util.Map.Entry;
 import java.util.WeakHashMap;
 import java.util.concurrent.ConcurrentHashMap;
 import java.util.concurrent.ExecutorService;
-import java.util.concurrent.Executors;
 import java.util.concurrent.ScheduledExecutorService;
 import java.util.concurrent.ScheduledThreadPoolExecutor;
 import java.util.concurrent.ThreadFactory;
@@ -25,6 +24,7 @@ import java.util.concurrent.ThreadPoolExecutor;
 import java.util.concurrent.TimeUnit;
 import java.util.concurrent.atomic.AtomicInteger;
 
+import org.eclipse.smarthome.core.internal.common.WrappedScheduledExecutorService;
 import org.osgi.service.component.annotations.Component;
 import org.slf4j.Logger;
 import org.slf4j.LoggerFactory;
@@ -114,7 +114,7 @@ public class ThreadPoolManager {
                 pool = pools.get(poolName);
                 if (pool == null) {
                     int cfg = getConfig(poolName);
-                    pool = Executors.newScheduledThreadPool(cfg, new NamedThreadFactory(poolName));
+                    pool = new WrappedScheduledExecutorService(cfg, new NamedThreadFactory(poolName));
                     ((ThreadPoolExecutor) pool).setKeepAliveTime(THREAD_TIMEOUT, TimeUnit.SECONDS);
                     ((ThreadPoolExecutor) pool).allowCoreThreadTimeOut(true);
                     pools.put(poolName, pool);
--- a/bundles/core/org.eclipse.smarthome.core/src/main/java/org/eclipse/smarthome/core/internal/common/WrappedScheduledExecutorService.java
+++ b/bundles/core/org.eclipse.smarthome.core/src/main/java/org/eclipse/smarthome/core/internal/common/WrappedScheduledExecutorService.java
@@ -1,0 +1,60 @@
+/**
+ * Copyright (c) 2014,2018 Contributors to the Eclipse Foundation
+ *
+ * See the NOTICE file(s) distributed with this work for additional
+ * information regarding copyright ownership.
+ *
+ * This program and the accompanying materials are made available under the
+ * terms of the Eclipse Public License 2.0 which is available at
+ * http://www.eclipse.org/legal/epl-2.0
+ *
+ * SPDX-License-Identifier: EPL-2.0
+ */
+package org.eclipse.smarthome.core.internal.common;
+
+import java.util.concurrent.CancellationException;
+import java.util.concurrent.ExecutionException;
+import java.util.concurrent.Future;
+import java.util.concurrent.ScheduledThreadPoolExecutor;
+import java.util.concurrent.ThreadFactory;
+
+import org.slf4j.Logger;
+import org.slf4j.LoggerFactory;
+
+/**
+ * This class wraps the ScheduledThreadPoolExecutor to implement the {@link #afterExecute(Runnable, Throwable)} method
+ * and log the exception in case the scheduled runnable threw an exception. The error will otherwise go unnoticed
+ * because an exception thrown in the runnable will simply end with no logging unless the user handles it. This
+ * wrapper removes the burden for the user to always catch errors in scheduled runnables for logging, and it also
+ * catches unchecked exceptions that can be the cause of very hard to catch bugs because no error is ever shown if the
+ * user doesn't catch the error in the runnable itself.
+ *
+ * @author Hilbrand Bouwkamp - Initial contribution
+ */
+public class WrappedScheduledExecutorService extends ScheduledThreadPoolExecutor {
+
+    final static Logger logger = LoggerFactory.getLogger(WrappedScheduledExecutorService.class);
+
+    public WrappedScheduledExecutorService(int corePoolSize, ThreadFactory threadFactory) {
+        super(corePoolSize, threadFactory);
+    }
+
+    @Override
+    protected void afterExecute(Runnable r, Throwable t) {
+        Throwable actualThrowable = t;
+        if (actualThrowable == null && r instanceof Future<?>) {
+            try {
+                ((Future<?>) r).get();
+            } catch (CancellationException ce) {
+                actualThrowable = ce;
+            } catch (ExecutionException ee) {
+                actualThrowable = ee.getCause();
+            } catch (InterruptedException ie) {
+                Thread.currentThread().interrupt();
+            }
+        }
+        if (actualThrowable != null) {
+            logger.warn("Scheduled runnable ended with an exception", actualThrowable);
+        }
+    }
+}


bundles/core/org.eclipse.smarthome.core/src/main/java/org/eclipse/smarthome/core/common/WrappedScheduledExecutorService.java
```diff
@@ -0,0 +1,60 @@
...
+import java.util.concurrent.ScheduledThreadPoolExecutor;
+import java.util.concurrent.ThreadFactory;
+
+import org.slf4j.Logger;
+import org.slf4j.LoggerFactory;
+
+/**
+ * This class wraps the ScheduledThreadPoolExecutor to implement the {@link #afterExecute(Runnable, Throwable)} method
+ * and log the error. The error will otherwise go unnoticed because on an exception thrown in the thread it will simply
+ * end with no logging or whatsoever unless the user handles it. This wrappers removes the burden by the user to always
```
maggu2810: Is there a type (whatsoever => whatsever)?

bundles/core/org.eclipse.smarthome.core/src/main/java/org/eclipse/smarthome/core/common/WrappedScheduledExecutorService.java
```diff
@@ -0,0 +1,60 @@
...
+ * catch errors in scheduled threads for logging and it also catches unhandled exceptions that can be the cause of very
+ * hard to catch bugs because no error is ever shown.
+ * The implementation follows the implementation suggestion as given in the JavaDoc of the method.
```
maggu2810: Hm, I assume this line can / should be removed.
We don't need to state that our implementation follows an implementation suggestion or that we do it correctly with respect to something.
WDYT?

bundles/core/org.eclipse.smarthome.core/src/main/java/org/eclipse/smarthome/core/common/WrappedScheduledExecutorService.java
```diff
@@ -0,0 +1,60 @@
...
+    @Override
+    protected void afterExecute(Runnable r, Throwable t) {
+        Throwable actualThrowable = t;
+        if (actualThrowable == null && r instanceof Future<?>) {
+            try {
+                ((Future<?>) r).get();
+            } catch (CancellationException ce) {
+                actualThrowable = ce;
+            } catch (ExecutionException ee) {
+                actualThrowable = ee.getCause();
```
maggu2810: What has been the reason to use the cause only? Shouldn't we keep the execution exception in the stack trace? Didn't we see "has caused by" and the cause exception in the log?

Hilbrand: The ExecutionException is thrown by the get method when the thread method was stopped on an exception. The cause is that exception. For logging I don't think it adds value to also keep te part of the ExecutionException as it doesn't given extra information. It's just a way the get method uses to pass the exception.
Related to your other comment regarding the implementation suggestion about the line to remove. The JavaDoc of the method contains the complete implementation and I just used that implementation suggestion. That's the reason I explicitly added the line (although it might not add value, so I removed it). But that implementation suggestion also uses the getCause(), hence I used it here too.

maggu2810: The question is as we are using
```
logger.error("Scheduled thread ended with an exception: ", actualThrowable);
```

does need the developer that wants to fix a bug needs to know if `actualThrowable` is the one that has been received by the function argument (`t`) or is the one that has been the cause of the `ExecutionException` of a future?

Does he "see" any difference?

Hilbrand: Actually now that I think of it the log message should probably not log to `error` but to `warn`.

bundles/core/org.eclipse.smarthome.core/src/main/java/org/eclipse/smarthome/core/common/WrappedScheduledExecutorService.java
```diff
@@ -0,0 +1,59 @@
...
+            } catch (InterruptedException ie) {
+                Thread.currentThread().interrupt();
+            }
+        }
+        if (actualThrowable != null) {
+            logger.error("Scheduled thread ended with an exception: ", actualThrowable);
```
maggu2810: I assume you should not use double point. AFAIK it depends on the logging configuration if the last non parameterized exception is logged or not (but I could be wrong).

maggu2810:
Will wait for a comment of the others committers, if they agree.

bundles/core/org.eclipse.smarthome.core/src/main/java/org/eclipse/smarthome/core/common/WrappedScheduledExecutorService.java
```diff
@@ -0,0 +1,60 @@
...
+            } catch (CancellationException ce) {
+                actualThrowable = ce;
+            } catch (ExecutionException ee) {
+                actualThrowable = ee.getCause();
+            } catch (InterruptedException ie) {
+                Thread.currentThread().interrupt();
+            }
+        }
+        if (actualThrowable != null) {
+            logger.warn("Scheduled thread ended with an exception", actualThrowable);
```
kaikreuzer: Scheduled runnable

bundles/core/org.eclipse.smarthome.core/src/main/java/org/eclipse/smarthome/core/common/WrappedScheduledExecutorService.java
```diff
@@ -0,0 +1,60 @@
...
+import java.util.concurrent.Future;
+import java.util.concurrent.ScheduledThreadPoolExecutor;
+import java.util.concurrent.ThreadFactory;
+
+import org.slf4j.Logger;
+import org.slf4j.LoggerFactory;
+
+/**
+ * This class wraps the ScheduledThreadPoolExecutor to implement the {@link #afterExecute(Runnable, Throwable)} method
+ * and log the exception in case the scheduled thread threw an exception. The error will otherwise go unnoticed because
```
kaikreuzer: scheduled thread -> scheduled runnable

bundles/core/org.eclipse.smarthome.core/src/main/java/org/eclipse/smarthome/core/common/WrappedScheduledExecutorService.java
```diff
@@ -0,0 +1,60 @@
...
+ * on an exception thrown in the thread it will simply end with no logging unless the user handles it. This wrapper
+ * removes the burden for the user to always catch errors in scheduled threads for logging, and it also catches
```
kaikreuzer: scheduled runnables

bundles/core/org.eclipse.smarthome.core/src/main/java/org/eclipse/smarthome/core/common/WrappedScheduledExecutorService.java
```diff
@@ -0,0 +1,60 @@
...
+ * unchecked exceptions that can be the cause of very hard to catch bugs because no error is ever shown if the user
+ * doesn't catch the error in the thread itself.
```
kaikreuzer: runnable

bundles/core/org.eclipse.smarthome.core/src/main/java/org/eclipse/smarthome/core/common/WrappedScheduledExecutorService.java
```diff
@@ -0,0 +1,60 @@
...
+ * See the NOTICE file(s) distributed with this work for additional
+ * information regarding copyright ownership.
+ *
+ * This program and the accompanying materials are made available under the
+ * terms of the Eclipse Public License 2.0 which is available at
+ * http://www.eclipse.org/legal/epl-2.0
+ *
+ * SPDX-License-Identifier: EPL-2.0
+ */
+package org.eclipse.smarthome.core.common;
```
kaikreuzer: Do we need/want to publicly expose this or could it be moved to an internal package?

kaikreuzer: @Hilbrand Did you see this question?

Hilbrand: I kind of assumed his was a question to the wider smarthome audience ;-)
I put it here as it might be useful in other places, although I have no specific use cases in mind. The class is already package private to limit visibility. Other options might be to make it an internal class ThreadPoolManager, but that would limit future usages outside that class. But I don't have a particular opinion in this case as my knowledge of the overall system is limited.

kaikreuzer: > I kind of assumed his was a question

No, it was a hidden order 😛 

> The class is already package private to limit visibility.

Moving it to an unexported package is a much stronger means of OSGi to restrict visibility. I'd therefore suggest to move it to ...internal.common.

If we one day have a request for reuse, we can still consider making it part of the API. Until then, it is always safer to keep the API as small as possible :-)

Hilbrand: Ok order executed :smile:

(That's what I mean by not know the whole system. I completely overlooked the internal packages, and only looked in the common package itself, and not one place higher)

Hilbrand:
Digit fixed.

kaikreuzer:
Thank you!

PR #4773, opened 2017-12-18
Adds a convenience method for scheduling jobs in things.

Hilbrand:
This pr is related to the discussion in #4543 and it is very much up for discussion and I've created this pr to get the discussion started on if the approach in this pr is the right way to go. Nevertheless it's a complete implementation and if considered a useful addition can be merged (with or within improvements of course). 

This implementation tries to improve on the following use cases:
- When a scheduled task stops on a runtime exception if it is not catched in the thread itself, the thread will simply die without any feedback. By using the scheduling method in this class it catches the error and update the binding state so the user is aware of this problem. Because if nothing is done the scheduling will simply stop without any information. (In pr #4772 an improvement for logging this added)
- Lots of bindings do the same: schedule task and remove in dispose. This implementation keeps track of scheduled tasks and disposes them. This makes binding code simpler. This does mean when bindings override dispose they should call super, which was not needed in the past. However as it is only relevant when using this new method this should not be a problem since when it's not called the super and not using this new method no harm is done.
- As scheduleWithFixedDelay is preferred over scheduleAtFixedRate by having this method available users are less likely to use the wrong scheduler.

### DIFF
--- a/bundles/core/org.eclipse.smarthome.core.thing/src/main/java/org/eclipse/smarthome/core/thing/binding/BaseThingHandler.java
+++ b/bundles/core/org.eclipse.smarthome.core.thing/src/main/java/org/eclipse/smarthome/core/thing/binding/BaseThingHandler.java
@@ -12,11 +12,15 @@
  */
 package org.eclipse.smarthome.core.thing.binding;
 
+import java.util.ArrayList;
 import java.util.HashMap;
+import java.util.List;
 import java.util.Map;
 import java.util.Map.Entry;
 import java.util.Objects;
 import java.util.concurrent.ScheduledExecutorService;
+import java.util.concurrent.ScheduledFuture;
+import java.util.concurrent.TimeUnit;
 
 import org.eclipse.jdt.annotation.NonNullByDefault;
 import org.eclipse.jdt.annotation.Nullable;
@@ -103,6 +107,7 @@ public abstract class BaseThingHandler implements ThingHandler {
     private ServiceTracker configDescriptionValidatorServiceTracker;
 
     private @Nullable ThingHandlerCallback callback;
+    private final List<ScheduledFuture<?>> scheduledFutures = new ArrayList<>();
 
     /**
      * Creates a new instance of this class for the {@link Thing}.
@@ -241,7 +246,8 @@ public abstract class BaseThingHandler implements ThingHandler {
 
     @Override
     public void dispose() {
-        // can be overridden by subclasses
+        scheduledFutures.stream().filter(sf -> !sf.isCancelled()).forEach(sf -> sf.cancel(true));
+        scheduledFutures.clear();
     }
 
     @Override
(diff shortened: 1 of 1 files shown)


triller-telekom:
I really like the approach of having the convenience method in the `BaseThingHandler`. But doesn't it supersede your PR #4772?

Also calling a method that can be overridden by the binding developer in case there is an exception is a nice feature!

Hilbrand:
pr #4772 is about the logging of the error and is implemented that is works for all schedule actions, like  `schedule`. While this pr also wraps the call it's only for a specific variant `scheduleWithFixedDelay`. Logging could also be done in this implementation, but then it's only for this specific variant. Because a repeating scheduler call is more relevant to update the state as it's a long running process I've added only this specific variant. Also I separated the pr's because this convenient method might or might not be desired, while logging should be less of a consideration.

maggu2810:
I like the idea to have an implementation to schedule tasks that execution is canceled on dispose.
I think this should be added for one time execution tasks (e.g. schedule one time execution in x seconds), too.
Should we create a ThingLifecycleScheduler for that tasks?

I am fine with a function wrapping and set the thing status to offline, but I don't think it should be a must. I would like to get also the option to add a function to the scheduler that is not wrapped again. Perhaps my function already uses a correct implementation. The additional wrap should be an option that could be opt-in or opt-out.

Hilbrand:
The `ThreadPoolManager` is not generic, not limited to things. Having a special ThingLifecycleScheduler would add some complexity to `ThreadPoolManager`. I'm not sure it that worth it.

You could still use the protected field `scheduler` directly, you would only miss out on the automatic dispose. I'm not that familiar with all use cases, so I don't know if adding an option to wrap will be useful or if it only is a feature to have the option just-in-case.

maggu2810:
We are not talking about `ThreadPoolManager` in this PR, or do I miss some point?
I think about a wrapper for the `scheduler` member that provides the methods of this one but cancel the futures on dispose...

Hilbrand:
I'm not sure what you mean? I was referring to that the field `scheduler` that is initialized with via `ThreadPoolManager`. Do you mean to put all code, including the Set to keep track of the futures in a separate class so `BaseThingHandler` doesn't get all the code?

maggu2810:
I will create some demo code in separate branch later. Then it is more easy to talk about and perhaps I think abou something that is not possible / good at all :wink:

maggu2810:
What I have in mind has been something similar to this one:
https://github.com/eclipse/smarthome/compare/master...maggu2810:thing-lifecycle-scheduler?expand=1

The scheduling method does not return a scheduled future to the caller as the future itself is managed by the thing lifecycle (but if desired, we could also return the future because of the cleanup code).

The thing lifecylce scheduler if used needs to be activated in initialize and deactivated in the dispose method.

And nothing speaks against adding a utility method e.g. for
```Java
        lifecycleScheduler.scheduleWithFixedDelay(() -> {
            try {
                // do some magic;
            } catch (RuntimeException e) {
                updateStatus(ThingStatus.OFFLINE, ThingStatusDetail.NONE, e.getMessage());
                throw e;
            }
        }, initialDelay, delay, unit);
```

WDYT?

Hilbrand:
I'm not sure I understand what problem you are trying to solve. My change was primarily to have a wrapper that updates the state on an error and/or let the binding know the scheduled job was stopped and secondly collect these scheduled tasks and clean them up. You seem to have turned this around.
Also I don't see why adding activate/deactivate adds any value. It only adds more complexity for the binding developer. Because if no tasks were scheduled and an implicit call to cleanup in the dispose (or in your code an implicit call to deactive without the check) would be made it will simply will iterate over an empty list, which is no issue at all. If the developer doesn't want an implicit clean up simple override dispose and not call super, but then the developer has to handle it self, in which case it's easier to just user the scheduler field directly. So why the explicit reason to activate?

maggu2810:
Hi @Hilbrand,
the intention of my changes has been to give you an idea what I am talking about.

> I will create some demo code in separate branch later. Then it is more easy to talk about

So, not a complete solution already.

IMHO your changes contains two things:
* add an utility method to schedule a job and set the thing to offline if something went wrong
* discard / cancel scheduled jobs if the thing handler has been disposed

I would like to add a generic support for the second part of your changes first as (again IMHO) this is a fundamental part that is used in lot of bindings.

Hilbrand:
Ok, I see what you mean. When clicking the link you provided it presents the page to create a new pull request, which gave the assumption it was more ready then you intended.

I'm not sure if a separate class would be to much separation especially if adding the catching of the error would be added, which would still required additional convenience methods or circular references to the separate class. And would then possible provide 3 different ways to schedule tasks: 1) the current way via scheduler, 2) via the new scheduler with automatic dispose and 3) with scheduler that wraps to catch the error.
IMHO my first priority here was to let the user know a scheduled task crashed and make that easy for the developer to add. For simplicity if (implicit) removal of tasks should be handled differently I can remove that part from this pr so it can be implemented in a separate change. WDYT?

maggu2810:
I still don't know if we should add your two functions to the base thing handler.
The function itself makes really sense to get some feedback, I just don't know how "big" that base class should be.
So perhaps e.g. @sjka or @kaikreuzer could comment.

I also think we should not mix the two points (wrt my previous comment).
Discard scheduled jobs on dispose and the offline mark are IMHO two different aspects.

kaikreuzer:
I am also not having a clear opinion here - I also see the risk that the base handler is growing too large if we add stuff to it. Nonetheless, we should provide means to make common tasks as easy and error-free as possible.

I agree with @maggu2810 that the problematic part is the status update in here. From my experience, there are many different ways you would want to handle it, e.g. especially setting a certain sub-status, so many developers would have to overwrite the scheduledThreadAborted method. Then again, when handling the status correctly, they would not want the exception to be further thrown and logged by #4772, so they might end up not using the scheduleWithFixedDelay method at all.

I therefore think that #4772 is the important part - anything that a developer has missed to catch/handle in his runnables is now logged. Apart from this, there is not really a big benefit through these new methods in this PR here (and explicitly remembering what one has scheduled isn't such a bad thing either as it can prevent unreflected scheduling of jobs).

Hilbrand:
To summarize. I think it looks this was a bit of a stretch and doesn't cover enough use cases to have this added to the base  class. Having additions to the scheduler management does look to be something useful as mentioned but needs more thought and a different solution as implemented in this pr. 

For developers of scheduled tasks I would suggest that the RuntimeExceptions should be catched to update the state of the binding as to let the user know something is wrong without the user having to look into the log files after it looks the binding is not working anymore. Possible this can be added to the binding implementation documentation.

I'll therefor close my pr.

PR #4783, opened 2017-12-19
Default import of static Collectors members

sjka:
On the aftermath of having of having worked with Java streams a bit, IMHO especially the `Collectors.*` is superfluous and makes it hard to read - thanks @htreu for pointing this out! Therefore, this allows easy typing of

```java
return stream().collect(toList());
```

instead of

```java
return stream().collect(Collectors.toList());
```

and then needing to fight the automatic import organizer.


kaikreuzer:
May I ask you to create a similar PR for the OH2 IDE?

sjka:
Sure: https://github.com/openhab/openhab2-addons/pull/2996

PR #4784, opened 2017-12-19
Remove Guavas ImmutableSet from GroupItem

triller-telekom:
Inspired by #4743 this is a first attempt to remove the left overs.

### DIFF
--- a/bundles/core/org.eclipse.smarthome.core/src/main/java/org/eclipse/smarthome/core/items/GroupItem.java
+++ b/bundles/core/org.eclipse.smarthome.core/src/main/java/org/eclipse/smarthome/core/items/GroupItem.java
@@ -30,7 +30,6 @@ import org.slf4j.Logger;
 import org.slf4j.LoggerFactory;
 
 import com.google.common.base.Joiner;
-import com.google.common.collect.ImmutableSet;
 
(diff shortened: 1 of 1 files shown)


bundles/core/org.eclipse.smarthome.core/src/main/java/org/eclipse/smarthome/core/items/GroupItem.java
```diff
@@ -123,7 +122,7 @@ public GroupFunction getFunction() {
      * @return all members of this and all contained {@link GroupItem}s
      */
     public Set<Item> getAllMembers() {
-        return ImmutableSet.copyOf(getMembers((Item i) -> !(i instanceof GroupItem)));
+        return Collections.unmodifiableSet(getMembers((Item i) -> !(i instanceof GroupItem)));
```
htreu: to respect the former `copyOf`, should this also be wrapped in `new HashSet(...)`?

triller-telekom: Done, and to also make travis happy I used `LinkedHashSet` to preserve the item order.

PR #4785, opened 2017-12-19
Cache sitemap models in SitemapProviderImpl

### LINKED ISSUE #4716: Starred interaction on config and rules changes
Whenever I change an items or rules file, to process this update in OH can take up to 2 or 3 minutes (for a 1k lines rules file on a Pi2). java is runnning at 100% CPU (i.e. 1 core@100%, but 3 idle cores).
Meanwhile, the interaction via REST or Web UI is apparently HALTED until the current file is processed.
As requested by @kaikreuzer, here's a large rules file to help you reproduce it, but it happens on _any_ rule und items file (re)load: it takes the API and UI to react exactly as long as it takes to process the updated file, UI comes back to life as soon as the 'refreshing xxx.rules' log entry shows up.

[reloadtest.TXT](https://github.com/eclipse/smarthome/files/1547661/reloadtest.TXT)

htreu:
This fixes #4716 by providing a local sitemap model cache.

Also: replace service XML by DS annotations.


bundles/model/org.eclipse.smarthome.model.sitemap/src/org/eclipse/smarthome/model/sitemap/internal/SitemapProviderImpl.java
```diff
@@ -28,55 +37,77 @@
...
+    private static final String SITEMAP_MODEL_NAME = "sitemap";
+    protected static final String SITEMAP_FILEEXT = "." + SITEMAP_MODEL_NAME;
 
     private final Logger logger = LoggerFactory.getLogger(SitemapProviderImpl.class);
 
     private ModelRepository modelRepo = null;
 
+    private final Map<String, Sitemap> sitemapModelCache = new ConcurrentHashMap<>();
+
+    @Reference(cardinality = ReferenceCardinality.MANDATORY, policy = ReferencePolicy.STATIC)
```
triller-telekom: Isn't this the default anyway and we agreed on not putting default values here? So I guess `@Reference` should be the same, but this is a minor issue.

bundles/model/org.eclipse.smarthome.model.sitemap/src/org/eclipse/smarthome/model/sitemap/internal/SitemapProviderImpl.java
```diff
@@ -28,55 +37,77 @@
...
+        refreshSitemapModels();
     }
 
     public void unsetModelRepository(ModelRepository modelRepo) {
+        modelRepo.removeModelRepositoryChangeListener(this);
         this.modelRepo = null;
     }
 
+    @Deactivate
+    protected void deactivate() {
```
kaikreuzer: Would be nice to clear the sitemapModelCache here as well.

bundles/model/org.eclipse.smarthome.model.sitemap/src/org/eclipse/smarthome/model/sitemap/internal/SitemapProviderImpl.java
```diff
@@ -28,55 +37,77 @@
...
     private final Logger logger = LoggerFactory.getLogger(SitemapProviderImpl.class);
 
     private ModelRepository modelRepo = null;
 
+    private final Map<String, Sitemap> sitemapModelCache = new ConcurrentHashMap<>();
+
+    @Reference(cardinality = ReferenceCardinality.MANDATORY, policy = ReferencePolicy.STATIC)
     public void setModelRepository(ModelRepository modelRepo) {
         this.modelRepo = modelRepo;
+        modelRepo.addModelRepositoryChangeListener(this);
```
kaikreuzer: This can potentially lead to callbacks even before this component is activated.
Even though we do not care today about the activation state in this component, I would prefer to add the listener in the activate method (to be created) only.

bundles/model/org.eclipse.smarthome.model.sitemap/src/org/eclipse/smarthome/model/sitemap/internal/SitemapProviderImpl.java
```diff
@@ -28,55 +37,77 @@
...
+        refreshSitemapModels();
     }
 
     public void unsetModelRepository(ModelRepository modelRepo) {
+        modelRepo.removeModelRepositoryChangeListener(this);
```
kaikreuzer: Should not be necessary as you remove it in deactivate already.

bundles/model/org.eclipse.smarthome.model.sitemap/src/org/eclipse/smarthome/model/sitemap/internal/SitemapProviderImpl.java
```diff
@@ -28,55 +36,81 @@
...
     private final Logger logger = LoggerFactory.getLogger(SitemapProviderImpl.class);
 
     private ModelRepository modelRepo = null;
 
+    private final Map<String, Sitemap> sitemapModelCache = new ConcurrentHashMap<>();
+
+    @Reference
     public void setModelRepository(ModelRepository modelRepo) {
         this.modelRepo = modelRepo;
+        refreshSitemapModels();
```
kaikreuzer: Here again, it would be much nicer to do this in activate() when the component is really meant to become active.

PR #4786, opened 2017-12-19
Updated documentation on min/max/options in XML file

cdjackson:
This updates the documentation around the use of min/max values, and options. Since options can be outside the min/max range this needs to be made clear and this is the main purpose of this PR.

One issue I noted though is that min/max is stated as also being used to select the min/max number of values that may be selected in a list. I've removed this from here as I thought this was wrong, but maybe it has also been implemented? If so, this conflicts with the ```multipleLimit``` setting which sets the maximum number of selections.

```
multipleLimit | If multiple is true, sets the maximum number of options that can be selected (optional).
```

Given that a list could contain numbers, doesn't using the min/max as the number of options that can be selected actually conflict with preventing the user entering multiple values that are outside a specific range? With the definition in the documentation at least, we can't allow the user to enter a set of numbers between a certain range - I would suggest that maybe there should be a ```multipleLimitMin``` (or something similar) to go with the ```multipleLimit``` setting.

### DIFF
--- a/docs/documentation/development/bindings/xml-reference.md
+++ b/docs/documentation/development/bindings/xml-reference.md
@@ -77,8 +77,8 @@ The following HTML tags are allowed -: ```<b>, <br>, <em>, <h1>, <h2>, <h3>, <h4
   <tr><td>parameter</td><td>The description of a concrete configuration parameter (optional).</td></tr>
   <tr><td>parameter.name</td><td>The name of the configuration parameter (mandatory).</td></tr>
   <tr><td>parameter.type</td><td>The data type of the configuration parameter (mandatory).</td></tr>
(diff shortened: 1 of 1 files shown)


kaikreuzer:
> Since options can be outside the min/max range

Why? This is actually pretty unexpected as I would consider min/max as some "real" constraints of the value.

> this conflicts with the multipleLimit setting which sets the maximum number of selections

The documentation on min/max existed a long time before you had introduced multipleLimit - we seem to have missed this contradiction at that time. I don't know if it is anywhere checked in the code.

cdjackson:
Some time ago we agreed that options can fall outside of the min max so this only documents the current situation as already implemented in ESH. 

As you say, the fact that options can fall outside the min/max is not related to the multiple limit - this is another issue and maybe I shouldn't have mixed this up.

cdjackson:
I will close this and open separate PRs for the different issues to make this clearer.

PR #4787, opened 2017-12-19
Dropped ServiceBinder

sjsf:
Continuing my spring cleaning frenzy, I took my chances and removed some ugly heritage called `ServiceBinder`. While certainly invented once with all the best intentions, it turned out to be pretty much of a beast, circumventing DS and therefore causing all sorts of funny things when starting bundles in the wrong order.

Therefore this PR does all that was needed in order to get rid of it:

* Dropped `ServiceBinder`
* Dropped the related `Activator`s
* Dropped `TypeResolver` in favor of using the corresponding registry-classes directly
* Turned `ConfigDescriptionValidator` into an OSGi service

There os a down-side of course:
* Additional `ServiceTracker`s in `BaseThingHandler` (only continuing the already existing pattern - general cleanup is needed there, but out-of-scope here)
* Some very few occurrences of `FrameworkUtil.getBundle(this.getClass())` were needed in the remaining static helper methods - IMHO acceptable as we should get rid of these as well (out-of-scope here) and it's not any uglier than what we had before

I fear this change might not binary-compatible, therefore labeling it as APIBreaking. Haven't confirmed it yet though.

The only usage of `TypeResolver` within bindings I could find was within the Yamaha binding (which introduced a cycle anyway - PR to fix this here: https://github.com/openhab/openhab2-addons/pull/2997).


bundles/config/org.eclipse.smarthome.config.core/src/main/java/org/eclipse/smarthome/config/core/validation/ConfigValidationException.java
```diff
@@ -50,11 +53,12 @@
      *
      * @throws NullPointException if given bundle or configuration description validation messages are null
      */
-    public ConfigValidationException(Bundle bundle, Collection<ConfigValidationMessage> configValidationMessages) {
-        Preconditions.checkNotNull(bundle, "Bundle must not be null");
+    public ConfigValidationException(TranslationProvider translationProvider,
```
maggu2810: This class is instantiated by tests and ConfigDescriptionValidatorImpl.
ConfigDescriptionValidatorImpl itself is a component and so it can access the bundle without using `FrameworkUtil`.
We could keep the bundle as an argument and there is no use to introduce the usage of `FrameworkUtil` here just to drop it later.

bundles/core/org.eclipse.smarthome.core.thing/src/main/java/org/eclipse/smarthome/core/thing/type/ThingType.java
```diff
@@ -286,6 +302,10 @@ public String getBindingId() {
                     }
                 }
             }
+
+            if (ref != null) {
+                bundleContext.ungetService(ref);
```
maggu2810: Shouldn't we use a final block to ensure the service reference is released?

maggu2810:
I agree with this changes (as soon as Travis CI does).

We should add the point "remove the usage of ServiceTracker" to an agenda. :wink:

bundles/config/org.eclipse.smarthome.config.core/src/main/java/org/eclipse/smarthome/config/core/validation/ConfigValidationException.java
sjka: Good point, thanks!

bundles/core/org.eclipse.smarthome.core.thing/src/main/java/org/eclipse/smarthome/core/thing/type/ThingType.java
sjka: Yes

sjka:
> We should add the point "remove the usage of ServiceTracker" to an agenda. 😉

Done: #4790

maggu2810:
Great work

martinvw:
@sjka are you aware that this broke some things on our side, what is the the correct fix for this:

https://github.com/openhab/openhab2-addons/search?utf8=%E2%9C%93&q=TypeResolver&type=

jwveldhuis:
@sjka I used TypeResolver to add Channels dynamically based on the Thing's model capabilities.
https://github.com/jwveldhuis/openhab2-addons/blob/denon-marantz-binding/addons/binding/org.openhab.binding.denonmarantz/src/main/java/org/openhab/binding/denonmarantz/handler/DenonMarantzHandler.java#L153
What is the new correct way of doing this?

sjka:
Yes, that would _currently_ be the way of accessing the ChannelType information. However, I admit that this is quite some bloat in your case. Sure you don't want to set up this whole enchilada just to find out the itemType which is basically coming from the static XML within _your_ binding. 

Also, moving away from the XMLs completely just for this very reason will again create a lot of overhead. I'd suggest putting a simple, static mapping somewhere only for the channelType-to-itemType mapping. I mean, you ARE the binding, so I think it's at least acceptable to do so.

However, despite the fact that duplicating this bit of information might be the least painful solution currently, I think this absolutely sucks. You don't _want_ to do all that by hand, the framework should support something like "please create me a channel of the given type"... There is #4950 to address this purpose 😄

kreutpet:
hi , i am again working on the siebel eltron binding and need to find a solution to access ChannelGroupType propeerties and channelTypes in the thingTypeHandler.
How can i achive that without TypeResolver?
         
```
            List<ChannelGroupDefinition> channelGroups = thingType.getChannelGroupDefinitions();

            for (ChannelGroupDefinition channelGroup : channelGroups) {

                ChannelGroupTypeUID ChannelGroupTypeUID = channelGroup.getTypeUID();
                ChannelGroupType channelGroupType = TypeResolver.resolve(ChannelGroupTypeUID);

                List<ChannelDefinition> channelDefinitions = channelGroupType.getChannelDefinitions();
                String requestDescription = channelGroupType.getDescription();
                String requestName = channelGroup.getId();

                for (ChannelDefinition channelDefinition : channelDefinitions) {

                    if (!channelDefinition.getProperties().containsKey("requestByte")) {
                        // only iterate on channels with request property
                        continue;
                    }

                    RecordDefinition record = new RecordDefinition();
                    DataParser parser = new DataParser();

                    byte requestByte = parser
                            .shortToByte(Short.decode(channelDefinition.getProperties().get("requestByte")))[0];

                    ChannelTypeUID channelTypeUID = channelDefinition.getChannelTypeUID();

                    try {
                        ChannelType channelType = TypeResolver.resolve(channelTypeUID);
                        StateDescription state = channelType.getState();


```
i am defining different thingtypes based on same channelgrougtypes.
```
	<channel-group-type id="channelGroupTypeVersion" >
	    <label>Version</label>
	    <description>Version information of heat pump firmware</description>
	    <channels>
	    <!-- at least 1 channel-group with id = version must be included  -->
	        <channel id="version" typeId="version" >
                <properties>
                    <property name="dataType">Status</property>
                    <property name="requestByte">0xfd</property>
                    <property name="position">4</property>
                    <property name="length">2</property>
                    <property name="scale">0.01</property>
                </properties>
            </channel>
            <channel id="dumpResponse" typeId="dumpResponse"/>
	    </channels>
	</channel-group-type>

```
in the channegroups i do have channeltype referenced 

```
    <channel-type id="version">
        <item-type>String</item-type>
        <label>Version</label>
        <description>Firmware version of heat pump.</description>
        <state readOnly="true" />
    </channel-type>

```
how do i get access to the properties of the channelgroupttype?

PR #4788, opened 2017-12-19
Cancelled Firmware Update Does not Wait for the Update to Finish

dstivanov:
The firmware update service operations 'update' and 'cancel' should not be executed with the same identifier for the context of the async safe caller. When the identifier is not specified, the default identifier is the target object (in this case the handler) which causes the 'cancel' operation to be executed when the 'update' operation is finished. 
Implemented unit tests to reproduce the problem, which are passing with the provided fix.

### DIFF
--- a/bundles/core/org.eclipse.smarthome.core.thing.test/src/test/java/org/eclipse/smarthome/core/thing/firmware/FirmwareUpdateServiceTest.java
+++ b/bundles/core/org.eclipse.smarthome.core.thing.test/src/test/java/org/eclipse/smarthome/core/thing/firmware/FirmwareUpdateServiceTest.java
@@ -67,6 +67,7 @@ import org.osgi.framework.Bundle;
  *
  * @author Thomas Höfer - Initial contribution
  * @author Simon Kaufmann - converted to standalone Java tests
+ * @author Dimitar Ivanov - added a test for valid cancel execution during firmware update
  */
 public class FirmwareUpdateServiceTest extends JavaOSGiTest {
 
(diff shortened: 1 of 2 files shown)


bundles/core/org.eclipse.smarthome.core.thing.test/src/test/java/org/eclipse/smarthome/core/thing/firmware/FirmwareUpdateServiceTest.java
```diff
@@ -318,6 +319,39 @@ public void testCancelFirmwareUpdate() {
...
+        firmwareUpdateService.updateFirmware(THING1_UID, FW112_EN.getUID(), null);
+        firmwareUpdateService.cancelFirmwareUpdate(THING1_UID);
+        
+        // Be sure that the cancel is executed before the completion of the update
+        waitForAssert(() -> {
+            verify(handler1, times(1)).cancel();
+        },stepsTime * numberOfSteps,50);
+        
+        // Check that the update is not finished even if the cancel has been called within the timeout
+        assertThat("Cancel should stop long running updates", isUpdateFinished.get(), is(false));
```
sjka: Please drop the explanation string here - it makes the tests harder to read because it states exactly the opposite of what your are asserting here and hence completely neglects the "fluent" API of the hamcrest matchers.

bundles/core/org.eclipse.smarthome.core.thing.test/src/test/java/org/eclipse/smarthome/core/thing/firmware/FirmwareUpdateServiceTest.java
```diff
@@ -318,6 +319,39 @@ public void testCancelFirmwareUpdate() {
...
+        final long stepsTime = 200;
+        final int numberOfSteps = SEQUENCE.length;
+        final AtomicBoolean isUpdateFinished = new AtomicBoolean(false);
+        
+        doAnswer(invocation -> {
+            ProgressCallback progressCallback = (ProgressCallback) invocation.getArguments()[1];
+            progressCallback.defineSequence(SEQUENCE);
+            
+            // Simulate update steps with delay
+            for(int updateStepsCount=0; updateStepsCount < numberOfSteps; updateStepsCount++) {
```
sjka: There are some spaces missing. Please use the formatter of the Eclipse SmartHome IDE.

bundles/core/org.eclipse.smarthome.core.thing.test/src/test/java/org/eclipse/smarthome/core/thing/firmware/FirmwareUpdateServiceTest.java
```diff
@@ -318,6 +319,39 @@ public void testCancelFirmwareUpdate() {
...
+        }).when(handler1).updateFirmware(any(Firmware.class), any(ProgressCallback.class));
+        
+        // Execute update and cancel it immediately
+        firmwareUpdateService.updateFirmware(THING1_UID, FW112_EN.getUID(), null);
+        firmwareUpdateService.cancelFirmwareUpdate(THING1_UID);
+        
+        // Be sure that the cancel is executed before the completion of the update
+        waitForAssert(() -> {
+            verify(handler1, times(1)).cancel();
+        },stepsTime * numberOfSteps,50);
```
sjka: You could actually limit the waiting time to something lower here - in the end you want to make sure the cancel is executed more or less immediately, right? (Just give the CI servers enough time to mine enough bitcoins or whatever else keeps them so busy all the time 😉 )

bundles/core/org.eclipse.smarthome.core.thing.test/src/test/java/org/eclipse/smarthome/core/thing/firmware/FirmwareUpdateServiceTest.java
dstivanov: Thanks, I will apply the formatter :)

bundles/core/org.eclipse.smarthome.core.thing.test/src/test/java/org/eclipse/smarthome/core/thing/firmware/FirmwareUpdateServiceTest.java
dstivanov: OK, with 10ms it fails well with the previous identifier, so I will leave it like this :) Thanks for the suggestion!

bundles/core/org.eclipse.smarthome.core.thing.test/src/test/java/org/eclipse/smarthome/core/thing/firmware/FirmwareUpdateServiceTest.java
dstivanov: Okay, got it - no explanations :) We read code :) :)

sjka:
Thanks!

PR #4791, opened 2017-12-20
Corrected null annotations in binding builders

triller-telekom:

### DIFF
--- a/bundles/core/org.eclipse.smarthome.core.thing/src/main/java/org/eclipse/smarthome/core/thing/binding/builder/ChannelBuilder.java
+++ b/bundles/core/org.eclipse.smarthome.core.thing/src/main/java/org/eclipse/smarthome/core/thing/binding/builder/ChannelBuilder.java
@@ -37,7 +37,7 @@ public class ChannelBuilder {
 
     private final ChannelUID channelUID;
     private @Nullable final String acceptedItemType;
-    private @Nullable ChannelKind kind;
+    private ChannelKind kind;
     private @Nullable Configuration configuration;
     private Set<String> defaultTags;
     private @Nullable Map<String, String> properties;
@@ -139,12 +139,12 @@ public class ChannelBuilder {
      * @param kind kind.
      * @return channel builder
      */
-    public ChannelBuilder withKind(@Nullable ChannelKind kind) {
+    public ChannelBuilder withKind(ChannelKind kind) {
         if (kind == null) {
-            this.kind = ChannelKind.STATE;
-        } else {
-            this.kind = kind;
+            throw new IllegalArgumentException("kind must not be null");
         }
+
+        this.kind = kind;
         return this;
     }
 
--- a/bundles/core/org.eclipse.smarthome.core.thing/src/main/java/org/eclipse/smarthome/core/thing/binding/builder/ChannelBuilder.java
+++ b/bundles/core/org.eclipse.smarthome.core.thing/src/main/java/org/eclipse/smarthome/core/thing/binding/builder/ChannelBuilder.java
@@ -37,7 +37,7 @@ public class ChannelBuilder {
 
     private final ChannelUID channelUID;
     private @Nullable final String acceptedItemType;
-    private ChannelKind kind;
+    private @Nullable ChannelKind kind;
     private @Nullable Configuration configuration;
     private Set<String> defaultTags;
     private @Nullable Map<String, String> properties;
@@ -139,7 +139,12 @@ public class ChannelBuilder {
      * @param kind kind.
      * @return channel builder
      */
-    public ChannelBuilder withKind(ChannelKind kind) {
+    public ChannelBuilder withKind(@Nullable ChannelKind kind) {
+        if (kind == null) {
+            // this code is used by 3rd party developers and thus we cannot use nonnull annotation because we have to
+            // protect ourselves against {@code null} anyway
+            throw new IllegalArgumentException("kind must not be null");
+        }
         this.kind = kind;
         return this;
     }
--- a/bundles/core/org.eclipse.smarthome.core.thing/src/main/java/org/eclipse/smarthome/core/thing/binding/builder/ChannelBuilder.java
+++ b/bundles/core/org.eclipse.smarthome.core.thing/src/main/java/org/eclipse/smarthome/core/thing/binding/builder/ChannelBuilder.java
@@ -35,8 +35,8 @@ import org.eclipse.smarthome.core.thing.type.ChannelTypeUID;
 @NonNullByDefault
 public class ChannelBuilder {
 
-    private ChannelUID channelUID;
-    private @Nullable String acceptedItemType;
+    private final ChannelUID channelUID;
+    private @Nullable final String acceptedItemType;
     private ChannelKind kind;
     private @Nullable Configuration configuration;
     private Set<String> defaultTags;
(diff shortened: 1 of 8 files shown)


bundles/core/org.eclipse.smarthome.core.thing/src/main/java/org/eclipse/smarthome/core/thing/binding/builder/ChannelBuilder.java
```diff
@@ -140,10 +140,6 @@ public ChannelBuilder withDefaultTags(Set<String> defaultTags) {
      * @return channel builder
      */
     public ChannelBuilder withKind(ChannelKind kind) {
-        if (kind == null) {
```
triller-telekom: I am wondering if there was a reason why @sjka has not removed this check after he added the `@NonNullbyDefault` annotation to the class. I have removed the check for now, but if there was a good reason to keep it we should think about allowing `@Nullable` for the kind and keep the check.

maggu2810: I don't know what has been @sjka intentions and I agree that this could be removed...
... but didn't we have a discussion in another topic where you / Kai argued that we cannot rely on correct usage WRT to the nullness annotations for code that is called by third party ones (e.g. binding developers etc.)?
So, if we don't trust non ESH framework developers that he uses annotations classes correctly, isn't that exactly what you want?

triller-telekom: I am still struggling with which parts of the code are internal for the framework and which is used by others. I mean `public` doesn't mean its really exported by the manifest, so I would have to look it up to make that decision since I don't know the code by heart as you guys :)

So if this method is really "public public", I think we should add a `@Nullable` on the parameter and keep the check.

maggu2810:
> I think we should add a `@Nullable` on the parameter and keep the check.

Hm, if the intention is that `kind` is not null the annotation should be kept.
If we don't trust the caller, we should kept the check, too.
But just because someone does not follow the null annotations, there is no need to mark that this value could be nullable and then throwing an excpetion.
It is similar as you state "you can use null as an argument but I will throw an exception if you do it."

Normally we export every package that does not contain the "internal" wording.

triller-telekom:
So you say that I should keep, i.e. **add** in this case the `@Nullable` and also keep the check?

But I should remove the throwing of the exception? What should I do instead? Set the `kind` to `ChannelKind.STATE` which is the default anyway?

> Normally we export every package that does not contain the "internal" wording.

Thanks, but from the "normally" I read there might be some "exceptions" ;-)

maggu2810:
> Thanks, but from the "normally" I read there might be some "exceptions" ;-)

It shouldn't be any but it could some...
The question is which exported packages are to be used by third party developers and which one has only to be used by other core bundles.

No, you should keep the argument marked non-null (by set a default using `@NonNullByDefault` or explicit by using `@NonNull`). As this is what you also want to document by your annotation, the argument must not be null.
The additional check depends on your (I don't remember but IIRC it has been you and kai) decision if we trust the third party developer to follow the nullness annotations (use it correctly, so we can remove the additional check) or we don't trust the developers that he respect that annotatoins (and kept the additional check as a safe guardian).

triller-telekom:
> As this is what you also want to document by your annotation, the argument must not be null.

But this is to be used by 3rd party developers so we cannot guarantee that, right?

> The additional check depends on your (I don't remember but IIRC it has been you and kai) decision if we trust the third party developer to follow the nullness annotations (use it correctly, so we can remove the additional check) or we don't trust the developers that he respect that annotatoins (and kept the additional check as a safe guardian).

If I set it to `@NonNull` the Eclipse IDE tooling will mark the `null` check as unnecessary because the annotation already states that it is not `null`. So we are back to our discussion what we do with such warnings... And I guess we said that if it can be `null` because 3rd party developers use it, i.e. the value comes from outside of our framework, we accept `null` and thus mark it as `@Nullable` and also keep the check.

maggu2810:
The use case of nullness annotations are to state if something is allowed to be null or is not allowed to be null. If you mark it as `@Nullable` then you state, null is okay to be used as an argument. So, when you throw an exception because someone uses a allowed argument, then there is something completely wrong.

triller-telekom:
How about doing it like in my last commit?

```
public ChannelBuilder withKind(@Nullable ChannelKind kind) {
        if (kind == null) {
            this.kind = ChannelKind.STATE;
        } else {
            this.kind = kind;
        }
        return this;
    }
```

maggu2810:
That would be an option and IMHO at least correct behavior (WRT to annotations and its usage).
My personal preference would be to state that there must be a kind given, but don't let start the whole topic again...

triller-telekom:
I have reverted the changes in `ChannelBuilder` and left the warning in there which I tried to remove with my original commit. At least now there is the implicit `@NonNull` on the `kind` as you suggested.

bundles/core/org.eclipse.smarthome.core.thing/src/main/java/org/eclipse/smarthome/core/thing/binding/builder/ThingStatusInfoBuilder.java
```diff
@@ -82,8 +84,7 @@ private ThingStatusInfoBuilder(ThingStatus status, ThingStatusDetail statusDetai
      *
      * @throws IllegalArgumentException if thing status detail is null
      */
-    public @NonNull ThingStatusInfoBuilder withStatusDetail(@NonNull ThingStatusDetail statusDetail)
-            throws IllegalArgumentException {
+    public ThingStatusInfoBuilder withStatusDetail(ThingStatusDetail statusDetail) throws IllegalArgumentException {
```
kaikreuzer: Not related to your changes, but do you happen to know why there is a `throws IllegalArgumentException`?

bundles/model/org.eclipse.smarthome.model.thing/src/org/eclipse/smarthome/model/thing/internal/GenericThingProvider.xtend
```diff
@@ -209,7 +209,9 @@ class GenericThingProvider extends AbstractProvider<Thing> implements ThingProvi
 
         thingBuilder.withConfiguration(configuration)
         thingBuilder.withBridge(bridgeUID)
-        thingBuilder.withLabel(label)
+        if(label !== null) {
```
kaikreuzer: Why is location nullable, while label isn't?
Shouldn't we treat the Builders as consumer API and thus accept null values as parameter unless it causes real issues (as e.g. in the create methods)?

bundles/core/org.eclipse.smarthome.core.thing/src/main/java/org/eclipse/smarthome/core/thing/binding/builder/ThingStatusInfoBuilder.java
triller-telekom: Good catch. These were leftovers from before our `null` annotations, i.e. from before the first `null` annotations were added to the file. When they were introduced the `throw IAE` in the body of the methods was removed but not in its header. I have corrected it in this PR now.

PR #4792, opened 2017-12-20
Corrected null annotations in Channel and ItemFactory

triller-telekom:


maggu2810:
I restarted Travis CI because it's failing now.

triller-telekom:
@maggu2810 it was failing because I introduced a bug in the `equals` method :( I have fixed it and now Travis should succeed.

PR #4793, opened 2017-12-21
Method registerService of OSGiTest.groovy To Return Service Registrations

dstivanov:
The registerService methods of the base groovy class OSGiTest are now returning the service registration. Null assertions added into the methods as well as some javadoc


kaikreuzer:
I thought to remember some comment from @sjka about why it is at it is at the moment.
But as it anyhow only concerns the Groovy tests, which we are being removed step by step, it probably won't cause to much harm if I merge this :-)

PR #4800, opened 2017-12-22
Allow binding DSL Rules to all items contained in a group

jmsjones:
This change allows writing rules that reference groups but execute based on the items within the group. This is primarily intended to extend the "triggeringItem" functionality provided in PR #4526, but also works with the other implicit variables.

Consider rules based on two switches contained in a group.
```
Item SwitchGroup changed
```
will only execute when a switch change results in the group state changing. The script will not know which switch caused the change.

```
Item Switch01 changed or
Item Switch02 changed
```
and
```
ItemIn SwitchGroup changed
```
are now functionaly identical. The script will execute based on an individual switch state change without regard for the impact on the group.
The "ItemIn" version is advantageous because items can be added to or removed from the group without modifying the rule.

**"ItemIn" has been replaced with "Member of" in an update below.**


rkoshak:
What about `SwitchGroup received update` and `SwitchGroup received command`?

Will those now only trigger the Rule once and triggeringItem get populated or is the old behavior retained?

What happens if I have a rule triggered by `SwitchGroup received command` and I call `SwitchGroup.sendCommand(ON)` from another Rule (or the sitemap). Will the Rule trigger and triggeringItem end up null?

I'm just trying to understand the full impact of the changes because changes to how Groups work seems to have a pretty big ripple effect.

jmsjones:
The existing functionality remains unchanged, and any rules that are triggered on `Item GROUP_NAME CONDITION` will behave as they currently do. This PR introduces a new trigger option of `ItemIn GROUP_NAME CONDITION` that will handle the members of the group instead of the group itself.

"Changed", "received update", and "received command" are all supported.

On a side note: I know that "ItemIn" as the new trigger keyword can be hard to distinguish from the existing "Item" keyword. This was the best I came up with at 2 AM and am welcome to suggestions.

jmsjones:
I decided that "Member of" is definitely a better trigger identifier than "ItemIn" and committed a change to reflect this.

From the original examples:
```
Item Switch01 changed or
Item Switch02 changed
```
and
```
Member of SwitchGroup changed
```
are equivalent.

jmsjones:
Conflicts with a recently merged PR have been resolved.

bundles/model/org.eclipse.smarthome.model.rule.runtime/src/org/eclipse/smarthome/model/rule/runtime/internal/engine/RuleTriggerManager.java
```diff
@@ -273,12 +279,42 @@ public RuleTriggerManager(Injector injector) {
         }
     }
 
-    private Iterable<Rule> internalGetRules(TriggerTypes triggerType, Item item, Type oldType, Type newType) {
-        List<Rule> result = new ArrayList<>();
-        Iterable<Rule> rules = getAllRules(triggerType, item.getName());
+    private Iterable<Rule> getRulesOrEmptySet(TriggerTypes type, String name) {
```
sjka: IMHO it would be better to amend `getAllRules()` accordingly. There were only two usages of it - and both guard themselves anyway by creating an empty collection if there was null returned.

bundles/model/org.eclipse.smarthome.model.rule/src/org/eclipse/smarthome/model/rule/Rules.xtext
```diff
@@ -41,12 +44,24 @@ CommandEventTrigger:
 	'Item' item=ItemName 'received command' (command=ValidCommand)?
 ;
 
+GroupMemberCommandEventTrigger:
+    'Member' 'of' group=ItemName 'received command' (command=ValidCommand)?
```
sjka: I like this `Member of ...´-version much better too! 

Independently of the naming one general remark though - this adds `Member` and `of`  as new keywords to the language. Therefore this it is breaking all existing rule scripts which are using these words for anything like variable names. This isn't necessarily a problem - just everybody should at least be aware of this.

bundles/model/org.eclipse.smarthome.model.rule.runtime/src/org/eclipse/smarthome/model/rule/runtime/internal/engine/RuleTriggerManager.java
```diff
@@ -289,11 +325,11 @@ public RuleTriggerManager(Injector injector) {
             case UPDATE:
                 if (newType instanceof State) {
                     State state = (State) newType;
-                    for (Rule rule : rules) {
+                    for (Rule rule : getRulesOrEmptySet(UPDATE, itemName)) {
                         for (EventTrigger t : rule.getEventtrigger()) {
                             if (t instanceof UpdateEventTrigger) {
                                 UpdateEventTrigger ut = (UpdateEventTrigger) t;
-                                if (ut.getItem().equals(item.getName())) {
+                                if (ut.getItem().equals(itemName)) {
```
sjka: I wonder if it wouldn't make sense to extract this block into a separate method, so it can be shared with the group member code below. Just to reduce the risk of somebody later on changing one and forgetting to adapt the other one...

bundles/model/org.eclipse.smarthome.model.rule.runtime/src/org/eclipse/smarthome/model/rule/runtime/internal/engine/RuleTriggerManager.java
```diff
@@ -306,17 +342,37 @@ public RuleTriggerManager(Injector injector) {
...
                 if (newType instanceof State && oldType instanceof State) {
                     State newState = (State) newType;
                     State oldState = (State) oldType;
-                    for (Rule rule : rules) {
+                    for (Rule rule : getRulesOrEmptySet(CHANGE, itemName)) {
                         for (EventTrigger t : rule.getEventtrigger()) {
                             if (t instanceof ChangedEventTrigger) {
                                 ChangedEventTrigger ct = (ChangedEventTrigger) t;
-                                if (ct.getItem().equals(item.getName())) {
+                                if (ct.getItem().equals(itemName)) {
```
sjka: see above

bundles/model/org.eclipse.smarthome.model.rule.runtime/src/org/eclipse/smarthome/model/rule/runtime/internal/engine/RuleTriggerManager.java
```diff
@@ -336,16 +392,43 @@ public RuleTriggerManager(Injector injector) {
...
             case COMMAND:
                 if (newType instanceof Command) {
                     final Command command = (Command) newType;
-                    for (Rule rule : rules) {
+                    for (Rule rule : getRulesOrEmptySet(COMMAND, itemName)) {
                         for (final EventTrigger t : rule.getEventtrigger()) {
                             if (t instanceof CommandEventTrigger) {
                                 final CommandEventTrigger ct = (CommandEventTrigger) t;
-                                if (ct.getItem().equals(item.getName())) {
+                                if (ct.getItem().equals(itemName)) {
```
sjka: see above

bundles/model/org.eclipse.smarthome.model.rule.runtime/src/org/eclipse/smarthome/model/rule/runtime/internal/engine/RuleTriggerManager.java
```diff
@@ -273,12 +279,42 @@ public RuleTriggerManager(Injector injector) {
...
+                break;
+            case THINGCHANGE:
+                rules = thingChangedEventTriggeredRules.get(name);
+                break;
+            default:
+                break;
+        }
         if (rules == null) {
-            rules = new ArrayList<>();
+            rules = new HashSet<>();
```
htreu: please set rules to `Collections.emptySet()` here.

bundles/model/org.eclipse.smarthome.model.rule/src/org/eclipse/smarthome/model/rule/Rules.xtext
htreu: Maybe we can lower the conflict potential by defining just a single keyword, like `MemberOf`?

sjka: That looks awkward though. To be honest, I really like the proposal as it is.

ptweety:
Just a question re/the implicitly available variable `triggeringItem`: Will the group by the triggeringItem or will it be the Item in the group? I suppose that with your example give above

`Item SwitchGroup changed`

`SwitchGroup` will be the `triggeringItem`. But

```
Item Switch01 changed or
Item Switch02 changed
```

and

`ItemIn SwitchGroup changed`

will have either `Switch01` or `Switch02` as `triggeringItem`

sjka:
Yes, the `triggeringItem` variable is still filled with the item that the event was received for.
Indeed very good having thought about that, thanks for bringing this up!

jmsjones:
Thanks for the feedback everyone. I have updated the code accordingly.

bundles/model/org.eclipse.smarthome.model.rule/src/org/eclipse/smarthome/model/rule/Rules.xtext
kaikreuzer: @sjka Wouldn't it be nicer though to have `'Member of'`, just like we have `'received command'`? As after "Member" there always has to be the word "of", this would make the content completion quicker and easier to understand.

sjka: Afair, having spaces in keywords is considered bad style (whitespace there suddenly matters and the tokenizer being greedy). However, as you pointed out, this is not the first place where it was done like that, therefore: why should be care now, right? 
This would indeed minimized the risk of collisions then.

jmsjones:
I believe all feedback to this point has been addressed. Are there any other concerns, questions, or change requests related to this PR?

sjka:
Argh, my apologies, somehow your update slipped through. Thanks for the reminder! 
All good from my point of view.

PR #4802, opened 2017-12-22
OneWire4J binding

aploese:
New binding for 1-Wire network.


J-N-K:
What licence is the onewire4j library? Is that compatible with EPL?

aploese:
SPSW and OneWire4J are under LGPL licensed.

kaikreuzer:
No, it is LGPL (https://github.com/aploese/spsw/blob/master/LICENSE.txt), so it is not compatible with EPL, hence we cannot accept the PR here, sorry.

@aploese Please also note that we are reluctant to accept any binding that requires serial access as long as https://github.com/eclipse/smarthome/issues/4465 is not settled - you might want to involve yourself in that discussion.

The "de-facto standard" for the moment is gnu.io for ESH-compatible bindings, but they for the moment usually reside in openhab2-addons repo.

Please also note that there is already another PR for OneWire support (#4287), so you might first want to liaise with @J-N-K on that.

aploese:
@kaikreuzer My last post there is 17 days old ;-)

kaikreuzer:
Right, so you are aware of the discussion. And as I pointed out there, anything LGPL isn't really a preferred choice for ESH.

kaikreuzer:
Why, will you reimplement the libs under a more permissive license? :-)

aploese:
Which libs?

kaikreuzer:
> SPSW and OneWire4J are under LGPL licensed.

aploese:
Because I get no material profit out of it - I think the LGPL is here a good choice - You can use it even in a closed source environment. Only if you change the LGPL code itself you will have to honour the LGPL.
Or am I wrong?
As I said before I could dual license the SPSW API - If that helps for #4465.

kaikreuzer:
LGPL is a big debate where many clever lawyers were involved (I am not one of them). The fact simply is that it is not one of the approved licenses, see https://www.eclipse.org/legal/licenses.php.
There are some exceptions made as the last sentence of that page suggests, but this is afaik really only meant for rare cases and it would involve a lot of legal discussions.

But yes, dual licensing would therefore help, so you might want to leave such a comment on #4465.

Issue #4445, opened 2017-10-19
Switch Dynamic Icons should represent Item State


Dracrius:
It was brought to my attention while testing a problem on androids openHAB app that Basic UI only pays attention to the switch state of an item when a switch is used in a sitemap regardless of the item type. 

This is fine and the simple way to do things but say you have a Dimmer item with a switch and a slider in your sitemap. The slider will show say the 50% dynamic icon because it's set to 50% but the switch right beside it would show 100% or I am assuming more accurately the ON icon. This seems like a very quick and dirty solution. 

I post this as an issue because until I brought up androids in ability to show color item icons dynamically [#408](https://github.com/openhab/openhab.android/issues/408) android actually fetched the item's state so any switch I had for a light would show an icon that represented more then just ON or OFF this was beautiful and made a simple switch more informative. The person helping me found how Basic UI handles this and has reverted the android app to the simpler approach. This is highly disappointing as that feels like a step backwards simply because of a standard most likely unintentionally set by Basic UI!

htreu:
@lolodomo could you have a look into this?

lolodomo:
You have an icon for each item state, that means one item for slide 0%, one for slide 10%, ... one for slider 100%, and you have an icon for ON state and for OFF state.
So if the current state is 50%, you will get the icon for 50% for the dimmer control and the icon for ON for the switch control.
That makes sense for me.

Dracrius:
But this is an overly simplified way to look at it. It may work but its kind of cheating. If an item has two controls a slider and a switch a user may put them together. Even if not there is no reason to slack here! An icon is supposed to represent the items current state not just the sitemap elements state. Also if your going to have dynamic icons shouldnt they all be equaly dynamic? The information is there so why not use it. I'm simply asking for the change from checking a visual element when deciding which icon to use to using the items actual state! This would result in any switch used with a dimmer or color item having the same icons as its corisponding slider rather then a slightly jaring 100% icon for lights only set to 50% right beside a slider for the same light showing a diffrent icon! This is tandamount to autoupdate=off keeping sliders from being updated to 100% because an ON command was sent.

Dracrius:
It boils down to the idea that i think i should be able to tell from my light switches page which lights are set high or low when ON based on their icons. Its a simple thing but it makes all the diffrence in feeling like a more polished product vs just what works. Dynamic icons are there to present visual information why not provide all the information you can!

lolodomo:
[...]

PR #4810, opened 2017-12-24
Sitemap SSE: provide unconverted item state

### LINKED ISSUE #4670: Sitemap SSE: item state vs widget state
As I explained in the issue #4445, there is something that should be fixed in the SSE events the server is producing. The SSE event includes data for the linked item and in particular its state. But in fact, this state is overwritten by a state adjusted to the widget type. It is done at this line: https://github.com/eclipse/smarthome/blob/master/bundles/io/org.eclipse.smarthome.io.rest.sitemap/src/main/java/org/eclipse/smarthome/io/rest/sitemap/internal/PageChangeListener.java#L204
As a result, the UI caller has no information on the real item state but only the widget state.
This conversion is helpful because the UI like Basic UI has not to implement this conversion logic.

I think it will be better to distinguish in the SSE event the real item state from the computed widget state. That is why I propose to add a new field that will contain the computed widget state, leaving the item state with its real state.
This change will probably break any app using SSE. As an example, a littke change will be required in the Basic UI to use the right field in the JavaScript file.

The only difficulty is that I would like to avoid duplicating big data. This can only happen with image items. I don't want to duplicate the image raw data in the SSE event.

WDYT ?

lolodomo:
Fixes #4670

### DIFF
--- a/bundles/io/org.eclipse.smarthome.io.rest.sitemap/src/main/java/org/eclipse/smarthome/io/rest/sitemap/internal/PageChangeListener.java
+++ b/bundles/io/org.eclipse.smarthome.io.rest.sitemap/src/main/java/org/eclipse/smarthome/io/rest/sitemap/internal/PageChangeListener.java
@@ -198,10 +198,15 @@ public class PageChangeListener implements StateChangeListener {
                 event.valuecolor = itemUIRegistry.getValueColor(w);
                 event.widgetId = itemUIRegistry.getWidgetId(w);
                 event.visibility = itemUIRegistry.getVisiblity(w);
+                // event.item contains data from the item including its state (in event.item.state)
                 event.item = EnrichedItemDTOMapper.map(item, false, null, null);
 
-                // adjust the state according to the widget type
-                event.item.state = itemUIRegistry.getState(w).toFullString();
+                // event.state is an adjustment of the item state to the widget type.
+                event.state = itemUIRegistry.getState(w).toFullString();
+                // In case this state is identical to the item state, its value is set to null.
+                if (event.state != null && event.state.equals(event.item.state)) {
+                    event.state = null;
+                }
 
                 events.add(event);
             }
--- a/bundles/io/org.eclipse.smarthome.io.rest.sitemap/src/main/java/org/eclipse/smarthome/io/rest/sitemap/internal/SitemapWidgetEvent.java
+++ b/bundles/io/org.eclipse.smarthome.io.rest.sitemap/src/main/java/org/eclipse/smarthome/io/rest/sitemap/internal/SitemapWidgetEvent.java
@@ -28,6 +28,7 @@ public class SitemapWidgetEvent extends SitemapEvent {
     public String labelcolor;
     public String valuecolor;
     public boolean visibility;
+    public String state;
     public EnrichedItemDTO item;
 
     public SitemapWidgetEvent() {
--- a/extensions/ui/org.eclipse.smarthome.ui.basic/web-src/smarthome.js
+++ b/extensions/ui/org.eclipse.smarthome.ui.basic/web-src/smarthome.js
@@ -1751,6 +1751,7 @@
 
 			var
 				data = JSON.parse(payload.data),
+				state,
 				value,
 				title;
 
@@ -1770,9 +1771,15 @@
 				return;
 			}
 
+			if (data.state === undefined) {
+				state = data.item.state;
+			} else {
+				state = data.state;
+			}
+
 			value = _t.extractValueFromLabel(data.label);
 			if (value === null) {
-				value = data.item.state;
+				value = state;
 			}
 			title = _t.getTitleFromLabel(data.label);
 
@@ -1789,7 +1796,7 @@
 					});
 				}
 
-				widget.setValue(smarthome.UI.escapeHtml(value), data.item.state, data.visibility);
+				widget.setValue(smarthome.UI.escapeHtml(value), state, data.visibility);
 
 				[{
 					apply: widget.setLabel,


lolodomo:
@resetnow : please have a look as the change is impacting mainly Basic UI.

bundles/io/org.eclipse.smarthome.io.rest.sitemap/src/main/java/org/eclipse/smarthome/io/rest/sitemap/internal/PageChangeListener.java
```diff
@@ -201,7 +201,12 @@ public void stateUpdated(Item item, State state) {
                 event.item = EnrichedItemDTOMapper.map(item, false, null, null);
 
                 // adjust the state according to the widget type
-                event.item.state = itemUIRegistry.getState(w).toFullString();
+                String state = itemUIRegistry.getState(w).toFullString();
+                if (state.equals(event.item.state)) {
+                    event.state = "#item.state#";
```
resetnow: Static final string (or what other constants use). And maybe write add a comment why it was introduced

lolodomo: Ok for the constant.
Maybe I will do this indirection only for big data like image.

lolodomo:
The end goal is already explained in the issue but I can explain it again. We don't want a sitemap SSE that can only be used with Basic UI but rather a sitemap SSE that could be used by any client UI and in a certain way similar to the sitemap REST API. For that reason, each event should contain the real item state.
Unfortunately the item state was replaced in the SSE event by a widget state because it was helpful for Basic UI. implementation. As this widget state is needed by Basic UI and could be an helper for any client UI, I kept it but move it to a new field of the event.

With this change, the sitemap SSE could be used by the Android app. The Android app will only use the item state because the Android app already implements its own conversion logic between item state and widget state.
On the other side, Basic UI front end will continue using the widget state.

resetnow:
>With this change, the sitemap SSE could be used by the Android app. The Android app will only use the item state because the Android app already implements its own conversion logic between item state and widget state.

That's the part I didn't know. Thanks for explaining.

lolodomo:
In fact, that is certainly the case for every UI using the sitemap REST API. Even the Basic UI backend implements its own conversion logic.
PS: this new field (widget state) could even be added to the sitemap REST API as an helper for the UIs. But that is not the purpose of this PR.

bundles/io/org.eclipse.smarthome.io.rest.sitemap/src/main/java/org/eclipse/smarthome/io/rest/sitemap/internal/PageChangeListener.java
```diff
@@ -201,7 +201,12 @@ public void stateUpdated(Item item, State state) {
                 event.item = EnrichedItemDTOMapper.map(item, false, null, null);
 
                 // adjust the state according to the widget type
-                event.item.state = itemUIRegistry.getState(w).toFullString();
```
htreu: With this change we break the API of the sitemap REST interface. The `event.item.state` is now not a widget specific conversion any more but the raw state.
How about turning this around and provide the raw state in `event.state` (in case it is different to `event.item.state` and leave the widget specific state on the item?

lolodomo: Yes you're right but sitemap SSE is very probably only used in Basic UI and the windows app. So I will break only the windows app.
I think it would be confusing to put/keep the widget state in the item section and put the item state in the widget/main section.
Additionally the sitemap REST API provides the same item structure and in this case we get the item state. So it would be confusing for a developer to have 2 different behaviors. And my intention was to add the widget state to rhe REST sitemap API in a second time and this time without breaking the API.

I think it is reasonable to break the API as it will only impact an app not very used and still in a build state.
But if the backward compatibility is the first priority, yes I can exchange the two fields and we can choose the confusing solution.

htreu: I agree with you that overwriting `event.item.state` with a widget specific state feels wrong and confusing in the first place. And also "mixing" in the REST interface is not good either.
However we do not know if and how other solutions do actually use the REST interface. Basic UI and the windows app are  only the consumers we know of. There might be others depending on the interface.
@maggu2810, @kaikreuzer & @sjka what is your take on this?

lolodomo:
I updated the PR to consider @resetnow review comments, that is a constant and more comments.

lolodomo:
@htreu : I discovered that the REST sitemap SSE is just unusable by any app using only the REST API. It is usable in Basic UI only because it has another low level access to the server through its servlet. I will open an issue to describe the problem and propose a fix.
All this to say that there is no risk to break the REST sitemap SSE while we take care to not break Basic UI.

lolodomo:
@kaikreuzer : please merge this PR before the next ESH stable build for OH.

lolodomo:
@NicoVermeir : please take into account this change in case you are really using sitemap SSE in the openHAB Windows app.

lolodomo:
Could this PR be merged finally. i am waiting for it with next PRs in the pipe.

lolodomo:
Many thanks to @resetnow for his quick review and approval for this PR.
Why does it remain now in an ignore state without any feedback ?
This PR is important to have a clean and usable sitemap SSE for any UI consummer (in addition to the second fix that was merged quickly - thanks to @kaikreuzer for that).
I spent several hours to implement SSE in the Android app (major technical improvment) and fix the ESH framework but everything is blocked until this PR is merged. How frustating it is for me to see this PR just being ignored.
Is it possible at least to have a feedback what is blocking the merge ?

resetnow:
@lolodomo I understand your frustration but please note that the main reason I was able to review quickly is because it's holiday time now in eastern europe. So people have jobs, etc

bundles/io/org.eclipse.smarthome.io.rest.sitemap/src/main/java/org/eclipse/smarthome/io/rest/sitemap/internal/PageChangeListener.java
wborn: > Yes you're right but sitemap SSE is very probably only used in Basic UI and the windows app. 

It's also used in my [ESH Grafana Javascript library](https://github.com/wborn/smarthome-grafana).

lolodomo: And I hope you are not against my improvment ?!

wborn: The proposed changes are probably the best way to solve the issues in all UIs without having to deal with legacy data in the future.

Is there some versioning on the interface that can be used to determine which fields should be used? That way it would be possible to consume the old/new API with the same code. The Windows App may also need this when it keeps supporting older OH2 versions. Of course I can also apply the same kind of change in my library that you've applied in Basic UI. 

It will probably cost people some time to figure out why it no longer works and what they have to do to fix it. Announcements and release notes may help with that.

lolodomo:
@wborn : the code change to apply in client to be compatible with old and new servers is very easy. In case your client code is expecting the converted item state (widget state), you will have to replace:
`state = event.item.state`
by
```
if (event.state exists AND event.state != "#item.state#" then state = event.state else state = event.item.state
```

In the opposite, for UI consumer like the Android app that requires the fix ((the raw item state that is not present in current servers), the change is more difficult. In this, you will have to consider the REST API version. If REST API is higher than 0.10.0 or 0.10.0 built in 2018 then use SSE in the app or use another method.

kaikreuzer:
> How frustating it is for me to see this PR just being ignored.

@lolodomo I understand the frustration that when you are working on something and think you have finished it, you want to get it off your plate asap.
But I very much hope that you also have some sense of proportion here: We have >60 open PRs, >340 open issues with discussions and >130 contributors that are impatiently waiting for feedback whenever they post something. Be assured that nobody likes the fact that there are so long queues, but it simply is not possible to do everything at once, especially if most people do this in their spare time (and yes, even I _could_ think of other leisures on a Sunday night).
This PR has received multiple comments 2 days after its creation and also has a longer discussion on its associated issue. It appears to break backward compatibility and touches a rather fragile part of the code, so it imho is required to invest time and to also actively test it before merging.
If you look at most of the other PRs in comparison, I am sure you will agree that "ignoring" is not an adequate word here as it rather offends everyone who involved himself here.

lolodomo:
Of course I understand all this but that is frustrating to notice that I could contribute largely more if everything was more reactive.
I will be patient and put in standby my work for rhe Android app.

Now it looks like the backward compatibility is more important than a clean API. In this case, my PR could become only 2 lines + a big comment to explain the confused data storage. If this could help a fast merge, let me know and I go in this direction.

kaikreuzer:
> Now it looks like the backward compatibility is more important than a clean API.

No, that's not what I was saying. My point was that when doing breaking changes, we should be really sure to do the right thing (and thus spend some thoughts/time when reviewing) and also involving more people in the review round (as you did e.g. with @NicoVermeir, whose feedback is also still pending).

NicoVermeir:
right, forgot about this during holidays, sorry!

as for the Windows app: we open a stream to rest/events and listen for data coming in through there. We filter out what item has changed through the item name

lolodomo:
@NicoVermeir : so you are not using the /rest/sitemaps/events/subscribe ?

bundles/io/org.eclipse.smarthome.io.rest.sitemap/src/main/java/org/eclipse/smarthome/io/rest/sitemap/internal/PageChangeListener.java
```diff
@@ -198,10 +200,19 @@ public void stateUpdated(Item item, State state) {
...
                 event.item = EnrichedItemDTOMapper.map(item, false, null, null);
 
-                // adjust the state according to the widget type
-                event.item.state = itemUIRegistry.getState(w).toFullString();
+                // event.state is an adjustment of the item state to the widget type.
+                // In case this state is big and identical to the item state, an indirection to the
+                // item state is provided by setting its value to "#item.state#" in order to avoid
+                // duplicating in the event message a big amount of data.
+                String state = itemUIRegistry.getState(w).toFullString();
+                if ((state.length() > 128) && state.equals(event.item.state)) {
```
maggu2810: The client does need to handle `REDIRECT_TO_ITEM_STATE`, so why using the magic number 128?
If the client check if `event.state` is equal to `REDIRECT_TO_ITEM_STATE` and use `event.item.state` wouldn't already a length greater then `REDIRECT_TO_ITEM_STATE.length()` make sense?

bundles/io/org.eclipse.smarthome.io.rest.sitemap/src/main/java/org/eclipse/smarthome/io/rest/sitemap/internal/PageChangeListener.java
maggu2810: > @maggu2810, @kaikreuzer & @sjka what is your take on this?

I would prefer useful names instead of keeping backward compatibility.
A consumer of different version should check the used ESH version.

NicoVermeir:
@lolodomo no, only the rest/events endpoint is used

bundles/io/org.eclipse.smarthome.io.rest.sitemap/src/main/java/org/eclipse/smarthome/io/rest/sitemap/internal/PageChangeListener.java
lolodomo: My intention was to reduce the use of redirection in most of the cases (for reasonable size), for a maybe better readability and for a working in most cases for UI client that even forget to consider the redirection.
Your proposal is more for the optimal size.
I am ok to change that.

lolodomo:
Ok, perfect, even the Windows app is not concerned by this breaking API change.
Ifr I exclude the Basic UI which is already handled in my PR, the only known impacted client would be the ESH Grafana Javascript library.

lolodomo:
@maggu2810 review comment now integrated.

bundles/io/org.eclipse.smarthome.io.rest.sitemap/src/main/java/org/eclipse/smarthome/io/rest/sitemap/internal/PageChangeListener.java
```diff
@@ -198,10 +200,19 @@ public void stateUpdated(Item item, State state) {
...
                 event.item = EnrichedItemDTOMapper.map(item, false, null, null);
 
-                // adjust the state according to the widget type
-                event.item.state = itemUIRegistry.getState(w).toFullString();
+                // event.state is an adjustment of the item state to the widget type.
+                // In case this state is big and identical to the item state, an indirection to the
+                // item state is provided by setting its value to "#item.state#" in order to avoid
+                // duplicating in the event message a big amount of data.
+                String state = itemUIRegistry.getState(w).toFullString();
+                if ((state.length() > REDIRECT_TO_ITEM_STATE.length()) && state.equals(event.item.state)) {
```
kaikreuzer: This looks dodgy. Is this the coded translation of "In case this state is big and identical to the item state"?
If so, why is anything bigger than 12 already "big"?

bundles/io/org.eclipse.smarthome.io.rest.sitemap/src/main/java/org/eclipse/smarthome/io/rest/sitemap/internal/PageChangeListener.java
```diff
@@ -198,10 +200,19 @@ public void stateUpdated(Item item, State state) {
...
                 event.widgetId = itemUIRegistry.getWidgetId(w);
                 event.visibility = itemUIRegistry.getVisiblity(w);
+                // event.item contains data from the item including its state (in event.item.state)
                 event.item = EnrichedItemDTOMapper.map(item, false, null, null);
 
-                // adjust the state according to the widget type
-                event.item.state = itemUIRegistry.getState(w).toFullString();
+                // event.state is an adjustment of the item state to the widget type.
+                // In case this state is big and identical to the item state, an indirection to the
+                // item state is provided by setting its value to "#item.state#" in order to avoid
```
kaikreuzer: This looks like a very nasty hack to me. How is any client be supposed to know this? If a String item state changes by one character from 12 to 13 in length, the events will all of a sudden have a different content?

Those magic strings should imho be really avoided.

It would probably be better to simply say that the widget state is generally only provided if it differs from the item state. Clients would not have to check for any magic strings, but could simply implement a simple fallback to the item state. Or am I oversimplifying things here?

lolodomo: Yes I like very much this idea. I don't remember why I abandoned it :)
I think I already added in the past the transformed state in the item data which is optional, that is present only if different from the state. So it is clearly doable.

The only problem I see with that solution is that a client UI will not be able to distinguish old behaviour from new behaviour just with the event. It will probably require to check the API version.

bundles/io/org.eclipse.smarthome.io.rest.sitemap/src/main/java/org/eclipse/smarthome/io/rest/sitemap/internal/PageChangeListener.java
lolodomo: Because @maggu2810 suggested this change yesterday ;) Before that, it was set to 128 bytes.
But considering your other suggestion, we don't care about that now.

bundles/io/org.eclipse.smarthome.io.rest.sitemap/src/main/java/org/eclipse/smarthome/io/rest/sitemap/internal/PageChangeListener.java
lolodomo: I think I abandoned this idea due to @digitaldan comment: https://github.com/eclipse/smarthome/issues/4445#issuecomment-341996820

lolodomo: But as a logic is required by the way if we don't want to duplicate big data, I think this argument is no more valid.

bundles/io/org.eclipse.smarthome.io.rest.sitemap/src/main/java/org/eclipse/smarthome/io/rest/sitemap/internal/PageChangeListener.java
maggu2810: Yes, it has been suggested by me here: https://github.com/eclipse/smarthome/pull/4810#discussion_r160153919

WRT to @kaikreuzer comments above, not using a magic string constant at all would be better -- I agree.
But if we use a magic string constant, I still think it should be used as soon as it makes some difference and not only by a magic "128". :wink: 

So, if we can get rid off the magic string it would be better at all.

kaikreuzer: Ah, thanks, I had missed that comment.

But ok, it seems we all agree with abandoning the REDIRECT_TO_ITEM_STATE string completely.
@lolodomo Once you have done this change, I think we should be good to merge.

lolodomo: Ok.
I cannot do that before Friday in the best case.

kaikreuzer: Ok, no problem - I would then schedule a first ESH stable build since the OH 2.2 release for next Monday, so that this change is included.

lolodomo:
@kaikreuzer : final change done. I squashed everything.

wborn:
Is your [code fragment](https://github.com/eclipse/smarthome/pull/4810#issuecomment-355811421) to be compatible with both the old and new API still valid @lolodomo after you reworked the code?

lolodomo:
I give you the new instructions later today. It will be required to check the REST API version first.

lolodomo:
@wborn : sorry I forgot to answer to you.
First you have to check the REST API version. If this version is greater than 0.10.0.xxx or is 0.10.0.2018xxx, you have the changed behaviour inside the server. Then to get the converted state, use this: `if (event.state exists then state = event.state else state = event.item.state`

wborn:
Thanks for the update @lolodomo. I'll be testing the changes soon!

PR #4820, opened 2017-12-28
Fix widget ids generated by the sitemap REST API

### LINKED ISSUE #4819: Sitemap SSE vs sitemap REST API: diffent widget ID
For a client UI, the normal way to handle a page is first to download the page with all its widgets using the sitemap REST API and then to use the sitemap SSE to get updates of widget in the page. For that, the SSE event contains a widget id toidentify whart widget is updated.
The problem is that the two API are using different widget ID for the same widget ! So when you receive an event through the SSE, you are not able to identify which widget from the page retrieved with the sitemap REST API is concerned.

I will propose a fix to unify the widget IDs in the sitemap REST API and the sitemap SSE. I will keep the IDs from the sitemap SSE because they are taking care of the hidden widgets when building the ID while the sitemap REST API is ignoring the hidden widgets.

Note that the Sitemap SSE is used and working well in Basic UI only because it has an additional interface to the server through its servlet and so is able to retrieve the IDs that will then be used by the sitemap SSE.

lolodomo:
Same widget ids as the ones in the sitemap SSE events

Fixes #4819

### DIFF
--- a/bundles/io/org.eclipse.smarthome.io.rest.sitemap/src/main/java/org/eclipse/smarthome/io/rest/sitemap/internal/SitemapResource.java
+++ b/bundles/io/org.eclipse.smarthome.io.rest.sitemap/src/main/java/org/eclipse/smarthome/io/rest/sitemap/internal/SitemapResource.java
@@ -377,14 +377,12 @@ public class SitemapResource implements RESTResource, SitemapSubscriptionCallbac
         bean.leaf = isLeaf;
         bean.link = UriBuilder.fromUri(uri).path(PATH_SITEMAPS).path(sitemapName).path(pageId).build().toASCIIString();
         if (children != null) {
-            int cntWidget = 0;
             for (Widget widget : children) {
-                String widgetId = pageId + "_" + cntWidget;
+                String widgetId = itemUIRegistry.getWidgetId(widget);
                 WidgetDTO subWidget = createWidgetBean(sitemapName, widget, drillDown, uri, widgetId, locale);
                 if (subWidget != null) {
                     bean.widgets.add(subWidget);
                 }
-                cntWidget++;
             }
         } else {
             bean.widgets = null;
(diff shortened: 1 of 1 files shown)


lolodomo:
Hi @kaikreuzer.
To implement sitemap SSE in the Android app, I needed this fix (and the other one not yet merged).
When you will build a new stable version of the ESH framework including the two fixes I need, is it possible that you increase the REST API version ? This will help me to enable SSE in the Android app only when the REST API has this new minimal version.
Current version of the API is 0.10.0.b1 ?

kaikreuzer:
The version provided in the REST API is always the version of the ESH build, so it will be automatically updated with every ESH stable. At the moment, openHAB is still hard coded to use the b1 release, but that will be changed again.

lolodomo:
Ok, I see it now in https://github.com/openhab/openhab-core/blob/1d0c749514bacd58391fc988ffce932842e8a7d9/poms/pom.xml#L61
Can you already tell me what you plan as new version for the next build ?

kaikreuzer:
As I said: The version will be the ESH build version, e.g. 0.10.0.201801122359.

lolodomo:
You could switch to 0.10.1.xxxxx at the right time so that I can identify the version that includes all I need ?

kaikreuzer:
No, 0.10.1.xxx versions can clearly only be built once a 0.10.0 release is done.

lolodomo:
@NicoVermeir : please check the impact of this change in case you are really using sitemap SSE in the openHAB Windows app.

maniac103:
> As I said: The version will be the ESH build version, e.g. 0.10.0.201801122359.

@kaikreuzer Does that 'will be' statement apply to current builds already or - if not - will it apply to the next release?

kaikreuzer:
@maniac103 This is in place since years this way - what you see in openHAB are all the ESH stable builds (https://ci.eclipse.org/smarthome/job/SmartHomeDistribution-Stable/).

We might change the versioning soon, though, please have a look at https://github.com/eclipse/smarthome/pull/5032 and feel free to comment (I actually started that discussion on your (=Android app maintainer) behalf.

maniac103:
@kaikreuzer I was asking because the demo server (which runs 2.2, right?) returns '0.10.0.b1' ... what I was referring to was the point in time when the 'b1' (beta marker?) is replaced by the timestamp.

kaikreuzer:
openHAB 2.2 release uses the ESH 0.10.0.b1 release (and thus no snapshot, so that it does not include a timestamp). This is exactly the reason why I created #5032 - please read its description and tell us what you think/need.

maniac103:
Sorry, I should have read it before commenting. I'll comment there.

Issue #4825, opened 2017-12-30
Exceptions in Actions are not logged


smerschjohann:
Exceptions in Actions aren't logged, instead only the action id is mentioned.
Example: `Failed to execute rule '420aa8d5-3f8b-41a3-be5c-6d64f49d7148': Fail to execute action: 2`

Workaround: Set the log level of the automation.core to DEBUG.

Proposed solution: Add the Exception message behind "Fail[ed] to execute action"...

PR #4826, opened 2017-12-30
Include exception text in error message of failed actions

smerschjohann:
solves #4825


J-N-K:
Good idea.

PR #4830, opened 2018-01-01
(Sonos) Added note to ReadMe about the handling of the 'notificationsoundvolume'

JueBag:
As discussed in this OH community thread (https://community.openhab.org/t/sonos-notificationvolume-after-restart/19055), the 'notificationsoundvolume' has to be set by the user. 
The binding does set this volume to equal the master volume, however the observed sound volume of an notification with the same setting as the master volume is barely audible ( I use a master volume of about 5 and a 'notificationsoundvolume' of 20!).
The automatic setting of that value to the master volume after each restart of the system makes the use of a persistence service obsolete.
IMHO users should be noted to take care of that setting themself.

Signed-off-by: Jürgen Baginski <opus42@gmx.de> (github:JueBag)

### DIFF
--- a/extensions/binding/org.eclipse.smarthome.binding.sonos/README.md
+++ b/extensions/binding/org.eclipse.smarthome.binding.sonos/README.md
@@ -88,7 +88,7 @@ The devices support the following channels:
 
 All supported Sonos devices are registered as an audio sink in the framework.
 Audio streams are treated as notifications, i.e. they are fed into the `notificationsound` channel and changing the volume of the audio sink will change the `notificationvolume`, not the `volume`.
-Note that the 'notificationsoundvolume' is set by the binding during each start of the system to equal the master volume (resulting in a barely audible level), making the use of a persistance policy obsolete. In order to control the 'notificationsoundvolume' users have to set it manually or by a rule.
+Note that the `notificationvolume` is set by the binding during each start of the system to equal the master volume (resulting in a barely audible level), making the use of a persistence policy obsolete. In order to control the `notificationvolume` users have to set it manually or by a rule.
 Note that the Sonos binding has a limit of 20 seconds for notification sounds. Any sound that is longer than that will be cut off.
 
 URL audio streams (e.g. an Internet radio stream) are an exception and do not get sent to the `notificationsound` channel. Instead, these will be sent to the `playuri` channel.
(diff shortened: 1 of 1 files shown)


extensions/binding/org.eclipse.smarthome.binding.sonos/README.md
```diff
@@ -88,6 +88,7 @@ The devices support the following channels:
 
 All supported Sonos devices are registered as an audio sink in the framework.
 Audio streams are treated as notifications, i.e. they are fed into the `notificationsound` channel and changing the volume of the audio sink will change the `notificationvolume`, not the `volume`.
+Note that the 'notificationsoundvolume' is set by the binding during each start of the system to equal the master volume (resulting in a barely audible level), making the use of a persistance policy obsolete. In order to control the 'notificationsoundvolume' users have to set it manually or by a rule.
```
sjka: Looking at the above channels, you meant `notificationvolume` and not `notificationsoundvolume`, right?
Please also note the usage of the ` instead of ' for properly quoting the names.

sjka: persistance -> persistence

sjka:
Thanks, this indeed is a very useful addition! 

Regarding the IP validation error: Don't worry about the failed check, I confirmed manually that your ECA is valid. I guess the automated check just got confused by the trailing github handle...

sjka:
> This is my first one in here! 

No worries, all good. 
We're happy to have you here and looking forward to many more contributions!

Thanks!

kaikreuzer:
> Thanks, this indeed is a very useful addition!

@sjka, Question: While the hint is indeed useful, doesn't it though highlight an unexpected/buggy behavior of the binding?
The problem is that the channel state is actually not backed by any real physical device state (which could be re-read after a reboot), but that this is a pure "software feature". Shouldn't the binding ideally itself persist this setting, so that the user does not have to worry about it by configuring persistence or creating rules?

sjka:
Yes - the question is why this is a channel at all. It sounds like it should rather be a configuration parameter of the Thing (or even Binding?). On the otherhand it makes sense to have it as a channel so sliders can be linked against them. I opened #4832 to continue this discussion.

Still, in the meantime it was a good idea of JueBag to document the current pitfall properly!

kaikreuzer:
> Still, in the meantime it was a good idea of JueBag to document the current pitfall properly!

Absolutely!

PR #4834, opened 2018-01-02
Delayed model processing

### LINKED ISSUE #4824: Rules/Sitemaps/Scripts refreshed twice
When changing files (rules, scripts, sitemaps), this results in openHAB to Load and Refresh the file twice, causing unnecessary delays especially on embedded platforms.

**Expected behaviour**
When changing files, the ModelRepositoryImpl should only refresh known files. When a new file is added, it should be added and watched out for changes.

**Current behaviour**
When changing files (rules, scripts, sitemaps), this results in openHAB to Load and Refresh the file twice, causing unnecessary delays especially on embedded platforms.

**Possible solution**
None known. Workaround: Stop openhab, change files as needed, start openhab again.

**Steps to reproduce**
Change a configuration file (rule, sitemap, script) by using any editor. Look at the openhab.log and notice "Loading model XXX", "Refreshing model XXX", "Refreshing model XXX".

**My environment**
Debian 9.3 on raspberry pi 3 with openhab2 2.2.0-1 stable out of repo. 
root@openhab2:~# java -version
openjdk version "1.8.0_152"
OpenJDK Runtime Environment (Zulu Embedded 8.25.0.76-linux-aarch32hf) (build 1.8.0_152-b76)
OpenJDK Client VM (Zulu Embedded 8.25.0.76-linux-aarch32hf) (build 25.152-b76, mixed mode, Evaluation)

**Additional notes**
The bug may be caused by one of these places: https://github.com/eclipse/smarthome/blob/36c9e2717e4f89cf60a0d885e7ecc6069b338f48/bundles/model/org.eclipse.smarthome.model.core/src/main/java/org/eclipse/smarthome/model/core/internal/folder/FolderObserver.java#L242

https://github.com/eclipse/smarthome/blob/36c9e2717e4f89cf60a0d885e7ecc6069b338f48/bundles/model/org.eclipse.smarthome.model.core/src/main/java/org/eclipse/smarthome/model/core/internal/ModelRepositoryImpl.java#L91

Also discussed in the community
https://community.openhab.org/t/rules-still-loading-running-twice/22183

sjka:
...in order to skip duplicate file system events or intermediate
states where the file technically is empty for a very short
moment.

fixes #4824

### DIFF
--- a/bundles/config/org.eclipse.smarthome.config.xml/src/main/java/org/eclipse/smarthome/config/xml/osgi/XmlDocumentBundleTracker.java
+++ b/bundles/config/org.eclipse.smarthome.config.xml/src/main/java/org/eclipse/smarthome/config/xml/osgi/XmlDocumentBundleTracker.java
@@ -58,7 +58,7 @@ import org.slf4j.LoggerFactory;
  */
 public class XmlDocumentBundleTracker<T> extends BundleTracker<Bundle> {
 
-    public static final String THREAD_POOL_NAME = "xml-processing";
+    public static final String THREAD_POOL_NAME = "file-processing";
 
     private final Logger logger = LoggerFactory.getLogger(XmlDocumentBundleTracker.class);
     private final ScheduledExecutorService scheduler = ThreadPoolManager.getScheduledPool(THREAD_POOL_NAME);
@@ -73,7 +73,7 @@ public class XmlDocumentBundleTracker<T> extends BundleTracker<Bundle> {
 
     @SuppressWarnings("rawtypes")
     private BundleTracker relevantBundlesTracker;
-    private ReadyService readyService;
+    private final ReadyService readyService;
 
     /**
      * Creates a new instance of this class with the specified parameters.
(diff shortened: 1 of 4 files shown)


bundles/config/org.eclipse.smarthome.config.xml/src/main/java/org/eclipse/smarthome/config/xml/osgi/XmlDocumentBundleTracker.java
```diff
@@ -58,7 +58,7 @@
  */
 public class XmlDocumentBundleTracker<T> extends BundleTracker<Bundle> {
 
-    public static final String THREAD_POOL_NAME = "xml-processing";
+    public static final String THREAD_POOL_NAME = "file-processing";
```
kaikreuzer: You were probably already waiting for this question: Would it make sense to move this constant to some "common" place as it is declared twice now in different classes (but they are meant to be identical to each other)?

PR #4846, opened 2018-01-04
[Sonos] Added property for storing the 'notificationvolume' and removed 'notificationvolume' channel

### LINKED ISSUE #4832: [Sonos] notificationvolume is not persistent
...as it is a pure "binding" feature and not backed by the real device at all (see https://github.com/eclipse/smarthome/pull/4830#issuecomment-354727202).

Question is: why is this a channel and not a configuration parameter? And if it is to remain a channel, shouldn't the handler persist the state somewhere itself rather than relying on the user to set the notificationvolume again and again after every restart?

cweitkamp:
- Added a property for persisting the `notificationvolume`.
- Removed the `notificationvolume` channel
- Adapted audio sink implementation to ESH API

NOTE: This is an APIBreaking change.

Closes #4832 


extensions/binding/org.eclipse.smarthome.binding.sonos/src/main/java/org/eclipse/smarthome/binding/sonos/internal/handler/ZonePlayerHandler.java
```diff
@@ -119,11 +121,6 @@
      */
     private final Object notificationLock = new Object();
 
-    /**
-     * Separate sound volume used for the notification
-     */
-    private String notificationSoundVolume = null;
```
cweitkamp: TODO: updateChannel in initialize of handler after restart

extensions/binding/org.eclipse.smarthome.binding.sonos/ESH-INF/config/config.xml
```diff
@@ -1,23 +1,28 @@
...
+			<description>Set the volume applied to a notification sound</description>
+			<default>-1</default>
+		</parameter>
 		<parameter name="refresh" type="integer" unit="s">
 			<label>Refresh interval</label>
 			<description>Specifies the refresh interval in seconds</description>
 			<default>60</default>
 		</parameter>
 	</config-description>
+
```
triller-telekom: Can you please remove the unnecessary empty line here?

extensions/binding/org.eclipse.smarthome.binding.sonos/src/main/java/org/eclipse/smarthome/binding/sonos/internal/handler/ZonePlayerHandler.java
triller-telekom: Good that you found this one by yourself :) While reading your source I was already asking myself where the initial setting should come from since `handleConfigurationUpdate` is only called if the configuration is changed, right? If it is also called while starting the handler to load the intial config you can omit the implementation in `initialize`. Please check that first.

extensions/binding/org.eclipse.smarthome.binding.sonos/ESH-INF/config/config.xml
```diff
@@ -1,23 +1,28 @@
...
 		</parameter>
 		<parameter name="notificationTimeout" type="integer" unit="s">
 			<label>Notification Timeout</label>
 			<description>Specifies the amount of time in seconds for which the notification sound will be played</description>
 			<default>20</default>
 		</parameter>
+		<parameter name="notificationVolume" type="integer" min="0" max="100" step="1" unit="%">
+			<label>Notification Sound Volume</label>
+			<description>Set the volume applied to a notification sound</description>
+			<default>-1</default>
```
kaikreuzer: -1 is not a valid value. Why at all declare a default?

extensions/binding/org.eclipse.smarthome.binding.sonos/src/main/java/org/eclipse/smarthome/binding/sonos/internal/handler/ZonePlayerHandler.java
```diff
@@ -1541,33 +1554,37 @@ private SonosZoneGroup getCurrentZoneGroup() {
...
-    public void setNotificationSoundVolume(Command command) {
-        if (command != null) {
-            notificationSoundVolume = command.toString();
+    public void setNotificationSoundVolume(PercentType notificationSoundVolume) {
+        if (notificationSoundVolume != null) {
+            logger.debug("Sonos device {}: Update configuration for 'notificationVolume' to {}", getThing().getUID(),
+                    notificationSoundVolume);
+            Configuration config = editConfiguration();
+            config.put(ZonePlayerConfiguration.NOTIFICATION_VOLUME, notificationSoundVolume.toBigDecimal());
+            updateConfiguration(config);
```
kaikreuzer: This is imho pretty nasty and hints at the fact that this is rather a hack than a recommended way to model things. I would vote for removing the channel completely, see my https://github.com/eclipse/smarthome/issues/4832#issuecomment-355505603.

extensions/binding/org.eclipse.smarthome.binding.sonos/src/main/java/org/eclipse/smarthome/binding/sonos/internal/handler/ZonePlayerHandler.java
cweitkamp: This is not relevant anymore.

extensions/binding/org.eclipse.smarthome.binding.sonos/ESH-INF/i18n/sonos_de.properties
```diff
@@ -29,6 +29,8 @@ thing-type.config.sonos.zoneplayer.udn.label = Eindeutiger Name (UDN)
 thing-type.config.sonos.zoneplayer.udn.description = Der UDN zur Identifizierung des Sonos Ger�tes.
 thing-type.config.sonos.zoneplayer.notificationTimeout.label = Timeout f�r Benachrichtigungen
 thing-type.config.sonos.zoneplayer.notificationTimeout.description = Timeout f�r Benachrichtigungen (in Sekunden).
+thing-type.config.sonos.zoneplayer.notificationVolume.label = Lautst�rke der Benachrichtigung
+thing-type.config.sonos.zoneplayer.notificationVolume.description = Zeigt die Lautst�rke f�r Benachrichtigungen (in %) an.
```
maggu2810: Shouldn't it be "Die Lautstärke für ...". The refresh description uses "Intervall zur Abfrage" and not "Zeigt das ..."

extensions/binding/org.eclipse.smarthome.binding.sonos/ESH-INF/config/config.xml
```diff
@@ -1,19 +1,22 @@
...
-			<required>true</required>
 		</parameter>
 		<parameter name="notificationTimeout" type="integer" unit="s">
 			<label>Notification Timeout</label>
 			<description>Specifies the amount of time in seconds for which the notification sound will be played</description>
 			<default>20</default>
 		</parameter>
+		<parameter name="notificationVolume" type="integer" min="0" max="100" step="1" unit="%">
+			<label>Notification Sound Volume</label>
+			<description>Set the volume applied to a notification sound</description>
```
maggu2810: Shouldn't we remove the word "Set"?

extensions/binding/org.eclipse.smarthome.binding.sonos/src/main/java/org/eclipse/smarthome/binding/sonos/internal/handler/ZonePlayerHandler.java
```diff
@@ -1557,33 +1550,34 @@ private SonosZoneGroup getCurrentZoneGroup() {
...
      *
-     * @param command
+     * @param notificationSoundVolume
      */
-    public void setNotificationSoundVolume(Command command) {
-        if (command != null) {
-            notificationSoundVolume = command.toString();
+    public void setNotificationSoundVolume(PercentType notificationSoundVolume) {
+        if (notificationSoundVolume != null) {
+            // set new volume only if it is higher than the current volume
```
maggu2810: Why? My notification sounds are usually static files that volume range is known.
The different internet radio could require a louder volume level, so why not using the notification volume that has been defined and adding such a logic here?

extensions/binding/org.eclipse.smarthome.binding.sonos/src/main/java/org/eclipse/smarthome/binding/sonos/internal/handler/ZonePlayerHandler.java
```diff
@@ -2409,9 +2399,13 @@ private void handleEmptyQueue(Command notificationURL, ZonePlayerHandler coordin
      * @param coordinator - {@link ZonePlayerHandler} coordinator for the SONOS device(s)
      */
     private void applyNotificationSoundVolume() {
-        PercentType volume = getNotificationSoundVolume();
-        if (volume != null) {
-            setVolumeForGroup(volume);
+        PercentType notificationSoundVolume = getNotificationSoundVolume();
+        if (notificationSoundVolume != null) {
+            // set new volume only if it is higher than the current volume
```
maggu2810: see above

extensions/binding/org.eclipse.smarthome.binding.sonos/src/main/java/org/eclipse/smarthome/binding/sonos/internal/handler/ZonePlayerHandler.java
cweitkamp: I can tell you. I want to preserve a little bit of the previous functionality. The `notificationVolume` property covers maybe 90% of your notifications and will be used every time. You cannot change the value during runtime in any case. And then there is this one situation when you need a very high noise (e.g. an intruder alert). With this condition you are able to set the normal volume in a rule to the abnormal level and it prevents the binding from using your default notification volume.

cweitkamp:
@maggu2810 Thanks for your review. I changed the descriptions in english and german.

lolodomo:
Like @maggu2810, I think very tricky to apply the notification volume only when higher than the current volume.
With the next PR #4963, it will then be possible to request a volume for the notification, so I don't understand this specific rule.

cweitkamp:
Yes, finally I agree with the two of you. After PR #4963  this is not necessary anymore. But we have to do some more changes in this binding. Depends on the final solution of #4963 .

kaikreuzer:
FTR, #4963 is now merged. Do you thus want to rework this PR?

cweitkamp:
Yes, I want to do de more changes.

cweitkamp:
I rebased my branch and added some final changes.

extensions/binding/org.eclipse.smarthome.binding.sonos/README.md
```diff
@@ -87,11 +89,12 @@ The devices support the following channels:
 ## Audio Support
 
 All supported Sonos devices are registered as an audio sink in the framework.
-Audio streams are treated as notifications, i.e. they are fed into the `notificationsound` channel and changing the volume of the audio sink will change the `notificationvolume`, not the `volume`.
-Note that the `notificationvolume` is set by the binding during each start of the system to equal the master volume (resulting in a barely audible level), making the use of a persistence policy obsolete. In order to control the `notificationvolume` users have to set it manually or by a rule.
-Note that the Sonos binding has a limit of 20 seconds for notification sounds. Any sound that is longer than that will be cut off.
+Audio streams are treated as notifications, i.e. they are fed into the `notificationsound` channel and changing the volume of the audio sink will change the `notificationVolume` defined in the properties fo the thing, not the `volume`.
```
lolodomo: Typo: "of" rather than "fo".
I would use "settings" rather than "properties".

extensions/binding/org.eclipse.smarthome.binding.sonos/README.md
```diff
@@ -89,7 +89,7 @@ The devices support the following channels:
 ## Audio Support
 
 All supported Sonos devices are registered as an audio sink in the framework.
-Audio streams are treated as notifications, i.e. they are fed into the `notificationsound` channel and changing the volume of the audio sink will change the `notificationVolume` defined in the properties fo the thing, not the `volume`.
+Audio streams are treated as notifications, i.e. they are fed into the `notificationsound` channel and changing the volume of the audio sink will change the `notificationVolume` defined in the properties of the thing, not the `volume`.
```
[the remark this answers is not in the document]
cweitkamp: I changed it.

kaikreuzer:
Looks all wonderful - excellent work, @cweitkamp!

PR #4847, opened 2018-01-04
Paper UI support for multiple configuration contexts

triller-telekom:
A service defined as

`@Component(configurationPolicy = ConfigurationPolicy.REQUIRE, service = MyService.class, configurationPid = "myPid")`

can now be changed into one that can get multiple configurations via Paper UI.

To do so one has to create a service with `@Component(immediate = true)` that implements the `MultipleInstanceServiceInfo` interface and returns the `configurationPid` of the service that should be configured multiple times ("myPid") in the `getServicePID()` method.

This new service appears in Paper UI under the services tab and has a "manage" button where all its configurations can be created/edited/deleted.

Service contexts created via the `pid:myPid#context` notation in the file system do also show up and can be edited but will revert the changes to the content of the file once the system is restarted.

Also-by: Henning Treu <henning.treu@telekom.de>


htreu:
ip-validation complains can be ignored. all commits are authored and signed-off by @triller-telekom and me.

extensions/ui/org.eclipse.smarthome.ui.paper/web-src/js/controllers/controller.configurableService.dialog.js
```diff
@@ -29,6 +29,25 @@ configurableServiceDialogController.controller('ConfigurableServiceDialogControl
...
 
+    var addESHServiceContextParameter = function(configDescription) {
+        var found = false;
+        angular.forEach(configDescription.parameters, function(parameter) {
+            if (parameter.name === 'esh.servicecontext') {
+                found = true;
+            }
+        })
+        if (!found) {
+            var eshServiceContextParameter = {
```
sjka: Could you please add some description that guides the non-savvy user a little on what to do with this "Service Context" field?

bundles/test/org.eclipse.smarthome.magic/src/main/java/org/eclipse/smarthome/magic/service/MagicMultiInstanceService.java
```diff
@@ -0,0 +1,51 @@
...
+import org.slf4j.Logger;
+import org.slf4j.LoggerFactory;
+
+/**
+ * Testing service for multi-context configurations.
+ *
+ * @author Stefan Triller - initial contribution
+ *
+ */
+@Component(configurationPolicy = ConfigurationPolicy.REQUIRE, service = MagicMultiInstanceService.class, configurationPid = "org.eclipse.smarthome.magicMultiInstance")
```
sjka: this would be far more impressive if it had the `immediate = true` flag set, as nobody is using this service therefore it never gets activated. The `service = ...`  attribute however is not needed.

bundles/config/org.eclipse.smarthome.config.core/src/main/java/org/eclipse/smarthome/config/core/ConfigurableService.java
```diff
@@ -49,4 +49,9 @@
      */
     public static final String SERVICE_PROPERTY_CATEGORY = "service.config.category";
 
+    /**
+     * The factory pid of the service to be configured (for services with multiple contexts).
+     */
+    public static final String SERVICE_PROPERTY_FACTORY_PID = "service.factoryPid";
```
sjka: please `org.osgi.service.cm.ConfigurationAdmin.SERVICE_FACTORYPID` instead of defining your own constant here

bundles/config/org.eclipse.smarthome.config.core/src/main/java/org/eclipse/smarthome/config/core/MultipleInstanceServiceInfo.java
```diff
@@ -0,0 +1,55 @@
...
+
+/**
+ * Interface marking the service with service.pid as one that can be instantiated multiple times with different
+ * configurations
+ *
+ * @author Stefan Triller - initial contribution
+ *
+ */
+@NonNullByDefault
+public interface MultipleInstanceServiceInfo {
```
sjka: I understand that you need a "dummy" or "marker" instance of a service in order to let the framework know that there _is_ a type of service which can be configured multiple times.

For "singleton" services we are using service properties for the "meta-data" which need to be returned by the below methods. I'm not saying that it is the _better_ method of providing the meta-data, it's just _different_. And IMHO this difference is confusing. Can't they also be exposed as properties of such a dummy/marker service instance? Also, why does it need to implement this interface? Shouldn't a property suffice for that purpose?

bundles/config/org.eclipse.smarthome.config.core/src/main/java/org/eclipse/smarthome/config/core/MultipleInstanceServiceInfo.java
```diff
@@ -0,0 +1,55 @@
...
+     * Get the service.pid of the service that can be created multiple times with different configurations
+     *
+     * @return service.pid of the service
+     */
+    String getServicePID();
+
+    /**
+     * Get the description URI of this service
+     *
+     * @return the URI description of this service
```
sjka: URI description -> config description URI

bundles/config/org.eclipse.smarthome.config.dispatch/src/main/java/org/eclipse/smarthome/config/dispatch/internal/ConfigDispatcher.java
```diff
@@ -157,30 +158,33 @@ private void storeCurrentExclusivePIDList() {
...
         }
-        String pid = pidWithContext.split(SERVICE_CONTEXT_MARKER)[0];
-        String context = pidWithContext.split(SERVICE_CONTEXT_MARKER)[1];
+        String pid = pidWithContext.split(ConfigConstants.SERVICE_CONTEXT_MARKER)[0];
+        String context = pidWithContext.split(ConfigConstants.SERVICE_CONTEXT_MARKER)[1];
 
-        Configuration[] configs = configAdmin
-                .listConfigurations("(&(service.factoryPid=" + pid + ")(" + SERVICE_CONTEXT + "=" + context + "))");
+        Configuration[] configs = configAdmin.listConfigurations(
+                "(&(service.factoryPid=" + pid + ")(" + ConfigConstants.SERVICE_CONTEXT + "=" + context + "))");
```
sjka: `service.factoryPid` -> `ConfigurationAdmin.SERVICE_FACTORYPID`

bundles/io/org.eclipse.smarthome.io.rest.core/src/main/java/org/eclipse/smarthome/io/rest/core/internal/service/ConfigurableServiceResource.java
```diff
@@ -204,20 +223,59 @@ public Response deleteConfiguration(
         }
     }
 
+    private List<ConfigurableServiceDTO> getServicesByFilter(String filter) {
+        List<ConfigurableServiceDTO> services = new ArrayList<>();
+        ServiceReference<?>[] serviceReferences = null;
+        try {
+            serviceReferences = RESTCoreActivator.getBundleContext().getServiceReferences((String) null, filter);
+        } catch (InvalidSyntaxException ex) {
+            logger.error("Cannot get service references, because syntax is invalid: {}", ex.getMessage(), ex);
```
sjka: drop the comma

sjka: how about logging the `filter` string instead of the exception message?

sjka: Does logging the stacktrace _really_ adds some value here?

bundles/test/org.eclipse.smarthome.magic/src/main/java/org/eclipse/smarthome/magic/service/MagicMultiInstanceService.java
triller-telekom: If I omit the `service = ` attribute the services do not offer a service and they do not show up in paper UI...

I have set them to `immediate = true` though.

bundles/config/org.eclipse.smarthome.config.core/src/main/java/org/eclipse/smarthome/config/core/MultipleInstanceServiceInfo.java
triller-telekom: Alright. I agree that its better to have it consistent with the singleton services. So I dropped the new Interface and use the properties on the marker service now.

extensions/ui/org.eclipse.smarthome.ui.paper/web-src/js/controllers/controller.configurableService.dialog.js
```diff
@@ -29,6 +29,27 @@ configurableServiceDialogController.controller('ConfigurableServiceDialogControl
...
 
+    var addESHServiceContextParameter = function(configDescription) {
+        var found = false;
+        angular.forEach(configDescription.parameters, function(parameter) {
+            if (parameter.name === 'esh.servicecontext') {
+                found = true;
+            }
+        })
+        if (!found) {
+            // the service context field is the name of this configuration to help the user to distinguish between
```
sjka: I think you mis-understood me (and I though by now you know how I think about comments in code... 😉 ) 

With a *description*  for a *user*  I was referring to something in the sense of [ConfigDescriptionParameterDTO.description](https://github.com/eclipse/smarthome/blob/4204ce06bb28c28e5f711e720f87ef83beff2e27/bundles/config/org.eclipse.smarthome.config.core/src/main/java/org/eclipse/smarthome/config/core/dto/ConfigDescriptionParameterDTO.java#L33) like you have added [here](https://github.com/eclipse/smarthome/pull/4847/files#diff-d3ca81dfa99af7ad121aea1b1ece5c52R10), i.e. 

```js
var eshServiceContextParameter = {
   ...
   description : 'bla',
   ...
}
```

This is what the user currently sees:

image

and sorry, but that really doesn't give any hint about what that is meant to be!

triller-telekom: Indeed there was a misunderstanding.

I have added the description to the parameter but still think that the comment is also useful.

Note: javascript code is harder to read than java :) And the developer might wonder why we add that parameter there.

sjka:
Thanks!

Issue #3054, opened 2017-02-22
Documentation for config parameter contexts


rodneyrehm:
The changes made to the documentation in regard to parameter contexts in #2168 are missing a few things:

* The newly introduced context "Day of week" was not added to the [XML structure for Configuration Descriptions](https://github.com/qivicon/smarthome/blame/09b115c0738e8117db67baa8dbd66002c7648a1e/docs/documentation/development/bindings/xml-reference.md#L52) and [ConfigDescriptionParameter.getContext()](https://github.com/eclipse/smarthome/blob/6dd1df77136a70ec55da6746be05bf65b78ba15b/bundles/config/org.eclipse.smarthome.config.core/src/main/java/org/eclipse/smarthome/config/core/ConfigDescriptionParameter.java#L331-L347)
* The actual *value* for "Day of week" [is not mentioned](https://github.com/qivicon/smarthome/blame/09b115c0738e8117db67baa8dbd66002c7648a1e/docs/documentation/development/bindings/xml-reference.md#L114), I'd *assume* `weekday`, but [PaperUI include.config.html](https://github.com/eclipse/smarthome/blob/1121ecac6294f5ed396127ad8776862b99578bd4/extensions/ui/org.eclipse.smarthome.ui.paper/web-src/partials/include.config.html#L150) suggests `dayofweek`
* The [Supported Contexts section](https://github.com/qivicon/smarthome/blame/09b115c0738e8117db67baa8dbd66002c7648a1e/docs/documentation/development/bindings/xml-reference.md#L104-L115) is missing a number of values presented in the XML block above.

Furthermore the listed date/time contexts leave questions:

* `date` is supposed to be formatted as `DD:MM:YYYY`, while [ISO 8601](https://en.wikipedia.org/wiki/ISO_8601) specifies it as `YYYY-MM-DD`
* `time` apparently allows integer input (I assume [unix epoch](https://en.wikipedia.org/wiki/Epoch_(reference_date)#Notable_epoch_dates_in_computing)?) or `hh:mm`, while [ISO 8601 Dates](https://en.wikipedia.org/wiki/ISO_8601#Times) allows `hh:mm:ss` and `hh:mm`. Are these UTC or local time zone?
* "Day of week" isn't mentioned in ISO 8601, so the enum looks fine

The following tables are extracted from the docs I created for one of the parameter rendering engines used in Qivicon. They aren't perfect, but should be a good starting point for the ESH docs…

## Parameter Contexts

[...]

PR #4848, opened 2018-01-04
Allow configuration of a broadcast address

mherwege:
address.

See issue [#3054](https://github.com/openhab/openhab2-addons/issues/3054)

Signed-off-by: Mark Herwege <mark.herwege@telenet.be>

### DIFF
--- a/bundles/config/org.eclipse.smarthome.config.core/src/main/java/org/eclipse/smarthome/config/core/net/internal/NetworkConfigOptionProvider.java
+++ b/bundles/config/org.eclipse.smarthome.config.core/src/main/java/org/eclipse/smarthome/config/core/net/internal/NetworkConfigOptionProvider.java
@@ -14,8 +14,8 @@ package org.eclipse.smarthome.config.core.net.internal;
 
 import java.net.Inet4Address;
 import java.net.URI;
+import java.util.ArrayList;
 import java.util.Collection;
-import java.util.List;
 import java.util.Locale;
 import java.util.stream.Collectors;
 import java.util.stream.Stream;
@@ -51,7 +51,7 @@ public class NetworkConfigOptionProvider implements ConfigOptionProvider {
         }
 
         if (param.equals(PARAM_BROADCAST_ADDRESS)) {
-            List<String> broadcastAddrList = NetUtil.getAllBroadcastAddresses();
+            ArrayList<String> broadcastAddrList = new ArrayList<>(NetUtil.getAllBroadcastAddresses());
             broadcastAddrList.add("255.255.255.255");
             return broadcastAddrList.stream().map(a -> new ParameterOption(a, a)).collect(Collectors.toList());
         }
--- a/bundles/config/org.eclipse.smarthome.config.core/src/main/java/org/eclipse/smarthome/config/core/net/internal/NetworkConfigOptionProvider.java
+++ b/bundles/config/org.eclipse.smarthome.config.core/src/main/java/org/eclipse/smarthome/config/core/net/internal/NetworkConfigOptionProvider.java
@@ -15,6 +15,7 @@ package org.eclipse.smarthome.config.core.net.internal;
 import java.net.Inet4Address;
 import java.net.URI;
 import java.util.Collection;
+import java.util.List;
 import java.util.Locale;
 import java.util.stream.Collectors;
 import java.util.stream.Stream;
@@ -35,6 +36,7 @@ import org.osgi.service.component.annotations.Component;
 public class NetworkConfigOptionProvider implements ConfigOptionProvider {
     static final URI CONFIG_URI = URI.create("system:network");
     static final String PARAM_PRIMARY_ADDRESS = "primaryAddress";
+    static final String PARAM_BROADCAST_ADDRESS = "broadcastAddress";
 
     @Override
     public Collection<ParameterOption> getParameterOptions(URI uri, String param, Locale locale) {
@@ -46,6 +48,12 @@ public class NetworkConfigOptionProvider implements ConfigOptionProvider {
             Stream<CidrAddress> ipv4Addresses = NetUtil.getAllInterfaceAddresses().stream()
                     .filter(a -> a.getAddress() instanceof Inet4Address);
             return ipv4Addresses.map(a -> new ParameterOption(a.toString(), a.toString())).collect(Collectors.toList());
+        }
+
+        if (param.equals(PARAM_BROADCAST_ADDRESS)) {
+            List<String> broadcastAddrList = NetUtil.getAllBroadcastAddresses();
+            broadcastAddrList.add("255.255.255.255");
+            return broadcastAddrList.stream().map(a -> new ParameterOption(a, a)).collect(Collectors.toList());
         }
         return null;
     }
--- a/bundles/core/org.eclipse.smarthome.core.test/src/test/java/org/eclipse/smarthome/core/net/NetUtilTest.java
+++ b/bundles/core/org.eclipse.smarthome.core.test/src/test/java/org/eclipse/smarthome/core/net/NetUtilTest.java
@@ -1,5 +1,5 @@
 /**
- * Copyright (c) 2014,2017 Contributors to the Eclipse Foundation
+ * Copyright (c) 2014,2018 Contributors to the Eclipse Foundation
  *
  * See the NOTICE file(s) distributed with this work for additional
  * information regarding copyright ownership.
@@ -21,6 +21,7 @@ import org.junit.Test;
  * Tests for NetUtil class
  *
  * @author Stefan Triller - initial contribution
+ * @author Mark Herwege - added tests for broadcast address
  *
  */
 public class NetUtilTest {
@@ -29,27 +30,46 @@ public class NetUtilTest {
     public void testNetwork() {
         String network = NetUtil.getIpv4NetAddress("192.168.0.1", (short) 24);
         assertThat(network, is("192.168.0.0"));
+        String broadcast = NetUtil.getIpv4NetBroadcastAddress("192.168.0.1", (short) 24);
+        assertThat(broadcast, is("192.168.0.255"));
 
         network = NetUtil.getIpv4NetAddress("192.168.23.5", (short) 24);
         assertThat(network, is("192.168.23.0"));
+        broadcast = NetUtil.getIpv4NetBroadcastAddress("192.168.23.5", (short) 24);
+        assertThat(broadcast, is(" 192.168.23.255"));
 
         network = NetUtil.getIpv4NetAddress("172.16.42.23", (short) 16);
         assertThat(network, is("172.16.0.0"));
+        broadcast = NetUtil.getIpv4NetBroadcastAddress("172.16.42.23", (short) 16);
+        assertThat(broadcast, is("172.16.255.255"));
 
         network = NetUtil.getIpv4NetAddress("10.8.13.5", (short) 8);
         assertThat(network, is("10.0.0.0"));
+        broadcast = NetUtil.getIpv4NetBroadcastAddress("10.8.13.5", (short) 8);
+        assertThat(broadcast, is("10.255.255.255"));
 
         network = NetUtil.getIpv4NetAddress("192.168.5.8", (short) 23);
         assertThat(network, is("192.168.4.0"));
+        broadcast = NetUtil.getIpv4NetBroadcastAddress("192.168.5.8", (short) 23);
+        assertThat(broadcast, is("192.168.5.255"));
 
         network = NetUtil.getIpv4NetAddress("192.168.5.8", (short) 27);
         assertThat(network, is("192.168.5.0"));
+        broadcast = NetUtil.getIpv4NetBroadcastAddress("192.168.5.8", (short) 27);
+        assertThat(broadcast, is("192.168.5.31"));
 
         network = NetUtil.getIpv4NetAddress("192.168.5.8", (short) 29);
         assertThat(network, is("192.168.5.8"));
+        broadcast = NetUtil.getIpv4NetBroadcastAddress("192.168.5.8", (short) 29);
+        assertThat(broadcast, is("192.168.5.15"));
 
         try {
             network = NetUtil.getIpv4NetAddress("192.168.5.8", (short) 33);
+        } catch (IllegalArgumentException iae) {
+            assertThat(iae.getMessage(), is("Netmask '33' is out of bounds (1-32)"));
+        }
+        try {
+            network = NetUtil.getIpv4NetBroadcastAddress("192.168.5.8", (short) 33);
         } catch (IllegalArgumentException iae) {
             assertThat(iae.getMessage(), is("Netmask '33' is out of bounds (1-32)"));
         }
@@ -59,12 +79,27 @@ public class NetUtilTest {
             assertThat(iae.getMessage(), is("IP '192.168.58' is not a valid IPv4 address"));
         }
         try {
+            network = NetUtil.getIpv4NetBroadcastAddress("192.168.58", (short) 24);
+        } catch (IllegalArgumentException iae) {
+            assertThat(iae.getMessage(), is("IP '192.168.58' is not a valid IPv4 address"));
+        }
+        try {
             network = NetUtil.getIpv4NetAddress("SOME_TEXT", (short) 24);
         } catch (IllegalArgumentException iae) {
             assertThat(iae.getMessage(), is("IP 'SOME_TEXT' is not a valid IPv4 address"));
         }
         try {
+            network = NetUtil.getIpv4NetBroadcastAddress("192.168.58", (short) 24);
+        } catch (IllegalArgumentException iae) {
+            assertThat(iae.getMessage(), is("IP '192.168.58' is not a valid IPv4 address"));
+        }
+        try {
             network = NetUtil.getIpv4NetAddress("SOME_TEXT", (short) 42);
+        } catch (IllegalArgumentException iae) {
+            assertThat(iae.getMessage(), is("IP 'SOME_TEXT' is not a valid IPv4 address"));
+        }
+        try {
+            network = NetUtil.getIpv4NetBroadcastAddress("SOME_TEXT", (short) 42);
         } catch (IllegalArgumentException iae) {
             assertThat(iae.getMessage(), is("IP 'SOME_TEXT' is not a valid IPv4 address"));
         }
--- a/bundles/core/org.eclipse.smarthome.core.test/src/test/java/org/eclipse/smarthome/core/net/NetUtilTest.java
+++ b/bundles/core/org.eclipse.smarthome.core.test/src/test/java/org/eclipse/smarthome/core/net/NetUtilTest.java
@@ -97,7 +97,7 @@ public class NetUtilTest {
         try {
             broadcast = NetUtil.getIpv4NetBroadcastAddress("192.168.5.8", (short) 33);
         } catch (IllegalArgumentException iae) {
-            assertThat(iae.getMessage(), is("Netmask '33' is out of bounds (1-32)"));
+            assertThat(iae.getMessage(), is("Prefix '33' is out of bounds (1-32)"));
         }
         try {
             broadcast = NetUtil.getIpv4NetBroadcastAddress("192.168.58", (short) 24);
--- a/bundles/core/org.eclipse.smarthome.core.test/src/test/java/org/eclipse/smarthome/core/net/NetUtilTest.java
+++ b/bundles/core/org.eclipse.smarthome.core.test/src/test/java/org/eclipse/smarthome/core/net/NetUtilTest.java
@@ -30,46 +30,27 @@ public class NetUtilTest {
     public void testNetwork() {
         String network = NetUtil.getIpv4NetAddress("192.168.0.1", (short) 24);
         assertThat(network, is("192.168.0.0"));
-        String broadcast = NetUtil.getIpv4NetBroadcastAddress("192.168.0.1", (short) 24);
-        assertThat(broadcast, is("192.168.0.255"));
 
         network = NetUtil.getIpv4NetAddress("192.168.23.5", (short) 24);
         assertThat(network, is("192.168.23.0"));
-        broadcast = NetUtil.getIpv4NetBroadcastAddress("192.168.23.5", (short) 24);
-        assertThat(broadcast, is(" 192.168.23.255"));
 
         network = NetUtil.getIpv4NetAddress("172.16.42.23", (short) 16);
         assertThat(network, is("172.16.0.0"));
-        broadcast = NetUtil.getIpv4NetBroadcastAddress("172.16.42.23", (short) 16);
-        assertThat(broadcast, is("172.16.255.255"));
 
         network = NetUtil.getIpv4NetAddress("10.8.13.5", (short) 8);
         assertThat(network, is("10.0.0.0"));
-        broadcast = NetUtil.getIpv4NetBroadcastAddress("10.8.13.5", (short) 8);
-        assertThat(broadcast, is("10.255.255.255"));
 
         network = NetUtil.getIpv4NetAddress("192.168.5.8", (short) 23);
         assertThat(network, is("192.168.4.0"));
-        broadcast = NetUtil.getIpv4NetBroadcastAddress("192.168.5.8", (short) 23);
-        assertThat(broadcast, is("192.168.5.255"));
 
         network = NetUtil.getIpv4NetAddress("192.168.5.8", (short) 27);
         assertThat(network, is("192.168.5.0"));
-        broadcast = NetUtil.getIpv4NetBroadcastAddress("192.168.5.8", (short) 27);
-        assertThat(broadcast, is("192.168.5.31"));
 
         network = NetUtil.getIpv4NetAddress("192.168.5.8", (short) 29);
         assertThat(network, is("192.168.5.8"));
-        broadcast = NetUtil.getIpv4NetBroadcastAddress("192.168.5.8", (short) 29);
-        assertThat(broadcast, is("192.168.5.15"));
 
         try {
             network = NetUtil.getIpv4NetAddress("192.168.5.8", (short) 33);
-        } catch (IllegalArgumentException iae) {
-            assertThat(iae.getMessage(), is("Netmask '33' is out of bounds (1-32)"));
-        }
-        try {
-            network = NetUtil.getIpv4NetBroadcastAddress("192.168.5.8", (short) 33);
         } catch (IllegalArgumentException iae) {
             assertThat(iae.getMessage(), is("Netmask '33' is out of bounds (1-32)"));
         }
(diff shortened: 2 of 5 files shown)


bundles/core/org.eclipse.smarthome.core/src/main/java/org/eclipse/smarthome/core/net/NetUtil.java
```diff
@@ -183,11 +184,42 @@ private String getFirstLocalIPv4Address() {
         return broadcastAddresses;
     }
 
+    @Override
+    public String getPrimaryBroadcastAddress() {
```
triller-telekom: Could you please some test cases for this method in https://github.com/eclipse/smarthome/blob/master/bundles/core/org.eclipse.smarthome.core.test/src/test/java/org/eclipse/smarthome/core/net/NetUtilTest.java ?

`byte` magic is always tricky to debug and easy to break, so I think it would be good to have some test cases for that.

bundles/core/org.eclipse.smarthome.core/src/main/java/org/eclipse/smarthome/core/net/NetUtil.java
```diff
@@ -183,11 +184,42 @@ private String getFirstLocalIPv4Address() {
...
+
     /**
+     * Deprecated: Please use getPrimaryBroadcastAddress()
+     *
      * Get the first candidate for a broadcast address
      *
-     * @return broadcast address, null of no broadcast address is found
+     * @return broadcast address, null if no broadcast address is found
      */
+    @Deprecated
```
triller-telekom: Can you please add a comment that one should now use `getPrimaryBroadcastAddress` instead?

mherwege: The comment is in the java doc a few lines above.

bundles/core/org.eclipse.smarthome.core/src/main/java/org/eclipse/smarthome/core/net/NetUtil.java
mherwege: I will look into this. I have not done test cases so far, so need to get over the learning curve.

mherwege:
@triller-telekom I believe I have now addressed your comments. I have made the broadcast address a parameter. I don't know how to expose that in PaperUI though.

bundles/core/org.eclipse.smarthome.core/src/main/java/org/eclipse/smarthome/core/net/NetUtil.java
mherwege: I added test cases. I split out the main calculation logic in a static method, so the logic can be tested more easily.

bundles/core/org.eclipse.smarthome.core/src/main/java/org/eclipse/smarthome/core/net/NetUtil.java
triller-telekom: Sorry, my fault...

bundles/core/org.eclipse.smarthome.core.test/src/test/java/org/eclipse/smarthome/core/net/NetUtilTest.java
```diff
@@ -29,45 +30,79 @@
     public void testNetwork() {
         String network = NetUtil.getIpv4NetAddress("192.168.0.1", (short) 24);
         assertThat(network, is("192.168.0.0"));
+        String broadcast = NetUtil.getIpv4NetBroadcastAddress("192.168.0.1", (short) 24);
+        assertThat(broadcast, is("192.168.0.255"));
 
         network = NetUtil.getIpv4NetAddress("192.168.23.5", (short) 24);
         assertThat(network, is("192.168.23.0"));
+        broadcast = NetUtil.getIpv4NetBroadcastAddress("192.168.23.5", (short) 24);
+        assertThat(broadcast, is(" 192.168.23.255"));
```
triller-telekom: There is a space before the address, which is why the test case fails in travis (and locally) ;-)

bundles/core/org.eclipse.smarthome.core.test/src/test/java/org/eclipse/smarthome/core/net/NetUtilTest.java
```diff
@@ -29,45 +30,79 @@
     public void testNetwork() {
         String network = NetUtil.getIpv4NetAddress("192.168.0.1", (short) 24);
         assertThat(network, is("192.168.0.0"));
+        String broadcast = NetUtil.getIpv4NetBroadcastAddress("192.168.0.1", (short) 24);
```
triller-telekom: Thank you for creating these test cases!

Since they test something different than the ip address, I guess it would be better to separate them in an extra method.

Please move your lines into a new method:
```
@Test
public voidTestBroadcastAddress() {
//your code
}
```

The `@Test` annotation tells junit that it should treat this method as a test case and execute it.

mherwege: Done.

bundles/core/org.eclipse.smarthome.core.test/src/test/java/org/eclipse/smarthome/core/net/NetUtilTest.java
mherwege: Indeed, thank you for that. I struggle getting the tests to run locally, I get: Required plug-in 'org.eclipse.jdt.junit4.runtime' could not be found.

mherwege:
@triller-telekom I did not manage to get the test to run locally. So there might still be something wrong with it. Thank you for your support so far.

bundles/core/org.eclipse.smarthome.core/ESH-INF/config/networkinterface.xml
```diff
@@ -10,6 +10,12 @@
 			<description>A subnet (e.g. 192.168.1.0/24) <!-- or an IP address (e.g. 192.168.1.5) --></description>
 			<limitToOptions>true</limitToOptions>
 		</parameter>
+        <parameter name="broadcastAddress" type="text">
```
maggu2810: Have you applied the formatter (remember the Eclipse IDE does not support save actions and so auto formatting on save for XML files)?
Shouldn't XML files be indented by tabs?

bundles/core/org.eclipse.smarthome.core/src/main/java/org/eclipse/smarthome/core/net/NetUtil.java
```diff
@@ -40,18 +41,20 @@
  * @author Mark Herwege - Added methods to find broadcast address(es)
  * @author Stefan Triller - Converted to OSGi service with primary ipv4 conf
  */
-@Component(configurationPid = "org.eclipse.smarthome.network", property = { "service.pid=org.eclipse.smarthome.network",
-        "service.config.description.uri=system:network", "service.config.label=Network Settings",
-        "service.config.category=system" })
+@Component(immediate = true, configurationPid = "org.eclipse.smarthome.network", property = {
```
maggu2810: How is the `immediate` change related to the topic of this change?
A utility method does IMHO only be necessary if someone uses it.
Do you add some magic action that does need to be executed also if there is no consumer of this service?

triller-telekom: @maggu2810 This change was actually suggested by me. I was testing the paper UI configuration and noticed that the `modified` method of the `NetUtil` class was not called due to the fact that the service was not started yet. I was asking @kaikreuzer and @sjka if there is any reason to not start it immediately and they also said that there is no reason to not start it.

And since @mherwege was touching the code anyway I thought it won't hurt if he includes the change in this PR.

bundles/core/org.eclipse.smarthome.core.test/src/test/java/org/eclipse/smarthome/core/net/NetUtilTest.java
triller-telekom: You cannot run the test as a plain JUnit test. You have to right click on https://github.com/eclipse/smarthome/blob/master/bundles/core/org.eclipse.smarthome.core.test/CoreTests.launch

and run it as a `CoreTests` which will then run all core tests (including this one) as a J-Unit plug-in test. If that still fails you can edit this run configuration, click on the plugin tab and select the missing plugins there.

bundles/core/org.eclipse.smarthome.core/src/main/java/org/eclipse/smarthome/core/net/NetUtil.java
mherwege: @triller-telekom Thank you for the clarification.

bundles/core/org.eclipse.smarthome.core/ESH-INF/config/networkinterface.xml
mherwege: Done.

mherwege:
@triller-telekom I am still stuck trying to run the test. I believe there is still something wrong, but I don't manage to run it locally. I am getting the following error: Required plug-in 'org.eclipse.jdt.junit4.runtime' could not be found. I am running Eclipse Oxygen.2 Release (4.7.2). Let me know if you have suggestions.

triller-telekom:
@mherwege Did you follow my suggestions in https://github.com/eclipse/smarthome/pull/4848#discussion_r160086107 ?

I am able to run it like that...

But I get the following error:

```
java.lang.AssertionError: 
Expected: is "Netmask '33' is out of bounds (1-32)"
     but: was "Prefix '33' is out of bounds (1-32)"
	at org.junit.Assert.assertThat(Assert.java:956)
	at org.junit.Assert.assertThat(Assert.java:923)
	at org.eclipse.smarthome.core.net.NetUtilTest.testBroadcastAddress(NetUtilTest.java:100)
	at sun.reflect.NativeMethodAccessorImpl.invoke0(Native Method)
	at sun.reflect.NativeMethodAccessorImpl.invoke(NativeMethodAccessorImpl.java:62)
	at sun.reflect.DelegatingMethodAccessorImpl.invoke(DelegatingMethodAccessorImpl.java:43)
	at java.lang.reflect.Method.invoke(Method.java:498)
	at org.junit.runners.model.FrameworkMethod$1.runReflectiveCall(FrameworkMethod.java:50)
	at org.junit.internal.runners.model.ReflectiveCallable.run(ReflectiveCallable.java:12)
	at org.junit.runners.model.FrameworkMethod.invokeExplosively(FrameworkMethod.java:47)
	at org.junit.internal.runners.statements.InvokeMethod.evaluate(InvokeMethod.java:17)
	at org.junit.runners.ParentRunner.runLeaf(ParentRunner.java:325)
	at org.junit.runners.BlockJUnit4ClassRunner.runChild(BlockJUnit4ClassRunner.java:78)
	at org.junit.runners.BlockJUnit4ClassRunner.runChild(BlockJUnit4ClassRunner.java:57)
	at org.junit.runners.ParentRunner$3.run(ParentRunner.java:290)
	at org.junit.runners.ParentRunner$1.schedule(ParentRunner.java:71)
	at org.junit.runners.ParentRunner.runChildren(ParentRunner.java:288)
	at org.junit.runners.ParentRunner.access$000(ParentRunner.java:58)
	at org.junit.runners.ParentRunner$2.evaluate(ParentRunner.java:268)
	at org.junit.runners.ParentRunner.run(ParentRunner.java:363)
	at org.eclipse.jdt.internal.junit4.runner.JUnit4TestReference.run(JUnit4TestReference.java:86)
	at org.eclipse.jdt.internal.junit.runner.TestExecution.run(TestExecution.java:38)
	at org.eclipse.jdt.internal.junit.runner.RemoteTestRunner.runTests(RemoteTestRunner.java:459)
	at org.eclipse.jdt.internal.junit.runner.RemoteTestRunner.runTests(RemoteTestRunner.java:678)
	at org.eclipse.jdt.internal.junit.runner.RemoteTestRunner.run(RemoteTestRunner.java:382)
	at org.eclipse.pde.internal.junit.runtime.RemotePluginTestRunner.main(RemotePluginTestRunner.java:181)
	at org.eclipse.pde.internal.junit.runtime.CoreTestApplication.run(CoreTestApplication.java:24)
	at sun.reflect.NativeMethodAccessorImpl.invoke0(Native Method)
	at sun.reflect.NativeMethodAccessorImpl.invoke(NativeMethodAccessorImpl.java:62)
	at sun.reflect.DelegatingMethodAccessorImpl.invoke(DelegatingMethodAccessorImpl.java:43)
	at java.lang.reflect.Method.invoke(Method.java:498)
	at org.eclipse.equinox.internal.app.EclipseAppContainer.callMethodWithException(EclipseAppContainer.java:587)
	at org.eclipse.equinox.internal.app.EclipseAppHandle.run(EclipseAppHandle.java:198)
	at org.eclipse.core.runtime.internal.adaptor.EclipseAppLauncher.runApplication(EclipseAppLauncher.java:134)
	at org.eclipse.core.runtime.internal.adaptor.EclipseAppLauncher.start(EclipseAppLauncher.java:104)
	at org.eclipse.core.runtime.adaptor.EclipseStarter.run(EclipseStarter.java:388)
	at org.eclipse.core.runtime.adaptor.EclipseStarter.run(EclipseStarter.java:243)
	at sun.reflect.NativeMethodAccessorImpl.invoke0(Native Method)
	at sun.reflect.NativeMethodAccessorImpl.invoke(NativeMethodAccessorImpl.java:62)
	at sun.reflect.DelegatingMethodAccessorImpl.invoke(DelegatingMethodAccessorImpl.java:43)
	at java.lang.reflect.Method.invoke(Method.java:498)
	at org.eclipse.equinox.launcher.Main.invokeFramework(Main.java:653)
	at org.eclipse.equinox.launcher.Main.basicRun(Main.java:590)
	at org.eclipse.equinox.launcher.Main.run(Main.java:1499)
	at org.eclipse.equinox.launcher.Main.main(Main.java:1472)


```

mherwege:
OK, I fixed the error in the test. It hopefully runs through fine now. I still cannot start the test with your suggestion, so there must be something wrong with my setup. I will investigate further when I have more time. At least, I hope the test for this PR does not contain errors anymore now.

triller-telekom:
The tests are running fine now. Maybe you are lacking a plug-in. If you edit the run configuration of the `CoreTests` you can click on "validate plug-ins" and then on "add required-plugins" in case there a re some missing.

mherwege:
@triller-telekom I tried all that, but didn't get it to run.
I get the following in the test run log: 
```
!SESSION 2018-01-08 09:28:51.181 -----------------------------------------------
eclipse.buildId=unknown
java.version=1.8.0_121
java.vendor=Oracle Corporation
BootLoader constants: OS=win32, ARCH=x86_64, WS=win32, NL=en_US
Framework arguments:  -version 3 -port 63052 -testLoaderClass org.eclipse.jdt.internal.junit4.runner.JUnit4TestLoader -loaderpluginname org.eclipse.jdt.junit4.runtime -testNameFile C:\Users\MHERWEGE\AppData\Local\Temp\testNames9002395741831533917.txt -application org.eclipse.pde.junit.runtime.coretestapplication -testpluginname org.eclipse.smarthome.core
Command-line arguments:  -os win32 -ws win32 -arch x86_64 -consoleLog -version 3 -port 63052 -testLoaderClass org.eclipse.jdt.internal.junit4.runner.JUnit4TestLoader -loaderpluginname org.eclipse.jdt.junit4.runtime -testNameFile C:\Users\MHERWEGE\AppData\Local\Temp\testNames9002395741831533917.txt -application org.eclipse.pde.junit.runtime.coretestapplication -data C:\Users\MHERWEGE\openhab2-master\ws/../junit-workspace -dev file:C:/Users/MHERWEGE/openhab2-master/ws/.metadata/.plugins/org.eclipse.pde.core/pde-junit/dev.properties -os win32 -ws win32 -arch x86_64 -consoleLog -testpluginname org.eclipse.smarthome.core

!ENTRY org.eclipse.osgi 4 0 2018-01-08 09:28:52.829
!MESSAGE Error starting bundle: initial@reference:file:plugins/org.eclipse.jetty.osgi.alpn.fragment_9.3.22.v20171030.jar
!STACK 0
org.osgi.framework.BundleException: Invalid operation on a fragment.
	at org.eclipse.osgi.container.Module.checkFragment(Module.java:520)
	at org.eclipse.osgi.container.Module.start(Module.java:408)
	at org.eclipse.osgi.internal.framework.EquinoxBundle.start(EquinoxBundle.java:383)
	at org.eclipse.core.runtime.adaptor.EclipseStarter.startBundle(EclipseStarter.java:1055)
	at org.eclipse.core.runtime.adaptor.EclipseStarter.startBundles(EclipseStarter.java:1048)
	at org.eclipse.core.runtime.adaptor.EclipseStarter.loadBasicBundles(EclipseStarter.java:592)
	at org.eclipse.core.runtime.adaptor.EclipseStarter.startup(EclipseStarter.java:319)
	at org.eclipse.core.runtime.adaptor.EclipseStarter.run(EclipseStarter.java:239)
	at sun.reflect.NativeMethodAccessorImpl.invoke0(Native Method)
	at sun.reflect.NativeMethodAccessorImpl.invoke(NativeMethodAccessorImpl.java:62)
	at sun.reflect.DelegatingMethodAccessorImpl.invoke(DelegatingMethodAccessorImpl.java:43)
	at java.lang.reflect.Method.invoke(Method.java:498)
	at org.eclipse.equinox.launcher.Main.invokeFramework(Main.java:653)
	at org.eclipse.equinox.launcher.Main.basicRun(Main.java:590)
	at org.eclipse.equinox.launcher.Main.run(Main.java:1499)
	at org.eclipse.equinox.launcher.Main.main(Main.java:1472)

!ENTRY org.eclipse.osgi 4 0 2018-01-08 09:28:54.965
!MESSAGE Bundle org.eclipse.jetty.osgi.alpn.fragment_9.3.22.v20171030 [9] is not active.

!ENTRY org.eclipse.osgi 4 0 2018-01-08 09:28:55.136
!MESSAGE Application error
!STACK 1
java.lang.AbstractMethodError: org.eclipse.jdt.internal.junit4.runner.JUnit4TestLoader.loadTests([Ljava/lang/Class;Ljava/lang/String;[Ljava/lang/String;Lorg/eclipse/jdt/internal/junit/runner/RemoteTestRunner;)[Lorg/eclipse/jdt/internal/junit/runner/ITestReference;
	at org.eclipse.jdt.internal.junit.runner.RemoteTestRunner.runTests(RemoteTestRunner.java:444)
	at org.eclipse.jdt.internal.junit.runner.RemoteTestRunner.runTests(RemoteTestRunner.java:678)
	at org.eclipse.jdt.internal.junit.runner.RemoteTestRunner.run(RemoteTestRunner.java:382)
	at org.eclipse.pde.internal.junit.runtime.RemotePluginTestRunner.main(RemotePluginTestRunner.java:66)
	at org.eclipse.pde.internal.junit.runtime.CoreTestApplication.run(CoreTestApplication.java:24)
	at sun.reflect.NativeMethodAccessorImpl.invoke0(Native Method)
	at sun.reflect.NativeMethodAccessorImpl.invoke(NativeMethodAccessorImpl.java:62)
	at sun.reflect.DelegatingMethodAccessorImpl.invoke(DelegatingMethodAccessorImpl.java:43)
	at java.lang.reflect.Method.invoke(Method.java:498)
	at org.eclipse.equinox.internal.app.EclipseAppContainer.callMethodWithException(EclipseAppContainer.java:587)
	at org.eclipse.equinox.internal.app.EclipseAppHandle.run(EclipseAppHandle.java:198)
	at org.eclipse.core.runtime.internal.adaptor.EclipseAppLauncher.runApplication(EclipseAppLauncher.java:134)
	at org.eclipse.core.runtime.internal.adaptor.EclipseAppLauncher.start(EclipseAppLauncher.java:104)
	at org.eclipse.core.runtime.adaptor.EclipseStarter.run(EclipseStarter.java:388)
	at org.eclipse.core.runtime.adaptor.EclipseStarter.run(EclipseStarter.java:243)
	at sun.reflect.NativeMethodAccessorImpl.invoke0(Native Method)
	at sun.reflect.NativeMethodAccessorImpl.invoke(NativeMethodAccessorImpl.java:62)
	at sun.reflect.DelegatingMethodAccessorImpl.invoke(DelegatingMethodAccessorImpl.java:43)
	at java.lang.reflect.Method.invoke(Method.java:498)
	at org.eclipse.equinox.launcher.Main.invokeFramework(Main.java:653)
	at org.eclipse.equinox.launcher.Main.basicRun(Main.java:590)
	at org.eclipse.equinox.launcher.Main.run(Main.java:1499)
	at org.eclipse.equinox.launcher.Main.main(Main.java:1472)
```

bundles/config/org.eclipse.smarthome.config.core/src/main/java/org/eclipse/smarthome/config/core/net/internal/NetworkConfigOptionProvider.java
```diff
@@ -47,6 +49,12 @@
                     .filter(a -> a.getAddress() instanceof Inet4Address);
             return ipv4Addresses.map(a -> new ParameterOption(a.toString(), a.toString())).collect(Collectors.toList());
         }
+
+        if (param.equals(PARAM_BROADCAST_ADDRESS)) {
+            List<String> broadcastAddrList = NetUtil.getAllBroadcastAddresses();
+            broadcastAddrList.add("255.255.255.255");
```
triller-telekom: I just stumbled across this address again and by looking up what it means on wikipedia I found:

> It is the broadcast address of the zero network or 0.0.0.0, which in Internet Protocol standards stands for this network, i.e. the local network. Transmission to this address is limited by definition, in that it is never forwarded by the routers connecting the local network to other networks.

So by setting this we would be back to "random" because "this network" is not really defined if you have multiple network interfaces, right? So I guess we should remove this one.

bundles/core/org.eclipse.smarthome.core/src/main/java/org/eclipse/smarthome/core/net/NetUtil.java
maggu2810: There are services that brings some functionality on its own without any consumers. This ones should be started also if no other one reference this service. This ones should use `immediate = true`.
There are other services that provides some function for consumers. They do not need to be activated if no one uses it. This ones does not need to be activated immediately.

Surely we could set immediate for stuff regardless if it is used or not. Perhaps we already know, it will be used, but then there is still no need to set it explicit, because the framework will activate it if used.

So, I don't understand why we should set immediate here? "It makes no difference because the core will depend on that service" shouldn't be an argument. Shouldn't a component declaration be independent of its special use case and shouldn't we leave it up to the framework to do it correctly?

Or do I miss the point here and modified needs to be called without any consumer available?

bundles/config/org.eclipse.smarthome.config.core/src/main/java/org/eclipse/smarthome/config/core/net/internal/NetworkConfigOptionProvider.java
mherwege: It depends on the OS. Windows will broadcast to one interface. I believe that is true for all BSD derived systems, and windows got its TCP stack from BSD. My RPi broadcasts on all interfaces, and from what I know that would be the case for Linux. This has nothing to do with routing, as we are talking about multipe network interfaces on the same host. The standard does not define if each interface needs to broadcast in a multi-homed system, so it is implemented differently in different OS's. I added it because it allows an easy multipe interface broadcast on some systems (but sure not all). It is also the behaviour most bindings using broadcast to 255.255.255.255 currently have.
The better, OS independent, way is indeed not to use this. The default with this PR will be to use the broadcast address of the primary IP (with fallback to first if not defined). Secondly, selecting a value other than 255.255.255.255 here will broadcast to that specific interface. Then 255.255.255.255 is there as a last resort to support old behaviour with minimal change.
Ultimately, the best fix for broadcast to all interfaces would be to iterate over them. But that has a bigger impact on the bindings that use it than what this PR tries to do. @caffineehacker has a more complete solution in mind for this, adding methods to take care of the broadcasts and response callbacks. My PR was meant to be a minimum impact, but considerable improvement, making the behaviour more predictable, but leaving the possibility to still have the old behaviour.

triller-telekom: @mherwege Thank you for the clarification.

bundles/core/org.eclipse.smarthome.core/src/main/java/org/eclipse/smarthome/core/net/NetUtil.java
triller-telekom: If you are really sure that this service will always be started without the `immediate = true`, I am fine with removing this again.

@mherwege: Sorry about the noise, but could you please revert this?

bundles/core/org.eclipse.smarthome.core/src/main/java/org/eclipse/smarthome/core/net/NetUtil.java
```diff
@@ -183,11 +194,50 @@ private String getFirstLocalIPv4Address() {
...
+            broadcastAddress = getFirstIpv4BroadcastAddress();
+            LOGGER.warn(
+                    "Could not find broadcast address of primary IP, using broadcast address {} of first interface instead",
+                    broadcastAddress);
+        }
+        return broadcastAddress;
+    }
+
     /**
+     * Deprecated: Please use the NetworkAddressService with getConfiguredBroadcastAddress()
```
kaikreuzer: Please use the `@deprecated` annotation

bundles/config/org.eclipse.smarthome.config.core/src/main/java/org/eclipse/smarthome/config/core/net/internal/NetworkConfigOptionProvider.java
```diff
@@ -47,6 +49,12 @@
                     .filter(a -> a.getAddress() instanceof Inet4Address);
             return ipv4Addresses.map(a -> new ParameterOption(a.toString(), a.toString())).collect(Collectors.toList());
         }
+
+        if (param.equals(PARAM_BROADCAST_ADDRESS)) {
+            List<String> broadcastAddrList = NetUtil.getAllBroadcastAddresses();
```
kaikreuzer: You are lucky here that `getAllBroadcastAddresses` returns a modifiable list, otherwise your next line would fail. To make sure that your code survives potential refactorings of `NetUtil. getAllBroadcastAddresses()`, better construct your own List here.

kaikreuzer:
Perfect, thank you!

bundles/core/org.eclipse.smarthome.core/src/main/java/org/eclipse/smarthome/core/net/NetUtil.java
maggu2810: Seems we don't matter. Already merged.

mherwege: @triller-telekom @maggu2810 I missed this when doing the last changes requested by @Kai. Do you expect me to create a PR to revert this change?

kaikreuzer: ... and I didn't notice that it was missing - didn't mean to ignore it, sorry.
@mherwege, yes, please just create a small new PR for it - thanks!

PR #4854, opened 2018-01-05
switched to groovy snapshot repository

sjka:
Not ideal, but at least it exists...


kaikreuzer:
I remember that we often had issues with switching groovy version/repo (build/IDE/etc.).
Did you check whether this runs smooth?
Apart from this: Are you sure that the repo is permanently removed? That's really bad as it would mean that our latest 0.9.0 release build won't work anymore either. :-(

sjka:
It worked for the projects where I'm currently working on, but apparently not for all (as Travis just pointed out). 
No, I'm not sure about anything there.

sjka:
I opened https://github.com/groovy/groovy-eclipse/issues/406 to raise awareness

PR #4857, opened 2018-01-05
Basic UI with long polling update: management of dynamic widget visibility

### LINKED ISSUE #4558: Basic UI with Edge: change of visibility not dynamically taken into account
Using Microsoft Edge, the change of visibility of a sitemap element is not dynamically taken into account. You have to reopen the current page.
Of course, it is working in Firefox or Chrome.

lolodomo:
Fixes #4558

Cleanup code for the long polling update method with dataModelLegacy suppressed
and common code created in AbstractChangeListener

Signed-off-(by: Laurent Garnier <lg.hc@free.fr>

### DIFF
--- a/extensions/ui/org.eclipse.smarthome.ui.basic/web-src/smarthome.js
+++ b/extensions/ui/org.eclipse.smarthome.ui.basic/web-src/smarthome.js
@@ -1587,24 +1587,8 @@
 
 		_t.initControls = function() {
 			smarthome.dataModel = {};
-			smarthome.dataModelLegacy = {};
 
 			function appendControl(control) {
-				// dataModelLegacy keeps item → widgets binding for
-				// long-polling event listener
-				if (
-					(smarthome.dataModelLegacy[control.item] === undefined) ||
-					(smarthome.dataModelLegacy[control.item].widgets === undefined)
-				) {
-					if (control.item !== undefined) {
-						smarthome.dataModelLegacy[control.item] = { widgets: [] };
-					}
-				}
-
-				if (control.item !== undefined) {
-					smarthome.dataModelLegacy[control.item].widgets.push(control);
-				}
-
 				smarthome.dataModel[control.id] = control;
 			}
 
@@ -1733,6 +1717,47 @@
 
 			return title;
 		};
+
+		this.updateWidget = function(widget, update) {
+			var
+				value = this.extractValueFromLabel(update.label),
+				makeVisible = false;
+
+			if (widget.visible !== update.visibility) {
+				makeVisible = update.visibility;
+				smarthome.UI.layoutChangeProxy.push({
+					widget: widget,
+					visibility: update.visibility
+				});
+			}
+
+			if (makeVisible || update.state !== "NULL") {
+				if (value === null) {
+					value = update.state;
+				}
+				widget.setValue(smarthome.UI.escapeHtml(value), update.state, update.visibility);
+			}
+
+			[{
+				apply: widget.setLabel,
+				data: update.label,
+				fallback: null
+			}, {
+				apply: widget.setLabelColor,
+				data: update.labelcolor,
+				fallback: ""
+			}, {
+				apply: widget.setValueColor,
+				data: update.valuecolor,
+				fallback: ""
+			}].forEach(function(e) {
+				if (e.data !== undefined) {
+					e.apply(e.data);
+				} else if (e.fallback !== null) {
+					e.apply(e.fallback);
+				}
+			});
+		};
 	}
 
 	function ChangeListenerEventsource(subscribeLocation) {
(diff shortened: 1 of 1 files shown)


lolodomo:
It is only in WIP step only because I don't want an immediate merge and introduce a conflict with antoher PR having a higher priority.

@resetnow : it is purely frontend. It is working very well. You can already review it if you have time.

lolodomo:
Tested with Microsoft Edge.

extensions/ui/org.eclipse.smarthome.ui.basic/web-src/smarthome.js
```diff
@@ -1733,6 +1717,47 @@
 
 			return title;
 		};
+
+		this.updateWidget = function(widget, visibility, state, label, labelcolor, valuecolor) {
```
resetnow: would be better to make this an object:

```
function(widget, update)

// update
// {
//     state: ...,
//     label: ...,
//     And so on
// }

File: extensions/ui/org.eclipse.smarthome.ui.basic/web-src/smarthome.js [hunk 2]
```diff
@@ -1845,53 +1835,50 @@
 
 			function walkWidgets(widgets) {
 				widgets.forEach(function(widget) {
-					if (widget.item === undefined) {
+					if (widget.widgetId === undefined || smarthome.dataModel[widget.widgetId] === undefined) {
```
resetnow: linebreak between comparisons as line is too long

extensions/ui/org.eclipse.smarthome.ui.basic/web-src/smarthome.js
```diff
@@ -1845,53 +1835,50 @@
...
-						value = _t.extractValueFromLabel(widget.label),
-						visibility = widget.visibility,
-						labelcolor = widget.labelcolor,
-						valuecolor = widget.valuecolor;
-
-					if (value === null) {
-						value = state;
-					}
+						w = smarthome.dataModel[widget.widgetId],
+						state = "NULL";
```
resetnow: perhaps `null` is better?

extensions/ui/org.eclipse.smarthome.ui.basic/web-src/smarthome.js
lolodomo: Fixed

extensions/ui/org.eclipse.smarthome.ui.basic/web-src/smarthome.js
lolodomo: Done

extensions/ui/org.eclipse.smarthome.ui.basic/web-src/smarthome.js
lolodomo: If you look at the previous code, there was already a test again "NULL" and I think it was probably useful and I don't want to break that. If I add null here, I would have to handle null and "NULL".
I prefer a unique test against "NULL".

lolodomo:
@resetnow: PR rebased and your comments considered. Review and merge can be restarted.

lolodomo:
@resetnow : could you please have a final look so that it could be merged ?

I would like to switch to the next step (the PR that in in WIP stage).

extensions/ui/org.eclipse.smarthome.ui.basic/web-src/smarthome.js
```diff
@@ -1852,53 +1856,64 @@
...
-						value = _t.extractValueFromLabel(widget.label),
-						visibility = widget.visibility,
-						labelcolor = widget.labelcolor,
-						valuecolor = widget.valuecolor;
-
-					if (value === null) {
-						value = state;
-					}
+						w = smarthome.dataModel[widget.widgetId],
+						state = "NULL",
```
resetnow: Can this ever happen?

lolodomo: Normally a state is "NULL" when the item is not yet initialized.

resetnow:
@lolodomo could you please point out the PR where the visibility property was added to REST sitemap JSON response? I'm trying to understand whether my problems with testing are caused by incorrect launch configuration and/or build used.

lolodomo:
The visibility property is not provided by the REST API. The REST API returns only visible widgets.

The visibility property is provided in stemap SSE.

resetnow:
In this case what does `w.visible` refer to in `ChangeListenerLongpolling`? Have you tried this patch with default demo sitemap?

lolodomo:
Yes tested with an enhanced demo sitemap to cover different test cases.
w.visible refer to _t.visible set in Control function.

PR #4861, opened 2018-01-06
Update WallSwitch and add RemoteControl

cdjackson:
This clarifies that wall switches are not just for binary devices but could (for example) also be for dimmers.
It also adds a RemoteControl for portable / hand-held controllers.

### DIFF
--- a/docs/_data/categories_thing.csv
+++ b/docs/_data/categories_thing.csv
@@ -18,12 +18,13 @@ PowerOutlet,"Small devices to be plugged into a power socket in a wall which sti
 Projector,"Devices that project a picture somewhere",projector.png
 RadiatorControl,"Controls on radiators used to heat up rooms",
 Receiver,"Audio/Video receivers, i.e. radio receivers, satelite or cable receivers, recorders, etc.",receiver.png
+RemoteControl,"Any portable or hand-held device that controls the status of something, e.g. remote control, keyfob etc."
 Screen,"Devices that are able to show a picture",screen.png
 Sensor,"Device used to measure something",
 Siren,"Siren used by Alarm systems",siren.png
 SmokeDetector,"Smoke detectors",
 Speaker,"Devices that are able to play sounds",
-WallSwitch,"Any device attached to the wall that controls a binary status of something, for ex. a light switch",wallswitch.png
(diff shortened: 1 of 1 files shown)


docs/_data/categories_thing.csv
```diff
@@ -18,12 +18,13 @@ PowerOutlet,"Small devices to be plugged into a power socket in a wall which sti
...
 RadiatorControl,"Controls on radiators used to heat up rooms",
 Receiver,"Audio/Video receivers, i.e. radio receivers, satelite or cable receivers, recorders, etc.",receiver.png
+RemoteControl,"Any portable or hand-held device that controls the status of something, eg remote control, keyfob etc."
 Screen,"Devices that are able to show a picture",screen.png
 Sensor,"Device used to measure something",
 Siren,"Siren used by Alarm systems",siren.png
 SmokeDetector,"Smoke detectors",
 Speaker,"Devices that are able to play sounds",
-WallSwitch,"Any device attached to the wall that controls a binary status of something, for ex. a light switch",wallswitch.png
+WallSwitch,"Any device attached to the wall that controls the status of something, for ex. a light switch, dimmer",wallswitch.png
```
triller-telekom: Should we keep an additional entity that represents a "plain switch" i.e. something that only has a binary state? Or do we want to discard that?

kaikreuzer: What would be the purpose of it (or rather would it "look" anyhow different when being depicted)?

docs/_data/categories_thing.csv
```diff
@@ -18,12 +18,13 @@ PowerOutlet,"Small devices to be plugged into a power socket in a wall which sti
 Projector,"Devices that project a picture somewhere",projector.png
 RadiatorControl,"Controls on radiators used to heat up rooms",
 Receiver,"Audio/Video receivers, i.e. radio receivers, satelite or cable receivers, recorders, etc.",receiver.png
+RemoteControl,"Any portable or hand-held device that controls the status of something, eg remote control, keyfob etc."
```
maggu2810: As we are using "i.e." we should perhaps change "eg" to "e.g."

docs/_data/categories_thing.csv
triller-telekom: You are right, a "binary switch" could be seen as a special version of a general switch with multiple options and thus it would probably be the same icon.

PR #4864, opened 2018-01-06
Sonos: adjust log levels

### LINKED ISSUE #4850: [sonos] media channel WARN messages while thing is not added
The sonos binding shows warnings in the log while the corresponding thing is not yet added from the inbox (or in my case is hidden in the inbox).
I have a Playbar with two play 1's as left/right rear. The Playbar is added as Thing in openhab, the Play 1's are not. All works as expected except for these WARN messages in the log. The RINCON points to both play 1's. And that is a bit strange, because they are not used in openhab. No thing, no channels.

As far as i know there are two different warnings added to the log every time a different track is played on the Sonos Playbar:

[WARN ] [s.internal.handler.ZonePlayerHandler] - Cannot update media data for group member (Could not find handler for RINCON_XXXXXXXXXXXXX)
[WARN ] [s.internal.handler.ZonePlayerHandler] - Cannot update channel for group member (Could not find handler for RINCON_XXXXXXXXXXXXX)

This is with build 1162 and when i check my logs, this is happening for months, can't determine with what build it was introduced.

lolodomo:
Fixes #4850

### DIFF
--- a/extensions/binding/org.eclipse.smarthome.binding.sonos/src/main/java/org/eclipse/smarthome/binding/sonos/internal/handler/ZonePlayerHandler.java
+++ b/extensions/binding/org.eclipse.smarthome.binding.sonos/src/main/java/org/eclipse/smarthome/binding/sonos/internal/handler/ZonePlayerHandler.java
@@ -252,7 +252,7 @@ public class ZonePlayerHandler extends BaseThingHandler implements UpnpIOPartici
                             getCoordinatorHandler().stop();
                         }
                     } catch (IllegalStateException e) {
-                        logger.warn("Cannot handle stop command ({})", e.getMessage());
+                        logger.debug("Cannot handle stop command ({})", e.getMessage());
                     }
                     break;
                 case VOLUME:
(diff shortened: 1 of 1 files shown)


maggu2810:
> the problem is more that the binding tries to update all group members, for example the currently track information. As you have not these things, there is a log message.
But in this case (group slaves), I agree that we should just ignore them if the thing does not exist.
I will think about a fix.

If I understand this change correctly, you changed the log level if an illegal state exception has been catched. Is the reason "no such thing available" the only cause of that exception?

kaikreuzer:
Using info logging at these places is not in line with [our logging guidelines](https://www.eclipse.org/smarthome/documentation/development/guidelines.html#e-logging). So I would say that it should either stay a warning (in case this is a software or general setup bug and we should usually never get here) or it should be logged as debug only.

lolodomo:
Yes, "No such thing available" is the only cause of the exception.

And if INFO level should never be used except for important things, yes we should replace with DEBUG level.

lolodomo:
Review comments integrated.

Issue #3306, opened 2017-04-25
New thing status OFFLINE/THING_GONE required for removing things


cdjackson:
A thing handler (or at least a bridge) needs to be able to mark a thing under its control as "removed". This is needed when the user wants to remove a thing in systems where this must be done through the bridge, and only the bridge gets notified that the thing is gone.

Therefore, the bridge should be able to mark a thing as "OFFLINE" with detail "THING_GONE". It would also be desirable to have an "auto approve" option to allow the thing to be automatically removed if the binding puts it into this state (similar to the auto approve on inbox additions).

Discussion in #3205

PR #4865, opened 2018-01-06
Added THING_GONE enum for ThingStatusDetail

cdjackson:
Partly addresses #3306


sjka:
Such a new status detail should at least be mentioned in the [documentation](https://github.com/eclipse/smarthome/blob/master/docs/documentation/concepts/things.md).

cdjackson:
Added to docs...

rodneyrehm:
Hehe, nice to see #3199 is getting option 2 :)

May I suggest to rename the status from "THING_GONE" to just "GONE"? After all it's the upstream item that's vanished, the thing still exists.

Unrelated: By using GONE instead of VANISH my mental model shifted from [this](https://www.youtube.com/watch?v=rzpeH1xot5E) to [this](https://dev.mysql.com/doc/refman/5.7/en/gone-away.html) ;)

kaikreuzer:
> After all it's the upstream item that's vanished, the thing still exists.

.. which is correct, so the documentation texts of this new status should also be adapted accordingly.

Would it make sense to also implement a kind of (optional) "auto-removal" feature as a counterpart for the "auto-approve"?

cdjackson:
I'll update to say the device has been removed...

#3306 also suggests an auto-approve option which is why I've said that this PR only partly addresses that issue.

sjka:
Thanks! I very much like the point that @rodneyrehm brought up about the name: the `THING_` part indeed is confusing as it is exactly the opposite of the real situation. As it's also not adding any value I would indeed suggest dropping that prefix unless anybody has some objections to it.

cdjackson:
Fine with me - I'll update if there's no objections in the coming few hours....

bundles/core/org.eclipse.smarthome.core.thing/src/main/java/org/eclipse/smarthome/core/thing/ThingStatusDetail.java
```diff
@@ -29,7 +30,12 @@
     BRIDGE_OFFLINE,
     FIRMWARE_UPDATING,
     DUTY_CYCLE,
-    BRIDGE_UNINITIALIZED;
+    BRIDGE_UNINITIALIZED,
+    /**
+     * Thing has been removed. Used for example when the thing has been removed from its bridge and the thing handler
+     * should be removed.
+     */
+    THING_GONE;
```
kaikreuzer: You updated the docs, but here it should also be changed from THING_GONE to GONE, right ( + updating the JavaDoc comment)

cdjackson:
Updated...

bundles/core/org.eclipse.smarthome.core.thing/src/main/java/org/eclipse/smarthome/core/thing/ThingStatusDetail.java
```diff
@@ -16,6 +16,7 @@
  * {@link ThingStatusDetail} defines possible status details of a {@link ThingStatusInfo}.
  *
  * @author Stefan Bußweiler - Initial contribution, added new status details
+ * @author Chris Jackson - Added THING_GONE
```
kaikreuzer: And one more place to update...

PR #4869, opened 2018-01-07
edit syntax error in the documentation

khalil2535:
expression instead of epxression


kaikreuzer:
Thanks! 
(Merging as small patch exception.)

PR #4875, opened 2018-01-08
Add additional info how to configure the language parameter

mhaack:
Signed-off-by: Markus Haack <markus.haack@gmail.com> (github: mhaack)

### DIFF
--- a/extensions/binding/org.eclipse.smarthome.binding.weatherunderground/README.md
+++ b/extensions/binding/org.eclipse.smarthome.binding.weatherunderground/README.md
@@ -54,6 +54,8 @@ For the location parameter, different syntaxes are possible:
 
 It can happen that the service is not able to determine the station to use, for example when you select as location a city in which several stations are registered. In this case, the thing configuration will fail because the service will not return the data expected by the binding. The best solution in this case is to use as location latitude and longitude, the service will automatically select a station from this position.
 
+For the language parameter Weather Underground uses a special set of language codes which are different from ISO 639-1 standard, for example for German use `DL`  or Swedish use `SW`. See [Weather Underground language support documentation](https://www.wunderground.com/weather/api/d/docs?d=language-support) for a detailed list. 
+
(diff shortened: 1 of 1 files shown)


kaikreuzer:
Thanks for this clarification.
Does this actually mean that "the default is to use the language from the system locale" doesn't work as expected?

mhaack:
Yes that is actually the case, for all languages where Weather Underground uses a own language code.

For example my OH local is set to Germany/German, using the default aka. no language parameter provided I get the following:

```
2018-01-08 12:47:02.398 [DEBUG] [nd.handler.WeatherUndergroundHandler] - Use language from system locale: de
2018-01-08 12:47:02.400 [DEBUG] [nd.handler.WeatherUndergroundHandler] - URL = http://api.wunderground.com/api/<my key>/conditions/forecast10day/lang:DE/q/51.00000,12.0000.json
```

`lang:DE` is not correct here and ignored by Weather Underground and all text values are in english. Setting language parameter via `.thing` file to `DL` makes everything working as excepted.

htreu:
Would it make sense to have a mapping table which will _fix_ the default behaviour? In the light of _units of measurement_ with automatic unit conversion it would be nice to have the correct texts just from the default locale.

@mhaack please review your commit settings and have a look at the ip-validation check: you need to have a valid Eclipse Committer Agreement for the smarthome project to accept your contribution. See here for more information on the ECA: https://www.eclipse.org/legal/ECA.php

mhaack:
@htreu I signed the Eclipse Contributor Agreement, but after creating the PR. So re-triggering the check should be enough to make it pass but I don't know how this can be done (without a dummy commit)

sjka:
I created issue #4892 for implementing the mapping. 

Your Eclipse Contributor Agreement is all fine, thanks for signing it! The automatic check doesn't run again, but that shouldn't bother us. 

Thanks for your contribution! Looking forward to your next one 😉

PR #4877, opened 2018-01-09
Fix Channel default constructor to not create an invalid UID

htreu:
This commit (3e8e91ef39dc31fff63e75d) introduced a bug when the default constructor for Channel is reflectively invoked by a storage engine. The ChannelUID must not be created with less then 4 segments.
Imho we should not create a dummy uid in this case but live with the fact that uid might not have been initialized (so not final and not covered by null annotations).


htreu:
I will add a simple test for the default c'tor.

sjka:
Thanks!

PR #4882, opened 2018-01-10
Create state description merging service

lpapazow:
Created as a fix for #2267.

As suggested in this [discussion](https://github.com/eclipse/smarthome/pull/4428#issuecomment-340480706), created a **StateDescriptionService** that has a **getStateDescription** method, which iterates over all **StateDescriptionProviders** and sets the **StateDescription**'s fields if they haven't been set by a **StateDescriptionProvider** with higher ranking.

The list of **stateDescriptionProviders** is moved from the **ItemRegistryImpl** class to the **StateDescriptionService** and a **StateDescriptionService** reference is added to the **ItemRegistryImpl** which is later passed to the **GenericItem** class. 

Changed **GenericItem**'s **getStateDescription** method implementation to use **StateDescriptionService**'s **getStateDescription** method.

Added tests for the StateDescription service in the **StateDescriptionServiceImplTest** class.

### DIFF
--- a/bundles/core/org.eclipse.smarthome.core.test/src/test/java/org/eclipse/smarthome/core/service/StateDescriptionServiceImplTest.java
+++ b/bundles/core/org.eclipse.smarthome.core.test/src/test/java/org/eclipse/smarthome/core/service/StateDescriptionServiceImplTest.java
@@ -1,0 +1,179 @@
+/**
+ * Copyright (c) 2014,2018 Contributors to the Eclipse Foundation
+ *
+ * See the NOTICE file(s) distributed with this work for additional
+ * information regarding copyright ownership.
+ *
+ * This program and the accompanying materials are made available under the
+ * terms of the Eclipse Public License 2.0 which is available at
+ * http://www.eclipse.org/legal/epl-2.0
+ *
+ * SPDX-License-Identifier: EPL-2.0
+ */
+package org.eclipse.smarthome.core.service;
+
+import static org.hamcrest.CoreMatchers.is;
+import static org.junit.Assert.assertThat;
+import static org.mockito.Mockito.mock;
+import static org.mockito.Mockito.when;
+
+import java.math.BigDecimal;
+import java.util.Arrays;
+import java.util.Collections;
+import java.util.List;
+
+import org.eclipse.smarthome.core.internal.service.StateDescriptionServiceImpl;
+import org.eclipse.smarthome.core.library.items.NumberItem;
+import org.eclipse.smarthome.core.types.StateDescription;
+import org.eclipse.smarthome.core.types.StateDescriptionProvider;
+import org.eclipse.smarthome.core.types.StateOption;
+import org.junit.Test;
+
+/**
+ * Tests for the StateDescriptionService implementation
+ * 
+ * @author Lyubomir Papazov
+ *
+ */
+public class StateDescriptionServiceImplTest {
+
+	private static final String ITEM_NAME = "Item1";
+	private static final int STATE_DESCRIPTION_PROVIDER_DEFAULT_SERVICE_RANKING = 0;
+	private static final BigDecimal STATE_DESCRIPTION_PROVIDER_DEFAULT_MIN_VALUE = new BigDecimal("0");
+	private static final BigDecimal STATE_DESCRIPTION_PROVIDER_DEFAULT_MAX_VALUE = new BigDecimal("0");
+	private static final BigDecimal STATE_DESCRIPTION_PROVIDER_DEFAULT_STEP = new BigDecimal("0");
+	private static final String STATE_DESCRIPTION_PROVIDER_DEFAULT_PATTERN = "pattern1";
+	private static final boolean STATE_DESCRIPTION_PROVIDER_DEFAULT_IS_READONLY = false;
+	private static final List<StateOption> STATE_DESCRIPTION_PROVIDER_DEFAULT_OPTIONS = Collections.emptyList();
+
+	private StateDescriptionServiceImpl mergingService = new StateDescriptionServiceImpl();
+	private NumberItem item;
+
+	@Test
+	public void testServiceWithOneStateDescriptionProvider() {
+		StateDescriptionProvider stateDescriptionProviderDefault = mock(StateDescriptionProvider.class);
+		when(stateDescriptionProviderDefault.getRank()).thenReturn(STATE_DESCRIPTION_PROVIDER_DEFAULT_SERVICE_RANKING);
+		StateDescription stateDescription1 = new StateDescription(STATE_DESCRIPTION_PROVIDER_DEFAULT_MIN_VALUE,
+				STATE_DESCRIPTION_PROVIDER_DEFAULT_MAX_VALUE, STATE_DESCRIPTION_PROVIDER_DEFAULT_STEP,
+				STATE_DESCRIPTION_PROVIDER_DEFAULT_PATTERN, STATE_DESCRIPTION_PROVIDER_DEFAULT_IS_READONLY,
+				STATE_DESCRIPTION_PROVIDER_DEFAULT_OPTIONS);
+		when(stateDescriptionProviderDefault.getStateDescription(ITEM_NAME, null)).thenReturn(stateDescription1);
+
+		mergingService.addStateDescriptionProvider(stateDescriptionProviderDefault);
+
+		item = new NumberItem(ITEM_NAME);
+		item.setStateDescriptionService(mergingService);
+		StateDescription finalStateDescription = item.getStateDescription();
+
+		assertThat(finalStateDescription.getMinimum(), is(STATE_DESCRIPTION_PROVIDER_DEFAULT_MIN_VALUE));
+		assertThat(finalStateDescription.getMaximum(), is(STATE_DESCRIPTION_PROVIDER_DEFAULT_MAX_VALUE));
+		assertThat(finalStateDescription.getStep(), is(STATE_DESCRIPTION_PROVIDER_DEFAULT_STEP));
+		assertThat(finalStateDescription.getPattern(), is(STATE_DESCRIPTION_PROVIDER_DEFAULT_PATTERN));
+		assertThat(finalStateDescription.isReadOnly(), is(STATE_DESCRIPTION_PROVIDER_DEFAULT_IS_READONLY));
+		assertThat(finalStateDescription.getOptions(), is(STATE_DESCRIPTION_PROVIDER_DEFAULT_OPTIONS));
+	}
+
+	@Test
+	public void testMinValueMaxValueStepAndPatternTwoDescriptionProviders() {
+
+		StateDescription stateDescription1 = new StateDescription(new BigDecimal("-1"), new BigDecimal("-1"),
+				new BigDecimal("-1"), "pattern1", false, null);
+		StateDescription stateDescription2 = new StateDescription(new BigDecimal("-2"), new BigDecimal("-2"),
+				new BigDecimal("-2"), "pattern2", false, null);
+
+		int stateDescriptionProvider1ServiceRanking = -1;
+		int stateDescriptionProvider2ServiceRanking = -2;
+
+		StateDescription finalStateDescription = mergeStateDescriptions(stateDescription1, stateDescription2,
+				stateDescriptionProvider1ServiceRanking, stateDescriptionProvider2ServiceRanking);
+
+		assertThat(finalStateDescription.getMinimum(), is(stateDescription1.getMinimum()));
+		assertThat(finalStateDescription.getMaximum(), is(stateDescription1.getMaximum()));
+		assertThat(finalStateDescription.getStep(), is(stateDescription1.getStep()));
+		assertThat(finalStateDescription.getPattern(), is(stateDescription1.getPattern()));
+	}
+
+	@Test
+	public void testCorrectIsReadOnlyWhenTwoDescriptionProvidersHigherRankingIsNotReadOnly() {
+		StateDescription stateDescription1 = new StateDescription(null, null, null, null, false, null);
+		StateDescription stateDescription2 = new StateDescription(null, null, null, null, true, null);
+
+		int stateDescriptionProvider1ServiceRanking = -1;
+		int stateDescriptionProvider2ServiceRanking = -2;
+
+		StateDescription finalStateDescription = mergeStateDescriptions(stateDescription1, stateDescription2,
+				stateDescriptionProvider1ServiceRanking, stateDescriptionProvider2ServiceRanking);
+
+		assertThat(finalStateDescription.isReadOnly(), is(stateDescription1.isReadOnly()));
+	}
+
+	@Test
+	public void testCorrectIsReadOnlyWhenTwoDescriptionProvidersHigherRankingIsReadOnly() {
+		StateDescription stateDescription1 = new StateDescription(null, null, null, null,
+				true, null);
+		StateDescription stateDescription2 = new StateDescription(null, null, null, null,
+				false, null);
+		
+		int stateDescriptionProvider1ServiceRanking = -1;
+		int stateDescriptionProvider2ServiceRanking = -2;
+
+		StateDescription finalStateDescription = mergeStateDescriptions(stateDescription1, stateDescription2,
+				stateDescriptionProvider1ServiceRanking, stateDescriptionProvider2ServiceRanking);
+
+		assertThat(finalStateDescription.isReadOnly(), is(stateDescription1.isReadOnly()));
+	}
+
+	@Test
+	public void testCorrectOptionsWhenTwoDescriptionProvidersHigherRankingProvidesOptions() {
+		StateDescription stateDescription1 = new StateDescription(null, null, null, null, false,
+				Arrays.asList(new StateOption("value", "label")));
+		StateDescription stateDescription2 = new StateDescription(null, null, null, null, false,
+				Collections.emptyList());
+		
+		int stateDescriptionProvider1ServiceRanking = -1;
+		int stateDescriptionProvider2ServiceRanking = -2;
+
+		StateDescription finalStateDescription = mergeStateDescriptions(stateDescription1, stateDescription2,
+				stateDescriptionProvider1ServiceRanking, stateDescriptionProvider2ServiceRanking);
+
+		assertThat(finalStateDescription.getOptions(), is(stateDescription1.getOptions()));
+	}
+
+	@Test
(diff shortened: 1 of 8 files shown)


bundles/core/org.eclipse.smarthome.core/src/main/java/org/eclipse/smarthome/core/service/StateDescriptionServiceImpl.java
```diff
@@ -0,0 +1,109 @@
...
+        }
+        StateDescription result = foundNonNullStateDescription
+                ? new StateDescription(minimum, maximum, step, pattern, isReadOnly, stateOptions)
+                : null;
+        return result;
+    }
+
+    private static boolean shouldCheckNextDescriptionProvider(BigDecimal minimum, BigDecimal maximum, BigDecimal step,
+            String pattern, boolean isReadOnly, List<StateOption> stateOptions) {
+        return minimum == null || maximum == null || step == null || pattern == null || isReadOnly == false
```
triller-telekom: Should we document somewhere that if you want to create a "winning" `StateDescriptionProvider` that you have to set not only its priority to the highest value but **in addition* also provide values for **all* fields, i.e. minimum, maximum, step, pattern, isReadOnly + stateOptions?

bundles/core/org.eclipse.smarthome.core.test/src/test/java/org/eclipse/smarthome/core/service/StateDescriptionServiceImplTest.java
```diff
@@ -0,0 +1,265 @@
...
+
+		mergingService.addStateDescriptionProvider(stateDescriptionProvider1);
+		mergingService.addStateDescriptionProvider(stateDescriptionProvider2);
+		item = new NumberItem(ITEM_NAME);
+		item.setStateDescriptionService(mergingService);
+
+		StateDescription finalStateDescription = item.getStateDescription();
+
+		assertThat(
+				"expected minimum to be equal to the minumum from provider one because it has higher service ranking",
```
triller-telekom: Can you please remove the explanation string in the `assertThat` calls because they are hard to keep in sync and most of the time they are either stating the obvious or if they are meant as a error message for the failing test they are written to state the opposite of what the test actually checks.

bundles/core/org.eclipse.smarthome.core.test/src/test/java/org/eclipse/smarthome/core/service/StateDescriptionServiceImplTest.java
```diff
@@ -0,0 +1,265 @@
...
+		BigDecimal stateDescriptionProvider1MaxValue = new BigDecimal("-1");
+		BigDecimal stateDescriptionProvider2MaxValue = new BigDecimal("-2");
+
+		BigDecimal stateDescriptionProvider1Step = new BigDecimal("-1");
+		BigDecimal stateDescriptionProvider2Step = new BigDecimal("-2");
+
+		String stateDescriptionProvider1Pattern = "pattern1";
+		String stateDescriptionProvider2Pattern = "pattern2";
+
+		StateDescriptionProvider stateDescriptionProvider1 = mock(StateDescriptionProvider.class);
```
triller-telekom: Since the following lines down to `StateDescription finalStateDescription = item.getStateDescription();` are always the same for each test, how about creating a function that you pass the rank and description form 2 providers and return the `finalStateDescription`, to avoid the duplication of the code?

bundles/core/org.eclipse.smarthome.core.test/src/test/java/org/eclipse/smarthome/core/service/StateDescriptionServiceImplTest.java
```diff
@@ -0,0 +1,265 @@
...
+ * See the NOTICE file(s) distributed with this work for additional
+ * information regarding copyright ownership.
+ *
+ * This program and the accompanying materials are made available under the
+ * terms of the Eclipse Public License 2.0 which is available at
+ * http://www.eclipse.org/legal/epl-2.0
+ *
+ * SPDX-License-Identifier: EPL-2.0
+ */
+
```
htreu: empty line

bundles/core/org.eclipse.smarthome.core/src/main/java/org/eclipse/smarthome/core/internal/items/ItemRegistryImpl.java
```diff
@@ -388,31 +388,14 @@ protected void activate(final ComponentContext componentContext) {
...
-                }
-            });
-
-            for (Item item : getItems()) {
-                ((GenericItem) item).setStateDescriptionProviders(stateDescriptionProviders);
-            }
-        }
+    
+    @Reference(cardinality = ReferenceCardinality.MANDATORY, policy = ReferencePolicy.DYNAMIC)
+    protected void addStateDescriptionMergingService(StateDescriptionService stateDescriptionService) {
```
htreu: since we now have only one service which is injected & the service name is `StateDescriptionService` please rename the method to `setStateDescriptionService`.

bundles/core/org.eclipse.smarthome.core/src/main/java/org/eclipse/smarthome/core/internal/items/ItemRegistryImpl.java
```diff
@@ -388,31 +388,14 @@ protected void activate(final ComponentContext componentContext) {
...
-                    return provider2.getRank().compareTo(provider1.getRank());
-                }
-            });
-
-            for (Item item : getItems()) {
-                ((GenericItem) item).setStateDescriptionProviders(stateDescriptionProviders);
-            }
-        }
+    
+    @Reference(cardinality = ReferenceCardinality.MANDATORY, policy = ReferencePolicy.DYNAMIC)
```
htreu: The new service should have `cardinality=OPTIONAL` and then should be given to all existing items when set (just like a new StateDescriptionProvider before).

bundles/core/org.eclipse.smarthome.core/src/main/java/org/eclipse/smarthome/core/internal/items/ItemRegistryImpl.java
```diff
@@ -388,31 +388,14 @@ protected void activate(final ComponentContext componentContext) {
...
+        this.stateDescriptionService = stateDescriptionService;
     }
-
-    protected void removeStateDescriptionProvider(StateDescriptionProvider provider) {
-        stateDescriptionProviders.remove(provider);
-        for (Item item : getItems()) {
-            ((GenericItem) item).setStateDescriptionProviders(stateDescriptionProviders);
-        }
+    
+    protected void removeStateDescriptionMergingService(StateDescriptionService stateDescriptionService) {
```
htreu: please rename the method to `unsetStateDescriptionService` and also unset the StateDescriptionService` from all available items.

bundles/core/org.eclipse.smarthome.core/src/main/java/org/eclipse/smarthome/core/service/StateDescriptionService.java
```diff
@@ -0,0 +1,43 @@
...
+ * See the NOTICE file(s) distributed with this work for additional
+ * information regarding copyright ownership.
+ *
+ * This program and the accompanying materials are made available under the
+ * terms of the Eclipse Public License 2.0 which is available at
+ * http://www.eclipse.org/legal/epl-2.0
+ *
+ * SPDX-License-Identifier: EPL-2.0
+ */
+
```
htreu: remove empty line

bundles/core/org.eclipse.smarthome.core/src/main/java/org/eclipse/smarthome/core/service/StateDescriptionServiceImpl.java
```diff
@@ -0,0 +1,109 @@
...
+ * See the NOTICE file(s) distributed with this work for additional
+ * information regarding copyright ownership.
+ *
+ * This program and the accompanying materials are made available under the
+ * terms of the Eclipse Public License 2.0 which is available at
+ * http://www.eclipse.org/legal/epl-2.0
+ *
+ * SPDX-License-Identifier: EPL-2.0
+ */
+
```
htreu: remove empty line

bundles/core/org.eclipse.smarthome.core/src/main/java/org/eclipse/smarthome/core/service/StateDescriptionServiceImpl.java
```diff
@@ -0,0 +1,109 @@
...
+import org.eclipse.smarthome.core.types.StateDescription;
+import org.eclipse.smarthome.core.types.StateDescriptionProvider;
+import org.eclipse.smarthome.core.types.StateOption;
+import org.osgi.service.component.annotations.Component;
+import org.osgi.service.component.annotations.Reference;
+import org.osgi.service.component.annotations.ReferenceCardinality;
+import org.osgi.service.component.annotations.ReferencePolicy;
+
+/**
```
htreu: Please give a short explanation what this component is doing.

bundles/core/org.eclipse.smarthome.core/src/main/java/org/eclipse/smarthome/core/service/StateDescriptionServiceImpl.java
```diff
@@ -0,0 +1,109 @@
...
+        boolean isReadOnly = false;
+        List<StateOption> stateOptions = Collections.emptyList();
+
+        boolean foundNonNullStateDescription = false;
+
+        for (StateDescriptionProvider stateDescriptionProvider : stateDescriptionProviders) {
+            StateDescription currentDescription = stateDescriptionProvider.getStateDescription(itemName, locale);
+            if (currentDescription != null) {
+                foundNonNullStateDescription = true;
+                if (shouldCheckNextDescriptionProvider(minimum, maximum, step, pattern, isReadOnly, stateOptions)) {
```
htreu: matter of taste, but i would like this to be turned around, like `if(allPropertiesSet(...)) break;` and otherwise continue with the overwrite logic.

htreu: Besides that: the new merge logic might result in different behaviour and makes this API breaking I guess. wdyt?

bundles/core/org.eclipse.smarthome.core/src/main/java/org/eclipse/smarthome/core/service/StateDescriptionServiceImpl.java
htreu: Imho it there should not be such a "winning" ConfigDescriptionProvider: The Channel config description is always to be taken first, so if this one defines all fields from the beginning other would never be able to overwrite. This would result in the pattern from *.items files or *.sitemap files never to be taken into account.

My impression is that this PR should at first just introduce the new StateDescriptionService without changing the merging logic.

htreu:
Hey @lpapazow any update on this?

lpapazow:
I updated the PR, sorry for the delay. 

As suggested [in your comment](https://github.com/eclipse/smarthome/pull/4882#discussion_r161442979), I reverted the changes regarding the merging logic in `getStateDescription` and the PR now just introduces the new `StateDescriptionService`. 

However, the current version of this PR does not resolve the [orinial issue](https://github.com/eclipse/smarthome/issues/2267) that I was trying to fix.

@htreu Can you please start a discussion about an alternative way to merge StateDecriptions without losing relevant information from lower ranking description providers once this PR is merged.

bundles/core/org.eclipse.smarthome.core/src/main/java/org/eclipse/smarthome/core/internal/items/ItemRegistryImpl.java
```diff
@@ -159,9 +156,9 @@ private void replaceInGroupItems(Item oldItem, Item newItem, List<String> groupI
     /**
      * An item should be initialized, which means that the event publisher is
      * injected and its implementation is notified that it has just been
-     * created, so it can perform any task it needs to do after its creation.
+     *  created, so it can perform any task it needs to do after its creation.
```
htreu: this change (adding a blank character) should be reverted.

bundles/core/org.eclipse.smarthome.core/src/main/java/org/eclipse/smarthome/core/internal/items/ItemRegistryImpl.java
```diff
@@ -159,9 +156,9 @@ private void replaceInGroupItems(Item oldItem, Item newItem, List<String> groupI
...
      *
-     * @param item the item to initialize
+     * @param item  the item to initialize
```
htreu: this change (adding a blank character) should be reverted.

bundles/core/org.eclipse.smarthome.core/src/main/java/org/eclipse/smarthome/core/items/GenericItem.java
```diff
@@ -164,9 +163,9 @@ public void removeGroupName(String groupItemName) {
     public void setEventPublisher(EventPublisher eventPublisher) {
         this.eventPublisher = eventPublisher;
     }
-
-    public void setStateDescriptionProviders(List<StateDescriptionProvider> stateDescriptionProviders) {
-        this.stateDescriptionProviders = stateDescriptionProviders;
+
```
htreu: please use the default formatter or remove the blank chars manually.

bundles/core/org.eclipse.smarthome.core/src/main/java/org/eclipse/smarthome/core/service/StateDescriptionService.java
```diff
@@ -0,0 +1,43 @@
...
+ */
+@NonNullByDefault
+public interface StateDescriptionService {
+
+    /**
+     * Implementations of this method merge the StateDescriptions provided from
+     * multiple StateDescriptionProviders into one final StateDescription
+     * 
+     * @param itemName
+     *
```
htreu: Please sue the default ESH formatter here once to get rid of the obsolete empty lines.

lpapazow:
While rebasing the master branch, I noticed that #5022 introduces `Nullable` annotations for the `stateDescriptionProviders` and `StateDescription` in the `GenericItem` class. 
Should I add such annotations for the `stateDescriptionService` filed in the `GenericItem` class?

htreu:
@lpapazow yes, please add null annotations, thanks.

Regarding the original issue #2267: With your changes here any solution is now able to provide its own implementation of a StateDescriptionService. OSGi DS should then replace the ESH provided service with the custom solution service if its service rank is higher. For pure ESH this means the StateDescription calculation will not change.

RFC: @kaikreuzer, @sjka, @maggu2810

bundles/core/org.eclipse.smarthome.core/src/main/java/org/eclipse/smarthome/core/service/StateDescriptionService.java
```diff
@@ -0,0 +1,42 @@
...
+@NonNullByDefault
+public interface StateDescriptionService {
+
+    /**
+     * Implementations of this method merge the StateDescriptions provided from
+     * multiple StateDescriptionProviders into one final StateDescription
+     * 
+     * @param itemName
+     * @param locale (can be null)
+     *
```
maggu2810: Remove that empty line

bundles/core/org.eclipse.smarthome.core/src/main/java/org/eclipse/smarthome/core/service/StateDescriptionService.java
```diff
@@ -0,0 +1,42 @@
...
+ *
+ */
+@NonNullByDefault
+public interface StateDescriptionService {
+
+    /**
+     * Implementations of this method merge the StateDescriptions provided from
+     * multiple StateDescriptionProviders into one final StateDescription
+     * 
+     * @param itemName
```
maggu2810: Add description (also if it is pretty clear) to the parameters.

bundles/core/org.eclipse.smarthome.core/src/main/java/org/eclipse/smarthome/core/service/StateDescriptionService.java
```diff
@@ -0,0 +1,42 @@
...
+ * different StateDescriptionProviders into one StateDescription.
+ * 
+ * @author Lyubomir Papazov - Initial contribution
+ *
+ */
+@NonNullByDefault
+public interface StateDescriptionService {
+
+    /**
+     * Implementations of this method merge the StateDescriptions provided from
```
maggu2810: Shouldn't the first comment end with a dot?

bundles/core/org.eclipse.smarthome.core/src/main/java/org/eclipse/smarthome/core/items/GenericItem.java
```diff
@@ -370,32 +369,7 @@ public void setCategory(@Nullable String category) {
...
-        }
-
-        // we recreate the StateDescription if we found a valid one and state options are given:
-        if (result != null && !stateOptions.isEmpty()) {
-            result = new StateDescription(result.getMinimum(), result.getMaximum(), result.getStep(),
-                    result.getPattern(), result.isReadOnly(), stateOptions);
-        }
-
-        return result;
+        return stateDescriptionService.getStateDescription(this.name, locale);
```
kaikreuzer: This could result in an NPE as stateDescriptionService is nullable (and a dynamic+optional dependency in ItemRegistryImpl).

bundles/core/org.eclipse.smarthome.core/src/main/java/org/eclipse/smarthome/core/service/StateDescriptionServiceImpl.java
```diff
@@ -0,0 +1,90 @@
...
+ * See the NOTICE file(s) distributed with this work for additional
+ * information regarding copyright ownership.
+ *
+ * This program and the accompanying materials are made available under the
+ * terms of the Eclipse Public License 2.0 which is available at
+ * http://www.eclipse.org/legal/epl-2.0
+ *
+ * SPDX-License-Identifier: EPL-2.0
+ */
+package org.eclipse.smarthome.core.service;
```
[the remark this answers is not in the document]
maggu2810: Good point

PR #4884, opened 2018-01-10
add some nullness annotations for config core classes

maggu2810:


bundles/config/org.eclipse.smarthome.config.core/src/main/java/org/eclipse/smarthome/config/core/ConfigDescriptionParameterGroup.java
```diff
@@ -25,30 +28,26 @@
...
-     * @param description
-     *            a description that can be provided to the user
+     * @param name the name, used to link the group, to the parameter
+     * @param context a context string. Can be used to provide some context to the group
+     * @param advanced a flag that is set to true if this group contains advanced settings
+     * @param label the human readable group label
+     * @param description a description that can be provided to the user
      */
-    public ConfigDescriptionParameterGroup(String name, String context, Boolean advanced, String label,
+    public ConfigDescriptionParameterGroup(String name, @Nullable String context, Boolean advanced, String label,
```
kaikreuzer: Why is "advanced" a Boolean and not a boolean here? If we require NonNull, making it a boolean might be a worthwhile change here.

maggu2810: I asked me the same...

In the constructor the `Boolean` object unboxed without checks to the primitive `boolean` type.
So I assume the "non null by default" usage is correct for the argument.

I don't want to change the binary compatibility, so I don't touched the argument type.

kaikreuzer: Ok, you are probably right that it is not worth to break binary compatibility for this.

PR #4890, opened 2018-01-10
Paper UI: Use channel config description instead of channel-type

### LINKED ISSUE #4823: PaperUI is displaying configuration of channel-type rather than channel
Currently it seems that PaperUI is only displaying the configuration of the channel-type, and not the channel itself. 

For example -:
```
channel-type:zigbee:switch_level
```
and not
```
channel:zigbee:device:telegesis:0017880100dc880b:0017880100DC880B_11_switch_level
```

This means it's not possible to have channel specific configuration shown in PaperUI as the system supports. A direct call to the rest interface with the above config uri works fine.

I'm not sure if this is a bug in PaperUI, or maybe in the thing configure provider (or elsewhere).

htreu:
This fixes #4823.

@cdjackson please check this out.


cdjackson:
Seems to work - thanks @htreu

PR #4891, opened 2018-01-11
[WIP / RFC] move generic UID stuff to a seperate class

maggu2810:
Currently the UID is part of the core.thing bundle and states that every UID must start with the binding ID.
To reuse the non-binding, non-thing related UID implementation a new base class for a generic UID has been created.
It is part of the core bundle. So no need to depend on the thing bundle or to start the UID with a binding ID.

### DIFF
--- a/bundles/core/org.eclipse.smarthome.core.thing/src/main/java/org/eclipse/smarthome/core/thing/UID.java
+++ b/bundles/core/org.eclipse.smarthome.core.thing/src/main/java/org/eclipse/smarthome/core/thing/UID.java
@@ -12,89 +12,49 @@
  */
 package org.eclipse.smarthome.core.thing;
 
-import java.util.Arrays;
-import java.util.stream.Collectors;
+import java.util.List;
+
+import org.eclipse.jdt.annotation.NonNullByDefault;
+import org.eclipse.jdt.annotation.Nullable;
+import org.eclipse.smarthome.core.common.GenericUID;
 
 /**
- * {@link UID} is the base class for unique identifiers within the SmartHome
- * framework. A UID must always start with a binding ID.
+ * Base class for binding related unique identifiers within the SmartHome framework.
+ * A UID must always start with a binding ID.
  *
  * @author Dennis Nobel - Initial contribution
  * @author Oliver Libutzki - Added possibility to define UIDs with variable amount of segments
  * @author Jochen Hiller - Bugfix 455434: added default constructor, object is now mutable
+ * @author Markus Rathgeb - Moved base non binding / thing related UID stuff to a base class
  */
-public abstract class UID {
-
-    public static final String SEGMENT_PATTERN = "[A-Za-z0-9_-]*";
-    public static final String SEPARATOR = ":";
-    private final String[] segments;
+@NonNullByDefault
+public abstract class UID extends GenericUID {
 
     /**
+     * For reflection only.
      * Constructor must be public, otherwise it can not be called by subclasses from another package.
      */
     public UID() {
-        this.segments = null;
+        super();
     }
 
     /**
      * Parses a UID for a given string. The UID must be in the format
      * 'bindingId:segment:segment:...'.
      *
-     * @param uid
-     *            uid in form a string (must not be null)
+     * @param uid uid in form a string (must not be null)
      */
     public UID(String uid) {
-        this(splitToSegments(uid));
-    }
-
-    private static String[] splitToSegments(String uid) {
-        if (uid == null) {
-            throw new IllegalArgumentException("Given uid must not be null.");
-        }
-        return uid.split(SEPARATOR);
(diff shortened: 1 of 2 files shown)


triller-telekom:
I do understand your changes and moving the `UID` class out of the `thing` bundle certainly makes sense. But do you have an example for `UID`s that we use which do not start with the binding ID?

maggu2810:
> But do you have an example for `UID`s that we use which do not start with the binding ID?

Not an Eclipse SmartHome one.
I started using logic same as it is part of the `UID` class (e.g. limit the allowed character of an ID etc.).

If you think there will be no other usage, I can keep it in my bundles, too.
I just thought it would be perhaps also make sense for other consumers.

maggu2810:
Rules are using "simple" strings for its UID, so it seems the rule UID is allowed to contain any character (at least I don't find some limitations in the current code base). I don't know if this has been done by intention or if there should be a RuleUID class (some time), too.

sjka:
I don't have a good use-case in mind currently too. I don't think it's realistic to change the rule ID type, but who knows... 

In any case, I don't have any strong feelings about it. Would be okay for me to do that anyway. @kaikreuzer?

bundles/core/org.eclipse.smarthome.core.thing/src/main/java/org/eclipse/smarthome/core/thing/UID.java
```diff
@@ -12,129 +12,83 @@
...
     @Override
     public int hashCode() {
-        final int prime = 31;
-        int result = 1;
-        result = prime * result + Arrays.hashCode(segments);
-        return result;
-    }
-
-    @Override
-    public boolean equals(Object obj) {
```
htreu: out of curiosity: why are you able to remove `equals` here but leave `hashCode` with a call to super?

htreu: Never mind, I just saw the new `equals` implementation...

maggu2810: If a bundle uses the `UID` class and import `org.eclipse.smarthome.core.thing` but does not import `org.eclipse.smarthome.core.common` the IDE complains about e.g.:

```text
Access restriction: The method 'GenericUID.hashCode()' is not API (restriction on required project 'org.eclipse.smarthome.core')
```

That has been the reason I added that stubs at the moment.

maggu2810:
For me it does not care if I use an implementation in a separate bundle or in an ESH framework bundle.
If you think that there will be no need for others, no problem to close this PR.

bundles/core/org.eclipse.smarthome.core.thing/src/main/java/org/eclipse/smarthome/core/thing/UID.java
htreu: thanks for the explanation.

sjka:
This is really not meant as an objection to this PR. I only can't _think_ of any usages at the moment... 

I mean, we have all sorts on non-thing related entities which have unique identifiers (e.g. voice, transport, sitemap, persistence,...), but they're all using simple strings and there typically is only one instance of them in the system, so no segmented identifiers are required today. 

I still have the feeling that this might become useful one day, but I don't have any proof.

kaikreuzer:
> In any case, I don't have any strong feelings about it. Would be okay for me to do that anyway. @kaikreuzer?

I usually prefer to have stuff refactored only when it is really needed. I agree that the rule engine would be the best candidate for it, but for historic reasons it went for strings, which might be hard to change now.
I am also not too fond of the name "GenericUID". For the super interface, "UID" would imho be the preferred name, but having the same interface in two different ESH packages can also easily lead to confusion.

kaikreuzer:
@maggu2810 What is your suggestion on how to proceed here?

maggu2810:
I agree that GenericUID is not a good name. IMHO the current UID class is a binding specific ID as it requires the first segment to be the binding ID. So perhaps its name should be BindingUID and what I named GenericUID should be UID (as it use case is rather unspecific). Then there could be still the ThingUID as it is and some time in the future (or perhaps never) a RuleUID.

But if you don't think that refactor make sense as changing UID to BindingUID will involve all third party products etc. We can close this.
As long as I am the only one with the "need" to use some basic function in different non-binding related UIDs I can handle it in a custom UID implementation and no need to change the ESH one.

kaikreuzer:
I fully agree that it is unfortunate that the UID class is binding specific and not defined in the core.
But as you say: A refactoring would be a huge breaking change, which I would rather want to avoid, if there isn't really much pain that urges us to do so.
So with your approval, I'd suggest to close this PR.

PR #4894, opened 2018-01-11
Improve Bridge.getHandler() Javadoc

### LINKED ISSUE #4069: [Documentation/API Doc] BaseThingHandler.getBridge().getHandler()
The JavaDoc of `BaseThingHandler.getBridge().getHandler()` says "the handler can be null", but in recently added extensions I saw this:
```
    @Override
    public synchronized void initialize() {
        ... omitted ...
        TradfriGatewayHandler handler = (TradfriGatewayHandler) getBridge().getHandler();
        String uriString = handler.getGatewayURI() + "/" + id;
        ... omitted ...
    }
```
There is no null check. Is `getHandler()` guaranteed to return a handler if called in a thing `initialize`?
This issue is either/or:
* Change the JavaDoc to mention when and under which conditions the bridge handler is available.
* Correct the usage of the method in the above binding.

triller-telekom:
Fixes #4069

### DIFF
--- a/bundles/core/org.eclipse.smarthome.core.thing/src/main/java/org/eclipse/smarthome/core/thing/Bridge.java
+++ b/bundles/core/org.eclipse.smarthome.core.thing/src/main/java/org/eclipse/smarthome/core/thing/Bridge.java
@@ -36,7 +36,7 @@ public interface Bridge extends Thing {
     /**
      * Gets the bridge handler.
      *
-     * @return the handler the handler can be null for a Thing which is not initialized. Note that a Bridge is
+     * @return the handler which can be null for a Thing that is not initialized. Note that a Bridge is
      *         guaranteed to be initialized before its children. It is therefore safe to call getBridge().getHandler()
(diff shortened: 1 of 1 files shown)


bundles/core/org.eclipse.smarthome.core.thing/src/main/java/org/eclipse/smarthome/core/thing/Bridge.java
```diff
@@ -33,8 +36,11 @@
     /**
      * Gets the bridge handler.
      *
-     * @return the handler (can be null)
+     * @return the handler the handler can be null for a Thing which is not initialized. Note that a Bridge is
```
sjka: the handler the handler

triller-telekom: fixed

sjka:
Thanks!

PR #4896, opened 2018-01-11
[EXEC transformation] Update README.md

Josar:
Added transformation examples as somehow nobody really knows how it works.

Signed-off-by: josar <josuaarndt@live.de>

### DIFF
--- a/extensions/transform/org.eclipse.smarthome.transform.exec/README.md
+++ b/extensions/transform/org.eclipse.smarthome.transform.exec/README.md
@@ -9,6 +9,7 @@ The external program must either be in the executable search path of the server 
 **.items**
 
 This will replace the visible label in the UI with the transformation you apply with the command <YourCommand>.
+  
 ```java
 String yourItem "Some info  [EXEC(/absolute/path/to/your/TransformProgram %s):]"
 ```
@@ -16,6 +17,7 @@ String yourItem "Some info  [EXEC(/absolute/path/to/your/TransformProgram %s):]"
 **.thing**
 
 This example shows the transformation of the output channel of an item.
+
 ```java
 Thing yourThing [ ... , transform="EXEC(/absolute/path/to/your/TransformProgram %s)"]
 ```
@@ -35,6 +37,7 @@ end
 **Example with a program**
 
 Substitute the `/absolute/path/to/your/TransformProgram` with
+
 ```shell
 /bin/date -v1d -v+1m -v-1d -v-%s
 ```
--- a/extensions/transform/org.eclipse.smarthome.transform.exec/README.md
+++ b/extensions/transform/org.eclipse.smarthome.transform.exec/README.md
@@ -27,7 +27,7 @@ rule "Your Rule Name"
   when
     Item YourTriggeringItem changed
 then
-    var formatted = transform("EXEC","/absolute/path/to/your/TransformProgram ", YourTriggeringItem.state.toString)
+    var formatted = transform("EXEC","/absolute/path/to/your/TransformProgram", YourTriggeringItem.state.toString)
     yourFormattedItem.sendCommand(formatted.toString) 
 end
 ```
--- a/extensions/transform/org.eclipse.smarthome.transform.exec/README.md
+++ b/extensions/transform/org.eclipse.smarthome.transform.exec/README.md
@@ -11,7 +11,7 @@ The external program must either be in the executable search path of the server 
 This will replace the visible label in the UI with the transformation you apply with the command <YourCommand>.
   
 ```java
-String yourItem "Some info  [EXEC(/absolute/path/to/your/TransformProgram %s):]"
+String yourItem "Some info  [EXEC(/absolute/path/to/your/TransformProgram %s):%s]"
 ```
   
 **.thing**
@@ -26,7 +26,7 @@ Thing yourThing [ ... , transform="EXEC(/absolute/path/to/your/TransformProgram 
 
 ```java
 rule "Your Rule Name"
-  when
+when
     Item YourTriggeringItem changed
 then
     var formatted = transform("EXEC","/absolute/path/to/your/TransformProgram", YourTriggeringItem.state.toString)
--- a/extensions/transform/org.eclipse.smarthome.transform.exec/README.md
+++ b/extensions/transform/org.eclipse.smarthome.transform.exec/README.md
@@ -15,7 +15,7 @@ String yourItem "Some info  [EXEC(/absolute/path/to/your/TransformProgram %s):]"
   
 **.thing**
 
-This example shows the transformation of the input channel of the Exec Binding with "%2s" and the transformation of the return value of the execution, the output channel, with "%s".
+This example shows the transformation of the output channel of an item.
 ```java
 Thing yourThing [ ... , transform="EXEC(/absolute/path/to/your/TransformProgram %s)"]
 ```
--- a/extensions/transform/org.eclipse.smarthome.transform.exec/README.md
+++ b/extensions/transform/org.eclipse.smarthome.transform.exec/README.md
@@ -17,7 +17,7 @@ String yourItem "Some info  [EXEC(/absolute/path/to/your/TransformProgram %s):]"
 
 This example shows the transformation of the input channel of the Exec Binding with "%2s" and the transformation of the return value of the execution, the output channel, with "%s".
 ```java
-Thing exec:command:yourCommand [ command="/absolute/path/to/your/ExecProgram %2s" , transform="EXEC(/absolute/path/to/your/TransformProgram %s)"]
+Thing yourThing [ ... , transform="EXEC(/absolute/path/to/your/TransformProgram %s)"]
 ```
 
 **.rules**
@@ -34,8 +34,8 @@ end
 
 **Example with a program**
 
-Substitut the `/absolute/path/to/your/TransformProgram` with
-```java
+Substitute the `/absolute/path/to/your/TransformProgram` with
+```shell
 /bin/date -v1d -v+1m -v-1d -v-%s
 ```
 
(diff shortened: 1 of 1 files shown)


extensions/transform/org.eclipse.smarthome.transform.exec/README.md
```diff
@@ -6,14 +6,45 @@ The external program must either be in the executable search path of the server
...
-return a string showing the last weekday of the month.
+The execution returns a string showing the last weekday of the month.
 
-| input | output |
-|-------|--------|
+| input | output                         |
+|-------|--------------------------------|
 | `fri` | `Fri 31 Mar 2017 13:58:47 IST` |
+
+## Source
```
sjka: Please don't add any references to openHAB here in Eclipse SmartHome. While openHAB is _based_ upon Eclipse SmartHome, it is _just one_ of the existing solutions.

extensions/transform/org.eclipse.smarthome.transform.exec/README.md
```diff
@@ -6,14 +6,45 @@ The external program must either be in the executable search path of the server
...
 
-With the command line 
+**.items**
 
+This will replace the visible label in the UI with the transformation you apply with the command <YourCommand>.
+```java
+String yourItem "Some info  [EXEC(/absolute/path/to/your/TransformProgram %s):]"
 ```
+  
+**.thing**
```
sjka: While the items- and rules-examples absolutely make sense from my point of view, this one unfortunately does not - the exec binding is an openHAB binding and does not exist here in Eclipse SmartHome.

extensions/transform/org.eclipse.smarthome.transform.exec/README.md
```diff
@@ -6,14 +6,45 @@ The external program must either be in the executable search path of the server
...
+    Item YourTriggeringItem changed
+then
+    var formatted = transform("EXEC","/absolute/path/to/your/TransformProgram ", YourTriggeringItem.state.toString)
+    yourFormattedItem.sendCommand(formatted.toString) 
+end
+```
+
+**Example with a program**
+
+Substitut the `/absolute/path/to/your/TransformProgram` with
```
sjka: Substitute

extensions/transform/org.eclipse.smarthome.transform.exec/README.md
```diff
@@ -6,14 +6,45 @@ The external program must either be in the executable search path of the server
...
+```java
```
sjka: `java` -> `sh`

Josar:
May someone explain to me why not just changing typos and such is done directly, this cost a hell of time for everyone.  You finding it me correcting it. You write down the corrected version anyhow.

I allready spent time writting down lost documentation from the OH1 wiki. 
No i dont like to fill out a contributer form.

sjka:
> why not just changing typos and such is done directly

valid point!

> No i dont like to fill out a contributer form.

I fully respect that!

Unfortunately this is a requirement for every contributor to a project under the governance of the Eclipse Foundation (which this one is). This is the bare minimum to ensure the code-base is clean wrt license, intellectual property and other legal aspects. It's a pity, but unfortunately we cannot accept this PR then.

Josar:
you got that agreement.

Josar:
I created an Eclipse foundation account and signed the  Eclipse Contributor Agreement.
Why does the validation still fail?

kaikreuzer:
> Why does the validation still fail?

Probably because your latest commit misses the Signed-off-by line.
But never mind, we can manually check that you have a valid ECA and merge it anyhow.

extensions/transform/org.eclipse.smarthome.transform.exec/README.md
```diff
@@ -6,14 +6,41 @@ The external program must either be in the executable search path of the server
...
+This will replace the visible label in the UI with the transformation you apply with the command <YourCommand>.
+```java
+String yourItem "Some info  [EXEC(/absolute/path/to/your/TransformProgram %s):]"
 ```
+  
+**.thing**
+
+This example shows the transformation of the output channel of an item.
+```java
+Thing yourThing [ ... , transform="EXEC(/absolute/path/to/your/TransformProgram %s)"]
```
kaikreuzer: I am not aware of any such feature and would claim that this whole "things" section should be removed from here.

extensions/transform/org.eclipse.smarthome.transform.exec/README.md
```diff
@@ -6,14 +6,41 @@ The external program must either be in the executable search path of the server
 
 ## Example
 
-With the command line 
+**.items**
 
+This will replace the visible label in the UI with the transformation you apply with the command <YourCommand>.
+```java
```
kaikreuzer: Jekyll requires an empty line before a code section.

extensions/transform/org.eclipse.smarthome.transform.exec/README.md
```diff
@@ -6,14 +6,41 @@ The external program must either be in the executable search path of the server
...
+then
+    var formatted = transform("EXEC","/absolute/path/to/your/TransformProgram", YourTriggeringItem.state.toString)
+    yourFormattedItem.sendCommand(formatted.toString) 
+end
+```
+
+**Example with a program**
+
+Substitute the `/absolute/path/to/your/TransformProgram` with
+```shell
```
kaikreuzer: Jekyll requires an empty line before a code section.

extensions/transform/org.eclipse.smarthome.transform.exec/README.md
Josar: But it works and makes a lot nice things possible. 

https://community.openhab.org/t/2-openhab2-rpi-system-temperature-chart-with-persistence/35182

```php
Thing exec:command:cpuTemp [
        // Command to execute
        command="cat /sys/class/thermal/thermal_zone0/temp",
        // interval time in seconds
        interval=10,
        // should it run when input channel is changed
        autorun=false,
        // EXEC transformation to make 44007/1000 => 44.007
        // Using python to calculate
        transform="EXEC(python3 -c \"print(%s/1000)\")"]
```

kaikreuzer: But as @sjka mentioned above, the Exec binding does not exist in ESH, so you should not refer to it here. Such documentation is much better off in the Exec binding docs itself.

Josar: I do not refer to the exec Binding anymore, just to the transformation. Only the last three lines from the code above are relevant. It is embedded in a example to show how you may use it but schould be valid for all things.

kaikreuzer: You do refer to the exec binding, because the transformation parameter only exists for that and does not make any sense otherwise.

Josar: OK. Please elaborate as a do not understand.

I add some information to `org.eclipse.smarthome.transform.exec/README.md` which is  transfomartion based on the `openhab-transformation-exec` this can be used as a transformation service, same as the JSON, REGEX, JAVAScript, ... 
I use it as i would use a REGEX transformation. 

The EXEC Binding does not have "%s" it supports "%1$tY-%1$tm-%1$td" and "%2$s". To my limited knowledge.

This transformation can be used with things as it is done 
```
http="<[http://example.com/var/value:60000:EXEC(/absolute/path/to/my/program %s)]" }
```
in the openhab 1 wiki.
https://github.com/openhab/openhab1-addons/wiki/Transformations

extensions/transform/org.eclipse.smarthome.transform.exec/README.md
```diff
@@ -6,14 +6,44 @@ The external program must either be in the executable search path of the server
...
+
+```java
+Thing yourThing [ ... , transform="EXEC(/absolute/path/to/your/TransformProgram %s)"]
+```
+
+**.rules**
+
+```java
+rule "Your Rule Name"
+  when
```
maggu2810: It does not really matter, but I assume you can remove the space in front of `when`.

If `java` used as preformatted text interpretation by intention? Why not `text`?

extensions/transform/org.eclipse.smarthome.transform.exec/README.md
maggu2810: I assume this has been fixed already.

sjka: No - see https://github.com/eclipse/smarthome/pull/4896#discussion_r161493128

extensions/transform/org.eclipse.smarthome.transform.exec/README.md
Josar: Because it was demanded so from the openhab team.

extensions/transform/org.eclipse.smarthome.transform.exec/README.md
kaikreuzer: @Josar What I am talking about is the `transform=` parameter in this example. This does not exist for "Things", it only exists for the Exec binding (not being part of ESH) and thus should be documented in that binding and not here.

If we can agree to remove the "*.thing" example from here, it should imho be ready to be merged.

Josar: @kaikreuzer OK i thought the transformation is generally availabel for things I. Was not aware that it is not. Is there a reason why the transformation is not a generic feature of all things? I can imagine it could be usefull.

kaikreuzer: I agree that it makes sense for all things - and that's why it should not be a feature that every binding needs to support in their things (parameters are always specific to the binding and need to be handled by them).
My plan would be to have a "transformation profile" (see [the new profiles](https://www.eclipse.org/smarthome/documentation/concepts/profiles.html)), which could then be hooked on the link between any channel and the item and do their magic there.

Josar: I expected that the transformation prarameter would be inherited to all thing classes. Which sound a lot simpler than profiles.

I read the link and think i understand the words written but i can not imagine how i would define a profile. Besides that it sounds like there will be a additional `.profiles` file which i will have to edit.

PR #4897, opened 2018-01-11
[WIP] Toggle profile

sjka:
This PR introduces a "toggle" profile, inspired by the discussion in #4638 in order to formalize the current (hacky) solution for receiving `toggle`-"commands" via the REST API.  

It extends the `SystemDefaultProfile` by _also_ implementing the `TriggerProfile` interface, allowing state channels to also receive trigger events. 

Such events don't necessary need to come from a binding - the REST API could also be extended to also fire events from the outside (not yet implemented). 

This actually makes the conceptual misuse by this approach surface, as for the use-cases discussed in #4638 the natural way for users would be to fire such "toggle"-trigger-events for items, not for channels, while _normally_ trigger events are only sent by handlers for channels. Should we still "misuse" the `ChannelTriggeredEvent` for this? Or introduce a new `ItemTriggeredEvent` event for this purpose, which the `CommunicationManager` then listens to? And if so, should it then still call `TriggerProfile.onTriggerFromHandler(...)` for it (wrong profile type and wrong direction) or rather introduce a new method like `StateProfile.onTriggerFromItem(...)`? 

This seems to be lots of stuff-to-be-added just for this one use-case. And at least currently I don't really have any other use-cases on stock for sending arbitrary events to items. One other question would remain: How should such an `ItemTriggeredEvent` be handled, keeping in mind that there is one profile per `ItemChannelLink`, not per `Item`? So if there are two channel linked to the same item it doesn't make sense to involve all profiles, because then the state would be toggled twice... 

This doesn't really seem the natural way to go. Therefore, before continuing with this (or dropping it again), I'd like to first get it conceptually right and see if it makes sense at all...

#### Open Topics
- [x] implement a toggle profile
- [ ] decide whether it is viable to use it by default on applicable items
- [ ] decide what are the "toggle-able" items (i.e. including Color/Dimmer or not?)
- [ ] implement tests
- [ ] clarify sending events for items should be done conceptually
- [ ] extend the REST API accordingly

relates to #4638

### DIFF
--- a/bundles/core/org.eclipse.smarthome.core.thing/src/main/java/org/eclipse/smarthome/core/thing/internal/profiles/SystemProfileFactory.java
+++ b/bundles/core/org.eclipse.smarthome.core.thing/src/main/java/org/eclipse/smarthome/core/thing/internal/profiles/SystemProfileFactory.java
@@ -102,7 +102,11 @@ public class SystemProfileFactory implements ProfileFactory, ProfileAdvisor, Pro
         }
         switch (channelType.getKind()) {
             case STATE:
-                return SystemProfiles.DEFAULT;
+                if (SystemToggleProfile.SUPPORTED_ITEM_TYPES.contains(itemType)) {
+                    return SystemProfiles.TOGGLE;
+                } else {
+                    return SystemProfiles.DEFAULT;
+                }
             case TRIGGER:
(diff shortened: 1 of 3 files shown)


kaikreuzer:
Nice try, but it honestly feels pretty wrong to me to mix the "state" profiles with the trigger profiles.
And as you say, it doesn't even fit to the requirement, which would rather be to send a command to the item and not dealing with channels at all (actually, the toggle might want to be applied to an unlinked/virtual item).

I'd therefore rather vote against this PR.

PR #4900, opened 2018-01-12
Fix groovy osgitest

afuechsel:
Fixed OSGiTest.groovy according to the changes in #4715 


afuechsel:
PR was faulty

PR #4901, opened 2018-01-12
fixed OSGiTest.groovy according to the changes of JavaOSGiTest

afuechsel:
see also #4715 


sjka:
Thanks!

PR #4902, opened 2018-01-12
nail down groovy compiler level to 2.3

sjka:
...in the workspace configuration so that it matches the target platform.

At least mine was set to 2.4 after one of the recent restarts, which led to compiler version mismatch exceptions. Can anybody confirm this?

### DIFF
--- a/targetplatform/EclipseSmartHome.setup
+++ b/targetplatform/EclipseSmartHome.setup
@@ -22,6 +22,14 @@
         source="http://www.eclipse.org/oomph/setup/UserPreferences"/>
     <setupTask
         xsi:type="setup:CompoundTask"
+        name="org.codehaus.groovy.eclipse.compilerResolver">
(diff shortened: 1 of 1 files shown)


sjka:
Nevermind... I just did a fresh install to test our oompf configuration and the problem appears to be worse than I initially thought: There is no groovy 2.3 compiler installed automatically anymore. And event worse: it seems that the Groovy-Eclipse update-site for Oxygen doesn't contain the 2.3 compiler anymore. I'll figure out something else...

sjka:
For those who are following my monologue here: It's acually less of a problem, because if the 2.3 compiler is not installed then the groovy integration also uses the 2.4 libraries automatically in a fresh setup. Therefore the problem is not existent for new installations.

maggu2810:
> For those who are following my monologue here

Always :wink:

PR #4903, opened 2018-01-12
fix condition in ProgressCallbackImpl

sjka:
This fixes a mistake I did in #4743:

Original:

```java
Preconditions.checkState((this.progress != null && this.progress == 100) || 
    (this.progressIterator!=null && !progressIterator.hasNext()), "...")
```

Current version:

```java
if ((this.progress != null && this.progress < 100)
    || (this.progressIterator != null && progressIterator.hasNext())) {
```

This PR now replaces it by the correct negation:

```java
if ((this.progress == null || this.progress < 100)
    && (this.progressIterator == null || progressIterator.hasNext())) {
```

### DIFF
--- a/bundles/core/org.eclipse.smarthome.core.thing.test/src/test/groovy/org/eclipse/smarthome/core/thing/firmware/ProgressCallbackTest.groovy
+++ b/bundles/core/org.eclipse.smarthome.core.thing.test/src/test/groovy/org/eclipse/smarthome/core/thing/firmware/ProgressCallbackTest.groovy
@@ -117,6 +117,14 @@ public final class ProgressCallbackTest {
         sut.success()
     }
 
+    @Test
+    void 'assert success at 100 percent even if there is a remaining progress step'(){
+        sut.defineSequence(ProgressStep.DOWNLOADING, ProgressStep.TRANSFERRING)
+        sut.next()
+        sut.update(100)
+        sut.success()
+    }
+
     @Test(expected=IllegalArgumentException)
     void 'assert that update throws IllegalArgumentException if progress smaller 0'(){
         sut.update(-1)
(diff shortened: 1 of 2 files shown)


afuechsel:
We should add a test for more complex logic like this.

sjka:
Good point, I have added it.

@chriskn @tomhoefer it feels odd though that a firmware update can be successful although it hasn't run through all its steps - therefore could you please briefly confirm that this is as intended?

chriskn:
@sjka not all devices provide steps so yes

sjka:
Of course, that makes sense. 

However, if devices _don't_ provide steps at all, then I would expect `progressIterator` to be `null`. The case that we have run into here is somewhat different: They _do_ provide steps but haven't completed them _all_.

chriskn:
> if devices don't provide steps at all, then I would expect progressIterator to be null

correct, if there was no sequence defined the iterator is null. if also the percentage progress is not defined or is smaller 100 we throw an exception. 

Whats the issue? Maybe you should read the condition again.   


These are the cases in which an exception ist thrown: 
- progress == null && progressIterator == null	  ->  exception (no progress indicator)
- progress == null && progressIterator.hasNext()	->  exception (update not in final step)
- progress < 100 && progressIterator == null	    ->  exception (update not at 100%)
- progress < 100 && progressIterator.hasNext()	  ->  exception (update not at 100% and not in final step)

afuechsel:
The issue is: progress==100 && progressIndicator.hasNext. How could progress be 100 in this case? This should also throw an exception but currently it won‘t. If we fix this, several tests need to be adapted, where the mock just sets 100% bit does not forward the progressIndicator.

sjka:
> Whats the issue? Maybe you should read the condition again.

Maybe you should read my question again 😉 

`progress == 100 && progressIterator.hasNext()` -> no exception

to me this sound like a pretty unhealthy state as it reported 100% finished but didn't complete all the steps.

chriskn:
> progress == 100 && progressIterator.hasNext() -> no exception

Is valid because the steps are optional.

kaikreuzer:
If I understand the discussion right, this PR correctly fixes the issue that has been introduced with #4743 - the rest of the discussion is about whether or not that was a good behavior in the first place. But whatever the outcome is - it should be tackled in another PR.

PR #4907, opened 2018-01-12
Sitemap: validate step, minValue and maxValue of Setpoints

### LINKED ISSUE #4498: FR: Some new Setpoint checks
Currently it is possible to have
1. a negative step size
2. step size = 0
3. maxValue < minValue

Can you add new checks and let the sitemap load fail in such case?

triller-telekom:
Fixes #4498

### DIFF
--- a/bundles/model/org.eclipse.smarthome.model.sitemap/src/org/eclipse/smarthome/model/validation/SitemapValidator.xtend
+++ b/bundles/model/org.eclipse.smarthome.model.sitemap/src/org/eclipse/smarthome/model/validation/SitemapValidator.xtend
@@ -85,6 +85,12 @@ class SitemapValidator extends AbstractSitemapValidator {
 		}
 	}
 
+    @Check
+    def void checkSetpoints(Setpoint sp) {
+        if (sp.item === null) {
+            error("Setpoint without an item defined",
+                SitemapPackage.Literals.SETPOINT.getEStructuralFeature(SitemapPackage.SETPOINT));
+        }
 	@Check
 	def void checkSetpoints(Setpoint sp) {
 	    if(sp.item === null)
@@ -94,25 +100,19 @@ class SitemapValidator extends AbstractSitemapValidator {
             return
 	    }
 
-	    if(BigDecimal.ZERO.equals(sp.step))
-	    {
-	        error("Setpoint on item '" +sp.item+ "' has step size of 0",
-                    SitemapPackage.Literals.SETPOINT.getEStructuralFeature(SitemapPackage.SETPOINT__STEP));
-	        return
-	    }
-
-	    if(sp.step < BigDecimal.ZERO)
-        {
-            error("Setpoint on item '" +sp.item+ "' has negative step size",
-                    SitemapPackage.Literals.SETPOINT.getEStructuralFeature(SitemapPackage.SETPOINT__STEP));
-            return
+        if (BigDecimal.ZERO == sp.step) {
+            error("Setpoint on item '" + sp.item + "' has step size of 0",
+                SitemapPackage.Literals.SETPOINT.getEStructuralFeature(SitemapPackage.SETPOINT__STEP));
         }
 
-	    if(sp.minValue > sp.maxValue)
-	    {
-	        error("Setpoint on item '" +sp.item+ "' has larger minValue than maxValue",
-                    SitemapPackage.Literals.SETPOINT.getEStructuralFeature(SitemapPackage.SETPOINT__MIN_VALUE));
-            return
-	    }
-	}
+        if (sp.step < BigDecimal.ZERO) {
+            error("Setpoint on item '" + sp.item + "' has negative step size",
+                SitemapPackage.Literals.SETPOINT.getEStructuralFeature(SitemapPackage.SETPOINT__STEP));
+        }
+
+        if (sp.minValue > sp.maxValue) {
+            error("Setpoint on item '" + sp.item + "' has larger minValue than maxValue",
+                SitemapPackage.Literals.SETPOINT.getEStructuralFeature(SitemapPackage.SETPOINT__MIN_VALUE));
+        }
+    }
 }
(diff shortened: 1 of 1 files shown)


bundles/model/org.eclipse.smarthome.model.sitemap/src/org/eclipse/smarthome/model/validation/SitemapValidator.xtend
```diff
@@ -83,4 +85,34 @@ class SitemapValidator extends AbstractSitemapValidator {
 		}
 	}
 
+	@Check
+	def void checkSetpoints(Setpoint sp) {
+	    if(sp.item === null)
+	    {
+	        error("Setpoint without an item defined",
+                    SitemapPackage.Literals.SETPOINT.getEStructuralFeature(SitemapPackage.SETPOINT));
+            return
```
sjka: IIRC you can have multiple error markers for the same feature - so do you really want to return early?

bundles/model/org.eclipse.smarthome.model.sitemap/src/org/eclipse/smarthome/model/validation/SitemapValidator.xtend
```diff
@@ -83,4 +85,34 @@ class SitemapValidator extends AbstractSitemapValidator {
 		}
 	}
 
+	@Check
+	def void checkSetpoints(Setpoint sp) {
+	    if(sp.item === null)
+	    {
```
sjka: urgh, what happened to your formatter? Don't the brackets feel lonely like this?

bundles/model/org.eclipse.smarthome.model.sitemap/src/org/eclipse/smarthome/model/validation/SitemapValidator.xtend
```diff
@@ -83,4 +85,34 @@ class SitemapValidator extends AbstractSitemapValidator {
...
+	    }
+
+	    if(BigDecimal.ZERO.equals(sp.step))
+	    {
+	        error("Setpoint on item '" +sp.item+ "' has step size of 0",
+                    SitemapPackage.Literals.SETPOINT.getEStructuralFeature(SitemapPackage.SETPOINT__STEP));
+	        return
+	    }
+
+	    if(sp.step < BigDecimal.ZERO)
```
sjka: Why can you use `<` here but had to use `equals` above? If Xtend can handle this, then you probably also can use `0` directly, have you tried that?

bundles/model/org.eclipse.smarthome.model.sitemap/src/org/eclipse/smarthome/model/validation/SitemapValidator.xtend
triller-telekom: Seems like the formatter is not a save action for ".xtend" files. I had to run "CTRL + SHIFT + F" to make it work.

bundles/model/org.eclipse.smarthome.model.sitemap/src/org/eclipse/smarthome/model/validation/SitemapValidator.xtend
triller-telekom: Thanks, I was not aware of that, this little exercise was my first on xtext validation ;)

bundles/model/org.eclipse.smarthome.model.sitemap/src/org/eclipse/smarthome/model/validation/SitemapValidator.xtend
triller-telekom: Indeed I can, I tried that before by it did not work because it was friday and I was confusing `BigInteger` with `BigDecimal` :P

bundles/model/org.eclipse.smarthome.model.sitemap/src/org/eclipse/smarthome/model/validation/SitemapValidator.xtend
```diff
@@ -27,92 +27,84 @@ import java.math.BigDecimal
...
+            error("Setpoint on item '" + sp.item + "' has step size of 0",
+                SitemapPackage.Literals.SETPOINT.getEStructuralFeature(SitemapPackage.SETPOINT__STEP));
+        }
 
-	    if(sp.step < BigDecimal.ZERO)
-        {
-            error("Setpoint on item '" +sp.item+ "' has negative step size",
-                    SitemapPackage.Literals.SETPOINT.getEStructuralFeature(SitemapPackage.SETPOINT__STEP));
-            return
+        if (sp.step < BigDecimal.ZERO) {
```
sjka: Just played around with it a little: This and the below comparison fail with an NPE is the values are not explicitly set.

bundles/model/org.eclipse.smarthome.model.sitemap/src/org/eclipse/smarthome/model/validation/SitemapValidator.xtend
```diff
@@ -21,66 +21,90 @@ import org.eclipse.smarthome.model.sitemap.Sitemap
...
+    }
 
+    @Check
+    def void checkSetpoints(Setpoint sp) {
+        if (sp.item === null) {
+            error("Setpoint without an item defined",
+                SitemapPackage.Literals.SETPOINT.getEStructuralFeature(SitemapPackage.SETPOINT));
+        }
+
+        if (BigDecimal.ZERO == sp.step) {
```
htreu: should this also be compared by `===`?

sjka: No!

* `===` is checking for object identity
* `==` is using the equals implementation

htreu: thanks!

PR #4911, opened 2018-01-13
Add EntryControl thing category

cdjackson:
This adds ```EntryControl``` thing category meant for security keypads, alarm panels etc.

### DIFF
--- a/docs/_data/categories_thing.csv
+++ b/docs/_data/categories_thing.csv
@@ -5,6 +5,7 @@ Camera,"All kinds of cameras",camera.png
 Car,"Smart Cars",
 CleaningRobot,"Vacuum robots, mopping robots, etc.",
 Door,"Door",door.png
+EntryControl,"Entry control devices such as keypads, security control panels etc",
 FrontDoor,"Front Door",frontdoor.png
 GarageDoor,"Garage Door",garagedoor.png
 HVAC,"Air condition devices, Fans",


kaikreuzer:
How would you depict an "EntryControl"? You already introduced the RemoteControl recently, which is supposed to be used for keypads, so I see quite some overlap here. Also, people might think of locks as an entry control device as well, so it could be rather misleading.

cdjackson:
I see an entry control device being a security control panel type of thing - fixed keypads, often with a display - normally on the wall next to the door.

For me, remote control is a portable device.

I agree that in some view there could be considered some overlap, but then we have that already (eg we have "door" and "front door", "sensor" and "motion detector" - I think it's hard to differentiate these categories by look ;) ).

However, if you prefer not to add this, then I think we should try to better document devices that fit into each category so different bindings make a consistent categorisation.

kaikreuzer:
The main goal of the categories is [to get an idea how to render this thing](https://www.eclipse.org/smarthome/documentation/concepts/categories.html#thing-categories).
We should therefore aim to also come up with a suitable icon for the classic icon set for any newly introduced category. And I personally rather have problems thinking of a good `EntryControl` representation, so I would rather vote against it - unless you add a nice, easily understandable icon :-)
But note that any new icon set is supposed to provide a suitable icon as well - we are therefore really trying to keep the list as short as possible.

kaikreuzer:
@maggu2810 & @sjka Your votes please ;-)

maggu2810:
I don't know how a icon for an `EntryControl` should look like.
It could be a keypad without or with a display (as @cdjackson already mentioned), but it could also be a finger print sensor and so on.

If there is an icon to represent the `EntryControl` in a way, so it fits to different entry control systems, I am fine with adding that category if someone finds it useful.

cdjackson:
I’m not a graphics person, so probably not the best to ask about designing icons ;) I’m sure that someone can come up with an alarm/entry control icon though - I could suggest a keypad with a key symbol, or fingerprint as @maggu2810 suggested. Other thoughts could be an RFID sort of icon or a door with an alarm symbol...

A quick google search for inspiration -:

image
image
image
image


One thing I would challenge though (for thought/discussion - not just to be a complete pain ;) )... I know the stated goal of thing categories should be how they look, however I do think that we should think about slightly finer grained definition and also consider function in some areas. For example, a wall switch - while all wall switches might have the same icon, I think it would be useful to differentiate between (for example) a dimmer and a binary switch.

The rational is that another reason for categories could also be to allow searching for devices - the ZWave binding has around 700 things and maybe half will end up as wall switches. If people are looking through the list, possibly filtering the list looking for a device, I think a slightly finer definition should be considered. (but now I'm off topic ;) ).

sjka:
> This adds EntryControl thing category meant for security keypads, alarm panels etc.

This on my perspective is too much of a functional view on it, covering too many differently looking devices. As the above icon examples show nicely: There are keypads, card readers, fingerprint sensors, etc. We should stick to this level of abstraction here, and not fall back to a functional point of view. It simply doesn't matter that they _can_ be used to open doors. 

This just causes confusion on both, binding developers' and icon designers' sides. Keep in mind: All brandings are required to have an icon for all categories (and ideally they all need to get it right).

> to differentiate between (for example) a dimmer and a binary switch.

They look significantly different, so that wouldn't violate the "should be how they look"-rule as much. Here it's just a matter of optimization, i.e. whether we really need that. 350 ZWave devices are a pretty good start for the "pro" side, I'd say. But you're right: off-topic _here_.

cdjackson:
So how would you categorise an alarm panel? Kai suggested a remote control - for me, these are quite different. They look different, and they function differently. I don’t want to add loads of categories, but I think for categorisation to be useful, we need to have enough that people can tell the difference between things.

I personally think the fact that an alarm panel might be an RFID, or fingerprint, or keypad is not really that relevant - wall switches come in all shapes and sizes - many these days have no buttons but are touchscreen or even gesture controlled - but we still have a wall switch category.

I’m not trying to be difficult (honest ;) ) but I do think there should be some element of function in the definition.

PR #4912, opened 2018-01-13
REST sitemap API: widget state added

### LINKED ISSUE #4860: Basiuc UI with Edge: widget update not fully reliable
The widget update done by the Basic UI when run in a Microsoft Edge browser is not fully reliable due to the fact that Basic UI frontend is expecting an item state adjusted to the widget state while the long polling HTTP call is returning only the raw item state.
This is workin,g well in all browsers using SSE because the SSE events built by the server include this expected widget state.
The solution is easy: we just have to add a new field representing the widget state in the REST sitemap API to get something similar to what is provided by the sitemap SSE after the PR 4810.

I will provide a PR once the PR 4810 is merged.

lolodomo:
New helper field added representing the widget state.
This PR is not breaking the API.

This PR allows fiwing a bug in Basic UI when long polling HTTP is in use.

Fixes #4860


lolodomo:
This PR can be considered like the counterpart of the PR 4810.
The difference is that this change is not breaking any client consuming the current API as it only adds a new helper field (that Basic UI was lacking).

@resetnow : you are welcome to review this new change.

lolodomo:
Are you commenting the right PR ? I don't find the link between your comment and my code change ?

lolodomo:
Ah ok I think I understand what you mean. You're right, we need first the other PR that suppressed dataModelLegacy.

lolodomo:
This PR is put in standby.

lolodomo:
PR rebased.

This PR is the final step to have Basic UI working well in Microsoft Edge.

lolodomo:
I am developing new tests; they would be added tomorrow.

bundles/io/org.eclipse.smarthome.io.rest.sitemap.test/src/test/java/org/eclipse/smarthome/io/rest/sitemap/internal/SitemapResourceTest.java
```diff
@@ -270,4 +382,384 @@ public TestItem(String name) {
             return Collections.emptyList();
         }
     }
+
+    public class TestEClass implements EClass {
```
kaikreuzer: Can't you use a mock of `EClass` instead of this lengthy class definition here?

kaikreuzer:
Please also rebase this PR - thanks!

bundles/io/org.eclipse.smarthome.io.rest.sitemap.test/src/test/java/org/eclipse/smarthome/io/rest/sitemap/internal/SitemapResourceTest.java
lolodomo: Somethink like that ?
```
EClass testEClass = mock(EClass.class);
when(testEClass .getName()).thenReturn("slider");
Widget w1 = mock(Widget.class);
when(w1.eClass()).thenReturn(testEClass);
```
Is it possible to mock an interface class ?

lolodomo:
Test updated in order to suppress the class TestEClass.

PR rebased.

PR #4914, opened 2018-01-13
Fixes NPE in ThingResource when thing has no configuration

cdjackson:
As seen [here](https://github.com/openhab/org.openhab.binding.zwave/issues/802), ```ThingType.getConfigDescriptionURI``` will return null if there is no configuration, and ```ConfigDescriptionRegistry. getConfigDescription``` can not be passed a null.


alexus1211:
Hi Still having this issue after Openhabian latest upgrade to OH2 2.2.0_1. Could it be that this fix hasn't been ported to Openhabian?

kaikreuzer:
Clearly not in 2.2.0, which is the release from December. You'd need to switch to development/snapshot version in openHABian.

PR #4926, opened 2018-01-15
Fix GroupItem compilation error

htreu:
Missing import added.


sjka:
Although Travis failed twice due to some unstable tests (DMX & Wemo), the compilation went successfully well beyond the affected class.

PR #4930, opened 2018-01-15
[Tradfri] Added better version check

### LINKED ISSUE #4904: [Tradfri] Trådfri incorrectly reports old firmware on startup
#### Symptom:

The binding reports firmware incorrect on startup of openhab.

```
2018-01-12 11:05:43.095 [WARN ] [radfri.handler.TradfriGatewayHandler] - Gateway with old firmware - please consider upgrading to the latest version.
```

#### Reproduce:

Watch the `/var/log/openhab/openhab.log` while starting the system.

#### Expected outcome:

As my Ikea phone app reports up updated on all devices including the gateway (currently: 1.2.42) I don't expect this warn log.

#### System:

Openhab 2.2 running on an Pine64 device.

[Related forum thread](https://community.openhab.org/t/tradfri-wrong-upgrade-message/36316)

cweitkamp:
- [x] Added better version check
- [x] Add Test Units for version check

Closes #4904 

This is a first version which I started to implement in another binding. I guess it can be enhanced a lot. Please give me some hints if you have any.

### DIFF
--- a/extensions/binding/org.eclipse.smarthome.binding.tradfri/src/main/java/org/eclipse/smarthome/binding/tradfri/internal/model/TradfriVersion.java
+++ b/extensions/binding/org.eclipse.smarthome.binding.tradfri/src/main/java/org/eclipse/smarthome/binding/tradfri/internal/model/TradfriVersion.java
@@ -1,0 +1,78 @@
+/**
+ * Copyright (c) 2014,2018 Contributors to the Eclipse Foundation
+ *
+ * See the NOTICE file(s) distributed with this work for additional
+ * information regarding copyright ownership.
+ *
+ * This program and the accompanying materials are made available under the
+ * terms of the Eclipse Public License 2.0 which is available at
+ * http://www.eclipse.org/legal/epl-2.0
+ *
+ * SPDX-License-Identifier: EPL-2.0
+ */
+package org.eclipse.smarthome.binding.tradfri.internal.model;
+
+import org.eclipse.jdt.annotation.NonNull;
+
+/**
+ * The {@link TradfriVersion} class is a default implementation of comparing TRÅDFRI versions.
+ *
+ * @author Christoph Weitkamp - Initial contribution
+ */
+public class TradfriVersion implements Comparable<TradfriVersion> {
+    private static final String VERSION_PATTERN = "[0-9]+(\\.[0-9]+)*";
+    private static final String VERSION_DELIMITER = "\\.";
+    private final String version;
+
+    /**
+     * Create a new instance.
+     *
+     * @param version the version string
+     */
+    public TradfriVersion(@NonNull String version) {
+        if (version == null)
+            throw new IllegalArgumentException("TradfriVersion cannot be created as version is null.");
+        if (!version.matches(VERSION_PATTERN))
+            throw new IllegalArgumentException("TradfriVersion cannot be created as version has invalid format.");
+        this.version = version;
+    }
+
+    /**
+     * Returns the version string.
+     */
+    public String get() {
+        return version;
+    }
+
+    @Override
+    public int compareTo(@NonNull TradfriVersion other) {
+        if (other == null) {
+            return 1;
+        }
+        String[] thisParts = get().split(VERSION_DELIMITER);
+        String[] otherParts = other.get().split(VERSION_DELIMITER);
+        int length = Math.max(thisParts.length, otherParts.length);
+        for (int i = 0; i < length; i++) {
+            int thisPart = i < thisParts.length ? Integer.parseInt(thisParts[i]) : 0;
+            int otherPart = i < otherParts.length ? Integer.parseInt(otherParts[i]) : 0;
+            if (thisPart < otherPart) {
+                return -1;
+            }
+            if (thisPart > otherPart) {
+                return 1;
+            }
+        }
+        return 0;
+    }
+
+    @Override
+    public boolean equals(Object obj) {
+        if (this == obj) {
+            return true;
+        }
+        if (obj == null || getClass() != obj.getClass()) {
+            return false;
+        }
+        return compareTo((TradfriVersion) obj) == 0;
+    }
+}
--- a/extensions/binding/org.eclipse.smarthome.binding.tradfri/src/main/java/org/eclipse/smarthome/binding/tradfri/internal/model/TradfriVersion.java
+++ b/extensions/binding/org.eclipse.smarthome.binding.tradfri/src/main/java/org/eclipse/smarthome/binding/tradfri/internal/model/TradfriVersion.java
@@ -93,6 +93,6 @@ public class TradfriVersion implements Comparable<TradfriVersion> {
 
     @Override
     public String toString() {
-        return Arrays.toString(parts.toArray());
+        return parts.stream().map(String::valueOf).collect(Collectors.joining("."));
     }
 }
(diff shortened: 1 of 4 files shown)


extensions/binding/org.eclipse.smarthome.binding.tradfri/src/main/java/org/eclipse/smarthome/binding/tradfri/internal/model/TradfriVersion.java
```diff
@@ -0,0 +1,78 @@
...
+ *
+ * This program and the accompanying materials are made available under the
+ * terms of the Eclipse Public License 2.0 which is available at
+ * http://www.eclipse.org/legal/epl-2.0
+ *
+ * SPDX-License-Identifier: EPL-2.0
+ */
+package org.eclipse.smarthome.binding.tradfri.internal.model;
+
+import org.eclipse.jdt.annotation.NonNull;
```
maggu2810: You should not use `@NonNull` but `@NonNullByDefault` and `@Nullable`.

extensions/binding/org.eclipse.smarthome.binding.tradfri/src/main/java/org/eclipse/smarthome/binding/tradfri/internal/model/TradfriVersion.java
```diff
@@ -0,0 +1,78 @@
...
+
+    /**
+     * Create a new instance.
+     *
+     * @param version the version string
+     */
+    public TradfriVersion(@NonNull String version) {
+        if (version == null)
+            throw new IllegalArgumentException("TradfriVersion cannot be created as version is null.");
+        if (!version.matches(VERSION_PATTERN))
```
maggu2810: As you already dereference the version object I don't think you need to check above for null and throw a IAE. A NPE is also okay if you annotate that a non-null reference should be given.
Both are runtime exceptions, any reason to add a check and throw an IAE instead of the NPE here?

extensions/binding/org.eclipse.smarthome.binding.tradfri/src/main/java/org/eclipse/smarthome/binding/tradfri/internal/model/TradfriVersion.java
```diff
@@ -0,0 +1,78 @@
...
+    /**
+     * Returns the version string.
+     */
+    public String get() {
+        return version;
+    }
+
+    @Override
+    public int compareTo(@NonNull TradfriVersion other) {
+        if (other == null) {
```
maggu2810: So, why use `@NonNull` if you add code to handle it without any error?

extensions/binding/org.eclipse.smarthome.binding.tradfri/src/main/java/org/eclipse/smarthome/binding/tradfri/internal/model/TradfriVersion.java
```diff
@@ -0,0 +1,78 @@
...
+            return 1;
+        }
+        String[] thisParts = get().split(VERSION_DELIMITER);
```
maggu2810: I would prefer to split the version on object creation once.
If you hold a version reference for multiple comparisons this could speed up the process a little bit.

cweitkamp: Yes of course. I thought about that too. A StringTokenizer is not an option, am I right?

maggu2810: As long as the version uses only integers separated by the version delimiter wouldn't it be much easier to ... in the constructor:
* split the version using the version delimiter into a string array
* create an integer list (e.g. vector) using the same size as the string array
* parse every string part and store it to the correspondent integer part

The compare to could act similar to this one:

```Java
    public static void main(String[] args) {
        System.out.println(compare(Arrays.asList(1, 2, 0, 0), Arrays.asList(1, 2)));
        System.out.println(compare(Arrays.asList(1, 2, 1, 0), Arrays.asList(1, 2)));
        System.out.println(compare(Arrays.asList(1, 2, 0, 0), Arrays.asList(1, 2, 1)));
    }

    public static int compare(List<Integer> myParts, List<Integer> otherParts) {
        final int minPartsSize = Math.min(myParts.size(), otherParts.size());

        for (int i = 0; i < minPartsSize; ++i) {
            final int diff = myParts.get(i) - otherParts.get(i);
            if (diff == 0) {
                continue;
            } else if (diff > 0) {
                return 1;
            } else {
                return -1;
            }
        }

        for (int i = minPartsSize; i < myParts.size(); ++i) {
            if (myParts.get(i) != 0) {
                return 1;
            }
        }

        for (int i = minPartsSize; i < otherParts.size(); ++i) {
            if (otherParts.get(i) != 0) {
                return -1;
            }
        }

        return 0;
    }
```

extensions/binding/org.eclipse.smarthome.binding.tradfri/src/main/java/org/eclipse/smarthome/binding/tradfri/internal/model/TradfriVersion.java
```diff
@@ -0,0 +1,82 @@
...
+    public String[] get() {
+        return parts;
+    }
+
+    @Override
+    public int compareTo(@NonNull TradfriVersion other) {
+        String[] otherParts = other.get();
+        int length = Math.max(parts.length, otherParts.length);
+        for (int i = 0; i < length; i++) {
+            int thisPart = i < parts.length ? Integer.parseInt(parts[i]) : 0;
```
maggu2810: Wouldn't it be better to parse the strings to integer only once instead on every comparison?

```Java
private final List<Integer> parts;

public ...(final String version) {
    parts = Arrays.stream(version.split(VERSION_DELIMITER)).map(part -> Integer.parseInt(part))
}
```

extensions/binding/org.eclipse.smarthome.binding.tradfri/src/main/java/org/eclipse/smarthome/binding/tradfri/internal/model/TradfriVersion.java
```diff
@@ -0,0 +1,82 @@
...
+        if (!version.matches(VERSION_PATTERN)) {
+            throw new IllegalArgumentException("TradfriVersion cannot be created as version has invalid format.");
+        }
+        this.parts = version.split(VERSION_DELIMITER);
+    }
+
+    /**
+     * Returns an array of strings containing the version parts
+     */
+    public String[] get() {
```
maggu2810: Do we need that function at all?

extensions/binding/org.eclipse.smarthome.binding.tradfri/src/main/java/org/eclipse/smarthome/binding/tradfri/internal/model/TradfriVersion.java
maggu2810: Still missing

extensions/binding/org.eclipse.smarthome.binding.tradfri/src/main/java/org/eclipse/smarthome/binding/tradfri/internal/model/TradfriVersion.java
cweitkamp: I did. Thanks.

extensions/binding/org.eclipse.smarthome.binding.tradfri/src/main/java/org/eclipse/smarthome/binding/tradfri/internal/model/TradfriVersion.java
cweitkamp: Yes, I think so. How should we access the parts of the other `TradfriVersion` object in the `compareTo()` method?

extensions/binding/org.eclipse.smarthome.binding.tradfri/src/main/java/org/eclipse/smarthome/binding/tradfri/internal/model/TradfriVersion.java
cweitkamp: To be honest, I'm not very familiar with those annotations. Used them the first time for now. Please be patient with me on that. Hopefully I did it right now. :-)

cweitkamp:
@maggu2810 Thanks for your review. I picked it up at the same time you answered again.

PR #4932, opened 2018-01-15
Revert back to a standard executor service implementation

### LINKED ISSUE #4931: Things remaining in INITIALIZING state at startup
With the last ESH/OH build (snapshot 1186), we have the comeback of things not initializing correctly at startup.
Most of bridge things (hue, netatmo, ...) remains in a INITIALIZING state.
I have nothing special logged.
The problem is at each startup.

The problem was introduced after the snapshot 1140.

sjka:
...in order to not block the executor threads anymore and
gain time to analyze the root-cause of #4931.

partially reverts #4772
fixes #4931


kaikreuzer:
@Hilbrand FYI - we didn't fully understand why it isn't working, but as it completely breaks the runtime, we decided to revert to the standard executor service for the moment.

PR #4935, opened 2018-01-16
GenericItemUIProvider: get item from registry instead of model repository

### LINKED ISSUE #4716: Starred interaction on config and rules changes

htreu:
Fixes #4716.


kaikreuzer:
Let's take a step back here:
The "Generic" part of the name of GenericItemUIProvider was supposed to tell that it takes its wisdom from DSL files. With this PR, it does not at all depend on the model repository anymore.
Looking a bit closer, it merely proxies to the item registry for category and label and it returns null for the default widget. Sounds pretty useless, doesn't it?
Having had a short look in the code, I would claim that the ItemUIRegistryImpl could be made fully independent of ItemUIProvider as it can instead directly access the item registry.

htreu:
From what I understand from the code the ItemUIRegistryImpl holds a list of ItemUIProviders and tries to resolve label & category from this provider list. To me this looks like an extension point where implementations could hook up ItemUIProviders to resole those properties on their own.
Maybe I got you wrong but I don't see how we could live w/o this connection.

kaikreuzer:
> tries to resolve label & category from this provider list

Right, but meanwhile label & category are part of the item (which wasn't the case at the time when the ItemUIProvider interface was defined). So I do not really see a point in that interface anymore.

htreu:
I'm happily removing the whole ItemUIProvider interface infrastructure but this sounds API breaking as the `org.eclipse.smarthome.ui.items` package gets exported (and with it the ItemUIProvider interface). My suggestion is to merge this first, then decide on how to proceed. wdyt?

kaikreuzer:
Well, would you agree that you can at least remove the GenericItemUIProvider without breaking the API? As mentioned initially, with the current code of this PR, it isn't doing anymore what its name suggests.

htreu:
This would already imply that the ItemUIProvider infrastructure is dead. It will then not be used anymore to resolve label & category.
I will add a second commit which makes these change.

kaikreuzer:
> This would already imply that the ItemUIProvider infrastructure is dead.

No, it would only imply that we have one instance less. IF there is anybody else registering such services, they would still be used, i.e. no API/behavior would be broken.

htreu:
Thanks for the feedback, I was about to remove the _dependency_ on itemUIProviders in the registry. So only the useless GenericItemUIProvider will be gone then.

PR #4945, opened 2018-01-19
fix SafeCallerImplTest

sjka:
...as the `spawn()` helper method [actually blocked](https://github.com/eclipse/smarthome/compare/master...sjka:fixSafeCallerTests?expand=1#diff-c0c87bcca219c6210ea787bcc45cf0a1L564) the calling thread.

Therefore
* Fixed the `spawn()` [method](https://github.com/eclipse/smarthome/compare/master...sjka:fixSafeCallerTests?expand=1#diff-c0c87bcca219c6210ea787bcc45cf0a1R570) to really spawn a new thread without blocking
* Made the `scheduler` [injectable](https://github.com/eclipse/smarthome/compare/master...sjka:fixSafeCallerTests?expand=1#diff-c0c87bcca219c6210ea787bcc45cf0a1R94) by the test instead of just the `poolName`
* Making sure inner assertion errors and exceptions are heard within [@After](https://github.com/eclipse/smarthome/compare/master...sjka:fixSafeCallerTests?expand=1#diff-c0c87bcca219c6210ea787bcc45cf0a1R106)
* Adapted an [assertion](https://github.com/eclipse/smarthome/compare/master...sjka:fixSafeCallerTests?expand=1#diff-c0c87bcca219c6210ea787bcc45cf0a1R215) which was wrong because of the blocking behavior of `spawn()`

### DIFF
--- a/bundles/core/org.eclipse.smarthome.core.test/src/test/java/org/eclipse/smarthome/core/internal/common/SafeCallerImplTest.java
+++ b/bundles/core/org.eclipse.smarthome.core.test/src/test/java/org/eclipse/smarthome/core/internal/common/SafeCallerImplTest.java
@@ -51,13 +51,13 @@ public class SafeCallerImplTest extends JavaTest {
     private static final int THREAD_POOL_SIZE = 3;
 
     // the duration that the called object should block for
-    private static final int BLOCK = 1500;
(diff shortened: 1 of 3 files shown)


sjka:
Good point actually - yes, I will push another commit for it

PR #4948, opened 2018-01-19
fix NTP time calculation

### LINKED ISSUE #4867: NTP binding: not at time
I just discovered that my RPI is not at time at all but 11 minutes before the correct time.
The NTP binding is providing the server time as displayed in Basic UI.
Isn't it the role of the NTP binding to keep my server at time ?

sjka:
...as it was ignoring the offset and only returned the local time when the
response was received

fixes #4867

### DIFF
--- a/extensions/binding/org.eclipse.smarthome.binding.ntp/src/main/java/org/eclipse/smarthome/binding/ntp/handler/NtpHandler.java
+++ b/extensions/binding/org.eclipse.smarthome.binding.ntp/src/main/java/org/eclipse/smarthome/binding/ntp/handler/NtpHandler.java
@@ -61,7 +61,7 @@ import org.slf4j.LoggerFactory;
 
 public class NtpHandler extends BaseThingHandler {
 
-    private Logger logger = LoggerFactory.getLogger(NtpHandler.class);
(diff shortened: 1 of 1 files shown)


lolodomo:
Cool.
If mecessary, I can test this fix this evening. Let me know.

Does it mean the NTP binding never works ?

sjka:
Yes, looks like it 😄 

The code here was an adaption of the openHAB 1 binding - and there it was the same from the beginning in 2011: https://github.com/openhab/openhab1-addons/commit/2472a6c4d9e8c37b5d0aba590f4b5597cb3400b0#diff-c7af007978d885e1b87fb7e9178bf013R138

To be honest, I can't really believe myself that nobody ever noticed this! On the other hand - it is hard to see as I would assume that most computers are having NTP synchronizations for their system clock enabled anyway and the primary use-case for the NTP binding is to have an item with the time at all.

kaikreuzer:
Funny. Nice example of a badly named API in Apache Commons as I assume `getReturnTime()` was read as `getReturnedTime()` and thus it seemed to be the obvious method to call here :-) /cc @teichsta

PR #4951, opened 2018-01-19
handle indefinite results in JsonPath transformations

### LINKED ISSUE #4862: JSONPath transformation fails on indefinite paths
I am not able to filter an MQTT json response with the [JSONPath transformation](https://github.com/eclipse/smarthome/blob/master/extensions/transform/org.eclipse.smarthome.transform.jsonpath/src/main/java/org/eclipse/smarthome/transform/jsonpath/internal/JSonPathTransformationService.java) even though my expression is correct. My guess (*Edit: assessment, proof further down*) is that the transformation doesn't handle indefinite path expressions correctly. According to https://github.com/json-path/JsonPath#what-is-returned-when the jayway jsonpath implementation returns results as lists as soon as the path is indefinite, e.g. when a filter expression is used.

I did my testing with the following data:

```json
  {
    "Time": "2018-01-06T12:56:59",
    "DS18B20-1": {
      "Id": "000005FB25D0",
      "Temperature": 3.2
    },
    "DS18B20-2": {
      "Id": "031680634EFF",
      "Temperature": -18.9
    },
    "TempUnit": "C"
  }
```

My jayway jsonpath expression *(I want to select one specific sensor and the index "DS18B20-1" could change at some point)*:

```jsonpath
$.*[?(@.Id=='000005FB25D0')].Temperature
```

According to https://jsonpath.herokuapp.com the path returns a list rather than a value:

```json
[
   3.2
]
```

Finally here is a snippet from the DEBUG log confirming my hunch:

```
13:07:08.921 [DEBUG] [nternal.JSonPathTransformationService] - about to transform '{"Time":"2018-01-06T13:07:08","Uptime":1,"Vcc":3.187,"POWER":"OFF","Wifi":{"AP":1,"SSId":"HotelZurBirke","RSSI":70,"APMac":"24:65:11:BF:12:D8"}}' by the function '$.Wifi.RSSI'
13:07:08.928 [DEBUG] [nternal.JSonPathTransformationService] - transformation resulted in '70'

13:07:08.997 [DEBUG] [nternal.JSonPathTransformationService] - about to transform '{"Time":"2018-01-06T13:07:08","DS18B20-1":{"Id":"000005FB25D0","Temperature":3.2},"DS18B20-2":{"Id":"031680634EFF","Temperature":-18.9},"TempUnit":"C"}' by the function '$.*[?(@.Id=='000005FB25D0')].Temperature'
13:07:09.002 [DEBUG] [nternal.JSonPathTransformationService] - transformation resulted in '[3.2]'
13:07:09.007 [WARN ] [ab.core.events.EventPublisherDelegate] - given new state is NULL, couldn't post update for 'FK_KuehlschrankTemp_Kuehl'
```

@clinique @sjka would you be so kind to have a look? Thanks!
  
Suggestion: The JSONPath transformation should be able to handle indefinite paths / lists as JSONPath result. This means that resulting lists have to be processed. How about:

* Yield NULL when an empty list is returned
* Yield NULL and an log message on WARN level when a list with multiple elements is returned
* Yield the element when the list consists of exactly one element

?

sjka:
...as JsonPath may return a List if multiple elements matched,
even if they were correctly filtered down to a single element.

fixes #4862

### DIFF
--- a/extensions/transform/org.eclipse.smarthome.transform.jsonpath/src/main/java/org/eclipse/smarthome/transform/jsonpath/internal/JSonPathTransformationService.java
+++ b/extensions/transform/org.eclipse.smarthome.transform.jsonpath/src/main/java/org/eclipse/smarthome/transform/jsonpath/internal/JSonPathTransformationService.java
@@ -77,7 +77,9 @@ public class JSonPathTransformationService implements TransformationService {
             return list.get(0).toString();
         }
         if (list.size() > 1) {
-            logger.warn("The JsonPath expression yielded more than one result: {}", list);
+            logger.warn(
+                    "JsonPath expressions with more than one result are not allowed, please adapt your selector. Result: {}",
+                    list);
         }
         return UnDefType.NULL.toFullString();
     }
--- a/extensions/transform/org.eclipse.smarthome.transform.jsonpath/src/main/java/org/eclipse/smarthome/transform/jsonpath/internal/JSonPathTransformationService.java
+++ b/extensions/transform/org.eclipse.smarthome.transform.jsonpath/src/main/java/org/eclipse/smarthome/transform/jsonpath/internal/JSonPathTransformationService.java
@@ -12,6 +12,8 @@
  */
 package org.eclipse.smarthome.transform.jsonpath.internal;
 
+import java.util.List;
+
 import org.eclipse.smarthome.core.transform.TransformationException;
 import org.eclipse.smarthome.core.transform.TransformationService;
 import org.eclipse.smarthome.core.types.UnDefType;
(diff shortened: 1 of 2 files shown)


ThomDietrich:
Hey Simon,
during #4862 I realized that part of the issue is the jsonpath implementation used and the fact that it behaves slightly different then broadly documented. Could you please add a short paragraph to the README?

* Formal definition: http://goessner.net/articles/JsonPath/
* Implementation: https://github.com/json-path/JsonPath#functions
* Testing: https://jsonpath.herokuapp.com/ --> "Jayway"

extensions/transform/org.eclipse.smarthome.transform.jsonpath/src/main/java/org/eclipse/smarthome/transform/jsonpath/internal/JSonPathTransformationService.java
```diff
@@ -64,4 +72,14 @@ public String transform(String jsonPathExpression, String source) throws Transfo
 
     }
 
+    private String flattenList(List<?> list) {
+        if (list.size() == 1) {
+            return list.get(0).toString();
+        }
+        if (list.size() > 1) {
+            logger.warn("The JsonPath expression yielded more than one result: {}", list);
```
ThomDietrich: Imho this error doesn't imply that NULL is returned...

sjka: > What's the end user supposed to do here?

Check out the code from github and try to understand the program's behavior? 😉 

How about: 

```
Returning NULL because the JsonPath expression yielded more than one result: {} - did you forget to add a filter expression?
```

ThomDietrich: A statement by a true developer 😁 How about

> JsonPath expression with more than one result not allowed, please adapt your selector. Result: {}

maggu2810:
@ThomDietrich Would you like to approve that changes?

PR #4952, opened 2018-01-19
corrected null annotations

kaikreuzer:
I think #4884 introduced a bug here, which I discovered [here](https://github.com/openhab/openhab2-addons/pull/3140/files#diff-f717f341604676be676a8725bea23b99R306) - according to our schema, both label and description are optional and thus can be null. Our XML parser even feeds them in as null (but there isn't any error or warning in the IDE, since we have disabled pretty much all warnings...), so that this went unnoticed.

related to #4884


maniac103:
Don't you also need to annotate the constructor parameters accordingly?

kaikreuzer:
Thanks for noticing - added it!

PR #4953, opened 2018-01-19
make FSInternetRadioHandlerJavaTest independent of the OS language

sjka:
...as SAXParser returns localized error message, which is nice generally but
a little annoying if the tests check for the english message string.


sjka:
> I hate localised exceptions...

+1!

For this purpose, some really smart people invented the `Throwable.getLocalizedMessage()` method - back then, in JDK 1.1...

Issue #4360, opened 2017-09-30
Paper UI: new discovered things should be displayed as ignored when they are auto ignored


lolodomo:
When you ask for new discovery things for example for Sonos, new things are displayed at screen as "not ignored" while they are in fact ignored due to the automatic ignore mechanism (existing thing with the same representation property).
If you reload the inbox by clicking on the menu item "Inbox" then they are ignored as expected.

htreu:
This is the result from my analysis:
* `PersistentInbox` will pass a newly discovered thing to the `thingAdded` method from the `AutomaticInboxProcessor` which does the `autoIgnore` check.
* the `AutomaticInboxProcessor` will call `setFlag(thingUID, IGNORED)` on the `PersistentInbox`. This marks the discovery result from the storage with the flag, not the instance which is currently processed by the `PersistentInbox`.
* `setFlag` will immediately post an `updated` event for the discovery result (note: this happens before an `added` event is emitted).
* Paper UI will happily ignore the `updated` event since no entry for the given thingUID is present in the Paper UI inbox repository.
* `PersistentInbox` will then post an `added` event with the instance the discovery service created. This instance does not have the `IGNORED` flag set.
* Paper UI will pick up the `added` event and display the discovery result as a new result, just because the flag is still set to `NEW`.

A refresh of the Paper UI inbox will retrieve all discovery results from the backend. Since the `IGNORED` flag was correctly set to the instance from the storage Paper UI will then correctly hide the auto-ignored result.

htreu:
One solution would be to reload the discovered result from storage before emitting the first `add` event. This will then provide the correct flag.
The unused `updated` event is obsolete (in this scenario) but does not hurt as long as the client checks for existence of the given thingUID.

PR #4954, opened 2018-01-19
Paper UI: Update discovery results in inbox

### LINKED ISSUE #4921: [Paper UI] Inbox does not update entries upon "updated" event
When an inbox entry is updated there is no notification. This leads to the inbox display in UIs being wrong until the UI is refreshed.

DiscoveryListener should have an updated method to alert the UI that things have changed.

Alternatively, the added method could be called again (maybe) since the method called is always ```thingDiscovered``` in the discovery service.

htreu:
Fixes #4921.

### DIFF
--- a/extensions/ui/org.eclipse.smarthome.ui.paper/web-src/js/services.repositories.js
+++ b/extensions/ui/org.eclipse.smarthome.ui.paper/web-src/js/services.repositories.js
@@ -144,11 +144,16 @@ angular.module('PaperUI.services.repositories', []).factory('bindingRepository',
         var index = repository.findByIndex(function(result) {
             return discoveryResult.thingUID == result.thingUID;
         });
-        if (topic.indexOf("added") > -1 && index == -1) {
+        var command = topic.substring(topic.lastIndexOf('/') + 1);
+        if (command === "added" && index == -1) {
             repository.add(discoveryResult);
         }
-        if (topic.indexOf("removed") > -1 && index != -1) {
-            repository.remove(discoveryResult, index);
+        if (command === "removed" && index != -1) {
+            repository.remove(undefined, index);
+        }
+        if (command === "updated" && index != -1) {
+            repository.remove(undefined, index);
+            repository.add(discoveryResult);
         }
     });
     return repository;


extensions/ui/org.eclipse.smarthome.ui.paper/web-src/js/services.repositories.js
```diff
@@ -148,7 +148,11 @@ angular.module('PaperUI.services.repositories', []).factory('bindingRepository',
             repository.add(discoveryResult);
         }
         if (topic.indexOf("removed") > -1 && index != -1) {
-            repository.remove(discoveryResult, index);
+            repository.remove(undefined, index);
+        }
+        if (topic.indexOf("updated") > -1 && index != -1) {
+            repository.remove(undefined, index);
+            repository.add(discoveryResult);
```
sjka: Just a blind guess - is the order relevant in any way?

extensions/ui/org.eclipse.smarthome.ui.paper/web-src/js/services.repositories.js
```diff
@@ -148,7 +148,11 @@ angular.module('PaperUI.services.repositories', []).factory('bindingRepository',
             repository.add(discoveryResult);
         }
         if (topic.indexOf("removed") > -1 && index != -1) {
-            repository.remove(discoveryResult, index);
+            repository.remove(undefined, index);
+        }
+        if (topic.indexOf("updated") > -1 && index != -1) {
```
sjka: I know, it's been like that previously with the `removed` as well, but still - what happens if I (or rather: a binding) name(s) a Thing e.g. `mysterious_thing_updated`? When this now gets updated, the topic would be `smarthome/inbox/mysterious_thing_updated/removed`. Enjoy debugging 😬 
What I actually want to say: The pure string comparisons here are a little too simple.

extensions/ui/org.eclipse.smarthome.ui.paper/web-src/js/services.repositories.js
htreu: No, not relevant because `add` will append to the data array, `remove` will use the previously calculated index (by thingUID).

extensions/ui/org.eclipse.smarthome.ui.paper/web-src/js/services.repositories.js
htreu: agreed, the current string matching is not enough. I will update the regex to better match the actual topic.

lolodomo:
Maybe you could have a look at the issue #4360. It is not exactly the same issue but it concerns the display  of the updated inbox by Paper UI too.

extensions/ui/org.eclipse.smarthome.ui.paper/web-src/js/services.repositories.js
htreu: Now I got it. The UI orders the discovery results by label during rendering. So order in the repository does not matter.

HARDCYBER:
exactly

Issue #4957, opened 2018-01-20
Enhance HSBType to add additional conversions


cdjackson:
I'd like to suggest that ```HSBType``` is enhanced to provide conversions to/from XY, and potentially to/from CT.

A helper class for converting XY was recently added to the [ZigBee binding](https://github.com/openhab/org.openhab.binding.zigbee/blob/master/src/main/java/org/openhab/binding/zigbee/converter/ColorHelper.java), but I'm sure other bindings would also make use of this and I think it might be better suited as helpers in ```HSBType``` in the same way as I added for RGB some time back.

If this is acceptable I'll look to migrate this over

kaikreuzer:
If it is similar to RGB, I don't see any reason against adding this as well to HSBType.

cdjackson:
Closed by #4959

Issue #4939, opened 2018-01-18
[HUE] Support for Tradfri RGB Bulb


eiGelbGeek:
The Tradfri RGB Bulb supports no CIE Commands, only XY Commands! 
Please Add an Work A Round for this Bulb... perhaps transform the CIE Commands from Openhab to XY Commands. I don't look into the Code of the Tradfri Binding, but I think they do this Transformation to get it worked.

kaikreuzer:
Not sure what you are trying to say here. The hue Binding does all accesses through the hue bridge, which comes with its own REST API. And afaik, this REST API does not differ whether you are controlling a hue, a lightify or a tradfri bulb through it.

Do you have any log files that show that something isn't working as expected?

eiGelbGeek:
If I expected correct the Binding send the Commands in CIE Color Mode? Right? 
The Tradfri Bulbs don't support this Commands, they need Commands in XY Color Mode.

Here the link to the Thread

https://community.openhab.org/t/hue-colorpicker-not-working/38044/14

There are to Screenshots from me ... I hope you understand now the Problem :-)

There are no Error in Log Files.

kaikreuzer:
So what you are saying is that the REST API of the hue bridge indeed requires different kinds of requests for bulbs from different vendors?
The official docs (https://www.developers.meethue.com/documentation/lights-api#16_set_light_state) do not state anything like that. Maybe it would be worth to first discuss at their forum whether this is a bug or indeed intended that way?

eiGelbGeek:
Yes that is what I want to explain.

I think Point: - 1.6.2. Body arguments - in the Documentation - makes the difference. 
I think the HUE Binding sends hue,sat,bri and the Tradfri RGB Bulb don't support this. The Tradfri RGB Bulb needs the xy coordinates.

eiGelbGeek:
Controll the Tradfri RGB Bulb with the HUE App works perfectly.

kaikreuzer:
My point is that the client cannot/shouldn't really know about the exact bulb type - the Philips REST API should do the conversion in such a case.
If you find any discussion around that at the hue developer forum, it would be helpful.

BasvanH:
I have two types of Ikea Tradfri lamps bound to the Hue bridge.

TRADFRI_bulb_GU10_WS_400lm / 0220
TRADFRI_bulb_E27_CWS_opal_600lm / 0200

[...]

PR #4959, opened 2018-01-21
Add XY conversions to HSBType

cdjackson:
This adds XY conversion into the HSBType. Currently this is added as a helper in the ZigBee binding by @puzzle-star. I also see it in the Tradfri binding (I think by @hreichert) and maybe other bindings also use this. My feeling is that this should be placed into the HSBType in the same way as I did with the RGB conversion some time back.

Looking at the Tradfri code, it's very similar to the code that was added to ZigBee, but I think is slightly different interface.

I open this PR for discussion to see if it is a good idea to get these into the library as common code.

#4957 

### DIFF
--- a/bundles/core/org.eclipse.smarthome.core.test/src/test/java/org/eclipse/smarthome/core/library/types/HSBTypeTest.java
+++ b/bundles/core/org.eclipse.smarthome.core.test/src/test/java/org/eclipse/smarthome/core/library/types/HSBTypeTest.java
@@ -113,4 +113,17 @@ public class HSBTypeTest {
         assertEquals(null, new HSBType("100,100,100").as(PointType.class));
     }
 
+    @Test
+    public void testConversionToXY() {
+        HSBType hsb = new HSBType("220,90,50");
+        PercentType[] xy = hsb.toXY();
+        assertEquals(new PercentType("16.969364"), xy[0]);
+        assertEquals(new PercentType("12.379659"), xy[1]);
+    }
+
+    @Test
+    public void testCreateFromXY() {
+        HSBType hsb = HSBType.fromXY(5f, 3f);
+        assertEquals(new HSBType("11,100,100"), hsb);
+    }
 }
--- a/bundles/core/org.eclipse.smarthome.core/src/main/java/org/eclipse/smarthome/core/library/types/HSBType.java
+++ b/bundles/core/org.eclipse.smarthome.core/src/main/java/org/eclipse/smarthome/core/library/types/HSBType.java
@@ -46,6 +46,14 @@ public class HSBType extends PercentType implements ComplexType, State, Command 
     public static final HSBType GREEN = new HSBType("120,100,100");
     public static final HSBType BLUE = new HSBType("240,100,100");
 
+    // 1931 CIE XYZ to sRGB (D65 reference white)
+    private static float Xy2Rgb[][] = { { 3.2406f, -1.5372f, -0.4986f }, { -0.9689f, 1.8758f, 0.0415f },
+            { 0.0557f, -0.2040f, 1.0570f } };
+
+    // sRGB to 1931 CIE XYZ (D65 reference white)
+    private static float Rgb2Xy[][] = { { 0.4124f, 0.3576f, 0.1805f }, { 0.2126f, 0.7152f, 0.0722f },
+            { 0.0193f, 0.1192f, 0.9505f } };
+
     protected BigDecimal hue;
     protected BigDecimal saturation;
 
@@ -118,6 +126,36 @@ public class HSBType extends PercentType implements ComplexType, State, Command 
 
         return new HSBType(new DecimalType((int) tmpHue), new PercentType((int) tmpSaturation),
                 new PercentType((int) tmpBrightness));
+    }
+
+    /**
+     * Returns a HSBType object representing the provided xy color values in CIE XY color model.
+     * Conversion from CIE XY color model to sRGB using D65 reference white
+     * Returned color is set to full brightness
+     *
+     * @param x, y color information 0.0 - 1.0
+     *
+     * @return new HSBType object representing the given CIE XY color, full brightness
+     */
+    public static HSBType fromXY(float x, float y) {
+        float Yo = 1.0f;
+        float X = (Yo / y) * x;
+        float Z = (Yo / y) * (1.0f - x - y);
+
+        float r = X * Xy2Rgb[0][0] + Yo * Xy2Rgb[0][1] + Z * Xy2Rgb[0][2];
+        float g = X * Xy2Rgb[1][0] + Yo * Xy2Rgb[1][1] + Z * Xy2Rgb[1][2];
+        float b = X * Xy2Rgb[2][0] + Yo * Xy2Rgb[2][1] + Z * Xy2Rgb[2][2];
+
+        float max = r > g ? r : g;
+        if (b > max) {
+            max = b;
+        }
+
+        r = gammaCompress(r / max);
+        g = gammaCompress(g / max);
+        b = gammaCompress(b / max);
+
+        return HSBType.fromRGB((int) (r * 255.0f + 0.5f), (int) (g * 255.0f + 0.5f), (int) (b * 255.0f + 0.5f));
     }
 
     @Override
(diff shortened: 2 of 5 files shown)


bundles/core/org.eclipse.smarthome.core/src/main/java/org/eclipse/smarthome/core/library/types/HSBType.java
```diff
@@ -257,6 +295,57 @@ public boolean equals(Object obj) {
...
+    /**
+     * Returns the xyY values representing this object's color in CIE XY color model.
+     * Conversion from sRGB to CIE XY using D65 reference white
+     * xy pair contains color information
+     * Y represents relative luminance
+     *
+     * @param HSBType color object
+     * @return PercentType[x, y, Y] values in the CIE XY color model
+     */
+    public PercentType[] toXY() {
```
sjka: The `fromXY(...)` method above takes `float` arguments, while `toXY(...)` here returns a `PrecentType` array. Is this asymmetry intended?

bundles/core/org.eclipse.smarthome.core.test/src/test/java/org/eclipse/smarthome/core/library/types/HSBTypeTest.java
```diff
@@ -113,4 +113,17 @@ public void testConversionToPointType() {
...
+    public void testConversionToXY() {
+        HSBType hsb = new HSBType("220,90,50");
+        PercentType[] xy = hsb.toXY();
+        assertEquals(new PercentType("16.969364"), xy[0]);
+        assertEquals(new PercentType("12.379659"), xy[1]);
+    }
+
+    @Test
+    public void testCreateFromXY() {
+        HSBType hsb = HSBType.fromXY(5f, 3f);
```
sjka: wouldn't it make sense to use the same values as above in order to demonstrate that they are actually inverse operations?

maggu2810:
Okay for me to add that conversion to the HSB type.

Just a question if we should consider this at a feature point:
Once the QuantityType is added, should we move the conversion between different representations (HSB, RGB, CIE XY) to a more generic part?

sjka:
> Once the QuantityType is added, should we move the conversion between different representations (HSB, RGB, CIE XY) to a more generic part?

I wouldn't think so, as QuantityType is a specialized DecimalType and only works with NumberItems, so I don't see how this can relate to HSBTypes. Or I'm thinking the wrong way and don't get your point. /cc @htreu

maggu2810:
My point has been: Isn't it all about a color space / model and HSB is only one representation? So, a tuple of numeric values and a special measurement system. Sure, it does not fit to the current changes where it is one numeric value -- not a tuple -- and one unit. It has been only a question that comes in my mind reading that topic. I just want to left a short notice if we should open a reminder to do it one time. If this is completely different, no need to discuss it further.

htreu:
For QuantityType there is currently no unit available to express a color in a single value. It would be possible though to express HSB with three QuantityType states (e.g. during measurement of a specific color). Since QuantityType is about expressing physical features we can imho ignore it wrt HSBType.

cdjackson:
I think this is different isn't it?. Here the binding gets an HSB from the user - always. The HSB needs to be converted to a different 'format' for transferring to the device. 

Aren't the Quantity conversions there to allow the user to select how they want to interact? Or maybe I'm wrong as I've not followed the quantity conversions too closely.

htreu:
Yes, @cdjackson nailed it: the frameworks representation for color is always the HSBType. Conversion takes place in the binding when communicating with the devices. So at least in this scenario conversion into other types is out of scope.

maggu2810:
> I think this is different isn't it?

Perhaps, sorry, my intention was not to start an off-topic discussion.

Should I move that questions about the quantity type to another issue or perhaps use the forum?
I assume it is more related to that changes as to this PR.

Isn't the quantity stuff about
* a "what" that defines the allowed units (e.g. temperature => celsius, fahrenheit or kelvin, mass => kg, g, t, ..., and so on)
* and a value that

My understanding has been that e.g. a channel represents a temperature (e.g. a target temperature), it's value is represent by a quantity type that consists of a number -- the temperature -- and an temperature measurement system specific unit (e.g. celsius, fahrenheit or kelvin).
If an command is received, the quantity type could use the same different units (still e.g. celsius, fahrenheit or kelvin).

For a color my understand has been that it still a measurement system and a value for that measurement system. The only difference is the chosen measurement system subset (so instead of "temperature (celsius, fahrenheit, kelvin)" it is "color space (rgb, hue, cie xy) and that the value could be a tuple of values.

BasvanH:
When will this branch be committed to master ?

maggu2810:
@BasvanH for me it seems as there are two review comments of @sjka that has not been commented / changed yet:
* https://github.com/eclipse/smarthome/pull/4959#discussion_r162911478
* https://github.com/eclipse/smarthome/pull/4959#discussion_r162910229

cdjackson:
I had hoped that the two authors of ZigBee conversions and Tradfri converters would comment here. I am happy to address the comments and make the interface symetric.

@BasvanH I assume you have a need for this - do you have any comments on the interface. ZigBee and Tradfri are both slightly different so if you have comments, please make them.

BasvanH:
I'm using the Hue bridge with Tradfri bulbs. The color can't be controlled from Openhab due to missing HSB type in Hue binding.

https://github.com/eclipse/smarthome/issues/4939
and
https://community.openhab.org/t/ikea-tradfri-color-changing-bulb-on-hue-bridge-no-control-of-colors/35413/28

kaikreuzer:
No, #4939 is completely unrelated to this one here.

bundles/core/org.eclipse.smarthome.core/src/main/java/org/eclipse/smarthome/core/library/types/HSBType.java
puzzle-star: I implemeted this trying to be coherent with the existing fromRGB / toRGB couple of methods already in HSBtype (fromRGB takes ints, while toRGB returns Percentypes). Other than that, there is no particular reason. I would prefer to use only floats in the conversions, but it is like this just for consistency

BasvanH:
Ok, sorry for the confusion. I will move back to that issue

triller-telekom:
@cdjackson Since the zigbee authors have not responded yet, how about if you adjust the PR with the 2 changes from @sjka and ping them with their github names here and if there is no reaction we merge it?

cdjackson:
@triller-telekom the zigbee author did respond (it's inline above so not so easy to see). He stated -:

> I implemeted this trying to be coherent with the existing fromRGB / toRGB couple of methods already in HSBtype (fromRGB takes ints, while toRGB returns Percentypes). Other than that, there is no particular reason. I would prefer to use only floats in the conversions, but it is like this just for consistency

So, the design choice is to keep it the same concept as existing methods in the class or make them symmetrical? To me it does make sense to keep the same concept as these methods - wdyt?

triller-telekom:
Usually its better to have them symmetric, but I guess its hard to change it for RGB now, so I would say to keep it the same concept as with RGB.

triller-telekom:
@cdjackson I was curious if we can use this conversion in the Tradfri binding. It turns out that your implementation differs from the Tradfri one.

You are using these matrices that [convert from sRGB to x/y](https://en.wikipedia.org/wiki/SRGB#The_forward_transformation_(CIE_XYZ_to_sRGB)) in the code [here](https://github.com/eclipse/smarthome/pull/4959/files#diff-607f27ccdbeb189fdd32114c5269ed9bR49)

However, @hreichert has used these matrices in the Tradfri binding (plus normalization to 0-65535, but we can leave that out for now):

https://github.com/eclipse/smarthome/blob/7adbe25f1f192ef2cfd2ad5354eed523484bda12/extensions/binding/org.eclipse.smarthome.binding.tradfri/src/main/java/org/eclipse/smarthome/binding/tradfri/internal/TradfriColor.java#L174-L178

Following [the link](http://www.brucelindbloom.com/index.html?Eqn_RGB_XYZ_Matrix.html) in the Tradfri implementation I see that there are multiple matrices for a conversion from RGB to x/y.

I could not find which matrix @hreichert used from there and to be honest I am a bit confused now. As far as I can see is that there is NO such thing as **THE** conversion matrix, but it depends on two factors: 1. the colour space (sRGB, Adobe RGB, etc) and 2. (the reference white). This speaks against adding the conversion to `HSBType`.

@hreichert: Our implementation in the `HSBType.toRGB()` returns you a `sRGB` value:

https://github.com/eclipse/smarthome/blob/9829331107f28fead670b741bce912b51d78a070/bundles/core/org.eclipse.smarthome.core/src/main/java/org/eclipse/smarthome/core/library/types/HSBType.java#L156-L167

But you are applying a different matrix, so is this a bug in your code? Can you use the matrix from this PR in the Tradfri binding? If I use the matrix from this PR the values here:

https://github.com/eclipse/smarthome/blob/9829331107f28fead670b741bce912b51d78a070/extensions/binding/org.eclipse.smarthome.binding.tradfri.test/src/test/java/org/eclipse/smarthome/binding/tradfri/internal/TradfriColorTest.java#L99-L100

change to `x=41947` and `y=21625`. These should be the correct ones since you get the `sRGB` values from the `HSBType`. Can you please test that with your bulbs and see what difference it makes?

If we can agree on that our `reference white` is the `D65`, like we obviously agreed on that `sRGB` is our default RGB color space, I think we can add this conversion to the `HSBType`.

@cdjackson In case we come to the conclusion to merge this PR, please address @sjka https://github.com/eclipse/smarthome/pull/4959#discussion_r162911478 on the test code.

cdjackson:
Just a point to note that the ZigBee implementation was implemented specifically for Tradfri bulbs. I don’t know if the Tradfri hub does something different, but when the bulb is connected via ZigBee the conversion works (I assume - I don’t have a Tradfri color bulb myself, but it was tested extensively by @puzzle-star who does).

I’m happy to update the comment from @sjka once we agree on the merge ;)

hreichert:
@triller-telekom 
Thanks for your analysis. It's absolutely possible that I used a _wrong_ conversion matrix.
Back then I took a matrix that "looked good" on the bulbs and gave colors that I expected.
I can test the other matrix on the color bulbs, but next weekend at the earliest.

triller-telekom:
@hreichert: Great! Please test it with the matrix from this PR on the real hardware. If it works I would make a PR against this PR to integrate it into the Tradfri binding, so once this PR is merged we only have **one** `HSBType` to x/y conversion left in ESH :)

hreichert:
@triller-telekom Sounds cool! I do the tests as soon as possible.

triller-telekom:
@hreichert I have found a bulb in the office and tested it a bit. The X and Y values look good, they are even gamma corrected now with the code from @cdjackson.

But I am totally puzzled about the brightness and how you calculate it. There are many conversions involved, i.e from PercentType to 0-255, to HSBType etc.

I have put my changes in an extra branch, feel free to start from there if you try it with your hardware: https://github.com/triller-telekom/smarthome/tree/tradfriXY

triller-telekom:
@cdjackson I bought myself a Tradfri color bulb and created a PR against your PR so that Tradfri automatically uses your conversions: https://github.com/cdjackson/smarthome/pull/3

Please have a look at it.

triller-telekom:
@hreichert For testing, please have a look at my changes in https://github.com/cdjackson/smarthome/pull/3

kaikreuzer:
Looks as if this PR needs a rebase now...

cdjackson:
I've resolved the conflicts. @triller-telekom you probably should take a look as the issues were in the Tradfri code ;)

There's now an ip validation error, but it looks like everything is signed ok.

kaikreuzer:
@cdjackson Fyi, this is now available on openHAB side, so you could adapt the Zigbee implementation to use it.

cdjackson:
Thanks @kaikreuzer - I already plan to consolidate this, but I'm blocked with any ZigBee changes at the moment until we resolve the repository problems. Once this is done I'll refactor to account for this.

PR #4962, opened 2018-01-22
Reload discovery result from storage before emitting event

### LINKED ISSUE #4360: Paper UI: new discovered things should be displayed as ignored when they are auto ignored

htreu:
This fixes #4360 by reloading the discovery result from the storage before emitting an event. This guarantees that changes to the result (like setting the flag to IGNORED) are part of the event.


sjka:
Nice catch, thanks!

PR #4963, opened 2018-01-22
[audio] Improvements in Audio and Voice API

cweitkamp:
- Added annotiations for `AudioManager` and `VoiceManager` API 
- Added parameter to adjust the volume for `playSound()` and `say()` methods
- Moved handling of the volume to `AudioManagerImpl` and `VoiceManagerImpl`
- Added check if an `AudioSink` supports the given `AudioStream` before processing it (see https://github.com/eclipse/smarthome/pull/3764#issuecomment-312233080)
- Code cleanup
- Added documentation for audio & voice features (e.g. built-in functions within DSL rules or console commands)

See discussion in https://github.com/eclipse/smarthome/issues/4832#issuecomment-355991197.


bundles/core/org.eclipse.smarthome.core.audio/src/main/java/org/eclipse/smarthome/core/audio/AudioManager.java
```diff
@@ -44,27 +45,68 @@
...
     void play(AudioStream audioStream, String sinkId);
 
+    /**
+     * Plays the passed audio stream on the given sink.
+     *
+     * @param audioStream The audio stream to play
+     * @param sinkId The id of the audio sink to use or null
+     * @param volume The volume to be used or null
+     */
+    void play(AudioStream audioStream, String sinkId, PercentType volume);
```
amitjoy: As you have mentioned that `volume` and `sinkdId` can be null, what would happen if `audioStream` is null? Shouldn't it throw NPE or IAE? I believe such invariants should be clearly added in the javadoc as it is a public API.

bundles/core/org.eclipse.smarthome.core.audio/src/main/java/org/eclipse/smarthome/core/audio/AudioManager.java
```diff
@@ -44,27 +45,68 @@
...
     void playFile(String fileName) throws AudioException;
 
+    /**
+     * Plays an audio file with the given volume from the "sounds" folder using the default audio sink.
+     *
+     * @param fileName The file from the "sounds" folder
+     * @param volume The volume to be used or null
+     * @throws AudioException in case the file does not exist or cannot be opened
+     */
+    void playFile(String fileName, PercentType volume) throws AudioException;
```
amitjoy: Similarly here

bundles/core/org.eclipse.smarthome.core.audio/src/main/java/org/eclipse/smarthome/core/audio/AudioManager.java
```diff
@@ -44,27 +45,68 @@
...
 
+    /**
+     * Plays an audio file with the given volume from the "sounds" folder using the given audio sink.
+     *
+     * @param fileName The file from the "sounds" folder
+     * @param sink The id of the audio sink to use or null
+     * @param volume The volume to be used or null
+     * @throws AudioException in case the file does not exist or cannot be opened
+     */
+    void playFile(String fileName, String sink, PercentType volume) throws AudioException;
```
amitjoy: Actually my previous comment applies to all the methods but I am only pointing to the changes that you have incorporated.

bundles/core/org.eclipse.smarthome.core.audio/src/main/java/org/eclipse/smarthome/core/audio/internal/AudioManagerImpl.java
```diff
@@ -90,51 +91,80 @@ public void play(AudioStream audioStream) {
...
-                    sink.process(audioStream);
-                } catch (UnsupportedAudioFormatException | UnsupportedAudioStreamException e) {
-                    logger.error("Error playing '{}': {}", audioStream.toString(), e.getMessage());
+                // get current volume
+                PercentType oldVolume = getVolume(sinkId);
+                // set notification sound volume
+                if (volume != null) {
+                    setVolume(volume, sinkId);
+                }
+//                if (sink.getSupportedStreams().contains(audioStream.getClass())) {
```
amitjoy: If not required, you can actually remove such unused code comment blocks

bundles/core/org.eclipse.smarthome.core.voice/src/main/java/org/eclipse/smarthome/core/voice/VoiceManager.java
```diff
@@ -35,6 +37,14 @@
      */
     void say(String text);
 
+    /**
+     * Speaks the passed string with the given volume using the default TTS service and default audio sink.
+     *
+     * @param text The text to say
+     * @param volume The volume to be used or null
+     */
+    void say(String text, PercentType volume);
```
amitjoy: similarly here in this interface

bundles/core/org.eclipse.smarthome.core.voice/src/main/java/org/eclipse/smarthome/core/voice/internal/VoiceManagerImpl.java
```diff
@@ -167,29 +184,39 @@ public void say(String text, String voiceIda, String sinkId) {
...
-                        sink.process(audioStream);
-                    } catch (UnsupportedAudioFormatException | UnsupportedAudioStreamException e) {
-                        logger.error("Error saying '{}': {}", text, e.getMessage());
+                    // get current volume
+                    PercentType oldVolume = audioManager.getVolume(sinkId);
+                    // set notification sound volume
+                    if (volume != null) {
+                        audioManager.setVolume(volume, sinkId);
+                    }
+//                    if (sink.getSupportedStreams().contains(audioStream.getClass())) {
```
amitjoy: unused code comments

bundles/core/org.eclipse.smarthome.core.voice/src/main/java/org/eclipse/smarthome/core/voice/internal/VoiceManagerImpl.java
```diff
@@ -167,29 +184,39 @@ public void say(String text, String voiceIda, String sinkId) {
...
+                        try {
+                            sink.process(audioStream);
+                        } catch (UnsupportedAudioFormatException | UnsupportedAudioStreamException e) {
+                            logger.warn("Error saying '{}': {}", text, e.getMessage(), e);
+                        }
+//                    } else {
```
amitjoy: >>

bundles/core/org.eclipse.smarthome.core.audio/src/main/java/org/eclipse/smarthome/core/audio/AudioManager.java
cweitkamp: That's a good question. Currenlty the `AudioManagerImpl` drops it silently (see https://github.com/eclipse/smarthome/pull/4963/files#diff-93f4fd35ca9478d18d70f62ae3285f19R98). No logging, no exception. I suggest to throw an IAE in this case.

bundles/core/org.eclipse.smarthome.core.audio/src/main/java/org/eclipse/smarthome/core/audio/AudioManager.java
cweitkamp: Fair enough.

bundles/core/org.eclipse.smarthome.core.audio/src/main/java/org/eclipse/smarthome/core/audio/internal/AudioManagerImpl.java
cweitkamp: I don't want to remove them, I want to uncomment them. If I do Travis fails (see https://travis-ci.org/eclipse/smarthome/builds/331787920). But I'm not sure why. May I ask you to help me with that? I somehow have the feeling that it is related to your previous commt on the `audioSink` is null issue
(https://github.com/eclipse/smarthome/pull/4963/files#r163219285).

bundles/core/org.eclipse.smarthome.core.audio/src/main/java/org/eclipse/smarthome/core/audio/AudioManager.java
cweitkamp: I guess we have a NPE or `AudioException` here. The `fileName` will be extended with the path to the "sound" folder and directly passed to the constructor of a `File` object.

bundles/core/org.eclipse.smarthome.core.audio/src/main/java/org/eclipse/smarthome/core/audio/AudioManager.java
cweitkamp: Do you think we should add annotations to this API?

bundles/core/org.eclipse.smarthome.core.audio/src/main/java/org/eclipse/smarthome/core/audio/AudioManager.java
amitjoy: For granularity, it is always a good idea to throw specific exception, that is, for null scenario - NPE or IAE, usw.

bundles/core/org.eclipse.smarthome.core.audio/src/main/java/org/eclipse/smarthome/core/audio/AudioManager.java
amitjoy: could be a solution as well

bundles/core/org.eclipse.smarthome.core.audio/src/main/java/org/eclipse/smarthome/core/audio/internal/AudioManagerImpl.java
amitjoy: You mean when you remove them travis fails?

cweitkamp: Yes, I exaclty. But I found the problem now. The condition `sink.getSupportedStreams().contains(audioStream.getClass())` is to sloppy.

cweitkamp:
@amitjoy Thanks for your review. I added NPEs in every API.

bundles/core/org.eclipse.smarthome.core.audio/src/main/java/org/eclipse/smarthome/core/audio/internal/AudioManagerImpl.java
```diff
@@ -90,51 +91,89 @@ public void play(AudioStream audioStream) {
...
     }
 
     @Override
     public void playFile(String fileName, String sink) throws AudioException {
+        playFile(fileName, sink, null);
+    }
+
+    @Override
+    public void playFile(String fileName, String sink, PercentType volume) throws AudioException {
+        if (fileName == null) {
```
amitjoy: Just a suggestion: `Objects#requireNonNull(Object, String)`

bundles/core/org.eclipse.smarthome.core.audio/src/main/java/org/eclipse/smarthome/core/audio/internal/AudioManagerImpl.java
```diff
@@ -90,51 +91,89 @@ public void play(AudioStream audioStream) {
...
+    }
 
-        if (sink != null) {
-            try {
-                sink.process(audioStream);
-            } catch (UnsupportedAudioFormatException | UnsupportedAudioStreamException e) {
-                logger.error("Error playing '{}': {}", url, e.getMessage());
-            }
+    public void stream(String url, String sinkId, PercentType volume) throws AudioException {
+        if (url == null) {
```
amitjoy: `Objects#requireNonNull(Object, String)`

bundles/core/org.eclipse.smarthome.core.audio/src/main/java/org/eclipse/smarthome/core/audio/AudioManager.java
```diff
@@ -41,39 +43,100 @@
...
+
     /**
      * Stream audio from the passed url to the given sink
      *
      * @param url The url to stream from or null if streaming should be stopped
      * @param sinkId The id of the audio sink to use or null
      * @throws AudioException in case the url stream cannot be opened
+     * @throws NullPointerException If url is null
      */
     void stream(String url, String sinkId) throws AudioException;
```
amitjoy: javadoc says url can be null.

bundles/core/org.eclipse.smarthome.core.audio/src/main/java/org/eclipse/smarthome/core/audio/AudioManager.java
```diff
@@ -41,39 +43,100 @@
...
+    void playFile(String fileName, String sink, PercentType volume) throws AudioException;
+
     /**
      * Stream audio from the passed url using the default audio sink.
      *
+     * @param url The url to stream from or null if streaming should be stopped
      * @throws AudioException in case the url stream cannot be opened
+     * @throws NullPointerException If url is null
      */
     void stream(String url) throws AudioException;
```
amitjoy: javadoc says url can be null

bundles/core/org.eclipse.smarthome.core.voice.test/src/test/java/org/eclipse/smarthome/core/voice/internal/SinkStub.java
```diff
@@ -62,9 +67,7 @@ public void process(AudioStream audioStream) throws UnsupportedAudioFormatExcept
 
     @Override
     public Set<AudioFormat> getSupportedFormats() {
-        supportedFormats.add(AudioFormat.MP3);
-        supportedFormats.add(AudioFormat.WAV);
-        return supportedFormats;
+        return SUPPORTED_AUDIO_FORMATS;
```
amitjoy: You can return an immutable view

bundles/core/org.eclipse.smarthome.core.voice.test/src/test/java/org/eclipse/smarthome/core/voice/internal/SinkStub.java
```diff
@@ -89,7 +92,6 @@ public String toString() {
 
     @Override
     public Set<Class<? extends AudioStream>> getSupportedStreams() {
-        // this method will not be used in the tests
-        return null;
+        return SUPPORTED_AUDIO_STREAMS;
```
amitjoy: You can return an immutable view

bundles/core/org.eclipse.smarthome.core.audio/src/main/java/org/eclipse/smarthome/core/audio/internal/AudioManagerImpl.java
cweitkamp: Hm, ... I think when sca tool failed because of throwing NPEs it's not by intention of the designer to use a workaround instead. I changed my mind and will use annotations for both APIs.

cweitkamp:
Ready for final review.

bundles/core/org.eclipse.smarthome.core.audio/src/main/java/org/eclipse/smarthome/core/audio/AudioManager.java
```diff
@@ -42,28 +46,69 @@
...
+     * @param fileName The file from the "sounds" folder
      * @throws AudioException in case the file does not exist or cannot be opened
      */
     void playFile(String fileName) throws AudioException;
 
+    /**
+     * Plays an audio file with the given volume from the "sounds" folder using the default audio sink.
+     *
+     * @param fileName The file from the "sounds" folder
+     * @param volume The volume to be used or null
```
maggu2810: Here and on some other functions I miss a statement what will be done if the parameter is `null`.
On some places you already added it e.g.

```text
@param url The url to stream from or null if streaming should be stopped
```

but e.g. on the volume it is missing. Should it be "or null if the default notification volume should be used"?

bundles/core/org.eclipse.smarthome.core.audio/src/main/java/org/eclipse/smarthome/core/audio/internal/AudioConsoleCommandExtension.java
```diff
@@ -45,9 +46,9 @@ public AudioConsoleCommandExtension() {
     @Override
     public List<String> getUsages() {
         return Arrays.asList(new String[] {
-                buildCommandUsage(SUBCMD_PLAY + " <sink> <filename>",
+                buildCommandUsage(SUBCMD_PLAY + " <sink> <filename> <volume>",
```
maggu2810: It has not bee really correct before (IMHO), but as you touched it anyway shouldn't it be changed to

```text
" <sink> [<filename> [<volume>]]"
```

bundles/core/org.eclipse.smarthome.core.voice.test/src/test/java/org/eclipse/smarthome/core/voice/internal/SinkStub.java
```diff
@@ -28,19 +29,24 @@
...
- *
  * @author Velin Yordanov - migrated from groovy to java
- *
  */
 public class SinkStub implements AudioSink {
 
     private boolean isStreamProcessed;
-    private boolean isUnsupportedAudioFormatExceptionExpected;
-    private Set<AudioFormat> supportedFormats = new HashSet<AudioFormat>();
+    private static final Set<AudioFormat> SUPPORTED_AUDIO_FORMATS = new HashSet<>();
```
maggu2810: Perhaps we could drop the static block below and the unmodfiable set creation on the get method at all by using something like that here:

```Java
Collections.unmodifiableSet(Stream.of(AudioFormat.WAV, AudioFormat.MP3, ...).collect(Collectors.toSet())));
```

```Java
Collections.unmodifiableSet(new HashSet(Arrays.asList(AudioFormat.WAV, AudioFormat.MP3, ...)));
```

cweitkamp: Why do you think we should change that? Is the static initialization block not compliant to the guidelines?

maggu2810: Everytime `getSupportedForrmats()` and `getSupportedStreams()` a new unmodifiable collection / view is created. Is there any need for? Why not create that unmodifiable view once and reuse that reference at every get... call?

cweitkamp: Makes sense. I misunderstood your first comment.

cweitkamp:
Where is the right place to put some documentation?

sjka:
I'd say "it depends" 😉 

Documentation about how to use & configure it would go somewhere beneath the [features](https://www.eclipse.org/smarthome/documentation/features/index.html) section, any guidance wrt extending it and providing own implementations of certain audio/voice services would go beneath the [development](https://www.eclipse.org/smarthome/documentation/development/ide.html) section. They both don't have any "Audio & Voice" sub-sections yet, therefore feel free to create them.

kaikreuzer:
@cweitkamp Please ping us once you added some docs (or if you further want to discuss the right place for it).

cweitkamp:
@kaikreuzer Ping.

I added a short documentation for the features section. I hope I didn't mix up too much of ESH and OH in there. I'm not sure how to handle the structure for the menu. Will it be generated automatically? Feel free to extend what I have got so far.

bundles/core/org.eclipse.smarthome.core.audio/src/main/java/org/eclipse/smarthome/core/audio/internal/AudioManagerImpl.java
```diff
@@ -95,32 +95,33 @@ public void play(AudioStream audioStream, String sinkId) {
...
-                if (oldVolume != null) {
-                    setVolume(oldVolume, sinkId);
+        if (audioStream == null) {
+            throw new NullPointerException("Audio stream cannot be played as it is null.");
+        }
+
+        AudioSink sink = getSink(sinkId);
+        if (sink != null) {
+            // get current volume
+            PercentType oldVolume = getVolume(sinkId);
```
lolodomo: You should handle volume only if you process the stream.
So I will suggest to move your "volume" stuff inside the if just below.

bundles/core/org.eclipse.smarthome.core.voice/src/main/java/org/eclipse/smarthome/core/voice/internal/VoiceManagerImpl.java
```diff
@@ -191,18 +191,24 @@ public void say(String text, String voiceId, String sinkId, PercentType volume)
...
                     AudioStream audioStream = tts.synthesize(text, voice, audioFormat);
 
-                    try {
-                        // get current volume
-                        PercentType oldVolume = audioManager.getVolume(sinkId);
-                        // set notification sound volume
-                        if (volume != null) {
-                            audioManager.setVolume(volume, sinkId);
+                    // get current volume
+                    PercentType oldVolume = audioManager.getVolume(sinkId);
```
lolodomo: You should handle volume only if you process the stream.
So I will suggest to move your "volume" stuff inside the if just below.

lolodomo:
I am very surprised to not see any change in the Sonos binding while this binding is already handling the volume save/restore. That is certainly something o check first.

I added two comments about doing volume stuff only when the audio stream is processed.

bundles/core/org.eclipse.smarthome.core.audio/src/main/java/org/eclipse/smarthome/core/audio/internal/AudioConsoleCommandExtension.java
```diff
@@ -46,9 +46,9 @@ public AudioConsoleCommandExtension() {
     @Override
     public List<String> getUsages() {
         return Arrays.asList(new String[] {
-                buildCommandUsage(SUBCMD_PLAY + " <sink> <filename> <volume>",
+                buildCommandUsage(SUBCMD_PLAY + " <sink> [<filename> [<volume>]]",
```
lolodomo: Reading your code, I understand the case with only one parameter is not the sink but the filename.
So you should put two entries in this list.

bundles/core/org.eclipse.smarthome.core.audio/src/main/java/org/eclipse/smarthome/core/audio/internal/AudioConsoleCommandExtension.java
```diff
@@ -46,9 +46,9 @@ public AudioConsoleCommandExtension() {
...
                         "plays a sound file from the sounds folder through the optionally specified audio sink(s)"),
-                buildCommandUsage(SUBCMD_STREAM + " <sink> <url> <volume>",
+                buildCommandUsage(SUBCMD_STREAM + " <sink> [<url> [<volume>]]",
```
lolodomo: Same here

lolodomo:
Change looks good.

Another comment about the console commands.

lolodomo:
@cweitkamp : IMHO your code will not work at all with the Sonos binding because you first get the volume ffor the sink and then restore this volume at the end. But in the case of the Sonos binding, what you will get is the notification volume and so the restored volume at the end will be the notification volume, not the original volume.

Another problem is that you apply a change of volume on the original playback and then process the notification stream. So the user will listen its current playback with a new volume during a little time. Could be a problem if the notification volume is high.

My feeling is that the volume should be a parameter of the process method and the audio sinks should handle it. So it will not be the audio manager that change the volume but the audio sink when processing the stream.

lolodomo:
To not break the compatibility with all bindings using the audio feature, I would suggest a second `process` method in the interface with the stream as first parameter and a volume as second parameter.


For the first problem, the bug could be in the Sonos binding as the interface documentation says that getVolume() should return the current volume and this is not what the Sonos binding returns.

bundles/core/org.eclipse.smarthome.core.audio/src/main/java/org/eclipse/smarthome/core/audio/internal/AudioConsoleCommandExtension.java
cweitkamp: I see.

cweitkamp:
> IMHO your code will not work at all with the Sonos binding because you first get the volume ffor the sink and then restore this volume at the end. But in the case of the Sonos binding, what you will get is the notification volume and so the restored volume at the end will be the notification volume, not the original volume.

@lolodomo Thank you for your opinion and starting a discussion on this. I had my concerns, too. But IMHO the implementation of the `getVolume()` and `setVolume()` methods for the audio sink interface in the Sonos binding - and some others which used the Sonos binding as role model - are not done in the expected way. Don't get me wrong. I like the idea to have a separate volume for notifications. But we have some other DSL built-in function (e.g. `getMasterVolume()` and `setMasterVolume()`), which uses the same methods from the `AudioManager` and `AudioSink` interfaces to handle volumes in the framework.

> Another problem is that you apply a change of volume on the original playback and then process the notification stream. So the user will listen its current playback with a new volume during a little time.

Yes, that may happen and a possible solution for that could be different for each binding.

> I would suggest a second process method in the interface with the stream as first parameter and a volume as second parameter.

Good idea, but how does it solve the issue to be not API breaking for existing bindings? Let's look some steps ahead. If we add this new method to the `AudioSink` interface we break the bindings in the same way. That leads to a change in all existing bindings anyways. I suggest to introduce an `abstract` `AudioSink` implementation which can be used by the bindings to implement audio sinks and e.g. handles the volume in the proposed `process(AudioStream audioStream, PercentType volume)` method instead of doing it in the `AudioManagerImpl`. In this case each binding can take care on the above problem on its own if it is necessary. Additionally we can think about differ between basic audio handling like we have right now and handling notifications. Both features will blow up this PR.

@kaikreuzer May I ask for your opinion?

kaikreuzer:
I am not sure I can follow you in all details, but I agree with @lolodomo that the current implementation probably has issues with restoring the volume as it isn't ensured that the AudioManager knows when the playback of the audio stream is actually ending - which would be important in order to know when to restore the old volume. So this should be better left to the individual implementations, for which we would need the volume as an additional (while optional) parameter on the process method.
Whether something is a notification (and thus requires a certain volume) or not should indeed be left to the caller and thus out of scope of this PR.

bundles/core/org.eclipse.smarthome.core.audio.test/src/test/groovy/org/eclipse/smarthome/core/audio/test/AudioConsoleTest.groovy
```diff
@@ -62,6 +63,30 @@ public class AudioConsoleTest extends AudioOSGiTest {
         assertCompatibleFormat()
     }
 
+    @Test
+    public void 'audio console plays file for a specified sink with a specified volume'(){
```
kaikreuzer: In the light of https://github.com/eclipse/smarthome/issues/3305, please do not add new Groovy tests, but rather help migrating the few existing ones in here to Java - thanks!

bundles/core/org.eclipse.smarthome.core.audio/src/main/java/org/eclipse/smarthome/core/audio/AudioManager.java
```diff
@@ -42,28 +47,69 @@
...
+
+    /**
+     * Stream audio with the given volume from the passed url using the default audio sink.
+     *
+     * @param url The url to stream from or null if streaming should be stopped
+     * @param volume The volume to be used or null if the default notification volume should be used
      * @throws AudioException in case the url stream cannot be opened
      */
-    void stream(String url) throws AudioException;
+    void stream(@Nullable String url, @Nullable PercentType volume) throws AudioException;
```
kaikreuzer: I'm actually wondering if we should offer that for streams at all - a stream should never be a notification (but rather some background audio like radio) and it does not have an end when an old volume would be restored.
I'd therefore feel better to always play streams with the current volume and leave it to the user to adjust it at any time.

bundles/core/org.eclipse.smarthome.core.audio/src/main/java/org/eclipse/smarthome/core/audio/AudioManager.java
```diff
@@ -72,23 +118,33 @@
...
+
+    /**
+     * Stream audio with the given volume from the passed url to the given sink
+     *
+     * @param url The url to stream from or null if streaming should be stopped
+     * @param sinkId The id of the audio sink to use or null
+     * @param volume The volume to be used or null if the default notification volume should be used
+     * @throws AudioException in case the url stream cannot be opened
+     */
+    void stream(@Nullable String url, @Nullable String sinkId, @Nullable PercentType volume) throws AudioException;
```
kaikreuzer: see above

bundles/core/org.eclipse.smarthome.core.audio/src/main/java/org/eclipse/smarthome/core/audio/AudioManager.java
```diff
@@ -72,23 +118,33 @@
...
 
     /**
      * Retrieves the current volume of a sink
      *
-     * @param sinkId the sink to get the volume for
+     * @param sinkId the sink to get the volume for or null
      * @return the volume as a value between 0 and 100
      */
-    PercentType getVolume(String sinkId);
+    PercentType getVolume(@Nullable String sinkId);
```
kaikreuzer: Why Nullable? What is returned if null is passed?

bundles/core/org.eclipse.smarthome.core.audio/src/main/java/org/eclipse/smarthome/core/audio/AudioManager.java
```diff
@@ -72,23 +118,33 @@
...
 
     /**
      * Sets the volume for a sink.
      *
      * @param volume the volume to set as a value between 0 and 100
-     * @param sinkId the sink to set the volume
+     * @param sinkId the sink to set the volume or null
      */
-    void setVolume(PercentType volume, String sinkId);
+    void setVolume(PercentType volume, @Nullable String sinkId);
```
kaikreuzer: Why Nullable? What is returned if null is passed?

bundles/core/org.eclipse.smarthome.core.audio/src/main/java/org/eclipse/smarthome/core/audio/AudioManager.java
cweitkamp: You can use `null` to get the volume of the default audio sink. I will clarify it in the JavaDoc.

bundles/core/org.eclipse.smarthome.core.audio.test/src/test/groovy/org/eclipse/smarthome/core/audio/test/AudioConsoleTest.groovy
cweitkamp: I keep this in mind and submit a follow up PR.

htreu:
@cweitkamp may I ask you to have a look again:
The `AudioManager` defines a `play(url, sinkId)` method with url being Nullable (see [here](https://github.com/eclipse/smarthome/blob/master/bundles/core/org.eclipse.smarthome.core.audio/src/main/java/org/eclipse/smarthome/core/audio/AudioManager.java#L112)).
The `AudioManagerImpl` tries to create a AudioStream from the url but may be null afterwards. The `play(audioStream, sinkId)` method however is NonNull through the `@NonNullByDefault` annotation from the interface. My IDE complains about this discrepancy.
Also: The JavaDoc on the `play(url, sinkId)` says `@param url The url to stream from or null if streaming should be stopped` but the implementation does not live up to this promise.

lolodomo:
Does this merge finally breaks Sonos binding as I suspected ?

cweitkamp:
@htreu Yes, I will look into that.

@lolodomo No, I don't think so. The binding currently works like before. But you will see an unexpected behavior if you are using one of the new features from Karaf console or DSL rules.

cweitkamp:
@htreu You are right. My PR introduced a regression. Currently you are not able to stop any playing stream within the framework. I can fix that. The sad thing is no existing binding supports this feature. Every `AudioSink` implementation throws an exception (`UnsupportedAudioStreamException`) if the class of the audio stream isn't an instance of a supported stream.

cweitkamp:
I submitted a fix for it in PR #5241.

lolodomo:
@cweitkamp : if volume is not set, the Sonos binding should continue working except that you forgot to check if volume is not null at these lines when restoring the volume:
https://github.com/eclipse/smarthome/blob/master/bundles/core/org.eclipse.smarthome.core.audio/src/main/java/org/eclipse/smarthome/core/audio/internal/AudioManagerImpl.java#L115
https://github.com/eclipse/smarthome/blob/master/bundles/core/org.eclipse.smarthome.core.voice/src/main/java/org/eclipse/smarthome/core/voice/internal/VoiceManagerImpl.java#L211

lolodomo:
This is not the problem. When the volume save/restore is handled by the audio sink which is the case by the Sonos binding, the audio manager should not set the volume before processing the stream and should not restore the volume at the end (as correctly handled by the audio sink).

cweitkamp:
Alright, now I understand what you're meaning. That's true. We can think about skipping the `restoreVolume()` if we didn't set it before.

lolodomo:
Exactly.

lolodomo:
An issue certainly for you
https://community.openhab.org/t/web-audio-sink-does-not-support-volume-level-changes/41989

cweitkamp:
I see. A possible `IOException` from `getVolume()` or `setVolume()` methods isn't handled. I'll take care of that.

cweitkamp:
PR for both proposals #5253.

PR #4967, opened 2018-01-23
[Astro] Do not restart but reconfigure discovery service with new config

triller-telekom:
See https://github.com/eclipse/smarthome/pull/4079#issuecomment-359023592

### DIFF
--- a/extensions/binding/org.eclipse.smarthome.binding.astro/src/main/java/org/eclipse/smarthome/binding/astro/internal/discovery/AstroDiscoveryService.java
+++ b/extensions/binding/org.eclipse.smarthome.binding.astro/src/main/java/org/eclipse/smarthome/binding/astro/internal/discovery/AstroDiscoveryService.java
@@ -30,7 +30,10 @@ import org.eclipse.smarthome.core.i18n.TranslationProvider;
 import org.eclipse.smarthome.core.library.types.PointType;
 import org.eclipse.smarthome.core.thing.ThingTypeUID;
 import org.eclipse.smarthome.core.thing.ThingUID;
+import org.osgi.service.component.annotations.Activate;
 import org.osgi.service.component.annotations.Component;
+import org.osgi.service.component.annotations.Deactivate;
+import org.osgi.service.component.annotations.Modified;
 import org.osgi.service.component.annotations.Reference;
 import org.slf4j.Logger;
 import org.slf4j.LoggerFactory;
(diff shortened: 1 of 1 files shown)


triller-telekom:
close and reopen for travis restart due to strange dmx binding tests failure (could not open a socket...)

extensions/binding/org.eclipse.smarthome.binding.astro/src/main/java/org/eclipse/smarthome/binding/astro/internal/discovery/AstroDiscoveryService.java
```diff
@@ -61,15 +64,23 @@ public AstroDiscoveryService() {
     }
 
     @Override
+    @Activate
```
kaikreuzer: You do not need activate and deactivate annotations, it is only required for modified.

triller-telekom:
I thought it would make it consistent, but since we also keep the defaults for the parameters on `@Component` and `@Reference` without explicitly mentioning them, I think it is OK here to also only write what is required.

I have updated the PR.

extensions/binding/org.eclipse.smarthome.binding.astro/src/main/java/org/eclipse/smarthome/binding/astro/internal/discovery/AstroDiscoveryService.java
```diff
@@ -66,10 +67,16 @@ protected void activate(Map<String, Object> configProperties) {
     }
 
     @Override
+    @Modified
     protected void modified(Map<String, Object> configProperties) {
         super.modified(configProperties);
     }
 
+    @Override
+    protected void deactivate() {
```
kaikreuzer: As you aren't doing anything in here anymore, we do not have to keep this method as it does not carry any annotation anymore, right?

sjka: hang on - `super` does! It's just the annotation which is missing here

sjka:
> I thought it would make it consistent

I actually liked it 😞

sjka:
Okay, as I just learned: we are doing it like that everwhere, so nevermind my comment. 
Then make it consistent with the activate, at least

triller-telekom:
I have removed `activate` and `deactivate` because the OSGi framework will find them in the `super` class. Apparently only `modified` needs to be annotated in a child class to be found.

triller-telekom:
The methods are back again and I will also update the documentation in a separate PR. It makes sense to see all DS methods in the implementing class and not have them hidden in a super class.

PR #4971, opened 2018-01-23
[Documentation] Always use @Activate, @Deactivate and @Modified

triller-telekom:
See https://github.com/eclipse/smarthome/pull/4967#issuecomment-359793047

### DIFF
--- a/docs/documentation/development/guidelines.md
+++ b/docs/documentation/development/guidelines.md
@@ -22,7 +22,7 @@ Note that this list also serves as a checklist for code reviews on pull requests
 1. Generics must be used where applicable.
 1. Code should not show any warnings. Warnings that cannot be circumvented should be suppressed by using the @SuppressWarnings annotation. 
 1. For dependency injection, OSGi Declarative Services should be used.
-1. OSGi Declarative Services should be declared using annotations. The IDE will take care of the service *.xml file creation. See the official OSGi documentation for an [example here](http://enroute.osgi.org/services/org.osgi.service.component.html).
+1. OSGi Declarative Services should be declared using annotations. The IDE will take care of the service *.xml file creation. See the official OSGi documentation for an [example here](http://enroute.osgi.org/services/org.osgi.service.component.html). We always use `@Activate`, `@Deactivate` and `@Modified` if we define these methods, even if they exist in a super class, to make the code more readable.
 1. Packages that contain classes that are not meant to be used by other bundles should have "internal" in their package name.
 1. We are using [null annotations](https://wiki.eclipse.org/JDT_Core/Null_Analysis) from the Eclipse JDT project. Therefore every bundle should have an **optional** `Import-Package` dependency to `org.eclipse.jdt.annotation`.
 Classes should be annotated by `@NonNullByDefault` and return types, parameter types, generic types etc. are annotated with `@Nullable` only.


docs/documentation/development/guidelines.md
```diff
@@ -22,7 +22,7 @@ Note that this list also serves as a checklist for code reviews on pull requests
 1. Generics must be used where applicable.
 1. Code should not show any warnings. Warnings that cannot be circumvented should be suppressed by using the @SuppressWarnings annotation. 
 1. For dependency injection, OSGi Declarative Services should be used.
-1. OSGi Declarative Services should be declared using annotations. The IDE will take care of the service *.xml file creation. See the official OSGi documentation for an [example here](http://enroute.osgi.org/services/org.osgi.service.component.html).
+1. OSGi Declarative Services should be declared using annotations. The IDE will take care of the service *.xml file creation. See the official OSGi documentation for an [example here](http://enroute.osgi.org/services/org.osgi.service.component.html). We always use `@Activate`, `@Deactivate` and `@Modified` if we define these methods, even if they exist in a super class, to make the code more readable.
```
maggu2810: IIRC it has not been done because it is more readable but because the DS annotation service does in its default configuration not analyse super classes.
See e.g. http://enroute.osgi.org/faq/ds-inheritance.html

Or do I misinterpret your comment?

htreu: I agree.
Using the annotations is about readability when the methods are in fact implemented on the service (they would be called regardless of the annotations when named according to the DS standard.
But those methods would _not_ be called on super classes. @triller-telekom do you mind overhauling the sentence?

kaikreuzer: The discussion started in #4967, where I requested to remove Activate/Deactivate annotations, because they are optional - and without the need of annotations, there wasn't any need to override the methods from the super class either. Only Modified needs to be specified if the method should be used by DS, so there is the need to override+annotate.
While this is clear to OSGi DS experts, this asymmetry can be pretty confusing for others. Also, having (de)activate-methods hidden in a super class can be surprising.
To resolve the discussion, we simply asked @pkriens for his advice, who happened to stand next to my desk at that moment. His clear advice was to always have those methods present in a component and also annotate all of them and to not rely on inheritance.

sjka: I think there is some misunderstanding here - you'll have to distinguish the annotation processor which processes the annotations during the build and generates the XML files vs. the SCR which actually reads the XMLs and manages the service lifecycle. The former doesn't look at super-classes, while the SCR at runtime doesn't care at all _where_ the methods are defined. E.g. as long as there _is_ a method named `activate` which an appropriate signature, it will be called - regardless where it is defined nor if the XML is missing an `activate` attribute. Therefore IMHO the current comment is perfectly valid.

maggu2810: > His clear advice was to always have those methods present in a component and also annotate all of them and to not rely on inheritance.

I fully agree, it is what I do in all my components.

The question has been if "to make the code more readable" is the correct reason.
As @sjka explained, for `@Activate` and `@Deactivate` it does not matter as long as the function names fit to SRC's default ones. For `@Modified` the annotation is not chosen for make it readable but to use that function at all.

But if we all know why we do it that way, perhaps keep the documentation simple is okay.

kaikreuzer: Ok, got it. I agree that the chosen formulation is not correctly reflecting the reasons, but I think we can live with it.

PR #4975, opened 2018-01-23
GenericItemProvider: Compare GroupItems if an Item was changed

### LINKED ISSUE #4972: group membership not updated on every circumstances after group type change
I run into a problem in openHAB. 

First the group was defined as
Group:Switch:OR(ON, OFF)    gSwitchesAll
and works well.

I decided to changed it to
Group:Number:SUM            gSwitchesAll

and it looks like it is working too.

But the group did not get some updates from item changes and soem rule tigggers are not working. After restart of OH the error was gone.

It looks like the group update is not working correct sometimes.

triller-telekom:
Fixes #4972

### DIFF
--- a/bundles/model/org.eclipse.smarthome.model.item/src/org/eclipse/smarthome/model/item/internal/GenericItemProvider.java
+++ b/bundles/model/org.eclipse.smarthome.model.item/src/org/eclipse/smarthome/model/item/internal/GenericItemProvider.java
@@ -22,7 +22,6 @@ import java.util.List;
 import java.util.Locale;
 import java.util.Map;
 import java.util.Objects;
-import java.util.Set;
 import java.util.concurrent.ConcurrentHashMap;
 
 import org.apache.commons.lang.StringUtils;
@@ -395,7 +394,7 @@ public class GenericItemProvider extends AbstractProvider<Item>
         }
     }
 
-    private boolean hasItemChanged(Item item1, Item item2) {
+    protected boolean hasItemChanged(Item item1, Item item2) {
         return !Objects.equals(item1.getClass(), item2.getClass()) || //
                 !Objects.equals(item1.getName(), item2.getName()) || //
                 !Objects.equals(item1.getCategory(), item2.getCategory()) || //
@@ -403,24 +402,7 @@ public class GenericItemProvider extends AbstractProvider<Item>
                 !Objects.equals(item1.getLabel(), item2.getLabel()) || //
                 !Objects.equals(item1.getTags(), item2.getTags()) || //
                 !Objects.equals(item1.getType(), item2.getType()) || //
-                hasStateDescriptionChanged(item1.getName(), item2.getName()) || //
                 hasGroupItemChanged(item1, item2);
-    }
-
-    private boolean hasStateDescriptionChanged(String itemName1, String itemName2) {
-        StateDescription sd1 = stateDescriptions.get(itemName1);
-        StateDescription sd2 = stateDescriptions.get(itemName2);
-
-        if (sd1 == null && sd2 == null) {
-            return false;
-        }
-
-        else if ((sd1 != null && sd2 == null) || (sd1 == null && sd2 != null)) {
-            // A StateDescription was added or removed
-            return true;
-        }
-
-        return !Objects.equals(sd1.getPattern(), sd2.getPattern());
     }
 
     private boolean hasGroupItemChanged(Item item1, Item item2) {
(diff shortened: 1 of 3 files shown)


bundles/model/org.eclipse.smarthome.model.item/src/org/eclipse/smarthome/model/item/internal/GenericItemProvider.java
```diff
@@ -400,8 +402,76 @@ private boolean hasItemChanged(Item item1, Item item2) {
...
+        if (sd1 == null && sd2 == null) {
+            return false;
+        }
+
+        else if ((sd1 != null && sd2 == null) || (sd1 == null && sd2 != null)) {
+            // A StateDescription was added or removed
+            return true;
+        }
+
+        return !Objects.equals(sd1.getPattern(), sd2.getPattern());
```
sjka: Can't this whole method be simplified to 

```java
return !Objects.equals(sd1 != null ? sd1.getPattern() : null, sd2 != null ? sd2.getPattern() : null);
```

?

bundles/model/org.eclipse.smarthome.model.item/src/org/eclipse/smarthome/model/item/internal/GenericItemProvider.java
```diff
@@ -400,8 +402,76 @@ private boolean hasItemChanged(Item item1, Item item2) {
...
+                sameFunction = Arrays.equals(gf1.getParameters(), gf2.getParameters());
+            }
+        }
+
+        boolean sameMembers = true;
+        Set<Item> members1 = gItem1.getMembers();
+        Set<Item> members2 = gItem2.getMembers();
+        if (members1 != null && members2 != null) {
+            if (members1.size() != members2.size()) {
+                sameMembers = false;
```
sjka: Isn't that case already covered by `Objects.equals(members1, members2)`?

bundles/model/org.eclipse.smarthome.model.item/src/org/eclipse/smarthome/model/item/internal/GenericItemProvider.java
```diff
@@ -400,8 +402,76 @@ private boolean hasItemChanged(Item item1, Item item2) {
...
+        if (item2 instanceof GroupItem) {
+            gItem2 = (GroupItem) item2;
+        }
+
+        if (gItem1 == null && gItem2 == null) {
+            return false;
+        }
+
+        if ((gItem1 != null && gItem2 == null) || (gItem1 == null && gItem2 != null)) {
+            // Item was changed into a GroupItem
```
sjka: superfluous comment - and wrong

bundles/model/org.eclipse.smarthome.model.item/src/org/eclipse/smarthome/model/item/internal/GenericItemProvider.java
```diff
@@ -400,8 +402,76 @@ private boolean hasItemChanged(Item item1, Item item2) {
...
+            return true;
+        }
+
+        boolean sameBaseItemClass = false;
```
sjka: What if both don't have a base item? then `sameBaseItemClass ` _should_ be true!

If I'm not mistaken, then the below if-block could be replaced completely by 

```java
Objects.equals(gItem1, gItem2);
```

It will use the items' `equals()` methods which implicitly also to a type comparison. And will cover the both-`null`-case also.

bundles/model/org.eclipse.smarthome.model.item/src/org/eclipse/smarthome/model/item/internal/GenericItemProvider.java
```diff
@@ -400,8 +402,76 @@ private boolean hasItemChanged(Item item1, Item item2) {
...
+        if (gItem1.getBaseItem() != null && gItem2.getBaseItem() != null) {
+            sameBaseItemClass = Objects.equals(gItem1.getBaseItem().getClass(), gItem2.getBaseItem().getClass());
+        }
+
+        boolean sameFunction = false;
+        GroupFunction gf1 = gItem1.getFunction();
+        GroupFunction gf2 = gItem2.getFunction();
+        if (gf1 != null && gf2 != null) {
```
sjka: see above

bundles/model/org.eclipse.smarthome.model.item/src/org/eclipse/smarthome/model/item/internal/GenericItemProvider.java
```diff
@@ -400,8 +402,76 @@ private boolean hasItemChanged(Item item1, Item item2) {
...
+            if (Objects.equals(gf1.getClass(), gf2.getClass())) {
+                sameFunction = Arrays.equals(gf1.getParameters(), gf2.getParameters());
+            }
+        }
+
+        boolean sameMembers = true;
+        Set<Item> members1 = gItem1.getMembers();
+        Set<Item> members2 = gItem2.getMembers();
+        if (members1 != null && members2 != null) {
```
sjka: This whole block also could be replaced completely by 

```java
Objects.equals(gItem1.getMembers(), gItem2.getMembers());
```

bundles/model/org.eclipse.smarthome.model.item/src/org/eclipse/smarthome/model/item/internal/GenericItemProvider.java
triller-telekom: I have added an `else` branch to cover the "double null" case.

sjka:
Great improvement, thanks!

PR #4977, opened 2018-01-24
ignore point2point interfaces for local interface search

sobimibos:
PointToPoint interfaces are used to connect to the internet or to some other remote network. The ip address of these interface is used to connect from the remote network to this system. So this is not a local address and should not be in the list of local ip addresses, because

    it cannot be used to connect to other local devices and systems
    in case of an internet connection this is the public address. Opening a server socket on the public address is a serious security risk.


sjka:
Looks plausible to me, thanks

PR #4986, opened 2018-01-26
Bugfix/f wupdate tests

YordanDZhelev:
FirmwareUpdateServiceTest.testEvents() test fails once every several executions due to not delivered events.


YordanDZhelev:
Will be rebased on master.

PR #4987, opened 2018-01-26
FirmwareUpdateServiceTest.testEvents() test fails once every several …

YordanDZhelev:
FirmwareUpdateServiceTest.testEvents() test fails once every several executions due to not delivered events.

### DIFF
--- a/bundles/core/org.eclipse.smarthome.core.thing.test/src/test/java/org/eclipse/smarthome/core/thing/firmware/FirmwareUpdateServiceTest.java
+++ b/bundles/core/org.eclipse.smarthome.core.thing.test/src/test/java/org/eclipse/smarthome/core/thing/firmware/FirmwareUpdateServiceTest.java
@@ -321,7 +321,7 @@ public class FirmwareUpdateServiceTest extends JavaOSGiTest {
             verify(handler3, times(1)).cancel();
         });
     }
-    
+
     @Test
     public void testCancelFirmwareUpdateIntheMiddleOfUpdate() {
         final long stepsTime = 10;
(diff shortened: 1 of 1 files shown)


bundles/core/org.eclipse.smarthome.core.thing.test/src/test/java/org/eclipse/smarthome/core/thing/firmware/FirmwareUpdateServiceTest.java
```diff
@@ -647,7 +647,9 @@ public void testEvents() {
         events.get().addAll(eventCaptor.getAllValues());
         List<Event> list = events.get().stream().filter(event -> event instanceof FirmwareUpdateProgressInfoEvent)
                 .collect(Collectors.toList());
-        assertTrue(list.size() >= SEQUENCE.length);
+        waitForAssert(() -> {
```
htreu: I doubt that this will fix the test. Here you are waiting in case the event list does not have the correct size. But there is no background thread which does modify the list during waiting.
My assumption is that lines 644 - 649 could be the problem: the waitForAssert in line 644 verifies atLeast SEQUENCE.length times, and later (line 648) some of the captured events might be filtered out. If this is the case then the list will be shorter than SEQUENCE.length. wdyt?

YordanDZhelev: My assumption is that at line 639 firmwareUpdateService.getFirmwareStatusInfo() calls internally eventPublisher.post() which posts FirmwareStatusInfoEvent. At line 644 we are waiting for four events, but we don't filter their types. This way the first four events are one FirmwareStatusInfoEvent and three FirmwareUpdateProgressInfoEvent.

htreu: yes, this seems to be correct. Verifying on `SEQUENCE.length + 1` invocations of `mockPublisher#post` might do it then.

YordanDZhelev: PR updated regarding htreu's last comment.

bundles/core/org.eclipse.smarthome.core.thing.test/src/test/java/org/eclipse/smarthome/core/thing/firmware/FirmwareUpdateServiceTest.java
```diff
@@ -642,7 +642,7 @@ public void testEvents() {
         AtomicReference<List<Event>> events = new AtomicReference<>(new ArrayList<>());
         ArgumentCaptor<Event> eventCaptor = ArgumentCaptor.forClass(Event.class);
         waitForAssert(() -> {
-            verify(mockPublisher, atLeast(SEQUENCE.length)).post(eventCaptor.capture());
+            verify(mockPublisher, atLeast(SEQUENCE.length + 1)).post(eventCaptor.capture());
```
htreu: please leave a comment in the code why there should be 5 invocations. Maybe also on line 639 because its not obvious that there is an event generated internally.

YordanDZhelev: Comments added

PR #4990, opened 2018-01-26
Sat plugin new version

svilenvul:


maggu2810:
Would it make sense to fix all errors reported by the plugin and wait for a successful build of ESH in front of the CQ creation?

```text
[ERROR] /home/travis/build/eclipse/smarthome/bundles/test/org.eclipse.smarthome.magic/META-INF/MANIFEST.MF:[0]
The package org.eclipse.smarthome.magic.service should be marked as "internal" if it is not exported.
```

As long as ESH does not build completely successful we don't know, if we could detect an error in the plugin. If there is an error in the plugin, it needs to be fixed and another CQ has to be requested.
To reduce the work of the Eclipse IP team, I would prefer to fix all ESH violations the SAT found before.

WDYT?

svilenvul:
> Would it make sense to fix all errors reported by the plugin and wait for a successful build of ESH in front of the CQ creation?

Yes, of course. I opened the PR just to make it visible.

kaikreuzer:
@maggu2810 The SAT tool is an approved build-time only dependency, we do not require a new CQs when updating its version.

svilenvul:
Hmm, strange , the last build fails with:
```
[ERROR] Failed to execute goal org.apache.karaf.tooling:karaf-maven-plugin:4.0.3:verify (verify-esh) on project verify: Verification failures -> [Help 1]
```

Do you have any clue if it is related ?

kaikreuzer:
I doubt that it can be related - just retriggered Travis to check...

svilenvul:
From the logs, I see that some ESH Karaf features depend on `mvn:org.openhab/nrjavaserial/3.12.0.OH`

This seems to be in JCenter - https://jcenter.bintray.com/org/openhab/nrjavaserial/, so I guess this is the cause of the failure. 

How should we proceed in this case ? Should I bring back the JCenter repository ?

kaikreuzer:
Yes, you are right! I forgot that this was in JCenter as well.
So yes, please add the repo again.

PR #4535, opened 2017-11-12
[WIP] Bluetooth Binding

kaikreuzer:
This PR is a follow-up replacement of #3633. It includes heavy refactoring especially wrt discovery and lifecylcle. The namespace has also been changed from BLE to Bluetooth.

# Overview

This provides an initial PR for a Bluetooth binding, with a BlueGiga USB adapter and a YeelightBlue device support implemented. It is still WIP.

## Work to be completed

There is further testing, tidying and improving required.

Better connection management is needed to cope with disconnects. There is also some arbitration required in the BlueGiga handler to handle concurrency clashes between different thing handlers.

# Bluetooth Binding overview
 
The Bluetooth binding is implemented to allow bundles to extend the main Bluetooth bundle in order to add new Bluetooth adapter as well as device support. This architecture means that such extension bundles must utilise the binding name ```bluetooth```.

A base class structure is defined in the org.eclipse.smarthome.binding.bluetooth bundle. This includes the main classes required to implement Bluetooth -:
 
* BluetoothAdapter. This interface defines the main functionality required to be implemented by a Bluetooth adapter, including device discovery. Typically, this interface is implemented by a BridgeHandler and then registered as an OSGi service
* BluetoothDiscoveryParticipant. An interface to be implemented by services that can identify specific Bluetooth devices.
* BluetoothDevice. This implements a Bluetooth device. It manages the notifications of device notifications, Bluetooth service and characteristic management, and provides the main interface to communicate to a Bluetooth device.
* BluetoothService. Implements the Bluetooth service. A service holds a number of characteristics.
* BluetoothCharacteristic. Implements the Bluetooth characteristic. This is the basic component for communicating data to and from a Bluetooth device
* BluetoothDescriptor. Implements the Bluetooth descriptors for each characteristic.
 
## Implementing a new Bluetooth Adapter bundle
 
Bluetooth adapters are modelled as a Bridge in ESH. The bridge handler provides the link with the Bluetooth hardware (eg a dongle, or system Bluetooth API). An adapter bundle needs to implement two main classes: the BridgeHandler which should implement BluetoothAdapter (any be registered as a service), and a ThingFactory, which is required to instantiate the handler.
 
[...]

PR #4997, opened 2018-01-27
Initial contribution of the generic Bluetooth binding with BlueZ support

kaikreuzer:
This PR contains the first parts from #4535 - the generic Bluetooth binding, which defines the central service and API and comes already with a simple and generic device support.
Additionally, this PR includes the bundle for BlueZ support on ARMv6hf, so that the code can be used directly on a Raspberry Pi or similar.

This PR is ready for review. Other parts of #4535 still need polishing and can follow in separate PRs later on.

Besides a CQ for the sources in this PR itself, we will also require a CQ for the TinyB library. An earlier version is already in use at Eclipse Kura, so this should hopefully be no big effort.

Also-By: Chris Jackson <chris@cd-jackson.com>


kaikreuzer:
@vkolotov Please also feel cordially invited to review the code - as you are an expert on that matter with much more practical experience, your input is very welcome. I would be especially interested whether you think that we are missing any crucial features in the API (mainly the `BluetoothDevice`and `BluetoothAdapter` classes) that might be needed for common situations.

vkolotov:
Hi @kaikreuzer, from the first glance it is looking very good. Nice and tidy. I'll definitely have a look once I've cut a new release of the BT binding. 

BTW, why would you want to have two BT bindings? :) Just curious.

extensions/binding/org.eclipse.smarthome.binding.bluetooth/src/main/java/org/eclipse/smarthome/binding/bluetooth/BluetoothCharacteristic.java
```diff
@@ -0,0 +1,633 @@
...
+            case FORMAT_SINT8:
+                return unsignedToSigned(unsignedByteToInt(value[offset]), 8);
+
+            case FORMAT_SINT16:
+                return unsignedToSigned(unsignedBytesToInt(value[offset], value[offset + 1]), 16);
+
+            case FORMAT_SINT32:
+                return unsignedToSigned(
+                        unsignedBytesToInt(value[offset], value[offset + 1], value[offset + 2], value[offset + 3]), 32);
+        }
```
sjka: woudln't it make sense to have a `default` case as well, to let it fail loudly in case somebody extends it?

extensions/binding/org.eclipse.smarthome.binding.bluetooth/src/main/java/org/eclipse/smarthome/binding/bluetooth/BluetoothCharacteristic.java
```diff
@@ -0,0 +1,633 @@
...
+        if ((offset + getTypeLen(formatType)) > value.length) {
+            return null;
+        }
+
+        switch (formatType) {
+            case FORMAT_SFLOAT:
+                return bytesToFloat(value[offset], value[offset + 1]);
+
+            case FORMAT_FLOAT:
+                return bytesToFloat(value[offset], value[offset + 1], value[offset + 2], value[offset + 3]);
```
sjka: see above wrt to a `default`

extensions/binding/org.eclipse.smarthome.binding.bluetooth/src/main/java/org/eclipse/smarthome/binding/bluetooth/BluetoothCharacteristic.java
```diff
@@ -0,0 +1,633 @@
...
+    /**
+     * Return the stored value of this characteristic. This doesn't read the remote data.
+     *
+     */
+    public String getStringValue(int offset) {
+        if (value == null || offset > value.length) {
+            return null;
+        }
+        byte[] strBytes = new byte[value.length - offset];
+        for (int i = 0; i != (value.length - offset); ++i) {
```
sjka: `i < ...` is the widely accepted way. Should never ever happen to exceed the maximum, but you never know...

extensions/binding/org.eclipse.smarthome.binding.bluetooth/src/main/java/org/eclipse/smarthome/binding/bluetooth/BluetoothCharacteristic.java
```diff
@@ -0,0 +1,633 @@
...
+            strBytes[i] = (byte) value[offset + i];
+        }
+        return new String(strBytes);
```
sjka: `new String(byte[])` is the "I'm feeling lucky"-version wrt the character encoding. Rather give it _some_  fixed charset like UTF-8 instead of relying on the platform default.

extensions/binding/org.eclipse.smarthome.binding.bluetooth/src/main/java/org/eclipse/smarthome/binding/bluetooth/BluetoothCharacteristic.java
```diff
@@ -0,0 +1,633 @@
...
+    }
+
+    /**
+     * Updates the locally stored value of this characteristic.
+     *
+     */
+    public boolean setValue(int[] value) {
+        this.value = value;
+        return true;
```
sjka: what does this return value mean?

extensions/binding/org.eclipse.smarthome.binding.bluetooth/src/main/java/org/eclipse/smarthome/binding/bluetooth/BluetoothCharacteristic.java
```diff
@@ -0,0 +1,633 @@
...
+                this.value[offset] = (byte) (val & 0xFF);
+                this.value[offset + 1] = (byte) ((val >> 8) & 0xFF);
+                this.value[offset + 2] = (byte) ((val >> 16) & 0xFF);
+                this.value[offset + 2] = (byte) ((val >> 24) & 0xFF);
+                break;
+
+            default:
+                return false;
+        }
+        return true;
```
sjka: what is the meaning of this return value?

extensions/binding/org.eclipse.smarthome.binding.bluetooth/src/main/java/org/eclipse/smarthome/binding/bluetooth/ConnectedBluetoothHandler.java
```diff
@@ -0,0 +1,240 @@
...
+                // the channel does not exist yet, so let's add it
+                ThingBuilder updatedThing = editThing();
+                Channel channel = ChannelBuilder.create(new ChannelUID(getThing().getUID(), channelId), "Number")
+                        .withType(channelTypeUID).build();
+                updatedThing.withChannel(channel);
+                updateThing(updatedThing.build());
+                logger.debug("Added channel '{}' to Thing '{}'", channelId, getThing().getUID());
+            }
+            if (isLinked(channelId)) {
+                deviceCharacteristics.add(characteristic);
```
sjka: isn't the same necessary if a channel gets linked later on?

extensions/binding/org.eclipse.smarthome.binding.bluetooth/src/main/java/org/eclipse/smarthome/binding/bluetooth/discovery/internal/BluetoothDiscoveryService.java
```diff
@@ -0,0 +1,178 @@
...
+    private final Set<BluetoothAdapter> adapters = new CopyOnWriteArraySet<>();
+    private final Set<BluetoothDiscoveryParticipant> participants = new CopyOnWriteArraySet<>();
+    private final Map<UID, BluetoothDiscoveryListener> registeredListeners = new HashMap<>();
+
+    public BluetoothDiscoveryService() {
+        super(SEARCH_TIME);
+    }
+
+    @Override
+    protected void activate(Map<String, Object> configProperties) {
```
sjka: `@Activate`

extensions/binding/org.eclipse.smarthome.binding.bluetooth/src/main/java/org/eclipse/smarthome/binding/bluetooth/discovery/internal/BluetoothDiscoveryService.java
```diff
@@ -0,0 +1,178 @@
...
+    }
+
+    @Override
+    @Modified
+    protected void modified(Map<String, Object> configProperties) {
+        super.modified(configProperties);
+    }
+
+    @Override
+    public void deactivate() {
```
sjka: `@Deactivate`

extensions/binding/org.eclipse.smarthome.binding.bluetooth/src/main/java/org/eclipse/smarthome/binding/bluetooth/discovery/internal/BluetoothDiscoveryService.java
```diff
@@ -0,0 +1,178 @@
...
+@Component(immediate = true, service = DiscoveryService.class, configurationPid = "discovery.bluetooth")
+public class BluetoothDiscoveryService extends AbstractDiscoveryService {
+
+    private final Logger logger = LoggerFactory.getLogger(BluetoothDiscoveryService.class);
+
+    private final static int SEARCH_TIME = 15;
+
+    private final Set<BluetoothAdapter> adapters = new CopyOnWriteArraySet<>();
+    private final Set<BluetoothDiscoveryParticipant> participants = new CopyOnWriteArraySet<>();
+    private final Map<UID, BluetoothDiscoveryListener> registeredListeners = new HashMap<>();
```
sjka: Shouldn't this be a `ConcurrentHashMap`? Alternatively, the add... and remove... methods must be `synchronized`

extensions/binding/org.eclipse.smarthome.binding.bluetooth/src/main/java/org/eclipse/smarthome/binding/bluetooth/discovery/internal/BluetoothDiscoveryService.java
```diff
@@ -0,0 +1,178 @@
...
+        this.participants.add(participant);
+    }
+
+    protected void removeBluetoothDiscoveryParticipant(BluetoothDiscoveryParticipant participant) {
+        this.participants.remove(participant);
+    }
+
+    @Override
+    public Set<ThingTypeUID> getSupportedThingTypes() {
+        Set<ThingTypeUID> supportedThingTypes = new HashSet<>();
```
sjka: I _think_  it would make sense to construct this set whenever a BDP is added/removed and keep it ina a field, because IIRC this method gets called quite frequently.

extensions/binding/org.eclipse.smarthome.binding.bluetooth/src/main/java/org/eclipse/smarthome/binding/bluetooth/internal/BluetoothHandlerFactory.java
```diff
@@ -0,0 +1,60 @@
...
+import org.eclipse.smarthome.core.thing.binding.ThingHandlerFactory;
+import org.osgi.service.component.annotations.Component;
+import org.osgi.service.component.annotations.ConfigurationPolicy;
+
+/**
+ * The {@link BluetoothHandlerFactory} is responsible for creating things and thing handlers.
+ *
+ * @author Kai Kreuzer - Initial contribution and API
+ */
+@Component(service = ThingHandlerFactory.class, immediate = true, configurationPid = "binding.bluetooth", configurationPolicy = ConfigurationPolicy.OPTIONAL)
```
sjka: how is this configurable?

extensions/binding/org.eclipse.smarthome.binding.bluetooth/src/main/java/org/eclipse/smarthome/binding/bluetooth/notification/BluetoothScanNotification.java
```diff
@@ -0,0 +1,144 @@
...
+    public int getRssi() {
+        return rssi;
+    }
+
+    /**
+     * Sets the scan packet data
+     *
+     * @param data an int array containing the raw packet data;
+     */
+    public void setData(int[] data) {
```
sjka: do we really want to have arrays in our API?

sjka:
I created CQ 15487 (PB CQ13796) for the TinyB library

vkolotov:
Just a heads up, I've added a PR for @cdjackson to add support for service data broadcast messages: https://github.com/zsmartsystems/com.zsmartsystems.bluetooth.bluegiga/pull/6
And also slightly tweaked manufacturer data broadcast messages, this might affect you @kaikreuzer.

extensions/binding/org.eclipse.smarthome.binding.bluetooth.bluez/src/main/java/org/eclipse/smarthome/binding/bluetooth/bluez/handler/BlueZBridgeHandler.java
```diff
@@ -0,0 +1,267 @@
...
+    }
+
+    @Override
+    public BluetoothAddress getAddress() {
+        return address;
+    }
+
+    @Override
+    public BluetoothDevice getDevice(BluetoothAddress address) {
+        if (devices.containsKey(address.toString())) {
```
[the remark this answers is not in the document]
kaikreuzer: No, since reads should never be blocked and writes need to check whether a previous write already created the same device.

extensions/binding/org.eclipse.smarthome.binding.bluetooth.bluez/src/main/java/org/eclipse/smarthome/binding/bluetooth/bluez/internal/BlueZHandlerFactory.java
```diff
@@ -0,0 +1,84 @@
...
+import org.osgi.service.component.annotations.Component;
+import org.osgi.service.component.annotations.ConfigurationPolicy;
+
+/**
+ * The {@link BlueZHandlerFactory} is responsible for creating things and thing
+ * handlers.
+ *
+ * @author Kai Kreuzer - Initial contribution and API
+ */
+@Component(service = ThingHandlerFactory.class, immediate = true, configurationPid = "binding.bluetooth.bluez", configurationPolicy = ConfigurationPolicy.OPTIONAL)
```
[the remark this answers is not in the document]
kaikreuzer: This question would apply [to all bindings](https://github.com/eclipse/smarthome/blob/master/tools/archetype/binding/src/main/resources/archetype-resources/src/main/java/internal/__bindingIdCamelCase__HandlerFactory.java#L43), so I'll simply ignore it here ;-)

extensions/binding/org.eclipse.smarthome.binding.bluetooth.bluez/src/main/java/org/eclipse/smarthome/binding/bluetooth/bluez/internal/BlueZHandlerFactory.java
```diff
@@ -0,0 +1,84 @@
...
+    private synchronized void registerBluetoothAdapter(BluetoothAdapter adapter) {
+        this.serviceRegs.put(adapter.getUID(), bundleContext.registerService(BluetoothAdapter.class.getName(), adapter,
+                new Hashtable<String, Object>()));
+    }
+
+    @Override
+    protected synchronized void removeHandler(ThingHandler thingHandler) {
+        if (thingHandler instanceof BluetoothAdapter) {
+            UID uid = ((BluetoothAdapter) thingHandler).getUID();
+            ServiceRegistration<?> serviceReg = this.serviceRegs.get(uid);
```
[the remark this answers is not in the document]
kaikreuzer: Imho, I can only safe the last line (`serviceRegs.remove(uid);`) but that's already something.

extensions/binding/org.eclipse.smarthome.binding.bluetooth/ARCHITECTURE.md
```diff
@@ -0,0 +1,54 @@
+# Bluetooth Binding overview
+ 
+The Bluetooth binding is implemented to allow bundles to extend the main Bluetooth bundle in order to add new Bluetooth adapter as well as device support. This architecture means that such extension bundles must utilise the binding name ```bluetooth```.
```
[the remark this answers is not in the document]
kaikreuzer: No, this binding also serves as the glue between bridges and things and also handles the discovery centrally. A transport bundle would be "too low" for those purposes. We had that approach before and it didn't really work, see [these long discussions last year](https://github.com/eclipse/smarthome/pull/3531#issuecomment-305831722).

extensions/binding/org.eclipse.smarthome.binding.bluetooth/ESH-INF/binding/binding.xml
```diff
@@ -0,0 +1,10 @@
+<?xml version="1.0" encoding="UTF-8"?>
+<binding:binding id="bluetooth"
+				 xmlns:xsi="http://www.w3.org/2001/XMLSchema-instance"
+				 xmlns:binding="http://eclipse.org/smarthome/schemas/binding/v1.0.0"
+				 xsi:schemaLocation="http://eclipse.org/smarthome/schemas/binding/v1.0.0 http://eclipse.org/smarthome/schemas/binding-1.0.0.xsd">
+
+	<name>Bluetooth Binding</name>
+	<description>This binding supports the Bluetooth protocol.</description>
+
```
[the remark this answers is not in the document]
kaikreuzer: Too many (at least on the long run), thus leaving it out as it is optional.

extensions/binding/org.eclipse.smarthome.binding.bluetooth/ESH-INF/thing/thing-types.xml
```diff
@@ -0,0 +1,46 @@
+<?xml version="1.0" encoding="UTF-8"?>
+<thing:thing-descriptions bindingId="bluetooth" xmlns:xsi="http://www.w3.org/2001/XMLSchema-instance"
+	xmlns:thing="http://eclipse.org/smarthome/schemas/thing-description/v1.0.0"
+	xsi:schemaLocation="http://eclipse.org/smarthome/schemas/thing-description/v1.0.0 http://eclipse.org/smarthome/schemas/thing-description-1.0.0.xsd">
+
+	<thing-type id="beacon">
+		<supported-bridge-type-refs>
+			<bridge-type-ref id="bluegiga" />
```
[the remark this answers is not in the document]
kaikreuzer: Fully agree, at the moment this is rather a workaround. I already had many discussions about that in the past, but always missed to enter an according issue for it. Now here it is: https://github.com/eclipse/smarthome/issues/5010

extensions/binding/org.eclipse.smarthome.binding.bluetooth/src/main/java/org/eclipse/smarthome/binding/bluetooth/notification/BluetoothScanNotification.java
kaikreuzer: Well, those is the raw data of telegrams - so I would say that arrays are ok to use for this.
It is also [what the Android BLE API is doing](https://developer.android.com/reference/android/bluetooth/BluetoothAdapter.LeScanCallback.html#onLeScan(android.bluetooth.BluetoothDevice, int, byte[])).

The only thing I would claim should be changed is the int to byte - as the Bluetooth spec says that this content is all uInt8 - so just as Android does it, the best Java type seems to be byte (or was there any specific reason why you chose int, @cdjackson?).

kaikreuzer:
> BTW, why would you want to have two BT bindings? :) Just curious.

@vkolotov I assume this is just a rhetorical question, right? As discussed in the past, this one is a "pure" ESH approach with full IP treatment, which is also going to be used in commercial setups of ESH and not primarily targeted at openHAB. But yes, I would also still prefer to join any efforts, so please feel invited to contribute to it.

>  I've added a PR for @cdjackson's bluegiga library

Thanks for the heads-up, but for this initial PR here, it isn't relevant (you will notice that it does not contain the BlueGiga bundle at all ;-)). Will be important for subsequent additions, though.

kaikreuzer:
@sjka I have addressed all your comments and updated the PR accordingly.

vkolotov:
hi @kaikreuzer, I've just noticed that the binding id is "bluetooth". How we are going to resolve conflicts? I suspect that once this binding goes live and users update ESH, they will have some issues with their existing things.

sjka:
> I have addressed all your comments and updated the PR accordingly.

Thanks for that. Looks good to me now. From my side you can go ahead and squash the commits so we can file a CQ for it.

kaikreuzer:
@vkolotov I don't think that there are any conflicts as users should only have a single bluetooth binding installed.

@sjka Thanks - commits are squashed.

kaikreuzer:
FTR: 
- CQ 15487 for TinyB has been approved.
- CQ 15531 has been created for the code of this PR (as it exceeds 1000 loc), so we will wait for its approval.

extensions/binding/org.eclipse.smarthome.binding.bluetooth.bluez/src/main/java/org/eclipse/smarthome/binding/bluetooth/bluez/BlueZAdapterConstants.java
```diff
@@ -0,0 +1,35 @@
...
+import org.eclipse.smarthome.core.thing.ThingTypeUID;
+
+/**
+ * The {@link BlueZAdapterConstants} class defines common constants, which are
+ * used across the whole binding.
+ *
+ * @author Kai Kreuzer - Initial contribution and API
+ */
+@NonNullByDefault
+public class BlueZAdapterConstants {
```
amitjoy: A private constructor would be really useful. Inside the constructor we could throw `IllegalAccessError` as well, otherwise through reflection the instance can get created

extensions/binding/org.eclipse.smarthome.binding.bluetooth.bluez/src/main/java/org/eclipse/smarthome/binding/bluetooth/bluez/BlueZBluetoothDevice.java
```diff
@@ -0,0 +1,378 @@
...
+                    service.addCharacteristic(characteristic);
+                }
+                addService(service);
+            }
+            notifyListeners(BluetoothEventType.SERVICES_DISCOVERED);
+        }
+    }
+
+    @Override
+    public boolean connect() {
```
amitjoy: I believe it could be made thread-safe.

extensions/binding/org.eclipse.smarthome.binding.bluetooth.bluez/src/main/java/org/eclipse/smarthome/binding/bluetooth/bluez/BlueZBluetoothDevice.java
```diff
@@ -0,0 +1,378 @@
...
+                    logger.debug("Exception occurred when trying to connect device '{}': {}", device.getAddress(),
+                            e.getMessage());
+                }
+            }
+        }
+        return false;
+    }
+
+    @Override
+    public boolean disconnect() {
```
amitjoy: This could also be made thread-safe.

extensions/binding/org.eclipse.smarthome.binding.bluetooth.bluez/src/main/java/org/eclipse/smarthome/binding/bluetooth/bluez/handler/BlueZBridgeHandler.java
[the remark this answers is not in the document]
amitjoy: According to ReadWriteLock, the read lock may be held simultaneously by multiple reader threads, so long as there are no writers. The write lock is exclusive. It could be useful.

extensions/binding/org.eclipse.smarthome.binding.bluetooth/src/main/java/org/eclipse/smarthome/binding/bluetooth/BluetoothBindingConstants.java
```diff
@@ -0,0 +1,56 @@
...
+import org.eclipse.smarthome.core.thing.ThingTypeUID;
+
+/**
+ * The {@link BluetoothBindingConstants} class defines common constants, which are
+ * used across the whole binding.
+ *
+ * @author Chris Jackson - Initial contribution
+ * @author Kai Kreuzer - refactoring and extension
+ */
+public class BluetoothBindingConstants {
```
amitjoy: Private Constructor would make it non-instantiable.

extensions/binding/org.eclipse.smarthome.binding.bluetooth/src/main/java/org/eclipse/smarthome/binding/bluetooth/BluetoothCharacteristic.java
```diff
@@ -0,0 +1,658 @@
...
+ * <p>
+ * Characteristics are defined attribute types that contain a single logical value.
+ * <p>
+ * https://www.bluetooth.com/specifications/gatt/characteristics
+ *
+ * @author Chris Jackson - Initial contribution
+ * @author Kai Kreuzer - Cleaned up code
+ */
+public class BluetoothCharacteristic {
+    public static final int FORMAT_UINT8 = 0x11;
```
amitjoy: We could also extract these as cohesive types in Enumeration.

extensions/binding/org.eclipse.smarthome.binding.bluetooth/src/main/java/org/eclipse/smarthome/binding/bluetooth/discovery/BluetoothDiscoveryParticipant.java
```diff
@@ -0,0 +1,61 @@
...
+     * Defines the list of thing types that this participant can identify
+     *
+     * @return a set of thing type UIDs for which results can be created
+     */
+    public Set<ThingTypeUID> getSupportedThingTypeUIDs();
+
+    /**
+     * Creates a discovery result for a Bluetooth device
+     *
+     * @param device the Bluetooth device found on the network
```
amitjoy: @NotNull on device argument or invariant would make the API more comprehensive

extensions/binding/org.eclipse.smarthome.binding.bluetooth/src/main/java/org/eclipse/smarthome/binding/bluetooth/discovery/BluetoothDiscoveryParticipant.java
```diff
@@ -0,0 +1,61 @@
...
+     *
+     * @return the according discovery result or <code>null</code>, if device is not
+     *         supported by this participant
+     */
+    @Nullable
+    public DiscoveryResult createResult(BluetoothDevice device);
```
amitjoy: Similarly here

extensions/binding/org.eclipse.smarthome.binding.bluetooth/src/main/java/org/eclipse/smarthome/binding/bluetooth/notification/BluetoothScanNotification.java
```diff
@@ -0,0 +1,144 @@
...
+public class BluetoothScanNotification extends BluetoothNotification {
+    /**
+     * The receive signal strength for this beacon packet
+     */
+    private int rssi = Integer.MIN_VALUE;
+
+    /**
+     * The raw data
+     */
+    private byte[] data = null;
```
amitjoy: null would be set by default

extensions/binding/org.eclipse.smarthome.binding.bluetooth/src/main/java/org/eclipse/smarthome/binding/bluetooth/notification/BluetoothScanNotification.java
```diff
@@ -0,0 +1,144 @@
...
+
+    /**
+     * The manufacturer specific data
+     */
+    private byte[] manufacturerData = null;
```
amitjoy: Similarly here

extensions/binding/org.eclipse.smarthome.binding.bluetooth/src/main/java/org/eclipse/smarthome/binding/bluetooth/notification/BluetoothScanNotification.java
```diff
@@ -0,0 +1,144 @@
...
+
+    /**
+     * The beacon type
+     */
+    private BluetoothBeaconType beaconType = BluetoothBeaconType.BEACON_UNKNOWN;
+
+    /**
+     * The device name
+     */
+    private String name = new String();
```
amitjoy: Explicit String instance creation and hence the object instance cannot be interned. Assigning to empty value should suffice.

kaikreuzer:
Hey @amitjoy, thanks for your comments! Please note that (as mentioned in https://github.com/eclipse/smarthome/pull/4997#issuecomment-362285835) I cannot do any changes to this PR anymore as a CQ is already in process.
But I'll make sure to create a follow-up PR with changes that address your feedback afterwards.

amitjoy:
@kaikreuzer Didn't notice at all. Sorry for the noise. I will try to review the following PR. Thanks for the information.

amitjoy:
:)

kaikreuzer:
Not considered noise, please continue - it is still the best place to add comments.
Alternatively, you could of course also prepare a PR on top of this one yourself, in case that is easier than writing many comments and having to explain to me what you mean ;-)

amitjoy:
Really a good idea. will definitely do.

sjka:
CQ 15531 got approved. 
Hence I'll merge this one. @amitjoy please propose your mentioned improvements as part of another PR - it gets even easier then.

PR #5007, opened 2018-01-30
[core] Added 'isLinked' method with parameter 'channelUID'

cweitkamp:
- Added `isLinked` method with parameter `channelUID`

### DIFF
--- a/bundles/core/org.eclipse.smarthome.core.thing/src/main/java/org/eclipse/smarthome/core/thing/binding/BaseThingHandler.java
+++ b/bundles/core/org.eclipse.smarthome.core.thing/src/main/java/org/eclipse/smarthome/core/thing/binding/BaseThingHandler.java
@@ -665,7 +665,7 @@ public abstract class BaseThingHandler implements ThingHandler {
      * @param channelId channel ID (must not be null)
      * @return true if at least one item is linked, false otherwise
      */
-    public boolean isLinked(String channelId) {
+    protected boolean isLinked(String channelId) {
         Channel channel = thing.getChannel(channelId);
         if (channel != null) {
             return isLinked(channel.getUID());
@@ -682,7 +682,7 @@ public abstract class BaseThingHandler implements ThingHandler {
      * @param channelUID UID of the channel (must not be null)
      * @return true if at least one item is linked, false otherwise
      */
-    public boolean isLinked(ChannelUID channelUID) {
+    protected boolean isLinked(ChannelUID channelUID) {
         return linkRegistry != null ? !linkRegistry.getLinks(channelUID).isEmpty() : false;
     }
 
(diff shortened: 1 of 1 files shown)


kaikreuzer:
Could you comment on why you want to extend the API that way?

Note that there is still an architectural issue with `isLinked` as it requires the `linkRegistry` service internally, which we would rather want to get rid off - so extending it might rather complicate the path to a different solution.

sjka:
How about creating those two `isLinked`  methods in the `ThingHandlerCallback`? 

Its implementation is an inner class of the `ThingManager`, which has the required service references anyway. You could then drop the corresponding `ServiceTracker`-Hack from  `BaseThingHandler` and let the existing `isLinked` method delegate to the `ThingHandlerCallback` and mark it as `@Deprecated` because I don't think it's viable to drop it instantly.

(As a bonus we could leverage the same mechanism for the other `ServiceTrackers` too, but I don't want you to blow up this PR with it.)

cweitkamp:
Sry for my late response. I should've put my thoughts about this into the initial description but was in a hurry on that day as well.

@kaikreuzer To answer your questions: I had two ideas in my mind. First one was to add more flexibility to the method. We have other methods which can be used with `channelUID` parameter and `channelId` parameter in the same way (e.g. `updateState`, `triggerChannel` and `postCommand`). The second thought was to call this method from outside the handler. But the more I think about that the more I'm convinced my current use case for it has a bad design. I'm fine with reverting the exposure from `protected` to `public`.

I was not aware of the architectural issues. I want to add another question about this method. What was the intention to place it inside the `BaseThingHandler`? Isn't it more or less a util / helper method without any relation to the thing handler itself?

@sjka I have to look into that in detail but I think I can follow you.

kaikreuzer:
> I was not aware of the architectural issues.

We would like to get rid off the "bundleContext" within the handlers as this invites people to do all sorts of nasty stuff... See also https://github.com/eclipse/smarthome/pull/5161.

> I want to add another question about this method. What was the intention to place it inside the BaseThingHandler? Isn't it more or less a util / helper method without any relation to the thing handler itself?

Yes, it is, but it is a helper method that is required often within handlers and for which you need some services, which we don't want to expose directly to the handler developers. By now we think the `ThingHandlerCallback` is the better place to provide such features as it would allow us to completely hide the required services from the handler - see @sjka's comment.

As this PR only adds a slightly different variant of the existing method, I'm ok to merge it.
If you are interested in looking into a refactoring to deprecate those methods and move them to the `ThingHandlerCallback` instead, feel free to do so as a new PR :-)

htreu:
>If you are interested in looking into a refactoring to deprecate those methods and move them to the ThingHandlerCallback instead, feel free to do so as a new PR :-)

@cweitkamp do you have plans to further work on this refactoring?

cweitkamp:
@htreu yes, I do. I have put it onto my TODO list. But feel free to do it on your own.

htreu:
okay, thanks @cweitkamp. I created https://github.com/eclipse/smarthome/issues/5182 to track this issue. It would be nice to see your take on this :-)

PR #5009, opened 2018-01-31
deprecated the Channel constructors

sjka:
...in order to point people to using the builder.

Also, cleaned up all the tests which were still using the constructors.

### DIFF
--- a/bundles/core/org.eclipse.smarthome.core.thing/src/main/java/org/eclipse/smarthome/core/thing/Channel.java
+++ b/bundles/core/org.eclipse.smarthome.core.thing/src/main/java/org/eclipse/smarthome/core/thing/Channel.java
@@ -23,6 +23,7 @@ import org.eclipse.jdt.annotation.NonNullByDefault;
 import org.eclipse.jdt.annotation.Nullable;
 import org.eclipse.smarthome.config.core.Configuration;
 import org.eclipse.smarthome.core.items.Item;
+import org.eclipse.smarthome.core.thing.binding.builder.ChannelBuilder;
 import org.eclipse.smarthome.core.thing.type.ChannelKind;
 import org.eclipse.smarthome.core.thing.type.ChannelTypeUID;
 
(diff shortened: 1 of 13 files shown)


bundles/core/org.eclipse.smarthome.core.thing/src/main/java/org/eclipse/smarthome/core/thing/Channel.java
```diff
@@ -77,19 +82,35 @@ public Channel(ChannelUID uid, String acceptedItemType) {
...
+    @Deprecated
     public Channel(ChannelUID uid, String acceptedItemType, Configuration configuration, Set<String> defaultTags,
             Map<String, String> properties) {
         this(uid, null, acceptedItemType, ChannelKind.STATE, null, defaultTags, properties, null, null);
     }
 
+    /**
+     * @deprecated - use ChannelBuilder instead
+     */
+    @Deprecated
```
maggu2810: Hm, so the `ChannelBuilder` itself is using a `Deprecated` function...

sjka: Yes, and I don't intend to take it away from _him_.

In an ideal world, we would convert the `Channel` class to an interface and hide the implementation class in a private package, allowing the channel builder to access its then non-deprecated constructors (I'm pretty sure that's what you would have done too). But then a couple of people might set a bounty on my head when they discover the `IncompatibleClassChangeError`s  😉 

I'm really looking forward towards our yet outstanding, huge, big, breaking, disastrous API cleanup initiative that we will have to do before a potential 1.0 release (but hardly dare really doing it).

maggu2810: > I'm pretty sure that's what you would have done too

:+1: :wink:

Issue #4356, opened 2017-09-28
Circular reference between services


triller-telekom:
While starting my ESH instance I got this error:

```
!ENTRY org.apache.felix.scr 4 0 2017-09-28 16:54:40.558
!MESSAGE Circular reference detected trying to get service {org.eclipse.smarthome.core.events.EventSubscriber}={service.id=116, service.bundleid=99, service.scope=bundle, component.name=org.eclipse.smarthome.core.autoupdate, component.id=51}
 stack of references: ServiceReference: {org.eclipse.smarthome.core.events.EventSubscriber}={service.id=116, service.bundleid=99, service.scope=bundle, component.name=org.eclipse.smarthome.core.autoupdate, component.id=51}
ServiceReference: {org.osgi.service.event.EventHandler, org.eclipse.smarthome.core.events.EventPublisher}={service.id=115, service.bundleid=62, service.scope=bundle, event.topics=smarthome, component.name=org.eclipse.smarthome.core.internal.events.OSGiEventManager, component.id=75}

!STACK 0
java.lang.Exception: stack trace
	at org.apache.felix.scr.impl.ComponentRegistry.enterCreate(ComponentRegistry.java:510)
	at org.apache.felix.scr.impl.BundleComponentActivator.enterCreate(BundleComponentActivator.java:809)
	at org.apache.felix.scr.impl.manager.SingleComponentManager.getService(SingleComponentManager.java:817)
	at org.eclipse.osgi.internal.serviceregistry.ServiceFactoryUse$1.run(ServiceFactoryUse.java:212)
	at java.security.AccessController.doPrivileged(Native Method)
	at org.eclipse.osgi.internal.serviceregistry.ServiceFactoryUse.factoryGetService(ServiceFactoryUse.java:210)
	at org.eclipse.osgi.internal.serviceregistry.ServiceFactoryUse.getService(ServiceFactoryUse.java:111)
	at org.eclipse.osgi.internal.serviceregistry.ServiceConsumer$2.getService(ServiceConsumer.java:45)
	at org.eclipse.osgi.internal.serviceregistry.ServiceRegistrationImpl.getService(ServiceRegistrationImpl.java:508)
	at org.eclipse.osgi.internal.serviceregistry.ServiceRegistry.getService(ServiceRegistry.java:461)
	at org.eclipse.osgi.internal.framework.BundleContextImpl.getService(BundleContextImpl.java:624)
[...]

PR #5011, opened 2018-02-01
Reworked auto-update infrastructure completely

### LINKED ISSUE #595: Allow bindings to deactivate auto-update for channels

sjka:
This PR re-works the mechanism to automatically send item state events automatically when a command event was received. 

Bindings (i.e. `ThingHandlers` via the `ThingHandlerCallback`)  can set a [AutoUpdatePolicy](https://github.com/eclipse/smarthome/compare/master...sjka:autoUpdate?expand=1#diff-bc3977ef256cdaa4f785612631ccd536R22) for any channel:

| Policy | Description |
|:---|:---|
| VETO | It's not a good idea to send auto-updates, e.g. because the ThingHandler can (and will) do that much better. Makes sense especially for dimmers/rollershutters to have smooth transitions |
| DEFAULT | The ThingHandler has no specific needs or opinions - the default if no policy is set explicitly |
| RECOMMEND | It would be a good idea to send auto-updates (e.g. when a device doesn't expose its current state through its API) - unless another channel is linked to the same item which will take over this duty |

The [behavior](https://github.com/eclipse/smarthome/compare/master...sjka:autoUpdate?expand=1#diff-8f9fc030579e5c7cab577f57cd4f853cR75) of the AutoUpdateManager can be summarized like this:

| Case | Description | Event | Prediction |
|:---|:---|:---|:---|
| REQUIRED | No channels are linked to the item -> act as a "binding" | ✔️  |  - |  
| RECOMMENDED | Channels are linked but RECOMMENDed to send auto-updates | ✔️  | - | 
| OPTIMISTIC | At least one operational channel is linked with DEFAULT policy | - *) | `changeStateTo` |
| DONT | At least one linked channel cast a VETO | - | - |
| REVERT | All linked channel are not working | - | `keepCurrentState` |

Therefore, in comparison to the previous implementation there are two major improvements:

1. Sending automatic state updates before the binding can report back the real device status always happened based on the optimistic assumption that "everything will be alright". This doesn't hold true though when the framework already _knows_ that this can't work, e.g. when the channel or thing does not exist, the ThingHandler is not there (yet) or the thing is in OFFLINE mode. In such cases there will be no automatic state update sent anymore.

2. In the OPTIMISTIC case where the framework can assume that the binding eventually will send such a state update itself, by default now there won't be a automatic state update event sent out anymore. Instead, the prediction callback will be informed so that UIs _still_ can react nicely, but the "real" item state is not affected anymore. This has the huge advantage that e.g. persistence services won't get fed anymore with such "false", optimistic state updates but the item state rather represents the "real" values. RECOMMENDED and REQUIRED cases remain the same, events are still sent out as before.

As a rough summary, these are the changes on technical level:

* dropped the remainings of the AutoUpdateBindingConfigProvider
* dropped the autoupdate "binding" bundle
* Introduced the AutoUpdateManager as part of the core framework
* Allow bindings to RECOMMEND or VETO the auto-update for specific channels
* Created a new hook to allow UIs getting notified about "optimistic" state updates
* Do not really send auto-updates anymore unless no channel is linked or a binding RECOMMENDs to do so for a channel (so that auto-updates can't be mixed-up with "real" updates anymore and e.g. won't get persisted anymore)
* The  previous behavior of always sending item updates can still be configured for backward compatibility - the *) case above

This PR however temporarily drops the means to configure `autoUpdate=...` in item-DSL files. I'd suggest moving them to the item meta-data service which is developed in #4390 so that it will eventually become backward-compatible for end-users again. 

fixes #595
fixes https://github.com/eclipse/smarthome/issues/4356


bundles/core/org.eclipse.smarthome.core.thing/src/main/java/org/eclipse/smarthome/core/thing/binding/AutoUpdatePolicy.java
```diff
@@ -0,0 +1,41 @@
...
+ * A binding's recommendation to the framework whether a state update should be automatically sent to an item if a
+ * command was received.
+ *
+ * @author Simon Kaufmann - initial contribution and API
+ *
+ */
+public enum AutoUpdatePolicy {
+    /**
+     * No automatic state update should be sent by the framework. The handler will make sure it sends a state update and
+     * it can to it better than just converting the command to a state.
```
maggu2810: typo: ... can do it ...

bundles/core/org.eclipse.smarthome.core.thing/src/main/java/org/eclipse/smarthome/core/thing/binding/AutoUpdatePolicy.java
```diff
@@ -0,0 +1,41 @@
...
+     */
+    VETO,
+
+    /**
+     * The binding does not care and the framework may do what it deems to be right. The state update which the
+     * framework will send out normally will correspond the command state anyway. This is the default of no other policy
```
maggu2810: typo: ... default if no ...

bundles/core/org.eclipse.smarthome.core.thing/src/main/java/org/eclipse/smarthome/core/thing/binding/ThingHandlerCallback.java
```diff
@@ -88,4 +88,12 @@
      */
     void channelTriggered(Thing thing, ChannelUID channelUID, String event);
 
+    /**
+     * Give advice to the framework whether automatic state updates should be sent for the given channel or not.
+     *
+     * @param channelUID
```
maggu2810: A description should be added if it is ready for a final review (also if the comment is "the channel UID" only).
Same below and also in other classes...

bundles/io/org.eclipse.smarthome.io.rest.sitemap/src/main/java/org/eclipse/smarthome/io/rest/sitemap/SitemapSubscriptionService.java
```diff
@@ -314,4 +324,32 @@ public void modelChanged(String modelName, EventType type) {
...
+        return null;
+    }
+
+    @Override
+    public void receive(Event event) {
+        if (event instanceof ItemStatePredictedEvent) {
+            ItemStatePredictedEvent prediction = (ItemStatePredictedEvent) event;
+            Item item = itemUIRegistry.get(prediction.getItemName());
+            if (item instanceof GroupItem) {
+                return;
```
kaikreuzer: Could you leave a comment on why we ignore group items here?

bundles/io/org.eclipse.smarthome.io.rest.sitemap/src/main/java/org/eclipse/smarthome/io/rest/sitemap/internal/PageChangeListener.java
```diff
@@ -44,6 +47,7 @@
  */
 public class PageChangeListener implements StateChangeListener {
 
+    private final ScheduledExecutorService scheduler = ThreadPoolManager.getScheduledPool("ui");
```
kaikreuzer: Might this be a candidate for the new "common" pool?

bundles/io/org.eclipse.smarthome.io.rest.sitemap/src/main/java/org/eclipse/smarthome/io/rest/sitemap/internal/PageChangeListener.java
```diff
@@ -186,19 +194,24 @@ public void stateUpdated(Item item, State state) {
...
-                callback.onEvent(event);
-            }
-        }
+        constructAndSendEvents(item, state);
+    }
+
+    public void keepCurrentState(Item item) {
+        scheduler.schedule(() -> {
+            constructAndSendEvents(item, item.getState());
+        }, 200, TimeUnit.MILLISECONDS);
```
kaikreuzer: 200ms might be a bit tight, usually we go for 300ms for UI related feedback.

docs/documentation/development/bindings/thing-definition.md
```diff
@@ -349,3 +349,35 @@ The description can include limited HTML to enhance the display of this informat
...
+        <autoUpdatePolicy>default</autoUpdatePolicy>
+      </channel>
+    </channels>
+</thing-type>
+```
+
+The following policies are supported:
+
+* **veto**: No automatic state update should be sent by the framework. The thing handler will make sure it sends a state update and it can to it better than just converting the command to a state.
+* **default**: The binding does not care and the framework may do what it deems to be right. The state update which the framework will send out normally will correspond the command state anyway. This is the default of no other policy is set explicitly.
```
kaikreuzer: typo: [i]f no other

docs/documentation/development/bindings/thing-definition.md
```diff
@@ -349,3 +349,35 @@ The description can include limited HTML to enhance the display of this informat
...
+* **recommend**: An automatic state update should be sent by the framework because no updates will be sent by the binding. This usually is the case when devices don't expose their current state to the handler.
```
kaikreuzer: better: will be sent -> are sent

bundles/core/org.eclipse.smarthome.core.thing.xml/src/main/java/org/eclipse/smarthome/core/thing/xml/internal/ChannelTypeConverter.java
```diff
@@ -75,6 +77,14 @@ private String readCategory(NodeIterator nodeIterator) throws ConversionExceptio
         return (String) nodeIterator.nextValue("category", false);
     }
 
+    private AutoUpdatePolicy readAutoUpdatePolicy(NodeIterator nodeIterator) {
+        String string = (String) nodeIterator.nextValue("autoUpdatePolicy", false);
+        if (string != null) {
+            return AutoUpdatePolicy.valueOf(string.toUpperCase(Locale.ENGLISH));
```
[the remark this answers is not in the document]
sjka: No. Just (good) habit.

bundles/core/org.eclipse.smarthome.core.thing/src/main/java/org/eclipse/smarthome/core/thing/internal/AutoUpdateManager.java
```diff
@@ -0,0 +1,298 @@
...
+        if (command instanceof State) {
+            final State state = (State) command;
+
+            Recommendation autoUpdate = shouldAutoUpdate(itemName);
+
+            // consider user-override via item meta-data
+            MetadataKey key = new MetadataKey("autoupdate", itemName);
+            Metadata metadata = metadataRegistry.get(key);
+            if (metadata != null && !metadata.getValue().trim().isEmpty()) {
+                boolean override = Boolean.getBoolean(metadata.getValue());
```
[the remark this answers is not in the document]
sjka: Argh, thanks! Of course...

sjka:
@maggu2810 Sorry, I tried addressing your comments by making it a characteristic of the Channel/ChannelType, but clearly forgot to explicitly mention it here. 
Therefore you can:
1. as a binding, indicate the preference via the Channel (or rather the ChannelType, as its "template")
1. as a user, set the preference on the item's metadata (i.e. `autoupdate:<ITEM_NAME>`)
1. and last but not least, if you want to do this programmatically, you could also either set the corresponding metadata in the ManagedMetadataProvider or even have your own `MetadataProvider` which dynamically provides the corresponding metadata, e.g. for all items linked to a certain binding. In the latter case you'd just need to be sure to not interfere with the user and thereby have another `MetadataProvider` which provides metadata for the same keys.

Do you think one of these would cover your scenario?

kaikreuzer:
Just fresh from testing with the ESH demo setup:

```
smarthome send Weather_Temperature 1
Command has been sent successfully.
osgi> 17:25:31.834 INFO  s.event.ItemCommandEvent[:53] - Item 'Weather_Temperature' received command 1
17:25:31.838 DEBUG o.e.s.c.t.i.p.ProfileCallbackImpl[:69] - Delegating command '1 °C' for item 'Weather_Temperature' to handler for channel 'yahooweather:weather:berlin:temperature'
17:25:31.839 DEBUG o.e.s.c.i.c.InvocationHandlerSync[:55] - Already in a safe-call context, executing 'ThingHandler.handleCommand()' directly on 'org.eclipse.smarthome.binding.yahooweather.handler.YahooWeatherHandler@1e1cfe47'.
17:25:31.840 DEBUG o.e.s.b.y.h.YahooWeatherHandler[:157] - Command 1 °C is not supported for channel: temperature
17:25:31.843 ERROR o.e.s.c.i.events.EventHandler[:123] - Creation of ESH-Event failed, because one of the registered event factories has thrown an exception: Error getting class for simple name: 'nullType' using package name 'org.eclipse.smarthome.core.library.types.'.
java.lang.IllegalArgumentException: Error getting class for simple name: 'nullType' using package name 'org.eclipse.smarthome.core.library.types.'.
	at org.eclipse.smarthome.core.items.events.ItemEventFactory.parseSimpleClassName(ItemEventFactory.java:183)
	at org.eclipse.smarthome.core.items.events.ItemEventFactory.parseType(ItemEventFactory.java:158)
	at org.eclipse.smarthome.core.items.events.ItemEventFactory.getState(ItemEventFactory.java:136)
	at org.eclipse.smarthome.core.items.events.ItemEventFactory.createStatePredictedEvent(ItemEventFactory.java:123)
	at org.eclipse.smarthome.core.items.events.ItemEventFactory.createEventByType(ItemEventFactory.java:82)
	at org.eclipse.smarthome.core.events.AbstractEventFactory.createEvent(AbstractEventFactory.java:50)
	at org.eclipse.smarthome.core.internal.events.EventHandler.createESHEvent(EventHandler.java:121)
	at org.eclipse.smarthome.core.internal.events.EventHandler.handleEvent(EventHandler.java:95)
	at org.eclipse.smarthome.core.internal.events.EventHandler.handleEvent(EventHandler.java:72)
	at org.eclipse.smarthome.core.internal.events.ThreadedEventHandler.lambda$0(ThreadedEventHandler.java:67)
Caused by: java.lang.ClassNotFoundException: org.eclipse.smarthome.core.library.types.nullType cannot be found by org.eclipse.smarthome.core_0.10.0.qualifier
	at org.eclipse.osgi.internal.loader.BundleLoader.findClassInternal(BundleLoader.java:484)
	at org.eclipse.osgi.internal.loader.BundleLoader.findClass(BundleLoader.java:395)
	at org.eclipse.osgi.internal.loader.BundleLoader.findClass(BundleLoader.java:387)
	at org.eclipse.osgi.internal.loader.ModuleClassLoader.loadClass(ModuleClassLoader.java:150)
	at java.lang.ClassLoader.loadClass(ClassLoader.java:357)
	at java.lang.Class.forName0(Native Method)
	at java.lang.Class.forName(Class.java:264)
	at org.eclipse.smarthome.core.items.events.ItemEventFactory.parseSimpleClassName(ItemEventFactory.java:179)
	at org.eclipse.smarthome.core.items.events.ItemEventFactory.parseType(ItemEventFactory.java:158)
	at org.eclipse.smarthome.core.items.events.ItemEventFactory.getState(ItemEventFactory.java:136)
	at org.eclipse.smarthome.core.items.events.ItemEventFactory.createStatePredictedEvent(ItemEventFactory.java:123)
```

sjka:
Oh well, nasty little conflict-resolution-mistake I did when rebasing.
À propos: I will do another rebase 😩 Hang on...

kaikreuzer:
If you mean `autoupdate="false"` then simply keep doing the very same - nothing has changed about that. The feature is meant for bindings to be able to influence the behavior on their own.

lsafelix75:
Any binding example on how to apply this autoupdate feature? After reading this, i am still not clear what need to be done. Cheers

maggu2810:
It depends on your use case.
For the bindings I don't want the auto update mechanism I added this

```
<autoUpdatePolicy>veto</autoUpdatePolicy>
```

to my channel types in the XML.

lsafelix75:
> It depends on your use case.
> For the bindings I don't want the auto update mechanism I added this
> 
> ```
> <autoUpdatePolicy>veto</autoUpdatePolicy>
> ```
> 
> to my channel types in the XML.

ok.. thanks.

unparagoned:
It would be great if there was a global setting we could use to set to work the way it did before, or an easy way to make it work like it was before. I use custom items to trigger my rules, but that's not really working well now.

Issue #4014, opened 2017-08-10
PaperUI: Configuration field with limitToOptions false does not show stored value


triller-telekom:
A configuration field in PaperUI has the capability of delivering options via a `ConfigOptionProvider`. If you set the field to `limitToOptions = false`  and enter a custom value, this value will be stored. However if you leave the page and reenter the page this value is gone.

sjka:
When fixing this, please consider [this comment](https://github.com/eclipse/smarthome/issues/4895#issuecomment-356945467), i.e.

> It needs to ensure the value is "min <> max || options" in this case.

Issue #3556, opened 2017-06-01
[PaperUI] multiselect with limitToOptions=false select behaviour


htreu:
Follow up from #3246. Once #3555 is merged a the bug can easily be reproduced.

Original description from @cdjackson:

>The configuration parameter dropdown is rendered as a list only and it's not possible to enter values other than those in the list.

>My mistake - it does work, however once data is typed into the editbox, it's not possible to select from the list without emptying the editbox.

PR #5012, opened 2018-02-01
Config parameter: limitToOptions Sync code with documentation

### LINKED ISSUE #2844: default value of limitToOptions is unclear
The [documentation](https://www.eclipse.org/smarthome/documentation/development/bindings/xml-reference.html#xml-structure-for-configuration-descriptions) says that the default value of the "limitToOptions" property is true in case it is not specified. The [code](https://github.com/eclipse/smarthome/blob/master/bundles/config/org.eclipse.smarthome.config.core/src/main/java/org/eclipse/smarthome/config/core/ConfigDescriptionParameter.java#L91) sets it to false, though.

Only one of both can be right - which one?

triller-telekom:
Fixes #2844

Since limitToOptions = false is currently broken anyway, see #4014 I think we can now sync the code with the documentation to at least starting to fix all related issues (see also #3556).


kaikreuzer:
Note: I marked it as API breaking as it now changes the config descriptions where this option isn't mentioned. (@cdjackson fyi, just in case)

PR #5013, opened 2018-02-01
Remove some wrong @NonNull annotations in core.thing

triller-telekom:

### DIFF
--- a/bundles/core/org.eclipse.smarthome.core.thing/src/main/java/org/eclipse/smarthome/core/thing/ThingTypeMigrationService.java
+++ b/bundles/core/org.eclipse.smarthome.core.thing/src/main/java/org/eclipse/smarthome/core/thing/ThingTypeMigrationService.java
@@ -12,7 +12,7 @@
  */
 package org.eclipse.smarthome.core.thing;
 
-import org.eclipse.jdt.annotation.NonNull;
+import org.eclipse.jdt.annotation.NonNullByDefault;
 import org.eclipse.smarthome.config.core.Configuration;
 
 /**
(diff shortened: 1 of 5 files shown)


bundles/core/org.eclipse.smarthome.core.thing/src/main/java/org/eclipse/smarthome/core/thing/ThingTypeMigrationService.java
```diff
@@ -32,6 +33,6 @@
      *
      * @throws RuntimeException, if the new thing type is not registered in the registry
      */
-    void migrateThingType(@NonNull Thing thing, @NonNull ThingTypeUID thingTypeUID, Configuration configuration);
+    void migrateThingType(Thing thing, ThingTypeUID thingTypeUID, Configuration configuration);
```
sjka: Sorry for being too slow, but what about `@Nullable Configuration`?

triller-telekom: I wasn't 100% sure about it and there was no test that complained about it, i.e. no error because of one passing `null` to it. Is it possible to pass `null` here` if so I can create a follow-up PR.

sjka: Well, in this case you'll have to follow the white rabbit - then you will see that the implementation handles `null` values gracefully and that it is anticipated.

triller-telekom: The white rabbit was too quick for me :P Follow-up PR is here: #5016

sjka: Bloody rodent... 😉 
Thanks for the fix!

PR #5014, opened 2018-02-01
add time-only and date-only parsing from string

### LINKED ISSUE #5001: DateTimeType should also support dates and times separately
The valueOf method of `DateTimeType` currently requires a fully specified date+time combination.
This type is nonetheless also used for time-only use cases (e.g. an alarm clock for "08:00") or for date only use cases (e.g. next waste collection on "03.02.2018").

I do not see any strong reason why we should not support type creation with only date or time component, so I'd hope that this could be added.

J-N-K:
fixes #5001

### DIFF
--- a/bundles/core/org.eclipse.smarthome.core.test/src/test/java/org/eclipse/smarthome/core/library/types/DateTimeTypeTest.java
+++ b/bundles/core/org.eclipse.smarthome.core.test/src/test/java/org/eclipse/smarthome/core/library/types/DateTimeTypeTest.java
@@ -146,7 +146,9 @@ public class DateTimeTypeTest {
                 { new ParameterSet(TimeZone.getTimeZone("GMT-2"), initTimeMap(), TimeZone.getTimeZone("GMT-4"),
                         "2014-03-30T12:58:47.033-0200") },
                 { new ParameterSet(TimeZone.getTimeZone("UTC"), "10:58:47", "1970-01-01T10:58:47.000+0000") },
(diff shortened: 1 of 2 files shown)


bundles/core/org.eclipse.smarthome.core.test/src/test/java/org/eclipse/smarthome/core/library/types/DateTimeTypeTest.java
```diff
@@ -143,7 +144,11 @@ public ParameterSet(TimeZone defaultTimeZone, String inputTimeString, String exp
                 { new ParameterSet(TimeZone.getTimeZone("GMT-2"), initTimeMap(), TimeZone.getTimeZone("GMT+3"),
                         "2014-03-30T05:58:47.033-0200") },
                 { new ParameterSet(TimeZone.getTimeZone("GMT-2"), initTimeMap(), TimeZone.getTimeZone("GMT-4"),
-                        "2014-03-30T12:58:47.033-0200") }, });
+                        "2014-03-30T12:58:47.033-0200") },
+                { new ParameterSet(TimeZone.getTimeZone("UTC"), "10:58:47", "1970-01-01T10:58:47.000+0000") },
```
sjka: Does it work with only "10:58" too? This was the example given in the referenced issue and I reckon it's really important to support it!

J-N-K:
I‘ll check that.

sjka:
Wonderful, thanks!

PR #5018, opened 2018-02-02
normalize channel configurations in ThingResource

### LINKED ISSUE #4910: Configuration normalisation not performed on channel configuration
When saving channel configuration, normalisation is not performed on the channel configuration. Additionally, it looks like the defaults may not be of the correct type - defaults are specified as a ```String``` (ie in ConfigDescriptionParameterBuilder) and should be normalised before being passed to the binding.

For example, the following -:

```
            options = new ArrayList<ParameterOption>();
            options.add(new ParameterOption("65535", "Use On/Off transition time"));
            parameters.add(ConfigDescriptionParameterBuilder.create(CONFIG_ONTRANSITIONTIME, Type.INTEGER)
                    .withLabel("On Transition Time")
                    .withDescription("Time in 10ms intervals to transition from OFF to ON").withDefault("65535")
                    .withMinimum(new BigDecimal(0)).withMaximum(new BigDecimal(60000)).withOptions(options)
                    .withLimitToOptions(false).build());
```

will provide a String when the configuration is read from the channel.

sjka:
...in the same manner as with Thing configurations.

fixes #4910

### DIFF
--- a/bundles/io/org.eclipse.smarthome.io.rest.core/src/main/java/org/eclipse/smarthome/io/rest/core/internal/thing/ThingResource.java
+++ b/bundles/io/org.eclipse.smarthome.io.rest.core/src/main/java/org/eclipse/smarthome/io/rest/core/internal/thing/ThingResource.java
@@ -77,6 +77,9 @@ import org.eclipse.smarthome.core.thing.i18n.ThingStatusInfoI18nLocalizationServ
 import org.eclipse.smarthome.core.thing.link.ItemChannelLink;
 import org.eclipse.smarthome.core.thing.link.ItemChannelLinkRegistry;
 import org.eclipse.smarthome.core.thing.link.ManagedItemChannelLinkProvider;
+import org.eclipse.smarthome.core.thing.type.ChannelType;
+import org.eclipse.smarthome.core.thing.type.ChannelTypeRegistry;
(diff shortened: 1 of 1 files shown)


sjka:
@cdjackson would you mind verifying that his indeed fixes your use-case?

cdjackson:
Will look at this over the next few days - thanks @sjka

cdjackson:
@sjka I think this is working ok. There are some other issues I need to look at but the normalisation seems ok.

We miss some sort of ```handleConfigurationUpdate``` method when the channel data is updated...

bundles/io/org.eclipse.smarthome.io.rest.core/src/main/java/org/eclipse/smarthome/io/rest/core/internal/thing/ThingResource.java
```diff
@@ -350,6 +362,15 @@ public Response update(@HeaderParam(HttpHeaders.ACCEPT_LANGUAGE) @ApiParam(value
         // check configuration
         thingBean.configuration = normalizeConfiguration(thingBean.configuration, thing.getThingTypeUID(),
                 thing.getUID());
+        if (thingBean.channels != null) {
```
htreu: This code is very similar to the new block at lines 186 - 193. This could be refactored into a helper method.

PR #5021, opened 2018-02-02
Fix cyclic dependency bewteen ThingConfigDescProv and its registry

### LINKED ISSUE #4300: Circular dependency: ThingConfigDescriptionProvider & ConfigDescriptionRegistry
No idea why we haven't seen this before, but just ran into this one:

```
Circular reference detected trying to get service {org.eclipse.smarthome.config.core.ConfigDescriptionRegistry}={service.id=127, service.bundleid=86, service.scope=bundle, component.name=ConfigDescriptionRegistry, component.id=119}
 stack of references: ServiceReference: {org.eclipse.smarthome.config.core.ConfigDescriptionRegistry}={service.id=127, service.bundleid=86, service.scope=bundle, component.name=ConfigDescriptionRegistry, component.id=119}
ServiceReference: {org.eclipse.smarthome.config.core.ConfigDescriptionProvider}={service.id=128, service.bundleid=78, service.scope=bundle, component.name=org.eclipse.smarthome.core.thing.internal.ThingConfigDescriptionProvider, component.id=108}
```

Looks like SCR is right :-(

sjka:
...by obtaining a registry reference "on demand", due to the lack of a better idea
how to fix it without reworking pretty much everything.

If anybody has a better idea, I'd be more than happy to drop this. 

fixes #4300


kaikreuzer:
I would much prefer to resolve the cyclic dependency. Looking at the code, it feels pretty wrong to access the registry ourself here.
As the JavaDoc already sayss, this class is merely acting as a "proxy" or as I would call it, an "alias": It merely makes existing (thing-type or channel) config descs available under a different url.
I see two ways to solve that:
1. Quick&dirty: Do such aliasing (from thing to thing type) directly in the ConfigDescriptionRegistry as a fallback, if no provider has an answer.
2. Introduce a new kind of AliasHandler interface which can take a uri and return a different uri, which the registry will then use for the look up.

bundles/config/org.eclipse.smarthome.config.core/src/main/java/org/eclipse/smarthome/config/core/ConfigDescriptionAliasHandler.java
```diff
@@ -0,0 +1,38 @@
...
+import org.eclipse.jdt.annotation.Nullable;
+
+/**
+ * Can be implemented to point one config description URI to another.
+ *
+ * @author Simon Kaufmann - initial contribution and API
+ *
+ */
+@NonNullByDefault
+public interface ConfigDescriptionAliasHandler {
```
kaikreuzer: Any special reason why you chose to call it handler and not provider? After all, it does not handle aliases, but it provides them.

bundles/config/org.eclipse.smarthome.config.core/src/main/java/org/eclipse/smarthome/config/core/ConfigDescriptionRegistry.java
```diff
@@ -149,9 +170,18 @@ protected void removeConfigDescriptionProvider(ConfigDescriptionProvider configD
         List<ConfigDescriptionParameter> parameters = new ArrayList<ConfigDescriptionParameter>();
         List<ConfigDescriptionParameterGroup> parameterGroups = new ArrayList<ConfigDescriptionParameterGroup>();
 
+        URI aliasedURI = uri;
+        for (ConfigDescriptionAliasHandler aliasHandler : configDescriptionAliasHandlers) {
```
kaikreuzer: I'd think this code must be moved further below. The alias should only kick in as a fallback, if no other provider was able to provide a specific description. (request for thing -> no thing desc found -> therefore using thing-type desc of thing).

bundles/config/org.eclipse.smarthome.config.core/src/main/java/org/eclipse/smarthome/config/core/ConfigDescriptionRegistry.java
```diff
@@ -166,11 +196,11 @@ protected void removeConfigDescriptionProvider(ConfigDescriptionProvider configD
             List<ConfigDescriptionParameter> parametersWithOptions = new ArrayList<ConfigDescriptionParameter>(
                     parameters.size());
             for (ConfigDescriptionParameter parameter : parameters) {
-                parametersWithOptions.add(getConfigOptions(uri, parameter, locale));
+                parametersWithOptions.add(getConfigOptions(aliasedURI, parameter, locale));
```
kaikreuzer: Here you are missing the whole point of the aliases: The UI requests the config description of a thing (and not its thing-type directly) because there might be config options available that are only applicable for this thing, but not all of the same type. Therefore, you'd have to ask for the config options of the original uri - and yes, maybe here again, you could use the aliasedURI results as a fallback, if you do not get anything else.

bundles/config/org.eclipse.smarthome.config.core/src/main/java/org/eclipse/smarthome/config/core/ConfigDescriptionRegistry.java
```diff
@@ -166,11 +196,11 @@ protected void removeConfigDescriptionProvider(ConfigDescriptionProvider configD
...
             }
 
             // Return the new configuration description
-            return new ConfigDescription(uri, parametersWithOptions, parameterGroups);
+            return new ConfigDescription(aliasedURI, parametersWithOptions, parameterGroups);
```
kaikreuzer: Hm, again, shouldn't you return a config description for the uri that was originally requested?

sjka:
Of course, you were right with your feedback! 
I'm not quite sure if my brain is functioning any better now, but I added some [test-cases](https://github.com/eclipse/smarthome/pull/5021/commits/1b39d41c6e1580ebecb7288f99d0fdcaee166ee2) now, just in case... 😉

bundles/config/org.eclipse.smarthome.config.core/src/main/java/org/eclipse/smarthome/config/core/ConfigDescriptionRegistry.java
```diff
@@ -149,24 +174,24 @@ protected void removeConfigDescriptionProvider(ConfigDescriptionProvider configD
...
-                parameterGroups.addAll(config.getParameterGroups());
+        boolean found = fillFromProviders(uri, locale, parameters, parameterGroups);
+        URI alias = null;
+        if (!found) {
+            for (ConfigDescriptionAliasProvider aliasProvider : configDescriptionAliasProviders) {
+                alias = aliasProvider.getAlias(uri);
+                if (alias != null) {
+                    logger.debug("No config description found for '{}', using alias '{}' instead", uri, alias);
+                    found = fillFromProviders(alias, locale, parameters, parameterGroups);
+                    break;
```
maggu2810: Shouldn't this `break` only be called if `found == true`?

sjka: This was intended initially, but on a second thought I can't explain why it _should_ be like that. I certainly doesn't hurt to continue until the first "successful" alias, if there are multiple.

PR #5025, opened 2018-02-02
Provide preconfigured ChannelBuilders

### LINKED ISSUE #4950: Improve ThingHandler API to allow easy addition of Channels
In our [thing modelling FAQs](https://www.eclipse.org/smarthome/documentation/development/bindings/faq.html#structuring-things-and-thing-types), we suggest as (4) to dynamically add channels to a Thing within the handler.
As [discussed here](https://github.com/openhab/openhab2-addons/issues/2930#issuecomment-358709502), this is currently a bit cumbersome, though, as there isn't an easy way to construct a channel from a channel type, which means that the best code I could come up with is [this](https://github.com/eclipse/smarthome/pull/4535/files#diff-4cfd867ae80654e19a6695ee0d6b32c2R217).

We should simplify this either by improving the ChannelBuilder or possibly by directly supporting it in the ThingBuilder (not sure yet, what is the better option).

sjka:
...for any given ChannelTypeUID in the ThingHandlerCallback and
in the BaseThingHandler correspondingly.

Unlike what #4950 suggested, I didn't put this neither into `ChannelBuilder` nor `ThingBuilder` directly, but rather placed it like a "factory" method into the `ThingHandlerHelper` - mainly because it requires access to some OSGi-services. 

For the actual creation of the `ChannelBuilder` I reused the logic within `ThingFactoryHelper` as it knew already how to do it - only a little refactoring was needed in order to make it reusable. Bindings would use it as always via the new method in `BaseThingHandler`.

While I "was there" anyway, I also 
* ported the ThingBuilderTest to java
* inserted a check into ThingBuilder which ensures that the to-be-added ChannelUIDs match their ThingUID

fixes #4950


bundles/core/org.eclipse.smarthome.core.thing.test/src/test/java/org/eclipse/smarthome/core/thing/binding/ThingBuilderTest.java
```diff
@@ -0,0 +1,86 @@
...
+ */
+public class ThingBuilderTest {
+
+    private static final ThingTypeUID THING_TYPE_UID = new ThingTypeUID("test", "test");
+    private static final ThingUID THING_UID = new ThingUID(THING_TYPE_UID, "test");
+
+    @Test(expected = IllegalArgumentException.class)
+    public void testWithChannel_duplicates() {
+        ThingBuilder thingBuilder = ThingBuilder.create(THING_TYPE_UID, THING_UID);
+        thingBuilder.withChannel(new Channel(new ChannelUID(THING_UID, "channel1"), ""));
```
htreu: In the light of #5009: would you directly change here to the ChannelBuilder?

bundles/core/org.eclipse.smarthome.core.thing/src/main/java/org/eclipse/smarthome/core/thing/binding/builder/ThingBuilder.java
```diff
@@ -69,18 +70,22 @@ public ThingBuilder withLabel(@Nullable String label) {
...
+        validateChannelUIDs(Collections.singletonList(channel));
         ThingHelper.ensureUniqueChannels(mutableThingChannels, channel);
         mutableThingChannels.add(channel);
         return this;
     }
 
     public ThingBuilder withChannels(Channel... channels) {
-        ThingHelper.ensureUniqueChannels(channels);
-        this.thing.setChannels(new ArrayList<>(Arrays.asList(channels)));
+        List<Channel> newChannels = Arrays.asList(channels);
```
maggu2810: Wouldn't it be better to call `return withChannels(Arrays.asList(channels));` instead of adding the same implementation twice? Or do I miss something?

bundles/core/org.eclipse.smarthome.core.thing/src/main/java/org/eclipse/smarthome/core/thing/internal/ThingFactoryHelper.java
```diff
@@ -144,8 +143,27 @@ private static Channel createChannel(ChannelDefinition channelDefinition, ThingU
             channelBuilder = channelBuilder.withDescription(channelDefinition.getDescription());
         }
 
+        channelBuilder = channelBuilder.withProperties(channelDefinition.getProperties());
+
+        Channel channel = channelBuilder.build();
```
maggu2810: IIRC the JVM could do better optimization if you don't create a temporary (useless) variable.

htreu:
Travis is still unhappy:
```
[ERROR] /home/travis/build/eclipse/smarthome/bundles/core/org.eclipse.smarthome.core.thing.test/src/test/java/org/eclipse/smarthome/core/thing/internal/ThingManagerOSGiJavaTest.java:[206] 
[ERROR] 	Channel channel = channelBuilder.build();
[ERROR] 	^^^^^^^
[ERROR] Channel cannot be resolved to a type
[ERROR] -> [Help 1]
```

maggu2810:
```text
testCreateChannelBuilder(org.eclipse.smarthome.core.thing.internal.ThingManagerOSGiJavaTest)  Time elapsed: 0.182 sec  <<< ERROR!
java.lang.IllegalArgumentException: Channel type binding:channel is not known
	at org.eclipse.smarthome.core.thing.internal.ThingManager$1.createChannelBuilder(ThingManager.java:264)
	at org.eclipse.smarthome.core.thing.internal.ThingManagerOSGiJavaTest.testCreateChannelBuilder(ThingManagerOSGiJavaTest.java:205)
```

bundles/core/org.eclipse.smarthome.core.thing/src/main/java/org/eclipse/smarthome/core/thing/binding/BaseThingHandler.java
```diff
@@ -673,4 +675,20 @@ protected void changeThingType(ThingTypeUID thingTypeUID, Configuration configur
...
 
+    /**
+     * Create a {@link ChannelBuilder} which is preconfigured with values from the given channel type.
+     *
+     * @param channelId the ID of the channel to be created
+     * @param channelTypeUID the channel type UID for which the channel should be created
+     * @return a preconfigured ChannelBuilder
+     * @throw {@link IllegalArgumentException} if the referenced channel type is not known
+     */
+    protected ChannelBuilder createChannelBuilder(String channelId, ChannelTypeUID channelTypeUID) {
```
kaikreuzer: Do you really think that we should introduce this method on the BaseThingHandler?
Or wouldn't it suffice to provide this through the callback? After all, this method has nothing really to do with the current handler instance, but it is rather a helper/convenience/utility method that is provided by the framework.

kaikreuzer:
@sjka Any update on this?

bundles/core/org.eclipse.smarthome.core.thing/src/main/java/org/eclipse/smarthome/core/thing/binding/BaseThingHandler.java
sjka:
```java
public abstract class BaseThingHandler implements ThingHandler {
    // ...
    private @Nullable ThingHandlerCallback callback;
    // ...
}
```
How to you suggest using it in subclasses?

kaikreuzer: The same way as subclasses are supposed to use the callback for other things. ("Not at all" does not seem to be a valid answer...?)

sjka: Sorry, somehow I don't get it. Do you suggest to expose the callback instead to subclasses?

PR #5027, opened 2018-02-02
Add NonNull to firmware handler

### LINKED ISSUE #5026: FirmwareUpdateHandler.updateFirmware Nullable errors
As seen below, the parameters here show ```@Nullable```, but the javadoc says ```(not null)```.

image

image

PR on the way to resolve this.

cdjackson:
Fixes #5026 


kaikreuzer:
Could you maybe directly address the whole interface by adding `NonNullByDefault` (which is the normal case according to our coding guidelines)?
It would e.g. then also cover [this line](https://github.com/eclipse/smarthome/pull/5027/files#diff-e909c9ce5e17bf8d8a908b30e979733cR34) and the other methods.

cdjackson:
Updated...

PR #5032, opened 2018-02-05
switch to manually maintained REST API version

kaikreuzer:
We have so far used the bundle version of the io.rest bundle as a version that we expose through the REST API.
The primary use case is to allow REST clients to figure out, which version they are talking to, i.e. whether certain features are already available or not.

Unfortunately, there is a problem with this this approach: Our different builds result in different kinds of version syntax, e.g. 0.10.0, 0.10.0.201802051234, 0.10.0.b1 or 0.10.0.RC1.
If client comes across those versions, it has the problem that it has no clue how to order them. Is 0.10.0.b1 newer or older than 0.10.0.201802051234? There is no way to figure this out. (Note: this is exactly the situation for openHAB 2.2.0 release, which includes 0.10.0.b1).

Imho, the only satisfying solution for this is to manually maintain the REST API version. We should increase the minor version whenever we add new features to the REST API and increase the major version, whenever we break some existing feature on it.

My suggestion is to start with 1.0 as this is bigger than any previous 0.x.0 version.

@sjka & @maggu2810, wdyt, does this make sense?

/cc @lolodomo & @mueller-ma

### DIFF
--- a/bundles/io/org.eclipse.smarthome.io.rest/src/main/java/org/eclipse/smarthome/io/rest/internal/resources/beans/RootBean.java
+++ b/bundles/io/org.eclipse.smarthome.io.rest/src/main/java/org/eclipse/smarthome/io/rest/internal/resources/beans/RootBean.java
@@ -24,7 +24,7 @@ import java.util.List;
  */
 public class RootBean {
 
-    public final String version = "1.0";
+    public final String version = "1";
 
     public final List<Links> links = new ArrayList<Links>();
 
--- a/bundles/io/org.eclipse.smarthome.io.rest/src/main/java/org/eclipse/smarthome/io/rest/internal/resources/beans/RootBean.java
+++ b/bundles/io/org.eclipse.smarthome.io.rest/src/main/java/org/eclipse/smarthome/io/rest/internal/resources/beans/RootBean.java
@@ -15,18 +15,16 @@ package org.eclipse.smarthome.io.rest.internal.resources.beans;
 import java.util.ArrayList;
 import java.util.List;
 
-import org.eclipse.smarthome.io.rest.internal.RESTActivator;
-
 /**
  * This is a java bean that is used to define the root entry
  * page of the REST interface.
- * 
+ *
  * @author Kai Kreuzer - Initial contribution and API
  *
  */
 public class RootBean {
 
-    public final String version = RESTActivator.getContext().getBundle().getVersion().toString();
+    public final String version = "1.0";
 
     public final List<Links> links = new ArrayList<Links>();
 


maggu2810:
I understand the problem of openHAB but I don't think that the solution is to migrate from an automatic versioning to a manual one.
Isn't the problem that openHAB is using snapshots and so unreleased development versions of ESH?
I also don't like to start using different version for the different parts of ESH.

kaikreuzer:
No, it is not a problem that is specific to openHAB - I just used this example to point out that the problem is real.

> Isn't the problem that openHAB is using snapshots and so unreleased development versions of ESH?

No, not at all. The problem is actually with the release versions - be it alphas, betas, RCs or whatever. So the only working solution would be to avoid any of those, which is imho not desirable.
Additionally, I also think that clients must be able to work with snapshot builds of ESH as well, especially as there isn't such a thing as a snapshot for an OSGi bundle.

maggu2810:
IIRC OSGi versions uses a major, minor, micro and qualifier component. The qualifier could use letters, digits, underscores and hyphen. There is no rule that the qualifier needs to be a timestamp as it is done by our build tools (AFAIK). It could also be alpha, beta, ga, rc etc.
I don't know if there is a rule between the ordering of classifiers other then alphanumeric.

If we solve it for REST using manual versioning, the same problem remains for the bundles versions. Is the bundle version 0.10.0.b1 greater then 0.10.0.201802051234? This depends on the used comparator implementation, too. Didn't it?

Wouldn't it be better if we drop the bX version scheme at all and increase the micro number instead?
So we don't use non-nummeric version components at all.

* 0.9.0
* ...
* 0.10.0.YYYYMMDDhhmm
* 0.10.1 (formerly 0.10.0.b1)
* 0.10.1.YYYYMMDDhhmm

pfink:
Just for record: Maven implemented a an order:
<ul>
<li>mixing of '<code>-</code>' (dash) and '<code>.</code>' (dot) separators,</li>
<li>transition between characters and digits also constitutes a separator:
    <code>1.0alpha1 =&gt; [1, 0, alpha, 1]</code></li>
<li>unlimited number of version components,</li>
<li>version components in the text can be digits or strings,</li>
<li>strings are checked for well-known qualifiers and the qualifier ordering is used for version ordering.
    Well-known qualifiers (case insensitive) are:<ul>
    <li><code>alpha</code> or <code>a</code></li>
    <li><code>beta</code> or <code>b</code></li>
    <li><code>milestone</code> or <code>m</code></li>
    <li><code>rc</code> or <code>cr</code></li>
    <li><code>snapshot</code></li>
    <li><code>(the empty string)</code> or <code>ga</code> or <code>final</code></li>
    <li><code>sp</code></li>
    </ul>
    Unknown qualifiers are considered after known qualifiers, with lexical order (always case insensitive),
  </li>
<li>a dash usually precedes a qualifier, and is always less important than something preceded with a dot.</li>
</ul></p>

Source: https://github.com/apache/maven/blob/maven-3.3.9/maven-artifact/src/main/java/org/apache/maven/artifact/versioning/ComparableVersion.java#L32

kaikreuzer:
> If we solve it for REST using manual versioning, the same problem remains for the bundles versions. 

No, according to OSGi best practices, bundle versions are anyhow pure "marketing versions", while you should use package versioning in your dependencies (if you want to refer to a specific version). And there, we have so far the agreement that a solution has to make sure by itself that it packages only bundles that are compatible with each other as we are not using any package versioning.

> Is the bundle version 0.10.0.b1 greater then 0.10.0.201802051234? This depends on the used comparator implementation, too. Didn't it?

No, how could it? b1 was done at some point in time that you cannot figure out from "b1". There is no way to implement a comparator without hard-coding this information.

> Wouldn't it be better if we drop the bX version scheme at all and increase the micro number instead?

An alpha/beta/milestone/RC is a pretty different concept than a patch release, so these are not interchangable. A patch release is by definition newer than the release, while the others are by definition older.

> A bit hacky, but compliant with "the maven way" would be to detect whether it's a snapshot version and to append (or prepend) the word SNAPSHOT to the timestamp.

@pfink: No, you are missing that Maven and OSGi versioning are inherently incompatible - it does not matter with which solution you try to come up, you won't find one where you could use a Maven comparator on an OSGi version.

pfink:
> @pfink: No, you are missing that Maven and OSGi versioning are inherently incompatible - it does not matter with which solution you try to come up, you won't find one where you could use a Maven comparator on an OSGi version.

Sure, I was not talking about changing the OSGi version, just the version to be exposed via REST. In fact, REST clients can use the Maven comparator.

kaikreuzer:
You mean somehow getting hold of the Maven version and exposing that? Yes, that might be an option, but that also feels rather like a hack and the logic to implement on client side is still quite complex.

pfink:
Yes, or "mavenize" the OSGi snapshot version with a simple logic (which is even hackier, I admit). And you're right, the logic on client-side is complex. Personally, I don't think it's better than the approach suggested by this PR, just wanted to mention that it would be an option.

maggu2810:
> No, how could it? b1 was done at some point in time that you cannot figure out from "b1". There is no way to implement a comparator without hard-coding this information.

That's the reason I have written it depends on the comparator implementation. The Maven Comparator has its rule. I didn't say it needs to be correct -- I would like to say it could be also wrong -- as it missed the information "we" know.

> You mean somehow getting hold of the Maven version and exposing that? Yes, that might be an option,

But isn't this the current situation (if a timestamp is used it is a maven snapshot and in the other cases it fits to the maven one) that you want to change?

> An alpha/beta/milestone/RC is a pretty different concept than a patch release, so these are not interchangable. A patch release is by definition newer than the release, while the others are by definition older.

Sure, you are right, if we still want that the official release uses a micro number equal to zero.
There has been also products that uses odd minor number for the development only.

I still would prefer if we could use something that reflect the e.g. bundle version and does not need to hard code a version number (and the wired manual maintenance).

sjka:
If I understand the previous discussion correctly, there are several aspects mixed-in here. One is that the semantics for bundle versions are inherently incompatible between OSGi and maven (see e.g. https://bugs.eclipse.org/bugs/show_bug.cgi?id=382482). We won't be able to fix this or solve it by discussing it. 

There are workarounds to that - as @maggu2810 mentioned one could implement some even/odd semantics in the minor version for snapshots vs. releases. However, that doesn't help the "beta milestone"-problem because they explicitly must not be "official" releases because they are intended to just provide a path towards such a release with the full IP log enchilada. So not an option for us.

Also, what I see critical is the fact that the bundle versions are only "marketing" labels, i.e. they don't tell anything about the REST API version compatibility. I mean, the whole point of a version attribute in the REST API is about allowing clients to determine upfront if they will be able to use it or not. There are two approaches to encode such an information:

* **simplistic**: just a number which increases every time the REST API changes. So clients just need to know "I can use this feature since version 42".

* **advanced**: use semantic versioning, so clients also get the choice do set upper limits, e.g. work with "1.42.0 and upwards but only until 2.0.0". 

The remaining question is: how does the version get maintained. Here I'd like to second @maggu2810: maintaining it manually is cumbersome and error-prone and I would strongly advise against manually editing it! 

While with the simplistic number this might be doable. While in theory some automatic increase on every release could be be done, I don't think this is feasible because it would require the bundles to be "released" only  in the official build - if e.g. a solution provider does their own build (and even openHAB is doing so), this won't work. 

Semantic versioning on the other hand it is almost impossible to get right, considering that the REST API consists not only of the Resources, but also the DTOs and everything spread over several bundles. I would _only_ do that with some tooling in place, which _at least_ fails the build if the API changed and the version was not adapted correspondingly when compared to the previous release. 

One aspect that was neglected so far completely is the fact that the REST API is not made of one single bundle, but actually multiple ones. Some are even optional and may or may not be installed. And not all bundles necessarily need to be installed in the same version. How is the single semantic version supposed to reflect this?

To me, this whole things really sounds way too complicated to me when looking at the details. I would really prefer to keep it as simple as possible. How much granularity do the clients _really_ need? And how much can we compromise on that for the sake of simplicity?

kaikreuzer:
I would aim for a simple solution, no need to have any kind of (more) granularity - my goal is simply to fix the current feature that we have in place (and which does not work as expected).

Your "simplistic" solution should imho suffice as well (not offering the possibility for clients to find an upper bound they want to talk to, but that's imho not a must). But as you figured out, even the simplistic solution cannot be automated, but the version would need to be maintained manually. So from there, I do not see a big difference in effort to using a semantic version then.

maggu2810:
One question: How important is the "naming" / version of our "beta ref-tags" at all?

Would it be an option for you to force a special timestamp on the reftag creation instead of e.g. b1?
We could call the reftag tool with "201802080911" instead of "b1" (at least, I can try to change it) and state that there is a new beta-reftag-release called "201802080911".

We could create a Maven version number with a "b" prefix and drop that "b" on the bundle version.
So e.g.
* Maven version 0.10.0.b201802080911
* bundle version 0.10.0.201802080911

The Maven version use the "b", so the Maven Version comparator knows it is a "beta".
The bundle version is using (as every "non ESH release version") a timestamp part.

The only thing a consumer of the REST version needs to know is that "0.10.0.xxx" is lower then "0.10.0".

kaikreuzer:
@maggu2810 I guess you have missed that paragraph from @sjka above:

> With the simplistic number this might be doable. While in theory some automatic increase on every release could be be done, I don't think this is feasible because it would require the bundles to be "released" only in the official build - if e.g. a solution provider does their own build (and even openHAB is doing so), this won't work.

maggu2810:
No really missed, I assume I do not understand it

> I don't think this is feasible because it would require the bundles to be "released" only in the official build - if e.g. a solution provider does their own build (and even openHAB is doing so), this won't work.

I released a lot of inofficial bundle builds and don't see the problem.
Can you explain me why does it not work?

kaikreuzer:
Which version does the REST API show in your builds?

maggu2810:
> Which version does the REST API show in your builds?

It should not matter at all. It is using the version I set.
If someone different then ESH itself is doing an unofficial release it shouldn't make this release public so it conflicts with the official one. Or do we want that a "solution provider" could break other "solution providers" just by using an other / higher / wrong version scheme?

What perhaps need to be possible that the solution could switch from unofficial ones to the official ones without any breakage. So, the solution provider have to use something that fits to the "official version scheme". And if I as a solution provider would use above timestamped based version scheme it is about me to use timestamps for my releases that reflect the ESH release history.

kaikreuzer:
It is not about custom builds breaking/conflicting with official build ordering.
It is about UI compatibility across different solutions. If there is a ESH-compatible UI that uses the REST API, it simply needs to check the version of the API to decide how to communicate with it. If you ask solutions to switch to some completely different API versioning scheme, this would break this compatibility.
And as discussed above, the solution versions are merely marketing versions anyhow, while the version we expose in the REST API is meant to be a technical one.

kaikreuzer:
@maggu2810 & @sjka Any suggestions on how to proceed here? I honestly do not see any real alternative to this PR...

maggu2810:
For me it is still not clear what is wrong with https://github.com/eclipse/smarthome/pull/5032#issuecomment-364034469 (if technical possible).

> If you ask solutions to switch to some completely different API versioning scheme, this would break this compatibility.
> And as discussed above, the solution versions are merely marketing versions anyhow, while the version we expose in the REST API is meant to be a technical one.

I didn't get the point.

All semi-official releases (stuff that is build by the JIPP but not releases) will use an increasing bundle and so REST API version.

> The only thing a consumer of the REST version needs to know is that "0.10.0.xxx" is lower then "0.10.0".

Which "different API versioning scheme" that breaks compatibility do you refer to?
* It as the same as before just without "bX".
* A solution does not need to break compatibility, if it want to state which ESH API version is compatible it just need to ensure that the bundle version of its ESH REST bundle starts with the same 3-4 segments.

kaikreuzer:
> For me it is still not clear what is wrong with [...] (if technical possible).
> stuff that is build by the JIPP but not releases)

Exactly this is the problem: It only works for stuff that is built on JIPP. But many people are building ESH for their solutions on their own build servers. The "qualifier" is meant as a qualifier and should not be treated as a 4th segment of the version.

maggu2810:
I am using here an integer version only, but same would be true for multiple segments.

And how does a manually incremented version solve the problem with custom builds of solutions?

If ESH used a version 2 and a version 3 of the REST API and my solution uses "something between" because I only cherry picked some of the upstream changes?

What if the solution pick the changes that increments from 2 to 3 but not the one that 1 to 2?

kaikreuzer:
I am not talking of forking the project and making its set of functions a different one to the original project. That anyhow would never work and any version would be useless.

What I want to address is any build (be it on JIPP or somewhere else) of the official code base from the master branch at any point in time.

maggu2810:
> My suggestion is to start with 1.0 as this is bigger than any previous 0.x.0 version.

Why not keep at least the major and minor version in sync with ESH version? We have never used the third part (patch / revision / build number / ...) of the version different then zero. So the "third part" could be a candidate.

kaikreuzer:
> Why not keep at least the major and minor version in sync with ESH version? 

Because this would mean that the version is not following semver. If we say that we don't want to offer semver, we could also just use a number that we increment.

> We have never used the third part

... so far, but we might want to be able to release patch releases like 0.10.1, if we notice that we have a critical bug in a future 0.10.0 release that renders it almost useless. I would therefore not block us the way of doing patch releases.

maggu2810:
> I am not talking of forking the project
> What I want to address is any build (be it on JIPP or somewhere else) of the official code base from the master branch at any point in time.
> Because this would mean that the version is not following semver.

I didn't know that the intention has been also to start using the semantic version for the REST API.
Okay, perhaps can read this between the lines as you have written "1.0".

So, semver for the REST API, but not for the other parts (that are still not providing a stable API / using 0.x).

I am not happy with all that stuff (maintain the REST API version manually, using different API versions between the Java API and the REST one, the one needs to be following semver and should be stable (>= 1.0) the other one does not need to fulfill that requirement, ...). I could be wrong and its just my personal meaning / feeling. You are the PMC and I don't think we need to waste the time with that discussion anymore. If you think it makes the life easier for solution providers and the ESH project maintenance itself...

kaikreuzer:
> I didn't know that the intention has been also to start using the semantic version for the REST API.

You don't have to read between the lines, you just have to find [the right lines](https://github.com/eclipse/smarthome/pull/5032#issue-167164270):

_"We should increase the minor version whenever we add new features to the REST API and increase the major version, whenever we break some existing feature on it."_

Or [these](https://github.com/eclipse/smarthome/pull/5032#issuecomment-364033106):

_"I do not see a big difference in effort to using a semantic version then."_

> You are the PMC

... and as such, I do not have any more rights than you have, so I would like to come to an agreement that all core committers support or at least can live with.

I think the main underlying issue is that versioning is never fun, so I would also prefer not having to deal with that. But as a matter of fact, the current "feature" of having a version exposed in the REST API is broken and I do not see any other way to fix it than through this suggested manual approach.

Let's have @sjka have a say here as well and see whether he thinks we merge this or not.

sjka:
I would _love_ to have semantic versioning on all relevant, which mainly means Java packages and REST APIs. However, as I've written above, they are nearly impossible to be done correctly by hand. Considering how many classes have an effect on the REST APIs and how long some PRs are being discussed, this will end up in a mess sooner or later. We are not talking about the RESTResource implementations only, but also about all the DTOs, their mappers, the entities which are used to fill them, etc... So it MUST be automated. 

For Java APIs, unfortunately there is no viable option with our manifest-first approach. For that purpose it would be great to switch to bndtools, but that's another story and clearly out-of-scope here. 

For the REST API, I have not really seen a good tooling for that. There are some scripts flying around in the internet to at least verify API compatibility based on swagger JSONs. We are far away even being able to generate a swagger file at build time, not event thinking about deriving the semantic consequences from it. So I don't see how this should work. 

Nevertheless, I still think that the initial requirement, having to distinguish different snapshot/beta versions, reveals a bad smell already. This simply shouldn't be the case. And if we continue using our current versioning strategy of the bundles (i.e. the "marketing version"), then I don't see why UIs shouldn't be able to use it. Even the first two digits _should_ be sufficient. 

If that means we will need to release more often in order to enable downstream products to consume ESH properly, then IMHO we should rather discuss this instead.

sjka:
Just a clarification as we just had this discussion: As I have written in https://github.com/eclipse/smarthome/pull/5032#issuecomment-364030078: Maintaining a single (!) integer number might be doable IMHO. It's annoying, but better than nothing, I guess.

However, one thing must be clear: It MUST NOT carry any semantic meaning or make any promises about the REST API compatibility apart from making it comparable (which obviously is a given with natural numbers) and indicate _something_ in the REST API _might_ have changed. 

In that way you can increase the number "whenever you feel it makes sense". If that is what helps your use-case, I won't mind.

kaikreuzer:
Ok, so I have added https://github.com/eclipse/smarthome/pull/5032/commits/2a91af21a234b51d555fe016a8c15c56accc5308.
Is this something that everyone can live with?

maniac103:
From the perspective of the Android app, the integer number should work fine. One other idea for keeping the generation automatic while still being comparable would be using a timestamp:

```
public final String version = new SimpleDateFormat("YYYYMMddHHmm").format(new Date());
```

This has the obvious disadvantage of not working too well with point releases though...so your approach might work better.

kaikreuzer:
Hm, `version` in your example would always be the timestamp of the moment that your system started - it would be no way to tell, which build it actually is.

maniac103:
I knew my logic would be too simple :-/ The idea was inserting the compile/build timestamp; thinking about it one probably needs some maven script magic for that.

maggu2810:
> However, one thing must be clear: It MUST NOT carry any semantic meaning or make any promises about the REST API compatibility apart from making it comparable (which obviously is a given with natural numbers) and indicate something in the REST API might have changed.

> In that way you can increase the number "whenever you feel it makes sense". If that is what helps your use-case, I won't mind.

What about using "0.x" (and starting with "0.11" as it is bigger then all the previously used one) instead of x only.
As long as we are in a pre 1 we can then using the same mechanism ("increase whenever we feel it make sense", "not carry any semantic meaning", "only indicate something have changed").
But perhaps we keep the change open to use the same version as the bundles itself if we start using semver for the whole ESH.

Or do we really consider to use separate versions for packages or parts of ESH (e.g. binding API, core API, ...) some time?

kaikreuzer:
I guess IF we start semver one day this can be noticed by the fact that the version is then not a single integer anymore. I personally am fine with both, so I guess you have to settle it with @sjka...

sjka:
> Or do we really consider to use separate versions for packages or parts of ESH (e.g. binding API, core API, ...) some time?

For packages I see this a really valid option, considering this is a first-class notion in OSGi. 

The bundle-versions are and will be "marketing" versions anyway, I'm not sure if it will make sense to version them individually. As such, they won't carry any semantic meaning anyway.

I don't see any reasons or benefits of mixing these two aspects (bundle version and REST API "version"). If we think having the bundle version would be beneficial on client-side, then let's add it as a separate field.

PR #5036, opened 2018-02-06
BUG:  fix #5033

### LINKED ISSUE #5033: PaperUI for new Items name other properties are taken from channelType instead of channel
Linking a channel will use the properties from the channelType instead of the (maybe overwritten) properties from channel.

this is in file  extensions/ui/org.eclipse.smarthome.ui.paper/web-src/js/controllers/controller.things.view.js
line 79 definition of $scope.linkChannel.

aploese:
Signed-off-by: Arne Plöse <aploese@gmx.de>

### DIFF
--- a/extensions/ui/org.eclipse.smarthome.ui.paper/web-src/js/controllers/controller.things.view.js
+++ b/extensions/ui/org.eclipse.smarthome.ui.paper/web-src/js/controllers/controller.things.view.js
@@ -78,15 +78,14 @@ angular.module('PaperUI.controllers.things') //
 
     $scope.linkChannel = function(channelID, event, preSelect) {
         var channel = $scope.getChannelById(channelID);
-        var channelType = $scope.getChannelTypeByUID(channel.channelTypeUID);
 
         var params = {
             linkedItems : channel.linkedItems.length > 0 ? channel.linkedItems : '',
             acceptedItemTypes : channel.acceptedItemTypes,
-            category : channelType.category ? channelType.category : '',
-            suggestedName : getItemNameSuggestion(channelID, channelType.label),
-            suggestedLabel : channelType.label,
-            suggestedCategory : channelType.category ? channelType.category : '',
+            category : channel.category ? channel.category : '',
+            suggestedName : getItemNameSuggestion(channelID, channel.label),
+            suggestedLabel : channel.label,
+            suggestedCategory : channel.category ? channel.category : '',
             preSelectCreate : preSelect,
             allowNewItemCreation : $scope.advancedMode && channel.kind !== 'TRIGGER' // allow "Create new Item" in
             // advanced mode only, disable


extensions/ui/org.eclipse.smarthome.ui.paper/web-src/js/controllers/controller.things.view.js
```diff
@@ -78,15 +78,14 @@ angular.module('PaperUI.controllers.things') //
...
-        var channelType = $scope.getChannelTypeByUID(channel.channelTypeUID);
 
         var params = {
             linkedItems : channel.linkedItems.length > 0 ? channel.linkedItems : '',
             acceptedItemTypes : channel.acceptedItemTypes,
-            category : channelType.category ? channelType.category : '',
-            suggestedName : getItemNameSuggestion(channelID, channelType.label),
-            suggestedLabel : channelType.label,
-            suggestedCategory : channelType.category ? channelType.category : '',
+            category : channel.category ? channel.category : '',
```
sjka: Are you sure that the channel itself has a `category` property? Maybe the Paper UI is doing some magic somewhere, but to me it looks like neither [ChannelDTO](https://github.com/eclipse/smarthome/blob/4204ce06bb28c28e5f711e720f87ef83beff2e27/bundles/core/org.eclipse.smarthome.core.thing/src/main/java/org/eclipse/smarthome/core/thing/dto/ChannelDTO.java) nor [EnrichedChannelDTO](https://github.com/eclipse/smarthome/blob/38abcbb7d093b6fdf606d2f6d5600e8a3344aa9f/bundles/io/org.eclipse.smarthome.io.rest.core/src/main/java/org/eclipse/smarthome/io/rest/core/thing/EnrichedChannelDTO.java) have such a field. The channel entity itself in any case has no such field.

extensions/ui/org.eclipse.smarthome.ui.paper/web-src/js/controllers/controller.things.view.js
```diff
@@ -78,15 +78,14 @@ angular.module('PaperUI.controllers.things') //
...
+            suggestedName : getItemNameSuggestion(channelID, channel.label),
+            suggestedLabel : channel.label,
+            suggestedCategory : channel.category ? channel.category : '',
```
sjka: see above

aploese: Thats right, the category of an channel is governed by the channelType.
I can fix the fix - or makes it sense to specify the category at the channel too? - I'm not that deep into it.

extensions/ui/org.eclipse.smarthome.ui.paper/web-src/js/controllers/controller.things.view.js
htreu: omg, this simple change tricked me into being lazy. thanks for looking into this!

PR #5037, opened 2018-02-06
Magic bundle: Added thing that schedules a job to go ONLINE after a delay

triller-telekom:

### DIFF
--- a/bundles/test/org.eclipse.smarthome.magic/ESH-INF/thing/thing-types.xml
+++ b/bundles/test/org.eclipse.smarthome.magic/ESH-INF/thing/thing-types.xml
@@ -73,4 +73,12 @@
 		</config-description>
 	</thing-type>
 
+	<thing-type id="delayed-thing">
+        <label>Magic Delayed Online Thing</label>
+        <description>A thing that goes online after some time.</description>
+
+        <channels>
+            <channel id="switch" typeId="switch" />
+        </channels>
+    </thing-type>
 </thing:thing-descriptions>
--- a/bundles/test/org.eclipse.smarthome.magic/ESH-INF/thing/thing-types.xml
+++ b/bundles/test/org.eclipse.smarthome.magic/ESH-INF/thing/thing-types.xml
@@ -78,7 +78,7 @@
         <description>A thing that goes online after some time.</description>
 
         <channels>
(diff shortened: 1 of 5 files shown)


triller-telekom:
I have created this addition to the magic bundle in order to track down https://github.com/eclipse/smarthome/issues/4931 where we had to create the revert PR https://github.com/eclipse/smarthome/pull/4772 for the [desperately wanted feature](https://github.com/openhab/openhab2-addons/pull/2859#issuecomment-363222559) from PR: https://github.com/eclipse/smarthome/pull/4932

bundles/test/org.eclipse.smarthome.magic/ESH-INF/thing/thing-types.xml
```diff
@@ -73,4 +73,12 @@
 		</config-description>
 	</thing-type>
 
+	<thing-type id="delayed-thing">
+        <label>Magic Delayed Online Thing</label>
+        <description>A thing that goes online after some time.</description>
+
+        <channels>
+            <channel id="switch" typeId="switch" />
```
maggu2810: You could use a number and the implementation could use that number of seconds to delay the go online state update. A negative number could schedule a go offline state update. (Just an idea)

htreu: Yes, nice. And for a new command to this channel the Thing will go OFFLINE, wait the amount of seconds from the command (if positive) and go ONLINE again.

maggu2810: @htreu Okay for me, too. But my suggestion has been a different one :wink: 
My suggestion has been:
* Receiving a command does not change the state immediately.
* If the received number is > 0, then an online state update will be scheduled with a delay of number.
* If the received number is < 0, then an offline state update will be scheduled with a delay of -number.

So, in short:
* delay = Math.abs(number)
* state = number > 0 ? ONLINE : OFFLINE

triller-telekom: Good idea, I have changed the code to reflect that and as a bonus I can reproduce #4931 with that!

PR #5041, opened 2018-02-07
Channel fix

aploese:
Reverted to channelType.category.


maggu2810:
@aploese Thanks for the fix.
@htreu Thanks for the extensive testing.
:wink:

Issue #4440, opened 2017-10-18
Multiple / different GSON / Guava versions


martinvw:
After a [comment of @wborn](https://github.com/openhab/openhab2-addons/pull/2737#discussion_r142791801) I looked into my local configuration why I had errors which he did not.

Somehow his installation favored GSON 2.7 while mine used GSON 2.3.1.

> Looking at the line numbers of your [stacktrace](https://github.com/openhab/openhab2-
> addons/pull/2737#issuecomment-333316022) I think you were testing with openHAB 2.1.0?
> 
> openHAB 2.1.0 uses Gson 2.3.1 (2014-11-20) which only supports a very simple ISO 8601 format:
> 
> ```
>   private static DateFormat buildIso8601Format() {
>     DateFormat iso8601Format = new SimpleDateFormat("yyyy-MM-dd'T'HH:mm:ss'Z'", Locale.US);
>     iso8601Format.setTimeZone(TimeZone.getTimeZone("UTC"));
>     return iso8601Format;
>   }
> ```
> 
> So it will fail to parse the milliseconds part of the timestamps, e.g.:
> ```
>  "last_connection": "2017-09-30T11:45:55.599Z",
> ```
> 
> A better  ISO 8601 date type adapter was added in [Gson 2.4](https://github.com/google/gson/blob/master/CHANGELOG.md#version-24). 
> 
> I was testing with openHAB 2.2.0-SNAPSHOT which is using Gson 2.7.0 (2016-06-14) and also supports [more ISO 8601 timestamp variants](https://github.com/google/gson/blob/gson-parent-2.7/gson/src/main/java/com/google/gson/internal/bind/util/ISO8601Utils.java#L116).

After that I look into it and found out [this](https://github.com/openhab/openhab2-addons/pull/2737#discussion_r144098726):

 If I download the current openHAB snapshot and inflate that, it shows:

```
   ...
   creating: runtime/system/com/google/
   creating: runtime/system/com/google/guava/
   creating: runtime/system/com/google/guava/guava/
   creating: runtime/system/com/google/guava/guava/15.0/
   creating: runtime/system/com/google/guava/guava/18.0/
   creating: runtime/system/com/google/code/
   creating: runtime/system/com/google/code/findbugs/
   creating: runtime/system/com/google/code/findbugs/annotations/
   creating: runtime/system/com/google/code/findbugs/annotations/2.0.1/
   creating: runtime/system/com/google/code/gson/
[...]

PR #5043, opened 2018-02-07
Select only GSON 2.7 in SmartHome launch config

htreu:
* Both versions 2.4 and 2.7 of Google GSON where selected before.


maggu2810:
AFAIK we used 2.3.1 (see also https://github.com/eclipse/smarthome/issues/3680#issue-236529529)
Why do we want to use 2.7 in the launch configuration (used for testing) now?

kaikreuzer:
@maggu2810 See https://github.com/eclipse/smarthome/issues/4440. GSON 2.7 is required by the LSP feature, but only for this. So "plain" ESH is meant to be compatible with GSON 2.3.1, only solutions that have LSP installed will require 2.7.
We do not have an easy way to reflect that in the target platform - in order to allow LSP development&debugging, it is imho ok to have 2.7 in the IDE; the downside is that we won't notice if any other code accidentially starts using 2.7 features...

htreu:
Yes, @maggu2810 you are right.
First: 2.4 was my mistake, I was referring to Version 2.2.4 which is what our targetplatform defines right now, see https://github.com/eclipse/smarthome/blob/master/targetplatform/smarthome.target#L7

In the comment (https://github.com/eclipse/smarthome/issues/3680#issue-236529529) @kaikreuzer mentioned 2.3.1 but it seems that this version never made it into our targetplatform.
The 2.7.0 I selected in this PR is a transient dependency for LSP support, namely the `org.eclipse.lsp4j.jsonrpc` bundle. It requires GSON to be in a range [2.7.0,3.0.0).

How should we resolve this issue?
/cc @sjka

htreu:
FTR: the original issue I'm trying to fix is my IDE complaining about the version constraint for GSON is not matched when LSP support comes up. Sometimes only GSON 2.2.4 is available at that time.

maggu2810:
For me it would be okay bump the GSON requirements 2.7 for the whole ESH or just state that we officially don't support older versions anymore.
I don't like to keep compatibility for version x in the one part of the bundles but version y in the other ones.
If GSON is used by ESH bundles one version should be choosen.
Isn't it what we do for e.g. Jetty already? As we introduced a function that has been present in a more recent version, we used the more recent version for the whole third party dependency.
Sure, there could be a difference between "ESH core" and "ESH extensions" but in the long run, one requirement would be simpler to maintain and handle then different requirements for different parts of ESH.

kaikreuzer:
In general I agree, but please note that we had a lot of special stuff in the target platform for the designer in the past and no ESH bundle was supposed to make use of them.
In some way the LSP feature is the successor of the designer and it is probably used by only very few solutions that are built on ESH. Furthermore, it merely is a transitive dependency of those LSP libs that won't be required if you don't deploy LSP.

> For me it would be okay bump the GSON requirements to 2.7 for the whole ESH or just state that we officially don't support older versions anymore.

As always: If our code does not have any reason for this as it works with 2.2.4, we should be conservative and not require people to upgrade.

htreu:
My suggestion would be to select only 2.2.4 in the launch config and disable the LSP feature. In case you depend on the LSP feature during development you have to maintain a local launch config with the required dependencies.

maggu2810:
> As always: If our code does not have any reason for this as it works with 2.2.4, we should be conservative and not require people to upgrade.

Our (ESH) code includes the LSP feature (the successor of the designer) that is perhaps used by very few solutions and all the other one.
So, our code (the LSP one) does have any reasons for. In this case (let's repeat it) it is only one bundle of the dozen ones and this one bundle is mostly not used.
But I still think it is similar to the Jetty bump that has been done because (IIRC so correct my if I am wrong) one place -- the proxy servlet in the esh.ui bundle -- needs it (which is also optional for at least a few products I know).

At least I don't prefer any GSON version, I just don't like to use different ones in the ESH code base.
Regardless if there are solutions that don't need LSP or there are solutions that don't use the proxy servlet and are affected by the Jetty bump.

htreu:
I will change this PR to do the opposite of the original proposal: Select only the 2.2.4 version of Google GSON and also deselect the LSP bundle for now. My intent was to "fix" the startup behaviour in the IDE.
Lats take the GSON version bump discussion to a new issue.

htreu:
In fact, the ESH LSP bundle was not selected but only the org.eclipse.lsp* bundles.

htreu:
Moved discussion about supported versions here: #5077.

PR #5044, opened 2018-02-07
return NULL if PathNotFoundException is thrown

jschanz:
reverts a feature from f90852c
discussion also see #4376

works like described in the manual
https://docs.openhab.org/v2.2/addons/transformations/jsonpath/readme.html#example
and it's necessary to parse json which doesn't contain the path
every time you query the json, otherwise the logfile is filled up with the json
stuff
for example, see https://www.dwd.de/DWD/warnungen/warnapp/json/warnings.json

### DIFF
--- a/extensions/transform/org.eclipse.smarthome.transform.jsonpath/src/main/java/org/eclipse/smarthome/transform/jsonpath/internal/JSonPathTransformationService.java
+++ b/extensions/transform/org.eclipse.smarthome.transform.jsonpath/src/main/java/org/eclipse/smarthome/transform/jsonpath/internal/JSonPathTransformationService.java
@@ -65,7 +65,7 @@ public class JSonPathTransformationService implements TransformationService {
                 return transformationResult.toString();
             }
         } catch (PathNotFoundException e) {
-            throw new TransformationException("Invalid path '" + jsonPathExpression + "' in '" + source + "'");
+		return null;
         } catch (InvalidPathException | InvalidJsonException e) {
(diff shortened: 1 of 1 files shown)


triller-telekom:
Thank you for your first contribution.

However, before you can add code to Eclipse SmartHome you have to create an account at eclipse.org and sign the [ECA](https://www.eclipse.org/legal/ECA.php). This is to ensure that all code in ESH complies with the EPL.

You also have to put a blank line followed by a "Signed-off-by" in your commit message containign the same email address as your eclipse.org account that has signed the ECA.

```
Signed-off-by: FirstName LastName <your@email.tld>
```

Also you have used tabs instead of spaces, please use the Eclipse auto-formatter to format the code:
```
[ERROR] org.eclipse.smarthome.transform.jsonpath.internal.JSonPathTransformationService.java:[68]
File contains tab characters (this is the first instance).
```

triller-telekom:
@jschanz My comments were not targeted on keeping you away from the project. I am sorry if it sounded like that. The problem is that ESH is an Eclipse project and all its code has to be checked in terms of intellectual property rights because the code can be used in commercial products which do not have to worry about that people claim their rights on the code and sue them.

Thus, I am not sure whether you closed this issue because of that or if it is not valid (anymore)?

jschanz:
No problem ... I've done the developer registration process and had created a new pull request. It's a bit uncommon to contribute to the openhab / eclipse smarthome project, if you don't know how :-)

triller-telekom:
Alright. Just linking this to the new PR #5049.

PR #5045, opened 2018-02-07
Reactivate WrappedScheduledExecutorService that logs Exceptions of scheduled tasks

triller-telekom:


triller-telekom:
Thanks to @afuechsel and his https://github.com/eclipse/smarthome/issues/4931#issuecomment-357885341

PR #5047, opened 2018-02-07
ChannelItemprovider: Shutdown executor in deactivate

triller-telekom:


maggu2810:
What about this code that is executed if the component is still enabled but the configuration state it should not act on:

```Java
            for (ProviderChangeListener<Item> listener : listeners) {
                for (Item item : getAll()) {
                    listener.removed(this, item);
                }
            }
```

Would it make sense to move the deactivation code to a separate function that is called on "active but disabled (so one of the two modified branches)" and on "deactivate (component)"?

triller-telekom:
Close/reopen because of paperUI build failure which is independent of this change

maggu2810:
@velichkos Would you like to have a look at? Okay for you?

velichkos:
I like the disable logic in a separate method.
Looks good to me.

PR #5055, opened 2018-02-08
PersistenceModelManager: Make initialisation code paths consistent

mkhl:
NOTE: This PR includes and depends on the changes in PR #5054.

When PersistenceModelManager#activate is called, the ModelRepository is
present, but we cannot be sure whether it has loaded the persistence models
yet.

a) If it has not, PersistenceModelManager registers itself as a change
listener, then when a model is loaded ModelRepository calls
PersistenceModelManager#modelChanged which extracts its configuration and
calls PersistenceManager#addConfig with that, which in turn calls
PersistenceManager#startEventHandling.

b) If it has, PersistenceModelManager registers itself as a change
listener, but since the models are already loaded #modelChanged does not
get called.  Instead PersistenceModelManager loops over the models and
calls PersistenceManager#startEventHandling for each, but it never extracts
any configuration from the models, and since PersistenceManager is now
missing the configuration it does not actually set up any event handling
jobs.

In short, case b) is broken.

This change extracts methods for the behaviour that _should_ occur, namely
adding or removing the extracted configuration of each model to/from the
PersistenceManager (and _not_ calling
itself in #addConfig/#removeConfig), and calls these on both initialisation
and deinitialisation paths, ensuring that PersistenceManager knows the
configuration of each configured PersistenceService so it can actually
handle events.

### DIFF
--- a/bundles/model/org.eclipse.smarthome.model.persistence/src/org/eclipse/smarthome/model/persistence/internal/PersistenceModelManager.java
+++ b/bundles/model/org.eclipse.smarthome.model.persistence/src/org/eclipse/smarthome/model/persistence/internal/PersistenceModelManager.java
@@ -56,21 +56,19 @@ public class PersistenceModelManager implements ModelRepositoryChangeListener {
     public PersistenceModelManager() {
     }
 
-    protected void activate() {
-        modelRepository.addModelRepositoryChangeListener(this);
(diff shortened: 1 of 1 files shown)


maggu2810:
The build of the relevant parts has been succeeded. It failed on "Basic UI" because of an Node.js error.

PR #5056, opened 2018-02-08
Upgrade to Mockito 2

### LINKED ISSUE #5004: Mockito 2.13
As just seen in  #3918, openhab/org.openhab.binding.zwave#812 and others (and potentially in your IDE), it's been a struggle and required some luck to get plain unit tests running within the IDE. This whole mess can be solved with upgrading to mockito 2.x.

I will create CQs for it and its transitive dependencies:

* Mockito 2.13.0
* Byte-Buddy 1.7.9
* Byte-Buddy Agent 1.7.9
* Objenesis 2.6

(They'll be "Type B, full-review", so we can redistribute them via our own P2 update site)

sjka:
This now as a follow-up to #5050 integrates Mockito 2.13 (& friends) into the target platform.

Also I have adapted all the manifests of the test fragments and updated the launch configs.

fixes #5004


htreu:
Travis complains:
```
testApproveNormalization(org.eclipse.smarthome.config.discovery.internal.PersistentInboxTest)  Time elapsed: 0.139 sec  <<< ERROR!
	at org.eclipse.smarthome.config.discovery.internal.PersistentInboxTest.testApproveNormalization(PersistentInboxTest.java:147)
```

sjka:
What a whimperer... 😉 
I'll have a look.

sjka:
Apart from a mistake which I did in a manifest, there seems to be an interesting change in [ArgumentMatchers.any(Class)](https://static.javadoc.io/org.mockito/mockito-core/2.7.9/org/mockito/ArgumentMatchers.html#any(java.lang.Class)):

>* Since Mockito 2.1.0 this method will perform a type check thus null values are not authorized.
>* Since mockito 2.1.0 any() and anyObject() are not anymore aliases of this method.

It didn't hit us badly and IMHO it is for the better, but it definitely is worth knowing. Nevertheless, for this I actually had to touch a few lines of test-code.

bundles/core/org.eclipse.smarthome.core.thing.test/src/test/java/org/eclipse/smarthome/core/thing/firmware/FirmwareUpdateServiceTest.java
```diff
@@ -149,13 +149,14 @@ public void setup() {
...
-        when(mockProvider.getFirmware(eq(FW009_EN.getUID()), Matchers.any(Locale.class))).thenReturn(FW009_EN);
-        when(mockProvider.getFirmware(eq(FW111_EN.getUID()), Matchers.any(Locale.class))).thenReturn(FW111_EN);
-        when(mockProvider.getFirmware(eq(FW112_EN.getUID()), Matchers.any(Locale.class))).thenReturn(FW112_EN);
-        when(mockProvider.getFirmware(eq(FWALPHA_EN.getUID()), Matchers.any(Locale.class))).thenReturn(FWALPHA_EN);
-        when(mockProvider.getFirmwares(any(ThingTypeUID.class), any(Locale.class))).then(invocation -> {
+        when(mockProvider.getFirmware(eq(FW009_EN.getUID()), ArgumentMatchers.any())).thenReturn(FW009_EN);
+        when(mockProvider.getFirmware(eq(FW111_EN.getUID()), ArgumentMatchers.any())).thenReturn(FW111_EN);
+        when(mockProvider.getFirmware(eq(FW112_EN.getUID()), ArgumentMatchers.any())).thenReturn(FW112_EN);
+        when(mockProvider.getFirmware(eq(FWALPHA_EN.getUID()), ArgumentMatchers.any())).thenReturn(FWALPHA_EN);
+        when(mockProvider.getFirmwares(any(ThingTypeUID.class), any())).then(invocation -> {
```
htreu: Here we have a strange mix of ArgumentMatchers.any() and org.hamcrest.CoreMatchers.any(). Can this be reduced to one (preferably static) import?

bundles/core/org.eclipse.smarthome.core.test/META-INF/MANIFEST.MF
```diff
@@ -19,12 +19,10 @@ Import-Package:
  org.eclipse.smarthome.test.java,
  org.hamcrest;core=split,
  org.junit,
+ org.junit.rules;version="4.7.0",
```
htreu: A lot of other org.junit.rules imports are versioned with "4.0.0". Is this here 4.7.0 by accident?

targetplatform/smarthome.target
```diff
@@ -152,6 +148,16 @@
 <unit id="californium-osgi" version="0.0.0"/>
 <unit id="org.eclipse.californium.element-connector" version="0.0.0"/>
 <unit id="org.eclipse.californium.scandium" version="0.0.0"/>
+<unit id="osgi.annotation" version="0.0.0"/>
```
htreu: how is osgi.annotation related to mockito 2.0?

bundles/core/org.eclipse.smarthome.core.test/META-INF/MANIFEST.MF
sjka: No, this is by intention: see https://junit.org/junit4/javadoc/4.12/org/junit/rules/TestName.html (it was added with 4.7)

targetplatform/smarthome.target
sjka: Not at all - it was a follow-up on #4737 which was well overdue, which accidentally slipped into this branch here. Should have been done anyway, but you are right, it was intended for another PR.

bundles/core/org.eclipse.smarthome.core.thing.test/src/test/java/org/eclipse/smarthome/core/thing/firmware/FirmwareUpdateServiceTest.java
sjka: Probably yes. Let me check what the IDE did there and why it didn't do so consistently.

sjka:
> Did you check the archetype generated test bundle? It may also still have the require-bundle lines.

Yes, that's included: https://github.com/eclipse/smarthome/pull/5056/files#diff-e76929c22c8dd021339cb1c1153d30ea

> And we have a special sentence in the documentation for the allowed require-bundle lines which needs to be removed, too

Thanks, I missed that one. Stay tuned.

bundles/core/org.eclipse.smarthome.core.thing.test/src/test/java/org/eclipse/smarthome/core/thing/firmware/FirmwareUpdateServiceTest.java
sjka: Some more insights on that: `any()` _always_  comes from `ArgumentMatchers` and indeed, could be used as a static import. Hamcrest only has `any(Class)` - which Mockito also has and there static imports cannot be used (at least not if both are statically imported) because they're ambiguous. In any case, Hamcrest and Mockito matchers are not interchangeable.

I'll clean that up in the follow-up PR though.

sjka: PS: Correction - it works with `any(Class)` because they have different return types, but doesn't work with `isA(Class)`.

maggu2810:
Sorry for merging too early.

PR #5058, opened 2018-02-08
Implemented Startup Servlet that shows links to other Web UIs

### LINKED ISSUE #5052: Improve demo instance in IDE setup
Currently, the instance that is started up from within the ESH IDE is pretty minimal - it has hardly any items and UI elements and [it is hard for newbies to actually find the urls of the UIs](https://www.eclipse.org/forums/index.php?t=rview&goto=1781339).

As we have introduced the Magic Bundle for doing all kinds of functionality testing, I would first of all suggest to include this in the launch configuration and also to add some textual Thing definitions that include such "magic devices" in the demo setup.

Furthermore, the Magic Bundle could register a servlet that is called when http://localhost:8080/ is accessed and which provides links to the different UIs and the REST API.

triller-telekom:
Fixes #5052


bundles/test/org.eclipse.smarthome.magic/src/main/java/org/eclipse/smarthome/magic/binding/internal/servlet/MagicServlet.java
```diff
@@ -0,0 +1,128 @@
...
+ * @author Stefan Triller - initial contribution
+ *
+ */
+@Component(immediate = true)
+public class MagicServlet implements Servlet {
+
+    protected HttpService httpService;
+
+    /** the root path of this web application */
+    public static final String WEBAPP_ALIAS = "/";
```
maggu2810: Shouldn't static member be above non static ones?

Any reason to make the member public, couldn't it be private?

bundles/test/org.eclipse.smarthome.magic/src/main/java/org/eclipse/smarthome/magic/binding/internal/servlet/MagicServlet.java
```diff
@@ -0,0 +1,128 @@
...
+
+    private final Logger logger = LoggerFactory.getLogger(MagicServlet.class);
+
+    @Activate
+    protected void activate(Map<String, Object> configProps, BundleContext bundleContext) {
+        try {
+            httpService.registerResources(WEBAPP_ALIAS, "web", new MagicHttpContext(bundleContext.getBundle()));
+            logger.info("Started Magic UI at " + WEBAPP_ALIAS);
```
maggu2810: What's the reason for this log message (same error below and info in the deactivate)?
Shouldn't a error in activate already be logged by the caller logging mechanism?
Same for activate and deactivate in general.

bundles/test/org.eclipse.smarthome.magic/src/main/java/org/eclipse/smarthome/magic/binding/internal/servlet/MagicServlet.java
```diff
@@ -0,0 +1,128 @@
...
+
+    /** the root path of this web application */
+    public static final String WEBAPP_ALIAS = "/";
+
+    private final Logger logger = LoggerFactory.getLogger(MagicServlet.class);
+
+    @Activate
+    protected void activate(Map<String, Object> configProps, BundleContext bundleContext) {
+        try {
+            httpService.registerResources(WEBAPP_ALIAS, "web", new MagicHttpContext(bundleContext.getBundle()));
```
maggu2810: Do we need a custom HTTP context for the resources of the magic bundle, too?
IIRC it has been used because of a buggy Eclipse implementation if fragments are used?
Is the fragment support here important?

bundles/test/org.eclipse.smarthome.magic/src/main/java/org/eclipse/smarthome/magic/binding/internal/servlet/MagicServlet.java
```diff
@@ -0,0 +1,128 @@
...
+import org.slf4j.LoggerFactory;
+
+/**
+ * Registers itself under "/" and provides links to other web UIs
+ *
+ * @author Stefan Triller - initial contribution
+ *
+ */
+@Component(immediate = true)
+public class MagicServlet implements Servlet {
```
kaikreuzer: Why is this a servlet if all you do is registering a static resource?

bundles/test/org.eclipse.smarthome.magic/src/main/java/org/eclipse/smarthome/magic/binding/internal/servlet/MagicServlet.java
kaikreuzer: Is "web" a good url? Sounds pretty generic.

bundles/test/org.eclipse.smarthome.magic/web/index.html
```diff
@@ -0,0 +1,23 @@
+<html>
```
kaikreuzer: Could we add some love to that page?
This now looks like a 90ies webpage prototype in Netscape, this is not necessarily a good way to make newbies fall in love with the project...

bundles/test/org.eclipse.smarthome.magic/web/index.html
```diff
@@ -0,0 +1,23 @@
...
+<head>
+    <title>Eclipse SmartHome</title>
+</head>
+<body>
+    <h1>Welcome to Eclipse SmartHome</h1>
+	<p>
+	   Here are a few useful links to start exploring your local instance:
+	</p>
+	<p>
+	   <a href="basicui/app">BasicUI</a>
```
kaikreuzer: It is "Basic UI", "Classic UI" and "Paper UI".

distribution/smarthome/conf/things/demo.things
```diff
@@ -1,3 +1,6 @@
 yahooweather:weather:berlin [ location=638242 ]
 yahooweather:weather:losangeles "Los Angeles" [ location=2442047, refresh=120 ]
 
+magic:color-light:magicColor
```
kaikreuzer: Maybe worth to add some other devices than lights, like e.g. thermostats as well?
Maybe also assign them to a common location, add some labels and custom configurations?

bundles/test/org.eclipse.smarthome.magic/src/main/java/org/eclipse/smarthome/magic/binding/internal/servlet/MagicServlet.java
triller-telekom: I am logging that the interface has started like in the basic ui or classic ui servlet.

Regarding the error logging: I have to catch `NamespaceException` thrown by `httpService.registerRessources`.

bundles/test/org.eclipse.smarthome.magic/src/main/java/org/eclipse/smarthome/magic/binding/internal/servlet/MagicServlet.java
triller-telekom: I am registering everything under "/" now and jetty picks the context which has the closest match, i.e. it trys the full path and only falls back to "/" if no other servlet was responsible.

triller-telekom: The custom context is for returning the index page in case the user visits "/" without anything behind, so the index.html is shown to him.

bundles/test/org.eclipse.smarthome.magic/web/index.html
triller-telekom: Done

distribution/smarthome/conf/things/demo.things
triller-telekom: The magic bundle will get thermostats with PR #4818, afterwards this can be added to the demo setup. Also if one adds new stuff to the magic bundle we can extend this demo.

triller-telekom:
I have removed the logging and also created smaller images.

kaikreuzer:
The CSS files are pretty big and seem to contain a lot of unused stuff - any chance to reduce them to what is needed?

> also created smaller images.

At least the Paper UI image still has a width of 1000px. I think https://www.eclipse.org/smarthome/index.html#references proves that 400px width is more than enough (and imho it can be even much smaller for this index page).

triller-telekom:
I now have both css files in there, the original `materialize.css` and a `custom.css` containing only the stuff that is needed for the `index.html` file. I am unsure about the licence situation on what is allowed and what not. I am happy to change it to the smaller version or keep the original one, whichever fits in regards to legal terms.

triller-telekom:
This is how it looks now:

image

bundles/test/org.eclipse.smarthome.magic/src/main/java/org/eclipse/smarthome/magic/binding/internal/http/MagicHttpRessource.java
```diff
@@ -0,0 +1,95 @@
...
+@Component(immediate = true)
+public class MagicHttpRessource {
+    /** the root path of this web application */
+    private static final String WEBAPP_ALIAS = "/";
+
+    protected HttpService httpService;
+
+    @Activate
+    protected void activate(Map<String, Object> configProps, BundleContext bundleContext) throws NamespaceException {
+        httpService.registerResources(WEBAPP_ALIAS, WEBAPP_ALIAS, new MagicHttpContext(bundleContext.getBundle()));
```
maggu2810: Isn't it a little bit confusing to use the variable WEBAPP_ALIAS for the name (so, the resource prefix)?

bundles/test/org.eclipse.smarthome.magic/src/main/java/org/eclipse/smarthome/magic/binding/internal/http/MagicHttpRessource.java
```diff
@@ -0,0 +1,95 @@
...
+import org.osgi.service.http.NamespaceException;
+
+/**
+ * Registers itself under "/" and provides links to other web UIs
+ *
+ * @author Stefan Triller - initial contribution
+ *
+ */
+@Component(immediate = true)
+public class MagicHttpRessource {
```
kaikreuzer: typo: Resource

triller-telekom:
Alright, fixed.

PR #5061, opened 2018-02-09
Basic UI: fix live update (Microsoft Edge)

### LINKED ISSUE #5028: Basic UI with Microsoft Edge: page content hidden after an update event is received
Testing OH snapshot 1203, I discovered that each time an event update is received, the page content is make unvisible in Basiuc UI when using Microsoft Edge. F5 allows restoring the page.
It is certainly a consequence of my change in PR #4857.
Strange, it was working in my Eclipse dev environment.
I will provide a fix very soon as it is a major issue for any user using Microsoft Edge as browser.

### LINKED ISSUE #5029: Basic UI with Microsoft Edge: widgets not updating in Group Demo page
Using the demo sitemap with Basic UI and Microsoft Edge, I noticed that the page "Group Demo" is not refreshed.

lolodomo:
Fixes #5028
Fixes #5029

### DIFF
--- a/extensions/ui/org.eclipse.smarthome.ui.basic/web-src/smarthome.js
+++ b/extensions/ui/org.eclipse.smarthome.ui.basic/web-src/smarthome.js
@@ -407,6 +407,8 @@
 		};
 
 		_t.setValue = function() {};
+		_t.setLabelColor = function() {};
+		_t.setValueColor = function() {};
 		_t.suppressUpdate = function() {};
 	}
 
@@ -1589,7 +1591,9 @@
 			smarthome.dataModel = {};
 
 			function appendControl(control) {
-				smarthome.dataModel[control.id] = control;
+				if (control.id.length > 0) {
+					smarthome.dataModel[control.id] = control;
+				}
 			}
 
 			[].forEach.call(document.querySelectorAll(o.formControls), function(e) {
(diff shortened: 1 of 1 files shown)


lolodomo:
@resetnow @kaikreuzer : a quick review and merge would be appreciated as the current live update using Microsoft Edge is currentlmy "broken".

extensions/ui/org.eclipse.smarthome.ui.basic/web-src/smarthome.js
```diff
@@ -1589,7 +1591,9 @@
 			smarthome.dataModel = {};
 
 			function appendControl(control) {
-				smarthome.dataModel[control.id] = control;
+				if (control.id.length > 0) {
```
resetnow: Any reason to perform this check? Can widget ID ever be empty?

I can test it later today, as for now I only have this question.

lolodomo: Yes it happens when a frame is created when not defined in the sitemap, like for example in the "Astronomical Data" or 'Group Demo' pages in the demo sitemap.

resetnow:
Works for me.

Kind of my fault — I noticed this in a previous PR, but wasn't sure if it's due to my environment setup or it's actually a problem in the code, and since it takes forever to setup everything from ground up I didn't test it properly.

PR #5063, opened 2018-02-09
Remove interface inference

sjka:
As seen e.g. in https://github.com/openhab/org.openhab.binding.zigbee/issues/47#issuecomment-364414903, the `SafeCaller.create(Class)` method which infers the target interface has proven itself to be quite dangerous. Especially with target objects which are not in the framework's control (which is the main use-case for the SafeCaller), not all interfaces might be reachable from the current class loader. Hence I think it would be best to completely drop it. It was convenient, but not such a great improvement to start catching such exceptions and jumping through further hoops to make it possible.


sjka:
Thanks @maggu2810 - I though just realized that I had the types the wrong way round, therefore I had to push another commit. 

However, this doesn't really work nicely now with typed interfaces - see https://github.com/eclipse/smarthome/pull/5063/commits/3d2a6984ec0f0382fbf9080a51359285ede2886f#diff-c0c87bcca219c6210ea787bcc45cf0a1R360. Do you maybe have an idea how this could be done nicely?

PR #5067, opened 2018-02-10
Chart service: settings available in Paper UI

lolodomo:

### DIFF
--- a/bundles/ui/org.eclipse.smarthome.ui/ESH-INF/config/config.xml
+++ b/bundles/ui/org.eclipse.smarthome.ui/ESH-INF/config/config.xml
@@ -5,11 +5,6 @@
         http://eclipse.org/smarthome/schemas/config-description-1.0.0.xsd">
 
 	<config-description uri="ui:chart">
-		<parameter name="provider" type="text" required="false">
-			<label>Provider</label>
-			<description>Defines the chart provider to use.</description>
-			<default>default</default>
-		</parameter>
 		<parameter name="defaultHeight" type="text" required="false">
 			<label>Default Height</label>
 			<description>Defines the default height in pixels for the chart to build.</description>
@@ -24,10 +19,12 @@
 			<label>Scale</label>
 			<description>Defines the scale to apply to the requested chart height (0.5 to divide by 2 for example).</description>
 			<default>1</default>
+			<advanced>true</advanced>
 		</parameter>
 		<parameter name="maxWidth" type="text" required="false">
 			<label>Maximum Width</label>
 			<description>Defines the maximum width in pixels for the chart to build.</description>
+			<advanced>true</advanced>
 		</parameter>
 	</config-description>
 
--- a/bundles/ui/org.eclipse.smarthome.ui/ESH-INF/config/config.xml
+++ b/bundles/ui/org.eclipse.smarthome.ui/ESH-INF/config/config.xml
@@ -1,0 +1,34 @@
+<?xml version="1.0" encoding="UTF-8"?>
+<config-description:config-descriptions xmlns:xsi="http://www.w3.org/2001/XMLSchema-instance"
+	xmlns:config-description="http://eclipse.org/smarthome/schemas/config-description/v1.0.0"
+	xsi:schemaLocation="http://eclipse.org/smarthome/schemas/config-description/v1.0.0
+        http://eclipse.org/smarthome/schemas/config-description-1.0.0.xsd">
+
+	<config-description uri="ui:chart">
+		<parameter name="provider" type="text" required="false">
+			<label>Provider</label>
+			<description>Defines the chart provider to use.</description>
+			<default>default</default>
+		</parameter>
+		<parameter name="defaultHeight" type="text" required="false">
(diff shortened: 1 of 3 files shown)


lolodomo:
Take a look what we already have for Basic UI and Classic UI: "Defines ...".
But I agree that it is not the same "pattern" everywhere.

bundles/ui/org.eclipse.smarthome.ui/ESH-INF/config/config.xml
```diff
@@ -0,0 +1,34 @@
+<?xml version="1.0" encoding="UTF-8"?>
+<config-description:config-descriptions xmlns:xsi="http://www.w3.org/2001/XMLSchema-instance"
+	xmlns:config-description="http://eclipse.org/smarthome/schemas/config-description/v1.0.0"
+	xsi:schemaLocation="http://eclipse.org/smarthome/schemas/config-description/v1.0.0
+        http://eclipse.org/smarthome/schemas/config-description-1.0.0.xsd">
+
+	<config-description uri="ui:chart">
+		<parameter name="provider" type="text" required="false">
+			<label>Provider</label>
+			<description>Defines the chart provider to use.</description>
```
kaikreuzer: I wonder if we should have that displayed in the UI at all. There currently usually is only a single provider anyhow and it is rather a choice that is done by the solution and not so much by the user. We also do not have any mechanism to display available options to the user, so I would vote to leave that out (or at least to mark it advanced).

bundles/ui/org.eclipse.smarthome.ui/ESH-INF/config/config.xml
```diff
@@ -0,0 +1,34 @@
...
+			<default>240</default>
+		</parameter>
+		<parameter name="defaultWidth" type="text" required="false">
+			<label>Default Width</label>
+			<description>Defines the default width in pixels for the chart to build.</description>
+			<default>480</default>
+		</parameter>
+		<parameter name="scale" type="text" required="false">
+			<label>Scale</label>
+			<description>Defines the scale to apply to the requested chart height (0.5 to divide by 2 for example).</description>
```
kaikreuzer: Rather an advanced option, isn't it?

bundles/ui/org.eclipse.smarthome.ui/ESH-INF/config/config.xml
```diff
@@ -0,0 +1,34 @@
...
+			<default>1</default>
+		</parameter>
+		<parameter name="maxWidth" type="text" required="false">
```
kaikreuzer: Rather an advanced option, isn't it?

bundles/ui/org.eclipse.smarthome.ui/OSGI-INF/chartservlet.xml
```diff
@@ -17,6 +17,9 @@
    <implementation class="org.eclipse.smarthome.ui.internal.chart.ChartServlet"/>
    
    <property name="service.pid" type="String" value="org.eclipse.smarthome.chart"/>
+   <property name="service.config.description.uri" type="String" value="ui:chart"/>
+   <property name="service.config.label" type="String" value="Chart UI"/>
```
kaikreuzer: Hm, there isn't really such a thing as a "Chart UI" (compare to "Basic UI" etc.).

bundles/ui/org.eclipse.smarthome.ui/OSGI-INF/chartservlet.xml
```diff
@@ -17,6 +17,9 @@
...
+   <property name="service.config.category" type="String" value="ui"/>
```
kaikreuzer: The "ui" category is meant for UI extensions/addons. This here should imho rather be a "system" setting.

bundles/ui/org.eclipse.smarthome.ui/ESH-INF/config/config.xml
lolodomo: Ok

bundles/ui/org.eclipse.smarthome.ui/ESH-INF/config/config.xml
lolodomo: Ok

bundles/ui/org.eclipse.smarthome.ui/OSGI-INF/chartservlet.xml
lolodomo: "Chart" only ?

bundles/ui/org.eclipse.smarthome.ui/OSGI-INF/chartservlet.xml
lolodomo: Ok

bundles/ui/org.eclipse.smarthome.ui/ESH-INF/config/config.xml
lolodomo: Ok I will suppress it.

PR #5069, opened 2018-02-11
Sitemap REST API: include Location items in group members for Mapview…

### LINKED ISSUE #4989: Sitemap should include group members for location widgets
The location widget currently supports only a single location by taking a Location item name as its item parameter.
Technically, it is already possible to hand in a Group item name instead.
Using a Group might be a neat use case to show several locations on the same map - I'd therefore suggest that the sitemap should directly provide the group members (maybe filtered for only Location items?) in the response, so that UIs do not need to do the resolution by themselves.

lolodomo:
… widget

Fixes #4989


lolodomo:
Please note the build is failing in DMX binding tests, something totally independent of my PR.

lolodomo:
Once merged, I assume UIs will need updates to support group item in mapview widget.

bundles/io/org.eclipse.smarthome.io.rest.core/src/main/java/org/eclipse/smarthome/io/rest/core/item/EnrichedItemDTOMapper.java
```diff
@@ -42,12 +42,13 @@
      * @param uri the uri
      * @return item DTO object
      */
-    public static EnrichedItemDTO map(Item item, boolean drillDown, URI uri, Locale locale) {
+    public static EnrichedItemDTO map(Item item, boolean drillDown, String itemTypeFilter, URI uri, Locale locale) {
```
kaikreuzer: We would have to mark this PR "API Breaking", right?
And if we break the API, please let us consider making it more powerful and not just adding a very specific use case (it e.g. already won't cover a situation where you would want to filter for 2 types).

Wdyt about introducing a Predicate instead, which could be used for filtering?

bundles/io/org.eclipse.smarthome.io.rest.core/src/main/java/org/eclipse/smarthome/io/rest/core/item/EnrichedItemDTOMapper.java
```diff
@@ -64,7 +65,9 @@ private static EnrichedItemDTO map(Item item, ItemDTO itemDTO, URI uri, boolean
             if (drillDown) {
                 Collection<EnrichedItemDTO> members = new LinkedHashSet<>();
                 for (Item member : groupItem.getMembers()) {
-                    members.add(map(member, drillDown, uri, locale));
+                    if (itemTypeFilter == null || itemTypeFilter.equals(member.getType())) {
```
kaikreuzer: If there is an itemTypeFilter, you will never drill down any further now, is this expected?

bundles/io/org.eclipse.smarthome.io.rest.sitemap/src/main/java/org/eclipse/smarthome/io/rest/sitemap/internal/PageChangeListener.java
```diff
@@ -199,7 +200,10 @@ public void stateUpdated(Item item, State state) {
                 event.widgetId = itemUIRegistry.getWidgetId(w);
                 event.visibility = itemUIRegistry.getVisiblity(w);
                 // event.item contains data from the item including its state (in event.item.state)
-                event.item = EnrichedItemDTOMapper.map(item, false, null, null);
+                String widgetTypeName = w.eClass().getInstanceTypeName()
+                        .substring(w.eClass().getInstanceTypeName().lastIndexOf(".") + 1);
+                event.item = EnrichedItemDTOMapper.map(item, "mapview".equalsIgnoreCase(widgetTypeName),
```
kaikreuzer: Could you extract `"mapview".equalsIgnoreCase(widgetTypeName)` into a variable for readability?

bundles/io/org.eclipse.smarthome.io.rest.core/src/main/java/org/eclipse/smarthome/io/rest/core/item/EnrichedItemDTOMapper.java
lolodomo: The very specific use case was requested by you ;)
No, this change is not breaking the API. As you requested, I used exsisting fields.
Of course we can extend the initial need but my initial intention was just to implement ypour use case.

Can you explain what you mean by Predicate ?

bundles/io/org.eclipse.smarthome.io.rest.core/src/main/java/org/eclipse/smarthome/io/rest/core/item/EnrichedItemDTOMapper.java
lolodomo: Yes, that is correct.
Your use case was to get all Location items inside a group. I thinkmy code matches your use case.

bundles/io/org.eclipse.smarthome.io.rest.sitemap/src/main/java/org/eclipse/smarthome/io/rest/sitemap/internal/PageChangeListener.java
lolodomo: Ok

lolodomo:
@kaikreuzer : I have the feeling that you try to extend your initial need after it is implemented.
Why not but in this case, can you explain the expected feature ? Do you want a generic way to filter items ? Is it by type only ?

bundles/io/org.eclipse.smarthome.io.rest.core/src/main/java/org/eclipse/smarthome/io/rest/core/item/EnrichedItemDTOMapper.java
kaikreuzer: > No, this change is not breaking the API.

It is. It is a public method of an exported class. Or am I missing something?

> Can you explain what you mean by Predicate ?

Sure: I refer to a `java.util.function.Predicate`. Have a look at `GroupItem.getMembers(Predicate<Item> filterItem)` as an example.

kaikreuzer:
No, I don't think that I in any way extend the requirement.
My only concern is that the chosen implementation breaks the API and whenever we do so, we should carefully craft that change so that we do not need to break it again and again for every future small change.

lolodomo:
Of course, the answer from the REST call is different for a mapview widget attached to a group item. It is the object of this PR and the object of your change request. Is it what you mean when you say I am breaking the API ?
Except this change, everything else remains unchanged, including the REST item API.

I wuill take a look to Predicate...

kaikreuzer:
>  Is it what you mean when you say I am breaking the API ?

No, I mean what I [stated above](https://github.com/eclipse/smarthome/pull/5069#discussion_r167558087):

> [You have changed] a public method of an exported class [EnrichedItemDTOMapper]

lolodomo:
Ah ok, you're right. I will restore the old method parameters and add a new method. It will avoid breaking the API.

bundles/io/org.eclipse.smarthome.io.rest.core/src/main/java/org/eclipse/smarthome/io/rest/core/item/EnrichedItemDTOMapper.java
kaikreuzer: The method is called with drillDown=true, which means that it traverses the tree. I'd therefore say it is not the generally expected result (while we can discuss which way we want it for the mapview widget).

kaikreuzer:
Adding a new method results in situations [like this](https://github.com/eclipse/smarthome/blob/master/bundles/core/org.eclipse.smarthome.core.thing/src/main/java/org/eclipse/smarthome/core/thing/Channel.java#L73-L134).
Note that we deprecated all those and prefer other solutions. For adding filtering, we clearly favor Java8 streams and Predicates, so I wonder if you see any reason against my suggestion? If you are not feeling comfortable with implementing it with Predicates yourself, I can take care of that, if you like.

lolodomo:
Ok I let you do, it looks like you have a precise idea how you want this being implemented.

kaikreuzer:
@lolodomo I have pushed a commit with my suggested implementation, please have a look.

If it is ok for you, we would now only need @maggu2810 or @sjka for merging.

lolodomo:
I will have a look later today when I am at home.

lolodomo:
LGTM

kaikreuzer:
@maggu2810 Can you merge then?

maggu2810:
Sure, sorry, it seems I forget about this.

kaikreuzer:
> Once merged, I assume UIs will need updates to support group item in mapview widget.

Right! Would you be interested to look into this?

lolodomo:
Yes I can have a look what we can do in Basic UI and Classic UI.
Please note another remark regarding Basic UI, the position is not updated without a reload of the page. Is it a problem ?
In classic UI, as the page is reloaded at each item update, it is certainly working.

kaikreuzer:
I didn't yet notice that the Basic UI does not update the location. Yes, I would consider this a bug and it should actually do it.

lolodomo:
Having several markers with OSM is apparently not as easy to implement as having only one.

lolodomo:
The WEB link is built by the servlet only.

htreu:
From what I get the OpenStreetMap iframe integration is not capable of rendering multiple markers through URL parameters.
In Paper UI the locations now are displayed using the OpenLayers JavaScript library. With this its possible to add multiple markers.
This would mean Basic UI (and maybe Classic UI?) would have to be migrated to the OpenLayers lib for location display (& update). The basic implementation is already done in Paper UI and can be taken from there. It also supports drag&drop for markers but read-only would be a good first goal I guess.

kaikreuzer:
@htreu Can you say something about the size of that lib? How much would it blow up the Basic & Classic UI (as we cannot easily share it with the Paper UI)?

htreu:
This is the custom build we use in Paper UI right now:
`-rw-r--r--  1 henning  staff   234K Feb  9 09:16 node_modules/openlayers/ol-esh.js`.
See https://github.com/eclipse/smarthome/blob/master/extensions/ui/org.eclipse.smarthome.ui.paper/openlayers-esh.json for the current build configuration.

lolodomo:
I am sorry but in this case I prefer let a good JavaScript developer handle this change in Basic UI and Classic UI.

PR #5070, opened 2018-02-11
WU binding: add new channels containing the icon key

### LINKED ISSUE #5068: WeatherUnderground binding - iconUrl
https://github.com/eclipse/smarthome/blob/e1f424af668aded560c0473791349acd392f41be/extensions/binding/org.eclipse.smarthome.binding.weatherunderground/src/main/java/org/eclipse/smarthome/binding/weatherunderground/internal/json/WeatherUndergroundJsonForecastDay.java#L95

There's a reference about the iconUrl in the source code, this is not mentiond in the documentation. #iconUrl does not work in 2.3.0.SNAPSHOT . Can this be confirmed iconUrl can;t be used?

If so, I would like an aditional channel to get it, or even better to get the 'Icon name' mentiond in the API documentation of [WeatherUnderground](https://www.wunderground.com/weather/api/d/docs?d=resources/phrase-glossary&_ga=2.210180819.719075492.1518269308-1261285835.1518092632&MR=1).

The reason for this I can use custom icons, and line them up against this [set of icons](https://www.npmjs.com/package/weather-underground-icons).

In retun I will create a Habpanel widget (almost finished).

lolodomo:
Fixes #5068


htreu:
Hey @lolodomo: What is the meaning of the iconKey? What is the user supposed to do with it?

htreu:
Never mind, I simply read the corresponding issue :-)

PR #5075, opened 2018-02-11
bump tycho (1.1.0) and enable nullness checks again

maggu2810:
Tycho 1.1.0 has been released (2018-02-02).

Will add a revert for #4604 the next days.


kaikreuzer:
Why WIP? We could update Tycho in any case right away, can't we?

maggu2810:
The first reason has been because I started a local build and created the PR before it has been finished (I would like to go to sleep). So, I did not know if the build succeeds or not.

The other point is that we normally apply changes because there is a need for. At least bumps haven't been done without any reason (correct me if I am wrong). The revert of #4604 is AFAIK the only need to bump Tycho at all. So, I assume it is okay to use one PR for the bump and the changes that makes that bump necessary.

PR #5086, opened 2018-02-14
Fixed issues found by Static Code Analysis Tool (few checks)

kssimova:


bundles/automation/org.eclipse.smarthome.automation.api/src/main/java/org/eclipse/smarthome/automation/template/TemplateRegistry.java
```diff
@@ -26,7 +26,6 @@
  * @author Yordan Mihaylov - Initial Contribution
  * @author Ana Dimova - Initial Contribution
  * @author Vasil Ilchev - Initial Contribution
- * @param <T>
```
maggu2810: I agree that the removal of `@param <T>` is correct, but there should be a `@param <E>` inserted instead.

bundles/core/org.eclipse.smarthome.core/src/main/java/org/eclipse/smarthome/core/library/types/HSBType.java
```diff
@@ -227,27 +227,27 @@ public boolean equals(Object obj) {
         PercentType c = new PercentType(
                 value.multiply(BigDecimal.ONE.subtract((BigDecimal.ONE.subtract(f)).multiply(s))));
 
-        if (h_int == 0 || h_int == 6) {
+        if (hInt == 0 || hInt == 6) {
```
maggu2810: Isn't that a candidate for a switch case (as you already touched it anyway)?

bundles/io/org.eclipse.smarthome.io.console.karaf/META-INF/MANIFEST.MF
```diff
@@ -7,9 +7,9 @@ Bundle-SymbolicName: org.eclipse.smarthome.io.console.karaf
 Bundle-Vendor: Eclipse.org/SmartHome
 Bundle-Version: 0.10.0.qualifier
 Import-Package: 
- org.apache.karaf.shell.api.action.lifecycle;version="4.0.0",
- org.apache.karaf.shell.api.action;version="4.0.0",
- org.apache.karaf.shell.api.console;version="4.0.0",
+ org.apache.karaf.shell.api.action.lifecycle,
```
maggu2810: IIRC I added this sometime because it has been tested with >= 4.0.0 only. Do you test that is works with 3.x, too?
So, to be clear, it has been added by intention and not because the Eclipse IDE added it...

extensions/binding/org.eclipse.smarthome.binding.fsinternetradio.test/META-INF/MANIFEST.MF
```diff
@@ -10,9 +10,9 @@ Import-Package: javax.servlet,
  javax.servlet.http,
  org.apache.commons.io,
  org.eclipse.jdt.annotation;resolution:=optional,
- org.eclipse.jetty.server;version="[9.3.15,9.4.0)",
- org.eclipse.jetty.server.handler;version="[9.3.15,9.4.0)",
- org.eclipse.jetty.servlet;version="[9.3.15,9.4.0)",
+ org.eclipse.jetty.server,
```
maggu2810: This has been already discusses and also this one has been added (at the time of writing) by intention (and IIRC the discussion if this has to be removed has not been finished).

bundles/core/org.eclipse.smarthome.core/src/main/java/org/eclipse/smarthome/core/library/types/HSBType.java
kssimova: You are absolutely right! I'll do it right away.

kssimova:
@maggu2810 I updated this PR, it would be great if you take a look!

PR #5087, opened 2018-02-14
Tradfri: Show IP address in label of discovered gateway

triller-telekom:
Analog to the [hue binding](https://github.com/eclipse/smarthome/blob/master/extensions/binding/org.eclipse.smarthome.binding.hue/src/main/java/org/eclipse/smarthome/binding/hue/internal/discovery/HueBridgeNupnpDiscovery.java#L94) it is nice to see the IP address in the discovery result, especially if one has more than one gateway in the network.


kaikreuzer:
> Why not using the representation property that should be constant for a device.

I agree, this is what the representation property was actually created for - adding this to the label (as done in the hue binding at the time when this property didn't exist) is rather a workaround.

What we would need though is a change in the Paper UI to actually SHOW the representation property for the inbox results, which is currently not done.

kaikreuzer:
@triller-telekom Can you create an according issue for the Paper UI as discussed?

triller-telekom:
Done, see #5091

PR #5092, opened 2018-02-15
Paper UI: Update item state on state changed events

### LINKED ISSUE #4988: Paper UI: every item update refreshes UI item state
From the discussion and analysis in #4923 we can see that Paper UI does update its item states on every `ItemStateEvent` which is emitted regardless of a state change. Instead the UI should update its own item representation only on `ItemStateChangedEvent`.

htreu:
This fixes #4988.


extensions/ui/org.eclipse.smarthome.ui.paper/web-src/js/controllers.js
```diff
@@ -101,56 +103,26 @@ angular.module('PaperUI.controllers', [ 'PaperUI.constants' ]).controller('BodyC
...
-                });
-            }
-        }
-        var index = getItemIndex(itemName);
-        if (index !== -1) {
-            changeStateRecursively($rootScope.data.items[index]);
         }
     });
 
-    eventService.onEvent('smarthome/items/*/statechanged', function(topic, stateObject) {
```
kaikreuzer: I would have expected that you are actually adding a line to receive changed events and not removing it.
Aren't you know still listening for ALL state events and you do the detection of changes in the UI itself instead of merely listening on changed events?

htreu: I refactored the `state` handling to now handle `statechanged`events. See https://github.com/eclipse/smarthome/pull/5092/files#diff-86bb251ac89e6ba6cda9d10aecc6664dR77

kaikreuzer: Ah, thanks, I somehow overlooked that line above!

PR #5093, opened 2018-02-15
for annotated methods the cardinality is implicit mandatory

maggu2810:
Related to #4738

@lpapazow Can you check your formatter, it seems to work not reliable.
There are missing spaces `immediate=true`, tabs are used for intention instead of spaces, ...

@kaikreuzer Shouldn't SAT break the build if tabs are used for indentation?

### DIFF
--- a/bundles/io/org.eclipse.smarthome.io.rest.core/src/main/java/org/eclipse/smarthome/io/rest/core/internal/persistence/PersistenceResource.java
+++ b/bundles/io/org.eclipse.smarthome.io.rest.core/src/main/java/org/eclipse/smarthome/io/rest/core/internal/persistence/PersistenceResource.java
@@ -107,13 +107,13 @@ public class PersistenceResource implements RESTResource {
     protected void unsetPersistenceServiceRegistry(PersistenceServiceRegistry persistenceServiceRegistry) {
         this.persistenceServiceRegistry = null;
     }
-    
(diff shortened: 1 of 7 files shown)


kaikreuzer:
@lpapazow might actually be the one with the better knowledge about this.
What [I can see](https://github.com/openhab/static-code-analysis/pull/235/files#diff-63767b6e0071833a15e9215a887b807cR73) is that we have a check for XMLs, but even this will only issue a warning (and thus not break the build. Afaik, we do not have a check for Java files in place as we expect the format-on-save option to take care of that.

If you think we should enforce the spaces, we can discuss this.

maggu2810:
> If you think we should enforce the spaces, we can discuss this.

We should discuss some time, how we could enforce that the correct formatter (IDE setup) is used.

kaikreuzer:
Not sure if it is possible to have the full IDE formatting settings checked in the build.
Checking tabs vs. spaces might be a simple new check, though.

svilenvul:
Regarding the indentation.

As long as I remember we have two checks already for spacing:
- FileTabCharacter does not allow tabs in java files (it is a checkstyle  one);
- OnlyTabIdentationInXmlFilesCheck for xml files (our implementation);

As long I see the classes that have tabs are not in "src/main/java" folder. Currently Checkstyle will skip them.

PR #5095, opened 2018-02-15
fix licenses by running license:format

maggu2810:
@svilenvul @kaikreuzer Shouldn't SAT complain about that ones?


kaikreuzer:
I would have expected so...

svilenvul:
The pattern in the ESH repo is not strong enough :)

https://github.com/eclipse/smarthome/blob/master/tools/static-code-analysis/checkstyle/ruleset.properties

in comparison to the reference

https://github.com/openhab/static-code-analysis/blob/master/sat-plugin/src/main/resources/rulesets/checkstyle/rules.xml#L36


the "/d/d/d/d" should be replaced with "{0}". PR will follow up shortly :)

maggu2810:
Thanks

PR #5105, opened 2018-02-16
Fix nullannotations in ItemResource and possible NPEs in ItemDTOMapper

triller-telekom:

### DIFF
--- a/bundles/io/org.eclipse.smarthome.io.rest.core/src/main/java/org/eclipse/smarthome/io/rest/core/internal/item/ItemResource.java
+++ b/bundles/io/org.eclipse.smarthome.io.rest.core/src/main/java/org/eclipse/smarthome/io/rest/core/internal/item/ItemResource.java
@@ -501,7 +501,7 @@ public class ItemResource implements RESTResource {
     public Response createOrUpdateItem(
             @HeaderParam(HttpHeaders.ACCEPT_LANGUAGE) @ApiParam(value = "language") String language,
             @PathParam("itemname") @ApiParam(value = "item name", required = true) String itemname,
-            @ApiParam(value = "item data", required = true) @Nullable GroupItemDTO item) {
+            @ApiParam(value = "item data", required = true) GroupItemDTO item) {
         final Locale locale = LocaleUtil.getLocale(language);
 
         // If we didn't get an item bean, then return!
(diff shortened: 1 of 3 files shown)


bundles/io/org.eclipse.smarthome.io.rest.core/src/main/java/org/eclipse/smarthome/io/rest/core/internal/item/ItemResource.java
```diff
@@ -494,7 +501,7 @@ public Response removeTag(@PathParam("itemname") @ApiParam(value = "item name",
     public Response createOrUpdateItem(
             @HeaderParam(HttpHeaders.ACCEPT_LANGUAGE) @ApiParam(value = "language") String language,
             @PathParam("itemname") @ApiParam(value = "item name", required = true) String itemname,
-            @ApiParam(value = "item data", required = true) GroupItemDTO item) {
+            @ApiParam(value = "item data", required = true) @Nullable GroupItemDTO item) {
```
sjka: How can it be `required = true` and `@Nullable` at the same time?

bundles/io/org.eclipse.smarthome.io.rest.core/src/main/java/org/eclipse/smarthome/io/rest/core/internal/item/ItemResource.java
```diff
@@ -539,7 +546,8 @@ public Response createOrUpdateItem(
     @ApiOperation(value = "Adds a list of items to the registry or updates the existing items.")
     @ApiResponses(value = { @ApiResponse(code = 200, message = "OK", response = String.class),
             @ApiResponse(code = 400, message = "Item list is null.") })
-    public Response createOrUpdateItems(@ApiParam(value = "array of item data", required = true) GroupItemDTO[] items) {
+    public Response createOrUpdateItems(
+            @ApiParam(value = "array of item data", required = true) GroupItemDTO @Nullable [] items) {
```
sjka: Again, `required = true` and `@Nullable` are contradicting each other, aren't they?

triller-telekom:
I have removed the two `@Nullable` from the REST API interface and kept the `null` checks since this is an external interface where we cannot be sure to receive a non-null object.

maggu2810:
I triggered Travis CI again.

triller-telekom:
close -> reopen because of wemo test failure:

```
WemoLightHandlerOSGiTest
```

PR #5109, opened 2018-02-17
BasicUI: properly destroy controls before navigation

resetnow:
* Fix memory leaks in BasicUI from ESH code. Without proper event listeners cleanup, DOM nodes were left hanging detached in heap. This should improve long-term UI stability.
* Pause any media elements before navigation

### DIFF
--- a/extensions/ui/org.eclipse.smarthome.ui.basic/web-src/smarthome.js
+++ b/extensions/ui/org.eclipse.smarthome.ui.basic/web-src/smarthome.js
@@ -164,6 +164,25 @@
 		});
 	}
 
+	function EventMapper() {
+		var
+			_t = this;
+
+		function processEvents(callable, table) {
+			table.forEach(function(entry) {
+				var
+					element = entry[0],
+					event = entry[1],
+					handler = entry[2];
+
+				callable.call(element, event, handler);
+			});
+		}
+
+		_t.map = processEvents.bind(null, Node.prototype.addEventListener);
+		_t.unmap = processEvents.bind(null, Node.prototype.removeEventListener);
+	}
+
 	function HistoryStack() {
 		var
 			_t = this;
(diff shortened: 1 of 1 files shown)


lolodomo:
Here is the example I am using inside the demo sitemap:
`Video encoding="mjpeg" url="http://webcam.st-malo.com/axis-cgi/mjpg/video.cgi?resolution=640x480"
`
I tried your code and it does not help to stop video proxying by thez server, neither with Firefox nor with Chrome. It is only working in Edge. I did not test with Edge without your change so I cannot say that your change helped for Edge.
Note that proxying is immediately stopped if I trigger a refresh with F5 (in Firefox).

resetnow:
For me it does stop after a while or after I trigger garbage collection manually, so it makes sense. I'll update to manually stop the translation.

resetnow:
Updated with media support

lolodomo:
Quickly tested with my MJPEG video example and now it works in Firefox, Chrome and Edge. Bravo.
Needs to be tested with a video tag (MP4 file for example) too.

lolodomo:
For video like MP4 files, the handling seems to be different. Rather than having regular data transfer like for MJPEG, it looks like the browser is downloading and caching the data. So the data transfer is not regular. Browser caching leads to another problem, I will open a different issue.
By the way, it looks like you fixed the issue I noticed when using MJPEG video.

resetnow:
For MP4 videos I just added `pause()` call. Browser prefetching may occur after the call for a while I guess.

lolodomo:
I just tested the PR but did not review the code.

lolodomo:
Yes I will.

resetnow:
Any update on the PR?

maggu2810:
I cannot really review that changes, but perhaps @htreu can have a look at the JavaScript changes.
If you tested it yourself and also @lolodomo I am fine with give it a go.
@htreu Okay for you?

htreu:
All, sorry I'm so late. I have to admit that the Basic UI JavaScript code goes over my head. So I find it hard to ensure correctness here.

PR #5113, opened 2018-02-19
fix synchronization in QueueingThreadPoolExecutor

### LINKED ISSUE #5031: No events received from ESH
We are currently experiencing problems with receiving all kinds of events from ESH. It looks like for certain times we do not receive any events at all from ESH. If this happens, only events from ESH are affected, while at the same time we still receive events when the OSGI EventAdmin is used directly (we have some custom events in our own bundles, where the EventAdmin is used directly).

We have seen the issues from last December (#4591 and other issues referenced there) and think this could be related, but we do not see any log messages about the EventAdmin blacklisting certain services here. So we cannot say if there is a connection, but it seems likely.

We first observed this behaviour with the 0.9.0 release version.

sjka:
...as before it could end up with queued tasks and no monitoring thread
if tasks were added right at the moment when the monitoring thread
decided to  terminate after there were no new tasks queued withing
two seconds.

fixes #5031

### DIFF
--- a/bundles/core/org.eclipse.smarthome.core/src/main/java/org/eclipse/smarthome/core/common/QueueingThreadPoolExecutor.java
+++ b/bundles/core/org.eclipse.smarthome.core/src/main/java/org/eclipse/smarthome/core/common/QueueingThreadPoolExecutor.java
@@ -68,7 +68,7 @@ public class QueueingThreadPoolExecutor extends ThreadPoolExecutor {
     private final BlockingQueue<Runnable> taskQueue = new LinkedTransferQueue<>();
 
     /** The thread for processing the queued tasks */
-    private Thread queueThread;
+    private volatile Thread queueThread;
     private final ReadWriteLock lock = new ReentrantReadWriteLock(true);
 
     private final Object semaphore = new Object();
--- a/bundles/core/org.eclipse.smarthome.core/src/main/java/org/eclipse/smarthome/core/common/QueueingThreadPoolExecutor.java
+++ b/bundles/core/org.eclipse.smarthome.core/src/main/java/org/eclipse/smarthome/core/common/QueueingThreadPoolExecutor.java
@@ -20,6 +20,8 @@ import java.util.concurrent.ThreadFactory;
 import java.util.concurrent.ThreadPoolExecutor;
 import java.util.concurrent.TimeUnit;
 import java.util.concurrent.atomic.AtomicInteger;
+import java.util.concurrent.locks.ReadWriteLock;
+import java.util.concurrent.locks.ReentrantReadWriteLock;
 
 import org.slf4j.Logger;
 import org.slf4j.LoggerFactory;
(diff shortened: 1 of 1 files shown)


bundles/core/org.eclipse.smarthome.core/src/main/java/org/eclipse/smarthome/core/common/QueueingThreadPoolExecutor.java
```diff
@@ -179,7 +192,15 @@ public void run() {
                                 logger.trace("Executing queued task of thread pool '{}'.", threadPoolName);
                                 QueueingThreadPoolExecutor.super.execute(runnable);
                             } else {
-                                break;
+                                try {
+                                    lock.writeLock().lock();
```
htreu: I'm not 100% sure but shouldn't the readLock also be locked here?
Szenario: writeLock aquired here, then the Thread is paused. Switching to another Thread trying to add to the queue (line 166ff): readLock can be aquired, task added to queue, condition in line 119 is still false. When this thread continues the queuedTread is set to null and the queue again is stuck.

wdyt?

sjka: No. See https://docs.oracle.com/javase/7/docs/api/java/util/concurrent/locks/ReentrantReadWriteLock.html

TL;DR: The locks are interconnected - while many threads can have a read-lock at the same time, the write-lock is exclusive, i.e. only one thread can have it at a time and no read-locks are possible at the same time. That's the whole clue of the ReadWriteLocks.

maggu2810:
Can you please explain me a detail about the `QueueingThreadPoolExecutor`?

* The `EventHandler.dispatchESHEvent` method build a new `InvocationHandlerAsync` object (by calling the `build()` method and the use of `withAsync()`).
* The `receive` method is called, so the `invoke` method of the `InvocationHandlerAsync` object reference is called.
* The `SafeCallManagerImpl.enqueue` method is called.
* The `SafeCallManagerImpl.trigger` method is called.
* The `submit` method of the scheduler is called.
* `AbstractExecutorService.submit` calls the member function `execute`
* The `QueueingThreadPoolExecutor.execute` method checks if the `taskQueue` is empty and so calls (if empty) `ThreadPoolExecutor.execute` (`super.execute`)
* The `ThreadPoolExecutor` implementation itself is using the `workQueue` itself that has been supplied by the constructor.

The `workQueue` used by the `ThreadPoolExecutor` is set on construction.
The `QueueingTreadPoolExecutor` uses `new SynchronousQueue<Runnable>()`.
This one differs from the `taskQueue` used by the `QueueingTreadPoolExecutor` itself (`new LinkedTransferQueue<>()`).
Has it been implemented by intention that `workQueue != taskQueue`?
Makes the check in `QueueingTreadPoolExecutor.execute` sense?
Calling `QueueingTreadPoolExecutor.execute` multiple times will not change `taskQueue` at all.

maggu2810:
Ah, got it. The important point is the `QueueingRejectionHandler.rejectedExecution` method.

bundles/core/org.eclipse.smarthome.core/src/main/java/org/eclipse/smarthome/core/common/QueueingThreadPoolExecutor.java
htreu: Nice, thanks for the short explanation.

PR #5115, opened 2018-02-19
Fix 'use System.arraycopy' warnings generated by Static Code Analysis…

lpapazow:
As suggested by the **AvoidArrayLoops** in [pmd](http://pmd.sourceforge.net/pmd-5.0.0/rules/java/optimizations.html), copying data between two arrays is changed to be done with System.arraycopy method.

### DIFF
--- a/bundles/automation/org.eclipse.smarthome.automation.api/src/main/java/org/eclipse/smarthome/automation/parser/ParsingException.java
+++ b/bundles/automation/org.eclipse.smarthome.automation.api/src/main/java/org/eclipse/smarthome/automation/parser/ParsingException.java
@@ -67,10 +67,8 @@ public class ParsingException extends Exception {
         StackTraceElement[] st = new StackTraceElement[size];
         for (int n = 0; n < exceptions.size(); n++) {
             StackTraceElement[] ste = exceptions.get(n).getStackTrace();
-            for (int i = 0; i < ste.length; i++) {
-                st[index] = ste[i];
-                index++;
-            }
+            System.arraycopy(ste, 0, st, index, ste.length);
+            index += ste.length;
         }
         return st;
     }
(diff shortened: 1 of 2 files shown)


bundles/automation/org.eclipse.smarthome.automation.api/src/main/java/org/eclipse/smarthome/automation/parser/ParsingException.java
```diff
@@ -67,10 +67,8 @@ public String getMessage() {
         StackTraceElement[] st = new StackTraceElement[size];
         for (int n = 0; n < exceptions.size(); n++) {
             StackTraceElement[] ste = exceptions.get(n).getStackTrace();
-            for (int i = 0; i < ste.length; i++) {
-                st[index] = ste[i];
-                index++;
-            }
+            System.arraycopy(ste, 0, st, index, ste.length);
+            index += ste.length;
```
htreu: Edit: Please add a test case for this method prior to refactoring.

This fancy array engineering may with some even more fancy stream magic be refactored to:
```
List<StackTraceElement> stes = exceptions.stream()//
    .map(exception -> exception.getStackTrace())//
    .flatMap(array -> Arrays.stream(array))//

return stes.toArray(new StackTraceElement[0]);
```

or the old list fashioned way:

```
List<StackTraceElement> ste = new ArrayList<>();
for (ParsingNestedException e : exceptions) {
    ste.addAll(Arrays.asList(e.getStackTrace()));
}

return stes.toArray(new StackTraceElement[0]);
```

bundles/core/org.eclipse.smarthome.core.thing/src/main/java/org/eclipse/smarthome/core/thing/ThingUID.java
```diff
@@ -94,9 +94,7 @@ public ThingUID(String bindingId, ThingUID bridgeUID, String id) {
         String[] result = new String[3 + bridgeIds.length];
         result[0] = bindingId;
         result[1] = thingTypeId;
-        for (int i = 0; i < bridgeIds.length; i++) {
-            result[i + 2] = bridgeIds[i];
-        }
+        System.arraycopy(bridgeIds, 0, result, 2, bridgeIds.length);
```
htreu: Imho this also should be refactored to Lists.

PR #5117, opened 2018-02-19
Fixed issues found by Static Code Analysis Tool (OnlyTabIdentationInXmlFilesCheck)

tanyatanyatanya:
Changed spaces to tabs for some xml files.

Signed-off-by: Tanya D. Georgieva <tanya.dancheva.georgieva@gmail.com>

### DIFF
--- a/extensions/binding/org.eclipse.smarthome.binding.astro/ESH-INF/config/config.xml
+++ b/extensions/binding/org.eclipse.smarthome.binding.astro/ESH-INF/config/config.xml
@@ -2,7 +2,7 @@
 <config-description:config-descriptions xmlns:xsi="http://www.w3.org/2001/XMLSchema-instance"
 	xmlns:config-description="http://eclipse.org/smarthome/schemas/config-description/v1.0.0"
 	xsi:schemaLocation="http://eclipse.org/smarthome/schemas/config-description/v1.0.0
-        http://eclipse.org/smarthome/schemas/config-description-1.0.0.xsd">
+	http://eclipse.org/smarthome/schemas/config-description-1.0.0.xsd">
 
 	<config-description uri="thing-type:astro:config">
 		<parameter name="geolocation" type="text" required="true">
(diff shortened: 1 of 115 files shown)


maggu2810:
Should we use the same XML indentation style for really all XML files also if specific ones have their own style defined?

E.g. Maven has its own style for XML files:

```
The Maven style for XML files is mainly:

Indentation: Always use 2 space indents, unless you're wrapping a new XML tags line in which case you should indent 4 spaces.
```

J-N-K:
Please do not change the DMX binding files. This is alrteady incoroporated in #4812.

tanyatanyatanya:
@kaikreuzer wdyt? Should we change spaces to tabs for all XML files or leave the ones pointed out by @maggu2810?

kaikreuzer:
I'd agree with @maggu2810 that we do not have to enforce that rule on Maven files - we mainly defined that rule for XML files, which end up in the binary jars, e.g. everything we put under ESH-INF.

So the easiest thing might be to exclude any "pom.xml" from this check (sorry, as you have apparently invested quite some time in changing all of them for this PR...).

tanyatanyatanya:
@kaikreuzer I updated the PR and set suppression for any "pom.xml" files for OnlyTabIdentationInXmlFilesCheck instead of changing the logic for it. Also do you think its a good idea to change the severity from "warning" to "error" for this check?

J-N-K:
Another thing: The Eclipse formatter does not end up with XML files that follow these guidelines.

kaikreuzer:
@maggu2810 Good point about the license headers in XML. I don't know whether this can be addressed by the license Maven plugin. @tanyatanyatanya Would you like to have a look into that?
Once this works, I'd agree that the severity could be changed to error.

tanyatanyatanya:
@kaikreuzer @maggu2810 I have checked the documentation about the license maven plugin and I only found an option for changing the number of spaces with indentSize parameter. I do not think it is possible to change the indentation with this plugin to tabs. 

I have tried to find if it is possible with another plugin but the one that have an option for changing spaces to tabs is no longer maintainable.

And as for the license header in the XML files I managed to change the header style and set tab before each line. I will upload one more PR with changed files and I could change the pom because of the headers if you think that is a good idea and also to set the severity level to error.

maggu2810:
Let's give it a try. Thanks for your efforts.

PR #5119, opened 2018-02-19
make methods null safe

kaikreuzer:


kaikreuzer:
FTR: Most changes came from the auto-formatter, I only added the null checks, which is the only code change.

PR #5126, opened 2018-02-20
Adapted FSInternetRadio to use CommonHttpClient

kaikreuzer:


extensions/binding/org.eclipse.smarthome.binding.fsinternetradio/src/main/java/org/eclipse/smarthome/binding/fsinternetradio/handler/FSInternetRadioHandler.java
```diff
@@ -114,7 +116,7 @@ public void run() {
         }
     };
 
-    public FSInternetRadioHandler(Thing thing) {
+    public FSInternetRadioHandler(Thing thing, CommonHttpClient httpClient) {
```
maggu2810: Do you miss to set the assignment?

```
this.httpClient = httpClient;
```

Fields that are set by the constructor and immutable should be marked as final.
You would get an error that it is not set.

extensions/binding/org.eclipse.smarthome.binding.fsinternetradio/src/main/java/org/eclipse/smarthome/binding/fsinternetradio/internal/FSInternetRadioHandlerFactory.java
```diff
@@ -22,16 +22,24 @@
...
+import org.osgi.service.component.annotations.ConfigurationPolicy;
+import org.osgi.service.component.annotations.Reference;
 
 /**
  * The {@link FSInternetRadioHandlerFactory} is responsible for creating things and thing
  * handlers.
  *
  * @author Patrick Koenemann - Initial contribution
  */
+@Component(service = ThingHandlerFactory.class, immediate = true, configurationPid = "binding.fsinternetradio", configurationPolicy = ConfigurationPolicy.OPTIONAL)
```
maggu2810: Hm, IIRC we don't need `immediate` for the handler factory. If no one is interested in thing handler factories, no need to have it. But there should be in every ESH instance "someone" that collects the references...

kaikreuzer:
Hm, I am failing to adapt the pure unit tests, because I cannot get hold of an `CommonHttpClient` instance... We will have to figure out a good way to be able to use an HttpClient in the unit tests without requiring to do full OSGi tests....

maggu2810:
Hm, isn't that (as an example) already done in the `SecureHttpClientFactoryTest`?

kaikreuzer:
Well, that test has the advantage to be in a test fragment of the org.eclipse.smarthome.io.net.http bundle and thus has access to the non-exported classes...

maggu2810:
Hm, shouldn't the "is not exported" limitation exist only in OSGi?
I don't know how this is handled by Eclipse / Tycho but in pure Java tests (e.g. using Maven) I can access all classes in the classpath.
Need to have a look at, why a non-OSGi test respect the imports and exports at all.

kaikreuzer:
In theory, you are right, but in practice (at least in the IDE) the classpath is assembled through PDE, which only makes the exported classes available and thus you will see compile errors in your test if you use anything else.

maggu2810:
There could be a utility class (don't know a good name) that looks similar to this one:

```Java
package org.eclipse.smarthome.io.net.http;

import java.util.HashMap;
import java.util.Map;

import org.eclipse.smarthome.io.net.http.internal.SecureHttpClientFactory;

public class HttpTestUtils {

    private static class SecureHttpClientFactoryWrapper extends SecureHttpClientFactory {
        @Override
        public void activate(final Map<String, Object> parameters) {
            super.activate(parameters);
        }
    }

    public static HttpClientFactory createHttpClientFactory(final Map<String, Object> config) {
        // createConfigMap(10, 200, 60, 5, 10, 60)
        SecureHttpClientFactoryWrapper secureHttpClientFactory = new SecureHttpClientFactoryWrapper();
        // secureHttpClientFactory.setTrustmanagerProvider(trustmanagerProvider);
        secureHttpClientFactory.activate(config);
        return secureHttpClientFactory;
    }

    public static Map<String, Object> createConfig(int minThreadsShared, int maxThreadsShared,
            int keepAliveTimeoutShared, int minThreadsCustom, int maxThreadsCustom, int keepAliveTimeoutCustom) {
        Map<String, Object> configMap = new HashMap<>();
        configMap.put("minThreadsShared", minThreadsShared);
        configMap.put("maxThreadsShared", maxThreadsShared);
        configMap.put("keepAliveTimeoutShared", keepAliveTimeoutShared);
        configMap.put("minThreadsCustom", minThreadsCustom);
        configMap.put("maxThreadsCustom", maxThreadsCustom);
        configMap.put("keepAliveTimeoutCustom", keepAliveTimeoutCustom);
        return configMap;
    }

}
```

It's public interface is only using exported stuff. So it could be used in tests.

```Java
package org.eclipse.smarthome.binding.fsinternetradio.test;

import java.util.Map;

import org.eclipse.smarthome.io.net.http.CommonHttpClient;
import org.eclipse.smarthome.io.net.http.HttpClientFactory;
import org.eclipse.smarthome.io.net.http.HttpTestUtils;
import org.junit.Test;

public class SimpleTest {
    @Test
    public void demo() {
        final Map<String, Object> cfg = HttpTestUtils.createConfig(10, 200, 60, 5, 10, 60);
        HttpClientFactory factory = HttpTestUtils.createHttpClientFactory(cfg);
        CommonHttpClient httpClient = factory.getCommonHttpClient();
    }
}
```

My idea has been to add the utility class to a fragment (e.g. io.net.test) and add this fragment to the required bundles of the other test.
But I don't succeed to add the test fragment to required bundles.

maggu2810:
But wouldn't be the cleaner approach to use an OSGi test instead of adding that "workarounds"?

> without requiring to do full OSGi tests

Why does it matter?

kaikreuzer:
> Why does it matter?

Well, we have spent all the work of https://github.com/eclipse/smarthome/pull/5056 to enable just that. Many tests are just fine as plain Java and those are much easier and faster to execute and maintain. No need to struggle with broken launch configs, etc. I think lowering the bar to get tests working is important to have people write tests in the first place.

You're drafted utility class might solve the issue here, but it still feels a bit hacky (and not really obvious to find for people who are about to write a test).

I was not rather thinking about reverting one of my earlier ideas that the `CommonHttpClient` should be an ESH interface and only delegate to the Jetty `HttpClient` to instead have the `CommonHttpClient` extend the Jetty `HttpClient` and overwrite the setter-methods that we want to disallow by a no-op implementation (or throwing `MethodNotImplementedException`s). That way, a test could simply create a pure Jetty client instance and pass that in for tests - that would be straight and easy.

Another advantage would be that we do not have a mixed ESH+Jetty client API - this is in the current solution imho dangerous as we do not require a fixed version of the Jetty client to be present in a runtime - but anything >=9.3.15 and <10.0.0 should do. This makes delegation pretty fragile as we would rely on all method signatures of the installed Jetty client to be fully compatible with 9.3.15.

kaikreuzer:
> overwrite the setter-methods that we want to disallow by a no-op implementation (or throwing MethodNotImplementedExceptions).

That's a nice theory, which I just tried out - until I noticed that the `stop()` method is marked `final` and that's the most important one that people should be prevented to call 😢 .

kaikreuzer:
New idea, which really looks neat in the code: Simply wrap the shared `HttpClient` instance in a dynamic proxy. Unfortunately, the `HttpClient` is a class and not an interface and thus it isn't possible to create proxy instances for it.

sobimibos:
You can create proxies for classes with the cglib library. But you're not able to overwrite final methods. But what I have seen from the code of the HttpClient, overwriting stop is not really needed. It internally calls some do...-mthodes like doStop, which can be overwritten.

kaikreuzer:
Due to the lack of any other better options and after discussing it with @sobimibos and @sjka, I would suggest to simply return the original Jetty HttpClient instance as the shared instance and make sure through code reviews (and possibly static code analysis) that no consumer stops the instance or changes its settings through any of the setter methods.
After all, we have the same situation already with the scheduler instance in handlers, which is also a shared thread pool and no binding must ever call shutdown on it.

@maggu2810 Are you ok with that solution as well?

maggu2810:
I am okay with that solution.

What about my comment WRT the immediate setting?

extensions/binding/org.eclipse.smarthome.binding.fsinternetradio/src/main/java/org/eclipse/smarthome/binding/fsinternetradio/internal/FSInternetRadioHandlerFactory.java
kaikreuzer: I did that change in line with [the code that is generated by our archetype](https://github.com/eclipse/smarthome/blob/master/tools/archetype/binding/src/main/resources/archetype-resources/src/main/java/internal/__bindingIdCamelCase__HandlerFactory.java#L43).

kaikreuzer:
@maggu2810 Question answer inline above.

I fixed the tests, all should be good to merge now.

extensions/binding/org.eclipse.smarthome.binding.fsinternetradio/src/main/java/org/eclipse/smarthome/binding/fsinternetradio/internal/FSInternetRadioHandlerFactory.java
maggu2810: If you prefer to discuss it in a separate issue and then fixing this class (if necessary later), see #5163
I hoped we could fix this on the go...

PR #5127, opened 2018-02-21
Proxy service: add a timeout setting

### LINKED ISSUE #4873: Proxy timeout when streaming with video widget
When using the video widget in .sitemap with a video stream as input source like:
`Video encoding="mjpeg" url="http://cam:80/cgi-bin/hi3510/snap.cgi?&-getstream&-chn=1"`
(SV3C cam, but the result is the same for other streams), there is a proxy timeout after exactly 60 seconds, and the stream stops. The logs shows:

````
2018-01-07 15:38:10.713 [DEBUG] [.s.u.i.proxy.AsyncProxyServlet:257  ] - 734249762 asynchronous write of 704 bytes completed on HttpOutput@6c1f91a2{READY}
2018-01-07 15:38:10.718 [DEBUG] [.s.u.i.proxy.AsyncProxyServlet:73   ] - 734249762 proxying content to downstream: 1024 bytes
2018-01-07 15:38:10.718 [DEBUG] [.s.u.i.proxy.AsyncProxyServlet:251  ] - 734249762 asynchronous write start of 1024 bytes on HttpOutput@6c1f91a2{READY}
2018-01-07 15:38:10.718 [DEBUG] [.s.u.i.proxy.AsyncProxyServlet:257  ] - 734249762 asynchronous write of 1024 bytes completed on HttpOutput@6c1f91a2{READY}
2018-01-07 15:38:10.724 [DEBUG] [.s.u.i.proxy.AsyncProxyServlet:627  ] - 734249762 proxying failedjava.util.concurrent.TimeoutException: Total timeout 60000 ms elapsed
	at org.eclipse.jetty.client.TimeoutCompleteListener.run(TimeoutCompleteListener.java:71)

2018-01-07 15:38:10.726 [DEBUG] [.s.u.i.proxy.AsyncProxyServlet:240  ] - 734249762 proxying complete
````

Is this intended behavior, and if so - is there a setting to change that to make streams work indefinitely? Or is this a bug?

(This is against 'master' as of Jan 7, 2018)

Thanks for any help,
-Th

EDIT: Using the same url in a different client like VLC doesn't result in a timeout, it streams until VLC is closed.

lolodomo:
Default value is set to 0 which disables the timeout.
Disabling the timeout avoids stopping the video playback after the timeout.

Fixes #4873

### DIFF
--- a/bundles/ui/org.eclipse.smarthome.ui/src/main/java/org/eclipse/smarthome/ui/internal/proxy/ProxyServletService.java
+++ b/bundles/ui/org.eclipse.smarthome.ui/src/main/java/org/eclipse/smarthome/ui/internal/proxy/ProxyServletService.java
@@ -73,6 +73,7 @@ public class ProxyServletService extends HttpServlet {
     public static final String PROXY_ALIAS = "proxy";
 
     private static final String CONFIG_MAX_THREADS = "maxThreads";
+    private static final String CONFIG_TIMEOUT = "timeout";
     private static final int DEFAULT_MAX_THREADS = 8;
     public static final String ATTR_URI = ProxyServletService.class.getName() + ".URI";
     public static final String ATTR_SERVLET_EXCEPTION = ProxyServletService.class.getName() + ".ProxyServletException";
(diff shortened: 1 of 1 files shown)


triller-telekom:
I am not too sure about whether it is a good idea to disable the timeout in general. What if the client dies? The connection will be held open indefinitely.

In the issue you mention that we maybe could create a second proxy just for video streams which does not have this timeout, so we could keep the timeout value for the normal one.

Would that be an option?

lolodomo:
If the client dies, the connection is closed, no problem.

lolodomo:
@triller-telekom : the alternative solution would require a change in all UIs, even if easy to apply.  What alias would we use ion this case ? `proxyvideo` ?

triller-telekom:
To be honest I do not have too much insight in how our UIs interact with the proxyservice and if it is worth to change all of them. Maybe @kaikreuzer can comment on that.

lolodomo:
I close my PR as disabling the timeout could apparently introduce a security problem.

PR #5129, opened 2018-02-21
Optionally pass item name to TransformationHelper to improve error messages

### LINKED ISSUE #5120: Useless warning from TransformationHelper: Cannot format state...
The following log entry might be 100% correct, but at the same time it is 100% useless:

```
2018-02-19 20:45:48.781 [WARN ] [.core.transform.TransformationHelper] - Cannot format state '1519022658' to format '%.0f': f != java.lang.String
```

There is lacking _any_ detail where to start searching, e.g. the item (or widget?) name would be a good start.

triller-telekom:
Fixes #5120

### DIFF
--- a/bundles/core/org.eclipse.smarthome.core.transform/src/main/java/org/eclipse/smarthome/core/transform/TransformationHelper.java
+++ b/bundles/core/org.eclipse.smarthome.core.transform/src/main/java/org/eclipse/smarthome/core/transform/TransformationHelper.java
@@ -81,6 +81,19 @@ public class TransformationHelper {
      * @return the result of the transformation. If no transformation was done, the state is returned
      */
     public static String transform(BundleContext context, String stateDescPattern, String state) {
+        return transform(context, stateDescPattern, state, null);
+    }
+
+    /**
+     * Transforms a state string using transformation functions within a given pattern.
+     *
+     * @param context a valid bundle context, required for accessing the services
+     * @param stateDescPattern the pattern that contains the transformation instructions
+     * @param state the state to be formatted before being passed into the transformation function
+     * @param itemName the name if the item whose state should be converted or null
+     * @return the result of the transformation. If no transformation was done, the state is returned
+     */
+    public static String transform(BundleContext context, String stateDescPattern, String state, String itemName) {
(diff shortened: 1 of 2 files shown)


triller-telekom:
I kept the old method signature since it is a public static method. I did a search in my IDE from where it is called and only found the `EnrichedItemDTOMapper` which I have adjusted. But one never knows who else might call it.

bundles/core/org.eclipse.smarthome.core.transform/src/main/java/org/eclipse/smarthome/core/transform/TransformationHelper.java
```diff
@@ -81,6 +81,19 @@ public static TransformationService getTransformationService(BundleContext conte
...
+        return transform(context, stateDescPattern, state, null);
+    }
+
+    /**
+     * Transforms a state string using transformation functions within a given pattern.
+     *
+     * @param context a valid bundle context, required for accessing the services
+     * @param stateDescPattern the pattern that contains the transformation instructions
+     * @param state the state to be formatted before being passed into the transformation function
+     * @param itemName the name if the item whose state should be converted or null
```
htreu: The sentence is a little bit strange and it should indicate that itemName is optional.

triller-telekom:
Thanks, I actually had a typo in there "if" instead of "of" which made my comment even more confusing :)

maggu2810:
Hm, I don't know if it is a good practice to add method arguments just because that information will be added by the called function in logging statements as long as the method itself does not use it at all.
Now there is a method without that argument and one with.
What should the caller be told? Please use the other method that needs the item name to get a better logging information on failure.

Hm....

kaikreuzer:
I agree with @maggu2810. What I especially dislike is the fact that Transformations and the TransformationHelper are so far completely oblivious to items and these methods can be called in other situations. We'd e.g. planned to have a transformation profile soon.

triller-telekom:
I wasn't sure either if it is a good idea to implement it that way. And I agree with @maggu2810 that the choice for 2 methods isn't nice.

So I am closing this PR.

Issue #5099, opened 2018-02-16
Command options for channels


sjka:
There have been several discussions previously about channels which primary purpose is to send commands towards a device (i.e. the opposite direction of trigger channels). There are references in https://github.com/eclipse/smarthome/issues/4272#issuecomment-329713835, https://github.com/eclipse/smarthome/pull/3740#issuecomment-311005040, #4942 and probably some more. Also, there are numerous bindings which support this use-case.

While normal state channels so far have been a good fit for such use-cases, the bindings currently do not have any proper means to express the purpose of such channels and what commands/values are possible or expected: Bindings can only define _state_ options, i.e. a mapping between discrete values and their labels in order to give clues to the UIs how they should be rendered. The same clue is missing for _commands_ currently (which is not necessarily identical). 

If there would be something like "command options", then UIs could use them to render the corresponding controls, e.g. buttons instead of the plain number or a drop-down.

maggu2810:
> Bindings can only define state options, i.e. a mapping between discrete values and their labels in order to give clues to the UIs how they should be rendered. The same clue is missing for commands currently (which is not necessarily identical).

Correct, it seems there is something missing :wink:

Issue #4272, opened 2017-09-15
Trigger channels


davidgraeff:
This issue is either about poor documentation or extending the trigger channel.

The way that the documentation is layout out, with the trigger channels described first in the channels section among other channel types like switch, made me understand the concept differently than it was meant to be.

There are generally two types of channel types, those that are read-only and show information to the user and those that are meant to be used for interacting:

| Data type | Event from binding (Show to user)               | Action on binding (User interact) |
|-----------|-----------------------------|---------------|
| Boolean   | Contact                     | Switch        |
| Number    | Text (, Gauge, Progressbar) | Dimmer        |
| String    | Text                        | State Option  |
| Stateless | -                           | Trigger       |

As shown in the table, I understood the trigger channel to be the stateless way for interacting (the GUI would show a push button).

In reality the trigger channel is an event source and and is meant to be on the middle column of that table above.

I suggest to allow the trigger channel to be used in both ways by using the already existing "read-only" flag. If read-only is set, it is the event-source like now. If read-only is not set, it is meant to be used with a push button. The table from above would look like this:

| Data type | Event from binding (Show to user)               | Action on binding (User interact) |
|-----------|-----------------------------|---------------|
| Boolean   | Contact                     | Switch        |
| Number    | Text (, Gauge, Progressbar) | Dimmer        |
| String    | Text                        | State Option  |
| Stateless | Trigger (read-only)   | Trigger      |

What do you think?

## Use-case

In the OpenHAB Milight binding I have a use-case for stateless channels. There are certain features that are like modes, they can be activated but can't be switched on/off. Because I misunderstood the concept, I modeled those as trigger channels and now I have users complaining that those can't be used (https://community.openhab.org/t/the-new-milight-ibox-and-milight-binding/17621/259?u=david_graeff).

sjka:
Well, the key point is that trigger channels only _fire events_, but they neither maintain a state (i.e. don't send state updates) nor accept any commands. Therefore they cannot be linked to items. Their only purpose (currently) is to trigger rules by listening to these events. 

[...]

PR #5131, opened 2018-02-21
Extend ChannelType by command options

htreu:
This is WIP and open for discussion.

This addresses #5099 by adding command options as an alternative to a state description.
The implementation is naive and straigt forward.
Command options will be rendered as push buttons by UIs and send the corresponding command value as a command to the channel.
With this proposal, the state of the channel will not be represented in the UI, so ThingHandlers may not even update the state.


htreu:
Paper UI will now render push-buttons for Switch & String items when the corresponding ChannelType defines commandOptions.

htreu:
This is an example of the Sonos binding's "Clear Queue" channel:

screen shot 2018-02-22 at 12 03 16

triller-telekom:
I just tested this PR and it works for me. But before merging I would say that the feature should be "complete" in terms of that BasicUI and ClassicUI should also render buttons for command options.

bundles/core/org.eclipse.smarthome.core.thing.xml/src/main/java/org/eclipse/smarthome/core/thing/xml/internal/ChannelTypeConverter.java
```diff
@@ -142,6 +173,7 @@ protected ChannelTypeXmlResult unmarshalType(HierarchicalStreamReader reader, Un
         String description = super.readDescription(nodeIterator);
         String category = readCategory(nodeIterator);
         Set<String> tags = readTags(nodeIterator);
+        List<CommandOption> commandOptions = readCommandOptions(nodeIterator);
```
kaikreuzer: Even though we do not yet have anything besides the options - wouldn't it be more symmetrical to have a "commandDescription" next to the state and event descriptions?

bundles/core/org.eclipse.smarthome.core/src/main/java/org/eclipse/smarthome/core/types/CommandOption.java
```diff
@@ -0,0 +1,52 @@
...
+
+    /**
+     * The command which will be send to the Channel
+     */
+    private final String command;
+
+    /**
+     * The name of the command which will be displayed in the UI.
+     */
+    private final String name;
```
kaikreuzer: In StateOption, this is called "label".

extensions/binding/org.eclipse.smarthome.binding.sonos/ESH-INF/thing/channels.xml
```diff
@@ -31,9 +31,12 @@
 	</channel-type>
 
 	<channel-type id="clearqueue" advanced="true">
-		<item-type>Switch</item-type>
+		<item-type>String</item-type>
```
kaikreuzer: This change can break existing setups.

bundles/test/org.eclipse.smarthome.magic/ESH-INF/thing/channel-types.xml
```diff
@@ -51,13 +51,11 @@
 		<item-type>String</item-type>
 		<label>Alert</label>
 		<description>The alert channel allows a temporary change to the bulb’s state.</description>
-		<state>
```
kaikreuzer: This breaks backward compatibility as it removes the state options, which are still a valid information.

extensions/binding/org.eclipse.smarthome.binding.sonos/ESH-INF/thing/channels.xml
htreu: thanks, this was not intended is not even necessary. Switch items will also get the push buttons.

htreu:
> Wrt rendering buttons: So far, we render drop-down boxes, which is also a valid solution - it only fails for "pure command channels" which do not have any state and thus a selection should not stay active. But this is going to be addressed with #5011.

A drop-down box really fails at being a command input. Selecting a value from the list is more of a "change value" notion then pushing a button which has a meaning of "do it now".

What might be a discriminating property for "pure command channels" is the fact that only one command is available. This would mean the `<command-options>` list should rather be a single `<command-option value="clear">Clear Queue</...>` entry.

maggu2810:
> With this proposal, the state of the channel will not be represented in the UI, so ThingHandlers may not even update the state.

So it is exclusive? Using command options will hide the state in the UI?

kaikreuzer:
> Using command options will hide the state in the UI?

It definitely must not. There are clearly channels where you want command options, but which also have a state.

> A drop-down box really fails at being a command input.

Buttons really fail once you have too many of them...

> What might be a discriminating property for "pure command channels" is the fact that only one command is available. 

Why this? Id rather think that "pure command channels" usually have many options as it does not make sense to create 3 command channels for a thing, if you can create a single one that can handle 3 different commands.

htreu:
>> Using command options will hide the state in the UI?

>It definitely must not. There are clearly channels where you want command options, but which also have a state.

Then you are more close to select a new state from the options. Also: examples please.

>>A drop-down box really fails at being a command input.

>Buttons really fail once you have too many of them...

Here single-command-only-channels come in (see below). Besides: having too many buttons doesn't make drop downs any better ;-)

> Why this? Id rather think that "pure command channels" usually have many options as it does not make sense to create 3 command channels for a thing, if you can create a single one that can handle 3 different commands.

Why would you want 1 channel for arbitrary, unrelated commands? Having > 1 command on a channel may be an indicator for the channel to be a "normal" state channel.
I'd like to see some examples to see which channel category fits the different situations.

Example:
Sonos binding, command: "Clear Queue"
* This channel should not have other commands
* The channel does not change its own state

kaikreuzer:
> Then you are more close to select a new state from the options. Also: examples please.

Almost all that you can currently find. E.g. "alert" channel from the hue binding.

> Here single-command-only-channels come in (see below). Besides: having too many buttons doesn't make drop downs any better ;-)

I don't understand what kind of widget you then suggest for e.g. 5 command options?
I think the [Scene Selection](http://demo.openhab.org:8080/basicui/app?w=0202&sitemap=demo) works pretty good here.

> Why would you want 1 channel for arbitrary, unrelated commands? 

Why would you want to have multiple channels, when all that they provide is a distinct state (which you do not have)?

See https://github.com/openhab/openhab2-addons/pull/2345#discussion_r164434998:
_Anyhow, it does not make sense to have two commands (ON and OFF) for every single trigger here. If we say that the state does not make any sense (as it gets out of sync too easily), we should actually not model any separate channels at all, but rather have a single "command channel", which would take Strings like PLAY/PAUSE/RESET (and the user could have a single button for each of those options in the UI)._

htreu:
> E.g. "alert" channel from the hue binding

As I understand the hue `alert` channel its something which can be switched on/off and a UI should reflect the current state. I see it being modelled as a "normal" on/off channel linked to a Switch item (like it is right now).

> the Scene Selection works pretty good

No, ist does not: How do you select an already selected preset _again_? With push buttons you would simply press it once more. Also, selecting a preset does not mean a distinct state to change to the preset name. It would only be the status channel "last preset selected". A scene/preset selection changes state on a different channel.
Example: The tuning of an analog radio station would have the frequency channel and a command-options preset channel. Selecting on of the preset radio stations would change state on the frequency channel but leave no information on the preset channel. Here you _could_ mark a preset as selected when the frequency channel tunes into the corresponding frequency. I would consider this not to be necessary as the state is display elsewhere. And coming to the scene selection again you would not want to highlight a specific scene if by accident all corresponding states suddenly match.

The Lametric Time example is nice and another counter example to the single-command-only channel idea:
* The countdown app receives commands start/pause/reset
* There is no feedback channel, so no state which tells if the countdown is running or is in the reset state
* > 1 commands which belong together

I agree, a command channel with many options/buttons is not nice to render (right now) but I'm sure we will come up with a solution.

davidgraeff:
This is exactly what I meant with #4272. Hopefully this gets merged some time :)

mgeramb:
Hi all,
my binding [AmazonEchoControl](https://community.openhab.org/t/beta-openhab2-amazonechocontrol-binding-controlling-alexa-from-openhab2/37844) is also an example for the need of command channels. I have different command (start playing news flash, start playing weather information, start playing flash briefing). But I have no feedback from the device what is currently playing. Currently I use switch channels where the binding set the value automatically back to Off. But I think this is not a nice solution because the thing gets more and more channels. So I'am currently looking for a other solution. I think the command option and a single channel will help. 

But I'am not sure if the commands should be only static. I have other channels which get the available options from the device. E.g. playAlarmSound, the available sounds are provided with a DynamicStateDescriptionProvider, it would be nice if this feature could also be used for these channels in the future (Currently I have no need, because for these types of channels, I get a feedback from the device).

Tony-Hu:
I'm wondering that if the stateless channel is an essential part of some certain scenarios. Sometimes, either users or developers may not care about the current state a channel holds. However, shall we provide a way to block the state in lower level functions?
>This is an example of the Sonos binding's "Clear Queue" channel:

Perhaps we can consider the ``clearqueue`` channel has a state to indicate how many songs in the play queue by using non-negative numbers. And some negative numbers to indicate the queue is being cleared or it's failed. Although it's rarely happened for the failed execution of clear queue, it will make users unhappy. 
For those users who don't about some certain states, we can provide an option to hide those in both web UIs and mobile clients.
And for the developers, they can just ignore the state in various ways.

kaikreuzer:
Wdyt about the suggestion of a `DynamicCommandDescriptionProvider`?
Apart from this, is there anything else open after the discussion?

htreu:
I did rebase the branch and adopt the latest Paper UI refactorings.

Open issues:

* Rename `commandOptions` to `commandDescription`
* Provide Thing DSL enhancement for textual configuration
* provide and evaluate `DynamicCommandDescriptionProvider`
* Implement Classic UI & Basic UI

davidgraeff:
@htreu This would be nice to have as well for a lot of bindings ^^

htreu:
Since I#m about to implement at least some of the missing pieces: What is you expectation on how to access the `CommandDescription`? Right now it is only accessible through the ChannelType. I suppose we want the information about commands and a localized description to be present on the item domain? wdyt? /cc @kaikreuzer

htreu:
One more thing we need to consider using `DynamicCommandOptionProvider` to support dynamic commands: The handler must then also be able to execute these dynamic commands. This might need some additional work for a plugin mechanism.

htreu:
> I'd tend to saying that if a channel is readOnly=false and has state options, those should be treated as valid command options as well, unless there is a dedicated list of command options provided.

@kaikreuzer what do you think is the benefit of having state options along command options? With state options only we already have the rendering of drop downs and the correlation of state and the selected option.

> Wrt rendering buttons: So far, we render drop-down boxes, which is also a valid solution - it only fails for "pure command channels" which do not have any state and thus a selection should not stay active. But this is going to be addressed with #5011.

I see this PR to be used for "pure command channels" and to me it fits nice into the other channel "types" we already offer:

* trigger channels: the device triggers an event but has no state
* read-only channels: the state is only given by the device
* "normal" channels: state is given by the device but may be set by the framework/user
* "pure command" channels: the opposite of a trigger channel, commands are sent to the device but a state is not received on this channel. Other channels might be affected by the command and update their state accordingly.

kaikreuzer:
> what do you think is the benefit of having state options along command options? 

You mean, if both are identical or in general? In general, they can be different sets. If they are identical, my suggestion was that it should be enough for the user to provide state options and we automatically consider the command options to be the same.

> With state options only we already have the rendering of drop downs

Which I'd consider a bug/workaround of the current UIs. Drop-downs do not make any sense for states, only for commands. So the REST API should provide command options and those should be rendered by the UIs as drop-downs. We should not move the logic to UIs, i.e. they must not have to check whether there are command options and if not, they instead use state options as commands. Much better to have this centrally in the framework.

> I see this PR to be used for "pure command channels" and to me it fits nice into the other channel 

Not sure how your remark is related to my comment that drop-down boxes do not seem to be a good widget for command-only channels...?

htreu:
Thanks @kaikreuzer, that makes it clearer now.

What about different state and command options. What is a UI supposed to render if the current item state has a corresponding state option but no command option? In general, what are the state options then used for? We would need a new set of UI components which make a difference between state display and commands the user can issue. Like: 1st row displays all state options with the current state highlighted, 2nd row shows command buttons with the current state visualised as a pressed button (if any matches).

I will update the PR, so command options will be filled from state options when not explicitly given.

kaikreuzer:
> Like: 1st row displays all state options

There is no need to display state options at all - a "normal" UI could just ignore them, imho they are mostly relevant for rule editors to know what states they might come across and thus offer according triggers etc.

So only the current state is to be shown. And yes, if there is a state that is NOT a command at the same time, the drop-down widget isn't the right choice either. Or, similarly to command-only channels, it could be left with a "blank" option that is displayed and the state simply appears textually like it is actually already possible in e.g. the Basic UI with buttons:

screenshot 2019-01-18 at 11 15 20

PR #5133, opened 2018-02-22
add pmd suppressions file

lpapazow:
Added a PMD suppression file location property in the pom file.

Currently org.eclipse.smarthome.binding.hue.internal.HueBridge throws exception in a finally block and thus violates the DoNotThrowExceptionInFinally PMD check. This check is now ignored for this class.

### DIFF
--- a/pom.xml
+++ b/pom.xml
@@ -533,6 +533,7 @@
                 </execution>
               </executions>
               <configuration>
+                <pmdFilter>${basedirRoot}/tools/static-code-analysis/pmd/suppressions.properties</pmdFilter>
                 <checkstyleProperties>${basedirRoot}/tools/static-code-analysis/checkstyle/ruleset.properties</checkstyleProperties>
                 <checkstyleFilter>${basedirRoot}/tools/static-code-analysis/checkstyle/suppressions.xml</checkstyleFilter>
                 <findbugsExclude>${basedirRoot}/tools/static-code-analysis/findbugs/suppressions.xml</findbugsExclude>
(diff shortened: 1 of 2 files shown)


triller-telekom:
Instead of suppressing this check, I would rather ask why we do have the `throw` in a `finally` in there in the first place. @ddudnik it looks like this

https://github.com/eclipse/smarthome/blob/3462cdac46fe813fe29afc7a57934fb09ea00588/extensions/binding/org.eclipse.smarthome.binding.hue/src/main/java/org/eclipse/smarthome/binding/hue/internal/HueBridge.java#L686-L701

 is your code. Any specific reason to throw the `Exception` there instead of just logging an error or warning?

ddudnik:
@triller-telekom , unfortunately I can't provide a reason for that as my commit was an effort to just move 'Jue' library to become a part of Hue Binding without changing the actual source code of Jue. The orginal (unmaintained) library is here: https://github.com/Q42/Jue

The reasons for making Jue a part of Hue Binding could be looked up here: https://github.com/eclipse/smarthome/issues/2601

maggu2810:
I am not against the PMD suppression file at all, but I second @triller-telekom that changing the code would be a better idea.

It seems that exceptions itself should be ignored, so add changing it to `IOException | RuntimeException ex` should catch most cases (I assume all relevant ones).
Can't you move the code in the finally block below the try-catch-finally block at all then?

lpapazow:
Instead of adding suppressions, refactored the code in #5154

PR #5146, opened 2018-02-24
Update JsonPath Readme

Josar:
Content and structure

Signed-off-by: josar <josuaarndt@live.de>

### DIFF
--- a/extensions/transform/org.eclipse.smarthome.transform.jsonpath/README.md
+++ b/extensions/transform/org.eclipse.smarthome.transform.jsonpath/README.md
@@ -56,7 +56,7 @@ Returns `null` if the JsonPath expression could not be found.
 Compared to standard JSON the transformation it returns evaluated values when a single alement is retrieved from the querry.
 Means it does not return a valid JSON `[ 23.2 ]` but `23.2`, `[ "Outside" ]` but `Outside`.
 This makes it possible to use it in lables or output channel of things and get Numbers or strings instead of JSON arrays.
-A querry which returns multiple elements as lsit is not supported.
+A querry which returns multiple elements as list is not supported.
 
 ## Further Reading
 * An extended [introduction](https://www.w3schools.com/js/js_json_intro.asp) can be found at W3School.
--- a/extensions/transform/org.eclipse.smarthome.transform.jsonpath/README.md
+++ b/extensions/transform/org.eclipse.smarthome.transform.jsonpath/README.md
@@ -53,7 +53,7 @@ Now the resulting Number can also be used in the label to [change the color](htt
 ## Differences to standard JsonPath
 
 Returns `null` if the JsonPath expression could not be found.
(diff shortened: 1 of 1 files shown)


extensions/transform/org.eclipse.smarthome.transform.jsonpath/README.md
```diff
@@ -1,18 +1,63 @@
...
+
+Now the resulting Number can also be used in the label to [change the color](https://docs.openhab.org/configuration/sitemaps.html#label-and-value-colors) or in a rule as value to compare.
+
+## Differences to standard JsonPath
+
+Returns `null` if the JsonPath expression could not be found.
+Compared to standard JSON the transformation it returns evaluated values when a single alement is retrieved from the querry.
+Means it does not return a valid JSON `[ 23.2 ]` but `23.2`, `[ "Outside" ]` but `Outside`.
+This makes it possible to use it in lables or output channel of things and get Numbers or strings instead of JSON arrays.
+A querry which returns multiple elements as lsit is not supported.
```
htreu: typo: as list

extensions/transform/org.eclipse.smarthome.transform.jsonpath/README.md
```diff
@@ -1,18 +1,63 @@
...
+[{ "device": { "location": "Outside", "status": { "temperature": 23.2 }}}]
+```
 
-## Testing Tools
+the JsonPath expression `$.device.location` exstracts the JSON
 
-* [http://jsonpath.com/](http://jsonpath.com/)
-* [http://www.jsonquerytool.com/](http://www.jsonquerytool.com/)
+```
+[ "Outside" ]
```
htreu: Given the "Differences" description below this would result in `"Outside"`.

extensions/transform/org.eclipse.smarthome.transform.jsonpath/README.md
```diff
@@ -1,18 +1,63 @@
...
+```
+
+the JsonPath expression `$.device.status.temperature` exstracts the JSON
+
+```
+[ 23.2 ]
```
htreu: Given the "Differences" description below this would result in `23.2`.

extensions/transform/org.eclipse.smarthome.transform.jsonpath/README.md
Josar: This is a generall explanation of json and not relate to openhab/smarthome and so i decided to keep it as it would be in json on purpose. 

Same for the next.

htreu: To me it was confusing to first read this explanation and in the next paragraph see how the result (especially the `[ 23.2 ]`) could posted as an update to a Number item.
I get your point but imho the examples given here should directly reflect the behavior in the framework.

PR #5150, opened 2018-02-25
fixes "resumeAfter" in ChaserThing, move tests from OSGI to Java only

J-N-K:
Add some tests for the chaser thing and fixes a resum-after bug, where resuming the last state lead to channel lock

- [x] confirm this fixes the bug reported in the forum


extensions/binding/org.eclipse.smarthome.binding.dmx.test/src/test/java/org/eclipse/smarthome/binding/dmx/ChaserThingHandlerTest.java
```diff
@@ -112,4 +119,110 @@ public void initializationOfChaserThing() {
...
+            return thingHandler;
+        });
+
+        final ManagedItemChannelLinkProvider itemChannelLinkProvider = waitForAssert(() -> {
+            final ManagedItemChannelLinkProvider tmp = getService(ManagedItemChannelLinkProvider.class);
+            assertThat(tmp, is(notNullValue()));
+            return tmp;
+        });
+
+        switchItem = new DimmerItem(TEST_SWITCHITEM_NAME);
```
triller-telekom: At first I stumbled across this line because you assign a `Dimmeritem` to a variable called "switch....", but then I was thinking why you need the items at all for your tests?

Background of my question is that I see you testing basically our whole framework and we have a lot of JUnit tests "randomly failing" (in the past also some of your DMX tests) because of timing issues and how the framework behaves in tests.

Wouldn't it be better to just test your logic by calling your methods without the framework? and for the parts where you need the frame work do not create real registries or items but instead use mocks?

J-N-K: DimmerItem is obviously wrong :-). Can you point me to an example for the mocks? I dind't find out how to do that.

J-N-K: Found it. Will Update the checks.

triller-telekom: @J-N-K Regarding the mocks... You want to test your `ThingHandler` and we just found out that the framework does not allow easy mocking because we are using `ServiceTracker`s so you cannot easily mock away the registries for example. 

Since you found something, I am looking forward to what you have achieved, but otherwise I would suggest that you keep it as it is for now we will get back to you once we got rid of the `ServiceTracker`s and have provided a decent example. Then you could create a follow up PR to clean up your tests. But this is out of scope for this PR.

J-N-K: see [here](https://github.com/J-N-K/smarthome/blob/testing/extensions/binding/org.eclipse.smarthome.binding.dmx.test/src/test/java/org/eclipse/smarthome/binding/dmx/DimmerThingHandlerTest.java).

triller-telekom: Can you push it as a new commit onto this branch please?

J-N-K: Sure. Please see the last commit here.

J-N-K:
retriggering build (failed in core thing tests)

extensions/binding/org.eclipse.smarthome.binding.dmx.test/src/test/java/org/eclipse/smarthome/binding/dmx/ArtnetBridgeHandlerTest.java
```diff
@@ -44,64 +41,57 @@
...
-                .withBridge(bridge.getUID()).withConfiguration(new Configuration(thingProperties)).build();
+        ThingHandlerCallback mockCallback = mock(ThingHandlerCallback.class);
+        doAnswer(answer -> {
+            ((Thing) answer.getArgument(0)).setStatusInfo(answer.getArgument(1));
+            return null;
+        }).when(mockCallback).statusUpdated(any(), any());
+
+        bridgeHandler = new ArtnetBridgeHandler(bridge) {
+            @Override
+            protected void validateConfigurationParameters(Map<String, Object> configurationParameters) {
```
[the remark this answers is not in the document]
J-N-K: I would love to do so, however l. 318 in the BaseThingHandler.java (`ThingType thingType = thingTypeRegistry.getThingType(getThing().getThingTypeUID());`) throws a NPE. I assume this is due to the `thingTypeRegistry` is not available.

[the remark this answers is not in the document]
triller-telekom: That is correct. As I said before, for now it's not possible to full mock away all unnecessary stuff from the handlers :(

Apart from that, the code looks fine now.

I just wanted to double check with you if what you test is what you really wanted to test and not that you have removed half of you tests just because I mentioned that it can be done without inheriting from the `JavaOSGITest` class.

J-N-K:
after #5264 has been merged, retriggering build

maggu2810:
Build seems to succeed now.
@triller-telekom have you finished your review already?

htreu:
@triller-telekom @J-N-K is this one ready to go?

triller-telekom:
@maggu2810 @htreu Yes it looks good to me.

PR #5152, opened 2018-02-26
removed Activator(s) from core bundle

kaikreuzer:
I just found this and cannot imagine that we still need it in any way.
Or does anyone know a reason that it should be kept?


sjka:
Nope 😄

PR #5158, opened 2018-02-27
Fixed issues found by the Static Code Analysis Tool (few checks)

kssimova:
And suppressed *org.eclipse.smarthome.io.console.eclipse.internal.ConsoleSupportEclipse.java* and *org.eclipse.smarthome.io.console.rfc147.internal.CommandWrapper.java* for MethodNameCheck, because 
they contain methods that start with underscore (custom OSGi console commands)

### DIFF
--- a/bundles/test/org.eclipse.smarthome.magic/src/main/java/org/eclipse/smarthome/magic/binding/handler/MagicColorLightHandler.java
+++ b/bundles/test/org.eclipse.smarthome.magic/src/main/java/org/eclipse/smarthome/magic/binding/handler/MagicColorLightHandler.java
@@ -12,8 +12,6 @@
  */
 package org.eclipse.smarthome.magic.binding.handler;
 
-import static org.eclipse.smarthome.magic.binding.MagicBindingConstants.CHANNEL_COLOR;
-
 import org.eclipse.smarthome.core.thing.ChannelUID;
 import org.eclipse.smarthome.core.thing.Thing;
 import org.eclipse.smarthome.core.thing.ThingStatus;
@@ -34,8 +32,9 @@ public class MagicColorLightHandler extends BaseThingHandler {
 
     @Override
     public void handleCommand(ChannelUID channelUID, Command command) {
-        if (channelUID.getId().equals(CHANNEL_COLOR)) {
-        }
+        // process the command for the color channel here
+        // if (channelUID.getId().equals(MagicBindingConstants.CHANNEL_COLOR)) {
+        // }
     }
 
     @Override
--- a/bundles/test/org.eclipse.smarthome.magic/src/main/java/org/eclipse/smarthome/magic/binding/handler/MagicContactHandler.java
+++ b/bundles/test/org.eclipse.smarthome.magic/src/main/java/org/eclipse/smarthome/magic/binding/handler/MagicContactHandler.java
@@ -12,8 +12,6 @@
  */
 package org.eclipse.smarthome.magic.binding.handler;
 
-import static org.eclipse.smarthome.magic.binding.MagicBindingConstants.CHANNEL_CONTACT;
-
 import org.eclipse.smarthome.core.thing.ChannelUID;
 import org.eclipse.smarthome.core.thing.Thing;
 import org.eclipse.smarthome.core.thing.ThingStatus;
@@ -34,8 +32,9 @@ public class MagicContactHandler extends BaseThingHandler {
 
     @Override
     public void handleCommand(ChannelUID channelUID, Command command) {
-        if (channelUID.getId().equals(CHANNEL_CONTACT)) {
-        }
+        // process the command for the contact channel here
+        // if (channelUID.getId().equals(MagicBindingConstants.CHANNEL_CONTACT)) {
+        // }
     }
 
     @Override
--- a/bundles/test/org.eclipse.smarthome.magic/src/main/java/org/eclipse/smarthome/magic/binding/handler/MagicDimmableLightHandler.java
+++ b/bundles/test/org.eclipse.smarthome.magic/src/main/java/org/eclipse/smarthome/magic/binding/handler/MagicDimmableLightHandler.java
@@ -12,8 +12,6 @@
  */
 package org.eclipse.smarthome.magic.binding.handler;
 
-import static org.eclipse.smarthome.magic.binding.MagicBindingConstants.CHANNEL_BRIGHTNESS;
-
 import org.eclipse.smarthome.core.thing.ChannelUID;
 import org.eclipse.smarthome.core.thing.Thing;
 import org.eclipse.smarthome.core.thing.ThingStatus;
@@ -34,8 +32,9 @@ public class MagicDimmableLightHandler extends BaseThingHandler {
 
     @Override
     public void handleCommand(ChannelUID channelUID, Command command) {
-        if (channelUID.getId().equals(CHANNEL_BRIGHTNESS)) {
-        }
+        // process the command for the brightness channel here
+        // if (channelUID.getId().equals(MagicBindingConstants.CHANNEL_BRIGHTNESS)) {
+        // }
     }
 
     @Override
--- a/bundles/test/org.eclipse.smarthome.magic/src/main/java/org/eclipse/smarthome/magic/binding/handler/MagicLocationThingHandler.java
+++ b/bundles/test/org.eclipse.smarthome.magic/src/main/java/org/eclipse/smarthome/magic/binding/handler/MagicLocationThingHandler.java
@@ -12,8 +12,6 @@
  */
 package org.eclipse.smarthome.magic.binding.handler;
 
-import static org.eclipse.smarthome.magic.binding.MagicBindingConstants.CHANNEL_LOCATION;
-
 import org.eclipse.smarthome.core.thing.ChannelUID;
 import org.eclipse.smarthome.core.thing.Thing;
 import org.eclipse.smarthome.core.thing.ThingStatus;
@@ -34,8 +32,9 @@ public class MagicLocationThingHandler extends BaseThingHandler {
 
     @Override
     public void handleCommand(ChannelUID channelUID, Command command) {
-        if (channelUID.getId().equals(CHANNEL_LOCATION)) {
-        }
+        // process the command for the location channel here
+        // if (channelUID.getId().equals(MagicBindingConstants.CHANNEL_LOCATION)) {
+        // }
     }
 
     @Override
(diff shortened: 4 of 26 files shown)


extensions/binding/org.eclipse.smarthome.binding.wemo/src/main/java/org/eclipse/smarthome/binding/wemo/handler/WemoCoffeeHandler.java
```diff
@@ -213,8 +213,6 @@ public void handleCommand(ChannelUID channelUID, Command command) {
                                 e.getMessage());
                         updateStatus(ThingStatus.OFFLINE, ThingStatusDetail.COMMUNICATION_ERROR);
                     }
-                } else if (command.equals(OnOffType.OFF)) {
-                    // do nothing, as WeMo Coffee Maker cannot be switched off remotely
```
maggu2810: Hm, at least such a comment seems to my useful.
If you want to remove the empty branch, you could still keep a comment why we don't need to handle the OFF command.

extensions/binding/org.eclipse.smarthome.binding.sonos/src/main/java/org/eclipse/smarthome/binding/sonos/internal/handler/ZonePlayerHandler.java
```diff
@@ -969,95 +965,87 @@ protected void updateMediaInformation() {
...
-            // Do nothing
-        }
-
-        else if (currentURI.isEmpty()) {
-            // Reset data
-            needsUpdating = true;
-        }
-
-        else if (currentURI.contains(GROUP_URI)) {
-            // The Sonos is a slave member of a group, we do nothing
```
maggu2810: See below (useful command)

extensions/binding/org.eclipse.smarthome.binding.sonos/src/main/java/org/eclipse/smarthome/binding/sonos/internal/handler/ZonePlayerHandler.java
```diff
@@ -331,9 +330,6 @@ public void handleCommand(ChannelUID channelUID, Command command) {
                                 getCoordinatorHandler().previous();
                             }
                         }
-                        if (command instanceof RewindFastforwardType) {
-                            // Rewind and Fast Forward are currently not implemented by the binding
```
maggu2810: See below (useful command)

extensions/binding/org.eclipse.smarthome.binding.sonos/src/main/java/org/eclipse/smarthome/binding/sonos/internal/SonosXMLParser.java
```diff
@@ -612,8 +612,6 @@ public void startElement(String uri, String localName, String qName, Attributes
                 } else {
                     textField = null;
                 }
-            } else if (qName.equals("logo")) {
-                // logo = attributes.getValue("UUID");
```
maggu2810: Do you know why this instruction has been removed?

kssimova: No, but it hasn't been modified for 2 years and 5 months, so I thought that it is not used.

extensions/binding/org.eclipse.smarthome.binding.wemo/src/main/java/org/eclipse/smarthome/binding/wemo/handler/WemoCoffeeHandler.java
kssimova: Great idea! I'll remove them, but I'll leave the comments that explain why we don't need them. Thanks!

extensions/binding/org.eclipse.smarthome.binding.sonos/src/main/java/org/eclipse/smarthome/binding/sonos/internal/SonosXMLParser.java
maggu2810:
```
commit 9a2fc9cba501faa943f2e4aee3ce30ee67f6f77f (origin/pull/548)
Author: Karel Goderis <karel.goderis@me.com>
Commit: Karel Goderis <karel.goderis@me.com>

    Migration of the Sonos 2.0 binding from openHAB 2 to Eclipse Smarthome
    
    Also-by: Andre Fuechsel <a.fuechsel@telekom.de>
    Also-by: Dan Cunningham <dan@digitaldan.com>
    Also-by: David Wheeler <hippysoyboy@gmail.com>
```

@kgoderis It seems the line has been comment out as it has been added to the ESH repo.

Do you know why it is present but comment out?

bundles/test/org.eclipse.smarthome.magic/src/main/java/org/eclipse/smarthome/magic/binding/handler/MagicColorLightHandler.java
```diff
@@ -34,8 +32,6 @@ public MagicColorLightHandler(Thing thing) {
 
     @Override
     public void handleCommand(ChannelUID channelUID, Command command) {
-        if (channelUID.getId().equals(CHANNEL_COLOR)) {
-        }
```
triller-telekom: Since the magic bundle is supposed to be a "demo bundle" to play around with and show how to implement certain things I would vote for adding a comment inside the empty if branch instead of removing it. I.e. something like `//process the command for the color channel here`

bundles/test/org.eclipse.smarthome.magic/src/main/java/org/eclipse/smarthome/magic/binding/handler/MagicContactHandler.java
```diff
@@ -34,8 +32,6 @@ public MagicContactHandler(Thing thing) {
 
     @Override
     public void handleCommand(ChannelUID channelUID, Command command) {
-        if (channelUID.getId().equals(CHANNEL_CONTACT)) {
-        }
```
triller-telekom: see above

bundles/test/org.eclipse.smarthome.magic/src/main/java/org/eclipse/smarthome/magic/binding/handler/MagicDimmableLightHandler.java
```diff
@@ -34,8 +32,6 @@ public MagicDimmableLightHandler(Thing thing) {
 
     @Override
     public void handleCommand(ChannelUID channelUID, Command command) {
-        if (channelUID.getId().equals(CHANNEL_BRIGHTNESS)) {
-        }
```
triller-telekom: see above

bundles/test/org.eclipse.smarthome.magic/src/main/java/org/eclipse/smarthome/magic/binding/handler/MagicLocationThingHandler.java
```diff
@@ -34,8 +32,6 @@ public MagicLocationThingHandler(Thing thing) {
 
     @Override
     public void handleCommand(ChannelUID channelUID, Command command) {
-        if (channelUID.getId().equals(CHANNEL_LOCATION)) {
-        }
```
triller-telekom: see above

bundles/test/org.eclipse.smarthome.magic/src/main/java/org/eclipse/smarthome/magic/binding/handler/MagicOnOffLightHandler.java
```diff
@@ -34,8 +32,6 @@ public MagicOnOffLightHandler(Thing thing) {
 
     @Override
     public void handleCommand(ChannelUID channelUID, Command command) {
-        if (channelUID.getId().equals(CHANNEL_SWITCH)) {
-        }
```
triller-telekom: see above

extensions/binding/org.eclipse.smarthome.binding.lirc/src/main/java/org/eclipse/smarthome/binding/lirc/handler/LIRCRemoteHandler.java
```diff
@@ -59,9 +58,7 @@ public void handleCommand(ChannelUID channelUID, Command command) {
             return;
         }
         if (channelUID.getId().equals(LIRCBindingConstants.CHANNEL_TRANSMIT)) {
-            if (command instanceof RefreshType) {
-                // not supported
```
triller-telekom: Please keep this comment, otherwise other developers might wonder why the refresh command is not handled although every handler has to handle it.

bundles/test/org.eclipse.smarthome.magic/src/main/java/org/eclipse/smarthome/magic/binding/handler/MagicColorLightHandler.java
kssimova: Okay, but even if there is a comment in the empty if branch, it will still be reported by the SAT.

triller-telekom: Good point. How about adding that comment just before the branch instead and in addition commenting out the whole branch? This way developers will see that someone took care of this condition and the SAT tool will not complain because that branch doesn't exist anymore in code, but in a comment?

extensions/binding/org.eclipse.smarthome.binding.sonos/src/main/java/org/eclipse/smarthome/binding/sonos/internal/SonosXMLParser.java
kgoderis: @maggu2810 Nope - I can't remember doing it and for what reason. I think the UUID refers to some bitamp that is internally stored on the Sonos player, and since we do not use that in our own GUI's, I think it can be removed, and as @kssimova pointed out, nobody complained since 2.5 years....

kssimova: We can comment out the whole branch in case someday we want to use it?

sjka: Feel free to delete it completely. There is no need to keep commented-out code "in stock" 😄

kssimova:
@maggu2810 I've updated this PR, it would be great if you take a look when you have time :)

PR #5161, opened 2018-02-27
marked internal services as deprecated to make developers aware that those must not be used

kaikreuzer:


kaikreuzer:
> I assume the plan is to make this members private some time and remove the deprecated again. 

I'd rather prefer to remove them completely (together with the bundleContext) and make the methods that require those services a part of the `callback`, which is the better place to provide an API for handlers, which we can easily evolve over time.

frickua:
@kaikreuzer How i can obtain another thing from my thing, now? I mean `thingRegistry.get(uuid)`. Thanks for the help.

PR #5166, opened 2018-02-28
convert UID and ThingEventFactory tests to Java

sjka:
contributes to #3305


sjka:
Excellent question 😄 !

To be honest, the primary reason is that I was pretty much dead-minded simply fixing the test method names, making the methods public, adding semicolons, inserted missing brackets and fixed only the pitfalls where the groovy matchers deviate from the hamcrest matchers. 

On the technical side, I'm not aware of any differences. 

To my understanding, the only reason for using the hamcrest variant is that it gets as close to human-readable language as it can get - and that afaik was one of the groovy's promises: it helps writing "fluent" code better than in java. So groovy's

```groovy
assertThat x is(y)
```

without most of the syntactic sugar is supposed to be "better" than java's

```java
assertThat(x, is(y));
// or even
assertEquals(x, y);
```

Therefore my guess is that our test code-base is using it so much because these groovy-"best practices". 

However, I "grew up" with junit assertions and I personally am pretty happy with them and think they are equally clear. They don't resemble a prosa sentence that much, but they pretty directly state at the beginning what is to be expected. As developers we are very used to reading code, so that should be fine. And the goal that reading good code must be like reading a good novel to my understanding doesn't mean we must change our languages and APIs to resemble prosa text, but rather to structure it well and write "clean" code.

Hmmm.... Simple question, way too long answer. Nevertheless, I tend to say we should make more use of the junit assertions where suitable. Let me adapt this PR quickly.

maggu2810:
... and if we can get (most of the time) rid of another dependency ... :wink:

Thank you for your detailed answer.

Are the failing tests (Travis CI) caused by your changes?

sjka:
> Are the failing tests (Travis CI) caused by your changes?

Errr... :disappointed: copy'n'paste is an art of its own.

PR #5168, opened 2018-03-01
[WIP/POC] make jsonpath accerssible directly in rules/scripts

### LINKED ISSUE #5132: JSONPATH is no JSONPATH anymore.
https://community.openhab.org/t/jsonpath-transformation-service-vs-jsonpath/40439/5

As discussed there i propose to add an additional transformation called jsonobj whihc returns an evaluated object. In means of only one single element, without the list and slashes.

This can be used to parse json in item return values, labels etc.

And to make jsonpath transformation a full fledged jsonpath. As one would expect of it.

sjka:
This PR makes jsonpath usable directly within rules & scripts.

It is currently a quick draft only, it lacks some tests and further proofing. In any case it look on the first glance as if classloading with such an extension is working well enough, although the rule engine is not having it in its own classloader directly. 

It is though using the `ActionService` interface, which currently has some very prominent usages, but no clear future when thinking of the move to the new automation engine.

fixes #5132


kaikreuzer:
@sjka Would you agree that we won't need this PR after the change we introduced with https://github.com/eclipse/smarthome/pull/5049#issuecomment-392046527?

PR #5169, opened 2018-03-01
Paper UI: display representation property in Things view & Inbox

### LINKED ISSUE #5091: Paper UI: Show representation property of a thing
Imagine 2 hue bridges or 2 tradfri bridges in the same network. if you want to distinguish the two devices of the same type it would be useful to see their `representation property` in the thing configuration page.

By seeing  this distinct property one can tell which device is which.

htreu:
This fixes #5091.

Edit: For testing purposes this also introduces a simple MagicDiscoveryService which adds a single discovery entry per scan.


htreu:
A first screen to see how the UI should be done:
screen shot 2018-03-01 at 17 32 02

Here the magic configurable thing is extended by a serialNumber property which is the representation Property. It is displayed in small text beneath the thing label.

kaikreuzer:
I only notice now that the issue says "in the thing configuration page" - to me it was clear that the issue must be about the Inbox instead as this is the place where the user cannot differentiate between two entries. Once he approves an inbox entry, he can give the Thing an individual label, so there isn't a problem anymore. Also, the representation-property was introduced specifically for Inbox usage. Might I ask you to adapt you PR?

htreu:
Yes of cause.
What do you think about exposing the representation property on the thing resource in general? Instead of giving it a prominent displaying we could mark the specific property in the properties list. wdyt?

sjka:
> Instead of giving it a prominent displaying we could mark the specific property in the properties list. 

I like.

lolodomo:
I like too  ;)

triller-telekom:
+1 :)

htreu:
Two new screens:

The properties list with a marker for the unique representation property:
screen shot 2018-03-02 at 15 11 58

The inbox entry showing the representation property next to the label:
screen shot 2018-03-02 at 15 01 34

triller-telekom:
Regarding the Inbox screen: as far as I can see appears the representation property twice. You show it next to the large label "Magic Thing" and you also use it within the name "magic-configurable-thing:31a...." is this duplication intended?

htreu:
@triller-telekom this is just from the simple design of the new `MagicDiscoveryService`. I create a random UID and use it for both, the representation property and the last segment of the thing UID.

kaikreuzer:
>>>> Instead of giving it a prominent displaying we could mark the specific property in the properties list.

>>> I like.

>> I like too ;)

> +1 :)

Sorry, I have to stand out of the crowd as so often: I don't like it at all.

Reasons: 
1. The flag was introduced to be a hint at UIs which info is important for differentiation when showing two things side by side (as in the inbox). It does not have any real value to the user - other properties might actually be more interesting to him, so I do not see any good reason to make one stand out on the Thing page and especially waste another full line of screen space.
2. The representation property refers to properties of the discovery result, NOT of the Thing. As a result, it can easily (and probably often) be a Thing configuration parameter.

htreu:
> it can easily (and probably often) be a Thing configuration parameter

Nevertheless it will be part of the Thing's properties after inbox approval.
Anyway, since I have no strong emotions regarding the thing view, and with the argument that the representation property was manly designed to support discovery result discrimination, I´m happy to remove it from the thing resource and the thing view.

further opinions @maggu2810, @sjka & @triller-telekom?

kaikreuzer:
> Nevertheless it will be part of the Thing's properties after inbox approval.

Wrong. The discovery result property will become a Thing configuration parameter.

sjka:
> The discovery result property will become a Thing configuration parameter.

This indeed is a valid point. Unfortunately I missed that one... 😩 

> It does not have any real value to the user

You mean it _did not_ have any real value? Isn't #5091 exactly about sticking it into people's faces in the inbox? With that it gets some value: recognizing how it was discovered. 

In any case, your other argument makes this discussion completely useless.

htreu:
> Wrong. The discovery result property will become a Thing configuration parameter.

Unless the thing-type does not define a config parameter with the representation property's name. Then it will end up in the thing properties as I described above.

Anyway I will remove the whole Thing DTO, REST and Paper UI display. The inbox feature seems to be sufficient here.
Any comments on the UI?

kaikreuzer:
> Any comments on the UI?

I don't find it optimal to have it shown next to the label with a font style that is otherwise only used in the other two lines. It somehow destroys the "clean" layout. But I am no designer and I don't have any better suggestions, so what to do... But at least, the spacing between the label and the property should imho slightly be increased.

maggu2810:
> Wrong. The discovery result property will become a Thing configuration parameter.

@kaikreuzer I am a little bit confused, doesn't it depend on the thing type as @htreu comment?

> Unless the thing-type does not define a config parameter with the representation property's name. Then it will end up in the thing properties as I described above.

This has been (without looking at the code) my understand. If a map entry of the discovery result is a property or a configuration of the thing depends on the key used in the map and if it is defined as a configuration.

Has this been changed?

htreu:
@kaikreuzer I agree, the "extended" label somehow kills the design. I tried to put the rep. property below as the 2nd line which results in 4 lines being displayed. This is also not ideal and the material design style guide also only suggests 3 lines for a list entry.

for reference:
screen shot 2018-03-05 at 14 59 23

kaikreuzer:
Maybe we could drop the UID as this is not really important to the user here?

htreu:
I did thought about that, too. Since Paper UI is the admin interface maybe the UID is important? Otherwise it will be visible once the Thing gets accepted, so we could give it a try:

screen shot 2018-03-05 at 15 23 22

kaikreuzer:
I'd be for it as I've so far never come across a situation where I'd be specifically interested in the discovery result UID...

htreu:
Then this is ready for a review.

kaikreuzer:
Thanks - could you check, which existing ESH bindings might need to be adapted know? (Coming to my mind is the hue binding, which currently shows the IP address as a part of the label.)

kaikreuzer:
Also: Should there be any update of the developer docs to tell people to NOT put something like IP/MACs into the discovery label, but rather use the representation property instead?

htreu:
@kaikreuzer I will have a look.

kaikreuzer:
@htreu Don't forget it :-)

htreu:
See #5278.

kaikreuzer:
Thanks - but note that #5278 only addresses https://github.com/eclipse/smarthome/pull/5169#issuecomment-373287890, but not yet https://github.com/eclipse/smarthome/pull/5169#issuecomment-373287272.
(I'd expected PRs like https://github.com/eclipse/smarthome/pull/5279 for that.)

htreu:
All other bindings do not use the representation property in the label. Hue binding does indeed show the ip address, but representation property is the serial number.

PR #5175, opened 2018-03-02
Obey mandatory channel configs before initializing thing handlers

### LINKED ISSUE #5170: ThingManager initialized thing handlers despite of missing required channel config parameters
The `ThingManager` doesn't initialize `ThingHandler` if a `Thing`'s config description defines some required parameters which are yet missing via `isInitializable(...)`. It does not consider mandatory channel config parameters in that method. 

I don't see any reason why it shouldn't - considering that the `ThingHandler` presumably relies on these parameters to be present. So Things with missing required config description parameters should also stay in `OFFLINE/CONFIGURATION_ERROR` until the config gets fixed. 

Or am I missing something here?

sjka:
If things have missing mandatory configuration parameters, then the `ThingManager` does not initialize handlers but sets the thing status to `UNINITIALIZED/HANDLER_CONFIGURATION_PENDING` instead. 

This PR now amends the mechanism to also look for missing mandatory **channel** configuration parameters.

fixes #5170

### DIFF
--- a/bundles/core/org.eclipse.smarthome.core.thing/src/main/java/org/eclipse/smarthome/core/thing/internal/ThingManager.java
+++ b/bundles/core/org.eclipse.smarthome.core.thing/src/main/java/org/eclipse/smarthome/core/thing/internal/ThingManager.java
@@ -614,11 +614,12 @@ public class ThingManager implements ThingTracker, ThingTypeMigrationService, Re
     /**
      * Determines if all 'required' configuration parameters are available in the configuration
      *
-     * @param prototype
-     * @param targetUID
-     * @param configDescriptionURIFunction
-     * @param configuration
-     * @return
+     * @param prototype the "prototype", i.e. thing type or channel type
+     * @param targetUID the UID of the thing or channel entity
+     * @param configDescriptionURIFunction a function to determine the the config description UID for the given
+     *            prototype
+     * @param configuration the current configuration
+     * @return true if all required configuration parameters are given, false otherwise
      */
     private <T extends Identifiable<?>> boolean isComplete(T prototype, UID targetUID,
             Function<T, URI> configDescriptionURIFunction, Configuration configuration) {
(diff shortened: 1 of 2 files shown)


sjka:
rebased

bundles/core/org.eclipse.smarthome.core.thing/src/main/java/org/eclipse/smarthome/core/thing/internal/ThingManager.java
```diff
@@ -593,24 +596,49 @@ private void applyDefaultConfiguration(Thing thing, ThingType thingType) {
...
         }
 
         List<String> requiredParameters = getRequiredParameters(description);
-        Map<String, Object> properties = thing.getConfiguration().getProperties();
+        Map<String, Object> properties = configuration.getProperties();
         if (logger.isDebugEnabled()) {
-            logger.debug("Configuration of thing {} needs {}, has {}.", thing.getUID(), requiredParameters,
-                    thing.getConfiguration().getProperties().keySet());
+            logger.debug("Configuration of '{}' needs {}, has {}.", targetUID, requiredParameters,
+                    configuration.getProperties().keySet());
```
maggu2810: As you already store `configuration.getProperties()` to `properties`, no need to call `getProperties()` again.

bundles/core/org.eclipse.smarthome.core.thing/src/main/java/org/eclipse/smarthome/core/thing/internal/ThingManager.java
```diff
@@ -593,24 +596,49 @@ private void applyDefaultConfiguration(Thing thing, ThingType thingType) {
...
+    }
+
+    /**
+     * Determines if all 'required' configuration parameters are available in the configuration
+     *
+     * @param prototype
+     * @param targetUID
+     * @param configDescriptionURIFunction
+     * @param configuration
+     * @return
```
maggu2810: Will you add the missing descriptions to the JavaDoc.

Issue #4503, opened 2017-11-05
Proposal/Discussion: Self-contained aAdd-on metadata


ThomDietrich:

maggu2810:
Adding some metadata that could be used by which code ever will need them some day is okay.
But I don't think we need to at redundant information.
Why do we need to add the version here? All add-ons are OSGi bundles and so the version is already placed (supported by build tools etc.) in that metainfo.
Isn't the name already be part of the metainfo?
Do you expect that the author of the addon differ from the author of the binding?
Or do we make author in the binding XML optional and use the one in add-on instead?

ThomDietrich:
Hey Markus,
I've intentionally started with "Proposal/Discussion" 😉  The above is a first design and I'd gladly remove all redundant information. Please keep in mind that I am not an experienced binding/add-on developer.

* **Version:** Sounds logical. If you are certain the field is superfluous I can surely remove it. Would you be so kind and point me towards the existing version in the OSGi bundle?
* **Name/Authors:** These are indeed already available in `binding.xml`. I did not specifically propose that but yes of course, the new file should ideally replace the binding specific `binding.xml`. Would that be an option?

kaikreuzer:
I don't think we should add urls at all in the extensions. The idea of ESH extensions is that they are potentially compatible with many different solutions that are based on ESH. So the link to the a suitable documentation will highly depend on the solution and thus cannot be given within the extension.
That's why currently the link to documentation and icons is provided by the extension service implementation that makes the extensions available to the solution. openHAB e.g. does that [here](https://github.com/openhab/openhab-core/blob/master/bundles/org.openhab.core.karaf/src/main/java/org/openhab/core/karaf/internal/KarafExtensionService.java#L109-L115).

maggu2810:
> I've intentionally started with "Proposal/Discussion"

That has been my reading. :wink: I only want to give you a comment.

> point me towards the existing version in the OSGi bundle?

Hm, look at the `META-INF/MANIFEST.MF` file that is part of every OSGi bundle.
There should be a manifest entry `Bundle-Version`.
See e.g.: https://www.osgi.org/bundle-headers-reference/

ThomDietrich:
@kaikreuzer I had that thought in mind. I believe you are talking about `documentation_url` and/or `homepage_url` only? The idea was to provide thouse for e.g. externally developed add-ons. A developer should be able to provide a link to his private/corporate add-on-related resources?

[...]

PR #5176, opened 2018-03-03
[i18n] Added translation for author of binding

cweitkamp:
- Added translation for author of binding
- Corrected linters in documentation 

In OH2 we already have some bindings which are written / maintained by more than one author (e.g. Netatmo). To provide fully internationalization we should translate the author tag as well.

https://github.com/openhab/openhab2-addons/blob/14dacc3104dc8076310433e635f9f0a1d591b9a9/addons/binding/org.openhab.binding.netatmo/ESH-INF/binding/binding.xml#L9


lolodomo:
And I can probably be added for the netatmo binding ;)

maggu2810:
:+1: for the translation support.

But to be honest, I don't like the author for bindings in the official ESH repo.
That is a meta information for the binding but what is the intention?
I assume the author don't like to be contacted personally but prefer bug reports to the ESH repo.
And is the one that started the word (initial contribution) really the author of the whole binding?
Why not using "Eclipse SmartHome project" for bindings that are maintained by the Eclipse SmartHome community?

kaikreuzer:
Sorry, this PR does not make any sense to me.
First of all, names of people are usually not translated into other languages. Their name is their name.
Second, it is "author" and not "authors", hence for bindings where there are more than a single author, we already usually simply removed that information.
Third, I agree with @maggu2810, it was rather a bad idea of including it in the first place - it does not really have any value here. I'd therefore also rather suggest to phase its use out and move this information to the "technical meta-data" as discussed in https://github.com/eclipse/smarthome/issues/4503.

cweitkamp:
Alright, I didn't see that one coming. If we don't want to have the "author", we don't need a translation for it.

Issue #5177, opened 2018-03-03
Support for LATEST/EARLIEST group functions


rmichalak:
It is currently not possible to create group with items of DateTime type, where group's state will represent latest/earliest date from its children states.
MAX/MIN group functions are limited to decimal types and extending them for that purpose does not seem like a good idea.
Proposal: create new LATEST and EARLIEST group functions.

PR #5178, opened 2018-03-03
added LATEST and EARLIEST group functions for DateTime groups

rmichalak:
Fixes: #5177
Signed-off-by: Robert Michalak <rbrt.michalak@gmail.com>

### DIFF
--- a/bundles/core/org.eclipse.smarthome.core/src/main/java/org/eclipse/smarthome/core/library/types/DateTimeGroupFunction.java
+++ b/bundles/core/org.eclipse.smarthome.core/src/main/java/org/eclipse/smarthome/core/library/types/DateTimeGroupFunction.java
@@ -1,0 +1,117 @@
+/**
+ * Copyright (c) 2014,2018 Contributors to the Eclipse Foundation
+ *
+ * See the NOTICE file(s) distributed with this work for additional
+ * information regarding copyright ownership.
+ *
+ * This program and the accompanying materials are made available under the
+ * terms of the Eclipse Public License 2.0 which is available at
+ * http://www.eclipse.org/legal/epl-2.0
+ *
+ * SPDX-License-Identifier: EPL-2.0
+ */
+package org.eclipse.smarthome.core.library.types;
+
+import java.time.ZonedDateTime;
+import java.util.Set;
+
+import org.eclipse.smarthome.core.items.GroupFunction;
+import org.eclipse.smarthome.core.items.Item;
+import org.eclipse.smarthome.core.types.State;
+import org.eclipse.smarthome.core.types.UnDefType;
+
+/**
+ * This interface is a container for group functions that require {@link DateTimeType}s for its calculations.
+ * 
+ * @author Robert Michalak - Initial contribution
+ *
+ */
+public interface DateTimeGroupFunction extends GroupFunction {
+
+    /**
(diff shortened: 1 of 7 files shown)


htreu:
Hi @rmichalak thanks a lot for this proposal.

Taken from #5177:

> MAX/MIN group functions are limited to decimal types and extending them for that purpose does not seem like a good idea.

Why not extending MIN/MAX? The mapping is obvious and we would have the same functionality with less code.
For Paper UI there is a base item specific selection of group functions when creating a new group. Please see [controllers.items.js#262](https://github.com/eclipse/smarthome/blob/master/extensions/ui/org.eclipse.smarthome.ui.paper/web-src/js/items/controllers.items.js#L262) for details and extend it for DateTime, too. 

Also: The new class `DateTimeGroupFunction` is missing in the commit.

rmichalak:
Hi @htreu,

Whoops! missing file... I'm ashamed... Will fill that gap in the evening.
My initial idea was to extend MIN/MAX for that purpose, but what about groups containing both Numeric and DateTime items? We could in theory base our logic on group type and perform calculation either on DateTime or Numeric items, but AFAIK we don't have access to this information inside `calculate` method(?).
This is why I decided to create new functions, dedicated to DateTime types.

htreu:
> We could in theory base our logic on group type and perform calculation either on DateTime or Numeric items, but AFAIK we don't have access to this information inside calculate method(?).

Thats indeed a good argument! We could apply some heuristic what MIN/MAX would probably mean but thats hard to understand/explain and may in the end be wrong anyhow.
Please push the missing class and have a look at the Paper UI group function selection. There is another file [shared.properties.js](https://github.com/eclipse/smarthome/blob/master/extensions/ui/org.eclipse.smarthome.ui.paper/web-src/js/shared.properties.js#L148) which you might need to extend for the DateTime function selection.

rmichalak:
Sure, will do. I will also take a closer look at PaperUI, cause this part of the project was off my radar so far.
Thanks for your suggestions.

bundles/core/org.eclipse.smarthome.core/src/main/java/org/eclipse/smarthome/core/library/types/DateTimeGroupFunction.java
```diff
@@ -0,0 +1,118 @@
...
+import java.time.ZonedDateTime;
+import java.util.Set;
+
+import org.eclipse.smarthome.core.items.GroupFunction;
+import org.eclipse.smarthome.core.items.Item;
+import org.eclipse.smarthome.core.types.State;
+import org.eclipse.smarthome.core.types.UnDefType;
+
+/**
+ * This interface is a container for dimension based functions that require {@link DateTypeType}s for its calculations.
```
htreu: `for dimension based functions`
These group functions are not dimension based but specifically designed for the DateTimeItem base item type.

rmichalak: Good catch. Too many copy-pastes ;). Fixed.

rmichalak:
It looks like CI failure is not related to my changes. Is there a way to run it again?

kaikreuzer:
I retriggered Travis.

PR #5179, opened 2018-03-03
Proxy service: adjustments for video widgets

### LINKED ISSUE #4873: Proxy timeout when streaming with video widget

### LINKED ISSUE #5111: Browser caching video
Due to browser caching (Firefox, Chrome, ...), if you change the URL attached to a video element in your sitemap and save your sitemap, then when you reopen the page in Basic UI for example, it will display the old video. The reason is that the used URL by Basic UI is the proxied URL and this proxied URL is unchanged. So the browser is using the video it has already cached for this URL.
To get the new video, you need to clear the browser cache.

@resetnow : should we add a timestamp parameter to proxied video URL like it is done for proxied image URLs, in order to bypass browser caching ?

lolodomo:
Disable timeout for proxied video widgets
Avoid browser caching for video widgets

Fixes #4873
Fixes #5111

### DIFF
--- a/bundles/ui/org.eclipse.smarthome.ui/src/main/java/org/eclipse/smarthome/ui/internal/proxy/AsyncProxyServlet.java
+++ b/bundles/ui/org.eclipse.smarthome.ui/src/main/java/org/eclipse/smarthome/ui/internal/proxy/AsyncProxyServlet.java
@@ -13,6 +13,7 @@
 package org.eclipse.smarthome.ui.internal.proxy;
 
 import java.util.Objects;
+import java.util.concurrent.TimeUnit;
 
 import javax.servlet.http.HttpServletRequest;
 import javax.servlet.http.HttpServletResponse;
@@ -49,6 +50,21 @@ public class AsyncProxyServlet extends org.eclipse.jetty.proxy.AsyncProxyServlet
     @Override
     protected HttpClient newHttpClient() {
         return new HttpClient(new SslContextFactory());
+    }
+
+    @Override
+    protected void sendProxyRequest(HttpServletRequest clientRequest, HttpServletResponse proxyResponse,
+            Request proxyRequest) {
+        if (service.proxyingVideoWidget(clientRequest)) {
+            // We disable the timeout for video
+            proxyRequest.timeout(0, TimeUnit.MILLISECONDS);
+
+            // We request the browser to not cache the video
+            proxyResponse.setHeader("Cache-Control", "no-cache, no-store, must-revalidate");
+            proxyResponse.setHeader("Pragma", "no-cache");
+            proxyResponse.setHeader("Expires", "0");
+        }
+        super.sendProxyRequest(clientRequest, proxyResponse, proxyRequest);
     }
 
     /**
--- a/bundles/ui/org.eclipse.smarthome.ui/src/main/java/org/eclipse/smarthome/ui/internal/proxy/ProxyServletService.java
+++ b/bundles/ui/org.eclipse.smarthome.ui/src/main/java/org/eclipse/smarthome/ui/internal/proxy/ProxyServletService.java
@@ -305,6 +305,54 @@ public class ProxyServletService extends HttpServlet {
     }
 
     /**
+     * Determine if the request is relative to a video widget.
+     *
+     * @param request the servlet request
+     * @return true if the request is relative to a video widget
+     */
+    boolean proxyingVideoWidget(HttpServletRequest request) {
+
+        boolean proxyingVideo = false;
+
+        try {
+            String sitemapName = request.getParameter("sitemap");
+            if (sitemapName == null) {
+                throw new ProxyServletException(HttpServletResponse.SC_BAD_REQUEST,
+                        "Parameter 'sitemap' must be provided!");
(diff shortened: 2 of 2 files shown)


lolodomo:
@maggu2810  for review.

lolodomo:
I don't apply the same fix to the blocking proxy servlet because it could lead to a real blocking...
So the change is applied only to the async proxy servlet.

lolodomo:
@resetnow for review too.

bundles/ui/org.eclipse.smarthome.ui/src/main/java/org/eclipse/smarthome/ui/internal/proxy/AsyncProxyServlet.java
```diff
@@ -51,6 +52,21 @@ protected HttpClient newHttpClient() {
...
     }
 
+    @Override
+    protected void sendProxyRequest(HttpServletRequest clientRequest, HttpServletResponse proxyResponse,
+            Request proxyRequest) {
+        if (service.proxyingVideoWidget(clientRequest)) {
+            // We disable the timeout for video
+            proxyRequest.timeout(0, TimeUnit.MILLISECONDS);
+
+            // We request the browser to not cache the video
```
maggu2810: The proxy request has not been sent yet.
Are you sure, that it is intended to change the proxy response header already?

Shouldn't the `cache-control` values set to the respective values for all request?
Makes it sense to cache any result?

lolodomo: I don't know if the implementation allows intercepting the answer to update its headers. My current implementation allows presetting headers; the headers from the source servers are apparently concatenated with them.

Regarding your question if we should avoid browser caching for images too, that is a good question. In fact all UIs running in browser use a workaround to avoid browser caching, they add a timestamp parameter to the proxied URL. So that could make sense to apply this change for all content (image and video).

kaikreuzer:
> @resetnow for review too.

@resetnow Short friendly ping :-)

resetnow:
Sorry for the delay.

Overall it should fix the problem for MJPEG videos, however, one thing that should be checked is whether or not the proxy service passes headers through by default.

>Shouldn't the cache-control values set to the respective values for all request?

I agree — this code here could help fix broken MJPEG streams and make user experience better, but overall it's the stream that should provide the headers and the proxy should relay them.

htreu:
@lolodomo friendly ping. Are you going to incorporate the latest comments?

lolodomo:
Please let me kmow what you want exactly to be changed.

htreu:
@lolodomo thanks for the feedback, if you all agree on this to be good then we are ready to merge I guess.
@maggu2810 please proceed :-)

maggu2810:
In general this PR looks got to me.

But what about the cache-conrol?
It seems that @lolodomo @resetnow and me considered this should be set for all requests (or misunderstood I some answer?). Currently this is only set for video widgets.

lolodomo:
All clients UI are already handling correctly the images through the proxy, without the need of this patch.
As the solution consisting in overriding the response headers is more a patch than a clean solution, I would like to apply it only for video.

PR #5183, opened 2018-03-05
 convert types before passing them to the handler

sjka:
...as so far types which were incompatible with the channel were forwarded to the binding anyway if they made it into the system.

Channels can indicate which types they accept indirectly by specifying an acceptedItemType.

This should enable bindings (like e.g. [KNX 2](https://github.com/openhab/openhab2-addons/pull/2323)) to make stronger assumptions about the possible state/command types they can expect to receive from the framework for any given channel. 

### DIFF
--- a/bundles/core/org.eclipse.smarthome.core.thing/src/main/java/org/eclipse/smarthome/core/thing/internal/CommunicationManager.java
+++ b/bundles/core/org.eclipse.smarthome.core.thing/src/main/java/org/eclipse/smarthome/core/thing/internal/CommunicationManager.java
@@ -104,6 +104,7 @@ public class CommunicationManager implements EventSubscriber, RegistryChangeList
     private SafeCaller safeCaller;
     @NonNullByDefault({})
     private ItemStateConverter itemStateConverter;
+    private final Set<ItemFactory> itemFactories = new CopyOnWriteArraySet<>();
 
     // link UID -> profile
     private final Map<String, Profile> profiles = new ConcurrentHashMap<>();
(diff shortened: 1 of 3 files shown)


sjka:
PS: Two remarks:

* The `Command` interface doesn't define a `Command as(Class<? extends Command> type)` method, hence the awkward `instanceof` checking and casting is [here](https://github.com/eclipse/smarthome/pull/5183/files#diff-1001ffb3900ef3d71b7af0a98671581fR340).
* While I believe that this change makes sense and helps making bindings more robust, I see that there is a risk that some workarounds or misusing scenarios might become broken because it will not possible anymore to forward "wrong" types to the bindings. I'm not aware of any, but people might be more creative than what the concepts anticipated.

maggu2810:
> I see that there is a risk that some workarounds or misusing scenarios might become broken because it will not possible anymore to forward "wrong" types to the bindings.

IMHO breaking wrong usage is not a problem and should be accepted.

bundles/core/org.eclipse.smarthome.core.thing/src/main/java/org/eclipse/smarthome/core/thing/internal/CommunicationManager.java
```diff
@@ -253,51 +293,63 @@ private void receiveCommand(ItemCommandEvent commandEvent) {
...
-                ThingHandler handler = thing.getHandler();
-                if (handler != null) {
-                    Profile profile = getProfile(link, item, thing);
-                    if (profile instanceof StateProfile) {
-                        safeCaller.create(((StateProfile) profile), StateProfile.class).withAsync()
-                                .withIdentifier(thing).withTimeout(THINGHANDLER_EVENT_TIMEOUT).build()
-                                .onCommandFromItem(command);
+                Channel channel = thing.getChannel(channelUID.getId());
+                if (channel != null) {
+                    @Nullable
```
maggu2810: Do we need this annotation? Shouldn't a local variable's nullness state be identified by the respective return value?

bundles/core/org.eclipse.smarthome.core.thing/src/main/java/org/eclipse/smarthome/core/thing/internal/CommunicationManager.java
```diff
@@ -470,6 +514,27 @@ public void unsetItemStateConverter(ItemStateConverter itemStateConverter) {
         this.itemStateConverter = null;
     }
 
+    @Reference(cardinality = ReferenceCardinality.AT_LEAST_ONE, policy = ReferencePolicy.DYNAMIC)
+    protected void addItemFactory(ItemFactory itemFactory) {
+        for (String itemTypeName : itemFactory.getSupportedItemTypes()) {
+            Item item = itemFactory.createItem(itemTypeName, "tmp");
+            if (item != null) {
+                acceptedCommandTypeMap.put(itemTypeName, item.getAcceptedCommandTypes());
+                acceptedStateTypeMap.put(itemTypeName, item.getAcceptedDataTypes());
```
maggu2810: Hm, you are using cardinality "at least one".
Your code in remove item factory removes the item type name from the respective maps.
So, the logic in removed acts as a item type name exists in one item factory only.
If two item factories use the same item type name, it exists only once in the map and is removed if one item factory is removed.
A simple solution would be to maintain a collection with all injected item factories and rebuild the maps all the time a factory is added and removed.

sjka: > So, the logic in removed acts as a item type name exists in one item factory only.

Yes, this was my assumption. I doubt though that there will be any deterministic outcome if multiple factories handle the same type. In any case you are right, it is safer that way.

PR #5184, opened 2018-03-05
Increase TestPortUtil reusability

wborn:
Move `TestPortUtil` from
`org.eclipse.smarthome.binding.fsinternetradio.test` to
`org.eclipse.smarthome.test` so it can be reused by other test bundles.

See also: https://github.com/openhab/openhab2-addons/pull/3150#issuecomment-370232976


wborn:
I've made the code/comments less specific now @maggu2810. 

We could also add a unit test that allocates hundreds of ports to see if the ports are really free and unique. But such tests usually create resource conflicts with other parallel running tests. WDYT?

maggu2810:
> We could also add a unit test that allocates hundreds of ports to see if the ports are really free and unique. 

I don't think that's necessary.

PR #5185, opened 2018-03-05
Optimize thing update handling

sjka:
only update things if they really changed, like it is done in the lazyRetryRunnable. This caused superfluous ThingUpdatedEvents when restarting a binding.

And while being there anyway I also fixed a bazillion of warnings by using the triple-operators for `null` comparisons. 


kaikreuzer:
```
Failed tests: 
  GenericThingProviderTest.assert that updating an embedded Thing causes the right behavior wrt adding and updating and no removed events are sent:501->OSGiTest.waitForAssert:-1->OSGiTest.waitForAssert:223->OSGiTest.waitForAssert:252 
Expected: is <2>
     but: was <1>
```

sjka:
Yes, I obviously missed that this test was relying on that... 
It actually brings up a good question: If a thing is updated which has a bridge, should there be a ThingUpdatedEvent for the bridge? My assumption is no. Any objections?

kaikreuzer:
No.

PR #4468, opened 2017-10-30
Solves [Automation] circular service reference #3635

adimova:
Signed-off-by: Ana Dimova <a.dimova@prosyst.bg>

Closes #3635

PR #5191, opened 2018-03-06
Added DISABLED ThingStatusDetail. Added persistence of ThingStatusInfos.

YordanDZhelev:
1. New thing staus detail introduced with value `DISABLED`.
2. `ThinngManager` interface created in package `org.eclipse.smarthome.core.thing` with one boolean method `setDisabled(Thing thing, boolean disabled)`.
3. `ThingManager` class renamed to `ThingManagerImpl` and now implements the new `ThingManager` interface. 
4. New method `boolean isEnabled()` introduced in `org.eclipse.smarthome.core.thing.Thing`.
5. Mechanism for persistance of thing statuses introduced .

Signed-off-by: YordanDZhelev <zhelev.yordan@gmail.com>


bundles/core/org.eclipse.smarthome.core.thing/src/main/java/org/eclipse/smarthome/core/thing/internal/ThingManagerImpl.java
```diff
@@ -1033,6 +1043,32 @@ private void setThingStatus(Thing thing, ThingStatusInfo thingStatusInfo) {
...
+            if (handler != null) {
+                handler.dispose();
+            }
+            setThingStatus(thing, buildStatusInfo(ThingStatus.UNINITIALIZED, ThingStatusDetail.DISABLED));
+        } else {
+            handler.initialize();
+        }
+
+        // Persist the status of the updated thing.
+        managedThingStatusInfoProvider.update(new ManagedThingStatusInfo(thing.getUID(), thing.getStatusInfo()));
```
tomhoefer: because dispose and initialize might have async implementations we cannot easily persist the status here... additionally we cannot set the DISABLED status after disposing the handler

bundles/core/org.eclipse.smarthome.core.thing/src/main/java/org/eclipse/smarthome/core/thing/internal/ThingManagerImpl.java
```diff
@@ -1033,6 +1043,32 @@ private void setThingStatus(Thing thing, ThingStatusInfo thingStatusInfo) {
...
+    }
+
+    private boolean isThingDisabled(Thing thing) {
+        ManagedThingStatusInfo persistedStatus = managedThingStatusInfoProvider.get(thing.getUID());
```
tomhoefer: Would it make sense to remove the status once a thing has been "really" removed from the system?

bundles/core/org.eclipse.smarthome.core.thing/src/main/java/org/eclipse/smarthome/core/thing/Thing.java
```diff
@@ -206,4 +207,10 @@
      */
     void setLocation(@Nullable String location);
 
+    /**
+     * Returns {@code true} if the thing is enabled. Return {@code false} otherwise.
+     * 
+     * @return Returns {@code true} if the thing is enabled. Return {@code false} otherwise.
+     */
+    boolean isEnabled();
```
tomhoefer: Is it expected this information is also exposed by the REST interface?
Is it expected to have also REST resources to enable / disable a thing?

bundles/core/org.eclipse.smarthome.core.thing/src/main/java/org/eclipse/smarthome/core/thing/ThingStatusDetail.java
```diff
@@ -35,7 +35,8 @@
      * Device has been removed. Used for example when the device has been removed from its bridge and
      * the thing handler should be removed.
      */
-    GONE;
+    GONE,
+    DISABLED;
```
tomhoefer: We should also document this new status in the official ESH documentation...

bundles/core/org.eclipse.smarthome.core.thing/src/main/java/org/eclipse/smarthome/core/thing/internal/ThingImpl.java
```diff
@@ -244,6 +245,11 @@ public void setLocation(@Nullable String location) {
         this.location = location;
     }
 
+    @Override
+    public boolean isEnabled() {
+        return ThingStatusDetail.DISABLED != getStatusInfo().getStatusDetail();
```
tomhoefer: This could result in a potential NPE

bundles/core/org.eclipse.smarthome.core.thing/src/main/java/org/eclipse/smarthome/core/thing/internal/ThingManagerImpl.java
sjka: Right, not the thing status - but the information that it should be disabled.

bundles/core/org.eclipse.smarthome.core.thing/src/main/java/org/eclipse/smarthome/core/thing/internal/ThingManagerImpl.java
sjka: How can you tell?

tomhoefer: To be honest ... I dont know ... The Problem I see is that a new created thing that was already previously part of the system would be disabled after re-paring by default

YordanDZhelev:
Thanks for the review @sjka. Could you please point me to what is not alligned with the rule engine PR from architecrual point of view, so I could update this PR?

sjka:
> Could you please point me to what is not alligned with the rule engine PR 

No, _I_ cannot (yet). @kaikreuzer wrote in #4468:

> If nobody vetos, I would try to refactor the automation component accordingly next week

There wasn't any commit since then on that PR, so to me it looks pretty much impossible to do that right now, isn't it? 😉 That's why I asked you to please align with _him_ (i.e. @kaikreuzer), not look at a PR where there is nothing much to see yet in that respect.

kaikreuzer:
>  I would try to refactor the automation component accordingly next week

I am a bit behind schedule as I fell ill last week, but it is high on my to-do list...

bundles/core/org.eclipse.smarthome.core.thing/src/main/java/org/eclipse/smarthome/core/thing/ThingManager.java
```diff
@@ -0,0 +1,37 @@
...
+     * {@code UNINITIALIZED.DISABLED}.
+     * If {@code disabled} parameter is set to {@code false} {@link ThingHandler#initialize()} will be invoked for the
+     * given {@link Thing}. Enabling is implemented as a non-blocking operation and the method will return without
+     * waiting for the {@link Thing}'s status to be updated.
+     * 
+     * @param thing the {@link Thing} to be disabled or enabled.
+     * @param disabled if set to {@code true} the corresponding thing will be disabled. If set to {@code false} the
+     *            thing will be enabled.
+     */
+    void setDisabled(final @NonNull Thing thing, final boolean disabled);
```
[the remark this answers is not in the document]
YordanDZhelev: If we decide that when the thing is NOT with status detail DISABLED, then, it can be considered as *enabled*, then I could rename the method to not have *disabled* in its name. Maybe just *enableThing()*.

bundles/core/org.eclipse.smarthome.core.thing/src/main/java/org/eclipse/smarthome/core/thing/internal/ThingManagerImpl.java
YordanDZhelev: It seems too early to persist the status indeed. As the disposal is delegated to the handler, we are not sure what will happen in each dispose() method and when the async disposal will finish. Anyway, the `dispose()` method of the handler might be an empty one, therefore with the current implementation there is no way to know if the dispose is finished or not. 
@sjka , @kaikreuzer what do you think about providing a defaoult implementaion or some sort of a callback in `BaseThingHandler`? If we want to be certain that the disposal is finished we havae to provide this information somehow.

sjka: The `ThingHandler.dispose()` method is called synchronously from the framework. Therefore you can safely persist the boolean flag that this thing should be disabled. Then, later on in `doUnregisterHandler(...)` you can use this information to set the thing to `UNINITIALIZED/DISABLED` instead of `UNINITIALIZED/HANDLER_MISSING_ERROR`.

YordanDZhelev:
PR updated regarding previous comments from @tomhoefer @sjka @kaikreuzer . Rest endpoint included for thing enabling.

YordanDZhelev:
PR rebased on the latest changes in the mastar brancyh, @sjka @kaikreuzer could you, please, take a look at the PR once again.

bundles/io/org.eclipse.smarthome.io.rest.core/src/main/java/org/eclipse/smarthome/io/rest/core/internal/thing/ThingResource.java
```diff
@@ -786,6 +832,15 @@ protected void unsetFirmwareUpdateService(FirmwareUpdateService firmwareUpdateSe
         this.firmwareUpdateService = null;
     }
 
+    @Reference(cardinality = ReferenceCardinality.OPTIONAL, policy = ReferencePolicy.DYNAMIC)
+    protected void setThingManager(ThingManager thingManager) {
```
kaikreuzer: Don't forget to also check it in `isSatisfied`

bundles/io/org.eclipse.smarthome.io.rest.core/src/main/java/org/eclipse/smarthome/io/rest/core/internal/thing/ThingResource.java
```diff
@@ -461,6 +463,50 @@ public Response getStatus(@HeaderParam(HttpHeaders.ACCEPT_LANGUAGE) String langu
...
+            @PathParam("thingUID") @ApiParam(value = "thing") String thingUID,
+            @ApiParam(value = "enabled") String enabled) throws IOException {
+        final Locale locale = LocaleUtil.getLocale(language);
+
+        ThingUID thingUIDObject = new ThingUID(thingUID);
+
+        // Check if the Thing exists, 404 if not
+        Thing thing = thingRegistry.get(thingUIDObject);
+        if (null == thing) {
+            logger.info("Received HTTP PUT request for disable thing at '{}' for the unknown thing '{}'.",
```
kaikreuzer: -> for set enabled at....

bundles/io/org.eclipse.smarthome.io.rest.core/src/main/java/org/eclipse/smarthome/io/rest/core/internal/thing/ThingResource.java
```diff
@@ -461,6 +463,50 @@ public Response getStatus(@HeaderParam(HttpHeaders.ACCEPT_LANGUAGE) String langu
...
+                    uriInfo.getPath(), thingUID);
+            return getThingNotFoundResponse(thingUID);
+        }
+
+        // ask whether the Thing exists as a managed thing, so it can get
```
kaikreuzer: Why this check? We can also disable unmanaged things as this information is not persisted with the Thing entities!

bundles/io/org.eclipse.smarthome.io.rest.core/src/main/java/org/eclipse/smarthome/io/rest/core/internal/thing/ThingResource.java
```diff
@@ -461,6 +463,50 @@ public Response getStatus(@HeaderParam(HttpHeaders.ACCEPT_LANGUAGE) String langu
...
+        // updated, 409 otherwise
+        if (managedThingProvider.get(thingUIDObject) == null) {
+            logger.info("Received HTTP PUT request for update configuration at '{}' for an unmanaged thing '{}'.",
+                    uriInfo.getPath(), thingUID);
+            return getThingResponse(Status.CONFLICT, thing, locale,
+                    "Cannot update Thing " + thingUID + " as it is not editable.");
+        }
+
+        if(thingManager == null) {
```
kaikreuzer: This can never happen - remove this check.

bundles/core/org.eclipse.smarthome.core.thing/src/main/java/org/eclipse/smarthome/core/thing/ThingManager.java
```diff
@@ -0,0 +1,31 @@
...
+ */
+public interface ThingManager {
+
+    /**
+     * Enable or disable the thing.
+     * 
+     * @param thingUID the {@link ThingUID} uniquely describing the {@link Thing} to be enabled or disabled.
+     * @param enabled  {@code true} if the thing should be enabled, {@code false} if the thing should be disabled.
+     */
+    void enableThing(final @NonNull ThingUID thingUID, final boolean enabled);
```
kaikreuzer: For maximum alignment with the RuleManager, I would suggest to define exactly [these two methods](https://github.com/kaikreuzer/smarthome/blob/circular_dependencies/bundles/automation/org.eclipse.smarthome.automation.api/src/main/java/org/eclipse/smarthome/automation/RuleManager.java#L30-L52) here as well - feel free to copy it over (but keep using ThingUID instead of a String).

bundles/core/org.eclipse.smarthome.core.thing/src/main/java/org/eclipse/smarthome/core/thing/internal/ManagedThingStatusInfo.java
```diff
@@ -0,0 +1,37 @@
...
+ * SPDX-License-Identifier: EPL-2.0
+ */
+package org.eclipse.smarthome.core.thing.internal;
+
+/**
+ * {@link ManagedThingStatusInfo} is used for persistence of information regarding thing's status.
+ * 
+ * @author Yordan Zhelev - Initial contribution
+ */
+public class ManagedThingStatusInfo {
```
kaikreuzer: We do not need a class to persist a single boolean flag - please remove it.

bundles/core/org.eclipse.smarthome.core.thing/src/main/java/org/eclipse/smarthome/core/thing/internal/ThingManagerImpl.java
```diff
@@ -145,6 +151,8 @@
     private final Set<String> loadedXmlThingTypes = new CopyOnWriteArraySet<>();
     private SafeCaller safeCaller;
     private volatile boolean active = false;
+    private StorageService storageService;
+    private volatile Storage<ManagedThingStatusInfo> storage;
```
kaikreuzer: This can be a `Storage<Boolean>` see [here](https://github.com/kaikreuzer/smarthome/blob/circular_dependencies/bundles/automation/org.eclipse.smarthome.automation.core/src/main/java/org/eclipse/smarthome/automation/core/internal/RuleEngineImpl.java#L147) - and not need to declare it as volatile.

bundles/core/org.eclipse.smarthome.core.thing/src/main/java/org/eclipse/smarthome/core/thing/internal/ThingManagerImpl.java
```diff
@@ -599,13 +607,20 @@ private void initializeHandler(Thing thing) {
             }
             ThingType thingType = getThingType(thing);
             applyDefaultConfiguration(thing, thingType);
-            if (isInitializable(thing, thingType)) {
-                setThingStatus(thing, buildStatusInfo(ThingStatus.INITIALIZING, ThingStatusDetail.NONE));
-                doInitializeHandler(thing.getHandler());
+
+            if (isPersistedAsEnabled(thing.getUID())) {
```
kaikreuzer: This should simply be an `isEnabled` call.

bundles/core/org.eclipse.smarthome.core.thing/src/main/java/org/eclipse/smarthome/core/thing/internal/ThingManagerImpl.java
```diff
@@ -599,13 +607,20 @@ private void initializeHandler(Thing thing) {
...
+                            thing.getUID());
+                    setThingStatus(thing, buildStatusInfo(ThingStatus.UNINITIALIZED,
+                            ThingStatusDetail.HANDLER_CONFIGURATION_PENDING));
+                }
             } else {
-                logger.debug("Thing '{}' not initializable, check required configuration parameters.", thing.getUID());
-                setThingStatus(thing,
-                        buildStatusInfo(ThingStatus.UNINITIALIZED, ThingStatusDetail.HANDLER_CONFIGURATION_PENDING));
+                // the thing is persisted as disabled. Will not attempt to initialize its handler.
+                setThingStatus(thing, new ThingStatusInfo(ThingStatus.UNINITIALIZED, ThingStatusDetail.DISABLED, null));
```
kaikreuzer: It should have been in exactly this status before already - so no need to set it here.

bundles/core/org.eclipse.smarthome.core.thing/src/main/java/org/eclipse/smarthome/core/thing/internal/ThingManagerImpl.java
```diff
@@ -929,7 +946,7 @@ protected synchronized void activate(ComponentContext componentContext) {
         readyService.registerTracker(this, new ReadyMarkerFilter().withType(XML_THING_TYPE));
         for (ThingHandlerFactory factory : thingHandlerFactories) {
             handleThingHandlerFactoryAddition(getBundleName(factory));
-        }
+    }
```
kaikreuzer: What has happened here?

bundles/core/org.eclipse.smarthome.core.thing/src/main/java/org/eclipse/smarthome/core/thing/internal/ThingManagerImpl.java
```diff
@@ -939,8 +956,8 @@ protected synchronized void addThingHandlerFactory(ThingHandlerFactory thingHand
         logger.debug("Thing handler factory '{}' added", thingHandlerFactory.getClass().getSimpleName());
         thingHandlerFactories.add(thingHandlerFactory);
         if (active) {
-            handleThingHandlerFactoryAddition(getBundleName(thingHandlerFactory));
-        }
+        handleThingHandlerFactoryAddition(getBundleName(thingHandlerFactory));
+    }
```
kaikreuzer: What has happened here?

bundles/core/org.eclipse.smarthome.core.thing/src/main/java/org/eclipse/smarthome/core/thing/internal/ThingManagerImpl.java
```diff
@@ -1018,7 +1035,7 @@ protected synchronized void deactivate(ComponentContext componentContext) {
         thingRegistry.removeThingTracker(this);
         for (ThingHandlerFactory factory : thingHandlerFactories) {
             removeThingHandlerFactory(factory);
-        }
+    }
```
kaikreuzer: What has happened here?

bundles/core/org.eclipse.smarthome.core.thing/src/main/java/org/eclipse/smarthome/core/thing/internal/ThingManagerImpl.java
```diff
@@ -1111,6 +1128,61 @@ private void setThingStatus(Thing thing, ThingStatusInfo thingStatusInfo) {
         }
     }
 
+    @Override
+    public void enableThing(ThingUID thingUID, boolean enabled) {
```
kaikreuzer: Check if you can do it as closely as possible to [the rule engine](https://github.com/kaikreuzer/smarthome/blob/circular_dependencies/bundles/automation/org.eclipse.smarthome.automation.core/src/main/java/org/eclipse/smarthome/automation/core/internal/RuleEngineImpl.java#L873-L889) - it currently looks wrong to me that you get hold of the `thingHandler` as this should actually never be set for a disabled thing.

bundles/core/org.eclipse.smarthome.core.thing/src/main/java/org/eclipse/smarthome/core/thing/internal/ThingManagerImpl.java
```diff
@@ -1111,6 +1128,61 @@ private void setThingStatus(Thing thing, ThingStatusInfo thingStatusInfo) {
...
+     * <p>
+     * <ul>
+     * <li>case {@code true} the record for that key will be removed from the storage. Thing should
+     * be considered as enabled.
+     * <li>case {@code false} a boolean value false with a key the UID of the thing will be
+     * persisted. Thing should be considered as explicitly disabled.
+     * </ul>
+     * <p>
+     */
+    private void persistEnabledStatusForThing(final ThingUID thingUID, boolean enabled) {
```
kaikreuzer: Not sure if this is worth a separate method - especially as it even requires a parameter to differentiate between the two lines that could be executed.

bundles/core/org.eclipse.smarthome.core.thing/src/main/java/org/eclipse/smarthome/core/thing/internal/ThingManagerImpl.java
```diff
@@ -1111,6 +1128,61 @@ private void setThingStatus(Thing thing, ThingStatusInfo thingStatusInfo) {
...
+        // Persist only disabled things. Thing with no corresponding persisted value will be considered as enabled.
+        if (enabled) {
+            storage.remove(thingUID.getAsString());
+        } else {
+            storage.put(thingUID.getAsString(), new ManagedThingStatusInfo(enabled));
+        }
+    }
+
+    private boolean isPersistedAsEnabled(ThingUID thingUID) {
+        ManagedThingStatusInfo persistedStatus = storage.get(thingUID.getAsString());
```
kaikreuzer: This method should simply be based on the ThingStatusDetail, see the [rule engine as an example](https://github.com/kaikreuzer/smarthome/blob/circular_dependencies/bundles/automation/org.eclipse.smarthome.automation.core/src/main/java/org/eclipse/smarthome/automation/core/internal/RuleEngineImpl.java#L910-L913).
You should imho add a check in the `thingAdded` method, whether there is a disabled flag persisted in the storage and set the Thing status there already (and avoid a call of `registerAndInitializeHandler` at the same time.

kaikreuzer:
> What is missing is the adaption of the EnrichedThingDTO, which should also have an enabled flag

Update: After discussion with @htreu, we decided that this "enabled" flag is problematic as it is not sent through status update events. Adding it there is ugly and introducing a new event seems inadequate, too. So after all, the UIs should simply check for status detail "disabled" and we do not require the enabled flag in the DTO.

bundles/core/org.eclipse.smarthome.core.thing/src/main/java/org/eclipse/smarthome/core/thing/internal/ThingManagerImpl.java
YordanDZhelev: The public method `isEnabled(ThingUID)` will be with the same signature, so will have to keep this one as it is, or we can think of some other name.

YordanDZhelev:
PR updated in regard to @kaikreuzer comments.

tomhoefer:
@sjka / @kaikreuzer would you be so kind to review again?

bundles/core/org.eclipse.smarthome.core.thing/src/main/java/org/eclipse/smarthome/core/thing/internal/ThingManagerImpl.java
```diff
@@ -1117,6 +1142,34 @@ private void setThingStatus(Thing thing, ThingStatusInfo thingStatusInfo) {
...
+                storage.remove(thingUID.getAsString());
+            }
+            setThingStatus(thing, buildStatusInfo(ThingStatus.UNINITIALIZED, ThingStatusDetail.NONE));
+            registerAndInitializeHandler(thing, findThingHandlerFactory(thing.getThingTypeUID()));
+        } else {
+            logger.debug("Thing {} will be disabled.", thing.getUID().getAsString());
+            if (storage != null) {
+                storage.put(thingUID.getAsString(), enabled);
+            }
+            setThingStatus(thing, buildStatusInfo(ThingStatus.UNINITIALIZED, ThingStatusDetail.DISABLED));
```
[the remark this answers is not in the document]
YordanDZhelev: Here https://github.com/eclipse/smarthome/pull/5191/files#diff-5523918cb0a734afd589ef6edfeeced8R779 we rely on `thing.isEnabled()` to determne the status of the thing. If we remove line 1162, then in `doUnregisterHandler()` have to check the storage once again. Or to leave line 1162 and add additional code in `doUnregisterHandler()` to not set the same status twice.

bundles/core/org.eclipse.smarthome.core.thing/src/main/java/org/eclipse/smarthome/core/thing/internal/ThingManagerImpl.java
```diff
@@ -401,11 +410,19 @@ public void thingAdded(Thing thing, ThingTrackerEvent thingTrackerEvent) {
         this.things.add(thing);
         logger.debug("Thing '{}' is tracked by ThingManager.", thing.getUID());
 
+        // Disable thing if persisted as disabled
+        if (/* Storage is available */storage != null
```
sjka: please remove these ugly comments - they just make the code hard to read

bundles/core/org.eclipse.smarthome.core.thing/src/main/java/org/eclipse/smarthome/core/thing/internal/ThingManagerImpl.java
```diff
@@ -401,11 +410,19 @@ public void thingAdded(Thing thing, ThingTrackerEvent thingTrackerEvent) {
...
+                && /* Record for this thing is available */storage.get(thing.getUID().getAsString()) != null
```
sjka: anything speaking against using `storage.containsKey(...)`?

bundles/core/org.eclipse.smarthome.core.thing/src/main/java/org/eclipse/smarthome/core/thing/internal/ThingManagerImpl.java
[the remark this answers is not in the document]
sjka: You're setting here the thing-status to UNINITIALIZED **before** the handler actually was disposed. This completely messes up the lifecycle, therefore it is a NO-GO! Don't do it here.

>  [...] then in doUnregisterHandler() have to check the storage once again.

Yes, please do so.

bundles/core/org.eclipse.smarthome.core.thing/src/main/java/org/eclipse/smarthome/core/thing/internal/ThingManagerImpl.java
```diff
@@ -1117,6 +1142,34 @@ private void setThingStatus(Thing thing, ThingStatusInfo thingStatusInfo) {
...
+            }
+            setThingStatus(thing, buildStatusInfo(ThingStatus.UNINITIALIZED, ThingStatusDetail.NONE));
+            registerAndInitializeHandler(thing, findThingHandlerFactory(thing.getThingTypeUID()));
+        } else {
+            logger.debug("Thing {} will be disabled.", thing.getUID().getAsString());
+            if (storage != null) {
+                storage.put(thingUID.getAsString(), enabled);
+            }
+            setThingStatus(thing, buildStatusInfo(ThingStatus.UNINITIALIZED, ThingStatusDetail.DISABLED));
+            unregisterAndDisposeHandler(findThingHandlerFactory(thing.getThingTypeUID()), thing, thing.getHandler());
```
sjka: What if the thing doesn't have a handler at this point in time? 😉

YordanDZhelev:
PR rebased on current master. @kaikreuzer should I consider it ready to squash commits, or better wait for @sjka approval?

YordanDZhelev:
@sjka could you please check once again if you havce more comments. PR is rebase on master several days ago.

bundles/core/org.eclipse.smarthome.core.thing/src/main/java/org/eclipse/smarthome/core/thing/internal/ThingManagerImpl.java
```diff
@@ -1116,6 +1144,40 @@ private void setThingStatus(Thing thing, ThingStatusInfo thingStatusInfo) {
...
+        } else {
+            logger.debug("Thing {} will be disabled.", thing.getUID().getAsString());
+            if (storage != null) {
+                storage.put(thingUID.getAsString(), enabled);
+            }
+            ThingHandlerFactory thingHandlerFactory = findThingHandlerFactory(thing.getThingTypeUID());
+            if (thing.getHandler() != null) {
+                unregisterAndDisposeHandler(thingHandlerFactory, thing, thing.getHandler());
+            } else {
+                thingHandlerFactory.unregisterHandler(thing);
```
sjka: You are calling `unregisterHandler(...)` here directly on the thing handler factory. You must not do that. There is a `ThingManager.unregisterHandler(...)` method which you should use instead as it does a lot more than that internally.

bundles/core/org.eclipse.smarthome.core.thing/src/main/java/org/eclipse/smarthome/core/thing/internal/ThingManagerImpl.java
```diff
@@ -1116,6 +1144,40 @@ private void setThingStatus(Thing thing, ThingStatusInfo thingStatusInfo) {
...
+    public void setEnabled(ThingUID thingUID, boolean enabled) {
+        Thing thing = getThing(thingUID);
+
+        if (enabled) {
+            // Enable a thing
+            logger.debug("Thing {} will be enabled.", thing.getUID().getAsString());
+            if (storage != null) {
+                storage.remove(thingUID.getAsString());
+            }
+            if (!thing.getStatus().equals(ThingStatus.ONLINE)) {
```
sjka: What is the purpose of this condition? Do you maybe mean `if (!ThingHandlerHelper.isHandlerInitialized(thing)) {...}`?

bundles/core/org.eclipse.smarthome.core.thing/src/main/java/org/eclipse/smarthome/core/thing/internal/ThingManagerImpl.java
```diff
@@ -1116,6 +1144,40 @@ private void setThingStatus(Thing thing, ThingStatusInfo thingStatusInfo) {
...
+                setThingStatus(thing, buildStatusInfo(ThingStatus.UNINITIALIZED, ThingStatusDetail.NONE));
```
sjka: This look like pretty broad big-hammer-strategy to me. What happens if e.g.
* the thing was e.g. in `REMOVING` before?
* the thing was e.g. in `OFFLINE`/`COMMUNICATION_ERROR` before? 
* somebody calls `setEnabled(..., true) on an already enabled thing?

bundles/core/org.eclipse.smarthome.core.thing/src/main/java/org/eclipse/smarthome/core/thing/internal/ThingManagerImpl.java
```diff
@@ -400,11 +409,17 @@ public void thingAdded(Thing thing, ThingTrackerEvent thingTrackerEvent) {
         this.things.add(thing);
         logger.debug("Thing '{}' is tracked by ThingManager.", thing.getUID());
 
+        if (storage != null && storage.containsKey(thing.getUID().getAsString())
+                && !storage.get(thing.getUID().getAsString())) {
+            setEnabled(thing.getUID(), false);
+        }
+
```
sjka: The `setEnabled(...)`-method calls `registerAndInitializeHandler(...)` for enabled things. When looking at this from this place in the code:
* it is pretty confusing when looking at the method name "setEnabled", don't you think so?
* Isn't the below code useless now?

bundles/core/org.eclipse.smarthome.core.thing/src/main/java/org/eclipse/smarthome/core/thing/internal/ThingManagerImpl.java
```diff
@@ -1116,6 +1144,40 @@ private void setThingStatus(Thing thing, ThingStatusInfo thingStatusInfo) {
...
+                setThingStatus(thing, buildStatusInfo(ThingStatus.UNINITIALIZED, ThingStatusDetail.NONE));
+                registerAndInitializeHandler(thing, findThingHandlerFactory(thing.getThingTypeUID()));
+            }
+        } else {
+            logger.debug("Thing {} will be disabled.", thing.getUID().getAsString());
+            if (storage != null) {
+                storage.put(thingUID.getAsString(), enabled);
+            }
+            ThingHandlerFactory thingHandlerFactory = findThingHandlerFactory(thing.getThingTypeUID());
+            if (thing.getHandler() != null) {
```
sjka: I think the more meaningful method in this case would be `isHandlerRegistered(...)` as it is the primary source for that information.

sjka: You have created a funny little asymmetry here: While `unregisterAndDisposeHandler(...)` also unregisters and disposes all child handlers in case it is a bridge, your `else` block enters the game at a place where this does not happen anymore. This leads to quite surprising side-effects, I reckon.

bundles/core/org.eclipse.smarthome.core.thing/src/main/java/org/eclipse/smarthome/core/thing/internal/ThingManagerImpl.java
YordanDZhelev: !thing.getStatus().equals(ThingStatus.ONLINE) changed to !ThingHandlerHelper.isHandlerInitialized(thing) as suggested, this will prevent the attempt to initialize already initialized thing. Otherwise will change the storage entry and try to initialize the thing.

bundles/core/org.eclipse.smarthome.core.thing/src/main/java/org/eclipse/smarthome/core/thing/internal/ThingManagerImpl.java
YordanDZhelev: thing.getHandler() != null changed to isHandlerRegistered(thing) as suggested.

YordanDZhelev: The for checking the status of the thing is moved to ```registerAndInitializeHandler(...)``` method. This  also covers ```thingUpdated(...)``` entry point.

kaikreuzer:
In https://github.com/eclipse/smarthome/pull/5942, @maggu2810 actually convinced us that the `isEnabled`method on the entity itself is a bad idea and that we should only have it in the manager service, resp. the callback. I have updated the [comparison list](https://github.com/eclipse/smarthome/pull/4468#issuecomment-367781063) accordingly.
To stick to the goal to align Things and Rules, I would suggest to change this PR the same way as well - this should have little impact as the method already exists on the ThingManager and we would mainly remove it on the Thing itself. Wdyt?

tomhoefer:
yes - I think this makes sense

before we change the PR accordingly I suggest to wait for @sjka `s review feedback

bundles/core/org.eclipse.smarthome.core.thing/src/main/java/org/eclipse/smarthome/core/thing/internal/ThingManagerImpl.java
```diff
@@ -1116,6 +1143,38 @@ private void setThingStatus(Thing thing, ThingStatusInfo thingStatusInfo) {
         }
     }
 
+    @Override
+    public void setEnabled(ThingUID thingUID, boolean enabled) {
+        Thing thing = getThing(thingUID);
+
+        if (enabled) {
+            // Enable a thing
+            logger.debug("Thing {} will be enabled.", thing.getUID().getAsString());
```
sjka: minor optimization: `thing.getUID().getAsString()` -> `thingUID`

bundles/core/org.eclipse.smarthome.core.thing/src/main/java/org/eclipse/smarthome/core/thing/internal/ThingManagerImpl.java
```diff
@@ -1116,6 +1143,38 @@ private void setThingStatus(Thing thing, ThingStatusInfo thingStatusInfo) {
...
+            if (storage != null) {
+                storage.remove(thingUID.getAsString());
+            }
+            if (!ThingHandlerHelper.isHandlerInitialized(thing)) {
+                setThingStatus(thing, buildStatusInfo(ThingStatus.UNINITIALIZED, ThingStatusDetail.NONE));
+                registerAndInitializeHandler(thing, findThingHandlerFactory(thing.getThingTypeUID()));
+            }
+        } else {
+            logger.debug("Thing {} will be disabled.", thing.getUID().getAsString());
```
sjka: minor optimization: `thing.getUID().getAsString()` -> `thingUID`

bundles/core/org.eclipse.smarthome.core.thing/src/main/java/org/eclipse/smarthome/core/thing/internal/ThingManagerImpl.java
```diff
@@ -1116,6 +1143,38 @@ private void setThingStatus(Thing thing, ThingStatusInfo thingStatusInfo) {
...
+    public void setEnabled(ThingUID thingUID, boolean enabled) {
+        Thing thing = getThing(thingUID);
+
+        if (enabled) {
+            // Enable a thing
+            logger.debug("Thing {} will be enabled.", thing.getUID().getAsString());
+            if (storage != null) {
+                storage.remove(thingUID.getAsString());
+            }
+            if (!ThingHandlerHelper.isHandlerInitialized(thing)) {
```
sjka: You didn't answer [my question](https://github.com/eclipse/smarthome/pull/5191#discussion_r201306064) about the purpose if this if-statement but just changed it to this. 

I assume you want to prevent initializing an already existing thing handler for the second time? 

If that's the case and this really _could_ happen, then I see a problem here: `isHandlerInitialized(...)` only tells you if the initialization phase of a potentially existing handler is _finished_  or not. It doesn't tell you if there already exists a handler which might still be INITIALIZING. Therefore this will lead to awkward race conditions, at worst leaving two handlers working on the same device. For that matter, `isHandlerRegistered(...)` (which you have used below too) would have been the better choice, I think.

On the other hand, if you follow the code flow a little, you will see in `registerHandler(...)` that the code already _is_ protected against such a race-condition itself. Also, `initializeHandler(...)` is protected against such a scenario. Leaving ugly warnings in the log though - so it might still be good to do have this kind of protection here too. 

Nevertheless, the most "dangerous" part in this context is setting the thing status to `UNINITIALIZED`/`NONE` blindly, regardless of its previous status. Don't you think it would be better to check for the previous status first and make sure that it _really_ is `UNINITIALIZED`/`DISABLED` and _only then_ change it to `UNINITIALIZED`/`NONE` and call `registerAndInitializeHandler(...)`?

bundles/core/org.eclipse.smarthome.core.thing/src/main/java/org/eclipse/smarthome/core/thing/internal/ThingManagerImpl.java
```diff
@@ -1116,6 +1143,38 @@ private void setThingStatus(Thing thing, ThingStatusInfo thingStatusInfo) {
...
+                setThingStatus(thing, buildStatusInfo(ThingStatus.UNINITIALIZED, ThingStatusDetail.NONE));
+                registerAndInitializeHandler(thing, findThingHandlerFactory(thing.getThingTypeUID()));
+            }
+        } else {
+            logger.debug("Thing {} will be disabled.", thing.getUID().getAsString());
+            if (storage != null) {
+                storage.put(thingUID.getAsString(), enabled);
+            }
+            ThingHandlerFactory thingHandlerFactory = findThingHandlerFactory(thing.getThingTypeUID());
+            if (isHandlerRegistered(thing)) {
```
sjka: You wrote the new enablement status into the storage and call `unregisterAndDisposeHandler(...)` below, if a handler exists. You rely on `doUnregisterHandler(...)` to set the thing to `UNINITIALIZED`/`DISABLED`. 

However, what if there is not handler, e.g. because of a problem leading to `UNINITIALIZED`/`HANDLER_MISSING_ERROR`? Then suddenly the disabled-state will not be reflected anywhere.

bundles/core/org.eclipse.smarthome.core.thing/src/main/java/org/eclipse/smarthome/core/thing/internal/ThingManagerImpl.java
```diff
@@ -1116,6 +1143,38 @@ private void setThingStatus(Thing thing, ThingStatusInfo thingStatusInfo) {
...
+        Thing thing = getThing(thingUID);
+
+        if (enabled) {
+            // Enable a thing
+            logger.debug("Thing {} will be enabled.", thing.getUID().getAsString());
+            if (storage != null) {
+                storage.remove(thingUID.getAsString());
+            }
+            if (!ThingHandlerHelper.isHandlerInitialized(thing)) {
+                setThingStatus(thing, buildStatusInfo(ThingStatus.UNINITIALIZED, ThingStatusDetail.NONE));
```
sjka: All other handler-lifecycle-related thing status changes are guarded by a thing specific lock in order to prevent any race conditions. It somehow feels wrong to not do so here, doesn't it?

bundles/core/org.eclipse.smarthome.core.thing/src/main/java/org/eclipse/smarthome/core/thing/internal/ThingManagerImpl.java
YordanDZhelev: My assumption here was that, when we call `setEnabled(thingUID, true)` we want to bring the thing to `ONLINE` status regardless of its previous status. That is why the condition here was `if (!thing.getStatus().equals(ThingStatus.ONLINE))` before my last commit. So, besides your comments on the code, does this sounds like the correct approach to you?

bundles/core/org.eclipse.smarthome.core.thing/src/main/java/org/eclipse/smarthome/core/thing/internal/ThingManagerImpl.java
YordanDZhelev: It will not be reflected in the thing's status, but the storage will keep the information for that thing. Later when `registerAndInitializeHandler(...)` is called to register handler for that thing the information in the storage will be taken in consideration.

YordanDZhelev:
@sjka could you please check the [tests](https://github.com/eclipse/smarthome/pull/5191/files#diff-b0466ced3dcf55c156cde959a38b5a31) if some scenrio is not covered yet.

bundles/core/org.eclipse.smarthome.core.thing/src/main/java/org/eclipse/smarthome/core/thing/ThingManager.java
```diff
@@ -0,0 +1,43 @@
...
+     * This method is used for changing <b>enabled</b> state of the {@link Thing}
+     * The only {@link ThingStatus} which is NOT enabled is {@link ThingStatus} with
+     * {@link ThingStatusDetail#DISABLED}.
+     *
+     * @param thingUID UID of the {@link Thing}.
+     * @param isEnabled a new <b>enabled / disabled</b> state of the {@link Thing}.
+     */
+    public void setEnabled(ThingUID thingUID, boolean isEnabled);
+
+}
```
htreu: missing new line

bundles/core/org.eclipse.smarthome.core.thing/src/main/java/org/eclipse/smarthome/core/thing/ThingStatusDetail.java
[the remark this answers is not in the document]
htreu: this one is still open.

bundles/core/org.eclipse.smarthome.core.thing/src/main/java/org/eclipse/smarthome/core/thing/internal/ThingManagerImpl.java
```diff
@@ -420,14 +429,13 @@ private void waitForRunningHandlerRegistrations(ThingUID thingUID) {
...
         logger.debug("Thing '{}' is tracked by ThingManager.", thing.getUID());
-
         if (!isHandlerRegistered(thing)) {
             registerAndInitializeHandler(thing, getThingHandlerFactory(thing));
         } else {
             logger.debug("Handler of tracked thing '{}' already registered.", thing.getUID());
         }
     }
-
+
```
htreu: obsolete spaces introduced

bundles/core/org.eclipse.smarthome.core.thing/src/main/java/org/eclipse/smarthome/core/thing/internal/ThingManagerImpl.java
```diff
@@ -625,13 +640,15 @@ private void initializeHandler(Thing thing) {
...
                 setThingStatus(thing, buildStatusInfo(ThingStatus.INITIALIZING, ThingStatusDetail.NONE));
                 doInitializeHandler(thing.getHandler());
             } else {
-                logger.debug("Thing '{}' not initializable, check required configuration parameters.", thing.getUID());
-                setThingStatus(thing,
-                        buildStatusInfo(ThingStatus.UNINITIALIZED, ThingStatusDetail.HANDLER_CONFIGURATION_PENDING));
+                logger.debug("Thing '{}' not initializable, check required configuration parameters.",
+                        thing.getUID());
+                setThingStatus(thing, buildStatusInfo(ThingStatus.UNINITIALIZED,
+                        ThingStatusDetail.HANDLER_CONFIGURATION_PENDING));
```
htreu: lines 648 - 651 do not seem to have been changed. please restore the original format.
it seems your formatter is not ESH default, please check and adopt.

bundles/core/org.eclipse.smarthome.core.thing/src/main/java/org/eclipse/smarthome/core/thing/internal/ThingManagerImpl.java
```diff
@@ -1146,6 +1172,66 @@ private void setThingStatus(Thing thing, ThingStatusInfo thingStatusInfo) {
...
+        Thing thing = getThing(thingUID);
+        if (enabled) {
+            // Enable a thing
+            // Clear the disabled thing storage. Otherwise the handler will NOT be initialized later.
+            if (storage != null) {
+                storage.remove(thingUID.getAsString());
+            }
+
+            if (thing.getStatus().equals(ThingStatus.ONLINE)) {
+
```
htreu: remove empty line

bundles/core/org.eclipse.smarthome.core.thing/src/main/java/org/eclipse/smarthome/core/thing/internal/ThingManagerImpl.java
```diff
@@ -1146,6 +1172,66 @@ private void setThingStatus(Thing thing, ThingStatusInfo thingStatusInfo) {
...
+            if (thing.getStatus().equals(ThingStatus.ONLINE)) {
+                
+                logger.debug("Thing {} is already in the required state.", thingUID);
+                return;
+            }
+
+            logger.debug("Thing {} will be enabled.", thingUID);
+
+            if (isHandlerRegistered(thing)) {
+
```
htreu: remove empty line

bundles/core/org.eclipse.smarthome.core.thing/src/main/java/org/eclipse/smarthome/core/thing/internal/ThingManagerImpl.java
```diff
@@ -1146,6 +1172,66 @@ private void setThingStatus(Thing thing, ThingStatusInfo thingStatusInfo) {
...
+            }
+
+            logger.debug("Thing {} will be enabled.", thingUID);
+
+            if (isHandlerRegistered(thing)) {
+                
+                // A handler is already registered for that thing. Try to initialize it.
+                initializeHandler(thing);
+            } else {
+
```
htreu: remove empty line

bundles/core/org.eclipse.smarthome.core.thing/src/main/java/org/eclipse/smarthome/core/thing/internal/ThingManagerImpl.java
```diff
@@ -1146,6 +1172,66 @@ private void setThingStatus(Thing thing, ThingStatusInfo thingStatusInfo) {
...
+                registerAndInitializeHandler(thing, findThingHandlerFactory(thing.getThingTypeUID()));
+            }
+        } else {
+            // Mark the thing as disabled in the storage.
+            if (storage != null) {
+                storage.put(thingUID.getAsString(), enabled);
+            }
+
+            if (!thing.isEnabled()) {
+
```
htreu: remove empty line

bundles/core/org.eclipse.smarthome.core.thing/src/main/java/org/eclipse/smarthome/core/thing/internal/ThingManagerImpl.java
```diff
@@ -1146,6 +1172,66 @@ private void setThingStatus(Thing thing, ThingStatusInfo thingStatusInfo) {
...
+            if (!thing.isEnabled()) {
+                
+                logger.debug("Thing {} is already in the required state.", thingUID);
+                return;
+            }
+
+            logger.debug("Thing {} will be disabled.", thingUID);
+
+            if (isHandlerRegistered(thing)) {
+
```
htreu: remove empty line

bundles/core/org.eclipse.smarthome.core.thing/src/main/java/org/eclipse/smarthome/core/thing/internal/ThingManagerImpl.java
```diff
@@ -1146,6 +1172,66 @@ private void setThingStatus(Thing thing, ThingStatusInfo thingStatusInfo) {
...
+            logger.debug("Thing {} will be disabled.", thingUID);
+
+            if (isHandlerRegistered(thing)) {
+                
+                // Dispose handler if registered.
+                ThingHandlerFactory thingHandlerFactory = findThingHandlerFactory(thing.getThingTypeUID());
+                unregisterAndDisposeHandler(thingHandlerFactory, thing, thing.getHandler());
+            } else {
+
```
htreu: remove empty line

bundles/core/org.eclipse.smarthome.core.thing/src/main/java/org/eclipse/smarthome/core/thing/internal/ThingManagerImpl.java
```diff
@@ -1202,4 +1288,18 @@ protected void unsetSafeCaller(SafeCaller safeCaller) {
...
+        }
+    }
+
+    protected void unsetStorageService(StorageService storageService) {
+        if (this.storageService == storageService) {
+            this.storageService = null;
+            this.storage = null;
+        }
+    }
+}
```
htreu: missing new line.

bundles/io/org.eclipse.smarthome.io.rest.core/src/main/java/org/eclipse/smarthome/io/rest/core/internal/thing/ThingResource.java
```diff
@@ -473,6 +475,34 @@ public Response getStatus(@HeaderParam(HttpHeaders.ACCEPT_LANGUAGE) String langu
         return Response.ok().entity(thingStatusInfo).build();
     }
 
+    @PUT
+    @RolesAllowed({ Role.USER, Role.ADMIN })
+    @Path("/{thingUID}/enable")
+    @ApiOperation(value = "Sets the thing enabled status.")
+    @ApiResponses(value = { @ApiResponse(code = 200, message = "OK", response = String.class),
+            @ApiResponse(code = 400, message = "Thing could not be disabled as it does not have a handler."),
```
htreu: this response will never be returned.

bundles/core/org.eclipse.smarthome.core.thing/src/main/java/org/eclipse/smarthome/core/thing/ThingStatusDetail.java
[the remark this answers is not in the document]
YordanDZhelev: Should be closed since the documentation is updated with this PR

htreu:
Thanks @YordanDZhelev for the quick update. @sjka please have a final look at this.

maggu2810:
If a thing is removed, should the status still be present in the storage?

bundles/core/org.eclipse.smarthome.core.thing/src/main/java/org/eclipse/smarthome/core/thing/internal/ThingManagerImpl.java
```diff
@@ -1238,4 +1317,18 @@ protected void unsetSafeCaller(SafeCaller safeCaller) {
         this.safeCaller = null;
     }
 
+    @Reference
```
sjka: As with #3617 the storage service became dynamically replaceable in all other places, then we should do so here to, shouldn't we? At least I don't see any reason to deviate from that strategy here, especially as the code is prepared for this. 

So, unless I missed anything, this should become

```java
@Reference(policy = ReferencePolicy.DYNAMIC)
```

YordanDZhelev: Implemented as suggested.

sjka:
PS: Travis passed successfully about 15 hours ago. No idea why this isn't reflected here.

tomhoefer:
Wonderful :)

kaikreuzer:
Quick question: How are we supposed to use this feature now?
Afaics, the Paper UI does not support disabling of Things yet, nor is there a console command available. Imho, those two features need to be added. @YordanDZhelev would you volunteer on that? For the Paper UI part, I guess @htreu will happily support you :-)

PR #5194, opened 2018-03-07
Fixed issues found by Static Code Analysis Tool for OnlyTabIdentationInXmlFilesCheck 

tanyatanyatanya:
Changed the rest of the XML files with indentation from spaces to tabs.

### DIFF
--- a/bundles/auth/org.eclipse.smarthome.auth.jaas/OSGI-INF/jaas.xml
+++ b/bundles/auth/org.eclipse.smarthome.auth.jaas/OSGI-INF/jaas.xml
@@ -1,22 +1,22 @@
 <?xml version="1.0" encoding="UTF-8"?>
 <!--
 
-    Copyright (c) 2014,2018 Contributors to the Eclipse Foundation
+	Copyright (c) 2014,2018 Contributors to the Eclipse Foundation
 
-    See the NOTICE file(s) distributed with this work for additional
-    information regarding copyright ownership.
+	See the NOTICE file(s) distributed with this work for additional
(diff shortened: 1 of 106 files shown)


BenediktNiehues:
@maggu2810 didn`t you recognize the compile error in UnitUtils introduced with this PR? Maybe @tanyatanyatanya, could change SI and US in lines 136 and 137 to lowercase, too which @htreu introduced a few hours before your commit ;-)

htreu:
Thanks @BenediktNiehues. See #5195 for a fix.

maggu2810:
> @maggu2810 didn`t you recognize the compile error in UnitUtils introduced with this PR? 

Hopefully! Otherwise it woudln't make sense to merge a PR if I realized that it breaks the build. :wink:

Are you sure this PR is the correct one?
I have not expected that changing indentation breaks something.

Do you refer to this change:
https://github.com/eclipse/smarthome/pull/5158/files#diff-efdbb74ae42de9e6a9978e01fa4da831

@tanyatanyatanya I don't know if it has been a typo in your issue title (then we don't need to care) or in the SAT rule itself (then it should be fixed), but shouldn't "OnlyTabIdentationInXmlFilesCheck" (Identation) be changed into "OnlyTabIndentationInXmlFilesCheck" (Indentation)?

tanyatanyatanya:
@maggu2810 It is not a typo in the issue title, it was a typo when the check was merged, so I decided to leave it like that in the title as well. But you are right we should fix it. I will upload a PR with this fix.

Issue #5198, opened 2018-03-07
Firefox rises xml error "no root element found"


Maw-Jung:
FF 52.6.0 (64-Bit) throws console errors when for example approving a hue lamp.
The approve request has no mediatype set in the Response. Thus firefox tries to
guess it's xml and the error is thrown since there is no xml root element.

PR #5199, opened 2018-03-07
[5198] prevent Responses to be interpreted as xml

Maw-Jung:
The ResponseBuilder "guesses" the best mimetype for the Response by
interpreting the accept headers from the Request. When the client
sets no preferation and xml is valid, the response will be interpreted
as xml. If FF doesn't find the xml root element in this case, which happens
when the Response is empty, a console error is printed out.

Bug: https://github.com/eclipse/smarthome/issues/5198
Signed-off-by: Marcel Walwei <maw.albrecht.jung@gmail.com>

### DIFF
--- a/bundles/automation/org.eclipse.smarthome.automation.rest/src/main/java/org/eclipse/smarthome/automation/rest/internal/RuleResource.java
+++ b/bundles/automation/org.eclipse.smarthome.automation.rest/src/main/java/org/eclipse/smarthome/automation/rest/internal/RuleResource.java
@@ -171,7 +171,7 @@ public class RuleResource implements RESTResource {
             return Response.status(Status.NOT_FOUND).build();
         }
 
-        return Response.ok().build();
+        return Response.ok(null, MediaType.APPLICATION_JSON).build();
     }
 
     @PUT
@@ -190,7 +190,7 @@ public class RuleResource implements RESTResource {
             return Response.status(Status.NOT_FOUND).build();
         }
 
-        return Response.ok().build();
+        return Response.ok(null, MediaType.APPLICATION_JSON).build();
     }
 
     @GET
@@ -229,7 +229,7 @@ public class RuleResource implements RESTResource {
         } else {
             rule.setConfiguration(new Configuration(config));
             ruleRegistry.update(rule);
-            return Response.ok().build();
+            return Response.ok(null, MediaType.APPLICATION_JSON).build();
         }
     }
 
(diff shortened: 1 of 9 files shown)


eclipsewebmaster:
Issue tracker reference:
https://bugs.eclipse.org/bugs/show_bug.cgi?id=5198

maggu2810:
@Maw-Jung Is this really necessary if the method itself is annotated this way
```
@Produces(MediaType.APPLICATION_JSON)
```

Maw-Jung:
From the docs @ https://docs.oracle.com/cd/E19798-01/821-1841/gipxf/index.html:

> The @Produces annotation is used to specify the MIME media types or representations a resource can produce and send back to the client. If @Produces is applied at the class level, all the methods in a resource can produce the specified MIME types by default. If applied at the method level, the annotation overrides any @Produces annotations applied at the class level.

> If no methods in a resource are able to produce the MIME type in a client request, the JAX-RS runtime sends back an HTTP “406 Not Acceptable” error.

The annotation will only limit acceptable requests to the api. Since the paperui accepts the annotated mime type, the response will be successful (200). The mimetype however in an empty response is never set, so the client is free to guess, which fails in FF. A quick test with the paperui prints the exact same error with the annotation set.

maggu2810:
@Maw-Jung You are right.
On an empty response (e.g. `Content-Length: 0`) the header `Content-Type: application/json` is missing.
I don't know if it is a good practice of Firefox to try to parse the content as a XML if it already knows that the response is empty. But if we can avoid errors in browsers -- which seems to much simpler that forcing a user base to update the browser -- I am fine with that change.

Can you post a response (header) before and after your change?
Would a utility function make sense that could be used in all that files?

Maw-Jung:
Before:
`Content-Length: 0`
After:
`Content-Length: 0
Content-Type: application/json`

maggu2810:
@kaikreuzer @sjka Any objections? I am fine with merging this one but would like to know your opinion.

kaikreuzer:
Those kind of problems qualify as "we cannot be the only one" and indeed the web seems to be full of reports about that behavior of Firefox.

According to [this SO answer](https://stackoverflow.com/questions/45195047/angular-2-http-properly-ignore-empty-results/46956926#46956926), the proper solution on our end would not adding a content type (as it is done now in this PR), but to use `Response.noContent()` instead, which results in an HTTP 204, instead of an HTTP 200.

maggu2810:
> which results in an HTTP 204, instead of an HTTP 200

But this would be a breaking change of all the REST API return codes.
Doesn't it involve necessary changes on every REST client?

kaikreuzer:
You are right, if clients rely on 200 return codes, this would mean a breaking change for them, which isn't good.

Sending `application/json`is [incorrect](https://stackoverflow.com/a/40624629), though, so setting it to `text/plain` should be the better choice (this seems to be what [others do as well](https://github.com/go-swagger/go-swagger/pull/1249).

PR #5201, opened 2018-03-07
use generic type argument

maggu2810:
The "as" and "getStateAs" functions should return an instance of the
given type (or null).
It does not make sense that someone calls e.g.
`item.getStateAs(DecimalType.class)` and needs to cast the return value.
The return value should already use the correct type.

### DIFF
--- a/bundles/core/org.eclipse.smarthome.core.test/src/test/java/org/eclipse/smarthome/core/library/types/DecimalTypeTest.java
+++ b/bundles/core/org.eclipse.smarthome.core.test/src/test/java/org/eclipse/smarthome/core/library/types/DecimalTypeTest.java
@@ -16,7 +16,6 @@ import static org.junit.Assert.*;
 
 import java.text.DecimalFormatSymbols;
 
-import org.eclipse.smarthome.core.types.UnDefType;
 import org.junit.Test;
 
 /**
@@ -101,14 +100,14 @@ public class DecimalTypeTest {
     public void testConversionToOpenCloseType() {
         assertEquals(OpenClosedType.OPEN, new DecimalType("1.0").as(OpenClosedType.class));
         assertEquals(OpenClosedType.CLOSED, new DecimalType("0.0").as(OpenClosedType.class));
-        assertEquals(UnDefType.UNDEF, new DecimalType("0.5").as(OpenClosedType.class));
+        assertNull(new DecimalType("0.5").as(OpenClosedType.class));
     }
 
     @Test
     public void testConversionToUpDownType() {
         assertEquals(UpDownType.UP, new DecimalType("0.0").as(UpDownType.class));
         assertEquals(UpDownType.DOWN, new DecimalType("1.0").as(UpDownType.class));
-        assertEquals(UnDefType.UNDEF, new DecimalType("0.5").as(OpenClosedType.class));
+        assertNull(new DecimalType("0.5").as(OpenClosedType.class));
     }
 
     @Test
--- a/bundles/core/org.eclipse.smarthome.core.test/src/test/java/org/eclipse/smarthome/core/library/types/PercentTypeTest.java
+++ b/bundles/core/org.eclipse.smarthome.core.test/src/test/java/org/eclipse/smarthome/core/library/types/PercentTypeTest.java
@@ -12,10 +12,9 @@
  */
 package org.eclipse.smarthome.core.library.types;
 
-import static org.junit.Assert.assertEquals;
+import static org.junit.Assert.*;
 
 import org.eclipse.smarthome.core.library.unit.SmartHomeUnits;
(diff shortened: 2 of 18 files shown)


sjka:
This looks beautiful! And on the fist glance it _even_ is binary compatible, thanks to the type restriction. Or did I miss anything?

Did you see the failing test-case? It looks like it discovered a little flaw - PercentType actually tries to return an UnDefType (because that's how it is defined - it _is_ undefined in that range), which is not the requested type. This is not possible anymore with the strong typing, is it?

maggu2810:
@sjka Didn't realize it yet, as I have to stop the work before the build has been finished. So I would like to check it later.

> It looks like it discovered a little flaw - PercentType actually tries to return an UnDefType (because that's how it is defined - it is undefined in that range), which is not the requested type.

IMHO the current implementation is wrong.

The API states:

> @return the calculated group state of the requested type or null, if type is not supported

> @return the current state in the requested type or null, if state cannot be provided as the requested type

> @return the {@link State}'s value in the given type's representation, or <code>null</code> if the conversion was not possible

So, if would assume if the value is not null it is an instance of the given type / class.
And I am pretty sure I am not the only one: https://github.com/eclipse/smarthome/pull/5178/files#diff-30e97c26aa67e196c6ece6f7375ce825R44

If the API states that null if returned if the state cannot be provided as the requested type and we don't want to use a fallback percent type, then we need to return null and not a instance of another class (UnDefType).

> This is not possible anymore with the strong typing, is it?

Hopefully it is not and I am pretty sure, you know that this is not possible :wink:

maggu2810:
I have added a new commit that return null instead of UnDefType. WDYT?

sjka:
As far as I understood so far, `UnDefType.UNDEF` was invented for this very purpose: To express that a State is not defined. Which somewhat is the case here!

 Is it a different meaning than "the state is not set" a.k.a. `UnDefType.NULL`. Therefore these are not equivalent. Of course, `UnDefType.NULL` is not the same as `null`, therefore your refactoring might still be valid - although clearly a change but in compliance with the javadoc. Hence to me is the question whether the javadoc actually is correct or not. 

@kaikreuzer do you remember the thoughts that historically went into this?

maggu2810:
Isn't it about a conversion to another class / type?
If the conversion is not possible, null should be returned.
If the caller uses `UnDefType.NULL` if null is returned, it is okay, as the caller needs to know what he wants to do.
But if I say, convert to class_a and class_b is returned, this is a very confusing API.

kaikreuzer:
>  UnDefType.UNDEF was invented for this very purpose: To express that a State is not defined.

To be precise: That it cannot be expressed as a value of that type - so yes, in theory it fits as a return value (and when returning `State`, it also feels fully ok).

I think the original design was to return UNDEF, whenever in theory the internal state is compatible with the requested type, but the current value (like 0.5) cannot be mapped - this way, `getStateAs()` ensured to be non-null (if being called with an "accepted type"), just as `getState()`.

I admit that the JavaDoc does not reflect his behavior, though, and that I would also have difficulties to properly formulate the behavior - so returning `null` seems to be indeed the simpler and more straight-forward solution.

I cannot really assess the breaking implications, though. I can imagine that this will specifically break rules of existing users, if they do not handle null values yet. Also, we should check all existing places within ESH whether they already properly deal with null return values.

maggu2810:
So, it depends which API we want to provide.

* a method that provides an state implementation that could be `null`, an instance of the requested type or `UnDefType.UNDEF`
* a method that provides an instance of the requested type or null

If we want to keep the first behavior, this PR does not make any sense.
If we want the second behavior, I assume this PR fits the need and we need to check if all callers handles null correctly.

A null value has been already a possible return value, so normally the caller should already care of. But if someone think of "return UNDEF, whenever in theory the internal state is compatible with the requested type, but the current value (like 0.5) cannot be mapped" then he perhaps does not check for null at all.

How do we want to proceed?
Would we want to differ between "no conversion between the types possible => null", "conversion is possible but not for the given value => UNDEF", "conversion done => requested type" or should we return null if the conversion was not possible and the caller has to handle it.

I would prefer the approach implemented by this PR, because the caller should not bring any knowledge that conversion A to B is possible and so a non-null value is returned. He should always check for null as the return value is nullable.
And then I would prefer to write

```Java
SomeType foo = otherType.as(SomeType.class);
if (foo == null) {
    ...
} else {
   ...
}
```

instead of

```Java
State foo = otherType.as(SomeType.class);
if (foo == null) {
    ...
} else if (foo instanceof UnDefType) {
   ...
} else if (foo instanceof SomeType) {
   ...
} else {
    throw new IllegalStateException();
}
```

htreu:
Although potentially painful I second this PR and the clean API resulting from it.
@kaikreuzer how severe do you rate the openHAB script issue?

PR #5206, opened 2018-03-08
Synthetic script encoding

sjka:
The merge of #5190 made a little funny problem surface when the build is run on a machine which doesn't have UTF-8 set as its default encoding. 

The Xtext based script engine currently uses the default platform encoding when parsing scripts. This somewhat makes sense when scrips are loaded from the file system. So far, so good. 

However, whenever the scripts are passed into the script engine as strings (e.g. from the OSGi console, from rules or from test-cases), they are for sure UTF-8 encoded (as this is that Java strings are). Luckily these strings are made available as "synthetic" resources, i.e. they can be distinguished at runtime. 

This PR adds an encoding provide which uses this knowledge to force Xtext into using UTF-8 for such resources.


sjka:
`computeUnusedUri(...)` creates such synthetic URIs. It is a private method and the only caller is `parseScriptIntoXTextEObject(String)`. 

The method `executeScript(String)` that you mentioned calls `newScriptFromString(String)` which in turn calls `parseScriptIntoXTextEObject(String)` as well. q.e.d.

PR #5207, opened 2018-03-08
AbstractRegistry updates/removes existing instances

### LINKED ISSUE #5189: Statechangelistener prevents GroupItem to be disposed
When a GroupItem registers a StateChangeListener in form of another GroupItem, the deletion of the latter won't remove the StateChangeListener, because the toElement method in ManagedItemProvider creates a new instance of the GroupItem to perform the tidy up task on. Later the removeStateChangeListener of the newly created GroupItem will be called, but this will fail since there are none. The Java GarbageCollector will remove the GroupItem from the ItemRegistryImpl but stops at the old GroupItem, since it's still referenced from the GroupItem which holds the reference on it's listeners. When creating a new GroupItem with the same uid the GroupItem won't add the new GroupItem as a listener, since the overriden equals method of a GenericItem will state them the same, because the comparison isn't done by object id. Thus the listener of the old GroupItem will be used and the new GroupItem won't get updates.
If questions arise i'll be happy to provide additional informations.

Greetings,
maw

htreu:
Fixes #5189.

During update/remove of provider backed entities the AbstractRegistry would have handled deserialized entities. This leads to improper clean up since the live instances where not being touched. Listeners and supporting services where not being cleared.

This ensures that the AbstractRegistry operates on existing instances before passing them to the onUpdatelElement/onRemoveElement methods.

The existing ItemRegistryOSGiTest could not detect this issue because all add/update/remove operations were done on the ProviderChangeListener methods directly on the ItemRegistryImpl.

Also:
* Port ManagedItemProviderOSGiTest, ItemRegistryOSGiTest and ItemTagTest to Java.


bundles/core/org.eclipse.smarthome.core/src/main/java/org/eclipse/smarthome/core/common/registry/AbstractRegistry.java
```diff
@@ -168,8 +168,11 @@ public void removed(Provider<E> provider, E element) {
...
         if (elements != null) {
             try {
-                onRemoveElement(element);
-                elements.remove(element);
+                // the given "element" might not be the live instance but
+                // loaded from storage. operate on the real element:
+                E existingElement = get(element.getUID());
+                onRemoveElement(existingElement);
+                elements.remove(existingElement);
                 notifyListenersAboutRemovedElement(element);
```
sjka: Just a quick first browse-through: Shouldn't `notifyListenersAboutRemovedElement(...)` also be called with `existingElement`?

htreu: absolutely!

htreu: I edited the lines once or twice to see the effect of the test. And while I'm typing this: The tests could also check that the listeners are called with the real entity. wdyt?

htreu: done :-)

sjka:
This somewhat looks plausible. And for sure it is the safest solution and it _should_ be done exactly as you did. So nothing against this change itself.

Nevertheless, something bothered me and after contemplating a little I think I found what it is: The registry as methods where it expected the `oldElement` (i.e. `remove(...)` and `update(...)`). There however the managed providers don't pass the old element, but just a fake copy of it. Isn't that cheating a little? I mean, the API _expects_ the entity. So wouldn't it be cleaner to additionally change that behavior in the managed provider? 
The way you fixed it now basically uses the "fake oldEntity" only to get the UID. So consequently the method signature could be changed to require the identifier only. 
Either way, from my point of view their current behavior doesn't follow the principle of least surprise... Wdyt?

htreu:
@sjka you are completely right: My first attempt was to pass the live instance to the registry as the method signature suggests. Unfortunately the provider (ManagedItemProvider in this case) only has access to the persisted entity. Do you have an idea on how to access the "original" entity?
We could at least introduce a new method with signatures `update/remove(oldElementKey, newElement)`.

htreu:
@sjka may I ask you on your opinion on how to proceed here?

sjka:
> We could at least introduce a new method with signatures update/remove(oldElementKey, newElement).

Sounds like a plan. I'd deprecate but keep the old ones though for compatibility reasons and lt them delegate to the new ones.

htreu:
Unfortunately the plan has some major issue:
A new method `removed(Provider<E> provider, Object key)` on the `ProviderChangeListener` interface would look exactly like the existing method `removed(Provider<E> provider, E element)` due to type erasure. To overcome this situation the `Provider` interface would have to be defined as `Provider<E extends Identifiable<K>, K> to differentiate between the element `E` and its key `K`.
This would be a major refactoring and also API breaking.

sjka:
That's indeed not worth it. Let's keep it as it is - sorry for causing this little detour!

PR #5208, opened 2018-03-08
Some Bluetooth binding improvements

kaikreuzer:
- added support for manufacturer data scan notifications in BlueZ
- only update RSSI if a value is set
- introduced a Utils class for some generally useful methods when dealing with Bluetooth data

### DIFF
--- a/bundles/core/org.eclipse.smarthome.core.test/src/test/java/org/eclipse/smarthome/core/util/HexUtilsTest.java
+++ b/bundles/core/org.eclipse.smarthome.core.test/src/test/java/org/eclipse/smarthome/core/util/HexUtilsTest.java
@@ -1,0 +1,53 @@
+/**
+ * Copyright (c) 2014,2018 Contributors to the Eclipse Foundation
+ *
+ * See the NOTICE file(s) distributed with this work for additional
+ * information regarding copyright ownership.
+ *
+ * This program and the accompanying materials are made available under the
+ * terms of the Eclipse Public License 2.0 which is available at
+ * http://www.eclipse.org/legal/epl-2.0
+ *
+ * SPDX-License-Identifier: EPL-2.0
+ */
+package org.eclipse.smarthome.core.util;
+
+import static org.junit.Assert.assertEquals;
+
+import org.junit.Test;
+
+/**
+ * Test for {@link HexUtils}.
+ *
+ * @author Kai Kreuzer - Initial implementation
+ */
+public class HexUtilsTest {
+
+    @Test
+    public void test_bytesToHex_noParams() {
+        byte[] bytes = "ABCD".getBytes();
+        String result = HexUtils.bytesToHex(bytes);
+        assertEquals("41424344", result);
+    }
+
+    @Test
+    public void test_bytesToHex_withDelimiter() {
+        byte[] bytes = "ABCD".getBytes();
+        String result = HexUtils.bytesToHex(bytes, " ", null);
+        assertEquals("41 42 43 44", result);
+    }
+
+    @Test
+    public void test_bytesToHex_withPrefix() {
+        byte[] bytes = "ABCD".getBytes();
+        String result = HexUtils.bytesToHex(bytes, null, "0x");
+        assertEquals("0x410x420x430x44", result);
+    }
+
+    @Test
+    public void test_bytesToHex_withDelimiterAndPrefix() {
+        byte[] bytes = "ABCD".getBytes();
+        String result = HexUtils.bytesToHex(bytes, "-", "0x");
+        assertEquals("0x41-0x42-0x43-0x44", result);
+    }
+}
--- a/bundles/core/org.eclipse.smarthome.core.test/src/test/java/org/eclipse/smarthome/core/util/HexUtilsTest.java
+++ b/bundles/core/org.eclipse.smarthome.core.test/src/test/java/org/eclipse/smarthome/core/util/HexUtilsTest.java
@@ -33,21 +33,14 @@ public class HexUtilsTest {
     @Test
     public void test_bytesToHex_withDelimiter() {
         byte[] bytes = "ABCD".getBytes();
-        String result = HexUtils.bytesToHex(bytes, " ", null);
+        String result = HexUtils.bytesToHex(bytes, " ");
         assertEquals("41 42 43 44", result);
     }
 
     @Test
-    public void test_bytesToHex_withPrefix() {
+    public void test_bytesToHex_withMultiCharDelimiter() {
         byte[] bytes = "ABCD".getBytes();
-        String result = HexUtils.bytesToHex(bytes, null, "0x");
-        assertEquals("0x410x420x430x44", result);
-    }
-
-    @Test
-    public void test_bytesToHex_withDelimiterAndPrefix() {
-        byte[] bytes = "ABCD".getBytes();
-        String result = HexUtils.bytesToHex(bytes, "-", "0x");
-        assertEquals("0x41-0x42-0x43-0x44", result);
+        String result = HexUtils.bytesToHex(bytes, "-:-");
+        assertEquals("41-:-42-:-43-:-44", result);
     }
 }
--- a/bundles/core/org.eclipse.smarthome.core/META-INF/MANIFEST.MF
+++ b/bundles/core/org.eclipse.smarthome.core/META-INF/MANIFEST.MF
@@ -30,7 +30,8 @@ Export-Package:
  org.eclipse.smarthome.core.service,
  org.eclipse.smarthome.core.storage,
  org.eclipse.smarthome.core.types,
- org.eclipse.smarthome.core.types.util
+ org.eclipse.smarthome.core.types.util,
+ org.eclipse.smarthome.core.util
 Ignore-Package: org.eclipse.smarthome.core.internal.items,org.eclipse.sm
  arthome.core.internal,org.eclipse.smarthome.core.internal.events,org.ec
  lipse.smarthome.core.internal.logging
--- a/bundles/core/org.eclipse.smarthome.core/src/main/java/org/eclipse/smarthome/core/util/HexUtils.java
+++ b/bundles/core/org.eclipse.smarthome.core/src/main/java/org/eclipse/smarthome/core/util/HexUtils.java
@@ -10,30 +10,30 @@
  *
  * SPDX-License-Identifier: EPL-2.0
  */
-package org.eclipse.smarthome.binding.bluetooth;
+package org.eclipse.smarthome.core.util;
 
 /**
- * Static utility methods that are helpful when dealing with Bluetooth data.
+ * Static utility methods that are helpful when dealing with hex data and byte arrays.
  *
  * @author Kai Kreuzer - Initial contribution and API
  *
  */
-public class BluetoothUtils {
+public class HexUtils {
 
     // used for hex conversions
     private final static char[] hexArray = "0123456789ABCDEF".toCharArray();
 
     // private constructor as we only have static methods
-    private BluetoothUtils() {
+    private HexUtils() {
     }
 
     /**
-     * Converts a byte array into a hex string (in format "01 23 45 67 89 0A BC DE").
+     * Converts a byte array into a hex string (in format "01 23 45 67 89 AB CD EF").
      *
      * @param bytes the byte array
      * @return the corresponding hex string
      */
-    public static String bytesToHex(byte[] bytes) {
+    public static String bytesToHexWithSpaces(byte[] bytes) {
         char[] hexChars = new char[bytes.length * 3];
         for (int j = 0; j < bytes.length; j++) {
             int v = bytes[j] & 0xFF;
@@ -42,5 +42,15 @@ public class BluetoothUtils {
             hexChars[j * 3 + 2] = ' ';
         }
         return new String(hexChars).trim();
+    }
+
+    /**
+     * Converts a byte array into a hex string (in format "0123456789ABCDEF").
+     *
+     * @param bytes the byte array
+     * @return the corresponding hex string
+     */
+    public static String bytesToHex(byte[] bytes) {
+        return bytesToHexWithSpaces(bytes).replaceAll(" ", "");
     }
 }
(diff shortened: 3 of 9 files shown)


extensions/binding/org.eclipse.smarthome.binding.bluetooth/src/main/java/org/eclipse/smarthome/binding/bluetooth/BluetoothUtils.java
```diff
@@ -0,0 +1,46 @@
...
+    private BluetoothUtils() {
+    }
+
+    /**
+     * Converts a byte array into a hex string (in format "01 23 45 67 89 0A BC DE").
+     *
+     * @param bytes the byte array
+     * @return the corresponding hex string
+     */
+    public static String bytesToHex(byte[] bytes) {
```
maggu2810: Isn't such a method already present / accessible by the bluetooth binding? IIRC The matthew utils brings such a functionality.
If not or if we want to provide our own implementation, wouldn't it make sense to place it in another package (e.g. core)? I don't think bytesToHex is a BT only utility class but rather unrelated to BT only.
Other bindings parsing incoming bytes would be happy to use that function instead of adding an implementation itself.

Shouldn't the last trailing space at the end of the string removed?

If I understand the code correctly it does not return "01 23 45 67 89 0A BC DE" but "01 23 45 67 89 0A BC DE ".

kaikreuzer: >  IIRC The matthew utils brings such a functionality.

We do not have the mattew libs in the Bluetooth binding. I am not aware of any other place where we have such a method.

In Bluetooth, you have to work extensively with byte arrays, so this seemed to be the most logical place to me in the first step. Yes, there might be a few other bindings that could use it as well, but it isn't very common, I guess.

So if you think it would make sense to have it in smarthome.core that is fine for me and I will move it there.

kaikreuzer: > Shouldn't the last trailing space at the end of the string removed?

Absolutely, I've fixed it.

sjka:
> I am not aware of any other place where we have such a method.

A quick search brought up the following bindings: max, mihome, pentair, silvercrestwifisocket, knx2 (see https://github.com/openhab/openhab2-addons/search?utf8=%E2%9C%93&q=0123456789ABCDEF&type=) and presumably there are several more which which just use a different algorithm and therefore are not as easy to find.

> So if you think it would make sense to have it in smarthome.core that is fine for me and I will move it there.

Yes!

sjka:
PS: just noticed it's used as an "ENCRIPTION_KEY [sic!]" in the silvercrestwifisocket  🤐
Nevertheless it still would make sense, I think

kaikreuzer:
Ok, I have now moved the utility class to the core bundle and also added a version that does not add spaces (as this is what some of the bindings above will need). (I also experimented with a single method that would take a "delimiter" parameter, but that made it all rather more ugly and I don't think we need the flexibility to have any other delimiters than spaces).

bundles/core/org.eclipse.smarthome.core/src/main/java/org/eclipse/smarthome/core/util/HexUtils.java
```diff
@@ -0,0 +1,56 @@
...
+     * @param bytes the byte array
+     * @return the corresponding hex string
+     */
+    public static String bytesToHexWithSpaces(byte[] bytes) {
+        char[] hexChars = new char[bytes.length * 3];
+        for (int j = 0; j < bytes.length; j++) {
+            int v = bytes[j] & 0xFF;
+            hexChars[j * 3] = hexArray[v >>> 4];
+            hexChars[j * 3 + 1] = hexArray[v & 0x0F];
+            hexChars[j * 3 + 2] = ' ';
```
sjka: I actually see use-cases for other delimiters - e.g. `:` for MAC addresses. I'd therefore very much prefer to have the delimiter as an (optional) argument. I care less about _one_ implementation that we hopefully will never have to touch again than having an awkward API with several similar methods or, even worse, people starting to `result = result.replaceAll(" ", ":")`. And while thinking about it, another argument for a prefix would make sense, so the result can be prefixed with e.g. `0x` or `#` optionally.

bundles/core/org.eclipse.smarthome.core/META-INF/MANIFEST.MF
```diff
@@ -30,7 +30,8 @@ Export-Package:
  org.eclipse.smarthome.core.service,
  org.eclipse.smarthome.core.storage,
  org.eclipse.smarthome.core.types,
- org.eclipse.smarthome.core.types.util
+ org.eclipse.smarthome.core.types.util,
+ org.eclipse.smarthome.core.util
```
sjka: This package should also be mentioned in the documentation as "allowed for bindings" and be included as an import in the archetype, considering it would be a perfect place for those few collection utilities that would finally make guava unnecessary. Just mentioning - this IMHO doesn't need to be in this PR!

bundles/core/org.eclipse.smarthome.core/src/main/java/org/eclipse/smarthome/core/util/HexUtils.java
```diff
@@ -0,0 +1,56 @@
...
+ */
+package org.eclipse.smarthome.core.util;
+
+/**
+ * Static utility methods that are helpful when dealing with hex data and byte arrays.
+ *
+ * @author Kai Kreuzer - Initial contribution and API
+ *
+ */
+public class HexUtils {
```
sjka: Such stuff IMHO should also have test-cases, primarily to prevent accidental regressions.

extensions/binding/org.eclipse.smarthome.binding.bluetooth.bluez/src/main/java/org/eclipse/smarthome/binding/bluetooth/bluez/BlueZBluetoothDevice.java
```diff
@@ -136,6 +138,18 @@ private void enableNotifications() {
             notification.setRssi(n);
             notifyListeners(BluetoothEventType.SCAN_RECORD, notification);
         });
+        device.enableManufacturerDataNotifications(n -> {
+            for (Map.Entry<Short, byte[]> entry : n.entrySet()) {
+                BluetoothScanNotification notification = new BluetoothScanNotification();
+                byte[] data = new byte[entry.getValue().length + 2];
+                data[0] = (byte) (entry.getKey() & 0xFF);
+                data[1] = (byte) (entry.getKey() >> 8);
```
sjka: don't you need to erase the unwanted bits first by either `& 0xFF00`-ing it before or `& 0xFF`-ing it after the shift?

extensions/binding/org.eclipse.smarthome.binding.bluetooth.bluez/src/main/java/org/eclipse/smarthome/binding/bluetooth/bluez/BlueZBluetoothDevice.java
```diff
@@ -136,6 +138,18 @@ private void enableNotifications() {
...
+                System.arraycopy(entry.getValue(), 0, data, 2, entry.getValue().length);
+                logger.debug("Received manufacturer data for '{}': {}", address, HexUtils.bytesToHexWithSpaces(data));
```
sjka: `if (logger.isDebugEnabled()) {`?

bundles/core/org.eclipse.smarthome.core/src/main/java/org/eclipse/smarthome/core/util/HexUtils.java
kaikreuzer: Do you have a suggestion for the method signature if you want to have prefixes and suffixes at any length? Pass two nullable Strings as additional parameters? Define a formatting (which syntax) for the single values?

extensions/binding/org.eclipse.smarthome.binding.bluetooth.bluez/src/main/java/org/eclipse/smarthome/binding/bluetooth/bluez/BlueZBluetoothDevice.java
kaikreuzer: No, because they are dropped by shifting.

bundles/core/org.eclipse.smarthome.core/src/main/java/org/eclipse/smarthome/core/util/HexUtils.java
sjka: How about 

```java
public static String bytesToHex(byte[] bytes, @Nullable CharSequence prefix, @Nullable CharSequence delimiter) {
  // ...
}
```

? Of course, a convenience signature like 

```java
public static String bytesToHex(byte[] bytes) {
  return bytesToHex(bytes, null, null);
}
```

makes sense too.

extensions/binding/org.eclipse.smarthome.binding.bluetooth.bluez/src/main/java/org/eclipse/smarthome/binding/bluetooth/bluez/BlueZBluetoothDevice.java
sjka: No, technically they are not all dropped - see https://docs.oracle.com/javase/tutorial/java/nutsandbolts/op3.html. The `>>` operator replicates the highest order bit when shifting. If you want to fill the higher order bits with `0`s, then you'd have to use the `>>>` operator.

kaikreuzer: Thanks for the lesson. That used to be different in C when I did that 30 years ago ;-)

bundles/core/org.eclipse.smarthome.core/src/main/java/org/eclipse/smarthome/core/util/HexUtils.java
kaikreuzer: What should the convenience method return? The string with space delimiters as right now?

kaikreuzer:
@sjka I think I have addressed all your comments.

bundles/core/org.eclipse.smarthome.core.test/src/test/java/org/eclipse/smarthome/core/util/HexUtilsTest.java
```diff
@@ -0,0 +1,53 @@
...
+        byte[] bytes = "ABCD".getBytes();
+        String result = HexUtils.bytesToHex(bytes, " ", null);
+        assertEquals("41 42 43 44", result);
+    }
+
+    @Test
+    public void test_bytesToHex_withPrefix() {
+        byte[] bytes = "ABCD".getBytes();
+        String result = HexUtils.bytesToHex(bytes, null, "0x");
+        assertEquals("0x410x420x430x44", result);
```
sjka: The prefix was meant to be applied only once, not in every segment, i.e. the expectation here should be `0x41424344`

bundles/core/org.eclipse.smarthome.core.test/src/test/java/org/eclipse/smarthome/core/util/HexUtilsTest.java
```diff
@@ -0,0 +1,53 @@
...
+    }
+
+    @Test
+    public void test_bytesToHex_withDelimiterAndPrefix() {
+        byte[] bytes = "ABCD".getBytes();
+        String result = HexUtils.bytesToHex(bytes, "-", "0x");
+        assertEquals("0x41-0x42-0x43-0x44", result);
```
sjka: see above

extensions/binding/org.eclipse.smarthome.binding.bluetooth.bluez/src/main/java/org/eclipse/smarthome/binding/bluetooth/bluez/BlueZBluetoothDevice.java
```diff
@@ -136,6 +138,21 @@ private void enableNotifications() {
...
+        device.enableManufacturerDataNotifications(n -> {
+            for (Map.Entry<Short, byte[]> entry : n.entrySet()) {
+                BluetoothScanNotification notification = new BluetoothScanNotification();
+                byte[] data = new byte[entry.getValue().length + 2];
+                data[0] = (byte) (entry.getKey() & 0xFF);
+                data[1] = (byte) (entry.getKey() >>> 8);
+                System.arraycopy(entry.getValue(), 0, data, 2, entry.getValue().length);
+                if (logger.isDebugEnabled()) {
+                    logger.debug("Received manufacturer data for '{}': {}", address,
+                            HexUtils.bytesToHexWithSpaces(data));
```
sjka: `bytesToHexWithSpaces(...)` doesn't exist anymore

extensions/binding/org.eclipse.smarthome.binding.bluetooth.bluez/src/main/java/org/eclipse/smarthome/binding/bluetooth/bluez/BlueZBluetoothDevice.java
```diff
@@ -150,7 +167,12 @@ private void enableNotifications() {
             }
         });
         device.enableServiceDataNotifications(data -> {
-            logger.debug("Received service data for '{}': {}", address, data);
+            if (logger.isDebugEnabled()) {
+                logger.debug("Received service data for '{}':", address);
+                for (Map.Entry<String, byte[]> entry : data.entrySet()) {
+                    logger.debug("{} : {}", entry.getKey(), HexUtils.bytesToHexWithSpaces(entry.getValue()));
```
sjka: dito

bundles/core/org.eclipse.smarthome.core.test/src/test/java/org/eclipse/smarthome/core/util/HexUtilsTest.java
kaikreuzer: Hm, then I do not really see much sense in the prefix at all. The caller can easily add it to the string himself...

kaikreuzer: I have removed it.

sjka:
Fine by me, thanks!

PR #5209, opened 2018-03-08
Initial contribution of Blukii Smart Beacon support for Bluetooth

kaikreuzer:


kaikreuzer:
This is the first specific-device-add-on for the Bluetooth binding.
It brings support for [Blukii Smart Beacon](http://www.blukii.com/en/produkte/smartbeacons.html)s.

The build is still expected to fail as this PR requires #5208 to be merged first.

extensions/binding/org.eclipse.smarthome.binding.bluetooth.blukii/src/main/java/org/eclipse/smarthome/binding/bluetooth/blukii/BlukiiBindingConstants.java
```diff
@@ -0,0 +1,45 @@
...
+import org.eclipse.smarthome.binding.bluetooth.BluetoothBindingConstants;
+import org.eclipse.smarthome.core.thing.ThingTypeUID;
+
+/**
+ * The {@link BlukiiBindingConstants} class defines common constants, which are
+ * used across the whole binding.
+ *
+ * @author Kai Kreuzer - Initial contribution
+ */
+public class BlukiiBindingConstants {
```
amitjoy: Just a Geschmacksache:

```java
private BlukiiBindingConstants() {
     throw new IllegalAccessError("Cannot be instantiated");
}
```

extensions/binding/org.eclipse.smarthome.binding.bluetooth.blukii/src/main/java/org/eclipse/smarthome/binding/bluetooth/blukii/BlukiiBindingConstants.java
```diff
@@ -0,0 +1,45 @@
...
+
+    // List of all Thing Type UIDs
+    public static final ThingTypeUID THING_TYPE_BEACON = new ThingTypeUID(BluetoothBindingConstants.BINDING_ID,
+            "blukii_beacon");
+
+    public final static String BLUKII_PREFIX = "blukii ";
```
amitjoy: We could reorder the modifiers to comply with the Java Language Specification.

public final static ==> public static final

amitjoy: applies to other constants as well.

sjka:
It's always easy to review what _is_ there but I just realized I missed something which is _not_ here: Shouldn't this binding have a `README.md`?

PR #5214, opened 2018-03-09
Change max connections per host to 2

sobimibos:
Changing max connections per host to 2 avoids EofExceptions.


kaikreuzer:
Please clean up your PR and make sure that there is a single commit only!

maggu2810:
Why 2, why not 1 or 3?
Can you please add more details about the chooen number?

sjka:
[RFC 2616](https://www.ietf.org/rfc/rfc2616.txt) (Hypertext Transfer Protocol) says:

> Clients that use persistent connections SHOULD limit the number of
   simultaneous connections that they maintain to a given server. A
   single-user client SHOULD NOT maintain more than 2 connections with
   any server or proxy.

To me this reads like 2 is the "official" recommendation.

maggu2810:
> To me this reads like 2 is the "official" recommendation.

Me to.
It would be great if the commit message contains such details or at least the PR description itself (or refer to an issue).
I don't think we should expect that everybody have the RFC 2616 in mind. :wink:

kaikreuzer:
I agree that some comment would be helpful. @sobimibos Wdyt? Could you add one?

PR #5216, opened 2018-03-10
WIP: Add properties to config description

cdjackson:
I would like to propose to add properties to config parameters. This is to support some level of parameterisation of configuration.

As a concrete example, in the zwave binding, in order to describe a parameter, I need to use the parameter name to embed some data -:

```
      <parameter name="config_17_1" type="integer" groupName="configuration">
      </parameter>
```

This embeds a parameter number and a size into the name - ie parameter 17, 1 byte. I don't really like this, but there's currently no alternative. I should add that in some instances, a lot more data is added to this name so this is a simple example, and I can have this sort of thing ```config_17_4_0000FF00_wo```.

My proposal is to add properties, so instead of the above, we would have -:
```
      <parameter name="config" type="integer" groupName="configuration">
        <properties>
          <property name="parameter">17</property>
          <property name="length">1</property>
        </properties>
      </parameter>
```

I'm about to start doing the same sort of thing in the ZigBee binding, and I would prefer not to if we can agree on something along the lines proposed here.

This PR just adds the XSD to present the proposal - it would be extended if agreed.

### DIFF
--- a/bundles/config/org.eclipse.smarthome.config.xml/config-description-1.0.0.xsd
+++ b/bundles/config/org.eclipse.smarthome.config.xml/config-description-1.0.0.xsd
@@ -56,6 +56,7 @@
             <xs:element name="verify" type="xs:boolean" minOccurs="0" />
 			<xs:element name="multipleLimit" type="xs:integer" minOccurs="0" />
             <xs:element name="unitLabel" type="xs:string" minOccurs="0" />
+            <xs:element name="properties" type="config-description:properties" minOccurs="0"/>
 		</xs:all>
 		<xs:attribute name="name" type="xs:string" use="required" />
 		<xs:attribute name="type" type="config-description:parameterType"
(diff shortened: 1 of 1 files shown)


kaikreuzer:
I am not really clear on what this is exactly about - it again feels like the config descriptions are misused for something they are not meant for...

What does "in order to describe a parameter" mean? By definition, the config description IS the description of the parameters - that is, the keys of parameters must be clearly defined and they are not a dynamic set.

I assume that your requirement is rather in the line of https://github.com/eclipse/smarthome/issues/3484 - so we should actually move forward on that issue.

cdjackson:
The config description describes the configuration only from the user perspective - it doesn't necessarily describe it from a device perspective. The config description shows what we put on the UI, but somehow, we need to use the data and that's what I'm doing here. So, the properties tells the thing handler that a specific config description is related to a specific end device parameter.

Yes, I do agree it's related to this other issue, but I would expect (hope even) that when we come up with a system to configure devices, that the config descriptions should be the same right? You're not planning a new way to describe configuration are you? I expect that config descriptions is THE reusable component in ESH for describing parameters - internal to the thing, channel, device, etc. 

So, while my use case is for a thing, I would have thought that we still need to do something in the config description?

htreu:
Hi @cdjackson, I just wanted to let you know that I'm now working on the subject of #3484. The requirements you described here will be considered. I hope to come up with a proposal in the-next-days(tm).

cdjackson:
Cool - thanks @treu. Do you plan to use the existing config descriptions? I guess so, but just in a separate definition section of the XML?

I'd be happy to look over the PR - ZWave has over 700 device descriptions ;)

sjka:
> I'm now working on the subject of #3484. The requirements you described here will be considered.

Okay, so I reckon we can close this one.

PR #5222, opened 2018-03-12
Fix link to QuantityType in UoM blog post

### LINKED ISSUE #5221: Broken link to QuantityType in UoM blog post
Hi,

I just noticed that the link to the _QuantityType_ inside the [UoM blog post](https://www.eclipse.org/smarthome/blog/2018/02/22/units-of-measurement.html) is broken. Stepping a directory back won't work because of the URL pattern.

@htreu: Don't know whats best to fix this. Absolute URL?

Regards
Ingo

sjka:
fixes #5221

### DIFF
--- a/docs/_config.yml
+++ b/docs/_config.yml
@@ -5,7 +5,7 @@ description: > # this means to ignore newlines until "baseurl:"
   Write an awesome description for your new site here. You can edit this
   line in _config.yml. It will appear in your document head meta (for
   Google search results) and in your feed.xml site description.
-baseurl: "/smarthome/" # the subpath of your site, e.g. /blog/
+baseurl: "/smarthome" # the subpath of your site, e.g. /blog/
 url: "https://www.eclipse.org/smarthome" # the base hostname & protocol for your site
 twitter_username: jekyllrb
 github_username:  jekyll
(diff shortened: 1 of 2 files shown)


docs/_config.yml
```diff
@@ -5,7 +5,7 @@ description: > # this means to ignore newlines until "baseurl:"
   Write an awesome description for your new site here. You can edit this
   line in _config.yml. It will appear in your document head meta (for
   Google search results) and in your feed.xml site description.
-baseurl: "/smarthome/" # the subpath of your site, e.g. /blog/
+baseurl: "/smarthome" # the subpath of your site, e.g. /blog/
```
kaikreuzer: Maybe worth to remove the trailing slash from the "e.g. /blog/" as well? But then again, why don't we follow the example?

sjka: > Maybe worth to remove the trailing slash from the "e.g. /blog/" as well?

That one comes from jekyll itself as an example. I wouldn't change it, but rather remove it completely.

> But then again, why don't we follow the example?

Because we would end up with 

```html
<a href="/smarthome//documentation/concepts/units-of-measurement.html">
```

No idea why though - Jekyll sometimes feels like pure magic.

Issue #5224, opened 2018-03-12
[Test failure] NTP binding test fails


sjka:
Since today/yesterday, the NTP binding tests fail with

```
java.lang.AssertionError: The dateTime channel calendar was not updated with the right timezone
Expected: is <-08:00>
     but: was <-07:00>
	at org.hamcrest.MatcherAssert.assertThat(MatcherAssert.java:20)
	at org.junit.Assert.assertThat(Assert.java:956)
	at org.eclipse.smarthome.binding.ntp.test.NtpOSGiTest.testDateTimeChannelCalendarTimeZoneUpdate(NtpOSGiTest.java:245)
	at sun.reflect.NativeMethodAccessorImpl.invoke0(Native Method)
	at sun.reflect.NativeMethodAccessorImpl.invoke(NativeMethodAccessorImpl.java:62)
	at sun.reflect.DelegatingMethodAccessorImpl.invoke(DelegatingMethodAccessorImpl.java:43)
	at java.lang.reflect.Method.invoke(Method.java:498)
	at org.junit.runners.model.FrameworkMethod$1.runReflectiveCall(FrameworkMethod.java:50)
	at org.junit.internal.runners.model.ReflectiveCallable.run(ReflectiveCallable.java:12)
	at org.junit.runners.model.FrameworkMethod.invokeExplosively(FrameworkMethod.java:47)
	at org.junit.internal.runners.statements.InvokeMethod.evaluate(InvokeMethod.java:17)
	at org.junit.internal.runners.statements.RunBefores.evaluate(RunBefores.java:26)
	at org.junit.internal.runners.statements.RunAfters.evaluate(RunAfters.java:27)
	at org.junit.runners.ParentRunner.runLeaf(ParentRunner.java:325)
	at org.junit.runners.BlockJUnit4ClassRunner.runChild(BlockJUnit4ClassRunner.java:78)
	at org.junit.runners.BlockJUnit4ClassRunner.runChild(BlockJUnit4ClassRunner.java:57)
	at org.junit.runners.ParentRunner$3.run(ParentRunner.java:290)
	at org.junit.runners.ParentRunner$1.schedule(ParentRunner.java:71)
	at org.junit.runners.ParentRunner.runChildren(ParentRunner.java:288)
	at org.junit.runners.ParentRunner.access$000(ParentRunner.java:58)
	at org.junit.runners.ParentRunner$2.evaluate(ParentRunner.java:268)
[...]

PR #1050, opened 2016-02-16
Proposition of evolution for issue #992 : DateTimeType is not reconst…

clinique:
…ructable from its String

Signed-off-by: Gaël L'hopital <glhopital@gmail.com>

Issue #992, opened 2016-02-07
DateTimeType is not reconstructable from its String serialization


kaikreuzer:
For all Types, we have the general assumption that they should be reconstructable from their string representation, i.e. `typeX.equals(new TypeX(typeX.toString()))` should be true.

This is currently not the case for the DateTimeType, since the toString methods cuts off the milliseconds, which are contained within the internal Calendar object.

watou:
> For all Types, we have the general assumption that they should be reconstructable from their string representation, i.e. typeX == new TypeX(typeX.toString()) should be true.

Do you mean `typeX.equals(new TypeX(typeX.toString())`?

> This is currently not the case for the DateTimeType, since the toString methods cuts off the milliseconds, which are contained within the internal Calendar object.

Other DateTime information is also not preserved in the string format.  It's worth considering fixing the omission of time zone as well as milliseconds in the string format using ISO8601 formatting.  Currently no time zone is serialized, so DateTimeType is unusable by itself for use with multiple timezones.

kaikreuzer:
> Do you mean typeX.equals(new TypeX(typeX.toString())?

Of course, corrected it above now!

clinique:
This would mean changing the actual contract (and according tests) of DateTimeType.toString() - right ?

kaikreuzer:
Right.

kaikreuzer:
Fixed with https://github.com/eclipse/smarthome/pull/1050

marcelrv:
as @watou indicated:
> Other DateTime information is also not preserved in the string format. It's worth considering fixing the omission of time zone as well as milliseconds in the string format using ISO8601 formatting. Currently no time zone is serialized, so DateTimeType is unusable by itself for use with multiple timezones.

This is indeed an issue for other bindings as well... see e.g. 
https://github.com/eclipse/smarthome/issues/1666

PR #5226, opened 2018-03-12
Corrected NTP test failures

doandzhi:
#5224

Extended the DateTimeType API with a method that returns the zonedDateTime instance for a given time zone. The test failures are because of time zone change in LA (test zone) and the .isDaylightSavingTime() for System Zone (which is Europe/Bucharest).

### DIFF
--- a/bundles/core/org.eclipse.smarthome.core/src/main/java/org/eclipse/smarthome/core/library/types/DateTimeType.java
+++ b/bundles/core/org.eclipse.smarthome.core/src/main/java/org/eclipse/smarthome/core/library/types/DateTimeType.java
@@ -115,6 +115,13 @@ public class DateTimeType implements PrimitiveType, State, Command {
         return zonedDateTime;
     }
 
+    /*
+     * Returns the zonedDateTime object for a given time zone
+     */
+    public ZonedDateTime getZonedDateTimeAtZone(ZoneId zone) {
+        return zonedDateTime.withZoneSameInstant(zone);
+    }
+
     public static DateTimeType valueOf(String value) {
         return new DateTimeType(value);
     }
--- a/extensions/binding/org.eclipse.smarthome.binding.ntp.test/src/test/java/org/eclipse/smarthome/binding/ntp/test/NtpOSGiTest.java
+++ b/extensions/binding/org.eclipse.smarthome.binding.ntp.test/src/test/java/org/eclipse/smarthome/binding/ntp/test/NtpOSGiTest.java
@@ -14,7 +14,7 @@ package org.eclipse.smarthome.binding.ntp.test;
 
 import static org.hamcrest.CoreMatchers.*;
 import static org.junit.Assert.assertThat;
-import static org.mockito.Matchers.any;
+import static org.mockito.ArgumentMatchers.any;
 import static org.mockito.Mockito.*;
 
 import java.time.ZoneId;
@@ -217,7 +217,8 @@ public class NtpOSGiTest extends JavaOSGiTest {
         String testItemState = getItemState(ACCEPTED_ITEM_TYPE_DATE_TIME).toString();
         assertFormat(testItemState, DateTimeType.DATE_PATTERN_WITH_TZ_AND_MS);
         ZonedDateTime timeZoneFromItemRegistry = ((DateTimeType) getItemState(ACCEPTED_ITEM_TYPE_DATE_TIME))
-                .getZonedDateTime();
+                .getZonedDateTimeAtZone(ZoneId.of(TEST_TIME_ZONE_ID));
+        ;
         ZoneOffset expectedOffset;
         if (timeZoneFromItemRegistry.getZone().getRules().isDaylightSavings(timeZoneFromItemRegistry.toInstant())) {
             expectedOffset = ZoneOffset.of("-07:00");
@@ -235,7 +236,7 @@ public class NtpOSGiTest extends JavaOSGiTest {
         configuration.put(NtpBindingConstants.PROPERTY_TIMEZONE, TEST_TIME_ZONE_ID);
         initialize(configuration, NtpBindingConstants.CHANNEL_DATE_TIME, ACCEPTED_ITEM_TYPE_DATE_TIME, null, null);
         ZonedDateTime timeZoneIdFromItemRegistry = ((DateTimeType) getItemState(ACCEPTED_ITEM_TYPE_DATE_TIME))
-                .getZonedDateTime();
+                .getZonedDateTimeAtZone(ZoneId.of(TEST_TIME_ZONE_ID));
         ZoneOffset testZoneId;
         if (timeZoneIdFromItemRegistry.getZone().getRules().isDaylightSavings(timeZoneIdFromItemRegistry.toInstant())) {
             testZoneId = ZoneOffset.of("-07:00");


extensions/binding/org.eclipse.smarthome.binding.ntp.test/src/test/java/org/eclipse/smarthome/binding/ntp/test/NtpOSGiTest.java
```diff
@@ -217,7 +217,8 @@ public void testDateTimeChannelTimeZoneUpdate() {
         String testItemState = getItemState(ACCEPTED_ITEM_TYPE_DATE_TIME).toString();
         assertFormat(testItemState, DateTimeType.DATE_PATTERN_WITH_TZ_AND_MS);
         ZonedDateTime timeZoneFromItemRegistry = ((DateTimeType) getItemState(ACCEPTED_ITEM_TYPE_DATE_TIME))
-                .getZonedDateTime();
+                .getZonedDateTimeAtZone(ZoneId.of(TEST_TIME_ZONE_ID));
```
sjka: To be honest, I don't really get this change. The purpose of this test is to verify the inverse of "The dateTime channel was not updated with the right timezone", i.e. that the channel _was_ updated with the right time zone. Now you have to manually convert the time zone to the desired one. Isn't that obfuscating some other flaw?

bundles/core/org.eclipse.smarthome.core/src/main/java/org/eclipse/smarthome/core/library/types/DateTimeType.java
```diff
@@ -115,6 +115,13 @@ public ZonedDateTime getZonedDateTime() {
         return zonedDateTime;
     }
 
+    /*
+     * Returns the zonedDateTime object for a given time zone
+     */
+    public ZonedDateTime getZonedDateTimeAtZone(ZoneId zone) {
```
sjka: Is this method really needed? (see below)

sjka:
PS: I think I now understood the problem...

In the test-case, the following expression doesn't have the time zone name anymore, just the "-7", hence making wrong assumptions about whether is is daylight saving or not:

```
timeZoneIdFromItemRegistry.getZone().getRules().isDaylightSavings(timeZoneIdFromItemRegistry.toInstant())
```

The test therefore must not use the returned date time as a basis to infer whether the returned daytime has the correct time zone, but rather should hard-code expectations properly, i.e.

```
ZoneOffset expectedOffset = ZoneId.of(TEST_TIME_ZONE_ID).getRules()
                .getOffset(timeZoneFromItemRegistry.toInstant());
```

lpapazow:
I was trying to understand why the timezone name gets lost and only the "-7" remains and I saw that when the NTP handler updates the dateTimeChannel [here](https://github.com/eclipse/smarthome/blob/master/extensions/binding/org.eclipse.smarthome.binding.ntp/src/main/java/org/eclipse/smarthome/binding/ntp/handler/NtpHandler.java#L221), the `toString()` method of DateTimeType gets implicitly called, then the `toString()` method calls `toFullString()` which returns the zonedDateTime which formats with the zonedDateTime with the RFC formatter [here](https://github.com/eclipse/smarthome/blob/master/bundles/core/org.eclipse.smarthome.core/src/main/java/org/eclipse/smarthome/core/library/types/DateTimeType.java#L142).

The RFCFormatter has the following pattern: `yyyy-MM-dd'T'HH:mm[:ss[.SSS]]Z`

I tested how the date is displayed when applying different formatters and I saw that when 
`yyyy-MM-dd'T'HH:mm[:ss[.SSS]]Z` is applied, the date is displayed like this: `2018-03-14T12:13:57.941+0200`
and when the the 
`yyyy-MM-dd'T'HH:mm[:ss[.SSS]]z` is applied, the date is displayed like this: `2018-03-14T12:13:57.945EET`

I noticed that the [parse method](https://github.com/eclipse/smarthome/blob/master/bundles/core/org.eclipse.smarthome.core/src/main/java/org/eclipse/smarthome/core/library/types/DateTimeType.java#L178) also prioritizes the formatterTzRFC. 

After changing the formatter in the `toFullString` method, these two tests worked but many other broke.

Do you know if there is any specific reason to prioritize this formatter over the other? @sjka

sjka:
I'm not that much into the details - I think it was first added in #1050. The main objective was #992.

From a framework's perspective it is important that `dateTimeType.equals(new DateTimeType(dateTimeType.toFullString()))` keeps holding true. 

From a user's point - I'm not sure what the uproar will be if we change that format. It's been like that for 2 years now, but I'm not sure how much the actual format matters in real-world use-cases.

sjka:
@doandzhi did you see my comment? After thinking little more about this, I think definitively would make sense to make that change. Parsing a string should of course stay backwards-compatible, so it must still understand the previous format. Nevertheless you'd have to check if the PaperUI as a receiver of events can handle the new format.

@ghys do you know if changing the serialized DateTimeType format like outlined in https://github.com/eclipse/smarthome/pull/5226#issuecomment-372971820 would be an issue for HABPanel?

ghys:
> @ghys do you know if changing the serialized DateTimeType format like outlined in #5226 (comment) would be an issue for HABPanel?

I think both Paper UI and HABPanel use the AngularJS date filter to deserialize the dates, which in turn simply use the JavaScript Date object constructor accepting a string (https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Date).

> The string should be in a format recognized by the Date.parse() method (IETF-compliant RFC 2822 timestamps and also a version of ISO8601).

If you go to a JavaScript console here's what you get:
```
> new Date('2018-03-14T12:13:57.941+0200')
Wed Mar 14 2018 11:13:57 GMT+0100 (Central Europe Standard Time)
> new Date('2018-03-14T12:13:57.945EET')
Invalid Date
```
So yes, the second format would be an issue I think.

sjka:
I think @ghys brought up a really important point in https://github.com/eclipse/smarthome/pull/5306#issuecomment-376211841: The abbreviated time zone names are not standardized - or at least not unique (see e.g. https://en.wikipedia.org/wiki/List_of_time_zone_abbreviations). Therefore they are not usable for that purpose. Therefore it would have to be [ISO_ZONED_DATE_TIME](https://docs.oracle.com/javase/8/docs/api/java/time/format/DateTimeFormatter.html#ISO_ZONED_DATE_TIME) like e.g. `2011-12-03T10:15:30+01:00[Europe/Paris]`. 

Which actually throws us a step back again: There are two valid interpretations of the problem we are discussing here:

1. The `DateTypeType` as it is today suits its needs: Its string representation describes perfectly well a point in time, including the GMT offset. That leaves no question about when this is anywhere in the world. Why include any additional meta-data?

2. The string representation is incomplete in the sense that it doesn't state anymore which time-zone it originates from, therefore in cannot be told whether there currently is a daylight active or in some cases what would be the rules to apply if that timestamp is shifted. 

So far I haven't seen that the second aspect was a requirement anywhere. Actually, everybody is happy with the timestamp as it is, apart from the test-cases which wrongly try to derive this meta-information from the result instead of making a fixed, predefined assumption. Therefore I reckon it would be okay to just fix the test as outlined in https://github.com/eclipse/smarthome/pull/5226#issuecomment-372588755, especially as changing the format wouldn't do any JS based UI any good.

lpapazow:
@maggu2810, relevant discussion to #5306

sjka:
With #5306 being merged I'm closing this PR with respect to its original purpose.

PR #5228, opened 2018-03-12
don't start custom client

sobimibos:

### DIFF
--- a/bundles/io/org.eclipse.smarthome.io.net/src/main/java/org/eclipse/smarthome/io/net/http/HttpClientFactory.java
+++ b/bundles/io/org.eclipse.smarthome.io.net/src/main/java/org/eclipse/smarthome/io/net/http/HttpClientFactory.java
@@ -25,7 +25,7 @@ public interface HttpClientFactory {
 
     /**
      * Creates a new jetty http client.
-     * The returned client is already started but must be stopped
+     * The returned client is must be started before and stopped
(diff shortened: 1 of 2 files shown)


bundles/io/org.eclipse.smarthome.io.net/src/main/java/org/eclipse/smarthome/io/net/http/HttpClientFactory.java
```diff
@@ -25,7 +25,7 @@
 
     /**
      * Creates a new jetty http client.
-     * The returned client is already started but must be stopped
+     * The returned client is must be started before and stopped
```
maggu2810: Shouldn't you remove the `is`, too?
Perhaps is could be changed to:
```
The returned client is not started yet. You have to start it yourself before using. Don't forget to stop a started client again after its usage.
```

sobimibos: The custom client is intended to be used by consumers with special configuration requirements for the client. Unfortuntely some part of the configuration like a customized thread pool can only be applied before the client is started. So it's better to not start it in the factory and give the consumer full access to the configuration.
I will change the comment.

maggu2810:
@kaikreuzer Do you agree, too?

Issue #5211, opened 2018-03-09
Sitemap SSE channel doesn't emit events if 'sitemap' and 'pageid' parameters aren't given


maniac103:
Those parameters are marked as optional in the docs though.

Repro steps:
- `curl -X POST http://demo.openhab.org:8080/rest/sitemaps/events/subscribe`
- `curl http://demo.openhab.org:8080/rest/sitemaps/events/{subid from above}` -> change e.g. 'Ceiling' switch in 'Bathroom' and observe no event
- `curl http://demo.openhab.org:8080/rest/sitemaps/events/{subid from above}?sitemap=demo&pageid=FF_Bath` -> flip same switch and observe event
- `curl http://demo.openhab.org:8080/rest/sitemaps/events/{subid from above}` -> flip same switch and now observe events as well, despite sitemap and pageid not being given

For the OH Android app, it would be great to be able to actually omit the parameters, as it would allow getting away with a single SSE connection instead of one per shown page.

lolodomo:
All the SSE stuff on the server side is managed per page.
When you provide sitemap/page as URL parameter, the server is linking your subscription to this page.
When you provide no sitemap/page as URL parameter, the server continue considering the previously page set.

maniac103:
In that case, the docs should be updated to say that sitemap and pageid are mandatory.

lolodomo:
Do you know if any documentation exists for that ?
Not at my knowledge.

maniac103:
It's shipped with the distribution. Example: demo.openhab.org -> REST API -> sitemaps -> /sitemaps/events/subscribe

lolodomo:
You mean we should update the REST docs ?

kaikreuzer:
> For the OH Android app, it would be great to be able to actually omit the parameters, as it would allow getting away with a single SSE connection instead of one per shown page.

Could you elaborate on this? The sitemap event subscription had been specifically introduced to reduce the event stream to what is relevant for the client - sending events for potentially all pages of a huge sitemap pretty much contradicts its purpose....

lolodomo:
The Android app is displaying a maximum of 2 pages at the same time so 2 SSE subscriptions should be enough.
In my current implementation, before the code was massively refactored, I was using more SSE subscriptions depending on the sitemap depth.

maniac103:
> In my current implementation, before the code was massively refactored, I was using more SSE subscriptions depending on the sitemap depth.

[...]

PR #5229, opened 2018-03-12
sitemap SSE: event resource: sitemap name and page ID are required

maggu2810:
Fixes: https://github.com/eclipse/smarthome/issues/5211


maniac103:
I think that's the wrong method to change. https://github.com/maggu2810/smarthome/blob/b89e8d774ffe036a28e2e9051ebd50853db9e574/bundles/io/org.eclipse.smarthome.io.rest.sitemap/src/main/java/org/eclipse/smarthome/io/rest/sitemap/internal/SitemapResource.java#L260 is the one I was referring to in #5211 (AFAICT from looking at the code that I'm otherwise unfamiliar with)

Issue #5197, opened 2018-03-07
Unable to use JSON structured mqtt message payload in the lwt property of MQTTService


doandzhi:
As Last Will Testament is configured with the `lwt` property of the MqttService it has a specific formatting that uses which doesn't allow JSON structured messages since the separation between parametters is done using `:`. I see it as a valid use case to use JSON structure in my LWT message but the property is split by `:` which makes it impossible. Does it make sense to change the `:` to something else or it is against some ESH guidelines that I don't know yet?

_example usage of lwt property:_ 
`org.eclipse.smarthome.mqtt:mosquitto.lwt=mqttTopic:mqttMessagePayload:QoS:retainFlag`

_wanted usage of lwt property:_ 
`org.eclipse.smarthome.mqtt:mosquitto.lwt=mqtt/clients/client1:{"clientID":"client1","status":"online"}:2:1`

It takes "client1" as Quality Of Service and throws Exception because it is not a number at all.

Any thoughts on this one?

maggu2810:
AFAIK also the topic is allowed to contain `:`.

What about this one:
* If there is a property `lwt.topic`, use that topic
* If there is a property `lwt.message`, use that message
* If there is a property `lwt.qos`, use that QoS
* If there is a property `lwt.retain`, use that retain flag
* If some of the properties above are missing, try to parse the `lwt` property and use the missing part encoded in that property.

doandzhi:
Thank you for the response @maggu2810 ! Having a look in the [MqttWillAndTestament](https://github.com/eclipse/smarthome/blob/master/bundles/io/org.eclipse.smarthome.io.transport.mqtt/src/main/java/org/eclipse/smarthome/io/transport/mqtt/MqttWillAndTestament.java) I see that this `fromString(String string)` method splits the given parameter by `:` .  Concerning the last will setting after the session has been started I already mentioned this in #5162 . So this is impossible to build any logic around this at the moment (parsing the parameter and setting the lwt later) since it is set when the connection is established.

   Sorry my ignorace if I don't understand something properly but at the moment it is impossible for me to use  the `:` in my message payload because the property is not processed properly and there seems to be no alternative to use instead of this configuration property.

maggu2810:
But there is also a constructor that could be used:
```Java
public MqttWillAndTestament(String topic, byte[] payload, int qos, boolean retain)
```

[...]

PR #5231, opened 2018-03-12
MQTT last will and testament

maggu2810:
Add support to configure LWT topic, message, qos and retain flag explicit.

Fixes: https://github.com/eclipse/smarthome/issues/5197

### DIFF
--- a/bundles/io/org.eclipse.smarthome.io.transport.mqtt/src/main/java/org/eclipse/smarthome/io/transport/mqtt/MqttService.java
+++ b/bundles/io/org.eclipse.smarthome.io.transport.mqtt/src/main/java/org/eclipse/smarthome/io/transport/mqtt/MqttService.java
@@ -12,6 +12,7 @@
  */
 package org.eclipse.smarthome.io.transport.mqtt;
 
+import java.nio.charset.StandardCharsets;
 import java.util.Collection;
 import java.util.HashMap;
 import java.util.Iterator;
@@ -62,6 +63,10 @@ public class MqttService {
         public final @Nullable Integer qos;
         public final @Nullable Boolean retain;
         public final @Nullable String lwt;
+        public final @Nullable String lwtTopic;
+        public final byte @Nullable [] lwtMessage;
+        public final @Nullable Integer lwtQos;
+        public final @Nullable Boolean lwtRetain;
 
         public Config(final Map<String, String> cfg) {
             name = cfg.get(NAME_PROPERTY);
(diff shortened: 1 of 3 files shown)


maggu2810:
@doandzhi What do you think about?

bundles/io/org.eclipse.smarthome.io.transport.mqtt/src/main/java/org/eclipse/smarthome/io/transport/mqtt/MqttService.java
```diff
@@ -228,28 +295,31 @@ public boolean addBrokerConnection(MqttBrokerConnection connection) {
     }
 
     /**
-     * Add a broker by a configuration key-value map. You need to provide at least a "name" and an "url".
+     * Add a broker by a configuration.
+     *
+     * <p>
+     * You need to provide at least a "name" and an "url".
      * Additional properties are "user","pwd","qos","retain","lwt","keepAlive","clientId", please read the
```
doandzhi: Maybe it would be useful to mention the new properties here ( lwtTopic, lwtMessage lwtQos lwtRetain)

maggu2810:
> I assume that it is working fine.

Would you like to give it a try.
I don't use MQTT at the moment myself.

bundles/io/org.eclipse.smarthome.io.transport.mqtt/src/main/java/org/eclipse/smarthome/io/transport/mqtt/MqttService.java
maggu2810: The implementation has been changed to use a config object that properties are "fixed". As all fields are nullable and so optional the documentation needs to state which ones are required.
I don't know if it make sense to sync the additional properties all the time.
Perhaps we could remove the "Additional properties ..." part at all if this PR will be accepted (and before it is merged).

bundles/io/org.eclipse.smarthome.io.transport.mqtt/src/main/java/org/eclipse/smarthome/io/transport/mqtt/MqttService.java
```diff
@@ -228,28 +295,31 @@ public boolean addBrokerConnection(MqttBrokerConnection connection) {
...
      * service configuration documentation for a detailed description.
      *
-     * @param brokerConnectionConfig The configuration key-value map.
+     * @param cfg The configuration key-value map.
```
maggu2810: it is not a key-value map anymore

doandzhi:
> Would you like to give it a try.
I don't use MQTT at the moment myself.

Sure, I can do that! I will give feedback in a while.

doandzhi:
I can confirm that the proposed changes resolve issue #5197 . It is able to use JSON payload and no problems were faced. Thanks for the contribution @maggu2810 !

PR #5233, opened 2018-03-13
Display the time which was returned from the NTP server

sjka:
When trying to understand the root-cause of #5226, I happened to spot these innocent looking lines of code:

https://github.com/eclipse/smarthome/blob/0aad48c9b40aacf9292ac32432b3965fbe853257/extensions/binding/org.eclipse.smarthome.binding.ntp/src/main/java/org/eclipse/smarthome/binding/ntp/handler/NtpHandler.java#L217-L222

Note the `LocalDateTime.now()` - this actually is a come-back of #4867: It leads to the **local machine's time** to be displayed, not the time which was returned from the NTP server 😩.

Commit "[Update channel with NTP time, not LocalDateTime.now()](https://github.com/eclipse/smarthome/commit/6eee6aa0f306261e67973906ecdecec0a4d2085d)" fixes this. 

And while being there anyway I also applied some cosmetics:

* added the missing UoM bundles to the launch config
* applied the formatter
* dropped the inverse and therefore misleading asserThat-comments
* replaced hamcrest matchers with their simple junit siblings where possible
* use mockito for mocking


extensions/binding/org.eclipse.smarthome.binding.ntp.test/src/test/java/org/eclipse/smarthome/binding/ntp/test/NtpOSGiTest.java
```diff
@@ -267,8 +264,7 @@ public void testStringChannelDefaultTimeZoneUpdate() {
 
         String timeZoneFromItemRegistry = getStringChannelTimeZoneFromItemRegistry();
 
-        assertThat("The string channel was not updated with the right timezone", timeZoneFromItemRegistry,
-                is(anyOf(equalTo(expectedTimeZoneEEST), equalTo(expectedTimeZoneEET))));
+        assertThat(timeZoneFromItemRegistry, is(anyOf(equalTo(expectedTimeZoneEEST), equalTo(expectedTimeZoneEET))));
```
htreu: same as above: Is EEST and EET different and why is this asserted the one _or_ the other?

sjka:
> I know you are not the original author

Thanks for bothering _me_ with these questions anyway 😒 

I'd suggest opening another issue for these. Especially in the realm of #3950 where we have the NTP . server's answer fully under our control there is no reason for getting surprised by any daylight savings etc.

extensions/binding/org.eclipse.smarthome.binding.ntp.test/src/test/java/org/eclipse/smarthome/binding/ntp/test/NtpOSGiTest.java
```diff
@@ -499,25 +499,18 @@ private void assertEventIsReceived(UpdateEventType updateEventType, String chann
...
+        registerService(eventSubscriberMock);
 
         if (updateEventType.equals(UpdateEventType.HANDLE_COMMAND)) {
             ntpHandler.handleCommand(new ChannelUID("ntp:test:chan:1"), new StringType("test"));
         } else if (updateEventType.equals(UpdateEventType.CHANNEL_LINKED)) {
             ntpHandler.channelLinked(new ChannelUID("ntp:test:chan:1"));
         }
         waitForAssert(() -> {
-            assertTrue(eventSubscriberMock.isEventReceived);
+            verify(eventSubscriberMock).receive(ArgumentMatchers.any(Event.class));
```
maggu2810: I am not such familiar with Mockito (shame on me)... Does this also work if `receive` is also called in fron of the `waitForAssert` block? Or rephrase: Does this verify check if the method has been called "before"?

maggu2810:
Sorry for flooding this PR. I wanted to give it a try to review the commits separately, but it seems not be supported as the approval of one commit seems to be displayed identically to the approval of the whole PR.

extensions/binding/org.eclipse.smarthome.binding.ntp.test/src/test/java/org/eclipse/smarthome/binding/ntp/test/NtpOSGiTest.java
sjka: To my understanding, mockito simply records _all_ interactions and with the `verify(..)` call you simply make sure that _some_ matching interaction happened until now. So it doesn't block at this point and it also doesn't "eat up" one interaction if the `verify(...)` would be called twice.

By the way, if it would be relevant how many times a certain interaction happened, then there is `verify(mock, times(n))` etc.

sjka:
> Sorry for flooding this PR. 

No worries. I always wanted to try this too 😉 

> it seems not be supported as the approval of one commit seems to be displayed identically to the approval of the whole PR.

That's a pity, indeed!

extensions/binding/org.eclipse.smarthome.binding.ntp.test/src/test/java/org/eclipse/smarthome/binding/ntp/test/NtpOSGiTest.java
maggu2810: > By the way, if it would be relevant how many times a certain interaction happened, then there is `verify(mock, times(n))` etc. 

Ah, thanks.

maggu2810:
One remaining question:
You used good separated commits :+1: 
Should we keep that separation in the history (rebase and merge) or should we combine all that changes (squash and merge).
If someone later looks at the history, why a line has been changed, I would prefer to keep them separated.

sjka:
Afaik we are not _forced_ to squash - it's merely a safety measure to not accidentally include any malicious or ip infringing intermediate states into the history. But to _my_ knowledge it would to okay to rebase-and-merge whenever you think it make sense, especially since you have reviewed the commits individually.

PS: Now I only have to convince the maven build on Travis of my changes - the IDE already was happy 😩

maggu2810:
As Mockito is using reflection (AFAIK) it is the same black magic then Groovy :wink:

extensions/binding/org.eclipse.smarthome.binding.ntp.test/src/test/java/org/eclipse/smarthome/binding/ntp/test/NtpOSGiTest.java
sjka: Argh 😒 

Minor (but important) correction:

```
Wanted 1 time:
-> at org.eclipse.smarthome.binding.ntp.test.NtpOSGiTest.lambda$5(NtpOSGiTest.java:512)
But was 4 times. Undesired invocation:
-> at sun.reflect.GeneratedMethodAccessor11.invoke(Unknown Source)
	at org.eclipse.smarthome.binding.ntp.test.NtpOSGiTest.lambda$5(NtpOSGiTest.java:512)
```

I just learned that `verify(mock)` is equivalent to `verify(mock, times(1))`. My assumption that it is the same as `verify(mock, atLeastOnce())` apparently was wrong. Your question was pretty much to the point. Sorry!

sjka:
@maggu2810 as you were planning to rebase-merge this, I amended the last commit with a change according to https://github.com/eclipse/smarthome/pull/5233#discussion_r174427558, i.e. I inserted a  `atLeastOnce()` only [here](https://github.com/eclipse/smarthome/pull/5233/files#diff-2de3655377d64808b696466f342977e4R512).

maggu2810:
All fine @sjka, will merge it after Travis CI succeed.
Just to point this out again: I like your semantic separation of the changes very much.

What is your opinion about the short commit message (first line), I personally would like a verbose (as much as possible) short description that contains most possible information.
E.g. for this one, I would prefix all with "NTP: " (or "NTP test" or similar), so a `git log` identify that changing the launch configuration, mockito refactor, ... is done on the NTP (test) binding only and not the whole repo. Without such a prefix you need to show at the changed files to...

Don't get me wrong, I don't want to be fastidious, just want to improve our all skills :wink:

sjka:
> Don't get me wrong

No, I don't - and actually fully agree. 

Under the assumption that they're gonna get squashed away anyway it didn't make any sense to put many thoughts into the commit messages... I have updated the messages accordingly (no code changes).

maggu2810:
:+1:

PR #5234, opened 2018-03-13
fix reference mess when updating uninitialized things

### LINKED ISSUE #5215: BaseBridgeHandler.childHandlerInitialized will not be called if the bridge will be configured at runtime
**Steps to reproduce:**
Create a bridge device in PaperUI
Create a child device in PaperUI, but do not select the bridge
Open the child device configuration again and select now the bridge

**Expected result:**
The function childHandlerInitialized of the bridge will be called after the initialize function of child was called

**Current result:**
Only the initialize function of the child thing will be called.

sjka:
Whenever there was a ThingHandler which failed to initialize,
e.g. because of a missing/non-configured bridge, then the framework
would not updated this ThingHandler's Thing reference as soon
as there was an update on that Thing. As a result, the ThingHandler
operated on the old Thing instance with all its consequences.

fixes #5215

### DIFF
--- a/bundles/core/org.eclipse.smarthome.core.thing.test/src/test/java/org/eclipse/smarthome/core/thing/internal/ThingManagerOSGiJavaTest.java
+++ b/bundles/core/org.eclipse.smarthome.core.thing.test/src/test/java/org/eclipse/smarthome/core/thing/internal/ThingManagerOSGiJavaTest.java
@@ -25,6 +25,7 @@ import java.util.Dictionary;
 import java.util.Hashtable;
 import java.util.List;
 import java.util.Map;
+import java.util.concurrent.Semaphore;
 import java.util.concurrent.atomic.AtomicReference;
 import java.util.function.Function;
 
@@ -40,6 +41,7 @@ import org.eclipse.smarthome.core.common.SafeCaller;
 import org.eclipse.smarthome.core.items.ItemRegistry;
 import org.eclipse.smarthome.core.service.ReadyMarker;
 import org.eclipse.smarthome.core.service.ReadyService;
+import org.eclipse.smarthome.core.thing.Bridge;
 import org.eclipse.smarthome.core.thing.Channel;
 import org.eclipse.smarthome.core.thing.ChannelUID;
 import org.eclipse.smarthome.core.thing.ManagedThingProvider;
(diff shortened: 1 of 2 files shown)


bundles/core/org.eclipse.smarthome.core.thing.test/src/test/java/org/eclipse/smarthome/core/thing/internal/ThingManagerOSGiJavaTest.java
```diff
@@ -257,6 +265,228 @@ public void handleCommand(@NonNull ChannelUID channelUID, @NonNull Command comma
...
+        Bridge bridge = BridgeBuilder.create(BRIDGE_TYPE_UID, BRIDGE_UID).build();
+        managedThingProvider.add(bridge);
+        waitForAssert(() -> {
+            assertEquals(ThingStatus.ONLINE, bridge.getStatus());
+        });
+
+        Thing thing = ThingBuilder.create(THING_TYPE_UID, THING_UID).build();
+        managedThingProvider.add(thing);
+        waitForAssert(() -> {
+            assertEquals(ThingStatus.UNINITIALIZED, thing.getStatus());
```
maggu2810: The bridge is not required, correct?
So, wouldn't it be possible that we miss the uninitialized status (timing)?
Shouldn't we wait for online as this status should be reached and the thing should stay in?

maggu2810:
In the tests the bridge is not required. Correct?

Can give me some details about the expected behavior:
* bridge is created / added
* thing1 without an assigned bridge is created / added
* thing1 will not be initialized because a bridge is required and missing
* thing2 with the same UID as thing1 is created, thing2 has an assigned bridge (the above one)
* managed thing provider's update method is called with thing2
* childHandlerInitialized(...) must be called
* thingUpdated(...) is not called - because thing1 that required but miss a bridge has not been initialized before

Just to be sure: Is thing updated only be called if an initialized thing is updated by another initialized one?

bundles/core/org.eclipse.smarthome.core.thing.test/src/test/java/org/eclipse/smarthome/core/thing/internal/ThingManagerOSGiJavaTest.java
sjka: It technically (currently) is not required. But if it's missing, the handler throws a RuntimeException, therefore the ThingManager puts it back into UNINITIALIZED/HANDLER_INITIALIZATION_ERROR. We can't miss it here.

maggu2810: Ah ****, I missed the different `initialize` implementation as I tried to find the difference to the test below that waits for `ONLINE`.
Thanks and sorry :wink:

sjka:
> Just to be sure: Is thing updated only be called if an initialized thing is updated by another initialized one?

Correct. The assumption is that a handler which is not yet initialized is not capable of doing anything properly, therefore it also must not be bothered with thing updates:

https://github.com/eclipse/smarthome/blob/ec3cc58a493da1293c166513928ce92b8c85eeff/bundles/core/org.eclipse.smarthome.core.thing/src/main/java/org/eclipse/smarthome/core/thing/internal/ThingManager.java#L433-L435

It has been like that for quite a long time now.

You are implicitly bringing up a good point though: Currently we _are_ trying to initialize a handler, even though we could _know_ that its ThingType declared some supportedBridgeTypeUIDs. I actually think it would be safe to say that a thing handler with a type which declares such supportedBridgeTypeUIDs actually _requires_ a bridge. At least I'm not aware of any binding where the bridge is optional. So I think I'm gonna change that in [ThingManager.isInitializable(...)](https://github.com/eclipse/smarthome/blob/ec3cc58a493da1293c166513928ce92b8c85eeff/bundles/core/org.eclipse.smarthome.core.thing/src/main/java/org/eclipse/smarthome/core/thing/internal/ThingManager.java#L598) which makes the contract clearer and also eliminates one possible source of errors for ThingHandler implementations in bindings.

maggu2810:
> So I think I'm gonna change that

Would you like to add that change to this PR or to another one?

sjka:
I'll do that in another PR.

PR #5235, opened 2018-03-13
removed bridge-type config description URI from docu

sjka:
...in order to adapt the documentation to #50.


maggu2810:
Thanks

PR #5236, opened 2018-03-14
Paper UI: On item changed filter cached items

### LINKED ISSUE #5232: [PaperUI] performance issue
I'm not sure if this an issue or not. But I've got several "things" added in OH (72 channels in total for all of those things linked), one of the channels is constantly changing its value (every second, just a test) which causes a "items?recursive=false" request on every change. 

The problem is, all the responses for the "items?recursive=false" request contain ALL 72 links, e.g. even those 71 that never changes. 

I've been monitoring it for 1 hour, and it has downloaded 77mb of the "items?recursive=false" responses.

Is this something expected?

It is definitely making everything very slow, especially when the rest of the 72 channels start changing, because literally for every change PaperUI makes another request to get ALL 72 links.

Thanks

htreu:
Fixes #5232.

On item changes events filter items in the front end repository instead reloading the whole list.

Also:
* Refactor front end repositories to depend on abstract repository factory.
* Refactor repositories to be angularjs components.

### DIFF
--- a/extensions/ui/org.eclipse.smarthome.ui.paper/web-src/js/repositories/repositories-module.js
+++ b/extensions/ui/org.eclipse.smarthome.ui.paper/web-src/js/repositories/repositories-module.js
@@ -37,7 +37,7 @@
 
         function getAll(callback, refresh) {
             if (typeof callback === 'boolean') {
-                refresh = true;
+                refresh = callback;
                 callback = null;
             }
             var deferred = $q.defer();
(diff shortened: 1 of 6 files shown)


extensions/ui/org.eclipse.smarthome.ui.paper/web-src/js/repositories/repositories-module.js
```diff
@@ -0,0 +1,172 @@
...
+            setDirty : setDirty
+        };
+
+        function setDirty() {
+            self.dirty = true;
+        }
+
+        function getAll(callback, refresh) {
+            if (typeof callback === 'boolean') {
+                refresh = true;
```
sjka: for the `getAll(false)` case `refresh` should be assigned with the `callback` value

htreu:
Travis fails due to SafeCallerImplTest...

PR #5238, opened 2018-03-14
Define StateDescription readOnly optional - API breaking

### LINKED ISSUE #5237: DSL defined items miss readOnly setting from channel type
In Paper UI all item defined in the DSL are readOnly = false. This way temperature readings can be altered, etc.
The problem is that the first StateDescription found will define this setting. The GenericItemProvider in this case defines a default of `false`.

htreu:
Fixes #5237.

This makes the readOnly property of `StateDescription` optional as it is now modelled as `Boolean` and thus may be null.
The `ChannelConverter` reading the XML ChannelTypes defines a default value of false. This way the ChannelStateDescriptionProvider will as a last resort fall back to `false`.


bundles/core/org.eclipse.smarthome.core/src/main/java/org/eclipse/smarthome/core/types/StateDescription.java
```diff
@@ -98,7 +100,7 @@ public String getPattern() {
      *
      * @return true, if the state can only be read but not written
      */
-    public boolean isReadOnly() {
+    public @Nullable Boolean isReadOnly() {
```
maggu2810: As there is now a third possible value (`null`) that could be returned, we should change the JavaDoc and explain what is the intention of `null`.

maggu2810:
@kaikreuzer How should we handle such APIBreaking?
I normally update all ESH bundles or none, so the binary compatibility is for me not such important.
But IIRC it is important for some of you.

So, should we wait for some acks of different people in front of merging such stuff?

kaikreuzer:
@maggu2810 Right, such small changes can have a big impact as I have just [recently seen on openHAB](https://community.openhab.org/t/unit-support-knx-2-karaf-4-1-5-upgrade-and-more/41904/6?u=kai).
Also DT needs internal coordination on such changes - hence it would be welcome to only merge once at least one person from DT also approved it.
In general, we might want to leave a grace time of 2-3 days for breaking PRs, so that everyone has a chance to veto/comment before a merge.

sjka:
I fear in this case I will have to cast a veto!

This is indeed a pretty harmful change, as it not only breaks bindings but also "applications" sitting on top of the framework, doing something smart with it. A few `grep`s and asking around revealed several places which will cause unnecessary trouble for DT.

Conceptually this PR change absolutely makes sense, but I would suggest evolving the API in a compatible way, allowing a grace period having `@Deprecated` signatures. The constructor shouldn't be much of a problem (despite of looking awkward, `boolean` and `Boolean` essentially are different type). For the return type, you could simply add another method (e.g. `getReadOnly()`) and deprecate the `isReadOnly()` one, having it piggy-back on the new one but falling back to the previous default of `false` in case of `null`.

htreu:
I authored a new commit with a duplicated API just like @sjka suggested.

kaikreuzer:
I'd like to challenge the whole idea: On a closer look, I would claim that we are actually just mixing two different interfaces into a single one:
1. On the one side, we have the "state description providers", which are delivering partial/incomplete information about state descriptions. Those require a mechanism to express that they do NOT provide a certain information (that's why we want to use Boolean instead of boolean). For all those, the current way of using a constructor with xxx parameters where most are null is anyhow rather unpleasant - we should rather provide a builder for those.
2. On the other side there is the "state description service" which delivers complete state descriptions where "isReadOnly" will always return a boolean, but never null. By mixing both interfaces into one, we would not be able to express this.

I'd therefore think that the cleanest option would be to introduce a new `StateDescriptionFragment?/Part?/Xxx?` interface together with a builder. The `StateDescriptionService` could be kept fully backward compatible by querying the existing (old) `StateDescriptionProviders`as well as newly introduced `StateDescriptionFragmentProvider`s while returning the unchanged `StateDescription` interface and thus avoiding any API breakage.

htreu:
According to @kaikreuzer's comment I introduced a new infrastructure. The `StateDescriptionFragment` has all fields `@Nullable` and allows to provide only the parts of a StateDescription which are known to the current provider. The new `StateDescriptionFragmentProvider` replaces the legacy `StateDescriptionProvider` which is now marked deprecated.
The `StateDescriptionFragmentBuilder` should be used to create the new bean. It is also capable of merging StateDescription[Fragment]s via subsequent calls to `builder.withStateDescription[Fragment]`.

The old API is fully intact.

bundles/core/org.eclipse.smarthome.core.test/META-INF/MANIFEST.MF
```diff
@@ -9,6 +9,7 @@ Fragment-Host: org.eclipse.smarthome.core
 Import-Package: 
  groovy.lang,
  javax.measure.quantity,
+ net.bytebuddy,
```
maggu2810: Is this still necessary (WRT #5319)? I also don't identify the place you are using that package.

bundles/core/org.eclipse.smarthome.core.thing.xml/src/main/java/org/eclipse/smarthome/core/thing/xml/internal/StateDescriptionConverter.java
```diff
@@ -58,15 +58,15 @@ private BigDecimal toBigDecimal(Map<String, String> attributes, String attribute
...
+                throw new ConversionException(
+                        "The attribute '" + attribute + "' has not a valid decimal number format!", nfe);
             }
         }
 
         return defaultValue;
     }
 
-    private boolean toBoolean(Map<String, String> attributes, String attribute, boolean defaultValue) {
+    private boolean extractReadOnly(Map<String, String> attributes, String attribute, Boolean defaultValue) {
```
maggu2810: Why has this method be renamed from `toBoolean` to `extractReadOnly`?
Reading the implementation of the method, I don't see any specific `readOnly` attribute stuff.
Sure, it is used for the `readOnly` atrribute, but the attribute itself is still a parameter and so not limited to that only use case.

bundles/core/org.eclipse.smarthome.core/src/main/java/org/eclipse/smarthome/core/types/StateDescriptionFragment.java
```diff
@@ -0,0 +1,80 @@
...
+ *
+ */
+public interface StateDescriptionFragment {
+
+    /**
+     * Returns the minimum value of an item state.
+     *
+     * @return minimum value of an item state
+     */
+    public @Nullable BigDecimal getMinimum();
```
maggu2810: Methods defined in interfaces are already implicit `public`.

bundles/core/org.eclipse.smarthome.core/src/main/java/org/eclipse/smarthome/core/types/StateDescriptionFragmentBuilder.java
```diff
@@ -0,0 +1,172 @@
...
+    private @Nullable String pattern;
+    private @Nullable Boolean readOnly;
+    private @Nullable List<StateOption> options;
+
+    public static StateDescriptionFragmentBuilder instance() {
+        return new StateDescriptionFragmentBuilder();
+    }
+
+    public StateDescriptionFragment build() {
+        return new StateDescriptionFragment() {
```
maggu2810: Hm, I would expect that if I call `build` on a fragment and I change the fragment after that, that it does not change the previously build reference...

extensions/binding/org.eclipse.smarthome.binding.digitalstrom/src/main/java/org/eclipse/smarthome/binding/digitalstrom/internal/providers/DsChannelTypeProvider.java
```diff
@@ -25,6 +25,7 @@
 import java.util.Set;
 
 import org.eclipse.smarthome.binding.digitalstrom.DigitalSTROMBindingConstants;
+import org.eclipse.smarthome.binding.digitalstrom.internal.lib.structure.devices.Device;
```
maggu2810: I don't see how this is related with this PR (as only an import is added but no other code change is done).

bundles/core/org.eclipse.smarthome.core/src/main/java/org/eclipse/smarthome/core/types/StateDescriptionFragmentBuilder.java
```diff
@@ -0,0 +1,179 @@
...
+    }
+
+    /**
+     * Set the builder fields from all non-null fields of the given {@link StateDescriptionFragment}.
+     *
+     * @param fragment a {@link StateDescriptionFragment} this builder should incorporate.
+     * @return the builder.
+     */
+    public StateDescriptionFragmentBuilder withStateDescriptionFragment(StateDescriptionFragment fragment) {
+        if (fragment.getMinimum() != null) {
```
htreu: This logic seems to be wrong: The builder is supposed to merge in the given StateDescriptionFragment. So it should not override its internal fields if already set. It should only take values for still unknown values.

J-N-K: What if I have a state description (let‘s say min, max, readonly) and want to change readonly, but leave the others untouched. 
As far as I understood I should create a fragment with all fields null, except the readonly field. If already set fields are not overridden, this change will not take effect. Is that intended?

maggu2810: @J-N-K I would expect:

```Java
StateDescriptionFragmentBuilder builder = StateDescriptionFragmentBuilder.instance().withStateDescription(stateDescription).withReadOnly(newReadOnlyValue);
```

htreu: Yes, correct.
With the latest code and behaviour as described above I did refactor the `withStateDescription...` methods to `mergeStateDescription...` and also adopted the JavaDoc. I hope with this it is clear that `with...` does override but `merge...` does not.

htreu:
@maggu2810 this is ready for the next review. Thanks.

bundles/core/org.eclipse.smarthome.core/src/main/java/org/eclipse/smarthome/core/types/StateDescriptionFragmentBuilder.java
```diff
@@ -0,0 +1,239 @@
...
+    private @Nullable String pattern;
+    private @Nullable Boolean readOnly;
+    private @Nullable List<StateOption> options;
+
+    /**
+     * Return a fresh builder instance.
+     *
+     * @return a fresh {@link StateDescriptionFragmentBuilder} instance.
+     */
+    public static StateDescriptionFragmentBuilder instance() {
```
sjka: Maybe it's personal bias, but to me the name `instance()` implies that I will get the singleton instance of it. We have `ThingBuilder.create(...)` and `ChannelBuilder.create(...)`. Admittedly I overlooked  `ThingTypeBuilder.instance(...)`, so it became inconsistent already, but my personal preference would be to keep it as consistent as possible and stay with the majority...

bundles/core/org.eclipse.smarthome.core/src/main/java/org/eclipse/smarthome/core/types/StateDescriptionFragmentBuilder.java
```diff
@@ -0,0 +1,239 @@
...
+
+    /**
+     * Merge the given {@link StateDescription}. Set all unset ({@code null}) fields of this builder to the values from
+     * the
+     * given {@link StateDescription}.
+     *
+     * @param legacy a {@link StateDescription} this builder should merge in.
+     * @return the builder.
+     */
+    public StateDescriptionFragmentBuilder mergeStateDescription(StateDescription legacy) {
```
sjka: Wouldn't that rather be a canditate for a `public static ... create(StateDescription)` method? Or did I miss a use-case?

bundles/core/org.eclipse.smarthome.core/src/main/java/org/eclipse/smarthome/core/types/StateDescriptionFragmentBuilder.java
```diff
@@ -0,0 +1,239 @@
...
+    /**
+     * Merge the given {@link StateDescriptionFragment}. Set all unset ({@code null}) fields of this builder to the
+     * values
+     * from the
+     * given {@link StateDescriptionFragment}.
+     *
+     * @param fragment a {@link StateDescriptionFragment} this builder should merge in.
+     * @return the builder.
+     */
+    public StateDescriptionFragmentBuilder mergeStateDescriptionFragment(StateDescriptionFragment fragment) {
```
sjka: IMHO this method mixes the to use-cases again: One is the "providing" side, i.e. a convenient API to express what aspects an actor "knows" about the state description. The other one is to then merge all this information in the StateDescriptionService - and this is internal logic and shouldn't be exposed to bindings et al.

bundles/core/org.eclipse.smarthome.core/src/main/java/org/eclipse/smarthome/core/types/StateDescriptionFragmentBuilder.java
```diff
@@ -0,0 +1,239 @@
...
+     * @param options the {@link StateOption}s for the resulting {@link StateDescriptionFragment}.
+     * @return this builder.
+     */
+    public StateDescriptionFragmentBuilder withOptions(List<StateOption> options) {
+        this.options = options;
+        return this;
+    }
+
+    /**
+     * Merge the given {@link StateDescriptionFragment}. Set all unset ({@code null}) fields of this builder to the
```
sjka: The fact that it only merges yet unfilled parameters is surprising - the intuitive way would be to use all non-null values of `fragment` and override `this`'s fields with them.

To me this is rather an indicator that this shouldn't be a `merge...()`-method but rather a `static create(SDF)`  which pre-initializes the builder with the given values.

bundles/core/org.eclipse.smarthome.core/src/main/java/org/eclipse/smarthome/core/types/StateDescriptionFragmentBuilder.java
htreu: agreed.

bundles/core/org.eclipse.smarthome.core/src/main/java/org/eclipse/smarthome/core/types/StateDescriptionFragmentBuilder.java
htreu: Thanks for pointing me to it. I had the feeling too but was not able to turn it into code ;-)

htreu: For the `create(StateDescription legacy)` method your arguments hold true and I will refactor it that way.
For the StateDescriptionFragment merging I will try to refactor it in a way which is less surprising. Maybe an extra class or directly in the StateDescriptionService.

htreu:
@sjka friendly ping :-)

bundles/core/org.eclipse.smarthome.core/src/main/java/org/eclipse/smarthome/core/internal/types/StateDescriptionFragmentImpl.java
```diff
@@ -0,0 +1,171 @@
...
+
+    /**
+     * Merge the given {@link StateDescriptionFragment}. Set all unset ({@code null}) fields of this instance to the
+     * values from the given {@link StateDescriptionFragment}.
+     *
+     * @param fragment a {@link StateDescriptionFragment} this instance should merge in.
+     * @return this instance with the fields merged.
+     */
+    public StateDescriptionFragment merge(StateDescriptionFragment fragment) {
+        if (this.minimum == null) {
```
sjka: I'm still surprised 😉 Wouldn't the intuitive way be 

```java
if (fragment.getMinimum() != null) {
  this.minimum = fragement.getMinimum();
}
```

?
At least _I_ would expect a "newest wins" strategy here (and for all others below). Which would be in compliance with 

> the higher [the rank], the more like it [the provided fragment] is going to win

in [StateDescriptionFragmentProvider.getRank()](https://github.com/eclipse/smarthome/pull/5238/files#diff-71d98e857913c7aaf3a5650f924b1a68R51).

htreu: See also this test case: https://github.com/eclipse/smarthome/pull/5238/files#diff-df777383266487f98708b37c495e6960R175

htreu:
rebased on master.

bundles/core/org.eclipse.smarthome.core/src/main/java/org/eclipse/smarthome/core/internal/types/StateDescriptionFragmentImpl.java
sjka: You're right:

https://github.com/eclipse/smarthome/blob/a31b18d194fe4310d29885db0dd924955989c7eb/bundles/core/org.eclipse.smarthome.core/src/main/java/org/eclipse/smarthome/core/internal/service/StateDescriptionServiceImpl.java#L43-L49

Had to read it three times until it clicked... Wich brings me back to my most favorite "principle of least surprise": Shouldn't we also swap the comparison in order to sort it the other way round then?

htreu: Aren't these implementation details? I don't mind turning both logics around, but I would like to have this one merged first. And then concentrate on this specific topic on a new PR. wdyt?

PR #5240, opened 2018-03-15
Renamed methods and local variables reported by the Static Code Analysis Tool

kssimova:
Renamed methods and local variables reported by the Static Code Analysis Tool and suppressed *org.eclipse.smarthome.binding.astro.internal.calc.MoonCalc.java*, *org.eclipse.smarthome.core.library.unit.MetricPrefix.java* for MethodNameCheck, as they contain specific methods with calculations and renaming them will make the code hard to read.

### DIFF
--- a/bundles/automation/org.eclipse.smarthome.automation.provider.file/src/main/java/org/eclipse/smarthome/automation/internal/provider/file/WatchServiceUtil.java
+++ b/bundles/automation/org.eclipse.smarthome.automation.provider.file/src/main/java/org/eclipse/smarthome/automation/internal/provider/file/WatchServiceUtil.java
@@ -25,7 +25,7 @@ import java.util.Map;
 @SuppressWarnings("rawtypes")
 public class WatchServiceUtil {
 
-    static Map<AbstractFileProvider, Map<String, AutomationWatchService>> WATCH_SERVICES = new HashMap<AbstractFileProvider, Map<String, AutomationWatchService>>();
+    private static final Map<AbstractFileProvider, Map<String, AutomationWatchService>> WATCH_SERVICES = new HashMap<AbstractFileProvider, Map<String, AutomationWatchService>>();
 
     public static void initializeWatchService(String watchingDir, AbstractFileProvider provider) {
         AutomationWatchService aws = null;
--- a/bundles/automation/org.eclipse.smarthome.automation.providers/src/main/java/org/eclipse/smarthome/automation/internal/core/provider/AbstractResourceBundleProvider.java
+++ b/bundles/automation/org.eclipse.smarthome.automation.providers/src/main/java/org/eclipse/smarthome/automation/internal/core/provider/AbstractResourceBundleProvider.java
@@ -82,7 +82,7 @@ public abstract class AbstractResourceBundleProvider<E> {
      * This static field provides a root directory for automation object resources in the bundle resources.
      * It is common for all resources - {@link ModuleType}s, {@link RuleTemplate}s and {@link Rule}s.
      */
-    protected static String PATH = "ESH-INF/automation";
+    protected static final String ROOT_DIRECTORY = "ESH-INF/automation";
 
     /**
      * This field holds a reference to the service instance for internationalization support within the platform.
--- a/bundles/automation/org.eclipse.smarthome.automation.sample.extension.java/src/main/java/org/eclipse/smarthome/automation/sample/extension/java/internal/type/AirConditionerTriggerType.java
+++ b/bundles/automation/org.eclipse.smarthome.automation.sample.extension.java/src/main/java/org/eclipse/smarthome/automation/sample/extension/java/internal/type/AirConditionerTriggerType.java
@@ -27,7 +27,7 @@ import org.eclipse.smarthome.automation.type.TriggerType;
  */
 public class AirConditionerTriggerType extends TriggerType {
 
-    public static String UID = "AirConditionerTrigger";
+    public static final String UID = "AirConditionerTrigger";
 
     public static TriggerType initialize() {
         List<Output> output = new ArrayList<Output>();
--- a/bundles/core/org.eclipse.smarthome.core.audio/src/main/java/org/eclipse/smarthome/core/audio/AudioFormat.java
+++ b/bundles/core/org.eclipse.smarthome.core.audio/src/main/java/org/eclipse/smarthome/core/audio/AudioFormat.java
@@ -24,20 +24,20 @@ import java.util.Set;
 public class AudioFormat {
 
     // generic mp3 format without any further constraints
-    public static AudioFormat MP3 = new AudioFormat(AudioFormat.CONTAINER_NONE, AudioFormat.CODEC_MP3, null, null, null,
-            null);
-
-    // generic wav format without any further constraints
-    public static AudioFormat WAV = new AudioFormat(AudioFormat.CONTAINER_WAVE, AudioFormat.CODEC_PCM_SIGNED, null,
-            null, null, null);
-
-    // generic OGG format without any further constraints
-    public static AudioFormat OGG = new AudioFormat(AudioFormat.CONTAINER_OGG, AudioFormat.CODEC_VORBIS, null, null,
+    public static final AudioFormat MP3 = new AudioFormat(AudioFormat.CONTAINER_NONE, AudioFormat.CODEC_MP3, null, null,
             null, null);
 
+    // generic wav format without any further constraints
+    public static final AudioFormat WAV = new AudioFormat(AudioFormat.CONTAINER_WAVE, AudioFormat.CODEC_PCM_SIGNED,
+            null, null, null, null);
+
+    // generic OGG format without any further constraints
+    public static final AudioFormat OGG = new AudioFormat(AudioFormat.CONTAINER_OGG, AudioFormat.CODEC_VORBIS, null,
+            null, null, null);
+
     // generic AAC format without any further constraints
-    public static AudioFormat AAC = new AudioFormat(AudioFormat.CONTAINER_NONE, AudioFormat.CODEC_AAC, null, null, null,
-            null);
+    public static final AudioFormat AAC = new AudioFormat(AudioFormat.CONTAINER_NONE, AudioFormat.CODEC_AAC, null, null,
+            null, null);
 
     /**
      * {@link AudioCodec} encoded data without any container header or footer,
--- a/bundles/core/org.eclipse.smarthome.core.thing/src/main/java/org/eclipse/smarthome/core/thing/ThingStatusDetail.java
+++ b/bundles/core/org.eclipse.smarthome.core.thing/src/main/java/org/eclipse/smarthome/core/thing/ThingStatusDetail.java
@@ -37,9 +37,9 @@ public enum ThingStatusDetail {
      */
     GONE;
 
-    public static OnlineStatus ONLINE = new OnlineStatus();
+    public static final OnlineStatus ONLINE = new OnlineStatus();
 
-    public static OfflineStatus OFFLINE = new OfflineStatus();
+    public static final OfflineStatus OFFLINE = new OfflineStatus();
 
     public static final class OnlineStatus {
         private OnlineStatus() {
--- a/bundles/core/org.eclipse.smarthome.core.thing/src/main/java/org/eclipse/smarthome/core/thing/firmware/ProgressCallbackImpl.java
+++ b/bundles/core/org.eclipse.smarthome.core.thing/src/main/java/org/eclipse/smarthome/core/thing/firmware/ProgressCallbackImpl.java
@@ -37,7 +37,7 @@ import org.osgi.framework.FrameworkUtil;
  */
 final class ProgressCallbackImpl implements ProgressCallback {
 
-    private static String UPDATE_CANCELED_MESSAGE_KEY = "update-canceled";
+    private static final String UPDATE_CANCELED_MESSAGE_KEY = "update-canceled";
 
     /**
      * Handler instance is needed to retrieve the error messages from the correct bundle.
--- a/bundles/core/org.eclipse.smarthome.core/src/main/java/org/eclipse/smarthome/core/library/unit/ImperialUnits.java
+++ b/bundles/core/org.eclipse.smarthome.core/src/main/java/org/eclipse/smarthome/core/library/unit/ImperialUnits.java
@@ -38,7 +38,7 @@ import tec.uom.se.unit.Units;
 @NonNullByDefault
 public class ImperialUnits extends SmartHomeUnits {
 
-    private static ImperialUnits INSTANCE = new ImperialUnits();
+    private static final ImperialUnits INSTANCE = new ImperialUnits();
 
     private ImperialUnits() {
         // avoid external instantiation
--- a/bundles/core/org.eclipse.smarthome.core/src/main/java/org/eclipse/smarthome/core/library/unit/SIUnits.java
+++ b/bundles/core/org.eclipse.smarthome.core/src/main/java/org/eclipse/smarthome/core/library/unit/SIUnits.java
@@ -36,7 +36,7 @@ import tec.uom.se.unit.Units;
 @NonNullByDefault
 public class SIUnits extends SmartHomeUnits {
 
-    private static SIUnits INSTANCE = new SIUnits();
+    private static final SIUnits INSTANCE = new SIUnits();
 
     private SIUnits() {
         // avoid external instantiation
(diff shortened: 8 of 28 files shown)


bundles/automation/org.eclipse.smarthome.automation.providers/src/main/java/org/eclipse/smarthome/automation/internal/core/provider/AbstractResourceBundleProvider.java
```diff
@@ -82,7 +82,7 @@ public AbstractResourceBundleProvider() {
      * This static field provides a root directory for automation object resources in the bundle resources.
      * It is common for all resources - {@link ModuleType}s, {@link RuleTemplate}s and {@link Rule}s.
      */
-    protected static String PATH = "ESH-INF/automation";
+    protected static String rootDirectory = "ESH-INF/automation";
```
htreu: I suspect this to be a constant, so `ROOT_DIRECTORY` would fit it best.

bundles/automation/org.eclipse.smarthome.automation.sample.extension.java/src/main/java/org/eclipse/smarthome/automation/sample/extension/java/internal/type/AirConditionerTriggerType.java
```diff
@@ -27,7 +27,7 @@
  */
 public class AirConditionerTriggerType extends TriggerType {
 
-    public static String UID = "AirConditionerTrigger";
+    public static String uid = "AirConditionerTrigger";
```
htreu: This also seems to be a constant.

bundles/core/org.eclipse.smarthome.core.audio/src/main/java/org/eclipse/smarthome/core/audio/AudioFormat.java
```diff
@@ -24,19 +24,19 @@
 public class AudioFormat {
 
     // generic mp3 format without any further constraints
-    public static AudioFormat MP3 = new AudioFormat(AudioFormat.CONTAINER_NONE, AudioFormat.CODEC_MP3, null, null, null,
+    public static AudioFormat mp3 = new AudioFormat(AudioFormat.CONTAINER_NONE, AudioFormat.CODEC_MP3, null, null, null,
```
htreu: constant

bundles/core/org.eclipse.smarthome.core.audio/src/main/java/org/eclipse/smarthome/core/audio/AudioFormat.java
```diff
@@ -24,19 +24,19 @@
...
             null);
 
     // generic wav format without any further constraints
-    public static AudioFormat WAV = new AudioFormat(AudioFormat.CONTAINER_WAVE, AudioFormat.CODEC_PCM_SIGNED, null,
+    public static AudioFormat wav = new AudioFormat(AudioFormat.CONTAINER_WAVE, AudioFormat.CODEC_PCM_SIGNED, null,
```
htreu: constant

bundles/core/org.eclipse.smarthome.core.audio/src/main/java/org/eclipse/smarthome/core/audio/AudioFormat.java
```diff
@@ -24,19 +24,19 @@
...
             null, null, null);
 
     // generic OGG format without any further constraints
-    public static AudioFormat OGG = new AudioFormat(AudioFormat.CONTAINER_OGG, AudioFormat.CODEC_VORBIS, null, null,
+    public static AudioFormat ogg = new AudioFormat(AudioFormat.CONTAINER_OGG, AudioFormat.CODEC_VORBIS, null, null,
```
htreu: constant

bundles/core/org.eclipse.smarthome.core.audio/src/main/java/org/eclipse/smarthome/core/audio/AudioFormat.java
```diff
@@ -24,19 +24,19 @@
...
             null, null);
 
     // generic AAC format without any further constraints
-    public static AudioFormat AAC = new AudioFormat(AudioFormat.CONTAINER_NONE, AudioFormat.CODEC_AAC, null, null, null,
+    public static AudioFormat aac = new AudioFormat(AudioFormat.CONTAINER_NONE, AudioFormat.CODEC_AAC, null, null, null,
```
htreu: constant

bundles/core/org.eclipse.smarthome.core.thing/src/main/java/org/eclipse/smarthome/core/thing/ThingStatusDetail.java
```diff
@@ -37,9 +37,9 @@
      */
     GONE;
 
-    public static OnlineStatus ONLINE = new OnlineStatus();
+    public static OnlineStatus online = new OnlineStatus();
```
htreu: constant

bundles/core/org.eclipse.smarthome.core.thing/src/main/java/org/eclipse/smarthome/core/thing/ThingStatusDetail.java
```diff
@@ -37,9 +37,9 @@
...
 
-    public static OfflineStatus OFFLINE = new OfflineStatus();
+    public static OfflineStatus offline = new OfflineStatus();
```
htreu: constant

bundles/core/org.eclipse.smarthome.core.thing/src/main/java/org/eclipse/smarthome/core/thing/firmware/ProgressCallbackImpl.java
```diff
@@ -37,7 +37,7 @@
  */
 final class ProgressCallbackImpl implements ProgressCallback {
 
-    private static String UPDATE_CANCELED_MESSAGE_KEY = "update-canceled";
+    private static String updateCanceledMessageKey = "update-canceled";
```
htreu: constant

kssimova:
@htreu I've updated this PR, thank you for your comments! :)

bundles/automation/org.eclipse.smarthome.automation.provider.file/src/main/java/org/eclipse/smarthome/automation/internal/provider/file/WatchServiceUtil.java
```diff
@@ -25,15 +25,15 @@
 @SuppressWarnings("rawtypes")
 public class WatchServiceUtil {
 
-    static Map<AbstractFileProvider, Map<String, AutomationWatchService>> WATCH_SERVICES = new HashMap<AbstractFileProvider, Map<String, AutomationWatchService>>();
+    static Map<AbstractFileProvider, Map<String, AutomationWatchService>> watchServices = new HashMap<AbstractFileProvider, Map<String, AutomationWatchService>>();
```
htreu: This should also be `private final static Map<...> WATCH_SERVICES` as it is only used in this util class. Further this util class is not supposed to be extended by inheritance.

bundles/core/org.eclipse.smarthome.core/src/main/java/org/eclipse/smarthome/core/library/unit/ImperialUnits.java
```diff
@@ -35,7 +35,7 @@
  */
 public class ImperialUnits extends SmartHomeUnits {
 
-    private static ImperialUnits INSTANCE = new ImperialUnits();
+    private static ImperialUnits instance = new ImperialUnits();
```
htreu: this is also a constant, `private static final ... INSTANCE` please.

bundles/core/org.eclipse.smarthome.core/src/main/java/org/eclipse/smarthome/core/library/unit/SIUnits.java
```diff
@@ -33,7 +33,7 @@
  */
 public class SIUnits extends SmartHomeUnits {
 
-    private static SIUnits INSTANCE = new SIUnits();
+    private static SIUnits instance = new SIUnits();
```
htreu: this is also a constant, `private static final ... INSTANCE` please.

bundles/core/org.eclipse.smarthome.core/src/main/java/org/eclipse/smarthome/core/library/unit/SmartHomeUnits.java
```diff
@@ -64,7 +64,7 @@
  */
 public class SmartHomeUnits extends AbstractSystemOfUnits {
 
-    private static SmartHomeUnits INSTANCE = new SmartHomeUnits();
+    private static SmartHomeUnits instance = new SmartHomeUnits();
```
htreu: this is also a constant, `private static final ... INSTANCE` please.

bundles/test/org.eclipse.smarthome.magic/src/main/java/org/eclipse/smarthome/magic/binding/handler/MagicDelayedOnlineHandler.java
```diff
@@ -31,7 +31,7 @@
 @NonNullByDefault
 public class MagicDelayedOnlineHandler extends BaseThingHandler {
 
-    public static int DELAY = 15;
+    public static int delay = 15;
```
htreu: this is also a constant, `private static final int DELAY` please.

extensions/binding/org.eclipse.smarthome.binding.digitalstrom/src/main/java/org/eclipse/smarthome/binding/digitalstrom/handler/CircuitHandler.java
```diff
@@ -59,7 +59,7 @@
     /**
      * Contains all supported thing types of this handler, will be filled by DsDeviceThingTypeProvider.
      */
-    public static Set<ThingTypeUID> SUPPORTED_THING_TYPES = new HashSet<ThingTypeUID>();
+    public static Set<ThingTypeUID> circuitHandlerSupportedThingTypes = new HashSet<ThingTypeUID>();
```
htreu: this is also a constant: No other entity is allowed to change the reference. So: `public static final Set<ThingTypeUID> SUPPORTED_THING_TYPES` please.

extensions/binding/org.eclipse.smarthome.binding.digitalstrom/src/main/java/org/eclipse/smarthome/binding/digitalstrom/handler/DeviceHandler.java
```diff
@@ -83,10 +83,10 @@
     /**
      * Contains all supported thing types of this handler, will be filled by DsDeviceThingTypeProvider.
      */
-    public static Set<ThingTypeUID> SUPPORTED_THING_TYPES = new HashSet<ThingTypeUID>();
+    public static Set<ThingTypeUID> deviceHandlerSupportedThingTypes = new HashSet<ThingTypeUID>();
```
htreu: see above.

extensions/binding/org.eclipse.smarthome.binding.digitalstrom/src/main/java/org/eclipse/smarthome/binding/digitalstrom/handler/ZoneTemperatureControlHandler.java
```diff
@@ -72,7 +72,7 @@
     /**
      * Contains all supported thing types of this handler
      */
-    public static Set<ThingTypeUID> SUPPORTED_THING_TYPES = new HashSet<>(
+    public static Set<ThingTypeUID> zoneTempControlHandlerSupportedThingTypes = new HashSet<>(
```
htreu: see above.

extensions/binding/org.eclipse.smarthome.binding.sonos/src/main/java/org/eclipse/smarthome/binding/sonos/internal/SonosXMLParser.java
```diff
@@ -45,7 +45,7 @@
 
     static final Logger LOGGER = LoggerFactory.getLogger(SonosXMLParser.class);
 
-    private static MessageFormat METADATA_FORMAT = new MessageFormat(
+    private static MessageFormat metadataFormat = new MessageFormat(
```
htreu: constant, `private static final MessageFormat METADATA_FORMAT` please.

extensions/binding/org.eclipse.smarthome.binding.wemo/src/main/java/org/eclipse/smarthome/binding/wemo/handler/WemoLightHandler.java
```diff
@@ -81,7 +81,7 @@
     /**
      * The default refresh initial delay in Seconds.
      */
-    private static int DEFAULT_REFRESH_INITIAL_DELAY = 15;
+    private static int defaultRefreshInitialDelay = 15;
```
htreu: constant, `private static final int DEFAULT_REFRESH_INITIAL_DELAY` please.

extensions/ui/iconset/org.eclipse.smarthome.ui.iconset.classic/src/main/java/org/eclipse/smarthome/ui/iconset/classic/internal/ClassicIconProvider.java
```diff
@@ -37,7 +37,7 @@
 
     private final Logger logger = LoggerFactory.getLogger(ClassicIconProvider.class);
 
-    static String ICONSET_ID = "classic";
+    static String iconSetId = "classic";
```
htreu: constant, should at least be `static final String ICONSET_ID`

extensions/binding/org.eclipse.smarthome.binding.wemo/src/main/java/org/eclipse/smarthome/binding/wemo/handler/WemoLightHandler.java
kssimova: It can not be a constant, because in [WemoLightHandlerOSGiTest](https://github.com/MusalaSoft/smarthome/blob/632e93ef832b30ea29cd0efed4ca77a09a237e11/extensions/binding/org.eclipse.smarthome.binding.wemo.test/src/main/groovy/org/eclipse/smarthome/binding/wemo/handler/test/WemoLightHandlerOSGiTest.groovy#L53) for the test we set it to 1 second.

htreu: From a coding style pov the test must not access this field (since its `private` anyway, but only groovy doesn't care).
We could just drop the line from the test which will result in very slow execution (waiting 15s for each of the 7 test cases).
The other option would be to rework this and add a configuration property. @hmerk do you have any suggestion on how to deal with this issue? background: static analysis complains about this "variable" being all caps. And from the looks it should be a constant.

hmerk: Yes, it is a constant, therefore it is capitalized and somebody askedd me to change it a while ago. But if code analasys has an issue, we could remove it and hard code it in the according method.
I can put it on my long list of needed/wanted changes but can not tell when to have my updates ready.

htreu: Thanks for the very quick reply @hmerk.
It was indeed introduced to override it during test execution. My suggestion is to revert it to a constant and remove the overriding in the test.
@kssimova could you go on with this? The tests will be super slow then but we should address that in a separate PR if needed.

maggu2810: I agree, we should improve the test case later.
Perhaps we could also mitgrate it from Groovy to pure Java.

bundles/core/org.eclipse.smarthome.core.audio/src/main/java/org/eclipse/smarthome/core/audio/AudioFormat.java
```diff
@@ -24,19 +24,23 @@
 public class AudioFormat {
 
     // generic mp3 format without any further constraints
-    public static AudioFormat MP3 = new AudioFormat(AudioFormat.CONTAINER_NONE, AudioFormat.CODEC_MP3, null, null, null,
+    public static final AudioFormat MP3 = new AudioFormat(AudioFormat.CONTAINER_NONE, AudioFormat.CODEC_MP3, null, null,
+            null,
```
htreu: please run the auto formatter for this file once. Lines 28 & 29 should be on one line. And so should the other "hanging" parameters.

kssimova: Done :)

PR #5241, opened 2018-03-16
[audio] Fixed stop currently playing streams

cweitkamp:
- Added `@Nullable` for `play(audioStream, ...)` methods in `AudioManager` API
- Added annotations for `AudioSink` API
- Removed  redundant `ǹull` checks

This fix reverts a regression introduced with #4963 . It wasn't possible to stop currently playing stream anymore.

### DIFF
--- a/bundles/core/org.eclipse.smarthome.core.audio.test/src/test/groovy/org/eclipse/smarthome/core/audio/test/AudioManagerTest.groovy
+++ b/bundles/core/org.eclipse.smarthome.core.audio.test/src/test/groovy/org/eclipse/smarthome/core/audio/test/AudioManagerTest.groovy
@@ -37,16 +37,19 @@ public class AudioManagerTest extends AudioOSGiTest {
     }
 
     @Test
-    public void 'null streams are not processed'(){
+    public void 'null streams are processed'(){
         registerSink()
 
         audioManager.play(null, audioSinkFake.getId())
 
-        waitForAssert({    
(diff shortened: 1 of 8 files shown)


bundles/core/org.eclipse.smarthome.core.audio.test/src/test/groovy/org/eclipse/smarthome/core/audio/test/AudioManagerTest.groovy
```diff
@@ -36,10 +36,17 @@ public class AudioManagerTest extends AudioOSGiTest {
...
+    @Test
     public void 'null streams are not processed'(){
         registerSink()
+
         audioManager.play(null, audioSinkFake.getId())
+
+        waitForAssert({    
+            assertThat "null stream was processed",    
+                    audioSinkFake.isStreamProcessed,
+                    is(false)
```
htreu: The JavaDoc says that `null` streams should stop the current running stream. Isn't this also a sort of processing? Maybe the `AudioSinkFake` should indicate that the fake stream was stopped and the test should not talk about "not processed". wdyt?

cweitkamp: Yes, I saw this too. We should indicate whether a processed stream is playing or not.

Issue #5182, opened 2018-03-05
Remove ServiceTracker from BaseThingHandler


htreu:
The abstract `BaseThingHandler` uses OSGi ServiceTrackers to gain access to the Thing-, Link- and TypeTypeRegistry and to the ConfigDescriptionValidatorService. It also holds a reference to the BundleContext.
We don't want bindings to access these services directly but restrict the access and encapsulate it in the ThingHandlerCallback.

lolodomo:
I search which are the bindings using thingRegistry from `BaseThingHandler` and I found only 2:

- Sonos binding in ESH: it is to get the thing handler of a Sonos group coordinator
- Allplay binding in OH2

I found 3 other bindings in OH2 using directly `ThingRegistry`:

- Powermax binding: it is for the console extension => normal I think
- amazon echo control binding
- neeo IO binding: I don't know why but this binding requires all the ESH registries !

This would be interesting to check if there is not a conception problem with the neeo binding.

lolodomo:
@tmrobert8 : can you comment regarding the neeo binding please ?

kaikreuzer:
The NEEO binding is not a binding in the classical sense - it also includes an IO bundle, i.e. it exposes the entire openHAB to an external system; for this reason it requires access to quite some registries.
It definitely is no code people should look at when they want to learn how to implement a pure binding!

tmrobert8:
@lolodomo - @kaikreuzer pretty succinctly stated the IO portion usage and I don't have much more to add.  Please let me know if you need any more details..

lolodomo:
I found 2 other bindings using the `ThingRegistry` (not through `BaseThingHandler`):

- Zigbee binding: it is for the console extension (like the Powermax binding) => normal I think
- Zwave binding

To comeback to the original topic (removing `thingRegistry` from `BaseThingHandler` or not exposing it), I can find 2 possible workarounds for the Sonos and AllPlay bindings:

1. Declare the link to ThingRegistry in the binding itself
2. Add and expose a method `getAllThings` in the `BaseThingHandler` that will return the list of all things supported by the binding. This solution suppose that `thingRegistry` is not removed from `BaseThingHandler` but only made hidden (private).

So, to be clear, if we want to remove `ThingRegistry` from `BaseThingHandler` quickly, we can first implement the solution 1 to the Sonos and AllPlay bindings and then it can be done.

htreu:
[...]

PR #5244, opened 2018-03-17
[core] Removed access to ItemChannelLinkRegistry from BaseThingHandler

cweitkamp:
- Moved `ItemChannelLinkRegistry` from `BaseThingHandler` to `ThingHandlerCallback`

A start for #5182 to make sure my approach goes into the right direction.

CC @htreu @sjka wdyt?

### DIFF
--- a/bundles/core/org.eclipse.smarthome.core.thing/src/main/java/org/eclipse/smarthome/core/thing/binding/BaseThingHandler.java
+++ b/bundles/core/org.eclipse.smarthome.core.thing/src/main/java/org/eclipse/smarthome/core/thing/binding/BaseThingHandler.java
@@ -36,7 +36,6 @@ import org.eclipse.smarthome.core.thing.ThingTypeUID;
 import org.eclipse.smarthome.core.thing.ThingUID;
 import org.eclipse.smarthome.core.thing.binding.builder.ThingBuilder;
 import org.eclipse.smarthome.core.thing.binding.builder.ThingStatusInfoBuilder;
-import org.eclipse.smarthome.core.thing.link.ItemChannelLinkRegistry;
 import org.eclipse.smarthome.core.thing.type.ThingType;
 import org.eclipse.smarthome.core.thing.type.ThingTypeRegistry;
 import org.eclipse.smarthome.core.thing.util.ThingHandlerHelper;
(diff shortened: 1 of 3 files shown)


cweitkamp:
Sounds good. Yes, I will do that.

cweitkamp:
I found some (four) occurrences in the OH2 Z-Way Binding ([ZWayDeviceHandler.java](https://github.com/openhab/openhab2-addons/blob/master/addons/binding/org.openhab.binding.zway/src/main/java/org/openhab/binding/zway/handler/ZWayDeviceHandler.java)) which could be handled by a new method `getChannelLinkedItems()` on the `ThingHandlerCallback`.

https://github.com/openhab/openhab2-addons/blob/4427e1a48a905b4fc42d3bd9d7601b7765c97e20/addons/binding/org.openhab.binding.zway/src/main/java/org/openhab/binding/zway/handler/ZWayDeviceHandler.java#L117

Additionally I found access to the `ItemChannelLinkRegistry` in the unit tests of the ESH Hue Binding ([HueLightHandlerOSGiTest.groovy](https://github.com/eclipse/smarthome/blob/master/extensions/binding/org.eclipse.smarthome.binding.hue.test/src/test/groovy/org/eclipse/smarthome/binding/hue/test/HueLightHandlerOSGiTest.groovy)). I am not sure if we have to remove those now. Maybe in the light of #3305.

maggu2810:
> which could be handled by a new method `getChannelLinkedItems()` on the `ThingHandlerCallback`.

Should the base framework really provide a mechanism that a handler can access the linked items for a channel?
Isn't it something that should normally not be done and so solved by the special handler itself?

htreu:
When I understand the ZWay code right it is about registering the openHAB item names in the physical device's properties. To me this seems to be a nice-to-have feature but not necessarily required to successfully run the binding.
Imho one goal from removing the protected fields someday is also to prohibit access to the items from a binding. So the ZWay code already behaves bad here.

kaikreuzer:
> Should the base framework really provide a mechanism that a handler can access the linked items for a channel?

No, removing the registry from access of the handler is actually specifically in order to avoid that this is done. The only required information should be whether it is linked or not.

cweitkamp:
Let's sum up what has to be done next:
- This PR is fine and I can either extend it or we can merge this one before I will proceed working on #5182.
- Nothing has to be done in the ESH Hue Binding unit tests, yet.
- We will ask the developer(s) of the OH2 Z-Way Binding to remove the use of the `linkRegistry` or shall I do it without permission?

Wdyt?

maggu2810:
As we agree that the handler should normally not access that registry, I think we could merge this PR.
But it would be fair to notify the Z-Way developers that they know about that change.

kaikreuzer:
> shall I do it without permission?

You could create according PRs and have the authors approve it - I would then merge them as soon as the ESH change is available on openHAB side.

kaikreuzer:
@cweitkamp Did you check whether the failing test on Travis is related to this PR?

cweitkamp:
@kaikreuzer Yes, I did. It is not related. Travis complains about #5248 which is solved now. Shall I rebase my branch?

kaikreuzer:
No need, I will simply restart Travis (which by itself does a rebase on master for the build).

PR #5254, opened 2018-03-19
[webaudio] Switched to auto-generated DS descriptor and minor Improvements

cweitkamp:
- Switched to auto-generated DS descriptor
- Minor code improvements


sjka:
Thanks!

Issue #5248, opened 2018-03-18
Thing tests repeatedly fail


J-N-K:
in PR #5150 Travis failed three times in row in the thing tests (also affected #5244, #5246) 

```assert BaseThingHandlerFactory registers config status provider(org.eclipse.smarthome.core.thing.binding.BindingBaseClassesOSGiTest)  Time elapsed: 0.022 sec
assert configuration is rolled-back on error(org.eclipse.smarthome.core.thing.binding.BindingBaseClassesOSGiTest)  Time elapsed: 0.021 sec
assert properties can be updated from ThingHandler(org.eclipse.smarthome.core.thing.binding.BindingBaseClassesOSGiTest)  Time elapsed: 0.024 sec
assert BaseThingHandler handles bridge status updates correctly(org.eclipse.smarthome.core.thing.binding.BindingBaseClassesOSGiTest)  Time elapsed: 0.083 sec
assert BaseThingHandlerFactory registers handler and BaseThingHandlers getBridge works(org.eclipse.smarthome.core.thing.binding.BindingBaseClassesOSGiTest)  Time elapsed: 0.021 sec
assert thing can be updated from ThingHandler(org.eclipse.smarthome.core.thing.binding.BindingBaseClassesOSGiTest)  Time elapsed: 0.015 sec
assert configuration parameters are validated(org.eclipse.smarthome.core.thing.binding.BindingBaseClassesOSGiTest)  Time elapsed: 0.012 sec
assert config status is propagated(org.eclipse.smarthome.core.thing.binding.BindingBaseClassesOSGiTest)  Time elapsed: 0.134 sec
assert configuration will be updated by default implementation(org.eclipse.smarthome.core.thing.binding.BindingBaseClassesOSGiTest)  Time elapsed: 0.021 sec
assert BaseThingHandler notifies ThingManager about configuration updates(org.eclipse.smarthome.core.thing.binding.BindingBaseClassesOSGiTest)  Time elapsed: 0.014 sec

Results :

Failed tests: 
  ThingManagerOSGiJavaTest.testChildHandlerInitialized_replacedUnitializedThing:337 expected:<0> but was:<1>

Tests run: 257, Failures: 1, Errors: 0, Skipped: 0
```

I'm lost, since my local maven build is working fine. Rebasing to latest master didn't help either.

Full log: [https://api.travis-ci.org/v3/job/354958442/log.txt](https://api.travis-ci.org/v3/job/354958442/log.txt)

PR #5255, opened 2018-03-19
fix ThingManagerOSGiJavaTest

sjka:
...as in testChildHandlerInitialized_replacedUnitializedThing sometimes the
test execution was faster than the framework realizing the handler failed.


maggu2810:
Thank you

J-N-K:
Unfortunately this is not fixing #5248. A new build of #5150 still fails with the same error as before.

PR #5265, opened 2018-03-20
Refactor service trackers

lpapazow:
As discussed [here](https://github.com/openhab/static-code-analysis/issues/265), I added these abstract classes as suppressions for the `DeclarativeServicesDependencyInjectionCheck`. 

This check was also triggered by the `RESTActivator` class. Instead of having a `localeProviderTracker` in the `RESTActivator` and have a `LocaleUtil` class that uses the RESTActivator's `getLocale`, I changed `LocaleUtil` to use DS and removed the `RESTActivator`.

If you think this change is unnecessary or potentially dangerous, I could revert it and just add the RESTActivator class as a suppression.


maggu2810:
As you already refactored it, why not doing it a little bit more the OSGi way.
* drop immediate activation of `LocaleService`
* make the `Locale getLocale(String acceptLanguageHttpHeader)` method non static
* the components / REST resources could require a mandatory reference for `LocaleService` (as the locale service itself is using a dynamic reference to the provider it can be activated all the time)

No need for an activator or access a static method of a service class.

WDYT?

lpapazow:
@maggu2810, I agree completely, having a mandatory reference and non staitic methods in the LocaleService seems like a way better idea.

@htreu, after this [commit](https://github.com/eclipse/smarthome/commit/b3933fa4ca29e6624db8f8b98f35552afbf3a3bb) the `getContext` method is no longer used and the only place the `getLocale` method is used is the LocaleUtil class. Could you please clarify if this is what you mean by API breaking and do you think refactoring the code in the way maggu proposed would be ok?

htreu:
@lpapazow yes you are right. I just realised that RESTActivator is internal. And I second @maggu2810's idea to drop static stiff completely and make use of the new component throughout the REST components.

lpapazow:
I added a mandatory reference for LocaleService in the REST resources. Some of them didn't use the `@Component` annotation but used the xml files in `OSGI-INF` folder instead. I refactored all of the classes that now require LocaleService to use annotations , and then I refactored all the other classes in those bundles to use annotations, deleted all the redundant .xml files and added .gitignore files in a seperate commit.

bundles/automation/org.eclipse.smarthome.automation.rest/META-INF/MANIFEST.MF
```diff
@@ -17,5 +17,6 @@ Import-Package:
  org.eclipse.smarthome.config.core,
  org.eclipse.smarthome.core.common.registry,
  org.eclipse.smarthome.io.rest,
+ org.osgi.service.component.annotations,
```
maggu2810: IIRC there has been a commit some time ago that removes all that imports.
Annotations does not need to be present at runtime (by the JVM) and the one here are known to be used at build time only.

bundles/automation/org.eclipse.smarthome.automation.rest/src/main/java/org/eclipse/smarthome/automation/rest/internal/ModuleTypeResource.java
```diff
@@ -73,6 +80,15 @@ protected void unsetModuleTypeRegistry(ModuleTypeRegistry moduleTypeRegistry) {
         this.moduleTypeRegistry = null;
     }
 
+    @Reference(cardinality=ReferenceCardinality.MANDATORY, policy=ReferencePolicy.STATIC)
```
maggu2810: `cardinality=ReferenceCardinality.MANDATORY, policy=ReferencePolicy.STATIC` is default, so no need to set it explicit.
Can you remove it here and on other places

bundles/io/org.eclipse.smarthome.io.rest.core/src/main/java/org/eclipse/smarthome/io/rest/core/internal/item/ItemResource.java
```diff
@@ -176,6 +178,15 @@ protected void unsetDTOMapper(DTOMapper dtoMapper) {
         this.dtoMapper = dtoMapper;
     }
 
+    @Reference(cardinality=ReferenceCardinality.MANDATORY, policy=ReferencePolicy.STATIC)
+    protected void setLocaleService(LocaleService localeService) {
+        this.localeService = localeService;
+    }
+    
+    protected void unsetLocaleService(LocaleService localeService) {
+        this.localeService = null;
```
htreu: with this you should have an error in your IDE: `Null type mismatch: required '@NonNull LocaleService' but the provided value is null`

bundles/io/org.eclipse.smarthome.io.rest.core/src/main/java/org/eclipse/smarthome/io/rest/core/internal/item/ItemResource.java
```diff
@@ -130,6 +131,7 @@
     @NonNullByDefault({})
     private DTOMapper dtoMapper;
     private final Set<ItemFactory> itemFactories = new HashSet<>();
+    private LocaleService localeService;
```
htreu: Here you should have an error in your IDE: `The @NonNull field localeService may not have been initialized`
You can circumvent this by adding the `@NonNullByDefault({})` annotation here. This will basically exclude the field from the null annotation checks.

bundles/io/org.eclipse.smarthome.io.rest.core/src/main/java/org/eclipse/smarthome/io/rest/core/internal/item/ItemResource.java
```diff
@@ -65,7 +65,8 @@
 import org.eclipse.smarthome.core.types.TypeParser;
 import org.eclipse.smarthome.io.rest.DTOMapper;
 import org.eclipse.smarthome.io.rest.JSONResponse;
-import org.eclipse.smarthome.io.rest.LocaleUtil;
+import org.eclipse.smarthome.io.rest.LocaleServiceImpl;
```
htreu: this import is reported to be unsed, please remove.

bundles/io/org.eclipse.smarthome.io.rest/src/main/java/org/eclipse/smarthome/io/rest/LocaleService.java
```diff
@@ -0,0 +1,9 @@
+package org.eclipse.smarthome.io.rest;
```
htreu: license header missing

bundles/io/org.eclipse.smarthome.io.rest/src/main/java/org/eclipse/smarthome/io/rest/LocaleService.java
```diff
@@ -0,0 +1,9 @@
...
+
+import java.util.Locale;
+
+public interface LocaleService {
```
htreu: JavaDoc & author tag missing

bundles/io/org.eclipse.smarthome.io.rest/src/main/java/org/eclipse/smarthome/io/rest/LocaleService.java
```diff
@@ -0,0 +1,9 @@
...
+
+    public Locale getLocale(String acceptLanguageHttpHeader);
```
htreu: methods on public interfaces are public by default, please remove the keyword.

htreu:
Lets wait for rebase and Travis, I guess we are good to go then.

sjka:
@lpapazow friendly ping

lpapazow:
Thanks for the reminder, I've rebased the branch.

maggu2810:
@htreu Have your requested changes been applied.

maggu2810:
> All injected services to a REST resource have to be referenced optional/dynamic to allow the resource to resolve anytime.

Why?

sjka:
To avoid having jersey rebuilds the entire universe again and again - see #2320

htreu:
>> All injected services to a REST resource have to be referenced optional/dynamic to allow the resource to resolve anytime.

> Why?

The REST framework jersey is not able to restart single resources in case a dependency gets registered/unregistered. So all resources get restarted every time dependent services register/unregister. The relevant change was introduced with https://github.com/eclipse/smarthome/pull/2320.

maggu2810:
Thank you both for point me the respective PR.
Now I remember again. :wink:

lpapazow:
I made the requested changes, sorry for taking so long.

bundles/io/org.eclipse.smarthome.io.rest/src/main/java/org/eclipse/smarthome/io/rest/LocaleServiceImpl.java
```diff
@@ -0,0 +1,93 @@
...
+ * @author Markus Rathgeb - Use locale provider
+ * @author Martin Herbst - Support of different language definition variants
+ * @authro Lyubomir Papazov - Add component annotation, rename the class to LocaleService and add method tryGetLocale
+ */
+@Component
+public class LocaleServiceImpl implements LocaleService {
+
+    private static final Logger logger = LoggerFactory.getLogger(LocaleServiceImpl.class);
+
+    private static LocaleProvider localeProvider;
```
maggu2810: * remove `static`
* add `volatile`

bundles/io/org.eclipse.smarthome.io.rest/src/main/java/org/eclipse/smarthome/io/rest/LocaleServiceImpl.java
```diff
@@ -0,0 +1,93 @@
...
+ *
+ * @author Dennis Nobel - Initial contribution
+ * @author Markus Rathgeb - Use locale provider
+ * @author Martin Herbst - Support of different language definition variants
+ * @authro Lyubomir Papazov - Add component annotation, rename the class to LocaleService and add method tryGetLocale
+ */
+@Component
+public class LocaleServiceImpl implements LocaleService {
+
+    private static final Logger logger = LoggerFactory.getLogger(LocaleServiceImpl.class);
```
maggu2810: * remove `static`

bundles/automation/org.eclipse.smarthome.automation.rest/src/main/java/org/eclipse/smarthome/automation/rest/internal/TemplateResource.java
```diff
@@ -48,13 +52,16 @@
  */
 @Path("templates")
 @Api("templates")
+@Component(service = { RESTResource.class, TemplateResource.class })
```
maggu2810: Why do you need to export the the implementation class itself as service (has not been done before)?

bundles/automation/org.eclipse.smarthome.automation.rest/src/main/java/org/eclipse/smarthome/automation/rest/internal/ModuleTypeResource.java
```diff
@@ -58,13 +62,16 @@
  */
 @Path("module-types")
 @Api("module-types")
+@Component(service = { RESTResource.class, ModuleTypeResource.class })
```
maggu2810: Why do you declare and so export the class itself as provided service?

bundles/automation/org.eclipse.smarthome.automation.rest/src/main/java/org/eclipse/smarthome/automation/rest/internal/RuleResource.java
```diff
@@ -76,6 +80,7 @@
  */
 @Path("rules")
 @Api("rules")
+@Component(service = { RESTResource.class, RuleResource.class })
```
maggu2810: Why do you declare and so export the class itself as provided service?

bundles/io/org.eclipse.smarthome.io.rest.voice/src/main/java/org/eclipse/smarthome/io/rest/voice/internal/VoiceResource.java
```diff
@@ -55,7 +55,7 @@
  *
  * @author Kai Kreuzer - Initial contribution and API
  */
-@Component
+@Component(service = { RESTResource.class, VoiceResource.class })
```
maggu2810: Why do you declare and so export the class itself as provided service?

bundles/io/org.eclipse.smarthome.io.rest/src/main/java/org/eclipse/smarthome/io/rest/LocaleServiceImpl.java
htreu: Do you claim `volatile` here because `localeProvider` is a dynamic reference and can be changed any time? Other threads holding a LocaleService need to check with the main copy then, correct?

maggu2810: Correct.
For dynamic references volatile or atomic references should be used.
If you want to spend your time in reading some discussion about: https://github.com/eclipse/smarthome/pull/3617
:wink:

kaikreuzer:
@lpapazow Any update on this?

bundles/automation/org.eclipse.smarthome.automation.rest/src/main/java/org/eclipse/smarthome/automation/rest/internal/TemplateResource.java
lpapazow: I changed it because of this [comment](https://github.com/eclipse/smarthome/pull/5265#discussion_r188225858)

bundles/automation/org.eclipse.smarthome.automation.rest/src/main/java/org/eclipse/smarthome/automation/rest/internal/ModuleTypeResource.java
lpapazow: Changed because of [this](https://github.com/eclipse/smarthome/pull/5265#discussion_r188215435)

bundles/automation/org.eclipse.smarthome.automation.rest/src/main/java/org/eclipse/smarthome/automation/rest/internal/TemplateResource.java
htreu: Sorry for the confusion. @maggu2810 is completely right, the REST services should be declared to provide only the RESTResource interface. Please update accordingly @lpapazow. Thanks.

kaikreuzer: Sorry to comment so late, but I would actually prefer if you could remove all changes to the automation engine from this PR, because it will cause huge conflicts for #4468, which is already addressing these components as well.

lpapazow: If I want to keep the LocaleService logic, I should injected the locale into the automation RESTResources. Would it be okay to add refererences in the .xml files in the OSGI-INF folder?

kaikreuzer: Oh, I didn't look closely enough to notice that it all is linked to the LocaleService...
No, changing the XML won't be any better wrt conflicts. I guess it is then ok to leave this PR as it is and rather rebase #4468 later on and sync the changes again. @adimova do you agree?

bundles/io/org.eclipse.smarthome.io.rest.sitemap/META-INF/MANIFEST.MF
```diff
@@ -32,6 +32,8 @@ Import-Package:
  org.eclipse.smarthome.ui.items,
  org.glassfish.jersey.media.sse,
  org.glassfish.jersey.server,
+ org.osgi.service.component.annotations,
```
htreu: This dependency is not needed, please remove.

bundles/io/org.eclipse.smarthome.io.rest.sitemap/src/main/java/org/eclipse/smarthome/io/rest/sitemap/internal/SitemapResource.java
```diff
@@ -108,7 +108,7 @@
  * @author Chris Jackson
  * @author Yordan Zhelev - Added Swagger annotations
  */
-@Component(service = { SitemapResource.class, RESTResource.class })
+@Component
```
htreu: This will expose the service with all attached interfaces. Please use `@Component(service = RESTResource.class )` here.

bundles/ui/org.eclipse.smarthome.ui.icon/src/main/java/org/eclipse/smarthome/ui/icon/internal/IconServlet.java
```diff
@@ -41,6 +45,7 @@
  *
  * @author Kai Kreuzer - Initial contribution
  */
+@Component
```
htreu: The original service xml stated also `configuration-policy="optional"` and `modified="modified"`.

bundles/model/org.eclipse.smarthome.model.script/OSGI-INF/voiceaction.xml
```diff
@@ -1,22 +0,0 @@
-<?xml version="1.0" encoding="UTF-8"?>
```
htreu: This file is removed, but I'm missing the annotation at `VoiceActionService`.

htreu:
Only Travis complains. Unfortunately I don't understand why.

kaikreuzer:
I have restarted Travis, now it is green.
@maggu2810 As you still have a "request changes" review open, I leave merging to you.

bundles/ui/org.eclipse.smarthome.ui.icon/src/main/java/org/eclipse/smarthome/ui/icon/internal/IconServlet.java
```diff
@@ -41,6 +47,7 @@
  *
  * @author Kai Kreuzer - Initial contribution
  */
+@Component(configurationPolicy = ConfigurationPolicy.OPTIONAL)
```
maggu2810: > If not specified, the OPTIONAL configuration policy is used. 

On no other places we set the configuration policy to optional explicit, why here?

bundles/io/org.eclipse.smarthome.io.rest.sitemap/META-INF/MANIFEST.MF
```diff
@@ -35,3 +35,4 @@ Import-Package:
  org.osgi.framework,
  org.slf4j
 Service-Component: OSGI-INF/*.xml
+Bundle-ActivationPolicy: lazy
```
maggu2810: Not needed

bundles/ui/org.eclipse.smarthome.ui.icon/src/main/java/org/eclipse/smarthome/ui/icon/internal/IconServlet.java
htreu: It was defined like that in the original service XML, see https://github.com/eclipse/smarthome/pull/5265#discussion_r197353799.

In case its optional I suggest addressing this in another PR for better traceability.

maggu2810: @htreu What do you want to trace?
We migrate to the component annotations.
The annotations have default values, that does not need to be set explicitly.
We have requested to drop implicit values all the time (in this PR and all the others) if we found their usage in the review. Don't you remember on the comments about dropping the mandatory (1..1) reference and the static policy because it is the default value of the annotation? It does not matter, if this has been part of the XML before (e.g. https://github.com/eclipse/smarthome/pull/5265/files#diff-46cb4a3fb85d0b0057a3f9264fc8da74L18).

htreu: Alright, thanks for the implicit explanation that `configurationPolicy = ConfigurationPolicy.OPTIONAL` is the default :-)
@lpapazow please drop it as requested.

htreu:
Thanks @lpapazow for the update.
@maggu2810 this one should be good to go.

sjka:
I flagged this as APIBreaking because of the vanished `LocaleUtil`.

PR #5268, opened 2018-03-20
Fix QuantityType parsing for exponential numbers with units

htreu:
Fixes #5247.

Signed-off-by: Henning Treu <henning.treu@telekom.de>

PR #5276, opened 2018-03-20
QuantityType: parse exponential values more stable

htreu:
This addresses the remaining issues from #5268.


maggu2810:
Great.
I have to admit that I am not such familiar with that lookahead and lookbehind expressions. So, if Travis CI succeeds let's give it ago.

PR #5277, opened 2018-03-20
removed todo

### LINKED ISSUE #5274: Static Code Analysis Tool reports unaddressed TODO comments in SonosAudioSink
I'm currently fixing issues in Smarthome, found by [the Static Code Analysis Tool](https://github.com/openhab/static-code-analysis/blob/master/README.md). The included check from [Checkstyle](https://github.com/checkstyle/checkstyle/blob/master/README.md) - *TodoCommentCheck*, which checks for left `// todo comments` in javadoc, reports [SonosAudioSink](https://github.com/eclipse/smarthome/blob/4204ce06bb28c28e5f711e720f87ef83beff2e27/extensions/binding/org.eclipse.smarthome.binding.sonos/src/main/java/org/eclipse/smarthome/binding/sonos/internal/SonosAudioSink.java#L119) and I can not find an open issue with discussion about it.

Comment had been introduced with [this](https://github.com/eclipse/smarthome/commit/d87b1532643a1fbcdebe338e934db8b8c955507c) commit.

@kaikreuzer Аs an author, do you think that this is still an issue or we can remove the TODO comment?

kaikreuzer:
As the comment suggests, handling an unknown stream is rather risky, so I would say that the current implementation is just fine.
Nonetheless, I would like to keep the comment in there as an explanation.

Fixes #5274


maggu2810:
Why there is a [Tidelift] check now?

[Tidelift]: https://tidelift.com/builds/649847

kaikreuzer:
I wondered the same myself...

PR #5280, opened 2018-03-20
prevent potential NPE on dispose

kaikreuzer:

### DIFF
--- a/extensions/binding/org.eclipse.smarthome.binding.bluetooth/src/main/java/org/eclipse/smarthome/binding/bluetooth/BeaconBluetoothHandler.java
+++ b/extensions/binding/org.eclipse.smarthome.binding.bluetooth/src/main/java/org/eclipse/smarthome/binding/bluetooth/BeaconBluetoothHandler.java
@@ -12,6 +12,8 @@
  */
 package org.eclipse.smarthome.binding.bluetooth;
 
+import java.util.concurrent.locks.ReentrantLock;
+
 import org.eclipse.jdt.annotation.NonNull;
 import org.eclipse.smarthome.binding.bluetooth.notification.BluetoothConnectionStatusNotification;
 import org.eclipse.smarthome.binding.bluetooth.notification.BluetoothScanNotification;
@@ -39,6 +41,7 @@ public class BeaconBluetoothHandler extends BaseThingHandler implements Bluetoot
     protected BluetoothAdapter adapter;
     protected BluetoothAddress address;
     protected BluetoothDevice device;
+    protected ReentrantLock deviceLock;
 
     public BeaconBluetoothHandler(@NonNull Thing thing) {
         super(thing);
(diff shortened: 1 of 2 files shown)


extensions/binding/org.eclipse.smarthome.binding.bluetooth/src/main/java/org/eclipse/smarthome/binding/bluetooth/ConnectedBluetoothHandler.java
```diff
@@ -78,10 +78,14 @@ public void initialize() {
...
             connectionJob.cancel(true);
+            connectionJob = null;
         }
-        if (device != null) {
-            scheduler.submit(() -> device.disconnect());
-        }
+        scheduler.submit(() -> {
+            if (device != null) {
+                device.disconnect();
+                device = null;
```
maggu2810: Do we care about threading?
`device` is set in the thread that calls `initialize` (in the parent class), `device` is unset (set to `null`) in a thread used by `scheduler`. So, should we use volatile, some synchronization, ...?

kaikreuzer: You are right, I'll rework this...

kaikreuzer: Using a reentrant lock now to make sure that different threads are correctly synchronized when accessing the device.

extensions/binding/org.eclipse.smarthome.binding.bluetooth/src/main/java/org/eclipse/smarthome/binding/bluetooth/BeaconBluetoothHandler.java
```diff
@@ -67,18 +70,23 @@ public void initialize() {
         }
 
         adapter = (BluetoothAdapter) bridgeHandler;
+
+        deviceLock.lock();
```
sjka: locks should generally be guarded by `try {} finally {}`-blocks - especially when some (foreign) non-trivial code is called while the lock is in place.

kaikreuzer: thx. updated.

maggu2810:
One question:
As we don't hold the lock over functions etc. wouldn't an sync. object (e.g. `deviceSync`) and a `synchronized (deviceSync) { ... }` simpler / safer (you cannot forget an unlock) then `try { ...lock(); ... } finally { ...unlock(); }`?

kaikreuzer:
I had chosen the ReentrantLock in order to allow sub-classes to potentially do more complex things, if needed. I agree that with the current implementation as simply object would suffice for the synchronisation.

maggu2810:
As #5303 has been merged now, should we (you) rebase your branch, so the build doesn't fail?

kaikreuzer:
Which build fails...? All green for me :-)

maggu2810:
I re-triggered the failed build immediately before I post the comment.

PR #5284, opened 2018-03-21
Add some more imperial units

ivivanov-bg:


htreu:
Hi @ivivanov-bg, thanks for extending the units!
Could I ask you to also provide some small test cases to prove the conversion is correct? There are already some tests in `SmartHomeUnitsTest` but you may also create a new `ImperialUnitsTest`.

ivivanov-bg:
Done

bundles/core/org.eclipse.smarthome.core/src/main/java/org/eclipse/smarthome/core/library/unit/ImperialUnits.java
```diff
@@ -69,21 +72,70 @@ public static SystemOfUnits getInstance() {
...
+    public static final Unit<Length> FOOT = addUnit(
+            new TransformedUnit<>("ft", ImperialUnits.INCH, new MultiplyConverter(12.0)));
+    
+    public static final Unit<Length> YARD = addUnit(
+            new TransformedUnit<>("yd", ImperialUnits.FOOT, new MultiplyConverter(3.0)));
+    
+    public static final Unit<Length> CHAIN = addUnit(
+            new TransformedUnit<>("ch", ImperialUnits.YARD, new MultiplyConverter(22.0)));
+    
+    public static final Unit<Length> FURLON = addUnit(
```
9037568: "FURLONG"

bundles/core/org.eclipse.smarthome.core/src/main/java/org/eclipse/smarthome/core/library/unit/ImperialUnits.java
```diff
@@ -69,21 +72,70 @@ public static SystemOfUnits getInstance() {
...
-        SimpleUnitFormat.getInstance().label(INCH, "in");
-        SimpleUnitFormat.getInstance().label(MILE, "mi");
+        SimpleUnitFormat.getInstance().label(INCH_OF_MERCURY, INCH_OF_MERCURY.getSymbol());
+        SimpleUnitFormat.getInstance().label(FAHRENHEIT, FAHRENHEIT.getSymbol());
+        SimpleUnitFormat.getInstance().label(MILES_PER_HOUR, MILES_PER_HOUR.getSymbol());
+        SimpleUnitFormat.getInstance().label(INCH, INCH.getSymbol());
+        SimpleUnitFormat.getInstance().label(FOOT, FOOT.getSymbol());
+        SimpleUnitFormat.getInstance().label(YARD, YARD.getSymbol());
+        SimpleUnitFormat.getInstance().label(CHAIN, CHAIN.getSymbol());
+        SimpleUnitFormat.getInstance().label(FURLON, FURLON.getSymbol());
```
9037568: "FURLONG"

htreu:
The introduction of the imperial gallon with this PR brings up the following issue:
ESH right now uses the Imperial system for US (and Liberia) since all units up to now are identical to the US customary system.
With the imperial gallon introduced here this would lead to confusing conversions when having locale `US`: The US Gallon is different from the imperial gallon [as described here](https://en.wikipedia.org/wiki/Comparison_of_the_imperial_and_US_customary_measurement_systems).
@ivivanov-bg May I ask you to introduce a new measurement system `USCustomary` with corresponding units? I think the new Units may extend the ImperialUnits and define US_GALLON, etc.
These new measurement system should then replace the imperial system for the locale `US` (and maybe also Liberia).

ivivanov-bg:
Or - remove the gallon from the PR (so that it doens't cause a confusion) and leave the `USCustomary `  measurement system for a further PR. Because it will require more time for implementation and I don't have it currently

htreu:
Yes, I'm fine with that. I will create a new issue for US customary measurement system.

htreu:
Issue for the US customary system: #5289.

PR #5287, opened 2018-03-21
add nullness annotations to ConfigI18nLocalizationService

maggu2810:
The locale is optional and currently already called with a null.
ConfigDescriptionProviderImpl provides a nullable locale.
If we follow the function calls a getText method of the
TranslationProvider is called and this one already states (currently in
the JavaDoc only) that the locale is nullable.

Will come up with the changes for the TranslationProvider in another PR.


htreu:
Only waiting for Travis...

Issue #5291, opened 2018-03-21
core.config.ConfigurationService IllegalArgumentException with Magic binding


davidgraeff:
When setting the value of "Multiselect Integer - limit" of the Magic binding within PaperUI an exception is thrown.
![bildschirmfoto vom 2018-03-21 15-28-53](https://user-images.githubusercontent.com/66436/37715703-fa41a45a-2d1c-11e8-8b35-4cb7231527ee.png)

I get this stacktrace:
```
16:58:51.959 DEBUG o.e.s.i.r.JSONResponse$ExceptionMapper[:203] - exception during REST Handling
java.lang.IllegalArgumentException: java.math.BigDecimal in java.util.Vector
	at org.eclipse.equinox.internal.cm.ConfigurationDictionary.validateValue(ConfigurationDictionary.java:58)
	at org.eclipse.equinox.internal.cm.ConfigurationDictionary.put(ConfigurationDictionary.java:109)
	at org.eclipse.equinox.internal.cm.ConfigurationImpl.updateDictionary(ConfigurationImpl.java:338)
	at org.eclipse.equinox.internal.cm.ConfigurationImpl.update(ConfigurationImpl.java:310)
	at org.eclipse.smarthome.io.rest.core.config.ConfigurationService.update(ConfigurationService.java:132)
	at org.eclipse.smarthome.io.rest.core.config.ConfigurationService.update(ConfigurationService.java:71)
	at org.eclipse.smarthome.io.rest.core.internal.service.ConfigurableServiceResource.updateConfiguration(ConfigurableServiceResource.java:193)
	at sun.reflect.NativeMethodAccessorImpl.invoke0(Native Method)
```

kaikreuzer:
@davidgraeff Has this been fixed meanwhile or is it still open? (I cannot fully interpret your commit comment on #5292...)

davidgraeff:
I just mentioned this bug. If nobody worked on the magic binding, it may still exist. I have no Eclipse at hand right now to verify.

PR #5292, opened 2018-03-21
Configuration: Move static methods to ConfigUtil. Add documentation.

davidgraeff:
* getAllFields, as methods moved from Configuration to internal ConfigMapper
* Add String->BigDecimal conversion in ConfigMapper.as
* Add configuration holder object to MagicService of the magic binding to see if the mapping works for all kind of configuration values.
  (Except for the case of bug #5291)

### DIFF
--- a/bundles/config/org.eclipse.smarthome.config.core/src/main/java/org/eclipse/smarthome/config/core/ConfigUtil.java
+++ b/bundles/config/org.eclipse.smarthome.config.core/src/main/java/org/eclipse/smarthome/config/core/ConfigUtil.java
@@ -21,26 +21,24 @@ import java.util.List;
 import java.util.Map;
 import java.util.Map.Entry;
 
+import org.eclipse.jdt.annotation.Nullable;
 import org.eclipse.smarthome.config.core.internal.normalization.Normalizer;
 import org.eclipse.smarthome.config.core.internal.normalization.NormalizerFactory;
 import org.eclipse.smarthome.config.core.validation.ConfigDescriptionValidator;
 
 /**
- * This class provides some useful static methods for handling configurations
+ * The configuration admin service provides us with a map of key->values. Values can be any
+ * primitive type like String, int, double as well as the object variants of the number types
+ * like Integer, Double etc. You find normalization utility methods in here to convert all
+ * number Types to BigDecimal and all Collections<type> to List<type>. This works on a best-effort
+ * strategy (e.g. "3" will always end up being a BigDecimal, never a String),
+ * except if a {@link ConfigDescriptionParameter} is given. In the latter case,
+ * a conversion according to the type given in the description is performed.
  *
  * @author Kai Kreuzer - Initial API and implementation
  * @author Thomas Höfer - Minor changes for type normalization based on config description
  */
 public class ConfigUtil {
-    /**
-     * We do not want to handle or try to normalize OSGi provided configuration parameters
-     *
-     * @param name The configuration parameter name
-     */
-    private static boolean isOSGiConfigParameter(String name) {
-        return name.equals("objectClass") || name.equals("component.name") || name.equals("component.id");
-    }
-
     /**
      * Normalizes the types to the ones allowed for configurations.
      *
@@ -53,20 +51,10 @@ public class ConfigUtil {
             String name = parameter.getKey();
             Object value = parameter.getValue();
             if (!isOSGiConfigParameter(name)) {
-                convertedConfiguration.put(name, normalizeType(value));
+                convertedConfiguration.put(name, normalizeType(value, null));
             }
         }
         return convertedConfiguration;
-    }
-
-    /**
-     * Normalizes the type of the parameter to the one allowed for configurations.
-     *
-     * @param value the value to return as normalized type
-     * @return corresponding value as a valid type
-     */
-    public static Object normalizeType(Object value) {
-        return normalizeType(value, null);
     }
 
     /**
@@ -77,7 +65,7 @@ public class ConfigUtil {
      * @return corresponding value as a valid type
      * @throws IllegalArgumentException if a invalid type has been given
      */
-    public static Object normalizeType(Object value, ConfigDescriptionParameter configDescriptionParameter) {
+    public static Object normalizeType(Object value, @Nullable ConfigDescriptionParameter configDescriptionParameter) {
         if (configDescriptionParameter != null) {
             Normalizer normalizer = NormalizerFactory.getNormalizer(configDescriptionParameter);
             return normalizer.normalize(value);
@@ -91,30 +79,6 @@ public class ConfigUtil {
         }
         throw new IllegalArgumentException(
                 "Invalid type '{" + value.getClass().getCanonicalName() + "}' of configuration value!");
-    }
-
-    /**
-     * Normalizes a collection.
-     *
-     * @param collection the collection that entries should be normalized
-     * @return a collection that contains the normalized entries
-     * @throws IllegalArgumentException if the type of the normalized values differ or an invalid type has been given
-     */
-    private static Collection<Object> normalizeCollection(Collection<?> collection) throws IllegalArgumentException {
-        if (collection.size() == 0) {
-            return Collections.emptyList();
-        } else {
-            final List<Object> lst = new ArrayList<>(collection.size());
-            for (final Object it : collection) {
-                final Object normalized = normalizeType(it);
-                lst.add(normalized);
-                if (normalized.getClass() != lst.get(0).getClass()) {
-                    throw new IllegalArgumentException(
-                            "Invalid configuration property. Heterogeneous collection value!");
-                }
-            }
-            return lst;
-        }
     }
 
     /**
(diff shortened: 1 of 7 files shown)


bundles/test/org.eclipse.smarthome.magic/src/main/java/org/eclipse/smarthome/magic/binding/internal/MagicServiceImpl.java
```diff
@@ -49,4 +59,25 @@
...
+    @Activate
+    public void activate(Map<String, Object> properties) {
+        modified(properties);
+    }
+
+    @Modified
+    public void modified(Map<String, Object> properties) {
+        // Print changed values
+        MagicServiceConfig config = new Configuration(properties).as(MagicServiceConfig.class);
+        List<Field> l = ConfigUtil.getAllFields(MagicServiceConfig.class);
```
htreu: Maybe instead of exposing the reflection based method `getAllFields` you could give a `toString` to MagicServiceConfig.

davidgraeff: True. Regarding the annotations: I actually introducted `UseKey` because the configuration key is unfortunately called "boolean".

I'm honestly a big fan of this property-map to class object mapping, partly because I do not find that powerful reflection API in many other languages. However I'd like to see this as a default way of handling service configuration and it should work in every case.

So we either need this `UseKey` annotation or we need to forbid certain configuration key names (Java language keywords).

htreu:
For the time being I suggest to have the refactoring (and magic extension) only.

davidgraeff:
Agreed. I'll open another issue for discussion. I'll rename the magic bindings configuration name from boolean to "bool".

davidgraeff:
@htreu Done

davidgraeff:
And I have added an issue for the annotations #5300.

bundles/config/org.eclipse.smarthome.config.core/src/main/java/org/eclipse/smarthome/config/core/ConfigUtil.java
```diff
@@ -161,4 +171,130 @@ public static Object normalizeType(Object value, ConfigDescriptionParameter conf
         return convertedConfiguration;
     }
 
+    /**
+     * Return fields of the given class as well as all super classes.
+     *
+     * @param clazz The class
+     * @return A list of Field objects
+     */
+    public static List<Field> getAllFields(Class<?> clazz) {
```
htreu: please make this `private` again.

htreu: And it would be nice to have the public API on top and the supporting private methods at the bottom of the class.

bundles/test/org.eclipse.smarthome.magic/src/main/java/org/eclipse/smarthome/magic/binding/internal/MagicServiceImpl.java
```diff
@@ -49,4 +56,14 @@
...
     }
 
+    @Activate
+    public void activate(Map<String, Object> properties) {
+        modified(properties);
+    }
+
+    @Modified
+    public void modified(Map<String, Object> properties) {
+        MagicServiceConfig config = new Configuration(properties).as(MagicServiceConfig.class);
```
htreu: Since you introduced the new ConfigUtil#as method you can directly use it here.

davidgraeff: That would be the only place of all services where it would be different. I'd like to wait for the discussion and outcome of #5300, if that is ok for you.

bundles/config/org.eclipse.smarthome.config.core/src/main/java/org/eclipse/smarthome/config/core/ConfigUtil.java
davidgraeff: Yup, I usually would do that, but in this case all fields and methods of the ConfigUtil class are widely mixed. If I sort them now, it would be a hard job reviewing. Should be done in another PR, if that's ok.

davidgraeff:
@htreu done

bundles/test/org.eclipse.smarthome.magic/src/main/java/org/eclipse/smarthome/magic/binding/internal/MagicServiceImpl.java
htreu: agreed.

bundles/config/org.eclipse.smarthome.config.core/src/main/java/org/eclipse/smarthome/config/core/ConfigUtil.java
htreu: agreed, thats a pity.
But then please order at least the refactored code, so that public comes first, private comes last. And please make #getAllFields `private`.

davidgraeff:
I have sorted public/private now. But did two more things as well:
* Added a nullable to this signature: `normalizeType(Object value, @Nullable ConfigDescriptionParameter configDescriptionParameter)`.
* Deprecated `normalizeType(Object value)`

htreu:
> Deprecated normalizeType(Object value)

Why? Everything just looked so good ;-)

davidgraeff:
Haha yes I know. Trouble maker /me :D 
Let me explain my intention. Think of what `normalizeTypes` actually does: It converts all number types to BigDecimal and all collection types to Lists. Usually it would consider the configuration specification for the correct conversion. If this is not provided, it does this only on a best-effort approach.

The consumer of the API should clearly realize that he is using an API that only works best-effort and is used in a way that it originally was not intended to be used.  The one argument only method hides this fact and the best way to realize the API intention is by the mandatory second argument `ConfigDescriptionParameter`.

htreu:
Yes, I can follow your intention. Please revert the deprecation here so we can merge the refactoring. Lets discuss this new aspect on a separate PR where others can join the discussion. I don't want to blow this one up unnecessarily. Thanks :-)

bundles/config/org.eclipse.smarthome.config.core/src/main/java/org/eclipse/smarthome/config/core/ConfigUtil.java
```diff
@@ -161,4 +128,175 @@ public static Object normalizeType(Object value, ConfigDescriptionParameter conf
...
+     * </pre>
+     *
+     *
+     * @param properties The configuration map.
+     * @param configurationClass The configuration holder class. An instance of this will be created so make sure that
+     *            a default constructor is available.
+     * @return The configuration holder object. All fields that matched a configuration option are set. If a required
+     *         field is not set, null is returned.
+     */
+    public static <T> @Nullable T as(Map<String, Object> properties, Class<T> configurationClass) {
```
sjka: Do you have any particular reasons why this method must/should be exposed as API? I'd rather have expected to see in in a class within an internal package. In that way the `Configuration` class can use it and everybody else can do a `new Configuration(Map).as(Class)` if needed. What did I miss?

davidgraeff: @sjka I tried to explain this in the PR message :) 
`new Configuration(Map).as(Class)` does more than we need for service configuration objects. It "normalizes" values, which basically means all numbers -> BigDecimal, Collections -> Lists.
But the ConfigUtils.as method is way smarter, because it converts the types based on the field type. So yes, it makes sense to have it public API and use ConfigUtils.as within services.

sjka: > I tried to explain this in the PR message

If I didn't get that from your explanation, how should a user of the API figure that out? That is pretty surprising. 

Another thing worth mentioning is that static utility methods only got us into trouble, both on testing/mocking side as well as with respect to evolving the APIs. We have been working in the past on getting rid of several such static helper methods and that always was painful on some end. Therefore I'd like to object to adding yet another one.

> Maybe it should be marked as deprecated so that nobody starts to use it before this is discussed.

I don't know which "as" method you mean: 
* Configuration.as(...) must stay where it is, that cannot be moved anymore.
* ConfigUtil.as.(...) must not be introduced publicly if we already suspect it will have to move. Why does it have to be public then if nobody should start using it?

davidgraeff:
It's hard to mock static methods true. 

And yes documentation of this class is not very rich. The pr addresses some methos. 

Please tell me how to proceed here. The mqtt stuff is once again ready and API changes of this pr are a requirement for my other PRs. The "as" method does not need to be moved in this PR, but it leads to inefficient code if it is not addressed. (Double type conversion of config values)

htreu:
@davidgraeff, @sjka to get this one finally approved I suggest to provide the APIs just as they are available now. Please refactor the `public static as()` method into some internal package. This gives us room for discussion and the rest of MQTT can finally move on.

davidgraeff:
@htreu, @sjka Done

htreu:
@davidgraeff Travis fails due to missing author-tag & JavaDoc on the new ConfigMapper. Please add this asap. Thanks :-)

htreu:
Never mind, fixed it with #5320.

PR #5293, opened 2018-03-21
add nullness annotations for i18n interfaces and impl

maggu2810:
I added WIP, because there are some question / preconditions that must be fixed:

---

First one question about the `getText` method of the `TranslationProvider`, more precise about the `arguments` argument.
The JavaDoc states:
```JavaDoc
@param arguments the arguments to be injected into the translation (could be null)
```
What do we expect / allow for the value of `arguments`:
1. it can be null and if present its entries can be nullable, too (so a `@Nullable Object @Nullable []`)
1. it can be null but if present its entries must be non-null (so a `@NonNull Object @Nullable []`)
1. itself must be non-null but can contain nullable entries (so a `@Nullable Object @NonNull []`)
1. itself must be non-null and it entries must be non-null (so a `@NonNull Object @NonNull []`)

From the reading of the JavaDoc I would expect 1 or 2.

---

It seems there is a "bug" in SAT. It breaks on annotated optional arguments (openhab/static-code-analysis#276).

I added the exclusion that will be used by SAT in a new release.
Until SAT has released a new version and that one is used by ESH, I added a work around to disable that for the respective bundle.

---

The current implementation of `private void setLocation(final @Nullable String location)` keeps the old location if the `location` is set to `null`. That differs from the other `set` functions that unset the explicit value.
I changed that implementation now, as I have assumed this has not been done differently for this attribute by intention.
If it has been done by intention please comment or better, comment in every case.

### DIFF
--- a/bundles/core/org.eclipse.smarthome.core/pom.xml
+++ b/bundles/core/org.eclipse.smarthome.core/pom.xml
@@ -20,4 +20,22 @@
     <bundle.namespace>org.eclipse.smarthome.core</bundle.namespace>
   </properties>
 
+  <build>
+    <pluginManagement>
+      <plugins>
+        <plugin>
+          <groupId>org.openhab.tools.sat</groupId>
+          <artifactId>sat-plugin</artifactId>
+          <version>${sat.version}</version>
+          <executions>
+            <execution>
+              <id>sat-all</id>
+              <phase>none</phase>
+            </execution>
+          </executions>
+        </plugin>
+      </plugins>
+    </pluginManagement>
+  </build>
+
 </project>
--- a/bundles/core/org.eclipse.smarthome.core/src/main/java/org/eclipse/smarthome/core/i18n/LocaleProvider.java
+++ b/bundles/core/org.eclipse.smarthome.core/src/main/java/org/eclipse/smarthome/core/i18n/LocaleProvider.java
@@ -14,11 +14,14 @@ package org.eclipse.smarthome.core.i18n;
 
 import java.util.Locale;
 
+import org.eclipse.jdt.annotation.NonNullByDefault;
+
 /**
  * The interface describe a provider for a locale.
  *
  * @author Markus Rathgeb - Initial contribution and API
  */
+@NonNullByDefault
 public interface LocaleProvider {
 
     /**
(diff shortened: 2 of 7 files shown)


maggu2810:
@triller-telekom Can you have a look at?

triller-telekom:
Regarding your first question about the `TranslationProvider`:

https://github.com/eclipse/smarthome/blob/master/bundles/core/org.eclipse.smarthome.core.thing/src/main/java/org/eclipse/smarthome/core/thing/i18n/ThingStatusInfoI18nLocalizationService.java#L87

uses `Description.args` which is set to `null` here:

https://github.com/eclipse/smarthome/blob/master/bundles/core/org.eclipse.smarthome.core.thing/src/main/java/org/eclipse/smarthome/core/thing/i18n/ThingStatusInfoI18nLocalizationService.java#L117

So it is obviously allowed to pass `null`. Question is whether it is also allowed to pass a non-null array with `null` values.

A brief test with

```
String s = MessageFormat.format("Value-0: {0} Value-1 {1}", new Object[] { "val1", null });
System.out.println("s: " + s);
```

results in: 
```
s: Value-0: val1 Value-1 null
```
So I would vote for your option 1: `@Nullable Object @Nullable []`

---

Regarding the `setLocation` issue: This method was added in PR #4256 by @doandzhi and it is only called from the `modified` method. From what is implemented there I read that if the service configuration does not have a value for location set (and thus `null`), it will not be set for the service, so the service keeps whatever it has as a location.

I think this is not what was intended by that logic, i.e. it was forgotten to set the location of the service to `null` if it was deleted from the configuration. Am I right @doandzhi ?

maggu2810:
> So I would vote for your option 1: `@Nullable Object @Nullable []`

Done

> Am I right @doandzhi ?

You have gotten a :+1: from Erdoan, so this should be okay, too.

From my side the "WIP" is removed now.

The SAT workaround could be removed as soon as a new version is available / integrated.

maggu2810:
I assume it conflict has been caused by #5405.
Will rebase this branch.

Is there anything missing?

maggu2810:
I would like to postpone that change, because I am not sure (at least I don't like it) if the backend should be polluted with cleanup of input. The parser should have a well defined input and not bring in any magic to make a "wrong" config a valid one.

triller-telekom:
Alright, let's postpone that fix and discuss it elsewhere.

This PR is fine and can be merged!

triller-telekom:
@maggu2810 FYI: I have created #5408 as a follow-up for our discussion.

Issue #5294, opened 2018-03-21
CronExpressionTest: Cannot find a time after


maggu2810:
I cannot build the branch anymore.
It fails on the "core test".

```text
Results :

Failed tests: 
  CronExpressionTest.findNext:124 Cannot find a time after '2018-03-25 03:01:01.000' for expression '0 0 * ? * *'
```

Log: [log.txt](https://github.com/eclipse/smarthome/files/1834226/log.txt)

maggu2810:
DST Starts in Europe Sunday, March 25, 2018, see https://www.timeanddate.com/news/time/europe-starts-dst-2018.html
So it seems that the change is not handled correctly.
@kgoderis Would it be possible that you have a look at?

maggu2810:
@kaikreuzer Should we ignore that test until it is fixed?

kaikreuzer:
Yes, ok to ignore it - although I fear that it might indeed hint at an issue that users might have (again) on March 25... But it would anyhow too late to roll out a fix now :-/

PR #5296, opened 2018-03-22
CronExpression: start using fixed test input

maggu2810:
The current test execution depends on the TimeZone and current time of
the system the test is executed. That is not optimal as tests could
succeed on the one machine but fail on the other one.
Added another test (currently ignored, too), that uses the TimeZone and
start time the error occurs.

Related to: https://github.com/eclipse/smarthome/issues/5294

### DIFF
--- a/bundles/core/org.eclipse.smarthome.core.test/src/test/java/org/eclipse/smarthome/core/scheduler/CronExpressionTest.java
+++ b/bundles/core/org.eclipse.smarthome.core.test/src/test/java/org/eclipse/smarthome/core/scheduler/CronExpressionTest.java
@@ -20,6 +20,7 @@ import java.util.Arrays;
 import java.util.Calendar;
 import java.util.Date;
 import java.util.List;
+import java.util.TimeZone;
 
 import org.junit.Assert;
 import org.junit.Test;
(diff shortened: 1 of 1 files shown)


kaikreuzer:
This could be superseded by https://github.com/eclipse/smarthome/pull/5299 - If #5299 is successful, you might want to adapt this PR to not ignore the tests, but only add the new one to the suite.

kaikreuzer:
@maggu2810 #5299 has been merged.

maggu2810:
@kaikreuzer I changed the commit:

```text
    CronExpression: start using fixed test input
    
    The current test execution depends on the TimeZone and current time of
    the system the test is executed. That is not optimal as tests could
    succeed on the one machine but fail on the other one.
    Added another test (currently ignored, too), that uses a explicit set
    TimeZone and time.
    We use "millis from Epoc" to set the time, so we can easily difference
    between e.g. 2:30 nDST and 2:30 DST.
```

maggu2810:
* `findNext` still exists and is perhaps useful to generate some random test input
* `findNextSpecial` will use some special / fixed test input, so we can add ("from time to time") known "difficult" input

PR #5302, opened 2018-03-22
Intern UID segment strings

sjka:
...in order to avoid cluttering the heap with hundreds identical strings. Especially binding and thing type IDs are repeated all over the place.

UID derived classes are often held in maps within the framework and therefore actually are quite persistent in the heap. A quick analysis showed that something in the range of a little over 100K heap space could be saved just by that.

To me this looks like significant enough to consider, especially as [since Java 7 interned strings are stored in the heap](https://bugs.java.com/bugdatabase/view_bug.do?bug_id=6962931) (as opposed to the PermGen space where they were before).

### DIFF
--- a/bundles/core/org.eclipse.smarthome.core.thing/src/main/java/org/eclipse/smarthome/core/thing/UID.java
+++ b/bundles/core/org.eclipse.smarthome.core.thing/src/main/java/org/eclipse/smarthome/core/thing/UID.java
@@ -69,6 +69,7 @@ public abstract class UID {
         for (int i = 0; i < segments.length; i++) {
             String segment = segments[i];
             validateSegment(segment, i, segments.length);
+            segments[i] = segment.intern();
         }
         this.segments = segments;
     }


htreu:
Did you consider the maximum number of distinct Strings we would be able to intern w/o performance loss? According to [this blog post](http://java-performance.info/string-intern-in-java-6-7-8/) with Java 7u40+ the String table size is set to 60013 which allows storing 30000 distinct Strings.
Is this an issue?

sjka:
> Did you consider [...]

No

> Is this an issue?

Before: 

```
47205 interned Strings occupying 3889576 bytes.
```

After (on a plain ESH instance with very few bindings):

```
50371 interned Strings occupying 4169432 bytes.
```

=> Yes!

htreu:
What I took away from the blog post: It is in fact possible to store more then the mentioned 30000 entries. Its only a loss of performance since every map entry handles a bucket with LinkedList search performance. This will go down when there a re two many entries...
Quote from the blog (about the default value 60013):
> This value allows you to have approximately 30.000 distinct strings in the pool **before your start experiencing collisions**.

There is however `-XX:StringTableSize=N` which can be used to control the String table size (obviously). Maybe this is an option and we could in the end after all benefit from this PR.

kaikreuzer:
> There is however -XX:StringTableSize=N which can be used to control the String table size 

Which would be really ugly, if we'd require any ESH user/stack to set special JVM parameters in order to start at all. Imho not worth for saving a few hundred KBs...

htreu:
> in order to start at all.

I don't think this is about getting the system up and running but really sophisticated performance tuning. So on hardware where a small footprint is required/CPU performance is limited this can be fine tuned. imho all the others wouldn't notice a difference.

kaikreuzer:
What happens if you exceed 60013 distrinct strings? Will the system still start without the JVM param?

As you can see from @sjka's figures above, even a small ESH runtime that is made for embedded use with constraint resources is already taking 75% of the default String table size...

htreu:
> Will the system still start without the JVM param?

Yes it will. As @sjka already pointed out, interned Strings in Java 7+ are stored in the Heap and the storage (hashmap) can grow. With Java 6 you would have encountered OutOfMemoryError when interning too many Strings. Not so with java 7+.
The `StringTableSize` is the bucket count of the hashmap storing the interned Strings. When the hashmap detects has collisions the buckets will fill up.
So it will safe us space! And it _may_ degrade performance when the buckets get filled too much. With the default of 60013 the blog post showed how 1000000 Strings got interned w/o significant performance loss.

maybe this is something @JochenHiller would also comment on.

kaikreuzer:
Ok, thanks for the explanation, reading ALL details of the blog post indeed helps ;-)

@sjka might have closed the PR too early then...

maggu2810:
I haven't read all the technical details, blogs, spec etc. (I didn't know about that).
Just a question:
If my reading has been correct Java 8 Update 20 (or at least about that version) introduced `-XX:+UseStringDeduplication` shouldn't this do the job without any code changes? ESH itself already requires Java 8.

JochenHiller:
@htreu AFAI understand the blog post: the default size is just about a good default value which should fit to typical installations. The numbers @sjka pointed out in a basic test would fit to this range. So the fear from @kaikreuzer should be not a problem: it will always work, but might have performance impacts if the number of intern'd strings would increase too much, and "too" means really another dimension (10x more).
In such a case it can be even by optimized using `-XX:StringTableSize=N` by a specific solution if it knows it specific sizing.
So I would recommend to open that PR again and take care of about memory optimization >100 kB.

Interesting comment from @maggu2810 comment about `-XX:+UseStringDeduplication`:
Some background information: We ran across these optimizations and analysis by checking usage of a bigger amount of memory in our solution, where the root cause of memory comsumption were "equal" Strings but not identical Strings in a variant of an IdentityHashMap. This can be solved by intern that Strings which gave us a beneficial improvement. During analysis @sjka came across similar statistics about Strings heavily used in Eclipse SmartHome classes.

@sobimibos Could you repeat your tests with the option `-XX:+UseStringDeduplication` if this would solve our memory issues without a code change? I am just curious about that comment from @maggu2810 This would mean that such an optimization can also change behavior, where I am not sure if this is good or bad.

maggu2810:
I started a Wiki page so we could add lecture / comments about that topic (and later other such tasks) https://github.com/eclipse/smarthome/wiki/Java-Lecture

maggu2810:
I added `-XX:+UseG1GC -XX:+UseStringDeduplication -XX:StringDeduplicationAgeThreshold=3 -XX:+PrintStringDeduplicationStatistics` to a setup to get some statistics.

Here a test on a small setup with 53 things only:

```text
[GC concurrent-string-deduplication, 174.3K->88.5K(85.8K), avg 37.1%, 0.0015265 secs]
   [Last Exec: 0.0015265 secs, Idle: 136.6226873 secs, Blocked: 0/0.0000000 secs]
      [Inspected:            1939]
         [Skipped:              0(  0.0%)]
         [Hashed:             894( 46.1%)]
         [Known:               79(  4.1%)]
         [New:               1860( 95.9%)    174.3K]
      [Deduplicated:          979( 52.6%)     85.8K( 49.2%)]
         [Young:              979(100.0%)     85.8K(100.0%)]
         [Old:                  0(  0.0%)      0.0B(  0.0%)]
   [Total Exec: 32/0.4934939 secs, Idle: 32/270.8758200 secs, Blocked: 16/0.0747618 secs]
      [Inspected:          270935]
         [Skipped:              0(  0.0%)]
         [Hashed:          175684( 64.8%)]
         [Known:            34319( 12.7%)]
         [New:             236616( 87.3%)     22.0M]
      [Deduplicated:       130889( 55.3%)   8356.1K( 37.1%)]
         [Young:            21501( 16.4%)   1340.7K( 16.0%)]
         [Old:             109388( 83.6%)   7015.5K( 84.0%)]
   [Table]
      [Memory Usage: 4266.7K]
      [Size: 131072, Min: 1024, Max: 16777216]
      [Entries: 138357, Load: 105.6%, Cached: 0, Added: 139476, Removed: 1119]
      [Resize Count: 7, Shrink Threshold: 87381(66.7%), Grow Threshold: 262144(200.0%)]
      [Rehash Count: 0, Rehash Threshold: 120, Hash Seed: 0x0]
      [Age Threshold: 3]
   [Queue]
      [Dropped: 0]
[GC concurrent-string-deduplication, deleted 0 entries, 0.0000007 secs]
```

And another system with xxx things:

```text
[GC concurrent-string-deduplication, 456.0B->408.0B(48.0B), avg 32.9%, 0.0000470 secs]
   [Last Exec: 0.0000470 secs, Idle: 0.8319669 secs, Blocked: 0/0.0000000 secs]
      [Inspected:              72]
         [Skipped:              0(  0.0%)]
         [Hashed:               2(  2.8%)]
         [Known:               68( 94.4%)]
         [New:                  4(  5.6%)    456.0B]
      [Deduplicated:            1( 25.0%)     48.0B( 10.5%)]
         [Young:                1(100.0%)     48.0B(100.0%)]
         [Old:                  0(  0.0%)      0.0B(  0.0%)]
   [Total Exec: 38/0.2769161 secs, Idle: 38/409.4653517 secs, Blocked: 22/0.1043092 secs]
      [Inspected:          314076]
         [Skipped:              0(  0.0%)]
         [Hashed:          215413( 68.6%)]
         [Known:            35077( 11.2%)]
         [New:             278999( 88.8%)     26.8M]
      [Deduplicated:       143548( 51.5%)   9038.5K( 32.9%)]
         [Young:            37739( 26.3%)   2401.2K( 26.6%)]
         [Old:             105809( 73.7%)   6637.2K( 73.4%)]
   [Table]
      [Memory Usage: 4997.5K]
      [Size: 131072, Min: 1024, Max: 16777216]
      [Entries: 169052, Load: 129.0%, Cached: 486, Added: 172146, Removed: 3094]
      [Resize Count: 7, Shrink Threshold: 87381(66.7%), Grow Threshold: 262144(200.0%)]
      [Rehash Count: 0, Rehash Threshold: 120, Hash Seed: 0x0]
      [Age Threshold: 3]
   [Queue]
      [Dropped: 0]
[GC concurrent-string-deduplication, deleted 0 entries, 0.0000007 secs]
```

Would be great to get some statistics on systems with much more things, but cannot play with them as they are productive.

kaikreuzer:
How shall we proceed here? Shall we vote to use it or not? I myself am quite undecided...

htreu:
My suggestion is to close here and make String interning an implementation specific topic. Not a fixed solution in the framework.

kaikreuzer:
Makes sense to me. As solutions can use general jvm parameters, there isn't really a need to do it within the code itself.

maggu2810:
I agree with @htreu

PR #5306, opened 2018-03-23
fix some ntp tests not taking into account DaylightSavingTime

lpapazow:
Fix the two tests that are currently not working and another one that could potentially fail after DST takes effect.


extensions/binding/org.eclipse.smarthome.binding.ntp.test/src/test/java/org/eclipse/smarthome/binding/ntp/test/NtpOSGiTest.java
```diff
@@ -230,15 +231,17 @@ public void testDateTimeChannelTimeZoneUpdate() {
...
-    @Ignore("https://github.com/eclipse/smarthome/issues/5224")
     public void testDateTimeChannelCalendarTimeZoneUpdate() {
         Configuration configuration = new Configuration();
         configuration.put(NtpBindingConstants.PROPERTY_TIMEZONE, TEST_TIME_ZONE_ID);
         initialize(configuration, NtpBindingConstants.CHANNEL_DATE_TIME, ACCEPTED_ITEM_TYPE_DATE_TIME, null, null);
         ZonedDateTime timeZoneIdFromItemRegistry = ((DateTimeType) getItemState(ACCEPTED_ITEM_TYPE_DATE_TIME))
                 .getZonedDateTime();
         ZoneOffset testZoneId;
-        if (timeZoneIdFromItemRegistry.getZone().getRules().isDaylightSavings(timeZoneIdFromItemRegistry.toInstant())) {
+
```
lpapazow: I want to change the ZoneId instance on which getRules() method is called from timeZoneFromItemRegistry.getZone() to ZoneId.of(timeZoneInConfiguration)

Even though both of these are ZoneId instances, the second one provides more information:
printing timeZoneIdFromItemRegistry.getZone().getId() results in -07:00 and 
printing ZoneId.of(timeZoneInConfiguration).getId() results in 'America/Los_Angeles`

This information loss happens because of the formatter used in DateTimeType see this comment and the formatter can't easily be changed because it will affect PaperUI | see this comment

getRules() method returns a ZoneRules instance which has isDaylightSavings(Instant instant) method.
The way this method works is by calling the results of the getStandardOffset(Instant instant) and getOffset(Instant instant) methods for the ZoneRules instance. If both methods return different offsets, Java assumes DST is used in that instant.

As the JavaDoc states

The standard offset is the offset before any daylight saving time is applied. This is typically the offset applicable during winter.

Printing timeZoneIdFromItemRegistry.getZone().getRules().getStandardOffset(timeZoneIdFromItemRegistry.toInstant()) results in -07:00 
Printing ZoneId.of(timeZoneInConfiguration).getRules().getStandardOffset(timeZoneIdFromItemRegistry.toInstant()) results in -08:00 for timeZoneIdFromItemRegistry="2018-03-23T07:17:10.879-07:00" which is in DST in the Pacific Time Zone.

This test doesn't work because timeZoneIdFromItemRegistry.getZone() doesn't have the full information which causes timeZoneIdFromItemRegistry.getZone().getRules().isDaylightSavings(timeZoneIdFromItemRegistry.toInstant()) to give wrong result.

If there is no Timezone property available in the configuration, then the Default Timezone is used and using TimeZone.getDefault().toZoneId().getRules().isDaylightSavings(timeZoneIdFromItemRegistry.toInstant()) should return the correct result

maggu2810: @lpapazow But don't we need to change the `DateTimeType` and the UI interpretation of the format, then? If the `ZoneId` information of the `DateTimeType` loss information, it can cause bugs on other places sometime, too.

lpapazow: According to this [comment](https://github.com/eclipse/smarthome/pull/5226#issuecomment-375390293):

>both Paper UI and HABPanel use the AngularJS date filter to format the dates, which in turn simply uses the JavaScript Date object constructor accepting a string to deserialize them.

and the JavaScript Date constructor doesn't recognize string patterns with more specific timezone information than the offset. 

Do you think it will be easy to implement this changes in the UI?

maggu2810:
> I'm not very familiar with the UI, do you think it will be easy to implement this changes?

I didn't dig deeper into the topic, but as long as we are loosing information, we cannot workaround it correctly without fixing it.
In the test itself, you can perhaps rely on the timezone of the registry, because you know that this one is used, but it does not need to be correct for all the other usages of that type (correct me if I am wrong).

Can we rely on "correct information holding" at all as long as this test fail?

```Java
private static boolean isDst(final ZonedDateTime time) {
    return time.getZone().getRules().isDaylightSavings(time.toInstant());
}

private static ZonedDateTime fromTo(final ZonedDateTime time) {
    return new DateTimeType(time).getZonedDateTime();
}

private static boolean cmpDst(final ZonedDateTime time) {
    return isDst(time) == isDst(fromTo(time));
}

@Test
public void test() {
    ZonedDateTime zdtDst = ZonedDateTime.of(2018, 10, 28, 2, 30, 0, 0, ZoneId.of("Europe/Berlin"));
    ZonedDateTime zdtNonDst = zdtDst.plusHours(1);

    assertTrue(cmpDst(zdtDst));
    assertTrue(cmpDst(zdtNonDst));
}
```

lpapazow:
I agree it's not a good idea to use the timezone from the configuration even for test cases.

@ghys @sjka, Do you think changing the UI so that a more exhaustive formatter that doesn't lose any information could be used in the DateTimeType class?

ghys:
https://momentjs.com/timezone/ could probably do it, but it's an additional dependency and IMHO feels overkill.

See note 2 here: https://www.ecma-international.org/ecma-262/5.1/#sec-15.9.1.15

> There exists no international standard that specifies abbreviations for civil time zones like CET, EST, etc. and sometimes the same abbreviation is even used for two very different time zones. For this reason, ISO 8601 and this format specifies numeric representations of date and time.

extensions/binding/org.eclipse.smarthome.binding.ntp.test/src/test/java/org/eclipse/smarthome/binding/ntp/test/NtpOSGiTest.java
```diff
@@ -220,7 +219,9 @@ public void testDateTimeChannelTimeZoneUpdate() {
         ZonedDateTime timeZoneFromItemRegistry = ((DateTimeType) getItemState(ACCEPTED_ITEM_TYPE_DATE_TIME))
                 .getZonedDateTime();
         ZoneOffset expectedOffset;
-        if (timeZoneFromItemRegistry.getZone().getRules().isDaylightSavings(timeZoneFromItemRegistry.toInstant())) {
+        String timeZoneInConfiguration = (String) configuration.getProperties().get(NtpBindingConstants.PROPERTY_TIMEZONE);
```
triller-telekom: I second @sjka's explanation in https://github.com/eclipse/smarthome/pull/5226#issuecomment-376446165 and would vote for not reading out any configuration values but instead use fixed dates. Our string representation in https://github.com/eclipse/smarthome/blob/master/bundles/core/org.eclipse.smarthome.core/src/main/java/org/eclipse/smarthome/core/library/types/DateTimeType.java#L48 seems fine.

PR #5311, opened 2018-03-25
New system channel types for weather data

lolodomo:
Weather Underground binding updated.
Yahoo weather binding updated.


lolodomo:
@htreu @cweitkamp : can you have a first look and give a feedback ?
The work is done only partially.
@cweitkamp : I even tried a translation in German ;)

triller-telekom:
Adding some more system channel definitely makes sense!

However I am wondering whether we need several different `temperature` channels

```
channel-type.system.current-temperature.label = Aktuelle Temperatur
channel-type.system.current-temperature.description = Aktuelle Temperatur
channel-type.system.minimum-temperature.label = Mindesttemperatur
channel-type.system.minimum-temperature.description = Mindesttemperatur
channel-type.system.maximum-temperature.label = Maximale Temperatur
channel-type.system.maximum-temperature.description = Maximale Temperatur
channel-type.system.feeling-temperature.label = Gefühlstemperatur
channel-type.system.feeling-temperature.description = Gefühlstemperatur
channel-type.system.dew-point-temperature.label = Taupunkttemperatur
channel-type.system.dew-point-temperature.description = Taupunkttemperatur
channel-type.system.wind-chill-temperature.label = Windchill Temperatur
channel-type.system.wind-chill-temperature.description = Windchill Temperatur
```

This makes 6 temperature channels. How about adding only one with a unit etc. and the user can override its label and description?

I am not sure if a `heat-index` is something that is widely used. Even potential weather bindings do not all provide this information, so maybe we should not add this one as a system channel.

```
channel-type.system.heat-index.label = Hitzeindex
channel-type.system.heat-index.description = Hitzeindex
```

I guess we can keep `humidity` since this one seems to be rather popular for weather bindings or sensor bindings.

```
+channel-type.system.relative-humidity.label = Relative Luftfeuchtigkeit
+channel-type.system.relative-humidity.description = Relative Luftfeuchtigkeit in %
```

For pressure I would argue like I did for temperature, how about creating one `pressure` channel and the user can override its label and description?

```
+channel-type.system.barometric-pressure.label = Luftdruck
+channel-type.system.barometric-pressure.description = Luftdruck
+channel-type.system.absolute-pressure.label = Absoluter Druck
+channel-type.system.absolute-pressure.description = Absoluter Druck
```

kaikreuzer:
Instead of specifying all kinds of detailed system channels that are specific to weather, I would very much prefer to first pick up the discussion on https://github.com/eclipse/smarthome/issues/3756 and come up with a good set of initial system channels (which also contain some weather specific stuff, but not only those). Would that make sense?

lolodomo:
@triller-telekom : regarding the channel type granularity, defining one unique channel type for any temperature was my first approach when I implemented the WU binding but @kaikreuzer rejected it explaining what should be the expected granularity of a channel type. He even updated the documentation with a concrete example with temperature : https://github.com/eclipse/smarthome/pull/3419/files

lolodomo:
The other issue is already opened since almost one year. So I am trying to make a new step focussing on weather, now that UoM is available, considering the different bindings handling such data: WU, Yahoo weather and netatmo.

lolodomo:
Finally I made the effort to finish the creation of system channel types for the main weather data that could be handled by bindings.

lolodomo:
The two existing ESH weather bindings are updated to use the new system channels.
Once this PR is merged, I will then update the Netatmo openHAB binding.

sjka:
In the light of #3756, let's close this one for now to de-duplicate efforts.

PR #5316, opened 2018-03-26
Fix isChannelLinked logic

triller-telekom:
PR #5244 moved the ItemChannelLinkRegistry from the BaseThingHandler to
the ThingHandlerCallback but inverted its logic. This PR fixes the logic
again.


cweitkamp:
@triller-telekom What a pitty from my side. I just thought it is already April's Fool. Thank you very much.

@kaikreuzer Yes, I will add a unit test.

Issue #3335, opened 2017-04-28
Model validation fails with NPE


kaikreuzer:
I sometimes find exceptions like the following in my log:
```
17:21:16.762 [ERROR] [xbase.resource.BatchLinkableResource] - resolution of uriFragment '|::0.2.0.2.0.0.0.6.7::0::/1' failed.
java.lang.NullPointerException
	at org.eclipse.xtext.xbase.resource.BatchLinkingService.resolveBatched(BatchLinkingService.java:62)
	at org.eclipse.xtext.xbase.resource.BatchLinkingService.resolveBatched(BatchLinkingService.java:41)
	at org.eclipse.xtext.xbase.resource.BatchLinkableResource.getEObject(BatchLinkableResource.java:117)
	at org.eclipse.emf.ecore.resource.impl.ResourceSetImpl.getEObject(ResourceSetImpl.java:223)
	at org.eclipse.emf.ecore.util.EcoreUtil.resolve(EcoreUtil.java:199)
	at org.eclipse.emf.ecore.util.EcoreUtil.resolve(EcoreUtil.java:259)
	at org.eclipse.emf.ecore.impl.BasicEObjectImpl.eResolveProxy(BasicEObjectImpl.java:1477)
	at org.eclipse.xtext.xbase.impl.XAbstractFeatureCallImplCustom.getFeature(XAbstractFeatureCallImplCustom.java:48)
	at org.eclipse.xtext.xbase.impl.XAbstractFeatureCallImpl.eGet(XAbstractFeatureCallImpl.java:490)
	at org.eclipse.xtext.xbase.impl.XFeatureCallImpl.eGet(XFeatureCallImpl.java:256)
	at org.eclipse.emf.ecore.impl.BasicEObjectImpl.eGet(BasicEObjectImpl.java:1011)
	at org.eclipse.emf.ecore.impl.BasicEObjectImpl.eGet(BasicEObjectImpl.java:1003)
	at org.eclipse.emf.ecore.util.EContentsEList$FeatureIteratorImpl.hasNext(EContentsEList.java:439)
	at org.eclipse.emf.ecore.util.EObjectValidator.validate_EveryReferenceIsContained(EObjectValidator.java:782)
	at org.eclipse.emf.ecore.util.EObjectValidator.validate_EveryDefaultConstraint(EObjectValidator.java:355)
	at org.eclipse.emf.ecore.util.EObjectValidator$DynamicEClassValidator.validate(EObjectValidator.java:1420)
[...]

PR #5318, opened 2018-03-27
Fixed AvoidCatchingNPE and AvoidThrowingRawExceptionTypes issues found by PMD

VelinYordanov:
Fixed pmd issues found is smarthome - AvoidCatchingNPE and AvoidThrowingRawExceptionTypes.

### DIFF
--- a/bundles/automation/org.eclipse.smarthome.automation.module.script/src/main/java/org/eclipse/smarthome/automation/module/script/internal/handler/AbstractScriptModuleHandler.java
+++ b/bundles/automation/org.eclipse.smarthome.automation.module.script/src/main/java/org/eclipse/smarthome/automation/module/script/internal/handler/AbstractScriptModuleHandler.java
@@ -90,23 +90,25 @@ public abstract class AbstractScriptModuleHandler<T extends Module> extends Base
     private void loadConfig() {
         Object type = module.getConfiguration().get(SCRIPT_TYPE);
         Object script = module.getConfiguration().get(SCRIPT);
-        if (type == null || !(type instanceof String) || ((String) type).trim().isEmpty()) {
-            throw new RuntimeException(
-                    String.format("Type is missing in the configuration of module '%s'.", module.getId()));
-        } else if (script == null || !(script instanceof String) || ((String) script).trim().isEmpty()) {
-            throw new RuntimeException(
-                    String.format("Script is missing in the configuration of module '%s'.", module.getId()));
+        if (!isValid(type)) {
+            throw new IllegalStateException(String.format("Type is missing in the configuration of module '%s'.", module.getId()));
+        } else if (!isValid(script)) {
+            throw new IllegalStateException(String.format("Script is missing in the configuration of module '%s'.", module.getId()));
         } else {
             this.type = (String) type;
             this.script = (String) script;
         }
+    }
+    
+    private boolean isValid(Object parameter) {
+        return parameter != null && parameter instanceof String && !((String) parameter).trim().isEmpty();
     }
 
     /**
      * Adds the passed context variables of the rule engine to the context scope of the ScriptEngine, this should be
      * updated each time the module is executed
      *
-     * @param engine the scriptengine that is used
+     * @param engine the script engine that is used
      * @param context the variables and types to put into the execution context
      */
     protected void setExecutionContext(ScriptEngine engine, Map<String, ?> context) {
--- a/bundles/io/org.eclipse.smarthome.io.rest/src/main/java/org/eclipse/smarthome/io/rest/JSONResponse.java
+++ b/bundles/io/org.eclipse.smarthome.io.rest/src/main/java/org/eclipse/smarthome/io/rest/JSONResponse.java
@@ -164,7 +164,7 @@ public class JSONResponse {
             PipedInputStream in = new PipedInputStream(out);
             rp.entity(in);
         } catch (IOException e) {
-            throw new RuntimeException(e);
+            throw new IllegalStateException(e);
         }
 
         Thread writerThread = new Thread(() -> {
--- a/extensions/binding/org.eclipse.smarthome.binding.hue/src/main/java/org/eclipse/smarthome/binding/hue/handler/HueBridgeHandler.java
+++ b/extensions/binding/org.eclipse.smarthome.binding.hue/src/main/java/org/eclipse/smarthome/binding/hue/handler/HueBridgeHandler.java
@@ -515,16 +515,19 @@ public class HueBridgeHandler extends ConfigStatusBridgeHandler {
 
         boolean colorModeIsEqual = true;
         boolean effectIsEqual = true;
-        try {
-            colorModeIsEqual = state1.getColorMode().equals(state2.getColorMode());
-        } catch (NullPointerException npe) {
+        
+        if (state1.getColorMode() == null) {
             logger.trace("Light does not support color mode.");
+        } else {
+            colorModeIsEqual = state1.getColorMode().equals(state2.getColorMode());
         }
-        try {
-            effectIsEqual = state1.getEffect().equals(state2.getEffect());
-        } catch (NullPointerException npe) {
+        
+        if (state1.getEffect() == null) {
             logger.trace("Light does not support effect.");
+        } else {
+            effectIsEqual = state1.getEffect().equals(state2.getEffect());
         }
+        
         return colorModeIsEqual && effectIsEqual;
     }
 
--- a/extensions/binding/org.eclipse.smarthome.binding.hue/src/main/java/org/eclipse/smarthome/binding/hue/internal/HueBridge.java
+++ b/extensions/binding/org.eclipse.smarthome.binding.hue/src/main/java/org/eclipse/smarthome/binding/hue/internal/HueBridge.java
@@ -873,6 +873,10 @@ public class HueBridge {
                 }
 
                 for (ErrorResponse error : errors) {
+                    if (error == null) {
+                        continue;
+                    }
+                    
                     switch (error.getType()) {
                         case 1:
                             username = null;
@@ -893,8 +897,6 @@ public class HueBridge {
                 }
             } catch (JsonParseException e) {
                 // Not an error
-            } catch (NullPointerException e) {
-                // Object that looks like error
             }
         }
     }
--- a/tools/static-code-analysis/pmd/suppressions.properties
+++ b/tools/static-code-analysis/pmd/suppressions.properties
@@ -3,3 +3,5 @@ org.eclipse.smarthome.io.console.karaf.internal.OSGiConsole=SystemPrintln
 org.eclipse.smarthome.io.console.rfc147.internal.CommandWrapper=SystemPrintln
 org.eclipse.smarthome.io.console.rfc147.internal.OSGiConsole=SystemPrintln
 org.eclipse.smarthome.core.internal.events.ThreadedEventHandler=EmptyIfStmt
+org.eclipse.smarthome.model.core.internal.ModelRepositoryImpl=AvoidCatchingNPE
+org.eclipse.smarthome.io.net.http.internal.SecureHttpClientFactory=AvoidThrowingRawExceptionTypes
--- a/bundles/config/org.eclipse.smarthome.config.xml/src/main/java/org/eclipse/smarthome/config/xml/ConfigDescriptionConverter.java
+++ b/bundles/config/org.eclipse.smarthome.config.xml/src/main/java/org/eclipse/smarthome/config/xml/ConfigDescriptionConverter.java
@@ -65,9 +65,14 @@ public class ConfigDescriptionConverter extends GenericUnmarshaller<ConfigDescri
         }
 
         URI uri = null;
+        if (uriText == null) {
+            throw new ConversionException(
+                    "No URI provided");
+        }
+
         try {
             uri = new URI(uriText);
-        } catch (NullPointerException | URISyntaxException ex) {
+        } catch (URISyntaxException ex) {
             throw new ConversionException(
                     "The URI '" + uriText + "' in node '" + reader.getNodeName() + "' is invalid!", ex);
         }
@@ -82,7 +87,7 @@ public class ConfigDescriptionConverter extends GenericUnmarshaller<ConfigDescri
 
         // iterate through the nodes, putting the different types into their
         // respective arrays
-        while (nodeIterator.hasNext() == true) {
+        while (nodeIterator.hasNext()) {
             Object node = nodeIterator.next();
             if (node instanceof ConfigDescriptionParameter) {
                 configDescriptionParams.add((ConfigDescriptionParameter) node);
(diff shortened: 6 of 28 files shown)


bundles/automation/org.eclipse.smarthome.automation.module.script/src/main/java/org/eclipse/smarthome/automation/module/script/internal/handler/AbstractScriptModuleHandler.java
```diff
@@ -29,20 +29,28 @@
 import org.slf4j.LoggerFactory;
 
 /**
- * This is an abstract class that can be used when implementing any module handler that handles scripts.
+ * This is an abstract class that can be used when implementing any module
```
htreu: Please check your formatter settings. The maximum line width for ESH is 120.

bundles/automation/org.eclipse.smarthome.automation.module.script/src/main/java/org/eclipse/smarthome/automation/module/script/internal/handler/AbstractScriptModuleHandler.java
```diff
@@ -90,24 +98,27 @@ public void dispose() {
...
         Object type = module.getConfiguration().get(SCRIPT_TYPE);
         Object script = module.getConfiguration().get(SCRIPT);
-        if (type == null || !(type instanceof String) || ((String) type).trim().isEmpty()) {
-            throw new RuntimeException(
+        if (isNotValid(type) || isNotValid(script)) {
+            throw new IllegalStateException(
                     String.format("Type is missing in the configuration of module '%s'.", module.getId()));
-        } else if (script == null || !(script instanceof String) || ((String) script).trim().isEmpty()) {
-            throw new RuntimeException(
-                    String.format("Script is missing in the configuration of module '%s'.", module.getId()));
```
htreu: With this removal the individual error description gets lost.

bundles/io/org.eclipse.smarthome.io.net/src/main/java/org/eclipse/smarthome/io/net/http/internal/SecureHttpClientFactory.java
```diff
@@ -174,7 +174,7 @@ public Void run() {
                 if (cause instanceof RuntimeException) {
                     throw (RuntimeException) cause;
                 } else {
-                    throw new RuntimeException(cause);
```
htreu: With the lines above this looks like the RuntimeException is thrown on purpose here. @sobimibos can you clarify here what we can do alternatively? Thanks.

bundles/io/org.eclipse.smarthome.io.net/src/main/java/org/eclipse/smarthome/io/net/http/internal/SecureHttpClientFactory.java
```diff
@@ -232,7 +232,7 @@ public HttpClient run() {
             if (cause instanceof RuntimeException) {
                 throw (RuntimeException) cause;
             } else {
-                throw new RuntimeException(cause);
```
htreu: Same as above.

bundles/io/org.eclipse.smarthome.io.rest/src/main/java/org/eclipse/smarthome/io/rest/JSONResponse.java
```diff
@@ -164,7 +164,7 @@ private Response createResponse(Status status, Object entity) {
             PipedInputStream in = new PipedInputStream(out);
             rp.entity(in);
         } catch (IOException e) {
-            throw new RuntimeException(e);
+            throw new IllegalStateException("I/O error occurred"+e);
```
htreu: please add the original exception as a parameter here.

extensions/binding/org.eclipse.smarthome.binding.hue/src/main/java/org/eclipse/smarthome/binding/hue/handler/HueBridgeHandler.java
```diff
@@ -493,15 +509,17 @@ private void notifyLightStatusListeners(final FullLight fullLight, final String
...
+     * first we check for the common properties which are set for every light type.
+     * If they equal, we additionally try to check the colorMode. If we get an NPE,
+     * the light does not support color mode and the common properties equality is
+     * our result: true. Otherwise if no NPE occurs the equality of colorMode is our
+     * result.
      *
-     * @param state1 Reference state
-     * @param state2 State which is checked for equality.
+     * @param state1
+     *            Reference state
```
htreu: this line break is also not standard with the ESH code formatter.

extensions/binding/org.eclipse.smarthome.binding.hue/src/main/java/org/eclipse/smarthome/binding/hue/internal/HueBridge.java
```diff
@@ -161,24 +168,25 @@ public boolean isAuthenticated() {
         String lastScan = safeFromJson(result.getBody(), NewLightsResponse.class).lastscan;
 
         switch (lastScan) {
-            case "none":
+        case "none":
```
htreu: this is also not ESH formatter conform.

bundles/io/org.eclipse.smarthome.io.net/src/main/java/org/eclipse/smarthome/io/net/http/internal/SecureHttpClientFactory.java
VelinYordanov: I discarded the changes and added an exception for this class for now.

bundles/automation/org.eclipse.smarthome.automation.module.script/src/main/java/org/eclipse/smarthome/automation/module/script/internal/handler/AbstractScriptModuleHandler.java
```diff
@@ -90,23 +90,25 @@ public void dispose() {
...
+            throw new IllegalStateException(String.format("Type is missing in the configuration of module '%s'.", module.getId()));
+        } else if(isNotValid(script)) {
+            throw new IllegalStateException(String.format("Script is missing in the configuration of module '%s'.", module.getId()));
         } else {
             this.type = (String) type;
             this.script = (String) script;
         }
     }
+    
+    private boolean isNotValid(Object parameter) {
```
maggu2810: I see that we would like to check if the parameter is not valid in the cases above.
But I don't think that methods that check that something is not true should be used.
If we want to check if a collection is not empty, we call `!collection.isEmpty()` instead of `collection.isNotEmpty()`.
I would also prefer to provide a method `isValid(Object parameter)` and use `!isValid(type)` above.

bundles/automation/org.eclipse.smarthome.automation.module.script/src/main/java/org/eclipse/smarthome/automation/module/script/internal/handler/AbstractScriptModuleHandler.java
```diff
@@ -90,23 +90,25 @@ public void dispose() {
...
+        return (parameter == null) || !(parameter instanceof String) || ((String) parameter).trim().isEmpty();
```
maggu2810: Why do we add braces that are not necessary? At least the one around `parameter == null` could be removed.

extensions/binding/org.eclipse.smarthome.binding.hue/src/main/java/org/eclipse/smarthome/binding/hue/internal/HueBridge.java
```diff
@@ -873,6 +873,10 @@ private void handleErrors(Result result) throws IOException, ApiException {
                 }
 
                 for (ErrorResponse error : errors) {
+                    if(error == null) {
```
maggu2810: It seems the formatter is still not correctly applied to all changes.

tools/static-code-analysis/pmd/suppressions.properties
```diff
@@ -3,3 +3,8 @@ org.eclipse.smarthome.io.console.karaf.internal.OSGiConsole=SystemPrintln
 org.eclipse.smarthome.io.console.rfc147.internal.CommandWrapper=SystemPrintln
 org.eclipse.smarthome.io.console.rfc147.internal.OSGiConsole=SystemPrintln
 org.eclipse.smarthome.core.internal.events.ThreadedEventHandler=EmptyIfStmt
+org.eclipse.smarthome.config.xml.ConfigDescriptionConverter=AvoidCatchingNPE
```
maggu2810: Is it still necessary to catch NPE?
Can you point me to the places where it cannot be checked before?

VelinYordanov: I think the check should be suppressed for ModelRepositoryImpl and JavaTest. The only reason for the rest is that removing the catching of NPE would lead to longer and a bit uglier code. I could refactor them as well if you prefer.

VelinYordanov:
I also fixed all SimplifyBooleanExpressions issues.

tools/static-code-analysis/pmd/suppressions.properties
maggu2810: Sorry for the delay...

> The only reason for the rest is that removing the catching of NPE would lead to longer and a bit uglier code. I could refactor them as well if you prefer.

I cannot judge without having a look at.
Test for null and prevent NPEs in front of instead of catching it normally result into a cleaner (perhaps this depends on the POV) code. I will have a short look at that classes and try to understand why the code gets "uglier".

tools/static-code-analysis/pmd/suppressions.properties
```diff
@@ -3,3 +3,8 @@ org.eclipse.smarthome.io.console.karaf.internal.OSGiConsole=SystemPrintln
...
+org.eclipse.smarthome.io.rest.core.internal.thing.ThingResource=AvoidCatchingNPE
+org.eclipse.smarthome.test.java.JavaTest=AvoidCatchingNPE
+org.eclipse.smarthome.model.core.internal.ModelRepositoryImpl=AvoidCatchingNPE
```
maggu2810: WRT https://github.com/eclipse/smarthome/issues/3335 and the Xtext magic, I don't care -- it's black magic for me :wink:

tools/static-code-analysis/pmd/suppressions.properties
```diff
@@ -3,3 +3,8 @@ org.eclipse.smarthome.io.console.karaf.internal.OSGiConsole=SystemPrintln
 org.eclipse.smarthome.io.console.rfc147.internal.CommandWrapper=SystemPrintln
 org.eclipse.smarthome.io.console.rfc147.internal.OSGiConsole=SystemPrintln
 org.eclipse.smarthome.core.internal.events.ThreadedEventHandler=EmptyIfStmt
+org.eclipse.smarthome.config.xml.ConfigDescriptionConverter=AvoidCatchingNPE
+org.eclipse.smarthome.io.rest.core.internal.thing.ThingResource=AvoidCatchingNPE
+org.eclipse.smarthome.test.java.JavaTest=AvoidCatchingNPE
```
maggu2810: Perhaps okay for relaxed test writings, but hey, we are using pure Java tests and no Groovy, so should be easy to fix (you could also do it in a separate PR and create an tracker issue now if too much tests needs to be fixed).

tools/static-code-analysis/pmd/suppressions.properties
```diff
@@ -3,3 +3,8 @@ org.eclipse.smarthome.io.console.karaf.internal.OSGiConsole=SystemPrintln
 org.eclipse.smarthome.io.console.rfc147.internal.CommandWrapper=SystemPrintln
 org.eclipse.smarthome.io.console.rfc147.internal.OSGiConsole=SystemPrintln
 org.eclipse.smarthome.core.internal.events.ThreadedEventHandler=EmptyIfStmt
+org.eclipse.smarthome.config.xml.ConfigDescriptionConverter=AvoidCatchingNPE
+org.eclipse.smarthome.io.rest.core.internal.thing.ThingResource=AvoidCatchingNPE
```
maggu2810: Shouldn't this be easy to fix? At which places do we expect that an NPE is thrown?

tools/static-code-analysis/pmd/suppressions.properties
maggu2810: This one is rather easy to fix.
The URI constructor will throw a NPE if the provided text is null only (AFAIK).

maggu2810: Shouldn't this already do the job?

```diff
@@ -62,12 +62,15 @@ public class ConfigDescriptionConverter extends GenericUnmarshaller<ConfigDescri
             // the URI could be overridden by a context field if it could be
             // automatically extracted
             uriText = (String) context.get("config-description.uri");
+            if (uriText == null) {
+                throw new ConversionException("Missing URI.");
+            }
         }
 
         URI uri = null;
         try {
             uri = new URI(uriText);
-        } catch (NullPointerException | URISyntaxException ex) {
+        } catch (URISyntaxException ex) {
             throw new ConversionException(
                     "The URI '" + uriText + "' in node '" + reader.getNodeName() + "' is invalid!", ex);
         }
```

VelinYordanov: Yes, I updated the PR with the required changes.

tools/static-code-analysis/pmd/suppressions.properties
```diff
@@ -3,3 +3,5 @@ org.eclipse.smarthome.io.console.karaf.internal.OSGiConsole=SystemPrintln
 org.eclipse.smarthome.io.console.rfc147.internal.CommandWrapper=SystemPrintln
 org.eclipse.smarthome.io.console.rfc147.internal.OSGiConsole=SystemPrintln
 org.eclipse.smarthome.core.internal.events.ThreadedEventHandler=EmptyIfStmt
+org.eclipse.smarthome.model.core.internal.ModelRepositoryImpl=AvoidCatchingNPE
+org.eclipse.smarthome.io.net.http.internal.SecureHttpClientFactory=AvoidThrowingRawExceptionTypes
```
maggu2810: One question remaining about that one (so, this is perhaps more addresses to the author of that file and other ones caring about the API):
* Why not using a more specific exception instead of encapsulate all ones into `RuntimeException` (at the relevant places)?
* Why not using a checked exception instead of an unchecked? If there is a potential error the caller should handle it correctly. Now, the caller gets not informed (without reading the called function implementation) that he needs to handle that error.

VelinYordanov:
I added a suppression for the `SecureHttpClientFactory` in regards to @htreu 's review

> With the lines above this looks like the RuntimeException is thrown on purpose here. @sobimibos can you clarify here what we can do alternatively? Thanks.

maggu2810:
> * Why not using a more specific exception instead of encapsulate all ones into `RuntimeException` (at the relevant places)?
> * Why not using a checked exception instead of an unchecked? If there is a potential error the caller should handle it correctly. Now, the caller gets not informed (without reading the called function implementation) that he needs to handle that error.

 @sobimibos can you please give us some details?

maggu2810:
@VelinYordanov I added another issue, so this one could be merged as soon as Travis CI succeeded.

PR #5331, opened 2018-03-31
Extend hex utils to include the reverse operation

martinvw:
Relates to https://github.com/openhab/openhab2-addons/pull/3430

openHAB needs this because we want to get rid of using the javax DatatypeConverter

### DIFF
--- a/bundles/core/org.eclipse.smarthome.core/src/main/java/org/eclipse/smarthome/core/util/HexUtils.java
+++ b/bundles/core/org.eclipse.smarthome.core/src/main/java/org/eclipse/smarthome/core/util/HexUtils.java
@@ -22,16 +22,19 @@ import org.eclipse.jdt.annotation.Nullable;
  * Static utility methods that are helpful when dealing with hex data and byte arrays.
  *
  * @author Kai Kreuzer - Initial contribution and API
- *
+ * @author Martin van Wingerden - Implemented the reverse operation
  */
 @NonNullByDefault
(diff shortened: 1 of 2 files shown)


bundles/core/org.eclipse.smarthome.core/src/main/java/org/eclipse/smarthome/core/util/HexUtils.java
```diff
@@ -64,4 +67,57 @@ public static String bytesToHex(byte[] bytes) {
...
+     *
+     * @param hexString the hex string
+     * @return the corresponding byte array
+     */
+    public static byte[] hexToBytes(String hexString) {
+        return hexToBytes(hexString, "(?<=\\G.{2})");
+    }
+
+    /**
+     * Convert an upper case hex character to an integer
```
kaikreuzer: -> ... to a byte
Would be the better fit

kaikreuzer: Assuming you agree to that change, I just did it myself, so that it can be merged right away :-)

martinvw:
Thanks, can you ping me once it’s included in openHAB it allows me to kill all references to javax.xml

kaikreuzer:
Ping! https://ci.eclipse.org/smarthome/job/SmartHomeDistribution-Stable/262/

PR #5342, opened 2018-04-03
Fixed CompareObjectsWithEquals issues found by PMD

VelinYordanov:
Fixed [CompareObjectsWithEquals ](https://pmd.github.io/pmd-5.8.1/pmd-java/rules/java/design.html#CompareObjectsWithEquals)issues found by PMD.

I noticed that refactoring from `==` to `equals` breaks existing tests in **org.eclipse.smarthome.automation.core.internal.RuleRegistryImpl** and
**org.eclipse.smarthome.core.thing.internal.ChannelItemProvider** so I assume using `==` there is intentional.

### DIFF
--- a/tools/static-code-analysis/pmd/suppressions.properties
+++ b/tools/static-code-analysis/pmd/suppressions.properties
@@ -5,3 +5,9 @@ org.eclipse.smarthome.io.console.rfc147.internal.OSGiConsole=SystemPrintln
 org.eclipse.smarthome.core.internal.events.ThreadedEventHandler=EmptyIfStmt
 org.eclipse.smarthome.model.core.internal.ModelRepositoryImpl=AvoidCatchingNPE
 org.eclipse.smarthome.io.net.http.internal.SecureHttpClientFactory=AvoidThrowingRawExceptionTypes
+org.eclipse.smarthome.automation.core.internal.RuleRegistryImpl=CompareObjectsWithEquals
+org.eclipse.smarthome.core.thing.internal.ChannelItemProvider=CompareObjectsWithEquals
+org.eclipse.smarthome.core.thing.internal.ThingManager=CompareObjectsWithEquals
+org.eclipse.smarthome.core.internal.common.SafeCallManagerImpl=CompareObjectsWithEquals
+org.eclipse.smarthome.core.internal.events.ThreadedEventHandler=CompareObjectsWithEquals
+
--- a/bundles/config/org.eclipse.smarthome.config.xml/src/main/java/org/eclipse/smarthome/config/xml/util/GenericUnmarshaller.java
+++ b/bundles/config/org.eclipse.smarthome.config.xml/src/main/java/org/eclipse/smarthome/config/xml/util/GenericUnmarshaller.java
@@ -52,7 +52,7 @@ public abstract class GenericUnmarshaller<T> implements Converter {
     @SuppressWarnings("rawtypes")
     @Override
     public final boolean canConvert(Class paramClass) {
-        return (clazz == paramClass);
+        return (clazz.equals(paramClass));
     }
 
     @Override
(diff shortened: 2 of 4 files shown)


tools/static-code-analysis/pmd/suppressions.properties
```diff
@@ -5,3 +5,6 @@ org.eclipse.smarthome.io.console.rfc147.internal.OSGiConsole=SystemPrintln
 org.eclipse.smarthome.core.internal.events.ThreadedEventHandler=EmptyIfStmt
 org.eclipse.smarthome.model.core.internal.ModelRepositoryImpl=AvoidCatchingNPE
 org.eclipse.smarthome.io.net.http.internal.SecureHttpClientFactory=AvoidThrowingRawExceptionTypes
+org.eclipse.smarthome.automation.core.internal.RuleRegistryImpl=CompareObjectsWithEquals
```
maggu2810: Which line? I only found ` == null` and ` != null` in that file.

maggu2810:
> org.eclipse.smarthome.core.thing.internal.ChannelItemProvider

I agree in `beforeAdding` and `afterRemoving` the check for the same reference is used by intention.

tools/static-code-analysis/pmd/suppressions.properties
VelinYordanov: The line is 513.

bundles/core/org.eclipse.smarthome.core.thing/src/main/java/org/eclipse/smarthome/core/thing/internal/ThingManager.java
```diff
@@ -475,7 +475,7 @@ public void thingUpdated(final Thing thing, ThingTrackerEvent thingTrackerEvent)
 
     private ThingHandler replaceThing(Thing oldThing, Thing newThing) {
         final ThingHandler thingHandler = thingHandlers.get(newThing.getUID());
-        if (oldThing != newThing) {
+        if (oldThing.equals(newThing)) {
```
sjka: Nope, that was by intention!

We currently don't create copies and equals only goes against the UID - which by definition won't change.

bundles/core/org.eclipse.smarthome.core/src/main/java/org/eclipse/smarthome/core/internal/common/SafeCallManagerImpl.java
```diff
@@ -123,7 +123,7 @@ private void trigger(Object identifier) {
     private void handlePotentialTimeout(Invocation invocation) {
         Object identifier = invocation.getIdentifier();
         Invocation activeAsyncInvocation = activeAsyncInvocations.get(identifier);
-        if (activeAsyncInvocation == invocation) {
+        if (activeAsyncInvocation.equals(invocation)) {
```
sjka: nope!

bundles/core/org.eclipse.smarthome.core/src/main/java/org/eclipse/smarthome/core/internal/events/ThreadedEventHandler.java
```diff
@@ -65,7 +65,7 @@ public ThreadedEventHandler(final Map<String, EventFactory> typedEventFactories,
                     final Event event = queue.poll(1, TimeUnit.HOURS);
                     if (event == null) {
                         logger.debug("Hey, you have really very few events.");
-                    } else if (event == notifyEvent) {
+                    } else if (event.equals(notifyEvent)) {
```
sjka: Nope - we have a de-facto singleton instance of notifyEvent and are comparing it against events of potentially unknown sources where we don't know what they will do in their `equals()`-implementations, hence I'd suggest leaving this as it is.

Issue #4964, opened 2018-01-22
Remove Firmware UID - Firmware provider should provide firmwares for things


tomhoefer:
In ESH, firmwares are identified by their FirmwareUID, which consists of the firmware's thing type UID, and the firmware version. The firmware UID must be unique amongst all firmwares. I.e., each firmware for a given thing type must have a unique version. This is not the case for the following scenario:

In our project there are two firmwares for a special thing type with version 1, and two firmwares for that thing type with version 2 (two firmwares because of two different models). In consequence, for each version only one firmware is considered (and displayed as result of the 'firmware list' console command). Hence the solution introduced by #4446 works only if the different models have also different firmware versions. 

For this reason we cannot consider the firmware UID as unique. I would like to remove the Firmware UID completely and refactor the firmware provider / registry in the manner that it provides firmwares for things instead of thing types. Then the provider can determine which firmware is to be provided for the given thing (even if there are multiple firmwares available for the same thing type and version). This would also fix the issue reported here https://www.eclipse.org/forums/index.php?t=rview&goto=1779174.

Should also fix #4647

htreu:
Hi @tomhoefer. Your issue, as well as the linked discussions all seem to agree that a refactoring of the current firmware API makes sense. Since there are no other consumers (at least known) of the firmware API I guess we could change and go from thing-type to the thing entity here. This will even add more information available to the FirmwareProvider, so migrating should be quite simple.
It would be nice to see a PR proposing the changes.
Any objections @kaikreuzer, @sjka & @maggu2810?

cdjackson:
> Since there are no other consumers (at least known) of the firmware API

Please note that any change here will break the ZigBee binding. I've no objection to this and had raised the same issue with @tomhoefer but I'd like to coordinate the release so as not to break the binding.

tomhoefer:
@cdjackson I am aware of the FirmwareProvider inside the ZigBee binding. Currently I consider that provider as a development tool that we need to adapt accordingly. There are some more providers used in our project that need to be adapted, too.

cdjackson:
Thanks @tomhoefer - I know you're aware, but I just wanted to comment here for everyone elses visibility...

maggu2810:
> Any objections @kaikreuzer, @sjka & @maggu2810?

[...]

PR #5343, opened 2018-04-03
Firmware Provider Provides Firmwares for Things

dstivanov:
The following changes are made in regard to  #4964 :

- Initial refactoring on the provisioning the firmwares by thing instead of a thing type;
- All the firmware events, that contained thingUID are now containing only info objects, which consolidate the whole information for the event;
- The javadoc is updated;
- All the existing tests are adapted to the changes;
- Implemented new tests for the FirmwareEventFactory;
- Adapted the PaperUI controllers in regard to the changes in the thing type and thing REST endpoints.


sjka:
It is API-Breaking, but as outlined in https://github.com/eclipse/smarthome/issues/4964#issuecomment-360738963 this should be okay. This change was done in the realm of the DT project and no other objections have been raised so far.

bundles/core/org.eclipse.smarthome.core.thing/src/main/java/org/eclipse/smarthome/core/thing/internal/console/FirmwareUpdateConsoleCommandExtension.java
```diff
@@ -180,7 +188,7 @@ private FirmwareUpdateHandler getFirmwareUpdateHandler(ThingUID thingUID) {
     @Override
     public List<String> getUsages() {
         return Arrays.asList(new String[] {
-                buildCommandUsage(SUBCMD_LIST + " <thingTypeUID>", "lists the available firmwares for a thing type"),
+                buildCommandUsage(SUBCMD_LIST + " <thingUID>", "lists the available firmwares for a thing with specified UID"),
```
sjka: ~~with specified UID~~

(Keep it consistent with below)

bundles/io/org.eclipse.smarthome.io.rest.core/src/main/java/org/eclipse/smarthome/io/rest/core/internal/thing/ThingResource.java
```diff
@@ -522,16 +525,41 @@ public Response updateFirmware(
...
+    @Path("/{thingUID}/firmwares")
+    @Produces(MediaType.APPLICATION_JSON)
+    @ApiOperation(value = "Get all available firmwares for provided thing UID", response = StrippedThingTypeDTO.class, responseContainer = "Set")
+    @ApiResponses(value = { @ApiResponse(code = 200, message = "OK"),
+            @ApiResponse(code = 204, message = "No firmwares found.") })
+    public Response getFirmwares(@PathParam("thingUID") @ApiParam(value = "thingUID") String thingUID,
+            @HeaderParam(HttpHeaders.ACCEPT_LANGUAGE) @ApiParam(value = HttpHeaders.ACCEPT_LANGUAGE) String language) {
+        ThingUID aThingUID = new ThingUID(thingUID);
+        Thing thing = thingRegistry.get(aThingUID);
+        Collection<Firmware> firmwares = firmwareRegistry.getFirmwares(thing, LocaleUtil.getLocale(language));
```
sjka: Don't you want to return a `404` in case`thing == null`? At least in the console command you're checking that and the `FirmwareRegistry` doesn't like `null`-Things...

bundles/io/org.eclipse.smarthome.io.rest.core/src/main/java/org/eclipse/smarthome/io/rest/core/internal/thing/ThingTypeResource.java
```diff
@@ -164,25 +160,6 @@ public Response getByUID(@PathParam("thingTypeUID") @ApiParam(value = "thingType
         }
     }
 
-    @GET
-    @Path("/{thingTypeUID}/firmwares")
-    @Produces(MediaType.APPLICATION_JSON)
-    @ApiOperation(value = "Get all available firmwares for provided thingType", response = StrippedThingTypeDTO.class, responseContainer = "Set")
-    @ApiResponses(value = { @ApiResponse(code = 200, message = "OK"),
-            @ApiResponse(code = 204, message = "No firmwares found.") })
-    public Response getFirmwares(@PathParam("thingTypeUID") @ApiParam(value = "thingTypeUID") String thingTypeUID,
```
sjka: With the drop of this method you can also remove the entire dependency to the `FirmwareRegistry`, can't you?

bundles/core/org.eclipse.smarthome.core.thing/src/main/java/org/eclipse/smarthome/core/thing/binding/firmware/Firmware.java
```diff
@@ -231,7 +233,7 @@ public String getMd5Hash() {
                 try (DigestInputStream dis = new DigestInputStream(inputStream, md)) {
                     bytes = IOUtils.toByteArray(dis);
                 } catch (IOException ioEx) {
-                    LOGGER.error("Cannot read firmware with UID {}.", uid, ioEx);
+                    LOGGER.error("Cannot read firmware {}.", this, ioEx);
```
sjka: Btw, why is this logger static? Could you please make it non-static & lower-case?

bundles/core/org.eclipse.smarthome.core.thing/src/main/java/org/eclipse/smarthome/core/thing/firmware/FirmwareRegistry.java
```diff
@@ -43,6 +43,7 @@
  * <ul>
  *
  * @author Thomas Höfer - Initial contribution
+ * @author Dimitar Ivanov - The firmwares are provided by thing and version
```
sjka: While refactoring it anyway, could you please turn this class into an interface and move the implementation into a class named `org.eclipse.smarthome.core.thing.internal.firmware.FirmwareRegistryImpl`?

bundles/core/org.eclipse.smarthome.core.thing/src/main/java/org/eclipse/smarthome/core/thing/firmware/FirmwareUpdateService.java
```diff
@@ -67,6 +66,7 @@
  * @author Thomas Höfer - Initial contribution
  * @author Dimitar Ivanov - update and cancel operations are run with different safe caller identifiers in order to
  *         execute asynchronously
+ * @author Dimitar Ivanov - Firmware update is done for thing
```
sjka: same as with the `FirmwareRegistry`: Could you please turn this class into an interface and move the implementation to `org.eclipse.smarthome.core.thing.internal.firmware.FirmwareUpdateServiceImpl`?

bundles/core/org.eclipse.smarthome.core.thing/src/main/java/org/eclipse/smarthome/core/thing/binding/firmware/Firmware.java
```diff
@@ -376,7 +386,8 @@ private boolean isInt(String s) {
      */
     public static final class Builder {
```
sjka: Please factor this out into a dedicated `FirmwareBuilder` class, so that it gets symmetric to all the other builders we have.

bundles/core/org.eclipse.smarthome.core.thing/src/main/java/org/eclipse/smarthome/core/thing/binding/firmware/Firmware.java
```diff
@@ -67,6 +68,7 @@
  * device is required afterwards) so that {@link FirmwareUpdateHandler}s can handle this information accordingly.
  *
  * @author Thomas Höfer - Initial contribution
+ * @author Dimitar Ivanov - FirmwareUID is replaced by ThingTypeUID and firmware version
```
sjka: Please turn this class into an Interface and move the implementation into `org.eclipse.smarthome.core.thing.internal.firmware.FirmwareImpl`

bundles/core/org.eclipse.smarthome.core.thing/src/main/java/org/eclipse/smarthome/core/thing/firmware/ProgressCallbackImpl.java
```diff
@@ -34,6 +34,7 @@
  *
  * @author Thomas Höfer - Initial contribution
  * @author Christoph Knauf - Introduced pending, canceled, update and InternalState
+ * @author Dimitar Ivanov - Callback contains firmware domain object
```
sjka: Despite it is package-private, could you please move it to `org.eclipse.smarthome.core.thing.internal.firmware.ProgressCallbackImpl`?

sjka:
@tomhoefer would you mind having a look too?

bundles/core/org.eclipse.smarthome.core.thing/src/main/java/org/eclipse/smarthome/core/thing/binding/firmware/Firmware.java
tomhoefer: Originally I wanted to avoid having a new logger created for each new instance of this class - especially due to potential performance issues on embedded devices.

bundles/core/org.eclipse.smarthome.core.thing/src/main/java/org/eclipse/smarthome/core/thing/firmware/FirmwareRegistry.java
```diff
@@ -54,107 +55,113 @@
...
     /**
-     * Returns the collection of available firmwares for the given thing type UID using the locale provided by the
+     * Returns the collection of available firmwares for the given thing using the locale provided by the
      * {@link LocaleProvider}. The collection is sorted in descending order, i.e. the latest firmware will be the first
      * element in the collection.
      *
-     * @param thingTypeUID the thing type UID for which the firmwares are to be provided (not null)
-     * @return the collection of available firmwares for the given thing type UID (not null)
-     * @throws NullPointerException if given thing type UID is null
+     * @param thing - the thing for which the firmwares are to be retrieved (not null)
```
tomhoefer: Is there a ```-```to be removed?

bundles/core/org.eclipse.smarthome.core.thing.test/src/test/groovy/org/eclipse/smarthome/core/thing/firmware/FirmwareRegistryOSGiTest.groovy
```diff
@@ -325,29 +324,29 @@ final class FirmwareRegistryOSGiTest extends OSGiTest {
...
+    void 'assert that firmware properties are immutable'(){
+        def fw = firmwareRegistry.getFirmware(thing1, FW112_EN.getVersion())
         assertThat fw, is(notNullValue())
         fw.getProperties().put("test", null)
     }
 
     @Test
-    void 'void assert that registry returns null for unknown thing type uid and getLatestFirmware operation()'() {
-        def firmware = firmwareRegistry.getLatestFirmware(UNKNOWN_THING_TYPE_UID);
+    void'void assert that registry returns null for unknown thing type uid and getLatestFirmware operation'() {
```
tomhoefer: I think we can remove the void term in the operation name and add an additional blank between ```void 'assert...```

tomhoefer:
I have only a few comments

dstivanov:
@sjka 
> As it is API breaking anyway, I would prefer moving the implementation classes into an internal package
I would consider moving them, didn't come to my mind.

> Additionally, is there anything preventing you from using the @NonNullByDefault annotations on class-level?
There is no impediments to do it so I will apply it to the refactored code. Good point!

I will also adapt the code to your comments.

triller-telekom:
@dstivanov Before adding the null annotations and replacing all `Objects.requireNonNull` with an annotation, please have a look at PR #5360 where I explain when we still want to have a runtime check such as  `Objects.requireNonNull`.

bundles/core/org.eclipse.smarthome.core.thing/src/main/java/org/eclipse/smarthome/core/thing/binding/firmware/Firmware.java
maggu2810: See also:
* https://github.com/jupnp/jupnp/pull/68#pullrequestreview-3429591
* https://stackoverflow.com/a/10456613

bundles/io/org.eclipse.smarthome.io.rest.core/src/main/java/org/eclipse/smarthome/io/rest/core/internal/thing/ThingResource.java
dstivanov: Yes, there is even a private method that I can reuse.

dstivanov: And will also check the firmware version for null or empty as previously this check was done for the FirmwareUID.

bundles/io/org.eclipse.smarthome.io.rest.core/src/main/java/org/eclipse/smarthome/io/rest/core/internal/thing/ThingTypeResource.java
dstivanov: Yes, you are right. Thanks for noticing that.

bundles/core/org.eclipse.smarthome.core.thing/src/main/java/org/eclipse/smarthome/core/thing/binding/firmware/Firmware.java
dstivanov: I will fix it as @sjka suggested, thak you for the participation into the discussion!

bundles/core/org.eclipse.smarthome.core.thing/src/main/java/org/eclipse/smarthome/core/thing/firmware/FirmwareRegistry.java
dstivanov: Yes, all these issues were reported by the SCAT tool of ESH, so I will fix them with the next update on the PR.

dstivanov:
@sjka @triller-telekom @maggu2810 @tomhoefer I have addressed your review comments. Can you please take a look at the recent version of the PR ?

bundles/core/org.eclipse.smarthome.core.thing.test/src/test/java/org/eclipse/smarthome/core/thing/internal/firmware/FirmwareUpdateServiceTest.java
```diff
@@ -68,6 +76,7 @@
  * @author Thomas Höfer - Initial contribution
  * @author Simon Kaufmann - converted to standalone Java tests
  * @author Dimitar Ivanov - added a test for valid cancel execution during firmware update
+ * @author Dimitar Ivanov - replaced Firmware UID with thing UID and firmware version
```
tomhoefer: I am not sure - is it common that we add the author twice?

bundles/core/org.eclipse.smarthome.core.thing.test/src/test/java/org/eclipse/smarthome/core/thing/internal/firmware/FirmwareUpdateServiceTest.java
```diff
@@ -118,25 +118,34 @@
...
 
-        firmwareUpdateService = new FirmwareUpdateService();
+        firmwareUpdateService = new FirmwareUpdateServiceImpl();
 
         safeCaller = getService(SafeCaller.class);
         assertNotNull(safeCaller);
 
         firmwareUpdateService.setSafeCaller(safeCaller);
 
+        firmwareUpdateService.setFirmwareRegistry(mockFirmwareRegistry);
```
tomhoefer: since it is an osgi test why dont we register the mock as a service?

bundles/core/org.eclipse.smarthome.core.thing.test/src/test/java/org/eclipse/smarthome/core/thing/internal/firmware/FirmwareUpdateServiceTest.java
```diff
@@ -831,13 +920,14 @@ private void updateInvalidConfigAndAssert(int period, int delay, TimeUnit timeUn
 
     private void assertResultInfoEvent(ThingUID thingUID, Firmware firmware, String messageKey, Locale locale,
             String text, int expectedEventCount) {
-        firmwareUpdateService.removeFirmwareUpdateHandler(handler1);
-        firmwareUpdateService.addFirmwareUpdateHandler(handler1);
+        // FIXME See why remove and add
```
tomhoefer: I guess this should be checked

bundles/core/org.eclipse.smarthome.core.thing.test/src/test/java/org/eclipse/smarthome/core/thing/internal/firmware/FirmwareUpdateServiceTest.java
```diff
@@ -850,8 +940,9 @@ private void assertResultInfoEvent(ThingUID thingUID, Firmware firmware, String
 
     private void assertCancellationMessage(String messageKey, String expectedEnglishMessage, Locale locale,
             int expectedEventCount) {
-        firmwareUpdateService.removeFirmwareUpdateHandler(handler3);
-        firmwareUpdateService.addFirmwareUpdateHandler(handler3);
+        // FIXME See why remove and add
```
tomhoefer: here as well - fix me?

bundles/core/org.eclipse.smarthome.core.thing/src/main/java/org/eclipse/smarthome/core/thing/binding/firmware/Firmware.java
```diff
@@ -67,150 +57,89 @@
...
-        this.internalPrerequisiteVersion = this.prerequisiteVersion != null ? new Version(this.prerequisiteVersion)
-                : null;
-    }
-
     /**
-     * Returns the UID of the firmware.
+     * Returns the thing type UID, that this firmware is associated with.
      *
-     * @return the UID of the firmware (not null)
+     * @return the thing type UID
```
tomhoefer: i guess this is never null, isnt it?

bundles/core/org.eclipse.smarthome.core.thing/src/main/java/org/eclipse/smarthome/core/thing/binding/firmware/FirmwareBuilder.java
```diff
@@ -0,0 +1,227 @@
...
+    private boolean modelRestricted;
+    private String description;
+    private String prerequisiteVersion;
+    private String changelog;
+    private URL onlineChangelog;
+    private transient InputStream inputStream;
+    private String md5Hash;
+    private Map<String, String> properties;
+
+    public ThingTypeUID getThingTypeUID() {
```
tomhoefer: why does the builder provide the getter operations?

bundles/core/org.eclipse.smarthome.core.thing/src/main/java/org/eclipse/smarthome/core/thing/firmware/FirmwareStatusInfo.java
```diff
@@ -37,76 +45,104 @@ protected FirmwareStatusInfo() {
...
+        Objects.requireNonNull(thingUID, "Thing UID must not be null.");
+        this.thingUID = thingUID;
+
         Objects.requireNonNull(firmwareStatus, "Firmware status must not be null.");
         this.firmwareStatus = firmwareStatus;
-        this.updatableFirmwareUID = updatableFirmwareUID;
+
+        // Keep only the feasible information from the firmware to pass it through event later
+        if (updatableFirmware != null) {
+            this.thingTypeUID = updatableFirmware.getThingTypeUID();
```
tomhoefer: I wonder why we need the thing type uid here

bundles/core/org.eclipse.smarthome.core.thing/src/main/java/org/eclipse/smarthome/core/thing/firmware/FirmwareStatusInfo.java
```diff
@@ -37,76 +45,104 @@ protected FirmwareStatusInfo() {
...
-     * @return the firmware UID (only set if firmware status is {@link FirmwareStatus#UPDATE_EXECUTABLE})
+     * @return the thing type UID (only set if firmware status is {@link FirmwareStatus#UPDATE_EXECUTABLE})
      */
-    public FirmwareUID getUpdatableFirmwareUID() {
-        return updatableFirmwareUID;
+    public ThingTypeUID getUpdatableFirmwareThingTypeUID() {
+        return this.thingTypeUID;
+    }
+
+    public ThingUID getThingUID() {
```
tomhoefer: Should we add some javadoc here?

bundles/core/org.eclipse.smarthome.core.thing/src/main/java/org/eclipse/smarthome/core/thing/firmware/FirmwareUpdateProgressInfo.java
```diff
@@ -72,37 +72,81 @@ protected FirmwareUpdateProgressInfo() {
...
+
+    /**
+     * Returns the Thing type UID of the firmware that is updated.
      *
-     * @return the UID of the firmware that is updated (not null)
+     * @return the Thing type UID ofthe firmware that is updated (not null)
      */
-    public FirmwareUID getFirmwareUID() {
-        return firmwareUID;
+    public ThingTypeUID getThingTypeUID() {
```
tomhoefer: do we need to expose this information?

bundles/core/org.eclipse.smarthome.core.thing/src/main/java/org/eclipse/smarthome/core/thing/internal/firmware/FirmwareUpdateServiceImpl.java
```diff
@@ -0,0 +1,556 @@
...
+
+    @Deactivate
+    protected void deactivate() {
+        cancelFirmwareUpdateStatusInfoJob();
+        firmwareStatusInfoMap.clear();
+        progressCallbackMap.clear();
+    }
+
+    /**
+     * Returns the {@link FirmwareStatusInfo} for the thing having the given thing UID.
```
tomhoefer: maybe we can remove the javadoc here

bundles/core/org.eclipse.smarthome.core.thing/src/main/java/org/eclipse/smarthome/core/thing/internal/firmware/FirmwareUpdateServiceImpl.java
```diff
@@ -0,0 +1,556 @@
...
+        FirmwareStatusInfo firmwareStatusInfo = getFirmwareStatusInfo(firmwareUpdateHandler, latestFirmware);
+
+        processFirmwareStatusInfo(firmwareUpdateHandler, firmwareStatusInfo, latestFirmware);
+
+        return firmwareStatusInfo;
+    }
+
+    /**
+     * Updates the firmware of the thing having the given thing UID by invoking the operation
+     * {@link FirmwareUpdateHandler#updateFirmware(Firmware, ProgressCallback)} of the thing´s firmware update handler.
```
tomhoefer: maybe we can remove the javadoc here, too

bundles/core/org.eclipse.smarthome.core.thing/src/main/java/org/eclipse/smarthome/core/thing/internal/firmware/FirmwareUpdateServiceImpl.java
```diff
@@ -0,0 +1,556 @@
...
+                            "Unexpected exception occurred for firmware update of thing with UID {} and firmware {}.",
+                            thingUID, firmware, e.getCause());
+                    progressCallback.failedInternal("unexpected-handler-error");
+                }).build().updateFirmware(firmware, progressCallback);
+    }
+
+    /**
+     * Cancels the firmware update of the thing having the given thing UID by invoking the operation
+     * {@link FirmwareUpdateHandler#cancel()} of the thing´s firmware update handler.
+     *
```
tomhoefer: maybe we can remove the javadoc here, too

tomhoefer:
just some minor comments - lgtm in general

bundles/core/org.eclipse.smarthome.core.thing/src/main/java/org/eclipse/smarthome/core/thing/firmware/FirmwareStatusInfo.java
```diff
@@ -14,21 +14,29 @@
...
  * The {@link FirmwareStatusInfo} represents the {@link FirmwareStatus} of a {@link Thing}. If the firmware status is
- * {@link FirmwareStatus#UPDATE_EXECUTABLE} then the information object will also provide the {@link FirmwareUID} of the
- * latest updatable firmware for the thing.
+ * {@link FirmwareStatus#UPDATE_EXECUTABLE} then the information object will also provide the thing type UID and the
+ * version of the latest updatable firmware for the thing.
  *
  * @author Thomas Höfer - Initial contribution
+ * @author Dimitar Ivanov - Consolidated all the needed information for firmware status events
  */
 public final class FirmwareStatusInfo {
```
sjka: `@NonNullByDefault`

bundles/core/org.eclipse.smarthome.core.thing/src/main/java/org/eclipse/smarthome/core/thing/firmware/FirmwareUpdateProgressInfo.java
```diff
@@ -15,18 +15,25 @@
...
 import org.eclipse.smarthome.core.thing.binding.firmware.ProgressStep;
 
 /**
  * The {@link FirmwareUpdateProgressInfo} represents the progress indicator for a firmware update.
  *
  * @author Thomas Höfer - Initial contribution
  * @author Christoph Knauf - Added progress and pending
+ * @author Dimitar Ivanov - Consolidated all the needed information for firmware status events
  */
 public final class FirmwareUpdateProgressInfo {
```
sjka: `@NonNullByDefault`

bundles/core/org.eclipse.smarthome.core.thing/src/main/java/org/eclipse/smarthome/core/thing/firmware/FirmwareUpdateResultInfo.java
```diff
@@ -14,34 +14,34 @@
...
+import org.eclipse.jdt.annotation.NonNull;
+import org.eclipse.smarthome.core.thing.ThingUID;
+
 /**
  * The {@link FirmwareUpdateResultInfo} contains information about the result of a firmware update.
  *
  * @author Thomas Höfer - Initial contribution
+ * @author Dimitar Ivanov - Consolidated all the needed information for firmware status events
  */
 public final class FirmwareUpdateResultInfo {
```
sjka: `@NonNullByDefault`

bundles/core/org.eclipse.smarthome.core.thing/src/main/java/org/eclipse/smarthome/core/thing/firmware/FirmwareUpdateResultInfo.java
```diff
@@ -54,6 +54,23 @@ protected FirmwareUpdateResultInfo() {
...
+     *
+     * @param thingUID thingUID of the thing being updated
+     * @param result the result of the firmware update (must not be null)
+     * @param errorMessage the error message in case of result is {@link FirmwareUpdateResult#ERROR} (must not be null
+     *            or empty for erroneous firmware updates; ignored for successful firmware updates)
+     * @return FirmwareUpdateResultInfo (not null)
+     * @throws NullPointerException if the thingUID or firmware update result is null
+     * @throws IllegalArgumentException if error message is null or empty for erroneous firmware updates
+     */
+    @NonNull
```
sjka: see https://www.eclipse.org/smarthome/documentation/development/conventions.html

bundles/core/org.eclipse.smarthome.core.thing/src/main/java/org/eclipse/smarthome/core/thing/firmware/FirmwareUpdateService.java
```diff
@@ -12,152 +12,23 @@
...
-        }
-    }
-
-    @Deactivate
-    protected void deactivate() {
-        cancelFirmwareUpdateStatusInfoJob();
-        firmwareStatusInfoMap.clear();
-        progressCallbackMap.clear();
-    }
+public interface FirmwareUpdateService {
```
sjka: `@NonNullByDefault`

bundles/core/org.eclipse.smarthome.core.thing/src/main/java/org/eclipse/smarthome/core/thing/internal/firmware/FirmwareImpl.java
```diff
@@ -0,0 +1,517 @@
...
+ *
+ * <p>
+ * A {@link Firmware} consists of various meta information like a version, a vendor or a description. Additionally
+ * {@link FirmwareProvider}s can specify further meta information in form of properties (e.g. a factory reset of the
+ * device is required afterwards) so that {@link FirmwareUpdateHandler}s can handle this information accordingly.
+ *
+ * @author Thomas Höfer - Initial contribution
+ * @author Dimitar Ivanov - FirmwareUID is replaced by ThingTypeUID and firmware version
+ */
+public final class FirmwareImpl implements Firmware {
```
sjka: `@NonNullByDefault`

bundles/core/org.eclipse.smarthome.core.thing/src/main/java/org/eclipse/smarthome/core/thing/internal/firmware/FirmwareRegistryImpl.java
```diff
@@ -0,0 +1,138 @@
...
+import org.slf4j.LoggerFactory;
+
+/**
+ * Default implementation of {@link FirmwareRegistry}.
+ *
+ * @author Thomas Höfer - Initial contribution
+ * @author Dimitar Ivanov - The firmwares are provided by thing and version
+ */
+@Component(immediate = true, service = FirmwareRegistry.class)
+public final class FirmwareRegistryImpl implements FirmwareRegistry {
```
sjka: `@NonNullByDefault`

bundles/core/org.eclipse.smarthome.core.thing/src/main/java/org/eclipse/smarthome/core/thing/internal/firmware/FirmwareUpdateServiceImpl.java
```diff
@@ -0,0 +1,556 @@
...
+/**
+ * Default implementation of {@link FirmwareUpdateService}.
+ *
+ * @author Thomas Höfer - Initial contribution
+ * @author Dimitar Ivanov - update and cancel operations are run with different safe caller identifiers in order to
+ *         execute asynchronously
+ * @author Dimitar Ivanov - Firmware update is done for thing
+ */
+@Component(immediate = true, service = { EventSubscriber.class, FirmwareUpdateService.class })
+public final class FirmwareUpdateServiceImpl implements FirmwareUpdateService, EventSubscriber {
```
sjka: `@NonNullByDefault`

bundles/core/org.eclipse.smarthome.core.thing/src/main/java/org/eclipse/smarthome/core/thing/internal/firmware/FirmwareUpdateServiceImpl.java
```diff
@@ -0,0 +1,556 @@
...
+
+    private static final String THREAD_POOL_NAME = FirmwareUpdateServiceImpl.class.getSimpleName();
+    private static final Set<String> SUPPORTED_TIME_UNITS = ImmutableSet.of(TimeUnit.SECONDS.name(),
```
sjka: `Collections.unmodifiableSet(Stream.of(...).toSet())` - see https://www.eclipse.org/smarthome/documentation/development/guidelines.html#c-language-levels-and-libraries

bundles/core/org.eclipse.smarthome.core.thing/src/main/java/org/eclipse/smarthome/core/thing/binding/firmware/FirmwareBuilder.java
```diff
@@ -0,0 +1,227 @@
...
+    public Map<String, String> getProperties() {
+        return properties;
+    }
+
+    /**
+     * Creates a new builder.
+     *
+     * @param thingTypeUID the thing type UID that is associated with this firmware (must not be null)
+     * @param firmwareVersion the version of the firmware to be created (must not be null)
+     * @throws NullPointerException if given thing type UID is null
```
sjka: this is stating the obvious - non need to document NPEs explicitly

bundles/core/org.eclipse.smarthome.core.thing/src/main/java/org/eclipse/smarthome/core/thing/binding/firmware/FirmwareBuilder.java
```diff
@@ -0,0 +1,227 @@
...
+     * @throws IllegalArgumentException if given firmware version is null or empty
+     */
+    public FirmwareBuilder(ThingTypeUID thingTypeUID, String firmwareVersion) {
+        Objects.requireNonNull(thingTypeUID, "Thing type UID must not be null.");
```
sjka: isn't it somewhat inconsistent to throw an NPE in case on argument is `null` and throw an IAE in case the other is `null`?

I won't repeat it, but this commend basically applies to the whole firmware area - it's pretty inconsistent in this respect. The rest of ESH avoids throwing NPEs "on purpose" - usually in such cases IAE are thrown in order to distinguish and "accidental programming bug" from a "hey, gotcha, you've been passing crap"-situation.

sjka: PS: I of course know that it was like that previously - I just reckon this would be a good point in time to streamline this.

bundles/core/org.eclipse.smarthome.core.thing/src/main/java/org/eclipse/smarthome/core/thing/binding/firmware/FirmwareBuilder.java
dstivanov: I was thinking of another solution now - what about changing the signature to 
```
public FirmwareBuilder(@NonNull ThingTypeUID thingTypeUID, String firmwareVersion) {
```

bundles/core/org.eclipse.smarthome.core.thing.test/src/test/java/org/eclipse/smarthome/core/thing/internal/firmware/FirmwareUpdateServiceTest.java
dstivanov: I checked and you are right - the contributions are comma separated instead of on new line. I will change this.

bundles/core/org.eclipse.smarthome.core.thing.test/src/test/java/org/eclipse/smarthome/core/thing/internal/firmware/FirmwareUpdateServiceTest.java
dstivanov: I found this test with part of the services mocked and set manually and the others were injected. As the most important services were mocked I decided to continue the test like this and mock as much as I can. 
I agree that the better approach would be to register the mocks as services, but this was not possible, because of the large number of mostly static dependencies, which would have introduced more sophisticated operations on the Firmware update service component.

At the end, now I am mocking all the dependencies and setting them manually (instead of the SafeCaller that can be mocked as well, but will generate API restriction warning and since we are still JavaOsgiTest it does not make sense), but there is a `FrameworkUtil.getBundle(clazz)` call in the `ProgressCallbackImpl`, which stops from doing JavaTest instead of JavaOsgiTest.

I would be happy to discuss with @tomhoefer and @sjka for a posible better approach.

P.S. I did some small adjustments on the test, so please let me update the PR before starting the discussion.

bundles/core/org.eclipse.smarthome.core.thing.test/src/test/java/org/eclipse/smarthome/core/thing/internal/firmware/FirmwareUpdateServiceTest.java
dstivanov: I have rechecked that and it seems to be there for legacy reasons. The tests are behaving well without these statements so I will delete them.

bundles/core/org.eclipse.smarthome.core.thing.test/src/test/java/org/eclipse/smarthome/core/thing/internal/firmware/FirmwareUpdateServiceTest.java
dstivanov: Commented above.

bundles/core/org.eclipse.smarthome.core.thing/src/main/java/org/eclipse/smarthome/core/thing/binding/firmware/Firmware.java
dstivanov: Yes, it is not null. I will apply the @NonNullByDefault annotations according  to the coding guidelines.

bundles/core/org.eclipse.smarthome.core.thing/src/main/java/org/eclipse/smarthome/core/thing/binding/firmware/FirmwareBuilder.java
dstivanov: This is commented below by Simon so in the next adaption of the PR it will be fixed. Thank you for noticing that.

bundles/core/org.eclipse.smarthome.core.thing/src/main/java/org/eclipse/smarthome/core/thing/firmware/FirmwareStatusInfo.java
dstivanov: In general we can identify the thing type UID by the thing UID so I agree that we can remove the thing type uid from here.

bundles/core/org.eclipse.smarthome.core.thing/src/main/java/org/eclipse/smarthome/core/thing/firmware/FirmwareStatusInfo.java
dstivanov: For sure! I will add javadoc here.

bundles/core/org.eclipse.smarthome.core.thing/src/main/java/org/eclipse/smarthome/core/thing/firmware/FirmwareUpdateProgressInfo.java
dstivanov: The same as above, the thing type UID can be found by the thing UID so I will remove it from here as well

bundles/core/org.eclipse.smarthome.core.thing/src/main/java/org/eclipse/smarthome/core/thing/internal/firmware/FirmwareUpdateServiceImpl.java
dstivanov: Yes, I found all such places (where I copy pasted javadoc during the interface separation) and cleaned the odd javadoc.

bundles/core/org.eclipse.smarthome.core.thing/src/main/java/org/eclipse/smarthome/core/thing/binding/firmware/FirmwareBuilder.java
```diff
@@ -0,0 +1,227 @@
...
+
+    /**
+     * Creates a new builder.
+     *
+     * @param thingTypeUID the thing type UID that is associated with this firmware (must not be null)
+     * @param firmwareVersion the version of the firmware to be created (must not be null)
+     * @throws NullPointerException if given thing type UID is null
+     * @throws IllegalArgumentException if given firmware version is null or empty
+     */
+    public FirmwareBuilder(ThingTypeUID thingTypeUID, String firmwareVersion) {
```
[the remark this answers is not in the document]
dstivanov: Thanks for noticing that! i refactored it to be the same as the other builders. Will be visible in the next PR update.

bundles/core/org.eclipse.smarthome.core.thing/src/main/java/org/eclipse/smarthome/core/thing/firmware/FirmwareUpdateResultInfo.java
dstivanov: Thanks! Will be cleared out in the next PR update.

bundles/core/org.eclipse.smarthome.core.thing/src/main/java/org/eclipse/smarthome/core/thing/internal/firmware/FirmwareUpdateServiceImpl.java
dstivanov: Thanks for reminding me to use standard java 8 instead of guava.

bundles/core/org.eclipse.smarthome.core.thing/src/main/java/org/eclipse/smarthome/core/thing/internal/firmware/FirmwareRegistryImpl.java
```diff
@@ -0,0 +1,158 @@
...
+ */
+@Component(immediate = true, service = FirmwareRegistry.class)
+@NonNullByDefault
+public final class FirmwareRegistryImpl implements FirmwareRegistry {
+
+    private final Logger logger = LoggerFactory.getLogger(FirmwareRegistryImpl.class);
+
+    private final List<FirmwareProvider> firmwareProviders = new CopyOnWriteArrayList<>();
+
+    private @Nullable LocaleProvider localeProvider;
```
hsudbrock: As the `LocaleProvider` is a static mandatory reference, you could annotate it as `@NonNullByDefault({})` (instead of `@Nullable`), as indicated here: https://www.eclipse.org/smarthome/documentation/development/conventions.html#null-annotations

dstivanov: You are absolutely right! I will also add `@NonNullByDefault({})` into the FirmwareUpdateServiceImpl, where I have several static references as well.

dstivanov:
@sjka I have addressed the review comments and adapted the code to the recent changes. Can you please take another look ?

dstivanov:
Adapted the PR due to some conflicts.

sjka:
Great, thanks!

I will file a CQ for this PR as it exceeds 1000 LoC. Please don't make any changes to it anymore.

sjka:
FTR: [CQ 16200](https://dev.eclipse.org/ipzilla/show_bug.cgi?id=16200)

kaikreuzer:
CQ has been approved!

Issue #5344, opened 2018-04-03
ThingStatusDetail.GONE is listed as ThingStatus in the Things Documentation


Xrygramming:
The [ThingStatusDetail.GONE](https://github.com/eclipse/smarthome/blob/68ce82a5d61b7576fec47aaf82cb5336bd8fd2cc/bundles/core/org.eclipse.smarthome.core.thing/src/main/java/org/eclipse/smarthome/core/thing/ThingStatusDetail.java#L38) is incorrectly listed in the [Things documentation](https://github.com/eclipse/smarthome/blob/master/docs/documentation/concepts/things.md). It should be listed as status detail instead of a status.

PR #5345, opened 2018-04-03
Adapted Things Documentation so that the ThingStatusDetail.GONE is listed correctly

Xrygramming:
In this PR, we changed the Things documentation because the ThingStatusDetail.GONE was incorrectly listed as a ThingStatus.

See [#5344](https://github.com/eclipse/smarthome/issues/5344).

Signed-off-by: Hannes Hofmann <hannes.hofmann@telekom.de>


kaikreuzer:
FTR: I have manually checked the ECA, which is valid.

PR #5336, opened 2018-04-02
add QuantityType<Density> and QuantityType<Energy> and units for that

J-N-K:
fixes #5333
fixes #5341 

Signed-off-by: Jan N. Klug <jan.n.klug@rub.de>

PR #5346, opened 2018-04-03
adds kW·h as Unit

### LINKED ISSUE #5341: kWh are not recognized as Energy values
The recently introduced unit support has an "Energy" dimension defined, which uses by default the SI unit "J" (Joule).
In practice, the alternative "Wh" are used much more frequently and thus I would expect this unit to be usable for items with `Energy` dimension.

I defined an Item like this:
```
Number:Energy Energie			"Energie [%.1f kWh]"
```
Sending a value on the console
```
smarthome send Energie 123.456
```
results in
```
2018-04-03 13:27:05.491 [DEBUG] [.e.s.core.types.util.UnitUtils:108  ] - Unknown unit from pattern: kWh
```
and no unit being assigned to the item state.

As energy is likely the most common use case besides weather for units, I'd consider this a rather critical issue that needs to be addressed quickly. See also https://github.com/openhab/openhab2-addons/pull/2167#discussion_r178578851.

J-N-K:
fixes #5341

### DIFF
--- a/bundles/core/org.eclipse.smarthome.core.test/src/test/java/org/eclipse/smarthome/core/library/types/QuantityTypeTest.java
+++ b/bundles/core/org.eclipse.smarthome.core.test/src/test/java/org/eclipse/smarthome/core/library/types/QuantityTypeTest.java
@@ -20,6 +20,7 @@ import java.math.BigDecimal;
 
 import javax.measure.Unit;
 import javax.measure.quantity.Dimensionless;
+import javax.measure.quantity.Energy;
 import javax.measure.quantity.Length;
 import javax.measure.quantity.Temperature;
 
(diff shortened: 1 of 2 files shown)


J-N-K:
Travis fails in unrelkated JSON storage test.

J-N-K:
Retriggering build

J-N-K:
I don‘t understand why so many builds fail lately. This is clearly unrelated, unfortunately before the core tests.

maggu2810:
I assume this error is raised because of #5318.
NPEs in tests are not catched anymore.
IMHO this change is correct but it identify that some tests need more love (e.g. the related one seems to be very timing dependent).

I assume this one (the JSON storage one) will be fixed by #5353

J-N-K:
retriggering build after tests are finally fixed

J-N-K:
done. and integrated in #5336 to avoid merge conflicts

Issue #3320, opened 2017-04-26
Extend the Marketplace extension service to support the "UI" package type


ghys:
Hi,
Last week I added the ability to provision custom widgets to the HABPanel user interface for openHAB with OSGi bundles (registering additional resources and templates during the bundle's activation).
Today a first widget in this format was reviewed and accepted in the Eclipse IoT marketplace: 
https://marketplace.eclipse.org/content/gauge-widget-habpanel
which incidentally is the first having "UI" as a package type.

Since the `MarketplaceExtension` and `MarketplaceExtensionService`  only support bindings and rule templates at the moment. I would therefore like to propose - if the time and approach is right - to extend them to support the "UI" package type. An implementation of `MarketplaceExtensionHandler` (similar to `BindingExtensionHandler`) to install and uninstall those bundles would also be needed.

(I'm aware HABPanel is an openHAB project and not all ESH-based products include it; the bundle will simply be useless then, but I understand it might confuse users if they see it appearing in their extension list without the "parent" software available).

Looking forward to your feedback, thanks!

kaikreuzer:
Thanks @ghys! The custom widgets are a very nice feature of HABPanel.
I agree that the Eclipse IoT Marketplace looks like a good match for distributing such custom widgets.

A few things to maybe discuss before we move forward (additionally to [my comments here](https://community.openhab.org/t/new-feature-globally-provisioned-widgets-i-e-with-osgi-bundles/26994/5?u=kai)):
- So far the "UI" category was meant for "complete" UIs and not extensions for a certain UI. I am also wondering if it makes sense to show these widgets in the "UI" tab of openHAB "Add-ons" besides all the "real" UIs or if there shouldn't be a dedicated tab for the custom widgets.
- As you point out, the widgets have dependencies to openHAB and thus are not "ESH-compatible". I don't think it is a must that any listing here can be used by any ESH-based solution, but we should nonetheless think about how we could potentially remove dependencies to a specific solution and thus make it in general more re-usable.
- As HABPanel is still under heavy development (I assume you do not consider the API as stable?), I wonder how to deal with the compatibility and the version range that the widgets define as being compatible. We have a similar problem for the other listing types as well, but I postponed the solution for the moment as I hoped that it won't be relevant too soon - for the custom widgets, I am not so sure about that anymore, though.

ghys:
[...]

PR #5349, opened 2018-04-03
Support installation of Voice extensions from the IoT Marketplace.

jsichi:
This PR is for the discussion in 

https://community.openhab.org/t/moving-beyond-bindings-in-the-iot-addon-marketplace/42194

I tried to make the code a bit more data-driven (via enum PackageType) to make it easier to keep adding more extension types.  For the moment, I only added Voice and UI, but I can add the others (actions, persistence, transformations) if desired.

I renamed BindingExtensionHandler to BundleExtensionHandler for generality.  I also arranged for migration of the old installedBindingsMap.csv to the new installedBundlesMap.csv.  This might be unnecessary since I think this file gets discarded during an upgrade anyway, but it seems like the safest thing to do.

I tested by enabling plugin org.eclipse.smarthome.extensionservice.marketplace in my Eclipse workspace configuration.  For UI, I was able to test https://marketplace.eclipse.org/content/gauge-widget-habpanel.  With this change, it now shows up in the UI tab under extensions in the Paper UI, and I was able to successfully install+uninstall it.

For Voice, I am not able to do full testing since the Marketplace server does not currently allow for this package type when publishing an addon.  So I'll need someone's help to make that happen.

Bug: https://github.com/eclipse/smarthome/issues/3320

Signed-off-by: John Sichi <jsichi@gmail.com>

### DIFF
--- a/extensions/extensionservice/org.eclipse.smarthome.extensionservice.marketplace.test/src/test/java/org/eclipse/smarthome/extensionservice/marketplace/internal/MarketplaceExtensionServiceTest.java
+++ b/extensions/extensionservice/org.eclipse.smarthome.extensionservice.marketplace.test/src/test/java/org/eclipse/smarthome/extensionservice/marketplace/internal/MarketplaceExtensionServiceTest.java
@@ -59,6 +59,22 @@ public class MarketplaceExtensionServiceTest {
     }
 
     @Test
+    public void shouldParseUiExtensionIdFromValidURI() throws Exception {
+        String url = BASE_PATH + "?mpc_install=3414678";
+        String extensionId = marketplaceService.getExtensionId(new URI(url));
+
+        assertThat(extensionId, is("market:ui-3414678"));
+    }
+
+    @Test
+    public void shouldParseVoiceExtensionIdFromValidURI() throws Exception {
+        String url = BASE_PATH + "?mpc_install=396962433";
+        String extensionId = marketplaceService.getExtensionId(new URI(url));
+
+        assertThat(extensionId, is("market:voice-396962433"));
+    }
+
+    @Test
     public void shouldParseExtensionIdFromValidURIWithMultipleQueryParams() throws Exception {
         String url = BASE_PATH + "?p1&p2=foo&mpc_install=3305842&p3=bar";
         String extensionId = marketplaceService.getExtensionId(new URI(url));
@@ -94,6 +110,8 @@ public class MarketplaceExtensionServiceTest {
         List<Node> nodes = new ArrayList<>(2);
         nodes.add(createBindingNode());
         nodes.add(createRuleNode());
+        nodes.add(createUiNode());
+        nodes.add(createVoiceNode());
         return nodes;
     }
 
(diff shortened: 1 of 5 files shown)


maggu2810:
I really like to add support for other types, too.


Just to link another comment from https://github.com/eclipse/smarthome/pull/4112#issuecomment-355006491

>> Shouldn't we enhance that bundle to support a transport type, too?

> There isn't such a thing as a "transport add-on/extension". Transport bundles are internal bundles that are technical dependencies, which get automatically installed (e.g. through Karaf if they are declared in a Karaf feature). For the marketplace, we do not have a dependency mechanism in place. What is prepared is to have other packaging formats than "bundle" in future, thus as "kar" - this would allow packaging multiple bundles that are required for a single add-on. But this is yet something to implement for the marketplace support.
> Until then, we only have "binding" and "rule templates" as valid categories.

I assume that UI and Voice are not "technical dependencies" only, but extensions that an user wants to add explicitly. Do you agree @kaikreuzer

jsichi:
Sure, just let me know exactly which types make sense.  For voice, I think it's clear that these are extensions a user wants to add; we already have two which should be recategorized (they are currently listed as bindings):

https://marketplace.eclipse.org/content/pollytts-voice-service
https://marketplace.eclipse.org/content/shlurd-human-language-interpreter

For UI, I noticed that there was some discussion regarding the need to distinguish complete interfaces (such as HabPanel) from widgets (such as the gauge widget at https://marketplace.eclipse.org/content/gauge-widget-habpanel).  So maybe we should have a corresponding ui-widget package type?

I'll wait to hear from @kaikreuzer and then update this PR accordingly.

jsichi:
Ping?

jsichi:
Resolved merge conflicts.  @maggu2810 could you tell me the best way to get a code review started for this?

maggu2810:
@jsichi I am waiting for some comment of the other committers (that they agree it is a good idea). Then the review itself shouldn't be the problem.

htreu:
Hi @jsichi, please take another look at the merge you did. We avoid merging branches but do rebases instead. 
Please remove the merge commit and do a `git rebase master` and `git push --force` to update this PR.

kaikreuzer:
@jsichi Sorry for my late reply and thanks for picking up that issue.

Wrt "voice": As you have correctly noticed, it currently is not possible to publish bundles with that category - I will have to ask Eclipse admins to add that in the marketplace first. But yes, I agree that this category makes sense.

Wrt "ui": We currently do not have any UIs in the marketplace, so I do not see a need for this right now. HABPanel widgets actually do not belong in the marketplace at all - all add-ons are meant to be fully ESH-compatible and not openHAB-specific; and as long as HABPanel itself is a proprietary openHAB UI, its extensions simply do not belong in the Eclipse Marketplace.

jsichi:
@htreu got it, will do.

@kaikreuzer OK, I will remove the ui addon support.

jsichi:
I've rebased the pull request and removed the UI support.

kaikreuzer:
I've created a request for "Voice" type in the Marketplace: https://bugs.eclipse.org/bugs/show_bug.cgi?id=534469

jsichi:
A month later, I don't see any response to the request you created...how often do those get processed?

kaikreuzer:
Don't ask me - better leave a comment on that request yourself and ask there - maybe this helps to get someones attention...

jsichi:
OK, the marketplace request has been carried out, and I've successfully tested it with my branch + extension.  Extensions which use the new Voice type will only be installable from ESH instances running a version which includes this PR (after it gets merged).  For older versions, such an extension won't be visible at all.

PR #5356, opened 2018-04-04
Add guava avoidance remark to coding guidelines

sjka:
In the realm of #4743 and #2801 et al. it makes sense to
not accept any new contributions adding unecessary usages
of guava again, therefore the documentation is updated
accordingly.


sjka:
> Hopefully people read this documentation 
Not needed - but I will have something I can point them to 😉

maggu2810:
I fully agree!

PR #5360, opened 2018-04-05
Clarify reasons for null annotations in coding guidelines

triller-telekom:

### DIFF
--- a/docs/documentation/development/guidelines.md
+++ b/docs/documentation/development/guidelines.md
@@ -26,7 +26,13 @@ To speed up the contribution process, we therefore advice to go through this che
 1. For dependency injection, OSGi Declarative Services should be used.
 1. OSGi Declarative Services should be declared using annotations. The IDE will take care of the service *.xml file creation. See the official OSGi documentation for an [example here](http://enroute.osgi.org/services/org.osgi.service.component.html). We always use `@Activate`, `@Deactivate` and `@Modified` if we define these methods, even if they exist in a super class, to make the code more readable.
 1. Packages that contain classes that are not meant to be used by other bundles should have "internal" in their package name.
-1. [Null annotations](https://wiki.eclipse.org/JDT_Core/Null_Analysis) are used from the Eclipse JDT project. Therefore every bundle should have an **optional** `Import-Package` dependency to `org.eclipse.jdt.annotation`.
+1. [Null annotations](https://wiki.eclipse.org/JDT_Core/Null_Analysis) are used from the Eclipse JDT project.
+Our intention with these annotations is to transfer a method's contract written in its JavaDoc into the code to be processed by tools.
+We are aware that these annotations can be used for **static** checks, but **not** at runtime.  
+Thus for publicly exposed methods that belong to our API and are (potentially) called by external callers, we cannot omit a `null` check although a method parameter is marked to be not `null` via an annotation.
+We will get a warning in the IDE for this check, but we decided to live with that.
+For private methods or methods in an internal package we agreed to respect the annotations and omit an additional `null` check.  
+To use the annotaions, every bundle should have an **optional** `Import-Package` dependency to `org.eclipse.jdt.annotation`.
 Classes should be annotated by `@NonNullByDefault` and return types, parameter types, generic types etc. are annotated with `@Nullable` only.
 Fields that get a static and mandatory reference injected through OSGi Declarative Services can be annotated with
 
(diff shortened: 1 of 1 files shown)


triller-telekom:
@maggu2810 In the light of https://github.com/eclipse/smarthome/issues/4321#issuecomment-378714537 can you please have a look on my explanation?

docs/documentation/development/guidelines.md
```diff
@@ -26,7 +26,13 @@ To speed up the contribution process, we therefore advice to go through this che
...
-1. [Null annotations](https://wiki.eclipse.org/JDT_Core/Null_Analysis) are used from the Eclipse JDT project. Therefore every bundle should have an **optional** `Import-Package` dependency to `org.eclipse.jdt.annotation`.
+1. [Null annotations](https://wiki.eclipse.org/JDT_Core/Null_Analysis) are used from the Eclipse JDT project.
+Our intention with these annotations is to transfer a method's contract written in its JavaDoc into the code to be processed by tools.
+We are aware that these annotations can be used for **static** checks, but **not** at runtime.  
+Thus for publicly exposed methods that belong to our API and are (potentially) called by external callers, we cannot omit a `null` check although a method parameter is marked to be not `null` via an annotation.
+We will get a warning in the IDE for this check, but we decided to live with that.
+For private methods or methods in an internal package we agreed to respect the annotations and omit an additional `null` check.  
+To use the annotaions, every bundle should have an **optional** `Import-Package` dependency to `org.eclipse.jdt.annotation`.
```
IOOOTAlan: Spotted little typo: annotaions -> annotations

triller-telekom: @IOOOTAlan Thanks, I have fixed it in #5363

tmrobert8:
Few comments on this
1.  Can you clarify whether the usage of package-info.java to specify NonNullByDefault as the default for all classes in the package is required, recommended, optional or disallowed?  I personally use it but don't have strong opinions one way or the other whether it should be required or recommended or optional or even not allowed (for those that think it takes away from readability).
2. Can you clarify if additional null checks can be optional however?  I prefer to have the checks on all methods (public and private) because it prevents maintenance issues down the road.  Example: let's say I have a private method that shouldn't allow nulls, will screw up in various ways if it's passed a null but I went with these guidelines and didn't put a null check in.  No problems since I own all the calling code and know not to pass a null.  Now let's say I get hit by a bus and someone picks up maintaining the code.  In some future change, they need to call that method and either don't understand the null contract or does some quick refactoring ("Change visibility of method to public") without understanding the implication and suddenly things blow up.  To me, private or public is simply an accessibility marker (as it should be) and makes no difference in the contracts for the method (ie nullable or not).   While I think the guideline for private is going to create issues for people down the road - I'd like my code well protected and following my method contracts - so I'd like to make null checks optional (I'd really like to make them required - but for some reason people think private methods are somehow different than public methods even though they may be public in the future)...

maggu2810:
> Can you clarify whether the usage of package-info.java to specify NonNullByDefault as the default for all classes in the package is required, recommended, optional or disallowed? I personally use it but don't have strong opinions one way or the other whether it should be required or recommended or optional or even not allowed (for those that think it takes away from readability).

We have decided to not use `package-info.java` but annotate every class.
If `package-info.java` is used on some places to mark the whole package and the respective class on some other places it is not such easy to identify the classes that are already annotated and the ones that are not (because you have also to look at the package level).

About your point (2). As we are a community projects with multiple committers it is unlikely that all that ones leave the projects at the same time. :wink: 
Validate every argument on every function would perhaps make the code more robust, but will make it unusable on slow machines if this overhead is added to every function.

tmrobert8:
Note: my perspective is more from the openhab addon writer (which these guidelines are being applied to).

1.  I'm fine with package-info.java being disallowed - just needs to be documented then (would this still apply to addon writers?)
2.  Point taken for the framework but it doesn't make any sense for an addon (which is typically owned by a singular person) and since this is also being applied to addon code - I'd like to have null checks be optional.  BTW - I fully disagree with the performance overhead as a *general* statement.  There are cases where that might be true but you are talking about something in the .001% range of code

maggu2810:
So, if you would like to discuss the nullness annotations on openHAB addons you should perhaps discuss it on the respective board. This one is about the ESH code.
If your addon code is something in the "0.001%" range of code it is another story, but the framework methods are called more frequently.

maggu2810:
> ah - @triller-telekom asked me to post over and I thought this was this was also being applied to the openhab guidelines as well. Sorry about that - my misunderstanding.

Perhaps openHAB applies the ESH guidelines to their project. @kaikreuzer can answer that in more details.
But if some of the ESH guidelines should not be applied to openHAB because it does not fit, or openHAB would like to add other guidelines, that is IMHO something of the respective project.

As some of the openHAB addons has been migrated to ESH and the projects are pretty similar it would perhaps make sense to not force something impossible for OH in the ESH guidelines, but e.g. the point "an addon (which is typically owned by a singular person)" should not be true for addons that has been migrated to ESH.
Addons that are such exotic that they are only used (and maintained) by a few persons are perhaps not best placed into the ESH repo at all (there has been some discussion sometime about the correct place of addons -- but IIRC this has been an offline discussion).

PR #5361, opened 2018-04-05
Simple Magic Firmware Update ThingType

dstivanov:
Related to #4964. The following changes are introduced:
- Used to test the refactoring in PR #5343 and is now rebased, because #5343 is merged;
- Added simple magic firmware update thing type
- The model of the magic firmware update thing can be configured during pairing 
- Implemented simple firmware provider which provides several hardcoded firmware for different models;
- Updated the README.MD file to some of the previous changes + the new ones.

### DIFF
--- a/bundles/test/org.eclipse.smarthome.magic/ESH-INF/thing/thing-types.xml
+++ b/bundles/test/org.eclipse.smarthome.magic/ESH-INF/thing/thing-types.xml
@@ -1,5 +1,6 @@
 <?xml version="1.0" encoding="UTF-8"?>
-<thing:thing-descriptions bindingId="magic" xmlns:xsi="http://www.w3.org/2001/XMLSchema-instance"
+<thing:thing-descriptions bindingId="magic"
+	xmlns:xsi="http://www.w3.org/2001/XMLSchema-instance"
 	xmlns:thing="http://eclipse.org/smarthome/schemas/thing-description/v1.0.0"
 	xsi:schemaLocation="http://eclipse.org/smarthome/schemas/thing-description/v1.0.0 http://eclipse.org/smarthome/schemas/thing-description-1.0.0.xsd">
 
@@ -100,5 +101,30 @@
 		<channels>
 			<channel id="number" typeId="number" />
 		</channels>
+	</thing-type>
+
+	<thing-type id="firmware-update">
+		<label>Magic Firmware Updatable Thing</label>
+		<description>A thing which allows updating its firmware</description>
+
+		<properties>
+			<property name="firmwareVersion">0.1.0</property>
+		</properties>
+
+		<config-description uri="thing-type:magic:update-config">
+			<parameter name="updateModel" type="text" required="true">
+				<required>true</required>
+				<options>
+					<option value="Alohomora">Alohomora</option>
+					<option value="Colloportus">Colloportus</option>
+					<option value="Lumos">Lumos</option>
+					<option value="Nox">Nox</option>
+				</options>
+				<label>Model ID</label>
+				<description>The model of the magic firmware update thing</description>
+				<default>Not specified</default>
+			</parameter>
+		</config-description>
+
 	</thing-type>
 </thing:thing-descriptions>
--- a/bundles/test/org.eclipse.smarthome.magic/README.md
+++ b/bundles/test/org.eclipse.smarthome.magic/README.md
@@ -4,7 +4,6 @@ The Magic Bundle is a virtual device bundle which provides different Things, Cha
 
 Future plans:
 
-* Firmware update support
 * Simulate communication errors
 * Provide REST API to update thing status from _outside_
 * Provide REST API to temporarily create new Channels/Things
@@ -14,6 +13,13 @@ Future plans:
 * Magic Light - On/Off
 * Magic Light - Dimmable
 * Magic Light - Color
+* Magic Sensor - Door/Window Contact
+* Magic Location
+* Magic Configurable Thing
+* Magic Thermostat
+* Magic Delayed Online Thing - goes online after some time
+* Magic Firmware Updatable Thing - can be firmware updated, depending on the model
+
 
 ## Discovery
 
@@ -34,6 +40,11 @@ Available channels:
 * switch - the on/off toggle maps to a Switch item.
 * brightness - the brightness value maps to a Dimmer item.
 * color - the color maps to a Color item.
+* alert - the alert function of the color light.
+* contact - the contact of the door/window contact.
+* location - the location of the magic location.
+* temperature - the temperature of the magic thermostat.
+* number - the delay in seconds for the delayed thing to go online.
 
 ## Full Example
 
--- a/bundles/test/org.eclipse.smarthome.magic/src/main/java/org/eclipse/smarthome/magic/binding/handler/MagicFirmwareUpdateThingHandler.java
+++ b/bundles/test/org.eclipse.smarthome.magic/src/main/java/org/eclipse/smarthome/magic/binding/handler/MagicFirmwareUpdateThingHandler.java
@@ -1,0 +1,99 @@
+/**
+ * Copyright (c) 2014,2018 Contributors to the Eclipse Foundation
+ *
+ * See the NOTICE file(s) distributed with this work for additional
+ * information regarding copyright ownership.
+ *
+ * This program and the accompanying materials are made available under the
+ * terms of the Eclipse Public License 2.0 which is available at
+ * http://www.eclipse.org/legal/epl-2.0
+ *
+ * SPDX-License-Identifier: EPL-2.0
+ */
+package org.eclipse.smarthome.magic.binding.handler;
+
+import org.eclipse.jdt.annotation.NonNullByDefault;
+import org.eclipse.smarthome.core.thing.ChannelUID;
+import org.eclipse.smarthome.core.thing.Thing;
+import org.eclipse.smarthome.core.thing.ThingStatus;
+import org.eclipse.smarthome.core.thing.ThingStatusDetail;
+import org.eclipse.smarthome.core.thing.binding.BaseThingHandler;
+import org.eclipse.smarthome.core.thing.binding.firmware.Firmware;
+import org.eclipse.smarthome.core.thing.binding.firmware.FirmwareUpdateHandler;
+import org.eclipse.smarthome.core.thing.binding.firmware.ProgressCallback;
+import org.eclipse.smarthome.core.thing.binding.firmware.ProgressStep;
+import org.eclipse.smarthome.core.types.Command;
+import org.eclipse.smarthome.magic.binding.MagicBindingConstants;
+
+/**
+ * Handler for firmware updatable magic things. Defines full progress sequence and simulates firmware update with small
+ * delays between the steps.
+ *
+ * @author Dimitar Ivanov - Initial contribution
+ *
+ */
+@NonNullByDefault
+public class MagicFirmwareUpdateThingHandler extends BaseThingHandler implements FirmwareUpdateHandler {
+
+    private static final int STEP_DELAY = 100;
+
+    public MagicFirmwareUpdateThingHandler(Thing thing) {
+        super(thing);
+    }
+
+    @Override
+    public void handleCommand(ChannelUID channelUID, Command command) {
+        // Nothing to do
+    }
+
+    @Override
+    public void initialize() {
+        String updateModel = (String) getThing().getConfiguration().get(MagicBindingConstants.UPDATE_MODEL_PROPERTY);
+        switch (updateModel) {
+            case MagicBindingConstants.MODEL_ALOHOMORA:
+            case MagicBindingConstants.MODEL_COLLOPORTUS:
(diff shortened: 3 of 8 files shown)


dstivanov:
As #5343 is finally merged I adapted the code of the magic bundle here. @sjka @htreu : Can you please  take a look?

dstivanov:
As #5343 is finally merged I adapted the code of the magic bundle here. @sjka @htreu : Can you please  take a look?

bundles/test/org.eclipse.smarthome.magic/src/main/java/org/eclipse/smarthome/magic/binding/internal/firmware/MagicFirmwareProvider.java
```diff
@@ -0,0 +1,102 @@
...
+ *
+ * @author Dimitar Ivanov - Initial contribution
+ */
+@Component(service = FirmwareProvider.class)
+public class MagicFirmwareProvider implements FirmwareProvider {
+
+    private final Map<ThingTypeUID, Set<Firmware>> magicFirmwares = new HashMap<ThingTypeUID, Set<Firmware>>();
+
+    /**
+     * So far only this thing type is used for the firmware update feature of the magic binding. If this changes, we can
```
sjka: To my understanding the code here is not intended to promote a pattern anyway (like: _if_ you implement a firmware provider, this is your copy'n'paste template), right? As the code has to be changed _anyway_ in such a case, I'd prefer to keep it as simple as possible, i.e. also turn the `magicFirmwares` map into a set and drop the `addFirmware(...)` method for now.

bundles/test/org.eclipse.smarthome.magic/src/main/java/org/eclipse/smarthome/magic/binding/internal/firmware/MagicFirmwareProvider.java
```diff
@@ -0,0 +1,102 @@
...
+     * add a parameter to the {@link #createFirmware(String, String, boolean)} and
+     * {@link #addFirmware(Firmware)} methods.
+     */
+    private final ThingTypeUID thingTypeUID = MagicBindingConstants.THING_TYPE_FIRMWARE_UPDATE;
```
sjka: This should be a constant

bundles/test/org.eclipse.smarthome.magic/src/main/java/org/eclipse/smarthome/magic/binding/internal/firmware/MagicFirmwareProvider.java
```diff
@@ -0,0 +1,102 @@
...
+
+/**
+ * Provides firmware for the magic thing type for firmware update.
+ *
+ * @author Dimitar Ivanov - Initial contribution
+ */
+@Component(service = FirmwareProvider.class)
+public class MagicFirmwareProvider implements FirmwareProvider {
+
+    private final Map<ThingTypeUID, Set<Firmware>> magicFirmwares = new HashMap<ThingTypeUID, Set<Firmware>>();
```
sjka: Make it a set and think about initializing it statically

bundles/test/org.eclipse.smarthome.magic/src/main/java/org/eclipse/smarthome/magic/binding/internal/MagicHandlerFactory.java
```diff
@@ -83,6 +84,9 @@ protected ThingHandler createHandler(Thing thing) {
         if (thingTypeUID.equals(THING_TYPE_THERMOSTAT)) {
             return new MagicThermostatThingHandler(thing);
         }
+        if(thingTypeUID.equals(THING_TYPE_FIRMWARE_UPDATE)) {
```
sjka: What happened to your formatter?

bundles/test/org.eclipse.smarthome.magic/src/main/java/org/eclipse/smarthome/magic/binding/handler/MagicFirmwareUpdateThingHandler.java
```diff
@@ -0,0 +1,96 @@
...
+
+    @Override
+    public void handleCommand(ChannelUID channelUID, Command command) {
+        // Nothing to do
+    }
+
+    @Override
+    public void initialize() {
+        String updateModel = (String) getThing().getConfiguration().get(MagicBindingConstants.UPDATE_MODEL_PROPERTY);
+        switch (updateModel) {
```
sjka: What is the point of this switch statement?

bundles/test/org.eclipse.smarthome.magic/src/main/java/org/eclipse/smarthome/magic/binding/handler/MagicFirmwareUpdateThingHandler.java
```diff
@@ -0,0 +1,96 @@
...
+import org.eclipse.smarthome.magic.binding.MagicBindingConstants;
+
+/**
+ * Handler for firmware updatable magic things. Defines full progress sequence and simulates firmware update with small
+ * delays between the steps.
+ *
+ * @author Dimitar Ivanov - Initial contribution
+ *
+ */
+public class MagicFirmwareUpdateThingHandler extends BaseThingHandler implements FirmwareUpdateHandler {
```
sjka: `@NonNullByDefault`

bundles/test/org.eclipse.smarthome.magic/src/main/java/org/eclipse/smarthome/magic/binding/handler/MagicFirmwareUpdateThingHandler.java
```diff
@@ -0,0 +1,96 @@
...
+
+    private static final int DELAY_BETWEEN_FIRMWARE_UPDATE_PROGRESS_STEPS = 100;
```
sjka: `STEP_DELAY ` (or the like) would suffice - looks so much like groovy otherwise 😉

bundles/test/org.eclipse.smarthome.magic/src/main/java/org/eclipse/smarthome/magic/binding/internal/MagicHandlerFactory.java
```diff
@@ -45,7 +46,7 @@
 
     private static final Set<ThingTypeUID> SUPPORTED_THING_TYPES_UIDS = Sets.newHashSet(THING_TYPE_EXTENSIBLE_THING,
             THING_TYPE_ON_OFF_LIGHT, THING_TYPE_DIMMABLE_LIGHT, THING_TYPE_COLOR_LIGHT, THING_TYPE_CONTACT_SENSOR,
-            THING_TYPE_CONFIG_THING, THING_TYPE_DELAYED_THING, THING_TYPE_LOCATION, THING_TYPE_THERMOSTAT);
+            THING_TYPE_CONFIG_THING, THING_TYPE_DELAYED_THING, THING_TYPE_LOCATION, THING_TYPE_THERMOSTAT,THING_TYPE_FIRMWARE_UPDATE);
```
sjka: missing space

bundles/test/org.eclipse.smarthome.magic/README.md
```diff
@@ -34,6 +40,11 @@ Available channels:
 * switch - the on/off toggle maps to a Switch item.
 * brightness - the brightness value maps to a Dimmer item.
 * color - the color maps to a Color item.
+* alert - the alert function of the color light.
+* contact - the contact of the Door/Window Contact.
```
sjka: lowercase: `door/window contact`

bundles/test/org.eclipse.smarthome.magic/ESH-INF/thing/thing-types.xml
```diff
@@ -101,4 +102,29 @@
...
+
+		<properties>
+			<property name="firmwareVersion">0.1.0</property>
+		</properties>
+
+		<config-description uri="thing-type:magic:update-config">
+			<parameter name="updateModel" type="text" required="true">
+				<required>true</required>
+				<options>
+					<option value="Not specified">Not specified</option>
```
sjka: Wouldn't it make sense to have something different as values, like e.g. numbers? Just a thought, to keep it a little more realistic...

bundles/test/org.eclipse.smarthome.magic/ESH-INF/thing/thing-types.xml
```diff
@@ -101,4 +102,29 @@
 			<channel id="number" typeId="number" />
 		</channels>
 	</thing-type>
+
+	<thing-type id="firmware-update">
+		<label>Magic Firmware Updatable Thing</label>
+		<description>A thing that can be firmware updated.</description>
```
sjka: How about `A thing which allows updating its firmware`?

No trailing `.` anyway, please.

bundles/test/org.eclipse.smarthome.magic/src/main/java/org/eclipse/smarthome/magic/binding/internal/firmware/MagicFirmwareProvider.java
dstivanov: My intention here was to leave it open for extensions such as a REST resource that can add firmwares during runtime for easier testing.  @sjka : If you think that such functionality will be too much for the magic bundle - I will make leave a simple set here.

bundles/test/org.eclipse.smarthome.magic/src/main/java/org/eclipse/smarthome/magic/binding/internal/firmware/MagicFirmwareProvider.java
dstivanov: Yes, you are right. I would even prefer to use the `MagicBindingConstants.THING_TYPE_FIRMWARE_UPDATE` everywhere, because it does make sense to define extra constant here.

bundles/test/org.eclipse.smarthome.magic/src/main/java/org/eclipse/smarthome/magic/binding/internal/firmware/MagicFirmwareProvider.java
dstivanov: See my comment above. If you think that we do not need extra extensions here - I will leave it as simple as possible.

bundles/test/org.eclipse.smarthome.magic/ESH-INF/thing/thing-types.xml
dstivanov: Yes, will change them a bit

bundles/test/org.eclipse.smarthome.magic/ESH-INF/thing/thing-types.xml
dstivanov: Your description looks more accurate, I will use it.

bundles/test/org.eclipse.smarthome.magic/src/main/java/org/eclipse/smarthome/magic/binding/internal/firmware/MagicFirmwareProvider.java
htreu: I'm with @sjka. Provide an impplementation which is as simple as possible in the magic bundle.

dstivanov: OK, I will do it simple :)

bundles/test/org.eclipse.smarthome.magic/src/main/java/org/eclipse/smarthome/magic/binding/internal/firmware/MagicFirmwareProvider.java
dstivanov: As @htreu is also with you on that I will make it as simple as possible.

dstivanov:
@htreu @sjka I simplified the firmware provider and addressed your comments. Can you please take another look at the PR?

bundles/test/org.eclipse.smarthome.magic/ESH-INF/thing/thing-types.xml
sjka: Little misunderstanding, sorry... I was rather referring to something like this:

```xml
<option value="1">Abracadabra</option>
<option value="2">Alohomora</option>
<option value="3">Lumos</option>
```

dstivanov: OK, I will ca choose some Harry Potter spells, starting with different letter to make them distinguishable :relaxed: And will remove the "not specified"  value, we will just have one model without specific firmwares.

dstivanov:
@sjka Updated models and tested in Smarthome runtime - works "as a charm". It is magic :)

sjka:
Awesome!

PR #5364, opened 2018-04-05
MqttConnection: Add timeout

davidgraeff:
Refactors:
* Use `ConnectionCallbacks` private class analoge to `ClientCallbacks` private class. Make both classes static. Implicit shared parent objects will not work with test mocks.

Enhancements:
* Add timeout functionality. At the moment if no connection can be established, just nothing happens. The underlying paho library works with seconds only, which is not feasible for tests. We therefore have out own timeout mechanism with the frameworks `ScheduledExecutorService`.
* Add test for timeouts.
* Emit the CONNECTING state change. The state existed already but was never emitted.


davidgraeff:
@maggu2810 Can you have a look at this? "Mqtt Broker configuration binding #3839" would need this timeout functionality.

bundles/io/org.eclipse.smarthome.io.transport.mqtt/src/main/java/org/eclipse/smarthome/io/transport/mqtt/MqttBrokerConnection.java
```diff
@@ -164,10 +240,33 @@ public void setReconnectStrategy(AbstractReconnectStrategy reconnectStrategy) {
...
+     *
+     * @param executor One timer will be created when a connection attempt happens
+     * @param timeoutInMS Timeout in milliseconds
+     */
+    public void setTimeoutExecutor(@Nullable ScheduledExecutorService executor, int timeoutInMS) {
+        timeoutExecutor = executor;
+        this.timeout = timeoutInMS;
+    }
+
+    public @Nullable Executor getTimeoutExecutor() {
```
maggu2810: Do we really need that getters? What is the intention? The executor should not be of any interest after it has been set and we should prevent from reuse the executor of an instance by a consumer for something different.

bundles/io/org.eclipse.smarthome.io.transport.mqtt/src/main/java/org/eclipse/smarthome/io/transport/mqtt/MqttBrokerConnection.java
```diff
@@ -164,10 +240,33 @@ public void setReconnectStrategy(AbstractReconnectStrategy reconnectStrategy) {
...
+        return timeoutExecutor;
+    }
+
+    public int getTimeout() {
```
maggu2810: The one that calls `set` knows about and I assume no one other needs to get it (or can you give an example).

davidgraeff:
@htreu Can you have a look please?

bundles/io/org.eclipse.smarthome.io.transport.mqtt/src/main/java/org/eclipse/smarthome/io/transport/mqtt/MqttBrokerConnection.java
```diff
@@ -84,66 +103,121 @@
...
+                if (c.reconnectStrategy != null) {
+                    c.reconnectStrategy.lostConnection();
+                }
+            }
+        }
+    }
+
+    /** Client callback object. Package local, for testing. */
+    protected MqttCallback clientCallbacks = new ClientCallbacks(this);
+    /** Connection callback object. Package local, for testing. */
```
htreu: its actually protected

bundles/io/org.eclipse.smarthome.io.transport.mqtt/src/main/java/org/eclipse/smarthome/io/transport/mqtt/MqttBrokerConnection.java
```diff
@@ -84,66 +103,121 @@
...
+                c.isConnecting = false;
+                if (c.reconnectStrategy != null) {
+                    c.reconnectStrategy.lostConnection();
+                }
+            }
+        }
+    }
+
+    /** Client callback object. Package local, for testing. */
+    protected MqttCallback clientCallbacks = new ClientCallbacks(this);
```
htreu: its actually protected

bundles/io/org.eclipse.smarthome.io.transport.mqtt/src/main/java/org/eclipse/smarthome/io/transport/mqtt/MqttBrokerConnection.java
```diff
@@ -530,52 +598,79 @@ public synchronized void start() throws MqttException, ConfigurationException {
...
         serverURI.append(":");
         serverURI.append(port);
 
-        // Close client if there is still one existing
-        if (client != null) {
-            try {
-                client.close();
-            } catch (org.eclipse.paho.client.mqttv3.MqttException ignore) {
+        // Storage
+        Path tmpDir = Paths.get(ConfigConstants.getUserDataFolder());
```
htreu: when running from the IDE my tmpDir is always `userdata` and not absolute.
This can be fixed by adding `toAbsolutePath()` to the `Paths#get` call. wdyt?

bundles/io/org.eclipse.smarthome.io.transport.mqtt/src/main/java/org/eclipse/smarthome/io/transport/mqtt/MqttBrokerConnection.java
```diff
@@ -530,52 +598,79 @@ public synchronized void start() throws MqttException, ConfigurationException {
...
+        try {
+            if (!tmpDir.isAbsolute()) {
```
htreu: with the above addition this should not fail anymore.

bundles/io/org.eclipse.smarthome.io.transport.mqtt/src/main/java/org/eclipse/smarthome/io/transport/mqtt/MqttBrokerConnection.java
davidgraeff: The user dir is always absolute in production environment. I added this precheck to not make the class create directories in various test working dirs.

It is actually a bug in one of the tests if this method is not overwritten with a no-op.

maggu2810: > The user dir is always absolute in production environment.

Why?

The code looks like:

```Java
public static final String DEFAULT_USERDATA_FOLDER = "userdata";

// ...

public static String getUserDataFolder() {
    String progArg = System.getProperty(USERDATA_DIR_PROG_ARGUMENT);
    if (progArg != null) {
        return progArg;
    } else {
        return DEFAULT_USERDATA_FOLDER;
    }
}
```

So for example if `progArg == null` then `userdata`  is returned.
This is not absolute.

Also the `USERDATA_DIR_PROG_ARGUMENT` property does not need to be an absolute path.

htreu: > It is actually a bug in one of the tests if this method is not overwritten with a no-op.
tests are fine.

From my pov you can just create the folders on the user directory. @maggu2810 any objections?

davidgraeff: Hm, but tests shouldn't alter the working directory, should they? I only checked with my OH environment and the path was absolute, I thought that's the differentiation of testing Vs production systems.

maggu2810: > From my pov you can just create the folders on the user directory. @maggu2810 any objections?

The user data folder needs to be a writable location.
Why do you think the user directory needs to be a writable one?
This has till now never be a requirement for an ESH based solution (AFAIK).
And why should we fill the user directory with ESH stuff at all? Isn't that the intention of the user data folder, that ESH runtime data is stored at that place?
This will also break solutions that runs multiple ESH instances as the same OS user.

> but tests shouldn't alter the working directory, should they

I agree that the CWD should not be changed at runtime.
Using absolute or relative should not care, because if the CWD is changed also the conversion to an absolute path will result into a wrong directory.
I just want to point out, that `getUserDataFolder` can return an relative or an absolute directory, that depends on the system properties.

So, I am fine using the path returned by `getUserDataFolder()` and create a subfolder `mqtt` in that place (perhaps we should use `io.transport.mqtt` to not use the same directory for different MQTT bundles some time).

htreu: The test run well in the ESH IDE. No folders get created afaik.
But running ESH with io.transport.mqtt from the IDE results in the Exception for every new service instance I configure. So I just suggest to make the path absolute. Or do I overlook a problem here?

htreu: so, to clear this up:

> So, we have to use getUserDataFolder() and not the user's home directory.

Agreed. My suggestion was misleading due to imprecise wording. I was also talking about `userdata` which is the default user data folder.

> If the test only succeeds only if this directory is absolute

as @davidgraeff mentioned, unit tests do override the corresponding `createAndConnectClient` method and do not create any folders. Its only the ESH execution in the IDE (which falls back to the default `userdata` relative path).
With the use of `toAbsolutePath()` we should be all golden.

davidgraeff:
@htreu Everything addressed

PR #5365, opened 2018-04-05
Added null annotations to transformation services

triller-telekom:
While adding the annotations I noticed an inconsistency in the
RegExTransformationService. It validated the transformation contract
because it returned null. Now it returns UnDefType.NULL.toFullString()
just like the JSonPathTransformationService.

### DIFF
--- a/bundles/core/org.eclipse.smarthome.core.transform/src/main/java/org/eclipse/smarthome/core/transform/AbstractFileTransformationService.java
+++ b/bundles/core/org.eclipse.smarthome.core.transform/src/main/java/org/eclipse/smarthome/core/transform/AbstractFileTransformationService.java
@@ -54,7 +54,7 @@ import org.slf4j.LoggerFactory;
 @NonNullByDefault
 public abstract class AbstractFileTransformationService<T> implements TransformationService {
 
-    @NonNullByDefault({})
+    @Nullable
     private WatchService watchService = null;
 
     protected final Map<String, T> cachedFiles = new ConcurrentHashMap<>();
--- a/bundles/core/org.eclipse.smarthome.core.transform/src/main/java/org/eclipse/smarthome/core/transform/AbstractFileTransformationService.java
+++ b/bundles/core/org.eclipse.smarthome.core.transform/src/main/java/org/eclipse/smarthome/core/transform/AbstractFileTransformationService.java
@@ -120,7 +120,7 @@ public abstract class AbstractFileTransformationService<T> implements Transforma
      * @throws TransformationException
      */
     @Override
-    public String transform(String filename, String source) throws TransformationException {
+    public @Nullable String transform(String filename, String source) throws TransformationException {
         if (filename == null || source == null) {
             throw new TransformationException("the given parameters 'filename' and 'source' must not be null");
         }
--- a/bundles/core/org.eclipse.smarthome.core.transform/src/main/java/org/eclipse/smarthome/core/transform/AbstractFileTransformationService.java
+++ b/bundles/core/org.eclipse.smarthome.core.transform/src/main/java/org/eclipse/smarthome/core/transform/AbstractFileTransformationService.java
@@ -29,6 +29,8 @@ import java.util.Map;
 import java.util.concurrent.ConcurrentHashMap;
 
 import org.apache.commons.io.FilenameUtils;
+import org.eclipse.jdt.annotation.NonNullByDefault;
+import org.eclipse.jdt.annotation.Nullable;
 import org.eclipse.smarthome.config.core.ConfigConstants;
 import org.eclipse.smarthome.core.i18n.LocaleProvider;
 import org.osgi.framework.BundleContext;
@@ -49,8 +51,10 @@ import org.slf4j.LoggerFactory;
  * @author Kai Kreuzer - File caching mechanism
  * @author Markus Rathgeb - Add locale provider support
  */
+@NonNullByDefault
 public abstract class AbstractFileTransformationService<T> implements TransformationService {
 
+    @NonNullByDefault({})
     private WatchService watchService = null;
 
     protected final Map<String, T> cachedFiles = new ConcurrentHashMap<>();
@@ -58,7 +62,9 @@ public abstract class AbstractFileTransformationService<T> implements Transforma
 
     private final Logger logger = LoggerFactory.getLogger(AbstractFileTransformationService.class);
 
+    @NonNullByDefault({})
     private LocaleProvider localeProvider;
+    @NonNullByDefault({})
     private ServiceTracker<LocaleProvider, LocaleProvider> localeProviderTracker;
 
     private class LocaleProviderServiceTrackerCustomizer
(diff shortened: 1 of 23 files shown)


bundles/core/org.eclipse.smarthome.core.transform/src/main/java/org/eclipse/smarthome/core/transform/AbstractFileTransformationService.java
```diff
@@ -49,16 +51,20 @@
  * @author Kai Kreuzer - File caching mechanism
  * @author Markus Rathgeb - Add locale provider support
  */
+@NonNullByDefault
 public abstract class AbstractFileTransformationService<T> implements TransformationService {
 
+    @NonNullByDefault({})
```
maggu2810: The watchService member could be null, so why disable the null checks instead of changing the code little bit and use the correct nullable annotation?

We could change the initializeWatchService method with something like this one that returns a non null watch service (just a non tested example):

```Java
private synchronized WatchService getWatchService() throws IOException {
    WatchService watchService = this.watchService;
    if (watchService == null) {
        try {
            watchService = this.watchService = FileSystems.getDefault().newWatchService();
            watchSubDirectory("");
        } catch (IOException e) {
            logger.error("Unable to start transformation directory monitoring");
        }
    }
    return watchService;
}
```

This also identifies an potential error in the current code:
If the watchService is null in the transform method, the initialize watch service method is called.
If `FileSystems.getDefault().newWatchService();` throws an IOException this is hidden from the caller and the watchService is still `null`.
After that `getLocalizedProposedFilename` is called, that itself could call `watchSubDirectory` that requires a non null watch service...

The annotation change does not make the runtime behavior different then it is now, but perhaps we should try to improve the implementation and error handling in another PR after that one.

WDYT?

maggu2810: Sorry, to get offtopic, but is there some guarantee that the transform method is not called multiple times from different thread? If this is potential possible, the watchService could be initialized twice...

maggu2810: @triller-telekom I tried to harden the watch service usage: #5367

triller-telekom: very good finding and thanks for changing the ode to be more robust with your PR.

extensions/transform/org.eclipse.smarthome.transform.regex/src/main/java/org/eclipse/smarthome/transform/regex/internal/RegExTransformationService.java
```diff
@@ -66,7 +69,7 @@ public String transform(String regExpression, String source) throws Transformati
             logger.debug(
                     "the given regex '^{}$' doesn't match the given content '{}' -> couldn't compute transformation",
                     regExpression, source);
-            return null;
+            return UnDefType.NULL.toFullString();
```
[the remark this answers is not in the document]
triller-telekom: As I said in the commit message: I have aligned it with the other transformation services.

Nevertheless we should mark this PR as API-breaking because people might have used it in rules and now have to test against "NULL" instead of `null`.

bundles/core/org.eclipse.smarthome.core.transform/src/main/java/org/eclipse/smarthome/core/transform/AbstractFileTransformationService.java
maggu2810: The PR #5367 has been merged, would you like to rebase your branch and mark this variable as `Nullable`

triller-telekom: Done

triller-telekom:
This `transform` method is also used for state transformations, if we return `null` here we would have a `null` value as a state. I thought we agreed that a `State` of an `Item` should **never** be null, but rather `UndefType.NULL`.

We could change the return type of the `transform` method to `State` and return a `StringType("NULL")` to indicate that the transformation was not successful, but this will break the API even worse.

I would keep it like it is now, so it is in sync with `JSONPathTransformationService` and can be used for state transformations as usual.

To use the transformation in rules or scripts I would vote for this POC PR https://github.com/eclipse/smarthome/pull/5168 and also extend that one to RegExTransformations.

How shall we proceed here?

kaikreuzer:
>  if we return null here we would have a null value as a state. 

Why? Isn't there a place where we could set UnDefType.UNDEF as a state if the transform returns null?

triller-telekom:
Since the ´ItemUIregistryImpl` catches the `TransformationException` to handle the case of a failing transformation, I am throwing this exception now instead of returning `null`, see:

https://github.com/eclipse/smarthome/blob/3ee7e5be0831345e48836e05a230017d5ed15778/bundles/ui/org.eclipse.smarthome.ui/src/main/java/org/eclipse/smarthome/ui/internal/items/ItemUIRegistryImpl.java#L498-L506

kaikreuzer:
After your latest refactorings, is this still API breaking?

extensions/transform/org.eclipse.smarthome.transform.regex/src/main/java/org/eclipse/smarthome/transform/regex/internal/RegExTransformationService.java
```diff
@@ -87,7 +91,11 @@ public String transform(String regExpression, String source) throws Transformati
             }
         }
 
-        return result;
+        if (result != null) {
+            return result;
+        } else {
+            return UnDefType.NULL.toFullString();
```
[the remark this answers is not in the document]
triller-telekom: This line is to be conform with https://github.com/eclipse/smarthome/blob/master/extensions/transform/org.eclipse.smarthome.transform.jsonpath/src/main/java/org/eclipse/smarthome/transform/jsonpath/internal/JSonPathTransformationService.java#L60

because this method should not return `null`.

bundles/core/org.eclipse.smarthome.core.transform/src/main/java/org/eclipse/smarthome/core/transform/actions/Transformation.java
```diff
@@ -37,23 +52,31 @@
...
-                logger.error("Error executing the transformation '{}': {}", type, e.getMessage());
-                result = value;
-            }
-        } else {
-            logger.warn("No transformation service '{}' could be found.", type);
+        String result;
+        try {
+            result = trans(type, function, value);
+        } catch (TransformationException e) {
+            logger.error("Error executing the transformation '{}': {}", type, e.getMessage());
```
kaikreuzer: You will now log an error, where before was a warning (if the service is not found).
As we actually declare that this function expects failures and treats them internally already, I would claim that we should rather have only DEBUG, INFO or at most WARN level here - I'd prefer DEBUG.

PR #5381, opened 2018-04-08
make ui label and value coloring compatible with QuantityType

### LINKED ISSUE #5368: Basic UI: set label/value color for QuantityType states
From this comment https://github.com/eclipse/smarthome/issues/5243#issuecomment-378665172:

> How does it work? I didn't find an example. What has to be changed in the following line?
`Text item=fritzWashingMachineDECT200Temperature label="FRITZ!DECT 200 (Waschmaschine) [%.1f %unit%]" valuecolor=[<15="blue", <=18="green", <=21="orange", >21="red"]`
// EDIT: Item type is Number:Temperature.

The label- and value-colors definitions are not matched correctly against QuantityType states.
Since the syntax leaves little room for an extension I suggest to compare to the scalar value of the widget unit (which falls back to the item unit if not explicitly defined, like in the example above).

J-N-K:
fixes #5368 

### DIFF
--- a/bundles/ui/org.eclipse.smarthome.ui.test/META-INF/MANIFEST.MF
+++ b/bundles/ui/org.eclipse.smarthome.ui.test/META-INF/MANIFEST.MF
@@ -9,7 +9,7 @@ Bundle-Vendor: Eclipse.org/SmartHome
 Bundle-Version: 0.10.0.qualifier
 Fragment-Host: org.eclipse.smarthome.ui
 Import-Package: 
- javax.measure.quantity;version="1.0.0",
+ javax.measure.quantity,
  org.eclipse.jdt.annotation;resolution:=optional,
  org.eclipse.smarthome.core.library.unit,
  org.eclipse.smarthome.model.sitemap,
--- a/bundles/ui/org.eclipse.smarthome.ui.test/META-INF/MANIFEST.MF
+++ b/bundles/ui/org.eclipse.smarthome.ui.test/META-INF/MANIFEST.MF
@@ -9,8 +9,10 @@ Bundle-Vendor: Eclipse.org/SmartHome
 Bundle-Version: 0.10.0.qualifier
 Fragment-Host: org.eclipse.smarthome.ui
 Import-Package: 
+ javax.measure.quantity;version="1.0.0",
  org.eclipse.jdt.annotation;resolution:=optional,
  org.eclipse.smarthome.core.library.unit,
+ org.eclipse.smarthome.model.sitemap,
  org.hamcrest;core=split,
  org.junit;version="4.0.0",
  org.mockito,
--- a/bundles/ui/org.eclipse.smarthome.ui/src/main/java/org/eclipse/smarthome/ui/internal/items/ItemUIRegistryImpl.java
+++ b/bundles/ui/org.eclipse.smarthome.ui/src/main/java/org/eclipse/smarthome/ui/internal/items/ItemUIRegistryImpl.java
@@ -913,37 +913,44 @@ public class ItemUIRegistryImpl implements ItemUIRegistry {
                     break;
             }
         } else {
-            if (DecimalType.class.isInstance(state)) {
+            if (state instanceof DecimalType || state instanceof QuantityType<?>) {
                 try {
+                    Double compareDoubleValue = Double.parseDouble(unquotedValue);
+                    Double stateDoubleValue;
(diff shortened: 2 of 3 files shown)


cweitkamp:
Thank you. Do you think we can cover these calculations with some unit tests?

cweitkamp:
Ok, I assume increasing its visibility to package-private is not an option, is it?

J-N-K:
I've added two test cases, one for DecimalType, one for QuantityType. They basically check if the conversion is working, not that all comparisons are correct. This would require a lot more work.

cweitkamp:
> I've added two test cases, one for DecimalType, one for QuantityType.
Great. That is a good beginning.

bundles/ui/org.eclipse.smarthome.ui.test/META-INF/MANIFEST.MF
```diff
@@ -9,8 +9,10 @@ Bundle-Vendor: Eclipse.org/SmartHome
 Bundle-Version: 0.10.0.qualifier
 Fragment-Host: org.eclipse.smarthome.ui
 Import-Package: 
+ javax.measure.quantity;version="1.0.0",
```
htreu: please remove the version constraint.

bundles/ui/org.eclipse.smarthome.ui/src/main/java/org/eclipse/smarthome/ui/internal/items/ItemUIRegistryImpl.java
```diff
@@ -913,37 +917,44 @@ private boolean matchStateToValue(State state, String value, String matchConditi
...
         } else {
-            if (DecimalType.class.isInstance(state)) {
+            if (state instanceof DecimalType || state instanceof QuantityType<?>) {
                 try {
+                    double compareDoubleValue = Double.parseDouble(unquotedValue);
+                    double stateDoubleValue;
+                    if (state instanceof DecimalType) {
+                        stateDoubleValue = ((DecimalType) state).doubleValue();
+                    } else {
+                        stateDoubleValue = ((QuantityType<?>) state).doubleValue();
```
htreu: This unfortunately only holds true when the label says `%unit%`, so that the unit displayed for the widget is equal to the item state unit.
In case the widget label says `%.1f K` all temperature values will be converted to kelvin prior to rendering. So before taking the double value the state may have to be converted to the widget unit. [Have a look here](https://github.com/eclipse/smarthome/blob/master/bundles/ui/org.eclipse.smarthome.ui/src/main/java/org/eclipse/smarthome/ui/internal/items/ItemUIRegistryImpl.java#L414) on how this is done for the rendering part.

bundles/ui/org.eclipse.smarthome.ui.test/META-INF/MANIFEST.MF
J-N-K: done

bundles/ui/org.eclipse.smarthome.ui/src/main/java/org/eclipse/smarthome/ui/internal/items/ItemUIRegistryImpl.java
J-N-K: But that is exactly how we discussed it? The value colors should be in the same unit as the label. So if K is selected for the label, the coloring values should be K, too. The `state` already has the converted unit. 

`matchStateToValue? is called from `processColorDefintion`, which is called either from `getLabelColor` or `getValueColor`, on both cases the state is determined by `getState(w)` which does the conversion.

htreu: I just looked through the code and the sitemap DSL and its a bit more complicated than that. The sitemap DSL defines a color definition to be `<item>?<condition><state>=<argument>`. So instead of the state of the item displayed by the widget it can also check on any other item's state here.
So I guess we have to convert nonetheless. When the state is already in the widget unit the conversion is a NOP.

htreu: Yes, you are right. Conversion of _another_ item's state does not make sense here. So The code is fine and good to go.

cweitkamp:
Works fine again. Thanks

PR #5383, opened 2018-04-08
Remove Service-Component from serial transport manifest

wborn:
Fixes:

```
18:45:14.319 [ERROR] [eclipse.smarthome.io.transport.serial] - Component descriptor entry 'OSGI-INF/*.xml' not found
```


maggu2810:
You have been faster then me :wink:

PR #5384, opened 2018-04-08
serial: remove unused service component declaration

maggu2810:
Eclipse Equinox shows a warning if "*.xml" is used in the service
component declaration but there is no file that match that pattern in
the "OSGI-INF" directory.
As the plain serial bundle does not contains a DS component we could
remove that declaration.

Related to: https://github.com/eclipse/smarthome/pull/5313#issuecomment-379501569


wborn:
Duplicates #5383 ;-)

PR #5385, opened 2018-04-08
add ppm and dB to units

J-N-K:

### DIFF
--- a/bundles/core/org.eclipse.smarthome.core.test/src/test/java/org/eclipse/smarthome/core/library/types/QuantityTypeTest.java
+++ b/bundles/core/org.eclipse.smarthome.core.test/src/test/java/org/eclipse/smarthome/core/library/types/QuantityTypeTest.java
@@ -31,7 +31,6 @@ import org.junit.Before;
 import org.junit.Test;
 
 import tec.uom.se.quantity.QuantityDimension;
-import tec.uom.se.unit.Units;
 
 /**
  * @author Gaël L'hopital - initial contribution
@@ -231,18 +230,6 @@ public class QuantityTypeTest {
     @Test(expected = ArithmeticException.class)
     public void testDivide_Zero() {
         new QuantityType<>("4 m").divide(QuantityType.ZERO);
-    }
-
-    @Test
-    public void testPpm() {
-        QuantityType<Dimensionless> ppm = new QuantityType<>("500 ppm");
-        assertEquals("0.05 %", ppm.toUnit(Units.PERCENT).toString());
-    }
-
-    @Test
-    public void testDb() {
-        QuantityType<Dimensionless> ratio = new QuantityType<>("100");
-        assertEquals("20.0 dB", ratio.toUnit("dB").toString());
     }
 
     @Test
--- a/bundles/core/org.eclipse.smarthome.core.test/src/test/java/org/eclipse/smarthome/core/types/SmartHomeUnitsTest.java
+++ b/bundles/core/org.eclipse.smarthome.core.test/src/test/java/org/eclipse/smarthome/core/types/SmartHomeUnitsTest.java
@@ -15,7 +15,7 @@ package org.eclipse.smarthome.core.types;
 import static org.eclipse.smarthome.core.library.unit.MetricPrefix.HECTO;
 import static org.hamcrest.CoreMatchers.is;
 import static org.hamcrest.number.IsCloseTo.closeTo;
-import static org.junit.Assert.assertThat;
+import static org.junit.Assert.*;
 
 import java.math.BigDecimal;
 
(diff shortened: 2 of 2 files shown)


J-N-K:
DECIBEL is adapted from [here](https://github.com/unitsofmeasurement/uom-systems/blob/master/common-java8/src/main/java/systems/uom/common/NonSI.java). However, I don't think that this line of code is trivial, so not IP relevant.

bundles/core/org.eclipse.smarthome.core/src/main/java/org/eclipse/smarthome/core/library/unit/SmartHomeUnits.java
```diff
@@ -116,6 +117,10 @@ public String getName() {
     public static final Unit<Speed> METRE_PER_SECOND = addUnit(Units.METRE_PER_SECOND);
     public static final Unit<Acceleration> METRE_PER_SQUARE_SECOND = addUnit(Units.METRE_PER_SQUARE_SECOND);
     public static final Unit<Dimensionless> PERCENT = addUnit(Units.PERCENT);
+    public static final Unit<Dimensionless> PPM = addUnit(
+            new TransformedUnit<>(ONE, new RationalConverter(1, 1000000)));
+    static final Unit<Dimensionless> DECIBEL = addUnit(
```
cweitkamp: Please make it `public`.

bundles/core/org.eclipse.smarthome.core/src/main/java/org/eclipse/smarthome/core/library/unit/SmartHomeUnits.java
```diff
@@ -116,6 +117,10 @@ public String getName() {
     public static final Unit<Speed> METRE_PER_SECOND = addUnit(Units.METRE_PER_SECOND);
     public static final Unit<Acceleration> METRE_PER_SQUARE_SECOND = addUnit(Units.METRE_PER_SQUARE_SECOND);
     public static final Unit<Dimensionless> PERCENT = addUnit(Units.PERCENT);
+    public static final Unit<Dimensionless> PPM = addUnit(
```
cweitkamp: Shouldn't we call it `PARTS_PER_MILLION` instead of `PPM`?

bundles/core/org.eclipse.smarthome.core.test/src/test/java/org/eclipse/smarthome/core/library/types/QuantityTypeTest.java
```diff
@@ -232,6 +233,18 @@ public void testDivide_Zero() {
         new QuantityType<>("4 m").divide(QuantityType.ZERO);
     }
 
+    @Test
+    public void testPpm() {
+        QuantityType<Dimensionless> ppm = new QuantityType<Dimensionless>("500 ppm");
```
cweitkamp: You can use the diamond operator here: `new QuantityType<>("500 ppm")`.

bundles/core/org.eclipse.smarthome.core.test/src/test/java/org/eclipse/smarthome/core/library/types/QuantityTypeTest.java
```diff
@@ -232,6 +233,18 @@ public void testDivide_Zero() {
...
+        assertEquals("0.05 %", ppm.toUnit(Units.PERCENT).toString());
+    }
+
+    @Test
+    public void testDb() {
+        QuantityType<Dimensionless> ratio = new QuantityType<Dimensionless>("100");
```
cweitkamp: You can use the diamond operator here: `new QuantityType<>("100")`.

bundles/core/org.eclipse.smarthome.core.test/src/test/java/org/eclipse/smarthome/core/library/types/QuantityTypeTest.java
```diff
@@ -232,6 +233,18 @@ public void testDivide_Zero() {
         new QuantityType<>("4 m").divide(QuantityType.ZERO);
     }
 
+    @Test
+    public void testPpm() {
```
htreu: The QuantityTypeTest should only test parts of the corresponding class. Please use the SmartHomeUnitsTest for conversion tests.

bundles/core/org.eclipse.smarthome.core.test/src/test/java/org/eclipse/smarthome/core/library/types/QuantityTypeTest.java
```diff
@@ -232,6 +233,18 @@ public void testDivide_Zero() {
...
+        QuantityType<Dimensionless> ppm = new QuantityType<>("500 ppm");
+        assertEquals("0.05 %", ppm.toUnit(Units.PERCENT).toString());
+    }
+
+    @Test
+    public void testDb() {
```
htreu: same here.

bundles/core/org.eclipse.smarthome.core.test/src/test/java/org/eclipse/smarthome/core/library/types/QuantityTypeTest.java
J-N-K: shall the new tests from #5336 be move there, too?

PR #5386, opened 2018-04-08
[Astro] Add channel group descriptions

### LINKED ISSUE #5310: [Astro] Many channel groups have the same description
Continued from https://github.com/openhab/openhab2-addons/issues/1520.

The Astro binding has lots of channels and groups. But unfortunately many groups are shown with the same description. 

@lolodomo: 
> Looking at the code, there is now a label but no descriptions are defined.

Here is a screenshot showing the groups and channels for a Sun thing:

astro

wborn:
Fixes #5310

Is this a proper fix for the issue @lolodomo ?


lolodomo:
Yes I think it is exactly what was expected.

PR #5389, opened 2018-04-09
Documentation: Move null annotation explanation section into separate page

triller-telekom:

### DIFF
--- a/docs/documentation/development/conventions.md
+++ b/docs/documentation/development/conventions.md
@@ -13,7 +13,8 @@ Our intention with these annotations is to transfer a method's contract written 
 We are aware that these annotations can be used for **static** checks, but **not** at runtime.
 
 Thus for publicly exposed methods that belong to our API and are (potentially) called by external callers, we cannot omit a `null` check although a method parameter is marked to be not `null` via an annotation.
-We will get a warning in the IDE for this check, but we decided to live with that. For private methods or methods in an internal package we agreed to respect the annotations and omit an additional `null` check.
+We will get a warning in the IDE for this check, but we decided to live with that.
+For private methods or methods in an internal package we agreed to respect the annotations and omit an additional `null` check.
 
 To use the annotations, every bundle should have an **optional** `Import-Package` dependency to `org.eclipse.jdt.annotation`.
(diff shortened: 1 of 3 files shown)


triller-telekom:
If this PR is fine, I will create a follow-up PR to update the openHAB documentation as well.

docs/documentation/development/conventions.md
```diff
@@ -0,0 +1,34 @@
...
+# Conventions
+
+## Null annotations
+
+[Null annotations](https://wiki.eclipse.org/JDT_Core/Null_Analysis) are used from the Eclipse JDT project.
+Our intention with these annotations is to transfer a method's contract written in its JavaDoc into the code to be processed by tools.
+We are aware that these annotations can be used for **static** checks, but **not** at runtime.
+
+Thus for publicly exposed methods that belong to our API and are (potentially) called by external callers, we cannot omit a `null` check although a method parameter is marked to be not `null` via an annotation.
+We will get a warning in the IDE for this check, but we decided to live with that. For private methods or methods in an internal package we agreed to respect the annotations and omit an additional `null` check.
```
htreu: new line for next sentence

docs/documentation/development/conventions.md
```diff
@@ -0,0 +1,34 @@
...
+@NonNullByDefault({})
+private MyService injectedService;
+```
+
+to skip the nullevaluation for these fields.
+Fields within `ThingHandler` classes that are initialized within the `initialize()` method may also be annotated like this, because the framework ensures that `initialize()` will be called before any other method.
+However please watch the scenario where the initialization of the handler fails, because then fields might not have been initialized and using them should be prepended by a `null` check.
+
+There is **no need** for a `@NonNull` annotation because it is set as default.
+Test classes do not have to be annotated.
```
htreu: Should be allow `SuppressWarnings("null")` for potential NPEs in tests then? I would mention this here then.

triller-telekom: Sounds plausible to me to allow it in tests as well. I have added it to the sentence.

kaikreuzer:
Please revise the wording, see also my comments on https://github.com/openhab/openhab-docs/pull/664#pullrequestreview-110717159.

PR #5375, opened 2018-04-07
[core] Prevent potential NPE in 'isLinked' methods

cweitkamp:
- Prevent potential NPE in `isLinked()` methods, callback can be `null`
- Added information about `IllegalStateException` to JavaDocs

Signed-off-by: Christoph Weitkamp <github@christophweitkamp.de>

PR #5391, opened 2018-04-09
Handle missing callbacks gracefully in BaseThingHandler

sjka:
...but on the other hand making sure that the handler gets the blame.

relates to #5375


maggu2810:
I like it.

PR #5393, opened 2018-04-09
added new serial bundles to p2 feature

kaikreuzer:

### DIFF
--- a/features/org.eclipse.smarthome.feature.runtime.core/feature.xml
+++ b/features/org.eclipse.smarthome.feature.runtime.core/feature.xml
@@ -172,6 +172,27 @@
          unpack="false"/>
 
    <plugin
+         id="org.eclipse.smarthome.io.transport.serial"
+         download-size="0"
+         install-size="0"
+         version="0.0.0"
+         unpack="false"/>
+
+   <plugin
+         id="org.eclipse.smarthome.io.transport.serial.javacomm"
+         download-size="0"
+         install-size="0"
+         version="0.0.0"
+         unpack="false"/>
+
+   <plugin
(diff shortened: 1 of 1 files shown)


features/org.eclipse.smarthome.feature.runtime.core/feature.xml
```diff
@@ -171,6 +171,27 @@
          version="0.0.0"
          unpack="false"/>
 
+   <plugin
+         id="org.eclipse.smarthome.io.transport.serial"
```
maggu2810: I am not such familiar with p2. Does the resolution work if there is no bundle available that satisfy the javacomm bundle requirements? Or should this be solved by the consumer of the p2 feature?

kaikreuzer: Very good question. This might indeed cause an issue for consumers, because usually target platforms should fully resolve.
What shall we do? Remove this bundle? Create a separate p2 feature for it for the ones who want to use it? Include soda in the feature?

maggu2810: > because usually target platforms should fully resolve

ESH itself does not use its own targetplatform, so no problem here.

The targetplatform of a consumer currently needs to provide a rxtx implementation, so it seems to be the same for javacomm, too.

We have dkcomm in the third party repo of ESH (github) if this one is part of the TP then all should be fine. Correct?

kaikreuzer: Right, we probably have a similar situation for other bundles like mapdb, where resolution only works if the 3rd party p2 repo is also part of the target platform.

So all should be fine then.

PR #5396, opened 2018-04-09
Initialize deviceLock

marcinshoemaker:
Initialize deviceLock to fix NullPointer on null object deviceLock when calling methods lock() and unlock()
Signed-off-by: Marcin Szewczyk <szewczykus@gmail.com>


maggu2810:
Duplicate of #5395?

szewczykus:
yes duplicate, beacuse i do not know how exactly include ECA

maggu2810:
I have checked your ECA manually and it is all fine for the moment.

```
There is a valid ECA on file for szewczykus@gmail.com.
```

AFAIK: In your commit message the `Signed-off-by` line should be separated from your detailed commit message by an empty line.

But I assume in your case the problem is that you are the "author" but not the "commit" of cf76e40797911967c6e90407835c6e9713836e55

```
commit cf76e40797911967c6e90407835c6e9713836e55
tree 5ad2ffd7b86a3d1bad4e9c6fac7e435cff49837e
parent 936287de21b7a6baabcf969899b9b62d2faac0bf
author Marcin Szewczyk <3087416+szewczykus@users.noreply.github.com> 1523301785 +0200
committer GitHub <noreply@github.com> 1523301785 +0200
```

maggu2810:
If we would mark the member variable `deviceLock` as `final` the compiler will throw an error if the variable has not been initialized. Perhaps we should add that keyword.

PR #5315, opened 2018-03-26
[5153] Discovery for USB devices with serial port

hsudbrock:
This PR targets https://github.com/eclipse/smarthome/issues/5153, i.e., a discovery for things that provide functionality of USB dongles, and where the binding uses a serial port to communicate with the USB dongle. Examples for such dongles are Zigbee, Z-Wave, HomeMatic, or HomeMatic IP USB dongles.

As discussed in https://github.com/eclipse/smarthome/issues/5153, there are different options to discover USB dongles (e.g., using a native wrapper around usblib, or, for Linux systems, scanning the sysfs, accessing dbus, or using a native wrapper around udev), as well as different options to discover the corresponding serial ports (for Linux systems, scanning the sysfs, accessing dbus, or using a native wrapper around udev).

To allow for different implementations for discovering USB devices and serial ports, this PR contains, firstly, a generic discovery for USB devices that can be accessed via a serial port (cf. bundle `org.eclipse.smarthome.config.discovery.usbserial`). This discovery can be extended from two sides: On the one side, bindings can contribute implementations of the interface `UsbSerialDiscoveryParticipant`, which, given information about serial port and USB device provided by the discovery, generate a DiscoveryResult in case they support the USB device. On the other side, different implementations of the actual scanning for USB devices and serial ports can be provided by implementing the interface `UsbSerialDiscovery`.

Secondly, this PR contains an implementation of the actual discovery (i.e., an implementation of the interface `UsbSerialDiscovery`) that can be used on Linux systems (cf. bundle `org.eclipse.smarthome.config.discovery.usbserial.linux.sysfs`). This implementation regularly scans the 'sysfs', a Linux pseudo file system that provides information about the Linux system, including information about serial ports and about USB devices.  (In https://github.com/eclipse/smarthome/issues/5153, I suggested to scan the sysfs for devices provided by the Linux usbserial driver; note that the actual implementation differs here by looking also for serial ports provided by USB devices that use another driver. This extends the set of discoverable USB dongles; an example here is the Aeotec Z-Stick for Z-Wave that uses the cdc-acm driver, and not the usbserial driver.)

I am not quite sure whether the bundle for the Linux discovery should actually go into the `bundles` subfolder, or whether this should rather go into the `extensions` subfolder of the Eclipse SmartHome project (or even somewhere else) - feedback here would be helpful for me. Of course, I would also be happy about any other feedback on this PR.

[...]

PR #5400, opened 2018-04-10
Add the bundles for discovery of usb devices providing a serial port to p2 and karaf features

hsudbrock:
The bundles `o.e.sh.config.discovery.usbserial` and `o.e.sh.config.discovery.usbserial.linuxsysfs` were recently added (in https://github.com/eclipse/smarthome/pull/5315), but they were not added to the p2 and karaf features. This PR adds them to p2 and karaf features.

I have one question for reviewers: I chose `esh-io-discovery-usb-serial` as Karaf feature name to be aligned with the existing (and somewhat similar) mDNS and UPnP discoveries, although both new bundles are in the `o.e.sh.config` namespace. Should this rather be `esh-config-discovery-usb-serial`, or left as is?

### DIFF
--- a/features/karaf/esh-core/src/main/feature/feature.xml
+++ b/features/karaf/esh-core/src/main/feature/feature.xml
@@ -330,6 +330,12 @@
     <bundle>mvn:org.eclipse.smarthome.config/org.eclipse.smarthome.config.discovery.upnp/${project.version}</bundle>
   </feature>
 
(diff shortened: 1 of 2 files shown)


kaikreuzer:
Lgtm, thanks.
Wrt the feature name, I would suggest to stay close to the package name, so `esh-config-discovery-usbserial` could be a good choice. @maggu2810, wdyt?

PR #5403, opened 2018-04-11
MqttBrokerConnection: Protocol selection. Fixes #5370

### LINKED ISSUE #5370: MQTT: Add protocol selection to MqttBrokerConnection
# Abstract
At the moment the MqttBrokerConnection class assumes that the API user wants a TCP (+SSL) connection. MQTT is also specified for Websockets nowadays though. There might be additional transport protocols specified for MQTTv5.

# Usecase
For example to talk to the amazon IOT (AWS) cloud that speaks MQTT on websockets.

# Regression
The old API just required an URI string like "tcp://1.1.1.1:1234", which defeats any means of type safety and required an extra portion of parsing, but was nicely extensible. 

# Suggestion
Add a second constructor for this probably not so often used feature of selecting another protocol. I have this in mind:
```java
enum Protocol {
   TCP,
   WEBSOCKET
};

public MqttBrokerConnection(Protocol protocol, String host, @Nullable Integer port, boolean secure, @Nullable String clientId) {
}

public MqttBrokerConnection(String host, @Nullable Integer port, boolean secure, @Nullable String clientId) {
}
```

I would not introduce a *Builder* here, because all parameters are related to each other and basically resemble an URI and are easy to grasp.

davidgraeff:


bundles/io/org.eclipse.smarthome.io.transport.mqtt/src/main/java/org/eclipse/smarthome/io/transport/mqtt/MqttBrokerConnection.java
```diff
@@ -641,7 +667,16 @@ public synchronized void start() throws MqttException, ConfigurationException {
      */
     protected MqttAsyncClient createAndConnectClient() throws MqttException, ConfigurationException {
         StringBuilder serverURI = new StringBuilder();
-        serverURI.append((secure ? "ssl://" : "tcp://"));
+        switch (protocol) {
+            case TCP:
+                serverURI.append((secure ? "ssl://" : "tcp://"));
```
maggu2810: Remove one of the braces.

bundles/io/org.eclipse.smarthome.io.transport.mqtt/src/main/java/org/eclipse/smarthome/io/transport/mqtt/MqttBrokerConnection.java
```diff
@@ -641,7 +667,16 @@ public synchronized void start() throws MqttException, ConfigurationException {
...
+                break;
+            case Websockets:
+                serverURI.append((secure ? "wss://" : "ws://"));
```
maggu2810: Remove one of the braces.

bundles/io/org.eclipse.smarthome.io.transport.mqtt/src/main/java/org/eclipse/smarthome/io/transport/mqtt/MqttBrokerConnection.java
```diff
@@ -229,6 +237,24 @@ public MqttBrokerConnection(String host, @Nullable Integer port, boolean secure,
...
+     * @param port A port or null to select the default port for a secure or insecure connection
+     * @param secure A secure connection
+     * @param clientId Client id. Each client on a MQTT server has a unique client id. Sometimes client ids are
+     *            used for access restriction implementations.
+     *            If none is specified, a default is generated. The client id cannot be longer than 65535 characters.
+     * @throws IllegalArgumentException If the client id or port is not valid.
+     */
+    public MqttBrokerConnection(Protocol protocol, String host, @Nullable Integer port, boolean secure,
+            @Nullable String clientId) {
+        this(host, port, secure, clientId);
```
maggu2810: * Make the protocol field final and remove the initialization on the declaration line.
* Move the implementation from the other constructor to this one.
* Add `this(Protocol.TCP, host, port, secure, clientId);` to the other constructor.

bundles/io/org.eclipse.smarthome.io.transport.mqtt/src/main/java/org/eclipse/smarthome/io/transport/mqtt/MqttBrokerConnection.java
```diff
@@ -77,20 +77,22 @@
...
     private @Nullable String password;
+
     /// Configuration variables
     private int qos = DEFAULT_QOS;
     private boolean retain = false;
     private @Nullable MqttWillAndTestament lastWill;
     private @Nullable AbstractReconnectStrategy reconnectStrategy;
     private SSLContextProvider sslContextProvider = new AcceptAllCertificatesSSLContext();
     private int keepAliveInterval = DEFAULT_KEEPALIVE_INTERVAL;
+    protected @Nullable Map<String, String> websocketHeaders;
```
maggu2810: There is now a field and a getter and setter to read / write that map.
But at which place this map is used?
Is an instance of that class now extended to be a "holder" for that map that is used be the consumer?

davidgraeff: paho just doesn't have the ability to set custom headers yet (https://github.com/eclipse/paho.mqtt.java/pull/514). But I only have small contribution windows and I want this to be in place and not bothered in a few weeks again. We can delay this PR of course.

maggu2810: I would prefer to delay it if the Paho version has been bumped as the holding is no useless without the bump.
If you contribution window is closed at the time of the new Paho release that contains your other PR, I assume this could also be done by another one.

davidgraeff: Yes of course. It should be a one line addition to set the custom headers, next to the paho version bump.

PR #5406, opened 2018-04-11
Fix formatting of number states and units

### LINKED ISSUE #5312: UoM: problem with percent value (humidity)
Value is not displayed in UI for the humidity channels of the WU and Yahoo weather bindings when no label is defined in item/sitemap definition. "Err" is then displayed.

```
Number:Dimensionless WUHumiditeExt <humidity> {channel="weatherunderground:weather:maison:current#relativeHumidity"}

Text item=WUHumiditeExt
```

if using `Text item=WUHumiditeExt label="Humidité [%d %%]"` then it works.

Of course, it is working for temperature or pressure for example.

The problem seems to be relative to the unit "percent" or maybe "Number:Dimensionless".

@htreu : can you have a look ?

htreu:
Fixes #5312.

### DIFF
--- a/bundles/ui/org.eclipse.smarthome.ui.test/META-INF/MANIFEST.MF
+++ b/bundles/ui/org.eclipse.smarthome.ui.test/META-INF/MANIFEST.MF
@@ -13,8 +13,8 @@ Import-Package:
  org.eclipse.jdt.annotation;resolution:=optional,
  org.eclipse.smarthome.core.library.unit,
  org.eclipse.smarthome.model.sitemap,
+ org.eclipse.smarthome.test,
  org.hamcrest;core=split,
  org.junit;version="4.0.0",
  org.mockito,
- org.mockito.stubbing,
- org.objenesis
+ org.mockito.stubbing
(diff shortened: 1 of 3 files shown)


htreu:
@lolodomo I would appreciate your feedback for this bugfix.

lolodomo:
I will try to test it this evening and think about your change in a part of the code I knew well as I already enhanced it in the past.

htreu:
@lolodomo thanks for your feedback.
Have a look in line 315, the state is now initialised with the item's state upfront. The matching condition will just overwrite it.

As for the conversion to DecimalType:
Before we did convert all states to DecimalType (except PercentType, so the value would not get converted in the 0 - 1 range) to correctly format the `%d`. Now we only convert states different from Number (which are Decimal, Percent, Quantity) because the Number states now how to format `%d`.
And to be sure formatting works we replace `%d` by `%.0f` so even non-integer decimals get formatted w/o errors.

For the group count issue there is a test case [here](https://github.com/htreu/smarthome/blob/69562d12fbc3213ff7439fcaad845fc7ab8605ac/bundles/ui/org.eclipse.smarthome.ui.test/src/test/java/org/eclipse/smarthome/ui/internal/items/ItemUIRegistryImplTest.java#L405).

lolodomo:
If you are very confident on your code changes, then let's go.
If not (the method is very complex and covers lots of cases), I can do more non regression testing.

htreu:
> Yes but this code was run only in particular conditions.

You are right! It was only done in the `if (updatedPattern != null) {` block. I will update the PR accordingly. Thanks.

lolodomo:
Last change looks ok to me.
Just one minor thing, in your test case, you are sometimes using UnitUtils.UNIT_PLACEHOLDER and sometimes %unit%.

htreu:
thanks @lolodomo, all `%unit%` occurrences replaced.

PR #5407, opened 2018-04-11
[WeMo] remove implementation of DiscoveryListener from handler. Fixes #4662

### LINKED ISSUE #4662: Wemo handlers depend on UPnP discovery
The Wemo binding uses the UPnP discovery (by implementing the DiscoveryListener interface in its handlers) in order to set things to ONLINE/OFFLINE. If I'm not mistaken then the real communication for updates and commands is done via HTTP and not via UPnP broadcast. 

Therefore its a pity that the devices _have_ to be found by UPnP, because that needlessly limits the use-cases to being within the same network segment. That e.g. prevents any docker based installations like [openhab/openhab](https://hub.docker.com/r/openhab/openhab/) and others from being able to use this binding. 

Discovery should be an optional thing. Normal operations should be completely independent of it. Or did I miss anything?

hmerk:


hmerk:
Yes, weird, but still working ....

PR #5408, opened 2018-04-12
PointType and HSBType: Trim whitespaces from constructor arguments

### LINKED ISSUE #4634: [PaperUI] Whitespace in location setting
In PaperUI Configuration>System>Regional Settings>Location I put accidently a space after the comma that separates longitude and latitude:

49.123456, 8.123456,123

There was no error message in the UI, and I searched a long time why the astro binding didn't auto discover the location.

In the log there was only one warning:
```
 [WARN ] [.core.internal.i18n.I18nProviderImpl] - Could not set new location, keeping old one:
 [INFO ] [.core.internal.i18n.I18nProviderImpl] - Locale set to de_DE, Location set to null, Time zone set to Europe/Berlin
```
But no reason why the location was not set.

Please either ignore whitespaces in the location string or give an explicit error message.

triller-telekom:
Fixes #4634

### DIFF
--- a/bundles/core/org.eclipse.smarthome.core/src/main/java/org/eclipse/smarthome/core/library/types/HSBType.java
+++ b/bundles/core/org.eclipse.smarthome.core/src/main/java/org/eclipse/smarthome/core/library/types/HSBType.java
@@ -71,9 +71,12 @@ public class HSBType extends PercentType implements ComplexType, State, Command 
         if (value != null) {
             String[] constituents = value.split(",");
             if (constituents.length == 3) {
-                this.hue = new BigDecimal(constituents[0]);
-                this.saturation = new BigDecimal(constituents[1]);
-                this.value = new BigDecimal(constituents[2]);
+                String constituent0 = constituents[0].trim();
+                String constituent1 = constituents[1].trim();
+                String constituent2 = constituents[2].trim();
+                this.hue = new BigDecimal(constituent0);
+                this.saturation = new BigDecimal(constituent1);
+                this.value = new BigDecimal(constituent2);
             } else {
                 throw new IllegalArgumentException(value + " is not a valid HSBType syntax");
             }
--- a/bundles/core/org.eclipse.smarthome.core/src/main/java/org/eclipse/smarthome/core/library/types/HSBType.java
+++ b/bundles/core/org.eclipse.smarthome.core/src/main/java/org/eclipse/smarthome/core/library/types/HSBType.java
@@ -14,8 +14,11 @@ package org.eclipse.smarthome.core.library.types;
 
 import java.math.BigDecimal;
 import java.math.RoundingMode;
+import java.util.Arrays;
+import java.util.List;
 import java.util.SortedMap;
 import java.util.TreeMap;
+import java.util.stream.Collectors;
 
 import org.eclipse.smarthome.core.types.Command;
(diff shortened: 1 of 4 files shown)


triller-telekom:
Note: Even `QuantityType` allows
```
new QuantityType<>("3km/h");
new QuantityType<>("3 km/h");
new QuantityType<>("3 km/h ");
new QuantityType<>(" 3 km/h ");
new QuantityType<>(" 3   km/h ");
```

So I think it is a good idea to also be more tolerant in these types. This way it does not matter where a (configuration) value comes from, i.e. whether its from paperUI, configadmin, etc. In all cases the value is accepted if the only difference from the accepted value is a couple of whitespaces.

WDYT?

maggu2810:
My personal meaning is that we should not try to be clever and add code at different places to workaround a non-well formatted input data.
Do we need trim, toLowerCases, perhaps some replacement of characters, ... at all places the user can put a configuration?
And then it perhaps work for the REST interface, for ConfigAdmin, but not for the Karaf console, ...
Why not just require that we get what we could work with?

triller-telekom:
> Why not just require that we get what we could work with?

I am just extending the requirement from just "x,y" to "x,y", "x, y", or "x    ,   y" etc. I.e. I would **allow** whitespaces. Same for `HSBType`.

And since this is an **extension**, everything still works as before. I do not understand why this should break anything, i.e. the Karaf console that you mention.

NorbertHD:
Could you also please trim altitude (parameter 3)?
```
  if (elements.length == 3) {
      String element2 = elements[2].trim();
      setAltitude(new DecimalType(elements2));
  } else if (elements.length > 3) {
```

triller-telekom:
@NorbertHD Thanks for reminding me of the altitude argument! I have added a `trim()` for that as well.

I have also created tests for the `HSBType` and `PointType` constructors with `String`.

bundles/core/org.eclipse.smarthome.core/src/main/java/org/eclipse/smarthome/core/library/types/HSBType.java
```diff
@@ -71,9 +71,12 @@ public HSBType(String value) {
         if (value != null) {
             String[] constituents = value.split(",");
```
maggu2810: As every part should be trimmed, why not using something like this?

```Java
List<String> constituents = Arrays.stream(value.split(",")).map(in -> in.trim()).collect(Collectors.toList());
```

triller-telekom: Alright, I have changed it to make use of the stream API.

triller-telekom: Makes sense. Done.

NorbertHD:
Why the helper variables element0-2? And element2 has a second trim.
How About:
``
            if (elements.size() >= 2) { 
                canonicalize(new DecimalType(elements.get(0)), new DecimalType(elements.get(1))); 
                 if (elements.size() == 3) { 
                     setAltitude(new DecimalType(elements.get(2))); 
                 } else if (elements.size() > 3) {

triller-telekom:
@NorbertHD: Good catch. Thanks.

PR #5260, opened 2018-03-19
suppres PMD warning 'AvoidCatchingThrowable' for some classes

lpapazow:
[This PMD check](https://pmd.github.io/pmd-5.8.1/pmd-java/rules/java/strictexception.html#AvoidCatchingThrowable) that is used in the [static analysis tool](https://github.com/openhab/static-code-analysis) triggers warning whenever `Throwable` errors are caught in the code.

After using git blame I found that in all of these cases there is a specific reason to use throwable.

In [HueBridgeHandler](https://github.com/eclipse/smarthome/blob/master/extensions/binding/org.eclipse.smarthome.binding.hue/src/main/java/org/eclipse/smarthome/binding/hue/handler/HueBridgeHandler.java#L160) the code handles catching Throwable and catching Exception differently.
The `ScriptInterpreter.java` class is generated by xtend and in [AbstractRegistry](https://github.com/eclipse/smarthome/commit/38ac359dce) and in [ProxyServletService](https://github.com/eclipse/smarthome/commit/a0e98012a1) Exception was changed to Throwable for a specific reason.

However, there are 3 more classes in which `catch (Throwable t)` is used: 
RuleEngine, AutomationResourceBundleEventQueue and ScriptImpl.

@danchom, I saw that you are the author of the [RuleEngine](https://github.com/eclipse/smarthome/commit/3d2b6e6453#diff-d496a3945090abb3b9b346c925d42370R946) class. I was wondering if there is a specific reason why `catch Throwable` is preferred over `catch Exception` [here](https://github.com/eclipse/smarthome/commit/3d2b6e6453#diff-d496a3945090abb3b9b346c925d42370R946)?

@adimova, I saw that you are the author of the [AutomationResourceBundlesEventQueue](https://github.com/eclipse/smarthome/commit/3d2b6e6453#diff-2f5febc89267b77022ae6c5890218c5bR138) class. I was wondering if there is any specific reason why `catch Throwable` is preferred over `catch Exception` [here](https://github.com/eclipse/smarthome/commit/3d2b6e6453#diff-2f5febc89267b77022ae6c5890218c5bR138)?

[...]

PR #5409, opened 2018-04-12
review comments for jetty client factory

sobimibos:

### DIFF
--- a/bundles/io/org.eclipse.smarthome.io.net/src/main/java/org/eclipse/smarthome/io/net/http/internal/SecureHttpClientFactory.java
+++ b/bundles/io/org.eclipse.smarthome.io.net/src/main/java/org/eclipse/smarthome/io/net/http/internal/SecureHttpClientFactory.java
@@ -62,8 +62,8 @@ public class SecureHttpClientFactory implements HttpClientFactory {
 
     private volatile TrustManagerProvider trustmanagerProvider;
 
-    private QueuedThreadPool threadPool = null;
-    private HttpClient commonHttpClient = null;
+    private QueuedThreadPool threadPool;
+    private HttpClient commonHttpClient;
 
     private int minThreadsShared;
     private int maxThreadsShared;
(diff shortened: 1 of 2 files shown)


bundles/io/org.eclipse.smarthome.io.net/src/main/java/org/eclipse/smarthome/io/net/http/internal/SecureHttpClientFactory.java
```diff
@@ -89,16 +89,16 @@ protected void modified(Map<String, Object> parameters) {
...
-        try {
-            if (commonHttpClient != null) {
+        if (commonHttpClient != null) {
+            try {
                 commonHttpClient.stop();
-                commonHttpClient = null;
-                threadPool = null;
-                logger.info("jetty shared http client stopped");
+            } catch (Exception e) {
+                logger.error("error while stopping shared jetty http client", e);
```
martinvw: This could be a debug log, its gone anyway

kaikreuzer:
> I don't know the position of maintainers in this? @maggu2810 @kaikreuzer @htreu

@martinvw I assume you refer to [this line](https://github.com/eclipse/smarthome/pull/5409/files#diff-9013c676a0557e3282f0996ed06ded62R269)? Afaics, we can come across `KeyManagementException` and `NoSuchAlgorithmException` exceptions at that place, both catched exceptions. Both extend `GeneralSecurityException` and I would actually claim that instead of catching those and re-throwing an (always unexpected) runtime exception, it would be better to add a `throws GeneralSecurityException` to the method itself.

martinvw:
That is indeed a viable solution, thanks.

But my question was even meant more generic, what is the opinion of the ESH maintainers with respect to catching and throwing raw-exceptions types. Just to be sure before I ever bump into a merged PR again 😇

kaikreuzer:
> what is the opinion of the ESH maintainers with respect to catching and throwing raw-exceptions types

Speaking just for myself, I would say that it should be avoided if possible, but in exceptional cases, it can also be accepted as a pragmatic solution. Wrt Throwable, we definitely do not accept catching those (see https://github.com/eclipse/smarthome/pull/5260).

sobimibos:
I added a custom runtime exception for all possible errors during the initialization of the jetty client.

martinvw:
LGTM

PR #5410, opened 2018-04-12
Fix thing creation for nested Things from different ThingHandlerFactory

### LINKED ISSUE #5404: Bridge nested syntax does not support different thingHandlerFactory for thing
This issue is a follow up from #5301 and refers to a somewhat experimental technology at the moment.

Given are two bundles that share the same binding id. Reasons for this are described in #5301 A prototype for this is the [blukii support](https://github.com/eclipse/smarthome/tree/master/extensions/binding/org.eclipse.smarthome.binding.bluetooth.blukii) or the modbus based [sunspec support](https://github.com/mrbig/openhab2-addons/tree/3216-binding-sunspec-modbus/addons/binding/org.openhab.binding.modbus.sunspec)

The issue arises when the bridge and the connected thing are from different bundles, and the nested syntax is used:

```
Bridge modbus:tcp:se4000 [ ... ] {
    inverter-single-phase myinverter [ ... ]
}
```

Currently these items are silently ignored, because the GenericThingProvider checks the proposed thing with the bridge's ThingHandlerFactory only. [See here](https://github.com/eclipse/smarthome/blob/5528ad5e460e2d8b6412bc34610530c0f2f06a28/bundles/model/org.eclipse.smarthome.model.thing/src/org/eclipse/smarthome/model/thing/internal/GenericThingProvider.xtend#L232)

I propose, that we should rescan every thingHandlerFactory each time the code is recursing into nested bridges/things. To avoid incorrect connection between different bindings may the binding id could been checked and if they don't match then ignore the thingHandlerFactory.

Note: a workaround for this is to ignore the nested syntax in such cases and use flat list with references, so the above could be rewritten as:
```
Bridge modbus:tcp:se4000 [ ... ]
Thing modbus:inverter-single-phase:se4000:myinverter (modbus:tcp:se4000) [...]
```

htreu:
Fixes #5404.

The first commit is only a reorganisation of the tests into sane packages.


bundles/model/org.eclipse.smarthome.model.thing/src/org/eclipse/smarthome/model/thing/internal/GenericThingProvider.xtend
```diff
@@ -160,23 +161,31 @@ class GenericThingProvider extends AbstractProvider<Thing> implements ThingProvi
...
+    
+    def private Iterable<ModelThing> flattenModelThings(Iterable<ModelThing> things) {
+        if (things === null || things.length === 0) {
+            return #[]
+        }
+        
+        // construct the thingUID up front: this will ensure that short notation things with reference to a bridge 
+        // do not end up with the bridge id in their id. 
+        things.forEach([thingId = thingId ?: constructThingUID.toString])
+        things.filter(typeof(ModelBridge)).forEach([val b = it; b.things.forEach[
```
sjka: ~~`val b = it;` -> `b |`~~

Nevermind, all correct. Still, here you got me - how about giving _both_  `it`s some explicit names?

bundles/model/org.eclipse.smarthome.model.thing/src/org/eclipse/smarthome/model/thing/internal/GenericThingProvider.xtend
```diff
@@ -160,23 +161,31 @@ class GenericThingProvider extends AbstractProvider<Thing> implements ThingProvi
...
+            bridgeUID = b.id
+            // for long notation (id is already set) leave it like this.
+            // for nested things in short notation, make sure the bridge id is part of the thing id.
+            id = id ?: getThingUID(new ThingUID(b.id)).toString;
+        ]])
+        things + flattenModelThings(things.filter(typeof(ModelBridge)).map(b | b.things).flatten);
```
sjka: just personal preference, but here I think a `return` would help readability

bundles/model/org.eclipse.smarthome.model.thing.tests/src/test/groovy/org/eclipse/smarthome/model/thing/test/hue/GenericThingProviderTest2.groovy
```diff
@@ -0,0 +1,99 @@
...
+        ] as ComponentContext
+
+        def hueThingHandlerFactory = new TestHueThingHandlerFactoryX(componentContextMock)
+
+        def things = thingRegistry.getAll()
+        assertThat things.size(), is(0)
+
+        String model =
+                '''
+			Bridge Xhue:Xbridge:myBridge [ XipAddress = "1.2.3.4", XuserName = "123" ] {
```
sjka: Aaaaaahhh, TABS!!! 😱

htreu: auto-formatter refuses to make all good, sorry :-(

bundles/model/org.eclipse.smarthome.model.thing/src/org/eclipse/smarthome/model/thing/internal/util/BundleNameResolver.java
```diff
@@ -0,0 +1,6 @@
+package org.eclipse.smarthome.model.thing.internal.util;
+
+public interface BundleNameResolver {
```
sjka: Why do you need this additional interface? Wouldn't a making the `getBundleName(...)` method protected be sufficient, so that it can be overridden in the test-case by like this?

```java
new BundleNameResolver() {
    @Override
    protected String getBundleName(...) {
        // ...
    }
}
```

htreu: done. I didn't like the solution either.

PR #5411, opened 2018-04-13
Convert Windows to Unix line endings

wborn:
Most of the code is using Unix line endings. Though some files are using Windows line endings. 
You can find these files using the command in https://github.com/openhab/static-code-analysis/issues/273. 
Similar to https://github.com/openhab/openhab2-addons/pull/3475, I've used the output of the command with dos2unix and committed the result in this PR.


wborn:
> Do you know whether this interferes with other currently open PRs or is git clever enough to auto-merge thus changes and not consider them as a conflict?

It will result in merge conflicts. There is a [discussion](https://stackoverflow.com/questions/861995/is-it-possible-for-git-merge-to-ignore-line-ending-differences) on Stack Overflow how it should be possible to resolve such merge conflicts. I haven't tested this myself though.

I can have a look at the open PRs and revert changes to files that would result in conflicts.

> If git is oblivious about it, what pain does that change actually address (why do you consider it to be important)?

* If it's not addressed with a PR like this one, you could continuously see these conversion being made in PRs which adds noise to reviews.
* It stops the proliferation of Windows line endings due to copy/pasting and editing files with Windows line endings.
* It makes it easier for tools/scripts to process source code files when they only have to deal with one kind of line ending.

> How can we avoid getting new files with Windows line endings in the repo? Should we add a .gitattributes file right away?

That's the topic of https://github.com/openhab/static-code-analysis/issues/273. I haven't used a .gitattributes file with Git repositories myself. So I'm not familiar with what issues could arise when you start using one. It would be helpful if people can comment pros/cons on https://github.com/openhab/static-code-analysis/issues/273 before we decide on how to deal with this.

According to GitHub the .gitattributes file is the way to address it, see: https://help.github.com/articles/dealing-with-line-endings/#platform-all

kaikreuzer:
Ok, but then I would suggest to first find a solution to the problem that Windows line ending actually get into the repo in the first place, before merging this PR that erases all those line endings that are already in - because otherwise, we will have to do the same over and over again.

So let's continue first on  openhab/static-code-analysis#273.

kaikreuzer:
Please allow me to close this PR for the moment - once we have finished the discussion on the SAT issue, we can update/reopen it.

PR #5414, opened 2018-04-13
Set suppression for XML files

tanyatanyatanya:
@wborn As you suggested I added a suppression filter for the generated XML files from DS annotations for 
NewlineAtEndOfFileCheck and OnlyTabIndentationInXmlFilesCheck.

Signed-off-by: Tanya D. Georgieva <tanya.dancheva.georgieva@gmail.com>

### DIFF
--- a/tools/static-code-analysis/checkstyle/suppressions.xml
+++ b/tools/static-code-analysis/checkstyle/suppressions.xml
@@ -8,19 +8,21 @@
     <suppress files=".+[\\/]internal[\\/].+\.java" checks="JavadocType|JavadocVariable|JavadocMethod|JavadocFilterCheck"/>
     <suppress files=".+DTO\.java" checks="JavadocType|JavadocVariable|JavadocMethod|JavadocFilterCheck" />
     <suppress files=".+Impl\.java" checks="JavadocType|JavadocVariable|JavadocMethod|JavadocFilterCheck"/>
-	<suppress files=".+\\pom\.xml" checks="OnlyTabIdentationInXmlFilesCheck"/>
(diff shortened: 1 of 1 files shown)


tools/static-code-analysis/checkstyle/suppressions.xml
```diff
@@ -8,19 +8,21 @@
     <suppress files=".+[\\/]internal[\\/].+\.java" checks="JavadocType|JavadocVariable|JavadocMethod|JavadocFilterCheck"/>
     <suppress files=".+DTO\.java" checks="JavadocType|JavadocVariable|JavadocMethod|JavadocFilterCheck" />
     <suppress files=".+Impl\.java" checks="JavadocType|JavadocVariable|JavadocMethod|JavadocFilterCheck"/>
-	<suppress files=".+\\pom\.xml" checks="OnlyTabIdentationInXmlFilesCheck"/>
+    <suppress files=".+\\pom\.xml" checks="OnlyTabIndentationInXmlFilesCheck"/>
```
wborn: Since you are touching this line you might as well also apply this fix from https://github.com/openhab/static-code-analysis/pull/289 here to remove these false positives for people with Unix paths.

tanyatanyatanya: @wborn Done.

wborn:
The suppression will not work for the DS component XML files generated for DmxHandlerFactory and DmxTestHandlerFactory because these files have a `@Component` annotation with a `name` property. When the `name` property is removed on both, the generated files will also match the suppression filter.

PR #5417, opened 2018-04-14
[Core] Avoid sending update notification for a chart widget having a refresh argument

### LINKED ISSUE #5062: Chart refresh
Chart widget in sitemap has a refresh parameter but any UI will be notified of a change each time the linked item is updated. As a consequence, the UIs will refresh the chart each time the value in the chart is updated. It could be a problem in case the item value is changed frequently. And the final user only expects a refresh when the delay set in the sitemap refresh parameter expires.

What do you thing about that ? What could be the solution to implement ?

https://github.com/openhab/openhab-android/issues/217

lolodomo:
Fixes #5062

### DIFF
--- a/bundles/io/org.eclipse.smarthome.io.rest.sitemap/src/main/java/org/eclipse/smarthome/io/rest/sitemap/internal/PageChangeListener.java
+++ b/bundles/io/org.eclipse.smarthome.io.rest.sitemap/src/main/java/org/eclipse/smarthome/io/rest/sitemap/internal/PageChangeListener.java
@@ -29,6 +29,7 @@ import org.eclipse.smarthome.core.library.CoreItemFactory;
 import org.eclipse.smarthome.core.types.State;
 import org.eclipse.smarthome.io.rest.core.item.EnrichedItemDTOMapper;
 import org.eclipse.smarthome.io.rest.sitemap.SitemapSubscriptionService.SitemapSubscriptionCallback;
+import org.eclipse.smarthome.model.sitemap.Chart;
 import org.eclipse.smarthome.model.sitemap.ColorArray;
 import org.eclipse.smarthome.model.sitemap.Frame;
 import org.eclipse.smarthome.model.sitemap.VisibilityRule;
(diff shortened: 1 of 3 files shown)


lolodomo:
Please note that this PR should fix any UI including the Classic UI which has its specific way to listen for item updates.
Of course, this PR is fixing the Basic UI in Firefox/Chrome (SSE) and in Microsoft Edge (long polling REST call).

bundles/io/org.eclipse.smarthome.io.rest.sitemap/src/main/java/org/eclipse/smarthome/io/rest/sitemap/internal/PageChangeListener.java
```diff
@@ -191,30 +192,38 @@ public void stateUpdated(Item item, State state) {
...
-                if (event.state != null && event.state.equals(event.item.state)) {
-                    event.state = null;
+            if (w.getItem() != null && w.getItem().equals(item.getName())) {
+                // We skip the chart widgets having a refresh argument
+                boolean skipWidget = false;
+                if (w instanceof Chart) {
+                    Chart chartWidget = (Chart) w;
+                    skipWidget = chartWidget.getRefresh() > 0;
+                }
+                if (!skipWidget || definesVisibility(w, item.getName())) {
```
htreu: maybe the visibility for a chart widget should also be checked in the chart specific block (lines 198 - 201)? Otherwise the chart widget will receive frequent updates just by defining a visibility rule.

lolodomo: Yes, an item change referenced in a visibility rule will update the chart as before.
But what is exactly your use case ? The most common known use case is a user selection to select between different charts and for example different periods for the chart. And so this is not something that changes very often but only when the user decides it. So that looks fine to me like it is now.

htreu: okay, I will try to rephrase my question:
Imagine three charts, all part of the user selection you mentioned. Every chart displays a different period of a very frequent changing item (lets say the item gets updates once every 10 seconds). Each chart has a refresh of 3600 defined, so the UI should refresh it only once in an hour.
Now with the `!skipCheck || definesVisibility(...)` the charts will get updated once every 10 seconds because they do define visibility, regardless of their refresh interval defined.
My suggestion is to also check the visibility in the chart specific block, so that it would receive an update only if its visibility changes. Is that possible?

lolodomo: This could be a problem only in the case the visibility of the chart depends on the value of a data inside the chart.  I don't imagine this use case in practice. For example, you will not display the chart when the temperature displayed in the chart is lower than a value ??? The most common case is to link the visibility to another item, a user selection.

By the way, you pointed something interesting, we are notifying the client about a change of the item involved in the visibility attribute and not about the change of visibility itself. It is the client that checks the visibility. I am not sure we can get at this place the previous visibility. But this enhancement would be a global enhancement to all widgets and not only chart widget. It is not the object of my current PR.

lolodomo:
Any chance to make progress on this PR?

bundles/io/org.eclipse.smarthome.io.rest.sitemap/src/main/java/org/eclipse/smarthome/io/rest/sitemap/internal/PageChangeListener.java
lolodomo: The chart will be refreshed in case the item involved in the visibility condition is updated. That is of course expected. You don't want to wait for 1 minute when you decide to hide/show a chart. You expect an immediate impact in UI.

Consider this usual example from the demo sitemap:
https://github.com/openhab/openhab-distro/blob/master/features/distro-resources/src/main/resources/sitemaps/demo.sitemap#L17
definesVisibility(w, item.getName()) will be true only when the state of the item Weather_Chart_Period is updated but will be false when any item incliuded in the chart is updated. That means it will be true only when the user change its selection between hour / day / week using the UI.

What you mention would be true in case the visibility condition would be based on one item displayed in the chart, for example the item Weather_Temp_Min. So what you say is true in theory but I still can't imagine a practical use case where the user will define a visibilioty condition based on what is displayed in the chart. Can you please explain a not stupid use case covering your case ?

lolodomo: And again, definesVisibility triggers an update each time the item involved in the condition is updated. This should be replaced, if possible, by a trigger only when the visibility condition state is updated (show to hide or hide to show). But this is another debate, global to all widgets and not only the chart widget. I can try to address this other enhancement in another PR.

lolodomo: And that's being said, I am surprised that there is no check for labelcolor and valuecolor conditions at this place (sitemap SSE) but only the visibility condition is handled. That should be a bug I have first to reproduce in Basic UI.

lolodomo:
I just rebased my PR.

bundles/io/org.eclipse.smarthome.io.rest.sitemap/src/main/java/org/eclipse/smarthome/io/rest/sitemap/internal/PageChangeListener.java
htreu: Thanks for the detailed explanation and your patience. You are right, I overlooked the fact that `definesVisibility` takes the item name into account.

PR #5419, opened 2018-04-16
Port MapDbPersistenceService

mkhl:
This changeset imports MapDbPersistenceService from opanhab1-addons, ports it to Eclipse SmartHome, and uses GSON instead of a custom MapDBSerializer.


htreu:
Hi @mkhl, thanks a lot for this contribution.
A first scan of the code already made a good impression to me, please expect an in-depth review soon.
btw: Are you able to provide some basic test cases for serialisation/deserialisation? Especially for the former custom state conversion it would be nice to see how it is handled now.

htreu:
Travis complains about a null type missmatch:

```
[ERROR] /home/travis/build/eclipse/smarthome/bundles/persistence/org.eclipse.smarthome.persistence.mapdb/src/main/java/org/eclipse/smarthome/persistence/mapdb/internal/MapDbPersistenceService.java:[152] 
[ERROR] 	store(item, null);
[ERROR] 	            ^^^^
[ERROR] Null type mismatch: required '@NonNull String' but the provided value is null
```

Do you use the latest official IDE with null-annotation checks? This error should also be visible within eclipse.

bundles/persistence/org.eclipse.smarthome.persistence.mapdb/OSGI-INF/mapdb.xml
```diff
@@ -0,0 +1,17 @@
+<?xml version="1.0" encoding="UTF-8"?>
```
htreu: This file may be removed completely and replaced by annotations. See my comment on MapDbPersistenceService.

bundles/persistence/org.eclipse.smarthome.persistence.mapdb/src/main/java/org/eclipse/smarthome/persistence/mapdb/internal/MapDbItem.java
```diff
@@ -0,0 +1,80 @@
+/**
+ * Copyright (c) 2010-2016 by the respective copyright holders.
```
htreu: Please update all license herders to EPL 2. Run the command 

```
mvn license:format -pl bundles/persistence/org.eclipse.smarthome.persistence.mapdb
```

from the project root directory to have this done automatically.

bundles/persistence/org.eclipse.smarthome.persistence.mapdb/src/main/java/org/eclipse/smarthome/persistence/mapdb/internal/MapDbPersistenceService.java
```diff
@@ -0,0 +1,276 @@
...
+
+/**
+ * This is the implementation of the MapDB {@link PersistenceService}. To learn
+ * more about MapDB please visit their <a
+ * href="http://www.mapdb.org/">website</a>.
+ *
+ * @author Jens Viebig
+ * @since 1.7.0
+ */
+public class MapDbPersistenceService implements QueryablePersistenceService {
```
htreu: please annotate as `@Component(service = PersistenceService.class)` to automatically generate the service XML file.

bundles/persistence/org.eclipse.smarthome.persistence.mapdb/src/main/java/org/eclipse/smarthome/persistence/mapdb/internal/MapDbItem.java
```diff
@@ -32,29 +35,29 @@
     private Date timestamp;
 
     @Override
-    public String getName() {
+    public @NonNull String getName() {
```
htreu: Instead of `@NonNull` please use `@NonNullByDefault` on class level and then `@Nullable` on the exceptions from the default. Read this part of the documentation for reasons: https://www.eclipse.org/smarthome/documentation/development/guidelines.html#a-code-style

htreu:
Also: Travis now fails with

```
[ERROR] Failed to execute goal org.eclipse.tycho:tycho-packaging-plugin:1.1.0:package-plugin (default-package-plugin) on project org.eclipse.smarthome.persistence.mapdb: /home/travis/build/eclipse/smarthome/bundles/persistence/org.eclipse.smarthome.persistence.mapdb/build.properties: bin.includes value(s) [OSGI-INF/] do not match any files. -> [Help 1]
```

Please create the folder `OSGI-INF` and place a `.gitignore` file with content `*.xml`. This way the folder exists in git and the generated service XML will not be treated by git.

mkhl:
That could look a bit like this: https://github.com/mkhl/smarthome/compare/persistence/mapdb...mkhl:persistence/mapdb-null

htreu:
Since we write the database ourselves and "know" that the values for `name`, `state` and `timestamp` have to be present its okay to log an error or throw an exception if the deserialised entry is null or has null values. We know that something went wring then.

htreu:
When GSON deserialises `MapDBItem` the annotations will not matter. Its only for tool support (IDE and build script).
So maybe we should sanity check the output from GSON and then throw an exception if fields are still null. You would then have to suppress the "redundant null check" warnings. In the end the persistence service should return only valid entires.

second thought: maybe its better to only log an error and return the empty collection instead of throwing exceptions.

maggu2810:
> Besides from that, this persistence service is only used to store the latest item value. Since it would be the first persistence implementation we have to discuss if this fits.

This is not written.
A persistence service should provide a method to store an item.
How the persistence service manages the stored items is out of scope and can be vary by the persistence service implementations.
E.g. store one item per item name, store only one item at all, store the last x items per name, store all and never delete anything, aggregate older values, ...

kaikreuzer:
> This is not written.

That's why @htreu suggested to discuss it.
I agree with him that it would be nice to have a persistence service implementation in ESH, which can be used for the "typical" features, which is time series data for charts and rules.
Using a fully-fledged database implementation like MapDB just for storing a single value per item feels a bit like "mit Kanonen auf Spatzen" to me... This could be easily done by a small text file as well.

bundles/persistence/org.eclipse.smarthome.persistence.mapdb/src/main/java/org/eclipse/smarthome/persistence/mapdb/internal/MapDbPersistenceService.java
```diff
@@ -0,0 +1,287 @@
...
+
+        logger.debug("store called for {}", alias);
+
+        State state = item.getState();
+        MapDbItem mItem = new MapDbItem();
+        mItem.setName(alias);
+        mItem.setState(state);
+        mItem.setTimestamp(new Date());
+        String json = serialize(mItem);
+        String oldJson = map.put(alias, json);
```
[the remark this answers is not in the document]
mkhl: With an `ExecutorService` we could `submit` a commit task for every `store`.

maggu2810:
Just to be clear: What should be discussed?
The persistence service interface or this specific service implementation?

I agree that there should be a persistence service implementation in ESH that stores a history / time series and not only the last value.

I don't think the persistence service interface should define that their implementations need to implement the features for time series data.
Perhaps someone would like to use a "RestoreOnStartupPersistenceService" and only store the last one.

So, an implementation that will be part of ESH should demonstrate and contain the typical features but the interface should be kept as it is.

bundles/persistence/org.eclipse.smarthome.persistence.mapdb/META-INF/MANIFEST.MF
```diff
@@ -0,0 +1,26 @@
...
+ org.eclipse.smarthome.config.core,
+ org.eclipse.smarthome.core.common,
+ org.eclipse.smarthome.core.items,
+ org.eclipse.smarthome.core.library.items,
+ org.eclipse.smarthome.core.library.types,
+ org.eclipse.smarthome.core.persistence,
+ org.eclipse.smarthome.core.types,
+ org.mapdb,
+ org.osgi.framework,
+ org.quartz,
```
htreu: please reomve the now unused quartz imports.

bundles/persistence/org.eclipse.smarthome.persistence.mapdb/src/main/java/org/eclipse/smarthome/persistence/mapdb/internal/MapDbPersistenceService.java
```diff
@@ -0,0 +1,176 @@
...
+    public void deactivate() {
+        logger.debug("MapDB persistence service deactivated");
+        if (db != null) {
+            db.close();
+        }
+        threadPool.shutdown();
+    }
+
+    @Override
+    public @NonNull String getId() {
```
htreu: please use `@NonNullByDefault` on class level and then use `@Nullabel` accordingly.

bundles/persistence/org.eclipse.smarthome.persistence.mapdb/src/main/java/org/eclipse/smarthome/persistence/mapdb/internal/MapDbPersistenceService.java
```diff
@@ -0,0 +1,188 @@
...
+
+    public void activate() {
+        logger.debug("MapDB persistence service is being activated");
+
+        threadPool = ThreadPoolManager.getPool(getClass().getSimpleName());
+
+        File folder = new File(DB_FOLDER_NAME);
+        if (!folder.exists()) {
+            if (!folder.mkdirs()) {
+              logger.error("Failed to create one or more directories in the path '{}'", DB_FOLDER_NAME);
```
htreu: Please only log warnings here. `error` is reserved for fatal framework errors.

mkhl:
> Can I further motivate you to also provide some test cases?

I understand the sentiment, and I’ll try to find the time to figure out how to and to add them, but I can’t make promises on the timeframe, and I hope that the tests are not an obstacle to accepting this.

mkhl:
I’ve written tests that ensure StateTypeAdapter roundtrips locally.
Testing MapDbPersistenceService is giving me trouble though, mostly since it doesn’t include a way to clear the persisted Items.
I _could_ write a single long test that tries to cover everything and then delete the userdata folder on exit, but that seems like a bad idea…
Any advice you can offer?

htreu:
@mkhl thanks a lot for taking the tome and providing some tests! For the `MapDbPersistenceService` you can take a look at the `TemporaryFolder` JUnit rule.
You define a field inside your test class like

```
@Rule
public TemporaryFolder tmpFolder = new TemporaryFolder();
```

and JUnit will take care of providing the `tmpFolder` and also remove it when the test terminates.

htreu:
Oh, and I just realized you did groovy tests. We decided some time ago to not add more groovy but migrate all existing tests to java. Could you also migrate your tests to Java please?

The move to pure Java tests is also part of the documentation: https://www.eclipse.org/smarthome/documentation/development/testing.html#groovy---deprecated

mkhl:
@htreu Alright, thanks.

The conversion to Java itself was easy enough, but I’m having trouble seeing how I can make MapDbPersistenceStorage use the TemporaryDirectory to store its MapDB.
Then I tried the route that StorageServiceOSGiTest took: [removing the `userdata` directory in `tearDown`](https://github.com/eclipse/smarthome/blob/master/bundles/storage/org.eclipse.smarthome.storage.mapdb.test/src/main/java/org/eclipse/smarthome/storage/mapdb/internal/StorageServiceOSGiTest.java#L97-L107), but since now the test order is random and unregistering/reregistering the PersistenceService in between tests does not seem to deactivate/reactivate it, the Map isn’t clean and the tests fail.

I’ll try to get the tests running inside Eclipse now so I can attach a debugger and try to figure out what to do next, but I’ll happily take any suggestions ;-)

htreu:
You could try to implement a MapDbPersistenceServiceTest which is _not_ an OSGiTest. This would make things a little more easy I guess. Just instantiate a new instance if the service and call the activate method. For the temporary test folder you could provide a property in the configuration. This would also allow a solution to modify the storage folder. In case the property is not set you could fall back to the static defined folder name. wdyt?

mkhl:
Thanks, I managed to get it working as an OSGiTest.

bundles/persistence/org.eclipse.smarthome.persistence.mapdb/src/main/java/org/eclipse/smarthome/persistence/mapdb/internal/MapDbPersistenceService.java
```diff
@@ -0,0 +1,186 @@
...
+        logger.debug("store called for {}", alias);
+
+        State state = item.getState();
+        MapDbItem mItem = new MapDbItem();
+        mItem.setName(alias);
+        mItem.setState(state);
+        mItem.setTimestamp(new Date());
+        String json = serialize(mItem);
+        map.put(alias, json);
+        commit();
```
wborn: In the openhab1-addons implementation there is at most 1 commit of the database every 5 seconds. I don't know the exact reasons for this but it could be an expensive operation and committing every change may also cause significant wear and tear on SD cards. Did you look into this and are there thread safety issues when there is are multiple threads in the pool?

mkhl: No I have not looked into this, the current way was suggested [above](https://github.com/eclipse/smarthome/pull/5419#discussion_r185429583), but I can see how there is a tradeoff between safety and storage wear here.

Looking at the MapDB code I don’t see any thread safety issues.

sjka:
I'm just trying to understand the current state of this PR. 

> Using a fully-fledged database implementation like MapDB just for storing a single value per item feels a bit like "mit Kanonen auf Spatzen" to me... This could be easily done by a small text file as well.

I fully agree - sounds a pretty much overkill. To me it doesn't make much sense like that.

> I agree that there should be a persistence service implementation in ESH that stores a history / time series and not only the last value.

In that context however it would make sense to take the current state and extend it to also support historic data. MapDB is capable of that, right? Would you agree?

mkhl:
> I'm just trying to understand the current state of this PR.

I believe I’ve incorporated all of @htreu’s review feedback.

@wborn raised another point about the tradeoffs between latency versus throughput of storage operations that I couldn’t address (as noted above, @htreu seemed to lean on the side of minimising latency).
@pwolfart suggested that I could reuse (or duplicate) the MapDB configuration parameters from MapDbStorageService since he’s been very happy with them on the SD cards we use, so I’ll look into that.

> > Using a fully-fledged database implementation like MapDB just for storing a single value per item feels a bit like "mit Kanonen auf Spatzen" to me... This could be easily done by a small text file as well.

> I fully agree - sounds a pretty much overkill. To me it doesn't make much sense like that.

I disagree. We want a key-value store here, and MapDB provides a very capable one.
I believe that chasing the short-term simplicity of text files here would lead to reimplementation of at least indexing and transactions sooner rather than later.

> > I agree that there should be a persistence service implementation in ESH that stores a history / time series and not only the last value.

> In that context however it would make sense to take the current state and extend it to also support historic data. MapDB is capable of that, right? Would you agree?

I’m not sure who you mean by “you”, but for what it’s worth:

1.	MapDB provides a key-value datastore that we can use like a `Map` and that we can persist to a file.
	Of course we _can_ use it like a `Multimap` (a `Map` whose values are `Collection`s), with all the drawbacks that come with using a `Map` like a `Multimap` in Java, but I don’t think that it’s obvious that we _should_.
	My suggestion is to let this pull request stand on its own and extract discussion about historic data into its own issue or pull request.

2.	This PR was supposed to be a straightforward port of the old MapDbPersistenceService from openHAB to Eclipse SmartHome.
	That’s what we currently have a use case for at @AlJungOrg. We do not currently have a use case for historic data.
	I’ve been happy to update the code to current ESH standards and add tests but I consider the extension to historic data out of scope.

mkhl:
> @pwolfart suggested that I could reuse (or duplicate) the MapDB configuration parameters from MapDbStorageService since he’s been very happy with them on the SD cards we use, so I’ll look into that.

`MapDbStorage` already `commit`s after _every_ `put` and `remove` operation so I’ll leave this as is for now.

kaikreuzer:
> We want a key-value store here, and MapDB provides a very capable one.

I guess you are right - I so far had MapDB as a fully-fledged database in mind, but it is rather meant as a persistable Java collections implementation. As such, it makes sense to use it for simple key value pairs and supporting time-series is not necessarily something that comes naturally with MapDB.

One concern that I have with this PR in general and which I would like to align with you (and the @eclipse/eclipse-smarthome committers) is the fact that I consider the current persistence API as legacy from openHAB 1 times, which would need to be looked at and potentially extended/adapted to nicely fit into the current ESH concepts. We so far do not have any "persistence extension" here in the repo, no documentation (or even a skeleton) for them, no coverage in the Paper UI, etc. Especially the support in the Paper UI might require changes to the API and e.g. additional UI-related meta-data to be provided in some way. I so far have no clue, how much impact that might have on the current API, but I think this needs to be looked at and we all should be aware that things might have to be changed. As the interface is pretty small, I wouldn't expect many changes, though. The main challenge for Paper UI support will probably be the `PersistenceServiceConfiguration`, which luckily has no direct relation to the `PersistenceService´ interface.

If we all agree that we treat this as a pilot for a persistence extension and do not necessarily consider it to be complete for a while, I'd be ok to proceed with it - but in any case, it should be moved from `bundles/persistence` to `/extensions/persistence`.

mkhl:
> If we all agree that we treat this as a pilot for a persistence extension and do not necessarily consider it to be complete for a while, I'd be ok to proceed with it

That’s fine with me.

> but in any case, it should be moved from bundles/persistence to /extensions/persistence.

I’ll do that.

htreu:
> If we all agree that we treat this as a pilot for a persistence extension and do not necessarily consider it to be complete for a while, I'd be ok to proceed with it

Agreed.

kaikreuzer:
> to also persist the current unit for QuantityTypes.

I would question this. Assuming that most persistence services are some kind of database where you would want to store raw numbers (and not numbers with units as strings), I do not think that it is feasible to require a unit being persisted with every single value. Instead, I think our persistence configuration should cover UoM and allow to specify, to which unit states should be converted before persisting them. This would be one of the things I meant above with "which would need to be looked at and potentially extended/adapted to nicely fit into the current ESH concepts".

kaikreuzer:
@mkhl Sorry for coming back to late on this, the last comments were right before my summer vacation and I didn't noticed that you had meanwhile pushed the updates to move the bundle. Code looks all good to me now, I only noticed that you have named "Jens Viebig" as the author of some initial contributions and he is missing as an "Also-By" in your commit message(s).

Could you reach out to him and check that he also has a valid ECA and add him as an "Also-By", possibly simply by squashing all your commits into one?

mkhl:
@JSurf Hey there, I ported your openHAB MapDB persistence service (openhab/openhab1-addons#2264) to Eclipse SmartHome and credited you as the initial author.
Do you have a valid Eclipse Contributor Agreement?
Are you OK with me adding your email as an “Also-By” to this change?
It seems like both are required before they can merge this pull request.

JSurf:
Hi,

Do you have a valid Eclipse Contributor Agreement? No
Are you OK with me adding your email as an “Also-By” to this change? Ok

I'm fine with everything you need to do to complete this pull request.

kaikreuzer:
Hey @JSurf, excellent, many thanks!

> Do you have a valid Eclipse Contributor Agreement? No

It would simplify the approval process a lot if you were so kind and sign the Eclipse Contributor Agreement - all you need is to create an account at Eclipse and tick a box here: https://accounts.eclipse.org/user/login?destination=user/eca

extensions/persistence/org.eclipse.smarthome.persistence.mapdb/src/main/java/org/eclipse/smarthome/persistence/mapdb/internal/MapDbPersistenceService.java
```diff
@@ -53,7 +53,7 @@
  * @author Martin Kühl - Port to Eclipse SmartHome
  */
 @NonNullByDefault
-@Component(service = QueryablePersistenceService.class)
+@Component(service = { PersistenceService.class, QueryablePersistenceService.class })
```
htreu: since `QueryablePersistenceService` extends `PersistenceService`, doesn't it qualify for service resolution also without the explicit naming?

mkhl: I was surprised to find that it does not.

extensions/persistence/org.eclipse.smarthome.persistence.mapdb/pom.xml
```diff
@@ -0,0 +1,23 @@
...
+  <artifactId>org.eclipse.smarthome.persistence.mapdb</artifactId>
+
+  <packaging>eclipse-plugin</packaging>
+
+  <name>Eclipse SmartHome MapDB Persistence</name>
+
+  <properties>
+    <bundle.symbolicName>org.eclipse.smarthome.persistence.mapdb</bundle.symbolicName>
+    <bundle.namespace>org.eclipse.smarthome.persistence.mapdb</bundle.namespace>
+  </properties>
```
wborn: Please remove these properties.

extensions/persistence/org.eclipse.smarthome.persistence.mapdb.test/pom.xml
```diff
@@ -0,0 +1,57 @@
...
+  <artifactId>org.eclipse.smarthome.persistence.mapdb.test</artifactId>
+
+  <packaging>eclipse-test-plugin</packaging>
+
+  <name>Eclipse SmartHome MapDB Persistence Tests</name>
+
+  <properties>
+    <bundle.symbolicName>org.eclipse.smarthome.persistence.mapdb.test</bundle.symbolicName>
+    <bundle.namespace>org.eclipse.smarthome.persistence.mapdb.test</bundle.namespace>
+  </properties>
```
wborn: Please remove these properties.

extensions/persistence/org.eclipse.smarthome.persistence.mapdb/pom.xml
mkhl: Please provide additional information.

wborn: They are no longer used. See #6060. :wink:

JSurf:
> It would simplify the approval process a lot if you were so kind and sign the Eclipse Contributor Agreement - all you need is to create an account at Eclipse and tick a box here: https://accounts.eclipse.org/user/login?destination=user/eca

@kaikreuzer Done

kaikreuzer:
Perfect, thanks!

> add him as an "Also-By", possibly simply by squashing all your commits into one?

@mkhl This next step is on you again - afterwards, I'll create the CQ.

mkhl:
@JSurf Thank you! I’ll need the email address you signed the ECA with. Feel free to email me at martin.kuehl@gmail.com if you’d prefer to keep it out of this thread.

mkhl:
@kaikreuzer Done

htreu:
FTR: CQ17330 created.
@mkhl & @JSurf: I stated 50% share of the work from each of you. If you think this is not correct please let me know.

extensions/persistence/org.eclipse.smarthome.persistence.mapdb.test/src/test/java/org/eclipse/smarthome/persistence/mapdb/MapDbPersistenceServiceOSGiTest.java
```diff
@@ -0,0 +1,123 @@
+package org.eclipse.smarthome.persistence.mapdb;
```
htreu: please add the Eclipse license header

extensions/persistence/org.eclipse.smarthome.persistence.mapdb.test/src/test/java/org/eclipse/smarthome/persistence/mapdb/StateTypeAdapterTest.java
```diff
@@ -0,0 +1,37 @@
+package org.eclipse.smarthome.persistence.mapdb;
```
htreu: please add the Eclipse license header

mkhl:
@htreu done

htreu:
CQ is approved.

I triggered travis once again due to unrelated test failure.

kaikreuzer:
Many thanks everyone!

kirantpatil:
@kaikreuzer, any plan to release new beta version  ref-0.10.0.b2 ?

Can we use the beta for deployment ?

Thanks,
Kiran.

maggu2810:
@kaikreuzer Do I recall correctly that we do not plan to release any new official "reftag" release anymore but keep that "inofficial" releases up to the downstream projects?

kaikreuzer:
No, the change was only that I do not need any beta releases anymore for building openHAB releases.
But whenever we think we have a rather stable situation there is demand for a new beta release from ESH users, it should be fine to build one.
At the moment I am just not sure whether we are in such a state, considering the open issues with the automation module...

kirantpatil:
@mkhl,

Does MapDB "RestoreOnStartup" feature, also sends commands to restore respective hardware of the items ?

I mean does it Restore both at Software and Hardware on Startup ?

If not then what needs to be modified such that "RestoreOnStartup" triggers for items at both Software client and respective Hardware (Light, FAN, AC ...) ?

Thanks,
Kiran.

mkhl:
This only restores the state of the Item and nothing else.

Hardware state should be handled by the ThingHandler of the hardware’s Things.

kirantpatil:
May I know, what purpose does it(RestoreOnStartup) serve, if ESH reboots and items restores to previous state (Light to ON) but hardware state is unchanged (Ex: Light state is OFF) ?

Are we not restoring wrong states to items when at the hardware it may not be in that state ?

htreu:
State update from real hardware is in the responsibility of the binding. The persistence `restoreOnStartup` restores item states, so that UIs, rules and other components dealing with items can start off where they did on shutdown.
All linked channels for hardware things will receive a `REFRESH` command once the handler is initialized. So After restart and all initialization, the bindings should automatically request the current hardware state.
There are however devices which do not respond to requests, but will only report state when they report them on their own (like battery powered window contacts).

PR #5425, opened 2018-04-18
Documentation Fix title and alt text for icon list

### LINKED ISSUE #5416: [Classic Icon Set] "presence-on" icon is missing
Hi,

This is a port from openhab/openhab-docs#660

It seems that somethings going wrong in https://github.com/eclipse/smarthome/blob/master/extensions/ui/iconset/org.eclipse.smarthome.ui.iconset.classic/README.sh and the html tags for the presence icon are generated wrong.

triller-telekom:
Fixes #5416


sjka:
How can this fix #5416 if it's not yet even clear what the bug report is about?

It might have been surprising to hide the information which state variants are available - but isn't this PR now dropping this information completely? How do you expect somebody to gain access to this information from the docu?

sjka:
> As far as I understood #5416 the problem is that the text below the icon is "presence" and if you hover over it or use a text browser which cannot display icons, "presence-off" is shown.

No. Please read on, that has been clarified in https://github.com/eclipse/smarthome/issues/5416#issuecomment-382403590

> this icon list is supposed to show the available categories itself.

Shouldn't a [page](https://www.eclipse.org/smarthome/documentation/features/ui/iconset/classic/readme.html) titled "Classic Icon Set" be about the capabilities and features of _this specific_ icon set rather than about available categories? 

IMHO it had some value to have _some_ way of figuring out what icon variants are available. I admit, this was an abuse of `alt="..."` attribute. But then there should be a proper alternative, don't you think?

triller-telekom:
Since this PR doesn't fix https://github.com/eclipse/smarthome/issues/5416#issuecomment-382644413 it probably creates more confusion than before. I will close it and once the fore-mentioned comment is discussed, i.e. a solution has been found, we can come up with a PR that implements it.

Issue #5420, opened 2018-04-17
I18n of thing status info description does not work when extending thing handlers in another bundle


hsudbrock:
**Context**
If a thing status info description is given in a special format (example: `@text/rate_limit`) then it is translated automatically by the `o.e.sh.core.thing.i18n.ThingStatusInfoI18nLocalizationService`, using the translations from the folder `ESH-INF/i18n` in the bundle of the thing handler.

**Issue**
If one extends a thing handler in another bundle, this translation mechanism does not work anymore. Instead of translating the description, the specially formatted string is kept in the description (and will be displayed in GUIs like, e.g., the PaperUI).

One example is an issue of the openhab ZigBee binding (https://github.com/openhab/org.openhab.binding.zigbee/issues/185): Quite recently, the ZigBee binding was split up into several bundles; the thing handlers for concrete ZigBee USB dongles are now in separate bundles, each extending an abstract thing handler in a core bundle. The abstract handler sets the thing status info description to, e.g., `@text/zigbee.status.offline_startupfail`, and provides translations for this in the `ESH-INF/i18n` folder. But this has no effect, as this is a different bundle from the bundles in which the actual handler implementations are located.

**First discussion of improving this**
In the example given above, one could, of course, copy the translations to every bundle containing a concrete  handler implementation. But this looks bad to me, since code has to be repeated. One could also think about using fragments containing the concrete handler implementations, but this has other disadvantages.

Another solution would be to extend the `ThingStatusInfoI18nLocalizationService`. This service currently attempts to translate the description using only translations from the bundle containing the handler implementation. One could extend this by looking for translations also in the bundle of the parent class of the handler (in case no translation can be found in the bundle of the handler itself) (continuing recursively until a translation is found or there is no further parent class).

Does this sound like a sensible solution? (If yes, I could provide a PR.) Or is there maybe a better solution?

htreu:
@hsudbrock thanks for the report and the nice analysis. Your proposed solution for this sounds good to me (although it would involve reflection...). Looking forward to see your PR.

PR #5429, opened 2018-04-19
[5420] Lookup translations for thing status description also in bundles of thing handler parent classes

hsudbrock:
This PR resolves https://github.com/eclipse/smarthome/issues/5420.

With this PR, bundles containing abstract thing handler implementations can provide translations for thing status descriptions that will be used in other bundles that extend the abstract thing handler implementation.

Example: Consider the ZigBee binding (https://github.com/openhab/org.openhab.binding.zigbee) which provides the abstract `ZigBeeCoordinatorHandler` in its core bundle, which is extended by concrete handler implementations in other bundles. The core bundle provides translations used in the abstract handler - and with this PR, these translations will actually be used.

Note that this PR contains an additional Groovy test class - I am aware that Java is preferred for tests now, but I wanted to keep in line with the existing tests for thing status description i18n, without making the existing Groovy test class even bigger.


bundles/core/org.eclipse.smarthome.core.thing.test/src/test/groovy/org/eclipse/smarthome/core/thing/i18n/ThingStatusInfoI18nLocalizationServiceMultipleBundlesOSGiTest.groovy
```diff
@@ -0,0 +1,125 @@
...
+            new ExtendingOtherBundleThingHandler(thing)
+        }
+    }
+
+    class ExtendingOtherBundleThingHandler extends MagicColorLightHandler {
+        public ExtendingOtherBundleThingHandler(Thing thing) {
+            super(thing)
+        }
+    }
+}
```
triller-telekom: Please add a new line at the end of the file.

bundles/core/org.eclipse.smarthome.core.thing/src/main/java/org/eclipse/smarthome/core/thing/i18n/ThingStatusInfoI18nLocalizationService.java
```diff
@@ -15,6 +15,8 @@
 import java.util.Arrays;
 import java.util.Locale;
 
+import org.eclipse.jdt.annotation.NonNull;
```
triller-telekom: We agreed to not use `@NonNull`. Please use `@NonNullByDefault` for the whole file, or if not feasible for this PR, because it would involve too many changes in depending classes, just remove this annotation for now.

bundles/core/org.eclipse.smarthome.core.thing/src/main/java/org/eclipse/smarthome/core/thing/i18n/ThingStatusInfoI18nLocalizationService.java
```diff
@@ -98,35 +98,76 @@ protected void unsetTranslationProvider(TranslationProvider i18nProvider) {
...
+    private String getTranslationForClass(@NonNull String i18nConstant, @Nullable Locale locale,
+            @Nullable Class<?> clazz, @Nullable Object... args) {
+        if (clazz == null) {
+            return i18nConstant;
+        }
+
+        Bundle bundle = FrameworkUtil.getBundle(clazz);
+
+        if (bundle == null) {
+            return getTranslationForClass(i18nConstant, locale, clazz.getSuperclass());
```
triller-telekom: Why don't you pass the `args` as a parameter to this method call?

bundles/core/org.eclipse.smarthome.core.thing.test/src/test/groovy/org/eclipse/smarthome/core/thing/i18n/ThingStatusInfoI18nLocalizationServiceMultipleBundlesOSGiTest.groovy
hsudbrock: OK, will be included in my next commit to this PR.

bundles/core/org.eclipse.smarthome.core.thing/src/main/java/org/eclipse/smarthome/core/thing/i18n/ThingStatusInfoI18nLocalizationService.java
hsudbrock: OK; I will remove the annotations, as surrounding classes are not annotated yet.

bundles/core/org.eclipse.smarthome.core.thing/src/main/java/org/eclipse/smarthome/core/thing/i18n/ThingStatusInfoI18nLocalizationService.java
hsudbrock: Good question :) The answer is that I simply forgot it (probably because this is a corner case for which I cannot imagine how it could happen - this case is relevant if, in the class hierarchy of the thing handler, there is a superclass not defined by a bundle class loader, which has a superclass within a bundle with i18n translations).

Thanks for pointing this out.

hsudbrock:
Thanks for your comments, @triller-telekom, I addressed them in my recent commit.

hsudbrock:
@triller-telekom and I discussed that it would be nice to also convert the existing Groovy tests for the fixed/extended feature to Java. So please wait with this PR until I am done with this :)

hsudbrock:
@triller-telekom When migrating the tests from Groovy to Java, I was wondering why there is no equivalent for the method `OSGiTest#setDefaultLocale` in the `JavaOSGiTest` class - and ended putting the implementation of this method into the test class here.

Do you know why there is no such equivalent, or whether this should also be added to the `JavaOSGiTest`?

triller-telekom:
@hsudbrock The `JavaOSGiTest` and the goovy `OSGiTest` should be identical, feature-wise. So please feel free to move the `setDefaultLocale` method into the `JavaOSGiTest`

hsudbrock:
As discussed offline, @triller-telekom, I have moved the `setDefaultLocale` method into a utility class, which is used by those three test classes in ESH that used `OSGiTest#setDefaultLocale` before.

In ESH, the method `OSGiTest#setDefaultLocale` should now no longer be used within ESH; I think that it should nevertheless stay in the code base, as it might be used by projects based on ESH.

bundles/core/org.eclipse.smarthome.core.thing.test/src/test/java/org/eclipse/smarthome/core/thing/i18n/ThingStatusInfoI18nLocalizationServiceOSGiTest.java
```diff
@@ -0,0 +1,290 @@
...
+            return true;
+        }
+
+        @Override
+        protected ThingHandler createHandler(Thing thing) {
+            return new SimpleThingHandler(thing);
+        }
+    }
+
+    private class SimpleThingHandler extends MagicColorLightHandler {
```
htreu: Is there any advantage from extending the magic handler? I'm not sure if we should start depending our unit tests on the magic bundle. What about extending the BaseThingHandler?

hsudbrock:
@JochenHiller just hinted me at the `org.eclipse.smarthome.test.SyntheticBundleInstaller` - I will give that a try for removing the dependency to the magic bundle.

hsudbrock:
I had a look at the `SyntheticBundleInstaller` from ESH, and also at the `SyntheticBundleBuilder` from the concierge project (cf. https://github.com/eclipse/concierge/blob/master/framework/org.eclipse.concierge/test/org/eclipse/concierge/test/util/SyntheticBundleBuilder.java).

They support installing bundles during test runtime - but they do not provide an easy way to add classes to those bundles (classes would have to be added as a class file, which must either be precompiled before the test, or generated by compiling it during test runtime). (Currently, the `SyntheticBundleInstaller` is only used to install bundles that only contain textual files like manifest or ESH XML files.)

One might be able to do this for this test (compiling a class file at runtime, and using some kind of reflection to extend this class file at runtime) - but I think that this is overkill here.

@htreu Hence, I still don't see a better solution for testing this except referencing the magic bundle (as currently done), or adding an additional bundle just for this test - and I prefer the first solution that I implemented. What do you think?

htreu:
@hsudbrock with #5466 merged you could try to refactor the tests to not use the full OSGi stack but only check whether the parent class object is used to resolve the translation.
Also: please rebase.

hsudbrock:
@htreu Thanks for implementing #5466; with my recent commit, the unit test no longer uses the magic binding, but rather a custom `BundleResolver` that returns for a certain class a bundle added using the `SyntheticBundleInstaller`.

bundles/core/org.eclipse.smarthome.core.thing.test/src/test/resources/test-bundle-pool/thingStatusInfoI18nTest.bundle/META-INF/MANIFEST.MF
```diff
@@ -0,0 +1,12 @@
+Manifest-Version: 1.0
+Bundle-ClassPath: .
+Bundle-ManifestVersion: 2
+Bundle-Name: Synthetic Test Bundle
+Bundle-RequiredExecutionEnvironment: JavaSE-1.8
+Bundle-SymbolicName: thingStatusInfoI18nTest.bundle;singleton:=true
+Bundle-Vendor: Eclipse.org/SmartHome
+Bundle-Version: 1.0.0.qualifier
+Import-Package: 
+ com.google.common.collect,
```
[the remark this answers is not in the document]
hsudbrock: I copied that from an existing usage of synthetic bundles... I think it can be safely removed here (let me update the PR in a couple of minutes).

PR #5431, opened 2018-04-19
Item tags become metadata

### LINKED ISSUE #5428: Use metadata Infrastructure for persisting item tags
Currently, tags are an integral part of item definitions and are defined and persisted along with the items for that reason.

This makes it impossible to maintain tags for items that are not stored in the database, but are rather provided through other means, be it textual/DSL files or the "virtual" ChannelItemProvider.

As we now have the metadata infrastructure in place, which allows providing and persisting additional information about items through independent means, it would make sense to use this for the tag management as well.

We most likely could move the tag management without impacting backward compatibility, if the item registry itself simply splits item and tag information internally and treats the persistence differently. When reading items and metadata, tag information could be automatically be merged together (this means that we potentially don't even have to adapt the GenericItemProvider at all).

While tags would be persisted through the metadata infrastructure (potentially under "esh" or "tag" namespace?), we should treat that specially and not make it accessible/visible through the metadata service to others - the standard way to access and edit tags should continue to be the item registry.

sjka:
With this change, the tags which previously were stored
together with the items now are stored as metadata. This
is completely transparent though, so that users won't
require to migrate any data.

They are, however, stored in a namespace (`_tags`)  which
by its prefix marks it as "internal" and therefore is not
exposed like all other, normal namespaces e.g. through the
REST API.

fixes #5428

### DIFF
--- a/bundles/core/org.eclipse.smarthome.core/src/main/java/org/eclipse/smarthome/core/internal/items/ItemRegistryImpl.java
+++ b/bundles/core/org.eclipse.smarthome.core/src/main/java/org/eclipse/smarthome/core/internal/items/ItemRegistryImpl.java
@@ -15,14 +15,22 @@ package org.eclipse.smarthome.core.internal.items;
 import static java.util.stream.Collectors.toList;
 
 import java.util.ArrayList;
+import java.util.Arrays;
 import java.util.Collection;
+import java.util.Collections;
 import java.util.List;
+import java.util.Set;
+import java.util.SortedSet;
+import java.util.TreeSet;
 import java.util.concurrent.CopyOnWriteArrayList;
+import java.util.stream.Stream;
 
+import org.eclipse.jdt.annotation.Nullable;
 import org.eclipse.smarthome.core.common.registry.AbstractRegistry;
 import org.eclipse.smarthome.core.common.registry.Provider;
 import org.eclipse.smarthome.core.events.EventPublisher;
 import org.eclipse.smarthome.core.i18n.UnitProvider;
+import org.eclipse.smarthome.core.items.ActiveItem;
 import org.eclipse.smarthome.core.items.GenericItem;
 import org.eclipse.smarthome.core.items.GroupItem;
 import org.eclipse.smarthome.core.items.Item;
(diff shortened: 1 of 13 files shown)


kaikreuzer:
There's a test failure that seems to be related:
```
Failed tests: 
  GenericItemProviderTest.assert that items have tags if specified:103 
Expected: is "TAG1, TAG2, TAG3, TAG4-WITH-DASHES, TAG5 String Tag"
     but: was ""
```

bundles/io/org.eclipse.smarthome.io.rest.core/src/main/java/org/eclipse/smarthome/io/rest/core/internal/item/ItemResource.java
```diff
@@ -236,7 +238,9 @@ public Response getItemData(@HeaderParam(HttpHeaders.ACCEPT_LANGUAGE) @ApiParam(
 
         final Locale locale = LocaleUtil.getLocale(language);
         final Set<String> namespaces = namespaceSelector == null ? Collections.emptySet()
-                : Arrays.stream(namespaceSelector.split(",")).collect(Collectors.toSet());
+                : Arrays.stream(namespaceSelector.split(",")) //
```
htreu: is this block and lines 216 - 218 worth a separate method?

bundles/io/org.eclipse.smarthome.io.rest.core/src/main/java/org/eclipse/smarthome/io/rest/core/internal/item/ItemResource.java
```diff
@@ -535,6 +523,11 @@ public Response addMetadata(@PathParam("itemname") @ApiParam(value = "item name"
             return Response.status(Status.NOT_FOUND).build();
         }
 
+        if (namespace.startsWith(MetadataRegistry.INTERNAL_NAMESPACE_PREFIX)) {
```
htreu: how about `isInternalNamespace(namespace))`?

bundles/core/org.eclipse.smarthome.core/src/main/java/org/eclipse/smarthome/core/internal/items/ItemRegistryImpl.java
```diff
@@ -347,6 +402,83 @@ public Item remove(String itemName, boolean recursive) {
...
+        return metadata;
+    }
+
+    private SortedSet<String> readTags(String itemName) {
+        MetadataKey key = new MetadataKey(TAG_NAMESPACE, itemName);
+        SortedSet<String> tags = new TreeSet<>();
+        if (metadataRegistry != null) {
+            Metadata metadata = metadataRegistry.get(key);
+            if (metadata != null) {
+                tags.addAll(Arrays.asList(metadata.getValue().split("\\" + TAG_SEPARATOR)));
```
afuechsel: As the string `"\\"` depends on the `TAG_SEPARATOR`, I would also externalize it as a constant.

afuechsel:
As the meta data registry now manages both meta data and tags, wouldn't it be a good idea to have some official API to access the tags directly from the meta data registry? Accessing the internal namespace `_tags`  doesn't sound as a good idea to me.

kaikreuzer:
@afuechsel The official API to access the tags is the ItemRegistry - to the outside world, we want to hide the fact the the metadata infrastructure is used internally as a persistence mechanism.

sjka:
Rebased to resolve merge conflicts

PR #5435, opened 2018-04-20
[Model Server] use managed thread pool

### LINKED ISSUE #4723: LSP ModelServer: potential Denial of service
As I reviewed #4722 I identified this line in the code: https://github.com/kaikreuzer/smarthome/blob/d893a85/bundles/model/org.eclipse.smarthome.model.lsp/src/main/java/org/eclipse/smarthome/model/lsp/internal/ModelServer.java#L81

Do we really want to create a new thread for every incoming connection?
If I run a loop that opens a connection to that port and tries to keep the socket open then for every loop a new thread is started on the server.
Why don't we use a thread pool?

sjka:
...instead of spawning threads by itself.

fixes #4723


bundles/model/org.eclipse.smarthome.model.lsp/src/main/java/org/eclipse/smarthome/model/lsp/internal/ModelServer.java
```diff
@@ -54,6 +56,7 @@
     static final String CONFIG_PID = "org.eclipse.smarthome.lsp";
     private static final String KEY_PORT = "port";
     private static final int DEFAULT_PORT = 5007;
+    private final ExecutorService pool = ThreadPoolManager.getPool("Language Server");
```
kaikreuzer: We usually use camel case names starting lowercase - like e.g. "thingHandler". Maybe change this here to "languageServer" or simply "lsp"?

PR #5437, opened 2018-04-20
Fix NPE in RuleEngine if startup rule has an error

### LINKED ISSUE #5436: Broken startup-rule leads to NPE
While trying to reproduce a bug where I had to create broken rules on purpose I stumbled across this NPE:

```
14:08:29.936 WARN  o.e.s.c.i.c.WrappedScheduledExecutorService[:67] - Scheduled runnable ended with an exception
	at org.eclipse.smarthome.model.rule.runtime.internal.engine.RuleEngineImpl.lambda$1(RuleEngineImpl.java:315)
```
Apparently the `ScriptExecutionException` does not have a cause but provides the error message directly via `getMessage()`.

https://github.com/eclipse/smarthome/blob/31ea3e1e315a1cba9176772fa27173528e6ac411/bundles/model/org.eclipse.smarthome.model.rule.runtime/src/org/eclipse/smarthome/model/rule/runtime/internal/engine/RuleEngineImpl.java#L312-L321

I will provide a PR for it.

triller-telekom:
Fixes #5436


triller-telekom:
So you suggest to catch both cases:

- The message is passed directly
- The message is within the cause of the Exception

Did I understand you correctly? I have adjusted the PR to reflect that.

PR #5442, opened 2018-04-23
Catch and handle exception from image generation

kaikreuzer:
fixes https://github.com/openhab/openhab-distro/issues/684
fixes https://github.com/openhab/openhab.ios/issues/89

### DIFF
--- a/bundles/ui/org.eclipse.smarthome.ui/src/main/java/org/eclipse/smarthome/ui/internal/chart/ChartServlet.java
+++ b/bundles/ui/org.eclipse.smarthome.ui/src/main/java/org/eclipse/smarthome/ui/internal/chart/ChartServlet.java
@@ -22,6 +22,7 @@ import java.util.Hashtable;
 import java.util.Map;
 import java.util.Objects;
 
+import javax.imageio.IIOException;
 import javax.imageio.ImageIO;
 import javax.imageio.stream.ImageOutputStream;
 import javax.servlet.Servlet;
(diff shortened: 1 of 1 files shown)


bundles/ui/org.eclipse.smarthome.ui/src/main/java/org/eclipse/smarthome/ui/internal/chart/ChartServlet.java
```diff
@@ -322,6 +323,10 @@ protected void doGet(HttpServletRequest req, HttpServletResponse res) throws Ser
         } catch (IllegalArgumentException e) {
             logger.warn("Illegal argument in chart: {}", e.getMessage());
             res.sendError(HttpServletResponse.SC_BAD_REQUEST, "Illegal argument in chart: " + e.getMessage());
+        } catch (IIOException e) {
+            // this can happen if the request is terminated while the image is streamed, see
+            // https://github.com/openhab/openhab-distro/issues/684
+            logger.debug("Failed writing image to response stream", e);
```
maggu2810: So, why not sending an error then, so the get request signals that something go wrong?

kaikreuzer: Because nothing is wrong here. It is a valid behavior of a client to close a connection (e.g. because the app is closed or whatever).

maggu2810: Sure, you are right. My reading of the code has been wrong and I assumed the "reading" of the chart (so not the writing to `res` produces that exception.

PR #5446, opened 2018-04-23
generated binding skeleton: run tests without errors

ermartens:
After building the skeleton, the tests should run without errors.

With these modifications, they should.

I added a few lines to method OSGITest.setUp to be more verbose in case something goes wrong.

The  HandlerTest.java still uses method handler.initialize() which is marked deprecated in BaseThingHandler.java - I am curious about what will be the replacement for that!

Regards,
Erik


tools/archetype/binding.test/src/main/resources/archetype-resources/src/test/java/__bindingIdCamelCase__HandlerTest.java
```diff
@@ -60,6 +60,8 @@ public void setUp() {
 
     @Test
     public void initializeShouldCallTheCallback() {
+        // mock the function to prevent null exception.
+        when(thing.getConfiguration()).thenReturn(new demo1Configuration());
```
htreu: `demo1` seems to be a left over

tools/archetype/binding/src/main/resources/archetype-resources/src/main/java/internal/__bindingIdCamelCase__Configuration.java
```diff
@@ -19,12 +19,14 @@
...
 
+import org.eclipse.smarthome.config.core.Configuration;
+
 /**
  * The {@link ${bindingIdCamelCase}Configuration} class contains fields mapping thing configuration parameters.
  *
  * @author ${author} - Initial contribution
  */
-public class ${bindingIdCamelCase}Configuration {
+public class ${bindingIdCamelCase}Configuration extends Configuration {
```
sjka: That's not the idea. Such classes should be merely POJOs, which then get be used in contexts like this:

```java
${bindingIdCamelCase}Configuration config = getThing().getConfiguration().as(${bindingIdCamelCase}Configuration.class);
```

Therefore they must _not_  extend `Configuration`.

tools/archetype/binding.test/src/main/resources/archetype-resources/src/test/java/__bindingIdCamelCase__OSGiTest.java
```diff
@@ -51,8 +51,12 @@
     @Before
     public void setUp() {
         registerService(volatileStorageService);
+
         managedThingProvider = getService(ThingProvider.class, ManagedThingProvider.class);
+        assertNotNull("Could not get ManagedThingProvider", managedThingProvider);
```
sjka: I understand the intention of the explanatory texts, but I personally dislike them because they state the opposite of what should be asserted. They are great if something goes wrong, but they actually make the code harder to read. 
It might not be such of a big issue here, but we have to keep in mind that this code should server as an example - and in that case I think it is leading towards bad style. 

Therefore I suggest to only add the assertions themselves because they indeed are quite helpful for figuring out what went wrong at the ealiest point possible, but drop the explanatory texts.

triller-telekom:
@ermartens 
> The HandlerTest.java still uses method handler.initialize() which is marked deprecated in BaseThingHandler.java - I am curious about what will be the replacement for that!

it is marked as deprecated because it should definitely be overridden by the sub class. To make it more clear: It is deprecated in a sub class to call `super.initialize()`, because this sets the thing straight to ONLINE which in most cases is wrong, because configuration parameters have to be checked, a connection to the device has to be established, etc.

tools/archetype/binding.test/src/main/resources/archetype-resources/src/test/java/__bindingIdCamelCase__HandlerTest.java
sjka: Thanks - this indeed looks like it was forgotten in #4771 (and went unnoticed since...).

tools/archetype/binding.test/src/main/resources/archetype-resources/src/test/java/__bindingIdCamelCase__OSGiTest.java
```diff
@@ -65,6 +69,6 @@ public void tearDown() {
     public void creationOf${bindingIdCamelCase}Handler() {
         assertThat(bridge.getHandler(), is(nullValue()));
         managedThingProvider.add(bridge);
-        waitForAssert(() -> assertThat(bridge.getHandler(), is(notNullValue())));
+        waitForAssert(() -> assertThat(managedThingProvider.get(bridge.getUID()), is(notNullValue())));
```
sjka: I know, the hamcrest matcher was used before, but while you're here anyway could you please change it to 

```java
waitForAssert(() -> assertNotNull(managedThingProvider.get(bridge.getUID())));
```

ermartens:
Thank you for your comments, also for clarifying the issue about `initialize()`
I comitted accordingly, hope that helps. How about the corresponding changes in openHAB? I have not found respective files in the repo.

cweitkamp:
Both scripts (ESH and OH2) are based on the same archetype skeleton which is located in this repo.

kaikreuzer:
@ermartens Did you see my comment? Would be great if you could update the PR accordingly!

ermartens:
Thank you @kaikreuzer , @htreu!
I followed all the necessary steps as mentioned in https://github.com/eclipse/smarthome/blob/master/CONTRIBUTING.md#contributing-to-eclipse-smarthome. 
I am a medium experienced enthusiast, not a professional programmer, and when it comes to vi-style editing, I am lost: I can see my commits in the result of `git rebase -i HEAD~200`, but If I press r with the cursor in the corresponding line, nothing happens.

ermartens:
Following @htreu 's instructions above, I made another push.

kaikreuzer:
I am afraid, something went terribly wrong there - this PR now shows 186 commits instead of a single one... One issue might be that you created the PR from your master branch - you should instead always create a separate feature branch for any PR.
If you are willing to go through some more git lessons, you might want to try the following:
1. Copy your changes to a safe place
2. Reset your master to the origin master
4. Do a single commit with your changes
5. Force push your master branch

kaikreuzer:
Ok, it seems @ermartens is struggling a bit with git, so I suggest to close this PR.
I have created https://github.com/eclipse/smarthome/pull/5987 as a replacement.

PR #5449, opened 2018-04-23
added deprecated info to ActiveItem

kaikreuzer:
This PR makes it clear that `ActiveItem` must not be considered public API - imho, we should move towards removing it and replacing it by the builder pattern as discussed for Rules and Things as well.


maggu2810:
... and if the setters should not be used (and later available) anymore, the `GenericItem` should not only mark it deprecated but also `final`, so we ensure no one implements its own setter that is not marked deprecated, too.

maggu2810:
@kaikreuzer @sjka WDYT?

afuechsel:
I agree to @maggu2810 's comments. Furthermore I do not understand how I can now set labels or add tags to an `Item`. Shall we introduce some kind of `ItemBuilder` instead? But tags can be added or removed later on and the label could be changed etc. so a builder pattern for a first construction would not be enough.

Issue #5445, opened 2018-04-23
The Dimensionless dimension selection results in values with " one" string appended to them


spikeygg:
I have found that when using the "Dimensionless" option on items that result in a Number, the resulting value is the numeric value followed by the string " one". This post was suggested as the result from the following forum post on the openhab forum: https://community.openhab.org/t/new-homeseer-wd200-switches-coming/39244/46. I have found that I can use the "Area" dimension as a replacement which will not apply any units and still allow for numerical comparisons.

Here is an example from my openhab2 events.log:

```
2018-04-21 23:43:32.083 [vent.ItemStateChangedEvent] - ZM_Server_CPULoad changed from NULL to 0.87 one
2018-04-21 23:43:43.487 [vent.ItemStateChangedEvent] - ZM_Server_CPULoad changed from 0.87 one to 0.86 one
2018-04-21 23:44:06.525 [vent.ItemStateChangedEvent] - ZM_Server_CPULoad changed from 0.86 one to 0.87 one
2018-04-21 23:44:29.452 [vent.ItemStateChangedEvent] - ZM_Server_CPULoad changed from 0.87 one to 0.88 one
2018-04-21 23:44:52.600 [vent.ItemStateChangedEvent] - ZM_Server_CPULoad changed from 0.88 one to 0.87 one
2018-04-21 23:45:04.150 [vent.ItemStateChangedEvent] - ZM_Server_CPULoad changed from 0.87 one to 0.86 one
2018-04-21 23:45:15.614 [vent.ItemStateChangedEvent] - ZM_Server_CPULoad changed from 0.86 one to 0.85 one
2018-04-21 23:45:27.163 [vent.ItemStateChangedEvent] - ZM_Server_CPULoad changed from 0.85 one to 0.84 one
```

When I updated the dimension from "Dimensionless" to "Area" the events.log file appeared like this:
```
2018-04-22 21:54:07.698 [vent.ItemStateChangedEvent] - ZM_Server_CPULoad changed from 0.73 one to 0.74 one
2018-04-22 21:54:19.195 [vent.ItemStateChangedEvent] - ZM_Server_CPULoad changed from 0.74 one to 0.73 one
2018-04-22 21:54:30.659 [vent.ItemStateChangedEvent] - ZM_Server_CPULoad changed from 0.73 one to 0.72 one
2018-04-22 21:54:41.952 [ome.event.ItemUpdatedEvent] - Item 'ZM_Server_CPULoad' has been updated.
[...]

PR #5453, opened 2018-04-24
don't serialize unit for unitless quantities

maggu2810:
Fixes: https://github.com/eclipse/smarthome/issues/5445

### DIFF
--- a/bundles/core/org.eclipse.smarthome.core/src/main/java/org/eclipse/smarthome/core/library/types/QuantityType.java
+++ b/bundles/core/org.eclipse.smarthome.core/src/main/java/org/eclipse/smarthome/core/library/types/QuantityType.java
@@ -268,7 +268,11 @@ public class QuantityType<T extends Quantity<T>> extends Number
 
     @Override
     public String toFullString() {
-        return quantity.toString();
+        if (quantity.getUnit() == AbstractUnit.ONE) {
+            return quantity.getValue().toString();
+        } else {
+            return quantity.toString();
+        }
     }
 
     @Override


sjka:
Makes sense. 

I reckon though it would be worth amending [QuantityTypeTest.testDimensionless()](https://github.com/eclipse/smarthome/blob/4157491527705cb281abeb8aa5c39affa39c988b/bundles/core/org.eclipse.smarthome.core.test/src/test/java/org/eclipse/smarthome/core/library/types/QuantityTypeTest.java#L49) accordingly.

htreu:
I had the discussion with @kaikreuzer during implementation phase. Imho this is nit the bug. Please also see my comment here: https://github.com/eclipse/smarthome/issues/5445#issuecomment-383916890

htreu:
It is to be discussed, if AbstractUnit.ONE should be the default unit for the dimension `Dimensionless`.  After all, NumberItems which are not linked to Channels providing QuantityType should not define a dimension.

maggu2810:
> It is to be discussed, if AbstractUnit.ONE should be the default unit for the dimension `Dimensionless`.  

Hm, isn't that the intention of that public static member?
* the `DefaultQuantityFactory` is using this member for the `Dimensionless.class` in its `CLASS_TO_METRIC_UNIT` map
* `DefaultQuantityFormat` checks if the unit if equal to `AbstractUnit.ONE` in its format method
* `NumberSpaceQuantityFormat` checks if the unit if equal to `AbstractUnit.ONE` in its format method
* ...

htreu:
Okay, to be more precise I was talking about the default defined in I18nProviderImpl in ESH:

https://github.com/eclipse/smarthome/blob/95830210a7f215823712f0fbca053e938e878950/bundles/core/org.eclipse.smarthome.core/src/main/java/org/eclipse/smarthome/core/internal/i18n/I18nProviderImpl.java#L370-L373

And imho its correctly defined as such.

That being said, the issue originates from the fact that we try to set a unit to dimensionized Number items also for DecimalType state updates. This is helpful when initialising the item from the REST API which may not yet be aware of the correct unit.
Forcing a "legacy" channel which does not support units yet is another reason to support this. When the user wants his DecimalType temperature value from a legacy binding expressed as QuantityType he may do so by defining the Temperature dimension and get the default unit for the current locale.

In the particular case of #5445 the problem seems to be 

> ...and still allow for numerical comparisons

which I found in the light of script support is a correct statement for SmartHomeUnits.ONE. In the script `NumberExtensions` we assumed that any QuantityType is not comparable to a simple scalar number. This holds true for all of them except SmartHomeUnits.ONE which should actually be comparable to a scalar (which it represents in the unit system).
I suggest to:
1. fix the NumberExtensions to allow for comparison between Number and SmartHomeUnit.ONE
2. evaluate what "numerical comparisons" was done. Switching the dimension to "Angle" should not fix the script problem but just "clean" the logs.

htreu:
We had an offline discussion about this issue/PR and a good argument for this PR was, that it should be possible to add a dimension to any Number item (with whatever dimension fits the channel purpose, also Dimensionless) and work with it without interruption. Therefore it makes sense to cut off the " one" prefix.
Otherwise, w/o the prefix we wouldn't have had such an easy time examining the issue at hand. And imho the console log alone does not make for a good argument to drop it.
I would like to wait for further feedback on #5445.

PR #5456, opened 2018-04-24
Remove default configurationPolicy = ConfigurationPolicy.OPTIONAL

wborn:
It's not necessary to add ConfigurationPolicy.OPTIONAL because it is defined in the `@interface Component` as:

```java
    ConfigurationPolicy configurationPolicy() default ConfigurationPolicy.OPTIONAL;
```


sjka:
I double-checked, just to be on the safe side: DS 1.1 (in OSGi 4.2) also contains this as a default:

```
<attribute name="configuration-policy"
type="scr:Tconfiguration-policy" default="optional" use="optional" />
```

Thanks!

PR #5457, opened 2018-04-24
[FSInternetRadio] Cleanup OSGI-INF XML files

wborn:
The FSInternetRadio Binding uses DS annotations so the component XML files should no longer be stored in Git.


sjka:
Thanks!

PR #5464, opened 2018-04-24
[DigitalSTROM] Use DS annotations

wborn:
Updates the DigitalSTROM Binding to use DS annotations.


extensions/binding/org.eclipse.smarthome.binding.digitalstrom/src/main/java/org/eclipse/smarthome/binding/digitalstrom/internal/discovery/BridgeDiscoveryService.java
```diff
@@ -39,6 +41,7 @@
  * @author Michael Ochel - Initial contribution
  * @author Matthias Siegele - Initial contribution
  */
+@Component(service = DiscoveryService.class, immediate = true, configurationPid = "discovery.digitalstrom")
```
sjka: The `activate()` and `modified()` methods also need to have `@Activate` and `@Modified` annotations

Especially for the `modified()` method this is crucial, as unlike `activate()`/`deactivate()` it will _not_ be called by naming convention implicitly!

extensions/binding/org.eclipse.smarthome.binding.digitalstrom/src/main/java/org/eclipse/smarthome/binding/digitalstrom/internal/providers/DsChannelTypeProvider.java
```diff
@@ -47,6 +51,7 @@
  * @author Matthias Siegele - Initial contribution
  *
  */
+@Component(service = ChannelTypeProvider.class, immediate = true)
```
sjka: The base class has an `activate()` method. Therefore this should override in in order to make it explicit:

```java
@Activate
@Override
public void activate(ComponentContext componentContext) {
    super.activate(componentContext);
}
```

The `deactivate()` method correspondingly.

Note: it correctly works as it is because the SCR implementations will look for such method signatures automatically. However, it is considered good practice to make it explicit - and therefore they need to be overridden because the XML generation processor does _not_ look at superclasses.

extensions/binding/org.eclipse.smarthome.binding.digitalstrom/src/main/java/org/eclipse/smarthome/binding/digitalstrom/internal/providers/DsDeviceThingTypeProvider.java
```diff
@@ -44,6 +47,7 @@
  * @author Michael Ochel - initial contributer
  * @author Matthias Siegele - initial contributer
  */
+@Component(service = ThingTypeProvider.class, immediate = true)
```
sjka: Please also override `activate()`/ `deactivate()` (see above)

extensions/binding/org.eclipse.smarthome.binding.digitalstrom/src/main/java/org/eclipse/smarthome/binding/digitalstrom/internal/providers/DsChannelTypeProvider.java
```diff
@@ -189,6 +194,17 @@ public static boolean isOutputChannel(String channelTypeID) {
         return supportedOutputChannelTypes.contains(channelTypeID);
     }
 
+    @Reference
```
sjka: In the XML it was declared 1..1 / **DYNAMIC**. The implementation actually can handle it being `null`, but as it is mandatory (1..1) anyway, I agree it does make more sense to declare it statically. Therefore: all good 😃

extensions/binding/org.eclipse.smarthome.binding.digitalstrom/src/main/java/org/eclipse/smarthome/binding/digitalstrom/internal/discovery/BridgeDiscoveryService.java
wborn: I've added the missing annotations @sjka.

Should we also fix this for all other classes extending the `AbstractDiscoveryService`? It would fix the issue that background discovery configuration changes are not immediately applied.

sjka: That would be awesome! Sorry, I noticed that too late... One the positive side: Now you won't need to amend bazillions of PRs but can do it in one.

PR #5465, opened 2018-04-25
Add Notification API support

ivivanov-bg:
Add support for the Notification API described here:
https://developer.bose.com/soundtouch-audio-notification-api/apis

### DIFF
--- a/extensions/binding/org.eclipse.smarthome.binding.bosesoundtouch/ESH-INF/config/config.xml
+++ b/extensions/binding/org.eclipse.smarthome.binding.bosesoundtouch/ESH-INF/config/config.xml
@@ -1,8 +1,8 @@
 <?xml version="1.0" encoding="UTF-8"?>
 <config-description:config-descriptions xmlns:xsi="http://www.w3.org/2001/XMLSchema-instance"
 	xmlns:config-description="http://eclipse.org/smarthome/schemas/config-description/v1.0.0"
-	xsi:schemaLocation="http://eclipse.org/smarthome/schemas/config-description/v1.0.0 http://eclipse.org/smarthome/schemas/config-description-1.0.0.xsd">
-
+	xsi:schemaLocation="http://eclipse.org/smarthome/schemas/config-description/v1.0.0 http://eclipse.org/smarthome/schemas/config-description-1.0.0.xsd"
+>
 	<config-description uri="thing-type:bosesoundtouch:config">
 		<parameter name="host" type="text" required="true">
 			<label>Host Address</label>
@@ -14,6 +14,44 @@
 			<description>The MAC Address used for communication to this device.
 			</description>
 		</parameter>
+		<parameter name="appKey" type="text" required="false">
+			<label>Authorization Key</label>
+			<description>An authorization key used to identify the client application.
+						Should be requested from the developer portal.
+			</description>
+		</parameter>
 	</config-description>
-
+	
+	<config-description uri="channel-type:bosesoundtouch:notificationSound">
+		<parameter name="notificationVolume" type="integer" min="10" max="70" step="1" unit="%" required="false">
+			<label>Notification Sound Volume</label>
+			<description>This indicates the desired volume level while playing the notification. 
+						The value represents a percentage (0 to 100) of the full audible range of the speaker device. 
+						A value less than 10 or greater than 70 will result in an error and not play the notification. 
+						Upon completion of the notification, the speaker volume will return to its original value. 
+						If not present, the notification will play at the existing volume level.
+			</description>
+		</parameter>
+		<parameter name="notificationService" type="text" required="true">
+			<label>Notification Service</label>
+			<description>This indicates the service providing the notification. 
+						This text will appear on the device display (when available) and the SoundTouch application screen.
+			</description>
+			<default>Notification</default>
+		</parameter>
+		<parameter name="notificationReason" type="text" required="false">
+			<label>Notification Reason</label>
+			<description>This indicates the reason for the notification. 
+						This text will appear on the device display (when available) and the SoundTouch application screen. 
+						If a reason string is not provided, the field with be blank.
+			</description>
+		</parameter>
+		<parameter name="notificationMessage" type="text" required="false">
+			<label>Notification Message</label>
+			<description>This indicates further details about the notification. 
+						This text will appear on the device display (when available) and the SoundTouch application screen. 
+						If a message string is not provided, the field with be blank.
+			</description>
+		</parameter>
+	</config-description>
 </config-description:config-descriptions>
--- a/extensions/binding/org.eclipse.smarthome.binding.bosesoundtouch/ESH-INF/thing/channels.xml
+++ b/extensions/binding/org.eclipse.smarthome.binding.bosesoundtouch/ESH-INF/thing/channels.xml
@@ -281,5 +281,11 @@
 			</options>
 		</state>
 	</channel-type>
-
+	<!-- Notification channels -->
+	<channel-type id="notificationsound" advanced="true">
+		<item-type>String</item-type>
+		<label>Notification Sound</label>
+		<description>Play a notification sound by a given URI</description>
+		<config-description-ref uri="channel-type:bosesoundtouch:notificationSound" />
+	</channel-type>
 </thing:thing-descriptions>
--- a/extensions/binding/org.eclipse.smarthome.binding.bosesoundtouch/README.md
+++ b/extensions/binding/org.eclipse.smarthome.binding.bosesoundtouch/README.md
@@ -29,13 +29,16 @@ The binding has no configuration options, all configuration is done at Thing lev
  
 All thing types have the same configuration parameters:
 
-| Parameter Name | Type   | Required | Description                                                  |
-|----------------|--------|----------|--------------------------------------------------------------|
-| host           | String | Yes      | The host name or IP address of the device                    |
-| macAddress     | String | Yes      | The MAC address of the used interface (format "123456789ABC")|
+| Parameter Name      | Type   | Required | Description                                                  |
+|---------------------|--------|----------|--------------------------------------------------------------|
+| host                | String | Yes      | The host name or IP address of the device                    |
+| macAddress          | String | Yes      | The MAC address of the used interface (format "123456789ABC")|
+| appKey              | String |  No      | An authorization key used to identify the client application |
 
-Both are set when using discovery. For manual configuration, these values can be found in the Bose smartphone app (Settings -> About -> Device Name).
+The required properties are set when using discovery. For manual configuration, these values can be found in the Bose smartphone app (Settings -> About -> Device Name).
 Note that the device might have two MAC addresses, one for ethernet and one for Wifi.
+
+The authorization key is used to identify the client application when using the Notification API. It should be requested from the developer portal.
  
 ## Channels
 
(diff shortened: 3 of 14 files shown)


ivivanov-bg:
Did some changes:
1. Moved Notification Volume and other info to the thing configuration
2. Renamed the notificationUrl channel to notificationsound (same as in Sonos)

extensions/binding/org.eclipse.smarthome.binding.bosesoundtouch/ESH-INF/config/config.xml
```diff
@@ -14,6 +14,27 @@
...
 		</parameter>
+		<parameter name="appKey" type="text" required="false">
+			<label>Consumer Key</label>
+			<description>The consumer key retrieved from the developer portal</description>
+		</parameter>
+		<parameter name="notificationVolume" type="integer" min="10" max="70" step="1" unit="%" required="false">
+			<label>Notification Sound Volume</label>
+			<description>Specifies the volume in percent applied to a notification sound</description>
+		</parameter>
+		<parameter name="notificationService" type="text" required="true">
```
kaikreuzer: What is this about? label+description are clearly wrong.

extensions/binding/org.eclipse.smarthome.binding.bosesoundtouch/ESH-INF/config/config.xml
```diff
@@ -14,6 +14,27 @@
...
+			<label>Notification Sound Volume</label>
+			<description>Specifies the volume in percent applied to a notification sound</description>
+			<default>Notification</default>
+		</parameter>
+		<parameter name="notificationReason" type="text" required="false">
```
kaikreuzer: same here

extensions/binding/org.eclipse.smarthome.binding.bosesoundtouch/ESH-INF/config/config.xml
```diff
@@ -14,6 +14,27 @@
...
+			<label>Notification Sound Volume</label>
+			<description>Specifies the volume in percent applied to a notification sound</description>
+		</parameter>
+		<parameter name="notificationMessage" type="text" required="false">
```
kaikreuzer: same here

extensions/binding/org.eclipse.smarthome.binding.bosesoundtouch/ESH-INF/thing/channels.xml
```diff
@@ -281,5 +281,10 @@
 			</options>
 		</state>
 	</channel-type>
-
+	<!-- Notification channels -->
+	<channel-type id="notificationsound">
```
kaikreuzer: Could you use the identical definition [as in Sonos](https://github.com/eclipse/smarthome/blob/master/extensions/binding/org.eclipse.smarthome.binding.sonos/ESH-INF/thing/channels.xml#L134-L138)?

extensions/binding/org.eclipse.smarthome.binding.bosesoundtouch/src/main/java/org/eclipse/smarthome/binding/bosesoundtouch/BoseSoundTouchBindingConstants.java
```diff
@@ -71,11 +72,29 @@
...
-
-    // Device configuration parameters;
-    public static final String DEVICE_PARAMETER_HOST = "host";
+    public static final String CHANNEL_NOTIFICATION_SOUND = "notificationsound";
 
     // Device information parameters;
     public static final String DEVICE_INFO_NAME = "INFO_NAME";
     public static final String DEVICE_INFO_TYPE = "INFO_TYPE";
+    
+    public static class BoseSoundTouchConfiguration {
```
kaikreuzer: We usually have those a standalone classes, not being embedded in the constants.

extensions/binding/org.eclipse.smarthome.binding.bosesoundtouch/src/main/java/org/eclipse/smarthome/binding/bosesoundtouch/handler/BoseSoundTouchHandler.java
```diff
@@ -222,6 +223,19 @@ public void handleCommand(ChannelUID channelUID, Command command) {
...
+                BoseSoundTouchConfiguration currentConfiguration = getConfigAs(BoseSoundTouchConfiguration.class);
+                if (currentConfiguration.appKey != null && !currentConfiguration.appKey.isEmpty()) {
+                    if (command instanceof StringType) {
+                        String url = command.toString();
+                        if (!url.isEmpty()) {
+                            commandExecutor.playNotificationSound(currentConfiguration, url);
+                        }
+                    }
+                } else {
+                    logger.warn("Missing app key - can not use notification api");
```
kaikreuzer: typo: cannot

extensions/binding/org.eclipse.smarthome.binding.bosesoundtouch/ESH-INF/config/config.xml
```diff
@@ -14,6 +14,27 @@
 			<description>The MAC Address used for communication to this device.
 			</description>
 		</parameter>
+		<parameter name="appKey" type="text" required="false">
+			<label>Consumer Key</label>
+			<description>The consumer key retrieved from the developer portal</description>
+		</parameter>
+		<parameter name="notificationVolume" type="integer" min="10" max="70" step="1" unit="%" required="false">
```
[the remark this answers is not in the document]
ivivanov-bg: This is API restriction: 
 > Element. This indicates the desired volume level while playing the notification. The value represents a percentage (0 to 100) of the full audible range of the speaker device. A value less than 10 or greater than 70 will result in an error and not play the notification. Upon completion of the notification, the speaker volume will return to its original value. If not present, the notification will play at the existing volume level.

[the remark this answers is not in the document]
kaikreuzer: Thanks for the clarification. Maybe worth to add a comment in the XML as well.

extensions/binding/org.eclipse.smarthome.binding.bosesoundtouch/src/main/java/org/eclipse/smarthome/binding/bosesoundtouch/BoseSoundTouchBindingConstants.java
ivivanov-bg: OK - but may I put it in an exported package?

kaikreuzer: Or move everything in internal packages (what we now do in the skeleton as well).

kaikreuzer:
What I had forgotten to mention: Please also update the README.md!

extensions/binding/org.eclipse.smarthome.binding.bosesoundtouch/src/main/java/org/eclipse/smarthome/binding/bosesoundtouch/BoseSoundTouchBindingConstants.java
ivivanov-bg: But I will need to access the config param names outside the binding (and anyway they are also visible in the config.xml) - I think it would be better to put the constants in an exported package

kaikreuzer: Sure, feel free.

ivivanov-bg: you mean - extract the constants from the ```BoseSoundTouchConfiguration``` and move them to ```BoseSoundTouchBindingConstants``` or externalize the ```BoseSoundTouchConfiguration``` class and put in an exported package ?

kaikreuzer: Jus externalize it in the same package (which already is exported).

ivivanov-bg:
Done.

extensions/binding/org.eclipse.smarthome.binding.bosesoundtouch/ESH-INF/config/config.xml
```diff
@@ -14,6 +14,41 @@
 			<description>The MAC Address used for communication to this device.
 			</description>
 		</parameter>
+		<parameter name="appKey" type="text" required="false">
+			<label>Authorization Key</label>
+			<description>An authorization key used to identify the client application.
+						Should be requested from the developer portal.
+			</description>
+		</parameter>
+		<parameter name="notificationVolume" type="integer" min="10" max="70" step="1" unit="%" required="false">
```
kaikreuzer: Hm, looking at this, might I come up with an idea?
All those 4 parameters are specific to the notificationSound channel - they are not useful for anything else.
So imho, it would make a lot of sense to move them to the channel as a configuration and not have them on the thing. The cool thing is, that we could then mark the Thing extensible by "notificationSound" channels, so that the user can add multiple channels of that type to his Thing - each having a different configuration, so you could have different volumes, texts, etc. for different scenarios. Wdyt?

ivivanov-bg:
Moved the configuration to the channel and updated command handling.
(but I can't test adding new channels because PaperUI doesn't show me such button (or at least I couldn't find it))

extensions/binding/org.eclipse.smarthome.binding.bosesoundtouch/README.md
```diff
@@ -29,13 +29,23 @@ The binding has no configuration options, all configuration is done at Thing lev
...
-| Parameter Name | Type   | Required | Description                                                  |
-|----------------|--------|----------|--------------------------------------------------------------|
-| host           | String | Yes      | The host name or IP address of the device                    |
-| macAddress     | String | Yes      | The MAC address of the used interface (format "123456789ABC")|
+| Parameter Name      | Type   | Required | Description                                                  |
+|---------------------|--------|----------|--------------------------------------------------------------|
+| host                | String | Yes      | The host name or IP address of the device                    |
+| macAddress          | String | Yes      | The MAC address of the used interface (format "123456789ABC")|
+| appKey              | String |  No      | An authorization key used to identify the client application |
+| notificationVolume  | String |  No      | Desired volume level while playing the notification          |
```
kaikreuzer: Please move those to the "Channels" section.

extensions/binding/org.eclipse.smarthome.binding.bosesoundtouch/src/main/java/org/eclipse/smarthome/binding/bosesoundtouch/handler/BoseSoundTouchHandler.java
```diff
@@ -223,9 +227,28 @@ public void handleCommand(ChannelUID channelUID, Command command) {
                 }
                 break;
             default:
-                logger.warn("{} : Got command '{}' for channel '{}' which is unhandled!", getDeviceName(), command,
-                        channelUID.getId());
-                break;
+                Channel channel = getThing().getChannel(channelUID.getId());
+                switch (channel.getChannelTypeUID().getId()) {
```
kaikreuzer: `channel` could be null, `channel.getChannelTypeUID()` as well.

ivivanov-bg:
Added a NULL check

kaikreuzer:
What about my second comment?

ivivanov-bg:
Totally missed it in the mail - sorry.
Will do it later today when I get to a PC.

ivivanov-bg:
Readme - updated

PR #5466, opened 2018-04-25
Provide BundleResolver service to decouple from static FrameworkUtil.getBundle

htreu:
This makes it more easy to write unit tests without a running OSGi framework.
Some places where `FrameworkUtil.getBundle` is still used could not be refactored easily, so left over for now.
This will also be used to reduce a test case in #5429 to a simple unit test.


bundles/core/org.eclipse.smarthome.core/src/main/java/org/eclipse/smarthome/core/util/BundleResolver.java
```diff
@@ -0,0 +1,40 @@
...
+ */
+public interface BundleResolver {
+
+    /**
+     * Resolve the bundle name associated with the given {@link Class}.
+     *
+     * @param clazz the {@link Class} to resolve the bundle name for.
+     * @return the bundle name associated with the given {@link Class}.
+     */
+    String resolveSymbolicName(Class<?> clazz);
```
maggu2810: Do we really need that function in the `BundleResolver`?
Why not using the `resolveBundle` in the interface only and the one that is interested in the symbolic name calls `getSymbolicName()` on the returned bundle itself?

The test mocks could return a mock bundle that `getSymbolicName()` that return `BUNDLE_NAME` (WRT below).

If we provide that separate functions one user of that interface could call `resolveBundle(someClass).getSymbolicName()`, the other one calls `resolveSymbicName()`.
As long as the methods return the same string, it does not matter, but the question would be why adding this redundant method then.
If the methods return different strings sometime, then we have also to provide a function `resolveSymbolicName(Bundle bundle)` that must be used instead of `bundle.getSymbolicName()`...

htreu: Yes, agreed. I will change the PR accordingly.

maggu2810:
At least some test failed (didn't check if it is related to your changes).

htreu:
It actually was related, lets see what Travis thinks about my update...

maggu2810:
It still fails.

htreu:
Those OSGiTests are killing me. ChangeThingTypeOSGiTest is next to be migrated...

maggu2810:
ping me once the test succeeded

maggu2810:
Thanks!

PR #5485, opened 2018-04-26
[DS Annotations] org.eclipse.smarthome.extensionservice.marketplace.a…

sjka:
…utomation


extensions/extensionservice/org.eclipse.smarthome.extensionservice.marketplace.automation/src/main/java/org/eclipse/smarthome/extensionservice/marketplace/automation/internal/MarketplaceRuleTemplateProvider.java
```diff
@@ -34,6 +38,8 @@
  * @author Kai Kreuzer - Initial contribution and API
  *
  */
+@Component(immediate = true, service = { MarketplaceRuleTemplateProvider.class,
+        RuleTemplateProvider.class }, property = { "Parser.target=(&(format=json)(parser.type=parser.template))" })
```
kaikreuzer: This `property` looks like an error here - we have it on the `setParser`method...

sjka: Looks indeed like my converter was fooled by what it got from SCR's parser - and I didn't notice it. Thanks for paying attention!

PR #5493, opened 2018-04-27
Hue: Cache sentColortemperature and sentBrightness in all cases 

### LINKED ISSUE #652: Hue binding should cache color temperature and brightness in all cases
migrated from Bugzilla [#479793](https://bugs.eclipse.org/bugs/show_bug.cgi?id=479793)
status UNCONFIRMED severity _minor_ in component _Binding_ for _---_
Reported in version _0.8.0_ on platform _All_
Assigned to: Project Inbox

On 2015-10-14 15:58:16 -0400, Dominic Lerbs wrote:

> As mentioned by Kai here (https://github.com/eclipse/smarthome/pull/342#issuecomment-112727297) the hue binding does not cache the color temperature and the brightness in case the values are changes using OnOffType or PercentType.
> 
> A clean solution to this problem probably needs some refactoring of the binding hence this seperate issue report.

triller-telekom:
Fixes #652 

### DIFF
--- a/extensions/binding/org.eclipse.smarthome.binding.hue/src/main/java/org/eclipse/smarthome/binding/hue/handler/HueLightHandler.java
+++ b/extensions/binding/org.eclipse.smarthome.binding.hue/src/main/java/org/eclipse/smarthome/binding/hue/handler/HueLightHandler.java
@@ -273,6 +273,8 @@ public class HueLightHandler extends BaseThingHandler implements LightStatusList
                 break;
         }
         if (lightState != null) {
+            lastSentBrightness = lightState.getBrightness();
+            lastSentColorTemp = lightState.getColorTemperature();
             hueBridge.updateLightState(light, lightState);
         } else {
             logger.warn("Command sent to an unknown channel id: {}", channelUID);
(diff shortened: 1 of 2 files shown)


extensions/binding/org.eclipse.smarthome.binding.hue/src/main/java/org/eclipse/smarthome/binding/hue/handler/HueLightHandler.java
```diff
@@ -273,6 +273,8 @@ public void handleCommand(ChannelUID channelUID, Command command) {
                 break;
         }
         if (lightState != null) {
+            lastSentBrightness = lightState.getBrightness();
```
kaikreuzer: I doubt that this is the correct fix - usually, lightState will only contain brightness OR colorTemperature information. Therefore I assume that you are actually deleting at least one cached information every time, which is not desired.

triller-telekom: Good point. I was not aware that either `colorTemperature` **OR** `brightness` where sent. I have changed my variables to `Integer` and only cache them if they have been set.

NorbertHD:
Is something wrong with this PR, that it is not yet commited?

kaikreuzer:
No, it is just that @triller-telekom seems to have missed to leave a comment that he pushed an update after the review... Will have another look at it.

PR #5497, opened 2018-04-27
Exclude auto generated XML for DMX binding from SAT checks

svilenvul:


kaikreuzer:
What is special about that file?

sjka:
As I was pointed to it in another context, I have fixed this in #5520.

PR #5503, opened 2018-04-29
Add the units mmHg and knot

lolodomo:

### DIFF
--- a/bundles/core/org.eclipse.smarthome.core/src/main/java/org/eclipse/smarthome/core/library/unit/SmartHomeUnits.java
+++ b/bundles/core/org.eclipse.smarthome.core/src/main/java/org/eclipse/smarthome/core/library/unit/SmartHomeUnits.java
@@ -12,6 +12,8 @@
  */
 package org.eclipse.smarthome.core.library.unit;
 
+import java.math.BigInteger;
+
 import javax.measure.Unit;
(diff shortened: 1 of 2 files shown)


bundles/core/org.eclipse.smarthome.core/src/main/java/org/eclipse/smarthome/core/library/unit/SmartHomeUnits.java
```diff
@@ -117,6 +120,8 @@ public String getName() {
     public static final Unit<RadiationDoseEffective> SIEVERT = addUnit(Units.SIEVERT);
     public static final Unit<CatalyticActivity> KATAL = addUnit(Units.KATAL);
     public static final Unit<Speed> METRE_PER_SECOND = addUnit(Units.METRE_PER_SECOND);
+    public static final Unit<Speed> KNOT = addUnit(new TransformedUnit<>("knot", Units.KILOMETRE_PER_HOUR,
```
htreu: according to the wikipedia the ISO standard symbol for knot is `kn`. See https://en.wikipedia.org/wiki/Knot_(unit)

lolodomo:
Change of unit symbol for knot done.

sjka:
With @htreu's comment being addressed, this LGTM.
Thanks!

PR #5505, opened 2018-04-29
[WU] Suppress hardcoded unit in channel description and documentation

lolodomo:


extensions/binding/org.eclipse.smarthome.binding.weatherunderground/ESH-INF/thing/thing-types.xml
```diff
@@ -344,26 +344,26 @@
 
 	<channel-type id="windDirection-degrees" advanced="true">
 		<item-type>Number:Angle</item-type>
-		<label>Wind Direction (°)</label>
-		<description>Wind direction in degrees</description>
+		<label>Wind Direction (angle)</label>
```
kaikreuzer: I would suggest to remove "(angle)" from the label - we do not have that on other labels either and the state will be shown next to it with a unit anyhow.

extensions/binding/org.eclipse.smarthome.binding.weatherunderground/ESH-INF/thing/thing-types.xml
```diff
@@ -344,26 +344,26 @@
...
 		<category>Wind</category>
 		<state readOnly="true" min="0" max="360" pattern="%.0f %unit%">
 		</state>
 	</channel-type>
 
 	<channel-type id="maxWindDirection-degrees" advanced="true">
 		<item-type>Number:Angle</item-type>
-		<label>Maximum Wind Direction (°)</label>
-		<description>Maximum wind direction in degrees</description>
+		<label>Maximum Wind Direction (angle)</label>
```
kaikreuzer: dito

htreu:
@lolodomo, please create another PR with Kai's comments addressed. Thanks.

extensions/binding/org.eclipse.smarthome.binding.weatherunderground/ESH-INF/thing/thing-types.xml
lolodomo: This is only to distinguish the 2 direction channels, one with an angle (90 for example) and the other with a text (NE for example).

kaikreuzer: I'd think that it is enough to show this distinction in the description - as the label is used in the UIs in the end (where this info is rather disturbing).

lolodomo: Then the user would see 2 channels with exactly the same label in Paper UI. Very confusing. And as getting the channel description in Paper UI is all except intuitive...

kaikreuzer: Hm, you are right - I thought that the description is shown on the Thing detail page, but it isn't (only hidden behind a long-press).
So ok, let's keep that distinction in the label then...

PR #5510, opened 2018-04-30
[metadata] remove existing metadata from generic provider

### LINKED ISSUE #5502: Metadata with key already exists from provider GenericMetadataProvider Error
While testing the newly added metadata infrastructure (#4390) in openHAB 2.3.0-SNAPSHOT Build #&#xfeff;1263, there seems to be a bug with refreshing/deleting metadata on already configured item. Each time I tried to update the metadata, it is failing with the below error and if I delete it, I can still see it present on that item when querying though the rest api. The only way to flush that information is to remove and re-add the item with the proper metadata.

**Error**
`[WARN ] [ore.common.registry.AbstractRegistry] - Metadata with key 'foo:TestSwitch' already exists from provider GenericMetadataProvider! Failed to add a second with the same UID from provider GenericMetadataProvider!
`

**Items Config**
`Switch TestSwitch "Test Switch" {foo="bar" [baz=42]}`

sjka:
...on model update. Before, it was not possible to update or delete meta-data (including tags) by changing the item DSL files.

fixes #5502

### DIFF
--- a/bundles/model/org.eclipse.smarthome.model.item.tests/src/org/eclipse/smarthome/model/item/internal/GenericItemProvider2Test.java
+++ b/bundles/model/org.eclipse.smarthome.model.item.tests/src/org/eclipse/smarthome/model/item/internal/GenericItemProvider2Test.java
@@ -243,4 +243,44 @@ public class GenericItemProvider2Test extends JavaOSGiTest {
         assertNull(res);
     }
 
+    @Test
+    public void testMetadataUpdate() {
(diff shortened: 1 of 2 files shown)


kaikreuzer:
No, they are not yet included, see https://github.com/eclipse/smarthome/wiki/ESH-used-in-openHAB.
Will do a new ESH stable tomorrow or Friday.

jsetton:
Great. Thanks for the clarification.

PR #5511, opened 2018-04-30
Activate ThingHandlerFactories on demand

### LINKED ISSUE #5163: Any need for immediate activation of ThingHandlerFactory services
The archetype currently sets `immediate = true` for the component implementation of `ThingHandlerFactory`.
Why do a `ThingHandlerFactory` service requires to be activated if there is no one that is interested in?

If there exists "always" an instance in every solution using ESH that tracks that services there is still no need to set it explicit.

sjka:
...instead of immediately.

Also:
* added `configurationPid=...` parameter where missing
* adapted archetype accordingly
* amended sample in the docu (binding how-to) accordingly

fixes #5163


lolodomo:
@sjka : Is it expected to not have configurationPid in the astro binding and to have name rather than configurationPid in the DMX binding ?

sjka:
No, thanks for spotting these! 
-> #5520

Issue #5460, opened 2018-04-24
[Sonos] Re-open {Fixes on UPnP infrastucture #4569} {UPnP action invocation results in an empty result, which sets the Thing to OFFLINE #4531}


kgoderis:
Just bumped my operational system to  2.3.0-SNAPSHOT Build 1263  and I ran into exactly the problems described in the issues mentioned here above.

kgoderis:
When trying to replicate the problem in the ESH IDE, it was present at first, but after some time all sonos devices got "resolved" as being online. The JUNP version in the IDE is 2.3.0.201711021118, whereas the latest snapshot on Karaf reports 2.3.0 

@kaikreuzer you initially rooted the problem to the Jetty client. Just doing the same, I found that the IDE runs Jetty 9.3.22, whereas the SNAPSHOT contains 9.3.15. Time for another upgrade of the Jetty bundles? (At the time you bumped 9.3.14 to 9.3.15 I am not mistaken)

maggu2810:
The used Jetty bundles are part of the product setup -- in your case openHAB.
Karaf 4.2.0 is already out that uses Jetty 9.4.x (works for me in general).
openHAB still uses Karaf 4.1.x (AFAIK) -- to bump the Jetty version only or the whole Karaf the openHAB distro repo would be the correct place.

kgoderis:
@maggu2810 Tx - posted it here as the original Issues originated here, and because it is related to the Sonos binding. @kaikreuzer I am not an expert on Karaf, i.e. not my decision on Karaf 4.2.0, but to me the issue is severe enough to either bump in the short term, or find a short term workaround for openHAB. wdyt?

kgoderis:
@kaikreuzer FYI I get a bunch of these:

```
2018-04-25 18:54:18.930 [DEBUG] [org.jupnp.transport.Router          ] - Sending via TCP unicast stream: (OutgoingActionRequestMessage) POST http://192.168.0.179:1400/DeviceProperties/Control
2018-04-25 18:54:18.950 [INFO ] [org.jupnp.transport.spi.StreamClient] - Timeout of 10 seconds while waiting for HTTP request to complete, aborting: (OutgoingActionRequestMessage) POST http://192.168.0.165:1400/DeviceProperties/Control
```

Something fishy is going on indeed

lolodomo:
No, I have not seen this issue again. I restarted my OH server this morniong and my Sonos things are online. I am running snapshot 1262.

[...]

PR #5516, opened 2018-04-30
changed hard dependency on httpclient feature to a requirement

kaikreuzer:
This is my first attempt to fix https://github.com/eclipse/smarthome/issues/5460, resp. https://github.com/openhab/openhab-distro/issues/692.

@maggu2810 I'd be grateful for your review/input.
What I noticed is that on an freshly installed openHAB distro (no addons), the feature `esh-tp-jupnp` is installed and active - and with it, the Jetty 9.3.15 bundles are already there. 
openHAB comes with its own [httpclient capability](https://github.com/openhab/openhab-core/blob/master/features/openhab-core/src/main/feature/feature.xml#L65-L76), so I'd hope that I can make this being picked up instead. What isn't clear to me yet though is why `esh-tp-jupnp` is at all installed, when no add-on is installed - I would have expected it to be something optional...

### DIFF
--- a/features/karaf/esh-tp/src/main/feature/feature.xml
+++ b/features/karaf/esh-tp/src/main/feature/feature.xml
@@ -156,6 +156,7 @@
 
   <feature name="esh-tp-jupnp" description="UPnP/DLNA library for Java" version="${project.version}">
     <capability>esh.tp;feature=jupnp;version=2.3.0</capability>
+    <requirement>esh.tp;filter:="(feature=httpclient)"</requirement>
     <feature dependency="true">http</feature>
     <feature dependency="true">scr</feature>
     <feature dependency="true">esh-tp-httpclient</feature>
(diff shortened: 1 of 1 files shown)


maggu2810:
The question is why is a bundle of `esh-tp-httpclient` necessary to be installed in your openHAB without addons? Does a bundle that is installed already require e.g. Jetty's http client implementation?
If this is the case, you should set the requirements of that imported Jetty packages to the version you are using in your openHAB product.

maggu2810:
```
 76 │ Active   │  80 │ 9.3.15.v20161220       │ mvn:org.eclipse.jetty/jetty-io/9.3.15.v20161220
 87 │ Active   │  80 │ 9.3.15.v20161220       │ mvn:org.eclipse.jetty/jetty-util/9.3.15.v20161220
 91 │ Active   │  80 │ 9.3.15.v20161220       │ mvn:org.eclipse.jetty.websocket/websocket-api/9.3.15.v20161220
 92 │ Active   │  80 │ 9.3.15.v20161220       │ mvn:org.eclipse.jetty.websocket/websocket-common/9.3.15.v20161220
```

I am pretty sure its about the WebSocket bundles. It seems you depend on it and the first one that suffers that requirements is the esh-tp-httpclient feature.

An easy solution is to set the min version to the one you would like to force on the imports.

kaikreuzer:
Thanks for the help @maggu2810.

I am closing this as it won't solve the issue at hand - as a workaround, I have instead created https://github.com/openhab/openhab-distro/pull/703.

Issue #5491, opened 2018-04-27
[WARN ] [ysfs.internal.PollingUsbSerialScanner] - A NoSuchFileException prevented a scan for USB serial devices: \sys\class\tty


PiscisSwimeatus:
Platform information:
OS: Windows 10 Enterprise (64bit)
Java Runtime Environment: Java Runtime Environment Version 8.0 Update 161 (32bit)
openHAB version: 2.3.0 SNAPSHOT 2018-04-23

Start from a clean copy
Add zwave binding
The following warning logs every 5 seconds
Except the warning everything seams to be working ok

[openhab.log](https://github.com/eclipse/smarthome/files/1953948/openhab.log)
[events.log](https://github.com/eclipse/smarthome/files/1953949/events.log)

kaikreuzer:
@hsudbrock Could you have a look at how that can be prevented?
Imho, the bundle should ideally not become active at all on Windows, right?

hsudbrock:
Thanks for pointing out this issue @PiscisSwimeatus.

@kaikreuzer I agree that the bundle should not become active at all on Windows. I think that, ideally, the bundle should not even be installed on a Windows system. I had hoped that this might be achievable using a requirement in a Karaf feature on the property `osgi.os`. However, I stumbled upon https://github.com/eclipse/smarthome/issues/1416, where it is stated that this property is not set within Karaf instances.

Hence, I see two points for improvement here:
1. Do net let the DS component `SysfsUsbSerialScanner` become active on Windows systems (or, more generally, on systems where the sysfs cannot be scanned). In consequence, the 5-second polling of the sysfs will not happen.
2. Move the bundle `org.eclipse.smarthome.config.discovery.usbserial.linuxsysfs` to its own Karaf feature (currently, it is contained in the Karaf feature that also contains `org.eclipse.smarthome.config.discovery.usbserial`).

I think that 1. should be implemented (and would also provide a PR for that). I am not yet sure whether, on top of the DS component not being active, the complete bundle should not be active.

When also implementing 2., I think that this would mean that one would have to explicitly install an additional feature for being able to use the USB-Serial discovery on Linux systems. This is why I am not sure whether this should be done (but could also provide a PR, if discussion result here is that we want this).

What do you think?

kaikreuzer:
I agree that option 2 is not ideal, because it would be the user who'd have to explicitly install that feature.

Ftr, I see the same warning on Mac as well:
```
[...]

PR #5521, opened 2018-05-02
[5491] Do not enable background discovery for sysfs-based USB-Serial discovery if scanned directories are not accessible

hsudbrock:
This PR targets https://github.com/eclipse/smarthome/issues/5491.

On non-Linux systems, the directories used by the sysfs-based USB-Serial discovery are not available (like, e.g., `/sys/class` or `/dev`). This commit disables the polling of the directory scanning for systems where the directories are not available or not readable.

### DIFF
--- a/bundles/config/org.eclipse.smarthome.config.discovery.usbserial.linuxsysfs/src/main/java/org/eclipse/smarthome/config/discovery/usbserial/linuxsysfs/internal/SysfsUsbSerialScanner.java
+++ b/bundles/config/org.eclipse.smarthome.config.discovery.usbserial.linuxsysfs/src/main/java/org/eclipse/smarthome/config/discovery/usbserial/linuxsysfs/internal/SysfsUsbSerialScanner.java
@@ -121,7 +121,7 @@ public class SysfsUsbSerialScanner implements UsbSerialScanner {
 
     @Override
     public boolean canPerformScans() {
-        return isReadAccessible(Paths.get(sysfsTtyDevicesDirectory)) && isReadAccessible(Paths.get(devDirectory));
+        return isReadable(Paths.get(sysfsTtyDevicesDirectory)) && isReadable(Paths.get(devDirectory));
     }
 
     /**
@@ -138,21 +138,13 @@ public class SysfsUsbSerialScanner implements UsbSerialScanner {
                 String serialPortName = sysfsTtyPath.getFileName().toString();
                 Path devicePath = Paths.get(devDirectory).resolve(serialPortName);
                 Path sysfsDevicePath = getSysfsDevicePath(sysfsTtyPath);
-                if (sysfsDevicePath != null && isReadWriteAccessible(devicePath)) {
+                if (sysfsDevicePath != null && isReadable(devicePath) && isWritable(devicePath)) {
                     result.add(new SerialPortInfo(devicePath, sysfsDevicePath));
                 }
             }
         }
 
         return result;
-    }
-
-    private boolean isReadWriteAccessible(Path path) {
-        return isReadAccessible(path) && isWritable(path);
-    }
-
-    private boolean isReadAccessible(Path path) {
-        return exists(path) && isReadable(path);
     }
 
     /**


bundles/config/org.eclipse.smarthome.config.discovery.usbserial.linuxsysfs.test/src/test/java/org/eclipse/smarthome/config/discovery/usbserial/linuxsysfs/internal/SysFsUsbSerialScannerTest.java
```diff
@@ -153,6 +153,32 @@ public void testNonUsbDeviceIsSkipped() throws IOException {
         assertThat(scanner.scan(), is(empty()));
     }
 
+    @Test
+    public void testCanPerformScans() {
+        // with given test setup, scans can be performed
+        assertThat(scanner.canPerformScans(), is(true));
```
triller-telekom: For simple true/false checks, can you please change it to `assertTrue`/`assertFalse`? Saves you a couple of characters and makes it more readable :) This also applies to the other occurrences.

bundles/config/org.eclipse.smarthome.config.discovery.usbserial.linuxsysfs/src/main/java/org/eclipse/smarthome/config/discovery/usbserial/linuxsysfs/internal/SysfsUsbSerialScanner.java
```diff
@@ -142,8 +147,12 @@ protected void modified(Map<String, Object> config) {
...
     }
 
-    private boolean isAccessible(Path devicePath) {
-        return exists(devicePath) && isWritable(devicePath) && isReadable(devicePath);
+    private boolean isReadWriteAccessible(Path path) {
+        return isReadAccessible(path) && isWritable(path);
+    }
+
+    private boolean isReadAccessible(Path path) {
+        return exists(path) && isReadable(path);
```
maggu2810: isReadable checks already if the path exists...
```
This method checks that a file exists and that this Java virtual machine has appropriate privileges that would allow it open the file for reading.
```

hsudbrock: Thanks for the hint, I will improve the code accordingly.

bundles/config/org.eclipse.smarthome.config.discovery.usbserial.linuxsysfs.test/src/test/java/org/eclipse/smarthome/config/discovery/usbserial/linuxsysfs/internal/SysFsUsbSerialScannerTest.java
hsudbrock: I disagree that the suggested change would render the code more readable, because I think that the phrase "assert that x is true" sounds more natural than "assert true x". I also think that this is question of personal preference, so I probably cannot convince you that my preference is the better choice :)

Is there some coding guideline regarding this for ESH? I just found the following on the [ESH testing page](https://www.eclipse.org/smarthome/documentation/development/testing.html) (which suggests the solution I chose):

> Using the the hamcrest matcher library is a good way to write expressive assertions. In contrast to the original assertion statements from JUnit the hamcrest matcher library allows to define the assertion in a more natural order

Should I still change this?

triller-telekom: Personal preference, leave it as it is :)

PR #5523, opened 2018-05-02
refresh internal widget collection on sitemap change

### LINKED ISSUE #5522: Sitemap updates the wrong widgets after reordering
Let's assume we have a sitemap like this:

```xtend
sitemap test label="Test Sitemap" {
    Frame label="Test Items" {
        Text item=testNumber
        Text item=testDimmer
    }
}
```

It's going to look like this in Basic UI:

image

Now I have a rule that increments the testNumber item's state on a regular basis. 

When I now change the order of the two widgets in the sitemap file like this:

```xtend
sitemap test label="Test Sitemap" {
    Frame label="Test Items" {
        Text item=testDimmer
        Text item=testNumber
    }
}
```

then the Basic UI refreshes automatically and correctly looks like this:

image

However, as soon as the numberItem state changes for the next time, it is going to look like this:

image

As you can see, the wrong widget gets updated.

After a quick analysis it looks like the `PageChangeListener` keeps operating with a stale collection of widgets and therefore calculates the widgetIDs wrong. 

This in fact is a regression introduced by #3846. I will push a fix which partially reverts #3846 in a few moments.

sjka:
This partially reverts #3846.

fixes #5522

### DIFF
--- a/bundles/model/org.eclipse.smarthome.model.sitemap/src/org/eclipse/smarthome/model/sitemap/SitemapProvider.java
+++ b/bundles/model/org.eclipse.smarthome.model.sitemap/src/org/eclipse/smarthome/model/sitemap/SitemapProvider.java
@@ -14,12 +14,13 @@ package org.eclipse.smarthome.model.sitemap;
 
 import java.util.Set;
 
+import org.eclipse.smarthome.model.core.ModelRepositoryChangeListener;
 
 public interface SitemapProvider {
 
     /**
      * This method provides access to sitemap model files, loads them and returns the object model tree.
-     * 
+     *
      * @param sitemapName the name of the sitemap to load
      * @return the object model tree, null if it is not found
      */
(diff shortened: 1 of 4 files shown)


sjka:
While doing some manual testing, I realized that there is another potential source for wrong widget ID calculations: The `SitemapSubscriptionService` and the `SitemapProvider` both registered themselves as `ModelRepositoryChangeListener`s at the `ModelRepository`. The former uses the latter to get a cached instance of the model. 
There is no guarantee though about the order in which they get their `modelChanged()` methods called. So in case the `SitemapSubscriptionService` got called first, it used the stale widget copy given by the `SitemapProvider`, which can't know anything about the new model at this time. 
I have added a second commit which fixes this aspect by letting the `SitemapSubscriptionService` subscribe at the `SitemapProvider` for changes instead of the `ModelRepository` directly, thereby making sure it only gets notified when its trusted provider also knows about the change.

bundles/model/org.eclipse.smarthome.model.sitemap/src/org/eclipse/smarthome/model/sitemap/SitemapProvider.java
```diff
@@ -14,21 +14,37 @@
...
      * @return names of provided sitemaps
      */
     public Set<String> getSitemapNames();
+
+    /**
+     * Add a listener which will be informed subsequently once a model has changed
+     *
+     * @param listener
+     */
+    public void addModelChangeListener(ModelRepositoryChangeListener listener);
```
triller-telekom: I know the other methods in this interface also have the `public` keyword, but they are superfluous in an interface...

PR #5524, opened 2018-05-02
MQTT embedded server: Export the service for testing of other artifacts

davidgraeff:
Introduce an EmbeddedBrokerService interface for the service implementation. Other artifacts like the MqttThings one could use the embedded broker for testing.


htreu:
Hi @davidgraeff. Sorry, I'm not convinced this is a good idea. Every bundle in the framework is now able to get hold of this service and stop the embedded broker. Also, is it necessary to fire up a full broker instance for binding tests? Aren't there callbacks which can be called in tests mocking the actual broker away? I guess we can assume that the broker works, right?

davidgraeff:
@htreu The unit tests are all done and working. But I liked one implementation test as well that ties everything MQTT together and is done already as well. But I need access to either Moquette to spin up a broker, or access to the service.

Because the serviceadds the broker connection to the MqttService,  in theory it is enough to request the connection from there. But at this point OSGI runlevels and timing gets important. I just wanted to avoid that situation. We do have a flaky MQTT unit test already (soon to be fixed), because of timing issues.

If you really don't like the idea to open up the embedded broker service a bit, I might just work with a higher timeout time (of 2sec or so). All necessary services and connections are probably ready by then.

davidgraeff:
It's enough to use MqttService. No need for this PR :)

PR #5526, opened 2018-05-03
Pass DiscoveryServiceCallback also to DiscoveryParticipants

### LINKED ISSUE #4630: ExtendedDiscoveryService not available for  MDNSDiscoveryParticipant
ExtendedDiscoveryService is not available for MDNSDiscoveryParticipant.
https://www.eclipse.org/forums/index.php/t/1090001/

In my use case it  means that in case where multiple discovery methods are available it can't be checked and double entries prevented.

_ExtendedDiscoveryService works only for DiscoveryService implementations currently (which in your case would be the MDSNDiscoveryService). The MDNSDiscoveryParticipants however only hook into them. I don't see why this same mechanism couldn't be extend for participants also. It would mean that the MDSNDiscoveryService needs to implement the ExtendedDiscoveryService interface and in turn applies the same mechanism as the DiscoveryServiceRegistryImpl to any MDNSDiscoveryParticipant - i.e. if these also implement ExtendedDiscoveryService then it forwards the DiscoveryServiceCallback instance to them (and accordingly for UPnP, of course). It just gotta be done cleanly in terms of OSGi so it won't leak any stale instances._

triller-telekom:
DiscoveryParticipants may also implement the ExtendedDiscoveryService
interface to obtain a DiscoveryServiceCallback.

Fixes #4630

### DIFF
--- a/bundles/config/org.eclipse.smarthome.config.discovery.mdns/src/main/java/org/eclipse/smarthome/config/discovery/mdns/internal/MDNSDiscoveryService.java
+++ b/bundles/config/org.eclipse.smarthome.config.discovery.mdns/src/main/java/org/eclipse/smarthome/config/discovery/mdns/internal/MDNSDiscoveryService.java
@@ -27,6 +27,8 @@ import org.eclipse.jdt.annotation.Nullable;
 import org.eclipse.smarthome.config.discovery.AbstractDiscoveryService;
 import org.eclipse.smarthome.config.discovery.DiscoveryResult;
 import org.eclipse.smarthome.config.discovery.DiscoveryService;
+import org.eclipse.smarthome.config.discovery.DiscoveryServiceCallback;
+import org.eclipse.smarthome.config.discovery.ExtendedDiscoveryService;
 import org.eclipse.smarthome.config.discovery.mdns.MDNSDiscoveryParticipant;
 import org.eclipse.smarthome.core.thing.ThingTypeUID;
 import org.eclipse.smarthome.core.thing.ThingUID;
@@ -48,13 +50,16 @@ import org.slf4j.LoggerFactory;
  * @author Andre Fuechsel - make {@link #startScan()} asynchronous
  */
 @Component(immediate = true, service = DiscoveryService.class, configurationPid = "discovery.mdns")
-public class MDNSDiscoveryService extends AbstractDiscoveryService implements ServiceListener {
+public class MDNSDiscoveryService extends AbstractDiscoveryService
+        implements ServiceListener, ExtendedDiscoveryService {
     private final Logger logger = LoggerFactory.getLogger(MDNSDiscoveryService.class);
 
     @Deprecated
     private final Set<org.eclipse.smarthome.io.transport.mdns.discovery.MDNSDiscoveryParticipant> oldParticipants = new CopyOnWriteArraySet<>();
 
     private final Set<MDNSDiscoveryParticipant> participants = new CopyOnWriteArraySet<>();
+
+    private DiscoveryServiceCallback discoveryServiceCallback;
 
     private MDNSClient mdnsClient;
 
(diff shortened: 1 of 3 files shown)


bundles/config/org.eclipse.smarthome.config.discovery.mdns/src/main/java/org/eclipse/smarthome/config/discovery/mdns/internal/MDNSDiscoveryService.java
```diff
@@ -147,14 +152,24 @@ private void scan() {
 
     @Reference(cardinality = ReferenceCardinality.MULTIPLE, policy = ReferencePolicy.DYNAMIC)
     protected void addMDNSDiscoveryParticipant(MDNSDiscoveryParticipant participant) {
+        if (participant instanceof ExtendedDiscoveryService) {
+            ((ExtendedDiscoveryService) participant).setDiscoveryServiceCallback(discoveryServiceCallback);
```
htreu: I doubt that this will work: when the participants are added to the MDNSDiscoveryService the `discoveryServiceCallback` is not necessarily set.

triller-telekom: That is true, but therefore I have the loop over all participants when the callback is set.

This code here is for the case that **later** at runtime a participant is added, so it receives the callback as well.

triller-telekom:
To answer your question: With the callback you can access previous results and do whatever you like with them, that depends on the use-case. But you are right, for detecting duplicates we have the `representationProperty`.

maggu2810:
Can you give us some examples of concrete use cases?
I just wonder why it has not been necessary before.

tmrobert8:
Use case - if the configuration contains an IP Address and that IP address changes, you need to update the prior result.

kaikreuzer:
@tmrobert8 That is a perfect example why we do NOT want to introduce this feature - because people believe they have to do such things. Instead, the framework already takes care of updating existing inbox results, when you re-discover the same Thing UID but with changed configuration parameters.

htreu:
We should reflect this in the docs. The mechanism of updating current inbox results and existing things is not part of the discovery service description. Also the ExtendedDiscoveryService is explicitly mentioned to get hold of existing results/things.

kaikreuzer:
Who is volunteering to do so?

tmrobert8:
I'd have no issues documenting this once it's been merged

kaikreuzer:
Maybe I have missed something, but as far as I understood, the documentation would actually document, why this PR here is NOT needed, so I don't see why we would have to wait for it to be merged...

tmrobert8:
@kaikreuzer 
No - my fault.  I thought you meant to document (ie javadoc) the existing classes to mention that inbox items will automatically be updated the thing UID.

As for this PR - I still think it's valuable to have (unless you tell me there is a better way!).  My comments about updating the config apply to the existing thing.  Example: let's say I have an MDNS item (the NEEO brain) that I have added to my system.  If the NEEO Brain acquires a new IP address, the MDNS discovery  will see the new IP address and I can update the thing's configuration directly.  Same applies to Sony when the stupid TV changes port numbers and the UPNP discovery is notified of the change.

triller-telekom:
@tmrobert8 What prevents you from doing this 

> If the NEEO Brain acquires a new IP address, the MDNS discovery will see the new IP address and I can update the thing's configuration directly

without this PR?  You simply create a new `DiscoveryResult` with the same `ThingUID` and your existing  `Thing` will get the new IP address.

tmrobert8:
@triller-telekom 
Never knew that (ie DiscoveryResult will update existing thing's configuration).  I have no idea what the use case is then for the PR if it updates both the inbox and existing thing items...

htreu:
Please see https://github.com/eclipse/smarthome/pull/5583 for an updated documentation.

With this I think we can close this one.

Issue #5483, opened 2018-04-26
Selectively auto-approve inbox entries


hsudbrock:
**Context**
Currently, the `o.e.sh.config.discovery.internal.AutomaticInboxProcessor` can be configured to automatically approve *all* added inbox entries (using the configuration flag `autoApprove`).

**Issue**
I would like to be able to automatically approve only selected inbox entries (in particular, only auto-approving inbox entries for selected thing types).

**Possible solutions**
I see two options for implementing this:
1. Adding configuration parameters for specifying which inbox entries should be auto-approved (E.g., a parameter specifying a whitelist of thing types for which inbox entries are auto-approved).
2. Providing a possibility to hook into the `AutomaticInboxProcessor`, so that bundles can participate in the decision whether an inbox result is automatically approved.

I prefer the second option, because it is quite flexible (unlike the first solution - unless one allows to encode complex expression for inbox entries in config parameters). 

**Suggested solution**
For the second solution, my proposal would be to:
* add an interface with a method `boolean autoApprove(discoveryResult)`;
* inject DS components implementing this interface into the `AutomaticInboxProcessor`;
* for each added inbox result, query each of these components whether the inbox result should be auto approved; auto approve it if at least one of the components votes for auto approval.

With this solution, the components implementing the interface specify a whitelist of discovery results that are to be automatically approved. 

These auto approvals would apply if the current property `autoApprove` is set to `false`; if `autoApprove` is set to `true`, all inbox entries will be auto-approved anyway. In consequence, the name `autoApprove` is not perfect anymore, as inbox entries can be auto approved even when this property is set to `false`. A better name for this property would then be `alwaysAutoApprove`, but this probably should not be changed as it would be API breaking.

kaikreuzer:
I'm ok with option 2 as long as the docs of this interface clearly state that it is only meant to be implemented by solutions, but not by add-ons.

> add an interface with a method boolean autoApprove(discoveryResult);

My suggestion would be that this interface extends `Predicate<DiscoveryResult>` - that makes it easy to be used within closures.

>  the components implementing the interface specify a whitelist of discovery results that are to be automatically approved.

The alternative would be to veto auto-approval. In that case `autoApprove=false` would be the same as before and `autoApprove=true` would be slightly enhanced by giving solutions a chance to not apply it blindly to any result. This would imho be the smoother extension of the current feature.

hsudbrock:
[...]

PR #5529, opened 2018-05-03
[5483] Selective auto-approval of inbox entries

hsudbrock:
This PR addresses https://github.com/eclipse/smarthome/issues/5483.

With this PR, bundles can hook into the auto-approval of discovery results performed by the `o.e.sh.config.discovery.internal.AutomaticInboxProcessor`. To hook into it, they must provide an implementation of the `o.e.sh.config.discovery.InboxAutoApprovePredicate`, which determines based on the discovery result whether it should be automatically approved.

As indicated in the JavaDoc of the `InboxAutoApprovePredicate`, this is intended to be used by solutions integrating Eclipse SmartHome, but not for Eclipse SmartHome addons (like, e.g., bindings).

One open question for me is how to deal with "evil" `InboxAutoApprovePredicate`s that throw `RuntimeException`s. In this PR, it is implemented such that if this happens, then all remaining `InboxAutoApprovePredicate`s will be skipped, as the exception is not caught. Should this exception be caught (using a `catch (RuntimeException e)` in `AutomaticInboxProcessor.isToBeAutoApproved`)? Or can we live with that - in particular as such predicates will only be provided by solutions and not by ESH addons? (There is a currently ignored test illustrating this (edge?) case - I will un-ignore respectively remove it depending on feedback in this PR.)

 I would be happy about any other feedback as well!

### DIFF
--- a/bundles/config/org.eclipse.smarthome.config.discovery.test/src/test/java/org/eclipse/smarthome/config/discovery/internal/AutomaticInboxProcessorTest.java
+++ b/bundles/config/org.eclipse.smarthome.config.discovery.test/src/test/java/org/eclipse/smarthome/config/discovery/internal/AutomaticInboxProcessorTest.java
@@ -15,12 +15,13 @@ package org.eclipse.smarthome.config.discovery.internal;
 import static org.eclipse.smarthome.config.discovery.inbox.InboxPredicates.withFlag;
 import static org.hamcrest.CoreMatchers.*;
 import static org.junit.Assert.assertThat;
-import static org.mockito.Matchers.any;
-import static org.mockito.Matchers.eq;
-import static org.mockito.Mockito.when;
+import static org.mockito.ArgumentMatchers.*;
+import static org.mockito.ArgumentMatchers.any;
+import static org.mockito.Mockito.*;
 import static org.mockito.MockitoAnnotations.initMocks;
 
 import java.util.Collections;
+import java.util.HashMap;
 import java.util.List;
 import java.util.Map;
 import java.util.stream.Collectors;
@@ -47,11 +48,13 @@ import org.eclipse.smarthome.core.thing.type.ThingTypeBuilder;
 import org.eclipse.smarthome.core.thing.type.ThingTypeRegistry;
 import org.eclipse.smarthome.test.storage.VolatileStorageService;
 import org.junit.Before;
+import org.junit.Ignore;
 import org.junit.Test;
 import org.mockito.Mock;
 
 /**
  * @author Andre Fuechsel - Initial contribution
+ * @author Henning Sudbrock - Added tests for auto-approving inbox entries
  */
 public class AutomaticInboxProcessorTest {
 
@@ -82,7 +85,7 @@ public class AutomaticInboxProcessorTest {
 
     private final static Configuration CONFIG = new Configuration(Collections.singletonMap(CONFIG_KEY, CONFIG_VALUE));
 
-    private AutomaticInboxProcessor inboxAutoIgnore;
+    private AutomaticInboxProcessor automaticInboxProcessor;
     private PersistentInbox inbox;
 
     @Mock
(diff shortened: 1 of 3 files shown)


eclipsewebmaster:
Issue tracker reference:
https://bugs.eclipse.org/bugs/show_bug.cgi?id=5483

bundles/config/org.eclipse.smarthome.config.discovery/src/main/java/org/eclipse/smarthome/config/discovery/inbox/InboxAutoApprovePredicate.java
```diff
@@ -0,0 +1,39 @@
...
+ * automatically approved (e.g., because another such {@link Predicate} returned <code>true</code>) - i.e., it is not
+ * possible to veto the automatic approval of a {@link DiscoveryResult}.
+ * <p/>
+ * Please note that this interface is intended to be implemented by solutions integrating Eclipse SmartHome. This
+ * interface is <em>not</em> intended to be implemented by Eclipse SmartHome addons (like, e.g., bindings).
+ *
+ * @author Henning Sudbrock - initial contribution
+ */
+public interface InboxAutoApprovePredicate extends Predicate<DiscoveryResult> {
+
```
triller-telekom: empty line

bundles/config/org.eclipse.smarthome.config.discovery.test/src/test/java/org/eclipse/smarthome/config/discovery/internal/AutomaticInboxProcessorTest.java
```diff
@@ -374,10 +380,85 @@ public void testThingWithConfigIsBeingRemoved() {
...
+    @Test
+    public void testAutomaticDiscoveryResultApprovalIfInboxEntriesExistBeforeApprovalPredicatesAreAdded() {
+        inbox.add(DiscoveryResultBuilder.create(THING_UID).build());
+        inbox.add(DiscoveryResultBuilder.create(THING_UID2).build());
+        inbox.add(DiscoveryResultBuilder.create(THING_UID3).build());
+
+        // Adding this inboxAutoApprovePredicate will auto-approve the first two discovery results as they have matching
+        // thing type UID.
+        automaticInboxProcessor.addInboxAutoApprovePredicate(
+                discoveryResult -> discoveryResult.getThingTypeUID().equals(THING_TYPE_UID));
```
triller-telekom: Why does a result with ID `THING_UID2` match `THING_UID`?

bundles/config/org.eclipse.smarthome.config.discovery.test/src/test/java/org/eclipse/smarthome/config/discovery/internal/AutomaticInboxProcessorTest.java
```diff
@@ -374,10 +380,85 @@ public void testThingWithConfigIsBeingRemoved() {
...
+        verify(thingRegistry, times(1)).add(argThat(thing -> thing.getUID().equals(THING_UID2)));
+        verify(thingRegistry, times(1)).add(argThat(thing -> thing.getUID().equals(THING_UID3)));
+    }
+
+    @Test
+    public void testAlwaysAutoApproveInboxEntries() {
+        // Before setting the always auto approve property, existing inbox results are not approved
+        inbox.add(DiscoveryResultBuilder.create(THING_UID).build());
+
+        verify(thingRegistry, never()).add(argThat(thing -> thing.getUID().equals(THING_UID)));
```
triller-telekom: Can you be sure that `thingRegistry.add` was never called here? It might take a while until the inbox informs the registry, i.e. calls `add` on it, doesn't it?

bundles/config/org.eclipse.smarthome.config.discovery.test/src/test/java/org/eclipse/smarthome/config/discovery/internal/AutomaticInboxProcessorTest.java
```diff
@@ -374,10 +380,85 @@ public void testThingWithConfigIsBeingRemoved() {
...
+
+        verify(thingRegistry, times(1)).add(argThat(thing -> thing.getUID().equals(THING_UID)));
+
+        // Newly added inbox results are also approved.
+        inbox.add(DiscoveryResultBuilder.create(THING_UID2).build());
+        verify(thingRegistry, times(1)).add(argThat(thing -> thing.getUID().equals(THING_UID2)));
+    }
+
+    @Test
+    @Ignore("Should this test pass? It will fail currently, as RuntimeExceptions are not explicitly caught in AutomaticInboxProcessor#isToBeAutoApproved")
```
triller-telekom: I think its OK for this test to fail. As you already stated before, this `Predicate` is to be implemented by solutions and if they create a failing `Predicate` due to a uncaught `RuntimeException` the framework continues to work. So I think that solutions should take care of these `RuntimeException`s.

bundles/config/org.eclipse.smarthome.config.discovery.test/src/test/java/org/eclipse/smarthome/config/discovery/internal/AutomaticInboxProcessorTest.java
hsudbrock: The predicate matches thing types UIDs, not thing UIDs; both `THING_UID` and `THING_UID2` have the same thing type UID.

bundles/config/org.eclipse.smarthome.config.discovery.test/src/test/java/org/eclipse/smarthome/config/discovery/internal/AutomaticInboxProcessorTest.java
hsudbrock: As far as I see it, this actually all happens within the same thread, so `thingRegistry.add` is called before `inbox.add` terminates.

PR #5535, opened 2018-05-05
Add MqttBrokerConnection future API. Add tests. Add MqttService interface

### LINKED ISSUE #5370: MQTT: Add protocol selection to MqttBrokerConnection

### LINKED ISSUE #5527: [test failures] MQTT: BrokerHandlerTest fails randomly
The `BrokerHandlerTest#handlerInit` test case fails randomly. This may be due to timing issues.

davidgraeff:
This is an API breaking commit. The current API has some serious design flaws that can cause timing issues. At the end of this post is a list of PRs that adapt all mqtt bundles to the new API.

### Design flaws:
* Subscribe (called addConsumer) is an async method. But there was no way to register a callback.
* Stop() calls an async disconnect method. But the method returns before the disconnect process is finished and there is no way to get notified about the process.
* MqttMessageSubscriber was not only responsible for receiving messages for a subscribed topic, but also needed to provide that topic. It was not possible to use one MqttMessageSubscriber for multiple topics of interest.
* It is supported to subscribe to topics and then connect to a broker. But the start() method didn't
  return any feedback if subscriptions are failing (see trySubscribe, it only logs).
* In messageArrived: The MqttMessageSubscribers list was looped and subscribers were directly notified
  of a message. If a subscriber altered the list by for example adding another subscriber or removing itself, a ConcurrentModification exception was to be expected.

### Is it possible to preserve the current API?
`stop()` and `subscribe()` can be turned into synchronous methods and documented as such. Because massive subscribing to topics (like with the homie mqtt convention thing support) is a valid use-case, a synchronous only API for subscribing is not satisfying though.

### Changes:
* Rename addConsumer/removeConsumer to subscribe/unsubscribe
* Separate MqttService and MqttServiceImpl
* Remove MqttPublishResult
* Simplify MqttMessageSubscriber
* Move private classes out of MqttBrokerConnection if applicable
* Introduce CompletableFuture based API for MqttBrokerConnection
* Add another publish method where qos and retain can be specified.
* Don't return the deliveryToken message ID for publish. This is paho specific,  and the API should stay library agnostic.
* Clean up in stop(), unsubscribe from all topics, correctly close the persistence datastore and client.
* Add protocol selection (Websocket, TCP)

### Test additions:
* Add wildcard (+,#) tests and fix related wildcard topic match bug
* Add test for connecting, disconnecting multiple times
* Add tests for subtopic, complicated utf8 topic name subscriptions before and after a connection is established

### Adapt other Mqtt bindings
* Embedded broker: #5536 
* Mqtt main binding (broker configuration): #5537 
* Mqtt generic things and homie things: #5450 

### API overview of MqttBrokerConnection
```java
    /**
     * Add a new message consumer to this connection. Multiple subscribers with the same
     * topic are allowed. This method will not protect you from adding a subscriber object
     * multiple times!
     *
     * @param consumer Consumer to add
     * @throws MqttException If connected and the subscribe fails, this exception is thrown.
     */
    public CompletableFuture<Boolean> subscribe(String topic, MqttMessageSubscriber subscriber);

    /**
     * Subscribes to a topic on the given connection, but does not alter the subscriber list.
     */
    protected CompletableFuture<Boolean> subscribeRaw(String topic);

    /**
     * Remove a previously registered consumer from this connection.
     *
     * @param subscriber to remove.
     */
    @SuppressWarnings({ "null", "unused" })
    public CompletableFuture<Boolean> unsubscribe(String topic, MqttMessageSubscriber subscriber);

    /**
     * Unsubscribes a topic on the given connection, but does not alter the subscriber list.
     *
     * @param client The client connection
     * @param topic The topic to unsubscribe from
     * @return Completes with true if successful. Exceptionally otherwise.
     */
    protected CompletableFuture<Boolean> unsubscribeRaw(MqttAsyncClient client, String topic);

    /**
     * Unsubscribe from all topics
     *
     * @return Returns a future that completes as soon as all subscriptions have been canceled.
     */
    public CompletableFuture<Void> unsubscribeAll() ;

    /**
     * This will establish a connection to the MQTT broker and if successful, notify all
     * publishers and subscribers that the connection has become active. This method will
     * do nothing if there is already an active connection.
     *
     * @return Returns a future that completes with true if already connected or connecting,
     *         completes with false if a connection timeout has happened and completes exceptionally otherwise.
     */
    public CompletableFuture<Boolean> start();

    /**
     * Unsubscribes from all subscribed topics, stops the reconnect strategy, disconnect and close the client.
     *
     * You can re-establish a connection calling {@link #start()} again. Do not call start, before the closing process
     * has finished completely.
     *
     * @return Returns a future that completes as soon as the disconnect process has finished.
     */
    public CompletableFuture<Boolean> stop();

    /**
     * Publish a message to the broker.
     *
     * @param topic The topic
     * @param payload The message payload
     * @return Returns a future that completes with a result of true if the publishing succeeded and completes
     *         exceptionally on an error or with a result of false if no broker connection is established.
     */
    public CompletableFuture<Boolean> publish(String topic, byte[] payload);

    /**
     * Publish a message to the broker with the given QoS and retained flag.
     *
     * @param topic The topic
     * @param payload The message payload
     * @param qos The quality of service for this message
     * @param retain Set to true to retain the message on the broker
     * @param listener An optional listener to be notified of success or failure of the delivery.
     * @return Returns a future that completes with a result of true if the publishing succeeded and completes
     *         exceptionally on an error or with a result of false if no broker connection is established.
     */
    public CompletableFuture<Boolean> publish(String topic, byte[] payload, int qos, boolean retain);
```

Fixes #5370
Fixes #5527 


sjka:
And here your daily build failure readout:

```
[ERROR] Failed to execute goal org.eclipse.tycho:tycho-compiler-plugin:1.1.0:compile (default-compile) on project org.eclipse.smarthome.binding.mqtt: Compilation failure: Compilation failure: 
[ERROR] /home/travis/build/eclipse/smarthome/extensions/binding/org.eclipse.smarthome.binding.mqtt/src/main/java/org/eclipse/smarthome/binding/mqtt/handler/AbstractBrokerHandler.java:[80] 
[ERROR] 	} catch (ConfigurationException | MqttException e) {
[ERROR] 	         ^^^^^^^^^^^^^^^^^^^^^^
[ERROR] Unreachable catch block for ConfigurationException. This exception is never thrown from the try statement body
[ERROR] /home/travis/build/eclipse/smarthome/extensions/binding/org.eclipse.smarthome.binding.mqtt/src/main/java/org/eclipse/smarthome/binding/mqtt/handler/AbstractBrokerHandler.java:[80] 
[ERROR] 	} catch (ConfigurationException | MqttException e) {
[ERROR] 	                                  ^^^^^^^^^^^^^
[ERROR] Unreachable catch block for MqttException. This exception is never thrown from the try statement body
[ERROR] -> [Help 1]
[ERROR] 
```

You've now got 5 MQTT related PRs open which are all failing because they are either simply broken or they are so entangled with each other. It's now the time to slow down a little and recover order, I'd say. Otherwise I don't see the MQTT stuff making it into the [openHAB 2.3](https://github.com/eclipse/smarthome/pull/5450#issuecomment-386522119) release.

davidgraeff:
Each pr touches a different bundle, and this one is the basic building block for all others to work. It is unfortunately the most contradicting one as it breaks API.

That's why I wrote a lengthy description to discuss the proposed changes. 

PRs that I push are always building and passing all tests with the eclipse build system. I didn't had time to run maven. It literally takes ages on my system unfortunately.

I'll resolve whatever Travis finds over time though.

davidgraeff:
@htreu This is the PR that adds the future API. I stripped the tests (and will add those in a later PR again) to bring it down to under 1k of additional LOCs. This is still an API breaking PR.

htreu:
What is the merge order you suggest then? We will not merge a broken aka none-compiling PR.
Either this one "fixes" all the existing stuff and also Travis agrees, or the MQTT things PR gets merged on the old API first (also with Travis giving it a go).

davidgraeff:
I see not benefit for the user if ESH MQTT including the Things is released with OpenHab but unstable. This will just cause a hell lot of bug reports, depending on how prominent you want to announce it. I can provide a different* set of PRs, but not this week as I'm going on vacation around midweek. 

*: At the moment all referenced PRs are required to satisfy travis. If I pull the embedded broker and broker connection configuration changes into this on to make travis happy, a CQ is required (>1200 LOC).

htreu:
@davidgraeff please merge your open refactoring (= new API) PRs into a single one so we can continue review.

davidgraeff:
I would say today is a good day to do the PR merge. Should I keep the two topics (future API and adaption + new Mqtt Things Bundle) separated or should I make one big MQTT PR? A CQ is probably either required for one or for two PRs as a result.

maggu2810:
As written above this PR needs to be "complete" to be merged.
Complete in that case means the master still builds and works.
From my understanding this PR has been in the state that it breaks the build. Has this been fixed?

davidgraeff:
@maggu2810 I have been unclear with my comment. I mean I will probably have time to merge my PRs into one (or two, that's what the question is about) to make the build work. It was meant to be a fast response to the question what the timeline looks like.

maggu2810:
@davidgraeff Thank you for your fast response :wink:

davidgraeff:
Could someone reopen this for travis?

kaikreuzer:
I've re-triggered Travis.

davidgraeff:
```
[INFO] Eclipse SmartHome MQTT Transport ................... SUCCESS [  2.016 s]
[INFO] Eclipse SmartHome Mqtt Transport I/O Tests ......... SUCCESS [  3.381 s]

[INFO] MQTT Binding ....................................... SUCCESS [  2.285 s]
[INFO] MqttBroker Binding Tests ........................... SUCCESS [  4.674 s]

[INFO] Eclipse SmartHome Embedded MQTT Broker ............. SUCCESS [  1.825 s]
[INFO] Eclipse SmartHome Embedded MQTT Broker Tests ....... SUCCESS [  5.811 s]

[INFO] Eclipse SmartHome Extensions Feature ............... SUCCESS [  0.258 s]
[INFO] Eclipse SmartHome Karaf Feature Verification ....... FAILURE [01:47 min]
[INFO] Eclipse SmartHome Products ......................... SKIPPED
```

All Mqtt related bundles work. The naming isn't consistent apparently.

davidgraeff:
Because moquette 0.11 has been released, I'd like to update the "Embedded MQTT Broker" bundle before the CQ.

htreu:
@davidgraeff the moquette update requires yet another CQ for the eclipse IP validation. And the CQ for 0.10 is still in a "awaiting_analysis" state. How would the update affect the new API? Why do you think its blocking this PR?

kaikreuzer:
I remember that @davidgraeff spoke about memory leaks in 0.10, why he very much awaited the 0.11 release.
Imho it should be fine to create a new CQ for 0.11 - we can leave a comment on the old one that no detailed triage is required and the IP team should instead look at the 0.11 version.

htreu:
Thanks @kaikreuzer, no problem on the CQ side then.
@davidgraeff awaiting your PR and happy to issue the CQ.

davidgraeff:
Can someone look at the features file please? Travis continues to report errors related to the embedded broker feature. See last unsuccessful build: https://travis-ci.org/eclipse/smarthome/builds/391605273

htreu:
Feature verification fails du to this package import: https://github.com/eclipse/smarthome/pull/5535/commits/1f20aa14aee20eece04f17b61c898dc3c295671e#diff-d812dd30820dc08af04947c62b49b159R23

We do not provide `org.osgi.service.device` with the targetplatform. And when I understand your code right you do not use it either.

Regarding the moquette update: Is it affecting the current code? Please author a separate PR for the update. We can then directly create appropriate CQs.

htreu:
If stuff is independent I like to have separate PRs so we can operate independent. Its much more easy to track changes later.

htreu:
FTR: CQ 16750 created for moquette-broker-0.11

htreu:
@davidgraeff any news on this one?

davidgraeff:
Has CQ 16750 been accepted? Or should I still exclude the 0.11 moquette update from this PR?

htreu:
I would like to treat the version bump independent from the API update.

FTR: CQ 16750 was withdrawn by request from the Eclipse IP Team and CQ 16090 was updated to moquette-broker-0.11.

CQ 16090 just received checkin approval, so please author a separate PR for the moquette version bump. We can then directly merge it.

davidgraeff:
@htreu Travis passes now

htreu:
A new review will start soon(tm).

extensions/binding/org.eclipse.smarthome.binding.mqtt.test/org.eclipse.smarthome.binding.mqtt.test.launch
```diff
@@ -13,6 +13,7 @@
 <stringAttribute key="configLocation" value="${workspace_loc}/.metadata/.plugins/org.eclipse.pde.core/pde-junit"/>
 <booleanAttribute key="default" value="false"/>
 <booleanAttribute key="includeOptional" value="false"/>
+>>>>>>> Adapt to future API. Make tests super solid
```
davidgraeff: omg

bundles/io/org.eclipse.smarthome.io.transport.mqtt/src/main/java/org/eclipse/smarthome/io/transport/mqtt/internal/TopicSubscribers.java
```diff
@@ -0,0 +1,35 @@
...
+ */
+public class TopicSubscribers extends ArrayList<MqttMessageSubscriber> {
+    private static final long serialVersionUID = -2969599983479371961L;
+    final String regexMatchTopic;
+
+    public TopicSubscribers(String topic) {
+        this.regexMatchTopic = StringUtils.replace(StringUtils.replace(Matcher.quoteReplacement(topic), "+", "[^/]*"),
+                "#", ".*");
+    }
+}
```
htreu: missing newline

extensions/io/org.eclipse.smarthome.io.mqttembeddedbroker.test/src/test/java/org/eclipse/smarthome/io/mqttembeddedbroker/internal/MqttEmbeddedBrokerServiceTest.java
```diff
@@ -84,7 +84,7 @@ public void checkServerTrusted(X509Certificate[] chain, String authType, SSLEngi
         }
     }
 
-    private EmbeddedBrokerService subject;
+    private EmbeddedBrokerServiceImpl subject;
```
htreu: should the class be renamed to `EmbeddedBrokerServiceImplTest` then?

htreu:
While you are at it, please also review the `NetworkDiscoveryService`:

https://github.com/davidgraeff/smarthome/blob/a308c59c67effc03d901ee9e64503983e7a24b23/extensions/binding/org.eclipse.smarthome.binding.mqtt/src/main/java/org/eclipse/smarthome/binding/mqtt/internal/discovery/NetworkDiscoveryService.java#L154-L156

These lines give me the following exception:

```
java.lang.IllegalArgumentException: Could not parse [192.168.0.112]
	at org.apache.commons.net.util.SubnetUtils.calculate(SubnetUtils.java:240)
	at org.apache.commons.net.util.SubnetUtils.<init>(SubnetUtils.java:52)
	at org.eclipse.smarthome.binding.mqtt.internal.discovery.NetworkDiscoveryService.lambda$2(NetworkDiscoveryService.java:155)
```

`SubnetUtils#calculate` expects an address in the form of `192.168.0.1/16`.

maggu2810:
> As far as I can overlook all the changes it looks good. Would be nice to have a 2nd opinion, maybe @maggu2810 can step in here, too.

Sure, should I wait after your comments (and the ones of @davidgraeff itselves) are incorporated? IMO would make sense...

davidgraeff:
Would be nice if you could review now, so that I only have one round of changes to check

maggu2810:
> Would be nice if you could review now, so that I only have one round of changes to check

I will give it a try the next days as soon as there is some spare time.
But to be clear: All the fixes that will be applied after our review, needs to be reviewed again. If it consists of more then only a few lines and change e.g. some logic, ... we need to review the whole PR again.
So, I hope the changes @htreu requested are only minor ones :wink:

extensions/io/org.eclipse.smarthome.io.mqttembeddedbroker/src/main/java/org/eclipse/smarthome/io/mqttembeddedbroker/internal/EmbeddedBrokerService.java
```diff
@@ -12,145 +12,17 @@
...
-    }
-
-    @Override
-    public void brokerRemoved(String brokerID, MqttBrokerConnection broker) {
-        if (broker == connection) {
-            service.addBrokerConnection(brokerID, broker);
-        }
-    }
-
+public interface EmbeddedBrokerService {
```
maggu2810: Can you mark that interface as `@NonNullByDefault`?

extensions/io/org.eclipse.smarthome.io.mqttembeddedbroker/src/main/java/org/eclipse/smarthome/io/mqttembeddedbroker/internal/EmbeddedBrokerService.java
```diff
@@ -161,130 +33,16 @@ public void brokerRemoved(String brokerID, MqttBrokerConnection broker) {
...
-        server.startServer(new MemoryConfig(properties), null, null, authentificator, authorizer);
-        // }
-
-        this.server = server;
-        metrics.setServer(server);
-        ScheduledExecutorService s = new ScheduledThreadPoolExecutor(1);
-        detectStart.startBrokerStartedDetection(port, s);
-    }
+    void startEmbeddedServer(Integer portParam, boolean secure, String username, String password,
+            @NonNull String persistenceFilenameParam) throws IOException;
```
maggu2810: We don't use `@NonNull`, we are using `@NonNullByDefault` on classes / interfaces and @Nullable on return values, arguments, ...

extensions/io/org.eclipse.smarthome.io.mqttembeddedbroker/src/main/java/org/eclipse/smarthome/io/mqttembeddedbroker/internal/EmbeddedBrokerService.java
```diff
@@ -161,130 +33,16 @@ public void brokerRemoved(String brokerID, MqttBrokerConnection broker) {
...
-            connectionStateChanged(MqttConnectionState.CONNECTED, null);
-        }
-
-    }
-
-    @Override
-    public void connectedClientIDs(Collection<String> clientIDs) {
-        logger.debug("Connected clients: {}", clientIDs.stream().collect(Collectors.joining(", ")));
-    }
+    public MqttBrokerConnection getConnection();
```
maggu2810: Methods of interface are implicit public, please remove it.

extensions/io/org.eclipse.smarthome.io.mqttembeddedbroker/src/main/java/org/eclipse/smarthome/io/mqttembeddedbroker/internal/EmbeddedBrokerServiceImpl.java
```diff
@@ -0,0 +1,304 @@
...
+    // private NetworkServerTls networkServerTls; //TODO wait for NetworkServerTls implementation
+
+    protected Server server;
+    private final Logger logger = LoggerFactory.getLogger(EmbeddedBrokerServiceImpl.class);
+    protected MqttEmbeddedBrokerDetectStart detectStart = new MqttEmbeddedBrokerDetectStart(this);
+    protected MqttEmbeddedBrokerMetrics metrics = new MqttEmbeddedBrokerMetrics(this);
+
+    private MqttBrokerConnection connection;
+
+    @Reference(cardinality = ReferenceCardinality.MANDATORY)
```
maggu2810: `cardinality = ReferenceCardinality.MANDATORY` is the default for `@Reference`, please remove

extensions/io/org.eclipse.smarthome.io.mqttembeddedbroker/src/main/java/org/eclipse/smarthome/io/mqttembeddedbroker/internal/EmbeddedBrokerServiceImpl.java
```diff
@@ -0,0 +1,304 @@
...
+            connection.setCredentials(config.username, config.password);
+        }
+
+        // Start embedded server
+        try {
+            startEmbeddedServer(config.port, config.secure, config.username, config.password, config.persistenceFile);
+        } catch (IOException e) {
+            logger.debug("Could not start embedded broker", e);
+            logger.debug(e.getLocalizedMessage());
+            return;
```
maggu2810: Does it make sense to hide the not startable embedded server and keep the OSGi framework informed that the activation has been done successfully? Other services that rely on this one are activated after that. If you throw the exception so the caller of the activate method (the DS implementation) knows that there has been something wrong and will not activate services that depends on this one...

extensions/io/org.eclipse.smarthome.io.mqttembeddedbroker/src/main/java/org/eclipse/smarthome/io/mqttembeddedbroker/internal/EmbeddedBrokerServiceImpl.java
```diff
@@ -0,0 +1,304 @@
...
+     * @param port The broker port.
+     * @param secure Allow only secure connections if true or only plain connections otherwise.
+     * @param username Broker authentication user name. May be null.
+     * @param password Broker authentication password. May be null.
+     * @param persistence_filename The filename were persistent data should be stored.
+     * @throws IOException If any error happens, like the port is already in use, this exception is thrown.
+     */
+    @Override
+    public void startEmbeddedServer(Integer portParam, boolean secure, String username, String password,
+            @NonNull String persistenceFilenameParam) throws IOException {
```
maggu2810: see interface comment

extensions/io/org.eclipse.smarthome.io.mqttembeddedbroker/src/main/java/org/eclipse/smarthome/io/mqttembeddedbroker/internal/EmbeddedBrokerServiceImpl.java
```diff
@@ -0,0 +1,304 @@
...
+     * Starts the embedded broker.
+     *
+     * @param port The broker port.
+     * @param secure Allow only secure connections if true or only plain connections otherwise.
+     * @param username Broker authentication user name. May be null.
+     * @param password Broker authentication password. May be null.
+     * @param persistence_filename The filename were persistent data should be stored.
+     * @throws IOException If any error happens, like the port is already in use, this exception is thrown.
+     */
+    @Override
```
maggu2810: If a method of an interface is implemented we already inherit the JavaDoc of the interface method.
Overwriting that comment makes only sense if you need to add additional information.
As a consumer of the interface implementation WRT to OSGi services does not know about the specific implementation it should be removed completely as the behavior should be exactly the ones described by the interface.

extensions/io/org.eclipse.smarthome.io.mqttembeddedbroker/src/main/java/org/eclipse/smarthome/io/mqttembeddedbroker/internal/EmbeddedBrokerServiceImpl.java
```diff
@@ -0,0 +1,304 @@
...
+        this.server = server;
+        metrics.setServer(server);
+        ScheduledExecutorService s = new ScheduledThreadPoolExecutor(1);
+        detectStart.startBrokerStartedDetection(port, s);
+    }
+
+    /**
+     * Stops the embedded broker, if it is started.
+     */
+    @Override
```
maggu2810: See above, remove comment.

extensions/io/org.eclipse.smarthome.io.mqttembeddedbroker/src/main/java/org/eclipse/smarthome/io/mqttembeddedbroker/internal/EmbeddedBrokerServiceImpl.java
```diff
@@ -0,0 +1,304 @@
...
+
+    @Override
+    public void connectedClientIDs(Collection<String> clientIDs) {
+        logger.debug("Connected clients: {}", clientIDs.stream().collect(Collectors.joining(", ")));
+    }
+
+    /**
+     * Returns the MQTT broker connection, connected to the embedded broker
+     */
+    @Override
```
maggu2810: See above, remove comment.

extensions/binding/org.eclipse.smarthome.binding.mqtt/src/main/java/org/eclipse/smarthome/binding/mqtt/internal/discovery/NetworkDiscoveryService.java
```diff
@@ -48,21 +46,22 @@
...
+    private int scannedIPcount = 0;
 
     public NetworkDiscoveryService() {
         super(Collections.singleton(MqttBindingConstants.BRIDGE_TYPE_BROKER), 0);
     }
 
     @Override
     @Activate
-    protected void activate(Map<String, Object> config) {
+    protected void activate(@Nullable Map<String, @Nullable Object> config) {
```
maggu2810: Why is this map nullable? Is this forced by some super declaration?

bundles/io/org.eclipse.smarthome.io.transport.mqtt/src/main/java/org/eclipse/smarthome/io/transport/mqtt/MqttService.java
```diff
@@ -12,84 +12,52 @@
...
  */
-@Component(immediate = true, service = {
-        MqttService.class }, configurationPid = "org.eclipse.smarthome.mqtt", property = {
-                Constants.SERVICE_PID + "=org.eclipse.smarthome.mqtt" })
 @NonNullByDefault
-public class MqttService {
-    private final Logger logger = LoggerFactory.getLogger(MqttService.class);
-    private final Map<String, MqttBrokerConnection> brokerConnections = new ConcurrentHashMap<String, MqttBrokerConnection>();
-    private final List<MqttServiceObserver> brokersObservers = new CopyOnWriteArrayList<>();
+public interface MqttService {
```
maggu2810: Remove the public keyword from all methods as methods are already public by definition.

extensions/io/org.eclipse.smarthome.io.mqttembeddedbroker/src/main/java/org/eclipse/smarthome/io/mqttembeddedbroker/internal/EmbeddedBrokerService.java
davidgraeff: A static check would be helpful to enforce this guideline

extensions/io/org.eclipse.smarthome.io.mqttembeddedbroker/src/main/java/org/eclipse/smarthome/io/mqttembeddedbroker/internal/EmbeddedBrokerServiceImpl.java
davidgraeff: I agree

extensions/binding/org.eclipse.smarthome.binding.mqtt/src/main/java/org/eclipse/smarthome/binding/mqtt/internal/discovery/NetworkDiscoveryService.java
davidgraeff: I see this code a lot:
```java
protected final void activate(final Map<String, Object> config) {
    if (config != null) {
        properties = config;
    }
```

Apparently `config` can be null, but I haven't found the standard that specifies OSGIs `activated`.

extensions/io/org.eclipse.smarthome.io.mqttembeddedbroker/src/main/java/org/eclipse/smarthome/io/mqttembeddedbroker/internal/EmbeddedBrokerService.java
maggu2810: You are right. Feel free to create a PR against the static-code-analyzer.

@svilenvul Can you integrate this one: https://pmd.github.io/pmd-6.2.0/pmd_rules_java_codestyle.html#unnecessarymodifier

extensions/binding/org.eclipse.smarthome.binding.mqtt/src/main/java/org/eclipse/smarthome/binding/mqtt/internal/discovery/NetworkDiscoveryService.java
maggu2810: https://osgi.org/download/r4v42/r4.cmpn.pdf

> 112.5.8 Activate Method

> The activate method can take zero or
 more arguments. Each argument must 
be of one of the following types

AFAIK all of that arguments are non null.

> Map  – The component instance will be passed an unmodifiable Map containing the component properties.

davidgraeff: AbstractDiscoveryService declares it as `protected void activate(@Nullable Map<String, @Nullable Object> configProperties)`, so I need to leave it as it is for now.

davidgraeff:
addressed all comments

extensions/io/org.eclipse.smarthome.io.mqttembeddedbroker.test/src/test/java/org/eclipse/smarthome/io/mqttembeddedbroker/internal/MqttEmbeddedBrokerServiceTest.java
davidgraeff: We are also testing the interface here. I'd leave the name for now.

extensions/binding/org.eclipse.smarthome.binding.mqtt/src/main/java/org/eclipse/smarthome/binding/mqtt/internal/discovery/NetworkDiscoveryService.java
maggu2810: That's another story.
The activate method of the `AbstractDiscoveryService` is not called by the SCR but by the developers of the specific discovery implementation.
Sure, it could be non-null here, too but it is out of scope here (currently the homematic discovery service seems not be configurable and calls the activate of the abstract discovery service with an null argument).

bundles/io/org.eclipse.smarthome.io.transport.mqtt/src/main/java/org/eclipse/smarthome/io/transport/mqtt/MqttService.java
```diff
@@ -5,78 +5,51 @@
  * information regarding copyright ownership.
  *
  * This program and the accompanying materials are made available under the
- * terms of the Eclipse Public License 2.0 which is available at
+ * terms of the Eclipse License 2.0 which is available at
```
htreu: what happened here? Travis liked the version including `Public ` better.

davidgraeff: I replaced all "public " occurrences with an empty string 😓

htreu:
Travis still unhappy:

```
[ERROR] Failed to execute goal org.eclipse.tycho:tycho-compiler-plugin:1.1.0:compile (default-compile) on project org.eclipse.smarthome.binding.mqtt.test: Compilation failure: Compilation failure: 
[ERROR] /home/travis/build/eclipse/smarthome/extensions/binding/org.eclipse.smarthome.binding.mqtt.test/src/test/java/org/eclipse/smarthome/binding/mqtt/handler/MqttBrokerConnectionEx.java:[49] 
[ERROR] 	connectionCallbacks = spy(new ConnectionCallbacks(o));
[ERROR] 	^^^^^^^^^^^^^^^^^^^
[ERROR] connectionCallbacks cannot be resolved to a variable
[ERROR] /home/travis/build/eclipse/smarthome/extensions/binding/org.eclipse.smarthome.binding.mqtt.test/src/test/java/org/eclipse/smarthome/binding/mqtt/handler/MqttBrokerConnectionEx.java:[49] 
[ERROR] 	connectionCallbacks = spy(new ConnectionCallbacks(o));
[ERROR] 	                              ^^^^^^^^^^^^^^^^^^^
[ERROR] ConnectionCallbacks cannot be resolved to a type
[ERROR] 2 problems (2 errors)
```

davidgraeff:
Travis passes :)

maggu2810:
> Travis passes :)

Have you seen my comment: https://github.com/eclipse/smarthome/pull/5535/files#r200014223

extensions/binding/org.eclipse.smarthome.binding.mqtt/src/main/java/org/eclipse/smarthome/binding/mqtt/internal/discovery/NetworkDiscoveryService.java
davidgraeff: The problem is, that I'm using the same activate signature. I either hide/shadow the base implementation, which is a bad practise, or I need to match the exact same signature. What do you prefer?

maggu2810: Big sorry, I am missed that you extend the `AbstractDiscoveryService` here. :innocent:
All fine...

maggu2810:
@htreu no remarks from my side. Are you happy and would you like to push the button? :wink:

htreu:
@maggu2810 & @davidgraeff thanks a lot! Makes me happy :-)

htreu:
@maggu2810 I think we merged this by accident. Lines count is `+1,592 −860`. we should issue a CQ for this one, wdyt?

maggu2810:
Yes, we should. I will remove the commit from master (history rewriting ASAP before any other thing is committed). You agree?

htreu:
I discussed with @sjka and we think the CQ could also be issued afterwards. I think we all agree that the contribution is fine IP wise. Since I'm the least experienced committer I leave this to you and @sjka.

maggu2810:
@skja don't we run into trouble if the CQ is not accepted (sure, the change is similar to zero)

sjka:
Yes, we would run into trouble - as we would with any CQ with preliminary check-in approval which would get rejected afterwards. So you are right, a quick history rewrite would be safest.
Nevertheless, I indeed suggested to accept that (very low) risk of a later re-write and make this transparent towards legal in the CQ. However, I'm okay with a re-write now as well. Clearly "go" from my side.

htreu:
I'm also okay with the safe option and re-write now. I will issue the CQ.

davidgraeff:
I leave my branch open, until this is sorted.

maggu2810:
@davidgraeff Thank you. Can you open a new PR?

maggu2810:
... I can open a new PR for you, too (as you already spend a lot of time) if you want.

davidgraeff:
done

Issue #5322, opened 2018-03-28
Prerequisite version not always honored by FirmwareUpdateService


hsudbrock:
**Context**
`o.e.sh.core.thing.binding.firmwareFirmware`s have an optional `prerequisiteVersion`. The prerequisite version indicates that the firmware can only be installed on a device on which a firmware with version greater or equal to the prerequisite version is already installed.

Example: If the prerequisite version of a firmware with version '3.0.0' is '2.0.0', then this firmware must not be installed on devices with firmware version '1.x.x', but can be installed on devices with firmware version '2.x.x'.

**Issue**
The prerequisite version is checked in the `o.e.sh.core.thing.firmware.FirmwareUpdateService` when updating the firmware on a device using the method `updateFirmware(thingUID, firmwareUID, locale)` (throwing an `IllegalArgumentException` if the firmware on the device is below the prerequisite version of the to-be-installed firmware).

In contrast, the `FirmwareStatusInfo` returned by `FirmwareUpdateService#getFirmwareStatusInfo(thingUID)` always reports the latest firmware version as installable for the thing, ignoring the prerequisite version. 

In consequence, trying to install the latest updatable firmware returned by `getFirmwareStatusInfo` using `updateFirmware` will fail with an `IllegalArgumentException`.

The following unit test illustrates this issue:
```
package org.eclipse.smarthome.core.thing.firmware;

import static org.hamcrest.CoreMatchers.equalTo;
import static org.junit.Assert.assertThat;
import static org.mockito.ArgumentMatchers.*;
import static org.mockito.Mockito.*;

import java.util.Arrays;
import java.util.HashSet;
import java.util.Locale;

import org.eclipse.smarthome.core.thing.Thing;
import org.eclipse.smarthome.core.thing.ThingTypeUID;
import org.eclipse.smarthome.core.thing.ThingUID;
import org.eclipse.smarthome.core.thing.binding.firmware.Firmware;
import org.eclipse.smarthome.core.thing.binding.firmware.FirmwareUID;
import org.eclipse.smarthome.core.thing.binding.firmware.FirmwareUpdateHandler;
import org.eclipse.smarthome.test.java.JavaOSGiTest;
import org.junit.Test;

import com.google.common.collect.ImmutableMap;

public class FirmwareUpdateServicePrerequisiteVersionTest extends JavaOSGiTest {

    @Test
    public void testGetFirmwareStatusInfoWithPrerequisiteVersion() {
[...]

PR #5544, opened 2018-05-07
[5322] Honor prerequisite version for firmwares

hsudbrock:
This PR addresses https://github.com/eclipse/smarthome/issues/5322, by adding the check for the prerequisite firmware version into the method `Firmware#isSuitableFor(thing)`. A firmware is no longer deemed suitable for a thing, if the firmware has a prerequisite version, and the firmware version currently installed on the thing has a lower version than this prerequisite version.

With this change, the `FirmwareRegistryImpl` will check prerequisite versions when getting the list of available firmwares for a thing (as it calls `isSuitableFor`) - and, in consequence, the `FirmwareStatusInfo` will no longer return firmwares for a thing that are not suitable due to prerequisite version restrictions.

(As `FirmwareUpdateServiceImpl#validateFirmwareSuitability(...)` calls `Firmware#isSuitableFor(thing)`, the explicit check for the prerequisite version in that method is removed with this PR.)

Besides this fix, this PR also contains small refactorings in the `FirmwareImpl`, and a change in the `Firmware` interface: The method `Firmare#isPrerequisiteVersion(firmwareVersion)` is removed, because (a) I see no need for that method anymore as prerequisite version constraints are now checked in `Firmware.isSuitableFor(thing)`, and (b) I found that method somewhat confusing, as it is not immediately clear in which direction the check of `firmware.isPrerequisiteVersion(firmwareVersion)` goes (and I think that the JavaDoc even described it the wrong way round, at least the way I interpreted it). This removal is an API-breaking change.

### DIFF
--- a/bundles/core/org.eclipse.smarthome.core.thing.test/src/test/java/org/eclipse/smarthome/core/thing/firmware/FirmwareUpdateServicePrerequisiteVersionTest.java
+++ b/bundles/core/org.eclipse.smarthome.core.thing.test/src/test/java/org/eclipse/smarthome/core/thing/firmware/FirmwareUpdateServicePrerequisiteVersionTest.java
@@ -1,0 +1,121 @@
+/**
+ * Copyright (c) 2014,2018 Contributors to the Eclipse Foundation
+ *
+ * See the NOTICE file(s) distributed with this work for additional
+ * information regarding copyright ownership.
+ *
+ * This program and the accompanying materials are made available under the
+ * terms of the Eclipse Public License 2.0 which is available at
+ * http://www.eclipse.org/legal/epl-2.0
+ *
+ * SPDX-License-Identifier: EPL-2.0
+ */
+package org.eclipse.smarthome.core.thing.firmware;
+
+import static org.hamcrest.CoreMatchers.*;
+import static org.junit.Assert.assertThat;
+import static org.mockito.ArgumentMatchers.any;
+import static org.mockito.ArgumentMatchers.eq;
+import static org.mockito.Mockito.*;
+
+import java.util.Arrays;
+import java.util.HashSet;
+import java.util.Locale;
+
+import org.eclipse.smarthome.core.thing.Thing;
+import org.eclipse.smarthome.core.thing.ThingTypeUID;
+import org.eclipse.smarthome.core.thing.ThingUID;
+import org.eclipse.smarthome.core.thing.binding.firmware.Firmware;
+import org.eclipse.smarthome.core.thing.binding.firmware.FirmwareBuilder;
+import org.eclipse.smarthome.core.thing.binding.firmware.FirmwareUpdateHandler;
+import org.eclipse.smarthome.test.java.JavaOSGiTest;
+import org.junit.Before;
+import org.junit.Test;
+
+import com.google.common.collect.ImmutableMap;
+
+/**
+ * Tests that the {@link FirmwareUpdateService} honors the prerequisite version of {@link Firmware}s.
+ *
+ * @author Henning Sudbrock - initial contribution
+ */
+public class FirmwareUpdateServicePrerequisiteVersionTest extends JavaOSGiTest {
+
+    private ThingTypeUID thingTypeUID;
+
+    private Firmware firmwareV1;
+    private Firmware firmwareV2;
+    private Firmware firmwareV3;
+
+    private Thing thingWithFirmwareV1;
+    private Thing thingWithFirmwareV2;
+
+    @Before
+    public void setup() {
+        thingTypeUID = new ThingTypeUID("binding:thingtype");
+
+        firmwareV1 = FirmwareBuilder.create(thingTypeUID, "1.0.0").build();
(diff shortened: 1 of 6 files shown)


eclipsewebmaster:
Issue tracker reference:
https://bugs.eclipse.org/bugs/show_bug.cgi?id=5322

bundles/core/org.eclipse.smarthome.core.thing.test/src/test/java/org/eclipse/smarthome/core/thing/firmware/FirmwareUpdateServicePrerequisiteVersionTest.java
```diff
@@ -0,0 +1,121 @@
...
+    }
+
+    private FirmwareUpdateHandler createFirmwareUpdateHandler(Thing thing) {
+        FirmwareUpdateHandler firmwareUpdateHandler = mock(FirmwareUpdateHandler.class);
+        when(firmwareUpdateHandler.getThing()).thenReturn(thing);
+        when(firmwareUpdateHandler.isUpdateExecutable()).thenReturn(true);
+        return firmwareUpdateHandler;
+    }
+
+}
```
triller-telekom: Please add a line break after the last line

bundles/core/org.eclipse.smarthome.core.thing/src/main/java/org/eclipse/smarthome/core/thing/internal/firmware/FirmwareImpl.java
```diff
@@ -254,18 +239,26 @@ private boolean hasRequiredModel(Thing thing) {
...
         private final String[] parts;
 
-        private Version(@Nullable String versionString) {
-            if (versionString == null) {
-                this.parts = new String[] {};
-            } else {
-                this.parts = versionString.split("-|_|\\.");
-            }
+        private Version(String versionString) {
+            ParameterChecks.checkNotNull(versionString, "versionString");
```
triller-telekom: Since this is a private class with a private constructor, I think it is sufficient to annotate the parameter via `@NonNullByDefault` and omit the check since it can only be called by framework code anyway, right? See https://www.eclipse.org/smarthome/documentation/development/conventions.html#null-annotations

bundles/core/org.eclipse.smarthome.core.thing.test/src/test/java/org/eclipse/smarthome/core/thing/firmware/FirmwareUpdateServicePrerequisiteVersionTest.java
```diff
@@ -0,0 +1,121 @@
...
+import org.eclipse.smarthome.core.thing.ThingTypeUID;
+import org.eclipse.smarthome.core.thing.ThingUID;
+import org.eclipse.smarthome.core.thing.binding.firmware.Firmware;
+import org.eclipse.smarthome.core.thing.binding.firmware.FirmwareBuilder;
+import org.eclipse.smarthome.core.thing.binding.firmware.FirmwareUpdateHandler;
+import org.eclipse.smarthome.test.java.JavaOSGiTest;
+import org.junit.Before;
+import org.junit.Test;
+
+import com.google.common.collect.ImmutableMap;
```
triller-telekom: Is there no alternative for adding the dependency to the google library here? I.e. for your test I think it does not matter if it would be a "normal" `HashMap`, right?

bundles/core/org.eclipse.smarthome.core.thing.test/src/test/java/org/eclipse/smarthome/core/thing/firmware/FirmwareUpdateServicePrerequisiteVersionTest.java
hsudbrock: Thanks, will do (and should have looked at the static code analysis one more time before creating this PR...).

bundles/core/org.eclipse.smarthome.core.thing/src/main/java/org/eclipse/smarthome/core/thing/internal/firmware/FirmwareImpl.java
hsudbrock: This sounds sensible, I will remove the `ParameterChecks...`.

bundles/core/org.eclipse.smarthome.core.thing.test/src/test/java/org/eclipse/smarthome/core/thing/firmware/FirmwareUpdateServicePrerequisiteVersionTest.java
hsudbrock: There are alternatives, they are just not as nice as (to my knowledge) there is no builder for "normal" maps. But thanks for the hint, I will of course remove this guava dependency here :)

hsudbrock:
Thanks for reviewing, @triller-telekom! I addressed your comments in my latest commit.

triller-telekom:
Thanks, LGTM now!

PR #5560, opened 2018-05-09
Add Serial-over-IP (rfc2217) support in ESH Serial API

msteigenberger:
The actual instance of the SerialPort is chosen at runtime depending on the port's name.
This allows to have different implementations for different protocols. This can be e.g. RXTX (default local) or RFC2217.
Implementations are discovered at runtime via OSGi DS.

A client application now only needs to set the correct port name.
e.g.
-  /dev/ttyUSB -> RxTxPort
- rfc2217://x.x.x.x:3001 -> TelnetSerialPort

Signed-off-by: Matthias Steigenberger <matthias.steigenberger@gmail.com>


msteigenberger:
@maggu2810: Please have a look it is an addition to your PR: https://github.com/eclipse/smarthome/pull/5313

I want to have an easy way in the framework itself to allow bindings to switch between serial port implementations.
Actually I first thought to have a own bundle which holds the registry and all other implementations (like javacomm) can also use it and provide a "SerialPortCreator". I left it because I first wanted to know your opinion in general for this.
Thanks!

kaikreuzer:
> I want to have an easy way in the framework itself to allow bindings to switch between serial port implementations.

Could you elaborate on why you want to let the bindings to do the selection?

In essence, the current serial API expects a single implementation to be present, while you are now asking to have the option to have multiple implementations being present at runtime.

Imho the simplest (and non-intrusive) enhancement would be to leave it to the implementations to claim certain ports for them - e.g. any port that starts with `rfc2217://` could be claimed by the TelnetSerialPort implementation. Would such a naming convention approach work?

@maggu2810 Any comments or ideas from your side?

msteigenberger:
> Imho the simplest (and non-intrusive) enhancement would be to leave it to the implementations to claim certain ports for them - e.g. any port that starts with rfc2217:// could be claimed by the TelnetSerialPort implementation. Would such a naming convention approach work?

But then every implementation would need its own logic to decide which serial port to use AND - which is even worse - the implementation would need a dependency to the specific port (like TelnetSerialport which comes from nrjavaserial).
The whole intention was to exactly not depend on any implementation but rather let the framework decide **depending on the users input** which serial port implementation to use. If the user wants to communicate via rfc2217 he/she only has to state the rfc2217:// URL and the binding implementation doesn't care about it.

maggu2810:
Hm, for me the distributor should decide which implementations should be supported and not the user itself.
If I think about ESH as a framework also for commercial products the company choose the serial port implementation (stability, licenses, ...) and the consumer will use that implementation without needing to knowing the implementation itself.

msteigenberger:
> Imho it makes only sense to consider those "rfc2217" ports as serial ports, if they actually behave exactly the same, so from a binding perspective, it actually should not matter whether the port is local or remote - any existing binding that requires a serial connection should directly also work with remote ports; is this realistic? If so, the only way to distinguish between local and remote ports (and thus do proper handling in the serial implementation) is a naming convention wrt the port names.

Exactly. Intention is that it makes no difference from the binding perspective whether it's a remote (rfc2217) or local serial port. Therefore the user can decide that (and is the only one that can do that because it's depending on his setup). Of course if the binding (or protocol) makes some timing critical things, it could happen that it is not working with rfc2217.

maggu2810:
Please have a look at: https://github.com/eclipse/smarthome/compare/master...maggu2810:serial-port-provider-id?expand=1

If we assume that `://` will not be part of a currently existing port identifier's name, this one could be a working solution.

maggu2810:
If an user do not want to take care of the implementation, he just does not need to specify `://` and the part in front of.

If the user want to force a special implementation, he can specify the `provider ID`.

So, e.g. if we don't want to differ between the specific implementation, but differ between types, we just need to change `provider ID` to `provider type` and the first one that matches will be used.

gersilex:
> you simply do not know, which devices on your network might be available with such a feature (unless we could identify them through mDNS or similar, which I doubt).

@bodiroba told me about https://github.com/Tknika/serial2tcp-gateway where he is working on a daemon that watches the USB ports for known devices, creates TCP listen ports that relay to the physical serial port of that devices and advertises them on mDNS. Worked well in my short test. That can make the devices discoverable.

> We could also think about providing the type of the port (local, telnet com port), so the user can choose the type but do not need to care about the implementation.

I agree on that, too. We could start by expecting only one implementation per provider (remote-port only uses rfc2217, local port only uses nrserial, ...).

Thinking about the socket naming. I think that keeping the unix-style *everything is a file system* approach (`/dev/ttyUSB0`) and for remote ports the URL-style *absolute location* approach (see [rfc1738 Section 3.1](https://tools.ietf.org/html/rfc1738#section-3.1) (`//serialTelnetServer:5555`).  

This would also allow the user to directly specify an implementation (a scheme) if he (or the binding) really insists on it by prepending the scheme like in said rfc1738 (`rfc2217://serialTelnetServer:5556` or even `rxtx:/dev/ttyUSB1`).

All of the URIs can be matched easily:
- `^\/\w+` matches `/dev/ttyUSB0`: Use classic serial behaviour, local serial port.
- `^\/{2}(\w+):(\d+)` matches `//serialTelnetServer:5555`: Use remote serial port with best matching provider/scheme
- `^(\w+):\/{2}(\w+):(\d+)` matches `rfc2217://serialTelnetServer:5556`: Use remote provider/scheme from URI.
- `^(\w+):([^:]+)$` matches `rxtx:/dev/ttyUSB1`: Use local provider/scheme from URI.

[Show these regular expressions on regex101.com](https://regex101.com/r/p4dkS9/1)

What do you think?

kaikreuzer:
> IMHO we just have to consider following cases:
> The solution provides the serial port implementation (rxtx, javacomm)
> The user provides the serial port type (local, telnet)

Makes sense to me like that.

Imho @gersilex suggestion looks pretty good - it would be fully backward compatible and "//" would tell us that we are talking about a remote port. If we agree that (for now) we leave it to the solution to choose the implementation, I would suggest to not support the optional scheme for now at all (as it only complicates the implementation and it can be easily added later without breaking anything).

> creates TCP listen ports that relay to the physical serial port of that devices and advertises them on mDNS. Worked well in my short test. That can make the devices discoverable.

Yes, but I am not sure how much that is gonna be used; we are talking here also about embedded devices that offer such a port, but where you cannot add any additional services yourself. So in any case, we will have to set `limitToOptions=false` for the serial port parameter, so that the user can freely enter a string.

@maggu2810 & @msteigenberger Would that solution be ok for you? If so, @msteigenberger, could you rework the PR accordingly?

maggu2810:
> I would suggest to not support the optional scheme for now at all (as it only complicates the implementation and it can be easily added later without breaking anything).

We already have the regular expression and we just need to expose the name in the specific implementation. I don't think it will complicate the handling. Or do I miss some point?

If it is just a few lines I would prefer to add it as it will help to test serial provider implementations a lot.
E.g. one receiver on serial interface 1, another receiver on serial interface 2 and different implementations chosen for the bridges... Sure, this is nothing the normal user will do and the framework does not need to be used to test serial provider implementations -- but if it can be added without blowing the code base :wink:

kaikreuzer:
>  I don't think it will complicate the handling. Or do I miss some point?

Well, you obviously need to be able to identify different implementations by ID (which influences the API), implement the selection logic, document what it is about, wonder whether you need end user documentation and what options you want to allow for solutions, add localisation and Paper UI support in case that solutions should be allowed to leave the selection of an implementation to the user, etc.

That's why I simply think that as long as we do not have any requirement for it, better just completely leave it out for now...

msteigenberger:
I've added a new commit.
However the strategy is now following:
- Every implementation provider has now a name/id which is used as the scheme ("rxtx", "rfc2217", "javacomm")
- The user **can** use the scheme to use the concrete implementation
- If no implementation for provided scheme is found, an appropriate one with same type (local or remote) is used
- The user can use just "//" to indicate remote connection
- In latter case the first found remote implementation is used otherwise a local implementation is used

kaikreuzer:
Travis fails with:
```
[ERROR] Cannot resolve dependencies of MavenProject: org.eclipse.smarthome.io:org.eclipse.smarthome.io.transport.serial.rxtx.rfc2217:0.10.0-SNAPSHOT @ /home/travis/build/eclipse/smarthome/bundles/io/org.eclipse.smarthome.io.transport.serial.rxtx.rfc2217/pom.xml: See log for details -> [Help 1]
```

msteigenberger:
> Travis fails with:

@kaikreuzer: yes, thats because the current used version of nrjavaserial (3.12.0-OH) has no rfc2217 packages included.
We need to integrate newer version of nrjavaserial to get that build working.

kaikreuzer:
> We need to integrate newer version of nrjavaserial to get that build working.

You formulated this nicely fuzzy. Digging into it, I realized that https://github.com/NeuronRobotics/nrjavaserial/pull/109 still isn't merged, which is a prerequisite for us - and apart from this, it would be really time for a new official release of nrjavaserial...

If no release is in sight, we would have to do a 3.13.0.OH build ourselves (but we would need somebody, who is capable of doing so...).

bundles/io/org.eclipse.smarthome.io.transport.serial.ctrl/src/main/java/org/eclipse/smarthome/io/transport/serial/internal/SerialPortRegistry.java
```diff
@@ -0,0 +1,85 @@
...
+import org.osgi.service.component.annotations.Reference;
+import org.osgi.service.component.annotations.ReferenceCardinality;
+import org.osgi.service.component.annotations.ReferencePolicy;
+import org.osgi.service.component.annotations.ReferencePolicyOption;
+import org.slf4j.Logger;
+import org.slf4j.LoggerFactory;
+
+/**
+ *
+ * @author MatthiasS
```
maggu2810: Should be something like: forname surename - Initial contribution (and API)

bundles/io/org.eclipse.smarthome.io.transport.serial.ctrl/src/main/java/org/eclipse/smarthome/io/transport/serial/internal/SerialPortRegistry.java
```diff
@@ -0,0 +1,85 @@
...
+ *
+ */
+@Component(immediate = true, service = SerialPortRegistry.class)
```
maggu2810: Why do we need an immediate activiation?

bundles/io/org.eclipse.smarthome.io.transport.serial.ctrl/src/main/java/org/eclipse/smarthome/io/transport/serial/internal/SerialPortRegistry.java
```diff
@@ -0,0 +1,85 @@
...
+        return first
+                .orElse(portCreators.stream()
+                        .filter(provider -> provider.getAcceptedProtocols()
+                                .filter(prot -> prot.getPathType().equals(pathType)).count() > 0)
+                        .findFirst().orElse(null));
+
+    }
+
+    public Collection<SerialPortProvider> getPortCreators() {
+        return Collections.unmodifiableCollection(portCreators);
```
maggu2810: The unmodifiable collection is a view and so every change to the `portCreators` will affect the user of the view.
This will potential lead to a CME if register or unregister is called while someone uses the view.

You should perhaps use something like `Collection.unmodifiableCollection(new HashSet<>(portCreators))`.

bundles/io/org.eclipse.smarthome.io.transport.serial.ctrl/src/main/java/org/eclipse/smarthome/io/transport/serial/internal/SerialPortRegistry.java
```diff
@@ -0,0 +1,85 @@
...
+    }
+
+    /**
+     * Registers a {@link SerialPortProvider}.
+     *
+     * @param creator
+     */
+    @Reference(cardinality = ReferenceCardinality.MULTIPLE, policy = ReferencePolicy.DYNAMIC, policyOption = ReferencePolicyOption.GREEDY)
+    protected void registerSerialPortCreator(SerialPortProvider creator) {
+        this.portCreators.add(creator);
```
maggu2810: As you are using a dynamic reference the member portCreators will be accessed by different threads.
You could e.g. use a concurrent collection or other synchronization statements.

bundles/io/org.eclipse.smarthome.io.transport.serial.javacomm/src/main/java/org/eclipse/smarthome/io/transport/serial/internal/JavaCommPortCreator.java
```diff
@@ -22,18 +23,47 @@
...
- * Specific serial port manager implementation.
  *
- * @author Markus Rathgeb - Initial contribution
+ * @author MatthiasS
+ *
  */
 @NonNullByDefault
-@Component
-public class SerialPortManagerImpl implements SerialPortManager {
+@Component(service = SerialPortProvider.class, immediate = true)
```
maggu2810: immediate?

bundles/io/org.eclipse.smarthome.io.transport.serial.rxtx.rfc2217/src/main/java/org/eclipse/smarthome/io/transport/serial/rxtx/rfc2217/internal/RFC2217PortCreator.java
```diff
@@ -0,0 +1,57 @@
...
+import org.eclipse.smarthome.io.transport.serial.ProtocolType.PathType;
+import org.eclipse.smarthome.io.transport.serial.SerialPortIdentifier;
+import org.eclipse.smarthome.io.transport.serial.SerialPortProvider;
+import org.osgi.service.component.annotations.Component;
+
+import gnu.io.rfc2217.TelnetSerialPort;
+
+/**
+ *
+ * @author MatthiasS
```
maggu2810: see above

bundles/io/org.eclipse.smarthome.io.transport.serial.rxtx/src/main/java/org/eclipse/smarthome/io/transport/serial/internal/RxTxPortCreator.java
```diff
@@ -20,20 +21,52 @@
...
- * Specific serial port manager implementation.
  *
- * @author Markus Rathgeb - Initial contribution
+ * @author MatthiasS
+ *
  */
 @NonNullByDefault
-@Component
-public class SerialPortManagerImpl implements SerialPortManager {
+@Component(service = SerialPortProvider.class, immediate = true)
```
maggu2810: immediate?

bundles/io/org.eclipse.smarthome.io.transport.serial.rxtx/src/main/java/org/eclipse/smarthome/io/transport/serial/internal/RxTxPortCreator.java
```diff
@@ -20,20 +21,52 @@
...
+import org.eclipse.smarthome.io.transport.serial.SerialPortProvider;
 import org.osgi.service.component.annotations.Component;
 
 import gnu.io.CommPortIdentifier;
 
 /**
- * Specific serial port manager implementation.
  *
- * @author Markus Rathgeb - Initial contribution
+ * @author MatthiasS
```
maggu2810: see above

bundles/io/org.eclipse.smarthome.io.transport.serial.rxtx/src/main/java/org/eclipse/smarthome/io/transport/serial/internal/SerialPortUtil.java
```diff
@@ -0,0 +1,59 @@
...
+package org.eclipse.smarthome.io.transport.serial.internal;
+
+import java.util.HashSet;
+import java.util.Set;
+import java.util.stream.Collectors;
+import java.util.stream.Stream;
+
+/**
+ *
+ * @author MatthiasS
```
maggu2810: see above

bundles/io/org.eclipse.smarthome.io.transport.serial/src/main/java/org/eclipse/smarthome/io/transport/serial/SerialPortProvider.java
```diff
@@ -0,0 +1,56 @@
...
+
+import java.net.URI;
+import java.util.stream.Stream;
+
+import org.eclipse.jdt.annotation.NonNullByDefault;
+import org.eclipse.jdt.annotation.Nullable;
+
+/**
+ *
+ * @author MatthiasS
```
maggu2810: see above

bundles/io/org.eclipse.smarthome.io.transport.serial/src/main/java/org/eclipse/smarthome/io/transport/serial/SerialPortProvider.java
```diff
@@ -0,0 +1,56 @@
...
+ *
+ * @param <T>
```
maggu2810: the interface does not contain a generic parameter

maggu2810 (commented):
added some comments, thanks for your effort

kaikreuzer:
Good news, we should soon have a 3.14.0 release of nrjavaserial, see https://github.com/NeuronRobotics/nrjavaserial/pull/109!

kaikreuzer:
I was just checking how we can upgrade the tp to nrjavaserial 3.14.0. We imho cannot add it to our ESH 3rd party p2 site as it is an LGPL "works-with" dependency. We cannot consume it directly from Maven Central as we need it on a p2 site. So I guess, I will have to best deploy a p2 site to the openhab bintray, just where we have the forked version right now as well.

@maggu2810, @sjka & @htreu Any better/other idea?

htreu:
I guess its due to the fact that source code was not IP checked for the "works-with" dependency? Then hosting this anywhere is the only solution. The openHAB repo guarantees a great level of stability in terms of unexpected updates, so okay for me.

maggu2810:
> @maggu2810, @sjka & @htreu Any better/other idea?

Too late :wink:

bundles/io/org.eclipse.smarthome.io.transport.serial.ctrl/.project
```diff
@@ -0,0 +1,33 @@
+<?xml version="1.0" encoding="UTF-8"?>
+<projectDescription>
+	<name>org.eclipse.smarthome.io.transport.serial.ctrl</name>
```
kaikreuzer: I do not think we need this separate project/bundle. After all, the classes are in the `org.eclipse.smarthome.io.transport.serial.internal` package, so they should best be part of the `org.eclipse.smarthome.io.transport.serial` bundle. It is good enough if API and Impl are separated by package, we do not require separate bundles (and we do not do this at other places in ESH).

kaikreuzer (changes requested):
@msteigenberger The ESH TP should now include the new nrjavaserial, so hopefully you can make this PR compile now. Let me know, once you have updated it!

msteigenberger:
@kaikreuzer: The build is now working with your changes to the TP. 
I resolved the review comments.

bundles/io/org.eclipse.smarthome.io.transport.serial.javacomm/src/main/java/org/eclipse/smarthome/io/transport/serial/internal/JavaCommPortCreator.java
```diff
@@ -22,18 +23,47 @@
...
 import org.eclipse.smarthome.io.transport.serial.SerialPortIdentifier;
-import org.eclipse.smarthome.io.transport.serial.SerialPortManager;
+import org.eclipse.smarthome.io.transport.serial.SerialPortProvider;
 import org.osgi.service.component.annotations.Component;
 
 /**
- * Specific serial port manager implementation.
  *
- * @author Markus Rathgeb - Initial contribution
+ * @author MatthiasS
```
kaikreuzer: Full name here as well please.
And as you seem to have copied/renamed this class, you might actually want to keep Markus as the original author.

bundles/io/org.eclipse.smarthome.io.transport.serial.rxtx/src/main/java/org/eclipse/smarthome/io/transport/serial/internal/SerialPortUtil.java
```diff
@@ -0,0 +1,59 @@
...
+     * Registers the given port as system property {@value #GNU_IO_RXTX_SERIAL_PORTS}. The method is capable of
+     * extending the system property, if any other ports are already registered.
+     *
+     * @param port the port to be registered
+     */
+    public synchronized static void appendSerialPortProperty(String port) {
+        String serialPortsProperty = System.getProperty(GNU_IO_RXTX_SERIAL_PORTS);
+        String newValue = initSerialPort(port, serialPortsProperty);
+        if (newValue != null) {
+
```
kaikreuzer: remove empty line

bundles/io/org.eclipse.smarthome.io.transport.serial.rxtx/src/main/java/org/eclipse/smarthome/io/transport/serial/internal/RxTxPortCreator.java
```diff
@@ -20,20 +21,52 @@
...
- * Specific serial port manager implementation.
  *
- * @author Markus Rathgeb - Initial contribution
+ * @author Matthias Steigenberger - Initial Contribution
+ *
  */
 @NonNullByDefault
-@Component
-public class SerialPortManagerImpl implements SerialPortManager {
+@Component(service = SerialPortProvider.class)
```
kaikreuzer: Imho no need to specify the service as interfaces are considered by default

bundles/io/org.eclipse.smarthome.io.transport.serial/src/main/java/org/eclipse/smarthome/io/transport/serial/SerialPortProvider.java
```diff
@@ -0,0 +1,55 @@
...
+package org.eclipse.smarthome.io.transport.serial;
+
+import java.net.URI;
+import java.util.stream.Stream;
+
+import org.eclipse.jdt.annotation.NonNullByDefault;
+import org.eclipse.jdt.annotation.Nullable;
+
+/**
+ *
```
kaikreuzer: please add some JavaDoc

bundles/io/org.eclipse.smarthome.io.transport.serial/OSGI-INF/org.eclipse.smarthome.io.transport.serial.internal.SerialPortManagerImpl.xml
```diff
@@ -0,0 +1,8 @@
+<?xml version="1.0" encoding="UTF-8"?>
```
kaikreuzer: do not commit these xmls, but add a .gitignore file in the OSGI-INF folder

bundles/io/org.eclipse.smarthome.io.transport.serial/src/main/java/org/eclipse/smarthome/io/transport/serial/ProtocolType.java
```diff
@@ -0,0 +1,55 @@
...
+ *
+ * SPDX-License-Identifier: EPL-2.0
+ */
+package org.eclipse.smarthome.io.transport.serial;
+
+import java.net.URI;
+
+/**
+ *
+ * @author MatthiasS
```
kaikreuzer: full name

bundles/io/org.eclipse.smarthome.io.transport.serial/src/main/java/org/eclipse/smarthome/io/transport/serial/ProtocolType.java
```diff
@@ -0,0 +1,55 @@
...
+ * terms of the Eclipse Public License 2.0 which is available at
+ * http://www.eclipse.org/legal/epl-2.0
+ *
+ * SPDX-License-Identifier: EPL-2.0
+ */
+package org.eclipse.smarthome.io.transport.serial;
+
+import java.net.URI;
+
+/**
```
kaikreuzer: Add some JavaDoc

bundles/io/org.eclipse.smarthome.io.transport.serial/src/main/java/org/eclipse/smarthome/io/transport/serial/ProtocolType.java
```diff
@@ -0,0 +1,55 @@
...
+public class ProtocolType {
+    public enum PathType {
+        NET,
+        LOCAL;
+        public static PathType fromURI(URI uri) {
+            return uri.getSchemeSpecificPart().startsWith("//") ? NET : LOCAL;
+        }
+    }
+
+    PathType pathType;
```
kaikreuzer: do these have to be package private?

bundles/io/pom.xml
```diff
@@ -44,6 +44,7 @@
     <module>org.eclipse.smarthome.io.transport.serial</module>
     <module>org.eclipse.smarthome.io.transport.serial.javacomm</module>
     <module>org.eclipse.smarthome.io.transport.serial.rxtx</module>
+    <module>org.eclipse.smarthome.io.transport.serial.rxtx.rfc2217</module>
```
kaikreuzer: It needs to be added to the [p2 feature](https://github.com/eclipse/smarthome/blob/master/features/org.eclipse.smarthome.feature.runtime.core/feature.xml) as well (and maybe to Karaf, but @maggu2810 is better to comment on that).

bundles/io/org.eclipse.smarthome.io.transport.serial/src/main/java/org/eclipse/smarthome/io/transport/serial/internal/SerialPortRegistry.java
```diff
@@ -0,0 +1,95 @@
...
+import org.eclipse.smarthome.io.transport.serial.SerialPortProvider;
+import org.osgi.service.component.annotations.Component;
+import org.osgi.service.component.annotations.Reference;
+import org.osgi.service.component.annotations.ReferenceCardinality;
+import org.osgi.service.component.annotations.ReferencePolicy;
+import org.osgi.service.component.annotations.ReferencePolicyOption;
+import org.slf4j.Logger;
+import org.slf4j.LoggerFactory;
+
+/**
```
kaikreuzer: add JavaDoc

bundles/io/org.eclipse.smarthome.io.transport.serial/src/main/java/org/eclipse/smarthome/io/transport/serial/internal/SerialPortRegistry.java
```diff
@@ -0,0 +1,95 @@
...
+ *
+ * @author Matthias Steigenberger - Initial Contribution
+ *
+ */
+@Component(service = SerialPortRegistry.class)
+public class SerialPortRegistry {
+
+    private final static Logger logger = LoggerFactory.getLogger(SerialPortRegistry.class);
```
kaikreuzer: remove "static"

kaikreuzer (changes requested):
Thanks, great that it now compiles!
Just added some further small review comments - once these are addressed, I'd leave the final merge to @maggu2810.

msteigenberger:
Thanks for the comments. I resolved them.

bundles/io/org.eclipse.smarthome.io.transport.serial/src/main/java/org/eclipse/smarthome/io/transport/serial/internal/SerialPortManagerImpl.java
```diff
@@ -59,6 +63,7 @@ protected void unsetSerialportRegistry(SerialPortRegistry registry) {
         URI portUri = URI.create(name);
         SerialPortProvider portCreator = registry.getPortProviderForPortName(portUri);
         if (portCreator == null) {
+            logger.warn("No SerialPortProvider found for: " + name);
```
kaikreuzer: Use parameterized logging, no string concatenation!

kaikreuzer (approved):

bundles/io/org.eclipse.smarthome.io.transport.serial.rxtx.rfc2217/src/main/java/org/eclipse/smarthome/io/transport/serial/rxtx/rfc2217/internal/SerialPortIdentifierImpl.java
```diff
@@ -0,0 +1,64 @@
...
+import org.eclipse.smarthome.io.transport.serial.SerialPort;
+import org.eclipse.smarthome.io.transport.serial.SerialPortIdentifier;
+import org.eclipse.smarthome.io.transport.serial.rxtx.RxTxSerialPort;
+
+import gnu.io.rfc2217.TelnetSerialPort;
+
+/**
+ * Specific serial port identifier implementation.
+ *
+ * @author Markus Rathgeb - Initial contribution
```
maggu2810: I was not aware of working on a RFC2217 file ;)

bundles/io/org.eclipse.smarthome.io.transport.serial.rxtx/src/main/java/org/eclipse/smarthome/io/transport/serial/internal/SerialPortUtil.java
```diff
@@ -0,0 +1,58 @@
...
+        String serialPortsProperty = System.getProperty(GNU_IO_RXTX_SERIAL_PORTS);
+        String newValue = initSerialPort(port, serialPortsProperty);
+        if (newValue != null) {
+            System.setProperty(GNU_IO_RXTX_SERIAL_PORTS, newValue);
+        }
+    }
+
+    static String initSerialPort(String port, String serialPortsProperty) {
+
+        String pathSeparator = System.getProperty("path.separator", ":");
```
maggu2810: Why not using `File.pathSeparator` here?

bundles/io/org.eclipse.smarthome.io.transport.serial.rxtx/src/main/java/org/eclipse/smarthome/io/transport/serial/internal/SerialPortUtil.java
```diff
@@ -0,0 +1,58 @@
...
+import java.util.Set;
+import java.util.stream.Collectors;
+import java.util.stream.Stream;
+
+/**
+ *
+ * @author Matthias Steigenberger - Initial Contribution
+ *
+ */
+public class SerialPortUtil {
```
maggu2810: New classes should contain the `@NonNullByDefault` and `@Nullable` annotations. Please add it here and in all new classes / interfaces / ...

bundles/io/org.eclipse.smarthome.io.transport.serial/META-INF/MANIFEST.MF
```diff
@@ -17,3 +17,5 @@ Import-Package:
  org.osgi.service.component,
  org.slf4j
 Automatic-Module-Name: org.eclipse.smarthome.io.transport.serial
+Service-Component: OSGI-INF/*.xml
+Bundle-ActivationPolicy: lazy
```
maggu2810: Remove the lazy activation policy.

bundles/io/org.eclipse.smarthome.io.transport.serial/src/main/java/org/eclipse/smarthome/io/transport/serial/ProtocolType.java
```diff
@@ -0,0 +1,61 @@
...
+ *
+ */
+public class ProtocolType {
+
+    /**
+     * Remote (NET) or Local path.
+     *
+     */
+    public enum PathType {
+    NET,
```
maggu2810: Are that lines correctly intended?

bundles/io/org.eclipse.smarthome.io.transport.serial/src/main/java/org/eclipse/smarthome/io/transport/serial/ProtocolType.java
```diff
@@ -0,0 +1,61 @@
...
+
+import java.net.URI;
+
+/**
+ * Holds the {@link PathType}, which specifies whether its a local or remote path and the scheme.
+ *
+ * @author Matthias Steigenberger - Initial Contribution
+ *
+ */
+public class ProtocolType {
```
maggu2810: nullness annotations

bundles/io/org.eclipse.smarthome.io.transport.serial/src/main/java/org/eclipse/smarthome/io/transport/serial/internal/SerialPortRegistry.java
```diff
@@ -0,0 +1,99 @@
...
+                            .findFirst()
+                    : Optional.empty();
+            // get a PortProvider which accepts exactly the port with its scheme. If there is none, just try a port with
+            // same type (local, net)
+            if (!first.isPresent() && scheme != null) {
+                return null;
+            }
+
+            return first.orElseGet(() -> {
+                return portCreators.stream()
```
maggu2810: You can omit the `{ return'` and the `}` around your expression as it is only one.

bundles/io/org.eclipse.smarthome.io.transport.serial/src/main/java/org/eclipse/smarthome/io/transport/serial/internal/SerialPortRegistry.java
```diff
@@ -0,0 +1,99 @@
...
+import org.osgi.service.component.annotations.ReferencePolicyOption;
+
+/**
+ * Registers all {@link SerialPortProvider}s which can be accessed here.
+ *
+ * @author Matthias Steigenberger - Initial Contribution
+ *
+ */
+@Component(service = SerialPortRegistry.class)
+public class SerialPortRegistry {
```
maggu2810: nullness annotations

bundles/io/pom.xml
maggu2810: @msteigenberger Do you need some support for the Karaf feature changes?

msteigenberger: @maggu2810: yes, would be good if you could give me some pointers

bundles/io/org.eclipse.smarthome.io.transport.serial/src/main/java/org/eclipse/smarthome/io/transport/serial/ProtocolType.java
msteigenberger: imho not, but this is done by eclipse formatter automatically which I use for openhab development. Shouldn't it be used here in esh? Is it a bug in the formatter?

bundles/io/pom.xml
maggu2810: If you want an "offline" communication send me an mail or something similar.
If you want to discuss here:
What do you need to know? What does not work? Have you already tried to modify the existing Karaf feature repos (features/karaf/...)?

bundles/io/org.eclipse.smarthome.io.transport.serial.rxtx.rfc2217/src/main/java/org/eclipse/smarthome/io/transport/serial/rxtx/rfc2217/internal/SerialPortIdentifierImpl.java
```diff
@@ -0,0 +1,64 @@
...
+import org.eclipse.smarthome.io.transport.serial.SerialPort;
+import org.eclipse.smarthome.io.transport.serial.SerialPortIdentifier;
+import org.eclipse.smarthome.io.transport.serial.rxtx.RxTxSerialPort;
+
+import gnu.io.rfc2217.TelnetSerialPort;
+
+/**
+ * Specific serial port identifier implementation for RFC2217.
+ *
+ * @author atthias Steigenberger - Initial contribution
```
kaikreuzer: "M"atthias

maggu2810:
@msteigenberger 
WRT https://github.com/eclipse/smarthome/pull/5885#issuecomment-404875046 (and follow) can you add the missing methods?
Have you seen my comment about the Karaf features?

msteigenberger:
@maggu2810:  changes were already done in #5907. RFC2217 uses the RxTxSerialPort (as this is just using the gnu.io.SerialPort interface).

Regarding the Karaf changes: I updated it to my best knowledge (which is not very accurate on this). 
Basically what I did is to add the rfc2217 feature to karaf with the appropriate requirement (same as for the rxtx feature). I also updated the nrjavaserial maven dependency in the esh-tp.
Furthermore I updated the Provide-Capabilites and Require-Capabilities Headers in the Manifest to adapt to the changes.
Can you please have a look?

bundles/io/org.eclipse.smarthome.io.transport.serial/src/main/java/org/eclipse/smarthome/io/transport/serial/ProtocolType.java
maggu2810: I consider this as a bug in the formatter. I have seen it on other projects, too.

bundles/io/org.eclipse.smarthome.io.transport.serial/src/main/java/org/eclipse/smarthome/io/transport/serial/ProtocolType.java
```diff
@@ -0,0 +1,64 @@
...
+import org.eclipse.jdt.annotation.NonNullByDefault;
+
+/**
+ * Holds the {@link PathType}, which specifies whether its a local or remote path and the scheme.
+ *
+ * @author Matthias Steigenberger - Initial Contribution
+ *
+ */
+@NonNullByDefault
+public class ProtocolType {
```
maggu2810: The setters are not used at the moment (AFAIK). I don't know about a reason why changing a once constructed `ProtocolType` instance (modify pathType or scheme). Can we remove the setters and make the members final?
It would make the instances also thread safe (immutable objects).

features/karaf/esh-core/src/main/feature/feature.xml
```diff
@@ -362,6 +363,17 @@
...
 
+  <feature name="esh-io-transport-serial-rfc2217" version="${project.version}">
+    <feature>esh-base</feature>
+
+    <requirement>esh.tp;filter:="(&amp;(feature=serial)(impl=rfc2217))"</requirement>
+    <feature dependency="true">esh.tp-serial-rxtx</feature>
+
+    <bundle>mvn:org.eclipse.smarthome.io/org.eclipse.smarthome.io.transport.serial/${project.version}</bundle>
+    <bundle>mvn:org.eclipse.smarthome.io/org.eclipse.smarthome.io.transport.serial.rxtx/${project.version}</bundle>
+	<bundle>mvn:org.eclipse.smarthome.io/org.eclipse.smarthome.io.transport.serial.rxtx.rfc2217/${project.version}</bundle>
```
maggu2810: wrong indentation

features/karaf/esh-tp/src/main/feature/feature.xml
```diff
@@ -150,9 +150,9 @@
 
   <feature name="esh.tp-serial-rxtx" version="${project.version}">
     <capability>esh.tp;feature=serial;impl=rxtx</capability>
-    <bundle>mvn:org.openhab/nrjavaserial/3.12.0.OH</bundle>
+    <bundle>mvn:com.neuronrobotics/nrjavaserial/3.14.0</bundle>
   </feature>
-
+
```
maggu2810: trailing whitespace

maggu2810 (commented):
some minor comments

maggu2810 (approved):
I am not fully sure about the feature dependencies of the RFC2217 support. But this could be improved later, too.
No reason to postpone this changes.

Thank you for your work.

codeaholics:
A couple of questions on this. Is it released? If so, what OH version?

I want to change an existing binding (the ZWave binding) to use this. I’m a very experienced Java developer and Eclipse user, but I’ve never really done PDE and the tooling confuses the hell out of me. Any pointers (from an OSGI/plugins perspective) of how to begin making these changes?

kaikreuzer:
Hi @codeaholics, yes, this change is available in openHAB - not yet in an official release, but in the nightly snapshots.
Wrt tooling, we have a one-click IDE installer for openHAB, see https://www.openhab.org/docs/developer/development/ide.html. In step (4), you will find an entry for the Z-Wave binding, which you need to select as well and you will find the code fully set up in your workspace and ready to be launched from there.

kaikreuzer:
Btw, before starting with any work on a PR, you might want to enter an issue in https://github.com/openhab/org.openhab.binding.zwave/issues and mention your plans, so that @cdjackson is aware of it and has the chance to comment upfront.

codeaholics:
Thanks. I’m already talking to a Chris. Right now, the ZWave binding has a text entry field for the port and it uses that to manually construct a serial port object. This needs to change to use this plugin as a serial port provider instead. I think that’s the only change, as I understand this is API compatible. But if not, it’s not exactly a complicated API to change. 

Can you point me towards some existing plugins/bindings that are correctly using this provider so that I can learn from them, please?

dastrix80:
Hi All, would this fix the issues I see when the console displays I/O errors because the virtual USB (mapped using ser2net & socat) has been removed or restarted? ie would the zwave binding be able to reconnect if that mapping was removed?

Right now, as soon as the USB mapping is removed a restart of the Zwave binding is required.

Sorry thats about as best I can explain the issue.

More info here: https://community.openhab.org/t/share-z-wave-dongle-over-ip-usb-over-ip-using-ser2net-socat-guide/34895/85

Thank you!

maggu2810:
> would this fix the issues I see when the console displays I/O errors because the virtual USB (mapped using ser2net & socat) has been removed or restarted?

I don't remember there has been any change in this PR that is related to your point(s).

codeaholics:
Yes, because OH is no longer opening a serial port. There is no `ser2net` or `socat` on the OH end - only on the remote device (ZWave stick) end. OH then natively talks to that using TCP/IP. So, assuming that this plugin correctly deals with connection drops and reconnects, then yes, this would solve the issue.

thefathefa:
Hi,

I was about to go this way : https://community.openhab.org/t/share-z-wave-dongle-over-ip-usb-over-ip-using-ser2net-socat-guide/34895 because I need such a setup but if there is a more "mainstream openhab way" that is coming, I'd rather wait... Is it possible to have an rough view of when this will be available in openhab 2.4 branches i.e. within weeks or months or quarters?
Thanks,

msteigenberger:
It is already available in latest snapshots of esh. I tested it in my binding:
https://github.com/openhab/openhab2-addons/pull/2167

dastrix80:
Hi msteigenberger, can you tell me how/ what this changes? I'm a little confused over how this helps the issue or what is now different with the latest snapshot. I'm constantly experiencing issues with my zwave stick connected to a remote device using socat/s2net and when that device has a restart, OH2 looses connection to the usb stick and requires a restart of the zwave binding.

msteigenberger:
@dastrix80: You don't need socat anymore that is located at the same machine as your OH2 instance. Instead OH2 now directly supports RFC2217 (which socat/ser2net is using).
When your remote device now disconnects/restarts, OH2 will fail to read the inputstream. This needs to be handled by the binding and a reconnection needs to be done.

lolodomo:
I will try this new feature with the Sony projector I am developing.
What is finally the required syntax ? Is the prefix rfc2217: required ?

Regarding the binding, is it necessary to suppress the context serial-port from the thing configuration setting to let the user enter a value such as rfc2217://192.168.0.10:3333 ?

msteigenberger:
//<ip> is possible and just means that it uses any remote implementation. As only rfc2217 is available, it would pick this. To be more precise you should rather use rfc2217://

You don't need to suppress the context in the configuration, as it would allow to enter manual values as well.

lolodomo:
Thank you for your answer.
As I got difficulty to make it work with a new binding, can you tell me what bindings were confirmed to work  with Rfc2217 ?
Regarding ser2net, shall I declare telnet or raw protocol ?

PR #5566, opened 2018-05-11
add missing presence-on icon

### LINKED ISSUE #5416: [Classic Icon Set] "presence-on" icon is missing

sjka:
...as there previously was a `presence` and `presence-off` icon, but no `presence-on` icon.

fixes #5416


sjka:
And again I learned something new... 
Sure, will do!

kaikreuzer:
> as the copies are committed to the repo as well

Hm, I missed that you omitted that part - I've created https://github.com/eclipse/smarthome/pull/5575 to fix that.

sjka:
Ouch, indeed, I overlooked that instruction.

Having a script to copy files once which then will be checked-in apparently is so weird that my brain neglected to acknowledge this piece of information...

PR #5572, opened 2018-05-14
Tradfri: Implement ConfigStatusBridgeHandler

ivivanov-bg:
Subclass ConfigStatusBridgeHandler to support configuration status

### DIFF
--- a/extensions/binding/org.eclipse.smarthome.binding.tradfri/src/main/java/org/eclipse/smarthome/binding/tradfri/handler/TradfriGatewayHandler.java
+++ b/extensions/binding/org.eclipse.smarthome.binding.tradfri/src/main/java/org/eclipse/smarthome/binding/tradfri/handler/TradfriGatewayHandler.java
@@ -18,6 +18,7 @@ import java.io.IOException;
 import java.net.InetSocketAddress;
 import java.net.URI;
 import java.net.URISyntaxException;
+import java.util.Collection;
 import java.util.Set;
 import java.util.UUID;
 import java.util.concurrent.CopyOnWriteArraySet;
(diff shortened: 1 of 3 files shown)


extensions/binding/org.eclipse.smarthome.binding.tradfri/src/main/java/org/eclipse/smarthome/binding/tradfri/handler/TradfriGatewayHandler.java
```diff
@@ -392,4 +428,20 @@ public void thingUpdated(Thing thing) {
...
         }
     }
+
+    @Override
+    public Collection<ConfigStatusMessage> getConfigStatus() {
+        while (getThing().getStatus() == ThingStatus.UNKNOWN || getThing().getStatus() == ThingStatus.INITIALIZING
+                || (getThing().getStatus() != ThingStatus.ONLINE && currentConfigStatus.getConfigStatusMessages().isEmpty())) {
+            try {
+                synchronized (this) {
+                    this.wait();
```
sjka: You don't want to block the calling thread, do you? 😉

ivivanov-bg: Actually - that's exactly what I want

htreu: without knowing all the details this sounds like it should be handled in an asynchrone way. The UI should idle as long as the desired state is not reached. The backend should not be blocked by this. wdyt?

PR #5579, opened 2018-05-15
Paper UI: show bridge selection in thing editor

### LINKED ISSUE #5570: [Paper UI] Bridge selection is hidden, if no bridge instance exists
When manually creating a Thing that requires a Bridge, the Bridge selection field is missing, if no Bridge instance exists so far:

screen shot 2018-05-13 at 17 20 42

If there is a Bridge, the selection is shown:

screen shot 2018-05-13 at 17 22 19

I would expect that the Paper UI clearly shows that this Thing requires a Bridge. If none exists, it would be nice if it could directly offer to re-direct to the creation of the Bridge first or to at least clearly communicate that this is required.

htreu:
This makes sure to show the bridge selection in the thing editor always if the thing requires a bridge. If no bridge is defined yet, provides a button to directly create one.

Fixes #5570.

Also: Some structural refactoring in the Paper UI things module (1st commit).

Update:
Added magic bridges & refactored thing editor.


sjka:
Works for me, thanks!

FTR: I took the freedom to put a commit on top which makes the bridge's label start with all capitals.

Issue #5578, opened 2018-05-15
[SNAPSHOT 1276] - Warning on "Members of" triggers


vzorglub:
Just updated to 1276
Ubuntu 16.04 intel
Repo installation

I get these warnings on ALL the rules with "Member of" triggers

==> /var/log/openhab2/openhab.log <==
```
2018-05-15 08:09:36.064 [WARN ] [pse.smarthome.core.items.GenericItem] - failed notifying listener 'org.eclipse.smarthome.model.rule.runtime.internal.engine.RuleEngineImpl@51c2828' about state update of item Bathroom_ThermostatTarget: null
java.lang.NullPointerException: null
	at org.eclipse.smarthome.model.rule.runtime.internal.engine.RuleTriggerManager.internalGetChangeRules(RuleTriggerManager.java:344) [136:org.eclipse.smarthome.model.rule.runtime:0.10.0.201805141751]
	at org.eclipse.smarthome.model.rule.runtime.internal.engine.RuleTriggerManager.internalGetRules(RuleTriggerManager.java:429) [136:org.eclipse.smarthome.model.rule.runtime:0.10.0.201805141751]
	at org.eclipse.smarthome.model.rule.runtime.internal.engine.RuleTriggerManager.getRules(RuleTriggerManager.java:198) [136:org.eclipse.smarthome.model.rule.runtime:0.10.0.201805141751]
	at org.eclipse.smarthome.model.rule.runtime.internal.engine.RuleEngineImpl.stateChanged(RuleEngineImpl.java:202) [136:org.eclipse.smarthome.model.rule.runtime:0.10.0.201805141751]
	at org.eclipse.smarthome.core.items.GenericItem$1.run(GenericItem.java:251) [101:org.eclipse.smarthome.core:0.10.0.201805141751]
	at java.util.concurrent.ThreadPoolExecutor.runWorker(ThreadPoolExecutor.java:1149) [?:?]
	at java.util.concurrent.ThreadPoolExecutor$Worker.run(ThreadPoolExecutor.java:624) [?:?]
	at java.lang.Thread.run(Thread.java:748) [?:?]
```
==> /var/log/openhab2/events.log <==

2018-05-15 08:09:36.070 [vent.ItemStateChangedEvent] - Bathroom_ThermostatTarget changed from 19.7 to 19.3

kaikreuzer:
FTR, this bug has been introduced by https://github.com/eclipse/smarthome/pull/4383

PR #5580, opened 2018-05-15
check for null in trigger states and added execution tests

kaikreuzer:
fixes https://github.com/eclipse/smarthome/issues/5578

### DIFF
--- a/bundles/model/org.eclipse.smarthome.model.rule.tests/src/org/eclipse/smarthome/model/rule/runtime/internal/engine/RuleExecutionTest.java
+++ b/bundles/model/org.eclipse.smarthome.model.rule.tests/src/org/eclipse/smarthome/model/rule/runtime/internal/engine/RuleExecutionTest.java
@@ -1,0 +1,163 @@
+/**
+ * Copyright (c) 2014,2018 Contributors to the Eclipse Foundation
+ *
+ * See the NOTICE file(s) distributed with this work for additional
+ * information regarding copyright ownership.
+ *
+ * This program and the accompanying materials are made available under the
+ * terms of the Eclipse Public License 2.0 which is available at
+ * http://www.eclipse.org/legal/epl-2.0
+ *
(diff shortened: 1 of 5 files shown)


sjka:
Cool to finally have some tests! 
Did you maybe forget to commit the actual fix?

bundles/model/org.eclipse.smarthome.model.rule.tests/src/org/eclipse/smarthome/model/rule/runtime/internal/engine/RuleExecutionTest.java
```diff
@@ -0,0 +1,163 @@
...
+ */
+public class RuleExecutionTest extends JavaOSGiTest {
+
+    private final static String TESTMODEL_NAME = "testModel.rules";
+    private ModelRepository modelRepository;
+    private ItemRegistry itemRegistry;
+    private EventPublisher eventPublisher;
+    private Event resultEvent;
+
+    private final EventSubscriber eventSubscriber = new EventSubscriber() {
```
sjka: Wouldn't this rather be an excellent candidate for a mock? You could grab the expected event with an `ArgumentCaptor`...

kaikreuzer:
> Did you maybe forget to commit the actual fix?

Haha, seems that got lost when updating the license headers...
I just wanted to proof that the tests fail without the fix ;-)

bundles/model/org.eclipse.smarthome.model.rule.tests/src/org/eclipse/smarthome/model/rule/runtime/internal/engine/RuleExecutionTest.java
kaikreuzer: I'd prefer to leave it as it is for now - feel free to refactor it later on, if you think it becomes more readable with a mock.

PR #5587, opened 2018-05-17
Documentation: FAQ: Explain how to path images through channels via RawType

### LINKED ISSUE #3989: New ImageType
I'm wondering if a binding generates or provides images in a channel for UIs, what is the type I should choose. I see those possible options at the moment:

* Store the image in the http/static folder and provide a link through a StringType
* Use a RawType and send the binary image. This is done here: https://github.com/eclipse/smarthome/blob/master/extensions/binding/org.eclipse.smarthome.binding.sonos/src/main/java/org/eclipse/smarthome/binding/sonos/handler/ZonePlayerHandler.java#L670

This issue is either about introducing something like an ImageType with the fields `binaryData` and a `imageType` enum (JPEG,PNG,BMP,WEBP) or extending the documentation where one of the options above is recommended instead.

triller-telekom:
Fixes #3989

### DIFF
--- a/docs/documentation/development/bindings/faq.md
+++ b/docs/documentation/development/bindings/faq.md
@@ -36,3 +36,9 @@ Here is a list of frequently asked questions around the development of bindings.
 5. _When creating a Thing through my ThingHanderFactory, does it exactly have to have the channels that are defined in the thing type description?_
  
     It must at least have the channels that are defined in its thing type (and they are already automatically added by the super() implementation). Nonetheless, you are free to add any number of additional channels to the thing.
+
+### Presenting device data to a user
+
+1. I have an image in my binding and want to pass it to the framework to be shown in UIs. What is the best way to achieve this?
+
+    The Thing that wants to provide the image should have a Channel defined with `<item-type>Image</item-type>`. The `ThingHandler` can update this Channel with a state of type `RawType` that represents the raw bytes of the image. If the image should be downloaded from a URL, the helper method `HttpUtil.downloadImage(URL url)` can be used. A user may link this Channel to an Item of type `Image` which can then be displayed.
--- a/docs/documentation/development/bindings/faq.md
+++ b/docs/documentation/development/bindings/faq.md
@@ -37,8 +37,12 @@ Here is a list of frequently asked questions around the development of bindings.
  
     It must at least have the channels that are defined in its thing type (and they are already automatically added by the super() implementation). Nonetheless, you are free to add any number of additional channels to the thing.
 
-### Presenting device data to a user
+### State Updates of Channels
 
-1. I have an image in my binding and want to pass it to the framework to be shown in UIs. What is the best way to achieve this?
+1. I have an image in my binding and want to pass it to the framework. What is the best way to achieve this?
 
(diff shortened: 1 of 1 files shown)


docs/documentation/development/bindings/faq.md
```diff
@@ -36,3 +36,9 @@ Here is a list of frequently asked questions around the development of bindings.
 5. _When creating a Thing through my ThingHanderFactory, does it exactly have to have the channels that are defined in the thing type description?_
  
     It must at least have the channels that are defined in its thing type (and they are already automatically added by the super() implementation). Nonetheless, you are free to add any number of additional channels to the thing.
+
+### Presenting device data to a user
```
kaikreuzer: We usally use capitals in headers.
Apart from this, this header is a bit fuzzy as it sounds as if it is about UIs.
Better change it to something like "State Updates of Channels".

docs/documentation/development/bindings/faq.md
```diff
@@ -36,3 +36,9 @@ Here is a list of frequently asked questions around the development of bindings.
...
+
+1. I have an image in my binding and want to pass it to the framework to be shown in UIs. What is the best way to achieve this?
```
kaikreuzer: Remove "to be shown in UIs" - you might want to have it sent by e-mail or be persisted in a database; it simply does not matter.

kaikreuzer: Add a line break after each sentence.

triller-telekom:
@lolodomo Good point! I have added this to the paragraph. I have also incorporated the changes from @kaikreuzer.

docs/documentation/development/bindings/faq.md
```diff
@@ -37,8 +37,12 @@ Here is a list of frequently asked questions around the development of bindings.
...
+1. I have an image in my binding and want to pass it to the framework. What is the best way to achieve this?
 
-    The Thing that wants to provide the image should have a Channel defined with `<item-type>Image</item-type>`. The `ThingHandler` can update this Channel with a state of type `RawType` that represents the raw bytes of the image. If the image should be downloaded from a URL, the helper method `HttpUtil.downloadImage(URL url)` can be used. A user may link this Channel to an Item of type `Image` which can then be displayed.
+    The Thing that wants to provide the image should have a Channel defined with `<item-type>Image</item-type>`.
+The `ThingHandler` can update this Channel with a state of type `RawType` that represents the raw bytes of the image.
+If the image should be downloaded from a URL, the helper method `HttpUtil.downloadImage(URL url)` can be used.
+A user may link this Channel to an Item of type `Image` which can then be displayed.
+Please note that data put as a `RawType` in a Channel will stay in **memory only**, i.e., this data will **not** be persisted anywhere.
```
lolodomo: Is it true ?
What is important to note IMHO is that the author of a binding should have in mind that such an image will eat memory on the server side.

lolodomo:
What you changed is not what I was mentionning. What I have in mind is that the memory used by such item types on the server side.

lolodomo:
To say that in a different way, a binding should avoid having 100  items of type RawType for example nad take care not having very very big images.

docs/documentation/development/bindings/faq.md
triller-telekom: Yes it is true, AFAIK. We do not store the data on the file-system. If you restart the server all `RawType` information will be lost.

I have added a sentence explaining that the memory of the server is used.

lolodomo: Ok, thank you.

Issue #4794, opened 2017-12-21
Hue addon not switching all lights in big groups


TheNetStriker:
Since the latest Hue firmware update (version 01038802) the Hue api seams to be blocking rapid http requests. This happens e.g. when switching on/off lights using a group with a lot of lights. (In my case from 7 lights up) This results in several lights that do not react to the command from OpenHab.

I've reproduced this problem simply by using the Chrome ARC plugin and uploaded a video showing the problem [here](https://www.youtube.com/watch?v=nbn4KN1kHCU).

The first commands are working but when I try to send multiple request in short time the requests get blocked by the api.

I also tested what happens if I control the lights individually in a rule and add a Thread:sleep(100) between the requests and this seams to solve the problem.

Is there a way to increase the concurrent requests on the Hue bridge or could you implement a setting to limit the concurrent requests to the hue bridge in the addon?

kaikreuzer:
Thanks for the analysis @TheNetStriker! This indeed sounds as if some throttling has to be implemented in the hue binding - we already have something similar for TRADFRI afaik.

TheNetStriker:
Hi Kai, as I just found out the problem in my video seams to be that the button turns into an abort button and if it is fast clicked the request gets aborted by the button. But I still have the problem that not all my Hue lights react when I switch a huge group. Can I provide you more information about the problem somehow?

kaikreuzer:
> Can I provide you more information about the problem somehow?

Well, you should make sure that your issue report contains all required information that is necessary to easily reproduce a problem. Once any developer wants to look into it (which won't be me any time soon, sorry), he might come back with further questions.

TheNetStriker:
Hi Kai, I now analyzed this using tcpdump to see which http packets get sent to the bridge using this command:

tcpdump -i eth0 host 172.17.4.13 and port 80 -n -s 0 -vvv -w /var/log/openhab2/http.pcap

After that I analyzed the result on https://packettotal.com and I did come to the following conclusion:

- I started with 22 of my 32 lights switched on. I then switched the lights off using the main light group containing other groups which contain all my lights: [Click here to see my group configuration](https://pasteboard.co/H1p0zEC.png)
[...]

PR #5591, opened 2018-05-18
Implement message delay

TheNetStriker:
This commit solves the problem described in [this issue](https://github.com/eclipse/smarthome/issues/4794) by implementing a message delay [as suggested by Philips](https://developers.meethue.com/documentation/hue-system-performance).

Signed-off-by: David Masshardt <david@masshardt.ch> (github: TheNetStriker)

### DIFF
--- a/extensions/binding/org.eclipse.smarthome.binding.hue/src/main/java/org/eclipse/smarthome/binding/hue/internal/HttpClient.java
+++ b/extensions/binding/org.eclipse.smarthome.binding.hue/src/main/java/org/eclipse/smarthome/binding/hue/internal/HttpClient.java
@@ -18,10 +18,16 @@ import java.io.InputStream;
 import java.io.OutputStreamWriter;
 import java.net.HttpURLConnection;
 import java.net.URL;
+import java.util.LinkedList;
+import java.util.concurrent.CompletableFuture;
+import java.util.concurrent.Future;
+import java.util.concurrent.ScheduledExecutorService;
 
 import org.apache.commons.io.IOUtils;
 import org.eclipse.jdt.annotation.NonNullByDefault;
 import org.eclipse.jdt.annotation.Nullable;
+import org.slf4j.Logger;
+import org.slf4j.LoggerFactory;
 
 /**
  *
@@ -31,6 +37,44 @@ import org.eclipse.jdt.annotation.Nullable;
 @NonNullByDefault
 public class HttpClient {
     private int timeout = 1000;
+    private final Logger logger;
+    private final LinkedList<AsyncPutParameters> commandsQueue;
+    private final Runnable commandExecutor;
+    @Nullable
+    private Future<?> job;
+
+    public HttpClient() {
+        logger = LoggerFactory.getLogger(getClass());
+        commandsQueue = new LinkedList<>();
+
+        commandExecutor = () -> {
+            while (true) {
+                try {
+                    int delayTime = 0;
+                    synchronized (this.commandsQueue) {
+                        AsyncPutParameters payloadCallbackPair = HttpClient.this.commandsQueue.poll();
+                        if (payloadCallbackPair != null) {
+                            logger.debug("Async sending put to address: {} delay: {} body: {}",
+                                    payloadCallbackPair.address, payloadCallbackPair.delay, payloadCallbackPair.body);
+                            try {
+                                Result result = HttpClient.this.put(payloadCallbackPair.address,
+                                        payloadCallbackPair.body);
+                                payloadCallbackPair.future.complete(result);
+                            } catch (IOException e) {
+                                payloadCallbackPair.future.completeExceptionally(e);
+                            }
+                            delayTime = payloadCallbackPair.delay;
+                        } else {
+                            return;
+                        }
+                    }
+                    Thread.sleep(delayTime);
+                } catch (InterruptedException e) {
+                    logger.debug("commandExecutorThread was interrupted", e);
+                }
+            }
+        };
+    }
 
     public void setTimeout(int timeout) {
         this.timeout = timeout;
@@ -46,6 +90,24 @@ public class HttpClient {
 
     public Result put(String address, String body) throws IOException {
         return doNetwork(address, "PUT", body);
+    }
+
+    public CompletableFuture<Result> putAsync(String address, String body, Integer delay,
+            ScheduledExecutorService scheduler) {
+        AsyncPutParameters asyncPutParameters = new AsyncPutParameters(address, body, delay);
+
+        synchronized (this.commandsQueue) {
+            if (this.commandsQueue.isEmpty()) {
+                this.commandsQueue.offer(asyncPutParameters);
+                if (job == null || job.isDone()) {
+                    job = scheduler.submit(commandExecutor);
+                }
+            } else {
+                this.commandsQueue.offer(asyncPutParameters);
+            }
+        }
+
+        return asyncPutParameters.future;
     }
 
     public Result delete(String address) throws IOException {
@@ -90,6 +152,20 @@ public class HttpClient {
 
         public int getResponseCode() {
             return responseCode;
+        }
+    }
+
+    public final class AsyncPutParameters {
+        public final String address;
+        public final String body;
+        public final CompletableFuture<Result> future;
+        public final Integer delay;
+
+        public AsyncPutParameters(String address, String body, Integer delay) {
+            this.address = address;
+            this.body = body;
+            this.future = new CompletableFuture<Result>();
+            this.delay = delay;
         }
     }
 }
--- a/extensions/binding/org.eclipse.smarthome.binding.hue/src/main/java/org/eclipse/smarthome/binding/hue/internal/HueBridge.java
+++ b/extensions/binding/org.eclipse.smarthome.binding.hue/src/main/java/org/eclipse/smarthome/binding/hue/internal/HueBridge.java
@@ -22,6 +22,8 @@ import java.util.ArrayList;
 import java.util.Date;
 import java.util.List;
 import java.util.Map;
+import java.util.concurrent.CompletableFuture;
+import java.util.concurrent.ScheduledExecutorService;
 
 import org.eclipse.jdt.annotation.NonNullByDefault;
 import org.eclipse.jdt.annotation.Nullable;
@@ -56,14 +58,16 @@ public class HueBridge {
 
     private final Gson gson = new GsonBuilder().setDateFormat(DATE_FORMAT).create();
     private HttpClient http = new HttpClient();
+    private ScheduledExecutorService scheduler;
 
     /**
      * Connect with a bridge as a new user.
      *
      * @param ip ip address of bridge
      */
-    public HueBridge(String ip) {
+    public HueBridge(String ip, ScheduledExecutorService scheduler) {
         this.ip = ip;
+        this.scheduler = scheduler;
     }
 
     /**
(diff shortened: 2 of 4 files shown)


htreu:
@TheNetStriker thanks for your first contribution to Eclipse SmartHome. Please use the `git commit -s` feature to also add the Signed-off-by footer to your commit message. The ip-validation checks the commit message, not the comments.

TheNetStriker:
I think it worked. Please check if the changes are good and give me a feedback if I have to change something.

extensions/binding/org.eclipse.smarthome.binding.hue/src/main/java/org/eclipse/smarthome/binding/hue/internal/HueBridge.java
```diff
@@ -862,7 +867,7 @@ private void requireAuthentication() {
     }
 
     // Used as assert in all requests to elegantly catch common errors
-    private void handleErrors(Result result) throws IOException, ApiException {
+    public void HandleErrors(Result result) throws IOException, ApiException {
```
maggu2810: Why do you change the first letter of the method name to upper case?

extensions/binding/org.eclipse.smarthome.binding.hue/src/main/java/org/eclipse/smarthome/binding/hue/internal/HttpClient.java
```diff
@@ -31,6 +37,46 @@
 @NonNullByDefault
 public class HttpClient {
     private int timeout = 1000;
+    private final Logger logger;
+    private final LinkedList<AsyncPutParameters> commandsQueue;
```
maggu2810: Use a `LinkedBlockingQueue` and remove the synchronized statements...

extensions/binding/org.eclipse.smarthome.binding.hue/src/main/java/org/eclipse/smarthome/binding/hue/internal/HttpClient.java
```diff
@@ -31,6 +37,46 @@
...
+                                payloadCallbackPair.future.completeExceptionally(e);
+                            }
+                            delayTime = payloadCallbackPair.delay;
+                        } else {
+                            return;
+                        }
+                    }
+                    Thread.sleep(delayTime);
+                } catch (InterruptedException e) {
+                    logger.debug("commandExecutorThread was interrupted", e);
```
maggu2810: The thread should be marked to be interrupted

extensions/binding/org.eclipse.smarthome.binding.hue/src/main/java/org/eclipse/smarthome/binding/hue/internal/HttpClient.java
```diff
@@ -31,6 +37,46 @@
...
+    private final Runnable commandExecutor;
+    @Nullable
+    private Future<?> job;
+
+    public HttpClient() {
+        logger = LoggerFactory.getLogger(getClass());
+        commandsQueue = new LinkedList<>();
+
+        commandExecutor = () -> {
+            while (true) {
```
maggu2810: The loop is left if a null reference is returned by poll.
At which place null is be added to the queue?

extensions/binding/org.eclipse.smarthome.binding.hue/src/main/java/org/eclipse/smarthome/binding/hue/internal/HueBridge.java
TheNetStriker: I always write pulic methods with a capital Letter, but I can change this if you have other guidelines.

maggu2810: I would claim that we follow the "normal" Java naming converntion.
http://www.oracle.com/technetwork/java/codeconventions-135099.html

> Methods should be verbs, in mixed case with the first letter lowercase,  with the first letter of each internal word capitalized.

extensions/binding/org.eclipse.smarthome.binding.hue/src/main/java/org/eclipse/smarthome/binding/hue/internal/HttpClient.java
TheNetStriker: I will change that.

extensions/binding/org.eclipse.smarthome.binding.hue/src/main/java/org/eclipse/smarthome/binding/hue/internal/HttpClient.java
TheNetStriker: What exactly do you mean by this? Can you give me a code example?

maggu2810: See e.g. https://stackoverflow.com/a/3976377

Another question: Shouldn't we leave the loop if someone interrupt that execution?

extensions/binding/org.eclipse.smarthome.binding.hue/src/main/java/org/eclipse/smarthome/binding/hue/internal/HttpClient.java
TheNetStriker: As discussed with Kai in the issue (https://github.com/eclipse/smarthome/issues/4794) I've copied this code from the Tradfri binding. (https://github.com/eclipse/smarthome/blob/4204ce06bb28c28e5f711e720f87ef83beff2e27/extensions/binding/org.eclipse.smarthome.binding.tradfri/src/main/java/org/eclipse/smarthome/binding/tradfri/internal/TradfriCoapClient.java)

Can you give me an example how this could be solved in a better way?

extensions/binding/org.eclipse.smarthome.binding.hue/src/main/java/org/eclipse/smarthome/binding/hue/internal/HttpClient.java
TheNetStriker: Good idea, or else we would have an infinite loop. I will make these changes.

TheNetStriker: On second thought I see that this isn't really necessary. If the exception is catched on one message the other messages in the loop should still be sent. And if there are no other messages the loop would be exited on the next run. So this should be okay. I will only make the change for the thread interruption.

extensions/binding/org.eclipse.smarthome.binding.hue/src/main/java/org/eclipse/smarthome/binding/hue/internal/HttpClient.java
maggu2810: Perhaps I was not clear...
The http client object is created here: https://github.com/TheNetStriker/smarthome/blob/350ec17ba5218bb650b158a8b2afd0ffc88ef45d/extensions/binding/org.eclipse.smarthome.binding.hue/src/main/java/org/eclipse/smarthome/binding/hue/internal/HueBridge.java#L60
At the first call to `putAsync` the command executor is scheduled:
https://github.com/TheNetStriker/smarthome/blob/350ec17ba5218bb650b158a8b2afd0ffc88ef45d/extensions/binding/org.eclipse.smarthome.binding.hue/src/main/java/org/eclipse/smarthome/binding/hue/internal/HttpClient.java#L105
Now the endless loop is entered.
My question is: At which place this endless loop is exit (e.g. if a null element has been added to the queue).
I don't identify the logic how the loop is stopped if e.g. the binding is removed at runtime.
If it is e.g. updated isn't there the loop from the old binding dangling and a new one is started, too?

maggu2810: Okay, sorry. Now I got it, too :wink:

extensions/binding/org.eclipse.smarthome.binding.hue/src/main/java/org/eclipse/smarthome/binding/hue/internal/HttpClient.java
maggu2810: IMHO the interrupt flag should be set all the time in interrupt exception has been catch, so if you debug it later you can identify the flag.
I don't think this will stop the entire thread.
https://docs.oracle.com/javase/8/docs/api/java/lang/Thread.html#interrupt--

But we can leave it as it is, too.

TheNetStriker: This is the documetion of interrupts from Oracle:

An interrupt is an indication to a thread that it should stop what it is doing and do something else. It's up to the programmer to decide exactly how a thread responds to an interrupt, but it is very common for the thread to terminate. 

I don't know how the ScheduledExecutorService will react to this, so I leave it as it is. Or else I also would have to test the whole binding again for a longer time.

TheNetStriker:
@maggu2810 I just commited the changed code. Please check again if everything is ok.

maggu2810:
Hi @TheNetStriker,
the comment about the synchronization and the LinkedBlockingQueue is not really try anymore after you shown me that use poll the element from the list and don't want to wait at all.
I assume we have to use the synchronization statement (as you have used before) on adding and(!) polling elements from that list.
Sorry that you changed it on my feedback -- you just revert that changes.

But I think the "job = null" should be placed into a synchronized statement, too.
Only this way you can ensure that the putAsync and "job = null" assignment does not lead into a wrong behavior (@sjka do you agree?)

TheNetStriker:
@maggu2810 Do you mean on line 101 in HttpClient.java? But this part is already synchronized on line 98. Or what do you mean?

TheNetStriker:
@maggu2810 I just changed the code back. Please check again if now everything is alright.

maggu2810:
@TheNetStriker The problem exists only if we reach the line where `job = null` is set (in the command executor).

If thread "commandExecutor" is immediately in front of `job = null` execution no synchronization block is entered anymore.
So, there could be another thread handled (e.g. because of a multi processor or a context switch) that execute the `putAsync` method. This method will check if `job == null` which will evaluate to `false` and check if `job.isDone()` which will evaluate to `false`, too. So, no new job is scheduled.
After that the `job = null` could be executed by some CPU and the offered parameters will not be executed as now no job is executed anymore until the next `putAsync` call.

If a `poll` of the command queue returns `null` then `return` is called and `job = null` will not be executed (it would be a problem if there has been `break` been used and the lines after the loop are executed).
If an interrupt execution is catch by the commandExecutor thread then the line will be executed. But I assume also that this will not be send to that thread at all.

So, not a real problem, but why calling `job = null` at all?
If job has never been set, the `job == null` in `putAsync` will evaluate to true, otherwise `job.isDone()`.
Is there any need to set it to null?
There could be another race condition if `job = null` is executed between `job == null` and `job.isDone()` as this could result in a NPE...
And the other point is that we set a variable in different threads without any synchronization if that variable (it should be volatile if it is read and written by different threads).

So, WDYT about removing `job = null`?
Perhaps I miss why this is necessary...

TheNetStriker:
@maggu2810 Ah now I see what you mean. Yes the job = null line is not really required, but I the line also doesn't do anything bad because this is only called in the initialization of the class and the commandExecutor is also only initialized at this moment and will be started later in the putAsync method. But I removed the line nevertheless. Please check again if everything looks good now.

maggu2810:
@TheNetStriker Thank you for your contribution and sorry for my first mis-understanding.

davidgraeff:
I think this PR broke travis, at least my MQTT PR fails because of the hue binding.

maggu2810:
You are right. Should be fixed by #5653.
Sorry for the trouble.

PR #5605, opened 2018-05-23
avoid exception on when playing a stream if sink volume cannot be determined

kaikreuzer:
Since the audio manager refactoring, we see [warnings in the log](https://community.openhab.org/t/warning-audio-sink-warning-master-volume-level/44789) when playing a sound if a sink is not able to determine its volume (which is the case for some sinks).

This PR handles this situation gracefully.


sjka:
Computer says no:

```
groovy.lang.MissingPropertyException: No such property: getId for class: org.eclipse.smarthome.core.audio.test.fake.AudioSinkFake
	at org.codehaus.groovy.runtime.ScriptBytecodeAdapter.unwrap(ScriptBytecodeAdapter.java:51)
	at org.codehaus.groovy.runtime.callsite.GetEffectivePogoPropertySite.getProperty(GetEffectivePogoPropertySite.java:84)
	at org.codehaus.groovy.runtime.callsite.AbstractCallSite.callGetProperty(AbstractCallSite.java:227)
	at org.eclipse.smarthome.core.audio.test.AudioManagerTest.getVolume method handles IOException(AudioManagerTest.groovy:185)
```

kaikreuzer:
@cweitkamp Makes sense - I have updated the PR accordingly.

sjka:
Just added the APIBreaking label because of the added checked exceptions on a method which is part of the public API.

sjka:
This PR from my point of view definitely makes sense as it at least streamlines the `AudioManager` with the `AudioSinks` and lets the caller know that it didn't work.  

Nevertheless, it makes me wonder why `AudioManager.play(...)/.stream(...)` as well as `AudioSink.process(...)` don't throw _any_ checked exceptions while the `AudioManager.playFile(...)` methods throw a dedicated `AudioException` (as opposed to an `IOException`). This doesn't look very consistent to me. Shouldn't we revisit this whole topic again separately? What do you think?

kaikreuzer:
Ok to look into streamlining the whole API - maybe @cweitkamp would be interested in that (as he did the lastest refactorings as well)?

cweitkamp:
Yes, I would love to. I'll put it onto my TODO list.

PR #5607, opened 2018-05-23
fix Item confusion

sjka:
In #5207 the AbstractRegistry was changed to call
onUpdateElement(...) with the current live instance
instead of the oldElement instance which got handed
in by the provider.

However, the given oldElement was guaranteed to contain
the previous values, while the live instance potentially
is subject to modifications. At least as long as there
are e.g. ActiveItem & GenericItem part of the public API
and allow modification of the entities.

### DIFF
--- a/bundles/core/org.eclipse.smarthome.core.test/src/test/java/org/eclipse/smarthome/core/items/ItemRegistryOSGiTest.java
+++ b/bundles/core/org.eclipse.smarthome.core.test/src/test/java/org/eclipse/smarthome/core/items/ItemRegistryOSGiTest.java
@@ -179,7 +179,7 @@ public class ItemRegistryOSGiTest extends JavaOSGiTest {
     }
 
     @Test
-    public void testGroupUpdateWithMofifivationOfLiveInstance() {
+    public void testGroupUpdateWithModificationOfLiveInstance() {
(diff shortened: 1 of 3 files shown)


sjka:
@afuechsel thank you very much for pointing me to this issue! 
Would you like to have a second look if this indeed fixes the problem you observed?

bundles/core/org.eclipse.smarthome.core.test/src/test/java/org/eclipse/smarthome/core/items/ItemRegistryOSGiTest.java
```diff
@@ -180,6 +178,23 @@ public void assertItemRegistrySetsAndRemovesMembersOfGroupItems() throws ItemNot
         assertThat(groupItem3.getMembers().contains(updatedSwitchItem), is(true));
     }
 
+    @Test
+    public void testGroupUpdateWithMofifivationOfLiveInstance() {
```
afuechsel: Mofifi .. what?

tomhoefer:
Thank you very much @sjka

afuechsel:
@sjka I can confirm that our problem is solved with this fix. Thx a lot.

PR #5577, opened 2018-05-14
added HTTP client dependency to Karaf feature

kaikreuzer:
Signed-off-by: Kai Kreuzer <kai@openhab.org>

PR #5612, opened 2018-05-23
improve karaf target platform feature repository

maggu2810:
This replaces #5577

The feature handling has been improved by removing injection of multiple target platform features in the base one.

The Swagger feature usage needs to be checked again, but ATM there are at least no hard dependencies (AFAI can identify).

### DIFF
--- a/features/karaf/esh-core/src/main/feature/feature.xml
+++ b/features/karaf/esh-core/src/main/feature/feature.xml
@@ -17,9 +17,15 @@
 
   <feature name="esh-base" description="ESH" version="${project.version}">
     <details>ESH base bundles</details>
-    <feature dependency="true">esh-tp</feature>
+
+    <requirement>esh.tp;filter:="(feature=base)"</requirement>
+    <feature dependency="true">esh.tp-base</feature>
+
     <requirement>esh.tp;filter:="(&amp;(feature=xtext)(version&gt;=2.12.0)(!(version&gt;=2.13.0)))"</requirement>
+    <feature dependency="true">esh.tp-xtext</feature>
+
     <requirement>esh.tp;filter:="(feature=jax-rs)"</requirement>
+    <feature dependency="true">esh.tp-jax-rs</feature>
 
     <bundle>mvn:org.eclipse.smarthome.config/org.eclipse.smarthome.config.core/${project.version}</bundle>
     <capability>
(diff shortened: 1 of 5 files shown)


maggu2810:
@kaikreuzer Friendly ping... Have you already had a look at? I assume this is much more what you want.
We could improve the situation from time to time to if we split the `esh-base` and the `esh.tp-base` feature into smaller ones. But that does not need to be done now as it will be another big refactor.

kaikreuzer:
Sorry for all the delays from my side on the different issues/PRs - I am travelling this week and doing 3 presentations, all of which still need to be prepared... I'll be back with full dedication here on Github next week, promised!

maggu2810:
```text
  <feature name="esh-io-transport-coap" version="${project.version}">
    <feature>esh-base</feature>

    <requirement>esh.tp;filter:="(feature=coap)"</requirement>
    <feature dependency="true">esh.tp-coap</feature>
</feature>
```

The requirements of this feature consists of the requirement of the contained bundles and features.
Additionally we add a special requirement `esh.tp;filter:="(feature=coap)"`.

This special capability is e.g. provided by the feature `esh.tp-coap`.
This feature is marked to be installed only if requirements are not already satisfied and the respective feature satisfy such requirements.

So, if someone already installed a feature that provides the capability `esh.tp;filter:="(feature=coap)"` (or if he install it later explicit), the feature `esh.tp-coap` will not be installed (or removed again).

Just a side-note: If there is another requirement (e.g. import package statement in a bundle -- here this cannot be the case as there is no separate bundle in that feature) that is satisfied by `esh.tp-coap` but not by the custom provided one, `esh.tp-coap` will be installed because it satisfy requirement that will not be satisfied otherwise. But then the custom feature is incomplete and wrong.



An additional task that should / could be done some time is to reduce the base targetplatform as much as possible.
We could IMHO drop some of the Apache Commons as we don't use them anymore (at least it seems so).

kaikreuzer:
> This feature is marked to be installed only if requirements are not already satisfied

Ah, that was the part I wasn't yet completely clear about, thanks!
And sorry for still requiring lessons in Karaf dependency management 🙄 .

maggu2810:
You are welcome.
There is no such thing as a stupid question.
:wink:

kaikreuzer:
@maggu2810 I am afraid, I again need your help wrt Karaf features - might it be possible that this PR (or some other) has somehow reverted [this fix](https://github.com/eclipse/smarthome/commit/83d0cbbec1da41dc41ca2d8fbcdb662dcfc03326) from you? [Here is a report](https://community.openhab.org/t/error-messages-while-using-the-uis-openhab-2-4-0-m2-1/50295) that this problem just surfaced again :-(

Issue #5616, opened 2018-05-23
deactivated XmlBindingInfoProvider is used sometimes


maggu2810:
Sometimes I see the following message (just need to restart the container a few times):

```text
The XML document '/ESH-INF/binding/binding.xml' in module 'org.eclipse.smarthome.binding....' could not be parsed: The XmlConfigDescriptionProvider must not be null!
```

The binding itself does not care.

Here a full stacktrace (line numbers of `XmlBindingInfoProvider` are changed because I already added some logging info as this stacktrace has been generated):

```text
23:40:54.705 WARN [ESH-file-processing-5] The XML document '/ESH-INF/binding/binding.xml' in module 'org.eclipse.smarthome.binding.bosesoundtouch' could not be parsed: The XmlConfigDescriptionProvider must not be null!
java.lang.IllegalArgumentException: The XmlConfigDescriptionProvider must not be null!
        at org.eclipse.smarthome.core.binding.xml.internal.BindingInfoXmlProvider.<init>(BindingInfoXmlProvider.java:59) ~[?:?]
        at org.eclipse.smarthome.core.binding.xml.internal.XmlBindingInfoProvider.createDocumentProvider(XmlBindingInfoProvider.java:152) ~[?:?]
        at org.eclipse.smarthome.config.xml.osgi.XmlDocumentBundleTracker.acquireXmlDocumentProvider(XmlDocumentBundleTracker.java:174) ~[110:org.eclipse.smarthome.config.xml:0.10.0.201805232047]
        at org.eclipse.smarthome.config.xml.osgi.XmlDocumentBundleTracker.addingObject(XmlDocumentBundleTracker.java:199) ~[110:org.eclipse.smarthome.config.xml:0.10.0.201805232047]
        at org.eclipse.smarthome.config.xml.osgi.XmlDocumentBundleTracker.parseDocuments(XmlDocumentBundleTracker.java:343) [110:org.eclipse.smarthome.config.xml:0.10.0.201805232047]
        at org.eclipse.smarthome.config.xml.osgi.XmlDocumentBundleTracker.processBundle(XmlDocumentBundleTracker.java:329) [110:org.eclipse.smarthome.config.xml:0.10.0.201805232047]
        at org.eclipse.smarthome.config.xml.osgi.XmlDocumentBundleTracker.access$3(XmlDocumentBundleTracker.java:324) [110:org.eclipse.smarthome.config.xml:0.10.0.201805232047]
[...]

PR #5617, opened 2018-05-23
xml document bundle tracker: execute submitted task only if opened

maggu2810:
Fixes: https://github.com/eclipse/smarthome/issues/5616


sjka:
👍

PR #5625, opened 2018-05-24
Added ItemRegistry#removeTags(String) to remove ALL tags for an item.

afuechsel:
Also fixed the bug for deleting all tags.


sjka:
Thanks!

PR #5627, opened 2018-05-25
Send PLAY_PAUSE command on player control channel

ivivanov-bg:
Apparently Bose SoundTouch doesn't properly handle separate PLAY and PAUSE commands

Further more - on the remote - both commands are combined in a single button.

As a result - when the player is powered off (it's in StandBy mode) - sending a PLAY command does not bring the device ON
Sending a PLAY_PAUSE command however does and the last playback is started (similar as when pressing the play_pause button on the remote)

### DIFF
--- a/extensions/binding/org.eclipse.smarthome.binding.bosesoundtouch/src/main/java/org/eclipse/smarthome/binding/bosesoundtouch/internal/CommandExecutor.java
+++ b/extensions/binding/org.eclipse.smarthome.binding.bosesoundtouch/src/main/java/org/eclipse/smarthome/binding/bosesoundtouch/internal/CommandExecutor.java
@@ -190,10 +190,8 @@ public class CommandExecutor implements AvailableSources {
      * @param command the command is Type of Command
      */
     public void postPlayerControl(Command command) {
-        if (command.equals(PlayPauseType.PLAY)) {
-            postRemoteKey(RemoteKeyType.PLAY);
-        } else if (command.equals(PlayPauseType.PAUSE)) {
-            postRemoteKey(RemoteKeyType.PAUSE);
+        if (command.equals(PlayPauseType.PLAY) || command.equals(PlayPauseType.PAUSE)) {
+            postRemoteKey(RemoteKeyType.PLAY_PAUSE);
         } else if (command.equals(NextPreviousType.NEXT)) {
             postRemoteKey(RemoteKeyType.NEXT_TRACK);
         } else if (command.equals(NextPreviousType.PREVIOUS)) {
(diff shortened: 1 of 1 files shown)


kaikreuzer:
I do not really understand that explanation.
I assume that if there is a pause command available that this also works.
Your PR now causes a PAUSE command to actually start playing, which does not make sense.
If all you want to address is the lacking wake-up, why don't you simply call a playpause for the PLAY command (potentially only when being in standby)?

ivivanov-bg:
The only way in PaperUI to send a PAUSE command is if the player is actually playing (as both states share the same button) - so it doesn't make sense to me to handle both commands separately - the player is either playing and you send pause or is paused and you send play - there is no way to send pause while it's paused and vice verse

kaikreuzer:
The Paper UI is just some pretty constrained development/admin tool. Commands can be sent from a million other places...

kaikreuzer:
@ivivanov-bg Any update from your end? How shall we proceed?

ivivanov-bg:
We will have a sync call with Bose later this week, but I think eventually we will end up with sending two commands - one for power-on and one for play (if you are OK with this)

kaikreuzer:
What's wrong with my suggestion in https://github.com/eclipse/smarthome/pull/5627#issuecomment-392060931?

ivivanov-bg:
That would also be possible - but I wanted to get official confirmation first, because I'm not sure how will the soundtouch 300 soundbar handle the PLAY_PAUSE command (I don't have such device and based on pics - it has dedicated buttons for PLAY and PAUSE on the remote - which makes me thing that it may not wake-up on a PLAY_PAUSE command).

Which will require to send two commands if we want to keep only one handler class for all players.

What do you think about that - does it makes sense to have only one handler implementation and adjust the API calls so that they work on all devices or having a separate one - probably only for the soundbar ?

kaikreuzer:
If the differences are minimal, I would try to keep going with a single handler implementation, yes. And if the "two command" solutions works reliably for all devices, I don't have any issue with that solution. So ok to wait for the advice from the Bose guys themselves.

kaikreuzer:
Any update here, @ivivanov-bg?

ivivanov-bg:
I was busy last few weeks with different stuff around IFA, but will try to follow-up here as well.
After a talk with Bose - they advised to not use only the PLAY_PAUSE command, but also to check the player state and based on it to either send 'PLAY' or 'PLAY_PAUSE'

kaikreuzer:
Ok, just ping me once you have updated the PR accordingly!

AlexanderKostadinov28:
@kaikreuzer I updated the PR on behalf of @ivivanov-bg with the suggested change, please have a look.

PR #5633, opened 2018-05-25
Add warning for advanced network configurations

J-N-K:

### DIFF
--- a/extensions/binding/org.eclipse.smarthome.binding.sonos/README.md
+++ b/extensions/binding/org.eclipse.smarthome.binding.sonos/README.md
@@ -2,6 +2,12 @@
 
 This binding integrates the [Sonos Multi-Room Audio system](http://www.sonos.com).
 
+**Attention:** 
+You might run into trouble if your Eclipse Smarthome instance is in another subnet than your Sonos device. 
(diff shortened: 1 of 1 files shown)


extensions/binding/org.eclipse.smarthome.binding.sonos/README.md
```diff
@@ -2,6 +2,12 @@
 
 This binding integrates the [Sonos Multi-Room Audio system](http://www.sonos.com).
 
+**Attention:** 
+You might run into trouble if your Eclipse Smarthome instance is in another subnet than your Sonos device.
```
kaikreuzer: Better talk of "your gateway" or something neutral - remember that those docs are e.g. taken over to the openHAB documentation website.

J-N-K: I was thinking about that, too. But „gateway“ is not a good idea either as that is widely used in another context in networking.

kaikreuzer: Then simply use "instance".

sjka:
After stuggling a bit stetting up my system using an IGMP proxy in between two VLANs I gave up. I couldn't get it to work, although the idea is pretty clear though... Do you maybe have a pointer to how you got it working?

PR #5643, opened 2018-05-28
ItemConsoleCommand: Distinguish between no item found and no item for pattern

### LINKED ISSUE #5641: Inconsistent karaf message from smarthome:items
When the command `smarthome:items list` is issued but no items are defined, the response is `No item found for this pattern.`.

But no pattern was supplied, so this message makes no sense.

Further, when the command `smarthome:things list` is issued but no things are defined, the response is `No things found.`

These two messages should be consistent with each other. The items command should result in a `No items found.` message.

triller-telekom:
Now consistent with output for smarthome things list

Fixes #5641

### DIFF
--- a/bundles/io/org.eclipse.smarthome.io.console/src/main/java/org/eclipse/smarthome/io/console/internal/extension/ItemConsoleCommandExtension.java
+++ b/bundles/io/org.eclipse.smarthome.io.console/src/main/java/org/eclipse/smarthome/io/console/internal/extension/ItemConsoleCommandExtension.java
@@ -151,7 +151,11 @@ public class ItemConsoleCommandExtension extends AbstractConsoleCommandExtension
                 console.println(item.toString());
             }
         } else {
-            console.println("No item found for this pattern.");
+            if (pattern == null || (pattern != null && pattern.isEmpty())) {
+                console.println("No item found");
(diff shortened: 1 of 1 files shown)


sjka:
Thanks!

bundles/io/org.eclipse.smarthome.io.console/src/main/java/org/eclipse/smarthome/io/console/internal/extension/ItemConsoleCommandExtension.java
```diff
@@ -151,7 +151,11 @@ private void listItems(Console console, String pattern) {
                 console.println(item.toString());
             }
         } else {
-            console.println("No item found for this pattern.");
+            if (pattern == null || (pattern != null && pattern.isEmpty())) {
```
maggu2810: Shouldn't this be:

```
if (pattern == null || pattern.isEmpty()) {
```

triller-telekom: Indeed, thanks. Was too early in the morning. For the fix please see #5645

PR #5645, opened 2018-05-28
Fix ItemConsoleCommand duplicate null check

triller-telekom:
Relates to #5643


triller-telekom:
It was too early in the morning for #5643...

PR #5647, opened 2018-05-29
added "dump" to metadata command

afuechsel:
Also changed "remove" command to get it working even if the item is no longer existing. 

### DIFF
--- a/bundles/io/org.eclipse.smarthome.io.console/src/main/java/org/eclipse/smarthome/io/console/internal/extension/MetadataConsoleCommandExtension.java
+++ b/bundles/io/org.eclipse.smarthome.io.console/src/main/java/org/eclipse/smarthome/io/console/internal/extension/MetadataConsoleCommandExtension.java
@@ -21,6 +21,7 @@ import java.util.stream.Collectors;
 import org.eclipse.smarthome.core.items.ItemRegistry;
 import org.eclipse.smarthome.core.items.Metadata;
 import org.eclipse.smarthome.core.items.MetadataKey;
+import org.eclipse.smarthome.core.items.MetadataPredicates;
 import org.eclipse.smarthome.core.items.MetadataRegistry;
 import org.eclipse.smarthome.io.console.Console;
 import org.eclipse.smarthome.io.console.extensions.AbstractConsoleCommandExtension;
@@ -39,7 +40,6 @@ import org.osgi.service.component.annotations.Reference;
 public class MetadataConsoleCommandExtension extends AbstractConsoleCommandExtension {
 
     private static final String SUBCMD_LIST = "list";
-    private static final String SUBCMD_DUMP = "dump";
     private static final String SUBCMD_ADD = "add";
     private static final String SUBCMD_REMOVE = "remove";
 
@@ -53,25 +53,21 @@ public class MetadataConsoleCommandExtension extends AbstractConsoleCommandExten
     @Override
     public List<String> getUsages() {
         return Arrays.asList(new String[] {
-                buildCommandUsage(SUBCMD_DUMP, "prints all existing metadata for all namespaces"),
-                buildCommandUsage(SUBCMD_LIST + " <itemName> <namespace>",
-                        "lists the metadata for the specific item in the given namespace"),
-                buildCommandUsage(SUBCMD_REMOVE + " <itemName> <namespace>",
-                        "removes metadata for the specific item in the given namespace"),
+                buildCommandUsage(SUBCMD_LIST + " [<itemName> [<namespace>]]",
+                        "lists all available metadata, can be filtered for a specifc item and namespace"),
+                buildCommandUsage(SUBCMD_REMOVE + " <itemName> [<namespace>]",
+                        "removes metadata for the specific item (for all namespaces or for the given namespace only)"),
                 buildCommandUsage(SUBCMD_ADD + " <itemName> <namespace> <value> [{key1=value1, key2=value2, ...}]",
                         "adds or updates metadata value (and optional config values) for the specific item in the given namespace") });
     }
 
     @Override
     public void execute(String[] args, Console console) {
-        if (args.length > 0 && SUBCMD_DUMP.equals(args[0]) || args.length > 2) {
+        if (args.length > 0) {
             String subCommand = args[0];
             switch (subCommand) {
-                case SUBCMD_DUMP:
-                    dumpMetadata(console);
-                    break;
                 case SUBCMD_LIST:
-                    listMetadata(console, args[1], args[2]);
+                    listMetadata(console, args.length > 1 ? args[1] : null, args.length > 2 ? args[2] : null);
                     break;
                 case SUBCMD_ADD:
                     if (args.length < 4) {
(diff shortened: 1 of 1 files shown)


afuechsel:
@htreu Please be aware, that this PR also contains a bug fix. So if the "dump" command will not be accepted, the bug fix is still needed.

sjka:
I'm struggling hard to make up my mind. Somehow no strong opinion makes any attempt to form. 
In general I don't think it hurts, could be quite useful for development. But I don't see any potential use-case for operations.

The only thing I'm wondering is whether it really makes sense to give it a new name `dump`. Wouldn't it rather be more intuitive to change the make the two parameters of `list` optional like this?

```
smarthome metadata list [<itemName> [<namespace>]]
```

Then the following would work:

* `smarthome metadata list`
  would return _all_ metadata
* `smarthome metadata list foo`
  would return all metadata of item `foo`
* `smarthome metadata list foo bar`
  would return the `bar` metadata of `foo`

Another thing I don't really like is that we are exposing internal metadata here. IMHO that should be kept internal. The REST resource also doesn't allow access to them. It's just an implementation detail that the metadata infrastructure is used for e.g. tags. As is the format in which they are stored there. I don't consider this API it is subject to change.

afuechsel:
Ok for list instead of dump. But namespace cannot be optional as there is no way to address metadata without a namespace. It’s part of MetadataKey. 

Exposing internal metadata - I don’t see an issue with that. As the console commands are mainly for developers, why not? May be with an optional parameter -internal? Inbox for example also allows to list hidden entries.

sjka:
> But namespace cannot be optional as there is no way to address metadata without a namespace. It’s part of MetadataKey.

Correct. How about implementing it like this?

```java
metadataRegistry.stream().filter(MetadataPredicates.ofItem(itemName))...
```

maggu2810:
> Another thing I don't really like is that we are exposing internal metadata here. IMHO that should be kept internal. The REST resource also doesn't allow access to them. It's just an implementation detail that the metadata infrastructure is used for e.g. tags.

Would it make sense to provide a function that returns only the non-internal ones?
If there are now at least two places that don't want to export the internal ones why should the caller filter it (and perhaps forget the filter)?

afuechsel:
@maggu2810 As a developer I'd like to see everything, even implementation details. If you want I could add a `listinternal` command (like `listignored` for inbox) that lists the internal ones and list only to return the non-internal ones. Sometimes it's even helpful to see that "implementation detail", as it revealed the fact to me, that deleting the items did actually not delete their tags, so I could fix that bug (#5625).

maggu2810:
@afuechsel My comment has not been about the console command. It is about about to make the usage of the API for Java Developers simpler and less error prune.
If every Java Developer that uses the ESH API needs to know that the metadata should be filtered to not export internal stuff to some external interface...
It would be more visible if there are two function (e.g. to get all metadata, to get only non-internal metadata) instead of rely that all developers knows that there are internal metadata and external one.

afuechsel:
@maggu2810 I agree but I'd suggest that API changes should be handled in a different PR. I will update my PR to support `list` and `listinternal`.

bundles/io/org.eclipse.smarthome.io.console/src/main/java/org/eclipse/smarthome/io/console/internal/extension/MetadataConsoleCommandExtension.java
```diff
@@ -91,21 +98,28 @@ public void execute(String[] args, Console console) {
...
+                Metadata metadata = metadataRegistry.get(key);
+                if (metadata != null) {
+                    console.println(metadata.toString());
+                }
             }
         }
     }
 
+    private boolean isInternal(@NonNull Metadata metadata, boolean internal) {
+        return metadataRegistry.isInternalNamespace(metadata.getUID().getNamespace()) == internal;
```
afuechsel: While this is working - I'd suggest to extend the Metadata API in a way to have `MetadataRegistry#isInternalNamespace` to be static and having a method `isInternalNamespace` added to  `MetadataKey` and `Metadata` itself.

afuechsel:
@maggu2810 More opinions regarding this PR?

bundles/io/org.eclipse.smarthome.io.console/src/main/java/org/eclipse/smarthome/io/console/internal/extension/MetadataConsoleCommandExtension.java
```diff
@@ -87,14 +98,28 @@ public void execute(String[] args, Console console) {
...
+            if (metadataRegistry.isInternalNamespace(namespace) == internal) {
+                Metadata metadata = metadataRegistry.get(key);
+                if (metadata != null) {
+                    console.println(metadata.toString());
+                }
+            }
         }
     }
 
+    private boolean isInternal(@NonNull Metadata metadata, boolean internal) {
```
maggu2810: We do not use `@NonNull` annotations instead we use `@NonNullByDefault` and `@Nullable`.

maggu2810:
This PR does not contain the changes for adding a dump (list, list internal) command only but also fixes bugs.
Would be great if we could split the separate changes into different PRs next time.

Git provides us a great "revert" feature that does not work if we mix different stuff in a commit.
Also the history does not really fit as I do not expect a bugfix in a "adding feature" commit.

afuechsel:
@maggu2810 you are right and I reverted the bugfix commit and created a separate PR: #5688

maggu2810:
> @maggu2810 you are right and I reverted the bugfix commit and created a separate PR: #5688

Great! Thanks!

One additional for the remove bugfix? Then we merge all that three changes separate.

afuechsel:
> One additional for the remove bugfix? Then we merge all that three changes separate.

?? Don't get it. The bugfix is removed from that PR and a separate PR created (where you already commented).

maggu2810:
Hm, Github still shows me changes in `removedMetadata`

afuechsel:
> Hm, Github still shows me changes in `removeMetadata` 

Sure, but as this PR is an extension of a newly introduced feature the change in `removeMetadata` is a bit related to the changes in the list command, where namespace is now optional, as @sjka suggested in on review comment. So I don't want to split this really small PR any further.

bundles/io/org.eclipse.smarthome.io.console/src/main/java/org/eclipse/smarthome/io/console/internal/extension/MetadataConsoleCommandExtension.java
```diff
@@ -87,14 +95,29 @@ public void execute(String[] args, Console console) {
...
+            if (metadataRegistry.isInternalNamespace(namespace) == internal) {
+                Metadata metadata = metadataRegistry.get(key);
+                if (metadata != null) {
+                    console.println(metadata.toString());
+                }
+            }
         }
     }
 
+    @NonNullByDefault
```
maggu2810: Missed it on review. We don't use `@NonNullByDefault` on methods but for the whole class.

Please have a look at our "conventions":
https://www.eclipse.org/smarthome/documentation/development/conventions.html#null-annotations

I will fix it in another PR.

PR #5650, opened 2018-05-29
json response: use status type interface instead of a specific impl.

maggu2810:


htreu:
Travis seems to fail unrelated with `[ERROR] The constructor HueBridge(String) is undefined`.

lolodomo:
Id this not API breaking ? No impact on existing code ?

htreu:
It should not break anything. `Response.StatusType` is the interface to `Response.Status` so every existing call is compatible.

PR #5652, opened 2018-05-30
Allow tag deletion of missing item

### LINKED ISSUE #5649: Removing tags for items
If I have added some tags to an item, that is not been created by a `ManagedProvider`, the tags will not be deleted in the `MetadataRegistry`, when the item is being removed. Adding the same item again it will have the same tags. This is by design AFAIK. 

Currently there is no easy way to delete those tags when the item is being deleted. Using `ItemRegistryChangeListener#removed` will be too late to call `ItemRegistry#removeTags` as the item is already removed in the registry and it will throw an `IllegalArgumentException`. `RegistryHook#afterRemoving` is also too late and there is no `RegistryHook#beforeRemoving`. 

It would be good extension to have such a `RegistryHook#beforeRemoving` method.

sjka:
Whenever a non-managed item was removed, the removal of its
tags should still be possible without the need to hook into
any of the item registries lifecycle methods.
Metadata has its own lifecycle, so tags should do too.

Also, this adapts the signature of

  ItemRegistry.removeTags(String)

to return a boolean like all the other tag related methods do.

fixes #5649

### DIFF
--- a/bundles/core/org.eclipse.smarthome.core.test/src/test/java/org/eclipse/smarthome/core/internal/items/ItemRegistryOSGiJavaTest.java
+++ b/bundles/core/org.eclipse.smarthome.core.test/src/test/java/org/eclipse/smarthome/core/internal/items/ItemRegistryOSGiJavaTest.java
@@ -169,9 +169,10 @@ public class ItemRegistryOSGiJavaTest extends JavaOSGiTest {
         assertTagsInItem(ITEM_NAME, "foo", "bar");
     }
 
-    @Test(expected = IllegalArgumentException.class)
+    @Test
     public void testAddTagToNonExistingItem() throws Exception {
         itemRegistry.addTag(ITEM_NAME, "hello");
+        assertTagsInMetadata(ITEM_NAME, "hello");
     }
 
     @Test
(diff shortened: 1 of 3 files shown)


bundles/core/org.eclipse.smarthome.core.test/src/test/java/org/eclipse/smarthome/core/internal/items/ItemRegistryOSGiJavaTest.java
```diff
@@ -194,9 +195,13 @@ public void testRemoveNonExistingTag() throws Exception {
...
     }
 
-    @Test(expected = IllegalArgumentException.class)
+    @Test
     public void testRemoveTagFromNonExistingItem() throws Exception {
+        prepareMetadata(ITEM_NAME, "hello");
+
         itemRegistry.removeTag(ITEM_NAME, "hello");
+
+        assertTagsInMetadata(ITEM_NAME);
```
afuechsel: Is this assertion correct? I mean, if you have removed the tag, the assertion will probably fail.

maggu2810: I assume (WRT the method name) you would like to remove the item and not the tag, correct?

maggu2810:
@sjka Please have a look at #5654

bundles/core/org.eclipse.smarthome.core.test/src/test/java/org/eclipse/smarthome/core/internal/items/ItemRegistryOSGiJavaTest.java
triller-telekom: No, he doesn't pass any `tags` to the `assertTagsInMetadata` method and thus this method verifies that for the given item there are no tags. So the assertion is correct.

sjka:
I have pushed a commit which adds locking.

I decided for a lock instead of pure synchronization as in this way parallel read-only won't be slowed down significantly. 

An even bigger solution would be to lock per item(name) only, but I think that would have been a little beyond what we need here, I think.

PR #5653, opened 2018-05-30
Add scheduler to HueBridgeHandlerOSGiTest

htreu:
Fixing the build.


extensions/binding/org.eclipse.smarthome.binding.hue.test/src/test/java/org/eclipse/smarthome/binding/hue/handler/HueBridgeHandlerOSGiTest.java
```diff
@@ -241,4 +247,17 @@ private void injectBridge(HueBridgeHandler hueBridgeHandler, HueBridge bridge) {
             throw new AssertionError(e);
         }
     }
+
+    private class HueHandlerExt extends HueBridgeHandler {
```
triller-telekom: Why do you need this class? I do not see that it is used within the test...

maggu2810:
To get the build green again, I take the freedom not to wait for Travis.

PR #5338, opened 2018-04-03
POM Code Convention

maggu2810:
There is a recommended ordering for all Maven POM files.
See: https://maven.apache.org/developers/conventions/code.html

The POM files has been "fixed" by using the "sortpom-maven-plugin".
The blank lines has been kept to keep the element separation for
readability.
The plugin also fixes indentation etc.
Have a look at: https://github.com/Ekryd/sortpom/wiki

Command that has been executed:
```sh
mvn com.github.ekryd.sortpom:sortpom-maven-plugin:sort -Dsort.keepBlankLines=true
```

PR #5657, opened 2018-05-30
POM Code Convention

maggu2810:
I realized that there is at least one pom.xml file that uses tabs
instead of spaces for indentation. So I started the POM cleanup again.

In #5338 I missed to set the predefined sort order profile.
IMHO we should use the "recommended_2008_06" profile that states:
The POM Code Convention that was chosen by Maven developers in 2008

The main difference to the current one and the reason so much files are
touched is that the order for GAV has been "corrected".
In the default profile the artifactId is above the groupId (which does
not make sense IMHO at all).

Sorry for not using the correct profile the first time.

Command that has been executed:

    mvn \
      com.github.ekryd.sortpom:sortpom-maven-plugin:sort \
      -Dsort.keepBlankLines=true \
      -Dsort.predefinedSortOrder=recommended_2008_06


maggu2810:
Should I split the PR into separate commits, one that fixes the indentation of the POM by not forcing the usage of "recommended_2008_06" and another one that only introduce the different sort order?

sjka:
I think it is fine as it is.

VelinYordanov:
Should we use spaces for indentation only for pom.xml files or for all .xml files?

maggu2810:
I don't really care which indentation style is used.
But I don't see any reason to change all our XML files and would simple keep the tabs as it is for all other XML files.

IIRC there has been some comments some time ago, that 1 tab consumes less bytes then e.g. 4 spaces and this could be considered for e.g. HTML files, ...

PR #5658, opened 2018-05-30
Created a builder like structure to improve the code using the HttpUtil

martinvw:
An initial setup of the builders which I proposed earlier, I'm currently working on the tests, but the rest could maybe be ready for a first version.

See also the issue https://github.com/eclipse/smarthome/issues/4257


martinvw:
@triller-telekom with all those static methods its a small hell to test it, do you have an opinion about testing, I can either drag in wiremock, but it seems that we avoided that till now. 

Or the alternatives are hooks specific for test code (and reducing the usefulness of the test) or no tests, or ...

martinvw:
I get the warning:

```
[WARNING] org.eclipse.smarthome.io.net.http.HttpClientInitializationException.java:[18]
First javadoc author should have "Initial contribution" contribution description.
```

But all other files in this part of the code have either:

- 'initial contribution and API'
- or 'initial API'

triller-telekom:
Regarding the testing: Yes indeed it is difficult because you cannot easily mock the static methods with the plain mockito that we use so far...

The only possibility i see is the use of reflection. You could prepare the `HttpRequestBuilder` and afterwards access its internal fields.

I mean in the end you make use of the `HttpUtil.executeUrl` which isn't covered by unit tests either (except for one method that doesn't do a connection request though)...

The warning that you get is from a new version of the static code analyser, but I do not know where the rule with the required "initial contribution" on the first author comes from. if you change your line from "initial API" to "initial contribution" the warning should disappear though.

martinvw:
Would wiremock be accepted / desirable, I think it would be the only good way to do this. And I have the experience and the will to go throug the effort to test it with this.

triller-telekom:
I just had a look at wiremock and they claim to be a "a simulator for HTTP-based APIs". We do not want to test the underlying http request, because this is handled in the jetty library. What we rather want to test is if we feed the correct data into the library.

In theory we could integrate a mocking framework such as powermock that would allow for mocking static calls, but this might encourage people to create static methods which is a bad style of coding...

So my opinion is that since our `HTTPUtil` is static you can test your builder via reflection, i.e. whether the private properties are initialized correctly by the builder methods. But that is just my opinion on this.

bundles/io/org.eclipse.smarthome.io.net/src/main/java/org/eclipse/smarthome/io/net/http/HttpRequestBuilder.java
```diff
@@ -0,0 +1,127 @@
...
+    }
+
+    /**
+     * Add a timeout for this request
+     *
+     * @param amount pas the value of timeout in the unit which is applied as the second parameter
+     * @param unit a TimeUnit eg. TimeUnit.SECONDS
+     * @return a request builder to construct and complete the request
+     */
+    public HttpRequestBuilder withTimeout(int amount, TimeUnit unit) {
```
amitjoy: Instead of amount and time unit, we could provide a method with `Duration`.

```java
public HttpRequestBuilder withTimeout(Duration duration) {
...
}
```

bundles/io/org.eclipse.smarthome.io.net/src/main/java/org/eclipse/smarthome/io/net/http/HttpUtil.java
```diff
@@ -324,13 +330,13 @@ private static boolean shouldUseProxy(String urlString, String nonProxyHosts) {
      *             <code>POST</POST> or <code>DELETE</code>
      */
     public static HttpMethod createHttpMethod(String httpMethodString) {
-        if ("GET".equals(httpMethodString)) {
```
amitjoy: Just a proposal:

```java
import static org.eclipse.jetty.http.HttpMethod.*;
...
...
public static HttpMethod createHttpMethod(String httpMethodString) {
    Predicate<HttpMethod> isMethodAllowed = m -> m == GET || m == POST || m == PUT || m == DELETE;
    return Optional.ofNullable(HttpMethod.fromString(httpMethodString))
        .filter(isMethodAllowed)
        .orElseThrow(new IllegalArgumentException("Given HTTP Method '" + httpMethodString + "' is unknown"));
}
```

bundles/io/org.eclipse.smarthome.io.net/src/main/java/org/eclipse/smarthome/io/net/http/HttpUtil.java
```diff
@@ -47,26 +47,32 @@
...
+    static final String METHOD_POST = "POST";
+    static final String METHOD_DELETE = "DELETE";
+
     private static class ProxyParams {
-        public String proxyHost = null;
-        public int proxyPort = 80;
-        public String proxyUser = null;
-        public String proxyPassword = null;
-        public String nonProxyHosts = null;
+        String proxyHost = null;
```
amitjoy: the nulls are default

bundles/io/org.eclipse.smarthome.io.net/src/main/java/org/eclipse/smarthome/io/net/http/HttpUtil.java
```diff
@@ -47,26 +47,32 @@
 import org.slf4j.LoggerFactory;
 
 /**
- * Some common methods to be used in both HTTP-In-Binding and HTTP-Out-Binding
+ * Some common methods to be used in HTTP-In-Binding, HTTP-Out-Binding and other bindings
+ *
+ * For advanced usage direct use of the Jetty client is prefered
```
amitjoy: preferred

bundles/io/org.eclipse.smarthome.io.net/src/main/java/org/eclipse/smarthome/io/net/http/HttpRequestBuilder.java
```diff
@@ -0,0 +1,127 @@
...
+import java.util.Properties;
+import java.util.concurrent.TimeUnit;
+
+/**
+ * Builder class to construct http requests
+ *
+ * @author Martin van Wingerden - initial API
+ */
+public class HttpRequestBuilder {
+    private static final int DEFAULT_TIMEOUT_IN_MS = 5_000;
```
amitjoy: You can also provide it using `Duration`:

```java
private static final Duration DEFAULT_TIMEOUT = Duration.ofSeconds(5);
```

bundles/io/org.eclipse.smarthome.io.net/src/main/java/org/eclipse/smarthome/io/net/http/HttpRequestBuilder.java
```diff
@@ -0,0 +1,127 @@
...
+    private Properties headers;
+    private String contentType;
+
+    /**
+     * Private constructor, to hide away the construction behind some factory methods
+     *
+     * @param method Http method to be used
+     * @param url url to fetch or post data to
+     */
+    private HttpRequestBuilder(String method, String url) {
```
amitjoy: The invariants:

```java
import static java.util.Objects.requireNonNull;
....
....
requireNonNull(method, "HTTP Method cannot be null");
requireNonNull(url, "URL cannot be null");
```

If you use JDT annotations, you would need to annotate the static methods.

bundles/io/org.eclipse.smarthome.io.net/src/main/java/org/eclipse/smarthome/io/net/http/HttpRequestBuilder.java
```diff
@@ -0,0 +1,127 @@
...
+        return new HttpRequestBuilder(HttpUtil.METHOD_POST, url);
+    }
+
+    /**
+     * Add content to this request
+     *
+     * @param content a string containing the data to be pushed to the url
+     * @return a request builder to construct and complete the request
+     */
+    public HttpRequestBuilder withContent(String content) {
```
amitjoy: The invariants (`Objects#requireNonNull(..)` or JDT Annotations)

bundles/io/org.eclipse.smarthome.io.net/src/main/java/org/eclipse/smarthome/io/net/http/HttpRequestBuilder.java
```diff
@@ -0,0 +1,127 @@
...
+    }
+
+    /**
+     * Add an additional header to the request
+     *
+     * @param header name of the header, eg. Content-Type
+     * @param value value of the header, eg. "application/json"
+     * @return a request builder to construct and complete the request
+     */
+    public HttpRequestBuilder withHeader(String header, String value) {
```
amitjoy: The invariants (`Objects#requireNonNull(..)` or JDT Annotations)

bundles/io/org.eclipse.smarthome.io.net/src/main/java/org/eclipse/smarthome/io/net/http/HttpRequestBuilder.java
```diff
@@ -0,0 +1,127 @@
...
+        return this;
+    }
+
+    /**
+     * Add a content type for this request
+     *
+     * @param contentType a string containing the data to be pushed to the url
+     * @return a request builder to construct and complete the request
+     */
+    public HttpRequestBuilder withContentType(String contentType) {
```
amitjoy: The invariants (`Objects#requireNonNull(..)` or JDT Annotations)

martinvw:
@amitjoy your comments are greatly appreciated but I'm pretty busy at the moment and personally I really think that the code should be tested properly and that reflection is in 99% of the cases not the way to test (this / anything). So I still think that wiremock makes the most sense here.

So I will be back but it might take a few weeks 😏

martinvw:
> So I will be back but it might take a few weeks 😏

Or sooner, I processed the review comments and added an initial wiremock test, to give you some taste I'm in doubt whether this is the way to add (such) a dependency but please let me know.

martinvw:
And the tests worked until I rebased 😢 , is there a proper way to have the HTTP Client Factory available or should I leverage this to use a fake http client factory to inject some mock.

I still think that WireMock will give us more elegant tests, but with HTTP Client Factory these costs might just have increased.

bundles/io/org.eclipse.smarthome.io.net.test/src/test/java/org/eclipse/smarthome/io/net/http/BaseHttpUtilTest.java
```diff
@@ -0,0 +1,51 @@
+package org.eclipse.smarthome.io.net.http;
```
martinvw: Copyright header is missing

bundles/io/org.eclipse.smarthome.io.net.test/src/test/java/org/eclipse/smarthome/io/net/http/BaseHttpUtilTest.java
```diff
@@ -0,0 +1,51 @@
...
+import java.util.concurrent.TimeUnit;
+
+import org.eclipse.jetty.client.HttpClient;
+import org.eclipse.jetty.client.api.ContentResponse;
+import org.eclipse.jetty.client.api.Request;
+import org.eclipse.jetty.http.HttpMethod;
+import org.junit.Before;
+import org.mockito.Mock;
+
+public class BaseHttpUtilTest {
```
martinvw: Docblock header

bundles/io/org.eclipse.smarthome.io.net/src/main/java/org/eclipse/smarthome/io/net/http/HttpRequestBuilder.java
martinvw: I added the annotations

bundles/io/org.eclipse.smarthome.io.net/src/main/java/org/eclipse/smarthome/io/net/http/HttpRequestBuilder.java
```diff
@@ -0,0 +1,143 @@
...
+        return new HttpRequestBuilder("GET", url);
+    }
+
+    /**
+     * Construct an http request builder to post data to a <code>url</code>
+     *
+     * @param url to post data to
+     * @return a request builder to construct and complete the request
+     */
+    public static HttpRequestBuilder postTo(String url) {
```
martinvw: Misses test coverage

bundles/io/org.eclipse.smarthome.io.net/src/main/java/org/eclipse/smarthome/io/net/http/HttpRequestBuilder.java
```diff
@@ -0,0 +1,143 @@
...
+    }
+
+    /**
+     * Add content to this request
+     *
+     * @param content a string containing the data to be pushed to the <code>url</code>
+     * @return a request builder to construct and complete the request
+     */
+    public HttpRequestBuilder withContent(String content) {
+        this.inputStream = new ByteArrayInputStream(content.getBytes(StandardCharsets.UTF_8));
```
martinvw: Misses coverage

bundles/io/org.eclipse.smarthome.io.net/src/main/java/org/eclipse/smarthome/io/net/http/HttpRequestBuilder.java
```diff
@@ -0,0 +1,143 @@
...
+
+    /**
+     * Add content with a specific type to this request
+     *
+     * @param content a string containing the data to be pushed to the <code>url</code>
+     * @param contentType the content type of the given <code>content</code>
+     * @return a request builder to construct and complete the request
+     */
+    public HttpRequestBuilder withContent(String content, String contentType) {
+        withContent(content);
```
martinvw: Misses coverage

bundles/io/org.eclipse.smarthome.io.net/src/main/java/org/eclipse/smarthome/io/net/http/HttpRequestBuilder.java
martinvw: Done

wborn:
Is this PR going to be a full replacement for the methods in `HttpUtil` @martinvw? If so, you might want to deprecate those and instruct people to use the builder.

martinvw:
> Is this PR going to be a full replacement for the methods in HttpUtil

No because those methods are partly intended for the rules engine, see also the linked issue, a quote from kai from there:

> My concern above merely want to warn you that HttpUtil is used within the rules and thus all static methods are implicitly made available there, so we must avoid adding anything that leads to confusion in rules.
>
>You might therefore rather want to introduce a new class like HttpRequestBuilder, which assembles an HttpRequest instance, which could then be passed to a new method HttpUtil.execute(HttpRequest) or something along these lines.

martinvw:
@amitjoy @triller-telekom the tests should be complete now and all review comments processed, I also squashed an rebased.

If you need anything else just let me know.

bundles/io/org.eclipse.smarthome.io.net/src/main/java/org/eclipse/smarthome/io/net/http/HttpRequestBuilder.java
martinvw: done

bundles/io/org.eclipse.smarthome.io.net/src/main/java/org/eclipse/smarthome/io/net/http/HttpRequestBuilder.java
martinvw: done

bundles/io/org.eclipse.smarthome.io.net/src/main/java/org/eclipse/smarthome/io/net/http/HttpRequestBuilder.java
martinvw: done

bundles/io/org.eclipse.smarthome.io.net.test/src/test/java/org/eclipse/smarthome/io/net/http/BaseHttpUtilTest.java
martinvw: done

amitjoy:
Looks good to me 👍

bundles/io/org.eclipse.smarthome.io.net.test/src/test/java/org/eclipse/smarthome/io/net/http/HttpUtilTest.java
```diff
@@ -30,4 +69,19 @@ public void testCreateHttpMethod() {
...
+    }
+
+    @Test
+    public void testCreateHttpMethodForUnsupportedTrace() {
+        thrown.expect(IllegalArgumentException.class);
+        thrown.expectMessage("Given HTTP Method 'TRACE' is unknown");
+
+        HttpUtil.createHttpMethod("TRACE");
+    }
+}
```
triller-telekom: newline missing

bundles/io/org.eclipse.smarthome.io.net.test/src/test/java/org/eclipse/smarthome/io/net/http/HttpUtilTest.java
```diff
@@ -13,14 +13,53 @@
...
+import org.junit.Rule;
 import org.junit.Test;
+import org.junit.rules.ExpectedException;
 
 /**
- * @author Thomas Eichstaedt-Engelen
+ * Tests for the HttpUtil
+ *
+ * @author Thomas Eichstaedt-Engelen - Initial contribution
+ * @author Martin van Wingerden - Added tests bases on HttpClientFactory
```
triller-telekom: Typo: base**d** on ...

bundles/io/org.eclipse.smarthome.io.net.test/pom.xml
```diff
@@ -15,23 +15,96 @@
...
+                <versionRange>0.0.0</versionRange>
+              </requirement>
+              <requirement>
+                <type>eclipse-plugin</type>
+                <id>org.eclipse.smarthome.core.thing</id>
+                <versionRange>0.0.0</versionRange>
+              </requirement>
+              <requirement>
+                <type>eclipse-plugin</type>
+                <id>org.eclipse.smarthome.core.voice</id>
```
triller-telekom: Are you sure that you need the voice, audio, etc. bundles as a dependency here?

bundles/io/org.eclipse.smarthome.io.net.test/pom.xml
```diff
@@ -15,23 +15,96 @@
...
+              <id>org.eclipse.smarthome.core.voice</id>
+              <level>4</level>
+              <autoStart>true</autoStart>
+            </bundle>
+            <bundle>
+              <id>org.eclipse.smarthome.core.thing</id>
+              <level>4</level>
+              <autoStart>true</autoStart>
+            </bundle>
+
```
triller-telekom: Please remove the blank line.

bundles/io/org.eclipse.smarthome.io.net.test/src/test/java/org/eclipse/smarthome/io/net/http/BaseHttpUtilTest.java
```diff
@@ -0,0 +1,68 @@
...
+import org.eclipse.jetty.client.api.ContentResponse;
+import org.eclipse.jetty.client.api.Request;
+import org.eclipse.jetty.http.HttpMethod;
+import org.junit.Before;
+import org.mockito.Mock;
+
+/**
+ * Base class for tests for the <code>HttpRequestBuilder</code> and <code>HttpUtil</code> to validate their behaviour
+ *
+ * @author Martin van Wingerden & Wouter Born - Initial contribution *
```
triller-telekom: There is a "*" too much at the end of this line

bundles/io/org.eclipse.smarthome.io.net/src/main/java/org/eclipse/smarthome/io/net/http/HttpRequestBuilder.java
```diff
@@ -0,0 +1,143 @@
...
+
+    /**
+     * Executes the build request
+     *
+     * @return the response body or <code>null</code> when the request went wrong
+     * @throws IOException when the request execution failed, timed out or it was interrupted
+     */
+    public String getContentAsString() throws IOException {
+        // this cast should be okay, a requested timeout should normally not exceed a long
+        int timeoutMillis = (int) timeout.toMillis();
```
triller-telekom: Why not using an `int` or `long` instead of `Duration` for the timeout if you need an `int` here?

bundles/io/org.eclipse.smarthome.io.net.test/pom.xml
martinvw: No, I have no idea why all other test bundles contain these, what I noticed was that it started all kinds unexpected processes like a plug wise discovery and I did not have this so I though I had to add it 😐

bundles/io/org.eclipse.smarthome.io.net/src/main/java/org/eclipse/smarthome/io/net/http/HttpRequestBuilder.java
martinvw: Because a long without any unit misses a lot of expression power, and there are a lot cases where it would make more sense to express it in seconds or in strange corner cases maybe in minutes

martinvw:
I updated the PR

martinvw:
Locally I'm having some problem retrieving dependencies, it reports maven.eclipse.org down, see also https://downforeveryoneorjustme.com/maven.eclipse.org

martinvw:
> Locally I'm having some problem retrieving dependencies, it reports maven.eclipse.org down, see also https://downforeveryoneorjustme.com/maven.eclipse.org

But from maven directly from the command line its fine 😕 

And travis is also happy, but the ip-validation not, but I do not see a reason why?

wborn:
Why sign off with 2 Martin e-mails? 

```
Also-by: Wouter Born <eclipse@maindrain.net>
Signed-off-by: Martin van Wingerden <martinvw@mtin.nl>
Signed-off-by: Martin van Wingerden <martin@martinvw.nl>
```

martinvw:
> Why sign off with 2 Martin e-mails?

Because of a check mark in sourcetree, let me fix that

martinvw:
Resolved the double sign-off

wborn:
Did your ECA expire? The ECA validation tool on https://accounts.eclipse.org/ reports "User not found" when I enter your email.

wborn:
martin@martinvw.nl is validated OK

martinvw:
I changed my email for my ECA to https://accounts.eclipse.org/search/profile/martin%40martinvw.nl

wborn:
So now **ip-validation** is complaining about an e-mail that's not even in the commit message. 
Perhaps it doesn't handle force pushes well. 
Maybe you can make a micro improvement so it is re-triggered. ;-)

martinvw:
No the problem was that after squashing the author of the commit still pointed to my old ECA email

bundles/io/org.eclipse.smarthome.io.net.test/pom.xml
```diff
@@ -15,17 +15,12 @@
...
-    <bundle.namespace>org.eclipse.smarthome.io.net.test</bundle.namespace>
-  </properties>
-
   <build>
     <plugins>
       <plugin>
         <groupId>${tycho-groupid}</groupId>
         <artifactId>target-platform-configuration</artifactId>
-        <configuration>
+        <configuration>
```
triller-telekom: A few spaces too much at the end of this line.

martinvw: I let Eclipse format the complete file

martinvw: Is the Eclipse configuration now broken, because it replaced spaces by tabs for this xml file :(

martinvw: Okay, resolved that as well, but we should fix the eclipse settings, maybe its because I (accidentally) upgraded to photon :-)

bundles/io/org.eclipse.smarthome.io.net/src/main/java/org/eclipse/smarthome/io/net/http/HttpRequestBuilder.java
```diff
@@ -0,0 +1,143 @@
...
+ * @author Martin van Wingerden - Initial contribution
+ */
+@NonNullByDefault
+public class HttpRequestBuilder {
+    private static final Duration DEFAULT_TIMEOUT = Duration.ofSeconds(5);;
+
+    private final String method;
+    private final String url;
+
+    @Nullable
```
htreu: We have a new policy, to add the null annotations in front of the type, like `private @Nullable InputStream inputStream;`. See https://github.com/eclipse/smarthome/pull/5839

bundles/io/org.eclipse.smarthome.io.net/src/main/java/org/eclipse/smarthome/io/net/http/HttpUtil.java
```diff
@@ -48,27 +51,28 @@
...
+ * For advanced usage direct use of the Jetty client is preferred
  *
  * @author Thomas Eichstaedt-Engelen
  * @author Kai Kreuzer - Initial contribution and API
  * @author Svilen Valkanov - replaced Apache HttpClient with Jetty
  */
 @Component(immediate = true)
 public class HttpUtil {
-
     private static Logger logger = LoggerFactory.getLogger(HttpUtil.class);
```
htreu: this one could be final.

martinvw: And is it then a `LOGGER`?

bundles/io/org.eclipse.smarthome.io.net/src/main/java/org/eclipse/smarthome/io/net/http/HttpUtil.java
```diff
@@ -61,7 +61,7 @@
  */
 @Component(immediate = true)
 public class HttpUtil {
-    private static Logger logger = LoggerFactory.getLogger(HttpUtil.class);
+    private static final Logger LOGGER = LoggerFactory.getLogger(HttpUtil.class);
```
htreu: actually we have `logger` all over the place.

htreu: is this documented somewhere?

martinvw: @kaikreuzer and I used to agree that if we use a static logger then its a `private static final LOGGER` but that is maybe an openHAB rule :-)

htreu: ah, good point. Its getting much too warm in here... I agree :-)

martinvw:
Thanks

Issue #4942, opened 2018-01-18
Feature request: Populate Selection elements in sitemaps automatically


cweitkamp:
In Paper UI you automatically have a select menu when the channel-type definition contains an `options` list. e.g.
```
<channel-type id="systemcommand" advanced="true">
	<item-type>String</item-type>
	<label>Send system command</label>
	<description>Sends a system command to Kodi. This allows to shutdown/suspend/hibernate/reboot Kodi</description>
	<state>
		<options>
			<option value="Shutdown">Shutdown</option>
			<option value="Suspend">Suspend</option>
			<option value="Hibernate">Hibernate</option>
			<option value="Reboot">Reboot</option>
		</options>
	</state>
</channel-type>
```
![unbenannt](https://user-images.githubusercontent.com/5131747/35121566-23724e26-fc9c-11e7-8644-735e3b4537eb.png)

This is not possible in a sitemap for Basic UI without using a mapping parameter. A mapping parameter is static which makes it impossible to use dynamically created channels in a sufficient way or handle manipulated state descriptions during runtime.
```
String kodiLivingRoomSystemCommand { channel="kodi:kodi:XXXXXXXX:systemcommand" }
```
```
sitemap demo label="Kodi Media Center" {
    Frame label="Wohnzimmer (System)" {
        Default item=kodiLivingRoomSystemCommand
        Selection item=kodiLivingRoomSystemCommand
        Selection item=kodiLivingRoomSystemCommand mappings=[Shutdown="Herunterfahren", Suspend="Bereitschaft", Reboot="Neustart"]
    }
}
```

kaikreuzer:
I agree with @lolodomo - this needs to be a feature of the sitemaps, which could dynamically provide the mapping information, which would make this available in Android, iOS, etc., too.

cweitkamp:
Absolutely. Even better if it's available in all UIs.

mgeramb:
I want use a DynamicStateDescriptionProvider in my amazonechoncontrol binding for selecting the bluetooth device and have the same problem that it is useless in sitemaps. If this can be don easily in the sitemap implementation, it would be nice.

mhilbush:
It also would be nice if the mappings could be populated from an Item.

[...]

PR #5666, opened 2018-06-02
[Classic UI] Consider item options for selection widget when no mapping is defined

lolodomo:
Relative to issue #4942

### DIFF
--- a/extensions/ui/org.eclipse.smarthome.ui.classic/src/main/java/org/eclipse/smarthome/ui/classic/internal/render/SelectionRenderer.java
+++ b/extensions/ui/org.eclipse.smarthome.ui.classic/src/main/java/org/eclipse/smarthome/ui/classic/internal/render/SelectionRenderer.java
@@ -20,6 +20,7 @@ import org.eclipse.smarthome.core.items.ItemNotFoundException;
 import org.eclipse.smarthome.core.library.items.NumberItem;
 import org.eclipse.smarthome.core.library.types.QuantityType;
 import org.eclipse.smarthome.core.types.State;
+import org.eclipse.smarthome.core.types.StateOption;
 import org.eclipse.smarthome.core.types.util.UnitUtils;
 import org.eclipse.smarthome.model.sitemap.Mapping;
 import org.eclipse.smarthome.model.sitemap.Selection;
@@ -60,6 +61,7 @@ public class SelectionRenderer extends AbstractWidgetRenderer {
         State state = itemUIRegistry.getState(w);
         Selection selection = (Selection) w;
         String mappedValue = "";
+        String rowMappedValue;
 
         Item item = null;
         try {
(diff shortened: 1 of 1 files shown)


lolodomo:
@cweitkamp : can you please review and approve this one too ? It is the same principle as for Basic UI.

extensions/ui/org.eclipse.smarthome.ui.classic/src/main/java/org/eclipse/smarthome/ui/classic/internal/render/SelectionRenderer.java
```diff
@@ -69,40 +72,16 @@ public boolean canRender(Widget w) {
...
-            } else {
-                rowSnippet = StringUtils.replace(rowSnippet, "%checked%", "");
+        } else {
+            for (Mapping mapping : selection.getMappings()) {
+                String rowSnippet = buildRow(selection, mapping.getLabel(), mapping.getCmd(), item, state);
+                rowSB.append(rowSnippet);
             }
-            rowSB.append(rowSnippet);
         }
         snippet = StringUtils.replace(snippet, "%label_header%", getLabel(w, mappedValue));
```
cweitkamp: Only a question for my personal understanding: What is the difference between `getLabel(w, mappedValue)` and `getLabel(w)`?
See https://github.com/eclipse/smarthome/pull/5667/files#diff-955d3a6399e421b68151ddb7351de8bbR107

extensions/ui/org.eclipse.smarthome.ui.classic/src/main/java/org/eclipse/smarthome/ui/classic/internal/render/SelectionRenderer.java
```diff
@@ -114,6 +93,42 @@ public boolean canRender(Widget w) {
...
+
+        State compareMappingState = state;
+        if (state instanceof QuantityType) { // convert the item state to the command value for proper
+                                             // comparison and "checked" attribute calculation
+            compareMappingState = convertStateToLabelUnit((QuantityType<?>) state, command);
+        }
+
+        if (compareMappingState.toString().equals(command)) {
+            rowSnippet = StringUtils.replace(rowSnippet, "%checked%", "checked=\"true\"");
+            mappedValue = (label != null) ? label : command;
```
cweitkamp: Why do we assign the `command` here if the `label` is `null`? We don't check for that in Basic UI.
See https://github.com/eclipse/smarthome/pull/5667/files#diff-955d3a6399e421b68151ddb7351de8bbR144

extensions/ui/org.eclipse.smarthome.ui.classic/src/main/java/org/eclipse/smarthome/ui/classic/internal/render/SelectionRenderer.java
```diff
@@ -44,6 +45,8 @@
 
     private final Logger logger = LoggerFactory.getLogger(SelectionRenderer.class);
 
+    private String mappedValue;
```
kaikreuzer: Do you really want that as an instance field? Isn't that an issue if there are concurrent requests that render different widgets with mappings at the very same moment?

extensions/ui/org.eclipse.smarthome.ui.classic/src/main/java/org/eclipse/smarthome/ui/classic/internal/render/SelectionRenderer.java
lolodomo: A label for a StateOption can be null, not its value.
For the mappings in the sitemap, the syntax forces you to define a label and a value.
So to be clear, this test was useless before my change but is now required.
So it is now ok here but I have to change Basic UI.

extensions/ui/org.eclipse.smarthome.ui.classic/src/main/java/org/eclipse/smarthome/ui/classic/internal/render/SelectionRenderer.java
lolodomo: That was an easy way as I have now a method that should return 2 data ... but not clean.
I will try something else.

lolodomo:
New try without mappedValue as instance field.

PR #5667, opened 2018-06-02
Basic UI: consider item options for selection widget when no mapping is defined

lolodomo:
Relative to issue #4942

### DIFF
--- a/extensions/ui/org.eclipse.smarthome.ui.basic/src/main/java/org/eclipse/smarthome/ui/basic/internal/render/SelectionRenderer.java
+++ b/extensions/ui/org.eclipse.smarthome.ui.basic/src/main/java/org/eclipse/smarthome/ui/basic/internal/render/SelectionRenderer.java
@@ -20,6 +20,7 @@ import org.eclipse.smarthome.core.items.ItemNotFoundException;
 import org.eclipse.smarthome.core.library.items.NumberItem;
 import org.eclipse.smarthome.core.library.types.QuantityType;
 import org.eclipse.smarthome.core.types.State;
+import org.eclipse.smarthome.core.types.StateOption;
 import org.eclipse.smarthome.core.types.util.UnitUtils;
 import org.eclipse.smarthome.model.sitemap.Mapping;
 import org.eclipse.smarthome.model.sitemap.Selection;
@@ -67,32 +68,16 @@ public class SelectionRenderer extends AbstractWidgetRenderer {
         return w instanceof Selection;
     }
 
-    /**
-     * Get command-label map for a Selection widget
-     *
-     * @return String representing JSON object
-     */
-    private String getMappingsJSON(Selection w) {
-        JsonObject resultObject = new JsonObject();
-        for (Mapping mapping : w.getMappings()) {
-            resultObject.addProperty(mapping.getCmd(), mapping.getLabel());
-        }
-        String result = resultObject.toString();
-        result = StringEscapeUtils.escapeHtml(result);
-        return result;
-    }
-
     @Override
     public EList<Widget> renderWidget(Widget w, StringBuilder sb) throws RenderException {
         String snippet = getSnippet("selection");
 
         snippet = preprocessSnippet(snippet, w);
-        snippet = StringUtils.replace(snippet, "%value_map%", getMappingsJSON((Selection) w));
-        snippet = StringUtils.replace(snippet, "%label_header%", getLabel(w));
 
         State state = itemUIRegistry.getState(w);
         Selection selection = (Selection) w;
         String mappingLabel = null;
+        String rowMappingLabel;
 
         Item item = null;
         try {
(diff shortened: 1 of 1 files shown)


lolodomo:
Please note that the list of values (options) for the selection is built when the page is opened and then not refreshed until the page is loaded again.
Of course, the selected value is updated in case of an item update.

extensions/ui/org.eclipse.smarthome.ui.basic/src/main/java/org/eclipse/smarthome/ui/basic/internal/render/SelectionRenderer.java
```diff
@@ -72,11 +75,18 @@ public boolean canRender(Widget w) {
      *
      * @return String representing JSON object
      */
-    private String getMappingsJSON(Selection w) {
+    private String getMappingsJSON(Selection w, Item item) {
         JsonObject resultObject = new JsonObject();
-        for (Mapping mapping : w.getMappings()) {
-            resultObject.addProperty(mapping.getCmd(), mapping.getLabel());
+        if (w.getMappings().size() == 0 && item != null && item.getStateDescription() != null) {
+            for (StateOption option : item.getStateDescription().getOptions()) {
```
cweitkamp: It looks like we iterate over all state options or mappings twice. One time in this method (getMappingsJSON) the second time in the renderWidget method. Do you think it is possible to speed up the process by eliminating one loop? E.g. is the getMappingsJSON method imho not really necessary.

lolodomo: Yes, that is possible of course.

lolodomo:
New version to avoid declaring mappingLabel as instance attribute.

extensions/ui/org.eclipse.smarthome.ui.basic/src/main/java/org/eclipse/smarthome/ui/basic/internal/render/SelectionRenderer.java
```diff
@@ -101,42 +86,28 @@ private String getMappingsJSON(Selection w) {
...
+                jsonObject.addProperty(mapping.getCmd(), mapping.getLabel());
+                rowMappingLabel = buildRow(selection, mapping.getLabel(), mapping.getCmd(), item, state, rowSB);
+                if (rowMappingLabel != null) {
+                    mappingLabel = rowMappingLabel;
+                }
             }
-            rowSB.append(rowSnippet);
         }
         snippet = StringUtils.replace(snippet, "%rows%", rowSB.toString());
+        snippet = StringUtils.replace(snippet, "%value_map%", StringEscapeUtils.escapeHtml(jsonObject.toString()));
```
kaikreuzer: I you'd continue to use the inherited `escapeHtml`, it would be easier to refactor, if we want to replace apache commons one day.

extensions/ui/org.eclipse.smarthome.ui.basic/src/main/java/org/eclipse/smarthome/ui/basic/internal/render/SelectionRenderer.java
```diff
@@ -146,6 +117,46 @@ private String getMappingsJSON(Selection w) {
...
+        String label = lab;
+
+        if (item instanceof NumberItem && ((NumberItem) item).getDimension() != null) {
+            String unit = getUnitForWidget(w);
+            command = StringUtils.replace(command, UnitUtils.UNIT_PLACEHOLDER, unit);
+            label = StringUtils.replace(label, UnitUtils.UNIT_PLACEHOLDER, unit);
+
+            // Special treatment for °C since uom library uses a single character: ℃
+            // This will ensure the current state matches the cmd and the buttonClass is set accordingly.
+            command = StringUtils.replace(command, "°C", "℃");
```
kaikreuzer: Hm, this might break through https://github.com/eclipse/smarthome/issues/5669 - it would probably better to replace the special char and always compare with "°C" here (same for the Classic UI, I noticed that too late).

extensions/ui/org.eclipse.smarthome.ui.basic/src/main/java/org/eclipse/smarthome/ui/basic/internal/render/SelectionRenderer.java
lolodomo: I changed nothing, the code was just moved from the method I suppressed (as requested in another review comment).
I don't know if there was a valid reason to use antoher function.

extensions/ui/org.eclipse.smarthome.ui.basic/src/main/java/org/eclipse/smarthome/ui/basic/internal/render/SelectionRenderer.java
lolodomo: One time again, this is only a move of the existing code. My PR is not fixing something relative to UoM. Maybe the fix should be applied globally in another PR if there is a problem with °Cellsius unit ?
If you prefer, I can rebase my PR after this problem is globally fixed ?

kaikreuzer: Ok, I didn't see that this was indeed in there before, sorry.
So let's keep it here, we just have to make sure to remember to address those places once #5669 is addressed (cc @htreu).

extensions/ui/org.eclipse.smarthome.ui.basic/src/main/java/org/eclipse/smarthome/ui/basic/internal/render/SelectionRenderer.java
kaikreuzer: Ok, it isn't identical to the line before the move, but you are right that this class was used before already. So nothing that must be fixed in the scope of this PR.

lolodomo:
Thank you.

PR #5676, opened 2018-06-04
Typo of Hektopascal

vzorglub:
Should be Hectopascal


kaikreuzer:
Merging as small patch exception without IP validation.

Issue #5660, opened 2018-06-01
IllegalArgumentException in AutomaticInboxProcessor


amitjoy:
```java
java.lang.IllegalArgumentException: The argument 'discoveryResult' must not be null.
	at org.eclipse.smarthome.core.events.AbstractEventFactory.checkNotNull(AbstractEventFactory.java:112)
	at org.eclipse.smarthome.config.discovery.inbox.events.InboxEventFactory.assertValidArgument(InboxEventFactory.java:120)
	at org.eclipse.smarthome.config.discovery.inbox.events.InboxEventFactory.createUpdatedEvent(InboxEventFactory.java:112)
	at org.eclipse.smarthome.config.discovery.internal.PersistentInbox.postEvent(PersistentInbox.java:476)
	at org.eclipse.smarthome.config.discovery.internal.PersistentInbox.notifyListeners(PersistentInbox.java:462)
	at org.eclipse.smarthome.config.discovery.internal.PersistentInbox.setFlag(PersistentInbox.java:381)
	at org.eclipse.smarthome.config.discovery.internal.AutomaticInboxProcessor.ignoreInInbox(AutomaticInboxProcessor.java:186)
	at org.eclipse.smarthome.config.discovery.internal.AutomaticInboxProcessor.checkAndIgnoreInInbox(AutomaticInboxProcessor.java:176)
	at org.eclipse.smarthome.config.discovery.internal.AutomaticInboxProcessor.autoIgnore(AutomaticInboxProcessor.java:168)
	at org.eclipse.smarthome.config.discovery.internal.AutomaticInboxProcessor.receiveTypedEvent(AutomaticInboxProcessor.java:115)
	at org.eclipse.smarthome.config.discovery.internal.AutomaticInboxProcessor.receiveTypedEvent(AutomaticInboxProcessor.java:1)
	at org.eclipse.smarthome.core.events.AbstractTypedEventSubscriber.receive(AbstractTypedEventSubscriber.java:54)
	at sun.reflect.GeneratedMethodAccessor11.invoke(Unknown Source)
	at sun.reflect.DelegatingMethodAccessorImpl.invoke(Unknown Source)
	at java.lang.reflect.Method.invoke(Unknown Source)
	at org.eclipse.smarthome.core.internal.common.AbstractInvocationHandler.invokeDirect(AbstractInvocationHandler.java:153)
	at org.eclipse.smarthome.core.internal.common.Invocation.call(Invocation.java:53)
[...]

PR #5677, opened 2018-06-04
improve nullness handling on PersistentInbox

maggu2810:
* Add NonNullByDefault to ThingRegistryChangeListener
* Add NonNullByDefault to PersistentInbox
* Add nullness checks where necessary (AFAIK)

Fixes: https://github.com/eclipse/smarthome/issues/5660

### DIFF
--- a/bundles/config/org.eclipse.smarthome.config.discovery/src/main/java/org/eclipse/smarthome/config/discovery/internal/PersistentInbox.java
+++ b/bundles/config/org.eclipse.smarthome.config.discovery/src/main/java/org/eclipse/smarthome/config/discovery/internal/PersistentInbox.java
@@ -176,11 +176,11 @@ public final class PersistentInbox implements Inbox, DiscoveryListener, ThingReg
     }
 
     @Override
-    public synchronized boolean add(final @Nullable DiscoveryResult rslt) throws IllegalStateException {
-        if (rslt == null) {
+    public synchronized boolean add(final @Nullable DiscoveryResult discoveryResult) throws IllegalStateException {
+        if (discoveryResult == null) {
             return false;
         }
-        final DiscoveryResult result = rslt;
+        final DiscoveryResult result = discoveryResult;
 
         ThingUID thingUID = result.getThingUID();
         Thing thing = this.thingRegistry.get(thingUID);
@@ -300,7 +300,7 @@ public final class PersistentInbox implements Inbox, DiscoveryListener, ThingReg
     @Override
     @NonNullByDefault({})
     public Stream<DiscoveryResult> stream() {
-        return this.discoveryResultStorage.getValues().stream().filter(discoveryResult -> discoveryResult == null);
+        return this.discoveryResultStorage.getValues().stream().filter(Objects::nonNull);
     }
 
     @Override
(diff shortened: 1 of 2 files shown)


bundles/config/org.eclipse.smarthome.config.discovery/src/main/java/org/eclipse/smarthome/config/discovery/internal/PersistentInbox.java
```diff
@@ -173,43 +176,46 @@ public Thing approve(ThingUID thingUID, String label) {
...
-
-            if (thing == null) {
-                DiscoveryResult inboxResult = get(thingUID);
-
-                if (inboxResult == null) {
-                    discoveryResultStorage.put(result.getThingUID().toString(), result);
-                    notifyListeners(result, EventType.added);
-                    logger.info("Added new thing '{}' to inbox.", thingUID);
+    public synchronized boolean add(final @Nullable DiscoveryResult rslt) throws IllegalStateException {
+        if (rslt == null) {
```
afuechsel: Can we use `discoveryResult` instead of `rslt`?

bundles/config/org.eclipse.smarthome.config.discovery/src/main/java/org/eclipse/smarthome/config/discovery/internal/PersistentInbox.java
```diff
@@ -289,12 +298,13 @@ public void addInboxListener(InboxListener listener) throws IllegalStateExceptio
     }
 
     @Override
+    @NonNullByDefault({})
     public Stream<DiscoveryResult> stream() {
-        return this.discoveryResultStorage.getValues().stream();
+        return this.discoveryResultStorage.getValues().stream().filter(discoveryResult -> discoveryResult == null);
```
afuechsel: if you filter that way you will end up in streaming nulls, won't you? Wasn't your idea to filter `discoveryResult != null` ?

maggu2810: `getValues()` is declarated that it returns a collection that contains potential null entries. So without the filter this method needs to be changed to a stream that contains nullable entries.

bundles/config/org.eclipse.smarthome.config.discovery/src/main/java/org/eclipse/smarthome/config/discovery/internal/PersistentInbox.java
maggu2810: If the Eclipse Nullness checker would be clever (as others) it would not be necessary to introduce another local variable at all.
After returning if the argument is null, the checker should know that all occurrences in that method (at least as the argument is also final) are known to be non-null.
But the checker will complain about...

Sure, I don't care how to name the argument. The method signature is defined by the interface, also the JavaDoc, so the name here that is used for the nullness check at all should not matter.

bundles/config/org.eclipse.smarthome.config.discovery/src/main/java/org/eclipse/smarthome/config/discovery/internal/PersistentInbox.java
afuechsel: @maggu2810 Yes, but still the filter should be the other way around: `discoveryResult != null`.

maggu2810:
I will have a look at the failing tests.

bundles/config/org.eclipse.smarthome.config.discovery/src/main/java/org/eclipse/smarthome/config/discovery/internal/PersistentInbox.java
maggu2810: You are surely right :rofl:

PR #5679, opened 2018-06-04
TransformationProfile and SystemOffsetProfile

triller-telekom:
This PR includes two new profiles.

The TransformationProfile is a new system profile which is registered
under the scope "transform". It offers the possibility to use all
TransformationServices on ChannelItemLinks. The Strings send from a device
will be transformed via the specified function in the "pattern" parameter.
The other direction, i.e. Item to device is left untouched.

The SystemOffsetProfile lives in the "system" scope of profiles and is able
to add or substract and offset from a value send from or to a device. This
profile works in both directions by applying the opposite offset for the
opposite direction. QuantityTypes are supported as well, i.e. one can put
an offset of "1°C" on a channel that sends Fahrenheit.

In addition there is a new "ChattyThing" fort he magic binding which sends
random values on a Number channel and random Strings on a StringType channel
based on a configurable interval. For profile testing this turned out to be
quite useful.


triller-telekom:
I have marked this PR as WIP because there is one thing left to discuss:

Currently I allow offsets without a unit and then implicitly assume its the same as the one send to the channel.
However, a user might not know this type or this type is changed in a future version of the binding and this will create confusing results. So we mighty only want to allow offsets without a unit if the binding only sends plain Number states without units.

bundles/core/org.eclipse.smarthome.core.thing/src/main/java/org/eclipse/smarthome/core/thing/internal/profiles/SystemOffsetProfile.java
```diff
@@ -0,0 +1,137 @@
...
+    public void onStateUpdateFromHandler(State state) {
+        logger.debug("State update from handler: '{}'", state);
+        callback.sendUpdate((State) applyOffset(state, true));
+    }
+
+    @SuppressWarnings({ "rawtypes", "unchecked" })
+    private Type applyOffset(Type state, boolean towardsThing) {
+        QuantityType finalOffset = offset;
+        if (finalOffset == null) {
+            logger.error("Offset not configured correctly, returning undef state.");
```
sjka: This just states what is is doing - as logs are directed towards (admin) users, do you think this could be made a little more helpful wrt to guiding the user to do it right?

bundles/core/org.eclipse.smarthome.core.thing/src/main/java/org/eclipse/smarthome/core/thing/internal/profiles/SystemOffsetProfile.java
```diff
@@ -0,0 +1,137 @@
...
+            return UnDefType.UNDEF;
+        }
+
+        if (!towardsThing) {
+            finalOffset = finalOffset.negate();
```
sjka: My assumption was that if I have a device which say delivers `20 °C` but the "true" environment temperature is `21 °C`, that I configure an offset of `+1`. This statement makes it the other way round, right?

Oh, wait, maybe the variable is just named the wrong way (see above...)

bundles/core/org.eclipse.smarthome.core.thing/src/main/java/org/eclipse/smarthome/core/thing/internal/profiles/SystemOffsetProfile.java
```diff
@@ -0,0 +1,137 @@
...
+    }
+
+    @Override
+    public void onStateUpdateFromHandler(State state) {
+        logger.debug("State update from handler: '{}'", state);
+        callback.sendUpdate((State) applyOffset(state, true));
+    }
+
+    @SuppressWarnings({ "rawtypes", "unchecked" })
+    private Type applyOffset(Type state, boolean towardsThing) {
```
sjka: Is this really `towardsThing`? Isn't it used exactly the opposite?

bundles/core/org.eclipse.smarthome.core.thing/src/main/java/org/eclipse/smarthome/core/thing/internal/profiles/SystemOffsetProfile.java
```diff
@@ -0,0 +1,137 @@
...
+        QuantityType finalOffset = offset;
+        if (finalOffset == null) {
+            logger.error("Offset not configured correctly, returning undef state.");
+            return UnDefType.UNDEF;
```
sjka: How about assuming an offset of  `0` in that case and simply returning the original value?

bundles/core/org.eclipse.smarthome.core.thing/src/main/java/org/eclipse/smarthome/core/thing/internal/profiles/SystemOffsetProfile.java
```diff
@@ -0,0 +1,137 @@
...
+                result = qtState.add(finalOffset);
+            } catch (UnconvertibleException e) {
+                logger.error("Cannot apply offset '{}' to state '{}' because types do not match.", finalOffset,
+                        qtState);
+            }
+        } else if (state instanceof DecimalType && finalOffset.getUnit().equals(AbstractUnit.ONE)) {
+            DecimalType decState = (DecimalType) state;
+            result = new DecimalType(decState.doubleValue() + finalOffset.doubleValue());
+        } else {
+            logger.error("State '{}' of type '{}' is not applicable for this profile.", state, state.getClass());
```
sjka: Why error, isn't this rather a warning? And why not falling back to returning the origin value instead of UnDefType.UNDEF?

bundles/core/org.eclipse.smarthome.core.thing/src/main/java/org/eclipse/smarthome/core/thing/internal/profiles/SystemOffsetProfile.java
```diff
@@ -0,0 +1,137 @@
...
+            try {
+                if (finalOffset.getUnit() == AbstractUnit.ONE) {
+                    // allow offsets without unit -> implicitly assume its the same as the one from the state
+                    // TODO: do we want this? User would have to look up the original unit on the channel in the
+                    // binding...
+                    finalOffset = new QuantityType<>(finalOffset.toBigDecimal(), qtState.getUnit());
+                }
+                result = qtState.add(finalOffset);
+            } catch (UnconvertibleException e) {
+                logger.error("Cannot apply offset '{}' to state '{}' because types do not match.", finalOffset,
```
sjka: A `warning` should suffice here

bundles/core/org.eclipse.smarthome.core.thing/src/main/java/org/eclipse/smarthome/core/thing/internal/profiles/SystemOffsetProfile.java
```diff
@@ -0,0 +1,137 @@
...
+        Type result = UnDefType.UNDEF;
+
+        if (state instanceof QuantityType) {
+            QuantityType qtState = (QuantityType) state;
+            try {
+                if (finalOffset.getUnit() == AbstractUnit.ONE) {
+                    // allow offsets without unit -> implicitly assume its the same as the one from the state
+                    // TODO: do we want this? User would have to look up the original unit on the channel in the
+                    // binding...
+                    finalOffset = new QuantityType<>(finalOffset.toBigDecimal(), qtState.getUnit());
```
sjka: This _can_ lead to quite funny results whenever the concrete units change. Or did I miss anything?

bundles/core/org.eclipse.smarthome.core.thing/src/main/java/org/eclipse/smarthome/core/thing/internal/profiles/SystemOffsetProfile.java
sjka: If I configure an offset of `-1 K` for a binding which doesn't yet support quantity types, I will also end up with this log entry. And this is pretty misleading for the user.

bundles/core/org.eclipse.smarthome.core.thing/src/main/java/org/eclipse/smarthome/core/thing/profiles/ProfileTypeUID.java
```diff
@@ -47,4 +49,23 @@ public String getId() {
         return getSegment(1);
     }
 
+    @Override
+    public boolean equals(@Nullable Object obj) {
```
sjka: Why is that? This class inherits the `equals()` implementation from `AbstractUID`, which essentially does the same. Or did I miss anything?

bundles/core/org.eclipse.smarthome.core.thing/src/main/java/org/eclipse/smarthome/core/thing/profiles/ProfileTypeUID.java
```diff
@@ -13,12 +13,14 @@
...
 
 import org.eclipse.jdt.annotation.NonNullByDefault;
+import org.eclipse.jdt.annotation.Nullable;
 import org.eclipse.smarthome.core.thing.UID;
 
 /**
  * Identifier of a profile type.
  *
  * @author Simon Kaufmann - initial contribution and API.
+ * @author Stefan Triller - added equals method
```
sjka: ...and forgot `hashcode()` 😉 . But please see below!

bundles/core/org.eclipse.smarthome.core.transform/META-INF/MANIFEST.MF
```diff
@@ -11,16 +11,24 @@ Bundle-Vendor: Eclipse.org/SmartHome
...
+ org.eclipse.smarthome.core.thing.type,
  org.eclipse.smarthome.core.transform,
  org.eclipse.smarthome.core.transform.actions,
+ org.eclipse.smarthome.core.types,
  org.osgi.framework,
  org.osgi.util.tracker,
  org.slf4j
 Private-Package: org.eclipse.smarthome.binding.http.internal
+Service-Component: OSGI-INF/*.xml
+Bundle-ActivationPolicy: lazy
```
sjka: That line is wrong, AFAIK. Or at least pointless in our non-RCP context.

bundles/core/org.eclipse.smarthome.core.transform/src/main/java/org/eclipse/smarthome/core/transform/internal/profiles/TransformationProfile.java
```diff
@@ -0,0 +1,137 @@
...
+ * {@link StateProfile} which transforms a state value send from the handler to the framework based on the provided
+ * {@link ProfileTypeUID}. The direction from the framework toward the handler is left untouched.
+ *
+ * @author Stefan Triller - initial contribution
+ *
+ */
+@NonNullByDefault
+public class TransformationProfile implements StateProfile {
+
+    private static Logger logger = LoggerFactory.getLogger(TransformationProfile.class);
```
sjka: `final`, not `static`

bundles/core/org.eclipse.smarthome.core.transform/src/main/java/org/eclipse/smarthome/core/transform/internal/profiles/TransformationProfile.java
```diff
@@ -0,0 +1,137 @@
...
+import org.eclipse.smarthome.core.transform.TransformationHelper;
+import org.eclipse.smarthome.core.transform.TransformationService;
+import org.eclipse.smarthome.core.types.Command;
+import org.eclipse.smarthome.core.types.State;
+import org.eclipse.smarthome.core.types.Type;
+import org.slf4j.Logger;
+import org.slf4j.LoggerFactory;
+
+/**
+ * {@link StateProfile} which transforms a state value send from the handler to the framework based on the provided
```
sjka: send -> sent

bundles/core/org.eclipse.smarthome.core.transform/src/main/java/org/eclipse/smarthome/core/transform/internal/profiles/TransformationProfile.java
```diff
@@ -0,0 +1,137 @@
...
+ * {@link ProfileTypeUID}. The direction from the framework toward the handler is left untouched.
```
sjka: toward -> towards

sjka: Is the "direction" untouched, or the state/value?

bundles/core/org.eclipse.smarthome.core.transform/src/main/java/org/eclipse/smarthome/core/transform/internal/profiles/TransformationProfile.java
```diff
@@ -0,0 +1,137 @@
...
+@NonNullByDefault
+public class TransformationProfile implements StateProfile {
+
+    private static Logger logger = LoggerFactory.getLogger(TransformationProfile.class);
+
+    public static final String PROFILE_NAME = "Transformation";
+    private final ProfileTypeUID profileTypeUID;
+    private final TransformationService service;
+
+    private static String PATTERN_PARAM = "pattern";
```
sjka: +`final`?

bundles/core/org.eclipse.smarthome.core.transform/src/main/java/org/eclipse/smarthome/core/transform/internal/profiles/TransformationProfile.java
```diff
@@ -0,0 +1,137 @@
...
+
+    public TransformationProfile(ProfileCallback callback, ProfileContext context, TransformationService service,
+            ProfileTypeUID typeUID) {
+
+        this.service = service;
+        this.profileTypeUID = typeUID;
+        this.callback = callback;
+
+        Object paramValue = context.getConfiguration().get(PATTERN_PARAM);
+        logger.debug("Value for parameter '{}' is '{}'", PATTERN_PARAM, paramValue);
```
sjka: see above

bundles/core/org.eclipse.smarthome.core.transform/src/main/java/org/eclipse/smarthome/core/transform/internal/profiles/TransformationProfile.java
```diff
@@ -0,0 +1,137 @@
...
+
+    @Override
+    public void onCommandFromItem(Command command) {
+        callback.handleCommand(command);
+    }
+
+    @Override
+    public void onCommandFromHandler(Command command) {
+        if (function == null || format == null) {
+            return;
```
sjka: Does being "inactive" mean just not transforming anything, or rather making the system useless? 😉

bundles/core/org.eclipse.smarthome.core.transform/src/main/java/org/eclipse/smarthome/core/transform/internal/profiles/TransformationProfile.java
```diff
@@ -0,0 +1,137 @@
...
+        }
+        callback.sendCommand((Command) transformState(command));
+    }
+
+    @Override
+    public void onStateUpdateFromHandler(State state) {
+        if (function == null || format == null) {
+            return;
```
sjka: see above

bundles/core/org.eclipse.smarthome.core.transform/src/main/java/org/eclipse/smarthome/core/transform/internal/profiles/TransformationProfile.java
```diff
@@ -0,0 +1,137 @@
...
+        } else {
+            logError();
+            function = null;
+            format = null;
+
+        }
+    }
+
+    private void logError() {
+        logger.error("Parameter '{}' has to be a String containing separator '{}'. Profile will be inactive.",
```
sjka: `warning`

bundles/core/org.eclipse.smarthome.core.transform/src/main/java/org/eclipse/smarthome/core/transform/internal/profiles/TransformationProfile.java
```diff
@@ -0,0 +1,137 @@
...
+        this.service = service;
+        this.profileTypeUID = typeUID;
+        this.callback = callback;
+
+        Object paramValue = context.getConfiguration().get(PATTERN_PARAM);
+        logger.debug("Value for parameter '{}' is '{}'", PATTERN_PARAM, paramValue);
+        if (paramValue instanceof String) {
+            String pattern = (String) paramValue;
+            if (pattern.contains(TransformationHelper.FUNCTION_VALUE_DELIMITER)) {
+                function = pattern.split(TransformationHelper.FUNCTION_VALUE_DELIMITER)[0];
```
sjka: I know, this doesn't save the world, but why do the split twice?

bundles/core/org.eclipse.smarthome.core.transform/src/main/java/org/eclipse/smarthome/core/transform/internal/profiles/TransformationProfile.java
```diff
@@ -0,0 +1,137 @@
...
+
+        this.service = service;
+        this.profileTypeUID = typeUID;
+        this.callback = callback;
+
+        Object paramValue = context.getConfiguration().get(PATTERN_PARAM);
+        logger.debug("Value for parameter '{}' is '{}'", PATTERN_PARAM, paramValue);
+        if (paramValue instanceof String) {
+            String pattern = (String) paramValue;
+            if (pattern.contains(TransformationHelper.FUNCTION_VALUE_DELIMITER)) {
```
sjka: If you concatenate this condition with the previous one, than you don't need the duplicate else-block and can eliminitate the awkward `logError()`-method.

bundles/core/org.eclipse.smarthome.core.transform/src/main/java/org/eclipse/smarthome/core/transform/internal/profiles/TransformationProfile.java
```diff
@@ -0,0 +1,137 @@
...
+                result = service.transform(function, newFormat);
+            } catch (TransformationException e) {
+                logger.debug("Could not transform state '{}' with function '{}' and format '{}'", state, function,
+                        format);
+            }
+            StringType resultType = new StringType(result);
+            logger.debug("Transformed '{}' into '{}'", state, resultType);
+            return resultType;
+        } else {
+            logger.debug("State '{}' is not of type StringType, skipping transformation.", state);
```
sjka: The limitation to StringTypes makes it quite use-less in reality, doesn't it? AFAIK, transformations can be applied to all sorts of item types, not only StringItems.

bundles/core/org.eclipse.smarthome.core.transform/src/main/java/org/eclipse/smarthome/core/transform/internal/profiles/TransformationProfileFactory.java
```diff
@@ -0,0 +1,99 @@
...
+ */
+@NonNullByDefault
+@Component(service = { ProfileFactory.class, ProfileTypeProvider.class })
+public class TransformationProfileFactory implements ProfileFactory, ProfileTypeProvider {
+
+    private static final String SERVICE_PROPERTY = "smarthome.transform";
+
+    private final HashMap<ProfileTypeUID, ProfileType> supportedProfileTypes = new HashMap<>();
+
+    // private final Logger logger = LoggerFactory.getLogger(TransformationProfileFactory.class);
```
sjka: ??

bundles/core/org.eclipse.smarthome.core.transform/src/main/java/org/eclipse/smarthome/core/transform/internal/profiles/TransformationProfileFactory.java
```diff
@@ -0,0 +1,99 @@
...
+ * @author Stefan Triller - initial contribution
+ *
+ */
+@NonNullByDefault
+@Component(service = { ProfileFactory.class, ProfileTypeProvider.class })
+public class TransformationProfileFactory implements ProfileFactory, ProfileTypeProvider {
+
+    private static final String SERVICE_PROPERTY = "smarthome.transform";
+
+    private final HashMap<ProfileTypeUID, ProfileType> supportedProfileTypes = new HashMap<>();
```
sjka: Either this collections needs to be a "concurrent" one, or the add/remove methods must be synchronized

bundles/core/org.eclipse.smarthome.core.transform/src/main/java/org/eclipse/smarthome/core/transform/internal/profiles/TransformationProfileFactory.java
```diff
@@ -0,0 +1,99 @@
...
+
+    // private final Logger logger = LoggerFactory.getLogger(TransformationProfileFactory.class);
+
+    private final HashMap<ProfileTypeUID, TransformationService> transformationServices = new HashMap<>();
```
sjka: see above

bundles/core/org.eclipse.smarthome.core.transform/src/main/java/org/eclipse/smarthome/core/transform/profiles/TransformationProfiles.java
```diff
@@ -0,0 +1,33 @@
...
+
+/**
+ * Profiles for state transformations
+ *
+ * @author Stefan Triller - initial contribution
+ *
+ */
+public interface TransformationProfiles {
+
+    static final String SCOPE = "transform";
```
sjka: A public interface just for this one package-private constant doesn't make sense, does it?

bundles/core/org.eclipse.smarthome.core.transform/src/main/java/org/eclipse/smarthome/core/transform/profiles/TransformationProfiles.java
```diff
@@ -0,0 +1,33 @@
...
+
+    static final ProfileTypeUID OFFSET = new ProfileTypeUID(SCOPE, "offset");
```
sjka: ??

bundles/core/org.eclipse.smarthome.core.transform/src/main/java/org/eclipse/smarthome/core/transform/profiles/TransformationProfiles.java
```diff
@@ -0,0 +1,33 @@
...
+
+    static final StateProfileType OFFSET_TYPE = ProfileTypeBuilder.newState(OFFSET, "Offset").build();
```
sjka: ??

docs/documentation/development/bindings/profiles.md
```diff
@@ -89,4 +89,29 @@ Two methods have to be implemented to achieve this:
...
+## Existing Profiles
+
+### OffsetProfile
+
+The `OffsetProfile` provides the possibility to adjust a value from a device before it arrives at the framework.
+An offset can be specified via the parameter `offset` which has to be a `QuantityType` or `DecimalType`.
+A positive offset is the amount of change from the device towards the framework and vice versa.
+
+```java
+Number MyItem { channel="<bindingID>:<thing-typeID>:MyThing:myChannel"[profile="offset", pattern="<transformFunction>:<valueFormat>"]}
```
sjka: The pattern looks like c'n'p from below

docs/documentation/development/bindings/profiles.md
```diff
@@ -89,4 +89,29 @@ Two methods have to be implemented to achieve this:
...
\ No newline at end of file
+```
+
+## Existing Profiles
+
+### OffsetProfile
+
+The `OffsetProfile` provides the possibility to adjust a value from a device before it arrives at the framework.
+An offset can be specified via the parameter `offset` which has to be a `QuantityType` or `DecimalType`.
+A positive offset is the amount of change from the device towards the framework and vice versa.
```
sjka: No, not "vice versa". You'll have to stick to explaining a single direction if this sentence should be of any help.

docs/documentation/development/bindings/profiles.md
```diff
@@ -89,4 +89,29 @@ Two methods have to be implemented to achieve this:
...
+```
+
+### TransformationProfile
+
+The `TransformationProfile` offers the functionality of the `TransformationService`s to `Profile`s.
+Each `TransformationService` is identified by a property called `smarthome.transform`.
+This property specifies the name of the transformation.
+The function and format of the value can be specified in the `pattern` parameter of the `TransformationProfile` as a string, separated by a colon.
+Note that these transformations are unidirectional, i.e. they are configured to transform the value send by a device towards the framework, only.
+Commands send by the framework towards the device are passed through the `Profile` without any adjustments.
```
sjka: send -> sent

sjka: \`Profile` -> profile

docs/documentation/development/bindings/profiles.md
sjka: You are mixing here BNF notation (e.g. `<bindingID>`) with concrete examples (e.g. `MyThing`). Both make sense, but I wouldn't mix them in the same example.

docs/documentation/development/bindings/profiles.md
```diff
@@ -89,4 +89,29 @@ Two methods have to be implemented to achieve this:
...
+
+```java
+String MyItem { channel="<bindingID>:<thing-typeID>:MyThing:myChannel"[profile="transform:<transformationName>", pattern="<transformFunction>:<valueFormat>"]}
```
sjka: see above

bundles/core/org.eclipse.smarthome.core.thing/src/main/java/org/eclipse/smarthome/core/thing/internal/profiles/SystemOffsetProfile.java
triller-telekom: That is why the TODO is in the code and in my comment on this PR ;-)

bundles/core/org.eclipse.smarthome.core.thing/src/main/java/org/eclipse/smarthome/core/thing/internal/profiles/SystemOffsetProfile.java
```diff
@@ -0,0 +1,137 @@
...
+        logger.debug("Parameter '{}' has value '{}' ", OFFSET_PARAM, paramValue);
+        if (paramValue instanceof String) {
+            try {
+                offset = new QuantityType<>((String) paramValue);
+            } catch (IllegalArgumentException e) {
+                logger.error("Cannot convert value '{}' of parameter '{}' into a valid offset of type QyantityType.",
+                        paramValue, OFFSET_PARAM);
+            }
+        } else {
+            logger.error("Parameter '{}' is not of type String", OFFSET_PARAM);
```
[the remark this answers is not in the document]
triller-telekom: It is of type String because I allow Quantitytypes, what else shall I use for configuration parameters like "3.5 °C"?

bundles/core/org.eclipse.smarthome.core.thing/src/main/java/org/eclipse/smarthome/core/thing/internal/profiles/SystemOffsetProfile.java
triller-telekom: Why is this misleading? A user wants to subtract `1 K` from a value on that channel, how should the framework know "how much" is should subtract in case the binding sends "23" but means "kg" ? Shall it subtract "1" and ignore the unit completely?

bundles/core/org.eclipse.smarthome.core.thing/src/main/java/org/eclipse/smarthome/core/thing/internal/profiles/SystemOffsetProfile.java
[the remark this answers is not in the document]
sjka: _Allowing_ strings in order to enable QuantityTypes is one thing and is great. _Not_ allowing numeric values and logging them as an _error_ is another. Isn't it?

bundles/core/org.eclipse.smarthome.core.thing/src/main/java/org/eclipse/smarthome/core/thing/internal/profiles/SystemOffsetProfile.java
sjka: Okay, let me give you an example:

```
... ERROR  State 23 of type DecimalType is not applicable for this profile.
```

How is this helpful? How is the user supposed to know how to handle this situation? Which profile? Which profile? And where does this state come from?

triller-telekom:
I have removed the WIP because for backward compatibility I now allow offsets of type `DecimalType` on states of type `QuantityType`, but log a warning that the user should check his setup.

triller-telekom:
I have set this PR to WIP again because there is again one thing to discuss:

For transformations I started from the item definitions which have `"Window [MAP(window_esp.map):%s]"` in their state descriptions for example. The question is whether the formatting of the state **before** the transformation (the %s) is useful to the profiles or not. I have implemented it in the `TransformationProfile` but i am not sure whether we only should transform the raw state without any possibility to format the state before it will be transformed.

I have also added configdescriptions for the transformation profiles.

WDYT?

triller-telekom:
Finally I implemented the special handling of temperatures so one can have "3°F" as an offset on a channel that sends values in "°C".

triller-telekom:
This PR is now ready for review.

triller-telekom:
Arrrg: Next time I'll leave a conceptual PR like this longer in the WIP state...

An update will follow: There should not be one single Transformationprofile implementation that picks up all available `TransformationService`s, but instead each `TransformationService` should provide its own factory and profile(s).

An update will follow.

extensions/transform/org.eclipse.smarthome.transform.xslt/ESH-INF/config/xsltProfile.xml
```diff
@@ -0,0 +1,19 @@
...
+<config-description:config-descriptions
+    xmlns:xsi="http://www.w3.org/2001/XMLSchema-instance"
+    xmlns:config-description="http://eclipse.org/smarthome/schemas/config-description/v1.0.0"
+    xsi:schemaLocation="http://eclipse.org/smarthome/schemas/config-description/v1.0.0
+    http://eclipse.org/smarthome/schemas/config-description-1.0.0.xsd">
+
+    <config-description uri="profile:transform:XSLT">
+        <parameter name="function" type="text" required="true">
+            <label>XSL Filename</label>
+            <description>XSL Filename in transform folder, containing the transformation expression.</description>
```
htreu: `filename` or better `file name`.

extensions/transform/org.eclipse.smarthome.transform.xslt/README.md
```diff
@@ -103,6 +103,21 @@ Fri, 18 Dec 2009 9:38 am PST
...
+## Usage as a Profile
+
+The functionality of this `TransformationService` can be used in a `Profile` on an `ItemChannelLink` too.
+To do so, it can be configured in the `.items` file as follows:
+
+```java
+String <itemName> { channel="<channelUID>"[profile="transform:XSLT", function="<xsltExpression>", sourceFormat="<valueFormat>"]}
+```
+
+The XSLT expression to be executed has to be set in the `function` parameter.
```
htreu: in the config description the function parameter is supposed to hold a filename. Here it should be the XSLT expression...

edit: the XsltTransformationService expects it to be the file name.

PR #5681, opened 2018-06-04
Update things.md

vampylouis:
Correct a minor grammar mistake.

### DIFF
--- a/docs/documentation/concepts/things.md
+++ b/docs/documentation/concepts/things.md
@@ -19,7 +19,7 @@ A physical light bulb might have a color temperature channel and a color channel
 For sources of information the thing might be the local weather with information from a web service with different channels like temperature, pressure and humidity.
 
 Channels are linked to items, where such links are the glue between the virtual and the physical layer.
-Once such a link is established, a thing reacts on events sent for an item that is linked to one of its channels.
+Once such a link is established, a thing reacts on events sent from an item that is linked to one of its channels.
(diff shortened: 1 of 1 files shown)


kaikreuzer:
Actually, I would claim that the current formulation is correct; events are sent FOR (or to) an item. Items themselve just represent the state, but they do not send out events.

tavalin:
I would also consider "...on events sent to an item linked..."

vampylouis:
Thanks for the clarification. I am new to Openhab and not knowing the fact that 'Item' never generate events but only consume/receive events from 'Thing' through its channel.

I misinterpreted this statement from the document:
_**"Once such a link is established, a thing reacts on events sent for an item that is linked to one of its channels."**_

In fact, this statement mean:
"Once such a link is established, a thing reacts on events and send them to an item that is linked to one of its channels."

Correct ?

sjka:
No, not really... Let me try to explain it in more detail: 

The user (or a rule) issues an event (like a command, e.g. `ON`) and directs it towards an item, as this is the functional abstraction that the user interacts with. With the means of the link between the item and a thing's channel, the thing (i.e. the physical abstraction of devices/sensors/actuators/...) is made aware of this and therefore can do whatever is necessary in order to switch on the real lamp. (The same obviously works in reverse if in the physical world something changes).

Therefore I'd say the sentence is correct as it it.

kaikreuzer:
So if we all agree that the current sentence is correct, we can happily close this PR :-)

Issue #5003, opened 2018-01-29
Sonos binding, partly implemented getFavoriteRadios getFavorites


BasvanH:
Hello,

Currently, the Sonos Binding does not have a channel for getting an list of the favorites. I would like to use this list in an HabPanel widget so you can choose an favorite to play.

The SonosPlayeHandler does have some parts of code towards getting the favorites, but it doesn't have it ported to an channel. Could this be implemented?

SonosPlayerHandler.java
https://github.com/eclipse/smarthome/blob/master/extensions/binding/org.eclipse.smarthome.binding.sonos/src/main/java/org/eclipse/smarthome/binding/sonos/internal/handler/ZonePlayerHandler.java

triller-telekom:
This sounds like a good use case for #4942.

You could have a dynamic "SelectionOptionProvider" for a favorites channel and show it as a selection element in a basic ui sitemap or whatever kind of widget in HabPanel.

BasvanH:
Kindly requesting an update.

maggu2810:
As long as there is no volunteer that would like to implement that feature...

lolodomo:
I will have a look as soon as my changes in Classic UI and Basic UI are merged.
A similar thing was apparently already implemented in the Kodi binding (building options dynamically).

lolodomo:
In fact, it could be applied to 3 channels: `favorite`, `playlist` and `radio`.

lolodomo:
I implemented that in the Sonos binding as a POC. Works well.
I have a problem to solve before pushing: when triggering a refresh ? I was expecting a notification through the UPnP ContentDirectory subnscription but it does not happen and I don't understand why (I know that iot was happening in the Vera plugin).

I have one worry: what will happen with very big lists ? Should we limit to a certain max list size ?

lolodomo:
Unfortunately there is first an issue to fix in the jupnp library
https://github.com/jupnp/jupnp/issues/97

sjka:
I just merged the PR - nevertheless, the above statement wrt a yet open change in jUPnP remains valid. See also https://github.com/eclipse/smarthome/pull/5682#issuecomment-396170949.

lolodomo:
I have not yet checked but the new feature should work starting with the snapshot 1305.

lolodomo:
I can now confirm that it is available.

PR #5680, opened 2018-06-04
Update jUPnP to version 2.4

triller-telekom:
Needs the upcoming 2.4 release before this PR can be merged.

Signed-off-by: Stefan Triller <stefan.triller@telekom.de>

PR #5682, opened 2018-06-04
Sonos: implement a dynamic state description option provider

lolodomo:
Used to define dynamically the state options for the channels radio, favorite and playlist

Fixes issue #5003

### DIFF
--- a/extensions/binding/org.eclipse.smarthome.binding.sonos/src/main/java/org/eclipse/smarthome/binding/sonos/internal/SonosStateDescriptionOptionProvider.java
+++ b/extensions/binding/org.eclipse.smarthome.binding.sonos/src/main/java/org/eclipse/smarthome/binding/sonos/internal/SonosStateDescriptionOptionProvider.java
@@ -32,8 +32,7 @@ import org.osgi.service.component.annotations.Deactivate;
  *
  * @author Laurent Garnier - Initial contribution
  */
-@Component(service = { DynamicStateDescriptionProvider.class,
-        SonosStateDescriptionOptionProvider.class }, immediate = true)
+@Component(service = { DynamicStateDescriptionProvider.class, SonosStateDescriptionOptionProvider.class })
 @NonNullByDefault
 public class SonosStateDescriptionOptionProvider implements DynamicStateDescriptionProvider {
     private final Map<ChannelUID, @Nullable List<StateOption>> channelOptionsMap = new ConcurrentHashMap<>();
--- a/extensions/binding/org.eclipse.smarthome.binding.sonos/src/main/java/org/eclipse/smarthome/binding/sonos/internal/SonosStateDescriptionOptionProvider.java
+++ b/extensions/binding/org.eclipse.smarthome.binding.sonos/src/main/java/org/eclipse/smarthome/binding/sonos/internal/SonosStateDescriptionOptionProvider.java
@@ -1,10 +1,14 @@
 /**
- * Copyright (c) 2010-2018 by the respective copyright holders.
+ * Copyright (c) 2014,2018 Contributors to the Eclipse Foundation
  *
- * All rights reserved. This program and the accompanying materials
- * are made available under the terms of the Eclipse Public License v1.0
- * which accompanies this distribution, and is available at
- * http://www.eclipse.org/legal/epl-v10.html
+ * See the NOTICE file(s) distributed with this work for additional
+ * information regarding copyright ownership.
+ *
+ * This program and the accompanying materials are made available under the
+ * terms of the Eclipse Public License 2.0 which is available at
+ * http://www.eclipse.org/legal/epl-2.0
+ *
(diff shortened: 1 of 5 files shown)


lolodomo:
Will work well only after https://github.com/jupnp/jupnp/pull/98 is merged and a new version of the JUPnP library including the fix is pushed in the target platform.

lolodomo:
Note that this PR can be merged before the JUPnP library is updated.

sjka:
No, it can't - for a different reason though 😉 

```
[INFO] --- sat-plugin:0.5.0:report (sat-all) @ org.eclipse.smarthome.binding.sonos ---
[INFO] Individual report appended to summary report.
[ERROR] Code Analysis Tool has found: 
 1 error(s)! 
 1 warning(s) 
 1 info(s)
[ERROR] org.eclipse.smarthome.binding.sonos.internal.SonosStateDescriptionOptionProvider.java:[2]
Header line doesn't match pattern ^ \* Copyright \(c\) 2014,2018 Contributors to the Eclipse Foundation$
[WARNING] org.eclipse.smarthome.binding.sonos.internal.SonosResourceMetaData.java:[21]
First javadoc author should have "Initial contribution" contribution description.
[INFO] Detailed report can be found at: file:////home/travis/build/eclipse/smarthome/extensions/binding/org.eclipse.smarthome.binding.sonos/target/code-analysis/report.html
```

lolodomo:
I made a copy from an OH 2 binding, kodi binding I think.
I will fix that.

extensions/binding/org.eclipse.smarthome.binding.sonos/src/main/java/org/eclipse/smarthome/binding/sonos/internal/SonosStateDescriptionOptionProvider.java
```diff
@@ -0,0 +1,65 @@
...
+import org.osgi.service.component.annotations.Component;
+import org.osgi.service.component.annotations.Deactivate;
+
+/**
+ * Dynamic provider of state options while leaving other state description fields as original.
+ *
+ * @author Laurent Garnier - Initial contribution
+ */
+@Component(service = { DynamicStateDescriptionProvider.class,
+        SonosStateDescriptionOptionProvider.class }, immediate = true)
```
maggu2810: Why do we need an immediate activation for that component?

lolodomo: Is it a valid reason if I tell that all (7) OH2 bindings implementing such provider have set immediate to true ?

lolodomo:
PR updated (wrong copyright).

lolodomo:
It was a little stupid to not use the existing switch case. I will update my PR a little bit.

extensions/binding/org.eclipse.smarthome.binding.sonos/src/main/java/org/eclipse/smarthome/binding/sonos/internal/SonosStateDescriptionOptionProvider.java
maggu2810: Not really. :wink: 
It could be set by mistake or perhaps the developer does not understand the instruction and copied it from another component...
Immediate should only be used if the component must be activated also if there is no consumer of that service.
On immediate false (default) the component is activated as soon as someone consumes its but it keeps deactivated as long as this situation is not true.

lolodomo: I will make a try without immediate = true.

sjka:
Code-wise it looks good to me.
Unfortunately I can't get it to work - but actually I'm having trouble receiving anything currently from my speaker (controlling it works), so I guess that has nothing to do with your change.
Anybody else having more luck?

lolodomo:
It can't really work without my updated version of the JUPnP library.

lolodomo:
UNtil the fixed version of the JUPnP library is included, there is no GENA event for the service ContentDirectory received by the binding and so the state options are never built.

lolodomo:
I can provide the JUPnP library for testing.

sjka:
Argh, sorry, I overlooked your hint above. Of course, that makes sense...
No need to provide anything, I can simply fetch the PR myself.

sjka:
With jupnp/jupnp#98 this feature now works like a charm. Without, it doesn't do any harm, therefore I don't see any point in leaving this hanging around as an open PR. Still:

⚠️ ATTENTION ⚠️ 
This feature will ONLY work once jupnp/jupnp#98 is merged AND a successor of jUPnP 2.3 is released  (e.g. 2.4) AND integrated here (preparation in #5680).

Issue #5674, opened 2018-06-03
[REST API] MessageBodyWriter not found for media type=text/plain


openhab-5iver:
All the details can be found in this issue in the Habmin repo... https://github.com/cdjackson/HABmin/issues/323#issuecomment-394179178, but this occurs for all devices (not just zwave). Please let me know if more info is needed.

This call...
```
http://localhost:8080/rest/things/kodi:kodi:kodifamilyroom/config/status
```
... returns a 500 status and logs the following...
```
2018-06-03 16:28:58.816 [ERROR] [.jersey.message.internal.WriterInterceptorExecutor] - MessageBodyWriter not found for media type=text/plain, type=class java.util.Collections$UnmodifiableCollection, genericType=class java.util.Collections$UnmodifiableCollection.
```
Possibly a separate issue, but this call...
```
http://fedora01.rushworth.us:8080/rest/config-descriptions/channel:kodi:kodi:kodifamilyroom:control
```
... always returns a 404, but no error is logged.

cdjackson:
The 404 is completely unrelated. It's caused by requesting config descriptions that don't exist. In one view, a 404 is correct - there is no description. Unfortunately the request I believe is correct in that the  channel is valid, so it's valid to make the request to find out if there are descriptions.

We had a similar issue in the past with #1268 which also used to return 404.

SiggiFR:
Hi @cdjackson , I can confirm to see same as Scott for all bindings I use (network, airquality, zwave, systeminfo, ntp). And the error is shown in red in logfile, which is confusing as it seems not really to be an error:
![grafik](https://user-images.githubusercontent.com/36704892/40918336-47d4b332-6806-11e8-9fbe-eb5af1c0fd6a.png)
So it would really be a help if this dummy error (?) will disappear :-)

maggu2810:
I assume there is only the produces annotation missing that declares that the media type is JSON.

```Java
    @Produces(MediaType.APPLICATION_JSON)
```

I will have a look at and if it is working create a PR.

maggu2810:
@openhab-5iver, @cdjackson, @SiggiFR 
Can you have a look at if #5683 fixes your issue?

SiggiFR:
Hi Markus,

 

I’m relatively new using OpenHAB. Can you give me a hint how I can test this change? Have I “only” to install the newest snapshot?

 

Thanks

Siggi

 

Von: Markus Rathgeb <notifications@github.com> 
[...]

PR #5683, opened 2018-06-05
use media type json for thing's config status REST response

maggu2810:
Fixes: https://github.com/eclipse/smarthome/issues/5674

Please check if this really fixes your issue...


maggu2810:
There are other lines in that file which use `MediaType.TEXT_PLAIN`.
I don't know why but I don't know if we "break" the some clients if we change them all.

kaikreuzer:
@maggu2810 No, all the others (which return no content) must stay as they are, see the result of our discussion at https://github.com/eclipse/smarthome/pull/5199#issuecomment-371762467.

maggu2810:
Ah, I remember again :wink:

Issue #5684, opened 2018-06-05
Bose Soundtouch Binding XML parsing errors


DerGute:
Hi! Have an Soundtouch 300 and i would like to switch the input. Ive read all the replies relating to this case but i did not find any solution. I installed the Binding from the GUI in 2.3.0.

Added Soundtouch 300 via things:
`bosesoundtouch:300:Soundtouch300 @ "Wohnzimmer"  [ host="192.168.2.88", macAddress="C4F3129FF042" ]
`
all the items for the thing:
`Switch  Soundtouch_Power                      "Power: [%s]"          <switch>      { channel="bosesoundtouch:300:Soundtouch300:power" }
Dimmer  Soundtouch_Volume                     "Volume: [%d %%]"      <volume>      { channel="bosesoundtouch:300:Soundtouch300:volume" }
Number  Soundtouch_Bass                       "Bass: [%d %%]"        <volume>      { channel="bosesoundtouch:300:Soundtouch300:bass" }
Switch  Soundtouch_Mute                       "Mute: [%s]"           <volume_mute> { channel="bosesoundtouch:300:Soundtouch300:mute" }
String  Soundtouch_OperationMode              "OperationMode: [%s]"  <text>        { channel="bosesoundtouch:300:Soundtouch300:operationMode" }
String  Soundtouch_PlayerControl              "Player Control: [%s]" <text>        { channel="bosesoundtouch:300:Soundtouch300:playerControl" }
String  Soundtouch_ZoneAdd                    "Zone add: [%s]"       <text>        { channel="bosesoundtouch:300:Soundtouch300:zoneAdd" }
String  Soundtouch_ZoneRemove                 "Zone remove: [%s]"    <text>        { channel="bosesoundtouch:300:Soundtouch300:zoneRemove" }
Number  Soundtouch_Preset                     "Preset: [%d]"         <text>        { channel="bosesoundtouch:300:Soundtouch300:preset" }
String  Soundtouch_PresetControl              "Preset Control: [%s]" <text>        { channel="bosesoundtouch:300:Soundtouch300:presetControl" }
Number  Soundtouch_SaveAsPreset               "Save as Preset: [%d]" <text>        { channel="bosesoundtouch:300:Soundtouch300:saveAsPreset" }
String  Soundtouch_KeyCode                    "Key Code: [%s]"       <text>        { channel="bosesoundtouch:300:Soundtouch300:keyCode" }
[...]

PR #5685, opened 2018-06-05
BoseSoundtouch: Fix NPE

triller-telekom:
Occurred in #5684


extensions/binding/org.eclipse.smarthome.binding.bosesoundtouch/src/main/java/org/eclipse/smarthome/binding/bosesoundtouch/internal/ContentItem.java
```diff
@@ -71,7 +71,7 @@ public boolean isValid() {
         if (getOperationMode() == OperationModeType.STANDBY) {
             return true;
         }
-        if (itemName.equals("") || source.equals("")) {
+        if (itemName == null || source == null || itemName.equals("") || source.equals("")) {
```
kaikreuzer: While you are at it, maybe you could also change the `.equals("")` to `.isEmpty()`?

triller-telekom: Good point. Done!

PR #5686, opened 2018-06-05
GroupItem: Prevent simple direct loop

### LINKED ISSUE #5635: [core] Group item can be added into itself
A Group item can be added into itself. Is this supposed to be possible?

demo.items
```
Group gHISTORY "Persistence" (gHISTORY)
```

console output
```
openhab> items list gHISTORY
gHISTORY (Type=GroupItem, Members=0, State=NULL, Label=Persistence, Category=null, Groups=[gHISTORY])
```

triller-telekom:
This PR fixes a direct loop, i.e. if one tries to add a GroupItem as a
member of itself.

Fixes #5635

### DIFF
--- a/bundles/core/org.eclipse.smarthome.core/src/main/java/org/eclipse/smarthome/core/items/GenericItem.java
+++ b/bundles/core/org.eclipse.smarthome.core/src/main/java/org/eclipse/smarthome/core/items/GenericItem.java
@@ -129,22 +129,42 @@ public abstract class GenericItem implements ActiveItem {
         if (groupItemName == null) {
             throw new IllegalArgumentException("Group item name must not be null!");
         }
+
+        if (checkForGroupLoop(groupItemName)) {
+            return;
+        }
+
         if (!groupNames.contains(groupItemName)) {
             groupNames.add(groupItemName);
         }
     }
 
+    private boolean checkForGroupLoop(String groupItemName) {
+        if (this instanceof GroupItem) {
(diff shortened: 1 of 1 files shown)


triller-telekom:
However, this does not fix the case where loops are nested over multiple items. For example:

Group1 (Group2)
Group2 (Group1)

The problem is that the `GenericItem#addGroupName(s)` methods take a `String` as argument and not an `Item` so it is not possible to ask which groups this one is part of.

The only way I see would be to add the the `ItemRegistry` as a dependency and fetch the item by name to query it for its group names, but I am not sure whether that is a good idea...

kaikreuzer:
> However, this does not fix the case where loops are nested over multiple items.

I thought to be pretty sure that @sjka already fixed that case a while ago and made sure that such cycles do not cause any harm.

sjka:
Yes, I can also remember I once fixed something related to such cycles, but for sure I did nothing to prevent them.
So to me the question is: Shall the framework really _prevent_ that? I thought it is actually dealing quite well with such cycles.

In any case, this PR is pretty in the sense that now `GenericItem` "knows" one of its subclasses (i.e. `GroupItem`).

triller-telekom:
Thinkign about it again: Breaking only one direction, i.e. GroupItem -> Member but not the reverse direction is buggy too.

So far we allow the user to create cycles and @sjka has fixed one occasion where the framework could not handle them. All other occasions are fine so far, so we can also agree on that if the user creates a cycle in its data model, it is his problem.

PR #5688, opened 2018-06-06
fixed a bug in metadata command line parsing for configuration arguments

afuechsel:

### DIFF
--- a/bundles/io/org.eclipse.smarthome.io.console/src/main/java/org/eclipse/smarthome/io/console/internal/extension/MetadataConsoleCommandExtension.java
+++ b/bundles/io/org.eclipse.smarthome.io.console/src/main/java/org/eclipse/smarthome/io/console/internal/extension/MetadataConsoleCommandExtension.java
@@ -16,6 +16,7 @@ import java.util.Arrays;
 import java.util.HashMap;
 import java.util.List;
 import java.util.Map;
+import java.util.stream.Collectors;
 
 import org.eclipse.jdt.annotation.NonNullByDefault;
 import org.eclipse.smarthome.core.items.ItemRegistry;
@@ -79,7 +80,9 @@ public class MetadataConsoleCommandExtension extends AbstractConsoleCommandExten
                     if (args.length < 4) {
                         printUsage(console);
                     } else {
-                        addMetadata(console, args[1], args[2], args[3], args.length > 4 ? args[4] : null);
+                        addMetadata(console, args[1], args[2], args[3],
+                                args.length > 4 ? Arrays.stream(args, 4, args.length).collect(Collectors.joining())
+                                        : null);
                     }
                     break;
                 case SUBCMD_REMOVE:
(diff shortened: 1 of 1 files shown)


bundles/io/org.eclipse.smarthome.io.console/src/main/java/org/eclipse/smarthome/io/console/internal/extension/MetadataConsoleCommandExtension.java
```diff
@@ -71,7 +72,9 @@ public void execute(String[] args, Console console) {
                     if (args.length < 4) {
                         printUsage(console);
                     } else {
-                        addMetadata(console, args[1], args[2], args[3], args.length > 4 ? args[4] : null);
+                        addMetadata(console, args[1], args[2], args[3],
+                                args.length > 4 ? Arrays.stream(args, 4, args.length).collect(Collectors.joining())
+                                        : null);
```
maggu2810: Just to get it right: The arguments 4-x are joined without any separator, correct?
So, if argument 4 is "1", argument 5 is "2" and argument 6 is "3", this will result into a string "123".

bundles/io/org.eclipse.smarthome.io.console/src/main/java/org/eclipse/smarthome/io/console/internal/extension/MetadataConsoleCommandExtension.java
```diff
@@ -116,11 +119,20 @@ private void addMetadata(Console console, String itemName, String namespace, Str
         if (config == null) {
             return null;
         }
+        String configStr = config;
+        if (configStr.startsWith("{")) {
```
maggu2810: Should we combine that to checks into one using "and"?

```Java
if (configStr.startsWith("{") && configStr.endsWith("}")) {
    configStr = configStr.substring(1, configStr.length() - 1);
}
```

I assume valid inputs are only start AND end with curly bracket or not using any.

afuechsel: Sure, makes sense.

bundles/io/org.eclipse.smarthome.io.console/src/main/java/org/eclipse/smarthome/io/console/internal/extension/MetadataConsoleCommandExtension.java
afuechsel: yes, because it parses lines like these: 
`{ key1 = value1, key2      =value2, key3=value3}` 
I even trimm the arguments left and right of the `=`. 

It does not make sense to have arguments like `{ key    1 = 1 2 3 }` - whoich would end up in `{key1=123}`. 

Surely the parsing could be made more error proof but I don't think this is needed here.

afuechsel: If that is the case we could also nearly drop this PR and live with the original implementation. Only the removal of the curly brackets would remain.

afuechsel:
Looks pretty small now, does it?

maggu2810:
It does.

PR #5689, opened 2018-06-06
MDNS discovery fix

### LINKED ISSUE #4644: mDNS discovery always considers all thing types supporting mDNS discovery
**Context**
The method `o.e.sh.config.discovery.DiscoveryServiceRegistry#startScan(thingTypeUID, scanListener)` starts a discovery based on a specific thing type. Running this method will not discover results for that thing type only. Instead, running this method starts a discovery using all discovery services that support that thing type (but possibly also many other thing types).

Example: If a thing can be discovered using mDNS, starting the discovery for that thing type will start a mDNS discovery, which will discover all things discoverable via mDNS.

**Issue**
The behavior described above matches the JavaDoc of the `startScan` method ('the Thing type UID pointing to collection of discovery services to be forced to start a discovery`). However, this behavior leads to rather lengthy discovery in the case of mDNS.

For mDNS discovery (cf. `o.e.sh.config.discovery.mdns.internal.MDNSDiscoveryService`) an mDNS scan is done sequentially for each `o.e.sh.config.discovery.mdns.DiscoveryParticipant`, each taking 6 seconds (the default timeout of the jmDNS library). If there are many bindings supporting mDNS discovery, each providing a `DiscoveryParticipant`, the scan takes rather long.

**Potential improvement**
I suggest to improve this, e.g., by allowing to start a scan for one (or a list of) thing type(s) that does not trigger scans for other thing types as a side effect, as is currently the case for the mDNS discovery (or, as another example, upnp discovery). For mDNS discovery, this would mean that the `MDNSDiscoveryService` could filter the set of `DiscoveryParticipant`s by thing type, only starting those `DiscoveryParticipant`s that support the requested thing type.

Is this a sensible improvement, or do I overlook something here? I would be happy about feedback.

**Side note**
For mDNS, the discovery timeout is currently 5 seconds, while the timeout for one jMDNS lookup is 6 seconds. I.e., even if there is only one `DiscoveryParticipant`, the discovery might take longer than the 5 seconds timeout. Should this be improved in the mDNS discovery as well, by aligning the timeout of the mDNS discovery to 6 seconds?

triller-telekom:
Fixes #4644

### DIFF
--- a/bundles/config/org.eclipse.smarthome.config.discovery.mdns/src/main/java/org/eclipse/smarthome/config/discovery/mdns/internal/MDNSDiscoveryService.java
+++ b/bundles/config/org.eclipse.smarthome.config.discovery.mdns/src/main/java/org/eclipse/smarthome/config/discovery/mdns/internal/MDNSDiscoveryService.java
@@ -135,7 +135,7 @@ public class MDNSDiscoveryService extends AbstractDiscoveryService implements Se
      * users may become impatient. The underlying reason is that the jmDNS
      * implementation {@code MDNSClient#list(String)} has a default timeout of 6
      * seconds when no ServiceInfo is found. When there are many participants,
-     * waiting 6 seconds for each non-existant type is too long.
+     * waiting 6 seconds for each non-existent type is too long.
      * 
      * @param isBackground
      *            true if it is background scan, false otherwise.
--- a/bundles/config/org.eclipse.smarthome.config.discovery.mdns/src/main/java/org/eclipse/smarthome/config/discovery/mdns/internal/MDNSDiscoveryService.java
+++ b/bundles/config/org.eclipse.smarthome.config.discovery.mdns/src/main/java/org/eclipse/smarthome/config/discovery/mdns/internal/MDNSDiscoveryService.java
@@ -150,7 +150,7 @@ public class MDNSDiscoveryService extends AbstractDiscoveryService implements Se
                 services = mdnsClient.list(participant.getServiceType(), FOREGROUND_SCAN_TIMEOUT);
             }
             logger.debug("{} services found for {}; duration: {}ms", services.length, participant.getServiceType(),
-                    (System.currentTimeMillis() - start));
+                    System.currentTimeMillis() - start);
             for (ServiceInfo service : services) {
                 DiscoveryResult result = participant.createResult(service);
                 if (result != null) {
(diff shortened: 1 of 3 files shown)


triller-telekom:
Replaces PR #5662 because the history was messed up

garytse:
Thanks Stefan!

bundles/config/org.eclipse.smarthome.config.discovery.mdns/src/main/java/org/eclipse/smarthome/config/discovery/mdns/internal/MDNSDiscoveryService.java
```diff
@@ -114,18 +117,40 @@ protected void stopBackgroundDiscovery() {
...
-            logger.debug("{} services found for {}", services.length, participant.getServiceType());
+            long start = System.currentTimeMillis();
+            ServiceInfo[] services;
+            if (isBackground) {
+                services = mdnsClient.list(participant.getServiceType());
+            } else {
+                services = mdnsClient.list(participant.getServiceType(), FOREGROUND_SCAN_TIMEOUT);
+            }
+            logger.debug("{} services found for {}; duration: {}ms", services.length, participant.getServiceType(),
+                    (System.currentTimeMillis() - start));
```
maggu2810: What's the reason for the braces around `System.currentTimeMillis() - start`? Aren't they rather useless?

bundles/config/org.eclipse.smarthome.config.discovery.mdns/src/main/java/org/eclipse/smarthome/config/discovery/mdns/internal/MDNSDiscoveryService.java
```diff
@@ -114,18 +117,40 @@ protected void stopBackgroundDiscovery() {
...
-    private void scan() {
+    /**
+     * Scan has 2 different behaviours. background/ foreground. Background scans can
+     * have much higher timeout. Foreground scans have only a short timeout as human
+     * users may become impatient. The underlying reason is that the jmDNS
+     * implementation {@code MDNSClient#list(String)} has a default timeout of 6
+     * seconds when no ServiceInfo is found. When there are many participants,
+     * waiting 6 seconds for each non-existent type is too long.
+     * 
+     * @param isBackground
```
maggu2810: Can you remove the line break between the parameter name and the description?

triller-telekom:
@garytse I took the freedom to remove the line-break, I guess that's alright :)

PR #5696, opened 2018-06-07
correctly support Dimmer and Color items in toggle profile

kaikreuzer:
The profile so far didn't register itself for Dimmer and Color items.
Furthermore, it didn't correctly map states to ON/OFF values and the tests were wrong, so they didn't notice it.
This PR registers the profile for Dimmer and Color, fixes the tests and adds some further tests.


sjka:
Oh boy, so many flaws... And so few lines to fix them 😄 
Thanks, looks good!

PR #5698, opened 2018-06-07
JSONStorage: Do not use invalid fileNames for some file-systems

### LINKED ISSUE #5687: JSON Storage must not use colons in file name
The files produced by the JSON storage have illegal names for Windows, see https://community.openhab.org/t/openhab-storage-api-generate-filenames-with-colons-which-are-invalid-on-windows/46028:

```
07:18:50.454 [ERROR] [ome.storage.json.internal.JsonStorage] - Error writing JsonDB to D:\OpenHABRoot\OpenHAB2\userdata\jsondb\amazonechocontrol:account:account1.json. Cause D:\OpenHABRoot\OpenHAB2\userdata\jsondb\amazonechocontrol:account:account1.json (Die Syntax für den Dateinamen, Verzeichnisnamen oder die Datenträgerbezeichnung ist falsch).
07:18:50.458 [ERROR] [ome.storage.json.internal.JsonStorage] - Error writing JsonDB to D:\OpenHABRoot\OpenHAB2\userdata\jsondb\backup\1528175930458–amazonechocontrol:account:account1.json. Cause D:\OpenHABRoot\OpenHAB2\userdata\jsondb\backup\1528175930458–amazonechocontrol:account:account1.json (Die Syntax für den Dateinamen, Verzeichnisnamen oder die Datenträgerbezeichnung ist falsch).
```

triller-telekom:
Windows has a rather long list of forbidden characters in filenames, see

https://msdn.microsoft.com/en-us/library/aa365247(VS.85).aspx

Linux-like systems only disallow / and the null-byte.

https://en.wikipedia.org/wiki/Filename#Reserved_characters_and_words

By URLencoding the string, all of these characters are safely transformed
into %<HexValue> and % is an allowed character. This encoding is even
reversible.

Given that we will not run on RT-11 systems we should be safe.

I use a maximum length of 127 to be safe even on android systems:
https://stackoverflow.com/questions/13204807/max-file-name-length-in-android

The down-side is that even spaces are now converted into %20, but I chose
to use URLencoder.encode instead of implementing the conversion by myself.

For existing storages I check whether a file exists and keep using it.

Fixes #5687

### DIFF
--- a/bundles/storage/org.eclipse.smarthome.storage.json/src/main/java/org/eclipse/smarthome/storage/json/internal/JsonStorageService.java
+++ b/bundles/storage/org.eclipse.smarthome.storage.json/src/main/java/org/eclipse/smarthome/storage/json/internal/JsonStorageService.java
@@ -13,6 +13,8 @@
 package org.eclipse.smarthome.storage.json.internal;
 
 import java.io.File;
+import java.io.UnsupportedEncodingException;
+import java.net.URLEncoder;
 import java.util.HashMap;
 import java.util.Map;
 
@@ -30,6 +32,8 @@ import org.slf4j.LoggerFactory;
  * @author Chris Jackson - Initial Contribution
  */
 public class JsonStorageService implements StorageService {
+
+    private static final int MAX_FILENAME_LENGTH = 127;
 
     private final Logger logger = LoggerFactory.getLogger(JsonStorageService.class);
 
@@ -98,8 +102,13 @@ public class JsonStorageService implements StorageService {
 
     @Override
     public <T> Storage<T> getStorage(String name, ClassLoader classLoader) {
-        File file = new File(dbFolderName, name + ".json");
+        File legacyFile = new File(dbFolderName, name + ".json");
+        File escapedFile = new File(dbFolderName, urlEscapeUnwantedChars(name) + ".json");
 
+        File file = escapedFile;
+        if (legacyFile.exists()) {
+            file = legacyFile;
+        }
         if (!storageList.containsKey(name)) {
             storageList.put(name, (JsonStorage<Object>) new JsonStorage<T>(file, classLoader, maxBackupFiles,
                     writeDelay, maxDeferredPeriod));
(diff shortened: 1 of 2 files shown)


bundles/storage/org.eclipse.smarthome.storage.json/src/main/java/org/eclipse/smarthome/storage/json/internal/JsonStorageService.java
```diff
@@ -98,8 +102,13 @@ protected void deactivate() {
...
     @Override
     public <T> Storage<T> getStorage(String name, ClassLoader classLoader) {
-        File file = new File(dbFolderName, name + ".json");
+        File legacyFile = new File(dbFolderName, name + ".json");
+        File escapedFile = new File(dbFolderName, urlEscapeUnwantedChars(name) + ".json");
 
+        File file = escapedFile;
+        if (legacyFile.exists()) {
+            file = legacyFile;
+        }
```
maggu2810: Instead of create two local file objects that are available in that scope you could outsource the logic to find the correct file object...

```Java
private File getStorageFile(final String name) {
    final File legacyFile = new File(dbFolderName, name + ".json");
    if (legacyFile.exists()) {
        return legacyFile;
    } else {
        return new File(dbFolderName, urlEscapeUnwantedChars(name) + ".json");
    }
}
```

maggu2810: Hm, thinking about it again, it does perhaps not make sense to create a function if it is used only on that place.
No need to change it...

sjka:
I know, `%`-characters technically _are_ allowed on windows related file systems. Nevertheless I think to remember they were causing some trouble using command line tools. Or am I mistaken?

sjka:
Ah, it's coming back... The `%`-characters are variable-delimiters (i.e. what `$` is to an operating system). Therefore it's probably okay if they are enclosed by `"` in command line tools. 
Nevermind then my previous comment.

PR #5700, opened 2018-06-07
Introduce an ItemBuilder

sjka:
In continuation of #5449, this PR now aims at adding an `ItemBuilder`. 

In order to facilitate for the dynamic plugging of `ItemFactories`, the `ItemBuilder` has to be somewhat "OSGi-aware". For that purpose, an instance can be obtained from the `ItemRegistry` - under the assumption that the using component will have to interact with the registry anyway. Only the GenericItemProvider suffers from that, as it cannot depend on it. Alternatively we could it make its own service - I'd leave this open for discussion.

At the same time, the API was cleaned of `GenericItem`s, exposing `Item`s instead. Only very few, core places need a checked cast to `GenericItem` at the moment. While these changes are clearly API-breaking, I think they will be acceptable: 
* The change in `ItemRegistry` should be  even binary compatible, thanks to type erasure
* The change in `ItemFactory` is only source-compatible, but _NOT_ binary-compatible. Nevertheless, I wouldn't reckon there are solutions which are binary-consuming extensions which bring their own item factory nor consume item factories directly. Please object if you see any issue with this!

relates to #5449

### DIFF
--- a/bundles/core/org.eclipse.smarthome.core.test/src/test/java/org/eclipse/smarthome/core/internal/items/ItemBuilderTest.java
+++ b/bundles/core/org.eclipse.smarthome.core.test/src/test/java/org/eclipse/smarthome/core/internal/items/ItemBuilderTest.java
@@ -1,0 +1,176 @@
+/**
+ * Copyright (c) 2014,2018 Contributors to the Eclipse Foundation
+ *
+ * See the NOTICE file(s) distributed with this work for additional
+ * information regarding copyright ownership.
+ *
+ * This program and the accompanying materials are made available under the
+ * terms of the Eclipse Public License 2.0 which is available at
+ * http://www.eclipse.org/legal/epl-2.0
+ *
+ * SPDX-License-Identifier: EPL-2.0
+ */
+package org.eclipse.smarthome.core.internal.items;
+
+import static org.junit.Assert.*;
+import static org.mockito.ArgumentMatchers.*;
+import static org.mockito.Mockito.*;
+import static org.mockito.MockitoAnnotations.initMocks;
+
+import java.util.Arrays;
+import java.util.Collections;
+
+import org.eclipse.smarthome.core.items.ActiveItem;
+import org.eclipse.smarthome.core.items.GroupFunction;
+import org.eclipse.smarthome.core.items.GroupItem;
+import org.eclipse.smarthome.core.items.Item;
+import org.eclipse.smarthome.core.items.ItemFactory;
+import org.junit.Before;
+import org.junit.Test;
+import org.mockito.Mock;
+
+/**
+ *
+ * @author Simon Kaufmann - initial contribution and API
+ *
+ */
+public class ItemBuilderTest {
+
+    private ItemRegistryImpl itemRegistry;
+    private @Mock ItemFactory mockFactory;
+    private @Mock ActiveItem mockItem;
+    private @Mock Item originalItem;
+
+    @Before
+    public void setup() {
+        initMocks(this);
+        itemRegistry = new ItemRegistryImpl();
+        itemRegistry.addItemFactory(mockFactory);
+    }
+
+    @Test
+    public void testMinimal() {
+        when(mockFactory.createItem(anyString(), anyString())).thenReturn(mockItem);
+
+        Item res = itemRegistry.newItemBuilder("String", "test").build();
+
+        assertSame(mockItem, res);
+        verify(mockFactory).createItem(eq("String"), eq("test"));
+        verify(mockItem).setLabel(isNull());
+        verify(mockItem).setCategory(isNull());
+        verify(mockItem).addGroupNames(eq(Collections.emptyList()));
+    }
+
+    @Test
+    public void testMinimalGroupItem() {
+        Item resItem = itemRegistry.newItemBuilder("Group", "test").build();
+
+        assertEquals(GroupItem.class, resItem.getClass());
+        GroupItem res = (GroupItem) resItem;
+        verifyNoMoreInteractions(mockFactory);
+        assertNull(res.getCategory());
+        assertEquals(Collections.emptyList(), res.getGroupNames());
+        assertNull(res.getLabel());
+        assertNull(res.getFunction());
+        assertNull(res.getBaseItem());
+    }
+
+    @Test
+    public void testFull() {
+        when(mockFactory.createItem(anyString(), anyString())).thenReturn(mockItem);
+
+        Item res = itemRegistry.newItemBuilder("String", "test") //
+                .withCategory("category") //
+                .withGroups("a", "b") //
+                .withLabel("label") //
+                .build();
+
+        assertSame(mockItem, res);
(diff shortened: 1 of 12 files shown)


openhab-5iver:
Does this change affect the DSL too? Specifically, will GenericItem not be able to be used (:-1: but possibly curtailed by the following...), and will an item be able to be construct by String name (:+1: :+1:)?

bundles/core/org.eclipse.smarthome.core.test/src/test/java/org/eclipse/smarthome/core/internal/items/ItemBuilderTest.java
```diff
@@ -0,0 +1,176 @@
...
+        verify(mockItem).addGroupNames(eq(Arrays.asList("a", "b")));
+        verify(mockItem).setLabel(eq("label"));
+    }
+
+    @Test
+    public void testFullGroupItem() {
+        Item baseItem = mock(Item.class);
+        GroupFunction mockFunction = mock(GroupFunction.class);
+
+        Item resItem = itemRegistry.newItemBuilder("Group", "test") //
```
triller-telekom: You could use `GroupItem.TYPE` instead of "Group" since tests are usually a good example that people copy.

bundles/core/org.eclipse.smarthome.core/src/main/java/org/eclipse/smarthome/core/internal/items/ItemBuilderImpl.java
```diff
@@ -0,0 +1,154 @@
...
+    }
+
+    @Override
+    public Item build() {
+        Item item = null;
+        if (GroupItem.TYPE.equals(itemType)) {
+            item = new GroupItem(name, baseItem, groupFunction);
+        } else {
+            for (ItemFactory itemFactory : itemFactories) {
+                item = itemFactory.createItem(itemType, name);
```
triller-telekom: So the last one wins? This is  a pretty non-deterministic behaviour, isn't it?

bundles/core/org.eclipse.smarthome.core.test/src/test/java/org/eclipse/smarthome/core/internal/items/ItemBuilderTest.java
```diff
@@ -0,0 +1,176 @@
...
+
+import static org.junit.Assert.*;
+import static org.mockito.ArgumentMatchers.*;
+import static org.mockito.Mockito.*;
+import static org.mockito.MockitoAnnotations.initMocks;
+
+import java.util.Arrays;
+import java.util.Collections;
+
+import org.eclipse.smarthome.core.items.ActiveItem;
```
triller-telekom: Since `ActiveItem` is not public API and you are introducing the builder pattern to create items, shall we move this class to an `internal` package and remove the `@Deprecated` annotation on it?

bundles/core/org.eclipse.smarthome.core/src/main/java/org/eclipse/smarthome/core/items/ItemBuilder.java
```diff
@@ -0,0 +1,93 @@
...
+ *
+ */
+@NonNullByDefault
+public interface ItemBuilder {
+
+    /**
+     * Creates an item with the currently configured values.
+     *
+     * @return an item
+     * @throws IllegalStateException in case no item factory can create the given item type
```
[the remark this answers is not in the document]
maggu2810: I would say that a failing build is an exception and I assume here a unchecked exception is okay as there is nothing you can do.
I would assume that you should mostly never run into that situation.

bundles/core/org.eclipse.smarthome.core/src/main/java/org/eclipse/smarthome/core/items/ItemBuilder.java
```diff
@@ -0,0 +1,93 @@
...
+     */
+    ItemBuilder withGroups(@Nullable Collection<String> groups);
+
+    /**
+     * Set the group membership of the item.
+     *
+     * @param groups the group names the item belongs to
+     * @return the builder itself
+     */
+    ItemBuilder withGroups(String... groups);
```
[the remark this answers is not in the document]
maggu2810: I agree. I would prefer to have the method consuming a collection only. If someone really needs an array, `Arrays.asList` should be its friend.

bundles/core/org.eclipse.smarthome.core/src/main/java/org/eclipse/smarthome/core/items/ItemBuilder.java
```diff
@@ -0,0 +1,93 @@
...
+     * Set the group function
+     *
+     * @param function the group function
+     * @return the builder itself
+     * @throws IllegalArgumentException in case this is not a group item
+     */
+    ItemBuilder withGroupFunction(@Nullable GroupFunction function);
+
+    @Deprecated
+    ItemBuilder withTags(String... tags);
```
maggu2810: Same as above (Collection should be enough). But let's wait for the answer about the introduction of deprecated methods.

bundles/core/org.eclipse.smarthome.core/src/main/java/org/eclipse/smarthome/core/internal/items/ItemBuilderImpl.java
maggu2810: I assume if item is non null we would like to break the loop, so the first factory that can handle the item type wins.

bundles/core/org.eclipse.smarthome.core.test/src/test/java/org/eclipse/smarthome/core/internal/items/ItemBuilderTest.java
sjka: No. I want this test to break if somebody ever had the stupid idea to change the constant's value!

bundles/core/org.eclipse.smarthome.core/src/main/java/org/eclipse/smarthome/core/internal/items/ItemBuilderImpl.java
sjka: yes, of course 😩

bundles/core/org.eclipse.smarthome.core.test/src/test/java/org/eclipse/smarthome/core/internal/items/ItemBuilderTest.java
sjka: No!
It currently _is_ public API. And moving it into an internal package will not only be "API"-breaking, but will also render all concrete item implementations non-usable, as this interface defines the public methods on them. 
The current Item-API design is flawed. There is nothing we can do about that, apart from rewriting major parts of it. But that'll come at a high price on consumers' side.

kaikreuzer:
@maggu2810 I leave the merge to you, if you find all your comments answered/addressed.

PR #5703, opened 2018-06-08
add stream support for storages

maggu2810:


sjka:
Nice addition!
Wouldn't it makes sense to also have `JsonStorage` and `MapDbStorage` override it? They can do much better than the default implementation...

maggu2810:
I started with, but as every value of a key needs to be deserialized. I assume the deserialization is process that eats the most time and so I left it as it is.
WDYT?

sjka:
Right - I just remembered they both keep a "map" inside. Having had a closer look, I totally agree.

PR #5711, opened 2018-06-09
Sonos: update alarm properties only when alarm state changes

lolodomo:


sjka:
Thanks!

PR #5714, opened 2018-06-10
[Basic UI] Consider item options for switch widget

lolodomo:
Render a switch widget with buttons when the linked item has a maximum of 4 state options

### DIFF
--- a/extensions/ui/org.eclipse.smarthome.ui.basic/src/main/java/org/eclipse/smarthome/ui/basic/internal/render/SwitchRenderer.java
+++ b/extensions/ui/org.eclipse.smarthome.ui.basic/src/main/java/org/eclipse/smarthome/ui/basic/internal/render/SwitchRenderer.java
@@ -22,6 +22,7 @@ import org.eclipse.smarthome.core.library.items.RollershutterItem;
 import org.eclipse.smarthome.core.library.types.OnOffType;
 import org.eclipse.smarthome.core.library.types.QuantityType;
 import org.eclipse.smarthome.core.types.State;
+import org.eclipse.smarthome.core.types.StateOption;
 import org.eclipse.smarthome.core.types.util.UnitUtils;
 import org.eclipse.smarthome.model.sitemap.Mapping;
 import org.eclipse.smarthome.model.sitemap.Switch;
@@ -50,6 +51,9 @@ public class SwitchRenderer extends AbstractWidgetRenderer {
 
     private final Logger logger = LoggerFactory.getLogger(SwitchRenderer.class);
 
+    private static final int MAX_BUTTONS = 4;
+    private static final int MAX_LABEL_SIZE = 9;
+
     @Override
     @Activate
     protected void activate(BundleContext bundleContext) {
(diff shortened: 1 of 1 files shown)


lolodomo:
I proposed this PR because I implemented it but to be honest I am myself not fully convinced by the result. The reason is that the layout of Basic UI is limiting the number of buttons because they are rendered on one unique line.
To be really interesting, I think we should change the layout to allow the display of buttons on several lines. As a consequence it will allow displaying more buttons.

WDYT ?

PS: note that this limit is already applicable to switch with mappings.

lolodomo:
@resetnow : take a look to the fix I have done in smarhome.js. I think I fixed a bug. It was required for the new feature I implemented. The example in the demo sitemap with switch widget and mappings does not trigger the bug, I have to understand why and find another example that trigger the bug.
To be clear, the bug is that when Basic UI receives an update (through SSE), all buttons are then marked as unselected rather than marking the right one.

lolodomo:
Please note that I created a seperate issue and proposed the fix for the issue I was explaining just before.
If this PR is never merged, at least the bug will be fixed.

kaikreuzer:
I'm in general ok with that idea, but maybe we should only go for rendering buttons if there are <=4 options available? Taking the first 4 out of a longer list seems arbitrary and imho the widget would be equally broken as if no buttons are rendered automatically, but with the difference that the user might not actually notice it. Wdyt?

lolodomo:
I am ok with that.
Do you think we should apply this limit to switch with mappings too ?

kaikreuzer:
No, if the user specified x mapping, the UI should try to render it - maybe the user has an incredibly wide screen ;-)

lolodomo:
PR rebased + rendering with buttons only when max 4 buttons have to be displayed.

lolodomo:
Once this PR is mereged, I will enhance Classic UI the same way.

extensions/ui/org.eclipse.smarthome.ui.basic/src/main/java/org/eclipse/smarthome/ui/basic/internal/render/SwitchRenderer.java
```diff
@@ -147,6 +137,44 @@ public boolean canRender(Widget w) {
...
+        String label = lab;
+
+        if (item instanceof NumberItem && ((NumberItem) item).getDimension() != null) {
+            String unit = getUnitForWidget(w);
+            command = StringUtils.replace(command, UnitUtils.UNIT_PLACEHOLDER, unit);
+            label = StringUtils.replace(label, UnitUtils.UNIT_PLACEHOLDER, unit);
+        }
+
+        if (label != null && maxLabelSize >= 3 && label.length() > maxLabelSize) {
+            label = label.substring(0, maxLabelSize - 3) + "...";
```
kaikreuzer: I wonder if we could save some more space by replacing the three dots by a single unicode character, see https://stackoverflow.com/questions/19997912/how-should-i-write-three-dots. Could you give that a try?

lolodomo: Good idea. I will check that when implementing the feature in Classic UI.

PR #5721, opened 2018-06-11
Serialize degree celsius with °C, drop special char

### LINKED ISSUE #5669: RfC: Change QuantityType serialisation for degree celsius
I would like to suggest changing the string serialisation for degree celsius from the special character `℃` to `°C` (two separate characters).

When working with such units, I noticed that:
- many fonts either do not support `℃` at all or render it in some weird fashion
- Many Text-to-speech engines cannot cope with `℃`, while they correctly process `°C`.
- as `℃` is not on the keyboard, people won't use it themselves (in rules, sitemaps, etc.) and thus we will always have a mix which might result in issues (e.g. when trying to do string comparison).

TL;DR: I think our life would be easier if we avoid using the very uncommon char `℃`.

htreu:
Fixes #5669.

Note: During parsing the special character `℃` is still supported.

### DIFF
--- a/bundles/core/org.eclipse.smarthome.core.test/src/test/java/org/eclipse/smarthome/core/library/types/QuantityTypeTest.java
+++ b/bundles/core/org.eclipse.smarthome.core.test/src/test/java/org/eclipse/smarthome/core/library/types/QuantityTypeTest.java
@@ -38,6 +38,7 @@ import tec.uom.se.quantity.QuantityDimension;
 /**
  * @author Gaël L'hopital - initial contribution
  */
+@SuppressWarnings({ "rawtypes", "unchecked", "null" })
 public class QuantityTypeTest {
 
     @Before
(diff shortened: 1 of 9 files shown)


sjka:
```
Failed tests: 
  QuantityTypeTest.toFullStringShouldParseToEqualState:193 
Expected: is "20 ℃"
     but: was "20 °C"
```

kaikreuzer:
Sounds as if the failing test confirms the correctness of the change 😎 .

htreu:
Test fixed, thanks guys.

sjka:
Computer says: No. 🤣 

```
Failed tests: 
  ScriptEngineOSGiTest.testAssignQuantityType:117 expected:<20.0 [℃]> but was:<20.0 [°C]>
```

sjka:
Ahhh, there we go. Thanks!

PR #5727, opened 2018-06-12
Documentation: Explain SystemFollowProfile

triller-telekom:

### DIFF
--- a/docs/documentation/development/bindings/profiles.md
+++ b/docs/documentation/development/bindings/profiles.md
@@ -90,3 +90,13 @@ Two methods have to be implemented to achieve this:
 ```java
 <item-type> MyItem { channel="<bindingID>:<thing-typeID>:MyThing:myChannel"[profile="MyScope:MyProfile"]}
 ```
+
+## Existing Profiles
+
+### FollowProfile
+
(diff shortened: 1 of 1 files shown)


docs/documentation/development/bindings/profiles.md
```diff
@@ -89,4 +89,14 @@ Two methods have to be implemented to achieve this:
...
+```
+
+## Existing Profiles
+
+### FollowProfile
+
+If one device should "follow" the actions of another device, the FollowProfile can be used. The term "follow" in this case means that any state that is sent to an `Item` will be forwarded from this `Item` to any linked channel with the FollowProfile. The FollowProfile takes state updates on an `Item` and sends them as a command onto the channel. In the direction from the ThingHandler towards the `Item`, the FollowProfile ignores state updates.
+
+```java
+<itemType> <itemName> { channel="<bindingID>:<thing-typeID>:<thingName>:<channelName>", channel="<bindingID>:<thing-typeID>:<thingName>:<followChannelName>"[profile="follow"]"}
```
kaikreuzer: Just a small comment: This implies that the thing type ID can be derived from channel UIDs, which is not correct.
I'd suggest to rather simply write `channel="<channelUID>"` instead.

kaikreuzer: Also, there is no `thingName`, so another reason to change this :-)

PR #5730, opened 2018-06-13
added Dual IR 3a to supported Devices

aquast:


aquast:
I'm sorry, but due to the privacy policy and some other issues of ECA which are not conform to the GDPR, I don't want to accept ECA. So if You can't cope with that please reject the PR. ;-)

kaikreuzer:
Thanks for this info - to avoid any issues, this PR is closed and the information is covered through #5733.

